PR #58, opened 2018-04-30
[modal] Add close button to modal layouts

matchu:
In this change, we create a `ModalCloseButton` component, and add it to the modal layout components.

This close button doesn't control modal launcher state yet; it just triggers the public callback `onClickCloseButton` on the modal layout component. (We'll probably need to set up some wires via context, in order to control the modal launcher. Coming soon!)

Some modals will not want to provide default close behavior at all, but that's not part of the scope of this change. (See WB-134.)

Test Plan:
Open Styleguidist, and take a look at those cute close buttons!
StandardModal screenshot
TwoColumnModal screenshot

Click them, and see an alert, because the Styleguidist examples have an alert handler hooked up to `onClickCloseButton`.
Alert screenshot

Additionally, use a ruler to confirm that the title and subtitle are indeed centered in the title bar, despite the addition of the close button.

Finally, use the resize handle to confirm that, when the viewport becomes small, the `StandardModal` title and subtitle text still wrap naturally.
Small StandardModal screenshot

PR #59, opened 2018-04-30
[modal] Automatically wire close buttons up to ModalLauncher state

matchu:
In #58, we added close buttons to the modal layout components.

In this change, we wire them up to `ModalLauncher`'s state automatically.

To do this, we use `React.cloneElement` to add an `onClickCloseButton` prop to the provided `modal`. Our two modal layout components (`StandardModal` and `TwoColumnModal`) both support this prop, so this is the only change necessary to integrate the two layers.

An alternative would have been to require the user to perform this wiring themselves, via function-as-children. This would've been super annoying!

Another alternative would've been to use context. This would've been a bit more complex internally, but grant the API a bit more freedom to wrap the `modal` element in other ways. However, our API doesn't _want_ users to wrap the modal in arbitrary ways: we want them to give us a modal layout component instance, and that's it. So, this extra flexibility wouldn't actually be a win right now - but, if we end up needing that flexibility later, maybe this would be a good direction! (Incidentally, I tried to use Flow types to enforce this constraint, but had trouble and decided to punt.)

Along the way, I narrowed our API to accept a `React.Element` rather than a `React.Node`, in order to simplify our usage of `cloneElement`. This also enabled us to remove some complexity in `ModalLauncherPortal`, which used a wrapper element that's no longer necessary.

Test Plan:
```npm test```

In Storybook, use the `ModalLauncher` example to launch a standard modal and a two-column modal. Confirm that both of their close buttons successfully close the launched modal.

### CODE REVIEW

**Comment by jeresig** [comment]

Built with commit f38ad60e48d57ff870a594cc6e8123be6ea1ef27

https://deploy-preview-59--wonder-blocks.netlify.com

File: packages/wonder-blocks-modal/components/modal-launcher.test.js [hunk 1]
```diff
@@ -59,4 +64,80 @@ describe("ModalLauncher", () => {
         wrapper.find("button").simulate("click");
         expect(wrapper.find("ModalLauncherPortal")).toHaveLength(1);
     });
+
+    test("Adds an `onClickCloseButton` prop to the provided `modal`", () => {
```
kevinbarabash: These tests to have a lot of implementation details in them.  I think we should skew towards more black box testing in the sense that we shouldn't know the implementation details of the components only what their behavior should be.  In the case of the `onClickCloseButton`, what I'd actually like to test is that clicking the `close` button on a modal will dismiss the modal.  This will require us to actually mount the components, but I think the tests will be more straight forward both to read and write and they'll be more robust to changes in the implementation details over time.

packages/wonder-blocks-modal/components/modal-launcher.js
```diff
@@ -6,6 +6,11 @@ import {View} from "wonder-blocks-core";
 import ModalLauncherPortal from "./modal-launcher-portal.js";
 import ModalBackdrop from "./modal-backdrop.js";
 
+// TODO(mdr): This should only match modal elements: StandardModal,
+//     TwoColumnModal, etc. But I had trouble getting the types to work for this
+//     union, especially when combined with `cloneElement`, so I'm punting.
+type ModalElement = React.Element<any>;
```
kevinbarabash: It's possible that developers could write another component to return one of these three components as well and that should would.

packages/wonder-blocks-modal/components/modal-launcher-portal.js
```diff
@@ -31,17 +31,8 @@ export default class ModalLauncherPortal extends React.Component<Props> {
...
-        // accept `React.Node` instead of `React.Element`.
-        //
-        // NOTE(mdr): Once we upgrade to React 16, we'll no longer need this
-        //     wrapper node!
-        const wrapper = (
-            <View data-modal-launcher-portal>{this.props.children}</View>
-        );
-
         // Render the modal into the destination node.
-        ReactDOM.render(wrapper, destination);
```
kevinbarabash: Was the wrapper not really necessary in the first place?

packages/wonder-blocks-modal/components/modal-launcher.js
```diff
@@ -76,13 +81,23 @@ export default class ModalLauncher extends React.Component<Props, State> {
...
+            modal = this.props.modal({
                 closeModal: this._closeModal,
             });
         } else {
-            return this.props.modal;
+            modal = this.props.modal;
         }
+
+        return React.cloneElement(modal, {
+            onClickCloseButton: () => {
```
jeresig: Same question as in the other PR - shouldn't it be `onClose`?

packages/wonder-blocks-modal/components/modal-launcher.test.js
jeresig: How much of a deal-breaker is that? IMO I think these are fine for the moment, albeit more brittle. Maybe we can go back later and try and improve them?

matchu: Yeah, I'm not sure where to do the trade-off here. The shallow approach is fragile in the sense that it will need to be updated if we refactor the relationship between `ModalLauncher` and `ModalLauncherPortal`. But the mount-based approach is fragile in the sense that almost any breakage in `ModalLauncherPortal` would also cause this test to break.

Both are prone to failure, but I think the latter case would be surprising and confusing if it broke, whereas I think the former case would be rather unsurprising if it broke, and I don't expect the former breakage to actually happen more than once or twice in our component's lifetime. But if there's a third version of the test that would be chill in both cases, I'd strongly prefer that!

kevinbarabash: Point taken about unrelated components causes breakages here.  I do think we should have some more full mount tests that verify basic things without worry about implementation details.  Since we already have unit tests we'll know which part of the code the breakage is in.  The benefit of the full mount tests it's easier to see what the behavior should when the component is actually in use.  They replace a certain level of manual testing.  I no longer have to click on the close button to verify that it does indeed close the modal.  It also provides a level of documentation about how all of the subcomponents are supposed to function together.  We can leave this for now, but we should definitely go back and add some of these tests before calling the modals done.

matchu: In hindsight, I suppose this exact same class of problem is what led me to create `ModalBackdrop` and test it separately. Maybe I should move `cloneElement` into there? I'll try that.

packages/wonder-blocks-modal/components/modal-launcher-portal.js
matchu: It was necessary when `modal` was defined as accepting `React.Node`, because `ReactDOM.render` expects a `React.Element` instead of a `React.Node`.

Now that `modal` accepts a `React.Element`, we no longer need a wrapper element in order to be able to pass a single element here; we can just pass the `modal` itself.

matchu:
@kevinbarabash @jeresig Can y'all give this a re-review? I moved the cloning behavior into `ModalBackdrop`, which simplifies this change and its tests a lot. What do you think?

packages/wonder-blocks-modal/components/modal-backdrop.js
```diff
@@ -26,14 +32,22 @@ export default class ModalBackdrop extends React.Component<Props> {
...
-            this.props.onClickBackdrop();
+            this.props.onCloseModal();
         }
     };
 
     render() {
+        const children = this.props.children;
+        const clonedChildren = React.cloneElement(children, {
+            onClickCloseButton: () => {
+                children.props.onClickCloseButton();
```
jeresig: Do you have to check and see if this exists already, or not? Or maybe since it's required in the Flow props we're ok?

matchu: Hmm good point! This worked out in testing because modal elements have a `onClickCloseButton` defined in `defaultProps`, but this won't work for wrapper components. I've added a test for it!

packages/wonder-blocks-modal/util/types.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+import * as React from "react";
+
+/**
+ * A `ModalElement` is a React element that should either itself be a modal
+ * (StandardModal, OneColumnModal, TwoColumnModal), or wrap a modal.
```
kevinbarabash: wrap => wrapper

kevinbarabash:
@matchu feel free to merge away.

PR #60, opened 2018-04-30
[modal] Pressing Escape closes a launched modal

matchu:
In this change, we add keyboard listeners to `ModalLauncher` to close the modal when Escape is pressed.

Test Plan:
```npm test```

In Styleguidist, use the `ModalLauncher` example to launch a standard modal and a two-column modal. For each, confirm that pressing Escape closes the modal.

### DIFF
--- a/packages/wonder-blocks-modal/components/modal-launcher.test.js
+++ b/packages/wonder-blocks-modal/components/modal-launcher.test.js
@@ -1,6 +1,6 @@
 // @flow
 import React from "react";
-import {shallow} from "enzyme";
+import {shallow, mount} from "enzyme";
 
 import ModalLauncher from "./modal-launcher.js";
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jeresig** [comment]

Built with commit 92f9e4fbf7573e1c25710f8e495f5fb03c5b1f9c

https://deploy-preview-60--wonder-blocks.netlify.com

File: packages/wonder-blocks-modal/components/modal-launcher.test.js [hunk 1]
```diff
@@ -59,4 +59,34 @@ describe("ModalLauncher", () => {
...
         expect(wrapper.find("ModalLauncherPortal")).toHaveLength(1);
     });
+
+    test("Pressing Escape closes the modal", (done) => {
+        let pressedEscape = false;
+
+        // On close, assert that this function was triggered _after_ Escape was
+        // pressed, and finish the test.
+        const onClose = () => {
+            expect(pressedEscape).toBe(true);
```
kevinbarabash: Verify this isn't all that important.  I'd rather verify that the modal is gone which we can do since we `mount`ed the `ModalLauncher` for real.

matchu: Yeah, I punted on that assertion, since it's a bit more complex to test. I suppose we can use a `ref` on the modal again in order to listen for mounts and unmounts, and keep that stored in a bool. I'll try that.

matchu: It's easy to assert that the modal is not present in the DOM via `querySelector`… but it's not a very helpful assertion if we don't also assert at some point that the modal _was_ in the DOM, and there's not an obvious hook for when to do that :(

matchu: I suppose it depends on the value proposition of asserting the modal's presence in the DOM. I don't think it actually simplifies the test, because I think we have to hook into lifecycle methods awkwardly in order to do that robustly. And we have other tests that assert that `onClose` is only called once the modal content will be unmounted, so I'm not worried about asserting that even more.

I'm also now starting to wonder whether a full mount is actually necessary here. I think Enzyme's shallow rendering _does_ trigger React lifecycle methods, so the event listeners will get registered anyway… maybe I should try that and see how it goes!

matchu: Hmm, thinking more about it, I think we should just punt the async rendering concerns. Fingers crossed that, once async rendering lands, we'll still be able to use synchronous rendering for unit tests!

kevinbarabash: Yeah, let's punt on them for now.  I hope we don't wait for async rendering in React before we have some more e2e tests for these components.

matchu:
@kevinbarabash Pushed a new version of the test that's simple and synchronous, with a note that async rendering might break it someday 😅

PR #62, opened 2018-05-02
Use the Grid for the one and two-column modal dialogs.

jeresig:
This diff adds in support for the one and two-column modal dialogs, using the new grid system. See #65 for the details of how the grid is ready.

A couple things that had fuzzy definitions in the design that I made a decision on:
* I added a 64px margin to the top and bottom of the modal, which seems to match the current design.
* I used a single 12-column grid for all two-column modals (which will be the same one I use for the standard modal, too).
* For the "11-column" modal the new design calls for a 128px space between the two pieces of text. I implemented that as: (in Sidebar: 32px space + 32px gutter + in Content: 64px space). This helps us to avoid having to implement a brand new type of grid just for this one modal.
* If the contents have a footer then I add a 48px padding to the bottom so that when the content scrolls up it looks nice (Zeplin shows it as 40px, but that's not one of our standard spacing sizes).

I haven't implemented mobile support yet, I'll handle that in a follow-up diff.

### DIFF
--- a/packages/wonder-blocks-modal/components/one-column-modal.js
+++ b/packages/wonder-blocks-modal/components/one-column-modal.js
@@ -10,12 +10,9 @@ import ModalCloseButton from "./modal-close-button.js";
 import ModalFooter from "./modal-footer.js";
 
 /**
- * A two-column modal layout.
+ * A one-column modal layout.
  */
-export default class TwoColumnModal extends React.Component<{
-    /** The sidebar column's content. */
-    sidebar: React.Node,
-
+export default class OneColumnModal extends React.Component<{
     /**
      * Called when the close button is clicked.
      *
@@ -30,57 +27,27 @@ export default class TwoColumnModal extends React.Component<{
      */
     onClickCloseButton: () => void,
 
-    /** The footer to show under the main column's content. */
+    /** The footer to show under the contents. */
     footer?: React.Node,
 
-    /** The main column's content. */
+    /** The contents of the modal. */
     content: React.Node,
-
-    /** Should the sidebar be made smaller? (e.g. 1 column wide) */
-    smallSidebar: boolean,
 }> {
     static defaultProps = {
         onClickCloseButton: () => {},
-        smallSidebar: false,
     };
 
-    renderSidebar(sidebar: React.Node, smallSidebar: boolean) {
-        if (smallSidebar) {
-            return (
-                <Cell
-                    smCols={4}
-                    largeCols={1}
-                    style={[
-                        styles.column,
-                        styles.bgSidebar,
-                        styles.bgSidebarSmall,
-                    ]}
-                >
-                    {sidebar}
-                </Cell>
-            );
-        }
-
-        return (
-            <FlexCell style={[styles.column, styles.bgSidebar]}>
-                <View style={styles.sidebar}>{sidebar}</View>
-            </FlexCell>
-        );
-    }
-
     render() {
-        const {sidebar, footer, content, smallSidebar} = this.props;
+        const {footer, content} = this.props;
 
         return (
             <View style={styles.container}>
                 <Grid spec={GRID_MODAL_SPEC}>
                     <Row flush>
-                        {this.renderSidebar(sidebar, smallSidebar)}
                         <FlexCell style={[styles.column, styles.bgContents]}>
                             <View
                                 style={[
                                     styles.contents,
-                                    smallSidebar && styles.contentsSmall,
                                     footer && styles.contentsHasFooter,
                                 ]}
                             >
@@ -96,7 +63,7 @@ export default class TwoColumnModal extends React.Component<{
                 </Grid>
                 <View style={styles.closeButton}>
                     <ModalCloseButton
-                        color="light"
+                        color="dark"
                         onClick={this.props.onClickCloseButton}
                     />
                 </View>
@@ -110,13 +77,10 @@ const styles = StyleSheet.create({
         margin: "0 auto",
         position: "relative",
 
-        width: "86.72%",
+        width: "64.65%",
 
         borderRadius: 4,
         overflow: "hidden",
-
-        // For sidebar
-        minHeight: 464,
     },
 
     closeButton: {
@@ -130,54 +94,21 @@ const styles = StyleSheet.create({
         paddingBottom: 64,
         flex: "1 1 50%",
         boxSizing: "border-box",
-
-        // For sidebar
-        minHeight: 464,
-    },
-
-    bgSidebar: {
-        background: Color.darkBlue,
-        color: Color.white,
-    },
-
-    bgSidebarSmall: {
-        flex: "0",
     },
 
     bgContents: {
         background: Color.white,
         position: "relative",
         flexDirection: "column",
-
-        // For sidebar
-        display: "flex",
-    },
-
-    sidebar: {
-        marginRight: 32,
     },
 
     contents: {
         overflow: "auto",
         flex: "1 0 0",
-
-        // For sidebar
-        marginLeft: 64,
-        maxWidth: 472,
     },
 
     contentsHasFooter: {
         paddingBottom: 48,
-    },
-
-    contentsSmall: {
-        // Center the contents when the sidebar is small
-        margin: "0 auto",
-
-        // Add in some extra padding to the contents to
-        // simulate the extra gutters used for spacing
-        paddingLeft: 64,
-        paddingRight: 32,
     },
 
     footer: {
--- a/packages/wonder-blocks-modal/components/one-column-modal.md
+++ b/packages/wonder-blocks-modal/components/one-column-modal.md
@@ -1,0 +1,119 @@
+**One-Column, with footer:**
+
+A one-column modal dialog with a footer on the contents. The contents expand to fill the container and it continues to grow in size. There is no minimum height to the modal.
+
+```js
+const {StyleSheet, css} = require("aphrodite");
+const {View} = require("wonder-blocks-core");
+const {Title, Body} = require("wonder-blocks-typography");
+const OneColumnModal = require("./one-column-modal.js").default;
+
+const styles = StyleSheet.create({
+    example: {
+        // Checkerboard background
+        backgroundImage: "linear-gradient(45deg, #ccc 25%, transparent 25%), linear-gradient(-45deg, #ccc 25%, transparent 25%), linear-gradient(45deg, transparent 75%, #ccc 75%), linear-gradient(-45deg, transparent 75%, #ccc 75%)",
+        backgroundSize: "20px 20px",
+        backgroundPosition: "0 0, 0 10px, 10px -10px, -10px 0px",
+
+        padding: 32,
+    },
+
+    title: {
+        marginBottom: 16,
+    },
+});
+
+const content = <View>
+    <Title style={styles.title}>Contents</Title>
+    <Body>
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed
+        do eiusmod tempor incididunt ut labore et dolore magna
+        aliqua. Ut enim ad minim veniam, quis nostrud exercitation
+        ullamco laboris nisi ut aliquip ex ea commodo consequat.
+        Duis aute irure dolor in reprehenderit in voluptate velit
+        esse cillum dolore eu fugiat nulla pariatur. Excepteur sint
+        occaecat cupidatat non proident, sunt in culpa qui officia
+        deserunt mollit anim id est.
+    </Body>
+    <Body>
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed
+        do eiusmod tempor incididunt ut labore et dolore magna
+        aliqua. Ut enim ad minim veniam, quis nostrud exercitation
+        ullamco laboris nisi ut aliquip ex ea commodo consequat.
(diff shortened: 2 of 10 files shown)


jeresig:

Built with commit e8ad0db004eeaaa5a053541bd39611a1f8023f9c

https://deploy-preview-62--wonder-blocks.netlify.com

jeresig:

Built with commit 0f2804cc1d27cc09dbeef810af62a17732536c1f

https://deploy-preview-62--wonder-blocks.netlify.com

matchu:
Eep, a bit surprising and scary to see so many changes to the modal baked into this one! 😬😬😬 Are these meant to be prototype-y to enable you to try out the grid, or are they meant to be finalized implementations of these layouts? That'll affect how carefully I review them.

jeresig:
@matchu I guess it depends on how you want to use what I implemented. I did try to implement all of the modal changes to the best of my ability and matching the spec as closely as possible - I do hope that we can use it! It was challenging as I had to make sure that the grid could work for every modal case we wanted to throw at it, which essentially meant implementing the entire two-column (and one-column) modal. If you want to just focus on the modal-related changes, that's fine, I can defer to @kevinbarabash for input on the Grid stuff. Up to you!

packages/wonder-blocks-modal/components/two-column-modal.js
```diff
@@ -27,6 +25,15 @@ type Props = {
      * clicks too.)
      */
     onClickCloseButton: () => void,
+
+    /** The footer to show under the main column's content. */
+    footer?: React.Node,
+
+    /** The main column's content. */
+    contents: React.Node,
```
matchu: I'm not sure this column actually is more "main" than the other. There's [some discussion][1] in the spec about this, and it's a bit in-flux. I proposed `leftColumn` and `rightColumn` to avoid the implied hierarchy, but then Kevin pointed out that maybe there _is_ some natural hierarchy because only the main column is present in the mobile design, then I pointed out that maybe that's not the intent of the design spec? We meant to ask Tabitha, but I'm not sure where we got with that. I've added a Zeplin comment now (51).

In any case, if we do decide to go with the hierarchical naming, I think this is `content` in the spec rather than `contents`.

[1]: https://docs.google.com/a/khanacademy.org/document/d/1__cz2bHvZGWtBIcf_W0IGpGtgu1HKgHIjS9mmYsjbxk/edit?disco=AAAABw4OXrQ

jeresig:
@matchu Just a heads-up that Tabitha is going to be reviewing the modals today, as well.

packages/wonder-blocks-modal/components/two-column-modal.js
```diff
@@ -35,33 +42,73 @@ type Props = {
...
+                style={[styles.column, styles.bgSidebar, styles.bgSidebarSmall]}
+            >
+                {sidebar}
+            </Cell>
+        ) : (
+            <FlexCell style={[styles.column, styles.bgSidebar]}>
+                <View style={styles.sidebar}>{sidebar}</View>
+            </FlexCell>
+        );
+
```
matchu: I wonder if maybe we could pull all this stuff into a separate `_renderSidebar` method? Then use `if` blocks and `return` statements to handle the `null`, `smallSidebar`, and `!smallSidebar` cases? In this form, it's hard for me to keep track of the flow of rendering.

packages/wonder-blocks-modal/components/two-column-modal.js
```diff
@@ -35,33 +42,73 @@ type Props = {
...
+                                styles.column,
+                                styles.bgContents,
+                                !sidebar && styles.bgContentsNoSidebar,
+                            ]}
+                        >
+                            <View
+                                style={[
+                                    styles.contents,
+                                    smallSidebar && styles.contentsSmall,
+                                    !!footer && styles.contentsHasFooter,
```
matchu: Surprised to see a double-bang here when the line above doesn't have one. Is that a Flow typing thing?

packages/wonder-blocks-modal/components/two-column-modal.js
```diff
@@ -35,33 +42,73 @@ type Props = {
...
         );
     }
 }
 
 const styles = StyleSheet.create({
     container: {
-        display: "flex",
-        flexDirection: "row",
-        alignItems: "stretch",
+        margin: "0 auto",
```
matchu: Curious what this self-centering rule is for.

packages/wonder-blocks-modal/components/two-column-modal.js
```diff
@@ -78,32 +125,85 @@ const styles = StyleSheet.create({
         top: 8,
     },
 
-    column: {
-        boxSizing: "border-box",
-        flex: "0 0 50%",
+    containerNoSidebar: {
+        minHeight: "auto",
```
matchu: What does this rule do?

packages/wonder-blocks-modal/components/two-column-modal.js
```diff
@@ -78,32 +125,85 @@ const styles = StyleSheet.create({
...
+    },
 
-        // TODO(mdr): Some call sites probably won't want built-in padding. How
-        //     should we handle that possibility?
+    column: {
         paddingTop: 64,
         paddingBottom: 64,
+        minHeight: 464,
+        flex: "1 1 50%",
```
matchu: I feel like this rule is a bit misleading, because the 50% flex-basis isn't really relevant for the small-sidebar and 1-column cases, but happens to work anyway due to clever subtleties. I'd kinda prefer that we break out the styles for these various cases into entirely different style classes, to help make the expected behavior clearer for future readers.

packages/wonder-blocks-modal/components/two-column-modal.js
```diff
@@ -78,32 +125,85 @@ const styles = StyleSheet.create({
...
+        paddingBottom: 48,
+    },
+
+    contentsNoSidebar: {
+        marginLeft: 0,
+        maxWidth: "auto",
+    },
+
+    contentsSmall: {
+        margin: "0 auto",
```
matchu: What does this do?

jeresig: In the "2-column with subordinated preview" (or in the implementation, when `smallSidebar` is `true`) the contents are essentially centered in the container - although I realize that I got the margins slightly wrong (as you noted where it dropped to 0) so I'll fix that.

packages/wonder-blocks-modal/components/two-column-modal.js
jeresig: Hmm, ok! I'll see if I can split apart some of these rules to make them clearer (rather than having styles that override each other).

packages/wonder-blocks-modal/components/two-column-modal.js
jeresig: That was overriding the minHeight defined in `container`, but I think I'm going to try and reformulate this so it isn't necessary.

packages/wonder-blocks-modal/components/two-column-modal.js
jeresig: I did this to center the modal inside the container. Although maybe I shouldn't be doing that and should leave it up to the container to take care of? Thoughts?

packages/wonder-blocks-modal/components/two-column-modal.js
jeresig: Oh - `smallSidebar` is a boolean and `footer` is potentially a node, or null, or whatever. I like trying to keep it clean so that I'm consistently using booleans - don't think Flow cares about this, TBH.

packages/wonder-blocks-modal/components/two-column-modal.js
jeresig: Good call! I can do that.

packages/wonder-blocks-modal/components/two-column-modal.js
matchu: Yeah, shrug! I would generally expect the modal to handle alignment itself, but this isn't necessarily a ridiculous default. (Then again, I'm not sure how easy it is for parents to override things like `margin: 0 auto`—if it's tricky, maybe it's worth erring toward giving the parent more freedom.)

packages/wonder-blocks-modal/components/two-column-modal.js
jeresig: Hmm - ok! So when I saw that the "sidebar" went away in the mobile view that definitely indicated to me that it was secondary. But yeah, it's also not always a sidebar as it goes to the top in the mobile view. @tyong any thoughts on what we should be naming these things?

jeresig:
@matchu I think I addressed most of your feedback! I also split out all of the Grid changes into a separate diff to make it easier to review. Also I made a new OneColumnGrid component and broke that out and that helped to simplify things. I also made a new private ModalFooter component which is shared between the One and Two-column grid. Would love a re-review, thank you!

packages/wonder-blocks-modal/components/one-column-modal.md
```diff
@@ -0,0 +1,119 @@
...
+A one-column modal dialog with a footer on the contents. The contents expand to fill the container and it continues to grow in size. There is no minimum height to the modal.
+
+```js
+const {StyleSheet, css} = require("aphrodite");
+const {View} = require("wonder-blocks-core");
+const {Title, Body} = require("wonder-blocks-typography");
+const OneColumnModal = require("./one-column-modal.js").default;
+
+const styles = StyleSheet.create({
+    example: {
```
matchu: I updated some of these styles in #64, so we'll probably want to keep that in mind when we merge these changes together. (Maybe it's approaching time to actually pull them into a shared example component!)

packages/wonder-blocks-modal/components/two-column-modal.js
```diff
@@ -27,49 +29,94 @@ type Props = {
...
- * A two-column modal layout.
- */
-export default class TwoColumnModal extends React.Component<Props> {
+    /** The footer to show under the main column's content. */
+    footer?: React.Node,
+
+    /** The main column's content. */
+    content: React.Node,
+
+    /** Should the sidebar be made smaller? (e.g. 1 column wide) */
```
matchu: nit: should this be an "i.e." ("in other words") instead of an "e.g." ("for example")? Or are we maybe using that to suggest that being exactly 1 column isn't part of the contract, so we're framing it as an example of what we might do?

packages/wonder-blocks-modal/components/two-column-modal.js
```diff
@@ -79,31 +126,63 @@ const styles = StyleSheet.create({
...
+        margin: "0 auto",
+
+        // Add in some extra padding to the contents to
+        // simulate the extra gutters used for spacing
+        paddingLeft: 64,
+        paddingRight: 32,
+    },
+
+    footer: {
+        position: "absolute",
```
matchu: Hmm gotcha, I think I see why depending on pure Flexbox is trickier in the two-column case than the one-column case: the grid component has padding, but we want the footer to escape that padding.

I'm still surprised that absolute positioning doesn't disrupt the Flex children here, but I guess, if it works, it works! Long-term, I wonder if this is a hint that we'll need a built-in escape hatch to allow certain grid children to be full-bleed, but I'm okay with this for the modal case if it's working consistently!

packages/wonder-blocks-modal/components/one-column-modal.js
```diff
@@ -0,0 +1,119 @@
...
+        overflow: "auto",
+        flex: "1 0 0",
+    },
+
+    contentsHasFooter: {
+        paddingBottom: 48,
+    },
+
+    footer: {
+        position: "absolute",
```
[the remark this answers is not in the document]
jeresig: Yeah, this is to escape the padding - and unfortunately it's not possible to move it outside (I would've loved to have done this!) as its positioning is also defined by the grid system 😭

packages/wonder-blocks-modal/components/one-column-modal.md
jeresig: Good idea! I'll wait on you and I can go back through and update these once landed.

packages/wonder-blocks-modal/components/two-column-modal.js
jeresig: Right - I'll fix this, I think it should probably just be `(exactly 1 column wide)`

packages/wonder-blocks-modal/components/two-column-modal.js
jeresig: Yeah, we're definitely going to want an "escape hatch"/"full-bleed" component. For now I just have the footer that I've hacked together but I'm not sure if it makes sense to use absolute positioning for this stuff all the time? I can think about this more and maybe I can come up with a better solution in a follow-up commit.

jeresig:
@matchu I'm investigating a fix for the scrollbar being too close - it definitely has to do with the grid but I'm seeing if I can come up with another way...

matchu:
Oh, wait, does this depend on #65 which isn't merged? I think I see now, sorry.

jeresig:
@matchu yeahhh - and now that we're not using the grid I'm likely going to start over. I think I'll wait until you've had a go at things though. Feel free to do whatever work you need to do today and I can pick this back up again on Monday.

matchu:
Okay, sorry it didn't work out! Thanks John!

PR #63, opened 2018-05-02
[Button v1] Create ButtonCore component

chenyejia:
In this change, we create a stateless `ButtonCore` component that will render all the different states of `Button`.

Implemented:
* Styling for all five states of `Button` (default, disabled, hovered, focused, pressed).

Not yet implemented:
* Rendering a spinner instead of content
* Rendering `Icon`s
* Custom `style`
* `ButtonBehavior`

Test Plan:
View on Styleguidist


jeresig:

Built with commit 719dd2f322c80d1da5253eb3220d60fa5fd37594

https://deploy-preview-63--wonder-blocks.netlify.com

jeresig:

Built with commit 7fa23a1f9f0b2cbeeee6f552374dfb2b27cbeeaf

https://deploy-preview-63--wonder-blocks.netlify.com

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -0,0 +1,271 @@
...
+        paddingRight: 16,
+        border: "none",
+        borderRadius: 4,
+        cursor: "pointer",
+        outline: "none",
+    },
+    disabled: {
+        cursor: "auto",
+    },
+    focusRing: {
```
jeresig: None of these focusRings appear to be used anywhere?

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -0,0 +1,271 @@
...
+                </button>
+            );
+        }
+        return (
+            <button
+                className={css(
+                    sharedStyles.shared,
+                    size === "small" && sharedStyles.small,
+                    buttonStyles.default,
+                    pressed && buttonStyles.active,
```
jeresig: Another way to write these two lines which might make it a little bit clearer?:

```
(pressed
    ? buttonStyles.active
    : (hovered || focused) && buttonStyles.focus)
```

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -0,0 +1,271 @@
...
+            light,
+            size,
+            disabled,
+            hovered,
+            focused,
+            pressed,
+        } = this.props;
+
+        const buttonStyles = _generateStyles(
+            color || Color.blue,
```
jeresig: Ah, I think it would be better to change this so that we have the following in the class, then we can just pass them in here and it'll work!

```
static defaultProps = {
    color: Color.blue,
    kind: "primary",
    light: false,
};
```

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -0,0 +1,271 @@
...
+import {StyleSheet, css} from "aphrodite";
+
+import {LabelLarge} from "wonder-blocks-typography";
+import {processStyleList} from "../../wonder-blocks-core/util/util.js";
+import Color, {mix, fade} from "wonder-blocks-color";
+import type {ValidTints} from "wonder-blocks-color";
+import type {SharedProps} from "../types.js";
+
+type Props = SharedProps & {
+    hovered?: boolean,
```
jeresig: I think it might make sense to have all three of these be required. ButtonCore is only used internally and we want to make sure that we never forget to pass in one of these options (that would be a bad sign if we forgot!)

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -0,0 +1,271 @@
...
+            pressed,
+        } = this.props;
+
+        const buttonStyles = _generateStyles(
+            color || Color.blue,
+            kind || "primary",
+            !!light,
+        );
+
+        if (disabled) {
```
jeresig: Hmm - it doesn't feel like we need a whole separate branch for the disabled setting. I think we can just do:

```
disabled && sharedStyles.disabled
...
disabled && buttonStyles.disabled
```

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -0,0 +1,271 @@
...
+                borderWidth: 2,
+            },
+            active: {
+                color: mix(fade(color, 0.32), Color.white),
+            },
+            disabled: {
+                color: mix(Color.white, fade(color, 0.32)),
+            },
+        };
+    }
```
jeresig: It feels like we should probably throw an exception if none of these options are selected - it would be a really bad sign if we got to that point!

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -0,0 +1,271 @@
+// @flow
+import React from "react";
+import {StyleSheet, css} from "aphrodite";
+
+import {LabelLarge} from "wonder-blocks-typography";
+import {processStyleList} from "../../wonder-blocks-core/util/util.js";
+import Color, {mix, fade} from "wonder-blocks-color";
+import type {ValidTints} from "wonder-blocks-color";
```
jeresig: This doesn't appear to be used in this file.

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -0,0 +1,271 @@
+// @flow
+import React from "react";
+import {StyleSheet, css} from "aphrodite";
+
+import {LabelLarge} from "wonder-blocks-typography";
+import {processStyleList} from "../../wonder-blocks-core/util/util.js";
```
jeresig: Not used.

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -0,0 +1,271 @@
...
+                        buttonStyles.disabled,
+                    )}
+                    disabled
+                >
+                    <LabelLarge>{this.props.children}</LabelLarge>
+                </button>
+            );
+        }
+        return (
+            <button
```
jeresig: @kevinbarabash It feels like we should be using a View here to be able to support easy styling - however View doesn't support custom tags. Should that be changed?

packages/wonder-blocks-button/components/button.js
```diff
@@ -1,52 +1,68 @@
...
-                )}
-                onClick={this.props.onClick}
+            <div
+                onClick={this.props.onClick && this.props.onClick}
+                href={this.props.href && this.props.href}
             >
-                <LabelLarge>Hello, world!</LabelLarge>
-            </a>
+                <ButtonCore
+                    icon={this.props.icon ? this.props.icon : undefined}
```
jeresig: Hmm... I think you can simplify a lot of this by doing: `{...this.props}` and removing all of the props that you're just passing through.

packages/wonder-blocks-button/components/button.js
```diff
@@ -1,52 +1,68 @@
...
     render() {
         return (
-            <a
-                className={css(
-                    styles.button,
-                    styles[this.props.color || "blue"],
-                )}
-                onClick={this.props.onClick}
+            <div
+                onClick={this.props.onClick && this.props.onClick}
```
jeresig: Both of these can just be: `onClick={this.props.onClick}`, you don't need to confirm that they exist, React handles null/false/undefined correctly.

packages/wonder-blocks-button/components/button.js
```diff
@@ -1,52 +1,68 @@
...
 export default class Button extends React.Component<Props> {
     render() {
         return (
-            <a
-                className={css(
-                    styles.button,
-                    styles[this.props.color || "blue"],
-                )}
-                onClick={this.props.onClick}
+            <div
```
jeresig: If you're accessing a couple properties in a file it might be worth it to get them all doing something like: `const {onClick, href} = this.props;` and then just use those variables.

packages/wonder-blocks-button/components/button.js
```diff
@@ -1,52 +1,68 @@
...
-        backgroundColor: Color.green,
-    },
-    gold: {
-        backgroundColor: Color.gold,
-    },
-    red: {
-        backgroundColor: Color.red,
-    },
-});
+const styles = StyleSheet.create({});
```
jeresig: I wonder if Button will never have any styling of its own? It doesn't feel like it would!

packages/wonder-blocks-button/types.js
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * If true, replaces the contents with a spinner.
+     */
+    // TODO(yejia): Implement once spinner is implemented.
+    // spinner: boolean,
+
+    /**
+     * The color of the button, either blue (default) or red.
+     */
+    color?: ValidTints,
```
jeresig: It feels like a number of these should be required and then we should set sensible defaults using `static defaultProps`. Specifically color (Color.blue), kind (primary), light (false), size (default), and disabled (false). If you do this inside the Button component then you won't have to inside ButtonCore, you can just use the values.

packages/wonder-blocks-button/types.js
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Whether the button is disabled. Default is false.
+     */
+    disabled?: boolean,
+
+    /**
+     * Test ID used for e2e testing.
+     */
+    testId: string,
```
jeresig: This should be optional.

styleguide.config.js
```diff
@@ -33,6 +33,11 @@ module.exports = {
                 "**/components/modal-launcher-portal.js",
             ],
         },
+        {
+            name: "Button",
+            content: "packages/wonder-blocks-button/docs.md",
+            components: "packages/wonder-blocks-button/components/*.js",
```
jeresig: You should ignore button-core.js (like we did for some of the modal files, above) as we aren't expecting the user to ever interact with it directly.

packages/wonder-blocks-button/docs.md
```diff
@@ -1,11 +1,6 @@
...
-<Button>default button</Button>
-```
+A `Button` is a clickable button element consisting of a `ButtonBehavior` (TODO(yejia)) surrounding a `ButtonCore`. `ButtonBehavior` handles interactions and state changes. `ButtonBehavior` is a stateless component which displays the different states the `Button` can take.
 
-Caution Button:
-```js
-<Button color={"red"}>caution</Button>
-```
+### Usage
+TODO(yejia)
```
jeresig: Can you move button-core.md into this file? It would make for a great demonstration of all the possible states/sizes.

packages/wonder-blocks-button/docs.md
```diff
@@ -1,11 +1,6 @@
 **WARNING: Button is not finished, do not use!**
 
-Default Button:
-```js
-<Button>default button</Button>
-```
+A `Button` is a clickable button element consisting of a `ButtonBehavior` (TODO(yejia)) surrounding a `ButtonCore`. `ButtonBehavior` handles interactions and state changes. `ButtonBehavior` is a stateless component which displays the different states the `Button` can take.
 
-Caution Button:
-```js
```
jeresig: Also, can you add some samples of the small sizes to button-core.md? I'm curious to see what they look like (especially for the focus/hover state).

packages/wonder-blocks-button/types.js
```diff
@@ -0,0 +1,85 @@
+// @flow
+import React from "react";
+import type {Element} from "react";
```
jeresig: This doesn't appear to be used.

packages/wonder-blocks-button/types.js
```diff
@@ -0,0 +1,85 @@
...
+
+import type {ValidTints} from "wonder-blocks-color";
+
+export type SharedProps = {
```
jeresig: Even though it helps to cut down on duplicate code I think we may want to move this in to button.js (and parts of it into button-core.js) and merge it in with their existing props in order to ensure that the styleguidist prop documentation is generated correctly.

chenyejia:
@jeresig I think I am working on the correct copy? I cloned this repo, ran `npm install`, and I do see pre-commit checks running when I try to commit. I don't see any errors when I try to run `npm run flow`?

packages/wonder-blocks-button/components/button-core.md
```diff
@@ -0,0 +1,521 @@
+ButtonCores:
+```js
```
kevinbarabash: I would break this up into multiple `js` blocks.  This will make the `view code` output more reasonable.

packages/wonder-blocks-button/components/button-core.js
kevinbarabash: Another option is to have a single required prop called `state`, e.g.
```
state: "default" | "hovered" | "pressed" | "focused" | "disabled"
```
That we we don't have to pass `focused={false} pressed={false} hovered={true}` or combinations thereof.

packages/wonder-blocks-button/components/button-core.js
kevinbarabash: We could also add `state: "default"` as part of the defaultProps.

packages/wonder-blocks-button/components/button-core.js
kevinbarabash: That's an interesting idea.  It seems like that would could in pretty useful.  I wonder if we should have `View` also clear the styling on those tags. 🤔

packages/wonder-blocks-button/components/button-core.js
kevinbarabash: It should be fine to do `disabled={disabled}` on the button.

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -0,0 +1,271 @@
...
+            active: {
+                background: mix(Color.offBlack32, color),
+                color: mix(fade(color, 0.32), Color.white),
+            },
+            disabled: {
+                background: Color.offBlack32,
+                color: Color.white64,
+            },
+        };
+    } else if (kind === "primary" && light) {
```
kevinbarabash: I think this could be collapsed with the `if` block above by using ternaries for the properties that are different, e.g.
```
newStyles = {
   default: {
      backgroundColor: light ? Color.white : color,
      color: light ? color : Color.white,
   },
   ...
```

kevinbarabash: Should probably keep `primary`, `secondary`, `tertiary` separate though.

packages/wonder-blocks-button/components/button.js
kevinbarabash: We want to stub out `preventDefault` and `stopPropagation` though.

packages/wonder-blocks-button/components/button.js
kevinbarabash: I don't think the `div` wrapper is necessary.  Also, we should be passing the `style` prop to the `ButtonCore` which should then pass it to `<View tag={href ? "a" : "button"} style={[styles.clearStyle, buttonStyle, stylel]}>`.  We should avoid adding extract wrappers around the button itself since those may interfere with styles passed in to the `Button` component.

kevinbarabash: It's fine to leave styling to another PR.

packages/wonder-blocks-button/components/button.js
kevinbarabash: It's fine to leave `preventDefault` and `stopPropagation` to another PR.

packages/wonder-blocks-button/components/button-core.js
chenyejia: My understanding is that a `Button` can be in combinations of those states? Or do you mean since `ButtonCore` only displays any one of those states at a time, we just pass in the state with most precedence in `ButtonBehavior`?

packages/wonder-blocks-button/components/button-core.js
chenyejia: Ohhh did not know you could do that. Thanks!

kevinbarabash:
There are some weird messages in the tests about `@media` that I don't quite follow.  The other messages about snapshots not matching are pretty easy to fix.  Run `npm run jest -- -u` to update the .snap files.

jeresig:
@chenyejia I got the build to pass - I ran `npm run jest -- -u` as @kevinbarabash mentioned but then added in the new snapshots directory: `git add packages/wonder-blocks-button/__snapshots__/` and that seems to have done the trick! Will re-review now.

PR #64, opened 2018-05-02
[modal] Mobile-optimized design for standard modal

matchu:
In this change, we add mobile-optimized styles to `StandardModal`. We also tweak the example styles, to more accurately showcase mobile design.

We're not tackling other modal layouts yet; that's WB-130. Instead, we're producing one mobile layout, to showcase overall mobile behavior and get high-level design feedback, without slowing down iteration on the other modal layouts with this additional layer of complexity.

We also don't yet prevent the document from scrolling when this modal is open. That's up next, in WB-149.

(When tweaking the examples, I noticed that callers are currently responsible for a number of content style decisions, like content area padding on desktop and mobile, which we might want to standardize. We _don't_ want to include this padding in _all_ modals, because many modals don't need it and want to be full-bleed—but maybe it should be a default padding controlled by a flag, or there should be a `ModalContent` component? Long-term food for thought!)

# Test Plan:
Styleguidist examples for `StandardModal` look great on mobile!
screenshot

Here it is launched on the page.
screenshot

Also, confirm that all the examples still look as expected on desktop.
screenshot

### DIFF
--- a/packages/wonder-blocks-modal/components/standard-modal.js
+++ b/packages/wonder-blocks-modal/components/standard-modal.js
@@ -11,6 +11,7 @@ import {
 } from "wonder-blocks-typography";
 
 import ModalCloseButton from "./modal-close-button.js";
+import {smOrSmaller} from "../util/util.js";
 
 type Props = {
     /**
@@ -114,6 +115,15 @@ const styles = StyleSheet.create({
         background: Color.white,
         borderRadius: 4,
         overflow: "hidden",
+
+        // On mobile, we consume the full screen size.
+        [smOrSmaller]: {
+            width: "100%",
+            height: "100%",
+            maxWidth: "none",
+            maxHeight: "none",
+            borderRadius: 0,
+        },
     },
 
     titlebar: {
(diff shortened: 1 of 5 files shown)


jeresig:

Built with commit 9a4cec13f6e319d2497fba4a057226a12de07918

https://deploy-preview-64--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/standard-modal.js
```diff
@@ -114,6 +115,15 @@ const styles = StyleSheet.create({
         background: Color.white,
         borderRadius: 4,
         overflow: "hidden",
+
+        // On mobile, we consume the full screen size.
+        [smOrSmaller]: {
+            width: "100%",
+            height: "100%",
+            maxWidth: "none",
```
matchu: Ooh, I bet we also need to clear out `minHeight` here, or else the modal will overflow on not-tall-enough screens!

matchu:
Ehehe fun! 😬 Yeah, I'm curious what the general strategy is for grids that respond to media query. Should be exciting!

packages/wonder-blocks-modal/components/standard-modal.js
kevinbarabash: Did this end up being necessary or not?

matchu: Whoops, I missed this, thank you! Fixed in #72.

PR #65, opened 2018-05-02
Get Grid ready for Modals.

jeresig:
This diff includes a number of changes related to grids to get them ready for modals. See #62 for the details of how it will be used.

Major changes:

* Improved the handling of fixed-width cells in the Grid so that they wouldn't forcefully grow if their contents were larger than the cell.
* Added a new `flush` property to Rows to make the contents do a couple things: 1) Margins are removed and the first and last cells expand to consume the margins (adding padding inside of them automatically). 2) Gutters are removed and the width of the gutter is added to the cell before as padding.
* Fixed a bug with modal grids where their "small" size wasn't being detected as the naming of the different grids conflicted with each other.
* Reduced the number of modal grid layouts down to just one.

### DIFF
--- a/packages/wonder-blocks-grid/components/flex-cell.js
+++ b/packages/wonder-blocks-grid/components/flex-cell.js
@@ -9,11 +9,11 @@ import type {GridSize} from "../util/types.js";
 
 type Props = {
     /** Should this cell be shown on a Small Grid? */
-    small?: boolean,
+    small: boolean,
     /** Should this cell be shown on a Medium Grid? */
-    medium?: boolean,
+    medium: boolean,
     /** Should this cell be shown on a Large Grid? */
-    large?: boolean,
+    large: boolean,
     /**
      * The child components to populate inside the cell. Can also accept a
      * function which receives the `gridSize` and `totalColumns` and should
--- a/packages/wonder-blocks-grid/components/grid.js
+++ b/packages/wonder-blocks-grid/components/grid.js
@@ -30,7 +30,7 @@ type Props = {
     spec: GridSpec,
 
     /** The `<Row>` components that will make up the grid */
-    children: React.ChildrenArray<Row>,
+    children: React.Node,
 };
 
 /**
@@ -66,7 +66,7 @@ export default class Grid extends React.Component<
     },
 > {
     static WATCHERS: {
-        [size: GridSize]: any,
+        [query: string]: any,
     } = {};
 
     static defaultProps = {
@@ -104,6 +104,11 @@ export default class Grid extends React.Component<
         this.watchHandlers = {};
 
         for (const size of VALID_GRID_SIZES) {
+            // Skip sizes that aren't defined
+            if (!spec[size]) {
+                continue;
+            }
+
             const {query} = spec[size];
 
             // Don't watch sizes that don't have an associated query
@@ -112,11 +117,11 @@ export default class Grid extends React.Component<
             }
 
             // Create a new matchMedia watcher if one doesn't exist yet
-            if (!Grid.WATCHERS[size]) {
-                Grid.WATCHERS[size] = window.matchMedia(query);
+            if (!Grid.WATCHERS[query]) {
+                Grid.WATCHERS[query] = window.matchMedia(query);
             }
 
-            const watcher = Grid.WATCHERS[size];
+            const watcher = Grid.WATCHERS[query];
 
             // Attach a handler that watches for the change, saving a
             // references to it so we can remove it later
@@ -154,7 +159,14 @@ export default class Grid extends React.Component<
         // We go through the component and remove all of the listeners
         // that this Grid attached.
         for (const size of VALID_GRID_SIZES) {
-            const watcher = Grid.WATCHERS[size];
+            // Skip sizes that aren't defined
+            if (!spec[size]) {
+                continue;
+            }
+
+            const {query} = spec[size];
+            const watcher = Grid.WATCHERS[query];
+
             if (watcher) {
                 const handler = this.watchHandlers[size];
                 watcher.removeListener(handler);
@@ -185,7 +197,7 @@ class GridContext extends React.Component<{
     spec: GridSpec,
 
     // The Row components that will make up the grid
-    children: React.ChildrenArray<Row>,
+    children: React.Node,
 }> {
     static childContextTypes = gridContextTypes;
 
--- a/packages/wonder-blocks-grid/components/row.js
+++ b/packages/wonder-blocks-grid/components/row.js
@@ -33,11 +33,11 @@ import type {GridSize} from "../util/types.js";
  */
 export default class Row extends React.Component<{
     /** Should this row be shown on a Small Grid? */
-    small?: boolean,
+    small: boolean,
     /** Should this row be shown on a Medium Grid? */
-    medium?: boolean,
+    medium: boolean,
     /** Should this row be shown on a Large Grid? */
-    large?: boolean,
+    large: boolean,
     /**
      * The child components to populate inside the row. Typically this will be
      * a mixture of [Cell](#cell), [FlexCell](#flexcell), and
@@ -52,6 +52,14 @@ export default class Row extends React.Component<{
               gridSize: GridSize,
               totalColumns: number,
           }) => React.Node),
+    /**
+     * Make the cell contents flush with the row. Removes the margins and
+     * gutters and adds their spacing as paddings to the cells. Margin
+     * spacing gets added as padding to the first and last cell inside the
+     * row. Gutter spacing gets added to the right-hand-side of all cells,
+     * except for the last one, as padding.
+     */
+    flush: boolean,
     /** The styling to apply to the row. */
     style?: any,
 }> {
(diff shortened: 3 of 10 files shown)


jeresig:

Built with commit f347ea51d3cfdd234701db1f25ac444c3060b36a

https://deploy-preview-65--wonder-blocks.netlify.com

jeresig:
@kevinbarabash I merged the modal stuff and it got put into this branch accidentally. You can ignore that and just focus on the grid stuff.

packages/wonder-blocks-grid/components/flex-cell.js
```diff
@@ -9,11 +9,11 @@ import type {GridSize} from "../util/types.js";
...
 type Props = {
     /** Should this cell be shown on a Small Grid? */
-    small?: boolean,
+    small: boolean,
     /** Should this cell be shown on a Medium Grid? */
-    medium?: boolean,
+    medium: boolean,
     /** Should this cell be shown on a Large Grid? */
-    large?: boolean,
+    large: boolean,
```
kevinbarabash: Why are these mandatory now?

packages/wonder-blocks-grid/components/grid.js
```diff
@@ -30,7 +30,7 @@ type Props = {
     spec: GridSpec,
 
     /** The `<Row>` components that will make up the grid */
-    children: React.ChildrenArray<Row>,
+    children: React.Node,
```
kevinbarabash: Why is this so general now?

packages/wonder-blocks-grid/components/flex-cell.js
kevinbarabash: I see, they should've been in the first place since you're specifying default values.  Cool.  I'm usually lazy when it comes to bools and don't bother b/c `undefined` is falsy.

packages/wonder-blocks-grid/components/grid.js
```diff
@@ -66,7 +66,7 @@ export default class Grid extends React.Component<
     },
 > {
     static WATCHERS: {
-        [size: GridSize]: any,
+        [query: string]: any,
```
kevinbarabash: Why can't we use `GridSize` here?

packages/wonder-blocks-grid/components/flex-cell.js
kevinbarabash: I'm glad to see that react-docgen handles this case correctly.

packages/wonder-blocks-grid/util/specs.js
```diff
@@ -39,8 +39,8 @@ export const GRID_INTERNAL_SPEC: GridSpec = {
     },
 };
 
-// The default used for modals
-export const GRID_MODAL_12_SPEC: GridSpec = {
+// The default grid used for modals
+export const GRID_MODAL_SPEC: GridSpec = {
```
kevinbarabash: nice

packages/wonder-blocks-grid/components/grid.js
jeresig: I changed it! It's no longer keyed on the size, it's keyed on the media query string (which is more accurate).

packages/wonder-blocks-grid/components/grid.js
jeresig: I realized that it was far too strict - if anyone wanted to make a component that returned a Row it wouldn't work - so I just deferred to the docs to encourage it instead.

packages/wonder-blocks-grid/components/flex-cell.js
jeresig: Yep, exactly!

packages/wonder-blocks-grid/components/grid.js
kevinbarabash: Yeah, that's something that's pretty hard to model.  It's not like the component is inheriting from `Row` and yet it's still technically a `Row`.  I wonder if we could change this to be:
```
children: React.ChildrenArray<
   Row | React.Component<{render(): Row}> | () => Row
>
```

kevinbarabash: We do need it to be a `Row` or a component that renders a `Row` for this to work ,right?

packages/wonder-blocks-modal/components/one-column-modal.js
```diff
@@ -0,0 +1,119 @@
...
+        onClickCloseButton: () => {},
+    };
+
+    render() {
+        const {footer, content} = this.props;
+
+        return (
+            <View style={styles.container}>
+                <Grid spec={GRID_MODAL_SPEC}>
+                    <Row flush>
```
kevinbarabash: Nice to see this in the examples.

packages/wonder-blocks-grid/components/grid.js
jeresig: Nope! The Grid component doesn't actually render anything, it just sets the context. I did it before to try and provide some encouragement but honestly a grid can hold anything - and it can even be set way up the tree and only later have the row be used.

packages/wonder-blocks-modal/components/one-column-modal.js
kevinbarabash: I was playing around with this property and deselecting it didn't do what I expected it to do.  With `flush` I assumed that it would have no gutter, but instead it has no gutter with extra padding.

packages/wonder-blocks-grid/components/row.js
```diff
@@ -88,33 +99,72 @@ export default class Row extends React.Component<{
...
-                !item.props ||
-                !item.type.shouldDisplay ||
-                (typeof item.type.shouldDisplay === "function" &&
-                    item.type.shouldDisplay(item.props, gridSize))
+                typeof item === "object" &&
+                (!item.type ||
+                    !item.props ||
+                    !item.type.shouldDisplay ||
+                    (typeof item.type.shouldDisplay === "function" &&
+                        item.type.shouldDisplay(item.props, gridSize)))
```
kevinbarabash: omg, the safe-navigation operator can not come soon enough

packages/wonder-blocks-grid/components/row.js
```diff
@@ -88,33 +99,72 @@ export default class Row extends React.Component<{
...
                         <Gutter key={`gutter-${filteredContents.length}`} />,
                     );
                 }
 
-                filteredContents.push(item);
+                let newItem: Object = item;
+
+                // If the contents are flush then we insert in new padding to
+                // the cell to implement the first margin and the gutter spacing.
+                if (flush) {
```
kevinbarabash: I feel like `flush` is probably not a good name for what this does.  When I think about flush, I think about the contents lines up with the container.  I see... it's that the contents don't have any extra padding so we're extra padding.  Maybe we should call it `addPaddingToRow` or something like that.

packages/wonder-blocks-grid/index.js
```diff
@@ -1,10 +1,10 @@
...
-export * from "./components/flex-cell.js";
-export * from "./components/grid.js";
-export * from "./components/gutter.js";
-export * from "./components/row.js";
+export {default as Cell} from "./components/cell.js";
+export {default as FixedWidthCell} from "./components/fixed-width-cell.js";
+export {default as FlexCell} from "./components/flex-cell.js";
+export {default as Grid} from "./components/grid.js";
+export {default as Gutter} from "./components/gutter.js";
+export {default as Row} from "./components/row.js";
```
kevinbarabash: Nice

packages/wonder-blocks-grid/components/row.js
```diff
@@ -88,33 +99,72 @@ export default class Row extends React.Component<{
...
+                                item.props &&
+                                item.props.style,
+                            {
+                                paddingLeft: lastCell ? 0 : marginWidth,
+                                paddingRight: gutterWidth,
+                            },
+                        ],
+                    });
+
+                    lastCell = newItem;
```
kevinbarabash: In this case `lastCell` is more like `prevCell`

packages/wonder-blocks-grid/components/row.js
```diff
@@ -88,33 +99,72 @@ export default class Row extends React.Component<{
...
+                (!item.type ||
+                    !item.props ||
+                    !item.type.shouldDisplay ||
+                    (typeof item.type.shouldDisplay === "function" &&
+                        item.type.shouldDisplay(item.props, gridSize)))
             ) {
-                if (hasVisibleCell) {
+                // Figure out if we need to try and insert a gutter in-between
+                // cells. If the contents are flush then we don't insert any.
+                if (hasVisibleCell && !flush) {
```
kevinbarabash: Does this mean that there won't be any spaces between the columns within this row?  I thought `flush` was all about the padding at the edges of the row.

packages/wonder-blocks-modal/components/one-column-modal.js
jeresig: Heya! I'm curious if this matches what I have explaining the `flush` prop in the docs? "Make the cell contents flush with the row. Removes the margins and gutters and adds their spacing as paddings to the cells. Margin spacing gets added as padding to the first and last cell inside the row. Gutter spacing gets added to the right-hand-side of all cells, except for the last one, as padding."

packages/wonder-blocks-grid/components/row.js
jeresig: Well, in the context of a modal the padding exists - it's just outside of the modal and is invisible :D by putting it inside it becomes visible! If this was done in another context where more background colors were at play it would be a bit more obvious.

packages/wonder-blocks-grid/components/row.js
jeresig: flush does both! See my comment above.

jeresig:
Ok, I'm going to abandon this as it's no longer high priority given that we don't need to be using Grid for the modal stuff right now.

PR #66, opened 2018-05-02
[modal] Prevent scrolling when modal is open

matchu:
In this change, we prevent the document from scrolling when a modal is open. This scrolling was especially problematic in the mobile case, where the document could scroll without you having _any_ visual feedback (since the modal is full-screen).

To do this, we copied `ScrollDisabler` from webapp's modal package. This component also covers some tricky edge cases, like mobile Safari. I didn't write unit tests for it, because it seemed high-cost, this component has been stress-tested in prod, and this file is likely to go untouched for a long time. Instead, I just test in `ModalLauncher` that we render `ScrollDisabler` at the correct times.

https://khanacademy.atlassian.net/browse/WB-149

Test Plan:
```npm test```

In Styleguidist, use the `ModalLauncher` examples to open a modal. Confirm that the document is not scrollable, in desktop or mobile.


jeresig:

Built with commit c746d23e472d1746b0969d601685705293fdb4bd

https://deploy-preview-66--wonder-blocks.netlify.com

PR #67, opened 2018-05-02
Add Spacing v1 package.

jeresig:
This adds the Spacing package and uses it in the Grid package. I'll use it in Modal once the dust settles there, as well. Accessing the sizes by name is a bit clunky - I find it very hard to remember and have to constantly look it up (like I do with the colors) but I'm not of a better solution? Hmm.


packages/wonder-blocks-grid/util/specs.js
```diff
@@ -11,20 +13,20 @@ export const GRID_DEFAULT_SPEC: GridSpec = {
     small: {
         query: "(max-width: 767px)",
         totalColumns: 4,
-        gutterWidth: 16,
-        marginWidth: 16,
+        gutterWidth: Spacing.m,
+        marginWidth: Spacing.m,
```
kevinbarabash: Thanks for updating these to use `Spacing`.

packages/wonder-blocks-spacing/docs.md
```diff
@@ -0,0 +1,29 @@
...
+            }}
+        >
+            <div style={{width: 250, paddingRight: 10, textAlign: "right"}}>
+                {spaceName}: {Spacing[spaceName]}px
+            </div>
+            <div
+                style={{
+                    backgroundColor: "black",
+                    width: Spacing[spaceName],
+                    height: Spacing[spaceName],
```
kevinbarabash: Can you update this to instead have two sets of divs?  One that highlights the use of `Spacing` for the `width` and a separate one for the `height`.  Ideally the `height` ones would be organized in a row, maybe using a table. The reason for this is I'd like people thinking about these values in terms of one dimension.

kevinbarabash: This is a nice to have.

packages/wonder-blocks-spacing/index.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+const Spacing = {
+    // Named
+    xxxs: 4,
+    xxs: 6,
+    xxs: 8,
+    s: 12,
+    m: 16,
+    l: 24,
```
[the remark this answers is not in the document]
kevinbarabash: This is blocking.

kevinbarabash:
Woohoo!  Thanks for updating the docs too!  :heart:

PR #68, opened 2018-05-02
Update View and Text to pass along unknown props

matchu:
In this change, we allow `View` and `Text` to accept arbitrary props to pass along to the wrapper HTML element. This enables event handlers like `onClick`, and accessibility features like `id` and `aria-hidden`.

# Test Plan:
In Styleguidist, look at the new example in the "Core" section.

Click the words "Click me!", and see an alert appear.

Use a screen reader (I used VoiceOver in Safari), and observe that the screen reader does not recognize the text "This text is hidden from screen readers.".

PR #69, opened 2018-05-02
[typography] Type components now accept an `id` prop

matchu:
In this change, we add an `id` prop to typography components. This enables us to identify these nodes to screen readers, for cases like `aria-labelledby`.

Depends on #68, which enables `Text` to accept arbitrary props.

# Test Plan:
Open the typography section in styleguidist. Confirm that all typography tags in the first example have their `id` set correctly.

### DIFF
--- a/packages/wonder-blocks-typography/__snapshots__/generated-snapshot.test.js.snap
+++ b/packages/wonder-blocks-typography/__snapshots__/generated-snapshot.test.js.snap
@@ -7,6 +7,7 @@ exports[`wonder-blocks-typography example 1 1`] = `
 >
   <h1
     className=""
+    id="example-Title"
     style={
       Object {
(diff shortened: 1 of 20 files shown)


jeresig:

Built with commit 6afbf395b8e6999e5971520be1b90f545f678c9a

https://deploy-preview-69--wonder-blocks.netlify.com

packages/wonder-blocks-typography/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -297,6 +313,7 @@ exports[`wonder-blocks-typography example 1 1`] = `
 exports[`wonder-blocks-typography example 2 1`] = `
 <h1
   className=""
+  id={undefined}
```
kevinbarabash: lol

PR #70, opened 2018-05-02
[modal] First draft of ARIA dialog behavior

matchu:
In this change, we add basic screen reader support for our modal. In particular, we give it the "dialog" ARIA role, mark its title+subtitle nodes as a label and description respectively, and we add very basic focus management.

I only added an ARIA label and description in the `StandardModal`, since `title` and `subtitle` are its props. For the `TwoColumnModal`, I only applied `role="dialog"`, since identifying the appropriate label is trickier. (We just receive a monolith of content as our props.)

Depends on #69, which enables typography components to accept `id`.

https://khanacademy.atlassian.net/browse/WB-138

# Test Plan:
Using VoiceOver in Safari, open the standard modal in the `ModalLauncher` example. Confirm that focus shifts to its "Close modal" button, and that VoiceOver reports that the user is currently in a dialog named "Title".

### DIFF
--- a/packages/wonder-blocks-modal/components/modal-backdrop.js
+++ b/packages/wonder-blocks-modal/components/modal-backdrop.js
@@ -1,5 +1,6 @@
 // @flow
 import * as React from "react";
+import * as ReactDOM from "react-dom";
 import {StyleSheet} from "aphrodite";
 
 import Color from "wonder-blocks-color";
@@ -35,6 +36,26 @@ export default class ModalBackdrop extends React.Component<Props> {
             this.props.onCloseModal();
         }
     };
+
+    componentDidMount() {
+        // Focus the last button in the modal, on the assumption that it'll be
+        // a sensible default action.
+        //
+        // TODO(mdr): Not sure how robust this is; or whether we'll sometimes
+        //     want the default to be something in the modal content, or a
+        //     different button, or something else.
+        const node: HTMLElement = (ReactDOM.findDOMNode(this): any);
+        if (!node) {
+            return;
+        }
+
+        const buttons = node.querySelectorAll("button");
+        const lastButton = buttons[buttons.length - 1];
+        if (!lastButton) {
+            return;
+        }
+        lastButton.focus();
+    }
 
     render() {
         const children = this.props.children;
--- a/packages/wonder-blocks-modal/components/standard-modal.js
+++ b/packages/wonder-blocks-modal/components/standard-modal.js
@@ -74,6 +74,10 @@ export default class StandardModal extends React.Component<Props> {
 
         // Generate a unique ID for this modal. Math.random() isn't great, but
         // collisions on the same page should be pretty unlikely.
+        //
+        // TODO(mdr): Once we add snapshot testing, will this break it? We might
+        //     need to figure out how to seed the PRNG, or mock Math.random, or
+        //     find a different approach altogether.
         const modalId = Math.random()
             .toString(36)
             .substr(2);
--- a/packages/wonder-blocks-modal/components/standard-modal.js
+++ b/packages/wonder-blocks-modal/components/standard-modal.js
@@ -11,7 +11,7 @@ import {
 } from "wonder-blocks-typography";
 
 import ModalCloseButton from "./modal-close-button.js";
-import {smOrSmaller} from "../util/util.js";
+import {smOrSmaller, generateUniqueId} from "../util/util.js";
 
 type Props = {
     /**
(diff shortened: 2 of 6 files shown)


jeresig:

Built with commit 2541744af0c57fc501f935244b4fd72c809ce9a4

https://deploy-preview-70--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/standard-modal.js
```diff
@@ -58,28 +58,73 @@ type Props = {
...
     static defaultProps = {
         onClickCloseButton: () => {},
     };
 
+    constructor(props: Props) {
+        super(props);
+
+        // Generate a unique ID for this modal. Math.random() isn't great, but
+        // collisions on the same page should be pretty unlikely.
+        const modalId = Math.random()
```
matchu: Interesting how this randomness doesn't break snapshot tests. I suppose they must seed the PRNG!

packages/wonder-blocks-modal/components/standard-modal.js
```diff
@@ -58,28 +58,73 @@ type Props = {
...
+
     _renderTitleAndSubtitle() {
         const {title, subtitle} = this.props;
 
         if (subtitle) {
             return (
                 <View>
-                    <HeadingSmall>{title}</HeadingSmall>
-                    <LabelSmall>{subtitle}</LabelSmall>
+                    {/* TODO(mdr): Should typography components be able to
```
matchu: Oh whoops, we no longer need these wrapper nodes! Fixing.

packages/wonder-blocks-modal/components/standard-modal.js
matchu: …oh, or maybe it's because we don't currently do snapshot testing for modals >_> Cross that bridge when we get there? :/

packages/wonder-blocks-modal/components/standard-modal.js
```diff
@@ -58,28 +58,62 @@ type Props = {
...
+        //     find a different approach altogether.
+        const modalId = Math.random()
+            .toString(36)
+            .substr(2);
+
+        // Use that unique ID number to generate unique ID strings for this
+        // modal's elements.
+        const idPrefix = `wonder-blocks-standard-modal-${modalId}`;
+        this._titleId = `${idPrefix}-title`;
+        this._subtitleId = `${idPrefix}-title`;
```
kevinbarabash: `-subtitle`

packages/wonder-blocks-modal/components/standard-modal.js
```diff
@@ -58,28 +58,62 @@ type Props = {
...
+
+        // Generate a unique ID for this modal. Math.random() isn't great, but
+        // collisions on the same page should be pretty unlikely.
+        //
+        // TODO(mdr): Once we add snapshot testing, will this break it? We might
+        //     need to figure out how to seed the PRNG, or mock Math.random, or
+        //     find a different approach altogether.
+        const modalId = Math.random()
+            .toString(36)
+            .substr(2);
```
kevinbarabash: Maybe we can extract this into a separate module and then do something similar to what we do for aphrodite styles with the `SNAPSHOT_INLINE_APHRODITE` environment variable in `package.json`.

packages/wonder-blocks-modal/components/standard-modal.js
```diff
@@ -58,28 +58,62 @@ type Props = {
...
         return (
-            <View style={styles.container}>
+            <View
+                style={styles.container}
+                role="dialog"
+                aria-labelledby={this._titleId}
+                // TODO(mdr): Is the subtitle a reliable "description" of the
+                //     modal? Or is it often used to convey more specific
+                //     information?
+                aria-describedby={this._subtitleId}
```
kevinbarabash: We could probably get away with just not having a `aria-describedby`.

packages/wonder-blocks-modal/components/standard-modal.js
kevinbarabash: The reason for suggesting a separate module, `wonder-blocks-utils` for example, is that we'll probably need to do the same thing in other modules/components as well.

kevinbarabash: I guess at the very least, put it in a separate file with a TODO to make it a separate component later.

packages/wonder-blocks-modal/components/modal-backdrop.js
```diff
@@ -36,6 +37,26 @@ export default class ModalBackdrop extends React.Component<Props> {
...
+        //
+        // TODO(mdr): Not sure how robust this is; or whether we'll sometimes
+        //     want the default to be something in the modal content, or a
+        //     different button, or something else.
+        const node: HTMLElement = (ReactDOM.findDOMNode(this): any);
+        if (!node) {
+            return;
+        }
+
+        const buttons = node.querySelectorAll("button");
```
kevinbarabash: This is a good start.  We'll want to grab all `input`, `a`, and `button` together and probably grab the first one inside of content and if that doesn't exist the last one inside the footer.  The footer behavior relies on no-one use `flexDirection: "row-reverse"`.

kevinbarabash: For focusing the button we may want to try to figure out a way to search for the "primary" `Button` in the footer.  `Button` can render though as `<button>` or `<a>` (or `<Link>` which I believe renders as `<a>` under the hood).

matchu: 👍🏻 I've got WB-151 written down, which talks about making this cleverer and more robust. Perhaps by reusing the new "focus first descendant" and "focus last descendant" functionality in #71?

PR #71, opened 2018-05-02
[modal] Trap focus in the modal, and wrap tab button presses

matchu:
In this change, we ensure that tab focus stays within the modal. If the user attempts to tab beyond the modal's boundaries, their tab focus wraps back around.

The general approach was adapted from the WAI-ARIA dialog behavior example, in which they implemented similar behavior for their modal.

Unfortunately, I couldn't get an automated test working :( I left my WIP version of the test in the suite as a `test.skip`, and performed testing manually.

This new login-form example showcases that maybe we want cleverer auto-focus behavior than #70 gives us. (We auto-focus the "log in" button, but we should probably auto-focus the username field.) Will try something cleverer in a later change!

Depends on #70, which added basic auto-focus behavior for ARIA's sake.

https://khanacademy.atlassian.net/browse/WB-132

# Test Plan:
In Styleguidist, launch the `TwoColumnModal`, which now has a login form in the right-hand column.

Observe that, as in #70, focus is automatically moved inside the modal.

Then, use Tab and Shift-Tab to rotate through the form fields, form buttons, and modal close button. Observe that they are a cycle: pressing Tab when "Log in" is focused takes you to the modal close button, and pressing Shift-Tab when the close button is focused takes you to "Log in", despite them being at start and end of the modal.

### DIFF
--- a/packages/wonder-blocks-modal/components/modal-backdrop.js
+++ b/packages/wonder-blocks-modal/components/modal-backdrop.js
@@ -71,13 +71,15 @@ export default class ModalBackdrop extends React.Component<Props> {
 
         return (
             <View style={styles.modalPositioner} onClick={this._handleClick}>
-                {/* To ensure that tab wrapping happens inside the modal area,
-                  * even if the modal is mounted at the start/end of the
-                  * document (end is likely in practice!), we add focusable
-                  * sentinel nodes. That way, tabbing out of the modal is more
-                  * likely to hit a sentinel node. They're not critical to
-                  * focus wrapping, though; we're resilient to any kind of focus
-                  * shift, whether they hit the sentinels or not! */}
+                {/* When you press Tab on the last focusable node of the
+                  * document, some browsers will move your tab focus outside of
+                  * the document. But we want to capture that as a focus event,
+                  * and move focus back into the modal! So, we add focusable
+                  * sentinel nodes. That way, tabbing out of the modal should
+                  * take you to a sentinel node, rather than taking you out of
+                  * the document. These sentinels aren't critical to focus
+                  * wrapping, though; we're resilient to any kind of focus
+                  * shift, whether it's to the sentinels or somewhere else! */}
                 <div tabIndex="0" />
                 <FocusTrap>{clonedChildren}</FocusTrap>
                 <div tabIndex="0" />
--- a/packages/wonder-blocks-modal/components/modal-backdrop.js
+++ b/packages/wonder-blocks-modal/components/modal-backdrop.js
@@ -70,9 +70,108 @@ export default class ModalBackdrop extends React.Component<Props> {
 
         return (
             <View style={styles.modalPositioner} onClick={this._handleClick}>
-                {clonedChildren}
+                {/* To ensure that tab wrapping happens inside the modal area,
+                  * even if the modal is mounted at the start/end of the
+                  * document (end is likely in practice!), we add focusable
+                  * sentinel nodes. That way, tabbing out of the modal is more
+                  * likely to hit a sentinel node. They're not critical to
+                  * focus wrapping, though; we're resilient to any kind of focus
+                  * shift, whether they hit the sentinels or not! */}
+                <div tabIndex="0" />
+                <FocusTrap>{clonedChildren}</FocusTrap>
+                <div tabIndex="0" />
             </View>
         );
+    }
+}
+
+/**
+ * This component ensures that focus stays within itself. If the user uses Tab
+ * at the end of the modal, or Shift-Tab at the start of the modal, then this
+ * component wraps focus to the start/end respectively.
+ *
+ * Adapted from the WAI-ARIA dialog behavior example.
+ * https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/dialog-modal/dialog.html
+ */
+class FocusTrap extends React.Component<{children: React.Node}> {
+    _lastNodeFocusedInRoot: ?Node = null;
+    _ignoreFocusChanges: boolean = false;
+
+    componentDidMount() {
+        window.addEventListener("focus", this._handleGlobalFocus, true);
+    }
+
+    componentWillUnmount() {
+        window.removeEventListener("focus", this._handleGlobalFocus, true);
+    }
+
+    _getRootNode(): Node {
+        return (ReactDOM.findDOMNode(this): any);
+    }
+
+    _tryToFocus(node: Node) {
+        if (typeof node.focus !== "function") {
+            return;
+        }
+
+        this._ignoreFocusChanges = true;
+        try {
+            node.focus();
+        } catch (e) {}
+        this._ignoreFocusChanges = false;
+
+        return document.activeElement === node;
+    }
+
+    _focusFirstElementIn(currentParent: Node) {
+        const children = currentParent.childNodes;
+        for (let i = 0; i < children.length; i++) {
+            const child = children[i];
+            if (this._tryToFocus(child) || this._focusFirstElementIn(child)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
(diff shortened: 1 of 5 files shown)


jeresig:

Built with commit a720675cc415782387bc21c54c6fe0bb09f43d96

https://deploy-preview-71--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-backdrop.js
```diff
@@ -70,12 +70,147 @@ export default class ModalBackdrop extends React.Component<Props> {
...
+                  * shift, whether they hit the sentinels or not! */}
+                <div tabIndex="0" />
+                <FocusTrap>{clonedChildren}</FocusTrap>
+                <div tabIndex="0" />
             </View>
         );
     }
 }
 
+/**
```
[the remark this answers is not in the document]
jeresig: Is this component copied from webapp? If so it might be good to state as such.

packages/wonder-blocks-modal/components/modal-backdrop.js
```diff
@@ -70,12 +70,147 @@ export default class ModalBackdrop extends React.Component<Props> {
...
+        return false;
+    }
+
+    /**
+     * Focus the last focusable descendant of the given node.
+     *
+     * Return true if we succeed. Or, if the given node has no focusable
+     * descendants, return false.
+     */
+    _focusLastElementIn(currentParent: Node) {
```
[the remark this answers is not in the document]
matchu: Hehe yeah, I was gonna try something like that! Might pull this focus behavior into util functions so that we can go even more granular—something like "focus the first element in the content area, or, if unsuccessful, focus the last element in the footer area". Fingers crossed that it's just easy!

packages/wonder-blocks-modal/components/modal-backdrop.js
```diff
@@ -70,12 +70,147 @@ export default class ModalBackdrop extends React.Component<Props> {
...
+        window.addEventListener("focus", this._handleGlobalFocus, true);
+    }
+
+    componentWillUnmount() {
+        window.removeEventListener("focus", this._handleGlobalFocus, true);
+    }
+
+    /** Get the outermost DOM node of our children. */
+    _getModalRoot(): Node {
+        return (ReactDOM.findDOMNode(this): any);
```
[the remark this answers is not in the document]
matchu: I think it was mainly because it could be null, but I'll check that. If that was it, then probably throwing on null is better.

packages/wonder-blocks-modal/components/modal-backdrop.js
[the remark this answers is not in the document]
matchu: (If so, though, I'll do that extraction during WB-151, rather than this change 👍🏻)

matchu:
In this change, we ensure that tab focus stays within the modal. If the user attempts to tab beyond the modal's boundaries, their tab focus wraps back around.

The general approach was adapted from the WAI-ARIA dialog behavior example, in which they implemented similar behavior for their modal.

Unfortunately, I couldn't get an automated test working :( I left my WIP version of the test in the suite as a `test.skip`, and performed testing manually.

This new login-form example showcases that maybe we want cleverer auto-focus behavior than #70 gives us. (We auto-focus the "log in" button, but we should probably auto-focus the username field.) Will try something cleverer in a later change!

Depends on #70, which added basic auto-focus behavior for ARIA's sake.

https://khanacademy.atlassian.net/browse/WB-132

# Test Plan:
In Styleguidist, launch the `TwoColumnModal`, which now has a login form in the right-hand column.

Observe that, as in #70, focus is automatically moved inside the modal.

Then, use Tab and Shift-Tab to rotate through the form fields, form buttons, and modal close button. Observe that they are a cycle: pressing Tab when "Log in" is focused takes you to the modal close button, and pressing Shift-Tab when the close button is focused takes you to "Log in", despite them being at start and end of the modal.

packages/wonder-blocks-modal/components/modal-backdrop.js
```diff
@@ -70,7 +71,18 @@ export default class ModalBackdrop extends React.Component<Props> {
...
+                  * document, some browsers will move your tab focus outside of
+                  * the document. But we want to capture that as a focus event,
+                  * and move focus back into the modal! So, we add focusable
+                  * sentinel nodes. That way, tabbing out of the modal should
+                  * take you to a sentinel node, rather than taking you out of
+                  * the document. These sentinels aren't critical to focus
+                  * wrapping, though; we're resilient to any kind of focus
+                  * shift, whether it's to the sentinels or somewhere else! */}
+                <div tabIndex="0" />
+                <FocusTrap>{clonedChildren}</FocusTrap>
```
kevinbarabash: I like that you pulled this out into its own component.  Eventually we'll need left/right caret buttons that appear on the backdrop.  We'll want to make those tabbable too (they currently aren't in webapp).  It shouldn't be too hard to modify this code to allow for that.

packages/wonder-blocks-modal/components/modal-backdrop.test.js
```diff
@@ -177,8 +177,57 @@ describe("ModalBackdrop", () => {
...
-        expect(focusedElement).toBeTruthy();
-        expect(focusedElement.hasAttribute("data-last-button")).toBe(true);
+        const lastButton = wrapper.find("[data-last-button]").getDOMNode();
+        expect(document.activeElement).toBe(lastButton);
+    });
+
+    // TODO(mdr): I haven't figured out how to actually simulate tab keystrokes
+    //     or focus events in a way that JSDOM will recognize, so triggering the
+    //     global focus handler isn't feasible. I had to do manual testing
+    //     instead :( Here's what I had, though!
```
kevinbarabash: According to https://github.com/jsdom/jsdom/issues/1634 some people have been able to trigger keyboard events within JSDOM.  I haven't tried it myself though.

kevinbarabash: This test seems to be sufficient though to test your code.

matchu: I tried simulating Tab keydown events in similar ways, but it didn't seem to trigger a focus change. It didn't seem to work in Firefox either, so it might be that simulated Tab presses aren't respected? Also might just mean my code was wrong 😬

The test as written doesn't work, even though it seems like it should. First off, I'm not sure that focus handlers would normally trigger synchronously when you call `focus()`. (Maybe?) But, either way, I found that `a.focus()` didn't seem to trigger the global focus event handler, even if I waited a couple seconds. I suspect that the global focus handler doesn't work in JSDOM?

It's also possible that Enzyme doesn't actually mount the component in the global `document`, but instead in its own sandboxed document. So, the listeners we register onto the global document aren't triggered. I hadn't considered that until now!

matchu: Just spent some time exploring that last theory; no luck. I tried updating `FocusTrap` to register the event on `this._getModalRoot().ownerDocument`, and I wrote a `simulateTabPress` function that triggered the Tab keydown on the button itself, in the correct owner document. I also updated the test to wait a second to see if the focus event happened. Nothing :(

```
function simulateTabPress(target) {
    const event: KeyboardEvent = (target.ownerDocument.createEvent(
        "Event",
    ): any);
    event.key = "Tab";
    event.initEvent("keydown", true, true);
    target.dispatchEvent(event);
}
```

joshwcomeau:
I saw an article about how focus-trapping within a modal can be achieved using `focus-within`, where only a single line of JS is needed: https://css-tricks.com/a-css-approach-to-trap-focus-inside-of-an-element/

No idea if it's suitable, only skimmed the article, but I think it would greatly simplify this logic if it works :)

cc @matchu

kevinbarabash:
Unfortunately `focus-within` isn't supported on Edge yet.  https://caniuse.com/#feat=css-focus-within

PR #72, opened 2018-05-04
[modal] Whoops, override `minHeight` in mobile view

matchu:
I mentioned this in code review for #64, but forgot to change it! In addition to resetting our `maxWidth` and `maxHeight` rules, we also need to reset our `minHeight`. Otherwise, the modal will overflow the screen when it's <200px tall.

Test Plan:
Launch a `StandardModal`, then manually resize the window to be quite narrow and quite short. Confirm that the modal continues to be full-screen, without overflowing.
image


jeresig:

Built with commit 869180a34680aa189c7ed694b764e549ad4c1620

https://deploy-preview-72--wonder-blocks.netlify.com

PR #73, opened 2018-05-04
ensure all relative imports have a file extension

kevinbarabash:
Webapp doesn't understand files without extensions so we need to ensure that all local requires have one.

### DIFF
--- a/packages/wonder-blocks-typography/index.js
+++ b/packages/wonder-blocks-typography/index.js
@@ -1,22 +1,22 @@
 // @flow
 import styles from "./util/styles.js";
 
-import Title from "./components/title";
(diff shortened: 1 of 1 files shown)


jeresig:

Built with commit db69d3c39888a5707e0590c5e74c16a2ee24de56

https://deploy-preview-73--wonder-blocks.netlify.com

kevinbarabash:
> Thanks! I wonder if we should have a linter for this...

It looks like that shouldn't be too hard to do, but we need to set up eslint first.

https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/extensions.md

PR #74, opened 2018-05-04
ensure index.js exist in wonder-blocks-modal package

kevinbarabash:
The update-wonder-blocks-shims.js script assumes that package.json's "main" entry exists.


packages/wonder-blocks-modal/index.js
```diff
@@ -0,0 +1,6 @@
+// @flow
+import ModalLauncher from "./components/modal-launcher.js";
```
jeresig: You could also rewrite each of these lines as:

`export {default as ModalLuauncher} from "./components/modal-launcher.js";`

kevinbarabash: I'll do a separate PR to do that across all packages.

jeresig:

Built with commit 239178e45da4c04405effa32ee990b77dd1b0fe4

https://deploy-preview-74--wonder-blocks.netlify.com

PR #75, opened 2018-05-04
[Button v1] Create ClickableBehavior component

chenyejia:
In this change, we create a ClickableBehavior component that handles interactions and passes hovered, focused, pressed state to ButtonCore. ClickableBehavior should also be used for future Link and IconButton components. We also update ButtonCore to use anchor tags if href is passed in.

Not yet implemented:
* Routing
* Custom style

Test plan:
* Interact with Button on Styleguidist, via mouse, touch, and keyboard


packages/wonder-blocks-button/components/clickable-behavior.js
```diff
@@ -0,0 +1,190 @@
...
+
+    handleMouseEnter = () => {
+        if (!this.props.disabled && !this.waitingForClick) {
+            this.setState({hovered: true});
+        }
+    };
+
+    handleMouseLeave = () => {
+        if (!this.props.disabled && !this.waitingForClick) {
+            // TODO(yejia): Is this what we want for pressed?
```
chenyejia: Was drag behavior specced out? Not sure if I should be setting pressed to false if a user drags outside the button area!

jeresig:

Built with commit fda5c4b0a35af93905da6d1f6290ed43232ee957

https://deploy-preview-75--wonder-blocks.netlify.com

packages/wonder-blocks-button/components/button.js
```diff
@@ -118,18 +120,23 @@ export default class Button extends React.Component<Props> {
...
-            </div>
+            <ClickableBehavior
+                disabled={sharedProps.disabled}
+                onClick={onClick}
+                href={href}
+            >
+                {(state, handlers) => {
+                    return (
+                        <View {...handlers} style={{display: "inline-block"}}>
+                            <ButtonCore {...sharedProps} href={href} {...state}>
```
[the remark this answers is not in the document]
kevinbarabash: We should remove all wrapper elements between around what `Button`'s `render()` method is return and the `<a>`/`<button>` tag at the heart of this component.

packages/wonder-blocks-button/components/clickable-behavior.js
```diff
@@ -0,0 +1,190 @@
...
+};
+
+type Handlers = {
+    onClick: (e: SyntheticMouseEvent<>) => void,
+    onMouseEnter: () => void,
+    onMouseLeave: () => void,
+    onMouseDown: () => void,
+    onMouseUp: () => void,
+    onTouchStart: () => void,
+    onTouchEnd: () => void,
```
kevinbarabash: We should also handle `onTouchCancel`.

packages/wonder-blocks-button/components/clickable-behavior.js
```diff
@@ -0,0 +1,190 @@
...
+            }
+        }
+    };
+
+    handleKeyUp = (event: KeyboardEvent) => {
+        const keyCode = event.which || event.keyCode;
+        const {disabled, href} = this.props;
+        if (!disabled) {
+            if (keyCode === keyCodes.tab) {
+                this.setState({focused: true});
```
kevinbarabash: I wasn't able to get the buttons to focus with the "tab" key in Safari.  It works like a charm in Chrome though.

kevinbarabash: I wonder if this is even necessary since tabbing should focus a button by default.  All we should need to do is track if the button has become focused because of a keyboard event.

packages/wonder-blocks-button/components/clickable-behavior.js
```diff
@@ -0,0 +1,190 @@
...
+    handleKeyDown = (event: KeyboardEvent) => {
+        const keyCode = event.which || event.keyCode;
+        const {disabled, href} = this.props;
+        if (!disabled) {
+            if (keyCode === keyCodes.tab) {
+                this.setState({focused: false});
+            } else if (
+                href ? keyCode === keyCodes.enter : keyCode === keyCodes.space
+            ) {
+                this.setState({pressed: true});
```
kevinbarabash: This worked for the first button but not for the second button for me when I was testing in Chrome.

kevinbarabash: BTW, this is really cool!

packages/wonder-blocks-button/components/clickable-behavior.js
```diff
@@ -0,0 +1,190 @@
...
+ *     }
+ * }
+ *
+ * This follows a pattern call 'Function as Child Components':
+ * https://medium.com/merrickchristensen/function-as-child-components-5f3920a9ace9
+ */
+import React from "react";
+
+type Props = {
+    children: (state: State, handlers: Handlers) => React$Element<*>,
```
kevinbarabash: I like that this is general enough that if people do need to make other things clickable in a cross platform way they can.

packages/wonder-blocks-button/components/clickable-behavior.js
chenyejia: I think what's happening for the second one is that pressing enter immediately clicks the link, which takes you to the section linked and takes you out of focus of the button, so you're no longer "pressing" within the button. I think this is standard behavior for a link, but should the "click" happen onKeyUp?

packages/wonder-blocks-button/components/clickable-behavior.js
chenyejia: Hmmm I don't know how to do that :( I don't think I can just check the type of event passed in to onFocus. In general I don't know how to check type of SyntheticEvents, but also I think SyntheticFocusEvents aren't SyntheticKeyboardEvents. Also tabbing does work for me in Safari, so not sure how to check what's going on x_x

kevinbarabash: It looks like it isn't turned on by default in Safari.  After turning it on, it works as expected.  I'm not sure why they'd turn it off by default.  :(

I was going to suggest tracking if there was a `mousedown` or `touchstart` event on the button and if there wasn't then we could assume the focus was caused by the keyboard.  Totally unnecessary since the problem was a Safari setting.

packages/wonder-blocks-button/components/clickable-behavior.js
```diff
@@ -0,0 +1,190 @@
...
+ * class MyClickableComponent extends React.Component<Props> {
+ *     render() {
+ *         return <ClickableBehavior
+ *             disabled={this.props.disabled}
+ *             onClick={this.onClick}
+ *         >
+ *             {({hovered}, handlers) =>
+ *                 <RoundRect
+ *                      textcolor='white'
+ *                      backgroundColor={hovered ? 'red' : 'blue'}/>}
```
jeresig: Extra `/>`

packages/wonder-blocks-button/components/clickable-behavior.js
```diff
@@ -0,0 +1,190 @@
...
+ * an object containing event handlers. The 'children' function should return
+ * a clickable React Element of some sort.
+ *
+ * Example:
+ *
+ * class MyClickableComponent extends React.Component<Props> {
+ *     render() {
+ *         return <ClickableBehavior
+ *             disabled={this.props.disabled}
+ *             onClick={this.onClick}
```
jeresig: Might be slightly better to have this be `this.props.onClick` as there isn't an `onClick` method in this component.

packages/wonder-blocks-button/components/clickable-behavior.js
jeresig: This makes me wonder if we should make this component public (exposed in the docs and through wonder-blocks-button) so that people can use it in their code?

packages/wonder-blocks-button/components/clickable-behavior.js
```diff
@@ -0,0 +1,190 @@
...
+};
+
+const keyCodes = {
+    tab: 9,
+    enter: 13,
+    space: 32,
+};
+
+export default class ClickableBehavior extends React.Component<Props, State> {
+    constructor(props: Props) {
```
jeresig: We should specify a `defaultProps` that includes `disabled` to `false` so that you don't always have to specify it.

packages/wonder-blocks-button/components/clickable-behavior.js
jeresig: Hmm - I don't think it was! But I think what you have specified here is probably fine  for now.

packages/wonder-blocks-button/components/clickable-behavior.js
```diff
@@ -0,0 +1,190 @@
...
+    };
+
+    handleBlur = () => {
+        if (!this.props.disabled) {
+            this.setState({focused: false, pressed: false});
+        }
+    };
+
+    render() {
+        const handlers = {
```
jeresig: Every single one of these functions has `if (!this.props.disabled...)` in it. It seems like it would be easier to do something like: `const handlers = disabled ? {} : { onClick: ..., etc. };`. Or you could stub out empty functions for each of them, make it a constant at the top of the file, and just pass that in instead. That will make all your handler functions a lot easier to read!

packages/wonder-blocks-button/components/clickable-behavior.js
kevinbarabash: :+1: it should be option

packages/wonder-blocks-button/components/button.md
```diff
@@ -1,10 +1,18 @@
 Button example:
+```js
+<Button 
+    spinner={false}
+    testId={"1"}
+    onClick={(e) => console.log("Hello, world!")}
```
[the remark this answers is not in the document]
chenyejia: It should work :( Which browser/device are you using?

packages/wonder-blocks-button/components/clickable-behavior.js
```diff
@@ -0,0 +1,190 @@
+// @flow
+/**
+ * Add hover, focus, and active status updates to a clickable component.
+ *
+ * Via mouse:
+ * 1. Hover over button -> hover state
+ * 2. Mouse down -> active state
+ * 3. Mouse up -> default state
+ * 4. Press tab -> focus state
```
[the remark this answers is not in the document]
chenyejia: Synced with @kevinbarabash offline. I'm going to punt on this for now because it's a bit complicated and lower priority than other tasks

kevinbarabash:
I don't want to hold this up any longer, but some changes that I think we should look at making to the button examples later:
- Split out each of the `Button examples, style specified (Support width, position, margin, and flex styles):` in its own separate example.  This will make it easier for developers to find the code for the particular example they're interested in.
- Use `View` instead of `div`

PR #76, opened 2018-05-04
Setup eslint

kevinbarabash:
I've copied the eslintrc files from khan-linter and added a top level .eslintrc file that inherits from them.  This will allow us to add additional rules such as https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/extensions.md in the future.  The PR also fixes all existing lint of which there was not very much thankfully.

### DIFF
--- a/eslint/README.md
+++ b/eslint/README.md
@@ -1,0 +1,3 @@
+The files in this directory were copied from https://github.com/Khan/khan-linter/tree/60889f7a1c86e9eb7f3154184a89504a3d2163f8.
+
+To update, copy over the files and update github link in this file.
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "index.js",
   "scripts": {
     "install": "for d in packages/*/ ; do (cd \"$d\" && npm install); done",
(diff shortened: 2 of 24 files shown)


jeresig:

Built with commit 80f906d2d9f20dee669ba1643f038d574755cf15

https://deploy-preview-76--wonder-blocks.netlify.com

package.json
```diff
@@ -56,6 +65,7 @@
     }
   },
   "pre-commit": [
+    "lint",
```
jeresig: We should add `lint` as part of test - that way it'll run in Travis CI, as well (and then remove it from the pre-commit list).

eslint/README.md
```diff
@@ -0,0 +1,3 @@
+The files in this directory were copied from https://github.com/Khan/khan-linter/tree/60889f7a1c86e9eb7f3154184a89504a3d2163f8.
```
jeresig: And they're all un-modified, right? We may want to explicitly state that they shouldn't be modified and if we want to add a rule we should modify `.eslintrc`, instead.

PR #77, opened 2018-05-05
add prepare script to sub-packages to build them so they can be required

kevinbarabash:
Up till now our wonder-blocks-* packages can't actually be required in other projects.  We've been relying on webapp's build system to compile our .js files.  We'd like to be able to require our modules in other projects such as khanalytics or a demo project that I'd like to include within this monorepo.  The demo project I've been working on use create-react-app which only compiles the code on demo/src which means we have to compile our packages/modules.


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "React components for Wonder Blocks design system.",
   "main": "index.js",
   "scripts": {
-    "install": "for d in packages/*/ ; do (cd \"$d\" && npm install); done",
+    "postinstall": "for d in packages/*/ ; do (cd \"$d\" && npm install); done",
```
kevinbarabash: I changed this to `postinstall` since the `prepare` scripts in each of the sub packages depends on some of the build tools that are installed at the base level.

jeresig:

Built with commit 0688e642ff04e8b75b7de811a1571db28b0951ca

https://deploy-preview-77--wonder-blocks.netlify.com

gen-webpack-config.js
```diff
@@ -0,0 +1,40 @@
...
+    return {
+        // TODO(kevinb): build both dev and prod
+        mode: "development",
+        entry: "./index.js",
+        output: {
+            libraryTarget: "commonjs2",
+            filename: "dist/index.js",
+            path: subPkgRoot,
+        },
+        externals: [...rootPkgDeps, ...subPkgDeps],
```
kevinbarabash: The reason why we have to generate a separate webpack config for each sub-package is that we to exclude different dependencies from inclusion in `dist/index.js` for each module.

kevinbarabash:
The downside of doing this is now whenever people switch branches in webapp we'll end up building everything.  The reason for this is that `wonder-blocks` is a submodule.  I think we should look at making `wonder-blocks` a dependency in `third_party/package.json` and instead point all of the `javascript/node_modules/wonder-blocks-*/index.js` to `third_party/node_modules/wonder-blocks-*/dist/index.js`.   I was planning to move other things out of `third_party/javascript-khansrc` and into `third_party/package.json` anyways so that we'd have a consistent way of doing deps between `kake` and the webpack build.

kevinbarabash:
Now that this is working I'm going to turn my attention to some dev UX things.  In particular, I'd like to be able to run a single `webpack` command and put it in "watch" mode so that it can continually build all of our components.  Right now you can build them all by running `npm i` in the root folder.  This runs `npm prepare` in each sub package which isn't really ideal.

packages/wonder-blocks-button/package.json
```diff
@@ -2,9 +2,11 @@
   "name": "wonder-blocks-button",
   "version": "0.0.1",
   "description": "",
-  "main": "index.js",
+  "main": "dist/index.js",
+  "source": "index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "prepare": "../../node_modules/.bin/webpack"
```
kevinbarabash: The problem with this is that it means that we'll end up re-compling all of our packages whenever someone changes the branch in webapp, b/c webapp is set up to run `npm install` in `third_party/javascript-khansrc/wonder-blocks`.  The reason we do this is to set up symlinks in node_modules just in case someone wants to make changes to wonder-blocks directly within webapp and test their changes in webapp.

To avoid recompiling more often than needed, we should instead be installing `wonder-blocks` as a node modules into third_party/node_modules and rely on the third_party/package.json to track the correct version.  This means that we'll only recompile wonder-blocks-* when switching branches if the version number differs between the two branches.

packages/wonder-blocks-button/package.json
```diff
@@ -2,9 +2,11 @@
   "name": "wonder-blocks-button",
   "version": "0.0.1",
   "description": "",
-  "main": "index.js",
+  "main": "dist/index.js",
+  "source": "index.js",
```
kevinbarabash: This is a hack so that webapp can compile wonder-blocks-button (et al) from source.  We'll need to update dev/tools/update-wonder-blocks-shims.js in webapp to use "source" instead of "main" since webapp compiles wonder-blocks for us.

packages/wonder-blocks-button/package.json
kevinbarabash: I've fixed this by changing the root package.json's postinstall script to run `npm install --ignore-scripts` in each sub package which prevents `prepare` from running.

demo/src/App.js
```diff
@@ -0,0 +1,26 @@
...
+    render() {
+        return (
+            <View style={styles.pinkSquare}>
+                <Button>Hello, world!</Button>
+            </View>
+        );
+    }
+}
+
+const styles = {
```
jeresig: 🤷‍♂️ Seems like this should be wrapped in `StyleSheet.create`

demo/src/App.js
```diff
@@ -0,0 +1,26 @@
+import React, {Component} from "react";
+import logo from "./logo.svg";
```
jeresig: 🛑 Not used? Also it doesn't appear to be added to the PR?

demo/src/App.js
```diff
@@ -0,0 +1,26 @@
+import React, {Component} from "react";
```
jeresig: 🛑 Should be `import * as React`

demo/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "React App",
+  "name": "Create React App Sample",
```
jeresig: 🤷‍♂️ It might be good to update these.

demo/src/App.js
jeresig: 🛑 Are our linters working against these files? It doesn't seem like they are, if this slipped through. We should make sure they are!

demo/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "name": "demo",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "aphrodite": "^2.2.0",
+    "react": "^16.3.2",
+    "react-dom": "^16.3.2",
+    "react-scripts": "1.1.4",
+    "wonder-blocks-button": "../packages/wonder-blocks-button",
+    "wonder-blocks-core": "../packages/wonder-blocks-core"
```
jeresig: 🤷‍♂️ It'd probably be good to import all of the packages that we provide, making them automatically available for all to use.

(Sidebar: I think we may want to have a script that auto-creates a new WB package and handles details like this.)

demo/src/registerServiceWorker.js
```diff
@@ -0,0 +1,120 @@
+// In production, we register a service worker to serve assets from local cache.
```
jeresig: 🛑 I assume this is third-party code? We should move this into a third-party directory of sorts and explain where it came from.

gen-webpack-config.js
```diff
@@ -0,0 +1,38 @@
+const path = require("path");
```
jeresig: 🛑 There should probably be a comment in this file explaining how this is used and where.

🛑 Also, it should probably be moved to the `utils/` directory with the other gen script.

demo/src/App.js
kevinbarabash: `create-react-app` has it's own style... updating to match ours. :+1:

demo/package.json
kevinbarabash: I'm not sure how to make them automatically available.  I'm going to add all of them manually right now and will create a script for adding new packages in the future that updates this.

demo/src/registerServiceWorker.js
kevinbarabash: I'm adding a warning message.  Unfortunately, the way create-react-app sets up a project, all files must be inside `src` (or some sub-directory).  I'd rather create sub-directories for different demo "pages" and keep the number of files in `src` proper relatively small.

jeresig: Ok! Maybe just inside `src/third-party`? That way it's clearer that it's not something that we created (and maybe use the README convention to explain where we got it from and how to update it).

demo/package.json
jeresig: Manually is fine for now!

demo/src/App.js
jeresig: Ah tricky, ok!

jeresig: 🛑 Oh - and we should have `// @flow` on this file, too!

demo/src/registerServiceWorker.js
kevinbarabash: I'll add a README.md with how it go created, but I don't actually know how to update it.  I'm not sure create-react-app has a good update story. :\

demo/src/App.js
```diff
@@ -0,0 +1,49 @@
...
+            <View style={styles.pinkSquare}>
+                <View style={styles.content}>
+                    <Title>Hello, world!</Title>
+                </View>
+                <View style={styles.row}>
+                    <ModalLauncher modal={modal}>
+                        {({openModal}) => (
+                            <Button onClick={openModal}>Open modal</Button>
+                        )}
+                    </ModalLauncher>
```
kevinbarabash: Closing the modal for some reason breaks.  I'll look into it in a separate PR.  Here's the stack trace:
```
bundle.js:23601 Uncaught TypeError: this.updater.enqueueCallback is not a function
    at ModalLauncher.../node_modules/react/lib/ReactBaseClasses.js.ReactComponent.setState (bundle.js:23601)
    at Object.ModalLauncher._this._closeModal [as onCloseModal] (bundle.js:31428)
    at ModalBackdrop._this._handleClick (bundle.js:31142)
    at HTMLUnknownElement.boundFunc (bundle.js:15476)
    at Object.../node_modules/react-dom/lib/ReactErrorUtils.js.ReactErrorUtils.invokeGuardedCallback (bundle.js:15482)
    at executeDispatch (bundle.js:9155)
    at Object.executeDispatchesInOrder (bundle.js:9178)
    at executeDispatchesAndRelease (bundle.js:8568)
    at executeDispatchesAndReleaseTopLevel (bundle.js:8579)
    at Array.forEach (<anonymous>)
    at forEachAccumulated (bundle.js:21298)
    at Object.processEventQueue (bundle.js:8779)
    at runEventQueueInBatch (bundle.js:15515)
    at Object.handleTopLevel [as _handleTopLevel] (bundle.js:15525)
    at handleTopLevelImpl (bundle.js:15611)
    at ReactDefaultBatchingStrategyTransaction.perform (bundle.js:20514)
    at Object.batchedUpdates (bundle.js:15233)
    at Object.batchedUpdates (bundle.js:18482)
    at dispatchEvent (bundle.js:15686)
```

kevinbarabash:
@jeresig this is ready for review again.

demo/src/App.js
```diff
@@ -0,0 +1,49 @@
+import * as React from "react";
```
jeresig: Missing `// @flow` and comment at top.

packages/wonder-blocks-modal/index.js
```diff
@@ -2,5 +2,6 @@
 import ModalLauncher from "./components/modal-launcher.js";
 import StandardModal from "./components/standard-modal.js";
 import TwoColumnModal from "./components/two-column-modal.js";
+import OneColumnModal from "./components/one-column-modal.js";
```
jeresig: Whoops, good catch!

packages/wonder-blocks-button/package.json
MiguelCastillo: is there a way to tell webapp to not compile and instead use the already compiled artifact?  I know im missing some details why we wouldnt do that...  do we recompile all third party code?

MiguelCastillo: meh - ignore. I see the problem.  Let me think thru this for a minute. BRB

kevinbarabash: I talked with @jeresig about that and we decided to maintain webapp's ability to use the source directly for the time being.  If we haven't used that ability in 6 months time then we'll re-evaluate.

PR #78, opened 2018-05-08
[Button v1] Allow custom styling for Buttons

chenyejia:
Allow width, position, margin, and flex styles to be passed in to Buttons. Styles that are not defined in ButtonCore may also be passed in, though we don't explicitly support this, and will be removing this once we strictly type the styles that are passed in.

Test Plan:
View on styleguidist

### DIFF
--- a/packages/wonder-blocks-button/components/button-core.js
+++ b/packages/wonder-blocks-button/components/button-core.js
@@ -1,11 +1,12 @@
 // @flow
 import React from "react";
-import {StyleSheet, css} from "aphrodite";
+import {StyleSheet} from "aphrodite";
 
 import {LabelLarge} from "wonder-blocks-typography";
 import Color, {mix, fade} from "wonder-blocks-color";
 import type {SharedProps} from "./button.js";
 import type {Handlers} from "./clickable-behavior.js";
+import {processStyleList} from "../../wonder-blocks-core/util/util.js";
 
 type Props = SharedProps &
     Handlers & {
@@ -19,7 +20,7 @@ export default class ButtonCore extends React.Component<Props> {
     render() {
         const {
             children,
-            icon,
+            icon, // eslint-disable-line no-unused-vars
             color,
             kind,
             light,
@@ -36,19 +37,25 @@ export default class ButtonCore extends React.Component<Props> {
 
         const buttonStyles = _generateStyles(color, kind, light);
         const Tag = href ? "a" : "button";
+
+        const styleAttributes = processStyleList([
+            sharedStyles.shared,
+            disabled && sharedStyles.disabled,
+            buttonStyles.default,
+            disabled && buttonStyles.disabled,
+            !disabled &&
+                (pressed
+                    ? buttonStyles.active
+                    : (hovered || focused) && buttonStyles.focus),
+            size === "small" && sharedStyles.small,
+            overridableStyles,
+            style,
+        ]);
+
         return (
             <Tag
-                className={css(
-                    sharedStyles.shared,
-                    disabled && sharedStyles.disabled,
-                    buttonStyles.default,
-                    disabled && buttonStyles.disabled,
-                    !disabled &&
-                        (pressed
-                            ? buttonStyles.active
-                            : (hovered || focused) && buttonStyles.focus),
-                    size === "small" && sharedStyles.small,
-                )}
+                style={styleAttributes.style}
+                className={styleAttributes.className}
                 href={href}
                 disabled={disabled}
                 data-test-id={testId}
@@ -60,11 +67,15 @@ export default class ButtonCore extends React.Component<Props> {
     }
 }
 
+const overridableStyles = {
+    position: "relative",
+    display: "inline-flex",
+    alignItems: "center",
+    justifyContent: "center",
+};
+
 const sharedStyles = StyleSheet.create({
     shared: {
-        position: "relative",
-        display: "inline-flex",
-        alignItems: "center",
         height: 40,
         paddingLeft: 16,
         paddingRight: 16,
--- a/packages/wonder-blocks-button/components/button.md
+++ b/packages/wonder-blocks-button/components/button.md
@@ -20,3 +20,127 @@ Disabled button example:
     disabled={true}
 >Label</Button>
 ```
+
+Button examples, `style` specified (Support width, position, margin, and flex styles):
+```js
+const {View} = require("wonder-blocks-core");
+<table>
+    <thead>
+        <tr>
+            <th style={{minWidth: '250px'}}>Styles</th>
+            <th style={{width: '100%'}}>Buttons</th>
+        </tr>
+    </thead>
+    <tbody>
+    <tr>
+        <td>width: '200px'</td>
+        <td>
+            <Button
+                onClick={(e) => console.log("Hello, world!")}
+                style={{width: '200px'}}
+            >Label</Button>
+        </td>
+    </tr>
+    <tr>
+        <td>width: '75%'</td>
+        <td>
+            <Button
+                onClick={(e) => console.log("Hello, world!")}
+                style={{width: '75%'}}
+            >Label</Button>
+        </td>
+    </tr>
+    <tr>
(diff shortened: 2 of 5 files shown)


jeresig:

Built with commit fda0575b71d521b4ebec2323a97477dbfbef4e69

https://deploy-preview-78--wonder-blocks.netlify.com

kevinbarabash:
I was really confused as to why a review is required, but it's b/c the target branch isn't `master`.  I wish there was a way to require reviews on all PRs instead of just protected branches.

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -60,11 +67,15 @@ export default class ButtonCore extends React.Component<Props> {
     }
 }
 
+const overridableStyles = {
```
kevinbarabash: Why pull these out of `sharedStyles.shared`?

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -36,19 +37,25 @@ export default class ButtonCore extends React.Component<Props> {
...
+        const styleAttributes = processStyleList([
+            sharedStyles.shared,
+            disabled && sharedStyles.disabled,
+            buttonStyles.default,
+            disabled && buttonStyles.disabled,
+            !disabled &&
+                (pressed
+                    ? buttonStyles.active
+                    : (hovered || focused) && buttonStyles.focus),
+            size === "small" && sharedStyles.small,
```
kevinbarabash: One example should include the `size="small"`.

packages/wonder-blocks-button/components/button.md
```diff
@@ -20,3 +20,127 @@ Disabled button example:
...
+                >Label</Button>
+                <div
+                    onClick={(e) => console.log("Hello, world!")}
+                    style={
+                        {
+                            background: '#d92916',
+                            textAlign: 'center',
+                            lineHeight: '100px',
+                            height: '100px',
+                            borderRadius: '4px',
```
kevinbarabash: We shouldn't be advertising the ability to change the `height` or `borderRadius` as a feature since we'll be disallowing when we get around to statically type checking `style`.

packages/wonder-blocks-button/components/button.md
```diff
@@ -20,3 +20,127 @@ Disabled button example:
...
+                </div>
+            </div>
+        </td>
+    </tr>
+    <tr>
+        <td>background: 'green'</td>
+        <td>
+            <Button
+                onClick={(e) => console.log("Hello, world!")}
+                style={{background: 'green'}}
```
kevinbarabash: We want people to use the `color` prop.  Please remove this example.

packages/wonder-blocks-button/components/button.md
```diff
@@ -20,3 +20,127 @@ Disabled button example:
...
+        <td>flexShrink: 2, width: '300px'</td>
+        <td>
+            <div style={{display: 'flex'}}>
+                <Button
+                    onClick={(e) => console.log("Hello, world!")}
+                    style={{flexShrink: 2, width: '300px'}}
+                >Label</Button>
+                <div
+                    onClick={(e) => console.log("Hello, world!")}
+                    style={
```
kevinbarabash: Instead of having a clickable "wide div", make it a button instead.  Since these two buttons are right beside each could you add some margin to one of them.  There are a number of layouts where we have buttons side by side with gaps and having such an example will give people confidence that they'll be able to do the same in their layouts.

packages/wonder-blocks-button/components/button.md
```diff
@@ -20,3 +20,127 @@ Disabled button example:
...
+    <tr>
+        <td>background: 'green'</td>
+        <td>
+            <Button
+                onClick={(e) => console.log("Hello, world!")}
+                style={{background: 'green'}}
+            >Label</Button>
+        </td>
+    </tr>
+    </tbody>
```
kevinbarabash: Could you add an example that use all three kinds of buttons side by side with spaces in-between?

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -36,19 +37,25 @@ export default class ButtonCore extends React.Component<Props> {
 
         const buttonStyles = _generateStyles(color, kind, light);
         const Tag = href ? "a" : "button";
+
+        const styleAttributes = processStyleList([
```
kevinbarabash: I think we should create an HOC that we can use to wrap bare `<a>` and `<button>` tags within button-core.js and elsewhere so that the resulting component does this for us.  An HOC isn't a huge win over the current code, but in the future we'll want to wrap other things like `<input>` and maybe even `<svg>` etc.  Could you add a TODO?

packages/wonder-blocks-button/components/button.md
chenyejia: I added this to show that you can't change the color :o I can be more explicit about that? Or remove it! Either is cool

packages/wonder-blocks-button/components/button-core.js
chenyejia: If we leave them in, Aphrodite adds "!important" to them, which you can't override with passed in styles (I think). I could also leave them in the same object and pull them all out of Aphrodite styling, but then the button could accept any styling (until ADR 47?)

packages/wonder-blocks-button/components/button.md
chenyejia: You can't change the height or border radius of a Button, but I wanted this element to show positioning.

packages/wonder-blocks-button/components/button.md
kevinbarabash: This should actually work, but it doesn't b/c of a bug in `processStyles` that have to fix.

chenyejia: Ohhh ok got it. Will remove :)

packages/wonder-blocks-button/components/button-core.js
kevinbarabash: https://github.com/Khan/wonder-blocks/pull/79 addresses the issue.

jeresig: Question: do we actually want anyone to be able to override these styles? It seems like they would be quite important to leave intact! (Otherwise the layout of the button will get messed up)

kevinbarabash: The plan was to use static type checking to disallow styles that shouldn't be override so that we could override some style props while not others.

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -1,11 +1,12 @@
 // @flow
 import React from "react";
-import {StyleSheet, css} from "aphrodite";
+import {StyleSheet} from "aphrodite";
 
 import {LabelLarge} from "wonder-blocks-typography";
 import Color, {mix, fade} from "wonder-blocks-color";
 import type {SharedProps} from "./button.js";
 import type {Handlers} from "./clickable-behavior.js";
+import {processStyleList} from "../../wonder-blocks-core/util/util.js";
```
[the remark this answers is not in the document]
kevinbarabash: If we want to use `processStyleList` manually we should export it from `index.js` in `wonder-blocks-core` and then `import {processStyleList} from "wonder-blocks-core"`.

packages/wonder-blocks-button/components/button-core.js
kevinbarabash: I'd rather use the HOC than changing the tag on `View`.  `View` now has some default styles on which we'd probably want to remove for other `tags`.

packages/wonder-blocks-button/components/button-core.js
kevinbarabash: There's a small button in the docs for #75 so we're good.

packages/wonder-blocks-button/components/button.md
kevinbarabash: If this is just a wide div for layout purposes we can also just remove the `onClick` handler.

PR #79, opened 2018-05-08
important-ify inline styles

kevinbarabash:
`aphrodite` adds `!important` to all its styles by default so in order for inline styles to override `aphrodite` styles we need to make the inline styles `!important` as well.

### DIFF
--- a/packages/wonder-blocks-core/util/util.js
+++ b/packages/wonder-blocks-core/util/util.js
@@ -33,6 +33,8 @@ export function processStyleList<T: Object>(style?: StyleType<T>) {
     flatten(style).forEach((child: T) => {
         // Check for aphrodite internal property
         if ((child: any)._definition) {
+            // TODO(kevinb): figure out how to test this with
+            // process.env.SNAPSHOT_INLINE_APHRODITE=0
             if (
                 typeof process !== "undefined" &&
                 process.env.SNAPSHOT_INLINE_APHRODITE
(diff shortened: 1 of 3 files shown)


jeresig:

Built with commit 06b095ab7316ead3821e58c0a8c010bc246b2008

https://deploy-preview-79--wonder-blocks.netlify.com

xymostech:
Are you going to be applying this to a React element using the `style` prop? AFAIK, React doesn't support setting `!important` in inline styles. https://github.com/facebook/react/issues/1881

kevinbarabash:
@xymostech thanks for raising this.  I didn't realize that it had to be set using `style.setProperty`.  That's super uncool.  Since that's the case we probably shouldn't bother.  It matches the current state we're in right now with inline styles.

kevinbarabash:
One thing that might actually be helpful is being able to detect whether there is a property on an inline style that won't get applied b/c there's an aphrodite style that also uses that property.  The hard part would dealing with conflicts between `padding` and `padding-left`.

jeresig:
@kevinbarabash it sounds like you might be abandoning this? But yeah, I think we probably want to generate warnings/errors instead. (Could we get away with disabling the important in Aphrodite? I know we still have some rules in our sheets but maybe those are less of an issue now?)

kevinbarabash:
Abandoning.  I think generating errors could be interesting, but if we can drop `!important` that'd be great.  I don't have a good enough sense for the interactions between our .less and aphrodite styles.  If all of our .less styles require a class or id and we can guarantee that we don't have any components using aphrodite styles that are descendants of elements with one of those classes or ids then we should be safe.  That's a lot of `if`s and I don't know how to verify that they're all true.  :(

PR #81, opened 2018-05-09
Re-implement modals to use low-level building blocks.

jeresig:
Now that we're free from our Grid restrictions it's really easy to build up a series of low-level building blocks to construct our modals! I built a collection of these components to build all of our modals, and to give users the power to construct their own and potentially override specific aspects of the layout, if need be.

Looking at an existing set of modals:

screenshot 2018-05-09 17 10 56

* ModalDialog (red): Holds the entire modal and all of the contents (a set of ModalContentPanes).
* ModalContentPane (orange): Holds a set of ModalTitleBar, ModalHeader, ModalContent, and ModalFooter and renders them all. Additionally is responsible for rendering the ModalCloseButton.
* ModalTitleBar (green): Renders the Title and Subtitle.
* ModalHeader (teal): Renders the section displayed beneath the titlebar and above the contents.
* ModalContent (purple): The main contents of the ModalContentPane (and the only required part of the content pane).
* ModalFooter (blue): An optional footer to display beneath the contents.

This diff also implements the following new additions to the existing modals:
* OneColumnModal is now available!
* StandardModal Headers are now available (they are dark and if provided cause the titlebar to turn dark too).
* StandardModal Preview (displayed next to the main contents and have less padding than a normal contents).

I think you'll find that by breaking everything up all of the components are much simpler now. All of the main modals are quite small and (IMO) easy to understand. If someone wanted to build a similar one they could easily copy the contents of an existing modal and modify it.

Not done yet and will do in a following diff:
* Get mobile support working.
* Write documentation for all of this!

### DIFF
--- a/packages/wonder-blocks-modal/components/modal-content-pane.js
+++ b/packages/wonder-blocks-modal/components/modal-content-pane.js
@@ -6,17 +6,16 @@ import Color from "wonder-blocks-color";
 import {View} from "wonder-blocks-core";
 
 import ModalCloseButton from "./modal-close-button.js";
+import ModalContent from "./modal-content.js";
+import ModalTitleBar from "./modal-title-bar.js";
+import ModalHeader from "./modal-header.js";
+import ModalFooter from "./modal-footer.js";
 
-import typeof ModalContent from "./modal-content.js";
-import typeof ModalTitleBar from "./modal-title-bar.js";
-import typeof ModalHeader from "./modal-header.js";
-import typeof ModalFooter from "./modal-footer.js";
-
-export default class ModalContentPane extends React.Component<{
-    content: React.Element<ModalContent>,
-    titlebar?: React.Element<ModalTitleBar>,
-    header?: React.Element<ModalHeader>,
-    footer?: React.Element<ModalFooter>,
+type Props = {
+    content: React.Element<typeof ModalContent> | React.Node,
+    titleBar?: React.Element<typeof ModalTitleBar>,
+    header?: React.Element<typeof ModalHeader> | React.Node,
+    footer?: React.Element<typeof ModalFooter> | React.Node,
     showCloseButton: boolean,
     color: "light" | "dark",
     style?: any,
@@ -34,7 +33,9 @@ export default class ModalContentPane extends React.Component<{
      * clicks too.)
      */
     onClickCloseButton: () => void,
-}> {
+};
+
+export default class ModalContentPane extends React.Component<Props> {
     static defaultProps = {
         showCloseButton: false,
         color: "light",
@@ -44,7 +45,7 @@ export default class ModalContentPane extends React.Component<{
     render() {
         const {
             content,
-            titlebar,
+            titleBar,
             header,
             footer,
             showCloseButton,
@@ -53,16 +54,30 @@ export default class ModalContentPane extends React.Component<{
             style,
         } = this.props;
 
-        const mainContent =
-            titlebar || footer
-                ? React.cloneElement(content, {
-                      style: [
-                          titlebar && styles.hasTitle,
-                          footer && styles.hasFooter,
-                          content.props.style,
-                      ],
-                  })
-                : content;
+        let mainContent =
+            content &&
+            typeof content === "object" &&
+            content.type === ModalContent ? (
+                ((content: any): React.Element<typeof ModalContent>)
+            ) : (
+                <ModalContent>{content}</ModalContent>
+            );
+
+        if (mainContent && (titleBar || footer || header)) {
+            mainContent = React.cloneElement(mainContent, {
+                header: header || mainContent.props.header,
+                style: [
+                    titleBar && styles.hasTitleBar,
+                    footer && styles.hasFooter,
+                    mainContent.props.style,
+                ],
+            });
+        }
+
+        // Figure out the background color that'll be behind the close button
+        // If a titlebar is specified then we use that, otherwise we use the
+        // default background color.
+        const topBackgroundColor = (titleBar && titleBar.props.color) || color;
 
         return (
             <View
@@ -72,19 +87,19 @@ export default class ModalContentPane extends React.Component<{
                     <View style={styles.closeButton}>
                         <ModalCloseButton
                             color={
-                                color === "dark" ||
-                                (titlebar && titlebar.props.color === "dark")
-                                    ? "light"
-                                    : "dark"
+                                topBackgroundColor === "dark" ? "light" : "dark"
                             }
                             onClick={onClickCloseButton}
                         />
                     </View>
                 )}
-                {titlebar}
-                {header}
+                {titleBar}
                 {mainContent}
-                {footer}
+                {!footer || footer.type === ModalFooter ? (
+                    footer
+                ) : (
+                    <ModalFooter>{footer}</ModalFooter>
+                )}
             </View>
         );
     }
@@ -111,7 +126,7 @@ const styles = StyleSheet.create({
         color: Color.white,
     },
 
-    hasTitle: {
+    hasTitleBar: {
         paddingTop: 32,
     },
 
--- a/packages/wonder-blocks-modal/components/modal-content-pane.js
+++ b/packages/wonder-blocks-modal/components/modal-content-pane.js
@@ -1,0 +1,121 @@
+// @flow
+import * as React from "react";
+import {StyleSheet} from "aphrodite";
+
+import Color from "wonder-blocks-color";
+import {View} from "wonder-blocks-core";
+
+import ModalCloseButton from "./modal-close-button.js";
+
+import typeof ModalContent from "./modal-content.js";
+import typeof ModalTitleBar from "./modal-title-bar.js";
+import typeof ModalHeader from "./modal-header.js";
+import typeof ModalFooter from "./modal-footer.js";
+
+export default class ModalContentPane extends React.Component<{
+    content: React.Element<ModalContent>,
+    titlebar?: React.Element<ModalTitleBar>,
+    header?: React.Element<ModalHeader>,
+    footer?: React.Element<ModalFooter>,
+    showCloseButton: boolean,
+    color: "light" | "dark",
+    style?: any,
+
+    /**
+     * Called when the close button is clicked.
+     *
+     * If you're using `ModalLauncher`, you probably shouldn't use this prop!
+     * Instead, to listen for when the modal closes, add an `onClose` handler
+     * to the `ModalLauncher`.
+     *
+     * This defaults to a no-op via `defaultProps`. (When used in a
+     * `ModalLauncher`, we'll automatically add an extra listener here via
+     * `cloneElement`, so that the `ModalLauncher` can listen for close button
+     * clicks too.)
+     */
+    onClickCloseButton: () => void,
+}> {
+    static defaultProps = {
+        showCloseButton: false,
+        color: "light",
+        onClickCloseButton: () => {},
+    };
+
+    render() {
+        const {
+            content,
+            titlebar,
+            header,
+            footer,
+            showCloseButton,
+            onClickCloseButton,
+            color,
+            style,
+        } = this.props;
+
+        const mainContent =
+            titlebar || footer
+                ? React.cloneElement(content, {
+                      style: [
+                          titlebar && styles.hasTitle,
+                          footer && styles.hasFooter,
+                          content.props.style,
+                      ],
+                  })
+                : content;
+
+        return (
+            <View
+                style={[styles.wrapper, color === "dark" && styles.dark, style]}
+            >
+                {showCloseButton && (
+                    <View style={styles.closeButton}>
+                        <ModalCloseButton
+                            color={
+                                color === "dark" ||
+                                (titlebar && titlebar.props.color === "dark")
+                                    ? "light"
+                                    : "dark"
+                            }
+                            onClick={onClickCloseButton}
+                        />
+                    </View>
+                )}
+                {titlebar}
+                {header}
+                {mainContent}
+                {footer}
+            </View>
+        );
+    }
+}
+
+const styles = StyleSheet.create({
+    wrapper: {
+        flex: "1 1 auto",
+        position: "relative",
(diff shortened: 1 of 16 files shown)


jeresig:

Built with commit 586b6472762f7724cc8d56076e3522d42ffaabd0

https://deploy-preview-81--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-content.js
```diff
@@ -0,0 +1,24 @@
...
+        const {style, children} = this.props;
+        return <View style={[styles.content, style]}>{children}</View>;
+    }
+}
+
+const styles = StyleSheet.create({
+    content: {
+        flex: "1",
+        padding: 64,
+        overflow: "auto",
```
kevinbarabash: I think we should have some way to easily override this styling.  One of the examples in the modals Zeplin has no left/right padding using the standard modal with preview.

kevinbarabash: Maybe we can have a prop on modals that toggles padding within content panes and by default it's on.

packages/wonder-blocks-modal/components/modal-header.js
```diff
@@ -0,0 +1,48 @@
...
+        boxSizing: "border-box",
+        maxHeight: 108,
+        paddingLeft: 64,
+        paddingRight: 64,
+        paddingTop: 8,
+        paddingBottom: 8,
+
+        display: "flex",
+        flexDirection: "row",
+        overflow: "auto",
```
kevinbarabash: I don't think the header should ever be scrollable.

packages/wonder-blocks-modal/components/modal-content-pane.js
```diff
@@ -0,0 +1,121 @@
...
+                style={[styles.wrapper, color === "dark" && styles.dark, style]}
+            >
+                {showCloseButton && (
+                    <View style={styles.closeButton}>
+                        <ModalCloseButton
+                            color={
+                                color === "dark" ||
+                                (titlebar && titlebar.props.color === "dark")
+                                    ? "light"
+                                    : "dark"
```
kevinbarabash:
```
const invert = color => color === "dark" ? "light" : "dark";  
const backgroundColor = color || (titlebar && titlebar.props.color) || "light";
const closeButtonColor = invert(backgroundColor);
```
maybe?

packages/wonder-blocks-modal/components/modal-dialog.js
```diff
@@ -0,0 +1,48 @@
...
+import {StyleSheet} from "aphrodite";
+
+import {View} from "wonder-blocks-core";
+
+import typeof ModalContentPane from "./modal-content-pane.js";
+
+export default class ModalDialog extends React.Component<{
+    children: React.ChildrenArray<React.Element<
+        ModalContentPane,
+    > | null | void>,
```
kevinbarabash: I'm not sure what prettier was thinking here.  Can you pull these props out into a separate type declaration?

packages/wonder-blocks-modal/components/modal-content.js
```diff
@@ -0,0 +1,24 @@
...
+}> {
+    render() {
+        const {style, children} = this.props;
+        return <View style={[styles.content, style]}>{children}</View>;
+    }
+}
+
+const styles = StyleSheet.create({
+    content: {
+        flex: "1",
```
kevinbarabash: This can be `flex: 1`.

packages/wonder-blocks-modal/components/modal-footer.js
```diff
@@ -0,0 +1,37 @@
...
+import * as React from "react";
+import {StyleSheet} from "aphrodite";
+
+import Color from "wonder-blocks-color";
+import {View} from "wonder-blocks-core";
+
+export default class ModalFooter extends React.Component<{
+    children: React.Node,
+    style?: any,
+}> {
```
kevinbarabash: Extract `type Props = {..}`.

packages/wonder-blocks-modal/components/modal-title-bar.js
```diff
@@ -0,0 +1,106 @@
...
+
+const styles = StyleSheet.create({
+    titlebar: {
+        flex: "0 0 auto",
+        boxSizing: "border-box",
+        minHeight: 64,
+        paddingLeft: 4,
+        paddingRight: 4,
+        paddingTop: 8,
+        paddingBottom: 8,
```
kevinbarabash: I wonder if we should update these later to using `spacing` once that lands?

packages/wonder-blocks-modal/components/one-column-modal.js
```diff
@@ -0,0 +1,62 @@
...
+                />
+            </ModalDialog>
+        );
+    }
+}
+
+const styles = StyleSheet.create({
+    wrapper: {
+        width: "64.65%",
+    },
```
kevinbarabash: The `height` should also be set.

packages/wonder-blocks-modal/components/standard-modal.js
```diff
@@ -59,163 +62,73 @@ type Props = {
...
-    titleBarSpacer: {
-        flex: "1 0 0",
-
-        // When the modal gets small, provide some minimal space here, to
-        // prevent the text from running up against the edge.
-        minWidth: 16,
+    previewContent: {
+        padding: "0 64px 0 0",
     },
 });
```
kevinbarabash: This became so much simpler.  Nice!

packages/wonder-blocks-modal/components/standard-modal.test.js
```diff
@@ -6,18 +6,10 @@ import StandardModal from "./standard-modal.js";
 
 describe("StandardModal", () => {
     test("Clicking the close button triggers `onClickCloseButton`", () => {
```
kevinbarabash: The test title no longer reflects what the test is doing.

packages/wonder-blocks-modal/components/modal-dialog.js
```diff
@@ -0,0 +1,48 @@
...
+                {children}
+            </View>
+        );
+    }
+}
+
+const styles = StyleSheet.create({
+    wrapper: {
+        display: "flex",
+        flexDirection: "row",
```
kevinbarabash: This might be a little surprising to someone who's doing build-your-own modal.  I think we should keep this `"column"` by default and override it in `two-column-modal.jsx`.

packages/wonder-blocks-modal/components/modal-content-pane.js
```diff
@@ -0,0 +1,121 @@
...
+                            }
+                            onClick={onClickCloseButton}
+                        />
+                    </View>
+                )}
+                {titlebar}
+                {header}
+                {mainContent}
+                {footer}
```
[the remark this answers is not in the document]
jeresig: To the first point: I disagree. I think we should explicitly make people pass in the element to render as they can use it to apply the styling. One alternative is that I could do something like:

`{titlebar instanceof ModalTitleBar ? titlebar : <ModalTitlebar>{titlebar}</ModalTitlebar>}`

Would you want that? And then if you want to override the styling of those parts explicitly then you pull in the component and override the `style`. At the very least we should be giving user the ability to pass these in and shouldn't always override the value.

As to the second part: good catch, I'll fix this.

packages/wonder-blocks-modal/components/modal-content-pane.js
```diff
@@ -0,0 +1,121 @@
...
+    static defaultProps = {
+        showCloseButton: false,
+        color: "light",
+        onClickCloseButton: () => {},
+    };
+
+    render() {
+        const {
+            content,
+            titlebar,
```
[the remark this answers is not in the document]
jeresig: Yep, will fix! (to titleBar)

packages/wonder-blocks-modal/components/modal-content.js
jeresig: Oh, to be clear that's exactly what I'm doing for the preview area (that I already implemented). I pass in a ModalContent and override the `style` to reset the padding to be what I want. I feel like every part of the system can be overridden right now so it's quite flexible!

packages/wonder-blocks-modal/components/modal-header.js
jeresig: Ok, will remove.

packages/wonder-blocks-modal/components/modal-content-pane.js
jeresig: Hmm, ok! Will change.

packages/wonder-blocks-modal/components/modal-dialog.js
jeresig: Ha, will do.

packages/wonder-blocks-modal/components/modal-content.js
jeresig: Oh interesting, ok!

packages/wonder-blocks-modal/components/modal-footer.js
jeresig: Will do.

packages/wonder-blocks-modal/components/modal-title-bar.js
jeresig: Yep, I'm planning on it!

packages/wonder-blocks-modal/components/one-column-modal.js
jeresig: I don't think so! At most a maxHeight should be set - unless I'm mis-understanding the design I believe they're supposed to keep growing. Either way I can always tweak this later.

packages/wonder-blocks-modal/components/standard-modal.test.js
jeresig: Ah, will fix.

packages/wonder-blocks-modal/components/modal-dialog.js
jeresig: I... disagree. Both TwoColumnModal and StandardModal need the flexDirection of row. I can't think of a situation where you would want two ModalContentPanes stacked on top of each other (which is what column would do). Not a single design has that, so I feel like having it default to row makes more sense.

packages/wonder-blocks-modal/components/modal-content-pane.js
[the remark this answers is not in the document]
kevinbarabash:
```
{titlebar instanceof ModalTitleBar ? titlebar : <ModalTitlebar>{titlebar}</ModalTitlebar>}
```
is a really good idea.  We get the best of both worlds!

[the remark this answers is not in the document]
kevinbarabash: Note that the second part makes your proposed solution to the first part more difficult to implement.

packages/wonder-blocks-modal/components/one-column-modal.js
kevinbarabash: I was under the impression that all modals were proportional in both height and width to the window by default.

packages/wonder-blocks-modal/components/modal-dialog.js
kevinbarabash: I was thinking of a situation where a person creates a modal using `ModalDialog` and fills it in using their own custom components.  Since that should be discouraged and probably pretty edge case-y I guess we can keep this as is.

matchu:
(FWIW, I'm partly skipping because this change includes a lot of moving parts, which means it'd be a larger time commitment to get all the context straight. But that's probably fine since you have Kevin as a dedicated reviewer who's already engaged in the problem space 👍🏻)

packages/wonder-blocks-modal/components/modal-content-pane.js
```diff
@@ -53,16 +54,30 @@ export default class ModalContentPane extends React.Component<{
...
+            typeof content === "object" &&
+            content.type === ModalContent ? (
+                ((content: any): React.Element<typeof ModalContent>)
+            ) : (
+                <ModalContent>{content}</ModalContent>
+            );
+
+        if (mainContent && (titleBar || footer || header)) {
+            mainContent = React.cloneElement(mainContent, {
+                header: header || mainContent.props.header,
```
kevinbarabash: I'm glad that wasn't too tricky.

packages/wonder-blocks-modal/components/modal-content.js
```diff
@@ -4,19 +4,39 @@ import {StyleSheet} from "aphrodite";
...
+export default class ModalContent extends React.Component<Props> {
     render() {
-        const {style, children} = this.props;
-        return <View style={[styles.content, style]}>{children}</View>;
+        const {header, style, children} = this.props;
+
+        return (
+            <View style={styles.wrapper}>
+                {!header || header.type === ModalHeader ? (
+                    header
```
kevinbarabash: I see... if we don't have a `header` we render nothing.

packages/wonder-blocks-modal/components/standard-modal.js
```diff
@@ -81,20 +79,16 @@ export default class StandardModal extends React.Component<Props> {
...
-                    header={
-                        header ? <ModalHeader>{header}</ModalHeader> : undefined
-                    }
-                    content={<ModalContent>{content}</ModalContent>}
-                    footer={
-                        footer ? <ModalFooter>{footer}</ModalFooter> : undefined
-                    }
+                    header={header}
+                    content={content}
+                    footer={footer}
```
kevinbarabash: Yes!

packages/wonder-blocks-modal/components/modal-content.js
```diff
@@ -33,6 +33,10 @@ const styles = StyleSheet.create({
     wrapper: {
         flex: 1,
         overflow: "auto",
+
+        // This helps to ensure that the paddingBottom is preserved when
+        // the contents start to overflow, this goes away on display: flex
+        display: "block",
```
kevinbarabash: This is such a non-obvious part of `flex-box`.  I don't understand why this decision was made in the spec.

PR #82, opened 2018-05-09
Add grid-size support to style property.

jeresig:
Add in support for passing a function to the `style` property of our HOC so that we can display certain styles based on the current grid size. This should replace the need for most media queries in our code base (we'll just want to wrap our site in one grid as a base but we can always override it further down).


jeresig:

Built with commit 0b53d096ee428045172ed3c90818a81a1877aae2

https://deploy-preview-82--wonder-blocks.netlify.com

packages/wonder-blocks-core/generated-snapshot.test.js
```diff
@@ -56,16 +56,10 @@ describe("wonder-blocks-core", () => {
...
-
-        const example = (
-            <StyledInput
-                style={styles.pink}
-                type="text"
-                placeholder="hello, world"
-            />
-        );
+        
+        const example = <StyledInput style={styles.pink} type="text" placeholder="hello, world"/>;
```
kevinbarabash: Why the change in formatting?

packages/wonder-blocks-core/generated-snapshot.test.js
```diff
@@ -56,16 +56,10 @@ describe("wonder-blocks-core", () => {
                 backgroundColor: "pink",
             },
         });
-
+
```
kevinbarabash: We shouldn't have extra whitespace.

packages/wonder-blocks-core/generated-snapshot.test.js
kevinbarabash: Can you run `npm run pretty-quick` manually?

packages/wonder-blocks-core/util/util.js
```diff
@@ -3,14 +3,20 @@ import {css} from "aphrodite";
...
 import type {StyleType} from "./types.js";
 
-function flatten<T: Object>(list?: StyleType<T>): Array<T> {
+export type GridSize = "small" | "medium" | "large";
+
+function flatten<T: Object>(list?: StyleType<T>, gridSize: GridSize): Array<T> {
     const result: Array<T> = [];
 
+    if (typeof list === "function") {
+        list = list(gridSize);
```
kevinbarabash: This makes sense.

packages/wonder-blocks-core/generated-snapshot.test.js
jeresig: pretty-quick doesn't do anything if there are no changes. I ended up having to run `./node_modules/.bin/prettier --write packages/wonder-blocks-*/generated-snapshot.test.js` - but that changed even more files (I think prettier isn't getting run against these files for some reason).

jeresig:
@kevinbarabash I fixed the formatting issue, should be good-to-go.

package.json
```diff
@@ -7,7 +7,7 @@
     "install": "for d in packages/*/ ; do (cd \"$d\" && npm install); done",
     "lint": "eslint --config ./eslint/eslintrc packages",
     "flow": "flow",
-    "gen-snapshot-tests": "node utils/gen-snapshot-tests.js",
+    "gen-snapshot-tests": "node utils/gen-snapshot-tests.js && prettier --write packages/*/generated-snapshot.test.js",
```
jeresig: @kevinbarabash I added this in and now files are being properly formatted all the time!

kevinbarabash:
@jeresig thanks for fixing the prettier issue.

PR #83, opened 2018-05-10
Update View to have to have display: flex; flexDirection: column

kevinbarabash:
The rationale for this change is:
- flex-box layouts can break when a <div> is inserted in the layout
  without setting display: flex;
- by default flexDirection is "row" on the web, but this is quite
  different from the default layout of children when display: flex
  isn't set
- these are defaults for React Native's View component after which
  our View component is modelled so it makes sense have similar
  defaults

I've also re-enabled display of .snap files in github reviews.

Test Plan:
- npm test
- npm run styleguidist, look at all of the components to verify that
  nothing's broken after this change

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,3 @@
 docs/* linguist-generated
-**/__snapshots__/** linguist-generated
 package-lock.json linguist-generated
+**/generated-snapshot.test.js
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,3 @@
 docs/* linguist-generated
 **/__snapshots__/** linguist-generated
 package-lock.json linguist-generated
-**/generated-snapshot.test.js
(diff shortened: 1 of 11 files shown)


jeresig:

Built with commit 64a3512a63513193a0a10603221f18d833ef8764

https://deploy-preview-83--wonder-blocks.netlify.com

.gitattributes
```diff
@@ -1,4 +1,3 @@
 docs/* linguist-generated
 **/__snapshots__/** linguist-generated
 package-lock.json linguist-generated
-**/generated-snapshot.test.js
```
jeresig: I think you did the wrong thing - I think you want to remove the __snapshots__ line above instead!

kevinbarabash:
@jeresig the bulk of the changes in the snapshots seem to be in the grid.  Can you have a look at styleguidist and let me know if the behavior of the grid matches what we had before?

.gitattributes
kevinbarabash: d'oh 🤦‍♂️

kevinbarabash:
> hope this doesn't impact Webapp to badly.

We haven't started using `View` that widely in `webapp`.  I'm going to double check all of the current call sites though.

PR #84, opened 2018-05-14
Add mobile support to modal dialogs.

jeresig:
This adds in mobile support for all of the modal dialogs. Some of these were easier to implement than other (like with the standard modal). The one and two column modals had to be restructured so that their footer would be positioned outside of the ModalContentPane so that everything else could be scrollable (content, close button, etc.)

I also went in and stripped out a lot of old Grid modal stuff that we're not using anymore (but we are using some of it!) I think there may be room for a new Grid component to make some of this process easier so that we don't have to use React Context but I'll save that for a future diff.

It's probably easiest to just look at the live example on Netlify to see how it works. The non-mobile version should work the same - and the mobile one should match what's proposed in Zeplin.

* All modals should consume the entire viewport, nothing behind it should be visible.
* The horizontal margins are tighter (16px) and the vertical are the same (32px).
* If there is no title bar (and always in one and two-column modals) the "X" is positioned above the rest of the content and scrolls with the content.
* All content, except for the title bar and the footer, scroll off of the screen.
* The preview area in the standard modal disappears.
* The sidebar in the two-column modal becomes a pseudo-header.


packages/wonder-blocks-grid/components/grid.js
```diff
@@ -161,8 +162,10 @@ export default class Grid extends React.Component<
     }
 
     render() {
+        // eslint-disable-next-line no-unused-vars
```
jeresig: All of these changes are super-annoying. I had to apply styling to the Grid component as having an un-styled flexbox element was messing up the layout. This will all get removed when we can use React fragments.

jeresig:

Built with commit ecd2497de9c5e06a3a594899dc6fbdef27bb12aa

https://deploy-preview-84--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/standard-modal.md
```diff
@@ -41,7 +40,7 @@ const styles = StyleSheet.create({
             title="Title"
             content={
                 <View style={styles.modalContent}>
-                    <Body tag="p">
```
jeresig: I removed these as paragraphs add vertical padding and made it confusing to reason about the visible spacing between things (like the header and the content).

packages/wonder-blocks-modal/util/util.js
```diff
@@ -1,21 +0,0 @@
-// @flow
```
jeresig: We're not using anything from this file anymore so it's safe to remove it!

kevinbarabash:
I was playing around with the demo and notice that the modal switches to mobile mode when the resizing the page width to be smaller than 768, but it only switches back to desktop when resizing the page width to be larger than 1024.

jeresig:
@kevinbarabash Yeah, I noticed that to! I think that's a larger issue with the Grid, potentially. I'll tackle that in a separate PR.

jeresig:
@matchu Hmm - that's definitely not expected, and I'm not seeing it on my end. I can confirm that it opens and focuses the button in the footer (causing the X to go out of view) but I can scroll back up and click the X just fine. We should probably re-consider what we're focusing and how if stuff like this is happening though (although I'm not sure how common a 448px tall viewport is?)

packages/wonder-blocks-grid/components/grid.js
```diff
@@ -103,6 +100,10 @@ export default class Grid extends React.Component<
         this.watchHandlers = {};
 
         for (const size of VALID_GRID_SIZES) {
+            if (!spec[size]) {
+                continue;
+            }
```
kevinbarabash: Would this happen if a grid spec only defined a single size, e.g. `large`?  Are there situations in which it makes sense to only define a single size?

matchu:
Hmm okay. I think the general issue isn't so much the focus or the viewport height, but that scrolling seems to not work. I'm trying on my real phone now, and confirmed that when I launch a modal in Firefox for Android, it renders larger than the viewport but I can't scroll it at all. (Interestingly, in Chrome for Android, I _can_ scroll, but the modal seems to render at desktop size.)

packages/wonder-blocks-grid/components/grid.js
kevinbarabash: I don't follow.  Is there supposed to be a `style` prop within `otherProps`?

matchu:
I also have the scrolling issue in the standalone layout examples for both mobile and desktop Firefox. The modal renders larger than the content area, but there's no scrollbar to let me see the remaining content or the footer.

packages/wonder-blocks-grid/components/grid.js
```diff
@@ -161,8 +162,10 @@ export default class Grid extends React.Component<
     }
 
     render() {
+        // eslint-disable-next-line no-unused-vars
+        const {ssrSize, size, ...otherProps} = this.props;
         return (
-            <GridContext size={this.state.size} spec={this.props.spec}>
+            <GridContext {...otherProps} size={this.state.size}>
                 {this.props.children}
             </GridContext>
```
kevinbarabash: Won't `children` get passed `twice` to `GridContext`?  I usually pull out `children` as well when destructuring `this.props`.

matchu:
I can figure out how to do mobile screencap videos if that helps!

matchu:
I think on mobile, I expect the modal header to always be at the top of the viewport, and the modal footer to always be at the bottom, and for the content area to resize and scroll if necessary. Not sure if that's actually the intended design, though, or just what I'm imagining 😅

packages/wonder-blocks-grid/components/grid.js
kevinbarabash: I see... `GridContext` is defined below.  Why isn't it in its own file?

packages/wonder-blocks-grid/docs.md
```diff
@@ -117,27 +117,13 @@ Additionally, the following grid size specs are also available:
 | ----- | ------- | ------ | ------ | -------------------------------------------- |
 | large | 12      | 32px   | 16px   | `min-width: 1px` (No maximum content width.) |
 
-**12-Column Modal (`GRID_MODAL_12_SPEC`)**
+**12-Column Modal (`GRID_MODAL_SPEC`)**
```
kevinbarabash: I thought we were completely abandoning the grid for modals.

packages/wonder-blocks-grid/index.js
```diff
@@ -1,10 +1,10 @@
...
-export * from "./components/flex-cell.js";
-export * from "./components/grid.js";
-export * from "./components/gutter.js";
-export * from "./components/row.js";
+export {default as Cell} from "./components/cell.js";
+export {default as FixedWidthCell} from "./components/fixed-width-cell.js";
+export {default as FlexCell} from "./components/flex-cell.js";
+export {default as Grid} from "./components/grid.js";
+export {default as Gutter} from "./components/gutter.js";
+export {default as Row} from "./components/row.js";
```
kevinbarabash: fancy... we may want to make this part of our style guide.

jeresig:
@matchu ah, ok! I can confirm the issues in Firefox on mobile (no scrolling and contents are taller than the viewport) - Firefox on desktop, Chrome on desktop, and Chrome on mobile all appear to work as expected, for me. I'll look into the Firefox on mobile issue - must be a problem with flex box there.

(I'm curious: is Firefox on mobile a browser that we support?)

packages/wonder-blocks-modal/components/modal-content-pane.js
```diff
@@ -63,12 +66,18 @@ export default class ModalContentPane extends React.Component<Props> {
...
+                scrollOverflow,
                 header: header || mainContent.props.header,
                 style: [
                     titleBar && styles.hasTitleBar,
                     footer && styles.hasFooter,
+                    showCloseButton &&
+                        !titleBar &&
+                        ((gridSize) =>
+                            gridSize === "small" &&
+                            styles.smallWithCloseButton),
```
kevinbarabash: Are we actually intending people to use the grid within the modals?  If not, we should probably just you media queries instead.

packages/wonder-blocks-modal/components/modal-content-pane.js
```diff
@@ -38,6 +39,7 @@ type Props = {
 export default class ModalContentPane extends React.Component<Props> {
```
kevinbarabash: I was a little confused when I saw this again b/c I thought there would just content within this component.  Maybe `ModalPanel` would be better.

packages/wonder-blocks-modal/components/modal-content-pane.js
kevinbarabash: It might also be worth adding a comment as to why we have to inject `mainContent`'s style generally speaking.

packages/wonder-blocks-modal/components/modal-content.js
```diff
@@ -8,22 +8,40 @@ import ModalHeader from "./modal-header.js";
...
                     header
                 ) : (
                     <ModalHeader>{header}</ModalHeader>
                 )}
-                <View style={[styles.content, style]}>{children}</View>
+                <View
+                    style={[
+                        styles.content,
+                        (gridSize) => gridSize === "small" && styles.small,
+                        style,
```
kevinbarabash: I would usually expect styling to go on the root element of a component.  This breaks some implicit assumptions (that we should probably make explicit) about `style`.  In particular adding `margin` would add spacing within the scroll container not around it.  Perhaps we should rename `ModalContent`'s `style` prop to something else.

packages/wonder-blocks-modal/components/modal-content.js
```diff
@@ -8,22 +8,40 @@ import ModalHeader from "./modal-header.js";
 
 type Props = {
     header?: React.Element<typeof ModalHeader> | React.Node,
+    scrollOverflow: boolean,
     children: React.Node,
     style?: any,
```
kevinbarabash: comments

packages/wonder-blocks-modal/components/modal-content-pane.js
```diff
@@ -17,6 +17,7 @@ type Props = {
     header?: React.Element<typeof ModalHeader> | React.Node,
     footer?: React.Element<typeof ModalFooter> | React.Node,
     showCloseButton: boolean,
+    scrollOverflow: boolean,
     color: "light" | "dark",
     style?: any,
```
kevinbarabash: comments

packages/wonder-blocks-modal/components/one-column-modal.js
```diff
@@ -27,6 +32,58 @@ type Props = {
...
+                        onClickCloseButton={onClickCloseButton}
+                        content={content}
+                        footer={footer}
+                    />
+                </View>
+            );
+        }
+
+        return (
+            <View style={styles.contentFooterWrapper}>
```
kevinbarabash: I like this separation.  I wonder if we should maybe split this component into `MobileModalContent` and `DesktopModalContents` and have `OneColumnModal` decide which to render.  The benefit is that we'd be able to render both the desktop and mobile versions of the modals in styleguidist at the same time and people wouldn't have to resize the browser to see each version.

packages/wonder-blocks-grid/components/grid.js
jeresig: Or if there are cases where there are two specs specified (like in the case of the modal - there is no medium size).

packages/wonder-blocks-grid/components/grid.js
jeresig: Hmm - perhaps? It doesn't seem to have an impact though. I can change it?

kevinbarabash:
screen shot 2018-05-15 at 6 36 11 pm

Overscrolling allow you to see some white in-between the titlebar and header.  Fixing it would be nice, I think you might have to fight with a bit so I wouldn't worry about it for now.  I'll log it with some other issue in JIRA tomorrow.

packages/wonder-blocks-grid/components/grid.js
jeresig: We really don't want anyone using it - it's only purpose is to create the React Context (which we will need to rewrite when we upgrade to React 16+)

packages/wonder-blocks-grid/docs.md
jeresig: Nope - still using it! We're just using it for its viewport switching abilities, but not really any of the lower-level primitives.

kevinbarabash:
screen shot 2018-05-15 at 6 39 51 pm

I found an issue with the footer when there isn't a lot of text.  Not sure if that's the same issue y'all were discussing above.

packages/wonder-blocks-modal/components/modal-content-pane.js
jeresig: I don't think we should be using media queries at all - the grid replaces it entirely. I could renamed the main `<Grid>` component to be `<MediaLayout>` (or something) as that's really what it is - it's implementing a set of media queries and layouts that will be followed - once the content is wrapped with that there's no longer any reason to be using media queries.

packages/wonder-blocks-modal/components/modal-content.js
jeresig: I'm not doing comments/docs until we can get this landed and then I'll do them all at once.

packages/wonder-blocks-modal/components/modal-content.js
jeresig: Hmm - I'm not sure I follow. All the wrapper element is doing is adding the overflow - all the other styling is being applied to the element that's representing the contents. It definitely makes more sense to apply it to the contents (as being represented by what's passed in). Applying the user's styles to the wrapper would create some really strange effects (and I'm not sure that renaming it would improve things).

packages/wonder-blocks-modal/components/one-column-modal.js
jeresig: Hmmm - it would be quite as easy as that as we'd also have to render the correct ModalDialog, too (and a lot of the styling is determined by the viewport/grid so we'd have to find a way to override that). I'd like to defer that for now, even though it would be nice, as it's certainly going to be pretty challenging.

jeresig:
@kevinbarabash I've addressed some of your comments and left replies to others. I'm not seeing the issue that you screenshotted above - what browser was this on? Where'd you see this? Testing this in Chrome it looked ok to me. If there are remaining issues let me know which ones are blockers, thanks!

packages/wonder-blocks-modal/components/modal-content-pane.js
kevinbarabash: This is a pretty big shift in thinking.  I think calling it `MediaLayout` would help communicate that.  And then a grid layout is implemented using a combination of rows and cells and that layout changes in response to changes to the media that `MediaLayout` then feeds down to descendant components via context.

packages/wonder-blocks-modal/components/modal-content.js
kevinbarabash: In the docs we use `style` to add margin to add spacing between components.  This will lead people to think that you can always use `style` for that purpose.  In this case, the `style` prop on `ModalContent` can't be used for that purpose (afaict) so it should given a different name so that people don't get the wrong impression.

packages/wonder-blocks-modal/components/one-column-modal.js
kevinbarabash: I see... nevermind then.  I guess people will have to resize their browsers.

kevinbarabash:
@jeresig I was testing in Safari.

packages/wonder-blocks-modal/components/modal-content.js
jeresig: Do you have a recommendation for the name? In your example, adding a custom marginTop to the content will move it farther away from the title bar/header, as intended. I can't think of a name that wouldn't just make it confusing (especially since we definitely don't want people adding styling to the outer wrapper).

jeresig:
@kevinbarabash hmm - I must've fixed this with the change that I made for Firefox as I'm not seeing the issue in Safari now!

packages/wonder-blocks-modal/util/util.js
matchu: Oh, how are we doing a11y now? Are we assuming that there will only be one modal component instance per page? If so, how does this interact with cases where a page uses a modal layout inline, but might also render a modal on top of it?

packages/wonder-blocks-modal/components/modal-content.js
kevinbarabash: I guess this situation is a bit different b/c `ModalContent` will always be used inside `ModalPane`.  If it was used elsewhere then this would probably be a real problem but it sounds like the way it's currently set up it will meet dev expectations which is the important thing.

packages/wonder-blocks-modal/util/util.js
kevinbarabash: We can tackle this in a follow up PR.  I'll add an issue in JIRA.

jeresig:
@kevinbarabash thank you for filing the follow-up issues! One thing that wasn't clear to me was how we should handle the case where there is no title for the modal, but I can dig in to that in a follow-up.

Thank you both @kevinbarabash + @matchu for your reviews!

PR #85, opened 2018-05-14
[Link v1] Create LinkCore

chenyejia:
In this PR, we create the `LinkCore` component that will act as a stateless component which renders a state of `Link`. In a separate PR, we will add `ClickableComponent` to `Link`.

Due to complexities around styling the `:visited` pseudo class and visited `Link` designs, we will be updating visited `Link` designs and implementing them in a separate PR.

Test Plan:
View on Stylguidist


packages/wonder-blocks-link/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "wonder-blocks-button",
```
jeresig: I saw the error you were getting - I wonder if you need to update this name? (I see that the dependencies are correct below, so I'm not sure what else it could be.

packages/wonder-blocks-link/components/link.js
```diff
@@ -0,0 +1,92 @@
...
+     * URL to navigate to.
+     */
+    href: string,
+
+    /**
+     * Whether to display a caret after the text.
+     */
+    caret: boolean,
+
+    kind: "primary" | "secondary",
```
jeresig: Whoops! Should add a comment to this one, too.

packages/wonder-blocks-link/components/link.js
```diff
@@ -0,0 +1,92 @@
...
+     */
+    children: string,
+
+    /**
+     * URL to navigate to.
+     */
+    href: string,
+
+    /**
+     * Whether to display a caret after the text.
```
jeresig: This isn't implemented yet, right?

packages/wonder-blocks-link/components/link-core.js
```diff
@@ -0,0 +1,87 @@
...
+            <StyledAnchor style={[defaultStyles, style]} href={href}>
+                {children}
+            </StyledAnchor>
+        );
+    }
+}
+
+const styles = {};
+
+const sharedStyles = {
```
jeresig: I remember this coming up in the ButtonCore review, as well - but was there a reason this isn't in an Aphrodite stylesheet? Given how many links we might have a on page it'd be useful to cut down on the code duplication.

chenyejia: Oh, nope! Will update :)

packages/wonder-blocks-link/components/link.js
chenyejia: No it's not. Will put a TODO

PR #86, opened 2018-05-15
[Link v1] Add ClickableBehavior to Link

chenyejia:
Move `ClickableBehavior` to `wonder-blocks-core` and add it to `Link`. Minor updates include fixing keyboard link behavior and updating focused state depending on last interaction.

Test Plan:
Tested Link behavior on styleguidist


jeresig:

Built with commit 49f9cce10cd0043d3d3c8096020b026b022b8744

https://deploy-preview-86--wonder-blocks.netlify.com

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -5,8 +5,8 @@ import {StyleSheet} from "aphrodite";
 import {LabelLarge} from "wonder-blocks-typography";
 import Color, {mix, fade} from "wonder-blocks-color";
 import {addStyle} from "wonder-blocks-core";
+import type {Handlers} from "wonder-blocks-core";
```
jeresig: Hmm - I feel like this name isn't descriptive enough. Maybe `ClickableHandlers` instead?

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -115,14 +115,13 @@ export default class ClickableBehavior extends React.Component<Props, State> {
     }
 
     handleClick = (e: SyntheticMouseEvent<>) => {
-        if (this.props.onClick) {
+        if (this.keyboardClick && e) {
```
jeresig: Hmm - when would `e` be undefined? Shouldn't it always be defined? (especially as stated by the Flow rule)

chenyejia: Double jk, it was actually because I wasn't passing in an object in the tests.

PR #87, opened 2018-05-15
Implement ProgressSpinner v1 (and CircularSpinner).

jeresig:
Implements the new ProgressSpinner v1, including the first spinner CircularSpinner. It's possible that future versions might include other spinners so I wanted to leave it open.

Compared with the implementation in webapp I made a few changes:

1) I removed the `inline` prop.
2) I added the `style` prop (making it possible to implement `inline` yourself).
3) I removed the use of external images and used inline SVG instead.

Most everything else I left the same!


packages/wonder-blocks-progress-spinner/components/circular-spinner.js
```diff
@@ -0,0 +1,114 @@
...
+import * as React from "react";
+import {StyleSheet} from "aphrodite";
+
+import {View, addStyle} from "wonder-blocks-core";
+import Color from "wonder-blocks-color";
+
+const heights = {
+    xsmall: 16,
+    small: 48,
+    default: 96,
```
kevinbarabash: These will need to change once we get the 24x24 side which matches the default icon size.

kevinbarabash: Instead of having a `default` size I think we should continue with the same `xsmall, small, medium, large, xlarge` that we've been use elsewhere.  We can use `defaultProps` to pick a "default".

kevinbarabash: Maybe:
```
xsmall: 16,
small: 24,
medium: 48,
large: 96,
```

jeresig: To clarify: this will mean that we can no longer switch existing ProgressSpinners to use this new one as they currently correlate to the sizes that I used here. I can change it, but it will just make the rollout more challenging.

kevinbarabash:
> I removed the inline prop.
> I added the style prop (making it possible to implement inline yourself).
> I removed the use of external images and used inline SVG instead.

All of these make a lot of sense to me.  The inline case is the extraordinary case so that should be the one we have to do extra work for.  :+1:

packages/wonder-blocks-progress-spinner/components/circular-spinner.js
```diff
@@ -0,0 +1,120 @@
...
+import {StyleSheet} from "aphrodite";
+
+import {View, addStyle} from "wonder-blocks-core";
+import Color from "wonder-blocks-color";
+
+const heights = {
+    xsmall: 16,
+    small: 24,
+    medium: 48,
+    large: 96,
```
kevinbarabash: LGTM

PR #88, opened 2018-05-15
flow check generate snapshot tests

kevinbarabash:


package.json
```diff
@@ -9,7 +9,7 @@
     "flow": "flow",
     "gen-snapshot-tests": "node utils/gen-snapshot-tests.js && prettier --write packages/*/generated-snapshot.test.js",
     "jest": "SNAPSHOT_INLINE_APHRODITE=1 jest",
-    "test": "npm run flow && npm run gen-snapshot-tests && npm run jest",
+    "test": "npm run lint && npm run gen-snapshot-tests && npm run flow && npm run jest",
```
kevinbarabash: May as well run `lint` as part of `test.  I've also move `flow` to come after `gee-snapshots-test` in order to catch any changes.

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -10,9 +10,9 @@ import type {SharedProps} from "./button.js";
 
 type Props = SharedProps &
     ClickableHandlers & {
-        hovered: boolean,
-        focused: boolean,
-        pressed: boolean,
+        hovered?: boolean,
+        focused?: boolean,
+        pressed?: boolean,
```
kevinbarabash: These aren't used in the `ButtonCore` examples in docs.md.

jeresig:

Built with commit a29587f392dba6d7e0ed8ca75a12cc21afc91bfc

https://deploy-preview-88--wonder-blocks.netlify.com

packages/wonder-blocks-color/color-with-name.js
```diff
@@ -0,0 +1,50 @@
+// @flow
```
kevinbarabash: moved from docs.md

packages/wonder-blocks-color/colored-text-with-background.js
```diff
@@ -0,0 +1,23 @@
+// @flow
```
kevinbarabash: moved from docs.md

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -63,17 +63,17 @@ type State = {
...
+    onMouseEnter?: () => void,
+    onMouseLeave?: () => void,
+    onMouseDown?: () => void,
+    onMouseUp?: (e: SyntheticMouseEvent<>) => void,
+    onTouchStart?: () => void,
+    onTouchEnd?: () => void,
+    onTouchCancel?: () => void,
+    onKeyDown?: (e: SyntheticKeyboardEvent<*>) => void,
+    onKeyUp?: (e: SyntheticKeyboardEvent<*>) => void,
+    onBlur?: (e: SyntheticFocusEvent<*>) => void,
```
kevinbarabash: These handlers aren't set in docs.md for `ButtonCore`.

packages/wonder-blocks-grid/components/grid.js
```diff
@@ -182,7 +181,7 @@ class GridContext extends React.Component<{
     spec: GridSpec,
 
     // The Row components that will make up the grid
-    children: React.ChildrenArray<Row>,
+    children: React.Node,
```
kevinbarabash: @jeresig I remember you saying something along the lines of the children could be anything.

utils/gen-snapshot-tests.js
```diff
@@ -102,7 +101,7 @@ for (const section of styleguideConfig.sections) {
         .filter((token) => token.type === "code")
         .map((token) => token.text);
     if (examples.length === 0) {
-        console.warn(`no examples for section ${section.name}`);
+        console.warn(`no examples for section ${section.name}`); // eslint-disable-line no-console
```
kevinbarabash: My editor picks up lint in this file.

package.json
jeresig: Great, thank you!

packages/wonder-blocks-button/components/button-core.js
jeresig: Hmm - it seems like we should just change the examples in docs.md rather than changing the structure of this component (we should probably always be specifying these properties - or provide defaults via defaultProps).

packages/wonder-blocks-color/color-with-name.js
jeresig: Could you move these helper files into a sub-directory? Perhaps just one named `docs` or `docutils`?

packages/wonder-blocks-core/components/clickable-behavior.js
jeresig: This also doesn't feel like something we want to have be optional - we really want all of these to be defined every time. It might be better to create placeholder handlers for these and stick them in docs.md (so we can do something like `{...defaultHandlers}`, perhaps).

packages/wonder-blocks-grid/components/grid.js
jeresig: Yep - I changed these in latest modal/grid diff!

packages/wonder-blocks-grid/components/row.js
```diff
@@ -46,12 +44,7 @@ export default class Row extends React.Component<{
      * Can also accept a function which receives the `gridSize` and
      * `totalColumns` and should return some React Nodes to render.
      */
-    children:
```
jeresig: Oh, what was wrong with this? I felt like this would still be helpful as it would define the arguments for the callback function.

packages/wonder-blocks-link/components/link.js
```diff
@@ -19,7 +19,7 @@ export type SharedProps = {
      * Whether to display a caret after the text.
      */
     // TODO(yejia): Add once we have an Icon component
-    caret: boolean,
+    caret?: boolean,
```
jeresig: Probably better to have a `defaultProps` for this and have it be `false`.

utils/gen-snapshot-tests.js
```diff
@@ -92,7 +91,7 @@ function generateTestFile(root, examples, componentFileMap) {
 
     const outPath = path.join(root, "generated-snapshot.test.js");
     fs.writeFileSync(outPath, data, "utf8");
-    console.log(`wrote ${outPath}`);
+    console.log(`wrote ${outPath}`); // eslint-disable-line no-console
```
jeresig: In the past I've had better luck using `eslint-disable-next-line` as Prettier will keep it intact and won't trap to wrap it. Might be worth trying that?

packages/wonder-blocks-button/components/button-core.js
kevinbarabash: I'll update to use `defaultProps`.

packages/wonder-blocks-color/color-with-name.js
kevinbarabash: Good idea.  I'll move these to `docutils`.

kevinbarabash:
There are a few more comments to address.

packages/wonder-blocks-link/components/link.js
kevinbarabash: It's already in `defaultProps`... removing the `?`

packages/wonder-blocks-grid/components/row.js
kevinbarabash: This wasn't necessary.  There were lots of flow errors when I started on this diff and I misinterpreted some of them.  Thanks for catching this.  Always better to have better types.

packages/wonder-blocks-core/components/clickable-behavior.js
kevinbarabash: Whoops... thanks for catching this.

PR #89, opened 2018-05-15
[IconButton v1] Create IconButton

chenyejia:
Create the `IconButton` component using the same `ClickableBehavior` + `_Core` pattern as `Button` and `Link`.

Note: Since we do not yet have an `Icon` component, we are rendering a plain SVG element, which assumes the size of the SVG to be 9.8px by 9.8px (which is the size of the icon used in the examples).

Test Plan:
View on Styleguidist


jeresig:

Built with commit 89de20a59ea5dfe10a5d2cd69f6efa23be578b5d

https://deploy-preview-89--wonder-blocks.netlify.com

packages/wonder-blocks-icon-button/components/icon-button-core.js
```diff
@@ -0,0 +1,143 @@
...
+import Color, {mix, fade} from "wonder-blocks-color";
+import {addStyle} from "wonder-blocks-core";
+import type {ClickableHandlers} from "wonder-blocks-core";
+import type {SharedProps} from "./icon-button.js";
+
+type Props = SharedProps &
+    ClickableHandlers & {
+        hovered: boolean,
+        focused: boolean,
+        pressed: boolean,
```
kevinbarabash: These should be `optional` since flow will complain about them not being used in the generated snapshot tests.  See https://github.com/Khan/wonder-blocks/pull/88.

packages/wonder-blocks-icon-button/components/icon-button-core.js
```diff
@@ -0,0 +1,143 @@
...
+    } else if (kind === "secondary") {
+        newStyles["default"] = {
+            color: offBlack,
+        };
+    } else if (kind === "tertiary") {
+        newStyles["default"] = {
+            color: offBlack64,
+        };
+    } else {
+        throw new Error("IconButton kind not recognized");
```
kevinbarabash: In practice this should never happen, but throwing an exception doesn't hurt. :)

packages/wonder-blocks-icon-button/components/icon-button-core.js
```diff
@@ -0,0 +1,143 @@
...
+            alt,
+            color,
+            kind,
+            light,
+            disabled,
+            testId,
+            style,
+            hovered,
+            focused,
+            pressed,
```
kevinbarabash: We should have `href` and `clientSideNav` props similar to `Button`.  Sorry that I left this out the IconButton spec doc.  I'll update it but basically copy the same behavior as `Button` for those two props.

packages/wonder-blocks-icon-button/components/icon-button.js
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * ```js
+ * <IconButton
+ *     icon={anIcon}
+ *     alt="An Icon"
+ *     onClick={console.log("Hello, world!")}
+ * />
+ * ```
+ * @version 1.0
+ * @since 1.0
```
kevinbarabash: I don't know how important `@since` is b/c the plan was to version each package independently.  @jeresig wdyt?

packages/wonder-blocks-icon-button/components/icon-button.js
```diff
@@ -0,0 +1,127 @@
...
+    render() {
+        const {onClick, ...sharedProps} = this.props;
+        return (
+            <ClickableBehavior
+                disabled={sharedProps.disabled}
+                onClick={onClick}
+            >
+                {(state, handlers) => {
+                    return (
+                        <IconButtonCore
```
kevinbarabash: This is so similar to the implementation for `Button`.  Could almost have an HOC for it.  The little differences between components would make that difficult though.

packages/wonder-blocks-icon-button/components/icon-button.md
```diff
@@ -0,0 +1,50 @@
...
+    8.8838835 9.94454365,9.3587572 9.65165045,9.65165045 C9.3587572,
+    9.94454365 8.8838835,9.94454365 8.59099025,9.65165045 L6.7196699,
+    7.7803301 C6.70573509,7.76639529 6.69246326,7.75204857 6.67985439,
+    7.73732912 C5.99121283,8.21804812 5.15353311,8.5 4.25,8.5 C1.90278981,
+    8.5 0,6.59721019 0,4.25 C0,1.90278981 1.90278981,0 4.25,0 C6.59721019,
+    0 8.5,1.90278981 8.5,4.25 C8.5,5.15353311 8.21804812,5.99121283
+    7.73732912,6.67985439 L7.73732912,6.67985439 Z M4.25,7.5 C6.04492544,
+    7.5 7.5,6.04492544 7.5,4.25 C7.5,2.45507456 6.04492544,1 4.25,1
+    C2.45507456,1 1,2.45507456 1,4.25 C1,6.04492544 2.45507456,7.5 4.25,
+    7.5 L4.25,7.5 Z`;
```
kevinbarabash: This is okay for now.  We'll want to optimize these paths once we start work on `wonder-blocks-icons`.

packages/wonder-blocks-icon-button/components/icon-button.md
```diff
@@ -0,0 +1,50 @@
...
+        marginRight: 10,
+    }
+});
+
+<View style={[styles.row]}>
+    <IconButton
+    	icon={searchIcon}
+    	alt="search"
+        onClick={console.log("hello")}
+        style={[styles.sideMargins]}
```
kevinbarabash: The alignment is funky.  I don't think prettier reformats code blocks within .md files.  It would be cool if it did.

packages/wonder-blocks-icon-button/docs.md
```diff
@@ -0,0 +1,581 @@
...
+                <IconButtonCore
+                    kind="primary"
+                    icon={searchIcon}
+                    alt="search"
+                    color={blue}
+                    light={false}
+                    hovered={false}
+                    focused={false}
+                    pressed={false}
+                    disabled={false}
```
kevinbarabash: Could probably just set the prop that's true instead of having to specify all these `false`s.

packages/wonder-blocks-icon-button/docs.md
```diff
@@ -0,0 +1,581 @@
...
+        </tr>
+    </tbody>
+</table>
+```
+
+IconButtonCores, Red:
+```js
+const Color = require("wonder-blocks-color").default;
+const IconButtonCore = require("./components/icon-button-core.js").default;
+const searchIcon = `M7.73732912,6.67985439 C7.75204857,6.69246326 7.76639529,
```
kevinbarabash: Maybe extract this out into a helper file somewhere.

packages/wonder-blocks-icon-button/components/icon-button-core.js
```diff
@@ -0,0 +1,143 @@
...
+                {...handlers}
+            >
+                <svg
+                    role="img"
+                    aria-label={alt}
+                    width="20px"
+                    height="20px"
+                    viewBox="0 0 9.8 9.8"
+                >
+                    <title>{alt}</title>
```
kevinbarabash: @somewhatabstract has been doing some work around labelling SVGs.  I believe the best practice is to set `aria-labelledby="title-id-123"` on the `svg` and then set `id="title-id-123"` on the `title` itself.

packages/wonder-blocks-icon-button/index.test.js
```diff
@@ -0,0 +1,27 @@
+// @flow
+import React from "react";
+import {mount} from "enzyme";
+
+import IconButton from "./index.js";
+
+describe("IconButton", () => {
```
kevinbarabash: It's nice that all the heavy lifting can be done by `ClickableBehavior` and that it can be reused between components.

packages/wonder-blocks-icon-button/components/icon-button.md
jeresig: It does if we re-enable prettier to run on .md files! It was failing so we had to disable it. Need to investigate getting it to run again.

packages/wonder-blocks-icon-button/docs.md
jeresig: I don't think we need to worry about that for now - it's all going to get re-written once the Icon implementation comes in!

packages/wonder-blocks-icon-button/components/icon-button.js
jeresig: Ah! I used that for Grid but I realize now that it doesn't really make sense. So yeah, I guess we could remove it :)

packages/wonder-blocks-icon-button/components/icon-button-core.js
chenyejia: Sorry I'm confused. They are used in `docs.md`?

packages/wonder-blocks-icon-button/components/icon-button-core.js
chenyejia: Will `onClick` be optional then?

packages/wonder-blocks-icon-button/components/icon-button-core.js
chenyejia: Hm, should we taking in a separate `labelId` or something then? Or would the `Icon` component handle this?

packages/wonder-blocks-icon-button/components/icon-button-core.js
kevinbarabash: My bad, disregard.  I ran into this issue with the existing buttons/links and I assumed that button links would have the same issue.

packages/wonder-blocks-icon-button/components/icon-button-core.js
kevinbarabash: @chenyejia could you ask in the a11y channel if that alternative works.  It would be nice to avoid having to generate unique ids if possible.

PR #90, opened 2018-05-16
update View default styles to match React Native defaults exactly

k4b7:
There's some really useful stuff in there, e.g. `box-sizing` and `minWidth`/`minHeight`.


kevinbarabash:
Looks like I forgot to update the snapshots.

kevinbarabash:
That was a little tricky.  I needed to rebase to get the ProgressSpinner changes and then update those snapshots.

PR #95, opened 2018-05-16
update flow 0.57.3

kevinbarabash:
We need to stay in lock-step with webapp b/c we use webapp's build system to build wonder-blocks.


jeresig:

Built with commit 407249c97b9dacfe0db7353a741a06c004d1fc65

https://deploy-preview-95--wonder-blocks.netlify.com

matchu:
I think Flow now gets upset if you try to read a field from an object that
isn't known to be on the object. Maybe checking `"type" in obj` first would
help? Seems unwieldy tho.

kevinbarabash:
The fix is to check `typeof(footer) === "object"` first since footer is a `React.Node` it could be a boolean or a string.  I'm going to add that check before merging this PR.

PR #97, opened 2018-05-17
[Button v1] Remove DefaultProps from ButtonCore

chenyejia:
(Sorry there's more!)
Remove so that we don't define in both `Button` and `ButtonCore`.

If we don't want to remove it, we should update `disable` to `disabled`!


jeresig:

Built with commit 4b001da092ada2e4ebf13d934628012eb1ab2620

https://deploy-preview-97--wonder-blocks.netlify.com

kevinbarabash:
Good catch.  I need to start looking at those .snap files more closely.

PR #100, opened 2018-05-22
experiment w/ dropdown that can be controlled and uncontrolled

kevinbarabash:
This is an experimental PR.  It is not meant to merge, but rather to explore the idea of making components that can act both as controlled and uncontrolled components.  The motivation for this is to reduce boilerplate by providing a controlled component by default but also allow devs to use the component in an uncontrolled mode where the parent provides all data as props and is also responsible update those props as necessary.


packages/wonder-blocks-forms/components/dropdown.js
```diff
@@ -0,0 +1,121 @@
+// @flow
```
kevinbarabash: The behavior of this component on the whole is really hacky, please focus on its ability to switch between being controlled vs uncontrolled as opposed to the user facing behavior of the component itself.

jeresig:

Built with commit 5c7818d98a350ae4fd86ec7dde6d8fc3e6818446

https://deploy-preview-100--wonder-blocks.netlify.com

packages/wonder-blocks-forms/components/dropdown.js
```diff
@@ -0,0 +1,121 @@
...
+    controlled: boolean,
+    items?: Array<{
+        label: string,
+        value: any,
+    }>,
+};
+
+type State = {
+    show: boolean,
+    selection: ?any,
```
kevinbarabash: It would be interesting to reuse this inside of props, e.g.
```
type State = {|
   show: boolean,
   selection: ?any,
|};

type Props = {|
   ...State,
   onClose: () => void,
   onOpen: () => void,
   ...
|};
```

packages/wonder-blocks-forms/components/dropdown.js
```diff
@@ -0,0 +1,121 @@
...
+        const show = !this.state.show;
+        this.setState({show});
+        if (show) {
+            this.props.onOpen();
+        } else {
+            this.props.onClose();
+        }
+    };
+
+    handleChange = (e: SyntheticEvent<>, value: any) => {
```
kevinbarabash: I'm wondering if we should always include the original event in callbacks just in case people need to do something with it like with buttons and confetti.

packages/wonder-blocks-forms/components/dropdown.js
```diff
@@ -0,0 +1,121 @@
...
+        this.setState({selection: value});
+        // $FlowFixMe
+        this.props.onChange(e, value);
+    };
+
+    render() {
+        const show = this.props.controlled ? this.state.show : this.props.show;
+        const selection = this.props.controlled
+            ? this.state.selection
+            : this.props.selection;
```
kevinbarabash: This could be simplified by doing:
```
const {show, selection} = this.props.controlled ? this.state : this.props;
```

packages/wonder-blocks-forms/generated-snapshot.test.js
```diff
@@ -0,0 +1,87 @@
...
+                    {label: "item 2", value: 2},
+                    {label: "item 3", value: 3},
+                ]}
+            />
+        );
+        const tree = renderer.create(example).toJSON();
+        expect(tree).toMatchSnapshot();
+    });
+    it("example 3", () => {
+        class Parent extends React.Component {
```
kevinbarabash: I had to remove `// @flow` from this file b/c we can add prop flow typing within docs.md and then this fails to pass `flow`'s type-checking... sadness.

kevinbarabash:
@jeresig this is ready for review again.

package.json
```diff
@@ -12,6 +12,7 @@
     "test": "npm run lint && npm run gen-snapshot-tests && npm run flow && npm run jest",
     "coverage": "SNAPSHOT_INLINE_APHRODITE=1 jest --coverage",
     "styleguidist": "styleguidist server",
+    "start": "npm run styleguidist",
```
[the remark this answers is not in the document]
joshwcomeau: ++++

packages/wonder-blocks-forms/components/dropdown.js
joshwcomeau: I think whenever the prop can be used to set an initial value, it should have a different name that suggests that (like `initialShow` instead of `show`).

My rationale for this is that we should make it clear that the prop only controls the initial value; changing the prop later on won't update the component's state.

But yeah that would foil the neat spreading state in props thing :/

kevinbarabash: I think being explicit is worth the price.

package.json
```diff
@@ -54,6 +55,7 @@
   },
   "dependencies": {
     "aphrodite": "^1.2.5",
+    "make-controllable": "0.0.3",
```
jeresig: 🛑 This isn't used anymore, it seems.

packages/wonder-blocks-button/generated-snapshot.test.js
```diff
@@ -1,4 +1,3 @@
-// @flow
```
jeresig: 🛑 Why did these go away?

packages/wonder-blocks-forms/components/dropdown.js
```diff
@@ -0,0 +1,85 @@
...
+            this.props.onClose();
+        }
+    };
+
+    handleItemClick = (item: Item, index: number) => {
+        this.setState({selection: index});
+        this.props.onChange(item, index);
+    };
+
+    render() {
```
jeresig: I like this, much easier to understand!

packages/wonder-blocks-forms/generated-snapshot.test.js
```diff
@@ -0,0 +1,94 @@
+// This file is auto-generated by gen-snapshot-tests.js
```
jeresig: 🛑 Also missing `// @flow`

packages/wonder-blocks-forms/index.js
```diff
@@ -0,0 +1,3 @@
+import TextField from "./components/text-field.js";
```
jeresig: 🛑 This looks like an extra file?

styleguide.config.js
```diff
@@ -81,6 +81,11 @@ module.exports = {
             components:
                 "packages/wonder-blocks-icon-button/components/icon-button.js",
         },
+        {
```
jeresig: 🛑 Extra file - also missing the dropdown stuff (although maybe that's intentional?)

utils/gen-snapshot-tests.js
```diff
@@ -29,7 +29,7 @@ const ignored = [
 
 function generateTestFile(root, examples, componentFileMap) {
     const lines = [
-        "// @flow",
+        // "// @flow",
```
jeresig: Ahhh - that should be fixed, I suspect!

packages/wonder-blocks-button/generated-snapshot.test.js
kevinbarabash: I can't create an example in `docs.md` that includes `extends React.Component<Props, State>` b/c `styleguidist` can't handle that.  The generated snapshot then contains code that `flow` won't pass.

packages/wonder-blocks-forms/index.js
kevinbarabash: Yeah... I started prototyping with that component, but it wasn't complicated enough to try everything out.

utils/gen-snapshot-tests.js
kevinbarabash: Yeah, I'll add it to the infra epic in JIRA.

kevinbarabash:
@jeresig it'll probably be easier to start from scratch, but I'll link to this PR from the on-boarding doc as a reference for the Component/ComponentCore pattern.

kevinbarabash:
Closing this since it's served its purpose of documenting a pattern we'd like to use for stateful components in wonder-blocks.

PR #101, opened 2018-05-23
Client nav button

kevinbarabash:
This is a first pass at enabling client-side navigation using react-router-dom.  I was originally think that we'd have the client-button.js wrapper live in webapp, but it's also useful here for test purposes so maybe we can have some sort of HOC that wraps existing components that have a `tag` and `href` prop to create a client-side nav version of that component.


kevinbarabash:
Going to recreate this with the proper base branch.

jeresig:

Built with commit 139bda501c99162612d1952902ed81e27d787668

https://deploy-preview-101--wonder-blocks.netlify.com

PR #102, opened 2018-05-23
prototype client-side routing for Button

kevinbarabash:
This is a first pass at enabling client-side navigation using react-router-dom. I was originally think that we'd have the client-button.js wrapper live in webapp, but it's also useful here for test purposes so maybe we can have some sort of HOC that wraps existing components that have a tag and href prop to create a client-side nav version of that component.

screen shot 2018-05-22 at 11 57 12 pm
screen shot 2018-05-22 at 11 57 16 pm
screen shot 2018-05-22 at 11 57 26 pm


jeresig:

Built with commit 139bda501c99162612d1952902ed81e27d787668

https://deploy-preview-102--wonder-blocks.netlify.com

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -52,14 +53,15 @@ export default class ButtonCore extends React.Component<Props> {
...
-        const Tag = href ? StyledAnchor : StyledButton;
+        const Tag = tag || (href ? StyledAnchor : StyledButton);
         const Label = size === "small" ? LabelSmall : LabelLarge;
         return (
             <Tag
                 style={[defaultStyle, style]}
                 disabled={disabled}
                 data-test-id={testId}
                 href={href}
+                to={href}
```
jeresig: This is going to cause problems if we're just using a normal `a` or `button`, React will complain about properties that it doesn't know how to handle.

packages/wonder-blocks-button/components/button.js
```diff
@@ -61,6 +61,8 @@ export type SharedProps = {
      */
     testId?: string,
 
+    tag?: React.ComponentType<*>,
```
jeresig: Hmmm... not a huge fan of this. The only time it should be used is for client-side nav so letting anything else through doesn't seem good.

packages/wonder-blocks-link/components/link.js
```diff
@@ -99,33 +102,33 @@ export default class Link extends React.Component<Props> {
...
+            : ClickableBehavior;
+
         return (
-            <ClickableBehavior
+            <Behavior
                 disabled={false}
                 onClick={onClick}
                 href={href}
-                clientSideNav={clientSideNav}
+                clientNav={clientNav}
```
kevinbarabash: Remove `clientNav`... now `history` is being pass by `withRouter`.

kevinbarabash:
@jeresig this is ready for review again.  Things are much cleaner with us just including `react-router-dom` in `wonder-blocks`, I'm glad we decided to switch.

packages/wonder-blocks-button/components/button.js
```diff
@@ -136,33 +139,33 @@ export default class Button extends React.Component<Props> {
...
-            href,
-            children,
-            clientSideNav,
-            ...sharedProps
-        } = this.props;
+        const {onClick, href, children, clientNav, ...sharedProps} = this.props;
+
+        const Behavior = clientNav
+            ? ClickableBehaviorWithRouter
+            : ClickableBehavior;
```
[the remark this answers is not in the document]
jeresig: Ok! So talked about this in Slack and came up with a new algorithm for how this should work (in pseudo-code):

```
if (externalURL(href)) {
    ClickableBehavior;
} else if (clientNav !== null) { 
	if (clientNav) {
		ClickableBehaviorWithRouter;
	} else {
		ClickableBehavior;
	}
} else if (insideReactRouter()) {
	ClickableBehaviorWithRouter;
} else {
	ClickableBehavior;
}
```

`insideReactRouter` can work by checking to see if `this.context.router` is defined. You'll need to specify: `static contextTypes = {router: propTypes.any;};` to get access to it.

For `isExternalURL` I found a couple versions here: https://stackoverflow.com/questions/6238351/fastest-way-to-detect-external-urls There is also this: https://www.npmjs.com/package/is-url-external but that requires the use of `document` and `createElement`, I think a regex might be better. Or, in fact, we could just assume that anything that starts with `http` is external and leave it at that.

kevinbarabash:
I'm going to have to create a new PR for this since I landed the parent already.

kevinbarabash:
I'm going to close this and open a new PR.

PR #103, opened 2018-05-24
change Button and IconButton's color prop to take 'default' or 'destructive'

kevinbarabash:
The follows what mobile has already done with their Button component.  Using semantic names for colors instead of what the actual color is allows us to update the actual color without having to update all of the call sites that use the color.  I've also updated the background color for the destructive button colors example as requested by @nefaurk.


jeresig:

Built with commit 8d5a8b2a6b7d22ec7e7a4147771b84b43b703ab0

https://deploy-preview-103--wonder-blocks.netlify.com

packages/wonder-blocks-button/components/button.js
```diff
@@ -26,7 +24,7 @@ export type SharedProps = {
     /**
      * The color of the button, either blue or red.
      */
-    color: ValidTints,
+    color: "default" | "destructive",
```
jeresig: 🤷‍♂️ Is "color" even the right term to use? intent? type? are we using color on mobile?

packages/wonder-blocks-color/index.js
```diff
@@ -31,9 +31,14 @@ const Color = {
     pink: "#fa50ae",
 };
 
+const SemanticColor = {
+    controlDefault: Color.blue,
+    controlDestructive: Color.red,
+};
+
 // Note: Only blue and red are valid tints for Buttons IconButtons. We
```
jeresig: 🛑 This should probably be removed, right?

🤷‍♂️ It might be worth replacing this with something like `export type ValidSemanticColors = "default" | "destructive";` and then you can use that in your button code.

packages/wonder-blocks-button/components/button.js
kevinbarabash: I believe we're using `color` on mobile.  I'm open to changing it.

PR #104, opened 2018-05-25
allow aria props on typography components

kevinbarabash:
I tested this by going to http://localhost:6060/#typography and using VoiceOver to verify that the `<Title>` component in the Accessibility section says "Accessibility Title".  I also verified that `aria-label="Accessibility Title"` was set on the underlying DOM element.

https://khanacademy.atlassian.net/browse/WB-207


jeresig:

Built with commit 659ebdec009045ad634955eb60cddc68362a6612

https://deploy-preview-104--wonder-blocks.netlify.com

packages/wonder-blocks-core/docs.md
```diff
@@ -33,7 +33,7 @@ Other props can be passed through `View` or `Text`, as if they were normal tags.
         Click me!
     </View>
 
-    <Text aria-hidden>
+    <Text aria-hidden="true">
```
kevinbarabash: The `flow` type-checking for `aria-` props found this.

packages/wonder-blocks-core/util/types.js
```diff
@@ -9,10 +9,133 @@
...
+    "tooltip",
+    "tree",
+    "treegrid",
+    "treeitem",
+    "widget", // abstract role
+    "window", // abstract role
+];
+
+// Source: https://www.w3.org/WAI/PF/aria-1.1/states_and_properties
+// TODO(kevinb): figure out how to type ID ref lists
```
kevinbarabash: Apparently some things can take multiple IDs.  In the example that I found of this online they use JS to set this to an array so maybe it's just `string | Array<string>`.

somewhatabstract: Yeah, there are a few that accept multiples. `aria-labelledby`, `aria-describedby`, `aria-controls`, etc.

packages/wonder-blocks-core/util/types.js
```diff
@@ -9,10 +9,133 @@
...
+    "tabpanel",
+    "textbox",
+    "timer",
+    "toolbar",
+    "tooltip",
+    "tree",
+    "treegrid",
+    "treeitem",
+    "widget", // abstract role
+    "window", // abstract role
```
somewhatabstract: I wonder if it's worth having an `abstractRoles` type and then combine that with the non-abstract list to define `roles`, rather than all these comments.

packages/wonder-blocks-core/util/types.js
somewhatabstract: Of course, they can be combined in a string delimited by spaces...but that's a pain. An array sounds like a nice API.

packages/wonder-blocks-core/util/types.js
```diff
@@ -9,10 +9,133 @@
...
+];
+
+// Source: https://www.w3.org/WAI/PF/aria-1.1/states_and_properties
+// TODO(kevinb): figure out how to type ID ref lists
+export type AriaProps = {
+    role?: roles,
+    "aria-activedescendant"?: string, // ID
+    "aria-atomic"?: "false" | "true",
```
somewhatabstract: This is only useful when `aria-live` is specified. I don't know that we can easily codify that, though.

packages/wonder-blocks-core/util/types.js
```diff
@@ -9,10 +9,133 @@
...
+    "aria-level"?: number, // integer
+    "aria-live"?: "assertive" | "off" | "polite",
+    "aria-multiline"?: "false" | "true",
+    "aria-multiselectable"?: "false" | "true",
+    "aria-orientation"?: "horizontal" | "vertical",
+    "aria-owns"?: string, // ID
+    "aria-posinset"?: number, // integer
+    "aria-pressed"?: "false" | "mixed" | "true" | "undefined",
+    "aria-readonly"?: "false" | "true",
+    "aria-relevant"?:
```
somewhatabstract: I wonder if we should deliberately not support this.

https://medium.com/dev-channel/why-authors-should-avoid-aria-relevant-5d3164fab1e3

kevinbarabash:
> I'd love it if we could somehow codify some of the rules around which aria- attributes go with which, but that might make the API less discoverable, so I'm happy with this.

You'll love the following rules from the jsx-a11y eslint plugin:
- [role-has-required-aria-props](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/role-has-required-aria-props.md)
- [role-supports-aria-props](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/role-supports-aria-props.md)

kevinbarabash:
It looks like those jsx-a11y rules only work on DOM elements.  We could add an option to allow check non-DOM nodes as well.

somewhatabstract:
That sounds like a great idea!

kevinbarabash:
I'm going to do this in a separate PR b/c I want to update existing components to have props that make sense for their role so that people can only add those aria props that are valid on `button` to `Button`, and `IconButton`.

PR #110, opened 2018-06-05
Component core tests

kevinbarabash:
We don't want people using ButtonCore, IconButtonCore, and LinkCore so there's no need for them to be in the docs.  I've moved their examples out of the docs and into separate test suites.  The tests are much shorter b/c we can use nested for loops to go through all possibilities.


jeresig:

Built with commit 6e64618d64fc00b323cd45096398800bc15f75bd

https://deploy-preview-110--wonder-blocks.netlify.com

kevinbarabash:
The code coverage when up a bit with this b/c we don't generate any tests for examples in `components/*.md` yet and this diff moved some of those examples to `docs.md` in the root.

packages/wonder-blocks-icon-button/index.test.js
```diff
@@ -25,3 +44,49 @@ describe("IconButton", () => {
         ).toThrowError("Light is only supported for primary IconButtons");
     });
 });
+
+describe("IconButtonCore", () => {
+    for (const kind of ["primary", "secondary", "tertiary"]) {
```
jeresig: 🤷‍♂️ It feels like it'd be nice to have a way to abstract out these permutations, but any solution I can think of would also be a lot of work, so it's probably not worth it!

packages/wonder-blocks-icon-button/index.test.js
```diff
@@ -1,22 +1,41 @@
...
+    7.73732912,6.67985439 L7.73732912,6.67985439 Z M4.25,7.5 C6.04492544,
+    7.5 7.5,6.04492544 7.5,4.25 C7.5,2.45507456 6.04492544,1 4.25,1
+    C2.45507456,1 1,2.45507456 1,4.25 C1,6.04492544 2.45507456,7.5 4.25,
+    7.5 L4.25,7.5 Z`;
 
 describe("IconButton", () => {
     test("render an icon", (done) => {
-        const wrapper = mount(
-            <IconButton icon="" alt="" onClick={() => done()} />,
+        const wrapper = shallow(
```
jeresig: 🤷‍♂️ Are you using shallow on these as we're now testing the internals below? I guess I'm not sure I see the value in not doing a full mount.

chenyejia:
I think `button.md` didn't get updated

packages/wonder-blocks-icon-button/index.test.js
kevinbarabash: It wasn't necessary in the first place.

packages/wonder-blocks-icon-button/index.test.js
kevinbarabash: It would be cool to be able generate these tests from flow types.  Maybe when I'm more familiar with react-docgen this will turn out be really easy.

kevinbarabash:
@chenyejia thanks for catching that.

PR #113, opened 2018-06-06
script to copy flow source files to dist

kevinbarabash:
This solves the problem of how to include .js.flow library definition files in published packages.  This diff just copies over the source files into the dist folder which is an awful solution, but there aren't any good alternatives since `flow gen-flow-files` doesn't work and writing library definitions from hand is really time consuming.  https://medium.com/netscape/shipping-flowtype-definitions-in-npm-packages-c987917efb65 has a good write up about this technique.


jeresig:

Built with commit 8e4cd9ed8cc42fdfea4c99009fd9939aaeeeed66

https://deploy-preview-113--wonder-blocks.netlify.com

kevinbarabash:
Another option is not to copy the files and instead set dist/index.js.flow to:
```
// @flow
export * from '../src';
```
and then include `src` in the files when publishing.  This approach feels functionally equivalent and means we can avoid the copy step.

utils/copy-flow-source.js
```diff
@@ -0,0 +1,17 @@
+const flowCopySource = require("flow-copy-source");
```
jeresig: 🛑 A quick comment here explaining what this file is used for would be good!

utils/copy-flow-source.js
```diff
@@ -0,0 +1,17 @@
...
+const path = require("path");
+const fs = require("fs");
+
+const packages = fs.readdirSync(path.join(__dirname, "../packages"));
+
+for (const pkg of packages) {
+    // eslint-disable-next-line
```
jeresig: 🛑 Would be good to include the rule that is being blocked, rather than blocking all errors.

PR #114, opened 2018-06-07
Rename Grid to MediaLayout. Add HOC for MediaLayoutWrapper.

jeresig:
This is a rewrite to change Grid to become a more generic "MediaLayout" component (housed in `wonder-blocks-core` instead of inside grid). There are more improvements that I would like to make but this is a good place to start.

The main changes:

* Moving grid.js out of the Grid package into Core, named media-layout.js.
* Renaming many of the Grid-centric things as "MediaLayout" or "Media" instead (largely automated).
* Adding a HOC to wrap a React Component to give it access to `mediaSize` and `mediaSpec` (making it so that components no longer need to know about React Context to get access to this information).

Looking at the components in styleguidist I can confirm that the Grid and Modal work the same as before.

After having implemented it and converted over to it I'm a bit mixed on using a HOC for this - I think I'd rather have a dedicated component for this (e.g. `<Layout>{({mediaSize}) => ...}</Layout>`).

I was actually thinking that the ideal version of this might be something that behave similar to the existing Grid components and also has the ability to something like `<Layout large={true}>...</Layout>` if you wanted to show something only at large layouts.

Additionally, we have support for layouts in style like so: `style={(mediaSize) => ...}` but I think I'd like to modify it so that we get access to the properties of the spec so you can do something like: `style={({marginSize}) => ({marginTop: marginSize})}` right now you have no way to get this info. Let me know what your thoughts are on these changes and I can make them in a follow-up PR.


jeresig:

Built with commit c92cadc02281761f9bce28123a9c2a76ce726f25

https://deploy-preview-114--wonder-blocks.netlify.com

packages/wonder-blocks-core/index.js
```diff
@@ -1,11 +1,15 @@
...
+    default as getClickableBehavior,
+} from "./util/get-clickable-behavior.js";
+export * from "./util/specs.js";
+export {MediaLayoutWrapper} from "./util/util.js";
 
-export {ClickableBehavior, Text, View, addStyle, getClickableBehavior};
-export type {AriaProps, ClickableHandlers, TextTag};
+import type {ClickableHandlers} from "./components/clickable-behavior.js";
+import type {AriaProps, TextTag, MediaSize, MediaSpec} from "./util/types.js";
+export type {AriaProps, ClickableHandlers, TextTag, MediaSize, MediaSpec};
```
kevinbarabash: It looks like `export type` doesn't support the shorthand yet.  Maybe someday in the future.

packages/wonder-blocks-core/util/specs.js
```diff
@@ -2,14 +2,14 @@
 
 import Spacing from "wonder-blocks-spacing";
 
-import type {GridSize, GridSpec} from "./types.js";
+import type {MediaSize, MediaSpec} from "./types.js";
 
-// All possible valid grid sizes
-export const VALID_GRID_SIZES: Array<GridSize> = ["small", "medium", "large"];
+// All possible valid media sizes
+export const VALID_MEDIA_SIZES: Array<MediaSize> = ["small", "medium", "large"];
```
kevinbarabash: nice

packages/wonder-blocks-core/util/util.js
```diff
@@ -60,3 +64,33 @@ export function processStyleList<T: Object>(
         className: css(...stylesheetStyles),
     };
 }
+
+export const mediaContextTypes = {
+    mediaSize: propTypes.string,
+    mediaSpec: propTypes.object,
+};
+
+export function MediaLayoutWrapper<Props: {}>(
```
kevinbarabash: Maybe add a TODO to change this to use the FAC pattern.

packages/wonder-blocks-grid/components/cell.js
```diff
@@ -15,21 +15,31 @@ type Props = {
...
     /** The styling to apply to the cell. */
     style?: any,
+    /**
+     * The size of the media layout being used. Populated by MediaLayoutWrapper.
+     * @ignore
+     */
+    mediaSize: MediaSize,
+    /**
+     * The current media layout spec being used. Populated by MediaLayoutWrapper.
+     * @ignore
```
kevinbarabash: Does this hide them in the generated docs?

packages/wonder-blocks-grid/docs.md
```diff
@@ -83,55 +83,36 @@ const cellStyles = [styles.cell, (size) => styles[size]];
...
 		</Row>
-	</Grid>
+	</MediaLayout>
 </View>;
 ```
 
 Grids are built using the following components:
 
-* [Grid](#grid-1): A Grid wraps all parts of a grid and tracks the browser viewport, toggling the layout of the grid based on viewport width changes. A Grid will hold the [Row](#row) components used for rendering the grid.
+* [MediaLayout](#medialayout): A MediaLayout wraps all parts of a grid and tracks the browser viewport, toggling the layout of the grid based on viewport width changes. A MediaLayout will likely hold almost the entire contents of the page. Rows can be direct children or distant descendants.
```
kevinbarabash: :+1:

packages/wonder-blocks-core/components/media-layout.js
```diff
@@ -0,0 +1,226 @@
+// @flow
```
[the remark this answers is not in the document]
jeresig: Yep - pretty much just copied and renamed.

packages/wonder-blocks-core/util/util.js
jeresig: Hmm - I could, although it's probably just going to be my next PR (and I'm not expecting anyone to use it in the meantime).

packages/wonder-blocks-grid/components/cell.js
jeresig: Yes! I find this to be convenient as it still explains what it is if you're reading the code but we don't need to describe it in the official docs as no one is really expected to use it.

PR #115, opened 2018-06-08
make wonder-blocks-core a dependency of wonder-blocks-spacing, add prop-types as a project-wide dep

kevinbarabash:
The addition of `prop-types` is b/c in a few places we were using it was being included in built files.  The webpack.config.js looks at package.json so if something's not explicitly listed but imported as a sub dependency of something else we don't mark it as external.  Adding an explicit dep for it to the root package.json avoid bundling prop-types in each of the packages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
   },
   "dependencies": {
     "aphrodite": "^1.2.5",
+    "prop-types": "^15.6.1",
     "react": "^15.3.2",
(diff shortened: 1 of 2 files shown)


jeresig:

Built with commit 0eae96616d8771670dedf17adef46498346ac6fc

https://deploy-preview-115--wonder-blocks.netlify.com

PR #116, opened 2018-06-11
set up lerna and publish script

kevinbarabash:


jeresig:

Built with commit e9d4e8cc7432be1d165e11cab75ebfc65deb929e

https://deploy-preview-116--wonder-blocks.netlify.com

kevinbarabash:
Curious that both travis-ci builds stalled.

package.json
```diff
@@ -45,6 +46,7 @@
     "flow-bin": "0.57.3",
     "glob": "^7.1.2",
     "jest": "^22.4.2",
+    "lerna": "^2.11.0",
```
MiguelCastillo: What do you think about adding `yarn` as a dev-dependency?  This gives us two things.  First, developers wont have to brew install yarn just to work on wonder-blocks.  Second, we can control what version of yarn is used, so if there is an issue in a particular version we wont have to worry about inconsistencies between dev environments.  We can add that as a separate PR too if you want to give it a try to make sure things work as expected.

kevinbarabash: I was planning to create a separate PR to install `yarn` as a dep.

somewhatabstract: We should consolidate our code and docs to specify `yarn` instead of `npm` too. No reason to have both.

package.json
```diff
@@ -12,13 +12,14 @@
     "flow": "flow",
     "gen-snapshot-tests": "node utils/gen-snapshot-tests.js && prettier --write packages/*/generated-snapshot.test.js",
     "jest": "SNAPSHOT_INLINE_APHRODITE=1 jest",
-    "test": "npm run lint && npm run gen-snapshot-tests && npm run flow && npm run jest",
+    "test": "npm run lint && npm run gen-snapshot-tests && npm run build:all && npm run flow && npm run jest",
```
somewhatabstract: `yarn build:all` and `yarn flow`. Let's standardize on `yarn` and drop all these `npm` refs.

PR #117, opened 2018-06-12
check in .vscode/settings.json

kevinbarabash:


jeresig:

Built with commit dbfb10619e8789e57d13475ba0bca7e8577d2565

https://deploy-preview-117--wonder-blocks.netlify.com

.vscode/settings.json
```diff
@@ -0,0 +1,12 @@
+{
+    "editor.formatOnSave": true,
```
jeresig: 🛑 Jeff made the fair point that onSave stuff should probably be up to us to specify, which is fair... so both this an the eslint one should probably get pulled.

.vscode/settings.json
```diff
@@ -0,0 +1,12 @@
...
+    "editor.renderWhitespace": "boundary",
+    "editor.rulers": [
+        80
+    ],
```
jeresig: 🤷‍♂️ I find both of these previous rules to be useful but maybe others wouldn't? We could just those, as well.

somewhatabstract: I  think that these are user preferences rather than workspace mandates.

If the ruler setting were used for enforcement of that width, then it should stay as it's a project coding style issue, but if it's just a UI nicety, definitely should be removed and left for user preference.

Also, we should consider having an `extensions.json` to specify recommended extensions for working on this project (a different PR, though).

.vscode/settings.json
somewhatabstract: Yeah, remove this.

.vscode/settings.json
```diff
@@ -0,0 +1,12 @@
...
+    "eslint.autoFixOnSave": true,
+    "eslint.enable": true,
+    "files.trimTrailingWhitespace": true,
+    "flow.useNPMPackagedFlow": true,
```
somewhatabstract: This is a great example of why we should have `extensions.json`, as that way we can recommend that the `flowtype` extension is installed.

PR #118, opened 2018-06-13
Add support for getting reference to modal host element when mounted in a modal launcher portal

somewhatabstract:
This change adds the method `maybeGetPortalMountedModalHostElement` to exports from the `wonder-blocks-modal` package. This is the first change for supporting the tooltip v1 work. Putting out a PR early so that other portal uses can benefit from this (such as the dropdown work).

These changes add a constant for the attribute name that identifies the modal launcher portal element so that it is more carefully controlled, and adds a method for walking the DOM to find the nearest one from a given element. This ensures it means to be a modal portal is encapsulated in the modal package, rather than leaking out all over the place.

The argument of the method is typed to match what `ReactDOM.findDOMNode` can return.

# Tests
There are some tests to validate the behaviour of the new method.


jeresig:

Built with commit d8e9c7e67cf125e45aff0003e3646e063bf8adca

https://deploy-preview-118--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-launcher-portal.js
```diff
@@ -29,7 +30,7 @@ export default class ModalLauncherPortal extends React.Component<Props> {
         // The data attribute is used in unit tests, to identify which
         // ancestors of `children` were created by `ModalLauncherPortal`.
         const destination = document.createElement("div");
-        destination.setAttribute("data-modal-launcher-portal", "");
+        destination.setAttribute(ModalLauncherPortalAttributeName, "");
```
somewhatabstract: Consolidating the attribute name into one place.

packages/wonder-blocks-modal/components/modal-launcher-portal.test.js
```diff
@@ -15,7 +16,7 @@ describe("ModalPortal", () => {
                 let parent = element.parentElement;
                 while (
                     parent &&
-                    (parent.hasAttribute("data-modal-launcher-portal") ||
+                    (parent.hasAttribute(ModalLauncherPortalAttributeName) ||
```
somewhatabstract: Consolidating the attribute name into one place.

packages/wonder-blocks-modal/index.js
```diff
@@ -3,5 +3,12 @@ import ModalLauncher from "./components/modal-launcher.js";
...
 import OneColumnModal from "./components/one-column-modal.js";
+import getNearestModalLauncherPortal from "./util/get-nearest-modal-launcher-portal.js";
 
-export {ModalLauncher, StandardModal, TwoColumnModal, OneColumnModal};
+export {
+    ModalLauncher,
+    StandardModal,
+    TwoColumnModal,
+    OneColumnModal,
+    getNearestModalLauncherPortal,
```
somewhatabstract: Adding the new method as an export

jeresig:

Built with commit 1aefa8822578c2284295e8544c9c74bf65704b56

https://deploy-preview-118--wonder-blocks.netlify.com

packages/wonder-blocks-modal/util/constants.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+
+/**
+ * The attribute used to identify a modal launcher portal.
+ */
+const ModalLauncherPortalAttributeName = "data-modal-launcher-portal";
+
+export {ModalLauncherPortalAttributeName};
```
somewhatabstract: This is now the only place that the attribute name is specified. So we won't have weird typo errors and such if we ever change this, and we now have something that links all these places to make it clear they are talking about the same thing.

packages/wonder-blocks-modal/util/get-nearest-modal-launcher-portal.js
```diff
@@ -0,0 +1,26 @@
...
+        candidateElement &&
+        !candidateElement.hasAttribute(ModalLauncherPortalAttributeName)
+    ) {
+        if (candidateElement === candidateElement.parentElement) {
+            return null;
+        }
+        candidateElement = candidateElement.parentElement;
+    }
+    return candidateElement || null;
+}
```
somewhatabstract: This is the new method that does the work. Based on the similar code already present in the modal tests. Nothing particularly complicated here.

packages/wonder-blocks-modal/util/get-nearest-modal-launcher-portal.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+export default function getNearestModalLauncherPortal(
+    element: null | Element | Text,
+) {
+    let candidateElement = element ? element.parentElement : null;
+    while (
+        candidateElement &&
+        !candidateElement.hasAttribute(ModalLauncherPortalAttributeName)
+    ) {
+        if (candidateElement === candidateElement.parentElement) {
```
somewhatabstract: This probably isn't needed. IIRC, some browsers would return `document.body` as its own parent, so I wanted to make sure we terminated in that case.

packages/wonder-blocks-modal/util/get-nearest-modal-launcher-portal.test.js
```diff
@@ -0,0 +1,81 @@
...
+import * as React from "react";
+import * as ReactDOM from "react-dom";
+import {mount} from "enzyme";
+
+import getNearestModalLauncherPortal from "./get-nearest-modal-launcher-portal.js";
+import ModalLauncher from "../components/modal-launcher.js";
+import StandardModal from "../components/standard-modal.js";
+
+describe("getNearestModalLauncherPortal", () => {
+    test("when candidate is null, returns null", () => {
```
somewhatabstract: Test case to ensure that we don't do anything bad if we're given a falsy element. Should probably rename to "when candidate is falsy, returns null".

packages/wonder-blocks-modal/util/get-nearest-modal-launcher-portal.test.js
```diff
@@ -0,0 +1,81 @@
...
+        const candidateElement = null;
+
+        // Act
+        const result = getNearestModalLauncherPortal(candidateElement);
+
+        // Assert
+        expect(result).toBe(null);
+    });
+
+    test("when candidate is not hosted in modal, returns null", () => {
```
somewhatabstract: Test case for when the element isn't in a modal.

packages/wonder-blocks-modal/util/get-nearest-modal-launcher-portal.test.js
```diff
@@ -0,0 +1,81 @@
...
+
+        // Act
+        const result = getNearestModalLauncherPortal(candidateElement);
+
+        // Assert
+        expect(result).toBe(null);
+        expect(candidateElement).not.toBe(null);
+    });
+
+    test("when candidate is hosted in modal, returns modal portal", (done) => {
```
somewhatabstract: Test case for when the candidate IS in a modal. :)

somewhatabstract:
@jeresig May as well wait. There needs to be a tweak so that this returns a slightly different element. I'll assign it back when I've fixed that.

somewhatabstract:
OK, I have updated this to now return the root element that hosts the modal, rather than the modal portal. My prototyping had identified that the tooltip needs to be hosted in here in order to appear above the modal, rather than just in the portal itself.

This latest change makes the tests more robust (like actually checking that we returned an element inside a modal portal, and not just any old element) and adds a test to specify the behavior that should occur in a non-portal-hosted modal.

Also renamed the method to be clearer that it might not return an element.

kevinbarabash:
> My prototyping had identified that the tooltip needs to be hosted in here in order to appear above the modal, rather than just in the portal itself.

This makes sense.  There shouldn't be any clipping issues with this b/c the root element of the modal portal has not clipping applied to it.

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {null | Element | Text} element The element whose ancestors are to be
+ * walked.
+ * @returns {Element | null} The nearest parent modal launcher portal or
+ * `null`.
+ */
+function maybeGetNextAncestorModalLauncherPortal(element: null | Element | Text) {
+    let candidateElement = element ? element.parentElement : null;
+    while (
+        candidateElement &&
+        !candidateElement.hasAttribute(ModalLauncherPortalAttributeName)
```
kevinbarabash: I thought you mentioned that in some browsers that eventually we might get to a situation where `candidateElement === candidateElement.parentElement`.  Was that actually not the case?

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.js
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * From a given element, finds its next ancestor that is a modal launcher portal
+ * element.
+ * @param {null | Element | Text} element The element whose ancestors are to be
+ * walked.
+ * @returns {Element | null} The nearest parent modal launcher portal or
+ * `null`.
+ */
+function maybeGetNextAncestorModalLauncherPortal(element: null | Element | Text) {
+    let candidateElement = element ? element.parentElement : null;
```
jeresig: 🤷‍♂️ It feels like the logic can be simplified a bit. Perhaps:

`let candidateElement = element && element.parentElement;`

and then later:

`return candidateElement;`

Does it particularly matter if null or undefined is returned?

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.js
```diff
@@ -0,0 +1,44 @@
+// @flow
+import {ModalLauncherPortalAttributeName} from "./constants.js";
+
+/**
+ * From a given element, finds its next ancestor that is a modal launcher portal
+ * element.
+ * @param {null | Element | Text} element The element whose ancestors are to be
```
jeresig: 🤷‍♂️ These comments are nice but I don't think we need to use the full `@param` & `@returns` syntax (we're not doing anything with it and I think the Flow types themselves are sufficient).

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.js
```diff
@@ -0,0 +1,44 @@
...
+ * walked.
+ * @returns {Element | null} The nearest parent modal launcher portal or
+ * `null`.
+ */
+function maybeGetNextAncestorModalLauncherPortal(element: null | Element | Text) {
```
jeresig: 🤷‍♂️ I feel like `element: ?(Element | Text)` might be slightly more idiomatic?

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.js
```diff
@@ -0,0 +1,44 @@
...
+export default function maybeGetPortalMountedModalHostElement(
+    element: null | Element | Text,
+) {
+    const modalLauncherPortal = maybeGetNextAncestorModalLauncherPortal(element);
+
+    if (modalLauncherPortal) {
+        // If we have a portal, then its first child will be the actual
+        // data-react-root that contains the modal, which is what we want to
+        // return.
+        return modalLauncherPortal.firstElementChild;
```
kevinbarabash: Is that actually true?  Is it not possible for it be the nth child?  Shouldn't `maybeGetNextAncestorModalLauncherPortal` already be return the element that with the `data-modal-launcher-portal` attribute set?

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.js
```diff
@@ -0,0 +1,44 @@
...
+    } else {
+        return null;
```
jeresig: 🤷‍♂️ Same comments for the above function make sense for this function, too - and if that's the case then this extra `else` is unnecessary.

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.js
somewhatabstract: After reading the docs on `parentElement` and `parentNode`, it is likely that I was confusing `parentElement` with `parentNode`. Based on re-reading the specs and some BrowserStack testing, I'm confident that we don't need that check.

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.js
somewhatabstract: `maybeGetNextAncestorModalLauncherPortal` gets the portal with that attribute set, but what we want is the node that is rendered into that portal. Hence this call to get the child. My testing showed that mounting as the last item in the portal itself still had the tooltip render below the modal and its backdrop, whereas rendering within the mounted data-react-root fixed that.

It is plausible that when fragments come with React 16, we might have more than one child, so I could have this always get the last element instead.

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.js
somewhatabstract: For the first point, definitely can change that. This code originally looked a little different and I didn't golf it down after it hit this stage.

Regarding the second point, let me check the usage. There was a reason why I had it returning null, but that may have since changed. Definitely seems like we should simplify, but I will make sure that anything relying on it returning `null` is fixed too.

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.js
somewhatabstract: It's the document extension in code. I type /** return and it adds almost all the boilerplate. Happy to get rid of it, but it's really nice in the intellisense of VSCode (at least I think so).

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.js
somewhatabstract: Sure.

somewhatabstract:
@jeresig For an example of usage, you can see my WIP tooltip work.

https://github.com/Khan/wonder-blocks/blob/tooltip/packages/wonder-blocks-tooltip/components/tooltip-portal.js#L119

It is very very likely that this call will move into a different component (I'm still prototyping), but the general principle remains. Given an anchor point, we want to know whether we're in a modal portal or not, so we use the anchor point and this method to find that out.

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.js
kevinbarabash: I wasn't looking at the tests carefully, I see now that we get the parent when we do the check for `data-modal-launcher-portal`.  I'm surprised that we have to render the tooltip into the first child instead of portal itself.

somewhatabstract: Yeah, if I don't then the tooltip appears below the modal and its backdrop.

somewhatabstract: I suspect it has a lot to do with stacking context

somewhatabstract:
OK @jeresig, I've addressed a couple of the comments you made.

@kevinbarabash Nothing functional changed here other than tidying the API as per John's comments

PR #119, opened 2018-06-13
Add ourselves a cheeky little logo

somewhatabstract:
Just a bit of fun. I had created a new Wonder Blocks emoji and thought it might be nice here. Added a bit of color to the README.md and the styleguide. Also updates the favicon for the demo app and adds one for the styleguide.

image

I will add some inline comments to explain changes.

(and totally understand if we don't use this, but at least we'll have it to refer to if a real/better logo comes along later)


styleguide.config.js
```diff
@@ -101,7 +102,23 @@ module.exports = {
             base: '"Lato", sans-serif',
         },
     },
+    title: "Wonder Blocks",
```
somewhatabstract: Removes the "Style Guide" off the end

styleguide.config.js
```diff
@@ -101,7 +102,23 @@ module.exports = {
...
     styles: {
+        StyleGuide: {
+            sidebar: {
+                width: "initial",
```
somewhatabstract: Removes the "200px" limitation. Even just with "Wonder Blocks" as title, the sidebar wanted to be a touch wider to not wrap the text.

jeresig:

Built with commit 297f4e271b274a341878e63c824013253a277814

https://deploy-preview-119--wonder-blocks.netlify.com

styleguide.config.js
```diff
@@ -101,7 +102,23 @@ module.exports = {
...
+        },
+        Logo: {
+            logo: {
+                background: "no-repeat url('logo.png')",
+                backgroundPosition: "0 0.25em",
+                paddingLeft: 40,
+                minHeight: "calc(32px + 0.5em)",
+                display: "flex",
+                alignItems: "center",
+            },
```
somewhatabstract: This adds the logo to the title, gives it some space, and also ensures the text looks half decent alongside it. I did some testing with font-size changes and it seems to cope correctly.

styleguide.config.js
```diff
@@ -112,7 +129,7 @@ module.exports = {
                 overflow: "auto",
 
                 // Allow the preview content to grow to the full size, using
-                // absolute-positioning, and left/right/top/bottom all set to 0.
+                // absolute--positioning, and left/right/top/bottom all set to 0.
```
somewhatabstract: Er...I didn't type anything here, so this change is a bit confusing. I'll try removing it.

styleguide.config.js
somewhatabstract: I assume this is a netlify config thing. I don't have access to that so someone else would need to look into it.

kevinbarabash: @somewhatabstract it should work as long as the icon gets copied to the `docs` folder.

jeresig: 🛑 Yeah, you should be able to update `build:styleguidist` in package.json to also copy these files over.

🛑 Also, would it be possible to put the two images into a folder named something like `static` and then copy that whole folder over? Would help to avoid cluttering the root directory.

somewhatabstract: Ah, yeah, I can definitely do those things :)

somewhatabstract: Well, I worked it out with only some config changes and a `static` dir. Now it works like a charm! :grin:

PR #120, opened 2018-06-13
prefix all wonder-blocks-* packages with @khanacademy

kevinbarabash:
Need to do this before we can publish anything.

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "wonder-blocks-demo",
+  "name": "@khanacademy/wonder-blocks-demo",
   "version": "0.1.0",
   "private": true,
   "dependencies": {
@@ -8,16 +8,16 @@
     "react-dom": "^16.3.2",
     "react-router-dom": "^4.2.2",
     "react-scripts": "1.1.4",
-    "wonder-blocks-button": "../packages/wonder-blocks-button",
-    "wonder-blocks-color": "../packages/wonder-blocks-color",
-    "wonder-blocks-core": "../packages/wonder-blocks-core",
-    "wonder-blocks-grid": "../packages/wonder-blocks-grid",
-    "wonder-blocks-icon-button": "../packages/wonder-blocks-icon-button",
-    "wonder-blocks-link": "../packages/wonder-blocks-link",
-    "wonder-blocks-modal": "../packages/wonder-blocks-modal",
-    "wonder-blocks-progress-spinner": "../packages/wonder-blocks-progress-spinner",
-    "wonder-blocks-spacing": "../packages/wonder-blocks-spacing",
-    "wonder-blocks-typography": "../packages/wonder-blocks-typography"
+    "@khanacademy/wonder-blocks-button": "0.0.1",
+    "@khanacademy/wonder-blocks-color": "0.0.1",
+    "@khanacademy/wonder-blocks-core": "0.0.1",
+    "@khanacademy/wonder-blocks-grid": "0.0.1",
+    "@khanacademy/wonder-blocks-icon-button": "0.0.1",
+    "@khanacademy/wonder-blocks-link": "0.0.1",
+    "@khanacademy/wonder-blocks-modal": "0.0.1",
+    "@khanacademy/wonder-blocks-progress-spinner": "0.0.1",
+    "@khanacademy/wonder-blocks-spacing": "0.0.1",
+    "@khanacademy/wonder-blocks-typography": "0.0.2"
   },
   "scripts": {
     "start": "react-scripts start",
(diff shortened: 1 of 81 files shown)


packages/wonder-blocks-button/package.json
```diff
@@ -1,8 +1,11 @@
 {
   "private": true,
-  "name": "wonder-blocks-button",
+  "name": "@khanacademy/wonder-blocks-button",
   "version": "0.0.1",
   "design": "v1",
+  "publishConfig": {
+    "access": "public"
+  },
```
kevinbarabash: This is needed so that the packages we publish are public.

jeresig:

Built with commit 46a0310d605d47034dcbb8399a62ae60dea6d15b

https://deploy-preview-120--wonder-blocks.netlify.com

demo/package.json
```diff
@@ -8,16 +8,16 @@
...
-    "wonder-blocks-color": "../packages/wonder-blocks-color",
-    "wonder-blocks-core": "../packages/wonder-blocks-core",
-    "wonder-blocks-grid": "../packages/wonder-blocks-grid",
-    "wonder-blocks-icon-button": "../packages/wonder-blocks-icon-button",
-    "wonder-blocks-link": "../packages/wonder-blocks-link",
-    "wonder-blocks-modal": "../packages/wonder-blocks-modal",
-    "wonder-blocks-progress-spinner": "../packages/wonder-blocks-progress-spinner",
-    "wonder-blocks-spacing": "../packages/wonder-blocks-spacing",
-    "wonder-blocks-typography": "../packages/wonder-blocks-typography"
+    "@khanacademy/wonder-blocks-button": "0.0.1",
```
kevinbarabash: For this to work we'd need to add the `demo` folder to the list of `workspaces` in the root package.json but that breaks the test runner probably b/c `react@16` and/or `react-scripts` are hoisted up a level.  I need to update the demo to not use `react-scripts` and to use the same version of `react` but that'll have to be in a separate PR.

packages/wonder-blocks-button/package.json
kevinbarabash: It avoids having to re-authenticate which is impossible when running `lerna`.  You still need to do `npm login` before running `yarn run publish`.

packages/wonder-blocks-spacing/package.json
```diff
@@ -1,19 +1,22 @@
 {
   "private": true,
-  "name": "wonder-blocks-spacing",
+  "name": "@khanacademy/wonder-blocks-spacing",
   "version": "0.0.1",
   "design": "v1",
+  "publishConfig": {
```
jeresig: 🤷‍♂️ What does it mean when a package has a public publishConfig but is marked as `"private": true`?

demo/package.json
jeresig: Oof, that's annoying!

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -3,10 +3,14 @@ import React from "react";
 import {StyleSheet} from "aphrodite";
 import {Link} from "react-router-dom";
 
-import {LabelLarge, LabelSmall} from "wonder-blocks-typography";
-import Color, {SemanticColor, mix, fade} from "wonder-blocks-color";
-import {addStyle} from "wonder-blocks-core";
-import type {ClickableHandlers} from "wonder-blocks-core";
+import {LabelLarge, LabelSmall} from "@khanacademy/wonder-blocks-typography";
```
jeresig: What commands did you run to do these replacements? Might be useful for the future!

kevinbarabash: I did a simple find and replace.  I replaced `"wonder-blocks-` with `"@khanacademy/wonder-blocks-`.

packages/wonder-blocks-spacing/package.json
kevinbarabash: Yeah... the terminology is a little wonky. `"private": true` prevents publishing completely, `"publishConfig"` sets specific settings at publish time only, in this case the "access" setting does the following:
> When publishing scoped packages, the access level defaults to restricted. If you want your scoped package to be publicly viewable (and installable) set --access=public. The only valid values for access are public and restricted. Unscoped packages always have an access level of public.

kevinbarabash:
> Seeing all these @khanacademy makes me wish we had @khan, but so be it :)

Part of me wishes we used `@khan` too b/c then it would go with our repo name.  We could potentially change our npmjs.org username to `khan` as well since it looks like someone is squatting on that username.

MiguelCastillo:
the idea to try to claim `@khan` has been bounced around...  someone is just squatting it.  I am happy to start the process :)

https://www.npmjs.com/org/khan
https://www.npmjs.com/policies/disputes

kevinbarabash:
I'll change `@khanacademy` to `@khan` before merging.

kevinbarabash:
I changed my mind... I'm going stick with the `@khanacademy` prefix for now since we're already publishing packages under that prefix.

PR #121, opened 2018-06-14
Avoid recursive imports

kevinbarabash:
We had an issue where `yarn run watch:all` resulting in an ever growing dist/index.js for `wonder-blocks-core`.  This PR fixes that by ensuring that `wonder-blocks-core` has the prop dependencies specified and that internal dependencies are imported directly.  This fix will also require us to publish `wonder-blocks-spacing` which is a dependency of `wonder-blocks-core`.

### DIFF
--- a/packages/wonder-blocks-core/components/media-layout.js
+++ b/packages/wonder-blocks-core/components/media-layout.js
@@ -1,6 +1,6 @@
 // @flow
 import * as React from "react";
-import {View} from "@khanacademy/wonder-blocks-core";
+import View from "./view.js";
 
 import {MEDIA_DEFAULT_SPEC, VALID_MEDIA_SIZES} from "../util/specs.js";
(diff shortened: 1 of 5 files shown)


packages/wonder-blocks-core/components/media-layout.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 import * as React from "react";
-import {View} from "@khanacademy/wonder-blocks-core";
+import View from "./view.js";
```
jeresig: 🤷‍♂️ Do we have a way to lint against this? It feels like this is going to be an ongoing issue!

jeresig:

Built with commit e28f318da1597eecf2c70bd46b1c54cb8c70b63c

https://deploy-preview-121--wonder-blocks.netlify.com

PR #124, opened 2018-06-14
WB-119: Make snapshots include all docs, not just root

somewhatabstract:
Jira: [WB-119](https://khanacademy.atlassian.net/browse/WB-119)

# Code Changes
## 1st commit
This should be reviewed.
[Review commit](https://github.com/Khan/wonder-blocks/pull/124/commits/c57d2c9c91673165cdfd4aed44e520a1bb2e3e7c)
This changes the snapshot test generation script so that it collates examples and file paths from all our docs. There is some extensive refactoring of the main script so that I could reuse various bits and keep methods easy to follow.

## 2nd commit
This should be reviewed.
[Review commit](https://github.com/Khan/wonder-blocks/pull/124/commits/c8280b4378ae7db4e962c8e6d271f584ec0d17cc)
This also then updates all the broken markdown that was referencing files from its own source so that we don't have to munge require paths to be relative from the right locations (styleguidist still works).

# Auto-generated Changes
## 3rd commit
This is autogenerated and probably shouldn't be reviewed in great detail.
[Review commit](https://github.com/Khan/wonder-blocks/pull/124/commits/f0483635b17964260a25ceb9aac8a1455507fabd)
We have generated new snapshot tests.

## 4th commit
This is autogenerated and doesn't need direct review.
[Review commit](https://github.com/Khan/wonder-blocks/pull/124/commits/e39f733a48d8075b93f2ca63c84bed6e9404e8ee)
We have generated new snapshots so that tests pass.


jeresig:

Built with commit dde278a0d083b62541c0c159297039e44d98b75a

https://deploy-preview-124--wonder-blocks.netlify.com

utils/gen-snapshot-tests.js
```diff
@@ -27,12 +27,13 @@ const ignored = [
...
 
-function generateTestFile(root, examples, componentFileMap) {
+function generateTestFile(root, {examples, componentFileMap}) {
     const lines = [
         "// @flow",
         "// This file is auto-generated by gen-snapshot-tests.js",
         "// Do not edit this file.  To make changes to these snapshot tests",
-        `// edit ${root}/docs.md and run \`npm run gen-snapshot-tests\`.`,
+        `// edit ${root}/docs.md and other documentation files in the package,`,
+        "// then run `npm run gen-snapshot-tests`.",
```
somewhatabstract: Updated the comment to reflect what now happens (should probably update this to yarn too, but I don't know that I want to do all that rebasing and such. Maybe as a future edit).

utils/gen-snapshot-tests.js
```diff
@@ -95,41 +96,100 @@ function generateTestFile(root, examples, componentFileMap) {
...
     const examples = tokens
         .filter((token) => token.type === "code")
         .map((token) => token.text);
-    if (examples.length === 0) {
-        // eslint-disable-next-line no-console
-        console.warn(`no examples for section ${section.name}`);
-        continue;
+
+    return examples;
+}
```
somewhatabstract: Helper function so we can reuse it. All it does is grab examples. Also encapsulates checking that the path is valid, just for ease of use.

utils/gen-snapshot-tests.js
```diff
@@ -95,41 +96,100 @@ function generateTestFile(root, examples, componentFileMap) {
...
+        const componentDoc = path.join(
+            path.dirname(file),
+            path.basename(file, path.extname(file)),
+        ) + ".md";
+
+        collation.examples = collation.examples.concat(
+            readExamplesFromDocument(componentDoc),
+        );
+    }
+}
```
somewhatabstract: This was the content of the old `glob` callback. It now takes a collation object and reads examples from each components docs, if they exist. We only collate once since we don't want the examples repeated.

utils/gen-snapshot-tests.js
```diff
@@ -95,41 +96,100 @@ function generateTestFile(root, examples, componentFileMap) {
...
+
+function extractExamplesAndComponentFiles(section, collation) {
```
somewhatabstract: This function is the bulk of the old generation method.

utils/gen-snapshot-tests.js
```diff
@@ -95,41 +96,100 @@ function generateTestFile(root, examples, componentFileMap) {
...
+    // Add any examples fromt he section content.
+    collation.examples = collation.examples.concat(
+        readExamplesFromDocument(section.content),
+    );
+
+    // Then process any component files and add their stuff too.
+    if (section.components) {
+        const componentGlobs = typeof section.components === "string"
+            ? [section.components]
+            : section.components;
```
somewhatabstract: Just normalizing the components value since it can be a single string or an array of strings.

utils/gen-snapshot-tests.js
```diff
@@ -95,41 +96,100 @@ function generateTestFile(root, examples, componentFileMap) {
...
+        for (const componentGlob of componentGlobs) {
+            const files = glob.sync(componentGlob, {ignore: ignored});
```
somewhatabstract: Synchronous `glob` call instead of async

utils/gen-snapshot-tests.js
```diff
@@ -95,41 +96,100 @@ function generateTestFile(root, examples, componentFileMap) {
...
-                        src.match(/export default class \s*(\w+)/) ||
-                        src.match(/export default \s*\w+\(\s*(\w+)/);
-                    componentFileMap[match[1]] = file;
-                }
-
-                generateTestFile(root, examples, componentFileMap);
-            });
-        } else {
-            generateTestFile(root, examples);
+    if (section.sections) {
```
somewhatabstract: This is the new bit to support situations like how the modal docs use sections to specify examples.

utils/gen-snapshot-tests.js
kevinbarabash: Nice!

utils/gen-snapshot-tests.js
```diff
@@ -95,41 +96,100 @@ function generateTestFile(root, examples, componentFileMap) {
...
-        continue;
+
+    return examples;
+}
+
+function extractExamplesAndComponentsForFile(file, collation) {
+    const src = fs.readFileSync(file, "utf8");
+    const match =
+        src.match(/export default class \s*(\w+)/) ||
+        src.match(/export default \s*\w+\(\s*(\w+)/);
```
kevinbarabash: This is to support things that aren't classes?

utils/gen-snapshot-tests.js
```diff
@@ -95,41 +96,100 @@ function generateTestFile(root, examples, componentFileMap) {
...
         .map((token) => token.text);
-    if (examples.length === 0) {
-        // eslint-disable-next-line no-console
-        console.warn(`no examples for section ${section.name}`);
-        continue;
+
+    return examples;
+}
+
+function extractExamplesAndComponentsForFile(file, collation) {
```
kevinbarabash: Can you document what `file` should be passed in.  It looks a like a .js file that exports a component.

utils/gen-snapshot-tests.js
```diff
@@ -95,41 +96,100 @@ function generateTestFile(root, examples, componentFileMap) {
...
+        ) + ".md";
+
+        collation.examples = collation.examples.concat(
+            readExamplesFromDocument(componentDoc),
+        );
+    }
+}
+
+function extractExamplesAndComponentFiles(section, collation) {
+    // Add any examples fromt he section content.
```
kevinbarabash: `from the section`

utils/gen-snapshot-tests.js
```diff
@@ -95,41 +96,100 @@ function generateTestFile(root, examples, componentFileMap) {
...
+        return;
+    }
+
+    // Now, let's collate all our examples and the components for the section
+    // they're in.
+    const collation = {
+        examples: [],
+        componentFileMap: {},
+    };
+    extractExamplesAndComponentFiles(section, collation);
```
kevinbarabash: Could `extractExamplesAndComponentFiles` return a `collation` instead of mutating the arguments.

utils/gen-snapshot-tests.js
```diff
@@ -95,41 +96,100 @@ function generateTestFile(root, examples, componentFileMap) {
...
-                generateTestFile(root, examples, componentFileMap);
-            });
-        } else {
-            generateTestFile(root, examples);
+    if (section.sections) {
+        // Repeat all this for each sub-section.
+        for (const childSection of section.sections) {
+            extractExamplesAndComponentFiles(
+                childSection,
+                collation,
```
kevinbarabash: This is awesome!  This should handle an level of nesting we do with sections and subsections.  Thanks for doing this.

utils/gen-snapshot-tests.js
somewhatabstract: Yeah, that was already there in the old code. It is grabbing things like our stuff that gets wrapped by MediaLayout (FixedCell, for example).

utils/gen-snapshot-tests.js
somewhatabstract: Ah yes, I can add some docs.

utils/gen-snapshot-tests.js
somewhatabstract: Great catch! My chubby hammy fisters must've messed that up.

utils/gen-snapshot-tests.js
somewhatabstract: I had played around with that. We could easily change to that model, i think. I'll take a look.

utils/gen-snapshot-tests.js
kevinbarabash: right, I remember now... man my memory is bad 😅

somewhatabstract:
OK, @kevinbarabash, I have made some changes:

1. Fixed spelling in that comment
1. Updated the main generated file comment to refer to yarn and try to conform to our 80 char width commentary standard
1. Updated to return the collation object rather than mutating things in place
1. Merged with master and updated modal snapshots to cater for close button moving
1. Added docs

somewhatabstract:
Sorry for the commit mess around. Did a rebase when I should've done something else and got things all over the place. Anyway, it's all good now :smile:

kevinbarabash:
@somewhatabstract are these commits safe to commit as is or should I squash them?

somewhatabstract:
@kevinbarabash All good now. I would commit as-is that way we have the nice separation (which is what I was striving for but somehow got all mixed up). 

We are all good now though. Triple checked. So assuming everything passes CI and you like the changes, we're good to go.

somewhatabstract:
I've updated the links in the PR discussion to match the commits so that they can at least be found if we want to revisit this for some reason. :)

kevinbarabash:
I've merged them via rebasing.

PR #125, opened 2018-06-14
WB-194: Make sure the close button is above the header and content

somewhatabstract:
Jira: [WB-194](https://khanacademy.atlassian.net/browse/WB-194)

This moves the close button to render above the title and content so that it is accessible via pointer devices.

This also refactors the close button into a method so that it's a little easier to read the code.

# Test Plan
I checked the styleguidist examples in both local and master.

### DIFF
--- a/packages/wonder-blocks-modal/components/modal-panel.js
+++ b/packages/wonder-blocks-modal/components/modal-panel.js
@@ -58,6 +58,39 @@ export default class ModalPanel extends React.Component<Props> {
         onClickCloseButton: () => {},
     };
 
+    maybeRenderCloseButton() {
+        const {
+            showCloseButton,
+            onClickCloseButton,
+            color,
+            titleBar,
+        } = this.props;
+
+        if (!showCloseButton) {
+            return null;
+        }
+
+        // Figure out the background color that'll be behind the close button
+        // If a titlebar is specified then we use that, otherwise we use the
+        // default background color.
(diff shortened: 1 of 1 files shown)


jeresig:

Built with commit 42e4a4780b3c7484f52aea44e692fc76c78aba77

https://deploy-preview-125--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-panel.js
```diff
@@ -102,34 +134,13 @@ export default class ModalPanel extends React.Component<Props> {
...
-                            color={
-                                topBackgroundColor === "dark" ? "light" : "dark"
-                            }
-                            onClick={onClickCloseButton}
-                        />
-                    </View>
-                )}
                 {titleBar}
                 {mainContent}
+                {this.maybeRenderCloseButton()}
```
kevinbarabash: Thanks for extracting that.  This is an interesting issue.  Even if we wrote a test with enzyme it wouldn't have picked this up b/c enzyme could've simulated an click event and sent it directly to the close button itself whereas a user would still be unable to click the button.

somewhatabstract: Yeah, I don't know how we could automate a test for something like this without a proper native test app (like something that runs natively and manipulates the OS pointer directly to simulate a real user).

PR #126, opened 2018-06-15
rename CircularSpinner to ProgressSpinner

kevinbarabash:
There's only one progress spinner so it should be called `ProgressSpinner` and be a default export since the package only exports one thing.


jeresig:

Built with commit 69ab465a7689663d6b03d8524449570fb8128d80

https://deploy-preview-126--wonder-blocks.netlify.com

jeresig:
Erm, I disagree. When we talked about this with the design team before there were going to be a bunch of other progress indicators, not just the spinner. Circular was just the first one. I think we should leave it open to add more later.

kevinbarabash:
I didn't realize that.

PR #130, opened 2018-06-17
Support manual testing of internal components and add versioning to prod docs

somewhatabstract:
This PR adds some pre-processing to production documentation. 

The reasons for these changes are:
1. More test coverage with auto-generated documentation-derived snapshots of private components when we want them.
1. Benefit from manual testing of isolated inner components. Useful when coding and when debugging issues.
1. Documentation giving some indication of versions to help know what is ready for use and what is released under what versions.

The changes are documented inline, but here is a summary.

1. Adds a new script `utils/gen-styleguide-prod-config.js`
1. Adds a corresponding `yarn` task, `yarn run gen-styleguide-prod-config` which executes this script and runs prettier on the output (just so we can read it and check it once in a while with a diff to the original config)
1. Adds the generated config to the list of ignored files
1. Updates the `build:styleguidist` step to use the new script prior to build
1. Adds a new `yarn start:prod` task that starts a styleguidist server on port 6061 so that you can see what the prod docs would be (useful if you want to check your new dev docs really got hidden, for example)
1. Adds the concept of "private" documentation. This is for sections of documentation that have developer use (like being able to manually test unexported components in isolation from their parent components), but we don't want in our main production documentation.
1. Adds package version information to our production documentation. This scans the root sections (and assumes each corresponds to a package...whcih I think is ok). It then tries to identify the path of the package, load its package.json and grab some info. Examples below, but before reviewing those or the code read the next section.

## Considerations for review
1. Should the design version go in the section name instead of the description, as in "Tooltip v1"?
1. Should the prototype state be appended to the section name instead of the description, such as "Grid (prototype)"?
1. Might look nicer to drop the `Package: ` from the front of the package name.

Note that in both of these cases, appending to the section name will mean that direct links to those component sections would break whenever we regen'd docs with a changed design spec version or took away a prototype state.

# Examples
## Output of script
The script outputs some helpful things to the console as follows:
```
16:20 $ yarn run gen-styleguide-prod-config
yarn run v1.7.0
$ node utils/gen-styleguide-prod-config.js && prettier --write styleguide.prod.config.js
Removing private section Wonder Blocks Documentation
Adding package information for Typography: {"name":"@khanacademy/wonder-blocks-typography","lastRelease":"1.0.1","design":"v2"}
Adding package information for Color: {"name":"@khanacademy/wonder-blocks-color","lastRelease":"1.0.1","design":"v2"}
Adding package information for Spacing: {"name":"@khanacademy/wonder-blocks-spacing","lastRelease":"1.0.0","design":"v1"}
Adding package information for Grid: {"name":"@khanacademy/wonder-blocks-grid","lastRelease":"0.0.2","design":"v1","prototype":true}
Adding package information for Core: {"name":"@khanacademy/wonder-blocks-core","lastRelease":"1.0.1","design":"v1"}
Adding package information for ProgressSpinner: {"name":"@khanacademy/wonder-blocks-progress-spinner","lastRelease":"0.0.2","design":"v1","prototype":true}
Adding package information for Modal: {"name":"@khanacademy/wonder-blocks-modal","lastRelease":"0.0.2","design":"v1","prototype":true}
Adding package information for Button: {"name":"@khanacademy/wonder-blocks-button","lastRelease":"0.0.2","design":"v1","prototype":true}
Adding package information for Link: {"name":"@khanacademy/wonder-blocks-link","lastRelease":"0.0.2","design":"v1","prototype":true}
Adding package information for IconButton: {"name":"@khanacademy/wonder-blocks-icon-button","lastRelease":"0.0.2","design":"v1","prototype":true}
styleguide.prod.config.js 321ms
✨  Done in 1.34s.
```
The script will also output an alert if a non-private root level section has no version info to output.

## Package Information
The information added to the docs about each package is pre-pended to the section's `description` property, which we don't currently use but might, hence prepending rather than replacing.

The information data included are:
1. Prototype package indication (package.json `private`prop; only present when `true`)
1. Package name (package.json `name` prop)
1. Design spec version (package.json `design` prop)
1. Last released version (package.json `version` prop)

Here are some examples of how that looks when rendered in the docs.

image

image
Note here that I removed (Prototype) from the Grid section title as we're now grabbing such info from the package.json.


.codecov.yml
```diff
@@ -19,5 +19,5 @@ comment:
   require_head: yes
 
 ignore:
-  - "gen-snapshot-tests.js"
+  - "utils/*.js"
```
somewhatabstract: I think this is the right thing to do, unless we're hoping to cover the tools too.

package.json
```diff
@@ -11,12 +11,14 @@
     "watch:all": "webpack -w",
     "flow": "flow",
     "gen-snapshot-tests": "node utils/gen-snapshot-tests.js && prettier --write packages/*/generated-snapshot.test.js",
+    "gen-styleguide-prod-config": "node utils/gen-styleguide-prod-config.js && prettier --write styleguide.prod.config.js",
```
somewhatabstract: New task to run just the config transformation.

jeresig:

Built with commit f8331e2abb672edf159aa03bcbee41488607283c

https://deploy-preview-130--wonder-blocks.netlify.com

package.json
```diff
@@ -11,12 +11,14 @@
...
     "jest": "SNAPSHOT_INLINE_APHRODITE=1 jest",
     "test": "yarn run lint && yarn run gen-snapshot-tests && yarn run build:all && yarn run flow && yarn run jest",
     "coverage": "SNAPSHOT_INLINE_APHRODITE=1 jest --coverage",
     "styleguidist": "styleguidist server",
     "start": "yarn run styleguidist",
-    "build:styleguidist": "styleguidist build",
+    "start:prod": "yarn run build:styleguidist && yarn run styleguidist --config styleguide.prod.config.js",
```
somewhatabstract: New task to run a prod-specific server on port 6061.

package.json
```diff
@@ -11,12 +11,14 @@
...
+    "build:styleguidist": "yarn run gen-styleguide-prod-config && styleguidist build --config styleguide.prod.config.js",
```
somewhatabstract: Updated task to build the docs. Now starts with the `gen-styleguide-prod-config` task, then does the build using the transformed config instead of the original.

styleguide.config.js
```diff
@@ -5,6 +5,11 @@ module.exports = {
     styleguideDir: "docs",
     assetsDir: "static",
     sections: [
+        {
+            private: true,
+            name: "Wonder Blocks Documentation",
+            description: "This section only appears to developers. If you want to include content that only appears in the development server, add `private: true` to the section configuration in `styleguide.config.js`."
+        },
```
somewhatabstract: An example `private` section.

utils/gen-snapshot-tests.js
```diff
@@ -12,19 +12,11 @@
 const marked = require("marked");
 const fs = require("fs");
 const path = require("path");
-const glob = require("glob");
 const babylon = require("babylon");
 
-const styleguideConfig = require("../styleguide.config.js");
+const {getComponentFilesFromSection} =  require("./styleguidist-config-utils.js");
```
somewhatabstract: I created a file to contain some shared code and updated this script to use that shared code.

utils/gen-styleguide-prod-config.js
```diff
@@ -0,0 +1,188 @@
...
+            lastRelease: pkgJson.version,
+            design: pkgJson.design,
+            prototype: pkgJson.private,
+        };
+    } else {
+        return null;
+    }
+}
+
+function genStyleguideProdConfig() {
```
somewhatabstract: This is the function that orchestrates the work.

utils/styleguidist-config-utils.js
```diff
@@ -0,0 +1,32 @@
...
+    } else if (typeof section.compontents === "function") {
+        return section.components();
+    } else {
+        return section.components || [];
+    }
+}
+
+module.exports = {
+    getComponentFilesFromSection,
+};
```
somewhatabstract: This is extracted from the `gen-snapshot-tests` script and updated to properly follow the styleguidist configuration spec, which states that `components` is:

> Type: String, Function or Array, default: src/components/**/*.{js,jsx,ts,tsx}
>
 > when String: a glob pattern that matches all your component modules.
 > when Function: a function that returns an array of module paths.
 > when Array: an array of module paths.

utils/gen-styleguide-prod-config.js
```diff
@@ -0,0 +1,188 @@
...
+        if (files.length) {
+            return getPackageDirFromPath(files[0]);
+        };
+    }
+
+    // We get here, we got nothing.
+    return null;
+}
+
+function getVersionInfoForSection(section) {
```
jeresig: Hmm - this gets a bit more than just the version info - would be good to clarify this (and add a comment explaining it).

utils/gen-styleguide-prod-config.js
jeresig: This doesn't really need to be a function, does it? It can just operate at the top-level. Might make things a bit clearer.

package.json
jeresig: 🛑 So I do have a worry: this means that we can no longer see changes to prototype components when we do a PR to Github. Unfortunately, the same commands are used to build both the PR and the final release so I'm not sure that adding an extra command will help. It's just a bummer as we can no longer share prototypes with each other for testing purposes.

somewhatabstract:
@jeresig I think I have an idea around that. Let me have a tinker and I'll put this back to you. Thanks for the review comments.

utils/gen-styleguide-prod-config.js
somewhatabstract: I find lots of code outside of a function to be a little strange and so wrapped it for clarity, but in the end, either way will work.

utils/gen-styleguide-prod-config.js
```diff
@@ -0,0 +1,188 @@
...
+            case "serverPort":
+                // Skip. We do a different port for prod so that prod and dev servers can coexist, if necessary.
+                break;
+
+            default:
+                lines.push(`    ${entry[0]}: ${JSON.stringify(entry[1])},`);
+                break;
+        }
+    });
+    // Let's specify the production server port.
```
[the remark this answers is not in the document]
somewhatabstract: I didn't do that because the server port setting doesn't exist by default (we use the default port) so that case won't fire necessarily. We want to set this regardless of whether it already exists in the config.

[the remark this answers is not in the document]
somewhatabstract: I'll update the commentary accordingly.

utils/gen-styleguide-prod-config.js
```diff
@@ -0,0 +1,188 @@
...
+        } else {
+            // And now, we move into the chilren of the section and repeat.
+            sections[sectionIndex].sections =
+                removePrivateSections(sections[sectionIndex].sections);
+        }
+    }
+    return sections;
+}
+
+function maybeGetPackagePathForSection(section) {
```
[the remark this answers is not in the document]
somewhatabstract: It is attempting to get a wonder blocks package folder for a given section of documentation. If it can, it returns it; if it cannot, it returns null.

somewhatabstract:
@jeresig This is all working and ready for review again.

PR #131, opened 2018-06-18
Make default coverage task open the results in the browser

somewhatabstract:
`yarn coverage` or `yarn run coverage` will now run the coverage generation and then open the index.html to see the results.

Travis now uses `build:coverage`, which does what the old `coverage` used to do.

I made this change because when I run code coverage explicitly, I want to see the results and see what lines did and did not get covered. I feel this is the default case for devs explicitly running coverage and certainly the most useful when writing tests.

Since the data is already generated by jest when running coverage, this change adds an `open` command on the `index.html` file after building the generation.


jeresig:

Built with commit f3390bc81b7366bd76ef16cffd2cca59d5265206

https://deploy-preview-131--wonder-blocks.netlify.com

kevinbarabash:
Can't wait to try this out.

PR #132, opened 2018-06-19
Add tests for getClickableBehavior

somewhatabstract:


packages/wonder-blocks-core/util/get-clickable-behavior.js
```diff
@@ -28,7 +28,7 @@ export default function getClickableBehavior(
     /**
      * router object added to the React context object by react-router-dom.
      */
-    router: any,
+    router?: any,
```
somewhatabstract: No point in the first two args being optional if this one isn't.

jeresig:

Built with commit 852ae44970f2aff8008a4d44c6440caaea2ec5c3

https://deploy-preview-132--wonder-blocks.netlify.com

packages/wonder-blocks-core/util/get-clickable-behavior.test.js
```diff
@@ -0,0 +1,97 @@
...
+                    const url = "/prep/lsat";
+                    const clientNav = undefined;
+                    const router = undefined;
+                    const expectation = ClickableBehavior;
+
+                    // Act
+                    const result = getClickableBehavior(url, clientNav, router);
+
+                    // Assert
+                    expect(result).toBe(expectation);
```
kevinbarabash: I feel like `expect(result).toBe(ClickableBehavior)` is a clearer assertion b/c you don't have to look for the definition of `expectation` to know what it's asserting.

packages/wonder-blocks-core/util/get-clickable-behavior.test.js
```diff
@@ -0,0 +1,97 @@
...
+                    const url = "/prep/lsat";
+                    const clientNav = undefined;
+                    const router = <MemoryRouter history={[]} />;
+                    const expectation = "withRouter(ClickableBehavior)";
+
+                    // Act
+                    const result = getClickableBehavior(url, clientNav, router);
+
+                    // Assert
+                    expect(result.displayName).toBe(expectation);
```
kevinbarabash: This is a really cool use of `displayName`.

packages/wonder-blocks-core/util/get-clickable-behavior.test.js
```diff
@@ -0,0 +1,97 @@
...
+
+            // Act
+            const result = getClickableBehavior(url, clientNav, router);
+
+            // Assert
+            expect(result).toBe(expectation);
+        });
+
+        describe("Internal URL", () => {
+            describe("Client navigation is undefined", () => {
```
kevinbarabash: Why differentiate between `clientNav` undefined vs. false?

somewhatabstract: Because the code does. When `clientNav` isn't specified, it defers to the existence of a `router`. When it's specified, it ignores the router and makes a decision base don the `clientNav` value.

IF that's not the behaviour we want, then the method will need updating. I just captured its current behaviour.

packages/wonder-blocks-core/util/get-clickable-behavior.test.js
somewhatabstract: Fair. I can make that change if you think it's important to do now. Otherwise, I'll take it and ensure that I do that in future work.

packages/wonder-blocks-core/util/get-clickable-behavior.test.js
somewhatabstract: Yeah, made things a lot easier!

packages/wonder-blocks-core/util/get-clickable-behavior.test.js
kevinbarabash: @somewhatabstract I didn't think that through carefully.  I think we still want to provide a way to create links that aren't client-side nav links when the router is present.  Although we can do that by explicitly setting `clientNav={false}` it's confusing b/c often with booleans not setting them is equivalent to explicitly setting them to false.  I also wonder if we should make `clientNav` mandatory even when the router is present otherwise people might be confused as to why they're seeing client side navigation when `clientNav` isn't present.  @jeresig what do you think.

packages/wonder-blocks-core/util/get-clickable-behavior.test.js
kevinbarabash: I think it's probably worth making the change here too in case other people are looking at these tests as a model to follow.  I do like the organization of the tests into three separate parts.

PR #133, opened 2018-06-19
[Icon v1] Create and employ Icon component

idreyn:
This PR adds the Icon component [(spec here)](https://docs.google.com/document/d/1PM5tzTN3yOAWuu-Rduuyz5DfZK2Mdg1nn8LqBd5ag2Y/edit#heading=h.yn0t44oo8b6) and replaces the ad-hoc icon-like elements in `ModalCloseButton` and `IconButton` with this new component!

Test plan: check Styleguidist


jeresig:

Built with commit 864fd421807656f55830947f0500ebc8e1bae71c

https://deploy-preview-133--wonder-blocks.netlify.com

packages/wonder-blocks-icon/components/icon.js
```diff
@@ -0,0 +1,130 @@
...
+    style?: any,
+};
+
+const StyledSVG = addStyle("svg");
+
+const viewportPixelsForSize: {[IconSize]: number} = {
+    small: 16,
+    medium: 24,
+    large: 48,
+    xlarge: 96,
```
kevinbarabash: Is design actually planning on providing assets at the `large` and `xlarge` sizes for some of these assets?

packages/wonder-blocks-icon/components/icon.js
```diff
@@ -0,0 +1,130 @@
...
+     */
+    "aria-label"?: string,
+    /**
+     * The color of the icon. Will default to off-black.
+     */
+    color?: string,
+    /**
+     * One of our named icons from icon-paths.js
+     */
+    name: IconName,
```
kevinbarabash: Relying on a `name` means that we'll have to update `wonder-blocks-icon` to add new icons.  Initiatives should be able to use their own custom icons if need be.  Maybe we could have an `icon` prop or `asset` prop that looks like:
```
{
       small?: string,
       medium?: string,
       large?: string,
       xlarge?: string,
}
```

packages/wonder-blocks-icon/util/icon-paths.js
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * The source SVG paths for our icons at various sizes
+ */
+
+type IconPaths = {
+    [string]: {
+        small?: string,
+        medium?: string,
+        large?: string,
+        xlarge?: string,
```
kevinbarabash: I worry about everything being optional.  Maybe we could say the type is:
```
({ small: string } | { medium: string } | { small: string, medium: string })
```
that way we guarantee that there's at least a `small` or `medium` size path.

packages/wonder-blocks-icon/components/icon.js
```diff
@@ -0,0 +1,130 @@
...
+const StyledSVG = addStyle("svg");
+
+const viewportPixelsForSize: {[IconSize]: number} = {
+    small: 16,
+    medium: 24,
+    large: 48,
+    xlarge: 96,
+};
+
+const getPathForIcon = (
```
kevinbarabash: This could be extracted into a utils.js file and then tested.

packages/wonder-blocks-icon/components/icon.js
```diff
@@ -0,0 +1,130 @@
...
+    if (!iconPaths[name]) {
+        throw new Error(`There is no icon named ${name}.`);
+    }
+    const paths = iconPaths[name];
+    if (paths && paths[size]) {
+        // Great, we have the IconSize we actually requested
+        return {assetSize: size, path: paths[size]};
+    } else {
+        // Oh, no, we don't have the right IconSize! Let's find the next best
+        // one...we prefer
```
kevinbarabash: Can you describe the criteria we use to select the next best icon path?

packages/wonder-blocks-icon/components/icon.js
```diff
@@ -0,0 +1,130 @@
...
+            const availablePixelSize = viewportPixelsForSize[availableSize];
+            const sizeDifference = availablePixelSize - desiredPixelSize;
+            return sizeDifference < 0 ? sizeDifference : 1;
+        });
+        const bestAssetSize = assetSizes[0];
+        if (bestAssetSize && paths[bestAssetSize]) {
+            return {assetSize: bestAssetSize, path: paths[bestAssetSize]};
+        } else {
+            throw new Error(
+                `No path available for icon ${name} at size ${size}`,
```
kevinbarabash: We should always return a path if there's at least one path with that name, right?  If that's the case, shouldn't the error be that there's "no icon with that name".

packages/wonder-blocks-icon/components/icon.js
```diff
@@ -0,0 +1,130 @@
...
+            <StyledSVG
+                style={[styles.svg, style]}
+                width={pixelSize}
+                height={pixelSize}
+                // There is a weird thing where Flow will only recognize a
+                // string-quoted prop name if it's in single quotes, but our
+                // tooling normalizes it to double-quotes on commit. So the
+                // aria-label prop isn't included in props validation.
+                // eslint-disable-next-line react/prop-types
+                aria-label={this.props["aria-label"]}
```
kevinbarabash: Can you open an issue with the eslint react plugin and link to the issue here?  It seems like this should be fixable since it's a string literal.

packages/wonder-blocks-icon/docs.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+While we don't currently have assets for sizes larger than `medium`, we can
+still render the icons at `large` and `xlarge` in a pinch.
+```jsx
+
+const Icon = require("@khanacademy/wonder-blocks-icon").default;
+const iconPaths = require("@khanacademy/wonder-blocks-icon/util/icon-paths.js").default;
+
+<Icon size="large" name="dismiss" />
```
kevinbarabash: I would move this an the next example up to the top since these examples are easy to understand as well as copy and paste.

packages/wonder-blocks-icon/docs.md
```diff
@@ -0,0 +1,135 @@
+All of the `small` icons by name:
+
+```jsx
```
kevinbarabash: These first two examples could be combined into a single one that uses a table with the first column being the names of the icons and the second and third being the small and medium icons.

kevinbarabash: I believe there's a way to disable the "editor".  I think we should do that for this one since the code isn't very usefully to look at from a documentation point of view.

packages/wonder-blocks-icon-button/components/icon-button.js
```diff
@@ -6,14 +6,14 @@ import IconButtonCore from "./icon-button-core.js";
 
 export type SharedProps = {
     /**
-     * An icon, supplied as an svg path string.
+     * The name of a Wonder Blocks icon.
```
kevinbarabash: Thanks for updating icon-button.

packages/wonder-blocks-icon-button/components/icon-button.js
```diff
@@ -6,14 +6,14 @@ import IconButtonCore from "./icon-button-core.js";
...
      */
-    icon: string,
+    iconName: string,
 
     /**
      * Text to display as the title of the svg element.
      */
-    alt: string,
+    "aria-label": string,
```
kevinbarabash: Both of these are breaking changes so we should bump the version in its package.json.  Since this hasn't been published yet, bumping the minor version is sufficient.  We'll also need to update `wonder-blocks-modal`'s package.json.  This isn't critical at this point, but I think we should get into the habit of following semver.

packages/wonder-blocks-icon/docs.md
kevinbarabash: The dismiss icon looks like the letter X, maybe a different icon that looks more icon-y.

packages/wonder-blocks-icon-button/components/icon-button.js
jeresig: @kevinbarabash hmm - this feels like a bit of an anti-pattern. We shouldn't be doing version bumps on every commit. Instead, I think it makes more sense to do commits and then do a release that takes into account everything that we're releasing - right? (I guess I'm curious, in your model, how we manage multiple simultaneous small changes to a module. They can't all bump to the same minor version in their commit so we'll end up with merge conflicts.) It seems like it'd be better to land stuff and then do periodic releases of the code that are well publicized and documented.

packages/wonder-blocks-icon/components/icon.js
jeresig: 🛑 Agreed! We should have tests for this.

packages/wonder-blocks-icon/components/icon.js
```diff
@@ -0,0 +1,130 @@
...
+};
+
+/**
+ * An Icon displays a small informational or decorative image as an SVG. Wonder
+ * Blocks has a fixed set of icons and icon sizes.
+ *
+ * ```js
+ * <Icon
+ *     name={correct}
+ *     color={Colors.green}
```
jeresig: 🛑 We should change the color to be something that we want to support (like Color.white).

packages/wonder-blocks-icon-button/components/icon-button.js
kevinbarabash: I believe that as long as the current version of `wonder-blocks-icon` satisfies each component that depends on it, those components will use the local version so we could leave things without bumping the versions.  Multiple packages can all bump their minor version if they want.  They'll all be published from the same commit and that commit will have multiple tags, e.g. `@khanacademy/wonder-blocks-foo@0.0.2`, `@khanacademy/wonder-blocks-bar@0.02` are what the tags would look like.  I don't think we should publish for every version bump, but if we don't bump the version in the commit then we have to remember to bump it later.  I also worry about inconsistencies where something says it uses version `0.1.0` but actually needs `0.2.0`.

packages/wonder-blocks-icon/components/icon.js
kevinbarabash: Okay, but if initiatives request initiative specific icons be added to wonder-blocks I think we should say "no".

packages/wonder-blocks-icon-button/components/icon-button.js
jeresig: @kevinbarabash Interesting, so I think there are two separate issues: Ensuring that our published/released code points to the right versions and ensuring that the code inside Github represents reality.

What if we did something like this:

- Write some code, submit as PR, land code.
- Publish module "foo" 0.0.1 -> 0.0.2, which is depended upon by "bar" so we bump that from 0.0.1 -> 0.0.2, as well.
- Immediately (as part of the publish process) we add a commit that changes the versions of these modules as something like 0.0.3-pre and leave them in that state until the next publish (at which point we change the version from 0.0.3-pre to 0.0.3 or 0.1.0 or whatever).

Alternatively, we could add a pre-commit hook and have it auto-bump the version to "0.0.3-pre" if a package has any code changed in it. That will still have the issue with merge conflicts, but at least it won't be as bad as they'll be attempting to go to the same version, rather than potentially different ones.

This way code inside Github, that other devs see, will always represent a WIP version of the component. If they want to see code at a specific release then they need to view the version tag. I guess the downside with this is that the version of what we publish to Netlify won't, necessarily, match what users can actually use -- but that's also true even if we're doing publishing (e.g. something that we publish to NPM won't be immediately available in webapp). We may want to consider having a branch or tag inside `wonder-blocks` that points to what we're currently using on webapp -- and then only publish that to the main Netlify (we could publish the WIP stuff to something like wonder-blocks-dev.netlify.com instead).

kevinbarabash: We talked about this IRL and decided that devs won't be responsible for version bumping.  The specifics of how version bumping will happen will be documented somewhere in a Google doc.

packages/wonder-blocks-icon/components/icon.js
```diff
@@ -0,0 +1,130 @@
...
+type Props = {
+    /**
+     * Passed transparently to the SVG. If not provided, we assume the SVG is
+     * purely decorative and it is invisible to screenreaders.
+     */
+    "aria-label"?: string,
+    /**
+     * The color of the icon. Will default to off-black.
+     */
+    color?: string,
```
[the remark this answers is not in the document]
kevinbarabash: Allowing `currentColor` is a pretty big escape hatch.  I wonder if we should use `currentColor` all of the time.

packages/wonder-blocks-icon/components/icon.js
idreyn: This is my understanding based off of the design spec and my conversation with Elizabeth.

packages/wonder-blocks-icon/components/icon.js
idreyn: My understanding at the time of submitting this PR is that we didn't actually need to support non-WB icons. That said, I rather like this pattern and I have since refactored the branch to support it. Maybe I should put it up so you can have a look?

packages/wonder-blocks-icon/util/icon-paths.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+};
+
+export default ({
+    "add": {
+        medium: "M11 11V7a1 1 0 0 1 2 0v4h4a1 1 0 0 1 0 2h-4v4a1 1 0 0 1-2 0v-4H7a1 1 0 0 1 0-2h4zm1 13C5.373 24 0 18.627 0 12S5.373 0 12 0s12 5.373 12 12-5.373 12-12 12zm0-2c5.523 0 10-4.477 10-10S17.523 2 12 2 2 6.477 2 12s4.477 10 10 10z",
```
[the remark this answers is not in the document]
idreyn: Shall I add that to `.gitignore`?

packages/wonder-blocks-icon/components/icon.js
kevinbarabash: @idreyn thanks for making this change.  I think the thing we didn't want was having the support the current icon format from webapp, but now we have a new format. 🙌

packages/wonder-blocks-icon/util/icon-paths.js
kevinbarabash: I guess this plan of ensure at least one item is present gets a little more complicated if design could provide us with assets for any one of the four sizes.

packages/wonder-blocks-icon/util/icon-paths.js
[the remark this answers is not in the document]
kevinbarabash: I don't follow.

packages/wonder-blocks-icon-button/components/icon-button.js
```diff
@@ -6,14 +6,14 @@ import IconButtonCore from "./icon-button-core.js";
 
 export type SharedProps = {
     /**
-     * An icon, supplied as an svg path string.
+     * The name of a Wonder Blocks icon.
      */
-    icon: string,
+    iconName: string,
```
kevinbarabash: This should be updated to match the way we specify icons to `Icon`.

packages/wonder-blocks-icon/docs.md
idreyn: 👍

packages/wonder-blocks-icon/components/icon.js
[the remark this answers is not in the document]
idreyn: I thought we discussed allowing any color for the time being. I have no strong opinion one way or the other, but is the other option that we constrain the color to values in `wonder-blocks-color` and additionally add `currentColor`?

packages/wonder-blocks-icon/components/icon.js
idreyn: I can look into it, but I would need to spend some time writing up a minimal repro. I think that's outside the scope of what I need to get done this sprint.

packages/wonder-blocks-icon/util/icon-paths.js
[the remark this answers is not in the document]
idreyn: Ah oops somehow that comment above ^^ was meant in reference to `package-lock.json`.

packages/wonder-blocks-icon/docs.md
idreyn: I consolidated the two examples, but I don't think I'm going to put any more time into figuring out how to disable the code preview, since it seems like a pretty minor nice-to-have.

packages/wonder-blocks-icon/components/icon.js
```diff
@@ -0,0 +1,100 @@
...
+ *     style={{margin: 4}}
+ * />
+ * ```
+ *
+ * Wonder Blocks comes with a fixed set of icons available by importing `icons`,
+ * but you can provide your own `IconAsset` in the following format:
+ *
+ * ```js
+ * {small?: string, medium?: string, large?: string, xlarge?: string}
+ * ```
```
kevinbarabash: Nice docs!

packages/wonder-blocks-icon-button/components/icon-button.js
```diff
@@ -2,18 +2,20 @@
...
 import IconButtonCore from "./icon-button-core.js";
 
 export type SharedProps = {
     /**
-     * An icon, supplied as an svg path string.
+     * A Wonder Blocks icon asset, an object specifing paths for one or more of
+     * the following sizes: small, medium, large, xlarge.
      */
-    icon: string,
+    icon: IconAsset,
```
kevinbarabash: yay

packages/wonder-blocks-icon/components/icon.js
kevinbarabash: This captured in https://khanacademy.atlassian.net/browse/WB-291.

packages/wonder-blocks-icon/generated-snapshot.test.js
```diff
@@ -0,0 +1,143 @@
...
+            tableBorder: {
+                border: "1px #DDD solid",
+            },
+        });
+
+        const example = (
+            <table
+                className={css(styles.table)}
+                style={{borderCollapse: "collapse"}}
+            >
```
kevinbarabash: This table makes it really clear which icons have only one size or the other.  Thanks.

packages/wonder-blocks-icon/util/icon-util.js
```diff
@@ -0,0 +1,58 @@
...
+    assetSize: IconSize,
+    path: string,
+} => {
+    if (icon[size]) {
+        // Great, we have the IconSize we actually requested
+        return {assetSize: size, path: icon[size]};
+    } else {
+        // Oh, no, we don't have the right IconSize! Let's find the next best
+        // one...we prefer to find a smaller icon and blow it up instead of
+        // using a larger icon and shrinking it such that detail may be lost.
```
kevinbarabash: :+1:

packages/wonder-blocks-icon/util/icon-util.js
```diff
@@ -0,0 +1,58 @@
...
+            return tooLargeByPixels > 0
+                ? Number.POSITIVE_INFINITY
+                : Math.abs(tooLargeByPixels);
+        };
+        const assetSizes = availableSizes.sort((a, b) => sortFn(a) - sortFn(b));
+        const bestAssetSize = assetSizes[0];
+        if (bestAssetSize && icon[bestAssetSize]) {
+            return {assetSize: bestAssetSize, path: icon[bestAssetSize]};
+        } else {
+            throw new Error("Icon does not contain any valid asset sizes!");
```
kevinbarabash: :+1:

kevinbarabash:
Hmm... that's curious that the netlify deploy didn't work.

kevinbarabash:
I got netlify to redeploy and this time it worked.  It looks like it was a network error on netlify's end.

kevinbarabash:
The preview site looks good.  I noticed a minor issue with the docs but that can be fixed up in a separate PR.

kevinbarabash:
Hmm... GitHub's UI is a little weird in that it doesn't indicate that a re-review is required.

kevinbarabash:
From the docs:
```
# Order is important; the last matching pattern takes the most
# precedence.
```
@jeresig we probably want to list ourselves on each package as code owners as well so that either us or the person who wrote the package initially can be approved reviewers.  GitHub's UI unfortunately makes it very opaque as to which code owner actually needs to do the review.  I'm going to force merge this as an admin and then fix up the CODEOWNERS file in a separate PR.

PR #134, opened 2018-06-19
Fix roles for aria role types

wishdasher:
There aren't currently being used anywhere. When I was adding aria roles to the new dropdown component, I noticed that flow broke, so I'm fixing the type definition here.

### DIFF
--- a/packages/wonder-blocks-core/util/types.js
+++ b/packages/wonder-blocks-core/util/types.js
@@ -12,80 +12,80 @@ export type StyleType<T: Object> = any; // eslint-disable-line no-unused-vars
 export type TextTag = "span" | "p" | "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
 
 // Source: https://www.w3.org/WAI/PF/aria-1.1/roles#role_definitions
-type roles = [
-    "alert",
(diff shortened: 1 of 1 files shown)


jeresig:

Built with commit 7c78a98f869bd15460fd52aae0ae74243f0ed59b

https://deploy-preview-134--wonder-blocks.netlify.com

packages/wonder-blocks-core/util/types.js
```diff
@@ -12,80 +12,80 @@ export type StyleType<T: Object> = any; // eslint-disable-line no-unused-vars
...
+    | "tabpanel"
+    | "textbox"
+    | "timer"
+    | "toolbar"
+    | "tooltip"
+    | "tree"
+    | "treegrid"
+    | "treeitem"
+    | "widget" // abstract role
+    | "window";
```
wishdasher: The comment `// abstract role`` is moved to line 15 when I commit...

jeresig: OK, that's awkward. What happens if you make it:

```
// abstract role
| "window";
```
?

kevinbarabash:
@wishdasher thanks!

PR #137, opened 2018-06-21
break dependency cycle between -core and -spacing

kevinbarabash:
This breaks the cyclic dep by introducing `wonder-blocks-layout` which contains `Spring` and `Strut` components.


jeresig:

Built with commit a0ca912aa59ab4ef1a5f4de0c3a6061d15e626d1

https://deploy-preview-137--wonder-blocks.netlify.com

kevinbarabash:
I could see it making sense in a number of places which makes me think it should probably stay where it is until there's a clear winner.

kevinbarabash:
> Seems good! Are you imagining that <Layout> should go in here, too? should <MediaLayout>?

That makes sense to me.  It'd be nice to keep `-core` really tight.  I can make that change before doing a publish.

PR #138, opened 2018-06-21
Prevent spacebar from calling props.onClick twice via clickable-behavior

wishdasher:
For components that use `ClickableBehavior` (e.g. `Button`), pressing spacebar was triggering the component's `onClick` prop twice. This is because in `ClickableBehavior`, `handleKeyUp` set the variable `keyboardClick` to `false` and then called `props.onClick`. However, the spacebar also triggers the `ClickableBehavior`'s `onClick` -> `handleClick` function. In `handleClick`, `keyboardClick` is `false`, so `props.onClick` is called again. This change moves `this.keyboardClick = false` to the correct location.

Now, pressing `enter` or `space` on a focused component only triggers its `onClick` callback once. Clicking the component with a mouse works the same as before.


jeresig:

Built with commit cf1a5c9e1952c7d51ca32a7b3bce94dec170beec

https://deploy-preview-138--wonder-blocks.netlify.com

wishdasher:
https://wonder-blocks.netlify.com/#button
Bring focus to a button and press spacebar. "Hello, World!" prints twice.

https://deploy-preview-138--wonder-blocks.netlify.com/#button
Now pressing spacebar only prints the message once. Verify that pressing enter or clicking on the button still works as expected. Verify that the `href`-ed button (one row below) also works as before.

packages/wonder-blocks-core/components/clickable-behavior.test.js
```diff
@@ -106,8 +106,8 @@ describe("ClickableBehavior", () => {
         );
         expect(button.state("focused")).toEqual(false);
         button.simulate("keydown", {keyCode: keyCodes.space});
-        button.simulate("click", {preventDefault: jest.fn()});
         button.simulate("keyup", {keyCode: keyCodes.space});
+        button.simulate("click", {preventDefault: jest.fn()});
```
wishdasher: This is the correct order of events. (Doesn't change the result of the test for neither the existing version nor my change though.)

packages/wonder-blocks-core/components/clickable-behavior.test.js
```diff
@@ -270,4 +270,37 @@ describe("ClickableBehavior", () => {
...
+            </ClickableBehavior>,
+        );
+        expect(onClick).not.toHaveBeenCalled();
+
+        button.simulate("mousedown");
+        button.simulate("mouseup");
+        button.simulate("click", {preventDefault: jest.fn()});
+        expect(onClick).toHaveBeenCalledTimes(1);
+
+        // Order of DOM events is different for space vs. enter
```
kevinbarabash: I didn't know that, cool!

PR #139, opened 2018-06-22
ignore changes to tests and documentation when publishing

kevinbarabash:
Unfortunately, this change doesn't impact any of the packages that lerna wants to bump since it still checks to see if devDependencies have changed when decided whether or not to bump a package's version.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,6 +3,14 @@
   "packages": [
     "packages/*"
   ],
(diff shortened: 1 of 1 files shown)


jeresig:

Built with commit 468a1419e8599741a5503460dda2866625232a16

https://deploy-preview-139--wonder-blocks.netlify.com

PR #140, opened 2018-06-22
Initial tooltip work

somewhatabstract:
This is far from complete, but this seems like a good set of working changes for us to then build upon. This way we can keep the reviews somewhat small.

In this change:
1. `TooltipPortalMounter`: This component is responsible for mounting the tooltip bubble in a portal
1. `TooltipBubble`: This component is responsible for all things bubbly. It handles rendering and positioning of the tooltip (using react-popper).
1. `TooltipAnchor`: This component is responsible for anchoring the tooltip to a given parent element and monitoring it for hover/focus events.
1. `Tooltip`: This component is the tooltip. The API is very sparsely implemented in this diff, but it brings together the other components to provide a very basic tooltip.

These components use a variety of util methods, primarily to determine if the anchor element is visible or not (i.e., whether we should show the tooltip if our anchor is focused, but scrolled of screen, or not).

## Changes to come
1. `TooltipContent` component for providing what the tooltip should show
1. Making the bubble only accept our `TooltipContent` component
1. All the props for initial positioning
1. Hopefully, some better boundary checking so we can flip the tooltip when the anchor starts getting hidden
1. Handle the WCAG 2.1 things (keeping the bubble open if it is hovered over, dismissing it on Escape key, etc.)
1. Only show one tooltip at a time
1. Tooltip delays per spec
1. Fix many of the TODOs (see Jira for all the tasks remaining)
1. More tests!

And probably more.


jeresig:

Built with commit 68a94949e88c3395aba9948c388b854495fb7a2d

https://deploy-preview-140--wonder-blocks.netlify.com

packages/wonder-blocks-button/generated-snapshot.test.js
```diff
@@ -6,6 +6,9 @@
 //   2. Run `yarn run gen-snapshot-tests`.
 import React from "react";
 import renderer from "react-test-renderer";
+
+// Mock react-dom as jest doesn't like findDOMNode.
+jest.mock("react-dom");
```
somewhatabstract: I'm moving this general change into a different PR, but needed it here to get tests to pass.

jeresig:

Built with commit 0961197cef27d8db5c53904f75b292fef7ef735d

https://deploy-preview-140--wonder-blocks.netlify.com

somewhatabstract:
@kevinbarabash I am breaking down this PR, however, I don't think this is a good idea.

> TooltipBubble... it'd be good to have some docs for this in this diff so that we can see what the bubble looks before the following PR

The bubble is a positioning element, not a rendering element. It doesn't govern appearance of things, only where they go. The `TooltipContent` and `TooltipArrow` components will govern appearance.

I can add a test doc for the `TooltipBubble`, but I think it will only be confusing and not necessarily add value (not to mention it requires a test component to wrap it so that we can give it an anchor point).

somewhatabstract:
OK, I have reworked this into PRs #145, #146, #148, #149, #150, and #151. Since they now cover this work, I'm going to close this PR and defer to the new ones for getting this reviewed and merged.

kevinbarabash:
> The bubble is a positioning element, not a rendering element. It doesn't govern appearance of things, only where they go. The TooltipContent and TooltipArrow components will govern appearance.

We should probably call it `TooltipPositioner` to avoid this confusion.  Thanks for splitting up this PR.  I'm looking forward to reviewing it tomorrow.

somewhatabstract:
@kevinbarabash indeed. When the rendering and positioning bits are split off at the styling stage, I'll be making that split

PR #141, opened 2018-06-22
Mock react-dom in snapshot tests

somewhatabstract:
This is needed for when we use things like `ReactDOM.findDOMNode` in our components, as they will cause jest to fail the test.

While no components do this yet, there are some coming (like the `Tooltip`: #140 ) that will require this change.

### DIFF
--- a/utils/gen-snapshot-tests.js
+++ b/utils/gen-snapshot-tests.js
@@ -44,6 +44,9 @@ function generateTestFile(root, examples, componentFileMap) {
         "//   2. Run `yarn run gen-snapshot-tests`.",
         `import React from "react";`,
         `import renderer from "react-test-renderer";`,
+        ``,
+        `// Mock react-dom as jest doesn't like findDOMNode.`,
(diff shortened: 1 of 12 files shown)


utils/gen-snapshot-tests.js
```diff
@@ -44,6 +44,9 @@ function generateTestFile(root, examples, componentFileMap) {
         "//   2. Run `yarn run gen-snapshot-tests`.",
         `import React from "react";`,
         `import renderer from "react-test-renderer";`,
+        ``,
+        `// Mock react-dom as jest doesn't like findDOMNode.`,
+        `jest.mock("react-dom");`,
```
somewhatabstract: This is the actual change. The rest of the changes are the auto-generated output.

jeresig:

Built with commit 46b8561334aed8c4e0bc3a646954872644ce5ac0

https://deploy-preview-141--wonder-blocks.netlify.com

PR #142, opened 2018-06-23
Make jest work idiomatically and add extensions.json

somewhatabstract:
This is a tooling-only PR.

In this PR there are two main changes:
1. `jest` now uses `globals` in the jest configuration to control aphrodite style flattening, rather than an environment variable. This allows us to use idiomatic jest tooling like the VS Code jest extension.
1. Adds an `extensions.json` file to recommend a few extensions to use when developing for Wonder Blocks. This adds a "star" swatch to the extensions in the extensions explorer to indicate they are recommend; they can also be viewed by searching extensions for `@recommended` and expanding the "Workspace Recommendations" section.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,19 +1,19 @@
 {
-	// See http://go.microsoft.com/fwlink/?LinkId=827846
-	// for the documentation about the extensions.json format
-	"recommendations": [
-		// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+    // See http://go.microsoft.com/fwlink/?LinkId=827846
+    // for the documentation about the extensions.json format
+    "recommendations": [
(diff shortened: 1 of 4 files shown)


jeresig:

Built with commit 6cead377832f54b64f4a0d68a323ead71b5758cc

https://deploy-preview-142--wonder-blocks.netlify.com

somewhatabstract:
See also #143

somewhatabstract:
@kevinbarabash Yeah, that is strange. I'll see if I can fix that to use spaces.

And yes, they are still recommendations. It won't force them on anyone.

somewhatabstract:
Hmm, seemed to be auto-detected and since VS Code added the initial file, it was just continuing to use the tabs that had already provided. I replaced them with spaces and now it is always using spaces. :)

PR #145, opened 2018-06-26
Add util methods that will be used by tooltip visibility checks

somewhatabstract:
This PR is a partial element of #140 so that we can break it down into more manageable chunks.

This adds two utilities to the `wonder-blocks-core` package.

1. `enumerateScrollAncestors`
    - This takes an element and provides an iterator that returns all the ancestor elements that can scroll the element. This is used indirectly by the tooltip to determine if any parent element has hidden a tooltip anchor.

2. `getElementIntersection`
    - This takes an element and an optional bounds element. If the bounds element isn't present, it finds the intersection of the given element against each scroll parent in turn (using `enumerateScrollAncestors` until one of them is found to be hiding the given element, or that none of them are hiding it.
    - In this iteration, this simply checks if the element is before, after, or within the bounds of the given `boundsElement` or its parents. In a future change, we will likely extend this to provide some indication of partial visibility to support things like tooltip flipping.


jeresig:

Built with commit a25edc4606e644977bb6d044e0351fa36111fb18

https://deploy-preview-145--wonder-blocks.netlify.com

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.js
```diff
@@ -0,0 +1,92 @@
+// @flow
+// Utility for finding the scroll ancestors of a child
+// Inspired by https://github.com/olahol/scrollparent.js
+// with modifications to make a generator that returns a sequence of all scroll
```
kevinbarabash: `generator => iterator`

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.js
```diff
@@ -0,0 +1,92 @@
...
+// ancestors.
+// Also modified for our standards (and commented a bit).
+const style = function(node: Element, prop) {
+    return getComputedStyle(node).getPropertyValue(prop);
```
kevinbarabash: Maybe `getStyleProp`, `style` as a verb indicates that's styling something which isn't the case.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.js
```diff
@@ -0,0 +1,92 @@
...
+};
+
+const overflow = function(node: Element) {
```
kevinbarabash: Maybe `getOverflowStyle` or something like that.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.js
```diff
@@ -0,0 +1,92 @@
...
+    return (
+        style(node, "overflow") +
+        style(node, "overflow-y") +
+        style(node, "overflow-x")
+    );
+};
+
+const scroll = function(node: Element) {
```
kevinbarabash: `isScrollable` or `canScroll`

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.js
```diff
@@ -0,0 +1,92 @@
...
+            this.parentElement = element.parentElement;
+        }
+    }
+
+    //$FlowFixMe: Computed property keys are not allowed by flow at the moment
+    [Symbol.iterator]() {
+        return this;
+    }
+
+    next(): IteratorResult<Element, void> {
```
kevinbarabash: TIL.  I didn't realize that an iterator could "return" something else in addition to what was yielded for each call to `.next()`.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.js
```diff
@@ -0,0 +1,92 @@
...
+
+        if (!scrollCandidate) {
+            // If we don't have a scroll candidate, we're done iterating.
+            // So let's remember that.
+            this.done = true;
+
+            // If we don't have a documentElement, we are done on this
+            // iteration.
+            if (!document.documentElement) {
+                return {done: true};
```
kevinbarabash: Is this to support tests?  If so, please include that in the comment.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.js
```diff
@@ -0,0 +1,92 @@
...
+        };
+    }
+}
+
+export default function enumerateScrollAncestors(
+    element: Element,
+): Iterable<Element> {
+    return Object.defineProperty({}, Symbol.iterator, {
+        value: () => new ScrollAncestorsIterator(element),
+        writable: false,
```
kevinbarabash: Why worry about people overwriting the value of `retVal[Symbol.iterator]`?

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.test.js
```diff
@@ -0,0 +1,99 @@
...
+        // Arrange
+        const enumerator = enumerateScrollAncestors((null: any));
+
+        // Act
+        // Flow doesn't want us to index  $FlowFixMe
+        const result = enumerator[Symbol.iterator]();
+
+        // Assert
+        expect(result).not.toBeUndefined();
+        expect(result).toBe(result[Symbol.iterator]());
```
kevinbarabash: This doesn't really test that it's an iterator only that it returns the same value.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.test.js
```diff
@@ -0,0 +1,99 @@
...
+                    <div ref={actAndAssert}>Test</div>
+                </div>
+            );
+
+            mount(nodes);
+        };
+
+        const actAndAssert = (ref) => {
+            if (ref) {
+                const domNode = ((ReactDOM.findDOMNode(ref): any): Element);
```
kevinbarabash: nit: the `ref` could be typed as a `Node`.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.test.js
```diff
@@ -0,0 +1,99 @@
...
+
+                // Act
+                const result = Array.from(enumerateScrollAncestors(domNode));
+
+                // Assert
+                expect(result.length).toBe(1);
+                expect(result[0]).toBe(document.documentElement);
```
kevinbarabash: Cool!

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.test.js
```diff
@@ -0,0 +1,99 @@
...
+
+                // Act
+                const result = Array.from(enumerateScrollAncestors(domNode));
+
+                // Assert
+                expect(result).toHaveLength(4);
+                expect(result[0].getAttribute("id")).toBe("parent");
+                expect(result[1].getAttribute("id")).toBe("parent2");
+                expect(result[2].getAttribute("id")).toBe("parent3");
+                expect(result[3]).toBe(document.documentElement);
```
kevinbarabash: Awesome!

packages/wonder-blocks-core/util/get-element-intersection.js
```diff
@@ -0,0 +1,130 @@
...
+import enumerateScrollAncestors from "./enumerate-scroll-ancestors.js";
+
+/**
+ * Indicates the intersection state of an item on a single axis.
+ */
+export type AxisIntersection =
+    /**
+     * The item is partly or fully within the bounds.
+     */
+    | "within"
```
kevinbarabash: nit: these could be line comments.

kevinbarabash: Also, you may want to include that `"within"` implies that the intersect rect is strictly within the bounds rect.

packages/wonder-blocks-core/util/get-element-intersection.js
```diff
@@ -0,0 +1,130 @@
...
+        return "before";
+    } else if (start(intersectingRect) >= end(boundsRect)) {
+        return "after";
+    } else {
+        return "within";
+    }
+}
+
+/**
+ * Determines whether of an element against a single other element.
```
kevinbarabash: I think there's a word missing here.

packages/wonder-blocks-core/util/get-element-intersection.js
```diff
@@ -0,0 +1,130 @@
...
+        const intersection = getElementIntersectionAgainstParent(
+            intersectingRect,
+            scrollParent,
+        );
+        if (
+            intersection.vertical !== "within" ||
+            intersection.horizontal !== "within"
+        ) {
+            // Stop looking, we've found something that is hiding the element.
+            return intersection;
```
kevinbarabash: Why not get rid fo the `if` statement and save some lines of code?

packages/wonder-blocks-core/util/get-element-intersection.js
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * The intersection on the vertical axis.
+     */
+    vertical: ?AxisIntersection,
+|};
+
+const UndeterminedIntersection: Intersection = {
+    horizontal: null,
+    vertical: null,
+};
```
kevinbarabash: I wonder if it's worth `freeze`-ing or just returning a new value each time.  Both have their own runtime costs, but returning a shared object feels dangerous.

packages/wonder-blocks-core/util/get-element-intersection.test.js
```diff
@@ -0,0 +1,400 @@
...
+                const element = (ReactDOM.findDOMNode(node): any);
+
+                // Act
+                const result = getElementIntersection(element);
+
+                // Assert
+                expect(result).toMatchObject({
+                    horizontal: "within",
+                    vertical: "within",
+                });
```
kevinbarabash: Nice!

packages/wonder-blocks-core/util/get-element-intersection.test.js
```diff
@@ -0,0 +1,400 @@
...
+                getElementIntersection(element, fakeParentElement);
+
+                // Assert
+                expect(enumerateScrollAncestors).not.toHaveBeenCalledWith(
+                    element,
+                );
+                done();
+            };
+
+            arrange(actAndAssert);
```
kevinbarabash: You have a number of similar tests, it might be worth making a helper, e.g.
```
testHelper(arrayOfScrollAncestors, underTestBounds, expectedResult);
```

somewhatabstract: I don't like linking tests that way if I can help it. It makes test maintenance much harder. Generally, keeping test implementations isolated from one another keeps things clear and unambiguous.

packages/wonder-blocks-core/util/get-element-intersection.js
somewhatabstract: Yeah, I wondered about that myself. Freezing sounds like a great idea, but we could also change this to be a function that returns the result, so we still encapsulate in one place.

packages/wonder-blocks-core/util/get-element-intersection.js
somewhatabstract: Yup, I'll fix that. Good find.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.js
somewhatabstract: Thanks. Leftover from my initial work.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.js
somewhatabstract: Fair.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.js
somewhatabstract: Yes, it can be mildly confusing at first.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.js
somewhatabstract: No, this is how iterators signal they have nothing left to give. They have a `done` flag in the result.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.js
somewhatabstract: I was copying an example someone else had given somewhere. I seem to recall not doing this and some other error cropped up. Seemed harmless to leave this. I will try changing it back to the writable version and see if it works; I could be misremembering some earlier attempts.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.test.js
somewhatabstract: Yes, but since other tests are testing the iterator part, and then this tests that it matches, then we are in turn, already testing that it returns an iterator.

packages/wonder-blocks-core/util/get-element-intersection.js
somewhatabstract: Not sure I understand. We are looping through each parent and stopping on the first one that is hiding the element. Without the `if` we would just return on the first parent result, even if it was not hiding the element.

Perhaps I am misunderstanding your suggestion. Can you elaborate?

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.test.js
somewhatabstract: I don't think that's right. If it is specified as a `Node`, `findDOMNode` gets quite upset if I do that. Since `flow` isn't complaining without the type, I am inclined to leave it untyped.

packages/wonder-blocks-core/util/get-element-intersection.js
somewhatabstract: Not sure I understand your second comment there, @kevinbarabash. Can you elaborate? The comment that I currently have on the `within` value states that the item can be partly or fully within the bounds, which is currently correct as to the function implemented.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.js
kevinbarabash: The question was about the `!document.documentElement` check, but I understand that's what terminates the iteration.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.js
kevinbarabash: No need to change this, just curious more than anything.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.test.js
kevinbarabash: interesting... I was sure that would work, I guess I have to update my understanding of how ref callbacks work.

packages/wonder-blocks-core/util/get-element-intersection.js
kevinbarabash: d'oh... my misunderstanding

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.js
somewhatabstract: Ah, yes, well, by the standards and my reading on `documentElement`, it can be `null` in some circumstances, including testing, but not limited to that, so I included the check for completeness.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.test.js
somewhatabstract: I believe the type for `findDOMNode` has to be `?(Element | React.Component<*>)`.

packages/wonder-blocks-core/util/get-element-intersection.js
kevinbarabash: I missed that this was inside a `for` loop and that it was causing the loop to terminate early.

somewhatabstract: Yeah, I noticed it was a little...hard to read. Hopefully the updated commentary helps make it clearer?

packages/wonder-blocks-core/util/get-element-intersection.js
kevinbarabash: Yeah, let's go with that.  We should do the same for `FullIntersection`.

PR #146, opened 2018-06-26
Created wonder-blocks-tooltip package

somewhatabstract:
This PR builds on #145 and is part of breaking #140 down into more manageable reviews.

Here we add the new `wonder-blocks-tooltip` package.

There's no functionality.


jeresig:

Built with commit eff285266ec7f8d3ff2e94a45ac3b64ab4ec8ee5

https://deploy-preview-146--wonder-blocks.netlify.com

packages/wonder-blocks-tooltip/.npmignore
```diff
@@ -0,0 +1,2 @@
+index_stories.js
+index_test.js
```
kevinbarabash: I wonder if we can pass the top-level .npmignore to `lerna` as part of the publish process.  No action required.

packages/wonder-blocks-tooltip/docs.md
```diff
@@ -0,0 +1,9 @@
+Tooltips should be used to help describe an on screen object.
+
+Tooltips:
+
+- contain text
+- may contain small graphic elements to complement the text
+- do not include actions
```
kevinbarabash: "are dismissed when focus/hover moves away" or something like that.

packages/wonder-blocks-tooltip/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "MIT",
+  "dependencies": {
+    "@khanacademy/wonder-blocks-color": "^1.0.0",
+    "@khanacademy/wonder-blocks-core": "^1.0.0",
+    "@khanacademy/wonder-blocks-modal": "^0.0.3",
+    "@khanacademy/wonder-blocks-typography": "^1.0.0"
```
kevinbarabash: These should match the current version, e.g. `"@khanacademy/wonder-blocks-color": "^1.0.2"` (or whatever it is).  If it isn't the same, then `lerna` won't bump the version automatically.  Probably not a huge deal, but some of the other packages are doing this so we should be consistent.

packages/wonder-blocks-tooltip/.npmignore
somewhatabstract: 🤔 That would be cool

packages/wonder-blocks-tooltip/package.json
somewhatabstract: Yeah, I think the versions changed since I made the PR. I'll make sure to refresh them. Thanks for bringing this up. 

I wonder, going forward, how we can ensure this is the case when we merge code. Perhaps there's a script we can add to our tests that validates the versions in new PRs?

packages/wonder-blocks-tooltip/docs.md
somewhatabstract: I took this text directly from the Zeplin board. Definitely needs fleshing out a bit more.

kevinbarabash:
@somewhatabstract I'll let you manage the merging of this and your other PRs since they're stacked and I don't want to mess things up.

PR #147, opened 2018-06-26
Add select box implementation to dropdown package

wishdasher:


wishdasher:
Will change the base branch to `master` once [PR 144](https://github.com/Khan/wonder-blocks/pull/144) is merged.

jeresig:

Built with commit 5e840dd3cbea8241c732bb46cbd70ce0711d3494

https://deploy-preview-147--wonder-blocks.netlify.com

kevinbarabash:
It's unfortunate that Github doesn't have a setting to make all branches protected branches.

kevinbarabash:
screen shot 2018-06-27 at 11 35 49 am

There are a number of differences from the design spec that need to be addressed:
- the background color should be white, not grey
- in the disabled state, the caret should be the same color as the text
- in the active state, the text should *not* change color to white

packages/wonder-blocks-dropdown/components/select-box.js
```diff
@@ -0,0 +1,231 @@
...
+                borderStyle: "solid",
+                borderWidth: 1,
+            },
+            focus: {
+                borderColor: blue,
+                borderStyle: "solid",
+                borderWidth: 2,
+                // These values are default padding (16 and 12) minus 1, because
+                // changing the borderWidth to 2 messes up the button width
+                // and causes it to move a couple pixels. This fixes that.
```
kevinbarabash: Thanks for adding a comment.  I'm confused as to why we minus 1 instead of `2` which is the actual border width.  I see that `Button` does the same thing.

packages/wonder-blocks-dropdown/components/select-box.md
```diff
@@ -0,0 +1,80 @@
...
+const styles = StyleSheet.create({
+    row: {
+        flexDirection: "row",
+    },
+});
+<View style={[styles.row]}>
+    <SelectBox disabled={true} onClick={() => console.log("error error!!")}>
+        Disabled
+    </SelectBox>
+</View>;
```
kevinbarabash: Thanks for putting all of these examples in `View`.

packages/wonder-blocks-dropdown/components/select-box.md
```diff
@@ -0,0 +1,80 @@
...
+            style={{width: 150}}
+        >
+            Light version
+        </SelectBox>
+    </View>
+</View>;
+```
+
+### Example: Selectbox with a placeholder
+Caret color will be fixed with incorporation of Icon element.
```
kevinbarabash: Cool.  Can you make this a TODO?

packages/wonder-blocks-dropdown/components/select-box.md
```diff
@@ -0,0 +1,80 @@
...
+});
+<View style={[styles.row]}>
+    <SelectBox
+        isPlaceholder={false}
+        onClick={() => console.log("regular selectbox selected")}
+        style={{width: 200}}
+    >
+        Regular selectbox
+    </SelectBox>
+    <View style={[styles.strutLike]} />
```
kevinbarabash: Why doesn't `Strut` work here?  We could also set `marginRight` on the first `SelectBox`.

packages/wonder-blocks-dropdown/components/select-box.js
```diff
@@ -0,0 +1,231 @@
...
+    if (light) {
+        newStyles = {
+            focus: {
+                ":before": {
+                    content: "''",
+                    position: "absolute",
+                    top: -3,
+                    left: -3,
+                    right: -3,
+                    bottom: -3,
```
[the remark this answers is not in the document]
wishdasher: The most semanticky way I can think of to calculate these is to set a constant for the focus ring width (which is 2) and then use `calc(-${focusRingWidth}-1)`. Would that be better?

packages/wonder-blocks-dropdown/components/select-box.js
wishdasher: I think it's because the default state has a border of width 1. In the focused state, the border width changes to 2, so the difference is 1, and that's the amount we need to adjust the padding.

wishdasher:
image

I'm confused by the grey backgrounds you're seeing. The text (on active state) also doesn't change to white for me.

packages/wonder-blocks-dropdown/components/select-box.js
kevinbarabash: That makes sense.  I forgot that we make the focus ring thicker in that case.

packages/wonder-blocks-dropdown/components/select-box.js
```diff
@@ -0,0 +1,231 @@
...
+                    borderColor: white,
+                    borderStyle: "solid",
+                    borderWidth: 2,
+                    borderRadius: 7,
```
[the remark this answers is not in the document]
wishdasher: Oops, the default radius IS 4, thought it was 3 for some reason, d'oh.

packages/wonder-blocks-dropdown/components/select-box.md
```diff
@@ -0,0 +1,83 @@
+### Example: Regular and placeholder select box versions
+These select boxes also have specified widths.
+TODO(sophie): Caret color on the placeholder box will be fixed with incorporation of Icon element.
+
+```js
+const {View} = require("@khanacademy/wonder-blocks-core");
+const {Strut} = require("@khanacademy/wonder-blocks-spacing");
```
somewhatabstract: This is the wrong package. It's causing tests to fail.

PR #148, opened 2018-06-26
Add TooltipBubble and supporting methods

somewhatabstract:
Here, we build on #146 and add the positioning component, `TooltipBubble`.

This component uses `react-popper` to provide the positioning support, adding our own visibility modifier built using the things we added in #145.

In a later update, this is likely to be split into a `TooltipBubblePositioner` component for the react-popper wrapping and a `TooltipBubble` component for styling within that positioning component.

This is part of breaking up #140.


jeresig:

Built with commit 3e433a5cd944c440b6782409c28bf91e4e1df179

https://deploy-preview-148--wonder-blocks.netlify.com

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
```diff
@@ -0,0 +1,81 @@
...
+import {css, StyleSheet} from "aphrodite";
+import * as React from "react";
+import {Popper} from "react-popper";
+
+import visibilityModifierDefaultConfig from "../util/visibility-modifier.js";
+
+import type {PopperChildrenProps} from "react-popper";
+import type {Placement} from "../util/types.js";
+
+type Props = {|
```
kevinbarabash: Thanks for taking the plunge and using exact types.  I haven't been pressing people on it since the version of flow we're on has issues with spreading props.  Admittedly we don't do that that often.

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
```diff
@@ -0,0 +1,81 @@
...
+
+export default class TooltipBubble extends React.Component<Props> {
+    _renderContent(childrenProps: PopperChildrenProps) {
+        const {children} = this.props;
+        const {
+            ref,
+            outOfBoundaries,
+            style,
+            placement,
+            //  arrowProps,
```
kevinbarabash: Will `arrowProps` be making a comeback in a subsequent PR?

packages/wonder-blocks-tooltip/__mocks__/popper.js.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+/**
+ *  Mock to allow us to run tests that need react-popper or popper.js
+ * Inspired by:
```
kevinbarabash: super nit: inconsistent spacing

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
```diff
@@ -0,0 +1,81 @@
...
+        } = childrenProps;
+
+        return (
+            <div
+                data-placement={placement}
+                ref={ref}
```
kevinbarabash: Can you add a comment as to why we're not using `View` here?

packages/wonder-blocks-tooltip/components/tooltip-bubble.test.js
```diff
@@ -0,0 +1,74 @@
...
+        const resultRef = (((
+            <div ref={(ref) => this.props.resultRef(ref)}>
+                This is a pretend string with a ref so we can detect it being
+                rendered
+            </div>
+        ): any): string);
+        return (
+            <div>
+                <div ref={(ref) => this.updateRef(ref)}>Anchor</div>
+                <TooltipBubble
```
kevinbarabash: The tests should be written using `View` since people will be using `View` in production.

packages/wonder-blocks-tooltip/package.json
```diff
@@ -15,6 +15,7 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
+    "react-popper": "^1.0.0-beta.6",
```
kevinbarabash: 🎵 living on the edge... 🎵

packages/wonder-blocks-tooltip/util/is-obscured.js
```diff
@@ -0,0 +1,71 @@
...
+    // going to end up hiding, so, you know, probably don't do that.
+    const anchorRect = element.getBoundingClientRect();
+    const style =
+        ((element: any).currentStyle: ?CSSStyleDeclaration) ||
+        window.getComputedStyle(element);
+    const anchorLeft =
+        anchorRect.left +
+        parseFloat(style.marginLeft) +
+        parseFloat(style.paddingLeft) +
+        parseFloat(style.borderLeftWidth);
```
kevinbarabash: Would this be affected by `box-sizing`?

packages/wonder-blocks-tooltip/util/is-obscured.js
```diff
@@ -0,0 +1,71 @@
...
+    // saying that we're visible if we hit a parent because we already
+    // checked them for visibility earlier.
+    const topLeftVisible =
+        topLeftElement &&
+        (element.contains(topLeftElement) || topLeftElement.contains(element));
+    const bottomRightVisible =
+        bottomRightElement &&
+        (element.contains(bottomRightElement) ||
+            bottomRightElement.contains(element));
+    return !topLeftVisible && !bottomRightVisible;
```
kevinbarabash: Make sense.

packages/wonder-blocks-tooltip/util/is-obscured.test.js
```diff
@@ -0,0 +1,112 @@
...
+                </div>
+            );
+            mount(nodes);
+        };
+
+        const actAndAssert = (ref) => {
+            const element = ((ReactDOM.findDOMNode(ref): any): Element);
+            // When not obscurred, elementFromPoint should return the element.
+            // Flow doesn't like us doing this to the document $FlowFixMe
+            document.elementFromPoint = jest.fn().mockReturnValue(element);
```
kevinbarabash: Does this resent `elementFromPoint` after the test?  Not a big deal if it doesn't... I'm curious more than anything.

packages/wonder-blocks-tooltip/util/is-obscured.test.js
```diff
@@ -0,0 +1,112 @@
...
+    test("element is obscured, returns true", (done) => {
+        // Arrange
+        let otherElement = null;
+        const arrange = (actAndAssert) => {
+            const nodes = (
+                <div>
+                    <div ref={(ref) => (otherElement = ref)}>
+                        Pretend this covers everything
+                    </div>
+                    <div ref={actAndAssert}>Unobscured</div>
```
kevinbarabash: nit: `Unobscured => Obscured`

packages/wonder-blocks-tooltip/util/is-obscured.test.js
```diff
@@ -0,0 +1,112 @@
...
+                </div>
+            );
+            mount(nodes);
+        };
+
+        const actAndAssert = (ref) => {
+            const element = ((ReactDOM.findDOMNode(ref): any): Element);
+            // When not obscurred, elementFromPoint should return the element.
+            // Flow doesn't like us doing this to the document $FlowFixMe
+            document.elementFromPoint = jest.fn().mockReturnValue(otherElement);
```
kevinbarabash: We should have some tests where `elementFromPoint` returns `otherElement` and the unobscured element is a child of `otherElement` or vice versa.

packages/wonder-blocks-tooltip/util/visibility-modifier.js
```diff
@@ -0,0 +1,70 @@
...
+
+    // If we're hidden, we mimic what the built-in hide method does,
+    // and set the hide flag and the OOB attribute with appropriate
+    // short-circuiting.
+    if (hide) {
+        if (data.hide) {
+            return data;
+        }
+        data.hide = true;
+        data.attributes["x-out-of-boundaries"] = "";
```
kevinbarabash: Is `"x-out-of-boundaries"` only used for test purposes?

packages/wonder-blocks-tooltip/util/visibility-modifier.js
```diff
@@ -0,0 +1,70 @@
...
+
+    // First, we see how the element intersects with its scroll parents.
+    // If it doesn't, then we should hide it.
+    // Otherwise, we check to see if anything else obscures the component (like
+    // a fixed or absolute positioned element).
+    const {horizontal, vertical} = getElementIntersection(anchorElement);
+    const hide =
+        horizontal !== "within" ||
+        vertical !== "within" ||
+        isObscured(anchorElement);
```
kevinbarabash: Nice!

packages/wonder-blocks-tooltip/util/visibility-modifier.js
```diff
@@ -0,0 +1,70 @@
...
+        if (!data.hide) {
+            return data;
+        }
+        data.hide = false;
+        data.attributes["x-out-of-boundaries"] = false;
+    }
+
+    // Always have to return the data object to ensure the modifier chain
+    // in popper.js is unbroken.
+    return data;
```
kevinbarabash: Maybe link to the docs for popper.js modifiers in the docstring at the beginning fo this file.

packages/wonder-blocks-tooltip/util/visibility-modifier.test.js
```diff
@@ -0,0 +1,185 @@
+// @flow
+import Popper from "popper.js";
+import * as Core from "@khanacademy/wonder-blocks-core";
+//import isObscured from "./is-obscured.js";
```
kevinbarabash: remove

packages/wonder-blocks-tooltip/util/visibility-modifier.test.js
```diff
@@ -0,0 +1,185 @@
...
+        // that the results return what we want.
+        describe("Obscured by one or more scroll parents", () => {
+            test("Obscured horizontally only, is not visible", () => {
+                // Arrange
+                const psuedoAnchorElement = {
+                    getBoundingClientRect: () => ({
+                        top: 0,
+                        left: 0,
+                        right: 100,
+                        bottom: 100,
```
kevinbarabash: The value of this is unimportant since we're mocking `Core.getElementIntersection` anyways, right?

packages/wonder-blocks-tooltip/util/is-obscured.js
```diff
@@ -0,0 +1,71 @@
...
+    // Before we assume we're visible let's check to see if something else
+    // is obscuring us.  Here we make the assumption that if our topleft and
+    // bottomright corners are covered by something, we're not visible.
+    // There are ways that this can still not work, such as different
+    // elements only covering those points and the remainder being visible,
+    // or if some covering element has none for pointer-events style, but
+    // those edge cases shouldn't bother the main usages for this method.
+
+    // NOTE: If the anchor element has `pointer-events: none`, we're always
+    // going to end up hiding, so, you know, probably don't do that.
```
kevinbarabash: I don't see any code that checks for `pointer-events: none`.  Could you explain more how this works?

packages/wonder-blocks-tooltip/util/visibility-modifier.test.js
```diff
@@ -0,0 +1,185 @@
...
+                // Act
+                const result = visibilityModifierDefaultConfig.fn(data);
+
+                // Assert
+                expect(result.hide).toBeTruthy();
+                expect(result.attributes["x-out-of-boundaries"]).toBe("");
+            });
+        });
+
+        test("Obscured by fixed or absolute components, is not visible", () => {
```
kevinbarabash: I see... we're making the assumption that by having `getElementIntersection` return `CompletelyWithin` and `isObscured` return `true` that we're simulating a similar situation to what would happen if the anchor was obscured by a fixed or absolute components.  Can you add a comment as it's not immediately obvious how the test title follows from the test?

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
somewhatabstract: Yes, in fact, it's in PR #155

packages/wonder-blocks-tooltip/util/is-obscured.js
somewhatabstract: `elementFromPoint` doesn't return things that have `pointer-events: none`. We could add a runtime check for that on our anchors but that seems a little heavy-handed for what amounts to a cornercase.

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
somewhatabstract: It's habit now and that's worked out great. The cases where we don't want exact make themselves apparent really quickly.

packages/wonder-blocks-tooltip/components/tooltip-bubble.test.js
somewhatabstract: Sadly, that creates a lot of heavy lifting with the refs since the popper.js stuff wants `HTMLElement` types.

In reality, this component will be internal to our other tooltip components, so I don't know that we gain anything by putting in the effort to use View here. Certainly, we should use public components for our public components (which for Tooltip, will be Tooltip and TooltipContent).

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
somewhatabstract: Sadly, that creates a lot of heavy lifting with the refs since the popper.js stuff wants `HTMLElement` types. I've added a comment in the code.

packages/wonder-blocks-tooltip/package.json
somewhatabstract: 😁

packages/wonder-blocks-tooltip/util/is-obscured.js
somewhatabstract: I am not certain at this point. We should definitely test more edge cases (no pun intended).

packages/wonder-blocks-tooltip/util/is-obscured.test.js
somewhatabstract: Hmm, I think so, but I should check!

somewhatabstract: I've added some code to reset it, just in case.

packages/wonder-blocks-tooltip/util/visibility-modifier.js
somewhatabstract: I've added a link in the code. Reproduced here:

https://github.com/FezVrasta/popper.js/blob/08c5d6010346bf9df06e9f81a54fa6c2c51e3639/packages/popper/src/modifiers/hide.js#L29-L42

packages/wonder-blocks-tooltip/util/visibility-modifier.js
somewhatabstract: Good idea.

somewhatabstract: https://popper.js.org/popper-documentation.html#modifiers

packages/wonder-blocks-tooltip/util/visibility-modifier.test.js
somewhatabstract: Yeah, I can add a comment to make that clear.

somewhatabstract: In fact, they're a hangover from before I split `isObscured` into it's own method. Now that we're mocking that method out, we don't need this fake bounds rect call. :grin:

packages/wonder-blocks-tooltip/util/is-obscured.js
somewhatabstract: I have updated the commentary for this.

packages/wonder-blocks-tooltip/util/is-obscured.test.js
```diff
@@ -0,0 +1,134 @@
...
+            document.elementFromPoint = jest.fn().mockReturnValue(element);
+
+            // Act
+            const result = isObscured(element);
+
+            // Assert
+            expect(result).toBeFalsy();
+            // Reset the document method to avoid side-effects.
+            // Flow doesn't like us doing this to the document $FlowFixMe
+            document.elementFromPoint = ogElementFromPoint;
```
kevinbarabash: I'm a little shaky with how `beforeEach` and `afterEach` interact with async tests so I'm cool with handling the reset inside each `test`.

packages/wonder-blocks-tooltip/util/is-obscured.test.js
kevinbarabash: @somewhatabstract friendly ping.

kevinbarabash:
@somewhatabstract friendly ping regarding the isObscured tests.  I had suggested in one of my comments of adding tests to check this situations where the unobscured element was a child of the other element or vice versa.  Thoughts?

packages/wonder-blocks-tooltip/util/is-obscured.test.js
somewhatabstract: Yes, I considered the before each and after each. I think they work fine for async (I think the after each only runs once done() or timeout occurs), but I wasn't sure a single line in each warranted adding them. I could go either way.

somewhatabstract: I've added them. I think it tidies things up considerably and appears to work.

somewhatabstract:
@kevinbarabash I've added an extra test.

packages/wonder-blocks-tooltip/util/is-obscured.test.js
```diff
@@ -123,9 +132,80 @@ describe("isObscured", () => {
...
+            );
+            mount(nodes);
+        };
+
+        const actAndAssert = () => {
+            if (!elementRef || !parentRef || !childRef) {
+                // We're not ready to act or assert yet.
+                // We need all three refs to be in.
+                return;
+            }
```
kevinbarabash: Wow.

packages/wonder-blocks-tooltip/util/is-obscured.test.js
```diff
@@ -123,9 +132,80 @@ describe("isObscured", () => {
...
+            expect(childElement).toBeTruthy();
+
+            // When not obscurred, elementFromPoint should return the element.
+            // So let's return the element for one corner but not the other.
             // Flow doesn't like us doing this to the document $FlowFixMe
-            document.elementFromPoint = ogElementFromPoint;
+            document.elementFromPoint = jest
+                .fn()
+                .mockImplementationOnce(() => parentElement)
+                .mockImplementationOnce(() => childElement);
```
kevinbarabash: Cool!

packages/wonder-blocks-tooltip/util/is-obscured.test.js
kevinbarabash: I wish Github would show more of the code.

somewhatabstract: I think if you drag select a group of lines, you can comment on the group and it should show them all.

somewhatabstract: Hmm, I am mistaken. :(

PR #149, opened 2018-06-26
Add TooltipPortalMounter component

somewhatabstract:
This builds on #148, adding our `TooltipPortalMounter` which is responsible for creating a portal into which we render the `TooltipBubble`. It is another part of #140.

The `TooltipPortalMounter` sits in the component hierarchy, rendering the tooltip anchor content as its children while mounting the bubble in the portal. This ensures that all the React events properly wire up.

PR #150, opened 2018-06-26
Add TooltipAnchor component

somewhatabstract:
This extends #149 to add the `TooltipAnchor` component. This component is responsible for controlling the active state of the tooltip so that we know when to show or hide the tooltip bubble.

This is another piece of #140.

### DIFF
--- a/packages/wonder-blocks-tooltip/components/tooltip-anchor.js
+++ b/packages/wonder-blocks-tooltip/components/tooltip-anchor.js
@@ -1,0 +1,169 @@
+// @flow
+/**
+ * This component turns the given content into an accessible anchor for
+ * positioning and displaying tooltips.
+ */
+import * as React from "react";
+import * as ReactDOM from "react-dom";
+
+import TooltipPortalMounter from "./tooltip-portal-mounter.js";
+
+type Props = {|
+    // A method that renders the content for anchoring the tooltip.
+    // This must return a TooltipPortalMounter component.
+    children: (active: boolean) => React.Element<typeof TooltipPortalMounter>,
+
+    // Callback to be invoked when the anchored content is mounted.
+    // This provides a reference to the anchored content, which can then be
+    // used for calculating tooltip bubble positioning.
+    anchorRef: (?Element) => mixed,
+
+    // When true, if a tabindex attribute is not already present on the element
+    // wrapped by the anchor, the element will be given tabindex=0 to make it
+    // keyboard focusable; otherwise, does not attempt to change the ability to
+    // focus the anchor element.
+    //
+    // Defaults to true.
+    //
+    // One might set this to false in circumstances where the wrapped component
+    // already can receive focus or contains an element that can.
+    // Use good judgement when overriding this value, the tooltip content should
+    // be accessible via keyboard in all circumstances where the tooltip would
+    // appear using the mouse, so verify those use-cases.
+    forceAnchorFocusivity?: boolean,
+|};
+
+type State = {|
+    active: boolean,
+|};
+
+export default class TooltipAnchor extends React.Component<Props, State> {
+    _weSetFocusivity: ?boolean;
+    _anchorNode: ?Element;
+    _focused: boolean;
+    _hovered: boolean;
+
+    static defaultProps = {
+        forceAnchorFocusivity: true,
+    };
+
+    state = {
+        active: false,
+    };
+
+    componentDidMount() {
+        const anchorNode = ReactDOM.findDOMNode(this);
+
+        // This should never happen, but we have this check here to make flow
+        // happy and ensure that if this does happen, we'll know about it.
+        if (anchorNode instanceof Text) {
+            throw new Error(
+                "TooltipAnchor must be applied to an Element. Text content is not supported.",
+            );
+        }
+
+        this._anchorNode = anchorNode;
+        this._updateFocusivity();
+        if (anchorNode) {
+            // TODO(somewhatabstract): Detect ESC key to deactivate
+            /**
+             * TODO(somewhatabstract): Work out how to allow pointer to go over
+             * the tooltip content to keep it active. This likely requires
+             * pointer events but that would break the obscurement checks we do.
+             * So, careful consideration required.
+             */
+
+            anchorNode.addEventListener("focusin", this._handleFocusIn);
+            anchorNode.addEventListener("focusout", this._handleFocusOut);
+            anchorNode.addEventListener("mouseenter", this._handleMouseEnter);
+            anchorNode.addEventListener("mouseleave", this._handleMouseLeave);
+
+            this.props.anchorRef(this._anchorNode);
+        }
+    }
+
+    componentDidUpdate(prevProps: Props) {
+        if (
+            prevProps.forceAnchorFocusivity !==
+                this.props.forceAnchorFocusivity ||
+            prevProps.children !== this.props.children
+        ) {
+            this._updateFocusivity();
+        }
+    }
+
+    componentWillUnmount() {
+        const anchorNode = this._anchorNode;
+        if (anchorNode) {
+            anchorNode.removeEventListener("focusin", this._handleFocusIn);
+            anchorNode.removeEventListener("focusout", this._handleFocusOut);
+            anchorNode.removeEventListener(
+                "mouseenter",
+                this._handleMouseEnter,
+            );
+            anchorNode.removeEventListener(
+                "mouseleave",
+                this._handleMouseLeave,
+            );
+        }
+    }
+
+    _updateFocusivity() {
+        const anchorNode = this._anchorNode;
+        if (!anchorNode) {
+            return;
+        }
+        const {forceAnchorFocusivity} = this.props;
+        const currentTabIndex = anchorNode.getAttribute("tabindex");
+
+        if (forceAnchorFocusivity && !currentTabIndex) {
+            // Ensure that the anchor point is keyboard focusable so that
+            // we can show the tooltip for visually impaired users that don't
+            // use pointer devices nor assistive technology like screen readers.
+            anchorNode.setAttribute("tabindex", "0");
+            this._weSetFocusivity = true;
+        } else if (!forceAnchorFocusivity && currentTabIndex) {
+            // We may not be forcing it, but we also want to ensure that if we
+            // did before, we remove it.
+            if (this._weSetFocusivity) {
+                anchorNode.removeAttribute("tabindex");
+                this._weSetFocusivity = false;
+            }
+        }
+    }
+
+    _updateActiveState(hovered: boolean, focused: boolean) {
+        // Take a snapshot of the old and new state.
+        const oldState = this.state.active;
+        const newState = hovered || focused;
+
+        // Update our stored values.
+        this._hovered = hovered;
+        this._focused = focused;
+
+        // If we changed state, call our subscriber and let them know.
+        if (oldState !== newState) {
+            this.setState({active: newState});
+        }
+    }
+
+    _handleFocusIn = () => {
+        this._updateActiveState(this._hovered, true);
+    };
+
+    _handleFocusOut = () => {
+        this._updateActiveState(this._hovered, false);
+    };
+
+    _handleMouseEnter = () => {
+        this._updateActiveState(true, this._focused);
+    };
+
+    _handleMouseLeave = () => {
+        this._updateActiveState(false, this._focused);
+    };
+
+    render() {
+        return this.props.children(this.state.active);
+    }
+}
--- a/packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
+++ b/packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
@@ -23,9 +23,9 @@ describe("TooltipAnchor", () => {
     test("on mount, subscribes to focus and hover events", () => {
         // Arrange
         const getFakeTooltipPortalMounter = (active) =>
-            (((
-                <View id="anchor">{active ? "true" : "false"}</View>
-            ): any): React.Element<typeof TooltipPortalMounter>);
+            ((<View>{active ? "true" : "false"}</View>: any): React.Element<
+                typeof TooltipPortalMounter,
+            >);
         const nodes = (
             <View>
                 <TooltipAnchor anchorRef={() => {}}>
@@ -64,9 +64,9 @@ describe("TooltipAnchor", () => {
     test("on unmount, unsubscribes from focus and hover events", () => {
         // Arrange
         const getFakeTooltipPortalMounter = (active) =>
-            (((
-                <View id="anchor">{active ? "true" : "false"}</View>
-            ): any): React.Element<typeof TooltipPortalMounter>);
+            ((<View>{active ? "true" : "false"}</View>: any): React.Element<
+                typeof TooltipPortalMounter,
+            >);
         const nodes = (
             <View>
                 <TooltipAnchor anchorRef={() => {}}>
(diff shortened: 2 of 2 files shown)


jeresig:

Built with commit b3b853dff4f0af9ee97a8cb2dea3213db95a87d4

https://deploy-preview-150--wonder-blocks.netlify.com

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -0,0 +1,168 @@
...
+
+    // Callback to be invoked when the anchored content is mounted.
+    // This provides a reference to the anchored content, which can then be
+    // used for calculating tooltip bubble positioning.
+    anchorRef: (?Element) => mixed,
+
+    // When true, the element wrapped by the anchor will be given tabindex=0
+    // to make it keyboard focusable. This value defaults to true. One might set
+    // this to false in circumstances where the wrapped component already can
+    // receive focus or contains an element that can.
```
kevinbarabash: It's too bad there isn't an easy way to check this.

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -0,0 +1,168 @@
...
+                "TooltipAnchor must be applied to an Element. Text content is not supported.",
+            );
+        }
+
+        this._anchorNode = anchorNode;
+        this._updateFocusivity();
+        if (anchorNode) {
+            // TODO(somewhatabstract): Detect ESC key to deactivate
+            // TODO(somewhatabstract): Work out how to allow pointer to go over
+            // the tooltip content to keep it active.
```
kevinbarabash: This seems tricky.  How important is this compared to other a11y considerations?

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -0,0 +1,168 @@
...
+            this._updateFocusivity();
+        }
+    }
+
+    componentWillUnmount() {
+        const anchorNode = this._anchorNode;
+        if (anchorNode) {
+            anchorNode.removeEventListener("focusin", () =>
+                this._handleFocusIn(),
+            );
```
kevinbarabash: Does this actually work?  I thought you had to pass in the exact same function when removing event listeners.

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -0,0 +1,168 @@
...
+                anchorNode.removeAttribute("tabindex");
+                this._weSetFocusivity = false;
+            }
+        }
+    }
+
+    _updateActiveState(hovered: boolean, focused: boolean) {
+        // Take a snapshot of the old and new state.
+        const oldState = this.state.active;
+        const newState = !!(hovered || focused);
```
kevinbarabash: The `!!` isn't necessary since `hovered` and `focused` are both booleans.

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
```diff
@@ -0,0 +1,560 @@
...
+            expect.any(Function),
+        );
+        expect(removeEventListenerSpy).toHaveBeenCalledWith(
+            "mouseenter",
+            expect.any(Function),
+        );
+        expect(removeEventListenerSpy).toHaveBeenCalledWith(
+            "mouseleave",
+            expect.any(Function),
+        );
```
kevinbarabash: I wish there a was a way to check what handlers are currently subscribed to events on an element.

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
```diff
@@ -0,0 +1,560 @@
...
+                    return;
+                }
+
+                // Act
+                const focusin = () =>
+                    doEvent(localRef, new FocusEvent("focusin"), focusout);
+                const focusout = () =>
+                    doEvent(localRef, new FocusEvent("focusout"), assert);
+                // Focus starts the event chain defined above.
+                focusin();
```
kevinbarabash: This is kind of hard to read.  Maybe `doEvent` could be rewritten to return a Promise that resolves when the timeout expires.  Then we could use `async/await` to write this code in linear fashion.

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
```diff
@@ -0,0 +1,560 @@
...
+
+    describe("forceAnchorFocusivity is true", () => {
+        test("if not set, sets tabindex on anchor target", (done) => {
+            // Arrange
+            const arrange = (actAndAssert) => {
+                const fakeTooltipPortalMounter = (((
+                    <div>This is the anchor</div>
+                ): any): React.Element<typeof TooltipPortalMounter>);
+                const nodes = (
+                    <div>
```
kevinbarabash: These tests should use `View`.

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
```diff
@@ -0,0 +1,560 @@
...
+                            {(active) => fakeTooltipPortalMounter}
+                        </TooltipAnchor>
+                    </div>
+                );
+                mount(nodes);
+            };
+
+            const actAndAssert = (ref) => {
+                if (!ref) {
+                    return;
```
kevinbarabash: Does this ever happen?  If so won't the test time out.  Should we may be have
```
expect(true).toBe(false);  // or some other way to force a test failure
done();
```

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
```diff
@@ -0,0 +1,560 @@
...
+                }
+
+                // Act
+                const result = ref.getAttribute("tabindex");
+
+                // Assert
+                expect(result).toBe("-1");
```
kevinbarabash: I would expect this to be "0" since
> When true, the element wrapped by the anchor will be given tabindex=0

Should the docs or the test be changed?

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
```diff
@@ -0,0 +1,560 @@
...
+            const actAndAssert = (ref) => {
+                if (!ref) {
+                    return;
+                }
+
+                // Act
+                const result = ref.getAttribute("tabindex");
+
+                // Assert
+                expect(result).toBeNull();
```
kevinbarabash: This should probably be in the docs as I thought `forceAnchorFocusivity={false}` wouldn't actually do anything.

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
somewhatabstract: Agreed. For now, I think this works well enough.

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
somewhatabstract: Yeah, I am not going to tackle the pointer part just yet. That might have to go into the backlog.

Like any a11y considerations, sometimes we can't do them all. We have to do the best we can. I'd rather think on it a little more before we commit to a solution. Especially since this requires pointer events and that breaks our "fixed/absolute" obscurement checks a little (as they are currently implemented).

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
somewhatabstract: I'll double check.

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
somewhatabstract: Yeah, they didn't used to be :grin: Hangover of old code.

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
somewhatabstract: Indeed.

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
somewhatabstract: Definitely the docs. We don't want to change the tabindex if it's already been set. The assumption should be that it was set for a good reason and we shouldn't interfere with that.

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
somewhatabstract: I was expecting the docs stating "defaults to true" to show that it should do something. Why else would the prop exist?

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
somewhatabstract: It's just to shut `flow` up. I'll change it to be less obvious.

kevinbarabash: `more obvious`?

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
somewhatabstract: I've updated the docs to clarify these.

kevinbarabash: I get what you're saying, but it wasn't obvious from the docs what should happen when the prop is `false`.

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
```diff
@@ -0,0 +1,488 @@
...
+        setTimeout(resolve, 0);
+    });
+};
+
+describe("TooltipAnchor", () => {
+    test("on mount, subscribes to focus and hover events", () => {
+        // Arrange
+        const getFakeTooltipPortalMounter = (active) =>
+            (((
+                <View id="anchor">{active ? "true" : "false"}</View>
```
kevinbarabash: Is the `id` necessary?

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
```diff
@@ -0,0 +1,488 @@
...
+            "mouseleave",
+            expect.any(Function),
+        );
+    });
+
+    test("on unmount, unsubscribes from focus and hover events", () => {
+        // Arrange
+        const getFakeTooltipPortalMounter = (active) =>
+            (((
+                <View id="anchor">{active ? "true" : "false"}</View>
```
kevinbarabash: Same

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
```diff
@@ -0,0 +1,488 @@
...
+
+                // Assert
+                expect(result).toBeNull();
+                done();
+            };
+
+            arrange(actAndAssert);
+        });
+
+        test("if we had added tabindex, removes it", (done) => {
```
kevinbarabash: This test doesn't add `tabindex` anywhere.

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
```diff
@@ -0,0 +1,488 @@
...
+
+            const actAndAssert = (ref) => {
+                // Act
+                // Need to do a timeout so that the mount can return and the
+                // wrapper can change the force prop to false.
+                setTimeout(() => {
+                    const result = ref && ref.getAttribute("tabindex");
+
+                    // Assert
+                    expect(result).not.toBeNull();
```
kevinbarabash: I was misreading these tests before.  I thought that `forceAnchorFocusivity={false}` was resulting in `tabIndex` being removed, but that isn't the case.

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
```diff
@@ -0,0 +1,488 @@
...
+                    </View>
+                );
+                mount(nodes);
+            };
+
+            const actAndAssert = (ref) => {
+                // Act
+                const act = doEvent(ref, new FocusEvent("focusin"))
+                    .then(() => doEvent(ref, new MouseEvent("mouseenter")))
+                    .then(() => doEvent(ref, new FocusEvent("focusout")));
```
kevinbarabash: This is better.  I was actually thinking:
```
test("if hovered, remains active", async () => {
   ...
   await doEvent(ref, new FocusEvent("focusin"));
   await doEvent(ref, new MouseEvent("mouseenter"));
   await doEvent(ref, new FocusEvent("focusout"));
   const result = ref & ref.innerHTML;
   expect(result).toBe("true");
});
```
See http://jestjs.io/docs/en/tutorial-async.html#async-await.  Let's leave it as a TODO for now.  I think using `async`/`await` could simplify some of the other tests here as well.

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
```diff
@@ -0,0 +1,488 @@
...
+                    </View>
+                );
+                mount(nodes);
+            };
+
+            const actAndAssert = (ref) => {
+                // Act
+                const act = doEvent(ref, new MouseEvent("mouseenter"))
+                    .then(() => doEvent(ref, new FocusEvent("focusin")))
+                    .then(() => doEvent(ref, new MouseEvent("mouseleave")));
```
kevinbarabash: Nice test!  All of these tests are really nice.

packages/wonder-blocks-tooltip/util/is-obscured.test.js
```diff
@@ -8,6 +8,24 @@ import {mount} from "enzyme";
 import isObscured from "./is-obscured.js";
 
 describe("isObscured", () => {
+    let ogElementFromPoint = null;
+    beforeEach(() => {
+        // Some tests will want to mock this, so let's ensure we can reset it
+        ogElementFromPoint = document.elementFromPoint;
+    });
+
+    afterEach(() => {
```
kevinbarabash: Thanks for extract the setup and cleanup.

packages/wonder-blocks-tooltip/util/is-obscured.test.js
```diff
@@ -123,9 +132,80 @@ describe("isObscured", () => {
...
             // Assert
             expect(result).toBeTruthy();
-            // Reset the document method to avoid side-effects.
+            done();
+        };
+
+        arrange(actAndAssert);
+    });
+
+    test("element is not obscured, but elementFromPoint returns parent or child, returns false", (done) => {
```
kevinbarabash: I though this test was part of a different PR.  I hope I'm not losing it.

somewhatabstract: You're not...must've got merged forward. Working on rebasing this onto the already reviewed things. Sorry about that.

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
somewhatabstract: Oh yeah, I forgot about that syntax getting added. Used to it from C#.

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
somewhatabstract: No, I don't think so.

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
somewhatabstract: Yes, it does. It will get added on mount.

somewhatabstract: I've added an extra assert to demonstrate that.

PR #151, opened 2018-06-26
Add Tooltip component

somewhatabstract:
This is the last PR that together replace #140.

This builds on #150, adding the actual `Tooltip` component. With this addition, we have something that can anchor to an element (without wrapping it), and show/hide a bubble based on focus/hover and whether parents or other content clips it (at least for most cases).

Upcoming work will focus adding an arrow and content that is driven by our typography as per the implementation spec, as well as various other improvements around styling and accessibility.


jeresig:

Built with commit cd339d7c15dd1431b68024d4a07ab36360d7de02

https://deploy-preview-151--wonder-blocks.netlify.com

packages/wonder-blocks-tooltip/components/tooltip.js
```diff
@@ -1,12 +1,87 @@
...
+    content: string,
+
+    // When true, the child element will be given tabindex=0
+    // to make it keyboard focusable. This value defaults to true. One might set
+    // this to false in circumstances where the wrapped component already can
+    // receive focus or contains an element that can.
+    // Use good judgement when overriding this value, the tooltip content should
+    // be accessible via keyboard in all circumstances where the tooltip would
+    // appear using the mouse, so very those use-cases.
+    forceAnchorFocusivity?: boolean,
```
kevinbarabash: I think we change this, but in a separate PR since there are other files involved.   Maybe `isAnchorFocusable`?

packages/wonder-blocks-tooltip/components/tooltip.js
```diff
@@ -1,12 +1,87 @@
 // @flow
 import * as React from "react";
 
-type Props = {||};
+import {Text as WbText} from "@khanacademy/wonder-blocks-core";
```
kevinbarabash: Please avoid the use of `WbText`.  Instead, use one of the typography components.

packages/wonder-blocks-tooltip/components/tooltip.js
```diff
@@ -1,12 +1,87 @@
...
+    _updateAnchorElement(ref: ?Element) {
+        if (ref && ref !== this.state.anchorElement) {
+            this.setState({anchorElement: ((ref: any): HTMLElement)});
+        }
+    }
+
+    _renderAnchorElement() {
+        const {children} = this.props;
+        if (typeof children === "string") {
+            return <WbText>{children}</WbText>;
```
kevinbarabash: Is this a placeholder before `TooltipContent` is introduced?

packages/wonder-blocks-tooltip/components/tooltip.md
```diff
@@ -0,0 +1,85 @@
...
+### Substring anchor in scrollable parent
+
+```js
+const React = require("react");
+const {View, Text} = require("@khanacademy/wonder-blocks-core");
+const {Tooltip} = require("@khanacademy/wonder-blocks-tooltip");
+
+<div>
+    <div style={{height: 100, overflow: "auto", border: "1px solid", margin: 10,}}>
+        <div style={{height: "200vh"}}>
```
kevinbarabash: Please use `View` in the examples.

packages/wonder-blocks-tooltip/components/tooltip.md
```diff
@@ -0,0 +1,85 @@
...
+            <Text>This is a big long piece of text with a </Text>
```
kevinbarabash: Please a specific typography component.  I'm planning to hide or change `Text` since it doesn't match React Native's `Text` component so we should avoid use it.

packages/wonder-blocks-tooltip/components/tooltip.test.js
```diff
@@ -0,0 +1,4 @@
+//@flow
+describe("Tooltip", () => {
+    test("TODO", () => {});
```
kevinbarabash: Will this be added in a future PR?

packages/wonder-blocks-tooltip/components/tooltip.js
somewhatabstract: No, this is so that if we try to anchor on text, we wrap it into something that we can actually anchor.

somewhatabstract: This code changes somewhat as we move forward.

packages/wonder-blocks-tooltip/components/tooltip.test.js
somewhatabstract: Yup. Eventually.

packages/wonder-blocks-tooltip/components/tooltip.js
somewhatabstract: I don't want to give the illusion that we're turning focus off ever.

How about `forceAnchorToBeFocusable`?

packages/wonder-blocks-tooltip/components/tooltip.js
somewhatabstract: This is specifically to avoid those. We don't want to apply any styling changes at all. This is to allow us to anchor on a bit of a string in the middle of a typography element. So we want to be as light as we can. I believe this is a valid use of `Text` here.

packages/wonder-blocks-tooltip/components/tooltip.js
somewhatabstract: It's deliberate so that we don't affect styling, but can anchor to the text.

packages/wonder-blocks-tooltip/components/tooltip.md
somewhatabstract: These are all changing in later PRs as more components are introduced, so I'm going leave as-is here.

packages/wonder-blocks-tooltip/components/tooltip.md
somewhatabstract: Again, these are all changing in one of the other PRs, so I would prefer to leave as-is here, and update there.

packages/wonder-blocks-tooltip/components/tooltip.js
kevinbarabash: Ugh... let's leave it for now.

packages/wonder-blocks-tooltip/components/tooltip.js
kevinbarabash: I just had a look at `Text` and it's a `<span>` which is `inline` which is great.  🙂  There was some talk about making it a `div` b/c all of the typography elements are block-level elements.

PR #152, opened 2018-06-27
Add TooltipContent component

somewhatabstract:
This PR implements the `TooltipContent` component that is used to render the content of the `TooltipBubble`. In this change, we add the new component and some test documentation for it, as well as update the other components to support it.

We also add some helper types to `wonder-blocks-typography` to simplify how we specify what elements are valid.

### DIFF
--- a/packages/wonder-blocks-tooltip/components/tooltip-content.js
+++ b/packages/wonder-blocks-tooltip/components/tooltip-content.js
@@ -1,6 +1,8 @@
 // @flow
 import * as React from "react";
 import {View} from "@khanacademy/wonder-blocks-core";
+import {Strut} from "@khanacademy/wonder-blocks-layout";
+import Spacing from "@khanacademy/wonder-blocks-spacing";
 import {HeadingSmall, LabelMedium} from "@khanacademy/wonder-blocks-typography";
 import type {Typography} from "@khanacademy/wonder-blocks-typography";
 
@@ -42,6 +44,7 @@ export default class TooltipContent extends React.Component<Props> {
         return (
             <View>
                 {this._renderTitle()}
+                <Strut size={Spacing.xxxSmall} />
                 {this._renderChildren()}
             </View>
         );
--- a/packages/wonder-blocks-tooltip/components/tooltip-content.js
+++ b/packages/wonder-blocks-tooltip/components/tooltip-content.js
@@ -1,0 +1,49 @@
+// @flow
+import * as React from "react";
+import {View} from "@khanacademy/wonder-blocks-core";
+import {HeadingSmall, LabelMedium} from "@khanacademy/wonder-blocks-typography";
+import type {Typography} from "@khanacademy/wonder-blocks-typography";
+
+type Props = {|
+    title?: string | React.Element<Typography>,
+    children:
+        | string
+        | React.Element<Typography>
+        | Array<React.Element<Typography>>,
+|};
+
+/**
+ * This component is used to provide the content that is to be rendered in the
+ * tooltip bubble.
+ */
+export default class TooltipContent extends React.Component<Props> {
+    _renderTitle() {
+        const {title} = this.props;
+        if (title) {
+            if (typeof title === "string") {
+                return <HeadingSmall>{title}</HeadingSmall>;
+            } else {
+                return title;
+            }
+        }
+        return null;
+    }
+
+    _renderChildren() {
+        const {children} = this.props;
+        if (typeof children === "string") {
+            return <LabelMedium>{children}</LabelMedium>;
+        } else {
+            return children;
+        }
+    }
+
+    render() {
+        return (
+            <View>
+                {this._renderTitle()}
+                {this._renderChildren()}
+            </View>
+        );
+    }
+}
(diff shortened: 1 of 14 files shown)


packages/wonder-blocks-tooltip/components/tooltip-content.js
```diff
@@ -0,0 +1,49 @@
+// @flow
+import * as React from "react";
+import {View} from "@khanacademy/wonder-blocks-core";
+import {HeadingSmall, LabelMedium} from "@khanacademy/wonder-blocks-typography";
+import type {Typography} from "@khanacademy/wonder-blocks-typography";
+
+type Props = {|
+    title?: string | React.Element<Typography>,
```
somewhatabstract: Question: Do we want to allow any old `Text` as well? Currently we're limiting to `Typography` components only. Just wondering if we should have the escape hatch of allowing any `Text` for these props.

jeresig:

Built with commit 233b2c92bdf379db58a83482fb8d5eb9a71091fc

https://deploy-preview-152--wonder-blocks.netlify.com

packages/wonder-blocks-typography/index.js
```diff
@@ -18,6 +18,45 @@ import Tagline from "./components/tagline.js";
...
+/**
+ * All typography components.
+ */
+export type Typography =
+    | Heading
+    | BodyText
+    | Label
+    | typeof Tagline
+    | typeof Caption
+    | typeof Footnote;
```
somewhatabstract: Couldn't put these types into the `types.js` as that would create a cyclic dependency among the components. Should we have an `external-types.js` file or something? Or are these ok here? I like them here since it means that if we add a new component, it's clear these need updating too.

packages/wonder-blocks-tooltip/components/tooltip-content.js
```diff
@@ -0,0 +1,52 @@
...
+            return children;
+        }
+    }
+
+    render() {
+        return (
+            <View>
+                {this._renderTitle()}
+                <Strut size={Spacing.xxxSmall} />
+                {this._renderChildren()}
```
kevinbarabash: I personally like this pattern, but feel free to add `paddingBottom` on the title if you'd like.

packages/wonder-blocks-tooltip/components/tooltip-content.js
```diff
@@ -0,0 +1,52 @@
...
+ * tooltip bubble.
+ */
+export default class TooltipContent extends React.Component<Props> {
+    _renderTitle() {
+        const {title} = this.props;
+        if (title) {
+            if (typeof title === "string") {
+                return <HeadingSmall>{title}</HeadingSmall>;
+            } else {
+                return title;
```
kevinbarabash: Cool.  I wasn't thinking we'd make the title configurable, but this works for me.

packages/wonder-blocks-tooltip/components/tooltip-content.md
```diff
@@ -0,0 +1,34 @@
...
+
+This shows how we can customize both the title and the content.
+
+```jsx
+const {Body, LabelSmall} = require("@khanacademy/wonder-blocks-typography");
+
+<TooltipContent title={<Body>Body text title!</Body>}>
+    <Body>Body text content!</Body>
+    <LabelSmall>And LabelSmall!</LabelSmall>
+</TooltipContent>
```
kevinbarabash: Nice examples!

packages/wonder-blocks-typography/index.js
kevinbarabash: These are okay here.  Thanks for adding these groupings.

packages/wonder-blocks-tooltip/components/tooltip.md
```diff
@@ -2,7 +2,6 @@
 
 ```js
-const {Tooltip} = require("@khanacademy/wonder-blocks-tooltip");
```
kevinbarabash: Thanks for cleaning this up.

packages/wonder-blocks-tooltip/components/tooltip.js
```diff
@@ -1,22 +1,27 @@
 // @flow
 import * as React from "react";
 
-import {Text as WbText} from "@khanacademy/wonder-blocks-core";
+import {Text} from "@khanacademy/wonder-blocks-core";
```
kevinbarabash: Use something from `wonder-blocks-typography` instead.

packages/wonder-blocks-tooltip/components/tooltip-content.js
somewhatabstract: I think this works great, though there's a bug here that I fix in the later styling PR #159

packages/wonder-blocks-tooltip/components/tooltip.js
somewhatabstract: See my comment in the other PR. I don't want layout nor styling to be changed by this, hence the use of `Text`.

somewhatabstract:
@kevinbarabash I don't think there's any changes to make here. See my comments re: use of `Text`.

packages/wonder-blocks-tooltip/components/tooltip.js
kevinbarabash: Sounds like the use of `Text` will eventually be replaced by something else in a future diff.

PR #153, opened 2018-06-27
Add SingleSelectMenu to dropdown package

wishdasher:


jeresig:

Built with commit c842b0643cfa00517493429da3514d7844e5cfc2

https://deploy-preview-153--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/single-select-menu.md
```diff
@@ -0,0 +1,114 @@
...
+    },
+    strutLike: {
+        width: 8,
+    }
+});
+<View style={[styles.row]}>
+    <SingleSelectMenu
+        items={[
+            {
+                type: "select",
```
kevinbarabash: This should always be `type: "select"` for `SingleSelectMenu` so we shouldn't need to pass this in.

packages/wonder-blocks-dropdown/components/single-select-menu.md
```diff
@@ -0,0 +1,114 @@
...
+            },
+            {
+                type: "select",
+                label: "Grape",
+                selected: false,
+                value: "grape"
+            },
+        ]}
+        light={false}
+        onChange={(s) => console.log(s)}
```
kevinbarabash: Users of this component will probably want to use the current selection later so they'll need to store it in `state` somewhere.  Since they're already storing it in state it makes sense to pass it as a prop to `SingleSelectMenu` so that it doesn't get out of sync with what `SingleSelectMenu` is displaying.  Can you update this example to define a parent component that tracks the selected value just like devs would have to do to use this this for real?

kevinbarabash: We may also want to replace the `select` property on each of the `items` with a single `value` prop on `SingleSelectMenu` itself.  There are a couple of components that do this, e.g. https://jedwatson.github.io/react-select/, https://www.npmjs.com/package/react-select.

packages/wonder-blocks-dropdown/components/single-select-menu.js
```diff
@@ -0,0 +1,185 @@
...
+        };
+    }
+
+    toggleMenu() {
+        this.setState((prevState) => ({
+            open: !prevState.open,
+        }));
+    }
+
+    handleSelected(selectedValue: string, selectedLabel: string) {
```
kevinbarabash: This can be moved to the example.  See my comments in docs.md.

packages/wonder-blocks-dropdown/components/single-select-menu.js
```diff
@@ -0,0 +1,185 @@
...
+    /**
+     * Placeholder for the opening component when there are no items selected.
+     */
+    placeholder: string,
+
+    /**
+     * Whether this menu should be left-aligned or right-aligned with the
+     * opener component. Defaults to left-aligned.
+     */
+    alignment?: "left" | "right",
```
kevinbarabash: The examples should show this in action.  Maybe make the labels for the examples a bit longer.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -0,0 +1,95 @@
...
+    rightAlign: {
+        alignItems: "flex-end",
+    },
+
+    dropdown: {
+        backgroundColor: Color.white,
+        position: "absolute",
+        borderRadius: 4,
+        marginTop: Spacing.xxxSmall,
+        marginBottom: Spacing.xxxSmall,
```
kevinbarabash: Is `marginBottom` actually necessary?

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -0,0 +1,95 @@
...
+
+import Color, {fade} from "@khanacademy/wonder-blocks-color";
+import Spacing from "@khanacademy/wonder-blocks-spacing";
+import {View} from "@khanacademy/wonder-blocks-core";
+
+type DropdownCoreProps = {
+    /**
+     * Items for the menu.
+     */
+    items: Array<React.Node>,
```
kevinbarabash: Can we do something like:
```
items: Array<Element<typeof ActionItem> | Element<typeof SelectItem> | ...>
```

packages/wonder-blocks-dropdown/components/select-item.js
```diff
@@ -47,10 +46,10 @@ type SelectProps = {
 
     /**
      * Callback for when this item is pressed to change its selection state.
-     * Passes value of the item and the new selection state of the item. Should
+     * Passes value of the item, its label, and its new selection state. Should
      * be handled by an implementation of Menu.
      */
-    onToggle: (value: string, selectionState: boolean) => void,
+    onToggle: (value: string, label: string, selectionState: boolean) => void,
```
kevinbarabash: The `label` doesn't seem all that useful since whatever component is passing down data can always look up the `label` based on the `value`.

packages/wonder-blocks-dropdown/components/single-select-menu.js
```diff
@@ -0,0 +1,185 @@
...
+import {StyleSheet} from "aphrodite";
+
+import DropdownCore from "./dropdown-core.js";
+import SelectBox from "./select-box.js";
+import SelectItem from "./select-item.js";
+import SeparatorItem from "./separator-item.js";
+
+import type {SelectItemProps, SeparatorProps} from "../utils/types.js";
+
+type ItemProps = SelectItemProps | SeparatorProps;
```
kevinbarabash: Can you inline this so that it shows up in the docs?

packages/wonder-blocks-dropdown/components/single-select-menu.js
```diff
@@ -0,0 +1,185 @@
...
+     * opener component. Defaults to left-aligned.
+     */
+    alignment?: "left" | "right",
+
+    /**
+     * Whether to display the "light" version of this component instead, for
+     * use when the item is used on a dark background.
+     */
+
+    light?: boolean,
```
kevinbarabash: Nit: extra space.

packages/wonder-blocks-dropdown/components/single-select-menu.js
```diff
@@ -0,0 +1,185 @@
...
+     */
+    open: boolean,
+
+    /**
+     * The value and label of the selected element.
+     */
+    selected: {
+        value: string,
+        label: string,
+    },
```
kevinbarabash: Move to example.  See comments in docs.md.  Also, we only really need the `value` since the parent can always look up the `label` if needed.

packages/wonder-blocks-dropdown/components/single-select-menu.md
```diff
@@ -0,0 +1,114 @@
...
+        placeholder={"Choose a juice"}
+        style={{
+            width: 170,
+            maxWidth: 170,
+        }}
+    />
+
+    <View style={[styles.strutLike]} />
+
+    <View style={[styles.darkBackgroundWrapper]}>
```
kevinbarabash: Probably best to have this in a separate example.

packages/wonder-blocks-dropdown/components/single-select-menu.md
wishdasher: This is like this because I wasn't sure how to denote where a `SeparatorItem` would go without passing it in here (`type: "separator`), so I used [this pattern](https://flow.org/en/docs/types/unions/) to account for that.

wishdasher: Actually, I'll leave the separator out of this one for now. I don't really see a use case where you would use a separator for a single select menu (will discuss with design).

packages/wonder-blocks-dropdown/components/single-select-menu.md
```diff
@@ -0,0 +1,227 @@
...
+                    label: "Apple",
+                    value: "apple"
+                },
+                {
+                    label: "Grape",
+                    value: "grape"
+                },
+            ]}
+            light={false}
+            onChange={(selected) => this.handleSelected(selected)}
```
kevinbarabash: super nit: `this.handleChange(selected)` as per the https://github.com/Khan/style-guides/blob/master/style/react.md#name-handlers-handleeventname.

packages/wonder-blocks-dropdown/components/single-select-menu.js
```diff
@@ -0,0 +1,169 @@
...
+            placeholder,
+            selectedValue,
+            style,
+        } = this.props;
+
+        const {open} = this.state;
+
+        const menuText = `${
+            selectedValue
+                ? items.filter((item) => item.value === selectedValue)[0].label
```
kevinbarabash: We can use `items.find(item => item.value === selectedValue)`.  Also, we should probably handle the case where we get passed a `selectedValue` that doesn't exist in `items`.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -0,0 +1,96 @@
...
+        // The space between the opener and the top of the menu
+        marginTop: Spacing.xxxSmall,
+        paddingTop: Spacing.xxxSmall,
+        paddingBottom: Spacing.xxxSmall,
+        border: `solid 1px ${Color.offBlack16}`,
+        boxShadow: `0px 8px 8px 0px ${fade(Color.offBlack, 0.1)}`,
+        top: 40,
+        // NOTE: This is the z-index of the currently existing menus in the
+        // webapp. Perhaps wonderblocks will design a z-index hierarchy?
+        zIndex: 1000,
```
kevinbarabash: We'll be able to get rid of this when we switch to the portal.  Maybe change this to a `TODO` to remove once the portal is in place.

packages/wonder-blocks-dropdown/components/single-select-menu.md
```diff
@@ -0,0 +1,227 @@
...
+        />;
+    }
+}
+
+<View style={[styles.row]}>
+    <ExampleWithPlaceholder />
+    <View style={[styles.strutLike]} />
+    <ExampleWithStartingSelection />
+    <View style={[styles.strutLike]} />
+    <DisabledExample />
```
kevinbarabash: Can you split these up into separate examples?

packages/wonder-blocks-dropdown/components/single-select-menu.js
wishdasher: Didn't know about this function, thanks.

packages/wonder-blocks-dropdown/components/single-select-menu.js
```diff
@@ -0,0 +1,168 @@
...
+            selectedValue,
+            style,
+        } = this.props;
+
+        const {open} = this.state;
+
+        const selectedItem = items.find((item) => item.value === selectedValue);
+        // If nothing is selected, or if the selectedValue doesn't match any
+        // item in the menu, use the placeholder.
+        const menuText = `${selectedItem ? selectedItem.label : placeholder}`;
```
kevinbarabash: The template string isn't necessary b/c `placeholder` and `selectedItem.label` are already strings.

PR #154, opened 2018-06-27
Adds Toolbar component

scotchfield:
This is a first pass implementing most of the spec outlined in our documentation. I still have to write tests, and to implement the responsive reformatting properly.


packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -0,0 +1,206 @@
...
+// TODO(scottgrant): Make sure the toolbars are responsive.
+// import type {MediaSize} from "@khanacademy/wonder-blocks-core";
+
+const dismissIcon =
+    "M12.003 10.588L7.707 6.293a1 1 0 0 0-1.414 1.414l4.295 4.296-4.295 4.295a1 1 0 0 0 1.414 1.414l4.296-4.295 4.295 4.295a1 1 0 0 0 1.414-1.414l-4.295-4.295 4.295-4.296a1 1 0 1 0-1.414-1.414l-4.295 4.295z";
+
+type Props = {
+    /**
+     * A list of nodes to render on the left side of the toolbar. In most
+     * cases, this will be empty.
```
kevinbarabash: Maybe mention that this is used for the close button on modals.

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -0,0 +1,206 @@
+// @flow
```
kevinbarabash: Top-level docstring.

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -0,0 +1,206 @@
...
+     * A list of nodes to render on the right side of the toolbar. This will
+     * typically include buttons, links, span elements with text, or nothing
+     * at all.
+     */
+    rightContent?: Array<React.Node>,
+
+    /**
+     * True, if we should show the dismiss icon on the left side o the toolbar.
+     */
+    showClose: boolean,
```
kevinbarabash: It feels weird to have both `leftContent` and `showClose` as they seems mutually exclusive.  The modal header should be responsible for passing an `IconButton` with the correct icon and the appropriate `click` handler to the `leftContent` prop of the `Toolbar`.

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -0,0 +1,206 @@
...
+
+    /**
+     * A list of nodes to render on the right side of the toolbar. This will
+     * typically include buttons, links, span elements with text, or nothing
+     * at all.
+     */
+    rightContent?: Array<React.Node>,
+
+    /**
+     * True, if we should show the dismiss icon on the left side o the toolbar.
```
kevinbarabash: nit: `...side of the toolbar`

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -0,0 +1,206 @@
...
+
+    /**
+     * How much vertical space to use for the toolbar. If this prop is not
+     * provided, the default is "medium".
+     */
+    size?: "small" | "medium",
+
+    /**
+     * Additional styling for the toolbar. If a color value is provided and
+     * the showClose prop is set to true, the close icon will use this color.
```
kevinbarabash: This seems overly permissive.  We should probably go with `color: "dark" | "light"` to follow what was done in the modals.  According to design the only two colors allowed are `white` and `darkBlue`.  Are there are any other styles besides color that we need to allow?

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -0,0 +1,206 @@
...
+        const fillColor = (style && style.color) || "rgba(33, 36, 44, 0.5)";
+        const closeButton = showClose && (
+            <View onClick={onClose}>
+                <svg role="img" width="24px" height="24px">
+                    <path fill={fillColor} d={dismissIcon} />
+                </svg>
+            </View>
+        );
+
+        // TODO(scottgrant): Handle aria-labelledby appropriately.
```
kevinbarabash: Is this just for the close button or something else?

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -0,0 +1,206 @@
...
+                        )}
+                </View>
+                <View style={[sharedStyles.column]}>
+                    <View
+                        style={[
+                            sharedStyles.middleColumn,
+                            sharedStyles.verticalAlign,
+                        ]}
+                    >
+                        <span
```
kevinbarabash: Please use components from `wonder-blocks-typography` instead.

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -0,0 +1,206 @@
...
+                            id="wb-toolbar-title"
+                            className={css(
+                                sharedStyles.title,
+                                !subtitle && sharedStyles.titleNoSubtitle,
+                            )}
+                        >
+                            {title}
+                        </span>
+                        {subtitle && (
```
kevinbarabash: same

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -0,0 +1,206 @@
...
+        padding: 8,
+        position: "relative",
+        width: "100%",
+    },
+    small: {
+        minHeight: 48,
+    },
+    verticalAlign: {
+        display: "flex",
+        flexDirection: "column",
```
kevinbarabash: `View`s have `display: "flex"; flexDirection: "column"` set by default.

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -0,0 +1,206 @@
...
+        justifyContent: "center",
+    },
+    column: {
+        display: "block",
+        position: "absolute",
+        left: "50%",
+        top: "50%",
+        transform: "translate(-50%, -50%);",
+        width: "80%",
```
kevinbarabash: How do we handle collisions between columns?  We might need different styles for the columns depending on what props have been passed in, e.g. if there's a title then the left/right columns are constrained to be a certain width.

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,192 @@
+> TODO(scottgrant): Finish building, and write more thorough docs.
+
+## Content toolbar
+
+### Standard view
+
+```js
+<Toolbar
+	contentKind="Video"
```
kevinbarabash: This isn't actually a prop on `Toolbar`.

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,192 @@
+> TODO(scottgrant): Finish building, and write more thorough docs.
+
+## Content toolbar
+
+### Standard view
```
kevinbarabash: Instead of giving these labels, descriptions are sufficient, e.g. `Toolbar with title`

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,192 @@
+> TODO(scottgrant): Finish building, and write more thorough docs.
+
+## Content toolbar
```
kevinbarabash: Grouping the examples into specific toolbar types isn't necessary.  It might've made sense if there was a specific `type` or `kind` prop but we don't have one.

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,192 @@
...
+
+```js
+<Toolbar
+	contentKind="Video"
+	showClose={true}
+	title="Counting with small numbers"
+/>
+```
+
+### Content completed view
```
kevinbarabash: `Toolbar with left and right content`

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+## Exercise toolbar
+
+### Generic header
+
+```js
+const Link = require("@khanacademy/wonder-blocks-link").default;
+
+// TODO(scottgrant): This is a placeholder icon. The icon in Zeplin was rotated.
```
kevinbarabash: FYI `Link` is eventually supposed to have a `caret` prop so this should be really easy in the future.

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,192 @@
...
+### Content completed view
+
+```js
+const Button = require("@khanacademy/wonder-blocks-button").default;
+
+const startExercise = <Button>Next Video</Button>;
+
+<Toolbar
+	contentKind="Video"
+	rightContent={[startExercise]}
```
kevinbarabash: I wonder if we should allow people to pass either an array or a single element.  Having to wrap single items in an array adds friction.

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,192 @@
...
+    </svg>
+);
+//const goToExercise = <Link>Go to exercise {caretIcon}</Link>;
+const goToExercise = <a href="#">Go to exercise</a>;
+
+<Toolbar
+	rightContent={[goToExercise]}
+	showClose={true}
+	subtitle="0 of 7 questions answered"
+	title="Exercise Title"
```
kevinbarabash: This example doesn't add much b/c we've already covered these props in other examples.

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,192 @@
...
+
+```js
+const Button = require("@khanacademy/wonder-blocks-button").default;
+
+const nextExercise = <Button>Next exercise</Button>;
+const questionCount = <span>7 questions</span>;
+const tryAgain = <Button kind="secondary">Try again</Button>;
+
+<Toolbar
+	rightContent={[questionCount, tryAgain, nextExercise]}
```
kevinbarabash: This is an important example b/c it shows off the ordering of elements.  I like this example better than the previous one b/c it has more buttons.  Since they're so similar let's drop the previous one.

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,192 @@
...
+
+<Toolbar
+	rightContent={[goToExercise]}
+	showClose={true}
+	subtitle="0 of 7 questions answered"
+	title="Exercise Title"
+/>
+```
+
+### Start of task card
```
kevinbarabash: `## Toolbar with multiple elements on the right`

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,192 @@
...
+	rightContent={[questionCount, startExercise]}
+	showClose={false}
+/>
+```
+
+```js
+const Button = require("@khanacademy/wonder-blocks-button").default;
+
+const nextExercise = <Button>Next exercise</Button>;
+const questionCount = <span>7 questions</span>;
```
kevinbarabash: Please use one of the typography components, maybe `LabelMedium` or `LabelLarge`.

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,192 @@
...
+		<span style={{marginRight: 12}}>Get 5 of 7 questions to level up to Familiar</span>
+		{circleComplete} {circleComplete}
+		{circleActive}
+		{circleIncomplete} {circleIncomplete} {circleIncomplete} {circleIncomplete}
+	</span>
+);
+const nextQuestion = <Button>Next question</Button>;
+
+<Toolbar
+	rightContent={[questionCount, nextQuestion]}
```
kevinbarabash: This example is overkill and doesn't add anything new.

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,192 @@
...
+
+// TODO(scottgrant): This is a placeholder icon. The icon in Zeplin was rotated.
+const caretPath = "M18.629 15.997l-7.083-7.081L13.462 7l8.997 8.997L13.457 25l-1.916-1.916z";
+const caretIcon = (
+    <svg role="img" width="24px" height="24px">
+        <path fill="#1865f2" d={caretPath} />
+    </svg>
+);
+//const goToExercise = <Link>Go to exercise {caretIcon}</Link>;
+const goToExercise = <a href="#">Go to exercise</a>;
```
kevinbarabash: Please use `Link` for this.  One purpose of the documentation is to encourage people to use wonder-blocks components whenever there's one available.

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,192 @@
...
+<Toolbar
+	contentKind="Exercise"
+	rightContent={[goToExercise]}
+	showClose={true}
+	subtitle="1 of 14 questions answered"
+	title="Patterns of migration and communal bird-feeding given the serious situation of things that will make this string long and obnoxious"
+/>
+```
+
+## Video & Article toolbars
```
kevinbarabash: These examples can be dropped b/c they don't show anything new.

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,192 @@
...
+	backgroundColor: "#0a2a66",
+	boxShadow: "0 1px 0 0 rgba(255, 255, 255, 0.64)",
+	color: "white",
+};
+
+<Toolbar
+	showClose={true}
+	style={style}
+	title="Title"
+/>
```
kevinbarabash: These examples should show `Toolbar` parented in a `View` since that's how they'll be used.

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -0,0 +1,206 @@
...
+
+const dismissIcon =
+    "M12.003 10.588L7.707 6.293a1 1 0 0 0-1.414 1.414l4.295 4.296-4.295 4.295a1 1 0 0 0 1.414 1.414l4.296-4.295 4.295 4.295a1 1 0 0 0 1.414-1.414l-4.295-4.295 4.295-4.296a1 1 0 1 0-1.414-1.414l-4.295 4.295z";
+
+type Props = {
+    /**
+     * A list of nodes to render on the left side of the toolbar. In most
+     * cases, this will be empty.
+     */
+    leftContent?: Array<React.Node>,
```
jeresig: 🛑 Why are these arrays? It seems like having them just be `React.Node` (which can also be an array of React Nodes) is fine.

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -0,0 +1,206 @@
...
+    /**
+     * True, if we should show the dismiss icon on the left side o the toolbar.
+     */
+    showClose: boolean,
+
+    /**
+     * How much vertical space to use for the toolbar. If this prop is not
+     * provided, the default is "medium".
+     */
+    size?: "small" | "medium",
```
jeresig: 🛑 I don't see the default specified in defaultProps (and this prop would then be required, not optional).

packages/wonder-blocks-toolbar/components/toolbar.js
jeresig: 🛑 Yeah, I think I agree with this, it would certainly make the implementation simpler!

packages/wonder-blocks-toolbar/components/toolbar.js
scotchfield: Thanks, I added a simple one at the top.

packages/wonder-blocks-toolbar/components/toolbar.js
scotchfield: Good call! I've removed the specific `onClose` and `showClose` props in favour of `leftContent` and updated the documentation.

packages/wonder-blocks-toolbar/components/toolbar.js
scotchfield: Good idea, I updated the comment.

packages/wonder-blocks-toolbar/components/toolbar.js
scotchfield: Eek! Thank you. This comment went away, but I still don't love the typo. 😳

packages/wonder-blocks-toolbar/components/toolbar.js
scotchfield: Thank you, good call. I've had a bit of back-and-forth over how general or specific we want these components to be. You're right though; since we only have the two styles, it makes sense to stick to those for now. I've updated the component here (and sadly stuck with the standardized spelling of `color` with no u)

packages/wonder-blocks-toolbar/components/toolbar.js
scotchfield: This is something we'll need to address with the modals, but I think for now this comment can be removed. I've left an `id` property on the title so it can be referenced using `aria-labelledby` if needed, and since we pass in other children as content items on the left and right, it's possible for parent components to add `id` properties if they're needed.

packages/wonder-blocks-toolbar/components/toolbar.js
scotchfield: Oops, thank you! Done.

packages/wonder-blocks-toolbar/components/toolbar.js
scotchfield: Thanks! I've removed these extra properties.

packages/wonder-blocks-toolbar/components/toolbar.js
scotchfield: I'm going to take another look at this. The variable width columns and requirement that the title is right in the middle was a bit of a snag. There are some overlap issues here with long titles, so I'll take a closer look now.

packages/wonder-blocks-toolbar/docs.md
scotchfield: Good eye! This was from an earlier prototype and it's removed now.

packages/wonder-blocks-toolbar/docs.md
scotchfield: I wasn't sure whether the intent was to reproduce every example from our design or to just take the core set. This makes sense though, thanks!

packages/wonder-blocks-toolbar/docs.md
scotchfield: Whew. That's great to hear!

packages/wonder-blocks-toolbar/docs.md
scotchfield: John mentioned this too. I've updated the component to accept a single component or an array now, and handle that appropriately in `renderContent`.

packages/wonder-blocks-toolbar/docs.md
scotchfield: 👍

packages/wonder-blocks-toolbar/docs.md
scotchfield: It was kind of fun to make though. 😂

packages/wonder-blocks-toolbar/docs.md
scotchfield: 👍

packages/wonder-blocks-toolbar/docs.md
scotchfield: I wrapped it in a `View` and added some placeholder text below to show how it's used.

packages/wonder-blocks-toolbar/components/toolbar.js
scotchfield: Good eye, thanks! I've updated the defaults.

packages/wonder-blocks-toolbar/components/toolbar.js
scotchfield: You're right, and Kevin brought this up too. I've updated the component so either is fine.

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,97 @@
...
+const Button = require("@khanacademy/wonder-blocks-button").default;
+const {LabelMedium} = require("@khanacademy/wonder-blocks-typography");
+
+const nextExercise = <Button>Next exercise</Button>;
+const questionCount = <LabelMedium>7 questions</LabelMedium>;
+const tryAgain = <Button kind="secondary">Try again</Button>;
+
+<Toolbar
+	rightContent={[questionCount, tryAgain, nextExercise]}
+	showClose={false}
```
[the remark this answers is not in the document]
scotchfield: Will do.

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,97 @@
...
+
+const style = {
+	backgroundColor: "#0a2a66",
+	boxShadow: "0 1px 0 0 rgba(255, 255, 255, 0.64)",
+	color: "white",
+	height: 200,
+};
+
+<View style={style}>
+	<Toolbar color="dark" showClose={true} title="Title" />
```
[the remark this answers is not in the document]
scotchfield: Same, thanks for catching these leftovers!

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,97 @@
+> TODO(scottgrant): Finish building, and write more thorough docs.
+
+## Toolbar with title
+
+```js
+const dismissIcon =
+    "M12.003 10.588L7.707 6.293a1 1 0 0 0-1.414 1.414l4.295 4.296-4.295 4.295a1 1 0 0 0 1.414 1.414l4.296-4.295 4.295 4.295a1 1 0 0 0 1.414-1.414l-4.295-4.295 4.295-4.296a1 1 0 1 0-1.414-1.414l-4.295 4.295z";
+const closeButton = (
```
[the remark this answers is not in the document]
scotchfield: I added a TODO so we don't forget about this.

packages/wonder-blocks-toolbar/components/toolbar.md
```diff
@@ -0,0 +1,11 @@
+### Small toolbar example:
```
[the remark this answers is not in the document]
scotchfield: No good reason! I'll delete and move these examples to the main doc.

packages/wonder-blocks-toolbar/components/toolbar.md
```diff
@@ -0,0 +1,11 @@
...
+
+```js
+<Toolbar size="small" />
```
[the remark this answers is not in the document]
scotchfield: Will do!

packages/wonder-blocks-toolbar/components/toolbar.md
```diff
@@ -0,0 +1,11 @@
...
+```
+
+### Medium toolbar example:
+
+```js
+<Toolbar size="medium" />
```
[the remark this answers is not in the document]
scotchfield: That's a fair point. I'll remove the medium example since we have so many other examples.

packages/wonder-blocks-toolbar/index.js
```diff
@@ -0,0 +1,4 @@
+// @flow
+import Toolbar from "./components/toolbar.js";
+
+export {Toolbar};
```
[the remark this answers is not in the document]
scotchfield: I'd expected to have a few cases originally, but not anymore! You're right; I'll make `Toolbar` be the default.

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,106 @@
+> TODO(scottgrant): Finish building, and write more thorough docs.
+
+## Small toolbar example:
+
+```js
+const Button = require("@khanacademy/wonder-blocks-button").default;
+const smallButton = <Button size="small">Small button</Button>;
+
+<Toolbar size="small" rightContent={[smallButton, smallButton]} />
```
kevinbarabash: There should only ever be one "primary" button in a toolbar.  Can you change one of these to be a "secondary" button?

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,106 @@
+> TODO(scottgrant): Finish building, and write more thorough docs.
```
kevinbarabash: For some reason why I click on the https://deploy-preview-154--wonder-blocks.netlify.com/#!/Toolbar link it doesn't show the examples like it does for the other docs.

packages/wonder-blocks-toolbar/docs.md
```diff
@@ -0,0 +1,106 @@
...
+const nextExercise = <Button>Next exercise</Button>;
+const questionCount = <LabelMedium>7 questions</LabelMedium>;
+const tryAgain = <Button kind="secondary">Try again</Button>;
+
+<Toolbar
+	rightContent={[questionCount, tryAgain, nextExercise]}
+/>
+```
+
+## Header overflow text
```
kevinbarabash: Does this do the right thing on mobile?  I wasn't able to test it on the styleguidist page.

packages/wonder-blocks-toolbar/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "main": "dist/index.js",
+  "source": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "MIT",
+  "dependencies": {
+    "@khanacademy/wonder-blocks-core": "^1.0.1"
+  }
```
kevinbarabash: You'll need to include `react` and `aphrodite` as `peerDependencies`.

packages/wonder-blocks-toolbar/docs.md
scotchfield: Yep, will do.

packages/wonder-blocks-toolbar/docs.md
scotchfield: Looks like styleguidist might only do this if the examples are associated with a doc in the components directory. I extracted the examples to `wonder-blocks-toolbar/components/toolbar.md` and they show up now.

packages/wonder-blocks-toolbar/docs.md
scotchfield: The examples here don't restructure themselves based on the docs in Zeplin and there's some overflow with long words on the right side (`exercise` overflows), but the spacing works as expected. There's a note about assuming components that use `Toolbar` will use a media wrapper to specify if we should render differently based on size. Do you think that seems okay?

packages/wonder-blocks-toolbar/package.json
scotchfield: Will do!

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -0,0 +1,178 @@
...
+        minHeight: 64,
+        padding: 8,
+        position: "relative",
+        width: "100%",
+    },
+    small: {
+        minHeight: 48,
+    },
+    dark: {
+        backgroundColor: "#0a2a66",
```
kevinbarabash: Use `Color.darkBlue` from `wonder-blocks-color`.

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -0,0 +1,178 @@
...
+        boxShadow: "0 1px 0 0 rgba(255, 255, 255, 0.64)",
+        color: "white",
+    },
+    verticalAlign: {
+        justifyContent: "center",
+    },
+    column: {
+        display: "flex",
```
kevinbarabash: Not necessary.

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -0,0 +1,178 @@
...
+        flex: 1,
+        justifyContent: "center",
+    },
+    wideColumn: {
+        flexBasis: "50%",
+    },
+    rightColumn: {
+        display: "flex",
```
kevinbarabash: Not necessary.

packages/wonder-blocks-toolbar/components/toolbar.md
```diff
@@ -0,0 +1,105 @@
...
+const smallButton = <Button size="small">Small button</Button>;
+const secondaryButton = <Button size="small" kind="secondary">Secondary button</Button>;
+
+<Toolbar size="small" rightContent={[smallButton, secondaryButton]} />
+```
+
+### Toolbar with title
+
+```js
+// TODO(scottgrant): Use wonder-blocks-icon and wonder-blocks-icon-button
```
kevinbarabash: This is in a separate PR.

packages/wonder-blocks-toolbar/components/toolbar.md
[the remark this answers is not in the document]
kevinbarabash: Friendly ping re moving this to `docs.md`.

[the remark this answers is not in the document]
scotchfield: Hmm, maybe I misunderstood your intent here--I actually moved these back to a separate file after moving them all to `docs.md` so we could access them via `http://localhost:6060/#!/Toolbar`. For example, if we look at `http://localhost:6060/#!/Button`, there's nothing there. Do you have a preference?

packages/wonder-blocks-toolbar/components/toolbar.js
scotchfield: Done.

packages/wonder-blocks-toolbar/components/toolbar.js
scotchfield: Thanks, removed.

packages/wonder-blocks-toolbar/components/toolbar.js
scotchfield: Removed.

packages/wonder-blocks-toolbar/components/toolbar.md
scotchfield: Sorry, are you saying to remove the TODO? I totally agree that it should be in a separate PR, but I thought a TODO made sense. Maybe not though! What's your preference here?

kevinbarabash: I was agreeing with the TODO and saying that this doesn't need to be fixed in this PR.

packages/wonder-blocks-toolbar/components/toolbar.md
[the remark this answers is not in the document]
kevinbarabash: I just checked an `#!/Toolbar` is working now.  Let's leave it as is.

jeresig:
Great work Scott! Really glad to see this land!

PR #155, opened 2018-06-27
Add TooltipArrow component

somewhatabstract:
This PR builds on #152 to add the `TooltipArrow` component.

In this work, we extend the `TooltipBubble` component to render the arrow, including style changes to ensure the arrow appears where we need it relative to the content.

Now that we know the overall structure, the next PR after this one will split the `TooltipBubble` into a rendering component and a positioning component. This should allow us to have some rendering docs for the entire bubble outside of the positioning code. This will also be when we add styling for the tooltip bubble.

The SVG code for drawing the arrow may need some tweaks to get the drop shadow radius looking like our designs. This can be done once the overall tooltip styling is done in the next PR.

## Appearance
Docs have been added for this internal component. These internal component docs are marked private in anticipation of #130.

### DIFF
--- a/packages/wonder-blocks-tooltip/components/tooltip-arrow.js
+++ b/packages/wonder-blocks-tooltip/components/tooltip-arrow.js
@@ -1,8 +1,10 @@
 // @flow
 import * as React from "react";
+import * as ReactDOM from "react-dom";
 import {css, StyleSheet} from "aphrodite";
 
 import Colors from "@khanacademy/wonder-blocks-color";
+import {View} from "@khanacademy/wonder-blocks-core";
 import Spacing from "@khanacademy/wonder-blocks-spacing";
 
 import type {Placement} from "../util/types.js";
@@ -14,16 +16,19 @@ type Props = {
 };
 
 export default class TooltipArrow extends React.Component<Props> {
-    lastRef: ?HTMLElement;
+    _lastRef: ?HTMLElement;
 
-    updateRef(ref: ?HTMLElement) {
+    updateRef(ref: ?(React.Component<*> | Element)) {
         const {popperArrowProps} = this.props;
         // We only want to update the popper's arrow reference if it is
         // actually changed. Otherwise, we end up in an endless loop of updates
         // as every render would trigger yet another render.
-        if (popperArrowProps && ref && ref !== this.lastRef) {
-            this.lastRef = ref;
-            popperArrowProps.ref(ref);
+        if (popperArrowProps && ref) {
+            const domNode = ReactDOM.findDOMNode(ref);
+            if (domNode instanceof HTMLElement && domNode !== this._lastRef) {
+                this._lastRef = domNode;
+                popperArrowProps.ref(domNode);
+            }
         }
     }
 
@@ -165,17 +170,17 @@ export default class TooltipArrow extends React.Component<Props> {
         const {placement, popperArrowProps} = this.props;
         const {style} = popperArrowProps || {};
         return (
-            <div
-                className={css(
+            <View
+                style={[
                     styles.arrowContainer,
                     styles[`container-${placement}`],
-                )}
-                style={style}
+                    style,
+                ]}
                 data-placement={placement}
                 ref={(r) => this.updateRef(r)}
             >
                 {this._renderArrow()}
-            </div>
+            </View>
         );
     }
 }
--- a/packages/wonder-blocks-tooltip/components/tooltip-arrow.js
+++ b/packages/wonder-blocks-tooltip/components/tooltip-arrow.js
@@ -96,6 +96,10 @@ export default class TooltipArrow extends React.Component<Props> {
             height,
             width,
         } = this._calculateDimensionsFromPlacement();
+
+        // TODO(somewhatabstract): Use unique IDs!
+        const dropShadowFilterId = `tooltip-dropshadow-${placement}`;
+
         return (
             <svg
                 className={css(styles.arrow, styles[`arrow-${placement}`])}
@@ -115,7 +119,7 @@ export default class TooltipArrow extends React.Component<Props> {
                   * same side will yield ID conflicts... but at least they'll
                   * be the same filter, so it doesn't matter which one we
                   * reference, so long as it's for the correct side. */}
-                <filter id={`tooltip-dropshadow-${placement}`} height="150%">
+                <filter id={dropShadowFilterId} height="150%">
                     <feOffset
                         dx={Spacing.xSmall}
                         dy={Spacing.xSmall}
@@ -151,7 +155,7 @@ export default class TooltipArrow extends React.Component<Props> {
                     fill={Colors.white}
                     points={points.join(" ")}
                     stroke={Colors.offBlack16}
-                    filter={`url(#tooltip-dropshadow-${placement})`}
+                    filter={`url(#${dropShadowFilterId})`}
                 />
             </svg>
         );
--- a/packages/wonder-blocks-tooltip/components/tooltip-arrow.js
+++ b/packages/wonder-blocks-tooltip/components/tooltip-arrow.js
@@ -1,0 +1,228 @@
+// @flow
+import * as React from "react";
+import {css, StyleSheet} from "aphrodite";
+
+import Colors from "@khanacademy/wonder-blocks-color";
+import Spacing from "@khanacademy/wonder-blocks-spacing";
+
+import type {Placement} from "../util/types.js";
+import type {PopperArrowProps} from "react-popper";
+
+type Props = {
+    placement: Placement,
+    popperArrowProps?: PopperArrowProps,
+};
+
+export default class TooltipArrow extends React.Component<Props> {
+    lastRef: ?HTMLElement;
+
+    updateRef(ref: ?HTMLElement) {
+        const {popperArrowProps} = this.props;
+        // We only want to update the popper's arrow reference if it is
+        // actually changed. Otherwise, we end up in an endless loop of updates
+        // as every render would trigger yet another render.
+        if (popperArrowProps && ref && ref !== this.lastRef) {
+            this.lastRef = ref;
+            popperArrowProps.ref(ref);
+        }
+    }
+
+    _calculateDimensionsFromPlacement() {
+        const {placement} = this.props;
+        const arrowWidth = Spacing.large;
+        const arrowHeight = Spacing.small;
+
+        // Calculate the three points of the arrow. Depending on the arrow's
+        // direction (i.e., the tooltip's "side"), we choose different points,
+        // and set our SVG's bounds differently.
+        //
+        // Note that `arrowWidth` and `arrowHeight` refer to the
+        // downward-pointing arrow (i.e. side="top"). When the arrow points to
+        // the left or right instead, the width/height are inverted.
+        switch (placement) {
+            case "top":
+                return {
+                    points: [
+                        "0,0",
+                        `${arrowWidth / 2},${arrowHeight}`,
+                        `${arrowWidth},0`,
+                    ],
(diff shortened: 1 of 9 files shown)


jeresig:

Built with commit b882eb518762c37118b469350d63551b574bdeb1

https://deploy-preview-155--wonder-blocks.netlify.com

packages/wonder-blocks-tooltip/components/tooltip-arrow.js
```diff
@@ -0,0 +1,232 @@
...
+
+export default class TooltipArrow extends React.Component<Props> {
+    lastRef: ?HTMLElement;
+
+    updateRef(ref: ?HTMLElement) {
+        const {popperArrowProps} = this.props;
+        // We only want to update the popper's arrow reference if it is
+        // actually changed. Otherwise, we end up in an endless loop of updates
+        // as every render would trigger yet another render.
+        if (popperArrowProps && ref && ref !== this.lastRef) {
```
kevinbarabash: How often are we re-rendering `TooltipArrow`?

packages/wonder-blocks-tooltip/components/tooltip-arrow.js
```diff
@@ -0,0 +1,232 @@
...
+                  * Also, scope its ID by side, so that tooltips with other
+                  * "side" values don't end up using the wrong filter from
+                  * elsewhere in the document. (The `height` value depends on
+                  * which way the arrow is turned!)
+                  *
+                  * In general, it's not *great* that multiple tooltips of the
+                  * same side will yield ID conflicts... but at least they'll
+                  * be the same filter, so it doesn't matter which one we
+                  * reference, so long as it's for the correct side. */}
+                <filter id={dropShadowFilterId} height="150%">
```
kevinbarabash: Every time I look at this filter my eyes glaze over.  How can CSS box-shadow be so simple and this be so complicated!?

packages/wonder-blocks-tooltip/components/tooltip-arrow.js
```diff
@@ -0,0 +1,232 @@
...
+                />
+
+                {/* Draw an outline around the tooltip arrow, and apply the
+                  * blur filter we created above, to produce a drop shadow
+                  * effect. */}
+                <polyline
+                    // Redraw the stroke on top of the background color,
+                    // so that the ends aren't extra dark where they meet
+                    // the border of the tooltip.
+                    fill={Colors.white}
```
kevinbarabash: If this also has `fill={Colors.white}` do we need the first polyline?

packages/wonder-blocks-tooltip/components/tooltip-arrow.js
```diff
@@ -0,0 +1,232 @@
...
+                />
+            </svg>
+        );
+    }
+
+    render() {
+        const {placement, popperArrowProps} = this.props;
+        const {style} = popperArrowProps || {};
+        return (
+            <div
```
kevinbarabash: `View`?

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
```diff
@@ -39,12 +36,18 @@ export default class TooltipBubble extends React.Component<Props> {
         // gain over just using a div.
         return (
             <div
```
kevinbarabash: Is the extract lifting calling `findDOMNode` or is there other work that's involved?

packages/wonder-blocks-tooltip/components/tooltip.md
```diff
@@ -1,30 +1,30 @@
...
 
 In this example, we're no longer forcing the anchor root to be focusable, since the text input can take focus.
 
 ```js
 const {View} = require("@khanacademy/wonder-blocks-core");
 
-<Tooltip forceAnchorFocusivity={false} content={"The tooltip text"}>
+<Tooltip forceAnchorFocusivity={false} content={"I'm at the top!"}>
     <View>
```
kevinbarabash: Maybe give this `View` a background color so that it's easier to see.

packages/wonder-blocks-tooltip/components/tooltip.md
```diff
@@ -34,7 +34,7 @@ const {View, Text} = require("@khanacademy/wonder-blocks-core");
     <div style={{height: 100, overflow: "auto", border: "1px solid", margin: 10,}}>
         <div style={{height: "200vh"}}>
             <Text>This is a big long piece of text with a </Text>
```
kevinbarabash: This is a place where we should definitely be using a typography style since we want to encourage devs to use them all the time.  Something like:
```
<Body>This is a big long piece of text with a 
   <Tooltip content={"Some long text"}>
         <Text style={{color: "red"}}>tooltip</Text>
   </Tooltip>
in the middle</Body>
```
I wonder if we could detect when `Tooltip`'s `children` is a `string` and automatically wrap it in a `Text` component.

Although this example shows off the flexibility of the tooltip it seems sub-optimal from an i18n perspective.

packages/wonder-blocks-tooltip/components/tooltip.md
```diff
@@ -34,7 +34,7 @@ const {View, Text} = require("@khanacademy/wonder-blocks-core");
     <div style={{height: 100, overflow: "auto", border: "1px solid", margin: 10,}}>
         <div style={{height: "200vh"}}>
```
kevinbarabash: These should be `View`s.  I'm guessing the rebase didn't come through yet.

packages/wonder-blocks-tooltip/components/tooltip.md
```diff
@@ -53,11 +53,9 @@ const {StandardModal, ModalLauncher} = require("@khanacademy/wonder-blocks-modal
 const scrollyContent = (
      <div style={{height: 100, overflow: "auto", border: "1px solid", margin: 10,}}>
         <div style={{height: "200vh"}}>
-            <Text>This is a big long piece of text with a </Text>
-            <Tooltip content={"Some long text"}>
+            <Tooltip content={"I'm on the left!"} placement="left">
                 <Text style={{color: "red"}}>tooltip</Text>
```
kevinbarabash: typography component

packages/wonder-blocks-tooltip/package.json
```diff
@@ -15,12 +15,12 @@
...
   "dependencies": {
-    "react-popper": "^1.0.0-beta.6",
     "@khanacademy/wonder-blocks-color": "^1.0.2",
     "@khanacademy/wonder-blocks-core": "^1.0.2",
     "@khanacademy/wonder-blocks-layout": "^0.0.2",
     "@khanacademy/wonder-blocks-modal": "^0.0.3",
     "@khanacademy/wonder-blocks-spacing": "^2.0.0",
-    "@khanacademy/wonder-blocks-typography": "^1.0.2"
+    "@khanacademy/wonder-blocks-typography": "^1.0.2",
+    "react-popper": "^1.0.0"
```
kevinbarabash: We'll have to install this in `webapp`.  When `npm` installs it is there a single bundle we can load?  If so, maybe we can install in `third_party/node_modules` and have `javascript/node_modules/react-popper/index.js` point at that bundle.

styleguide.config.js
```diff
@@ -103,6 +103,15 @@ module.exports = {
                 "packages/wonder-blocks-tooltip/components/tooltip.js",
                 "packages/wonder-blocks-tooltip/components/tooltip-content.js",
             ],
+            sections: [
+                {
+                    name: "Internal Components",
+                    private: true,
```
kevinbarabash: I forget, does this do anything yet?

packages/wonder-blocks-tooltip/components/tooltip-arrow.js
somewhatabstract: This is not to do with re-rendering from what I can tell. All I know is that without this, we are eternally called again because of calling the ref with the same value.

I suspect it may be down to how we're doing our anchoring without wrapping.

packages/wonder-blocks-tooltip/components/tooltip-arrow.js
somewhatabstract: I know. I hate it. I can't even get it exactly how I want. I will..I will fight until I do, but still. Ugh.

packages/wonder-blocks-tooltip/components/tooltip-arrow.js
somewhatabstract: Yes. If we take it away, the drop shadow is much much darker and prominent.

packages/wonder-blocks-tooltip/components/tooltip-arrow.js
somewhatabstract: I think this all changes in #159. If it doesn't, I'll make the change there to minimize changes here.

packages/wonder-blocks-tooltip/components/tooltip.md
somewhatabstract: We already are doing the detect if it's a string and wrap it thing.

somewhatabstract: You commented on that in an earlier diff and I explained that's what it does.

I wouldn't stress on the tooltip docs just yet as they aren't really finished. The tooltip component is the last thing to be polished up.

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
somewhatabstract: Oh wait, it's a typo: "extra lifting". OK, yeah, so this actually goes away in #159. I moved to a `View` there and use `findDOMNode`

packages/wonder-blocks-tooltip/components/tooltip-arrow.js
somewhatabstract: So my suspicion is this.

1. We render and we call the ref to tell the popper what we are
2. The popper takes measurements and then calls to render us again with the changed positions

If we don't check that we are the same element, we re-render and RECALL the ref, which then signals the popper to redo its work.

packages/wonder-blocks-tooltip/components/tooltip-arrow.js
somewhatabstract: Also, because our `offBlack` has opacity, if we don't draw the initial white line, we blend with whatever is under the arrow, rather than white.

packages/wonder-blocks-tooltip/components/tooltip-arrow.js
somewhatabstract: I was wrong, we don't change this in #159. I've updated it for this PR.

packages/wonder-blocks-tooltip/components/tooltip.md
somewhatabstract: It will have one when #159 is done.

packages/wonder-blocks-tooltip/components/tooltip.md
somewhatabstract: I just used the `Text` element to add a red-highlight. I can change it to anything we want; just wanted it to be clear where one should over.

packages/wonder-blocks-tooltip/components/tooltip.md
somewhatabstract: Could be merge flog. I'll take a look.

packages/wonder-blocks-tooltip/components/tooltip-arrow.js
kevinbarabash: Interesting... I wonder what this looks like over a dark background.

somewhatabstract: It looks great! ...though the shadow is less visible for obvious reasons.

PR #158, opened 2018-06-28
Skip tab navigation on disabled clickable components

wishdasher:
This builds `tabIndex = -1` for disabled components straight into the `ClickableBehavior` component. If we don't want to add it here, we have to add it for the components that use `ClickableBehavior` and can be disabled, especially if the base component doesn't take the `disabled` attribute.

### DIFF
--- a/packages/wonder-blocks-tooltip/__snapshots__/generated-snapshot.test.js.snap
+++ b/packages/wonder-blocks-tooltip/__snapshots__/generated-snapshot.test.js.snap
@@ -124,3 +124,242 @@ exports[`wonder-blocks-tooltip example 4 1`] = `
   </button>
 </div>
 `;
+
+exports[`wonder-blocks-tooltip example 5 1`] = `
+<div
+  className=""
+  style={
+    Object {
+      "alignItems": "stretch",
+      "borderStyle": "solid",
+      "borderWidth": 0,
+      "boxSizing": "border-box",
+      "display": "flex",
+      "flexDirection": "column",
+      "margin": 0,
+      "minHeight": 0,
+      "minWidth": 0,
+      "padding": 0,
+      "position": "relative",
+      "zIndex": 0,
+    }
+  }
+>
+  <div
+    className=""
+    style={
+      Object {
+        "MsFlexBasis": 4,
+        "MsFlexPreferredSize": 4,
+        "WebkitFlexBasis": 4,
+        "alignItems": "stretch",
+        "borderStyle": "solid",
+        "borderWidth": 0,
+        "boxSizing": "border-box",
+        "display": "flex",
+        "flexBasis": 4,
+        "flexDirection": "column",
+        "flexShrink": 0,
+        "margin": 0,
+        "minHeight": 0,
+        "minWidth": 0,
+        "padding": 0,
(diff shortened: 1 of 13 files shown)


chenyejia:
I'm not sure that the tab index is getting applied properly. Should it be `tabindex` or `tabIndex`? Can you add an example with disabled `Button` with `href`? Disabled `button`s are already not focusable, so I can't tell with the current examples.

wishdasher:
That's really weird. Earlier, my tab index would set on the button and `tabindex` was working, but the behavior is completely different now. I'm going to look into this and get back to you.

wishdasher:
I figured it out -- so I thought I'd be clever and use `tabindex` because that's what the HTML documentation says. My last commit was just converting all `tabIndex` -> `tabindex`. (I removed that commit.) I didn't catch the error because the system didn't rebuild all the way, maybe because it doesn't register for case sensitivity?

Anyways, you're right about `Button` already working correctly. However, components that use tags like `a` as their base component cannot take the `disabled` attribute (https://www.w3schools.com/tags/att_disabled.asp), but we can skip navigation with `tabIndex` instead.

chenyejia:
Oh interesting!

Yeah could you add an example of `Button` with `href`? So that it uses the anchor tag with negative tab index. 

At a high level, I wonder if we should reconsider using `Button`s with anchor tags. VoiceOver (and I assume other screenreaders) will read disabled buttons as "dimmed", but if there's no equivalent for anchor tags, it feels weird to allow disabling.

wishdasher:
I see what you mean -- I added a `Button` that uses `href` in `docs.md`. If you apply just this change on `master`, you need to press tab twice to reach the next focusable element. With this change, it's skipped over.

chenyejia:
Hm yeah I'm still able to click the link via screenreader, which makes me think that it should be up to the individual component to handle disabled behavior. Maybe the solution is to only allow "disabled" on tags that have the attribute :/

wishdasher:
The link still being clickable is actually a bug (see https://github.com/Khan/wonder-blocks/pull/161). If we want screen readers to skip a disabled element entirely, we have to set `aria-hidden` to `true`. I think this PR is still worth considering, because it skips quasi-disabled elements on tab navigation.

packages/wonder-blocks-button/docs.md
```diff
@@ -50,11 +49,30 @@ const Button = require("./index.js").default;
...
+        marginRight: 10,
+    }
+});
+
+<View style={[styles.row]}>
+    <Button
+        style={[styles.sideMargins]}
+        onClick={(e) => console.log("Hello, world!")}
+        disabled={true}
+    >Label</Button>
```
kevinbarabash: nit (for consistency):
```
>
    Label
</Button>
```

packages/wonder-blocks-button/docs.md
```diff
@@ -50,11 +49,30 @@ const Button = require("./index.js").default;
...
+    <Button
+        style={[styles.sideMargins]}
+        href={"https://khanacademy.org"}
+        disabled={true}
+    >Href button</Button>
```
kevinbarabash: Same

PR #157, opened 2018-06-28
Split TooltipBubble into TooltipBubble and TooltipPopper

somewhatabstract:
This builds on #155 where we introduced the `TooltipArrow` component. The main purpose here is to separate the `TooltipBubble` component into a positioning component (`TooltipPopper`) and a rendering component (`TooltipBubble`).

This also adds some test examples for the bubble. The next task will be to style the bubble content.

PR #159, opened 2018-06-28
Style the bubble properly

somewhatabstract:
This PR builds on #157. It makes the following changes:

1. Adds better dropshadow support to the `TooltipArrow`
2. Adds styling to the `TooltipBubble` component
3. Fixes a bug where the `TooltipContent` component included spacing between the title and content even if there was only title or content.
4. Fixes tests broken by snapshot states changing and by additional element added to bubble.


jeresig:

Built with commit 395c306558648e4202491915b7f882e9630ebd75

https://deploy-preview-159--wonder-blocks.netlify.com

packages/wonder-blocks-tooltip/components/tooltip-arrow.js
```diff
@@ -94,73 +119,136 @@ export default class TooltipArrow extends React.Component<Props> {
...
+
+    /**
+     * Create an SVG filter that applies a blur to an element.
+     * We'll apply it to a dark shape outlining the tooltip, which
+     * will produce the overall effect of a drop-shadow.
+     *
+     * Also, scope its ID by side, so that tooltips with other
+     * "side" values don't end up using the wrong filter from
+     * elsewhere in the document. (The `height` value depends on
+     * which way the arrow is turned!)
```
kevinbarabash: It's possible that I don't know what `width` and `height` do in the filter.  I assume that they controlled the spread of the gaussian blur.  Is that not the case?

kevinbarabash: I'm so surprised that you have to set the `width` and `height` explicitly on a filter.  SVG already has other ways to clip things.  I wonder what purpose this is serving.  🤔  No action required.

packages/wonder-blocks-tooltip/components/tooltip-arrow.js
```diff
@@ -94,73 +119,136 @@ export default class TooltipArrow extends React.Component<Props> {
...
-                        <feMergeNode />
-                        <feMergeNode in="SourceGraphic" />
-                    </feMerge>
-                </filter>
-
-                {/* Draw the background of the tooltip arrow. */}
+                {this._maybeRenderDropshadow(points)}
+
+                {/**
+                 * Draw the actual background of the tooltip arrow.
```
kevinbarabash: This is weird output for prettier.

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
```diff
@@ -94,4 +99,13 @@ const styles = StyleSheet.create({
     "content-left": {
         flexDirection: "row",
     },
-});
+
+    content: {
+        padding: "10px 16px 10px 16px",
```
kevinbarabash: super nit: this could be `"10px 16px"`

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
```diff
@@ -94,4 +99,13 @@ const styles = StyleSheet.create({
...
+        maxWidth: 472,
+        borderRadius: Spacing.xxxSmall,
+        border: `solid 1px ${Colors.offBlack16}`,
+        backgroundColor: Colors.white,
+        boxShadow: `0 ${Spacing.xSmall}px ${Spacing.xSmall}px 0 ${Colors.offBlack8}`,
```
kevinbarabash: This is what I'm talking about.  I wish you could do this on SVG elements.

packages/wonder-blocks-tooltip/components/tooltip-bubble.test.js
```diff
@@ -72,12 +72,14 @@ describe("TooltipBubble", () => {
...
-                expect(
-                    contentElement && contentElement.getAttribute("id"),
-                ).toBe("content");
+                setTimeout(() => {
+                    const contentElement = document.getElementById("content");
+                    expect(contentElement).toBeDefined();
+                    done();
+                }, 0);
+            } else {
+                done();
```
kevinbarabash: Shouldn't we assert a failure in the `else` case?

packages/wonder-blocks-tooltip/components/tooltip-content.js
```diff
@@ -1,9 +1,11 @@
 // @flow
 import * as React from "react";
+
 import {View} from "@khanacademy/wonder-blocks-core";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
 import Spacing from "@khanacademy/wonder-blocks-spacing";
 import {HeadingSmall, LabelMedium} from "@khanacademy/wonder-blocks-typography";
+
 import type {Typography} from "@khanacademy/wonder-blocks-typography";
```
kevinbarabash: Thanks for following webapp style.  I haven't been raising this as an issue in reviews, maybe I should be.

packages/wonder-blocks-tooltip/util/is-obscured.js
```diff
@@ -60,6 +60,9 @@ export default function isObscured(element: Element) {
         anchorRight,
         anchorBottom,
     );
+    // TODO(somewhatabstract): Need to cater to the case where the viewport is
+    // zoomed such that both corners are off screen but the rest isn't.
```
kevinbarabash: Can you create Jira tasks for these TODOs?

somewhatabstract:
@kevinbarabash WRT the arrow, I haven't looked into that yet. I don't want to hold up this main work until I have. It's on my list for an upcoming change as I also need to do some work on the arrow dropshadow.

packages/wonder-blocks-tooltip/components/tooltip-arrow.js
kevinbarabash: I guess when it comes to some filters like gaussians the result being rendered could be quite big so forcing people to pick a clipping rect probably makes sense from a perf point of view.  The dev ergonomics are not the greatest though.

PR #160, opened 2018-06-28
Fix select box docs

somewhatabstract:
This fixes the select-box.md so that the snapshots generate correct code.


jeresig:

Built with commit 1c8ee65cf4aebf15dd518fcda2993e5b40036031

https://deploy-preview-160--wonder-blocks.netlify.com

wishdasher:
Could you also update `package.json` with the `layout` package?

somewhatabstract:
@wishdasher Does it need it? Since this is only test docs, I don't think we need to put a package dep on there.

wishdasher:
Oops, good point. I got confused.

PR #161, opened 2018-06-28
Prevent disabled button-appearing anchors from working

wishdasher:
`Button` and `ActionItem` components for menu still redirect to the `href` even when they're disabled! 😱

We fix this by checking for the `disabled` prop before adding `href` onto the component. The `disabled` attribute doesn't work because the base component is `a`, not `button`, in the cases where `href` is defined.


wishdasher:
Will base off `master` when [PR 158](https://github.com/Khan/wonder-blocks/pull/158) is resolved.

jeresig:

Built with commit dffe11eb8efb00bee55645e641834acfcfcc1eca

https://deploy-preview-161--wonder-blocks.netlify.com

kevinbarabash:
For some reason click on the disabled items changes the color of the text to be darker than it was before clicking on the item.

wishdasher:
This is very spooky 👻. I've isolated the cause -- the text turns lighter when highlighted and then losing focus. I inspected the component in React devtools. The props of the `ButtonCore` never change, so our styles are not being reapplied as far as I can tell. The already-existing Button in the docs (the one on the left) exhibits the same behavior. It seems like Safari, rather than React, is causing this behavior.

Here's a screenshot of part of the text in the different color.

screen shot 2018-06-28 at 17 14 18

PR #162, opened 2018-06-29
try upgrading to React 16 to suss out the issues

kevinbarabash:
There are three failing errors that I wasn't able to resolve...yet.  Two of them stem from our use of `findDOMNode` and the third is `Enzyme Internal Error: unknown node with tag 12`.  We'll to disable that test until https://github.com/airbnb/enzyme/issues/1620 is fixed.

### DIFF
--- a/packages/wonder-blocks-core/package.json
+++ b/packages/wonder-blocks-core/package.json
@@ -16,7 +16,7 @@
   },
   "peerDependencies": {
     "aphrodite": "^1.2.5",
-    "react": "^15.3.2",
+    "react": "^16.4.1",
     "react-router-dom": "^4.2.2"
   },
   "author": "",
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "babel-preset-react": "^6.24.1",
     "babylon": "^6.18.0",
     "enzyme": "^3.3.0",
-    "enzyme-adapter-react-15": "^1.0.5",
+    "enzyme-adapter-react-16": "^1.1.1",
     "eslint": "^4.19.1",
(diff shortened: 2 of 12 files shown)


packages/wonder-blocks-core/package.json
```diff
@@ -16,7 +16,7 @@
   },
   "peerDependencies": {
     "aphrodite": "^1.2.5",
-    "react": "^15.3.2",
+    "react": "^16.4.1",
```
kevinbarabash: It took me a while for me figure out that these needed to be updated as well.

jeresig:

Built with commit 03638b1f09febead0f15cd6ec8ee229ffd38b10e

https://deploy-preview-162--wonder-blocks.netlify.com

kevinbarabash:
@somewhatabstract any thoughts about the `findDOMNode` related failures.

somewhatabstract:
@kevinbarabash The issue, at least for the modal test failures I saw, is that the nodes haven't been mounted at the point `findDOMNode` is called and so there are no DOM nodes to speak of. Those tests probably need re-writing to be similar to the ones I wrote for the tooltip components so that they mount first and get real mounted nodes to use with `findDOMNode`.

somewhatabstract:
I can take a closer look a little later today to probably delve into them, if you'd like.

kevinbarabash:
> Those tests probably need re-writing to be similar to the ones I wrote for the tooltip components so that they mount first and get real mounted nodes to use with `findDOMNode`.

I might give that a shot this weekend.

wishdasher:
I'm going to rebase and push

PR #163, opened 2018-06-29
Add NoSSR component

somewhatabstract:
This adds the `NoSSR` component that @jnetterf created.

I have also added additional tests for the SSR-specific behaviour.

## Why?
Having the ability to avoid rendering things on the server is valuable; especially when we need to have different behaviours between client and server, like when producing identifiers; initial renders much match on server and client, and so unique IDs have to be generated on the client AFTER intial render. This component can help with that. The next work will be to add support for unique identifiers and that will require this SSR/Client behaviour differentiation.

### DIFF
--- a/packages/wonder-blocks-core/components/no-ssr.test.js
+++ b/packages/wonder-blocks-core/components/no-ssr.test.js
@@ -1,6 +1,5 @@
 // @flow
 import * as React from "react";
-// import * as ReactDOM from "react-dom";
 import * as ReactDOMServer from "react-dom/server";
 
 import {mount} from "enzyme";
@@ -28,6 +27,9 @@ describe("NoSSR", () => {
 
         const assert = () => {
             // Assert
+            // This was called from within the NoSSR children render prop.
+            // Therefore, if the placeholder has been called, it must have
+            // been called first.
             expect(mockPlaceholder).toHaveBeenCalledTimes(1);
             done();
         };
--- a/packages/wonder-blocks-core/components/no-ssr.test.js
+++ b/packages/wonder-blocks-core/components/no-ssr.test.js
@@ -1,0 +1,70 @@
+// @flow
+import * as React from "react";
+// import * as ReactDOM from "react-dom";
+import * as ReactDOMServer from "react-dom/server";
+
+import {mount} from "enzyme";
+
+import NoSSR from "./no-ssr.js";
+
+describe("NoSSR", () => {
+    test("renders a placeholder first, then the actual content", (done) => {
+        // Arrange
+        const mockPlaceholder = jest.fn(() => null);
+
+        const arrangeAct = (assert) => {
+            const nodes = (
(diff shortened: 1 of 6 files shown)


jeresig:

Built with commit d025acd769c3cb5352d55c62786fac1eb1628aa9

https://deploy-preview-163--wonder-blocks.netlify.com

jeresig:

Built with commit db7a206ca8993c4a7382bea41439327d7a7b4e48

https://deploy-preview-163--wonder-blocks.netlify.com

packages/wonder-blocks-core/components/no-ssr.test.js
```diff
@@ -0,0 +1,70 @@
+// @flow
+import * as React from "react";
+// import * as ReactDOM from "react-dom";
```
kevinbarabash: remove

packages/wonder-blocks-core/components/no-ssr.test.js
```diff
@@ -0,0 +1,70 @@
...
+import * as ReactDOMServer from "react-dom/server";
+
+import {mount} from "enzyme";
+
+import NoSSR from "./no-ssr.js";
+
+describe("NoSSR", () => {
+    test("renders a placeholder first, then the actual content", (done) => {
```
kevinbarabash: How does this test assert the order?

packages/wonder-blocks-core/components/no-ssr.test.js
```diff
@@ -0,0 +1,70 @@
...
+            const nodes = (
+                <NoSSR placeholder={mockPlaceholder}>{mockChildren}</NoSSR>
+            );
+
+            // Act
+            ReactDOMServer.renderToStaticMarkup(nodes);
+
+            // Assert
+            expect(mockPlaceholder).toHaveBeenCalledTimes(1);
+            expect(mockChildren).toHaveBeenCalledTimes(0);
```
kevinbarabash: Cool!

packages/wonder-blocks-core/components/no-ssr.test.js
somewhatabstract: Because the  assert is called when the inside renders, and the assert checks that the placeholder was already called. Therefore, the placeholder HAD to have been called first.

packages/wonder-blocks-core/components/no-ssr.test.js
somewhatabstract: Good catch, thanks!

packages/wonder-blocks-core/components/no-ssr.test.js
kevinbarabash: Ah... I got it.  I was confused b/c I mistook `assert` for a general `assert` function that might of come from node's `assert` package.  Time to go for a walk and get some fresh air.

PR #164, opened 2018-06-29
Avoid setTimeout in portal

somewhatabstract:
This is a simple PR to remove our heavy-handed mount/unmount approach that used timeouts.

In this version, we separate the mount/unmount from the rendering of the actual portal content. Then we do the render in the `componentDidUpdate` as recommended by the react error messages. This works great!

Although `componentDidUpdate` is not called for the initial render, we don't need it to because on initial render the anchor cannot be focused or hovered, so the portal wouldn't be needed anyway. Since we show/hide the bubble by adding and removing the `children` prop of the portal, we force a render by changing that prop, and so it all works!


packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
```diff
@@ -31,23 +31,18 @@ export default class TooltipPortalMounter extends React.Component<Props> {
     _timeoutId: ?number;
     _rendered: boolean;
 
+    componentDidUpdate() {
+        this._renderChildren();
```
somewhatabstract: We changed! Update the kids

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
```diff
@@ -83,17 +78,13 @@ export default class TooltipPortalMounter extends React.Component<Props> {
         destination.setAttribute(TooltipPortalAttributeName, "");
         root.appendChild(destination);
 
+        // Save the destination node, so we can re-use it.
+        this._destination = destination;
```
somewhatabstract: Make sure we save this before we call to render kids for the first time

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
```diff
@@ -83,17 +78,13 @@ export default class TooltipPortalMounter extends React.Component<Props> {
...
         // See https://github.com/tajo/react-portal/blob/master/src/LegacyPortal.js
-        ReactDOM.unstable_renderSubtreeIntoContainer(
-            this,
-            children,
-            destination,
-        );
-
-        // Save the destination node, so we can remove it on unmount.
-        this._destination = destination;
+        this._renderChildren();
```
somewhatabstract: Render the bubble

jeresig:

Built with commit 22ab7bf8847dbb01322aacbb7fff41b81f5c8e23

https://deploy-preview-164--wonder-blocks.netlify.com

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
```diff
@@ -111,24 +102,31 @@ export default class TooltipPortalMounter extends React.Component<Props> {
...
     }
 
-    _timeoutDoMount() {
-        this._clearMountTimeout();
-        this._timeoutId = setTimeout(() => this._doMount(), 0);
-    }
+    _renderChildren() {
+        if (!this._destination) {
+            this._doMount();
+            return;
```
somewhatabstract: If we have nowhere to mount, abandon this render and mount our portal. It will call us to render again once it has.

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
```diff
@@ -111,24 +102,31 @@ export default class TooltipPortalMounter extends React.Component<Props> {
...
+        }
+
+        const {children} = this.props;
+        if (!children) {
+            this._doUnmount();
```
somewhatabstract: If we have nothing to render, unmount the portal

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
```diff
@@ -111,24 +102,31 @@ export default class TooltipPortalMounter extends React.Component<Props> {
...
+            return;
+        } else {
+            ReactDOM.unmountComponentAtNode(this._destination);
```
somewhatabstract: If we get here, the content of the bubble changed, so let's remove what we had and render something new.

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
```diff
@@ -111,24 +102,31 @@ export default class TooltipPortalMounter extends React.Component<Props> {
...
+        }
 
-    _timeoutDoUnmount() {
-        this._clearMountTimeout();
-        this._timeoutId = setTimeout(() => this._doUnmount(), 0);
+        ReactDOM.unstable_renderSubtreeIntoContainer(
```
somewhatabstract: Render the actual things

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
somewhatabstract: Not certain this is necessary, or if the later render call kinda handles this, but it seems appropriate and is certainly clearer

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
```diff
@@ -111,24 +102,31 @@ export default class TooltipPortalMounter extends React.Component<Props> {
...
+            this,
+            children,
+            this._destination,
+        );
     }
 
     render() {
-        const {children, anchor} = this.props;
+        const {anchor} = this.props;
+
```
somewhatabstract: Now our render method is nice and clean and less complicated.

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
kevinbarabash: Why not call `_doMount` from `componentDidMount`?

kevinbarabash: I see... we may need to call `_doMount` more than once b/c `children` can change multiple times over the lifetime of this portal.  Could you add some documentation at the time that describes the life time of this component and when things are mounted/unmounted?

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
kevinbarabash: These would be good comments to have in the code.

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
```diff
@@ -111,24 +99,34 @@ export default class TooltipPortalMounter extends React.Component<Props> {
...
+            ReactDOM.unmountComponentAtNode(this._destination);
+        }
 
-    _timeoutDoUnmount() {
-        this._clearMountTimeout();
-        this._timeoutId = setTimeout(() => this._doUnmount(), 0);
+        // We have to render the subtree like this so that everything works as
+        // expected.
+        // See https://github.com/tajo/react-portal/blob/master/src/LegacyPortal.js
+        ReactDOM.unstable_renderSubtreeIntoContainer(
```
kevinbarabash: Will this still be necessary if/when we switch to `ReactDOM.createPortal(child, container)`?

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
somewhatabstract: It's not just that. We won't have an anchor element in the mount. That only appears after rendering.

somewhatabstract: I realise that comment was confusing since render happens before mount.

Because the portal itself and its content are shown during a hover or focus, this generally happens after componentDidMount. However, there is a chance we might have content immediately, so I've reinstated a call in the componentDidMount and commented accordingly.

I think the renaming of methods and commentary will make things clearer.

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
somewhatabstract: No, I don't think so. Assuming that the react portal works how we need it to, we shouldn't need to do this magic.

PR #166, opened 2018-06-29
Remove flow from snapshots

somewhatabstract:
Until we can use flowtypes inside of our markdown or at least have `$FlowFixMe` comments make it to the snapshots, we don't want flow validation on our generated snapshot tests.


jeresig:

Built with commit 33119e888b0bd9b0ea07ebe3199d5a3c01025c81

https://deploy-preview-166--wonder-blocks.netlify.com

PR #167, opened 2018-06-29
Add test coverage for our package exports

somewhatabstract:
This adds an `index.test.js` for each package that verifies the exported items match what we expect. It's a shallow test to catch unexpected API changes at the package export level. The tests don't delve more deeply into each export, that's for other tests to do.

This also standardizes custom snapshot tests into a specific file and moves component tests to their component test file.

NOTE: adds dev dependency on `babel-plugin-transform-dynamic-import`


jeresig:

Built with commit 047b5bfd4acac54a0410393dd884c8753d9c8a11

https://deploy-preview-167--wonder-blocks.netlify.com

somewhatabstract:
@kevinbarabash You didn't approve the review. Are you waiting for @jeresig to take a look or was that an oversight?

PR #168, opened 2018-06-30
Added coverage for mix and fade

somewhatabstract:
This adds test coverage for the `fade` and `mix` utility methods in the `wonder-blocks-colors` package. This also moves the `utils.js` file into a `util` folder to match the file layout of other packages.

As part of this, a couple of functional changes were made:

1. The `fade` call will now combine the alpha of the `color` with that of the `background`. None of the existing wonder blocks callsites are using it this way so this shouldn't have immediate impact. However, I wrote the test expecting this behaviour and was surprised that it didn't have it, so assuming it was an oversight, I fixed it.
2. The `mix` method had an issue where it would sometimes output 5 digit colors instead of 3 or 6. I think the 3 digit versions it had outputted were an accident based on the method's commentary. So, I modified it to ensure that we always outputted 6 digit colors, if outputting in that format.


jeresig:

Built with commit 4d40c4f512f8441db05d7bb545c0a3849d536f41

https://deploy-preview-168--wonder-blocks.netlify.com

packages/wonder-blocks-color/util/utils.test.js
```diff
@@ -0,0 +1,190 @@
...
+describe("mix", () => {
+    test("color is not a string, throws", () => {
+        // Arrange
+        const color = ((null: any): string);
+
+        // Act
+        const underTest = () => mix(color, "#FFF");
+
+        // Assert
+        expect(underTest).toThrowErrorMatchingSnapshot();
```
kevinbarabash: I don't think there's much benefit in using snapshots for testing this.
```
expect(underTest).toThrow(new Error("Failed to parse color: #FFF"))
```
isn't much longer and you don't have to go look at another file to see what error we were expecting.

packages/wonder-blocks-color/util/utils.js
```diff
@@ -0,0 +1,99 @@
+// @flow
```
kevinbarabash: With Github it's really hard to see any changes that have been made to moved.  Can you create separate commits for moving files and for making changes to them?

packages/wonder-blocks-color/util/utils.test.js
somewhatabstract: That's the point of using a snapshot. It avoids having to go copying strings around. The error snapshots are there for exactly this kind of thing. I think we should use the tools we have.

Also, it's a test. We don't need to go see what it was expecting unless it fails.

kevinbarabash: Okay.

somewhatabstract:
@kevinbarabash I've split this into two commits. If you go to https://github.com/Khan/wonder-blocks/pull/168/commits/9d00cc749371ee67835ea7c5b0e051ba7e8f8994, you can see the changes to `utils.js`.

packages/wonder-blocks-color/util/utils.test.js
```diff
@@ -0,0 +1,190 @@
...
+            expect(result).toBe(expectation);
+        }
+    });
+
+    test("color has alpha, returns mix with background alpha", () => {
+        // Arrange
+        const testpoints = [
+            ["rgba(100,100,100,0.2)", "#FFF", "#e0e0e0"],
+            ["rgba(100,100,100,0.2)", "rgba(0,0,0,0.5)", "rgba(20,20,20,0.50)"],
+            ["rgba(100,100,100,0.2)", "rgba(0,0,0,0.5)", "rgba(20,20,20,0.50)"],
```
kevinbarabash: I would expect the alpha to be higher, but since we don't use this anywhere I okay to leave this as a `TODO`.

packages/wonder-blocks-color/util/utils.test.js
```diff
@@ -0,0 +1,190 @@
...
+
+    test("rgba color with less than 100% alpha, returns faded color", () => {
+        // Arrange
+        const color = "rgba(100,200,0, 0.6)";
+
+        // Act
+        const result = fade(color, 0.5);
+
+        // Assert
+        expect(result).toBe("rgba(100,200,0,0.30)");
```
kevinbarabash: Fancy.

packages/wonder-blocks-color/util/utils.test.js
somewhatabstract: This was the existing behavior. I didn't change that as `mix` seemed to be more explicit about how it handled alpha, whereas fade implied it did something that it didn't, which I then fixed.

somewhatabstract: That said, I notice that I duplicated a test point here. Should probably fix that.

PR #169, opened 2018-06-30
Incorporate multi selection menus!

wishdasher:


jeresig:

Built with commit 50281475752d42281c11cd56343fc193d3f42963

https://deploy-preview-169--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/multi-select-menu.js
```diff
@@ -0,0 +1,227 @@
...
+                    }
+                    selected={selectedItems.includes(item.value)}
+                    value={item.value}
+                    variant={"checkbox"}
+                />
+            );
+        });
+
+        if (displaySelectionShortcuts) {
+            // TODO(sophie): translate for i18n
```
kevinbarabash: I think as long as we have a version of `i18n._()` that just returns the string passed in we should be good.  @jeresig that should work right?  We should probably add a function to `wonder-blocks-core`.  That can be a separate PR.

packages/wonder-blocks-dropdown/components/multi-select-menu.js
```diff
@@ -0,0 +1,227 @@
...
+            light,
+            placeholder,
+            selectItemType,
+            selectedItems,
+            style,
+        } = this.props;
+
+        const {open} = this.state;
+
+        // TODO(sophie): figure out how to configure plurals for i18n
```
kevinbarabash: We'll need a copy of the `i18n.ngettext()` function from `i18n.js` in `webapp`.  Probably best to add it to `wonder-blocks-core`... or maybe we can create an `i18n` package.  This can be a separate PR.

packages/wonder-blocks-dropdown/components/multi-select-menu.md
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    handleChanges(changes) {
+        console.log("changes happened!");
+        const {selectedItems} = this.state;
+        changes.forEach((change) => {
+            if (change.selected) {
+                selectedItems.push(change.value);
+            } else {
+                selectedItems.splice(selectedItems.indexOf(change.value), 1);
```
kevinbarabash: If `change` was also an array then this would become:
```
this.setState({
   selectedItems: changes,
});
```

packages/wonder-blocks-dropdown/components/multi-select-menu.js
```diff
@@ -0,0 +1,227 @@
...
+                    disabled={selectedItems.length === 0}
+                    indent={true}
+                    onClick={() => this.handleSelectNone()}
+                />
+            );
+
+            const separator = <SeparatorItem key={"selectOptSep"} />;
+
+            // Add options at beginning, add a separator
+            menuItems.unshift([selectAll, selectNone, separator]);
```
kevinbarabash: Maybe move the `displaySelectionShortcuts` block above the call to `const menuItems = items.map((item, index) => {`.  This avoids `unshift()` which can be bad for perf... in this situation not a huge deal.  Maybe something like:
```
const menuItems = [...this.maybeGetShortcuts(), ...this.getMenuitems()];
```

packages/wonder-blocks-dropdown/components/multi-select-menu.js
```diff
@@ -0,0 +1,227 @@
...
+            />
+        );
+    }
+}
+
+const styles = StyleSheet.create({
+    // This is to add extra space on top of the menu options to separate the
+    // options from the opener component.
+    menuTopSpace: {
+        top: 48,
```
kevinbarabash: This is b/c the opener is `40` and the space between the opener and the menu is `8`.

packages/wonder-blocks-dropdown/components/multi-select-menu.md
```diff
@@ -0,0 +1,241 @@
+The multi select menu allows the selection of multiple items. Clients are
+responsible for keeping track of the selected items in the menu.
+
+TODO(sophie): keyboard navigation and dismissal
+
+The first menu starts with nothing selected and has no selection shortcuts.
+There is also a disabled item, and this menu has a set width.
+
+The second menu starts with one item selected and has selection shortcuts for
+select all and select none. The second menu does not have a predefined placeholder.
```
kevinbarabash: I noticed that the items in these menus have white borders on the left/right whereas the ones in the SingleSelectMenu do not.  These shouldn't have any white borders on the left/right either.

packages/wonder-blocks-dropdown/components/multi-select-menu.md
```diff
@@ -0,0 +1,241 @@
+The multi select menu allows the selection of multiple items. Clients are
+responsible for keeping track of the selected items in the menu.
+
+TODO(sophie): keyboard navigation and dismissal
+
+The first menu starts with nothing selected and has no selection shortcuts.
+There is also a disabled item, and this menu has a set width.
```
kevinbarabash: One of the examples should have a lot of choices so that we can show that the menu scrolls.

packages/wonder-blocks-dropdown/components/multi-select-menu.md
```diff
@@ -0,0 +1,241 @@
...
+        />;
+    }
+}
+
+<View style={[styles.row]}>
+    <ExampleNoneSelected />
+    <View style={[styles.strutLike]} />
+    <ExampleWithShortcuts />
+    <View style={[styles.strutLike]} />
+    <ExampleNoneSelected disabled={true} />
```
kevinbarabash: Can you move these into separate examples?  It will keep the code snippet shorter and easier for devs to see what's going on.

packages/wonder-blocks-dropdown/components/single-select-menu.md
```diff
@@ -197,15 +197,15 @@ class LightRightAlignedExample extends React.Component {
             items={[
                 {
                     label: "Regular milk tea with boba",
-                    value: "regular"
+                    value: "regular",
                 },
                 {
                     label: "Wintermelon milk tea with boba",
-                    value: "wintermelon"
+                    value: "wintermelon",
```
kevinbarabash: This sounds really yummy.

packages/wonder-blocks-dropdown/utils/types.js
```diff
@@ -0,0 +1,32 @@
...
+//     href?: string,
+//     /** Whether to use client-side navigation. */
+//     clientNav?: boolean,
+//     /** Callback on the action. */
+//     onClick?: () => void,
+// };
+
+export type SelectItemProps = {
+    type: "select",
+    /** Whether this item is disabled. Default false. */
```
kevinbarabash: super nit: `/* ... */`

packages/wonder-blocks-dropdown/utils/types.js
```diff
@@ -0,0 +1,32 @@
...
+    label: string,
+    /** Value of this item. Treat as a key. */
+    value: string,
+    /** Optional extra callback. Passes whether this item is selected. */
+    onClick?: (selected: boolean) => void,
+};
+
+// export type SeparatorProps = {
+//     type: "separator",
+// };
```
kevinbarabash: Should this be used somewhere since we do have separators in some of the examples?

packages/wonder-blocks-dropdown/components/multi-select-menu.js
```diff
@@ -0,0 +1,227 @@
...
+            style,
+        } = this.props;
+
+        const {open} = this.state;
+
+        // TODO(sophie): figure out how to configure plurals for i18n
+        const menuText =
+            selectedItems.length === 0 && placeholder
+                ? placeholder
+                : `${selectedItems.length} ${selectItemType}`;
```
kevinbarabash: This is tricky since `ngettext` calls should look like:
```
i18n.ngettext("%(num)s Cat", "%(num)s Cats", 3)
```
We should probably be passing in a callback as a prop, something like:
```
<MultiSelectMenu
    ...
    getMenuText={(count) => i18n.ngettext("%(num)s Cat", "%(num)s Cats", count)}
    ...
/>
```

packages/wonder-blocks-dropdown/components/multi-select-menu.js
```diff
@@ -0,0 +1,227 @@
...
+     * Type of the menu items.
+     * For example, if selectItemType is "student" and there are two students
+     * selected, the SelectBox would display "2 students"
+     */
+    selectItemType: string,
+
+    /**
+     * Whether to display shortcuts for Select All and Select None.
+     */
+    displaySelectionShortcuts?: boolean,
```
kevinbarabash: This could be shorter.  Maybe just `shortcuts?: boolean`.

packages/wonder-blocks-dropdown/components/multi-select-menu.md
kevinbarabash: This will also simplify some of the other methods in `MultiSelectMenu`.

wishdasher: I'm not sure what you mean here. `changes` is an array of items that have changed states. Do you mean that `changes` should return an array of all the selected items?

packages/wonder-blocks-dropdown/utils/types.js
wishdasher: Some other `type` files in the codebase use `/** ... */`. I will leave it like that to be consistent.

packages/wonder-blocks-dropdown/utils/types.js
wishdasher: This will be used in the upcoming `ActionMenu` pull request. I just didn't want to have to save these somewhere else when I committed the `types.js` file.

packages/wonder-blocks-dropdown/components/multi-select-menu.md
kevinbarabash: Yes, if it's just an array of selected items there's a lot of code that can be simplified.  Besides, this is what's being passed as a prop to `MultiSelectMenu`.

packages/wonder-blocks-dropdown/utils/types.js
kevinbarabash: I didn't realize that.

wishdasher:
How the `changes` parameter should work/function is the one unaddressed comment. There is also a bug with `ClickableBehavior` that causes multiple items to be in the hover state. (Look at the menu with selection shortcuts for an example: press `Select all` and then deselect an item.) I will look into this separately.

kevinbarabash:
> press Select all and then deselect an item.) I will look into this separately.

I think it's b/c the "Select all (button)" stops receiving events when it's disable so it doesn't update its state.  That shouldn't be too hard to fix.

packages/wonder-blocks-dropdown/components/multi-select-menu.js
kevinbarabash: That's probably a good idea.  It might be best to have `webapp` pass in strings instead of adding `i18n` functions to `wonder-blocks`.

packages/wonder-blocks-dropdown/components/multi-select-menu.md
kevinbarabash: The regular menu items look good now, but the shortcuts still have a slight white border on the left/right.

packages/wonder-blocks-dropdown/components/multi-select-menu.md
kevinbarabash: Friendly ping. 🙂

kevinbarabash:
Here's what I'm seeing in Safari with the shortcuts.
screen shot 2018-07-03 at 4 59 20 pm

PR #170, opened 2018-06-30
Improve TooltipTail rendering and other tweaks

somewhatabstract:
- Add margins so that the arrow doesn't invade the border radius
- Get the drop shadow right so we don't have weird clipping artifacts (and comment it so others can hopefully understand what I learned)
- Remove some unused bits
- Offset shadow on left/right so that it doesn't crash the bubble outline
- Rename `TooltipArrow` to `TooltipTail` to match design vernacular
- Update docs

PR #171, opened 2018-06-30
Abstract away react-popper types and idiosyncrasies

somewhatabstract:
This update builds on the work done in #170. In this change, we abstract away most of the `react-popper` idiosyncrasies so that only `TooltipPopper` really knows about them. Specifically we:

1. Create Props types for `TooltipBubble` and `TooltipTail` that are mostly independent of `react-popper` and allow for some optional things to simplify our test cases.
2. Move the `ref` tracking that was duplicated in `TooltipBubble` and `TooltipTail` so that it is shared code inside of `TooltipPopper`


jeresig:

Built with commit 10d60e4f8fc55f09547b06c3f96767fe34021795

https://deploy-preview-171--wonder-blocks.netlify.com

packages/wonder-blocks-tooltip/components/tooltip-popper.js
```diff
@@ -37,14 +29,43 @@ type Props = {|
...
+            style: popperProps.style,
+            updateBubbleRef: this._bubbleRefTracker.updateRef,
+            tailOffset: {
+                top: popperProps.arrowProps.style.top,
+                left: popperProps.arrowProps.style.left,
+            },
+            updateTailRef: this._tailRefTracker.updateRef,
+            outOfBoundaries: popperProps.outOfBoundaries,
+        };
+        return children(bubbleProps);
```
kevinbarabash: I see... there is where the conversion from popper props to bubble props happens.  It might be worth calling this out with a comment so that it's easy to tell where the boundary between our code and react-popper code is.

packages/wonder-blocks-tooltip/components/tooltip-popper.js
```diff
@@ -37,14 +29,43 @@ type Props = {|
 |};
 
 export default class TooltipPopper extends React.Component<Props> {
-    _renderPositionedContent(childrenProps: PopperJSChildrenProps) {
-        const {children, placement} = this.props;
+    _bubbleRefTracker = new RefTracker();
+    _tailRefTracker = new RefTracker();
+
+    _getUpdatedBubbleRefTracker(targetRefFn: (?HTMLElement) => void) {
```
kevinbarabash: Where is this called?

packages/wonder-blocks-tooltip/components/tooltip-popper.js
```diff
@@ -37,14 +29,43 @@ type Props = {|
...
+        this._bubbleRefTracker.setCallback(targetRefFn);
+        return this._bubbleRefTracker.updateRef;
+    }
+
+    _getUpdatedTailRefTracker(targetRefFn: (?HTMLElement) => void) {
```
kevinbarabash: Where it this called?

packages/wonder-blocks-tooltip/util/ref-tracker.js
```diff
@@ -0,0 +1,43 @@
...
+    updateRef: getRefFn;
+    _lastRef: ?HTMLElement;
+    _targetFn: ?(?HTMLElement) => void;
+
+    updateRef(ref: ?(React.Component<*> | Element)) {
+        if (ref) {
+            // We only want to update the reference if it is
+            // actually changed. Otherwise, we can trigger another render that
+            // would then update the reference again and just keep looping.
+            const domNode = ReactDOM.findDOMNode(ref);
```
kevinbarabash: What if `ref` is an `Element`... does `findDOMNode` handle that by returning the value it was passed?

packages/wonder-blocks-tooltip/util/ref-tracker.js
```diff
@@ -0,0 +1,43 @@
...
+ * update methods so that we can convert a regular React ref into a DOM node
+ * as react-popper expects, and also ensure we only update react-popper
+ * on actual changes, and not just renders of the same thing.
+ */
+import * as React from "react";
+import * as ReactDOM from "react-dom";
+
+import type {getRefFn} from "../util/types.js";
+
+export default class RefTracker {
```
kevinbarabash: It feels like `forwardRef` from React 16 does something similar.  Do you think we'll be able to use `forwardRef` to replace this in the future?

packages/wonder-blocks-tooltip/util/ref-tracker.test.js
```diff
@@ -0,0 +1,171 @@
...
+                    tracker.updateRef(ref);
+
+                    // Assert
+                    expect(targetFn).toHaveBeenCalledWith(domNode);
+                };
+
+                arrange(actAndAssert);
+            });
+
+            test("same ref, targetFn callback, does not call targetFn", () => {
```
kevinbarabash: nice!

packages/wonder-blocks-tooltip/components/tooltip-popper.js
somewhatabstract: Yes, fair. I will do that.

packages/wonder-blocks-tooltip/components/tooltip-popper.js
somewhatabstract: I don't think it is anymore. I'll delete it.

packages/wonder-blocks-tooltip/components/tooltip-popper.js
somewhatabstract: Left over from interim solution. I'll delete. Thanks!

packages/wonder-blocks-tooltip/util/ref-tracker.js
somewhatabstract: Yes, it handles it just fine.

packages/wonder-blocks-tooltip/util/ref-tracker.js
somewhatabstract: I don't know. It would be nice if `react-popper` had this check built in so we didn't have to worry about it.

somewhatabstract:
@kevinbarabash I think this is all good now.

PR #172, opened 2018-06-30
Get inline styles working

somewhatabstract:
This is a workaround for aphrodite adding `!important`. When that gets dropped, we can remove this.

In this change, if we're not snapshotting styles, we convert all the gathered inline styles into a CSS class using aphrodite and add it to the end of the stylesheet list. That way, the inline styles apply as intended.

I updated the docs to include an example of it working.

image


jeresig:

Built with commit 14a3d30da9deebe49893bca681ae7e544fd76e1d

https://deploy-preview-172--wonder-blocks.netlify.com

packages/wonder-blocks-core/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -34,6 +34,28 @@ exports[`wonder-blocks-core example 1 1`] = `
...
+        "minHeight": 0,
+        "minWidth": 0,
+        "padding": 4,
+        "position": "relative",
+        "zIndex": 0,
+      }
+    }
+  >
+    Even with an array and inline styles!
+  </div>
```
somewhatabstract: Added snapshot because of the docs change. Notice that all other snapshots are unchanged because the added class for inline styles is only done when not snapshotting.

packages/wonder-blocks-core/docs.md
```diff
@@ -1,9 +1,13 @@
-`wonder-blocks-core` provides building blocks for other `wonder-blocks` 
-components.  These components, `View` and `Text` can also be used on their 
-own.
+`wonder-blocks-core` provides building blocks for other `wonder-blocks` components.  These components, `View` and `Text` can also be used on their own.
```
somewhatabstract: Removed the newlines. Let the browser and things wordwrap for us.

packages/wonder-blocks-core/docs.md
```diff
@@ -1,9 +1,13 @@
...
 
 `View` roughly maps to `div` and `Text` roughly maps to `span`.
 
+These components can take styles in a variety of manners:
+
+- An inline style object
+- An `aphrodite` StyleSheet style
+- An array combining the above
```
somewhatabstract: Some extra docs on styling options.

packages/wonder-blocks-core/docs.md
```diff
@@ -22,6 +26,12 @@ const styles = StyleSheet.create({
     <Text style={styles.text}>
         Hello, world!
     </Text>
+    <View style={[
+        styles.container,
+        {backgroundColor: "goldenrod", padding: 4},
+    ]}>
+        Even with an array and inline styles!
+    </View>
```
somewhatabstract: Extended the example to show combining classes and inline styles, demonstrating this feature working.

packages/wonder-blocks-core/util/util.js
```diff
@@ -43,13 +43,14 @@ export function processStyleList<T: Object>(
         };
     }
 
+    // Check to see if we should inline all the styles for snapshot tests.
+    const shouldInlineStyles =
+        typeof global !== "undefined" && global.SNAPSHOT_INLINE_APHRODITE;
```
somewhatabstract: Moved this check for two reasons:

1. To reuse it
2. To make it clear what the check is for

packages/wonder-blocks-core/util/util.js
```diff
@@ -59,8 +60,22 @@ export function processStyleList<T: Object>(
...
+    // all the styles, remove this <ADD JIRA ISSUE HERE IF THIS PASSES REVIEW>
+    // If we're not snapshotting styles, let's create a class for the inline
+    // styles so that they can apply to the element even with aphrodite's
+    // use of !important.
+    if (inlineStyles.length > 0 && !shouldInlineStyles) {
+        const inlineStylesStyleSheet = StyleSheet.create({
+            inlineStyles: inlineStylesObject,
+        });
+        stylesheetStyles.push(inlineStylesStyleSheet.inlineStyles);
+    }
```
somewhatabstract: This is the code that creates the extra class for us.

packages/wonder-blocks-core/util/util.js
```diff
@@ -59,8 +60,22 @@ export function processStyleList<T: Object>(
...
+
     return {
-        style: Object.assign({}, ...inlineStyles),
+        style: shouldInlineStyles ? inlineStylesObject : {},
```
somewhatabstract: If we're not inlining styles, then give an empty object here as all our inline styles are now in a class.

packages/wonder-blocks-core/util/util.js
somewhatabstract: Note that it is pushed onto the end of the stylesheet list so that just as with inline styles, these take precedence over other classes.

packages/wonder-blocks-core/docs.md
somewhatabstract: Ah good to know, I can update that.

packages/wonder-blocks-core/util/util.js
kevinbarabash: This may have perf implications, but I know were doing similar things elsewhere in the codebase so maybe not it's not that bad.

somewhatabstract: I did some basic perf checks and didn't see anything that was so horrible as to push back on this. If there are, perhaps we can have an escape hatch and make this inline->class thing optional?

somewhatabstract: Like, if the style has a `_noClassConversion` flag set to `true`, we don't do this bit.

somewhatabstract: That way for animations, we can opt out.

kevinbarabash: Cool.  I also checked on slack and Emily and Matt aren't concerned about the perf implications.

PR #174, opened 2018-07-02
Add action menu to dropdown package

wishdasher:


jeresig:

Built with commit 4f5d33aa5b70046edcd2ceb010ff9a932c784027

https://deploy-preview-174--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -0,0 +1,237 @@
...
+    /**
+     * Text for the opener of this menu.
+     */
+    menuText: string,
+
+    /**
+     * Whether this menu contains select items. If true, user should also define
+     * the onChange and selectedItems props.
+     */
+    containsSelectionItems?: boolean,
```
kevinbarabash: This feels redundant.  Can't we assume `containsSelectionItems = Array.isArray(selectedValues)`?

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -0,0 +1,237 @@
...
+     * The values of the items that are currently selected. Use only if this
+     * menu contains select items (prop containsSelectionItems).
+     */
+    selectedValues?: Array<string>,
+
+    /**
+     * Whether this menu should be left-aligned or right-aligned with the
+     * opener component. Defaults to left-aligned.
+     */
+    alignment?: "left" | "right",
```
kevinbarabash: Since we have a default prop for this, this can be `alignment: "left" | "right"`... the benefit is then we don't have to check if `this.props.alignment` is defined before using it.

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -0,0 +1,237 @@
...
+
+    /**
+     * Whether this menu is disabled. A disabled menu may not be opened and
+     * does not support interaction. Defaults to false.
+     */
+    disabled?: boolean,
```
kevinbarabash: Same.

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -0,0 +1,237 @@
...
+        };
+    }
+
+    toggleMenu() {
+        this.setState((prevState) => ({
+            open: !prevState.open,
+        }));
+    }
+
+    handleSelected(selectedValue: string, newSelectionState: boolean) {
```
kevinbarabash: Could this be a toggle and then we wouldn't need the `newSelectionState` or are there cases where that isn't sufficient?

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -0,0 +1,237 @@
...
+        // If either of these are not defined, return.
+        if (!onChange || !selectedValues) {
+            return;
+        }
+
+        if (newSelectionState) {
+            onChange([...selectedValues, selectedValue].sort());
+        } else {
+            const updatedSelection = selectedValues;
+            updatedSelection.splice(selectedValues.indexOf(selectedValue), 1);
```
kevinbarabash: This still mutates `selectedValues`.  Instead we should do something like:
```
const index = selectedValues.indexOf(selectedValue);
const updatedSelection = [...selectedValues.slice(0, index), ...selectedValues.slice(index)];
```

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -0,0 +1,237 @@
...
+    handleSelected(selectedValue: string, newSelectionState: boolean) {
+        const {onChange, selectedValues} = this.props;
+
+        // If either of these are not defined, return.
+        if (!onChange || !selectedValues) {
+            return;
+        }
+
+        if (newSelectionState) {
+            onChange([...selectedValues, selectedValue].sort());
```
kevinbarabash: Why sort?  We may even want to switch to `Set` after the React 16 upgrade to make it more explicit that people should not be relying on the order of of this.

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -0,0 +1,237 @@
...
+                        key={index}
+                        disabled={item.disabled}
+                        indent={containsSelectionItems}
+                        label={item.label}
+                        href={item.href}
+                        clientNav={item.clientNav}
+                        onClick={item.onClick}
+                    />
+                );
+            }
```
kevinbarabash: It seems find to me to not have an `else` clause since `flow` keeps us safe.

packages/wonder-blocks-dropdown/components/action-menu.md
```diff
@@ -0,0 +1,208 @@
...
+that navigate to a link. There is also a separator.
+
+The second menu demonstrates a menu where the items only have onClick callbacks
+and do not navigate to a different menu. The onClick property may be used for
+purposes such as conversion logging.
+
+The third menu is disabled.
+
+The fourth menu is right-aligned.
+```js
```
kevinbarabash: Can you split this into separate examples?

packages/wonder-blocks-dropdown/components/action-menu.js
wishdasher: I changed `SelectItem` to pass back the old selection state instead. Technically, the menu has this information (but it would have to filter for the value). I think this information is useful because in the case of the single select menu, clicking on a selected item doesn't toggle it. So this signature from `SelectItem` allows each menu to decide what to do.

packages/wonder-blocks-dropdown/components/action-menu.js
wishdasher: (The current implementation sorts the list, but yeah there's no reason to do it so I'll remove it for now.)

packages/wonder-blocks-dropdown/components/action-menu.js
wishdasher: I'll leave this in because I think it's slightly easier to understand for someone new looking at the clause.

packages/wonder-blocks-dropdown/components/action-menu.md
```diff
@@ -0,0 +1,145 @@
...
+        flexDirection: "row",
+        height: 250,
+    },
+    wrapper: {
+        width: "100%",
+    }
+});
+<View style={[styles.row]}>
+    <View style={[styles.wrapper]}>
+    <ActionMenu
```
kevinbarabash: super nit: this needs to be indented more.

packages/wonder-blocks-dropdown/components/action-menu.md
```diff
@@ -0,0 +1,145 @@
...
+                {
+                    type: "select",
+                    label: "Show in-class assignments",
+                    onClick: (state) => console.log(`Show in-class assignments ${(!state).toString()}`),
+                    value: "in-class"
+                },
+
+            ]}
+            menuText={"Assignments"}
+            onChange={(selectedValues) => this.handleChanges(selectedValues)}
```
kevinbarabash: super nit: `handleChanges => handleChange`

packages/wonder-blocks-dropdown/components/action-menu.test.js
```diff
@@ -0,0 +1,81 @@
...
+        // Close menu with space
+        opener.simulate("keydown", {keyCode: keyCodes.space});
+        opener.simulate("keyup", {keyCode: keyCodes.space});
+        opener.simulate("click", {preventDefault: jest.fn()});
+        expect(menu.state("open")).toEqual(false);
+
+        // Open menu again with enter
+        opener.simulate("keydown", {keyCode: keyCodes.enter});
+        opener.simulate("click", {preventDefault: jest.fn()});
+        opener.simulate("keyup", {keyCode: keyCodes.enter});
```
kevinbarabash: It would be good to encapsulate these sequences for simulating `space` and `enter` into helpers so that we can reuse them, but that can be a separate PR.

packages/wonder-blocks-dropdown/components/action-menu.test.js
```diff
@@ -0,0 +1,81 @@
...
+    });
+
+    it("triggers actions and toggles select items as expected", () => {
+        expect(menu.state("open")).toEqual(true);
+
+        const actionItem = menu.find(ActionItem);
+        actionItem.simulate("mousedown");
+        actionItem.simulate("mouseup");
+        actionItem.simulate("click");
+        expect(onClick).toHaveBeenCalledTimes(1);
```
kevinbarabash: This doesn't appear to be called out in the spec, but I think the action menu should dismiss when you click an item.  Can you check with Kitt if that's the case and create a separate PR if necessary?

packages/wonder-blocks-dropdown/components/select-box.md
```diff
@@ -1,83 +0,0 @@
-### Example: Regular and placeholder select box versions
```
kevinbarabash: Good call deprecating this.

packages/wonder-blocks-dropdown/docs.md
```diff
@@ -1,51 +1 @@
-Temporary views of ActionItem, SelectItem, and SeparatorItem. Will be replaced
```
kevinbarabash: Can this file be removed if we remove it from styleguidist.config.js as well?

wishdasher:
Thanks for all your comments! I've addressed the nits and noted the TODO laters.

PR #177, opened 2018-07-03
Close menu on pressing escape key or clicking elsewhere

wishdasher:


wishdasher:
Uses refs to do the clicking to close. I wasn't able to figure out how to fix the flow errors, so I tyepd `handleClick` with `any` and added a `$FlowFixMe` somewhere.

jeresig:

Built with commit d12c724eda3c4bfa12467b6d4f966c2834e3ec70

https://deploy-preview-177--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -38,33 +43,98 @@ type DropdownCoreProps = {
 };
 
 export default class DropdownCore extends React.Component<DropdownCoreProps> {
+    node: ?HTMLDivElement;
+    // TODO(sophie): figure out flow typing
+    handleClick: (event: any) => void;
```
kevinbarabash: The `any` can be a `SyntheticMouseEvent<"click">` I think.  See https://www.saltycrane.com/flow-type-cheat-sheet/latest/.

kevinbarabash: I see, you have it defined below... you shouldn't have to define it here as well.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -38,33 +43,98 @@ type DropdownCoreProps = {
...
+
+    componentWillMount() {
+        document.addEventListener("click", this.handleClick, false);
+    }
+
+    componentWillUnmount() {
+        document.removeEventListener("click", this.handleClick, false);
+    }
+
+    handleClick(event: SyntheticEvent<>) {
```
kevinbarabash: This should be pre-bound since you're passing it directly to `add`/`removeEventListenere`.
```
handleClick = (event: SyntheticMouseEvent<"click">) => { ... }
```

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -38,33 +43,98 @@ type DropdownCoreProps = {
...
+        if (this.node && this.node.contains(event.target)) {
+            return;
+        }
+        this.props.onClose();
+    }
+
     render() {
         const {alignment, items, open, opener, style} = this.props;
 
+        // TODO(sophie): why doesn't ref work with View?
```
kevinbarabash: Because it's not a DOM element.  You'll need to call `ReactDOM.findDOMNode` on the `ref` to get the underlying DOM node.  We may change this in the future after we upgrade to React 16.

kevinbarabash: Please use `ref` on the `View`.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -38,33 +43,98 @@ type DropdownCoreProps = {
...
+                </View>
+            </div>
         );
     }
 }
 
+/** A component that, when mounted, calls `onClose` when Escape is pressed. */
+class DropdownKeypressListener extends React.Component<{
+    onClose: () => void,
+}> {
```
kevinbarabash: Why a different solution for handling the keyboard?

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -215,6 +221,7 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
             <DropdownCore
                 alignment={alignment}
                 items={menuItems}
+                onClose={() => this.closeMenu()}
```
kevinbarabash: How were we opening and closing things before?

packages/wonder-blocks-dropdown/components/multi-select-menu.js
```diff
@@ -216,6 +222,7 @@ export default class MultiSelectMenu extends React.Component<Props, State> {
                 alignment={alignment}
                 items={menuItems}
                 light={light}
+                onClose={() => this.closeMenu()}
```
kevinbarabash: nit: this should be:
```
onClose={() => this.handleClose()}
```
as per our React style guide.

packages/wonder-blocks-dropdown/components/single-select-menu.js
```diff
@@ -142,6 +148,7 @@ export default class SingleSelectMenu extends React.Component<Props, State> {
                 alignment={alignment}
                 items={menuItems}
                 light={light}
+                onClose={() => this.closeMenu()}
```
kevinbarabash: same

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -76,11 +76,15 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+        if (this.props.open) {
+            this.props.onClose();
+        }
     };
 
     _handleKeyup = (event: KeyboardEvent) => {
-        if (event.key === "Escape") {
+        if (this.props.open && event.key === "Escape") {
+            event.preventDefault();
+            event.stopImmediatePropagation();
```
kevinbarabash: This is to prevent also dismissing the modal?

PR #180, opened 2018-07-03
Add UniqueIDProvider

somewhatabstract:
This PR adds the `UniqueIDProvider` component for the provision of unique identifiers in a server-side render-friendly manner (for those with access, this is detailed in [ADR #60](https://docs.google.com/document/d/1BYQtb6GLZpruWEoYYQ74_7H8qZ41UbFuVsQBVGFew3o/edit#)).

The component uses the `NoSSR` component to control behavior based on whether something is being rendered for the first time or not, allowing us to ensure that initial rendering on both client and server will match.

This also adds documentation and test coverage for the component and the utilities that it uses.

NOTE: This is part of the `Tooltip` work as we need unique identifiers for the a11y solution.

### DIFF
--- a/packages/wonder-blocks-core/components/unique-id-provider.js
+++ b/packages/wonder-blocks-core/components/unique-id-provider.js
@@ -10,47 +10,56 @@ import type {IIdentifierFactory} from "../util/types.js";
 
 type Props = {|
     // A render prop that takes an instance of IIdentifierFactory and returns
-    // the content to be rendered. See the comments for placeholder prop for
-    // more details.
+    // the content to be rendered.
+    // If mockOnFirstRender is false, this is only called after
+    // the initial render has occurred and will always be called with the same
+    // IIdentifierFactory instance.
+    // If mockOnFirstRender is true, this is called once with
+    // a mock IIdentifierFactory for the initial render, and then a unique ID
+    // factory thereafter.
     children: (IIdentifierFactory) => React.Node,
 
-    // If provided, this render prop is called when rendering for the very first
-    // time. This is useful to support server-side rendering with different
-    // content than will eventually be rendered in the client. The first client
-    // render and the server render will use this prop.
-    //
-    // If not provided, the children render prop will be called on first render.
-    //
-    // Whether this render prop, or children, the first render will receive an
-    // instance of IIdentifierFactory where the id method returns the same
-    // string with which it is called, rather than guaranteeing uniqueness.
-    placeholder?: (IIdentifierFactory) => React.Node,
+    // If mockOnFirstRender is false, children is only called
+    // after the initial render has occurred.
+    // If mockOnFirstRender is true, children is called once with
+    // a mock IIdentifierFactory for the initial render, and then a unique ID
+    // factory thereafter.
+    mockOnFirstRender?: boolean,
 
     // If this prop is specified, any identifiers provided will contain the
     // given scope. This can be useful for making easily readable identifiers.
-    scope?: string,
+    +scope?: string,
 |};
 
 export default class UniqueIDProvider extends React.Component<Props> {
     _idFactory: IIdentifierFactory;
 
-    _performRender() {
-        const {children, placeholder, scope} = this.props;
-        if (this._idFactory) {
-            return children(this._idFactory);
-        }
-        // We haven't rendered yet. This must be our first render.
-        // Let's create our real id factory for use in subsequent renders
-        // and then call our placeholder prop with our SSR-friendly factory.
-        this._idFactory = new UniqueIDFactory(scope);
+    _performRender(firstRender: boolean) {
+        const {children, mockOnFirstRender, scope} = this.props;
 
-        return (placeholder || children)(SsrIDFactory);
+        // If this is our first render, we're going to stop right here.
+        if (firstRender) {
+            // We'll be needing this on the next render, so let's set it up.
+            this._idFactory = new UniqueIDFactory(scope);
+
+            if (mockOnFirstRender) {
+                // We're allowing an initial render, so let's pass our mock
+                // identifier factory to support SSR.
+                return children(SsrIDFactory);
+            }
+            return null;
+        }
+
+        // It's a regular render, so let's use our identifier factory.
+        return children(this._idFactory);
     }
 
     render() {
+        // Here we use the NoSSR component to control when we render and whether
+        // we provide a mock or real identifier factory.
         return (
-            <NoSSR placeholder={() => this._performRender()}>
-                {() => this._performRender()}
+            <NoSSR placeholder={() => this._performRender(true)}>
+                {() => this._performRender(false)}
             </NoSSR>
         );
     }
--- a/packages/wonder-blocks-core/components/unique-id-provider.js
+++ b/packages/wonder-blocks-core/components/unique-id-provider.js
@@ -9,25 +9,37 @@ import SsrIDFactory from "../util/ssr-id-factory.js";
 import type {IIdentifierFactory} from "../util/types.js";
 
 type Props = {|
-    // A render prop that takes an instance of IIdentifierFactory and returns
-    // the content to be rendered.
-    // If mockOnFirstRender is false, this is only called after
-    // the initial render has occurred and will always be called with the same
-    // IIdentifierFactory instance.
-    // If mockOnFirstRender is true, this is called once with
-    // a mock IIdentifierFactory for the initial render, and then a unique ID
-    // factory thereafter.
+    /**
+     * A render prop that takes an instance of IIdentifierFactory and returns
+     * the content to be rendered.
+     *
+     * If mockOnFirstRender is false, this is only called after
+     * the initial render has occurred and will always be called with the same
+     * IIdentifierFactory instance.
+     *
+     * If mockOnFirstRender is true, this is called once with
+     * a mock IIdentifierFactory for the initial render, and then a unique ID
+     * factory thereafter.
+     *
+     * Full type with `IIdentifierFactory` definition inlined is:
+     *
+     * `{get(id: string): string} => React.Node`
+     */
     children: (IIdentifierFactory) => React.Node,
 
-    // If mockOnFirstRender is false, children is only called
-    // after the initial render has occurred.
-    // If mockOnFirstRender is true, children is called once with
-    // a mock IIdentifierFactory for the initial render, and then a unique ID
-    // factory thereafter.
+    /**
+     * If mockOnFirstRender is false, children is only called
+     * after the initial render has occurred.
+     * If mockOnFirstRender is true, children is called once with
+     * a mock IIdentifierFactory for the initial render, and then a unique ID
+     * factory thereafter.
+     */
     mockOnFirstRender?: boolean,
 
-    // If this prop is specified, any identifiers provided will contain the
-    // given scope. This can be useful for making easily readable identifiers.
+    /**
+     * If this prop is specified, any identifiers provided will contain the
+     * given scope. This can be useful for making easily readable identifiers.
+     */
     +scope?: string,
 |};
 
--- a/packages/wonder-blocks-core/components/unique-id-provider.js
+++ b/packages/wonder-blocks-core/components/unique-id-provider.js
@@ -1,0 +1,57 @@
+// @flow
+import * as React from "react";
+
+import NoSSR from "./no-ssr.js";
+
+import UniqueIDFactory from "../util/unique-id-factory.js";
+import SsrIDFactory from "../util/ssr-id-factory.js";
+
+import type {IIdentifierFactory} from "../util/types.js";
+
+type Props = {|
+    // A render prop that takes an instance of IIdentifierFactory and returns
+    // the content to be rendered. See the comments for placeholder prop for
+    // more details.
+    children: (IIdentifierFactory) => React.Node,
+
+    // If provided, this render prop is called when rendering for the very first
+    // time. This is useful to support server-side rendering with different
+    // content than will eventually be rendered in the client. The first client
+    // render and the server render will use this prop.
+    //
+    // If not provided, the children render prop will be called on first render.
+    //
+    // Whether this render prop, or children, the first render will receive an
+    // instance of IIdentifierFactory where the id method returns the same
+    // string with which it is called, rather than guaranteeing uniqueness.
+    placeholder?: (IIdentifierFactory) => React.Node,
+
+    // If this prop is specified, any identifiers provided will contain the
+    // given scope. This can be useful for making easily readable identifiers.
+    scope?: string,
+|};
+
+export default class UniqueIDProvider extends React.Component<Props> {
+    _idFactory: IIdentifierFactory;
+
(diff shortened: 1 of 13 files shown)


packages/wonder-blocks-core/components/unique-id-provider.test.js
```diff
@@ -0,0 +1,119 @@
...
+        });
+
+        test("all renders get same unique id factory", () => {
+            // Arrange
+            const children = jest.fn(() => <View />);
+            const nodes = <UniqueIDProvider>{children}</UniqueIDProvider>;
+            const wrapper = mount(nodes);
+
+            // Act
+            wrapper.instance().forceUpdate();
```
kevinbarabash: In practice what causes this update to occur on the client in normal use?

packages/wonder-blocks-core/components/unique-id-provider.test.js
```diff
@@ -0,0 +1,119 @@
...
+            // Arrange
+            const children = jest.fn(() => null);
+            const nodes = <UniqueIDProvider>{children}</UniqueIDProvider>;
+
+            // Act
+            mount(nodes);
+
+            // Assert
+            expect(children).toHaveBeenCalledTimes(1);
+            expect(children).not.toHaveBeenCalledWith(SsrIDFactory);
```
kevinbarabash: Would
```
expect(children.mock.calls[0][0]).toBeInstanceOf(UniqueIDFactory);
```
be try in this test?

packages/wonder-blocks-core/components/unique-id-provider.test.js
```diff
@@ -0,0 +1,119 @@
...
+            );
+
+            // Act
+            mount(nodes);
+
+            // Assert
+            // Called twice; once for initial mock render, and again for real
+            // render.
+            expect(children).toHaveBeenCalledTimes(2);
+            expect(children).toHaveBeenCalledWith(SsrIDFactory);
```
kevinbarabash: Maybe check
```
expect(children.mock.calls[1][0]).toBeInstanceOf(UniqueIDFactory);
```
as well.

packages/wonder-blocks-core/components/unique-id-provider.md
```diff
@@ -0,0 +1,101 @@
...
+    <View style={{flexDirection: "row"}}><Button onClick={onClick}>Click Me to Rerender</Button><Spring /></View>
+    <Strut size={16} />
+    <HeadingSmall>The UniqueIDProvider:</HeadingSmall>
+    {provider}
+</View>
+```
+
+### mockOnFirstRender is true
+
+When specifying `mockOnFirstRender` to be `true`, the first render will use a mock identifier factory that doesn't guarantee identifier uniqueness.
```
kevinbarabash: `mockOnFirstRender` doesn't explain what's being mocked, maybe `nonUniqueIdOnFirstRender` or `nonUniqueIdOnMount`.

packages/wonder-blocks-core/components/unique-id-provider.md
```diff
@@ -0,0 +1,101 @@
...
+`UniqueIDProvider` ensures every identifier factory is unique using a unique number for each one. However, this isn't very readable when wanting to differentiate the types of things using unique identifiers. If we want to, we can provide a `scope` prop that adds some text to each identifier provided. This can be useful for providing some quick at-a-glance component identification to identifiers when there are multiple providers.
+
+```jsx
+const {Body, HeadingSmall, BodyMonospace} = require("@khanacademy/wonder-blocks-typography");
+const {Spring, Strut} = require("@khanacademy/wonder-blocks-layout");
+
+const children = ({id}) => (
+    <View>
+        <Body>
+            The id returned for "my-identifier": {id("my-identifier")}
```
kevinbarabash: If I want multiple identifiers within `children` should I actually use multiple `UniqueIDProviders` and split up `children` into separate subtrees?

packages/wonder-blocks-core/components/unique-id-provider.md
```diff
@@ -0,0 +1,101 @@
+The `UniqueIDProvider` component is how Wonder Blocks components obtain unique identifiers. This component ensures that server-side rendering and initial client rendering match while allowing the provision of unique identifiers for the client. It does this by not guaranteeing unique identifiers for the initial render, but instead allowing for a placeholder to be rendered in place of the real content, or for non-unique values to be provided as identifiers.
```
kevinbarabash: My understanding is that `UniqueIDProvider` only provides a single unique identifier.  It might be worth calling that out here.

packages/wonder-blocks-core/components/unique-id-provider.js
```diff
@@ -0,0 +1,66 @@
...
+        if (firstRender) {
+            // We'll be needing this on the next render, so let's set it up.
+            this._idFactory = new UniqueIDFactory(scope);
+
+            if (mockOnFirstRender) {
+                // We're allowing an initial render, so let's pass our mock
+                // identifier factory to support SSR.
+                return children(SsrIDFactory);
+            }
+            return null;
```
kevinbarabash: It would be nice to provide a way to specify a placeholder in this case.

packages/wonder-blocks-core/components/unique-id-provider.md
kevinbarabash: Can you provide a real world example of when people would want to use this prop?

packages/wonder-blocks-core/util/unique-id-factory.js
```diff
@@ -0,0 +1,58 @@
+// @flow
+import {IIdentifierFactory} from "../util/types.js";
+
+/**
+ * This is NOT for direct use. Instead, see the UniqueIDProvider component.
+ *
+ * Implements IIdentifierFactory to provide unique identifiers.
+ */
+export default class UniqueIDFactory implements IIdentifierFactory {
```
kevinbarabash: The naming of this is a little confusing.  I would expect a factory of unique ids to be able to return multiple ids each of which are unique.  If I run the following code:
```
const uniqueIdFactory = new UniqueIDFactory("foo");
const id1 = uniqueIdFactory.id("bar");
const id2 = uniqueIdFactory.id("bar");
```
I would expect `id1 !== id2` to be true, but that isn't actually the case.  If we want to use the factory pattern there should be a singleton that call `UniqueIDFactory` that returns instances of this class and this class should have a different name.

packages/wonder-blocks-core/util/unique-id-factory.js
```diff
@@ -0,0 +1,58 @@
...
+    _uniqueFactoryName: string;
+
+    static _factoryUniquenessCounter = 0;
+
+    /**
+     * Creates a UniqueIDFactory instance.
+     *
+     * @param {string} factoryName An optional case-insensitive name for the
```
kevinbarabash: `factoryName -> scope`

packages/wonder-blocks-core/util/unique-id-factory.js
```diff
@@ -0,0 +1,58 @@
...
+        if (!this._hasValidIdChars(normalizedScope)) {
+            throw new Error(`Invalid factory name: ${scope}`);
+        }
+        this._uniqueFactoryName = `uid-${normalizedScope}-${UniqueIDFactory._factoryUniquenessCounter++}`;
+    }
+
+    _hasValidIdChars(value: ?string) {
+        if (typeof value === "string") {
+            const invalidCharsReplaced = value.replace(/[^\d\w-]/g, "-");
+            if (value === invalidCharsReplaced) {
```
kevinbarabash: Why not
```
return typeof value === "string" ? !`/[^\d\w-]/g.test(value)` : false;
```

packages/wonder-blocks-core/util/unique-id-factory.js
```diff
@@ -0,0 +1,58 @@
...
+    constructor(scope?: string) {
+        scope = typeof scope === "string" ? scope : "";
+        const normalizedScope = scope.toLowerCase();
+        if (!this._hasValidIdChars(normalizedScope)) {
+            throw new Error(`Invalid factory name: ${scope}`);
+        }
+        this._uniqueFactoryName = `uid-${normalizedScope}-${UniqueIDFactory._factoryUniquenessCounter++}`;
+    }
+
+    _hasValidIdChars(value: ?string) {
```
kevinbarabash: Maybe `_isValidId`?

packages/wonder-blocks-core/util/unique-id-factory.test.js
```diff
@@ -0,0 +1,165 @@
...
+
+            // Act
+            const id1 = factory1.id("testname");
+            const id2 = factory2.id("testname");
+
+            // Assert
+            expect(id2).not.toEqual(id1);
+        });
+    });
+});
```
kevinbarabash: nice tests

packages/wonder-blocks-core/components/unique-id-provider.js
```diff
@@ -0,0 +1,66 @@
...
+type Props = {|
+    // A render prop that takes an instance of IIdentifierFactory and returns
+    // the content to be rendered.
+    // If mockOnFirstRender is false, this is only called after
+    // the initial render has occurred and will always be called with the same
+    // IIdentifierFactory instance.
+    // If mockOnFirstRender is true, this is called once with
+    // a mock IIdentifierFactory for the initial render, and then a unique ID
+    // factory thereafter.
+    children: (IIdentifierFactory) => React.Node,
```
kevinbarabash: Unfortunately, styleguidist won't replace `IIdentifierFactory` with the full signature of this callback which makes this bad from a documentation perspective.  There's a task in Jira to do something about this, but we haven't gotten around to it yet. :(  In the meantime can you use the full signature here?

packages/wonder-blocks-core/components/unique-id-provider.test.js
somewhatabstract: A render. All we're doing here is forcing the render to occur again so that we can see we always get the same factory. In reality rather than forcing a render, something would have changed that caused react to re-render.

packages/wonder-blocks-core/components/unique-id-provider.test.js
somewhatabstract: I've added some additional commentary and reworked the asserts for this.

packages/wonder-blocks-core/components/unique-id-provider.md
somewhatabstract: See ADR #60. I implemented this to that specification, hence the name of the prop. It also explains why the prop is there. I will carry across some of that info.

I did originally use a different prop name but realised we made a decision when ADR60 was approved, so it would behoove me to follow the decision.

packages/wonder-blocks-core/util/unique-id-factory.js
somewhatabstract: That's not right at all.

The point here is that the factory produces unique identifiers for a given instance. If we ask for `id("bar")` it will give us a unique ID for "bar". If we ask for `id("foo")` it will give us a unique ID for id "foo". Calls to ask for that ID will then always return the same identifier, allowing us to just have `id("foo")` in our code and not care about what it really means. 

That's the pattern we established for webapp and for ADR 60 and the prior ADR 58.

packages/wonder-blocks-core/util/unique-id-factory.js
somewhatabstract: Legibility?

packages/wonder-blocks-core/util/unique-id-factory.js
somewhatabstract: This was copied directly from our webapp implementation. I didn't want to mislead that we were checking anything but the character make-up of the ID. `_isValidId` implies it might check more about what makes an identifier valid (such as enforcing no leading numbers).

packages/wonder-blocks-core/util/unique-id-factory.js
somewhatabstract: I've reworked this method to be a little cleaner. I think it reads better now.

packages/wonder-blocks-core/components/unique-id-provider.js
somewhatabstract: Ugh, that's a maintenance nightmare as it decouples the type definitions from one another just to satisfy the docs. Is there not a better way? I think I'd rather hand type the interface definition into some part of the docs than weaked our code type checking for the sake of documentation.

packages/wonder-blocks-core/components/unique-id-provider.md
somewhatabstract: I've updated the docs with more detail.

somewhatabstract:
@kevinbarabash I've responded to your comments and made some changes.

packages/wonder-blocks-core/components/unique-id-provider.md
kevinbarabash: I think `mockOnFirstRender` makes sense for `KAProvider` b/c it's providing a mock version of the `KA` object.  It makes less sense for `UniqueIDProvider`, but that discussion should probably happen in the ADR.  ADR's aren't immutable are they?

packages/wonder-blocks-core/components/unique-id-provider.js
kevinbarabash: I see.  Can you add an example in the docs that shows how these two components might work together?

packages/wonder-blocks-core/components/unique-id-provider.md
somewhatabstract: I think once they are accepted, we should see them as immutable and have additional ADR's that change them. Otherwise, I think things get complicated.

However, I think it makes sense to have a way to control that first render. I wonder if, instead of not having it or providing a dummy factory, we should provide `null` for the factory instead. That would be my preference.

packages/wonder-blocks-core/components/unique-id-provider.js
somewhatabstract: I hand typed the interface definition into some part of the docs rather than weaken our code type checking for the sake of documentation. It was not fun though; I raised an issue on why.

packages/wonder-blocks-core/util/unique-id-factory.js
kevinbarabash: Thanks for pointing back to ADR 58.  Sorry I didn't mean to rehash previous decisions.  I do find the naming of things a bit confusing, but the tests and documentation and help.  Thanks for calling out that calling `.get("test")` will always return the same value.

packages/wonder-blocks-core/components/unique-id-provider.js
kevinbarabash: Yeah, typing it in the docs sounds like a good interim solution.

packages/wonder-blocks-core/util/unique-id-factory.js
somewhatabstract: Naming things is hard. I'm not particularly hooked on the names, so we could change them if we think of something better.

PR #181, opened 2018-07-04
Adding TooltipCoordinator

somewhatabstract:
This PR adds a new component for the tooltip; `TooltipCoordinator`. This new component is responsible for controlling which tooltip is visible and when. It coordinates all the tooltips wanting to be visible and controls which one wins. It also controls whether tooltips display instantly or after a short delay. As per the design specification, the delay is skipped if an existing tooltip is already displayed when another becomes active.

This also sets up things so that the `TooltipBubble` can know whether it should animate its appearance/disappearance or not.

This is a somewhat complicated feature to implement so here's a brief overview of the pieces (there are also comprehensive unit testing and a docs example):

1. `setFlushableTimeout`
    This is a new function exported from `wonder-blocks-core`. It wraps `setTimeout`, providing callers with the ability to clear (cancel the timeout) and flush (force the timeout function to happen immediately), as well as just wait for the timeout to occur. It also provides a promise that is resolved on clear, flush, or timeout, which also passes whether the timeout was cancelled or not as a boolean to the function that was to be called.
1. `SuppressionTracker`
    This is the main worker in this feature. It tracks suppressees (things that implement the `ICanBeSuppressed` interface) using a stack. When a new suppressee is tracked, it hides the active tracked suppressee and shows the new one. When the active suppressee is untracked, it hides it and shows the next one that is still being tracked. If a new suppressee is the first to be tracked, it is unsuppressed after an initial delay. If an untracked suppressee is the last to be untracked, it is suppressed after a final delay.
1. `TooltipCoordinator`
    This component sits between the `TooltipAnchor` and the `TooltipPortalMounter`. When the anchor tells the coordinator to show the tooltip, the coordinate uses a global instance of our new `SuppressionTracker` to ensure that any currently active tooltip is hidden and the new tooltip is shown. The global tracker then calls back into the relevant coordinators (which implement `ICanBeSurpressed` to tell them when to actually show/hide their tooltips. The coordinator changes its state accordingly and passes it on to its children via render function. This allows us to slip this component into the render chain for the tooltip without having to change other components already present.


jeresig:

Built with commit de425c557ab90c135c48ca01c5c4bdd00a9cf847

https://deploy-preview-181--wonder-blocks.netlify.com

somewhatabstract:
Hmm, something isn't right here. I'm going to try this from a different angle.

PR #182, opened 2018-07-04
update flow to 0.75

kevinbarabash:
Now that webapp is using flow 0.75 wonder-blocks should be using the same.  I added a couple of `$FlowFixMe`s.  The ones in action-item.js are confusing b/c we have the exact same code in button-core.js and other files which doesn't raise an error.


jeresig:

Built with commit 196460fbd5533f759e49a6b723db8a2974beccd2

https://deploy-preview-182--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/focus-trap.js
```diff
@@ -52,19 +52,17 @@ export default class FocusTrap extends React.Component<{children: React.Node}> {
 
     /** Try to focus the given node. Return true iff successful. */
     _tryToFocus(node: Node) {
-        if (typeof node.focus !== "function") {
-            return;
-        }
+        if (node instanceof HTMLElement) {
+            this._ignoreFocusChanges = true;
+            try {
+                node.focus();
```
kevinbarabash: @matchu why would this throw?

matchu: Not sure, copied this from the WAI-ARIA example and sorta took it for granted, sorry! 😅

We don't make any other attempt to check whether this element is focusable, so I imagine this is a safeguard in case the element is non-focusable.

But it seems like today's browsers silently ignore the request instead of throwing, so maybe this was to work around older browsers that aren't in circulation anymore?

PR #183, opened 2018-07-04
update some of tooltip tests to use async/await

kevinbarabash:
Unfortunately I wasn't able to convert all async tests to use async/await.  In particular I couldn't get the tests in tooltip-portal-mounter.test.js to work b/c for some reason `node` wasn't actually mounted.  I do believe that using async/await results in more readable tests so it would be good to figure out why things weren't working in that particular situation.

### DIFF
--- a/packages/wonder-blocks-modal/components/modal-launcher-portal.test.js
+++ b/packages/wonder-blocks-modal/components/modal-launcher-portal.test.js
@@ -6,41 +6,40 @@ import ModalLauncherPortal from "./modal-launcher-portal.js";
 import {ModalLauncherPortalAttributeName} from "../util/constants.js";
 
 describe("ModalPortal", () => {
-    test("Mounts its children in document.body", (done) => {
-        // Once the child mounts, check that it was mounted directly-ish into
-        // `document.body`, and finish the test.
-        const childrenRef = (element) => {
-            if (element) {
-                // Find the nearest parent, disregarding nodes created by
-                // ModalLauncherPortal and by `ReactDOM.render`.
-                let parent = element.parentElement;
-                while (
-                    parent &&
-                    (parent.hasAttribute(ModalLauncherPortalAttributeName) ||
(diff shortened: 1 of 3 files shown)


jeresig:

Built with commit 7381802d1b26689d282c4fc24114eaa1cd369745

https://deploy-preview-183--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-launcher-portal.test.js
```diff
@@ -6,41 +6,40 @@ import ModalLauncherPortal from "./modal-launcher-portal.js";
...
-                    parent = parent.parentNode;
-                }
-
-                // This nearest parent _should_ be document.body. It definitely
-                // should not be the `data-this-should-not-contain-the-children`
-                // element! That element shouldn't be in the ancestor chain at
-                // all.
-                expect(parent).toBe(document.body);
-                done();
+        if (element) {
```
kevinbarabash: @matchu if there is no `element` should the test fail?

somewhatabstract:
I have some similar changes coming in a PR so we may want to consolidate on some common approaches

kevinbarabash:
@somewhatabstract we can merge your changes first and go from there.

packages/wonder-blocks-modal/components/modal-launcher-portal.test.js
matchu: I think this was because we're effectively treating this as a `componentDidMount` handler, but `ref` is called on both mount and unmount, and we want to ignore the unmount case. Probably would be better to short-circuit if `element` is absent, with a comment explaining why. Sorry for the ambiguity!

kevinbarabash: That makes sense.  I keep forgetting that it's called on unmount as well.

kevinbarabash:
> Looking forward to when we can use async/await in our code!

Me too!

PR #184, opened 2018-07-05
Added some test coverage for the arrow

somewhatabstract:
This adds a little extra testing for the `TooltipTail` component.

Unusually, I have included some testing of internal pieces. This is for two reasons:
1. To validate that our error throwing that will indicate if we missed a placement option is working
1. To boost coverage

Normally I wouldn't do this as external testing should cover all cases, but for maintenance-related code, internal tests are useful.


jeresig:

Built with commit 4969be55401db1e7cf185b56dc556ee7d9dc24ec

https://deploy-preview-184--wonder-blocks.netlify.com

PR #185, opened 2018-07-05
Demonstrate how the Toolbar can be used in ModalTitleBar

scotchfield:
The title id reference might need to change, or at least be accessible for modification through a Tooltip prop. This seems to drop in without much overhead though!


jeresig:

Built with commit ebc7d21aa307acc59c07527fce6601dfd2b10b53

https://deploy-preview-185--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-title-bar.js
```diff
@@ -25,83 +18,17 @@ type Props = {
...
-            >
-                <View style={styles.titleAndSubtitle}>
-                    {this._renderTitleAndSubtitle()}
-                </View>
-            </View>
-        );
+        const {color, title, subtitle} = this.props;
+
+        // TODO(scottgrant): Set size appropriately using mediaSize.
+        return <Toolbar color={color} title={title} subtitle={subtitle} />;
```
kevinbarabash: I wonder if we even need the `ModalTitleBar` component now.  We could just use `Toolbar` directly.

kevinbarabash:
Coverage went down b/c you removed a file that had good coverage.  That's okay.

packages/wonder-blocks-modal/components/modal-panel.js
```diff
@@ -7,9 +7,9 @@ import {View} from "@khanacademy/wonder-blocks-core";
 
 import ModalCloseButton from "./modal-close-button.js";
 import ModalContent from "./modal-content.js";
-import ModalTitleBar from "./modal-title-bar.js";
 import ModalHeader from "./modal-header.js";
 import ModalFooter from "./modal-footer.js";
+import {Toolbar} from "@khanacademy/wonder-blocks-toolbar";
```
jeresig: 🛑 nit: I'd move this up to be with Color/View and the other things from other modules.

packages/wonder-blocks-modal/components/standard-modal.js
```diff
@@ -5,7 +5,7 @@ import {StyleSheet} from "aphrodite";
 import ModalDialog from "./modal-dialog.js";
 import ModalPanel from "./modal-panel.js";
 import ModalContent from "./modal-content.js";
-import ModalTitleBar from "./modal-title-bar.js";
+import {Toolbar} from "@khanacademy/wonder-blocks-toolbar";
```
jeresig: 🛑 Here, too!

PR #186, opened 2018-07-06
Fixup Tooltip prop docs

somewhatabstract:
This changes all the documentation on props and such to use block (`/** */`) comments instead of line (`//`) comments so that they get picked up by styleguidist.


jeresig:

Built with commit 6d2e6834b08a543362d144ec421fcf13ba839be4

https://deploy-preview-186--wonder-blocks.netlify.com

PR #187, opened 2018-07-06
add eslint rule for event handlers names

kevinbarabash:
Uses [react/jsx-handler-names](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-handler-names.md) to check that our event handlers conform to our react style guide guidelines for [event handlers](https://github.com/Khan/style-guides/blob/master/style/react.md#name-handlers-handleeventname).

### DIFF
--- a/packages/wonder-blocks-dropdown/components/multi-select-menu.js
+++ b/packages/wonder-blocks-dropdown/components/multi-select-menu.js
@@ -172,6 +172,7 @@ export default class MultiSelectMenu extends React.Component<Props, State> {
                     disabled={item.disabled}
                     key={item.value}
                     label={item.label}
+                    /* eslint-disable-next-line react/jsx-handler-names */
                     onClick={item.onClick}
                     onToggle={(value, state) =>
                         this.handleSelected(value, state)
--- a/packages/wonder-blocks-modal/components/modal-launcher.js
+++ b/packages/wonder-blocks-modal/components/modal-launcher.js
@@ -71,7 +71,7 @@ export default class ModalLauncher extends React.Component<Props, State> {
         this.setState({opened: true});
     };
 
-    _closeModal = () => {
+    handleCloseModal = () => {
         this.setState({opened: false}, () => {
             this.props.onClose && this.props.onClose();
         });
@@ -80,7 +80,7 @@ export default class ModalLauncher extends React.Component<Props, State> {
     _renderModal() {
         if (typeof this.props.modal === "function") {
             return this.props.modal({
-                closeModal: this._closeModal,
+                closeModal: this.handleCloseModal,
             });
         } else {
             return this.props.modal;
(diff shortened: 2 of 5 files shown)


packages/wonder-blocks-dropdown/components/multi-select-menu.js
```diff
@@ -166,6 +166,7 @@ export default class MultiSelectMenu extends React.Component<Props, State> {
                     disabled={item.disabled}
                     key={item.value}
                     label={item.label}
+                    /* eslint-disable-next-line react/jsx-handler-names */
                     onClick={item.onClick}
```
kevinbarabash: This would not have been a violation if we were doing `onClick={this.props.onClick}`.

jeresig:

Built with commit 6a1d7a7b53b1346653d15ab84cb0becb12086e68

https://deploy-preview-187--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-launcher.js
```diff
@@ -71,7 +71,7 @@ export default class ModalLauncher extends React.Component<Props, State> {
         this.setState({opened: true});
     };
 
-    _closeModal = () => {
+    handleCloseModal = () => {
```
kevinbarabash: We could change the prefix to `_handle` if we want to, but then we'd have to use `_`s for all handlers which is a bigger change.

matchu: Hmm if we have a style of underscoring private methods anyway, then making the prefix be `_handle` seems appropriate. But I'm not sure that's a style we use consistently, or whether we should…

kevinbarabash: It's not really clear to me what constitutes a private method in a React class. You almost never call a method directly on a React class so should all methods that aren't the constructor, render, or lifecycle methods be marked as private?

matchu: That's how I treat it, yeah :) You sometimes _do_ call methods directly on a React class, via refs, so I suppose that's the only place where the distinction really matters. (I also like it for further distinguishing between custom methods vs lifecycle methods, but I suppose that placement and other IDE tricks can do that just as effectively.)

So yeah, not sure what we want here!

matchu: That's how I treat it, yeah :) You sometimes _do_ call methods directly on a React class, via refs, so I suppose that's the only place where the distinction really matters. I also like it for consistency with how I name things in other JS classes, but we don't have many of those tbh. And I also like it for further distinguishing between custom methods vs lifecycle methods, but I suppose that placement and other IDE tricks can do that just as effectively!

So yeah, not sure what we want here!

kevinbarabash: Our React style guide makes no mention of private methods and the prefix it uses for handlers is `handle` so I think we should go with that for now until there's an ADR to change it.

jeresig: Yeah, I've never been sure why we were doing _ prefixing as it's really hard to dip into the methods of a React class in the first place!

packages/wonder-blocks-dropdown/components/multi-select-menu.js
jeresig: We'd have to pass in the item as well, presumably. Would this be ok: `onClick={() => this.handleClick(item)}`?

wishdasher: The menu doesn't do anything with this `onClick`. It's an optional prop on `SelectItem`, to allow the menu creator to attach something like a conversion to a specific item. Maybe we can have something like
```
handleClick(item) { item.onClick(); }
```

kevinbarabash: Unfortunately the lint rule doesn't handle the `onClick={() => this.handleClick(item)}` format so this rule it easy to get around.  It shouldn't be too hard to update it to handle this case as well.

PR #188, opened 2018-07-06
Add behaviour: Escape key closes active tooltip

somewhatabstract:
This is a pre-cursor to #181.

It does these things:
1. Adds handling of Escape key to hide active tooltip
1. Tidies up the `TooltipAnchor` tests to use `async` and nicer code
1. Moves some things to testing helper functions


jeresig:

Built with commit 3eb723156fee855450755e860af22b9ba9885cca

https://deploy-preview-188--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-launcher.js
```diff
@@ -125,7 +125,11 @@ class ModalLauncherKeypressListener extends React.Component<{
     }
 
     _handleKeyup = (e: KeyboardEvent) => {
-        if (e.key === "Escape") {
+        const escape = 27;
+        const keyCode = e.which || e.keyCode;
```
kevinbarabash: I thought `which` was only necessary for old versions of IE.

packages/wonder-blocks-modal/components/modal-launcher.js
```diff
@@ -125,7 +125,11 @@ class ModalLauncherKeypressListener extends React.Component<{
     }
 
     _handleKeyup = (e: KeyboardEvent) => {
-        if (e.key === "Escape") {
+        const escape = 27;
```
kevinbarabash: I like this.

packages/wonder-blocks-modal/components/modal-launcher.js
```diff
@@ -125,7 +125,11 @@ class ModalLauncherKeypressListener extends React.Component<{
...
+        const keyCode = e.which || e.keyCode;
+        if (keyCode === escape) {
+            e.preventDefault();
+            e.stopImmediatePropagation();
```
kevinbarabash: Can there be a situation where a tooltip is still active below a modal that's been opened?  I'm guessing that this is defensive.  Can you add a comment about the intent?

packages/wonder-blocks-modal/components/modal-launcher.test.js
```diff
@@ -73,10 +73,10 @@ describe("ModalLauncher", () => {
         wrapper.find("button").simulate("click");
         expect(document.querySelector("[data-modal-child]")).toBeTruthy();
 
-        // Simulate an Escape keypress. This will happen synchronously, because
-        // we're using `dispatchEvent`.
+        // Simulate an Escape keypress.
         const event: KeyboardEvent = (document.createEvent("Event"): any);
         event.key = "Escape";
+        event.which = 27;
```
kevinbarabash: What about `keyCode`?  Also, is `key` still necessary now that we're not checking for it?

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -144,6 +146,12 @@ export default class TooltipAnchor extends React.Component<Props, State> {
         // If we changed state, call our subscriber and let them know.
         if (oldState !== newState) {
             this.setState({active: newState});
+
+            if (newState) {
```
kevinbarabash: I wonder if we should use `newActive` and `oldActive`?

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -106,6 +105,9 @@ export default class TooltipAnchor extends React.Component<Props, State> {
                 this._handleMouseLeave,
             );
         }
+        if (this.state.active) {
+            document.removeEventListener("keyup", this._handleKeyUp);
+        }
```
kevinbarabash: Good catch.

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -163,6 +171,16 @@ export default class TooltipAnchor extends React.Component<Props, State> {
         this._updateActiveState(false, this._focused);
     };
 
+    _handleKeyUp = (e: KeyboardEvent) => {
+        const escape = 27;
+        const keyCode = e.which || e.keyCode;
+        if (keyCode === escape && this.state.active) {
+            e.preventDefault();
+            e.stopImmediatePropagation();
+            this._updateActiveState(false, false);
```
kevinbarabash: This is to prevent the modal from also closing.  Can you add a comment about as to why we do `preventDefault` and `stopImmediatePropagation` here?  Also, why `stopImmediatePropagation` and not `stopPropagation`?

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.test.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 import * as React from "react";
-import {mount as enzymeMount} from "enzyme";
+
+import {mount, unmountAll} from "../../../utils/testing/mount.js";
```
kevinbarabash: It's too bad enzyme doesn't automatically `unmountAll` after each test.  Thanks for extracting this an pulling it out to a top level.  At some point we may want to make a separate package in the packages/ folder for test helpers like this.  That way we can do:
```
import {mount, unmountAll} from "@khanacademy/wonder-blocks-test-helpers";
```
and even publish it if we wanted to.

packages/wonder-blocks-modal/components/modal-launcher.js
somewhatabstract: I don't think so. From the errors I was seeting, `keyCode` is becoming obsolete in favour of `which`. In the end, I just copied what the modal did.

packages/wonder-blocks-modal/components/modal-launcher.js
somewhatabstract: Not with the changes that make tooltips one at a time. And yes, I can add a comment.

packages/wonder-blocks-modal/components/modal-launcher.test.js
somewhatabstract: This is just a test file, so I wasn't sure. However, note that `keyCode` is obsolete (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode).

somewhatabstract: Looks like we should be using `code`. I'll make changes accordingly so that we support things properly.

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
somewhatabstract: This is changing in an upcoming PR (where we add all the delays and things), so I'd wait until after that.

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
somewhatabstract: Sure.

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.test.js
somewhatabstract: Yeah, I thought the same thing about having a wonder blocks testing package.

There's a lengthy issue on enzyme about why they don't have an `unmountAll`. It has to do with them not knowing what test runner is being used and stuff. I think they could still have the helper and then leave it to us to call it, but they don't.

packages/wonder-blocks-modal/components/modal-launcher.js
somewhatabstract: Changed back to use `e.key`. `which` and `keyCode` are obsolete and/or non-standard, and their replacement `code` is a string. Pretty sure key is what we want.

somewhatabstract: I've added a comment accordingly.

PR #190, opened 2018-07-06
Tooltip behavior: one at a time and delays

somewhatabstract:
This PR updates the `TooltipAnchor` to enforce tooltip behavior:
1. Only one tooltip visible at a time
2. Delay first tooltip appearance by 100ms
3. Delay last tooltip disappearance by 75ms
4. Switch tooltips instantly if going from one to another

### DIFF
--- a/packages/wonder-blocks-tooltip/components/tooltip-anchor.js
+++ b/packages/wonder-blocks-tooltip/components/tooltip-anchor.js
@@ -90,7 +90,7 @@ export default class TooltipAnchor extends React.Component<Props, State>
              * TODO(somewhatabstract): Work out how to allow pointer to go over
              * the tooltip content to keep it active. This likely requires
              * pointer events but that would break the obscurement checks we do.
-             * So, careful consideration required. See
+             * So, careful consideration required. See WB-302.
              */
             anchorNode.addEventListener("focusin", this._handleFocusIn);
             anchorNode.addEventListener("focusout", this._handleFocusOut);
@@ -112,7 +112,9 @@ export default class TooltipAnchor extends React.Component<Props, State>
     }
 
     componentWillUnmount() {
-        this._unsubscribeFromTracker && this._unsubscribeFromTracker();
+        if (this._unsubscribeFromTracker) {
+            this._unsubscribeFromTracker();
+        }
         this._clearPendingAction();
 
         const anchorNode = this._anchorNode;
--- a/packages/wonder-blocks-tooltip/components/tooltip-anchor.js
+++ b/packages/wonder-blocks-tooltip/components/tooltip-anchor.js
@@ -7,6 +7,13 @@ import * as React from "react";
 import * as ReactDOM from "react-dom";
 
 import TooltipPortalMounter from "./tooltip-portal-mounter.js";
+import ActiveTracker from "../util/active-tracker.js";
+import {
+    TooltipAppearanceDelay,
+    TooltipDisappearanceDelay,
+} from "../util/constants.js";
+
+import type {IActiveTrackerSubscriber} from "../util/active-tracker.js";
 
 type Props = {|
     /**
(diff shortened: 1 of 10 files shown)


jeresig:

Built with commit 330286b4efba5dd39ff99a500aea9ccffd61b2bb

https://deploy-preview-190--wonder-blocks.netlify.com

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -44,11 +51,17 @@ type State = {|
     active: boolean,
 |};
 
-export default class TooltipAnchor extends React.Component<Props, State> {
+const TRACKER = new ActiveTracker();
+
+export default class TooltipAnchor extends React.Component<Props, State>
+    implements IActiveTrackerSubscriber {
```
jeresig: Woah - an implements!

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -69,15 +82,15 @@ export default class TooltipAnchor extends React.Component<Props, State> {
...
         this._updateFocusivity();
         if (anchorNode) {
-            // TODO(somewhatabstract): Detect ESC key to deactivate
             /**
              * TODO(somewhatabstract): Work out how to allow pointer to go over
              * the tooltip content to keep it active. This likely requires
              * pointer events but that would break the obscurement checks we do.
-             * So, careful consideration required.
+             * So, careful consideration required. See
              */
```
jeresig: See...?

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -99,6 +112,9 @@ export default class TooltipAnchor extends React.Component<Props, State> {
     }
 
     componentWillUnmount() {
+        this._unsubscribeFromTracker && this._unsubscribeFromTracker();
```
jeresig: 🤷‍♂️ I'd prefer, makes the intent a little bit clearer:

```
if (this._unsubscribeFromTracker) {
    this._unsubscribeFromTracker();
}
```

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
```
jeresig: lol - we should disable this for test files, if we can?

packages/wonder-blocks-tooltip/util/active-tracker.js
```diff
@@ -0,0 +1,60 @@
+// @flow
```
jeresig: 🤷‍♂️ Can we have a comment explaining what this tracker does a bit better and when we should use it?

Also: is this something that we should move out of Tooltip, would we want to use this in other components as well?

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
somewhatabstract: Good catch! I raised an issue for this and need to add in the issue number. :grin:

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
somewhatabstract: Yes, however, the old flow complained about that which means capturing the variable locally first and then doing this (not a big deal).. I'll see if it complains now we've updated it and edit accordingly

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
somewhatabstract: Agreed

packages/wonder-blocks-tooltip/util/active-tracker.js
somewhatabstract: Re: moving out of tooltip package; I wondered about that and I don't know. It wasn't super clear if anything else would need this.

I will add some more commentary though.

PR #191, opened 2018-07-06
Reset a clickable component's state when it is set to disabled

wishdasher:
A clickable component whose `disabled` prop is set to `false` by its parent still retains its existing `hovered`, `focused`, and `pressed` states. Because that causes the handlers to be disabled, the state of the component, when it is re-enabled, is no longer accurate.

This change resets the state of the component when it is disabled.


wishdasher:
For an example of the current behavior, go to https://wonder-blocks.netlify.com/#multiselectmenu and open the second `MultiSelectMenu` (the one with the shortcuts). Click "Select all" and then uncheck "Red". Notice that the "Select all" button still retains its `hovered` state and styling.

jeresig:

Built with commit 8b61539bebac305065be661b4cb2ec550ae5ac22

https://deploy-preview-191--wonder-blocks.netlify.com

PR #192, opened 2018-07-06
Implement a11y solution

somewhatabstract:
This PR adds our main accessibility feature to the tooltip. It is stacked on #190 so that the next PR can add testing for everything in the `Tooltip` component and make use of the testing utilities introduced in #188.

Here we add an identifier to the `TooltipBubble` and then force the anchor point to have `aria-describedby` that points to that identifier. This means when the anchor gets focused, assistive tech will give the tooltip content as additional information.

This also adds `aria-hidden` to `Spring` and `Strut`, as well as provide an escape hatch on the `Tooltip` for situations where a more complex or different accessibility solution is required.

NOTE: Additional test coverage to be added in separate PR.

### DIFF
--- a/packages/wonder-blocks-layout/components/spring.js
+++ b/packages/wonder-blocks-layout/components/spring.js
@@ -11,7 +11,7 @@ import {View} from "@khanacademy/wonder-blocks-core";
 
 export default class Spring extends React.Component<{}> {
     render() {
-        return <View style={styles.grow} />;
+        return <View aria-hidden="true" style={styles.grow} />;
     }
 }
 
--- a/packages/wonder-blocks-tooltip/generated-snapshot.test.js
+++ b/packages/wonder-blocks-tooltip/generated-snapshot.test.js
@@ -18,7 +18,10 @@ describe("wonder-blocks-tooltip", () => {
         const React = require("react");
 
         const example = (
-            <Tooltip content={"I'm on the right!"} placement="right">
+            <Tooltip
+                content={"This is a text tooltip on the right"}
+                placement="right"
+            >
                 Some text
             </Tooltip>
         );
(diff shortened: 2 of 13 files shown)


jeresig:

Built with commit 2dd4ee584c30bf00bf7c99422fab1a66ff64c0f8

https://deploy-preview-192--wonder-blocks.netlify.com

jeresig:

Built with commit d0e1c16a5f9e799b687dd950651b406b0e7c2e17

https://deploy-preview-192--wonder-blocks.netlify.com

packages/wonder-blocks-layout/components/spring.js
```diff
@@ -11,7 +11,7 @@ import {View} from "@khanacademy/wonder-blocks-core";
 
 export default class Spring extends React.Component<{}> {
     render() {
-        return <View style={styles.grow} />;
+        return <View aria-hidden="true" style={styles.grow} />;
```
jeresig: Huh... does having these be hidden have any impact if there is nothing inside of them and they don't receive focus?

packages/wonder-blocks-tooltip/generated-snapshot.test.js
```diff
@@ -18,7 +18,10 @@ describe("wonder-blocks-tooltip", () => {
         const React = require("react");
 
         const example = (
-            <Tooltip content={"I'm on the right!"} placement="right">
+            <Tooltip
+                content={"This is a text tooltip on the right"}
```
jeresig: 🤷‍♂️ I notice that in a few places you have extra `{...}` around strings, which can be removed!

packages/wonder-blocks-layout/components/spring.js
somewhatabstract: Not necessarily, but since we can't test every screen reader or assistive tech, it seems prudent to be as explicit as we can.

The biggest effect this has is perf. The browser will stop searching at nodes marked as hidden. If there were more empty children here, it would go through all of them. Marking as hidden short-circuits that effort.

PR #193, opened 2018-07-07
Replace menu checks with Icon, fix flow typing

wishdasher:


wishdasher:
I think the checks in the checkboxes ☑️are slightly too big, but the Icon is set at a size of small = 16px. Also, the design uses a slightly smaller checkbox for the multi-select checkboxes vs. the regular checks ✔️in the single-select.

kevinbarabash:
If design wants slightly different icons from what's in `wonder-blocks-icon` I think that's okay as long as they're aware that it's different and it's intentional.  If that is the case we can still use the `Icon` component at least.

packages/wonder-blocks-dropdown/components/select-item.js
```diff
@@ -72,69 +68,46 @@ type CheckProps = {
...
+                styles.check,
                 styles.checkbox,
-                !selected && styles.borderedCheckbox,
+                !selected &&
+                    !(pressed || hovered || focused) &&
+                    styles.borderedCheckbox,
                 !selected &&
                     (pressed || hovered || focused) &&
                     styles.invertBackground,
+                {backgroundColor: bgColor},
```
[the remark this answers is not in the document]
wishdasher: Nice catch!

wishdasher:
The smaller check marks are intentional. I will open a separate PR to add the asset once I get one.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -9,12 +9,19 @@ import Color, {fade} from "@khanacademy/wonder-blocks-color";
 import Spacing from "@khanacademy/wonder-blocks-spacing";
 import {View} from "@khanacademy/wonder-blocks-core";
 
+import ActionItem from "./action-item.js";
```
jeresig: :shrug	I think you can actually make these `import typeof ...` and avoid having to import the whole module! (and then remove the typeofs below)

packages/wonder-blocks-dropdown/components/select-box.js
```diff
@@ -127,9 +107,11 @@ export default class SelectBox extends React.Component<SelectBoxProps> {
                                 {children}
                             </LabelMedium>
                             <View style={[styles.spacing]} />
```
jeresig: Hmm - just realizing that we can use the new Strut component for stuff like this! Might be worth it for us to do a pass back through at some point (don't need to do it now).

packages/wonder-blocks-dropdown/components/select-item.js
```diff
@@ -72,69 +68,46 @@ type CheckProps = {
...
-                    <path
-                        fill={pressed || hovered || focused ? white : offBlack}
-                        d={checkIcon}
-                    />
-                )}
-            </svg>
+        <View style={[styles.check, !selected && styles.hide]}>
+            <Icon
+                icon={icons.check}
+                size={"small"}
```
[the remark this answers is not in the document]
jeresig: 🤷‍♂️ Whoops - still has braces!

packages/wonder-blocks-dropdown/components/select-item.js
```diff
@@ -66,75 +71,73 @@ type SelectProps = {|
...
+                color={pressed || hovered || focused ? white : offBlack}
+            />
         </View>
     );
 };
 
-// TODO: replace with Choice component once it's implemented
-const Checkbox = (props: CheckProps) => {
-    const {selected, pressed, hovered, focused} = props;
+// This is a smaller check specifically for use in checkboxes
```
jeresig: 🛑 I'd expand this comment to explicitly warn devs to not copy it and to check with a designer first before using.

packages/wonder-blocks-dropdown/components/select-item.js
```diff
@@ -66,75 +71,73 @@ type SelectProps = {|
...
-                    viewBox="-2 -1 14 14"
-                    aria-hidden="true"
-                    style={{backgroundColor: bgColor, borderRadius: 3}}
-                >
-                    <path
-                        fill={
-                            hovered || focused
+                <Icon
+                    icon={checkboxCheck}
+                    size={"small"}
```
jeresig: 🤷‍♂️ Braces!

packages/wonder-blocks-dropdown/package.json
```diff
@@ -14,6 +14,7 @@
     "@khanacademy/wonder-blocks-button": "0.0.3",
     "@khanacademy/wonder-blocks-color": "^1.0.0",
     "@khanacademy/wonder-blocks-core": "^1.0.0",
+    "@khanacademy/wonder-blocks-icon": "0.0.1",
```
jeresig: (I think this will need to get updated - and we should un-pin it so that it updates in the future!)

packages/wonder-blocks-button/components/button.js
```diff
@@ -11,6 +11,7 @@ export type SharedProps = {|
      */
     children: string,
 
+    // TODO: incorporate
```
jeresig: 🤷‍♂️ It may also make sense to just comment out this prop since it doesn't do anything anyway!

packages/wonder-blocks-dropdown/components/dropdown-core.js
wishdasher: Woahh I didn't know that. Will definitely use this instead! (It took me long enough to figure out how to do `typeof Blah` in the first place.

PR #195, opened 2018-07-07
add flow types for react-router-dom

kevinbarabash:
This improves flow coverage in clickable-behaviour.js.  We still need to refactor the render method so that flow can better understand which props are going to which components that are being rendered.

### DIFF
--- a/packages/wonder-blocks-dropdown/components/action-item.js
+++ b/packages/wonder-blocks-dropdown/components/action-item.js
@@ -72,6 +72,7 @@ type ActionProps = {|
 
 const StyledAnchor = addStyle("a");
 const StyledButton = addStyle("button");
+// $FlowFixMe: need to ensure we're passing 'to' to Link in a way that flow understands
 const StyledLink = addStyle(Link);
 
 export default class ActionItem extends React.Component<ActionProps> {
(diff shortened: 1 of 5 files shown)


jeresig:

Built with commit 68bc78fb1478d95587f870ff026c1d77b7780c53

https://deploy-preview-195--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/action-item.js
```diff
@@ -72,6 +72,7 @@ type ActionProps = {
 
 const StyledAnchor = addStyle("a");
 const StyledButton = addStyle("button");
+// $FlowFixMe: need to ensure we're passing 'to' to Link in a way that flow understands
```
jeresig: Hmm... this is a bit concerning. How do we do this?

kevinbarabash: I think we'll end up having to pass props directly to `StyledLink`, `StyledButton`, or `StyleAnchor` instead of passing them to `Tag` which could be anyone one of those components.

kevinbarabash:
Note to self: don't include `$FlowFixMe` in command line commit messages.

PR #197, opened 2018-07-08
Add tests for tooltip.js and tidy up testing

somewhatabstract:
Here we add some final testing tweaks to the `Tooltip` component and related things, including validation of the a11y work in the previous PR on which this is stacked, #192.

In addition to the `Tooltip` coverage, this set of changes migrates all the tooltip fixtures to use the special version of `mount` that gives us an `unmountAll` method. This is important since tests share a DOM, and without unmounting all mounted things, side-effects can bleed across test boundaries (we should update all our testing to use this model).

Finally, we also rework many tests to use `async`/`await` for a much tidier set of tests.

### DIFF
--- a/packages/wonder-blocks-tooltip/components/tooltip.test.js
+++ b/packages/wonder-blocks-tooltip/components/tooltip.test.js
@@ -1,4 +1,397 @@
-//@flow
+// @flow
+import * as React from "react";
+import * as ReactDOM from "react-dom";
+
+import {mount, unmountAll} from "../../../utils/testing/mount.js";
+
+import {View} from "@khanacademy/wonder-blocks-core";
+import {Body, HeadingSmall} from "@khanacademy/wonder-blocks-typography";
+
+import Tooltip from "./tooltip.js";
+import TooltipBubble from "./tooltip-bubble.js";
+import TooltipContent from "./tooltip-content.js";
+
+const mockIDENTIFIER = "mock-identifier";
+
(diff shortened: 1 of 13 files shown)


kevinbarabash:
It looks like we're be able to merge both this and #183.

somewhatabstract:
@kevinbarabash I'd hold off on merging #183 since my tooltip PRs rewrite a bunch of tests, including making them async/await. I think otherwise, there'll be lots of merge conflicts to deal with.

packages/wonder-blocks-tooltip/components/tooltip.test.js
```diff
@@ -1,4 +1,397 @@
...
+import {View} from "@khanacademy/wonder-blocks-core";
+import {Body, HeadingSmall} from "@khanacademy/wonder-blocks-typography";
+
+import Tooltip from "./tooltip.js";
+import TooltipBubble from "./tooltip-bubble.js";
+import TooltipContent from "./tooltip-content.js";
+
+const mockIDENTIFIER = "mock-identifier";
+
+jest.mock("./tooltip-bubble.js");
```
jeresig: 🤷‍♂️ Does this have any impact beyond this file? Do we want to localize this mock to just being inside this test?

somewhatabstract: It is only this file that it impacts.

PR #198, opened 2018-07-08
update flow types to be exact object types

kevinbarabash:
Now that we've upgrade to flow 0.75 we can start using flow's exact object types everywhere.  This PR converts almost all of our types to exact object types.  Doing this revealed a few places where we had extra props that weren't necessary.  It also adds an eslint rule to remind people that we should be using exact object types go forward.  I'm also planning to write up an ADR to adopt this practice in webapp as well.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,8 @@
 {
     "extends": [
         "./eslint/eslintrc"
-    ]
+    ],
+    "rules": {
+        "flowtype/require-exact-type": [2, "always"]
+    }
 }
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "scripts": {
     "clean": "rm -rf packages/wonder-blocks-*/dist && rm -rf packages/wonder-blocks-*/node_modules",
     "lint": "eslint --config ./eslint/eslintrc packages",
+    "lint:watch": "esw --watch --config ./eslint/eslintrc packages",
     "build:all": "webpack && node utils/package-flow-types.js",
     "watch:all": "webpack -w",
     "flow": "flow",
@@ -47,6 +48,7 @@
     "eslint-plugin-prettier": "2.3.1",
     "eslint-plugin-react": "7.7.0",
     "eslint-plugin-react-native-animation-linter": "0.1.1",
+    "eslint-watch": "^4.0.1",
     "flow-bin": "0.75.0",
     "glob": "^7.1.2",
     "jest": "^23.3.0",
--- a/packages/wonder-blocks-button/components/button-core.js
+++ b/packages/wonder-blocks-button/components/button-core.js
@@ -13,12 +13,13 @@ import {addStyle} from "@khanacademy/wonder-blocks-core";
 import type {ClickableHandlers} from "@khanacademy/wonder-blocks-core";
 import type {SharedProps} from "./button.js";
 
-type Props = SharedProps &
-    ClickableHandlers & {
-        hovered: boolean,
-        focused: boolean,
-        pressed: boolean,
-    };
+type Props = {|
+    ...SharedProps,
+    ...ClickableHandlers,
+    hovered: boolean,
+    focused: boolean,
+    pressed: boolean,
+|};
 
 const StyledAnchor = addStyle("a");
 const StyledButton = addStyle("button");
@@ -84,8 +85,10 @@ export default class ButtonCore extends React.Component<Props> {
 
         if (!disabled && href) {
             if (clientNav) {
+                // $FlowFixMe
                 props.to = href;
             } else {
+                // $FlowFixMe
                 props.href = href;
             }
         }
--- a/packages/wonder-blocks-core/components/media-layout.js
+++ b/packages/wonder-blocks-core/components/media-layout.js
@@ -7,7 +7,7 @@ import {mediaContextTypes} from "../util/util.js";
 
 import type {MediaSize, MediaSpec} from "../util/types.js";
 
-type Props = {
+type Props = {|
     /**
      * Force the MediaLayout to be a particular size (ignoring the actual
      * dimensions of the viewport).
@@ -29,7 +29,16 @@ type Props = {
 
     /** The contents of the page */
     children: React.Node,
-};
+
+    /**
+     * Style the layout container
+     */
+    style?: any,
+
+    // TODO(kevinb): provide a way to "mixin" allowed roles and aria props
+    role?: "dialog",
+    "aria-labelledby"?: string,
+|};
 
 /**
  * A MediaLayout wraps all parts of a page and tracks the browser viewport, toggling
--- a/packages/wonder-blocks-core/util/types.js
+++ b/packages/wonder-blocks-core/util/types.js
@@ -91,7 +91,8 @@ type IdRef = string;
 type IdRefList = IdRef | Array<IdRef>;
 
 // Source: https://www.w3.org/WAI/PF/aria-1.1/states_and_properties
-// TODO(kevinb): figure out how to type ID ref lists
+// TODO(kevinb): convert to disjoint union of exact object types keyed on role
+// eslint-disable-next-line flowtype/require-exact-type
 export type AriaProps = {
     role?: roles,
     "aria-activedescendant"?: IdRef,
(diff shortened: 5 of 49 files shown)


jeresig:

Built with commit a509d4503030c10083bec09981cd9acf477b85fe

https://deploy-preview-198--wonder-blocks.netlify.com

packages/wonder-blocks-link/__snapshots__/custom-snapshot.test.js.snap
```diff
@@ -136,45 +99,10 @@ exports[`LinkCore kind:primary href:# light:false pressed 1`] = `
 </a>
 `;
 
-exports[`LinkCore kind:primary href:# light:true disabled 1`] = `
```
kevinbarabash: There's no such thing as a disabled link.

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -13,12 +13,13 @@ import {addStyle} from "@khanacademy/wonder-blocks-core";
...
-type Props = SharedProps &
-    ClickableHandlers & {
-        hovered: boolean,
-        focused: boolean,
-        pressed: boolean,
-    };
+type Props = {|
+    ...SharedProps,
+    ...ClickableHandlers,
+    hovered: boolean,
```
kevinbarabash: This syntax looks cleaner to me.

package.json
```diff
@@ -47,6 +48,7 @@
     "eslint-plugin-prettier": "2.3.1",
     "eslint-plugin-react": "7.7.0",
     "eslint-plugin-react-native-animation-linter": "0.1.1",
+    "eslint-watch": "^4.0.1",
```
kevinbarabash: This was super helpful when making this change b/c vscode only shows lint issues in the opened files where as this will show lint errors in all files.

.eslintrc
```diff
@@ -1,5 +1,8 @@
 {
     "extends": [
         "./eslint/eslintrc"
-    ]
+    ],
+    "rules": {
+        "flowtype/require-exact-type": [2, "always"]
+    }
```
kevinbarabash: I think eventually we'll want to back port some (all?) of these new rules to khan-linter.  There's quite a lot of process in doing so so unfortunately I don't see that happening any time soon. 😞

packages/wonder-blocks-core/components/media-layout.js
```diff
@@ -29,7 +29,16 @@ type Props = {
...
-};
+
+    /**
+     * Style the layout container
+     */
+    style?: any,
+
+    // TODO(kevinb): provide a way to "mixin" allowed roles and aria props
+    role?: "dialog",
+    "aria-labelledby"?: string,
```
kevinbarabash: We shouldn't be adding aria props like these, but this will have to do for now.

packages/wonder-blocks-dropdown/components/multi-select-menu.js
```diff
@@ -69,14 +69,14 @@ type Props = {
      * Optional styling to add.
      */
     style?: any,
-};
+|};
 
-type State = {
+type State = {|
```
kevinbarabash: Originally I was thinking of only making `Props` exact objects, but the eslint rule doesn't have a way to do that and there doesn't seem to be a downside from making `State` exact as well.

packages/wonder-blocks-icon-button/custom-snapshot.test.js
```diff
@@ -48,15 +48,12 @@ describe("IconButtonCore", () => {
...
                                         aria-label="search"
                                         kind={kind}
-                                        size={size}
                                         color={color}
                                         light={light}
                                         {...stateProps}
                                         {...defaultHandlers}
                                         tabIndex={stateProps.disabled ? -1 : 0}
-                                    >
-                                        Click me
```
kevinbarabash: `IconButton` doesn't support children.

packages/wonder-blocks-link/__snapshots__/custom-snapshot.test.js.snap
somewhatabstract: It's called a "dimmed" link in assistive tech worlds and is created by giving it a tabIndex="-1" and adding `aria-disabled` as `true`. You can see an example here in webapp (https://github.com/Khan/webapp/blob/ee1ad038b3467ea09822745ae43fce6c0cfee88b/javascript/test-prep-package/practice-plan/components/checkpoint-progress.jsx#L230-L242)

packages/wonder-blocks-dropdown/components/multi-select-menu.js
somewhatabstract: I like making state exact. It really helps remind me to remove things that are no longer used.

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -84,8 +85,10 @@ export default class ButtonCore extends React.Component<Props> {
 
         if (!disabled && href) {
             if (clientNav) {
+                // $FlowFixMe
```
wishdasher: Huh, now these get caught in `button-core` as well. I spent a short while trying to fix it in `action-item` but wasn't able to.

packages/wonder-blocks-button/components/button-core.js
jeresig: Yeah, this is great!

packages/wonder-blocks-core/components/media-layout.js
jeresig: The moment we have support for React 16 we can rewrite MediaLayout to just be a passthrough component so this won't be needed anymore!

packages/wonder-blocks-core/util/types.js
```diff
@@ -140,6 +141,8 @@ export type Props = AriaProps & {
 
 export type MediaSize = "small" | "medium" | "large";
 
+// TODO(kevinb): dedupe with wonder-blocks-grid
```
jeresig: 🤷‍♂️ how do you mean? I think it only lives here, right?

packages/wonder-blocks-core/util/types.js
```diff
@@ -91,7 +91,8 @@ type IdRef = string;
 type IdRefList = IdRef | Array<IdRef>;
 
 // Source: https://www.w3.org/WAI/PF/aria-1.1/states_and_properties
-// TODO(kevinb): figure out how to type ID ref lists
+// TODO(kevinb): convert to disjoint union of exact object types keyed on role
```
jeresig: I'm not sure I follow this - meaning that ensuring that each role only allows certain sets of aria props?

packages/wonder-blocks-grid/util/types.js
```diff
@@ -2,6 +2,8 @@
 
 export type MediaSize = "small" | "medium" | "large";
 
+// TODO(kevinb): dedupe with wonder-blocks-core
```
jeresig: Ohhh - it is still here! d'oh...

packages/wonder-blocks-link/custom-snapshot.test.js
```diff
@@ -46,7 +40,7 @@ describe("LinkCore", () => {
                                     light={light}
                                     {...stateProps}
                                     {...defaultHandlers}
-                                    tabIndex={stateProps.disabled ? -1 : 0}
```
jeresig: 🤷‍♂️ Hmm - this changed?

kevinbarabash:
@wishdasher `{| |}` is semantically different.  It's the syntax for exact object types.  It does not allow extra properties on an object that's typed as an exact object.  See https://flow.org/en/docs/types/objects/#toc-exact-object-types

packages/wonder-blocks-link/custom-snapshot.test.js
kevinbarabash: This is b/c our Link component doesn't actually support a disabled state. The reason for keeping tabIndex is a bit roundabout but it boils down to the ClickableHandlers type including tabIndex as a mandatory prop. Maybe ClickableHandlers isn't the right place for tabIndex. Since tabIndex is either 0 (disabled=false) or -1 (disabled=true) it can be derived from disabled and maybe each component that uses ClickableBehavior should be responsible for doing that mapping and assigning tabIndex to whatever components it renders.

PR #199, opened 2018-07-10
Put dropdown menu in a portal

wishdasher:
Main changes:
- change `onClose` prop to `onOpenChanged`
- use React 16 portal to mount the menu under `body`
- fix event handling bugs that arose from portaling the menu
- use positioning methods to position portalled menu correctly (currently not using `react-popper`)
- update tests to better practices


packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -176,6 +179,7 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
             <ActionMenuOpener
                 disabled={disabled}
                 onClick={() => this.toggleMenu()}
+                ref={node => {this.openerNode = ReactDOM.findDOMNode(node)}}
```
wishdasher: We are using the ref on the opener to determine its position later with getBoundingClientRect.

packages/wonder-blocks-dropdown/components/multi-select-menu.test.js
```diff
@@ -72,48 +72,41 @@ describe("MultiSelectMenu", () => {
...
+        // selectAll.simulate("click");
+        // expect(allChanges.pop().length).toEqual(3);
+        // // Select none of the items
+        // const selectNone = menu.find(ActionItem).at(1);
+        // selectNone.simulate("mousedown");
+        // selectNone.simulate("mouseup");
+        // selectNone.simulate("click");
+        // expect(allChanges.pop().length).toEqual(0);
+        // // Menu should still be open
+        // expect(menu.state("open")).toEqual(true);
```
wishdasher: This test case now gives me trouble because I cannot grab the portaled React elements to test and verify callbacks on. Dispatching events on the base `HTMLElement` does not invoke the callbacks on the corresponding components.

packages/wonder-blocks-dropdown/components/single-select-menu.js
```diff
@@ -1,7 +1,8 @@
 // @flow
-// A menu that consists of action items
+// A menu that consists of items to be selected, single-choice
```
wishdasher: Incorrect relic of a comment

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -61,78 +77,171 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+        }
+
+        // If the menu just changed from closed to open, and if we haven't
+        // found the correct menuWidth yet
+        if (!prevProps.open && this.props.open && !this.menuWidth) {
+            // There is a timeout because if we call getBoundingClientRect
+            // right away, it returns the value of the document width instead
+            // of the width of the just-mounted component, the menu.
+            setTimeout(() => {
+                // $FlowFixMe
```
jeresig: 🤷‍♂️ What's this hiding? I think it might work if you do: `if (this.menuRef) { ...`

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -61,78 +77,171 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+            this.updateEventListeners();
+        }
+
+        // If the menu just changed from closed to open, and if we haven't
+        // found the correct menuWidth yet
+        if (!prevProps.open && this.props.open && !this.menuWidth) {
+            // There is a timeout because if we call getBoundingClientRect
+            // right away, it returns the value of the document width instead
+            // of the width of the just-mounted component, the menu.
+            setTimeout(() => {
```
jeresig: 🤷‍♂️ I think doing `requestAnimationFrame` might be the preferred method of running something on the next render frame (and I think it's what Aphrodite does, too).

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -61,78 +77,171 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+
+        this.updatePosition();
+        if (open) {
+            return (
+                <View
+                    onClick={(event) => {
+                        // Stop propagation to prevent the mouse click listener on the
+                        // document from closing the menu.
+                        event.stopPropagation();
+                        event.nativeEvent.stopImmediatePropagation();
```
jeresig: 🤷‍♂️ Are both of these required? It seems like only one of them would be needed, I would've thought.

packages/wonder-blocks-dropdown/components/dropdown-portal-mounter.js
```diff
@@ -0,0 +1,158 @@
+// @flow
```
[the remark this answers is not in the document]
jeresig: Hmm, yeah. I think you can save it for a follow-up PR but I think this is a good candidate for a new package (especially since it depends upon wonder-blocks-modal). Maybe something like `wonder-blocks-portal-mounter`?

packages/wonder-blocks-dropdown/components/dropdown-portal-mounter.js
```diff
@@ -0,0 +1,158 @@
...
+                // to dismantle it and mount a new one.
+                //
+                // Not certain this is necessary, or if the later render call
+                // handles this, but it seems appropriate and is certainly
+                // clearer to do it here.
+                // NOTE(sophie): This is the one line I changed from the tooltip
+                // portal. It was causing the menu to be re-mounted with each
+                // change, which reset clickable behavior state for each menu
+                // item upon selection, which was causing strange visual bugs.
+                // ReactDOM.unmountComponentAtNode(this._destination);
```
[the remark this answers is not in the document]
jeresig: @somewhatabstract Thoughts on this and moving the code into a shared component?

packages/wonder-blocks-dropdown/components/multi-select-menu.test.js
jeresig: Hmm... @somewhatabstract thoughts on how to get these to work?

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -61,78 +77,171 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+                left: rect.left + window.pageXOffset,
+                right: rect.right,
+                width: rect.width,
+            };
         }
     };
 
+    findTranslatedPosition() {
+        if (this.openerBounds) {
+            const top = this.openerBounds.top;
```
jeresig: Can do: `const {top, left, width} = this.openerBounds;`

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -61,78 +77,171 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+            const left = this.openerBounds.left;
+            const width = this.openerBounds.width;
+            if (this.props.alignment === "left") {
+                return {
+                    transform: `translate3d(${left}px, ${top}px, 0)`,
```
jeresig: @somewhatabstract Should we be using Popper.js to handle the positioning on something like this?

packages/wonder-blocks-dropdown/components/dropdown-core.js
wishdasher: `this.menuRef` is typed as a `Node` because that's what `ReactDOM.findDOMNode` returns. Maybe I can check for the `Element` type here?

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -234,9 +234,10 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
                 alignment={alignment}
                 items={menuItems}
                 light={false}
-                onClose={() => this.handleClose()}
+                onDropdownChanged={(open) => this.handleDropdownChanged(open)}
```
kevinbarabash: `onDropdownChanged` feels a little too general as the contents of the what are in the dropdown don't change.  Maybe change this to `onOpenChanged` since that's exactly what's changing.

packages/wonder-blocks-dropdown/components/action-menu.test.js
```diff
@@ -11,36 +11,36 @@ const keyCodes = {
...
+                        onClick: () => onToggle(),
+                        value: "toggle",
+                    },
+                ]}
+                menuText={"Action menu!"}
+                onChange={(selectedValues) => onChange()}
+                selectedValues={[]}
+            />,
+        );
+        const opener = menu.find("ActionMenuOpener");
```
kevinbarabash: Good call moving this into this `it`.  I didn't realize that you had this code in the `describe`.  You should almost never have code inside a `describe` that isn't in an `it`.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -64,78 +79,167 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
-            this.props.onClose();
+            this.props.onDropdownChanged(false);
+        }
+    };
+
+    updatePosition = () => {
+        const {openerNode} = this.props;
+        if (openerNode) {
+            //$FlowFixMe typing is weird for the return value of ref callback
+            const rect = openerNode.getBoundingClientRect();
```
kevinbarabash: I think maybe it has to be cast to `Element` since `getBoundClientRect` is defined on that.  https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -53,7 +60,15 @@ type DropdownCoreProps = {|
...
-    node: ?Node;
+    openerBounds: ?{
+        top: number,
+        left: number,
+        width: number,
+    };
+    // We need the menu width to tell it how to position itself for
+    // right-aligned menus
+    menuWidth: ?number;
+    menuRef: ?Node;
```
kevinbarabash: This should probably be `menuRef: ?Element`.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -64,78 +79,167 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+        document.addEventListener("click", this._handleInteract);
+        document.addEventListener("touch", this._handleInteract);
+        // This is still keyup because we are listening for the escape key,
+        // which invokes events keydown and keyup and NOT keypress.
+        document.addEventListener("keyup", this._handleKeyup);
+    }
+
+    removeEventListeners() {
+        document.removeEventListener("click", this._handleInteract);
+        document.removeEventListener("touch", this._handleInteract);
```
kevinbarabash: Is there actually a `"touch"` event?  You might have to use `"touchend"` here.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -64,78 +79,167 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+            this.props.onDropdownChanged(false);
         }
     };
 
     _handleKeyup = (event: KeyboardEvent) => {
         if (this.props.open && event.key === "Escape") {
             event.preventDefault();
             event.stopImmediatePropagation();
-            this.props.onClose();
+            this.props.onDropdownChanged(false);
```
kevinbarabash: Where is the opening via keyboard handled?

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -64,78 +79,167 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+    updatePosition = () => {
+        const {openerNode} = this.props;
+        if (openerNode) {
+            //$FlowFixMe typing is weird for the return value of ref callback
+            const rect = openerNode.getBoundingClientRect();
+            this.openerBounds = {
+                top: rect.top + window.pageYOffset,
+                left: rect.left + window.pageXOffset,
+                width: rect.width,
+            };
```
kevinbarabash: `openerBounds` is only used in a single place, we could just return it and use the value there instead of storing it on `this`.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -64,78 +79,167 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
         }
     };
 
+    findTranslatedPosition() {
+        if (this.openerBounds) {
+            const {top, left, width} = this.openerBounds;
+            if (this.props.alignment === "left") {
+                return {
+                    transform: `translate3d(${left}px, ${top}px, 0)`,
```
kevinbarabash: I'm curious about why you chose to use 3D transforms for this.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -64,78 +79,167 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+        const {items, light, open, style} = this.props;
+
+        this.updatePosition();
+        if (open) {
+            return (
+                <View
+                    onClick={(event) => {
+                        // Stop propagation to prevent the mouse click listener on the
+                        // document from closing the menu.
+                        event.nativeEvent.stopImmediatePropagation();
```
kevinbarabash: Which handler would otherwise close the menu if we didn't have this call?

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -64,78 +79,167 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+                    }}
+                    ref={(node) => (this.menuRef = ReactDOM.findDOMNode(node))}
+                    style={[
+                        styles.dropdown,
+                        light && styles.light,
+                        (!open || !this.menuWidth) && styles.hide,
```
kevinbarabash: I find `!open || !this.menuWidth` hard to reason about.  Maybe do `!(open && this.menuWidth)` instead.

packages/wonder-blocks-dropdown/components/dropdown-portal-mounter.js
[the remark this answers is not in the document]
kevinbarabash: @jeresig are you worried too much stuff ending up in `wonder-blocks-core`?

packages/wonder-blocks-dropdown/components/multi-select-menu.test.js
kevinbarabash: A few thoughts:
- code that's shared between tests should appear in a `beforeEach` not inside of the `describe`.  The reason being that the code inside `describe` is only run once and you'll want the setup code to run once per `it`.
- you're doing a full `mount` which is good.  You'll have to grab a ref to the dropdown being rendered in the portal.  See https://github.com/Khan/wonder-blocks/blob/cbbd25522b105b4a22ae0a8fbb1132fad9d168b5/packages/wonder-blocks-tooltip/components/tooltip.test.js for details.

packages/wonder-blocks-dropdown/components/dropdown-portal-mounter.js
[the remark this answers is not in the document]
kevinbarabash: @wishdasher what kinds of changes were causing the refresh?  By "refresh" do you mean it caused the portal to be destroyed and recreated?

packages/wonder-blocks-dropdown/components/single-select-menu.js
```diff
@@ -148,9 +148,10 @@ export default class SingleSelectMenu extends React.Component<Props, State> {
                 alignment={alignment}
                 items={menuItems}
                 light={light}
-                onClose={() => this.handleClose()}
+                onDropdownChanged={(open) => this.handleDropdownChanged(open)}
```
kevinbarabash: Maybe change this to `onOpenChanged` and `handleOpenChanged`.

packages/wonder-blocks-dropdown/components/single-select-menu.test.js
```diff
@@ -3,7 +3,7 @@ import React from "react";
 import {mount} from "enzyme";
 
 import SelectBox from "./select-box";
-import SelectItem from "./select-item";
+// import SelectItem from "./select-item";
 import SingleSelectMenu from "./single-select-menu.js";
 
 const keyCodes = {
```
kevinbarabash: I can't comment on these lines, but you'll want to move your shared setup into a `beforeEach` block to isolate tests better.

packages/wonder-blocks-dropdown/components/dropdown-portal-mounter.js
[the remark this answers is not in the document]
somewhatabstract: I don't know that we necessarily should share code just because it looks the same. There's a lot to be said for not coupling things if we don't need to, and I'd lean toward that here.

packages/wonder-blocks-dropdown/components/dropdown-portal-mounter.js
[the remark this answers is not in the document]
somewhatabstract: I want to double check the DOM here and make sure we don't ever get weird node duplication.

I think long term, we should use React 16 portals if we can. I also think we should avoid code sharing as much as possible for internals like this. Coupling components on shared code may seem valuable, but that coupling introduces a tech debt and I don't know that we gain much for it with this specific component, especially if we replace much of this with a simpler React 16 approach.

If we were moving to a shared component, I think I'd rather use the third-party react-portal package that handles React 15 and 16, than maintain our own. However, my preference is to not couple components at all with regards to the portal and have each maintain their own usage, unless we see a really compelling reason to maintain the portal component as an API that we export from Wonder Blocks.

packages/wonder-blocks-dropdown/components/dropdown-portal-mounter.js
```diff
@@ -0,0 +1,159 @@
...
+        // We have to render the subtree like this so that everything works as
+        // expected.
+        // See https://github.com/tajo/react-portal/blob/master/src/LegacyPortal.js
+        // console.log("starting render");
+        ReactDOM.unstable_renderSubtreeIntoContainer(
+            this,
+            children,
+            this._destination,
+        );
+        // console.log("ending render");
```
somewhatabstract: Delete?

packages/wonder-blocks-dropdown/components/multi-select-menu.test.js
```diff
@@ -3,8 +3,8 @@ import React from "react";
 import {mount} from "enzyme";
 
 import SelectBox from "./select-box";
-import ActionItem from "./action-item";
-import SelectItem from "./select-item";
+// import ActionItem from "./action-item";
+// import SelectItem from "./select-item";
```
somewhatabstract: Delete?

packages/wonder-blocks-dropdown/components/multi-select-menu.test.js
```diff
@@ -3,8 +3,8 @@ import React from "react";
 import {mount} from "enzyme";
```
somewhatabstract: Don't use `mount` directly from `enzyme`. Use `mount` from our `utils/testing/mount.js` file and have an `afterEach` in the root describe that calls `unmountAll`.

packages/wonder-blocks-dropdown/components/multi-select-menu.test.js
somewhatabstract: Almost never put code in the `describe`. If you want something to run once before all tests use `beforeAll`, or `afterAll` for something to run once after all tests. If you want something to run before or after each test, use `beforeEach` or `afterEach` instead.

somewhatabstract: As @kevinbarabash said, you'll want to use a `ref` prop to grab the dropdown being rendered into the portal as in the example he linked.

packages/wonder-blocks-dropdown/components/single-select-menu.test.js
```diff
@@ -3,7 +3,7 @@ import React from "react";
 import {mount} from "enzyme";
```
somewhatabstract: Don't use `mount` directly. Avoid having to tidy up every mount by using our own `mount` and the handy `unmountAll` call. Otherwise, you run the risk of side effects between test cases because they share a DOM and there's no auto-unmounting between tests with enzyme directly.

packages/wonder-blocks-dropdown/components/dropdown-core.js
wishdasher: The openers use ClickableBehavior's `onClick` callback to toggle the menu open state.

packages/wonder-blocks-dropdown/components/dropdown-core.js
wishdasher: I am using `ReactDOM.findDOMNode` already, but I'm not sure how to cast something into an `Element`

kevinbarabash: https://flow.org/en/docs/types/casting/#toc-type-casting-through-any

packages/wonder-blocks-dropdown/components/dropdown-core.js
wishdasher: It's how `react-popper` positions things. I'm not sure if there's another way to position given where the menu ends up in the DOM tree.

packages/wonder-blocks-dropdown/components/dropdown-core.js
wishdasher: There's a callback on the opener and an event on the `document` to close the menu.

packages/wonder-blocks-dropdown/components/dropdown-portal-mounter.js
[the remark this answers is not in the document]
wishdasher: I'm using React's portal instead, so this will be removed!

packages/wonder-blocks-dropdown/components/multi-select-menu.test.js
wishdasher: Since React portals preserve the React tree, I think now I can change the test case back and it'd work.

packages/wonder-blocks-dropdown/components/action-menu.test.js
```diff
@@ -11,34 +12,40 @@ const keyCodes = {
     space: 32,
 };
 
+let menu;
+const onClick = jest.fn();
+const onToggle = jest.fn();
+const onChange = jest.fn();
+
+beforeAll(() => {
+    unmountAll();
```
somewhatabstract: This call should be in a `beforeEach`.

`beforeAll` runs once only for the describe
`beforeEach` runs once before each test case

packages/wonder-blocks-dropdown/components/action-item.js
```diff
@@ -164,6 +164,7 @@ const styles = StyleSheet.create({
         flexDirection: "row",
         alignItems: "center",
         display: "flex",
+        margin: 0,
```
wishdasher: This would actually be fixed after the PR to reset `button` styles in Safari gets merged.

packages/wonder-blocks-dropdown/components/action-menu.test.js
wishdasher: My intention was for the test cases in the `describe` to use the same component. I suppose it may be better to compartmentalize each one.

kevinbarabash: Depending on the same component from one test to another is a bad idea b/c one test can cause other tests to fail.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -117,91 +113,96 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+                    onMouseUp={(event) => {
+                        // Stop propagation to prevent the mouseup listener
+                        // on the document from closing the menu.
                         event.nativeEvent.stopImmediatePropagation();
                     }}
-                    ref={(node) => (this.menuRef = ReactDOM.findDOMNode(node))}
+                    ref={(node) =>
+                        (this.menuElement = ((ReactDOM.findDOMNode(
+                            node,
+                        ): any): Element))
```
kevinbarabash: This is kind of a pain, but at least the `cast` is isolated.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -211,14 +212,16 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
 
         return (
             <View
+                ref={(node) =>
+                    (this.node = ((ReactDOM.findDOMNode(node): any): Element))
```
kevinbarabash: Maybe rename `this.node` to `this.element` since it's being cast as a `Element`.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -117,91 +113,96 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+                        ): any): Element))
+                    }
                     style={[
                         styles.dropdown,
                         light && styles.light,
-                        (!open || !this.menuWidth) && styles.hide,
-                        this.findTranslatedPosition(),
+                        !(open && this.menuWidth) && styles.hide,
+                        translatedPosition,
                         style,
```
kevinbarabash: I'm not sure this is the element we want to be apply the `style`.  The purpose of providing a `style` element on things to be able to adding `margin` for spacing or `width`, `align-self`, etc. for sizing.

packages/wonder-blocks-dropdown/components/dropdown-portal-mounter.js
```diff
@@ -1,159 +0,0 @@
-// @flow
```
kevinbarabash: Yay!  Thanks for being the first to use the official React modals.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -117,91 +113,96 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+                transform: `translate3d(${left +
+                    width -
+                    (this.menuWidth || 0)}px, ${top}px, 0)`,
+            };
         }
     }
 
     maybeRenderMenu() {
         const {items, light, open, style} = this.props;
+        const body = document.getElementsByTagName("body");
```
kevinbarabash: If you use `document.querySelector("body")` here you won't have to do `body[0]` below.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -117,91 +113,96 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
-                        this.findTranslatedPosition(),
+                        !(open && this.menuWidth) && styles.hide,
+                        translatedPosition,
                         style,
                     ]}
                 >
                     {items}
-                </View>
+                </View>,
+                body[0],
```
kevinbarabash: I've been told in the past that one shouldn't render directly into `body`.  Even if that's been lifted, we probably want to remove the modal container as shown in https://reactjs.org/docs/portals.html#event-bubbling-through-portals.  Can you create a container for the modal?

packages/wonder-blocks-dropdown/components/multi-select-menu.test.js
```diff
@@ -72,41 +78,46 @@ describe("MultiSelectMenu", () => {
...
-        // // Menu should still be open
-        // expect(menu.state("open")).toEqual(true);
+        const noop = jest.fn();
+        const nativeEvent = {
+            nativeEvent: {stopImmediatePropagation: noop},
+        };
+
+        expect(menu.state("open")).toEqual(true);
+        expect(menu.prop("selectedValues")).toEqual(["2"]);
+        // Grab the second item in the list
```
kevinbarabash: Can you add an extra line between sections so that it's easier to read?

packages/wonder-blocks-dropdown/components/single-select-menu.test.js
```diff
@@ -61,25 +66,31 @@ describe("SingleSelectMenu", () => {
...
-        // // instead of the placeholder
-        // expect(opener.text()).toEqual("item 2");
+        const noop = jest.fn();
+        const nativeEvent = {
+            nativeEvent: {stopImmediatePropagation: noop},
+        };
+
+        const opener = menu.find(SelectBox);
+        menu.setState({open: true});
+        // Grab the second item in the list
```
kevinbarabash: Please add a blank line between sections.

packages/wonder-blocks-dropdown/components/dropdown-core.js
kevinbarabash: Also, one of the examples should show `style` in action.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -64,78 +74,175 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+                        styles.dropdown,
+                        light && styles.light,
+                        !(open && this.menuWidth) && styles.hide,
+                        translatedPosition,
+                        style,
+                    ]}
+                >
+                    {items}
+                </View>,
+                body[0],
```
kevinbarabash: I don't know where my comment went to about this.  I tried the test page out and it look like things are working correctly like this.  Maybe I'm misunderstanding the docs around needing a separate container.

packages/wonder-blocks-dropdown/components/action-menu.test.js
somewhatabstract: The `unmount` should occur after each test case, not at the very end. Otherwise, there will be side effects between tests.

wishdasher: Okay, thanks for all the feedback!

packages/wonder-blocks-dropdown/components/dropdown-core.js
wishdasher: This is one of the things I'm going to address with the refactoring, because that would allow users to apply styles more intuitively. I'll add a TODO.

packages/wonder-blocks-dropdown/components/action-item.js
kevinbarabash: @wishdasher I've merged #222.

packages/wonder-blocks-dropdown/components/multi-select-menu.test.js
```diff
@@ -72,15 +80,20 @@ describe("MultiSelectMenu", () => {
...
+
         expect(menu.prop("selectedValues")).toEqual(["2"]);
 
         // Grab the second item in the list
         const item = menu.find(SelectItem).at(0);
         expect(item.text()).toEqual("item 1");
         // Click the item 2, deselecting it
         item.simulate("mousedown");
-        item.simulate("mouseup");
+        item.simulate("mouseup", nativeEvent);
```
kevinbarabash: Interesting.  I'm surprised `simulate` doesn't does this automatically.  Can you open an issue with `enzyme` and/or `react-dom/test-utils`, maybe they'll add stubs for it eventually?

packages/wonder-blocks-dropdown/components/single-select-menu.js
kevinbarabash: Is this comment correct now?

wishdasher:
Splitting the changes to the API and the portal work

PR #201, opened 2018-07-11
Update select box display text for various selection states

wishdasher:
I missed part of the spec the first time 'round.


packages/wonder-blocks-dropdown/components/multi-select-menu.js
```diff
@@ -164,6 +164,31 @@ export default class MultiSelectMenu extends React.Component<Props, State> {
         }
     }
 
+    getMenuText() {
+        const {items, placeholder, selectItemType, selectedValues} = this.props;
+        // If there is nothing selected, use the placeholder if it exists
```
jeresig: 🛑 Could you add a comment here with the string "i18n" so that we know to come back later and annotate this so that it can be translated, potentially?

packages/wonder-blocks-dropdown/components/multi-select-menu.js
```diff
@@ -164,6 +164,31 @@ export default class MultiSelectMenu extends React.Component<Props, State> {
...
+        const noSelectionText = placeholder
```
jeresig: Also, this could be written as: ```const noSelectionText = placeholder || `0 ${selectItemType}`;```

packages/wonder-blocks-dropdown/components/multi-select-menu.js
```diff
@@ -164,6 +164,31 @@ export default class MultiSelectMenu extends React.Component<Props, State> {
...
+                // display text for the case where nothing is selected.
+                const selectedItem = items.find(
+                    (item) => item.value === selectedValues[0],
+                );
+                return selectedItem ? selectedItem.label : noSelectionText;
+            case items.length:
+                // TODO(sophie): Configure i18n for the word "All"
+                return `All ${selectItemType}`;
+            default:
+                return `${selectedValues.length} ${selectItemType}`;
```
jeresig: 🛑 And for this one, too!

PR #202, opened 2018-07-11
Cleanup docs for v1 releases.

jeresig:
This cleans up the Wonder Blocks docs in a number of ways:

* Marks a few WIP components as private so that we don't show them on master (Dropdown, Grid, and Tooltip).
* Upgrades React Styleguidist.
* Temporarily fixes bugs with exact object types causing the link, button, and iconbutton docs to not compile (it also causes tooltip-bubble to fail but that's still WIP so I'm leaving for now).
* Fixes a number of files so that they actually show their docs.
* Move doc files around so that components have descriptions and examples.
* Write some docs, in some cases, where there were none previously.
* Use the new Button component instead of the generic HTML button.

It's not perfect, but it's at least ok enough to share for the 1.0 release of these components.


jeresig:

Built with commit 1c126134f019b7e1c63a138cb5495c232fe809a0

https://deploy-preview-202--wonder-blocks.netlify.com

packages/wonder-blocks-button/components/button.md
```diff
@@ -0,0 +1,205 @@
...
+    >Label</Button>
+    <Button
+        style={styles.sideMargins}
+        href="https://khanacademy.org"
+        disabled={true}
+    >Button with href</Button>
+</View>
+```
+
+Button examples, `style` specified (Support width, position, margin, and flex styles):
```
chenyejia: "Supports", I think

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -154,6 +104,73 @@ const startState = {
...
+ *                 >
+ *                      {this.props.children}
+ *                 </RoundRect>
+ *             }
+ *         </ClickableBehavior>
+ *     }
+ * }
+ * ```
+ *
+ * This follows a pattern call [Function as Child Components]
```
chenyejia: "called"

packages/wonder-blocks-modal/components/two-column-modal.md
```diff
@@ -142,7 +143,7 @@ const styles = StyleSheet.create({
             }
             footer={
                 <View>
-                    <button>Sample Button</button>
+                    <Button>Sample Button</Button>
```
chenyejia: nit: other buttons examples say "Button (no-op)"

packages/wonder-blocks-core/components/no-ssr.js
```diff
@@ -41,6 +26,23 @@ type State = {|
     mounted: boolean,
 |};
 
+/**
+ * Defer or change rendering until the component did mount.
```
kevinbarabash: What's the rationale for moving doc strings to be after Props?

packages/wonder-blocks-core/components/text.js
```diff
@@ -30,6 +30,19 @@ const styles = StyleSheet.create({
     },
 });
 
+/**
+ * Text is a building block for constructing other components. `View` roughly
+ * maps to `div` and `Text` roughly maps to `span`. You can override which tag
+ * is used to render the component (for semantic purposes) by specifying the
+ * `tag` prop.
```
kevinbarabash: The `View` parts of this comment could be removed.  I think the docs for `Text` and `View` need improvement so this is something that can be addressed whenever we improve the docs for both of these components.  As I was reading through the docs for "Core" it would be nice if we could categorize exports into components, functions, constants.  But that would be a bigger general docs improvement change.

packages/wonder-blocks-core/components/text.md
```diff
@@ -0,0 +1,40 @@
...
+<View style={styles.container}>
+    <Text style={styles.text}>
+        Hello, world!
+    </Text>
+    <View style={[
+        styles.container,
+        {backgroundColor: "goldenrod", padding: 4},
+    ]}>
+        Even with an array and inline styles!
+    </View>
```
kevinbarabash: We'd probably want a more `Text` specific example here, especially one that better motivates when you'd want to use `Text` vs. our other typography components, e.g. showing how to make a word within a sentence bold or a different color.

packages/wonder-blocks-core/components/unique-id-provider.md
```diff
@@ -1,9 +1,3 @@
-The `UniqueIDProvider` component is how Wonder Blocks components obtain unique identifiers. This component ensures that server-side rendering and initial client rendering match while allowing the provision of unique identifiers for the client.
```
kevinbarabash: Why move this to the implementation file?  It would be good to codify guidelines around documentation in the Wonder Blocks dev guide.

packages/wonder-blocks-core/docs.md
```diff
@@ -1,73 +1 @@
...
-    pink: {
-        backgroundColor: "pink",
-    },
-});
-
-const StyledInput = addStyle("input", styles.input);
-
-<StyledInput style={styles.pink} type="text" placeholder="hello, world"/>;
-```
+`wonder-blocks-core` provides building blocks and utilities for constructing other `wonder-blocks` components. Many of these may be useful inside your applications, as well.
```
kevinbarabash: I see, we only include introductory material to the package in docs.md.

packages/wonder-blocks-grid/components/flex-cell.js
```diff
@@ -50,9 +50,6 @@ type Props = {|
  * By default (with no properties specified) it will display at all
  * grid sizes. If you specify the `small`, `medium`, or `large`
  * props then the component will only be shown at those grid sizes.
- *
- * @version 1.0
- * @since 1.0
```
kevinbarabash: Good call removing them.  Maintenance would be a nightmare.

packages/wonder-blocks-icon-button/components/icon-button.js
```diff
@@ -63,10 +63,11 @@ export type SharedProps = {|
...
     }>>,
     */
-|};
 
-type Props = {|
-    ...SharedProps,
+    // NOTE(jeresig): Currently React Docgen (used by Styleguidist) doesn't
+    // support ... inside of an exact object type. Thus we had to move the
+    // following propers into this SharedProps, even though they should be
+    // external. Once that's fixed we can split them back apart.
```
kevinbarabash: Sadness.  I'll try to make some time to investigate this.

packages/wonder-blocks-icon-button/generated-snapshot.test.js
```diff
@@ -31,21 +31,21 @@ describe("wonder-blocks-icon-button", () => {
                     icon={icons.search}
                     aria-label="search"
                     onClick={(e) => console.log("hello")}
-                    style={[styles.sideMargins]}
+                    style={styles.sideMargins}
```
kevinbarabash: Yes!

packages/wonder-blocks-icon/components/icon.js
```diff
@@ -44,7 +44,7 @@ const StyledSVG = addStyle("svg");
  * <Icon
  *     icon={icons.search}
  *     color={Color.white}
- *     size={"medium"}
+ *     size="medium"
```
kevinbarabash: I tried to find a eslint rule for this but I couldn't.  I'm kind of surprised that prettier isn't opinionated about this.

packages/wonder-blocks-layout/docs.md
```diff
@@ -4,25 +4,19 @@ is uncompressible and occupies a fixed amount of space specified by its `size` p
...
 const {StyleSheet} = require("aphrodite");
-const {View, addStyle} = require("@khanacademy/wonder-blocks-core");
+const {View} = require("@khanacademy/wonder-blocks-core");
 const Color = require("@khanacademy/wonder-blocks-color").default;
 const Spacing = require("@khanacademy/wonder-blocks-spacing").default;
+const Button = require("@khanacademy/wonder-blocks-button").default;
 const {Spring, Strut} = require("./index.js");
 
-// TODO(kevinb): replace with wonder-blocks-button once yarn workspaces land
-const Button = addStyle("button");
```
kevinbarabash: Thanks.

packages/wonder-blocks-modal/components/modal-launcher.md
```diff
@@ -45,10 +46,9 @@ const standardModal = ({closeModal}) => (
             </View>
         }
         footer={
-            // TODO(mdr): Use Wonder Blocks Button.
-            <button onClick={closeModal}>
+            <Button onClick={closeModal}>
                 Close modal
-            </button>
+            </Button>
```
kevinbarabash: Yay!

packages/wonder-blocks-button/components/button.md
```diff
@@ -0,0 +1,205 @@
...
+    </tr>
+    <tr>
+        <td>alignSelf: 'flex-end'</td>
+        <td>
+            <div style={{display: 'flex'}}>
+                <Button
+                    onClick={(e) => console.log("Hello, world!")}
+                    style={[{alignSelf: 'flex-end'}, styles.sideMargins]}
+                >Label</Button>
+                <div
```
somewhatabstract: Should use View in examples, not div.

packages/wonder-blocks-button/components/button.md
```diff
@@ -0,0 +1,205 @@
...
+        <td>
+            <div style={{display: 'flex'}}>
+                <Button
+                    onClick={(e) => console.log("Hello, world!")}
+                    style={[
+                        {flexShrink: 2, width: '300px'},
+                        styles.sideMargins
+                    ]}
+                >Label</Button>
+                <div
```
somewhatabstract: View

packages/wonder-blocks-button/components/button.md
```diff
@@ -0,0 +1,205 @@
...
+            <Button
+                onClick={(e) => console.log("Hello, world!")}
+                style={styles.centerInDiv}
+            >Label</Button>
+        </td>
+    </tr>
+    <tr>
+        <td>flexGrow: 1</td>
+        <td>
+            <div style={{display: 'flex'}}>
```
somewhatabstract: View

packages/wonder-blocks-button/components/button.md
```diff
@@ -0,0 +1,205 @@
...
+                    onClick={(e) => console.log("Hello, world!")}
+                    style={{flexGrow: 1}}
+                >Label</Button>
+            </div>
+        </td>
+    </tr>
+    <tr>
+        <td>flexShrink: 2, width: '300px'</td>
+        <td>
+            <div style={{display: 'flex'}}>
```
somewhatabstract: View

packages/wonder-blocks-button/components/button.md
```diff
@@ -0,0 +1,205 @@
...
+                        }
+                    }
+                >Not a button</div>
+            </div>
+        </td>
+    </tr>
+    <tr>
+        <td>alignSelf: 'flex-end'</td>
+        <td>
+            <div style={{display: 'flex'}}>
```
somewhatabstract: View

packages/wonder-blocks-button/components/button.md
```diff
@@ -0,0 +1,205 @@
...
+                        }
+                    }
+                >Not a button</div>
+            </div>
+        </td>
+    </tr>
+    <tr>
+        <td>justifySelf: 'flex-end'</td>
+        <td>
+            <div style={{display: 'flex'}}>
```
somewhatabstract: View

packages/wonder-blocks-button/components/button.md
```diff
@@ -0,0 +1,205 @@
...
+                <div style={{display: 'grid', width: '100%'}}>
```
somewhatabstract: View (you can override to be display grid with style prop)

packages/wonder-blocks-core/components/no-ssr.js
somewhatabstract: This attaches the docs to the type itself, rather than the file. I can see that being useful.

packages/wonder-blocks-core/components/unique-id-provider.md
somewhatabstract: Yes, please leave this in the docs too. I think having the styleguide be a manual of sorts in the dev version is really valuable and this is a component that we want people to know about as it is exported for reuse.

packages/wonder-blocks-icon-button/components/icon-button.js
somewhatabstract: 3.0 beta supports spreads. Maybe we should try that.

styleguide.config.js
```diff
@@ -26,6 +26,8 @@ module.exports = {
             components: "packages/wonder-blocks-core/components/*.js",
         },
         {
+            // NOTE(jeresig): Work-in-progress.
+            private: true,
```
somewhatabstract: Does it make sense to make this manual? Should we change the script behavior to exclude sections that map to a private/prototype repo when we're producing the prod docs? Seems like that would be the better approach, with less effort and less margin for error when something moves out of private/prototype.

packages/wonder-blocks-button/components/button.md
jeresig: Fixed!

packages/wonder-blocks-core/components/clickable-behavior.js
jeresig: Fixed!

styleguide.config.js
somewhatabstract: OK, my changes to the script are in master now, so this manual setting of WIP sections is no longer needed! 🎊

packages/wonder-blocks-modal/components/two-column-modal.md
jeresig: Fixed!

packages/wonder-blocks-core/components/text.js
jeresig: Removed the View part.

packages/wonder-blocks-core/components/text.md
jeresig: Hmm, ok - I think I'll hold off on that until we improve the Text docs in general.

packages/wonder-blocks-core/docs.md
jeresig: Yep! docs.md for examples, everything else in the file. We should probably rename docs.md to be "examples.md" or similar.

packages/wonder-blocks-button/components/button.md
jeresig: Fixed!

packages/wonder-blocks-button/components/button.md
jeresig: Fixed the Views.

styleguide.config.js
jeresig: Removed this!

jeresig:
Thanks for the feedback, everyone! @somewhatabstract: I think I've addressed all your feedback. We discussed this off-thread but I'm opting to leave out the introductions to the components from the example files as having them be in the component themselves both makes it easier to find but also makes the formatting in styleguidist must easier to understand. Let me know if you have any questions!

somewhatabstract:
LGTM!

Ship it!

PR #206, opened 2018-07-13
Add form package and basic checkbox

wishdasher:
This adds the basic checkbox component. The docs have examples of unselected and selected checkboxes in the default, disabled, and error states. The docs also show the styling of each checkbox under ClickableBehavior, and it should match the designs from the [zeplin](https://zpl.io/brrRQx5) spec.


jeresig:

Built with commit 536b56b4d2d25c3f17f09aa8cb2be027a630550d

https://deploy-preview-206--wonder-blocks.netlify.com

somewhatabstract:
Would love to see testing get added as we add new things. I think it's the right thing to do.

packages/wonder-blocks-form/components/checkbox-core.js
```diff
@@ -0,0 +1,225 @@
...
+    hovered: boolean,
+    focused: boolean,
+    pressed: boolean,
+|};
+
+const {blue, red, white, offWhite, offBlack16, offBlack32, offBlack50} = Color;
+
+const StyledInput = addStyle("input");
+
+const checkboxCheck: IconAsset = {
```
jeresig: Err - so all checkboxes supposed to have the smaller check, not just the ones in the menu? I feel like I'd super-like the menu checkbox to use this checkbox-core.js now, if at all possible (even if we have a separate set of focus/hover states)!

kevinbarabash:
@wishdasher https://codecov.io/gh/Khan/wonder-blocks/pull/206/diff shows where code coverage can be improved.

packages/wonder-blocks-form/components/checkbox.md
```diff
@@ -0,0 +1,79 @@
...
+    <Checkbox error={false} checked={false} style={[styles.marginRight]} onChange={checked => handleChanged(checked)} />
+    <Checkbox error={false} checked={true} style={[styles.marginRight]} onChange={checked => handleChanged(checked)} />
+    <Checkbox error={true} checked={false} style={[styles.marginRight]} onChange={checked => handleChanged(checked)} />
+    <Checkbox error={true} checked={true} style={[styles.marginRight]} onChange={checked => handleChanged(checked)} />
+    <Checkbox disabled={true} checked={false} style={[styles.marginRight]} onChange={checked => handleChanged(checked)} />
+    <Checkbox disabled={true} checked={true} style={[styles.marginRight]} onChange={checked => handleChanged(checked)} />
+</View>
+```
+
+Here are all the checkbox state and style combinations.
```
kevinbarabash: Can you move this into a .test.js file similar to what we do in `wonder-blocks-button/custom-snapshot.test.js`?

packages/wonder-blocks-form/index.js
```diff
@@ -0,0 +1,4 @@
+// @flow
+import Checkbox from "./components/checkbox.js";
+
+export {Checkbox};
```
kevinbarabash: I see, eventually there will be more things exported.

packages/wonder-blocks-form/util/types.js
```diff
@@ -0,0 +1,26 @@
...
+    disabled?: boolean,
+    /** Whether this component should show an error state */
+    error?: boolean,
+    /** Name for the checkbox or radio button group */
+    groupName?: string,
+    /** Unique identifier attached to the HTML input element. If used, need to
+     * guarantee that the ID is unique within everything rendered on a page. */
+    id?: string,
+    /** Optional test ID for e2e testing */
+    testId?: string,
```
kevinbarabash: Thanks for adding these to everything.

packages/wonder-blocks-form/components/checkbox.js
```diff
@@ -0,0 +1,34 @@
...
+
+export default class Checkbox extends React.Component<ChoiceProps> {
+    static defaultProps = {
+        disabled: false,
+        error: false,
+    };
+
+    render() {
+        const {onChange, ...coreProps} = this.props;
+        const ClickableBehavior = getClickableBehavior(this.context.router);
```
kevinbarabash: The first param should be an `href` but since we don't have one we should pass nothing.

packages/wonder-blocks-form/components/checkbox.js
```diff
@@ -0,0 +1,34 @@
...
+
+        return (
+            <ClickableBehavior
+                disabled={coreProps.disabled}
+                onClick={() => onChange(coreProps.checked)}
```
kevinbarabash: I think this should be `onChange(!coreProps.checked)`.  One of the examples should probably show this in action so that we have something to test even if we're going to use `CheckboxGroup` in the future.

packages/wonder-blocks-form/components/checkbox-core.js
```diff
@@ -0,0 +1,195 @@
...
+        base: red,
+        active: activeRed,
+    },
+};
+
+const styles = {};
+
+const _generateStyles = (checked, error) => {
+    // "hash" the parameters
+    const styleKey = checked + 2 * error;
```
kevinbarabash: I don't get the `2 * error` part of this.  I thought `checked` and `error` were booleans?

packages/wonder-blocks-form/components/checkbox-core.js
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+const sharedStyles = StyleSheet.create({
+    wrapper: {
+        outline: "none",
+    },
+
+    // Reset the default styled input element
+    inputReset: {
```
kevinbarabash: This seems like a good candidate to move into the `overrides` stylesheet in `add-style.js`.

packages/wonder-blocks-form/components/checkbox-core.js
```diff
@@ -0,0 +1,195 @@
...
+        "M11.263 4.324a1 1 0 1 1 1.474 1.352l-5.5 6a1 1 0 0 1-1.505-.036l-2.5-3a1 1 0 1 1 1.536-1.28L6.536 9.48l4.727-5.157z",
+};
+
+export default class CheckboxCore extends React.Component<Props> {
+    render() {
+        const {
+            checked,
+            disabled,
+            error,
+            id,
```
kevinbarabash: It seems like there are two routes we can go.  We can use the `for`/`id` pattern with `<label>`s or `aria-labelledby`/`id` pattern.  My preference would be for using `aria-labelledby`.  The reason being that we using `aria-labelledby` in other places and it would be nice to have fewer a11y patterns to think about.  cc @somewhatabstract

packages/wonder-blocks-form/components/checkbox-core.js
```diff
@@ -0,0 +1,195 @@
...
+    checkIcon: {
+        position: "absolute",
+        pointerEvents: "none",
+    },
+});
+
+const fadedBlue = mix(fade(blue, 0.16), white);
+const activeBlue = mix(offBlack32, blue);
+const fadedRed = mix(fade(red, 0.08), white);
+const activeRed = mix(offBlack32, red);
```
kevinbarabash: Not necessary for this PR, but I wonder if we can move some of these colours into `wonder-blocks-color` since they're used in a number of different components.

packages/wonder-blocks-form/components/checkbox.js
wishdasher: I agree, this is actually changed to `!coreProps.checked` in a later diff.

wishdasher: I'll add a test in the #236, which would use the new signature.

packages/wonder-blocks-form/components/checkbox-core.js
wishdasher: This style isn't in the normalize sheet. It's a reset on the browser-specific styles that are added to checkboxes and radio buttons. It may be the cause that we don't need them to apply for all `input` objects.

packages/wonder-blocks-form/components/checkbox-core.js
kevinbarabash: I did some digging and according https://www.w3.org/WAI/tutorials/forms/labels/ we should go with `<label>` and then use `for` and `id`.

PR #207, opened 2018-07-13
Try alternative focus ring approach

wishdasher:
Not asking for this to be merged -- just wanted to bring up a visual bug we have.

Currently, we create a focus ring by using the `::before` pseudo-selector and adding a pixel-positioned border around the element. This ring doesn't work at all zoom levels. For example, this is what the ring looks like around a primary Button at 125% on my system.
screen shot 2018-07-13 at 15 14 55
It's possible to use two box shadows around the element to simulate the focus ring, and that looks good at all zoom levels.
screen shot 2018-07-13 at 15 15 03
However, this approach involves setting an explicit `1px`-wide white box shadow around the component...which wouldn't work as well if the component is being used on a non-white background. As far as I can tell, there is not a way to leverage the "transparent" color to work for this.

Haven't found a better solution yet. :/


jeresig:

Built with commit f03631ba275e17c8cd19bcfa6880bd9fab76ec3a

https://deploy-preview-207--wonder-blocks.netlify.com

somewhatabstract:
Will an rgba color work with the alpha all the way transparent?

wishdasher:
Turns out there's a variable `currentColor` that solves all our problems. Thanks slim.

chenyejia:
Current color is the text color, which can still be different from the background color!

wishdasher:
@chenyejia You're right, I didn't fully understand what that variable was. 😞The select box on a dark background was misleading because the black looked similar to the dark blue.

@somewhatabstract Setting the alpha all the way to transparent lets the outer box shadow color show through instead.

wishdasher:
I got it to work by adding a wrapper and setting its background color to `inherit`, then using `getComputerStyle` after the component mounts to figure out what the parent background color should be. Seems slightly overkill / like an anti-pattern though.

wishdasher:
I made a potential fix in #206 for the Choice checkbox component. If that solution gets merged, I'll come back and fix the rings on Button and SelectBox.

PR #209, opened 2018-07-14
Update padding around content to match latest design spec

somewhatabstract:
This change moves the padding from `TooltipBubble` to a margin on `TooltipContent`. The move is because the `TooltipContent` component knows whether it has a heading or not, and we now have different top/bottom margins based on the presence of a tooltip title.

So, now, with a title, we have 16px margin all around, but without a title, we have 16px on left and right and 10px on top and bottom, as per spec: see note https://zpl.io/2j4zNjQ.


jeresig:

Built with commit 24950a95136cccb95d3b329e1d871946841f2026

https://deploy-preview-209--wonder-blocks.netlify.com

packages/wonder-blocks-tooltip/components/tooltip-content.js
```diff
@@ -53,12 +54,23 @@ export default class TooltipContent extends React.Component<Props> {
...
     }
 }
+
+const styles = StyleSheet.create({
+    withoutTitle: {
+        margin: "10px 16px",
+    },
+
+    withTitle: {
+        margin: 16,
```
jeresig: We should update the 16pxs to Spacing.medium -- I'm not sure why we're using 10px, it's not one of our standard spacing units!

somewhatabstract: That's what the design says. I didn't use `Spacing.medium` because the 10px was a spacing option and it seemed better to be consistent with that.

PR #212, opened 2018-07-17
add missing peerDependencies

kevinbarabash:
I noticed that some of the packages that were published to npm are bigger than they should be.  This PR adds react, aphrodite, and react-router-dom as peer deps where need which will cause them to be excluded from the bundle b/c our build script marks anything that appears in `dependencies` or `peerDependencies` as external in the webpack settings for that package.

### DIFF
--- a/packages/wonder-blocks-button/package.json
+++ b/packages/wonder-blocks-button/package.json
@@ -17,5 +17,10 @@
     "@khanacademy/wonder-blocks-color": "^1.0.4",
     "@khanacademy/wonder-blocks-core": "^1.0.4",
     "@khanacademy/wonder-blocks-typography": "^1.0.4"
+  },
+  "peerDependencies": {
+    "aphrodite": "^1.2.5",
+    "react": "^15.3.2",
+    "react-router-dom": "^4.2.2"
   }
 }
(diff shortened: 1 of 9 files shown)


jeresig:

Built with commit 8c9dd7c85403dc52077a01b1d6f8040eb9c0fdd8

https://deploy-preview-212--wonder-blocks.netlify.com

kevinbarabash:
`Code owner review required`, ugh.  @jeresig I have to update CODEOWNERS so that you and I appear in each entry.

MiguelCastillo:
@jeresig @kevinbarabash Im assuming by "help update these dependencies" we mean update the peerDependencies in each package.json for each wonder-blocks package.  That's more the job of `yarn` to manage dependencies. Normally, you would want to run `yarn upgrade-interactive` which will detect if you have a workspace and prompt you to update each package individually.  You define specific dependencies to update too.

`yarn upgrade-interactive react-router-dom --latest` will update only react-router-dom to the latest in each package.

However - the sad thing is that yarn currently isn't working very well with peer dependencies in workspaces :-/   But it would not be a big lift to put something together to keep those peer dependencies up to date without having to update each one manually.

PR #216, opened 2018-07-18
Fix modal tests after update to react 16

wishdasher:
I think the tests were failing because they were doing a shallow render, causing `ReactDOM.findDOMNode(this)` in `ModalBackdrop.componentDidMount` to fail. If we do a full mount instead, the tests pass. Alternatively, we can disable the lifecycle methods, which don't appear to be relevant what these particular tests are testing for.


jeresig:

Built with commit 8f3c0cc2a1e2d71e62994da480cf5124ee8107dc

https://deploy-preview-216--wonder-blocks.netlify.com

somewhatabstract:
#162 already has some work for upgrading React 16. Perhaps these can be combined?

jeresig:
@wishdasher Thank you for fixing this!

PR #217, opened 2018-07-20
Allow components to be focusable by focus event

wishdasher:
I tested this in the browser environment with the console by grabbing two buttons and setting timeouts to call `.focus()` on them at different times. As expected, one button gained the `focused` style first, and then lost it when the other button became `focused`. I wasn't able to test this in enzyme because the enzyme environment doesn't simulate this type of browser behavior.

Tabbing to focus component still works.


jeresig:

Built with commit 600794138b75797ea73c069322c7a22ffebdd20b

https://deploy-preview-217--wonder-blocks.netlify.com

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -236,9 +235,7 @@ export default class ClickableBehavior extends React.Component<Props, State> {
 
     handleKeyUp = (e: SyntheticKeyboardEvent<*>) => {
         const keyCode = e.which || e.keyCode;
-        if (keyCode === keyCodes.tab) {
-            this.setState({focused: true});
-        } else if (
```
wishdasher: I removed this because it's not necessary. When tabbing into or away from an element, `onFocus` and `onBlur` respectively fire for that element. Previously, components were updating their state to `focused: false` twice when tabbed away from, once from `handleKeyDown` and once from `handleBlur`. I verified this by adding a temporary logging statement in `componentDidUpdate`.

packages/wonder-blocks-core/components/clickable-behavior.test.js
```diff
@@ -79,22 +79,6 @@ describe("ClickableBehavior", () => {
...
-                }}
-            </ClickableBehavior>,
-        );
-        expect(button.state("focused")).toEqual(false);
-        button.simulate("keyup", {keyCode: keyCodes.tab});
-        expect(button.state("focused")).toEqual(true);
-        button.simulate("keydown", {keyCode: keyCodes.tab});
-        expect(button.state("focused")).toEqual(false);
-    });
-
```
wishdasher: This test is now obsolete because I removed the tab-related code from `handleKeyUp` and `handleKeyDown`. The tab-focus behavior is now handled by `handleBlur` and `handleFocus`.

chenyejia:
I remember I chose not to use this for a reason, but I can't remember why right now, and I'm going to be on vacation 'til Tuesday. Will take a look then if this is still open!

jeresig:
@chenyejia I've approved this but I would be interested to hear your thoughts on this, if you have the time when you return!

chenyejia:
Hm I think I was getting flow or type errors or something? But I can't remember why...

But also, I don't think this matches design spec (from my understanding). If you click a button, then unfocus either the browser or the tab and then come back, there is a focus ring around the button, which I _think_ is not what design wanted?

wishdasher:
I think the focus ring is still around the button when you come back because the browser returns focus to the last focused element on the page. This seems helpful for keyboard navigation. Is there a design spec to look at?

chenyejia:
Yup I think that's helpful for keyboard navigation as well!

Uhh, I think keyboard interaction specs are kind of scattered throughout Slack. Moving this convo to Slack to consolidate!

PR #219, opened 2018-07-20
Improve focus ring on button, add light version in documentation

wishdasher:
Using a box shadow to add the focus ring removes off-centeredness at different zoom levels. 
image image
An assumption being made here is that the regular button appears on white backgrounds, and the light version appears on dark blue backgrounds. Even if the background colors don't exactly match, I think the inner ring is small enough (1px wide) for it to not be a huge visual issue. The light version of the button was already designed for a dark blue background anyways.

An alternative method is to use refs and recursively find the background color of the parent element in the DOM tree. It would need to be recursive because most elements have a computed `background-color` of `transparent`. I think this method is overly complicated and may also introduce extra wrapper elements, but I can switch to this method if people think it's a safer way of doing this.


jeresig:

Built with commit 66f8b0e76d6b9a659d1835e93df54c862bb7937f

https://deploy-preview-219--wonder-blocks.netlify.com

wishdasher:
I talked to Elizabeth, and she said that generally buttons should only be on the white or dark blue background. However, marketing may deviate from these conventions.

wishdasher:
Oops, I accidentally closed this.

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -158,18 +158,9 @@ const _generateStyles = (color, kind, light) => {
...
-                    top: -3,
-                    left: -3,
-                    right: -3,
-                    bottom: -3,
-                    borderColor: light ? white : color,
-                    borderWidth: 2,
-                    borderStyle: "solid",
-                    borderRadius: 7,
-                },
+                boxShadow: `0 0 0 1px ${light ? darkBlue : white}, 0 0 0 3px ${
```
jeresig: Ah - this is a great solution! Very easy to understand. It might be worth just adding a comment here that we're making the assumption that the background will be a particular color.

kevinbarabash:
screen shot 2018-07-20 at 7 17 58 pm

Even if the background color is a different color it doesn't look awful.  It kind of reminds of stripes on track suits.

kevinbarabash:
I'm happy we were able to fix this.

PR #221, opened 2018-07-20
WB-175: update modal to use IconButton

kevinbarabash:
Modals were already using IconButton but for some reason the
focus/hover ring wasn’t showing up.  This PR gets rid of the
ModalCloseButton component and uses IconButton directly.
I also adjust the position of the icon to be closer to what the spec
is.  There's still some additional work to do to have it match spec 
complete which will be done in a follow up PR.

screen shot 2018-07-20 at 5 50 03 pm

screen shot 2018-07-20 at 5 45 28 pm

screen shot 2018-07-20 at 6 04 00 pm


jeresig:

Built with commit a5fa772a89fd2ce0ccb27c6e26f65205e5534fbd

https://deploy-preview-221--wonder-blocks.netlify.com

PR #222, opened 2018-07-20
WB-322: buttons on Safari have extra 2px left/right margin

kevinbarabash:
This PRs removes extra padding around Buttons that use <button> on Safari.  It also updates `yarn start` to run `styleguidist` and `webpack -w` concurrently.

Test Plan:
- load http://localhost:6060 in Safari
- inspect the buttons in the inspector, see that they don't have any left padding (the right padding is added by the example, not Safari's built-in styles)

Before:
screen shot 2018-07-20 at 6 45 44 pm

After:
screen shot 2018-07-20 at 6 45 07 pm


jeresig:

Built with commit ce7802610751045ffc86abd24f8647e2c5356fb4

https://deploy-preview-222--wonder-blocks.netlify.com

packages/wonder-blocks-core/util/add-style.js
```diff
@@ -29,3 +33,10 @@ export default function addStyle<T: Object>(
 
     return MediaLayoutWrapper(StyleComponent);
 }
+
+// TODO(kevinb): add more style resets
+const styles = StyleSheet.create({
+    button: {
+        margin: 0, // Safari adds 2px left/right margins
+    },
```
kevinbarabash: @wishdasher we could add other resets based on normalizer.css here if necessary.

jeresig: 🤷‍♂️ I'm... iffy on this solution. I could see how this would be useful but since we're applying it intermittently I could also see how it'd be really confusing. I think this is ok for now but I think we should either flush this out a lot more or move it to be in the components themselves.

I think it'd be good to add a comment explaining the way in which we intend this to be used, so that we don't add in non-browser-specific overrides (also probably improve the variable name so that it's clearer, too).

kevinbarabash: I'll add some comments and improve the naming so that people don't abuse this.

PR #225, opened 2018-07-21
setup circleci

kevinbarabash:
I've found circleci builds are faster.  Also, we can parallelize different parts of the build, e.g. run eslint, flow, etc. as one sub-job and the build itself along with all tests as a separate sub-job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,22 @@
+version: 2
+jobs:
+  one:
+    docker:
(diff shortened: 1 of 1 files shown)


jeresig:

Built with commit a12a38fdf7baae64b6e55fc5942b35058347e768

https://deploy-preview-225--wonder-blocks.netlify.com

kevinbarabash:
It looks like this needs to go on `master` before circleci will do anything.

PR #226, opened 2018-07-21
update .circleci/config.yml to actually lint our code and run the tests and coverage

kevinbarabash:
This PR replaces travis-ci with circleci.  Circleci will run on PRs or pushes but prevents duplicate builds when both occur in the case of PRs for branches that aren't in a fork.  It also allows multiple jobs to be run in parallel.  In this case we're running eslint and flow in one job called `lint` and building the code and running the tests along with coverage is in another job called `test`.


kevinbarabash:
Cool!  The changes are being picked up by circleci, time to get it to run the actual jobs we want it to run.

jeresig:

Built with commit 9946506356c5dd5f2089f92d1b0b28424e7cd470

https://deploy-preview-226--wonder-blocks.netlify.com

kevinbarabash:
There doesn't seem to be a big difference between using a cache and not.  I'm going to leave it though.

.circleci/config.yml
```diff
@@ -1,22 +1,42 @@
 version: 2
 jobs:
-  one:
+  lint:
     docker:
       - image: circleci/node:8-browsers
```
kevinbarabash: circleci runs all tests within docker containers which means you can actually run test tests in the exact same environment locally using their `circleci` CLI too.  You can also use the tool's `config validate` command to check that changes made to config.yml are valid.

PR #227, opened 2018-07-22
test out using vite to generate screenshots of components

kevinbarabash:
This PR is an experiment and should not be merged.  Before `vite` can be used reliably for visual regression testing we need to be able run it using Docker so that we can have consistent environment. There's also some additional work to do on the test harness itself like loading Lato and making sure that it finishes loading before taking any screenshots.


jeresig:

Built with commit a4aea195a808c39c6847a8c0b8c23e5119210a74

https://deploy-preview-227--wonder-blocks.netlify.com

fixtures/button.fixture.js
```diff
@@ -0,0 +1,34 @@
...
+        await expect(element).toMatchScreenshot();
+        await simulate({type: "mousemove", clientX: 50, clientY: 25});
+        await expect(element).toMatchScreenshot({outset: 4});
+        await simulate({type: "mousedown"});
+        await expect(element).toMatchScreenshot();
+    });
+
+    test("primary (tab to focus)", async () => {
+        const element = await render(<Button>Hello, world!</Button>);
+        await simulate({type: "keypress", key: "tab"});
```
kevinbarabash: @matchu we should be able to write a test for focus capture using this.  I'll try to write a test for this.

fixtures/button.fixture.js
```diff
@@ -0,0 +1,34 @@
...
+import {describe, expect, test, render, simulate} from "./helpers.js";
+
+describe("Button", () => {
+    test("primary", async () => {
+        const element = await render(<Button>Hello, world!</Button>);
+        await expect(element).toMatchScreenshot();
+        await simulate({type: "mousemove", clientX: 50, clientY: 25});
+        await expect(element).toMatchScreenshot({outset: 4});
+        await simulate({type: "mousedown"});
+        await expect(element).toMatchScreenshot();
```
kevinbarabash: We should probably have a way to pass labels to `toMatchScreenshot()` so that it's easier to identify the screenshots.

fixtures/helpers.js
```diff
@@ -0,0 +1,104 @@
...
+            setTimeout(() => resolve(container), 0);
+        });
+    });
+}
+
+let currentTest = null;
+
+export async function runTests() {
+    for (const test of tests) {
+        currentTest = test;
```
kevinbarabash: This is kind of hacky, but I couldn't figure out a better way of passing the test title to `toMatchScreenshot`.

fixtures/helpers.js
```diff
@@ -0,0 +1,104 @@
...
+export function expect(domElement) {
+    return {
+        toMatchScreenshot: async (options = {}) => {
+            const bounds = domElement.getBoundingClientRect();
+            const titlebarHeight = window.outerHeight - window.innerHeight;
+
+            const outset = options.outset || 0;
+            const title = currentTest.title
+                .replace(/, /g, "_")
+                .replace(" ", "_");
```
kevinbarabash: I should probably keep title as an array of strings and `join` then using appropriate separators where needed.

fixtures/link.fixture.js
```diff
@@ -0,0 +1,35 @@
...
+describe("Link", () => {
+    test("primary", async () => {
+        const element = await render(
+            <Link href={randomLink()}>Hello, world!</Link>,
+        );
+        await expect(element).toMatchScreenshot();
+        await simulate({type: "mousemove", clientX: 50, clientY: 20});
+        await expect(element).toMatchScreenshot();
+        await simulate({type: "click"});
+        await sleep(100);
```
kevinbarabash: This is necessary b/c the browser doesn't update the style of visited links until the URL changes which takes some time.  We should probably wait for the URL change instead of hardcoding a set duration.

fixtures/link.fixture.js
```diff
@@ -0,0 +1,35 @@
...
+import {describe, expect, test, render, simulate, sleep} from "./helpers.js";
+
+function randomLink() {
+    return "#" + String(Math.random()).slice(2);
+}
+
+describe("Link", () => {
+    test("primary", async () => {
+        const element = await render(
+            <Link href={randomLink()}>Hello, world!</Link>,
```
kevinbarabash: Random URLs are used to prevent subsequent test runs from interfering with each other.

package.json
```diff
@@ -52,7 +52,7 @@
     "eslint-watch": "^4.0.1",
     "flow-bin": "0.75.0",
     "glob": "^7.1.2",
-    "jest": "^23.3.0",
+    "jest": "^23.4.1",
```
kevinbarabash: This isn't actually necessary.  I was running into issues with some of `vite`'s deps interfering with `jest`.  I think it was `frametalk` which had an old version of `jest` as a dependency.

kevinbarabash:
The travis build is failing b/c one of `vite`'s dependencies requires `robot-js` which is a native module and it can't find X11 headers when trying to compile on `travis-ci`.  We only really need `robot-js` inside the docker container... I should investigate building a container that has the `vite` server installed and then have a separate package which contains the test runner and then helper functions.

kevinbarabash:
The reason why all the screenshots changed between the first and second commit is b/c I was using an external monitor for the first set and was using my laptop's screen for the second.  These consistency problems will be solved by using Xvfb in a Docker image.

fixtures/modal.fixture.js
```diff
@@ -0,0 +1,43 @@
...
+            </ModalLauncher>,
+        );
+        await simulate({type: "mousemove", clientX: 50, clientY: 25});
+        await simulate({type: "click"});
+        await sleep(100);
+        await expect(document.body).toMatchScreenshot();
+        await simulate({type: "keypress", key: "tab"});
+        await expect(document.body).toMatchScreenshot();
+        await simulate({type: "keypress", key: "tab"});
+        await expect(document.body).toMatchScreenshot();
```
kevinbarabash: @matchu checkout the modal screenshots below.  They show that focus returns to the button that had focus when the modal was first opened.

fixtures/modal.fixture.js
```diff
@@ -0,0 +1,64 @@
...
+            type: "mousemove",
+            clientX: bounds.left + bounds.width / 2,
+            clientY: bounds.top + bounds.height / 2,
+        });
+        await expect(document.body).toMatchScreenshot();
+
+        // test that closing the modal works
+        await simulate({type: "click"});
+        await sleep(100);
+        await expect(document.body).toMatchScreenshot();
```
kevinbarabash: This last screenshot isn't really necessary b/c we don't care what the page looks like after we dismiss the modal.

fixtures/modal.fixture.js
```diff
@@ -0,0 +1,64 @@
...
+        // get the location of the close button and move the mouse over it
+        const closeButton = document.querySelector(
+            `[data-test-id="close-modal"]`,
+        );
+        const bounds = closeButton.getBoundingClientRect();
+        await simulate({
+            type: "mousemove",
+            clientX: bounds.left + bounds.width / 2,
+            clientY: bounds.top + bounds.height / 2,
+        });
```
kevinbarabash: It would be nice to have a helper that computes the coordinates of the middle of an element.

fixtures/button.fixture.js
matchu: Ooh yay that would be so exciting!

kevinbarabash: I did write a test for this.  Check out modal.fixture.js down below.

PR #229, opened 2018-07-23
Use Toolbar instead of ModalTitleBar

scotchfield:
We can replace the existing ModalTitleBar using the new Toolbar component without too much trouble.

This is an updated version of https://github.com/Khan/wonder-blocks/pull/185.

By the way, are there issues referencing unpublished components in published ones? For example, will the line `"@khanacademy/wonder-blocks-toolbar": "^0.0.1",` cause problems in `wonder-blocks-modal`?


kevinbarabash:
I would merge #230 first.  I'm curious to see how the `Toolbar`s look in the `Modal`s once they start using `IconButton`.

PR #230, opened 2018-07-23
Use Wonder Blocks icons and IconButton in the Toolbar documentation

scotchfield:


jeresig:

Built with commit c38dcf763292487422372bf3237b8fb9d6096952

https://deploy-preview-230--wonder-blocks.netlify.com

packages/wonder-blocks-toolbar/components/toolbar.md
```diff
@@ -11,40 +11,34 @@ const secondaryButton = <Button size="small" kind="secondary">Secondary button</
...
-const closeButton = (
-    <svg role="img" width="24px" height="24px">
-        <path d={dismissIcon} />
-    </svg>
-);
+const Color = require("@khanacademy/wonder-blocks-color").default;
+const {icons} = require("@khanacademy/wonder-blocks-icon");
+const IconButton = require("@khanacademy/wonder-blocks-icon-button").default;
+
+const closeButton = <IconButton icon={icons.dismiss} style={{color: Color.offBlack}} />;
```
kevinbarabash: Instead of setting `style={{color: ...}}` use `kind="secondary"` (offBlack) or `kind="tertiary"`.  I think the latter fits better given the mocks of the modals.  The close button is more subtle there.

packages/wonder-blocks-toolbar/components/toolbar.md
```diff
@@ -59,30 +53,27 @@ const questionCount = <LabelMedium>7 questions</LabelMedium>;
...
-const dismissIcon =
-    "M12.003 10.588L7.707 6.293a1 1 0 0 0-1.414 1.414l4.295 4.296-4.295 4.295a1 1 0 0 0 1.414 1.414l4.296-4.295 4.295 4.295a1 1 0 0 0 1.414-1.414l-4.295-4.295 4.295-4.296a1 1 0 1 0-1.414-1.414l-4.295 4.295z";
-const closeButton = (
-    <svg role="img" width="24px" height="24px">
-        <path d={dismissIcon} />
-    </svg>
-);
+const closeButton = <IconButton icon={icons.dismiss} style={{color: Color.offBlack}} />;
 
 const goToExercise = <Link>Go to exercise</Link>;
```
kevinbarabash: This should have an `href` otherwise it becomes disabled.

scotchfield:
I tried to make the width of the `IconButton` 24px, but the focus border seems to alter the sizing of the Icon inside. Tatiana is doing a design review in https://github.com/Khan/wonder-blocks/pull/235. I also changed the padding around medium Toolbars there which addresses a design concern and helps with the spacing.

kevinbarabash:
I think we still want the focus ring to be square.  We should check with design though.  Let's pick this up in #235.

PR #237, opened 2018-07-24
Toolbar tests

scotchfield:
Adds some tests to make sure the component is receiving props as expected.


jeresig:

Built with commit 6f5fab4d8285200e86deb23b5c48c55a2c7cb4f9

https://deploy-preview-237--wonder-blocks.netlify.com

packages/wonder-blocks-toolbar/components/toolbar.test.js
```diff
@@ -0,0 +1,30 @@
+// @flow
+import * as React from "react";
+import {shallow} from "enzyme";
+
+import Toolbar from "./toolbar.js";
+
+describe("Toolbar", () => {
```
kevinbarabash: These tests seem to be very low value since we the snapshot tests from the documentation and we can look at those to verify that the left/right content props got where they needed to go.  I think think we can remove these.  In the future we'll be adding visual regression tests.  For components that are purely presentation that's where we'll get value from testing.

packages/wonder-blocks-toolbar/index.test.js
```diff
@@ -0,0 +1,13 @@
...
+describe("@khanacademy/wonder-blocks-toolbar", () => {
+    test("package exports default", async () => {
+        // Arrange
+        const importedModule = import("./index.js");
+
+        // Act
+        const result = await importedModule;
+
+        // Assert
+        expect(Object.keys(result).sort()).toEqual(["default"].sort());
```
kevinbarabash: Keep this test though.

PR #238, opened 2018-07-24
Rename dropdown to more closely match existing and base elements

wishdasher:
I originally named the components closely to what they were called in the design. I realized that those names don't match the existing components or the basic HTML elements that the components correspond to, so I'm trying to improve that.

Test plan:
The snapshot of this package doesn't change with this diff. Searched through the codebase for all instances of the names I changed.

### DIFF
--- a/packages/wonder-blocks-dropdown/components/action-menu.js
+++ b/packages/wonder-blocks-dropdown/components/action-menu.js
@@ -73,7 +73,7 @@ type ItemProps = ActionItemProps | OptionItemProps | SeparatorProps;
 
 type MenuProps = {|
     /**
-     * The items in this menu.
+     * The items in this dropdown.
      */
     items: Array<ItemProps>,
 
@@ -101,8 +101,8 @@ type MenuProps = {|
     alignment: "left" | "right",
 
     /**
-     * Whether this menu is disabled. A disabled menu may not be opened and
-     * does not support interaction. Defaults to false.
+     * Whether this component is disabled. A disabled dropdown may not be opened
+     * and does not support interaction. Defaults to false.
      */
     disabled: boolean,
 
@@ -114,7 +114,7 @@ type MenuProps = {|
 
 type State = {|
     /**
-     * Whether or not menu is open.
+     * Whether or not the dropdown is open.
      */
     open: boolean,
 |};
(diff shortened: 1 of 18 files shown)


jeresig:

Built with commit d312bf8e4d3c7b838e430d1c34d425efc2757221

https://deploy-preview-238--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/index.js
```diff
@@ -1,6 +1,6 @@
...
-// import SingleSelectMenu from "./components/single-select-menu.js";
-// import MultiSelectMenu from "./components/multi-select-menu.js";
-//
-// export {ActionMenu, SingleSelectMenu, MultiSelectMenu};
+// @flow
+import ActionMenu from "./components/action-menu.js";
+import SingleSelect from "./components/single-select.js";
+import MultiSelect from "./components/multi-select.js";
+
+export {ActionMenu, SingleSelect, MultiSelect};
```
kevinbarabash: Yay!  Can you add a test for the exports of this file similar to what we do in https://github.com/Khan/wonder-blocks/blob/master/packages/wonder-blocks-typography/index.test.js?

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -180,22 +177,27 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
         const opener = (
             <ActionMenuOpener
                 disabled={disabled}
-                onClick={() => this.toggleMenu()}
+                onClick={() => this.handleOpenChanged(!open)}
+                ref={(node) =>
+                    (this.openerElement = ((ReactDOM.findDOMNode(
+                        node,
+                    ): any): Element))
```
jeresig: Woah! Oh Flow... 😬

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -93,15 +96,9 @@ export default class MultiSelectMenu extends React.Component<Props, State> {
...
-        this.setState((prevState) => ({
-            open: !prevState.open,
-        }));
-    }
-
-    handleClose() {
+    handleOpenChanged(open: boolean) {
         this.setState({
-            open: false,
+            open: open,
```
jeresig: Can also do: `this.setState({open});`

PR #239, opened 2018-07-25
WB-320: update IconButton to handle clientNav, fix FlowFixMes

kevinbarabash:
`IconButton` had a `clientNav` prop, but its behaviour was never implemented.  This PR implements that prop bringing `IconButton` in line with `Button`.  I would like to add some automated test for `clientNav`, but I'll try to do this in a separate PR.  This PR also removes some `$FlowFixMe`s.

### DIFF
--- a/packages/wonder-blocks-button/__snapshots__/generated-snapshot.test.js.snap
+++ b/packages/wonder-blocks-button/__snapshots__/generated-snapshot.test.js.snap
@@ -417,6 +417,7 @@ exports[`wonder-blocks-button example 3 1`] = `
   <a
     className=""
     disabled={true}
+    href="https://khanacademy.org"
     onBlur={[Function]}
     onClick={[Function]}
     onFocus={[Function]}
--- a/packages/wonder-blocks-button/__snapshots__/generated-snapshot.test.js.snap
+++ b/packages/wonder-blocks-button/__snapshots__/generated-snapshot.test.js.snap
@@ -22,6 +22,7 @@ exports[`wonder-blocks-button example 1 1`] = `
 >
   <button
     className=""
+    disabled={false}
     onBlur={[Function]}
     onClick={[Function]}
     onFocus={[Function]}
@@ -81,6 +82,7 @@ exports[`wonder-blocks-button example 1 1`] = `
   </button>
   <button
     className=""
+    disabled={false}
     onBlur={[Function]}
     onClick={[Function]}
     onFocus={[Function]}
@@ -143,6 +145,7 @@ exports[`wonder-blocks-button example 1 1`] = `
   </button>
   <button
     className=""
+    disabled={false}
     onBlur={[Function]}
     onClick={[Function]}
     onFocus={[Function]}
@@ -202,6 +205,7 @@ exports[`wonder-blocks-button example 1 1`] = `
   </button>
   <button
     className=""
+    disabled={false}
     onBlur={[Function]}
     onClick={[Function]}
     onFocus={[Function]}
@@ -346,6 +350,7 @@ exports[`wonder-blocks-button example 3 1`] = `
   <button
     aria-disabled="true"
     className=""
+    disabled={true}
     onBlur={[Function]}
     onClick={[Function]}
     onFocus={[Function]}
@@ -381,6 +386,7 @@ exports[`wonder-blocks-button example 3 1`] = `
         "textDecoration": "none",
       }
     }
+    tabIndex={-1}
   >
     <span
       className=""
@@ -441,6 +447,7 @@ exports[`wonder-blocks-button example 3 1`] = `
         "textDecoration": "none",
       }
     }
+    tabIndex={-1}
   >
     <span
       className=""
@@ -498,6 +505,7 @@ exports[`wonder-blocks-button example 4 1`] = `
       <td>
         <button
           className=""
+          disabled={false}
           onBlur={[Function]}
           onClick={[Function]}
           onFocus={[Function]}
@@ -564,6 +572,7 @@ exports[`wonder-blocks-button example 4 1`] = `
       <td>
         <button
           className=""
+          disabled={false}
           onBlur={[Function]}
           onClick={[Function]}
           onFocus={[Function]}
@@ -630,6 +639,7 @@ exports[`wonder-blocks-button example 4 1`] = `
       <td>
         <button
           className=""
+          disabled={false}
           onBlur={[Function]}
           onClick={[Function]}
           onFocus={[Function]}
@@ -714,6 +724,7 @@ exports[`wonder-blocks-button example 4 1`] = `
         >
           <button
             className=""
+            disabled={false}
             onBlur={[Function]}
             onClick={[Function]}
             onFocus={[Function]}
@@ -800,6 +811,7 @@ exports[`wonder-blocks-button example 4 1`] = `
         >
           <button
             className=""
+            disabled={false}
             onBlur={[Function]}
             onClick={[Function]}
             onFocus={[Function]}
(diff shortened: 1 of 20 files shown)


jeresig:

Built with commit a9f0c84b8ca031d8cba669fa846fa9ad0b911768

https://deploy-preview-239--wonder-blocks.netlify.com

packages/wonder-blocks-button/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -399,6 +399,7 @@ exports[`wonder-blocks-button example 3 1`] = `
   <a
     className=""
     disabled={true}
+    href="https://khanacademy.org"
```
kevinbarabash: For some reason I thought that you had to remove the `href` when disabling a link, but apparently that's not the case.

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -23,37 +23,24 @@ type Props = {|
...
-        } else {
-            return StyledButton;
-        }
-    }
-
-    getProps() {
+    render() {
         const {
+            children,
+            clientNav,
```
kevinbarabash: sorted props

packages/wonder-blocks-link/components/link-core.js
```diff
@@ -48,29 +48,21 @@ export default class LinkCore extends React.Component<Props> {
                 : (hovered || focused) && linkStyles.focus,
         ];
 
-        const props = {
-            style: [defaultStyles, style],
+        const commonProps = {
             "data-test-id": testId,
+            style: [defaultStyles, style],
```
kevinbarabash: sorted props

packages/wonder-blocks-button/__snapshots__/generated-snapshot.test.js.snap
chenyejia: Hm, if we're not removing the `href`, we should add `aria-disabled` like @somewhatabstract mentioned.

kevinbarabash:
@chenyejia thanks for reviewing.  I would've expected you to be a blocking reviewer for these changes.  I'm going to have to read the CODEOWNER docs more closely and try to figure out what's up.

packages/wonder-blocks-button/__snapshots__/generated-snapshot.test.js.snap
somewhatabstract: Yes, we shouldn't be setting `disabled`. Instead, links should be disabled using `tabIndex="-1"` and `aria-disabled="true"`.

kevinbarabash: @somewhatabstract I conflated `aria-disabled` and `disabled`.  Thanks for the clarification.  I'll make this change.

somewhatabstract:
I would need to check, but I think, when disabling the link for aria, you have to remove any click handler or target. You may want to test to verify that.

kevinbarabash:
> I would need to check, but I think, when disabling the link for aria, you have to remove any click handler or target. You may want to test to verify that.

We're going to have to something b/c the "hand" (pointer) cursor shows up on the disabled button that uses a link under the hood.  I think removing the target is the way to go.

packages/wonder-blocks-button/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -58,7 +58,6 @@ exports[`wonder-blocks-button example 1 1`] = `
         "textDecoration": "none",
       }
     }
-    tabIndex={0}
```
kevinbarabash: `tabIndex={0}` is the default order so we don't need to add it.

packages/wonder-blocks-button/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -397,8 +392,9 @@ exports[`wonder-blocks-button example 3 1`] = `
     </span>
   </button>
   <a
+    aria-disabled="true"
```
kevinbarabash: It's okay to have `disabled={true}` as well on buttons since you can still navigate to it via VO.  This means we'll be able to get rid fo that `e.preventDefault()` hack I added to the disabled click handlers.

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -75,11 +75,11 @@ export type ClickableHandlers = {|
     onKeyUp: (e: SyntheticKeyboardEvent<*>) => void,
     onFocus: (e: SyntheticFocusEvent<*>) => void,
     onBlur: (e: SyntheticFocusEvent<*>) => void,
-    tabIndex: number,
+    tabIndex?: number,
 |};
 
 const disabledHandlers = {
-    onClick: () => void 0,
+    onClick: (e) => e.preventDefault(),
```
kevinbarabash: This prevents navigation on links.  The other option is to remove the `href`.   How do people feel about this change?

kevinbarabash:
> Thanks for fixing the Flow errors, too.

:+1: If I get ambitious I might set up flow-coverage to run and be uploaded to codecov so that we can monitor that as well.

packages/wonder-blocks-core/components/clickable-behavior.js
wishdasher: I think we should be removing the `href` because that's how browsers know not to navigate. This preventDefault feels like hidden behavior. I think adding an `onClick` to disabled items also made them focusable by mouse press -- the disabled menu items now receive a focus ring.

packages/wonder-blocks-button/__snapshots__/generated-snapshot.test.js.snap
wishdasher: I think we need this. Adding `tabindex=0` to a previously non-focusable element adds that element to the tab navigation order. Removing this caused some elements that used ClickableBehavior, such as `div` and `li`, to no longer be tabbable.

wishdasher: I think we have two options. We can leave it out here and tell users of ClickableBehavior that if the component that receives the handlers isn't naturally tabbable, they have to set a `tabIndex`. Else we can include it here and warn users that if they're wrapping an `input`/`a`/`button` type element, they need to make sure that the handlers go onto that element (otherwise they may end up with two tab stops).

kevinbarabash: Good catch!.  This raises an interesting question though about kinds of things we want to add `ClickableBehavior` to.  I'll add it back in for now.

packages/wonder-blocks-core/components/clickable-behavior.js
kevinbarabash: That's a good point about it being kind of hidden.  If you inspected the DOM the link would look like it should work but it wouldn't work and that would be confusing.

kevinbarabash:
@wishdasher thanks for the great feedback.  I'll fix these things Monday.

packages/wonder-blocks-button/__snapshots__/generated-snapshot.test.js.snap
kevinbarabash: See https://github.com/Khan/wonder-blocks/pull/251.

packages/wonder-blocks-core/components/clickable-behavior.js
kevinbarabash: See https://github.com/Khan/wonder-blocks/pull/251.

PR #243, opened 2018-07-26
WB-180: Give toolbar titles and modals an optional unique id prop

scotchfield:
Rather than enforce a static id value for Toolbar and StandardModal, provide an optional prop that can be used to pass a new string. This means that if we have multiple modals on one page, they can be referenced individually.


jeresig:

Built with commit c4f7539243c8d76916073d58820d2a6fd8ef12ba

https://deploy-preview-243--wonder-blocks.netlify.com

kevinbarabash:
Please re-open after updating to use `UniqueIdProvider` or open a new PR.

PR #244, opened 2018-07-26
WB-312: Adds font-family fallbacks for Lato and Inconsolata

scotchfield:
The fallback from Inconsolata matches Google's suggestion from https://fonts.google.com/specimen/Inconsolata?selection.family=Inconsolata.


jeresig:

Built with commit 807ac6bca2837c81d1e9a6d98a38a73a1c99bbe0

https://deploy-preview-244--wonder-blocks.netlify.com

packages/wonder-blocks-typography/util/styles.js
```diff
@@ -15,7 +15,7 @@ const common = {
 const styles = StyleSheet.create({
     Title: {
         ...common,
-        fontFamily: "Lato",
+        fontFamily: ["Lato", "sans-serif"],
```
jeresig: It feels like there's is a lot of repetition in here, but I think this is fine for now.

PR #245, opened 2018-07-26
WIP - Put dropdown in portal

wishdasher:
Put the dropdown menus in portals. Now they are positioned using `react-popper`.

This updates enzyme as well because of a bug introduced by using `react-popper` and the new React 16 portals.

Code coverage decrease is due to copying util files over from the tooltip package but not the identical tests.


packages/wonder-blocks-dropdown/components/action-item.js
```diff
@@ -173,6 +173,7 @@ const styles = StyleSheet.create({
         whiteSpace: "nowrap",
         overflow: "hidden",
         textOverflow: "ellipsis",
+        textAlign: "left",
```
wishdasher: To fix a small visual bug related to ellipsis

jeresig:

Built with commit a4118fd18b515628a0439cfa158661d3a70a2878

https://deploy-preview-245--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -93,30 +144,78 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+        // If we are not in a modal, we use body as the location to portal to.
+        const modalHost = maybeGetPortalMountedModalHostElement(openerElement) || document.querySelector("body");
+
+        if (modalHost) {
+            return ReactDOM.createPortal(
+                <Popper
+                    referenceElement={this.props.openerElement}
+                    // TODO(sophie): allow dropdown to be a dropup (rise-up?) --
+                    // to appear above the dropdown opener
+                    placement={alignment === "left" ? "bottom-start" : "bottom-end"}
```
wishdasher: My impression is that `bottom-end` is supposed to right-align the elements, but instead the menu starts where the opener ends. So I move it back left by the width of the menu by adding another translate via `getTranslatedPosition`.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -3,16 +3,20 @@
...
 import Color, {fade} from "@khanacademy/wonder-blocks-color";
+import {maybeGetPortalMountedModalHostElement} from "@khanacademy/wonder-blocks-modal";
 import Spacing from "@khanacademy/wonder-blocks-spacing";
 import {View} from "@khanacademy/wonder-blocks-core";
 
 import typeof ActionItem from "./action-item.js";
 import typeof SelectItem from "./select-item.js";
 import typeof SeparatorItem from "./separator-item.js";
 
+import visibilityModifierDefaultConfig from "../../wonder-blocks-tooltip/util/visibility-modifier.js";
```
wishdasher: This will probably be used by anything that uses popper.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -3,16 +3,20 @@
 
 import * as React from "react";
 import ReactDOM from "react-dom";
+import {Popper} from 'react-popper';
```
jeresig: Double-quotes!

packages/wonder-blocks-dropdown/components/dropdown-core.js
jeresig: 🛑 Oof, yeah, this isn't great. It feels like something that we should probably copy over or put in a shared location. I think for now copying it over and leaving comments on both of them saying we should find a way to re-use them makes sense.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -65,22 +79,59 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
     }
 
     componentDidMount() {
+        this.updateEventListeners();
+    }
+
+    componentDidUpdate(prevProps: DropdownCoreProps) {
+        if (prevProps.open !== this.props.open) {
```
jeresig: You have a few props in this method, might be good to do: `const {open, alignment} = this.props;` to save a few characters.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -65,22 +79,59 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
         document.removeEventListener("touchend", this._handleInteract);
         document.removeEventListener("keyup", this._handleKeyup);
     }
 
     _handleInteract = (event: {target: any}) => {
-        if (this.node && this.node.contains(event.target)) {
-            return;
-        }
-        if (this.props.open) {
+        const target: Node = event.target;
```
jeresig: Here, too, with props. I roughly keep the rule of "if more than one prop is used, then do destructuring to get the variables".

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -93,30 +144,78 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
         }
     };
 
+    getTranslatedPosition() {
+        if (this.props.alignment === "right" ) {
```
jeresig: Extra space? Is Prettier running?

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -65,22 +79,59 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+        this.updateEventListeners();
+    }
+
+    componentDidUpdate(prevProps: DropdownCoreProps) {
+        if (prevProps.open !== this.props.open) {
+            this.updateEventListeners();
+        }
+
+        // If the menu just changed from closed to open, and if we haven't
+        // found the correct menuWidth yet. Only applies to right aligned.
```
jeresig: I'm not enthused by this, but at least we only need it for this one case.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -65,22 +79,59 @@ export default class DropdownCore extends React.Component<DropdownCoreProps> {
...
+        if (this.props.alignment === "right" && !prevProps.open && this.props.open && !this.menuWidth) {
+            // There is a timeout because if we call getBoundingClientRect
+            // right away, it returns the value of the document width instead
+            // of the width of the just-mounted component, the menu.
+            requestAnimationFrame(() => {
+                if (this.menuElement) {
```
jeresig: It seems like this check can be moved above to avoid calling this function in the first place, potentially?

packages/wonder-blocks-dropdown/components/dropdown-core.js
jeresig: @somewhatabstract did you encounter this issue with the tooltips? It seems like an odd case!

packages/wonder-blocks-dropdown/components/multi-select-menu.md
```diff
@@ -87,15 +73,14 @@ const {StyleSheet} = require("aphrodite");
...
-        height: 350,
     },
 });
 
 class ExampleWithShortcuts extends React.Component {
     constructor() {
         super();
         this.state = {
-            selectedValues: ["1"],
+            selectedValues: ["wonderblocks 4ever"],
```
jeresig: hehehe

packages/wonder-blocks-dropdown/components/dropdown-core.js
somewhatabstract: I think there's a misinterpretation of the values happening here. That certainly shouldn't be necessary. Instead, there's a `flip` modifier that auto detects the position based off the bounds element. There should be no need to do all this extra work.

packages/wonder-blocks-dropdown/components/dropdown-core.js
somewhatabstract: Agreed. If this gets used outside of tooltip we should consider having a `wonder-blocks-popper` package or consider putting this stuff into the `wonder-blocks-layout` package to share it

packages/wonder-blocks-dropdown/components/dropdown-core.js
wishdasher: The check is to satisfy Flow, but in this method `this.menuElement` should be defined.

wishdasher: Since the context of the callback is different, Flow doesn't allow me to move it above.

packages/wonder-blocks-dropdown/components/dropdown-core.js
wishdasher: Turns out I was wrongly positioning something using `absolute` when I no longer needed to! So I was able to remove all this calculation stuff. 😁

packages/wonder-blocks-dropdown/components/dropdown-core.js
wishdasher: I figured it out -- I was positioning a child element with `absolute` when I didn't need to. Thanks for helping me find my error!

packages/wonder-blocks-dropdown/components/dropdown-core.js
wishdasher: Got rid of the other props here ;D, so leaving this.

PR #246, opened 2018-07-27
[feature-modal] WB-180: Give modals a unique id with UniqueIDProvider

scotchfield:
Provide an optional prop in `Toolbar` and use `UniqueIDProvider` in `StandardModal` to make sure the title ID is unique.


jeresig:

Built with commit c393b865c4320491ef3ed242b83c7c08819fe0d3

https://deploy-preview-246--wonder-blocks.netlify.com

packages/wonder-blocks-modal/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -539,7 +539,7 @@ exports[`wonder-blocks-modal example 2 1`] = `
             >
               <h4
                 className=""
-                id="wb-toolbar-title"
+                id="uid--0-wb-modal-title"
```
kevinbarabash: I love that these are both unique and stable.

packages/wonder-blocks-modal/components/standard-modal.js
```diff
@@ -114,6 +126,14 @@ export default class StandardModal extends React.Component<Props> {
             </ModalDialog>
         );
     }
+
+    render() {
+        return (
+            <UniqueIDProvider>
```
kevinbarabash: We may actually want to set `mockOnFirstRender={true}` for the use case of rendering modals on page load.

I just realized that for the other modals we have `aria-labelledby="wb-modal-title"` but we haven't set the id on the title inside the modal.  This case is more interesting b/c the `<Title>` element is user specified.  I wonder if we could use the Context API in React 16 and have the modal provide a unique id for the title and then the `Title` component could use a context consumer to grab that id.

cc @somewhatabstract

somewhatabstract: I would rather we did something less magical. Why not allow the user to specify the id prop for the title on the modal props whenever why provide a `Title`?

scotchfield: @kevinbarabash Maybe you could comment on this? I think you two have more context about the use-cases here. We could also provide an optional prop so a user could specify the id, and assign a unique one if not, but maybe that's too fancy.

kevinbarabash: A `titleId` prop is probably a good start place.  I worry about how we might lint for this.  If we can't lint for it then maybe we'll have to resort to magic.

somewhatabstract: We have a similar need in the `Tooltip` component, so perhaps we can follow the same pattern from there to be consistent?

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -49,6 +49,12 @@ type Props = {|
      * The main title rendered in larger bold text.
      */
     title?: string,
+
+    /**
+     * An optional id to provide a selector for the title element. Uses
+     * "wb-toolbar-title" as a default if titleId is not specified.
```
somewhatabstract: We should follow the pattern used for the `Tooltip` to be consistent.

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -107,7 +114,7 @@ export default class Toolbar extends React.Component<Props> {
                                 sharedStyles.center,
                             ]}
                         >
-                            <TitleComponent id="wb-toolbar-title">
+                            <TitleComponent id={titleId || "wb-toolbar-title"}>
```
somewhatabstract: IDs have to be unique so we certainly shouldn't hard code a fixed string as that would not by WCAG compliant.

kevinbarabash:
As part of supporting `OneColumnModal` and `TwoColumnModal` we should also update the examples and documentation with whatever solution we settle on, but let's settle on a solution first.

kevinbarabash:
I'm closing this b/c it's stale.  @jandrade hopefully we can do this as part of the modal v2 work. 🤞

PR #247, opened 2018-07-27
Update clickable components to retain focus ring in active state

wishdasher:
* Changed SelectOpener to use `box-shadow` method instead of `:after`
* Fixed bug where placeholder styles weren't being applied for SelectOpener
* Added disabled IconButton example (like I promised to do at some point)
* Added examples for IconButton, Button, Link on dark backgrounds


packages/wonder-blocks-dropdown/components/select-opener.js
```diff
@@ -222,13 +217,12 @@ const _generateStyles = (light, hovered, focused, pressed) => {
             active: {
                 background: mix(fade(blue, 0.32), white),
                 borderColor: mix(offBlack32, blue),
-                borderStyle: "solid",
-                borderWidth: 1,
+                borderWidth: 2,
+                paddingLeft: 15,
```
jeresig: 🤷‍♂️ FWIW, both here and above it seems like it'd be better to use our standard spacing and then do `- 1` (and including the comment above). Not high priority but makes the number a bit clearer.

packages/wonder-blocks-form/components/radio-core.js
```diff
@@ -23,8 +23,7 @@ const StyledInput = addStyle("input");
 
 /**
  * The internal stateless 🔘 Radio button
- */
-export default class RadioCore extends React.Component<Props> {
+ */ export default class RadioCore extends React.Component<Props> {
```
jeresig: Aieee - the formatting issue is back!

packages/wonder-blocks-link/components/link.md
```diff
@@ -1,8 +1,20 @@
 Link example:
 
 ```js
-<p>
-    Lorem ipsum <Link href="#nonexistent-link">Link</Link> dolor sit amet,
-    consectetur <Link href="#">Visited Link</Link> adipiscing elit
-</p>
+
+<React.Fragment>
```
jeresig: Hmm, I think we preferred wrapping everything in `<View>`s, instead.

PR #249, opened 2018-07-28
add flow-coverage script, add aphrodite flow types

kevinbarabash:
Adding aphrodite.flow.js bumped our flow coverage from 85% to 91%.  This was computed using the `flow-coverage-report` npm package.  It doesn't support lcov yet so we can't upload the reports to codecov.  See https://github.com/rpl/flow-coverage-report/issues/67.  In the meantime you can run it manually with `yarn run flow-coverage`.

TODO:
- [x] add `flowtype/no-types-missing-file-annotation` eslint rule
- [x] add `// @flow` to files missing it
- [x] fix new `flow` errors.


packages/wonder-blocks-dropdown/components/multi-select.md
```diff
@@ -15,9 +15,6 @@ const styles = StyleSheet.create({
     row: {
         flexDirection: "row",
     },
-    strutLike: {
```
jeresig: Yay for dead code removal!

packages/wonder-blocks-form/components/choice-internal.js
```diff
@@ -53,14 +53,28 @@ type Props = {|
...
+
 /**
  * This is a labeled 🔘 or ☑️ item. This is an internal component that's wrapped
  * by ChoiceField or Choice. Choice should be used in a CheckboxGroup or in a
  * RadioGroup. ChoiceField is the variant used outside of such a group. The two
  * are different to allow for more explicit flow typing. Choice has many of its
  * props auto-populated, but ChoiceField does not.
- */
-export default class ChoiceInternal extends React.Component<Props> {
+ */ export default class ChoiceInternal extends React.Component<Props> {
```
jeresig: This is the weird spacing issue that @wishdasher was seeing with Prettier, would be good to fix this! Maybe we need to upgrade Prettier?

package.json
```diff
@@ -29,6 +30,10 @@
   "author": "",
   "license": "MIT",
   "devDependencies": {
+    "//": [
```
jeresig: 🛑 Hmm - this got moved. Maybe you could move it out of `devDependencies` and update to specifically reference `enzyme-adapter-react-16`?

kevinbarabash: Thanks for noticing this.  I guess I'll avoid running `yarn add` in the future.

kevinbarabash: I guess if we follow the same pattern as the package.json in webapp this shouldn't be an issue.

packages/wonder-blocks-form/components/choice-internal.js
kevinbarabash: Updating prettier does not resolve this issue.

kevinbarabash: It seems like this would be pretty easy to fix, just add a rule that states code should not appear on a line after a block comment.

kevinbarabash: I'm going to leave this as is otherwise we'll end up with this change the next time someone edits this file.

kevinbarabash:
> 85.13% (-0.74%)

😭

PR #250, opened 2018-07-29
avoid unnecessary renders by pre-binding all callbacks

kevinbarabash:
Apparently arrow functions are bad b/c each render generates a new function which means the props aren't equal even if the callback is functionally the same.  This PR pre-binds all backs in non-test code and adds a lint rule to check for this going forward.

https://medium.freecodecamp.org/why-arrow-functions-and-bind-in-reacts-render-are-problematic-f1c08b060e36

### DIFF
--- a/packages/wonder-blocks-form/components/choice-internal.js
+++ b/packages/wonder-blocks-form/components/choice-internal.js
@@ -59,8 +59,7 @@ type Props = {|
  * RadioGroup. ChoiceField is the variant used outside of such a group. The two
  * are different to allow for more explicit flow typing. Choice has many of its
  * props auto-populated, but ChoiceField does not.
- */
-export default class ChoiceInternal extends React.Component<Props> {
+ */ export default class ChoiceInternal extends React.Component<Props> {
     static defaultProps = {
         checked: false,
         disabled: false,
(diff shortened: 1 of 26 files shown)


jeresig:

Built with commit 172a4e511a476e56b23693f76e8fb34b07e57815

https://deploy-preview-250--wonder-blocks.netlify.com

kevinbarabash:
We should run eslint after we run prettier in the pre-commit hook.

kevinbarabash:
Reading the article a bit more closely, it appears that this only matters if we were extending `PureComponent` instead of `Component` or if we were using `shouldComponentUpdate`.  We should probably do some investigation around the performance implications of using `PureComponent` or `shouldComponentUpdate`.

kevinbarabash:
Apparently the old context API didn't work with PureComponent so we should probably update to using the new Context API before doing anything with PureComponent.

packages/wonder-blocks-form/components/choice-internal.js
```diff
@@ -74,19 +73,36 @@ export default class ChoiceInternal extends React.Component<Props> {
...
+    };
+    handleLabelClick = (e: SyntheticEvent<>) => {
+        e.preventDefault();
+    };
     render() {
         const {
             label,
             description,
+            // we don't need this to go into coreProps
+            // eslint-disable-next-line no-unused-vars
```
jeresig: Heya, while we're at it let's turn on this option so we don't have to do this anymore!

https://eslint.org/docs/rules/no-unused-vars#ignorerestsiblings

kevinbarabash: This sounds really good.

kevinbarabash:
@jeresig I think this PR should be blocked until we update to using the Context API otherwise things will mysteriously break for anyone use `PureComponent` which people may want to do now that we're pre-binding callbacks.

packages/wonder-blocks-form/components/choice-internal.js
kevinbarabash: I'll open a separate PR for this rule since it's really independent of the rest of the changes in this one.

jeresig: Yep, that's fine by me!

kevinbarabash:
I'm going to close this for now and will open new PR that does this when I have time to see it through.

PR #251, opened 2018-07-30
allow all clickable components can be tabbed to

kevinbarabash:
I removed `tabIndex={0}` from components in a previous PR thinking that it wasn't needed b/c that's the default behaviour for buttons and anchors.  Then @wishdasher reminded me that we might want to use clickable behaviour for other things that aren't focusable by default so I've added it back in along with a comment explaining why it's there.  I also noticed some `$FlowFixMe`s in action-item.js so I cleaned those up as well.


jeresig:

Built with commit 40a45d9f5019e43b423945b32710e27f7457d7b3

https://deploy-preview-251--wonder-blocks.netlify.com

packages/wonder-blocks-icon-button/components/icon-button-core.js
```diff
@@ -102,12 +102,13 @@ export default class IconButtonCore extends React.Component<Props> {
...
         if (href) {
+            const actualHref = !disabled ? href : undefined;
             return clientNav ? (
-                <StyledLink {...commonProps} to={href}>
+                <StyledLink {...commonProps} to={actualHref}>
                     {child}
                 </StyledLink>
             ) : (
-                <StyledAnchor {...commonProps} href={href}>
+                <StyledAnchor {...commonProps} href={actualHref}>
```
jeresig: 🛑 You're doing this in a few places now and I'm not sure this is right... we'll end up with `<a>`s that have no `href` - which doesn't seem to be the intended effect? An anchor without a `href` behaves very differently (as it's meant to represent an anchor point in a page to jump to). If we don't want the link to do anything when clicked we either need to do `preventDefault` (which is what I think we had before) or turn the link into something else like `javascript: void 0`

kevinbarabash: I've change this to use `preventDefault` but on for anchors.

kevinbarabash: > An anchor without a href behaves very differently (as it's meant to represent an anchor point in a page to jump to)

The main change I saw in behaviour was that the cursor doesn't change to the "pointer" on mouseover.  What are the other changes?

packages/wonder-blocks-icon-button/components/icon-button-core.js
```diff
@@ -104,11 +110,19 @@ export default class IconButtonCore extends React.Component<Props> {
...
+                    to={actualHref}
+                >
                     {child}
                 </StyledLink>
             ) : (
-                <StyledAnchor {...commonProps} href={actualHref}>
+                <StyledAnchor
+                    {...commonProps}
+                    onClick={this.handleClick}
+                    href={actualHref}
```
jeresig: 🛑 At least for `StyledAnchor` we always want to pass in `href` (not `actualHref`). I'm not sure if that's the case for `StyledLink` as well, maybe it is? Now that there is the `handleClick` with the `preventDefault` it seems like we can just remove `actualHref` and use `href`, correct?

kevinbarabash: 😞 whoops, I'll fix this.

wishdasher:
Thank you so much for taking care of ActionItem as well!

PR #252, opened 2018-07-30
don't package deps

kevinbarabash:
I've added some deps, but it looks like we're still including popper related deps in the dist bundles for the dropdown and tooltip components.  Further investigation is required.


jeresig:

Built with commit f4a33de0da909be551e050aa1735b26d5b30e2cb

https://deploy-preview-252--wonder-blocks.netlify.com

kevinbarabash:
It looks like webpack.config.js is marking all `dependencies` as `externals`.  I'm not sure why `react` and `react-dom` aren't getting bundled.  This definitely warrants further investigation.

kevinbarabash:
I see... we're also looking at `dependencies` in the root `package.json`.  That's weird.  I'm going to update this to look at `peerDependencies` in each sub-package instead.

PR #253, opened 2018-07-30
rework Button docs to include a Best Practices section

kevinbarabash:
The key changes:
- there's more regularity between the main examples (each focuses on a particular prop)
- the best practices section covers issues that people might run into an common solutions as well as best practices
- I also changed the `size` prop to take `"small" | "medium" (default)` instead of `"small" | "default" (default)` so that it would mesh better with our other `size` props.

### DIFF
--- a/packages/wonder-blocks-button/components/button.md
+++ b/packages/wonder-blocks-button/components/button.md
@@ -16,20 +16,20 @@ const styles = StyleSheet.create({
 <View style={styles.row}>
     <Button
         style={styles.button}
-        onClick={(e) => console.log("Hello, world!")}
+        onClick={(e) => window.alert("Hello, world!")}
     >
         Primary
     </Button>
     <Button
         style={styles.button}
-        onClick={(e) => console.log("Hello, world!")}
+        onClick={(e) => window.alert("Hello, world!")}
         kind="secondary"
     >
         Secondary
     </Button>
     <Button
         style={styles.button}
-        onClick={(e) => console.log("Hello, world!")}
+        onClick={(e) => window.alert("Hello, world!")}
         kind="tertiary"
     >
         Tertiary
@@ -37,7 +37,8 @@ const styles = StyleSheet.create({
 </View>
 ```
 
-Buttons have a `color` that is either `"default"` (default) or `"destructive"`:
+Buttons have a `color` that is either `"default"` (the default, as shown
+above) or `"destructive"` (as can seen below):
 ```jsx
 const {View} = require("@khanacademy/wonder-blocks-core");
 const {StyleSheet} = require("aphrodite");
@@ -54,14 +55,14 @@ const styles = StyleSheet.create({
 <View style={styles.row}>
     <Button
         style={styles.button}
-        onClick={(e) => console.log("Hello, world!")}
+        onClick={(e) => window.alert("Hello, world!")}
         color="destructive"
     >
         Primary
     </Button>
     <Button
         style={styles.button}
-        onClick={(e) => console.log("Hello, world!")}
+        onClick={(e) => window.alert("Hello, world!")}
         kind="secondary"
         color="destructive"
     >
@@ -69,7 +70,7 @@ const styles = StyleSheet.create({
     </Button>
     <Button
         style={styles.button}
-        onClick={(e) => console.log("Hello, world!")}
+        onClick={(e) => window.alert("Hello, world!")}
         kind="tertiary"
         color="destructive"
     >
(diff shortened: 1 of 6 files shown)


jeresig:

Built with commit e4046dd33e1373e7192bf47f79061ad9045ede35

https://deploy-preview-253--wonder-blocks.netlify.com

packages/wonder-blocks-button/components/button.md
```diff
@@ -93,171 +141,280 @@ const styles = StyleSheet.create({
...
+        </Button>
+    </View>
+</View>
+```
+
+Layouts often specify a specific width of button.  When implementing such
+designs use `minWidth` instead of `width`.  It prevents a button from
+resizing to fit content.  This is important for international sites since
+sometimes strings for UI elements can be much longer than their English
+counterparts.
```
chenyejia: I'm confused by this part. Is "It prevents a button..." referring to `minWidth` or `width`? Is `badButton` supposed to be with `minWidth`?

packages/wonder-blocks-button/components/button.md
```diff
@@ -1,78 +1,126 @@
...
 </View>
 ```
 
-Button example, `href="#button-1"`:
-```js
-<Button
-    onClick={(e) => console.log("Hello, world!")}
-    href="#button-1"
->Label</Button>
+Buttons have a `color` that is either `"default"` (default) or `"destructive"`:
```
jeresig: Might be clearer to say something like `(the default, as shown above) or "destructive" (as can be seen below):`

packages/wonder-blocks-button/components/button.md
```diff
@@ -83,181 +131,325 @@ const styles = StyleSheet.create({
...
-    >Disabled</Button>
+        disabled={true}
+    >
+        Tertiary
+    </Button>
 </View>
 ```
 
-Button examples, `style` specified (Supports width, position, margin, and flex styles):
+Buttons have a `size` that's either `"medium"` (default) or `"small"`.
```
jeresig: Would be good to clarify that the following buttons are all "small".

packages/wonder-blocks-button/components/button.md
```diff
@@ -83,181 +131,325 @@ const styles = StyleSheet.create({
...
+<View style={styles.row}>
+    <Button
+        href="#button-1"
+        style={styles.button}
+    >
+        href
+    </Button>
+    <Button
+        kind="secondary"
+        onClick={(e) => console.log("Hello, world!")}
```
jeresig: When I clicked these I was kind of surprised when something didn't happen. Maybe an `alert` might be more obvious?

packages/wonder-blocks-button/components/button.md
```diff
@@ -1,78 +1,126 @@
...
-    >Button with href</Button>
+    >
+        Tertiary
+    </Button>
 </View>
 ```
 
-Buttons on a dark blue background:
-```js
+Buttons on a `darkBlue` background should be `light`.
```
jeresig: Maybe saying "should set `light` to `true`" would make it clearer.

packages/wonder-blocks-button/components/button.md
```diff
@@ -83,181 +131,325 @@ const styles = StyleSheet.create({
...
+            </Route>
+        </Switch>
+    </View>
+</MemoryRouter>
+```
+
+Buttons can have a `style` props which supports width, position, margin,
+and flex styles:
+
+TODO(kevinb): link to front matter section describing the `style` prop in
```
jeresig: I wish there was a way to include non-visible comments. It makes for a very awkward documentation reading experience. I think it'd be better to put these into Jira issues instead.

packages/wonder-blocks-button/components/button.md
```diff
@@ -83,181 +131,325 @@ const styles = StyleSheet.create({
...
+            label
+        </Button>
+        <Button
+            style={styles.goodButton}
+        >
+            label in a different language
+        </Button>
+    </View>
+    <View style={styles.gap} />
+    <View style={styles.row}>
```
jeresig: I think it'd be better to just remove the bad example from this as it makes it confusing as to which is the right one (or we should add text explaining which set of buttons are using which technique).

packages/wonder-blocks-button/components/button.md
```diff
@@ -83,181 +131,325 @@ const styles = StyleSheet.create({
...
+        <Button
+            style={styles.badButton}
+        >
+            Primary
+        </Button>
+    </View>
+</View>
 ```
+
+TODO(kevinb): add section on media queries.
```
jeresig: (Same goes for all of the in-docs TODOs)

packages/wonder-blocks-button/components/button.md
kevinbarabash: Will do.  I should note that this is a breaking change.  Should I bump button to `v2`?

jeresig: Yep, I think so!

kevinbarabash:
@jeresig could you have another look?

PR #254, opened 2018-07-31
Remove inner border from button elements for Firefox

chenyejia:
Firefox adds an additional inner focus ring for buttons. Update `Button` and `IconButton` styles so that the inner focus ring is not shown.

Test plan:
Used keyboard navigation to navigate to `Button` and `IconButton` examples on styleguidist and confirm that the inner focus ring does not show up.

### DIFF
--- a/packages/wonder-blocks-button/components/button-core.js
+++ b/packages/wonder-blocks-button/components/button-core.js
@@ -125,6 +125,9 @@ const sharedStyles = StyleSheet.create({
         outline: "none",
         textDecoration: "none",
         boxSizing: "border-box",
+        "::-moz-focus-inner": {
+            border: 0,
+        },
     },
     disabled: {
         cursor: "auto",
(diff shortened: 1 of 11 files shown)


jeresig:

Built with commit 93cfa6da635336c9815aa89dc7a2aa717a177d99

https://deploy-preview-254--wonder-blocks.netlify.com

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -125,6 +125,9 @@ const sharedStyles = StyleSheet.create({
         outline: "none",
         textDecoration: "none",
         boxSizing: "border-box",
+        "::-moz-focus-inner": {
```
jeresig: Would be good to add a brief comment explaining what this does and why we added it (so that we don't accidentally remove it later!)

PR #255, opened 2018-07-31
Refactor dropdown package

wishdasher:
* Move {ActionItem, OptionItem, SeparatorItem} to be children of the menus instead of as an `items` prop -- this brings the design closer to the one I used for Choice and removed the need for different but duplicate props definitions
* Distinguish additional opener styling from dropdown styling
* Move the check and checkbox out into their own file (and cleaned up the styling logic) -- this made `option-item.js` cleaner


jeresig:

Built with commit 44b333ad6881166eccde7d062eaa6def83f269c8

https://deploy-preview-255--wonder-blocks.netlify.com

kevinbarabash:
@wishdasher cool!  I'll have a look tomorrow.

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -162,14 +165,45 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
...
+                    key: index,
+                    onToggle: (value, state) =>
+                        this.handleSelected(value, state),
+                    selected: selectedValues
+                        ? selectedValues.includes(item.props.value)
+                        : false,
+                    variant: "check",
+                });
+            } else {
+                return React.cloneElement(item, {
```
jeresig: If you don't then you can just `return item;` here.

packages/wonder-blocks-dropdown/components/single-select.js
```diff
@@ -50,9 +49,14 @@ type Props = {|
     light: boolean,
 
     /**
```
jeresig: I'm mixed on the additional style props, but I'm struggling to think of a better option, so I think this is fine!

packages/wonder-blocks-dropdown/util/types.js
```diff
@@ -1,31 +0,0 @@
-// @flow
```
jeresig: Woohoo!

packages/wonder-blocks-dropdown/components/check-icons.js
```diff
@@ -0,0 +1,131 @@
...
+
+type CheckProps = {|
+    disabled: boolean,
+    selected: boolean,
+    pressed: boolean,
+    hovered: boolean,
+    focused: boolean,
+|};
+
+export const Check = (props: CheckProps) => {
```
jeresig: 🛑 Hmm - I think I'd prefer having these be split into two files (we've been good about keeping 1 React component per file, exported using `export default class ...`). It doesn't look like there is very much shared between these two components, save for `CheckProps` and I think it's fine to duplicate that.

packages/wonder-blocks-dropdown/components/check-icons.js
```diff
@@ -0,0 +1,131 @@
...
+    white,
+    offBlack,
+    offBlack16,
+    offBlack32,
+    offBlack50,
+    offWhite,
+} = Color;
+
+type CheckProps = {|
+    disabled: boolean,
```
jeresig: Would be good to add explicit comments to each of these to explain them `/** ... */`

packages/wonder-blocks-dropdown/components/check-icons.js
```diff
@@ -0,0 +1,131 @@
+// @flow
+
+/**
```
jeresig: The comment should be placed above the `export default class ...` so that it gets picked up by Styleguidist.

styleguide.config.js
```diff
@@ -27,7 +27,11 @@ module.exports = {
...
         {
             name: "Dropdown",
+            content: "packages/wonder-blocks-dropdown/docs.md",
             components: [
+                "packages/wonder-blocks-dropdown/components/action-item.js",
+                "packages/wonder-blocks-dropdown/components/separator-item.js",
+                "packages/wonder-blocks-dropdown/components/option-item.js",
                 "packages/wonder-blocks-dropdown/components/action-menu.js",
                 "packages/wonder-blocks-dropdown/components/single-select.js",
                 "packages/wonder-blocks-dropdown/components/multi-select.js",
```
jeresig: Would be worth having an "Internal Components" section here with Check/Checkbox, as well? (I'm not sure if there are any other internal components but maybe they would be included, too.)

wishdasher: I added the internal Dropdown component but not Check/Checkbox. The main reason I took those out was to clean up the `option-item.js` file, and I feel like they don't really serve as functional components that are worth pointing out because it's a bunch of styling that is very dependent on the OptionItem design anyways.

packages/wonder-blocks-dropdown/components/action-menu.md
```diff
@@ -26,43 +26,19 @@ const styles = StyleSheet.create({
...
-                }
-            ]}
-            menuText="Betsy Appleseed"
             alignment="right"
-        />
+            menuText="Betsy Appleseed"
+        >
+            <ActionItem label="Profile" href="http://khanacademy.org/profile" />
+            <ActionItem label="Teacher dashboard" href="http://khanacademy.org/coach/dashboard" />
+            <ActionItem label="Settings (onClick)" onClick={() => console.log("user clicked on settings")} />
```
kevinbarabash: This is a lot more compact and readable!

packages/wonder-blocks-dropdown/components/action-menu.md
```diff
@@ -100,49 +76,23 @@ class HybridMenu extends React.Component {
...
+            <OptionItem label="Show homework assignments" value="homework"
+                onClick={(state) => console.log(`Show homework assignments ${(!state).toString()}`)} />
+            <OptionItem label="Show in-class assignments" value="in-class"
+                onClick={(state) => console.log(`Show in-class assignments ${(!state).toString()}`)} />
+        </ActionMenu>
     }
 }
 
-<View style={[styles.row]}>
+<View style={styles.row}>
```
kevinbarabash: nice

packages/wonder-blocks-dropdown/components/checkbox.js
```diff
@@ -0,0 +1,111 @@
...
+        : selected && !clickInteraction
+            ? blue
+            : white;
+    const checkColor = disabled
+        ? offBlack32
+        : clickInteraction
+            ? pressed
+                ? activeBlue
+                : blue
+            : white;
```
kevinbarabash: Oh wow.  I know people don't like nested ternaries but I'm okay with them as long as they're formatted like this.  Thanks prettier.

packages/wonder-blocks-dropdown/components/checkbox.js
```diff
@@ -0,0 +1,111 @@
...
+            ]}
+        >
+            {selected && (
+                <Icon
+                    icon={checkboxCheck}
+                    size="small"
+                    color={checkColor}
+                    style={[
+                        disabled && selected && styles.disabledCheckFormatting,
+                    ]}
```
kevinbarabash: super nit: parens aren't necessary here

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -170,7 +169,10 @@ export default class Dropdown extends React.Component<DropdownProps> {
                     }
                     modifiers={{
                         wbVisibility: visibilityModifierDefaultConfig,
-                        preventOverflow: {boundariesElement: "viewport"},
+                        preventOverflow: {
+                            boundariesElement: "viewport",
+                            escapeWithReference: true,
```
kevinbarabash: What does this do?

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -174,41 +157,69 @@ export default class MultiSelect extends React.Component<Props, State> {
...
-                    value={item.value}
-                    variant="checkbox"
-                />
-            );
+    getShortcuts(): Array<
+        React.Element<typeof ActionItem | typeof SeparatorItem>,
+    > {
+        const {children, selectedValues, shortcuts} = this.props;
+        const numOptions = React.Children.count(children);
+        // TODO(sophie): translate for i18n
```
kevinbarabash: This almost feels like a STOPSHIP... we need to figure this out soon otherwise people will start using them and then we'll get complaints about English on i18n sites.

kevinbarabash: I think there may be another English string somewhere.  It would be nice to have a linter to find them.

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -67,9 +64,14 @@ type Props = {|
...
-     * Optional styling to add.
+     * Optional styling to add to the opener component.
      */
-    style?: any,
+    openerStyle?: any,
+
+    /**
+     * Optional styling to add to the dropdown wrapper.
+     */
+    dropdownStyle?: any,
```
kevinbarabash: When would we need to style the dropdown?

packages/wonder-blocks-dropdown/components/multi-select.md
```diff
@@ -140,6 +138,9 @@ const styles = StyleSheet.create({
         margin: 10,
         padding: 20,
     },
+    setWidth: {
+        width: 170,
```
kevinbarabash: We should use `minWidth` in our examples to promote best practices around `i18n`.

packages/wonder-blocks-dropdown/components/multi-select.md
```diff
@@ -159,25 +160,24 @@ class SimpleMultiSelect extends React.Component {
...
             onChange={(selectedValues) => this.handleChanges(selectedValues)}
             selectedValues={this.state.selectedValues}
             selectItemType="Great Houses"
-            style={{
-                width: 170,
-                maxWidth: 170,
-            }}
-        />;
+            openerStyle={styles.setWidth}
+            dropdownStyle={styles.setWidth}
```
kevinbarabash: I don't think we should ever be setting the width on the dropdown.

packages/wonder-blocks-dropdown/components/option-item.js
```diff
@@ -183,124 +106,74 @@ export default class OptionItem extends React.Component<OptionProps> {
...
+                            : (hovered || focused) && styles.focus,
                         disabled && styles.disabled,
-                        !disabled &&
-                            (pressed
-                                ? styles.active
-                                : (hovered || focused) && styles.focus),
                     ];
 
                     return (
-                        <StyledButton style={[defaultStyle]} {...handlers}>
```
kevinbarabash: I'm curious why the switch from `StyledButton` to `View`.

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -59,13 +59,13 @@ type DropdownProps = {|
      * Styling specific to the dropdown component that isn't part of the opener.
      */
     dropdownStyle?: any,
```
kevinbarabash: This should probably just be `style` now.

packages/wonder-blocks-dropdown/components/checkbox.js
wishdasher: Unfortunately, prettier reformats this part to use the parentheses.

packages/wonder-blocks-dropdown/components/dropdown.js
wishdasher: It keeps the anchor and popper element in the same relative position with each other. There's still some kinks in the popper positioning I'm looking at.

packages/wonder-blocks-dropdown/components/multi-select.js
wishdasher: I agree... We can potentially use props to pass in all the phrasing that we want, but we would need to be wary of phrasal ordering assumptions.

packages/wonder-blocks-dropdown/components/multi-select.md
wishdasher: Done! Wow, that was much simpler than what I was anticipating having to do and also makes a whole lot more sense 😅

packages/wonder-blocks-dropdown/components/checkbox.js
kevinbarabash: I sometimes forget that prettier is there in the background always tidying up our code.

packages/wonder-blocks-dropdown/components/option-item.js
kevinbarabash: This is fine, I was just curious. 🙂

PR #256, opened 2018-08-01
Resolve a number of modal-related issues. Fixes WB-325, WB-176, and WB-323.

jeresig:
This fixes three modal issues:

Size of modals doesn't match spec
https://khanacademy.atlassian.net/browse/WB-323

I fixed this by putting a `maxWidth: 960` on all modals (matching what was already done on the StandardModal). We can give devs the ability to override this in the future.

Not resizing properly/getting errors
https://khanacademy.atlassian.net/browse/WB-176

I fixed this by using the media query string as the key inside MediaLayout, rather than the size. This fixes the issue when there is a modal and non-modal layout on the same page (as the modal layout doesn't have a "medium" size). I confirmed that this worked by resizing the full Styleguidist page so that it was in-between 768 and 1024px (medium size) and then reloaded the browser. All of the modals loaded without error! Additionally I resized the browser down to mobile size and then up and it immediately popped up to full size once it hit 768px in width (rather than 1024px, like before).

TwoColumnLayout rendering at small
https://khanacademy.atlassian.net/browse/WB-325

This one ended up being quite tricky! As it turns out the React context *was* working, it had just been set up in a way that the right props weren't getting passed to the right places. Additionally, the MediaLayout component wasn't wrapping the right other components. I had to re-work OneColumnModal and TwoColumnModal to fix the bug that I found (there wasn't anything immediately wrong with OneColumnModal but it might've been breaking in subtle ways so it was good to fix it, too).

screenshot 2018-08-01 16 20 30


packages/wonder-blocks-core/components/media-layout.js
```diff
@@ -223,10 +213,6 @@ class MediaLayoutContext extends React.Component<{
     }
 
     render() {
-        // TODO(jeresig): Switch to be a React.Fragment once we upgrade to
-        // React 16.2+.
-        // eslint-disable-next-line no-unused-vars
-        const {size, spec, children, ...otherProps} = this.props;
-        return <View {...otherProps}>{children}</View>;
+        return <React.Fragment>{this.props.children}</React.Fragment>;
```
jeresig: Nice little bit of clean-up!

jeresig:

Built with commit ae636065ee33d05b97752ee1b41e323818aa413f

https://deploy-preview-256--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/two-column-modal.js
```diff
@@ -150,7 +155,7 @@ const styles = StyleSheet.create({
     },
 
     smallColumn: {
-        flex: "0 0",
+        flex: "1",
```
jeresig: After getting the TwoColumnModal to render it was looking really bad, this fixed it, mostly. There are still some other issues with how it displays but I'll file those as follow-up issues.

packages/wonder-blocks-modal/components/modal-dialog.js
```diff
@@ -13,17 +17,18 @@ export default class ModalDialog extends React.Component<Props> {
...
-                ]}
-                role="dialog"
-                aria-labelledby="wb-modal-title"
-            >
-                {children}
+            <MediaLayout spec={MEDIA_MODAL_SPEC}>
+                <View
+                    style={[
+                        styles.wrapper,
+                        (mediaSize) => mediaSize === "small" && styles.small,
```
kevinbarabash: I just realized that this means that `mediaSize` can only be use to change the `style` but devs may also want to use `mediaSize` to toggle whether a component is rendered at all or to perform other logic.  I think we should have a component that provides the `mediaSize` (and the `mediaSpec`) using the function-as-children pattern.  I'll put together a separate PR and see what you think.

packages/wonder-blocks-modal/components/modal-dialog.js
```diff
@@ -13,17 +17,18 @@ export default class ModalDialog extends React.Component<Props> {
...
-                    styles.wrapper,
-                    (mediaSize) => mediaSize === "small" && styles.small,
-                    style,
-                ]}
-                role="dialog"
-                aria-labelledby="wb-modal-title"
-            >
-                {children}
+            <MediaLayout spec={MEDIA_MODAL_SPEC}>
+                <View
```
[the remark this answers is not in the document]
kevinbarabash: Good catch.  We may want to look at renaming some of these components to follow the provider/consumer pattern established in the new Context API, but that could be a separate PR.

packages/wonder-blocks-modal/components/one-column-modal.js
```diff
@@ -89,10 +93,12 @@ class ContentWrapper extends React.Component<Props> {
     }
 }
 
+const WrappedContentWrapper = MediaLayoutWrapper(ContentWrapper);
```
kevinbarabash: Having an explicit `MediaConsumer` component that uses function-as-child would mean that we wouldn't have to wrap everything with `MediaLayoutWrapper`.  Things can get a little dicey when you have to use multiple wrappers on the same component too since sometimes they have to be applied in a specific order.  I'll tackle this in the provider/consumer PR that I've proposed.

packages/wonder-blocks-modal/components/one-column-modal.js
```diff
@@ -105,21 +111,20 @@ class OneColumnModal extends React.Component<Props> {
...
-export default MediaLayoutWrapper(OneColumnModal);
-
 const styles = StyleSheet.create({
     dialog: {
         background: Color.white,
     },
 
     largeDialog: {
         width: "64.65%",
+        maxWidth: 960,
```
kevinbarabash: Shouldn't this be `maxWidth: 662`?  Otherwise the lines may be two long.

I think not setting the height for `OneColumnModal` is a good call, but we should probably have a maxHeight just in case there's too much content for whatever reason, maybe `90%`.

packages/wonder-blocks-modal/components/two-column-modal.js
```diff
@@ -126,17 +132,16 @@ class TwoColumnModal extends React.Component<Props> {
...
         );
     }
 }
 
-export default MediaLayoutWrapper(TwoColumnModal);
-
 const styles = StyleSheet.create({
     dialog: {
         width: "86.72%",
+        maxWidth: 960,
```
kevinbarabash: This should be 888 to match the spec.

kevinbarabash: The StandardModal dimensions are also off, they should be:
```
width: "93.75%",
maxWidth: 960,
height: "81.25%",
maxHeight: 624,
```

packages/wonder-blocks-modal/components/one-column-modal.js
kevinbarabash: Also, in talking with other devs, people in general preferred the FaC pattern as opposed to the HoC pattern.

packages/wonder-blocks-modal/components/modal-dialog.js
kevinbarabash: Also, in talking with other devs, people in general preferred the FaC pattern as opposed to the HoC pattern, mainly due to its flexibility.

kevinbarabash: One additional though about this, allow callbacks to be passed `style` makes our `style` incompatible with React Native's `style` prop which was one of the reasons for adopting our current `style` prop in the first place.

Sorry for not thinking about these arguments when this callback was first introduced.

packages/wonder-blocks-modal/components/two-column-modal.js
kevinbarabash: 🛑 this is blocking.

packages/wonder-blocks-modal/components/one-column-modal.js
kevinbarabash: 🛑 this is blocking.

packages/wonder-blocks-core/components/media-layout.js
```diff
@@ -29,15 +28,6 @@ type Props = {|
 
     /** The contents of the page */
     children: React.Node,
-
-    /**
-     * Style the layout container
-     */
-    style?: any,
```
kevinbarabash: Nice!  This should prevent future issues with `style` on `MediaLayout`.

packages/wonder-blocks-modal/components/modal-dialog.js
jeresig: Nope, I though of this too! Which is why I wrote up the MediaLayout/Layout spec - I'll be pursuing that next! https://docs.google.com/document/d/1hdh0_tEn0RY7jkC71Sln2BS3xJ9alwFzW6qYRiKu3kU/edit

packages/wonder-blocks-modal/components/two-column-modal.js
jeresig: Fixed, I got rid of the `minHeight: 200`

packages/wonder-blocks-modal/components/one-column-modal.js
jeresig: Done! Updated to 662 and added `maxHeight: "90%"`

packages/wonder-blocks-modal/components/modal-dialog.js
kevinbarabash: Yes!  This is much better.  Thanks for iterating on this. :)

jeresig:
@kevinbarabash Finished fixing! Ready for another look.

packages/wonder-blocks-modal/components/modal-dialog.js
kevinbarabash: I totally forgot about that doc. 😅

PR #257, opened 2018-08-02
Fix missing router context.

jeresig:
I noticed a couple issues with how we were using the router context.

1) Everywhere, except in the button module, we had forgotten to include the `contextTypes` (meaning that React would never give us access to the context, and thus we'd never get the router).
2) This introduced a dependency on prop-types for link, icon-button, and dropdown so I went to add it as a peerDependency and also realized it was missing for button, so I added it there too.

### DIFF
--- a/packages/wonder-blocks-dropdown/components/action-item.js
+++ b/packages/wonder-blocks-dropdown/components/action-item.js
@@ -5,6 +5,7 @@
 import * as React from "react";
 import {StyleSheet} from "aphrodite";
 import {Link} from "react-router-dom";
+import PropTypes from "prop-types";
 
 import Color, {mix, fade} from "@khanacademy/wonder-blocks-color";
 import Spacing from "@khanacademy/wonder-blocks-spacing";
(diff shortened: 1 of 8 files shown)


jeresig:

Built with commit a00b7ce59620f2feec3a24672c4189af98bf2e83

https://deploy-preview-257--wonder-blocks.netlify.com

packages/wonder-blocks-button/package.json
```diff
@@ -20,6 +20,7 @@
   },
   "peerDependencies": {
     "aphrodite": "^1.2.5",
+    "prop-types": "^15.6.2",
```
kevinbarabash: good call

packages/wonder-blocks-dropdown/components/action-item.js
```diff
@@ -80,6 +81,8 @@ export default class ActionItem extends React.Component<ActionProps> {
         indent: false,
     };
 
+    static contextTypes = {router: PropTypes.any};
```
kevinbarabash: Thanks for fixing this.  What's interesting is that the clientNav tests were passing for all components not just Button.  I think the issue is https://github.com/Khan/wonder-blocks/blob/master/packages/wonder-blocks-core/util/get-clickable-behavior.js#L37-L39.  I think we should make `clientNav` mandatory when using the router.  Otherwise it's hard to know whether you'll get client-side navigation just by looking at a call-site whether you'll get client-side navigation or not.  🛑

PR #258, opened 2018-08-03
WB-338: exclude all external deps from built files

kevinbarabash:
This is still kind of error prone b/c I had to check things manually.  I would like to create a script to verify that all require/import state for an external dep appear in the modules "dependencies" or "peerDependencies".  That'll be a separate PR b/c I want to get some of these components published.  For now we can verify what's being bundled by looking at the output of webpack which we're now passing `--display-modules` to which will show all modules.  Ones that aren't bundled are marked as `external` and only occupy 42 bytes in the bundle.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,8 +81,10 @@
   "dependencies": {
     "aphrodite": "^1.2.5",
     "prop-types": "^15.6.2",
+    "popper.js": "^1.14.1",
     "react": "^16.4.1",
     "react-dom": "^16.4.1",
+    "react-popper": "^1.0.0",
     "react-router-dom": "^4.2.2"
   },
   "workspaces": [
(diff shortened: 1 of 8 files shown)


jeresig:

Built with commit a43a737f709d1fa2b1733229ca0118bc363ec1e5

https://deploy-preview-258--wonder-blocks.netlify.com

kevinbarabash:
Whoops.  I missed `react-router` in `wonder-blocks-form`.
```
packages/wonder-blocks-form/dist/index.js  377 KiB       0  [emitted]  [big]  main
...
[32] ./node_modules/react-router-dom/es/index.js + 30 modules 75.7 KiB {0} [built]
...
```

PR #261, opened 2018-08-03
Make client-side navigation the default.

jeresig:
If a React Router is available then we would be defaulting to using client-side navigation and force the dev to opt out of it explicitly (rather than the other way around). This simplifies that logic for using these components a bit. Simply: If a React Router exists (and the href is a non-external URL) then it'll try to use React Router. If you don't want to use React Router then you opt out of it by providing a `directNav={true}`. This replaces the use of `clientNav` (which was the opposite, opting in).

There is similar logic in ActionItem, Link, Button, and IconButton so there is a lot of duplicate code in this diff, but it's all pretty similar.

### DIFF
--- a/packages/wonder-blocks-button/components/button.js
+++ b/packages/wonder-blocks-button/components/button.js
@@ -70,17 +70,18 @@ export type SharedProps = {|
     href?: string,
 
     /**
-     * Whether to use client-side navigation.
+     * Whether to avoid using client-side navigation.
      *
      * If the URL passed to href is local to the client-side, e.g.
-     * /math/algebra/eval-exprs, then it uses react-router-dom's Link
-     * component which handles the client-side navigation.
+     * /math/algebra/eval-exprs, then it tries to use react-router-dom's Link
+     * component which handles the client-side navigation. You can set
+     * `directNav` to true avoid using client-side nav entirely.
      *
      * NOTE: All URLs containing a protocol are considered external, e.g.
      * https://khanacademy.org/math/algebra/eval-exprs will trigger a full
      * page reload.
      */
-    clientNav?: boolean,
+    directNav?: boolean,
 
     /**
      * The content of the modal, appearing between the titlebar and footer.
@@ -145,11 +146,11 @@ export default class Button extends React.Component<SharedProps> {
     static contextTypes = {router: PropTypes.any};
 
     render() {
-        const {onClick, href, children, clientNav, ...sharedProps} = this.props;
+        const {onClick, href, children, directNav, ...sharedProps} = this.props;
 
         const ClickableBehavior = getClickableBehavior(
             href,
-            clientNav,
+            directNav,
             this.context.router,
         );
 
@@ -165,7 +166,7 @@ export default class Button extends React.Component<SharedProps> {
                             {...sharedProps}
                             {...state}
                             {...handlers}
-                            clientNav={clientNav}
+                            directNav={directNav}
                             href={href}
                         >
                             {children}
--- a/packages/wonder-blocks-button/components/button.md
+++ b/packages/wonder-blocks-button/components/button.md
@@ -274,7 +274,7 @@ const styles = StyleSheet.create({
 </View>
 ```
 
-Buttons can do client-side navigation if `clientNav` is `true`.
+Buttons do client-side navigation by default, if React Router exists:
 ```jsx
 const {StyleSheet} = require("aphrodite");
 const {View} = require("@khanacademy/wonder-blocks-core");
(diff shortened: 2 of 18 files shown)


jeresig:

Built with commit a060ee0552b4177ec8d081f61480b293e986b67b

https://deploy-preview-261--wonder-blocks.netlify.com

packages/wonder-blocks-core/util/get-clickable-behavior.js
```diff
@@ -23,26 +20,22 @@ function isExternalUrl(url: string) {
...
-    } else if (isExternalUrl(href)) {
-        return ClickableBehavior;
-    } else if (clientNav != null) {
-        if (clientNav) {
-            return ClickableBehaviorWithRouter;
-        } else {
-            return ClickableBehavior;
-        }
-    } else if (router) {
+    if (router && directNav !== true && href && !isExternalUrl(href)) {
```
kevinbarabash: Nice cleanup.  Super minor nit: `!directNav` feels a bit easier to read.

packages/wonder-blocks-core/util/get-clickable-behavior.js
```diff
@@ -23,26 +20,22 @@ function isExternalUrl(url: string) {
 const ClickableBehaviorWithRouter = withRouter(ClickableBehavior);
 
 export default function getClickableBehavior(
+    /**
+     * The URL to navigate to.
+     */
```
kevinbarabash: Thanks for adding comments. 🎉

packages/wonder-blocks-button/components/button.js
```diff
@@ -70,17 +70,18 @@ export type SharedProps = {|
...
+     * /math/algebra/eval-exprs, then it tries to use react-router-dom's Link
+     * component which handles the client-side navigation. You can set
+     * `directNav` to true avoid using client-side nav entirely.
      *
      * NOTE: All URLs containing a protocol are considered external, e.g.
      * https://khanacademy.org/math/algebra/eval-exprs will trigger a full
      * page reload.
      */
-    clientNav?: boolean,
+    directNav?: boolean,
```
kevinbarabash: I'm glad we're making client-side navigation the default.  I wonder if we should call it something less benign, e.g. `skipClientNav`.  When we were trying to brainstorm names before I was trying to think of something short but now that I think of it being a bit longer is probably feature in this case.  Also, `skipClientNav` clearly communicates what it's doing whereas I'm sure many people will have to look up what `directNav` means. 🛑

packages/wonder-blocks-core/util/get-clickable-behavior.test.js
```diff
@@ -53,44 +53,44 @@ describe("getClickableBehavior", () => {
...
+                const router = <MemoryRouter />;
+                const expectation = ClickableBehavior;
 
                 // Act
-                const result = getClickableBehavior(url, clientNav, router);
+                const result = getClickableBehavior(url, directNav, router);
 
                 // Assert
-                expect(result.displayName).toBe(expectation);
+                expect(result).toBe(expectation);
```
kevinbarabash: nice

packages/wonder-blocks-button/components/button.md
```diff
@@ -274,7 +274,7 @@ const styles = StyleSheet.create({
 </View>
 ```
 
-Buttons can do client-side navigation if `clientNav` is `true`.
+Buttons do client-side navigation by default, if React Router exists:
```
kevinbarabash: We should include an example of how to avoid client-side navigation within a router if necessary.  Also, I wonder what happens if people set `target="_blank"` on a button with an `href`. 🤔

packages/wonder-blocks-core/util/get-clickable-behavior.test.js
```diff
@@ -53,44 +53,44 @@ describe("getClickableBehavior", () => {
...
-                    const result = getClickableBehavior(url, clientNav, router);
+                    const result = getClickableBehavior(url, directNav, router);
 
                     // Assert
                     expect(result.displayName).toBe(expectation);
                 });
             });
 
-            test("Client navigation is false, returns ClickableBehavior", () => {
+            test("Direct navigation is false, returns ClickableBehaviorWithRouter", () => {
```
kevinbarabash: `Direct navigation is false w/ Router...`?

packages/wonder-blocks-button/components/button.js
jeresig: Good suggestion - I've renamed it to `skipClientNav`

packages/wonder-blocks-button/components/button.md
jeresig: I added an example of a link with `skipClientNav` - but yeah, maybe we should make it so that if you set a `target` it automatically sets `skipClientNav={true}`?

kevinbarabash: It looks like `react-router`'s `Link` supports this out of the box so `skipClientNav` isn't necessary when `taget="_blank"` is set.  https://github.com/ReactTraining/react-router/pull/2193

PR #265, opened 2018-08-06
Update invalid style casing fix, apply -moz-focus-inner to all buttons

wishdasher:
I moved the `::-moz-focus-inner` fix to the `overrides` stylesheet in `add-style` because people may not remember to remove that ring whenever they make a `addStyle("button")`.

React was also complaining about the non camelCased style properties, so this resulted in changing the replace function in `processStyleList` to replace all instances of `/-[a-z]/` with the corresponding `[A-Z]`, so that all styles in `aphrodite.flow.js` would be converted automatically.

Also corrected the spelling of aphrodite for that flow file.

Verification:
* Running `yarn run jest` no longer gives warnings about the non camelCased property.
* Buttons don't have inner focus rings on Firefox.
* Media queries `-width` => `Width` is still happening correctly.


jeresig:

Built with commit 8d22a51c78b44c361a9678c93e1068ebf9b455fd

https://deploy-preview-265--wonder-blocks.netlify.com

packages/wonder-blocks-core/util/util.js
```diff
@@ -56,7 +56,14 @@ export function processStyleList<T: Object>(
                 // It doesn't accept kebab-case in media queries and instead
                 // prefers camelCase.
                 for (const [key, value] of Object.entries(child._definition)) {
-                    def[key.replace("-width", "Width")] = value;
+                    // This regex converts all instances of -{lowercaseLetter}
+                    // to the uppercase version of that letter, without the
+                    // leading dash.
+                    def[
+                        key.replace(new RegExp("-[a-z]", "g"), (match) =>
+                            match[1].toUpperCase(),
```
wishdasher: Couldn't manage to get back references or the `/pattern/g` syntax to work

wishdasher:
I just realized that #262 addresses a subset of the stuff here 😨

packages/wonder-blocks-core/util/util.js
kevinbarabash: I think you should be able to do:
```
key.replace(/-([a-z])/g, match => match[1].toUpperCase())
```
no back references required, only groups.

kevinbarabash:
Thanks for fixing the spelling of aphrodite.flow.js.

packages/wonder-blocks-core/util/util.js
wishdasher: Oh, I was putting the pattern in parentheses and didn't realize that was the issue. Thank you!

packages/wonder-blocks-core/util/util.js
```diff
@@ -56,7 +56,14 @@ export function processStyleList<T: Object>(
                 // It doesn't accept kebab-case in media queries and instead
                 // prefers camelCase.
                 for (const [key, value] of Object.entries(child._definition)) {
-                    def[key.replace("-width", "Width")] = value;
+                    // This regex converts all instances of -{lowercaseLetter}
+                    // to the uppercase version of that letter, without the
+                    // leading dash.
+                    def[
+                        key.replace(/-[a-z]/g, (match) =>
+                            match[1].toUpperCase(),
```
kevinbarabash: I don't understand why a group isn't necessary here... does the stuff inside `[ ]` count as a group.

wishdasher: The `[ ]` doesn't count as a group, but a group isn't necessary because there's no additional operations on that group.

kevinbarabash: Ah... it's grabbing the first character from the `match`.  That took me way too long to figure out. 😞

PR #266, opened 2018-08-07
Ensure that all of the mains point to the right index.js file.

jeresig:
This was causing problems when we tried to include them in webapp. I'm going to integrate a check for this into the pre-publish-check.js.


jeresig:

Built with commit afe7c6a5cf78cb860c0b4d8a6a057445e2c094ed

https://deploy-preview-266--wonder-blocks.netlify.com

PR #267, opened 2018-08-07
Check package main and NPM user.

jeresig:
These would've saved me some trouble this morning!


jeresig:

Built with commit ffbace09a7b27c8f326ec095cfdcbd254016d9d5

https://deploy-preview-267--wonder-blocks.netlify.com

utils/pre-publish-check.js
```diff
@@ -16,39 +18,55 @@ glob(
...
+                        },
+                    ])
+                    .then(({skipWarnings}) => {
+                        // If we're not skipping then we need to exit with an error
+                        if (!skipWarnings) {
+                            process.exit(1);
+                        }
+                    });
+            }
+        });
```
somewhatabstract: Perhaps this whole condition should be in a method like `_maybeContinueAfterWarnings()`.

Would've made diff a little cleaner as that big chunk of red on the left and this green would've more easily correlated.

utils/pre-publish-check.js
```diff
@@ -16,39 +18,55 @@ glob(
...
-        }
+        exec("npm whoami", (err, currentUser) => {
+            if (currentUser.trim() !== "khanacademy") {
+                console.error(
+                    `ERROR: You are not logged in to NPM as "khanacademy". ` +
+                        `Run "npm login" and use the password from: ` +
+                        `https://phabricator.khanacademy.org/K207`,
+                );
+                process.exit(1);
+            }
```
somewhatabstract: Should this be a method? `checkKhanLogin` or something.

utils/pre-publish-check.js
```diff
@@ -16,39 +18,55 @@ glob(
...
                 );
                 process.exit(1);
             }
 
+            if (pkgJson.main !== "dist/index.js") {
+                console.error(
+                    `ERROR: ${name} must have a "main" set to "dist/index.js".`,
+                );
+                process.exit(1);
+            }
```
somewhatabstract: I think all these error checks will be way easier to maintain if they were methods. The method names would document the purpose of the check, rather than relying on maintainers to insinuate the purpose from the implementation.

kevinbarabash: something like:
```
checkPublishConfig();
checkMainField();
...
```
?

utils/pre-publish-check.js
kevinbarabash: That error message is super helpful, nice!

kevinbarabash:
While I think the refactoring of the checks into functions would be nice, I'd rather merge this code so that we can start using it.  The refactoring can come later.

PR #268, opened 2018-08-08
Add keyboard accessibility to dropdown menus

wishdasher:
This change creates refs for each item in the Dropdown component, which handles focus navigation logic based on a keyboard event listener. Instead of the menu/select components cloning the user-supplied items, this now happens at the Dropdown component level so that the Dropdown component can also create the refs and manage indices without having to reclone each item.

Also now pre-binding callbacks.

Expected behavior:
* Opening the menu with mouse click doesn't automatically fous an item, but opening the menu with the keyboard does
* Action menus and MultiSelect start off with the first item focused, but SingleSelect starts off with the selected item focused
* Pressing space on an item no longer scrolls the page, and it selects the item as expected
* Both tab and escape keys should close the menu. Because the menu is portalled, tabbing would naturally go to the next element in the DOM, which is probably the very first item on the page. I had to account for this by explicitly shifting focus back to the opener.
* Focused item should still be consistent even when an item becomes disabled (and thus is no longer focusable). This took a bit of extra variable tracking/code.

Changes to other components:
* Resetting the internal variables in ClickableBehavior in `handleBlur`. Originally, pressing space on an item would not always trigger the `handleClick` handler, which caused `keyboardClick` to never be reset. So pressing space and then clicking with the mouse would not trigger a click the second time as expected.
There are still other bugs in ClickableBehavior that I want to look at and address, but that's a separate task.

Future TODO:
* Write more extensive tests for item callbacks and ActionMenu, SingleSelect, MultiSelect-level navigation.

### DIFF
--- a/packages/wonder-blocks-dropdown/components/action-menu.js
+++ b/packages/wonder-blocks-dropdown/components/action-menu.js
@@ -40,7 +40,6 @@ class ActionMenuOpener extends React.Component<OpenerProps> {
 
     handleClick = (e: SyntheticEvent<>) => {
         const {open} = this.props;
-        // TODO: also detect enter click events
         this.props.onOpenChanged(!open, e.type === "keyup");
     };
 
@@ -185,8 +184,11 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
         const containsOptionItems = Array.isArray(selectedValues);
 
         return React.Children.map(children, (item) => {
-            if (item.type === ActionItem) {
-                const {disabled} = item.props;
+            const {
+                type,
+                props: {disabled, value},
+            } = item;
+            if (type === ActionItem) {
                 return {
                     component: item,
                     focusable: !disabled,
@@ -194,8 +196,7 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
                         indent: containsOptionItems,
                     },
                 };
-            } else if (item.type === OptionItem) {
-                const {disabled, value} = item.props;
+            } else if (type === OptionItem) {
                 return {
                     component: item,
                     focusable: !disabled,
--- a/packages/wonder-blocks-dropdown/components/action-menu.js
+++ b/packages/wonder-blocks-dropdown/components/action-menu.js
@@ -10,7 +10,9 @@ import Icon, {icons} from "@khanacademy/wonder-blocks-icon";
 import Dropdown from "./dropdown.js";
 import ActionItem from "./action-item.js";
 import OptionItem from "./option-item.js";
-import SeparatorItem from "./separator-item.js";
+
+import type {Item} from "../util/types.js";
+import type {DropdownItem} from "../util/types.js";
 
 type OpenerProps = {|
     /**
@@ -24,7 +26,11 @@ type OpenerProps = {|
     /**
      * Callback for when the opener is pressed.
      */
-    onClick: () => void,
+    onOpenChanged: (open: boolean, keyboard: boolean) => void,
+    /**
+     * Whether the dropdown is open.
+     */
+    open: boolean,
 |};
 
 class ActionMenuOpener extends React.Component<OpenerProps> {
@@ -32,8 +38,14 @@ class ActionMenuOpener extends React.Component<OpenerProps> {
         disabled: false,
     };
 
+    handleClick = (e: SyntheticEvent<>) => {
+        const {open} = this.props;
+        // TODO: also detect enter click events
+        this.props.onOpenChanged(!open, e.type === "keyup");
+    };
+
     render() {
-        const {children, disabled, onClick} = this.props;
+        const {children, disabled} = this.props;
 
         return (
             // $FlowFixMe: button doesn't allow 'role' prop or Icon as children
@@ -43,7 +55,7 @@ class ActionMenuOpener extends React.Component<OpenerProps> {
                 disabled={disabled}
                 kind="tertiary"
                 light={false}
-                onClick={onClick}
+                onClick={this.handleClick}
                 style={styles.opener}
             >
                 {children}
@@ -61,11 +73,7 @@ type MenuProps = {|
     /**
      * The items in this dropdown.
      */
-    children: Array<
-        React.Element<
-            typeof ActionItem | typeof OptionItem | typeof SeparatorItem,
-        >,
-    >,
+    children: Array<React.Element<Item>>,
 
     /**
      * Text for the opener of this menu.
@@ -107,13 +115,17 @@ type State = {|
      * Whether or not the dropdown is open.
      */
     open: boolean,
+    /**
+     * Whether or not last open state change was triggered by a keyboard click.
+     */
+    keyboard?: boolean,
 |};
 
 /**
  * A menu that consists of various types of items.
  */
 export default class ActionMenu extends React.Component<MenuProps, State> {
-    openerElement: ?Element;
+    openerElement: ?HTMLElement;
 
     static defaultProps = {
         alignment: "left",
@@ -128,13 +140,25 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
         };
     }
 
-    handleOpenChanged(open: boolean) {
+    handleOpenChanged = (open: boolean, keyboard?: boolean) => {
         this.setState({
-            open: open,
+            open,
+            keyboard,
         });
-    }
+    };
 
-    handleSelected(selectedValue: string) {
+    handleItemSelected = () => {
+        // Bring focus back to the opener element.
+        if (this.openerElement) {
+            this.openerElement.focus();
+        }
+
+        this.setState({
+            open: false, // close the menu upon selection
+        });
+    };
+
+    handleOptionSelected = (selectedValue: string) => {
         const {onChange, selectedValues} = this.props;
 
         // If either of these are not defined, return.
@@ -153,49 +177,60 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
             // Item was newly selected
             onChange([...selectedValues, selectedValue]);
         }
-    }
+        this.handleItemSelected();
+    };
 
-    getMenuItems(): Array<
-        React.Element<
-            typeof ActionItem | typeof OptionItem | typeof SeparatorItem,
-        >,
-    > {
+    getMenuItems(): Array<DropdownItem> {
         const {children, selectedValues} = this.props;
         const containsOptionItems = Array.isArray(selectedValues);
 
-        return React.Children.map(children, (item, index) => {
+        return React.Children.map(children, (item) => {
             if (item.type === ActionItem) {
-                return React.cloneElement(item, {
-                    indent: containsOptionItems,
-                });
+                const {disabled} = item.props;
+                return {
+                    component: item,
+                    focusable: !disabled,
+                    populatedProps: {
+                        indent: containsOptionItems,
+                    },
+                };
             } else if (item.type === OptionItem) {
-                return React.cloneElement(item, {
-                    onToggle: (value) => this.handleSelected(value),
-                    selected: selectedValues
-                        ? selectedValues.includes(item.props.value)
-                        : false,
-                    variant: "check",
-                });
+                const {disabled, value} = item.props;
+                return {
+                    component: item,
+                    focusable: !disabled,
+                    populatedProps: {
+                        onToggle: this.handleOptionSelected,
+                        selected: selectedValues
+                            ? selectedValues.includes(value)
+                            : false,
+                        variant: "check",
+                    },
+                };
             } else {
-                return item;
+                return {
+                    component: item,
+                    focusable: false,
+                    populatedProps: {},
+                };
             }
         });
     }
 
+    handleOpenerRef = (node: any) => {
+        this.openerElement = ((ReactDOM.findDOMNode(node): any): HTMLElement);
+    };
+
     render() {
         const {alignment, disabled, menuText, style} = this.props;
-
         const {open} = this.state;
 
         const opener = (
             <ActionMenuOpener
                 disabled={disabled}
-                onClick={() => this.handleOpenChanged(!open)}
-                ref={(node) =>
-                    (this.openerElement = ((ReactDOM.findDOMNode(
-                        node,
-                    ): any): Element))
-                }
+                onOpenChanged={this.handleOpenChanged}
+                open={open}
+                ref={this.handleOpenerRef}
             >
                 {menuText}
             </ActionMenuOpener>
@@ -205,15 +240,15 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
             <Dropdown
                 alignment={alignment}
                 dropdownStyle={styles.menuTopSpace}
+                items={this.getMenuItems()}
+                keyboard={this.state.keyboard}
                 light={false}
-                onOpenChanged={(open) => this.handleOpenChanged(open)}
+                onOpenChanged={this.handleOpenChanged}
                 open={open}
                 opener={opener}
                 openerElement={this.openerElement}
                 style={style}
-            >
-                {this.getMenuItems()}
-            </Dropdown>
+            />
         );
     }
 }
--- a/packages/wonder-blocks-dropdown/components/action-menu.md
+++ b/packages/wonder-blocks-dropdown/components/action-menu.md
@@ -62,6 +62,7 @@ class HybridMenu extends React.Component {
         this.state = {
             selectedValues: ["homework"],
         };
+        // Styleguidist doesn't support arrow functions in class field properties
         this.handleChange = this.handleChange.bind(this);
     }
 
--- a/packages/wonder-blocks-dropdown/components/action-menu.md
+++ b/packages/wonder-blocks-dropdown/components/action-menu.md
@@ -62,6 +62,7 @@ class HybridMenu extends React.Component {
         this.state = {
             selectedValues: ["homework"],
         };
+        this.handleChange = this.handleChange.bind(this);
     }
 
     handleChange(update) {
@@ -73,7 +74,7 @@ class HybridMenu extends React.Component {
     render() {
         return <ActionMenu
             menuText="Assignments"
-            onChange={(selectedValues) => this.handleChange(selectedValues)}
+            onChange={this.handleChange}
             selectedValues={this.state.selectedValues}
         >
             <ActionItem label="Create..." onClick={() => console.log("create action")} />
--- a/packages/wonder-blocks-dropdown/components/action-menu.test.js
+++ b/packages/wonder-blocks-dropdown/components/action-menu.test.js
@@ -7,13 +7,10 @@ import ActionItem from "./action-item.js";
 import OptionItem from "./option-item.js";
 import SeparatorItem from "./separator-item.js";
 import ActionMenu from "./action-menu.js";
-
-const keyCodes = {
-    enter: 13,
-    space: 32,
-};
+import {keyCodes} from "../util/constants.js";
 
 describe("ActionMenu", () => {
+    global.scrollTo = jest.fn();
     let menu;
     const onClick = jest.fn();
     const onToggle = jest.fn();
@@ -23,16 +20,12 @@ describe("ActionMenu", () => {
         menu = mount(
             <ActionMenu
                 menuText={"Action menu!"}
-                onChange={(selectedValues) => onChange()}
+                onChange={onChange}
                 selectedValues={[]}
             >
-                <ActionItem label="Action" onClick={() => onClick()} />
+                <ActionItem label="Action" onClick={onClick} />
                 <SeparatorItem />
-                <OptionItem
-                    label="Toggle"
-                    value="toggle"
-                    onClick={() => onToggle()}
-                />
+                <OptionItem label="Toggle" value="toggle" onClick={onToggle} />
             </ActionMenu>,
         );
     });
@@ -75,6 +68,9 @@ describe("ActionMenu", () => {
         actionItem.simulate("mouseup", nativeEvent);
         actionItem.simulate("click");
         expect(onClick).toHaveBeenCalledTimes(1);
+
+        // Have to reopen menu because menu closes after an item is selected
+        menu.setState({open: true});
 
         const optionItem = menu.find(OptionItem);
         optionItem.simulate("mousedown");
--- a/packages/wonder-blocks-dropdown/components/action-menu.test.js
+++ b/packages/wonder-blocks-dropdown/components/action-menu.test.js
@@ -10,13 +10,13 @@ import ActionMenu from "./action-menu.js";
 import {keyCodes} from "../util/constants.js";
 
 describe("ActionMenu", () => {
-    global.scrollTo = jest.fn();
     let menu;
     const onClick = jest.fn();
     const onToggle = jest.fn();
     const onChange = jest.fn();
 
     beforeEach(() => {
+        window.scrollTo = jest.fn();
         menu = mount(
             <ActionMenu
                 menuText={"Action menu!"}
@@ -31,6 +31,7 @@ describe("ActionMenu", () => {
     });
 
     afterEach(() => {
+        window.scrollTo.mockClear();
         unmountAll();
     });
 
--- a/packages/wonder-blocks-dropdown/components/dropdown.js
+++ b/packages/wonder-blocks-dropdown/components/dropdown.js
@@ -12,15 +12,34 @@ import Spacing from "@khanacademy/wonder-blocks-spacing";
 import {View} from "@khanacademy/wonder-blocks-core";
 
 import visibilityModifierDefaultConfig from "../util/visibility-modifier.js";
-import typeof ActionItem from "./action-item.js";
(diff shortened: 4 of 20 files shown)


jeresig:

Built with commit 1ade7902df7bc37d682be9b100a777839745f5c6

https://deploy-preview-268--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/action-item.js
```diff
@@ -193,9 +190,10 @@ const styles = StyleSheet.create({
 
     label: {
         whiteSpace: "nowrap",
-        overflow: "hidden",
-        textOverflow: "ellipsis",
-        textAlign: "left",
```
wishdasher: I misread the spec, items in the dropdown don't have a limited width.

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -24,7 +24,7 @@ type OpenerProps = {|
     /**
      * Callback for when the opener is pressed.
      */
-    onClick: () => void,
+    onClick: (keyboard: boolean) => void,
```
wishdasher: The opener lets the menu know that a keyboard press was used to open it.

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -172,18 +200,26 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
...
 
         return React.Children.map(children, (item, index) => {
             if (item.type === ActionItem) {
+                const {disabled, onClick} = item.props;
                 return React.cloneElement(item, {
+                    ref: !disabled && this.getRef(),
                     indent: containsOptionItems,
+                    onClick: () => {
+                        onClick();
+                        this.handleItemClicked();
```
wishdasher: Gotta close the menu after a click.

packages/wonder-blocks-dropdown/components/action-menu.test.js
```diff
@@ -14,6 +14,7 @@ const keyCodes = {
 };
 
 describe("ActionMenu", () => {
+    global.scrollTo = jest.fn();
```
wishdasher: So that `window.scrollTo` doesn't throw an error.

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -16,33 +16,60 @@ import typeof ActionItem from "./action-item.js";
...
+    initialFocusedIndex: number,
 
     /**
-     * The component that opens the menu.
+     * A list of React component refs to the non-disabled items of this menu.
+     * The items should have a focusItem method, which is called by keyboard
+     * listeners in this component to provide keyboard accessible navigation.
      */
-    opener: React.Element<*>,
+    itemRefs: Array<{current: any}>,
```
wishdasher: This is being passed by the wrapper components around Dropdown. Two reasons it's being passed instead of generated by the Dropdown component itself: I didn't know if it was possible to render all the children here adding a ref via `React.createRef` without having to reclone all the elements, and the wrapper components are filtering out disabled items and separator components before passing this, which is a good thing.

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -102,13 +151,11 @@ export default class Dropdown extends React.Component<DropdownProps> {
     addEventListeners() {
         document.addEventListener("mouseup", this.handleInteract);
         document.addEventListener("touchend", this.handleInteract);
-        document.addEventListener("keyup", this.handleKeyup);
```
wishdasher: This was originally for closing on the menu on pressing escape, now there's a general keyboard listener that handles this and navigation.

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -138,21 +140,33 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
...
+        if (this.openerElement) {
+            this.openerElement.focus();
+        }
+
+        this.setState({
+            open: false, // close the menu upon selection
+        });
+    };
 
-    handleSelected(selectedValue: string, oldSelectionState: boolean) {
```
wishdasher: Slightly changed the API of OptionItem to reduce redundancy.

packages/wonder-blocks-dropdown/components/option-item.js
```diff
@@ -34,15 +33,14 @@ type OptionProps = {|
     /**
      * Optional user-supplied callback when this item is called.
      */
-    onClick?: (oldSelectionState: boolean) => void,
```
wishdasher: The user of the menu controls which items are selected, so they already have this information.

wishdasher:
@jeresig Thanks for the detailed description of all the issues! I believe I've fixed them all -- and plus one more that I encountered while investigating.

One issue was due to a bug in ClickableBehavior and another issue was due to how FocusTrap of the modals package behaved. I've included solutions for those fixes as well.

wishdasher:
Accidentally force pushed while this was closed, took a while to figure out how to reopen...

wishdasher:
@jeresig Thanks for the detailed description of all the issues! I believe I've fixed them all -- and plus one more that I encountered while investigating. One issue was due to a bug in ClickableBehavior and another issue was due to how FocusTrap of the modals package behaved. I've included solutions for those fixes as well.

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -12,15 +12,41 @@ import Spacing from "@khanacademy/wonder-blocks-spacing";
...
      * Items for the menu.
      */
-    children: Array<React.Element<ActionItem | OptionItem | SeparatorItem>>,
+    items: Array<DropdownItem>,
+
+    /**
+     * An index that represents the index of the focused element when the menu
+     * is opened.
+     */
+    initialFocusedIndex: number,
```
kevinbarabash: Is this to handle the case where you've already selected something from the dropdown and you're opening the dropdown a second time?

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,98 @@
...
+
+describe("Dropdown", () => {
+    global.scrollTo = jest.fn();
+    global.getComputedStyle = jest.fn();
+
+    afterEach(() => {
+        unmountAll();
+    });
+
+    it("handles keyboard behavior as expected", () => {
```
kevinbarabash: Given all of the different scenarios that dropdown.js handles I would expect more tests.  We probably want to cover the following scenarios in the test matrix:
- keyboard: false, initialFocusedIndex: 0
- keyboard: false, initialFocusedIndex: not 0
- keyboard: true, initialFocusedIndex: 0
- keyboard: true, initialFocusedIndex: not 0

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,98 @@
...
+        const dropdown = mount(
+            <Dropdown
+                initialFocusedIndex={0}
+                // mock the itemrefs
+                items={[
+                    {
+                        component: (
+                            <OptionItem label="item 1" value="1" key="1" />
+                        ),
+                        focusable: true,
```
kevinbarabash: There should be a test that checks that we skip unfocusable options correctly.

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -142,10 +352,33 @@ export default class Dropdown extends React.Component<DropdownProps> {
...
+                        // if (index === 0) {console.log(item);}
+                        if (item.focusable) {
+                            focusCounter += 1;
+                        }
+                        const focusIndex = focusCounter - 1;
+                        return React.cloneElement(item.component, {
+                            ...item.populatedProps,
+                            key: index,
+                            ref: item.focusable && this.getRef(index),
+                            onClick: () => {
```
kevinbarabash: This `onClick` overwrite the `onClick` from `item.populatedProps`.

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -163,57 +177,65 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
...
+        return React.Children.map(children, (item) => {
             if (item.type === ActionItem) {
-                return React.cloneElement(item, {
-                    indent: containsOptionItems,
-                });
+                const {disabled, onClick} = item.props;
+                return {
+                    component: item,
+                    focusable: !disabled,
+                    populatedProps: {
```
kevinbarabash: This might be a little cleaner if we did `cloneElement` here and passed these additional props here.  The logic around `focusable` could be consolidated in `Dropdown` by checking `item.props.disabled` and `item.type` against each of the item types.

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,98 @@
...
+                        populatedProps: {},
```
kevinbarabash: We should probably have a test that checks that the callbacks that can appear in `populatedProps` will get called at the right time.

packages/wonder-blocks-dropdown/util/types.js
```diff
@@ -1,31 +1,15 @@
...
-    href?: string,
-    /** Whether to avoid using client-side navigation. */
-    skipClientNav?: boolean,
-    /** Callback on the action. */
-    onClick?: () => void,
-|};
+import * as React from "react";
+import typeof ActionItem from "../components/action-item.js";
+import typeof OptionItem from "../components/option-item.js";
+import typeof SeparatorItem from "../components/separator-item.js";
```
kevinbarabash: I didn't realize you could do imports like that.

packages/wonder-blocks-dropdown/components/action-menu.test.js
```diff
@@ -14,6 +14,7 @@ const keyCodes = {
 };
 
 describe("ActionMenu", () => {
```
kevinbarabash: We should probably have a some tests that keyboard navigation within an `ActionMenu` as well as the `SingleSelect` and `MultiSelect` work.  It's more valuable testing higher up b/c you're hitting more layers.  I would prefer to have stronger testing at this level and less testing in dropdown.test.js.

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -56,7 +76,8 @@ type DropdownProps = {|
     light: boolean,
 
     /**
-     * Styling specific to the dropdown component that isn't part of the opener.
+     * Styling specific to the dropdown component that isn't part of the opener,
+     * passed by the specific implementation of the dropdown menu,
      */
     dropdownStyle?: any,
```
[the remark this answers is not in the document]
wishdasher: I'm applying the user-passed `style` at the dropdown level now, so this should probably stay. It's not user-facing, the different implementations use it to fine-tune the spacing between the menu and the opener.

packages/wonder-blocks-dropdown/components/dropdown.js
wishdasher: It's to handle the case for single select menus, where the focused item on open should be the selected one, if any.

packages/wonder-blocks-dropdown/components/dropdown.js
wishdasher: `populatedProps` no longer tries to set an `onClick`

packages/wonder-blocks-dropdown/components/action-menu.js
wishdasher: That's what I was originally doing, but I wasn't able to add the `ref` to `children` of `Dropdown` without recloning each element, and I thought having two separate cloning processes was bad.

wishdasher: That's what I was originally doing, but I wasn't able to add the `ref` to  each child in `children` of `Dropdown` without recloning each element, and I thought having two separate cloning processes was bad.

packages/wonder-blocks-dropdown/components/action-menu.test.js
wishdasher: I was already working on navigation tests on `Dropdown` because that's where I tried to contain the navigation behavior logic. I can write tests higher up as well, but will maybe not include this in this PR.

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -153,49 +177,60 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
...
                 disabled={disabled}
-                onClick={() => this.handleOpenChanged(!open)}
-                ref={(node) =>
-                    (this.openerElement = ((ReactDOM.findDOMNode(
-                        node,
-                    ): any): Element))
-                }
+                onOpenChanged={this.handleOpenChanged}
+                open={open}
+                ref={this.handleOpenerRef}
```
jeresig: I like changes like this, it's making the code cleaner and easier to read!

packages/wonder-blocks-dropdown/components/action-menu.md
```diff
@@ -62,6 +62,7 @@ class HybridMenu extends React.Component {
         this.state = {
             selectedValues: ["homework"],
         };
+        this.handleChange = this.handleChange.bind(this);
```
jeresig: Alternatively, and I think we may prefer this, you could do:

```
handleChange = (update) => {
  ...
}
```

in the method definition.

packages/wonder-blocks-dropdown/components/action-menu.test.js
```diff
@@ -7,13 +7,10 @@ import ActionItem from "./action-item.js";
...
 import ActionMenu from "./action-menu.js";
-
-const keyCodes = {
-    enter: 13,
-    space: 32,
-};
+import {keyCodes} from "../util/constants.js";
 
 describe("ActionMenu", () => {
+    global.scrollTo = jest.fn();
```
jeresig: Hmm... this doesn't feel right. I think this needs to be in a `beforeEach` and then removed again in the `afterEach` like we do now for mount. Otherwise this will blow away the global which might have negative ramifications somehow? (Even though we're just running this in Node)

Alternatively, you could change the code itself to do something like:

```
if (window.scrollTo) {
    window.scrollTo(...);
}
```

That way we don't have to stub it out at all.

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,379 @@
+//@flow
+import React from "react";
+import {mount, unmountAll} from "../../../utils/testing/mount.js";
+
+import OptionItem from "./option-item.js";
+import Dropdown from "./dropdown.js";
+import {keyCodes} from "../util/constants.js";
+
+describe("Dropdown", () => {
+    global.scrollTo = jest.fn();
```
jeresig: Same for these too.

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -168,50 +175,75 @@ export default class MultiSelect extends React.Component<Props, State> {
...
+                focusable: !selectAllDisabled,
+                populatedProps: {},
+            };
+
+            const selectNoneDisabled = selectedValues.length === 0;
+            const selectNone = {
+                component: (
+                    <ActionItem
+                        disabled={selectNoneDisabled}
+                        label="Select none"
```
jeresig: Might want to add another i18n comment here too, just so we don't fix one but not the other!

packages/wonder-blocks-dropdown/components/multi-select.md
```diff
@@ -23,9 +23,10 @@ class ExampleNoneSelected extends React.Component {
         this.state = {
             selectedValues: [],
         };
+        this.handleChange = this.handleChange.bind(this);
```
jeresig: Same for this as above.

packages/wonder-blocks-dropdown/components/multi-select.md
```diff
@@ -143,9 +147,10 @@ class SimpleMultiSelect extends React.Component {
...
```
jeresig: Here too!

packages/wonder-blocks-dropdown/components/multi-select.test.js
```diff
@@ -6,26 +6,23 @@ import SelectOpener from "./select-opener.js";
...
 import MultiSelect from "./multi-select.js";
-
-const keyCodes = {
-    enter: 13,
-    space: 32,
-};
+import {keyCodes} from "../util/constants.js";
 
 describe("MultiSelect", () => {
+    global.scrollTo = jest.fn();
```
jeresig: This too.

packages/wonder-blocks-dropdown/components/option-item.js
```diff
@@ -139,12 +132,14 @@ const {blue, white, offBlack, offBlack32} = Color;
 
 const styles = StyleSheet.create({
     itemContainer: {
+        display: "flex",
```
jeresig: Is this necessary? Don't all `View`s have this by default?

packages/wonder-blocks-dropdown/components/select-opener.js
```diff
@@ -69,8 +73,14 @@ export default class SelectOpener extends React.Component<SelectOpenerProps> {
 
     static contextTypes = {router: PropTypes.any};
 
+    handleClick = (e: SyntheticEvent<>) => {
+        const {open} = this.props;
+        // TODO: also detect enter click events
```
jeresig: Same comment here, too.

packages/wonder-blocks-dropdown/components/single-select.md
```diff
@@ -25,7 +25,8 @@ class ExampleWithPlaceholder extends React.Component {
         super();
         this.state = {
             selectedValue: null,
-        }
+        };
+        this.handleChange = this.handleChange.bind(this);
```
jeresig: Same!

packages/wonder-blocks-dropdown/components/single-select.md
```diff
@@ -75,7 +76,8 @@ class ExampleWithStartingSelection extends React.Component {
         super();
         this.state = {
             selectedValue: "banana",
-        }
+        };
+        this.handleChange = this.handleChange.bind(this);
```
jeresig: Same!

packages/wonder-blocks-dropdown/components/single-select.test.js
```diff
@@ -5,22 +5,16 @@ import {mount, unmountAll} from "../../../utils/testing/mount.js";
...
 import SingleSelect from "./single-select.js";
-
-const keyCodes = {
-    enter: 13,
-    space: 32,
-};
+import {keyCodes} from "../util/constants.js";
 
 describe("SingleSelect", () => {
+    global.scrollTo = jest.fn();
```
jeresig: Same!

packages/wonder-blocks-dropdown/docs.md
```diff
@@ -1 +1,12 @@
-NOTE: Do not use yet for longer dropdowns (like more than six items or so).
\ No newline at end of file
+Dropdowns consist of an opener and the part that contains the items. The latter
+component is portalled to a place in the DOM that prevents is from being clipped
+by an overflow: auto container. This would be either as a child of `body` or the
+modal launcher portal if the dropdown is used in a portal.
```
jeresig: "...is used in a modal." perhaps?

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -24,16 +26,26 @@ type OpenerProps = {|
...
 |};
 
 class ActionMenuOpener extends React.Component<OpenerProps> {
     static defaultProps = {
         disabled: false,
     };
 
+    handleClick = (e: SyntheticEvent<>) => {
+        const {open} = this.props;
+        // TODO: also detect enter click events
```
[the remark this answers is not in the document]
wishdasher: This was fixed by the ClickableBehavior diff, so removing comment!

packages/wonder-blocks-dropdown/components/action-menu.md
wishdasher: Unfortunately Styleguidish doesn't support arrow functions in class field properties -- I've added comments everywhere to explain the binding in the constructor.

packages/wonder-blocks-dropdown/components/action-menu.test.js
wishdasher: Using `window.scrollTo` is better, changed!

packages/wonder-blocks-dropdown/components/option-item.js
wishdasher: At some point it was a `addStyle("button")` but now it's a `View`, so I can remove it!

packages/wonder-blocks-dropdown/components/select-opener.js
wishdasher: Removing comment because we don't want it to detect enters.

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,379 @@
...
+        dropdown.setProps({
+            onOpenChanged: (open) => handleOpen(open),
+            open: true,
+        });
+
+        const event = new MouseEvent("mouseup");
+        document.dispatchEvent(event);
+
+        expect(handleOpen).toHaveBeenCalledTimes(1);
+        expect(handleOpen.mock.calls[0][0]).toBe(false);
```
kevinbarabash: This is b/c we're passing `false` to `handleOpen` b/c it's closed now.

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,379 @@
...
+    });
+
+    it("doesn't close on external mouse click if already closed", () => {
```
kevinbarabash: nice

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,379 @@
...
+            keyboard: false,
+            open: true,
+        });
+
+        dropdown.simulate("keydown", {keyCode: keyCodes.down});
+        dropdown.simulate("keyup", {keyCode: keyCodes.down});
+        expect(dropdown.instance().focusedIndex).toBe(1);
+    });
+
+    it("focuses correct item after clicking on option, then switching to keyboard", () => {
```
kevinbarabash: Cool

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,379 @@
...
+        const option0 = dropdown.find("OptionItem").at(0);
+        const option2 = dropdown.find("OptionItem").at(2);
+
+        expect(dropdown.instance().focusedIndex).toBe(0);
+
+        // Should select option2
+        option0.simulate("keydown", {keyCode: keyCodes.down});
+        option0.simulate("keyup", {keyCode: keyCodes.down});
+        expect(dropdown.instance().focusedIndex).toBe(1);
+        expect(dropdown.instance().focusedOriginalIndex).toBe(2);
```
kevinbarabash: excellent

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,379 @@
...
+                {
+                    component: <OptionItem label="item 2" value="2" key="2" />,
+                    focusable: true,
+                    populatedProps: {},
+                },
+            ],
+        });
+
+        // Should figure out that option1, which is now at index 0, is selected
+        expect(dropdown.instance().focusedIndex).toBe(0);
```
kevinbarabash: It keeps track of the select item while updating the items in the list?!  What happens if the item that was selected is removed from the listed while focused?

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,379 @@
...
+                },
+            ],
+            open: true,
+        });
+
+        const option1 = dropdown.find("OptionItem").at(1);
+
+        option1.simulate("click");
+        expect(onClick1).toHaveBeenCalledTimes(1);
+    });
```
kevinbarabash: These are awesome tests!

packages/wonder-blocks-dropdown/components/dropdown.test.js
wishdasher: Added a comment to clarify!

packages/wonder-blocks-dropdown/components/dropdown.test.js
wishdasher: Good call. What's supposed to happen is that the initial first item gains focus. I'll add a test.

wishdasher: Oh, that's the next test, but I'll make it more explicit.

PR #269, opened 2018-08-08
Use strings instead of arrays for fontFamily.

jeresig:
This was causing problems in older versions of Aphrodite when trying to merge in a `string` value. This fixes that!


jeresig:

Built with commit b25c706083827e6095bf0c192efd5a02869496e5

https://deploy-preview-269--wonder-blocks.netlify.com

PR #270, opened 2018-08-08
Fix OneColumnModal overflow scrolling. Fixes WB-349.

jeresig:


packages/wonder-blocks-modal/components/one-column-modal.md
```diff
@@ -56,8 +56,7 @@ const styles = StyleSheet.create({
 </View>;
 ```
 
-This example shows that `OneColumnModal` works with content and footer elements
-that aren't block level elements.
+This example shows a `OneColumnModal` with content and footer:
```
jeresig: This felt like a more useful example (I prefer to show examples of what we want people to do, rather than anti-patterns).

jeresig:

Built with commit c20eb9a7a866c943ebc09bae2fbe4535c0e7774a

https://deploy-preview-270--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/one-column-modal.md
kevinbarabash: It seems reasonable to pass a single button to the footer.

kevinbarabash: See, that's what you did... you just phrased it differently.  sgtm.  🙂

PR #271, opened 2018-08-08
Grid cell widths don't match specs at >= 1168w

andymatuschak:
http://zpl.io/VQW88JP suggests that at >= 1168w, column widths should be exactly 64px. Unfortunately, they're a few px less than that because the row's max width is set to 1120 (the "maximum content width" specified in the spec), yet the row's max width _includes_ the left and right margins, while the spec's 1120px does not. Adding the left and right margins to the row max width produces cells with the specced widths.

Test Plan:
1. yarn run styleguidist
2. Hit http://localhost:6060/#cell
3. Resize example until Web Inspector shows example interior width at >= 1168px.
4. Cells should be exactly 64px wide.
5. yarn test


jeresig:

Built with commit d509540ae336d02daff8466097bc59f5fe199f4a

https://deploy-preview-271--wonder-blocks.netlify.com

packages/wonder-blocks-grid/components/row.js
```diff
@@ -116,7 +116,13 @@ class Row extends React.Component<{
 
         return (
             <View style={[styles.rowWrap, style]}>
-                <View style={[styles.row, hasMaxWidth && styles.rowMaxWidth]}>
+                <View
+                    style={[
+                        styles.row,
+                        maxWidth && styles.rowMaxWidth,
```
jeresig: 🤷‍♂️ Minor nit: could you change these to be: `!!maxWidth && ...`? Right now it's potentially passing through `0` which I'm not sure we want, I think we want to enforce it to be a boolean.

andymatuschak: Oh, so I actually changed `maxWidth` to be a `number` (the width value itself, fixing your TODO) rather than a boolean. I think that means this is OK, but double-check for me?

jeresig: Yep - exactly! Sorry, to clarify: since `maxWidth` is a number now, if it fails we're going to be passing a number to the `style` prop. Namely, if `maxWidth` is `0` the results will be: `style={[styles.row, 0, ...]}`, which isn't what we want, we'd rather have `style={[styles.row, false, ...]}`. We can fix that by doing a simple `!!maxWidth` to cast the number to become a boolean. Not a big deal, just an edge case that we can fix.

jeresig: (If you do fix it you can just fix it before you land, no need for a re-review!)

andymatuschak: Oh, of course! Thank you for explaining; will do.

PR #276, opened 2018-08-13
Address a few react 16 related TODOs

wishdasher:
Use `getDerivedStateFromProps` in `ClickableBehavior` instead of deprecated method.
- Verified by the existing test case in `clickable-behavior.test.js`

Change a `View` to `React.Fragment` in `ModalLauncher`.
- Manual verification of the modal examples.

### DIFF
--- a/packages/wonder-blocks-core/components/clickable-behavior.js
+++ b/packages/wonder-blocks-core/components/clickable-behavior.js
@@ -177,18 +177,20 @@ export default class ClickableBehavior extends React.Component<Props, State> {
         disabled: false,
     };
 
+    static getDerivedStateFromProps(props: Props, state: State) {
+        // If new props are disabled, reset the hovered/focused/pressed states
+        if (props.disabled) {
+            return startState;
+        } else {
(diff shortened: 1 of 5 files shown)


packages/wonder-blocks-modal/components/modal-launcher.js
```diff
@@ -93,8 +91,7 @@ export default class ModalLauncher extends React.Component<Props, State> {
         });
 
         return (
-            // TODO(mdr): This can be a fragment once we're on React 16.
-            <View>
+            <React.Fragment>
```
kevinbarabash: This is way better as a fragment, thanks!

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -177,20 +177,22 @@ export default class ClickableBehavior extends React.Component<Props, State> {
         disabled: false,
     };
 
+    static getDerivedStateFromProps(props: Props, state: State) {
```
kevinbarabash: snazzy

PR #278, opened 2018-08-13
WB-357 Add example of full-bleed image in two column modal

wishdasher:
https://khanacademy.atlassian.net/browse/WB-357
https://deploy-preview-278--wonder-blocks.netlify.com/#!/TwoColumnModal/5

What I think is happening is that `ModalContent` was shrinking to what it perceived as the size of its inner content and not growing even with `flex: 1`. When the inner content was absolutely positioned, the sizing was also incorrect. This change sets a `minHeight` of `100%` on `ModalContent` so it can't shrink.

Verified that the `ModalContent` components in the other example modals still have the correct padding, in both scrolling contexts and narrow screen contexts.

(Also, if there's a better place to put the image asset, let me know.)

Note: The example I added doesn't match the design spec for this particular example completely, but I don't think the modal is currently built to capture that spec completely either. (For example, I don't see a tablet design in the modal v1 spec.)


packages/wonder-blocks-modal/components/two-column-modal.md
```diff
@@ -193,4 +193,73 @@ const styles = StyleSheet.create({
...
+        backgroundPosition: "center bottom",
+        width: "100%",
+        height: "100%",
+    },
+});
+
+<View style={styles.previewSizer}>
+    <View style={styles.modalPositioner}>
+        <TwoColumnModal
+            sidebar={<View style={styles.imageContainer} />}
```
kevinbarabash: In talking with Taty, she said that using the sidebar for images is the more common case.  I think we should remove the padding from this container and if people need to add text in here on the odd occasion, they use a `ModalPanel`.

wishdasher:
Perhaps I misunderstand the issue. I thought the problem was that we couldn't make the content attach itself to the bottom of the modal instead of the top. Is the goal to make the modal the same height as the larger of the two panes?

I remember some conversation a few weeks ago about how modals are currently specced to take up X% of the height of the viewport and whether they should be allowed to be smaller/shorter, but I don't remember if there was a conclusion.

kevinbarabash:
Here's the zeplin for designs: zpl://screen?sid=5b51186c879619fa662f00ae&pid=5894da12b441ab44040813db.  I believe the intent is that the modal keeps the same dimensions as in the mock even as the browser is resized.

screen shot 2018-08-13 at 8 07 01 pm

wishdasher:
Oh, I see. That does make much more sense in terms of aesthetics. But this means we can no longer obey these constraints on sizing that are currently built into `TwoColumnModal`?
image

kevinbarabash:
I think we should provide a `style` prop so that devs can style modals to match designs.  That being said, I think we also have to be careful to handle cases where there's more text in a text container than is room for.  Make sure that all text contains in modals have `overflow: auto` set should do the trick.

jeresig:
@kevinbarabash I worry about providing a style property that can override the height as it'll almost certainly break on other layouts (tablet, mobile). I think we should seriously consider an option for this (something that causes it to shrink in size to fit the height of the largest child).

kevinbarabash:
> I think we should seriously consider an option for this (something that causes it to shrink in size to fit the height of the largest child).

I think maybe a `fixed` option that matches the size exactly with what's in the Modal Spec.  I believe that one of the goals the design team has with these new modals is reducing the variability.

khirasaki:
The original intent was that the desktop height of the modal was a fixed amount (464px), rather than a variable % of the screen. If the design requires a different size, the designer should explicitly specify that.

However, on tablet/mobile, I agree that we'll need the behavior to be different and %-based. I was out last week and sick this week but am trying to get the tablet spec (and likely revised mobile spec) together today.

Does this help unblock y'all re: desktop sizes?

idreyn:
To add onto Kitt's comment, it would be very useful for our feature work if we could push a fix that allows a full-bleed left hand side while deferring the discussion about modal height. As is we're okay with the behavior in the image above where the image is stuck to the bottom of a very tall modal.

kevinbarabash:
Sorry for derailing this PR by bringing up my concern about size.  Getting back to issue of full-bleed.

> In talking with Taty, she said that using the sidebar for images is the more common case. I think we should remove the padding from this container and if people need to add text in here on the odd occasion, they use a ModalPanel.

I think this is the right path forward.  @wishdasher @jeresig any concerns with this approach?

This change would also bring the behaviour of TwoColumnModal more in line with what we do with the preview area in the StandardModal.

screen shot 2018-08-15 at 5 44 46 pm

khirasaki:
The one place where we currently use a bunch of white text over blue is in the /signup and /login faux-dals, and we've observed that people generally aren't reading that content. So in evolving those designs, we'll want to use a different solution from white-text-on-blue.

So should be reasonable to treat the blue area as primarily for illustration, and secondarily for some other widget/display (e.g. mastery progress). Confirming this in design guild meeting tomorrow, but that should be fine.

wishdasher:
@kevinbarabash ModalPanel isn't an export of the modal package.

One possibility is to add a `fullBleedSidebar` prop to the TwoColumnModal and have a conditional ModalPanel wrapping the sidebar content in there. We can use an `` as the sidebar and it would work.

I like the current solution because the image is a background image on a View, which I think is cleaner than using ``. I think this necessitates that its wrapper  is `position: absolute`, which effectively gives it the fullbleed effect anyways.

wishdasher:
Okay, I decided to add the `fullBleedSidebar` prop, default `true`.

packages/wonder-blocks-modal/generated-snapshot.test.js
```diff
@@ -710,6 +713,82 @@ describe("wonder-blocks-modal", () => {
...
+            },
+
+            imageContainer: {
+                position: "absolute",
+                width: "100%",
+                height: "100%",
+                backgroundImage: "url('pencilHand-800x.png')",
+                backgroundSize: "contain",
+                backgroundRepeat: "no-repeat",
+                backgroundPosition: "center bottom",
```
kevinbarabash: Can you change this to use an `` tag to show that `fullBleedSidebar` is working?

Offtopic: `fullBleedSidebar` would make an awesome name for a metal band.

wishdasher: I added an example with `` but also kept the original to show that a background image is good for circumstances where the image is purely background thing and shouldn't be dragged around.

wishdasher: `fullBleedSidebar` 🤘

PR #280, opened 2018-08-14
Update ModalLauncher to use ReactDOM.createPortal

kevinbarabash:
This is the first step in support tooltips in webapp.  We need both modal and tooltip to be using React 16 portals.  This PR update ModalLauncher to use this new API.  For portals hosted within the modal such as the dropdown portal, we use the modal backdrop as the container so that the dropdown isn't clipped by the modal dialog.

Test Plan:
- load http://localhost:6060/#!/MultiSelect
- open the modal
- click on the multi-select opener, see that the dropdown isn't clipped by the modal dialog

screen shot 2018-08-14 at 4 41 04 pm


jeresig:

Built with commit 6cff13fefa2337e3d6010417e6b55322e33cf5a0

https://deploy-preview-280--wonder-blocks.netlify.com

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.js
```diff
@@ -29,11 +29,5 @@ function maybeGetNextAncestorModalLauncherPortal(element: ?(Element | Text)) {
 export default function maybeGetPortalMountedModalHostElement(
```
jeresig: Looks like we can just get rid of this function - it's not doing anything now!

packages/wonder-blocks-modal/components/modal-launcher-portal.test.js
```diff
@@ -1,66 +0,0 @@
...
-        const wrapper = mount(
-            <ModalLauncherPortal>
-                <div ref={childrenRef} />
-            </ModalLauncherPortal>,
-        );
-
-        unmounted = true;
-        wrapper.unmount();
-    });
-});
```
jeresig: Love that all of this is going away now, great work!

packages/wonder-blocks-modal/components/modal-launcher.test.js
```diff
@@ -1,19 +1,26 @@
 // @flow
 import React from "react";
-import {shallow, mount} from "enzyme";
+import {shallow} from "enzyme";
 
+import {mount, unmountAll} from "../../../utils/testing/mount.js";
```
somewhatabstract: Yes! 🎉

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.js
somewhatabstract: I think maybe this can go away once everything is using new portals and we've confirmed the modal isn't weirdly clipping other portal'd things.

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.test.js
```diff
@@ -113,7 +113,7 @@ describe("maybeGetPortalMountedModalHostElement", () => {
                     // Assert
                     expect(result).toBeTruthy();
 
-                    const modalPortalElement = result && result.parentElement;
+                    const modalPortalElement = result;
```
somewhatabstract: Huh, interesting.

Why did the test need to change? This implies to me that the method is not returning the right element as the portal element should be nested in something, even if it's just the body.

wishdasher: I think it's because this new way of doing the portal removes an extra `div` that was originally there. (`maybeGetPortalMountedModalHostElement` was returning the `firstElementChild` but now doesn't do that)

somewhatabstract: It was doing that because, at least for the tooltip, if we didn't, the tooltip appeared obscured by the modal backdrop. If that's no longer happening, then awesome.

kevinbarabash: The dropdown portal is the last child of the backdrop.  @wishdasher I'm curious to see if this affects your keyboard navigation, hopefully it won't.

screen shot 2018-08-15 at 3 50 01 pm

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.js
kevinbarabash: I think we can get rid of this eventually but it's a non-trivial amount of work b/c we need to figure out a different way of figuring out what a portal's host/container element should be.

kevinbarabash: I'll add a TODO.

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.test.js
wishdasher: It does :( but even before this change, it was already the last child of the backdrop. There's some wonky behavior with the FocusTrap that I'm planning on looking into after the keyboard nav is merged (meaning that pull request #268 doesn't work with the modal but I am aware of it).

kevinbarabash: Interesting... maybe we can introduce another container that lives inside the tab trap and that's where we add the modal dialog, any portals, as well as the left/right bumpers whenever we get around to adding them.

kevinbarabash: I'm going to work on the proposal I made in my last comment.

wishdasher: Awesome! I have a bit of extra context. When I was doing this, one issue I ran into was that the dropdown had to be a sibling of the dialog to be able to extend past the dialog div, but then these lines in `FocusTrap`
```
        const modalRoot = this._getModalRoot();
        if (modalRoot.contains(target)) {
            // If the newly focused node is inside the modal, we just keep track
            // of that.
            this._lastNodeFocusedInModal = target;
        } else {
```
weren't able to trap the focus events happening on the dropdown, which is why if you look at the MultiSelect in a modal on netlify, clicking anything in the dropdown focuses the close icon.

I briefly tried changing `_getModalRoot` to get its parent instead, but I think that messed up the "trap" part and tabbing was able to escape to the actual browser.

kevinbarabash: I have a fix in https://github.com/Khan/wonder-blocks/pull/283.

PR #281, opened 2018-08-14
Port Tooltip to React 16 portals

kevinbarabash:
I still need to switch over to the React 16 portal, but I think by having the portal side-by-side with the anchor instead of nested it should make the switch easier.  The changes I’ve made seem to work.  Let me know if what I’m doing is totally silly.


jeresig:

Built with commit 78a8e2ae62fe3702e8e49eb0280bec40210a651d

https://deploy-preview-281--wonder-blocks.netlify.com

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
```diff
@@ -51,7 +52,7 @@ export default class TooltipPortalMounter extends React.Component<Props> {
     }
 
     _getPortalParent() {
-        const anchorNode = ReactDOM.findDOMNode(this);
+        const anchorNode = this.props.anchorNode; //  ReactDOM.findDOMNode(this);
```
somewhatabstract: This should still be a DOM node look up as the anchor could be anything including a component, hence why it was as it was before.

kevinbarabash:
The automated tests are currently failing.  I'll fix those later, but things seem to be working in the manually testing I did.
screen shot 2018-08-15 at 11 57 37 am
screen shot 2018-08-15 at 11 57 11 am
screen shot 2018-08-15 at 11 57 04 am

packages/wonder-blocks-tooltip/components/tooltip.js
```diff
@@ -177,18 +181,19 @@ export default class Tooltip extends React.Component<Props, State> {
...
+                    anchorRef={(r) => this._updateAnchorElement(r)}
+                    onActiveChanged={(active) => this.setState({active})}
+                >
+                    {this._renderAnchorElement(ids)}
+                </TooltipAnchor>
+                {this.state.active &&
+                    <TooltipPortalMounter anchorNode={this.state.anchorElement}>
+                        {this._renderPopper(this.state.active, ids)}
+                    </TooltipPortalMounter>}
+            </React.Fragment>
```
kevinbarabash: I've reworked this some more so that there's no nesting at all between the tooltip or the portal.  React.Fragment allows us to do this would introducing any sort of wrapper component around the anchor except in the case of strings where it's actually needed.

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
kevinbarabash: The second lookup shouldn't be necessary anymore since the anchor is passed directly as a child to TooltipAnchor now instead of via children-as-function.

kevinbarabash:
I still have to cleanup tooltip-anchor.test.js but things are working with new React portal.

packages/wonder-blocks-tooltip/components/tooltip.js
somewhatabstract: BTW, since we're using React 16, we can use the react-popper components as they were intended and drop the need for passing refs around the way we do.

I believe it's something like this.

```
<Manager>
    <Reference>
        ...anchor...
     </Reference>
     <Popper>
        ....portal bubble stuff....
     </Popper>
</Popper.Manager>
```

Full details on the react-popper repo and site.

somewhatabstract: https://github.com/FezVrasta/react-popper#usage

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
```diff
@@ -429,33 +421,41 @@ describe("TooltipAnchor", () => {
...
             const ref = await new Promise((resolve) => {
-                const getFakeTooltipPortalMounter = (active) =>
-                    (((
-                        <View>{active ? "true" : "false"}</View>
-                    ): any): React.Element<typeof TooltipPortalMounter>);
                 const nodes = (
-                    <TooltipAnchor anchorRef={resolve}>
-                        {getFakeTooltipPortalMounter}
-                    </TooltipAnchor>
+                    <View>
```
kevinbarabash: The `View` isn't necessary... I was copy pasting from an example where it was used.  I'm going to get rid of it so we don't have unnecessary code changes in this PR.

kevinbarabash:
This is ready for review now.

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -6,21 +6,23 @@
 import * as React from "react";
 import * as ReactDOM from "react-dom";
 
-import TooltipPortalMounter from "./tooltip-portal-mounter.js";
+import {Text as WBText} from "@khanacademy/wonder-blocks-core";
```
jeresig: Why WBText?? This seems really awkward.

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -6,21 +6,23 @@
...
 
 type Props = {|
     /**
-     * A method that renders the content for anchoring the tooltip.
-     * This must return a TooltipPortalMounter component.
+     * The content for anchoring the tooltip.
+     * This component will be used to position the tooltip.
      */
-    children: (active: boolean) => React.Element<typeof TooltipPortalMounter>,
+    children: React.Element<any> | string,
```
jeresig: It took me a little while to understand this. It would be good to call out what we're looking for in the comment for this prop as we only want a React Element or a string (very explicitly) and a string is being wrapped with Text (which we should make clear in the documentation).

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -63,13 +78,20 @@ export default class TooltipAnchor extends React.Component<Props, State>
...
         forceAnchorFocusivity: true,
     };
 
-    state = {
-        active: false,
-    };
+    constructor(props: Props) {
+        super(props);
+
+        this._focused = false;
```
jeresig: Why did you move these to be inside of a constructor instead of leaving them as a global like before?

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -268,6 +291,24 @@ export default class TooltipAnchor extends React.Component<Props, State>
...
+                <WBText>{children}</WBText>
+            ) : (
+                children
+            );
+
+        if (ids) {
+            return React.cloneElement(anchorableChildren, {
+                "aria-describedby": ids.get(TooltipAnchor.ariaContentId),
+            });
+        } else {
```
jeresig: This `else` isn't necessary, you can just do `return anchorableChildren;`

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
```diff
@@ -1,158 +0,0 @@
-// @flow
```
jeresig: Very excited to see this go away!

packages/wonder-blocks-tooltip/components/tooltip.js
```diff
@@ -175,20 +158,27 @@ export default class Tooltip extends React.Component<Props, State> {
     }
 
     _renderTooltipAnchor(ids?: IIdentifierFactory) {
-        const {forceAnchorFocusivity} = this.props;
+        const {children, forceAnchorFocusivity} = this.props;
+        const {active, anchorElement} = this.state;
+
+        const popperHost =
+            maybeGetPortalMountedModalHostElement(anchorElement) ||
+            document.querySelector("body");
```
jeresig: Probably better to just do `document.body` -- if you're worried about it being null then `document.querySelector` isn't going to be better (from a Flow perspective).

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
kevinbarabash: I'll add a comment but it's b/c there's a React.Text type... I guess I could check for that instead of `Text` and then this could be simply `Text`.

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
kevinbarabash: A string can be wrapped int a `Text` node but other things can like booleans, arrays, etc.  I guess we could but it's probably a mistake to be passing those things as children.

kevinbarabash: I'll add a comment.

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
kevinbarabash: I don't like where our linter requests class properties be initialized which is after the lifecycle methods.

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
kevinbarabash: Will do.

packages/wonder-blocks-tooltip/components/tooltip.js
kevinbarabash: yeah, good point

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
somewhatabstract: I had this issue somewhere too. I don't think `React.Text` worked for me, so `WBText` was the way I went. I recommend doing the same.

packages/wonder-blocks-tooltip/components/tooltip-portal-mounter.js
somewhatabstract: Also sad because of all the lessons and effort in getting it working.

somewhatabstract:
> The diff coverage is 100%.

😻

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -268,6 +291,24 @@ export default class TooltipAnchor extends React.Component<Props, State>
...
+        // If the content is just a string, we wrap it in a Text element
+        // so as not to affect styling or layout but still have an element
+        // to anchor to.
+        const {children, ids} = this.props;
+        const anchorableChildren =
+            typeof children === "string" ? (
+                <WBText>{children}</WBText>
+            ) : (
+                children
+            );
```
somewhatabstract: Can you put this into a separate method. I think it will be cleaner.

```
const children = this._getAnchorableChildren();
```

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -268,6 +291,24 @@ export default class TooltipAnchor extends React.Component<Props, State>
...
+
+        if (ids) {
+            return React.cloneElement(anchorableChildren, {
+                "aria-describedby": ids.get(TooltipAnchor.ariaContentId),
+            });
+        } else {
+            return anchorableChildren;
+        }
```
somewhatabstract: Again, I think this should be a method like:

```
return this._renderAccessibleChildren()
```

I think this helps make the code self-documenting without needing maintainers to read and understand every line.

packages/wonder-blocks-tooltip/components/tooltip.js
```diff
@@ -175,20 +158,27 @@ export default class Tooltip extends React.Component<Props, State> {
...
+                    anchorRef={(r) => this._updateAnchorElement(r)}
+                    onActiveChanged={(active) => this.setState({active})}
+                    ids={ids}
+                >
+                    {children}
+                </TooltipAnchor>
+                {popperHost &&
+                    active &&
+                    ReactDOM.createPortal(this._renderPopper(ids), popperHost)}
+            </React.Fragment>
```
somewhatabstract: Check out the `react-popper` docs. I think that this could all be simplified by using their React 16-friendly syntax.

```
<Manager>
    <Reference>
       ...anchor here
    </Reference>
    <Popper>
    </Popper>
</Reference>
```

I think their docs even mention working with portals.

That said, perhaps for a different diff. This seems like it would work ok.

packages/wonder-blocks-tooltip/components/tooltip.js
somewhatabstract: Wrapping this in a `_getHost` method makes the code a little easier to read and encapsulates any extra logic we might need (not that there should be any).

PR #282, opened 2018-08-15
Update ClickableBehavior behavior across browsers and components

wishdasher:
After some invesigation and discussion, I wrote up [this document](https://docs.google.com/document/d/1DG5Rg2f0cawIL5R8UqnPQpd7pbdObk8OyjO5ryYQmBM/edit?usp=sharing) to describe the expected behavior of ClickableBehavior.

The main change here is to allow components that use ClickableBehavior to indicate whether they should be triggered on enter, or space, or both. This change also fixes a bug where a mouse click that directly followed a space click sometimes wasn't registered and attempts to normalize the ClickableBehavior across major browsers.

I added unit tests for the new cases, but since browsers have slightly different behavior in where event defaults happen, they're not perfect or fully comprehensive.

Test plan prereq:
* On Safari, may have to enable Preferences -> Advanced -> Check "Press Tab to highlight each item..."
* On Firefox (macOS), may have to go into System Preferences -> Keyboard -> Shortcuts -> Change Full Keyboard Access to "All controls"

Edge was the least well tested due to lagginess in BrowserStack.

Basic manual test plan (on Chrome, Firefox, Safari, Edge):
### Button
- triggers once on space (not until keyup), on enter (not until keyup), on mouse click, on touch
- also test Button with both `href` and `onClick` to make sure both happen (fixing https://khanacademy.atlassian.net/browse/WB-369)
### Link
- triggers once on enter (not until keyup), on mouse click, on touch
- space _doesn't_ trigger click but does scroll
### Checkbox
- triggers once on space, on mouse click, on touch
- doesn't trigger on enter

### DIFF
--- a/packages/wonder-blocks-button/components/button-core.js
+++ b/packages/wonder-blocks-button/components/button-core.js
@@ -29,12 +29,6 @@ const StyledLink = addStyle(Link);
 export default class ButtonCore extends React.Component<Props> {
     static contextTypes = {router: PropTypes.any};
 
-    handleClick = (e: SyntheticEvent<>) => {
-        if (this.props.disabled) {
-            e.preventDefault();
-        }
-    };
-
     render() {
         const {
             children,
@@ -76,6 +70,7 @@ export default class ButtonCore extends React.Component<Props> {
         const commonProps = {
             "aria-disabled": disabled ? "true" : undefined,
             "data-test-id": testId,
+            role: "button",
             style: [defaultStyle, style],
             ...handlers,
         };
@@ -86,25 +81,21 @@ export default class ButtonCore extends React.Component<Props> {
 
         if (href) {
             return router && !skipClientNav ? (
-                <StyledLink
-                    {...commonProps}
-                    onClick={this.handleClick}
-                    to={href}
-                >
+                <StyledLink {...commonProps} to={href}>
                     {label}
                 </StyledLink>
             ) : (
-                <StyledAnchor
-                    {...commonProps}
-                    onClick={this.handleClick}
-                    href={href}
-                >
+                <StyledAnchor {...commonProps} href={href}>
                     {label}
                 </StyledAnchor>
             );
         } else {
             return (
-                <StyledButton {...commonProps} disabled={disabled}>
+                <StyledButton
+                    type="button"
+                    {...commonProps}
+                    disabled={disabled}
+                >
                     {label}
                 </StyledButton>
             );
--- a/packages/wonder-blocks-core/components/clickable-behavior.js
+++ b/packages/wonder-blocks-core/components/clickable-behavior.js
@@ -22,9 +22,9 @@ type Props = {|
      * A URL.
      *
      * If specified, we will assume the rendered component should react like an
-     * `<a>` tag and respond to enter/return key press. Otherwise, we will
-     * assume the rendered component should react like a `<button>` tag and
-     * respond to spacebar key press.
+     * `<a>` tag and respond to enter/return key press but not space key press.
+     * If not specified, the component will react to both enter and space key
+     * presses.
      */
     href?: string,
 
@@ -35,8 +35,26 @@ type Props = {|
 
     /**
      * Passed in by withRouter HOC.
+     * @ignore
      */
     history?: any,
+
+    /**
+     * Trigger onClick callback and href navigation, if present, on enter key
+     * press. Default true. Only set to false if the component should definitely
+     * NOT trigger onClick on enter. An example of a component that shouldn't
+     * trigger on enter is Checkbox.
+     * @ignore
+     */
+    triggerOnEnter: boolean,
+
+    /**
+     * Trigger onClick callback on space key press. Only set to false if the
+     * component should definitely NOT trigger onClick on space. An example of a
+     * component that shouldn't trigger on space is Link.
+     * @ignore
+     */
+    triggerOnSpace: boolean,
 |};
 
 type State = {|
@@ -126,6 +144,17 @@ const startState = {
  * 2. Keydown (spacebar/enter) -> active state
  * 3. Keyup (spacebar/enter) -> focus state
  *
+ * Warning: The event handlers returned (onClick, onMouseEnter, onMouseLeave,
+ * onMouseDown, onMouseUp, onTouchStart, onTouchEnd, onTouchCancel, onKeyDown,
+ * onKeyUp, onFocus, onBlur, tabIndex) should be passed on to the component
+ * that has the ClickableBehavior. You cannot override these handlers without
+ * potentially breaking the functionality of ClickableBehavior.
+ *
+ * There are internal props triggerOnEnter and triggerOnSpace that can be set
+ * to false if one of those keys shouldn't count as a click on this component.
+ * Be careful about setting those to false -- make certain that the component
+ * shouldn't process that key.
+ *
  * `ClickableBehavior` accepts a function as `children` which is passed state
  * and an object containing event handlers. The `children` function should
  * return a clickable React Element of some sort.
@@ -171,10 +200,12 @@ const startState = {
  */
 export default class ClickableBehavior extends React.Component<Props, State> {
     waitingForClick: boolean;
-    keyboardClick: boolean;
+    enterClick: boolean;
 
     static defaultProps = {
         disabled: false,
+        triggerOnEnter: true,
+        triggerOnSpace: true,
     };
 
     static getDerivedStateFromProps(props: Props, state: State) {
@@ -194,9 +225,8 @@ export default class ClickableBehavior extends React.Component<Props, State> {
     }
 
     handleClick = (e: SyntheticMouseEvent<>) => {
-        if (this.keyboardClick) {
-            this.keyboardClick = false;
-            e.preventDefault();
+        if (this.enterClick) {
+            return;
         } else if (this.props.onClick) {
             this.waitingForClick = false;
             this.props.onClick(e);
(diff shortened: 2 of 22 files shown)


packages/wonder-blocks-button/components/button-core.js
```diff
@@ -86,25 +81,21 @@ export default class ButtonCore extends React.Component<Props> {
 
         if (href) {
             return router && !skipClientNav ? (
-                <StyledLink
-                    {...commonProps}
-                    onClick={this.handleClick}
```
wishdasher: A few components were overriding the `onClick` callback after passing the handlers. I think this is from an earlier change where the default disabled `onClick` handler did something different. This was bad because the `onClick` prop was never called.

Verified that disabled components don't do anything.

packages/wonder-blocks-dropdown/components/action-menu.test.js
```diff
@@ -55,12 +55,10 @@ describe("ActionMenu", () => {
         // Close menu with space
         opener.simulate("keydown", {keyCode: keyCodes.space});
         opener.simulate("keyup", {keyCode: keyCodes.space});
-        opener.simulate("click", {preventDefault: jest.fn()});
```
wishdasher: Because these components are mounted, the `keydown`/`keyup` events trigger a click.

packages/wonder-blocks-icon-button/components/icon-button-core.js
```diff
@@ -113,25 +107,21 @@ export default class IconButtonCore extends React.Component<Props> {
...
-                >
+                <StyledAnchor {...commonProps} href={href}>
                     {child}
                 </StyledAnchor>
             );
         } else {
             return (
-                <StyledButton {...commonProps} disabled={disabled}>
+                <StyledButton
+                    type="button"
```
wishdasher: Default is `type="submit"`. `type="button"` feels more appropriate.

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -86,25 +81,21 @@ export default class ButtonCore extends React.Component<Props> {
...
-                >
+                <StyledAnchor {...commonProps} href={href}>
                     {label}
                 </StyledAnchor>
             );
         } else {
             return (
-                <StyledButton {...commonProps} disabled={disabled}>
+                <StyledButton
+                    type="button"
```
wishdasher: Default is `type="submit"`. `type="button"` feels more appropriate.

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -76,6 +70,7 @@ export default class ButtonCore extends React.Component<Props> {
         const commonProps = {
             "aria-disabled": disabled ? "true" : undefined,
             "data-test-id": testId,
+            role: "button",
```
wishdasher: Because we style the `Button` component to look like a button, even when the tag is `<a>`, we want the role to be `button` so users can expect that space also works as a way to trigger this.

packages/wonder-blocks-core/components/clickable-behavior.test.js
```diff
@@ -11,6 +12,18 @@ const keyCodes = {
...
+    // Note: window.location.assign needs a mock function in the testing
+    // environment.
+    window.location.assign = jest.fn();
+
+    beforeEach(() => {
+        unmountAll();
+    });
+
+    afterEach(() => {
+        window.location.assign.mockClear();
```
wishdasher: Because this is a global, need to reset it because one of the tests test how many times this was called.

packages/wonder-blocks-dropdown/components/select-opener.js
```diff
@@ -84,14 +84,19 @@ export default class SelectOpener extends React.Component<SelectOpenerProps> {
...
+                triggerOnEnter={false}
+            >
                 {(state, handlers) => {
                     const stateStyles = _generateStyles(light, {...state});
                     const {hovered, focused, pressed} = state;
                     return (
                         <StyledButton
                             disabled={disabled}
-                            role="menu"
+                            role="button"
```
jeresig: Hmm, this is an interesting change - just confirming that this makes sense?

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -35,8 +35,26 @@ type Props = {|
...
      * Passed in by withRouter HOC.
+     * @ignore
      */
     history?: any,
+
+    /**
+     * Trigger onClick callback and href navigation, if present, on enter key
+     * press. Default true. Only set to false if the component should definitely
+     * NOT trigger onClick on enter. An example of a component that shouldn't
+     * trigger on enter is Checkbox.
```
jeresig: "...is a Checkbox or other form element." perhaps?

packages/wonder-blocks-button/components/button-core.js
jeresig: This is a really awesome improvement, thank you!

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -126,6 +144,17 @@ const startState = {
...
+ * onMouseDown, onMouseUp, onTouchStart, onTouchEnd, onTouchCancel, onKeyDown,
+ * onKeyUp, onFocus, onBlur, tabIndex) should be passed on to the component
+ * that has the ClickableBehavior. You cannot override these handlers without
+ * potentially breaking the functionality of ClickableBehavior.
+ *
+ * There are internal props triggerOnEnter and triggerOnSpace that can be set
+ * to false if one of those keys shouldn't count as a click on this component.
+ * Be careful about setting those to false -- make certain that the component
+ * shouldn't process that key.
+ *
```
jeresig: It'd be awesome to link to the doc you wrote up in here, as well! I think people would find it to be very useful when trying to understand this problem better.

packages/wonder-blocks-button/components/button-core.js
kevinbarabash: Is this what was causing `onClick` not to be called when `href` was set?

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -22,9 +22,9 @@ type Props = {|
      * A URL.
      *
      * If specified, we will assume the rendered component should react like an
-     * `<a>` tag and respond to enter/return key press. Otherwise, we will
-     * assume the rendered component should react like a `<button>` tag and
-     * respond to spacebar key press.
+     * `<a>` tag and respond to enter/return key press but not space key press.
+     * If not specified, the component will react to both enter and space key
+     * presses.
```
kevinbarabash: How does this jibe with want the same behaviour for things that look like buttons but actually use <a> or <button> under the hood?

kevinbarabash: It might be worth calling out that `triggerOnEnter` and `triggerOnSpace` override this behaviour.

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -35,8 +35,26 @@ type Props = {|
...
     history?: any,
+
+    /**
+     * Trigger onClick callback and href navigation, if present, on enter key
+     * press. Default true. Only set to false if the component should definitely
+     * NOT trigger onClick on enter. An example of a component that shouldn't
+     * trigger on enter is Checkbox.
+     * @ignore
+     */
+    triggerOnEnter: boolean,
```
kevinbarabash: Why are these marked as `@ignore`?  Someone might want to make a clickable `View` that behaves like a button but has an href.  I wonder if having a `role` property that allows people to set an aria role would be better.

packages/wonder-blocks-core/components/clickable-behavior.test.js
```diff
@@ -316,4 +360,185 @@ describe("ClickableBehavior", () => {
...
+
+        // Navigation didn't happen with space
+        expect(window.location.assign).toHaveBeenCalledTimes(0);
+
+        // Enter press should trigger the onClick after keyup
+        link.simulate("keydown", {keyCode: keyCodes.enter});
+        expect(onClick).toHaveBeenCalledTimes(0);
+
+        // Navigation doesn't happen until after enter is released
+        expect(window.location.assign).toHaveBeenCalledTimes(0);
```
kevinbarabash: I was wondering how to write a test for that.  Nice!

packages/wonder-blocks-core/components/clickable-behavior.test.js
```diff
@@ -316,4 +360,185 @@ describe("ClickableBehavior", () => {
...
+
+        link.simulate("keyup", {keyCode: keyCodes.enter});
+        expect(onClick).toHaveBeenCalledTimes(1);
+
+        // Navigation happened after enter click
+        expect(window.location.assign).toHaveBeenCalledTimes(1);
```
kevinbarabash: Yahoo!

packages/wonder-blocks-core/components/clickable-behavior.test.js
```diff
@@ -316,4 +360,185 @@ describe("ClickableBehavior", () => {
...
+        button.simulate("keydown", {keyCode: keyCodes.space});
+        expect(onClick).toHaveBeenCalledTimes(1);
+        button.simulate("keyup", {keyCode: keyCodes.space});
+        expect(onClick).toHaveBeenCalledTimes(2);
+    });
+
+    // This tests the case where we attach the handlers to an element that is
+    // not canonically clickable (like a div). The browser doesn't naturally
+    // trigger keyboard click events for such an element.
+    it("calls onClick listener on space/enter with a non-usually clickable element", () => {
```
kevinbarabash: I love this title.

packages/wonder-blocks-core/components/clickable-behavior.test.js
```diff
@@ -316,4 +360,185 @@ describe("ClickableBehavior", () => {
...
+            </ClickableBehavior>,
+        );
+
+        // Enter press should not do anything
+        checkbox.simulate("keydown", {keyCode: keyCodes.enter});
+        // This element still wants to have a click on enter press
+        checkbox.simulate("click");
+        checkbox.simulate("keyup", {keyCode: keyCodes.enter});
+        expect(onClick).toHaveBeenCalledTimes(0);
+    });
```
kevinbarabash: Thank you for all of these awesome tests!

packages/wonder-blocks-dropdown/components/action-menu.test.js
kevinbarabash: I wondered about that when I first saw that.  The differences between `shallow` and `mount` a interesting.

packages/wonder-blocks-dropdown/components/select-opener.js
kevinbarabash: If this isn't a menu anymore, does the dropdown itself have role="menu"?

packages/wonder-blocks-button/components/button-core.js
wishdasher: Yeah.

packages/wonder-blocks-dropdown/components/select-opener.js
wishdasher: I did some googling, looks like it should be `listbox`, so I'll change that.

packages/wonder-blocks-core/components/clickable-behavior.js
wishdasher: Ooh I forgot to update this comment, thanks! (This was the previous way of doing this.)

packages/wonder-blocks-core/components/clickable-behavior.js
wishdasher: Okay, I decided to expose these.

packages/wonder-blocks-link/components/link.js
```diff
@@ -124,7 +124,12 @@ export default class Link extends React.Component<SharedProps> {
         );
 
         return (
-            <ClickableBehavior disabled={false} onClick={onClick} href={href}>
+            <ClickableBehavior
+                disabled={false}
+                onClick={onClick}
+                href={href}
+                triggerOnSpace={false}
```
kevinbarabash: I was thinking we could replace `triggerOnSpace` and `triggerOnEnter` with `role`.  This prop would  take only those aria roles ("button", "link", etc.) that can be activated by pressing space or enter.  `ClickableBehavior` would be responsible for making the decision (based on that prop) about when to trigger the `onClick` handler.  We could even pass the `role` as part of `handlers` which should probably be renamed `childProps` since it also include `tabindex`.

This is part of a bigger plan I have to allow standard aria props (including `role`) on more of our components.  Obviously not all components need `role` since for `Button` the role should always be button, but being able to set `aria-label`, `aria-labelledby`, etc. on things in a way that matches how it's done on tags should be a win.

I've been meaning to write this all up in a doc, but haven't had time.

wishdasher: I'm kind of worried that the list that I create wouldn't be exhaustive and since ClickableBehavior is exported as a component to use in webapp, if someone tries to use it for a non-existent but valid role, ClickableBehavior would have to be updated here and then republished.

I'm also worried about this PR getting too big, so it would be more optimal for me to merge this and work on role-related behavior separately.

If there's a concern about the `trigger` props leaking into webapp before the role stuff could be added, I can set `@ignore` on them again.

kevinbarabash: Yeah let's split off the `role` work.  I terms of people using it for unsupported roles we can document which roles it supports and expand it as necessary.  TBH, I don't see people using this component directly very often in webapp.

PR #283, opened 2018-08-15
move the backdrop inside the focus trap so that focus in dropdowns works properly

kevinbarabash:
The FocusTrap sets up "sentinel" elements in the DOM and works to guarantee that focus always stays within the sentinels.  Unfortunately, we were mounting portals for things like tooltips and dropdowns that live inside modals, outside of these sentinels.  The result is that clicking an item in a dropdown would immediately focus the close button.  This PR fixes that but moving the whole ModalBackdrop inside the FocusTrap (and inside the sentinels).  This will also come in handy when we get around to adding the prev/next bumpers for navigating between content modals.  This also makes the FocusTrap easier to reuse b/c now it's responsible for setting up the sentinels.

Test Plan:
- load http://localhost:6060/#!/MultiSelect
- open the modal
- open the dropdown
- select an item, see that the close button isn't focused
- click on the backdrop to dismiss the modal
- use the keyboard to reopen the modal
- press space to open the dropdown
- tab through the items and select some of them with space
- tab though all of the items and keep pressing tab until you get back to dropdown, at no point should focus leave the modal or its contents

Unfortunately the FocusTrap can't be tested with enzyme + jest.  We'll need to e2e tests using vite for that.


jeresig:

Built with commit 94366c6a9720bc8a4e858a2c1c2590fff66dd62d

https://deploy-preview-283--wonder-blocks.netlify.com

kevinbarabash:
@wishdasher good catch.  I need to fix how the FocusTrap gets the root of the modal.  Right now it's grabbing a node outside of the sentinels.  I just need to move it to be something inside.

packages/wonder-blocks-modal/components/focus-trap.js
```diff
@@ -141,6 +134,27 @@ export default class FocusTrap extends React.Component<{children: React.Node}> {
...
+                  * and move focus back into the modal! So, we add focusable
+                  * sentinel nodes. That way, tabbing out of the modal should
+                  * take you to a sentinel node, rather than taking you out of
+                  * the document. These sentinels aren't critical to focus
+                  * wrapping, though; we're resilient to any kind of focus
+                  * shift, whether it's to the sentinels or somewhere else! */}
+                <div tabIndex="0" />
+                <View
+                    ref={(node) =>
+                        (this.modalRoot = (ReactDOM.findDOMNode(node): any))
```
kevinbarabash: I should create a ref in the constructor and use it here.

kevinbarabash:
I wasn't able to call `findDOMNode` on a ref created with `createRef` without `flow` complaining so I ended up pre-binding the method to get the modalRoot.

packages/wonder-blocks-modal/components/focus-trap.js
```diff
@@ -18,48 +20,63 @@ import * as ReactDOM from "react-dom";
...
+    lastNodeFocusedInModal: ?Node;
 
     /**
      * Whether we're currently applying programmatic focus, and should therefore
      * ignore focus change events.
      */
-    _ignoreFocusChanges: boolean;
+    ignoreFocusChanges: boolean;
+
+    modalRoot: Node;
```
jeresig: Can you add a comment for `modalRoot` explaining what it's used for?

jeresig: I'm also surprised that Flow is ok with this. Why do we need a `?Node` for `lastNodeFocusedInModal` but not for `modalRoot`? It does seem like there is the potential for it to be undefined, correct?

packages/wonder-blocks-modal/components/focus-trap.js
```diff
@@ -113,34 +130,49 @@ export default class FocusTrap extends React.Component<{children: React.Node}> {
...
             // node as the last node focused in the modal.
-            this._lastNodeFocusedInModal = document.activeElement;
+            this.lastNodeFocusedInModal = document.activeElement;
         }
     };
 
     render() {
-        return this.props.children;
+        return (
+            <View>
```
jeresig: It seems like this could be a React.Fragment, correct?

jeresig: Glad that we've moved this logic to be inside of the FocusTrap! We may want to consider exposing FocusTrap publicly, I think people would find it to be really useful.

packages/wonder-blocks-modal/components/focus-trap.js
jeresig: We should probably also do `this.modalRoot = null;` in the constructor, too.

kevinbarabash: I can change it to ?Node... I wonder why `flow` doesn't complain about this.

packages/wonder-blocks-modal/components/focus-trap.js
kevinbarabash: I thought about it.  It felt weird, but that's such a squishy argument.  It's not like we're relying on the layout properties of the View as a container in this case.

kevinbarabash:
@matchu focus marshal looks interesting.  I'm curious why they need to render something at the top-level.

matchu:
I think it's probably to coordinate across multiple components with their own focus traps? So that if you render a modal on top of another modal, then move focus into the second modal, the first modal doesn't complain? 🤷‍♀️ just speculating

kevinbarabash:
So you could jump between different focus traps in different portals?  I wonder how often you'd need to do that.  I guess if we do need to do that we have a potential solution.

PR #287, opened 2018-08-16
WB-375 Button text shouldn't wrap, WB-376 Control elements shouldn't …

wishdasher:
…have selectable text

I went over the dicussion [here](https://ux.stackexchange.com/questions/83184/is-user-select-none-bad-ux) and particularly this paragraph
> The goal in employing user-select: none should always be limited to reducing user frustration from unintended selections resulting from interaction that can easily cause them when attempting to carry out the primary task purpose of a control it is used on.

to determine which existing elements to add `user-select: none` to. Those ended up being Button, ActionMenuOpener, SelectOpener, menu items, and Choice labels.

The potentially debatable one would be the Choice labels, as clicking directly on the checkbox / radio button avoids this issue.

Button no longer wraps, but it also truncates if it's long and there's a `maxWidth` set. The documentation warns against setting this, however.

### DIFF
--- a/packages/wonder-blocks-button/components/button.md
+++ b/packages/wonder-blocks-button/components/button.md
@@ -363,13 +363,14 @@ const styles = StyleSheet.create({
 </View>
 ```
 
-Layouts often specify a specific width of button.  When implementing such
-designs use `minWidth` instead of `width`.  `minWidth` allows the button
-to resize to fit the content whereas `width` does not.  This is important
+Layouts often specify a specific width of button. When implementing such
+designs use `minWidth` instead of `width`. `minWidth` allows the button
+to resize to fit the content whereas `width` does not. This is important
 for international sites since sometimes strings for UI elements can be much
-longer in other languages.  Both of the buttons below have a "natural" width
-of 144px.  The one on the right is wider but it accommodates the full string
-instead of wrapping or truncating it.
+longer in other languages. Both of the buttons below have a "natural" width
+of 144px. The one on the right is wider but it accommodates the full string
+instead of wrapping it. Try to avoid setting a `maxWidth`, but if one is set,
+the text truncates.
 ```jsx
 const {View} = require("@khanacademy/wonder-blocks-core");
 const {StyleSheet} = require("aphrodite");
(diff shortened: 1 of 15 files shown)


jeresig:

Built with commit 1b4b1e3549578c1681c295733a1be0d3e926635d

https://deploy-preview-287--wonder-blocks.netlify.com

packages/wonder-blocks-button/components/button.md
```diff
@@ -363,13 +363,14 @@ const styles = StyleSheet.create({
 </View>
 ```
 
-Layouts often specify a specific width of button.  When implementing such
-designs use `minWidth` instead of `width`.  `minWidth` allows the button
-to resize to fit the content whereas `width` does not.  This is important
```
kevinbarabash: Old habits die hard.

packages/wonder-blocks-button/components/button.md
```diff
@@ -363,13 +363,14 @@ const styles = StyleSheet.create({
...
+designs use `minWidth` instead of `width`. `minWidth` allows the button
+to resize to fit the content whereas `width` does not. This is important
 for international sites since sometimes strings for UI elements can be much
-longer in other languages.  Both of the buttons below have a "natural" width
-of 144px.  The one on the right is wider but it accommodates the full string
-instead of wrapping or truncating it.
+longer in other languages. Both of the buttons below have a "natural" width
+of 144px. The one on the right is wider but it accommodates the full string
+instead of wrapping it. Try to avoid setting a `maxWidth`, but if one is set,
+the text truncates.
```
kevinbarabash: It's probably worth calling out that the text will truncate as well if the container the buttons are in doesn't have enough room to accommodate the natural width.  If you view this example on mobile, the text in the right button will be truncated even without `maxWidth`.  We should probably drop the part about `maxWidth`.

packages/wonder-blocks-form/components/choice-internal.js
```diff
@@ -57,8 +57,7 @@ type Props = {|
  * and RadioGroup. This design allows for more explicit prop typing. For
  * example, we can make onChange a required prop on Checkbox but not on Choice
  * (because for Choice, that prop would be auto-populated by CheckboxGroup).
- */
-export default class ChoiceInternal extends React.Component<Props> {
+ */ export default class ChoiceInternal extends React.Component<Props> {
```
kevinbarabash: :trollface:

kevinbarabash:
I love the polish in this PR.

PR #288, opened 2018-08-16
Make sure we take overflow visible into account when calculating intersections

somewhatabstract:
This addresses an issue we found when integrating the `Tooltip` component with webapp. The bounds checking used to detect visibility did not take into account that regardless of parent element bounds, the anchor may still be visible if the parent has `overflow` set to `visible` for the relevant axis.

This fixes that. We also update all the tests to make sure they have overflow styles to look at and we add a new test case for this new behaviour.

### DIFF
--- a/packages/wonder-blocks-core/util/get-element-intersection.js
+++ b/packages/wonder-blocks-core/util/get-element-intersection.js
@@ -66,19 +66,29 @@ function getElementIntersectionAgainstParent(
     intersectingRect: ClientRect | DOMRect,
     boundsElement: Element,
 ): Intersection {
+    // We need to check that it matters if we're out of bounds. If the parent
+    // element isn't clipping or otherwise hiding things outside its bounds,
+    // then checking against bounds isn't going to be much use.
+    // So, let's get the style for the element and use the overflow values.
+    const style =
+        ((boundsElement: any).currentStyle: ?CSSStyleDeclaration) ||
+        window.getComputedStyle(boundsElement);
+
     const boundsRect = boundsElement.getBoundingClientRect();
 
-    const horizontal = getAxisIntersection(
-        intersectingRect,
-        boundsRect,
-        "horizontal",
-    );
+    // We assume we're within this specific bounds element if it's overflow is
+    // visible.
+    const horizontal =
+        style.overflowX === "visible"
+            ? "within"
+            : getAxisIntersection(intersectingRect, boundsRect, "horizontal");
 
-    const vertical = getAxisIntersection(
-        intersectingRect,
-        boundsRect,
-        "vertical",
-    );
+    // We assume we're within this specific bounds element if it's overflow is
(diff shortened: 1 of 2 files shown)


jeresig:

Built with commit a4c6a58152ef41f3f237f1bbc5447daedb052e03

https://deploy-preview-288--wonder-blocks.netlify.com

packages/wonder-blocks-core/util/get-element-intersection.js
```diff
@@ -66,19 +66,29 @@ function getElementIntersectionAgainstParent(
     intersectingRect: ClientRect | DOMRect,
     boundsElement: Element,
 ): Intersection {
+    // We need to check that it matters if we're out of bounds. If the parent
+    // element isn't clipping or otherwise hiding things outside its bounds,
+    // then checking against bounds isn't going to be much use.
+    // So, let's get the style for the element and use the overflow values.
+    const style =
+        ((boundsElement: any).currentStyle: ?CSSStyleDeclaration) ||
+        window.getComputedStyle(boundsElement);
```
kevinbarabash: TIL about `currentStyle`.  https://developer.mozilla.org/en-US/docs/Web/API/Element/currentStyle says it isn't part of any spec... I guess it doesn't matter since we also have a fallback.

packages/wonder-blocks-core/util/get-element-intersection.test.js
```diff
@@ -79,7 +79,13 @@ describe("getElementIntersection", () => {
             // First, let's fake out some scroll parent.
             // Flow doesn't like jest mocks $FlowFixMe
             enumerateScrollAncestors.mockImplementation(() => [
-                {getBoundingClientRect: jest.fn(() => rect(0, 100, 100, 0))},
+                {
+                    getBoundingClientRect: jest.fn(() => rect(0, 100, 100, 0)),
+                    currentStyle: {
+                        overflowX: "auto",
+                        overflowY: "auto",
+                    },
```
kevinbarabash: I see why you used `currentStyle`. 🙂

packages/wonder-blocks-core/util/get-element-intersection.js
somewhatabstract: Yeah, from my understanding, it's quicker than `getComputedStyle` when it's available, so things often use it just in case.

packages/wonder-blocks-core/util/get-element-intersection.test.js
somewhatabstract: And yes, it also makes testing slightly easier (wouldn't be hard to have a mock for `getComputedStyle` though...and perhaps we should have a path that tests that.

kevinbarabash: Nah, it's cool... you just be mocking it to run the same values.

PR #290, opened 2018-08-20
WIP: ProviderGroup component wrt ADR-77

somewhatabstract:
Thought I'd have a go at a component that can amalgamate other "provider"-style children-as-a-function (CaaF) components so that instead of deeply nested renders we can have a more flat style.

I'm not sure the types are right or not, so putting up this WIP for some feedback before this gets written up as an ADR for inclusion in Wonder Blocks (currently added to Core, but perhaps this shouldn't be Core?).

## Usage
Instead of:

```
return (
    <Provider1 myProp="value">
        {x => (
            <Provider2 myProp="value">
                {y => (
                    <ThingThatWantsXAndY prop1={x} prop2={y} />
                )}
            </Provider2>
        )}
    </Provider1>
);
```

We can do:

```
const providers = [
    {type: Provider1, props: {myProp: "value"}},
    {type: Provider2, props: {myProp: "value"}},
];

return (
    <ProviderGroup providers={providers}>
        {([x, y]) => <ThingThatWantsXAndY prop1={x} prop2={y} />}
    </ProviderGroup>
);
```

The downside here is that `x` and `y` are no longer strongly typed by flow, but instead provided as `any`. I couldn't work out how we could do that differently with flow, but I think this loss is acceptable for the cleaner nesting, since callers should know what types their providers are returning and can just reapply those types as needed.

Introducing a component like this allows us to leverage the provider-style approach for more things (like the pending ADR-77 work for implementing managed `timeout` and `interval` replacements) without driving consuming code to have deeply nested "provider" trees.

### DIFF
--- a/packages/wonder-blocks-core/components/provider-group.js
+++ b/packages/wonder-blocks-core/components/provider-group.js
@@ -1,0 +1,37 @@
+// @flow
+import * as React from "react";
+
+type Provider<P> = {|
+    type: React.ComponentType<P>,
+    props?: P,
+|};
+
+type Props = {|
+    providers: Array<Provider<any>>,
+    children: (Array<any>) => React.Node,
+|};
+
+export default class ProviderGroup extends React.Component<Props> {
+    render() {
+        const {providers, children} = this.props;
+        const provided = [];
+        let i = 0;
+        const buildProviders = (...args) => {
+            if (i > 0) {
+                provided.push(...args);
+            }
+            if (i < providers.length) {
+                const providerEl = React.createElement(providers[i].type, {
+                    ...providers[i].props,
+                    children: buildProviders,
+                });
+                i++;
+                return providerEl;
+            } else {
+                return children(provided);
+            }
+        };
+
+        return buildProviders();
+    }
+}
--- a/packages/wonder-blocks-core/components/provider-group.test.js
+++ b/packages/wonder-blocks-core/components/provider-group.test.js
@@ -1,0 +1,36 @@
+// @flow
+import * as React from "react";
+import {mount, unmountAll} from "../../../utils/testing/mount.js";
+
+import UniqueIDProvider from "./unique-id-provider.js";
+import ProviderGroup from "./provider-group.js";
+
+describe("ProviderGroup", () => {
+    afterEach(() => {
+        unmountAll();
+    });
+
+    test("provides values to children from specified providers", () => {
+        // Arrange
+        const gatherer = jest.fn((...args) => {
+            // eslint-disable-next-line no-console
+            console.log(args[0]);
+            return "Gathered";
+        });
+        const providers = [
+            {type: UniqueIDProvider, props: {}},
+            {type: UniqueIDProvider, props: {scope: "test"}},
+        ];
+        const underTest = (
+            <ProviderGroup providers={providers}>
+                {(...args) => gatherer(args)}
+            </ProviderGroup>
+        );
+
+        // Act
+        mount(underTest);
+
+        // Assert
+        expect(gatherer).toHaveBeenCalledTimes(1);
+    });
+});
--- a/packages/wonder-blocks-core/generated-snapshot.test.js
+++ b/packages/wonder-blocks-core/generated-snapshot.test.js
@@ -11,6 +11,7 @@ jest.mock("react-dom");
 import ClickableBehavior from "./components/clickable-behavior.js";
 import MediaLayout from "./components/media-layout.js";
 import NoSSR from "./components/no-ssr.js";
+import ProviderGroup from "./components/provider-group.js";
 import Text from "./components/text.js";
 import UniqueIDProvider from "./components/unique-id-provider.js";
 import View from "./components/view.js";


jeresig:

Built with commit 9493391d31ad69042b95731cae4a391f00e84efc

https://deploy-preview-290--wonder-blocks.netlify.com

packages/wonder-blocks-core/components/provider-group.js
```diff
@@ -0,0 +1,37 @@
...
+    type: React.ComponentType<P>,
+    props?: P,
+|};
+
+type Props = {|
+    providers: Array<Provider<any>>,
+    children: (Array<any>) => React.Node,
+|};
+
+export default class ProviderGroup extends React.Component<Props> {
```
somewhatabstract: Rather than `ProviderGroup`, perhaps this should be called `CaaFConcatenator` or something. Thoughts?

packages/wonder-blocks-core/components/provider-group.js
```diff
@@ -0,0 +1,37 @@
...
+    render() {
+        const {providers, children} = this.props;
+        const provided = [];
+        let i = 0;
+        const buildProviders = (...args) => {
+            if (i > 0) {
+                provided.push(...args);
```
somewhatabstract: The first iteration kicks things off and isn't actually from one of the providers, so we don't want to include its args (an empty array).

packages/wonder-blocks-core/components/provider-group.js
```diff
@@ -0,0 +1,37 @@
...
+            }
+            if (i < providers.length) {
+                const providerEl = React.createElement(providers[i].type, {
+                    ...providers[i].props,
+                    children: buildProviders,
+                });
```
somewhatabstract: For each provider, we tell it that its children is our little function. This allows us to gather its CaaF arguments into our array and then tell that provider that its child is the next provider.

packages/wonder-blocks-core/components/provider-group.js
```diff
@@ -0,0 +1,37 @@
...
+                i++;
+                return providerEl;
+            } else {
+                return children(provided);
```
somewhatabstract: If we have made it through all the providers, we return the result of the real children given the now gathered provided values (which are in the same order as the providers array passed to the props, ensuring an intuitive ordering for consumers).

packages/wonder-blocks-core/components/provider-group.js
somewhatabstract: @kevinbarabash had suggested `ProviderComposer` in ADR-77. Might I suggest, `CaaFComposer`?

packages/wonder-blocks-core/components/provider-group.js
```diff
@@ -0,0 +1,37 @@
+// @flow
+import * as React from "react";
+
+type Provider<P> = {|
+    type: React.ComponentType<P>,
+    props?: P,
```
somewhatabstract: Question: Should we consider catering to providers that need the value of an earlier provider to be bound to one of their props?

If so, what syntax might that take? An additional optional value here like:

```
type Binding = {|
    propName: string,
    bindingPath: string,
|};

type Provider<P> = {|
...
    bindings?: Array<Binding>,
|}
```

Where binding path is of the form:

```
providerIndex.argumentIndex.fieldName1.fieldName2
```

Where `providerIndex` is the provider index in the `providers` prop, `argumentIndex` is the index of the argument from that provider's children function call, and each `fieldNameX` is a dereference level in that specific argument object.

So an examples might be:

```
0.0
```

or 

```
0.0.get("test")
```

Deconstructing the string, we could look in the `provided` array at index 0, then grab the argument at index 0 from that specific subarray, then call the function with the arguments and bind that to the named prop.

   `provided[0][0]["get"]("test")`

Of course, if we never envisage a need for this, we can completely leave it out, but it seems like it might be useful to build in.

packages/wonder-blocks-core/components/provider-group.js
```diff
@@ -0,0 +1,37 @@
...
+|};
+
+type Props = {|
+    providers: Array<Provider<any>>,
+    children: (Array<any>) => React.Node,
+|};
```
kevinbarabash: I think it's possible to make the types more precise using https://flow.org/en/docs/types/utilities/#toc-tuplemap.

packages/wonder-blocks-core/components/provider-group.js
kevinbarabash: `CaaFConcatenator` feels better.  `Provider` is a very overloaded term.

packages/wonder-blocks-core/components/provider-group.js
kevinbarabash: The `Binding` feels complicated, maybe a callback the signature of the `children` function that would've been passed to the CaaF if it was being called normally.

packages/wonder-blocks-core/components/provider-group.js
```diff
@@ -0,0 +1,37 @@
+// @flow
+import * as React from "react";
+
+type Provider<P> = {|
+    type: React.ComponentType<P>,
```
kevinbarabash: `component` seems more appropriate than `type`.

packages/wonder-blocks-core/components/provider-group.js
kevinbarabash: It would be good to include this and the previous two comments.

packages/wonder-blocks-core/components/provider-group.js
kevinbarabash: Spreading makes it impossible to support CaaF components that might pass variable numbers of args.  Not spreading makes it awkward to use.  Not sure the best direction to go in.

somewhatabstract: I think that can be easily addressed

packages/wonder-blocks-core/components/provider-group.js
somewhatabstract: Yeah, I know. Was just faster to do it here for now :)

somewhatabstract:
@kevinbarabash I have a compromise solution for the cases where a CaaF component passes multiple args.

packages/wonder-blocks-core/components/provider-group.js
somewhatabstract: I'm not sure how your suggestion would work, @kevinbarabash. This is for if a given CaaF component has a prop that needs to be bound to the value of another CaaF in the list. Not sure how matching the signature of the `children` function achieves that.

packages/wonder-blocks-core/components/caaf-aggregator.js
```diff
@@ -0,0 +1,78 @@
+// @flow
+import * as React from "react";
+
+type CaaFComponent<P> = {|
+    type: React.ComponentType<P>,
+    props?: P,
+|};
```
kevinbarabash: This is worrisome b/c it means that props won't be specific to each component.

somewhatabstract: I don't follow.

PR #291, opened 2018-08-20
**Breaking change for Icon and IconButton** -- Add ariaLabel prop to checkbox / radio

wishdasher:
I noticed that some checkboxes in `webapp` try to use `ariaLabel` to provide a label for checkboxes that don't have a clear label e.g.
image
However, the current `Checkbox` in `webapp` doesn't actually support an `ariaLabel` prop (which is how I missed this during the implementation spec phase). So all those checkboxes in `webapp` don't really have aria labels attached...oops.

### DIFF
--- a/packages/wonder-blocks-form/components/radio.js
+++ b/packages/wonder-blocks-form/components/radio.js
@@ -28,6 +28,13 @@ type ChoiceComponentProps = {|
     onChange: (newCheckedState: boolean) => void,
 
     /**
+     * Optional label if it is not obvious from the context what the radio
+     * does. If the label and id props are defined, this props does not need to
+     * be provided as the label would be matched to this input.
+     */
+    ariaLabel?: string,
+
+    /**
      * Optional label for the field.
      */
     label?: string,
--- a/packages/wonder-blocks-form/util/types.js
+++ b/packages/wonder-blocks-form/util/types.js
@@ -12,6 +12,10 @@ export type ChoiceCoreProps = {|
     disabled: boolean,
     /** Whether this component should show an error state */
     error: boolean,
+    /** Optional label if it is not obvious from the context what the checkbox
+     * does. If the label and id props are defined, this props does not need to
+     * be provided as the label would be matched to this input. */
+    ariaLabel?: string,
     /** Name for the checkbox or radio button group */
     groupName?: string,
     /** Unique identifier attached to the HTML input element. If used, need to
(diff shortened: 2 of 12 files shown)


packages/wonder-blocks-form/components/radio.js
```diff
@@ -27,6 +27,13 @@ type ChoiceComponentProps = {|
      */
     onChange: (newCheckedState: boolean) => void,
 
+    /**
+     * Optional label if it is not obvious from the context what the radio
+     * does. If the label and id props are defined, this props does not need to
+     * be provided as the label would be matched to this input.
+     */
+    ariaLabel?: string,
```
kevinbarabash: Let's make this `aria-label`.

packages/wonder-blocks-form/util/types.js
```diff
@@ -12,6 +12,10 @@ export type ChoiceCoreProps = {|
     disabled: boolean,
     /** Whether this component should show an error state */
     error: boolean,
+    /** Optional label if it is not obvious from the context what the checkbox
+     * does. If the label and id props are defined, this props does not need to
+     * be provided as the label would be matched to this input. */
+    ariaLabel?: string,
```
kevinbarabash: same

jeresig:

Built with commit 14a40754b0f3357fb064c9bd1cbf924843dc77a8

https://deploy-preview-291--wonder-blocks.netlify.com

packages/wonder-blocks-form/components/radio.js
wishdasher: I just noticed that IconButton uses `aria-label`. Some components in `webapp`, though, use `ariaLabel`. Maybe we should add something to the style guide 🤔

wishdasher: I think the disadvantage of `aria-label` is that we can't unpack it in the regular props unpacking part.

wishdasher: Is bumping up the version a matter of manually editing `version` 2.0.1 => 3 in `package.json`?

kevinbarabash: We'll bump when we do a release.  Can you edit the title of PR to include `BREAKING CHANGE`.  It would be cool if we could use https://github.com/semantic-release/semantic-release, but currently it doesn't work with `lerna` so we'll have to track this manually for now.  cc @jeresig

kevinbarabash: Okay... so there's actually a way to destructure objects with `aria-label` in it... by renaming it in the destructuring assignment, e.g.
```
const {
   'aria-label': ariaLabel,
   ...otherProps,
} = props;
```

PR #297, opened 2018-08-21
Don't automatically focus on items changed for dropdown

wishdasher:
I noticed something annoying -- currently, if you go to http://deploy-preview-268--wonder-blocks.netlify.com/#multiselect and click "Select all" with the mouse, the updating automatically focuses the newly updated focused item.

I changed that behavior to align with the rest of the dropdown navigation behavior -- keyboard navigation focus only shows once you start navigating with the keyboard. So if you update "Select all" with the keyboard, "Select none" does become focused.

Also cleaned up `keyboardNavigationOn` logic, as it was trying to do two separate things (keep track of whether keyboard nav was on, and also whether the user started to make selections).

Added a test case too.

Also fixed a bug where if the dropdown opener was focused in a modal, escape wouldn't dismiss the modal.


jeresig:

Built with commit dfe800fc86ae1a93013f2708c8d8b7bf1fb654e9

https://deploy-preview-297--wonder-blocks.netlify.com

PR #298, opened 2018-08-22
Replace ClickableBehavior's trigger props with role prop

wishdasher:
`role` prop is optional, and it defaults to both space and enter being triggers (this was the default behavior before as well).

### DIFF
--- a/packages/wonder-blocks-core/components/clickable-behavior.js
+++ b/packages/wonder-blocks-core/components/clickable-behavior.js
@@ -1,6 +1,44 @@
 // @flow
 import React from "react";
 
+// NOTE: Potentially add to this as more cases come up.
+type ClickableRole =
+    | "button"
+    | "link"
+    | "checkbox"
+    | "radio"
+    | "listbox"
+    | "option"
+    | "menuitem";
+
+const getAppropriateTriggersForRole = (role: ?ClickableRole) => {
+    switch (role) {
+        // Triggers on ENTER, but not SPACE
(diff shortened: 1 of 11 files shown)


jeresig:

Built with commit 123895fc0cfc39bef1efe198d8f276dd81e2ab46

https://deploy-preview-298--wonder-blocks.netlify.com

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -224,6 +257,9 @@ export default class ClickableBehavior extends React.Component<Props, State> {
         super(props);
 
         this.state = startState;
+        this.waitingForClick = false;
+        this.enterClick = false;
+        this.triggers = getAppropriateTriggersForRole(this.props.role);
```
kevinbarabash: I can't see people changing the `role` of a component ever, but we should handle updating all props properly.  In this case we should probably just call `getAppropriateTriggersForRole` whenever we need `triggers` (it's not very expensive) instead of storing it in an instance variable.

packages/wonder-blocks-form/components/choice-internal.js
```diff
@@ -131,7 +130,7 @@ type Props = {|
                 <ClickableBehavior
                     disabled={coreProps.disabled}
                     onClick={this.handleClick}
-                    triggerOnEnter={false}
+                    role={variant}
```
kevinbarabash: neat

PR #303, opened 2018-08-27
Fix tooltip positioning

kevinbarabash:
When we upgraded our type checking for the style prop we missed copying the `transform` style prop which actually positions the tooltip on the page.


jeresig:

Built with commit 6b1565795277f8893021ef1dbbc71c3ed5f5af88

https://deploy-preview-303--wonder-blocks.netlify.com

PR #304, opened 2018-08-27
fix element intersection func to work with overflow-y: scroll hack

kevinbarabash:
The fix wasn't too bad, set the bounding rect's height to be the scrollHeight and add update its bottom to be consistent.  Unfortunately, this is hard to write a unit test for.  I'd rather spend time working on setting up image regression tests for this as those will be useful for other things that are hard (or impossible) to test as well.

I tested this by setting the `height: 100%` on `body, html` and `overflow-y: scroll` on `body`.  Then I navigated to http://localhost:6060/#tooltip-1 and hovered over items and saw their tooltips.

screen shot 2018-08-27 at 5 53 10 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "test:coverage": "yarn run test:common && yarn run jest --coverage",
     "coverage": "yarn run test:coverage && open coverage/lcov-report/index.html",
     "styleguidist": "styleguidist server",
-    "start": "concurrently --kill-others-on-fail \"yarn run styleguidist\" \"webpack -w\"",
+    "start": "concurrently --kill-others-on-fail \"yarn run styleguidist\" \"webpack -w\" \"rollup -c rollup.config.js -w\"",
     "start:prod": "yarn run build:styleguidist && yarn run styleguidist --config styleguide.prod.config.js",
     "build:styleguidist": "yarn run gen-styleguide-prod-config && styleguidist build --config styleguide.prod.config.js",
     "git-add-snapshot-tests": "git add packages/*/generated-snapshot.test.js",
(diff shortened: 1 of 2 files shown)


jeresig:

Built with commit 53fcbe38242f74ce47faa585fad01458ad2314b0

https://deploy-preview-304--wonder-blocks.netlify.com

packages/wonder-blocks-core/util/get-element-intersection.js
```diff
@@ -39,9 +39,18 @@ function FullIntersection() {
...
+    left: number,
+    right: number,
+    width: number,
+    height: number,
+|};
+
 function getAxisIntersection(
     intersectingRect: ClientRect | DOMRect,
-    boundsRect: ClientRect | DOMRect,
+    boundsRect: Rect,
```
kevinbarabash: You can't mutate a `ClientRect` or a `DOMRect` so this is the result.

package.json
```diff
@@ -22,7 +22,7 @@
     "test:coverage": "yarn run test:common && yarn run jest --coverage",
     "coverage": "yarn run test:coverage && open coverage/lcov-report/index.html",
     "styleguidist": "styleguidist server",
-    "start": "concurrently --kill-others-on-fail \"yarn run styleguidist\" \"webpack -w\"",
+    "start": "concurrently --kill-others-on-fail \"yarn run styleguidist\" \"webpack -w\" \"rollup -c rollup.config.js -w\"",
```
kevinbarabash: `styleguidist` is using the `dist/es/index.js` files.  I decided to run both webpack and rollup in watch mode just in case.

kevinbarabash:
> I thought we had an issue filed in Jira for this but I can't find it now. Thanks for digging in to this!

I had a look but couldn't find anything in Jira for this issue.  There was #288 which was related but different.

PR #306, opened 2018-08-28
Use Babel 7 to transform our code.

jeresig:
Adds support for Babel 7. Compiling with Webpack and Rollup works.

Jest is failing right now. It appears to be transforming the async stuff into generators which is failing because we don't include the runtime for it. When I tried including the runtime I got different failures when building webpack and rollup. All the browsers we support have support for async so we don't want the generators included. Thoughts?


jeresig:

Built with commit e69f6f6106de7df3c1fd3aec04d9bb0e86bca420

https://deploy-preview-306--wonder-blocks.netlify.com

jeresig:
Oh strange - I'm seeing a totally different error from Jest here than locally! Not sure what's happening or why babel-core is getting called instead of @babel/core

kevinbarabash:
> All the browsers we support have support for async so we don't want the generators included. Thoughts?

We should turn off the transform for async/await when running the tests.

jeresig:
@kevinbarabash right, that's what I thought I did by setting the `.browserlistrc`. It only supports platforms that support async. So I'm not sure why it's not picking that up!

kevinbarabash:
It only seems to be failing on dynamic `import()`s now.

package.json
```diff
@@ -36,16 +36,19 @@
...
-    "babel-core": "6.26.3",
+    "@babel/core": "7.0.0",
+    "@babel/plugin-proposal-class-properties": "7.0.0",
+    "@babel/plugin-proposal-export-default-from": "^7.0.0",
+    "@babel/plugin-proposal-object-rest-spread": "7.0.0",
+    "@babel/plugin-transform-flow-strip-types": "7.0.0",
+    "@babel/preset-env": "7.0.0",
+    "@babel/preset-flow": "7.0.0",
+    "@babel/preset-react": "7.0.0",
+    "babel-core": "^7.0.0-bridge.0",
```
kevinbarabash: I added this "bridge" version.

packages/wonder-blocks-form/components/checkbox-group.test.js
```diff
@@ -66,13 +66,13 @@ describe("CheckboxGroup", () => {
     it("calls onChange for each new selection", () => {
         // a is clicked
         const a = group.find(Choice).at(0);
-        const aTarget = a.find("ClickableBehavior");
+        const aTarget = a.find("clickable_behavior_ClickableBehavior");
```
jeresig: I don't understand these changes -- they're also in the snapshot below for some of the typography components!? I tried searching all around and I can't figure out where this change may have come from.

jeresig:
@somewhatabstract If you have the time or inclination, I'm hitting an odd issue with the Babel 7 upgrade. Specifically the `jest.mock("@khanacademy/wonder-blocks-core");` line inside `visibility-modifier.test.js` is now failing! For some reason Jest no longer likes generating a mock for the core module - and I don't know why! (they don't really provide debugging details) I tried to provide a mock stub and it didn't seem to work... so if you have any ideas, you're thoughts would be greatly appreciated!

jeresig:
@kevinbarabash I got things almost entirely passing, save for the one Jest mock failure. I don't know if the serialized React component names are bugs, or not. But I got the tests passing, at least...

somewhatabstract:
I'll take a look, @jeresig. The jest mocks are amazing since they mock before the code under test imports its imports somehow, probably why stub didn't work right. Do you have a specific goal date for getting this complete (will help me prioritize)?

MiguelCastillo:
babel 7!! 🥇

jeresig:
@somewhatabstract No timeline! Would just be a nice-to-have so whenever you get a moment. Thanks!

somewhatabstract:
If I edit [this line in `jest-mock`](https://github.com/facebook/jest/blob/master/packages/jest-mock/src/index.js#L189) to be:

```
} else if (isA('Object', ref) || isA('Module', ref)) {
```

Then everything passes. I don't know how to describe this issue to get the `jest` team to accept that as a PR though. Perhaps you can help word the issue, @jeresig?

In the meantime, is there a way to get the tests to run with the non-ES modules?

somewhatabstract:
@jeresig I have pushed a change to fix the test. I'd prefer it if we get jest to support ES modules in the automocks (it's an easy change if we can sell it to them - [see my previous comment)](https://github.com/Khan/wonder-blocks/pull/306#issuecomment-424188843).

jeresig:
@somewhatabstract That's fantastic, thank you so much for finding a fix for this! I agree that this is something that Jest should just fix, I'll see if I can find a way to surface this to them!

jeresig:
@kevinbarabash Ok, all the tests are passing - do you think the modified class names are blocking? Do you have any thoughts as to why this might be happening?

kevinbarabash:
@jeresig I'm not too concerned about the class names.  I am a little concerned about not compiling the classes down to ES5.  Are we thinking about shipping code that has `class`es in it?

jeresig:
@kevinbarabash Yes, I've set the browser config to only support the browsers that we're actually supporting - and all of them support classes natively! So we just get to use it, which is exciting.

PR #1, opened 2018-03-08
Cleanup repo and enable travis-ci

kevinbarabash:
The cleanup involved:
- removing duplicate files
- preferring .test.js and .stories.js to the _ versions
- fixing regex in .storybook/config.js so that we were only loading stories in packages/wonder-blocks-*
- add explicit "dependencies" in each sub package's package.json to point to relative directories
- add script to run "npm install" in each sub package

Test Plan: npm test

Reviewers: miguel idr john

PR #307, opened 2018-08-28
Add i18n to core and markup i18n cases.

jeresig:


jeresig:

Built with commit c911ef7922786fe7629ebd46b38b1e59c38a24b9

https://deploy-preview-307--wonder-blocks.netlify.com

packages/wonder-blocks-core/util/i18n.js
```diff
@@ -0,0 +1,266 @@
...
+    // In such cases, only the plural form ends up in the .po file, and
+    // then it gets sent to us for the i18n._() case too.  No problem, though:
+    // we'll just take the singular arg.
+    if (typeof str === "object" && str.messages) {
+        str = str.messages[0];
+    }
+
+    return str.replace(interpolationMarker, (match, key) => {
+        if (options && options.hasOwnProperty(key)) {
+            const replaceWith = options[key];
```
kevinbarabash: If `replaceWith` is a `number` it would be nice to call `toLocaleString` on it, it's really only need when called from `ngettext`.  We'd also have to pass the `lang` to `_()` to be able to use it.

packages/wonder-blocks-core/util/i18n.js
```diff
@@ -0,0 +1,266 @@
...
+
+type Options = {|
+    [key: string]: string | number,
+|};
+
+const ngettext = (
+    singular: string | Plurals,
+    plural: number | string,
+    num: number | Options,
+    options?: Options,
```
kevinbarabash: Thanks for figuring out the typing here.

jeresig:
@kevinbarabash I can move this to a new package and write tests. One question/concern that I have is that if we want to do grepping for `i18n._` then we'll need to export it in a way that leaves the `i18n` intact, otherwise we'll have to do: `const i18n = require("@khanacademy/wonder-blocks-i18n");` to appease Kake. We could export it as: `import {i18n} from "@khanacademy/wonder-blocks-i18n";` but that seems strange. The benefit of having it in Core meant that we could export it as `{i18n}` and it wouldn't feel awkward (necessarily).

jeresig:
@kevinbarabash Hmm - I think you may have mis-interpreted me. There are two separate issues.

1) We want Kake to extract our i18n strings from Wonder Blocks. In order to do so we want our strings to be marked up with `i18n._(...)`. In order to ensure that the string is intact we can't use `import i18n ...` because it'll get rewritten by Babel, which is the same exact issue that we're hitting right now in Webapp. So we need i18n to be exported as a fully-intact object so that it doesn't get rewritten.

2) I do think we want to use the i18n stuff inside webapp, and I do think it should be part of Wonder Blocks package. Having Wonder Blocks be a dependency of everything is what we should be doing anyway (and it won't matter when we switch to Webpack). Either way I don't think this is something that we need to worry about right now, we can do this later if we want (and should probably have an ADR for it).

I can definitely put in inside a wonder-blocks-i18n package but I don't know how to resolve #1. I'm less excited about having it in another i18n-js package as I'll have to setup all of the build/deploy/lint/test stuff for that package as well.

kevinbarabash:
> We want Kake to extract our i18n strings from Wonder Blocks. In order to do so we want our strings to be marked up with i18n._(...). In order to ensure that the string is intact we can't use import i18n ... because it'll get rewritten by Babel, which is the same exact issue that we're hitting right now in Webapp. So we need i18n to be exported as a fully-intact object so that it doesn't get rewritten.

I see, we should probably be running our regexes on the source but we can't b/c of `<$_>`, burgh.

> I do think we want to use the i18n stuff inside webapp, and I do think it should be part of Wonder Blocks package. Having Wonder Blocks be a dependency of everything is what we should be doing anyway (and it won't matter when we switch to Webpack). Either way I don't think this is something that we need to worry about right now, we can do this later if we want (and should probably have an ADR for it).

Okay.  I didn't realize we wanted that.  Sounds good.

> I can definitely put in inside a wonder-blocks-i18n package but I don't know how to resolve #1. I'm less excited about having it in another i18n-js package as I'll have to setup all of the build/deploy/lint/test stuff for that package as well.

I was working under a faulty set of assumptions.  Let's leave it where it is.

This is still blocked on tests.  My suggestion for incorporating `toLocaleString()` can be deferred to a subsequent PR.

kevinbarabash:
@jeresig I'm going to write up an ADR for this.  Given that we haven't heard about any translation issues yet it doesn't seem high enough priority warrant add code to `wonder-blocks-core` just to remove it later.  Also, now that we can use `yarn link` in webapp, it makes it really easy to test future changes to i18n.

PR #309, opened 2018-08-29
Don't render empty elements inside toolbar components

pamelafox:
I have a usecase for CS scratchpads where the toolbar editor variably shows different elements, ie. many of them are null at any given time. With the WB toolbar, it currently shows an empty View for each of those null elements, which is undesirable. This change filters out empty elements using Array.filter() before mapping them.

For testing, I manually edited the generated file in webapp/third_party/node_modules and verified it works for my use case. I did not see other obvious ways to test it, since the jest tests are auto-generated. I'm assuming those tests are run by a CI.

### DIFF
--- a/packages/wonder-blocks-toolbar/components/toolbar.js
+++ b/packages/wonder-blocks-toolbar/components/toolbar.js
@@ -60,9 +60,9 @@ export default class Toolbar extends React.Component<Props> {
     };
 
     renderContent(content: React.Node) {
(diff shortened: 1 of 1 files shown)


jeresig:

Built with commit 7a08582f1fec76c6d1814b4a9642ce5e99317ee7

https://deploy-preview-309--wonder-blocks.netlify.com

kevinbarabash:
Thanks!

PR #310, opened 2018-08-29
Left align multiple items in left column, add more examples from spec

pamelafox:
I noticed while using the toolbar component that my left items were stacking on top of each other instead of next to eachother. That was due to a difference in styles, so I matched the left column styles to the right column styles. This seems to match the spec examples, and it seems to work with the close icon examples from the docs. I also added two more examples to the doc, based on the ones in Zeplin.

I believe that there are no current users of toolbar (atleast in webapp master) so this seems fairly safe. 

I did need to re-generate the snapshots, since the styles changed.


jeresig:

Built with commit 28ef6de111ac0f1a4b15753c9c270ff487f2f9f0

https://deploy-preview-310--wonder-blocks.netlify.com

kevinbarabash:
@pamelafox super minor nit: these IconButtons are further apart than they should be.

netlify:
screen shot 2018-08-29 at 6 06 13 pm

zeplin:
screen shot 2018-08-29 at 6 07 22 pm

pamelafox:
I saw that, yes. All the toolbar components have margins on the side. Do
you think that there should be special casing for toolbar icons inside
components?

kevinbarabash:
We shouldn't be adding margins automatically b/c in some case we don't need them.  It's always easier to padding this afterwards.  I've added an issue in zeplin: https://khanacademy.atlassian.net/browse/WB-426.

pamelafox:
Okay, sounds like that's a general design principle for WB. Thanks for
pointing that out!

PR #311, opened 2018-08-29
Layout

jeresig:
Implement the new `<Layout>` component. This is added to the Layout package along with MediaLayout (which has been moved from core).

Grid and Modals have been rewritten to use Layout instead of other techniques.

I need to finish updating the MediaLayout v1 spec to represent what I've done here: https://docs.google.com/document/d/1hdh0_tEn0RY7jkC71Sln2BS3xJ9alwFzW6qYRiKu3kU/edit

In short: Layout is responsible for all of the viewport size tracking now. MediaLayout is only used for changing default settings.

Layout can do a couple things:
- It can toggle its visibility based on a series of booleans (small/medium/large and mdOrSmaller/mdOrLarger).
- It can take a function as a child and pass through the mediaSize, mediaSpec, and styles to the function for rendering.
- It can consume some Aphrodite stylesheets and pass them through to the function child. It filters the stylesheets automatically based upon the current mediaSize making it really easy to use.

In some future PRs I want to make some more changes:

- I want to remove FixedWidthCell, FlexCell, and Gutter from Grid. All of these are not that useful in a world in which we have Layout.
- I want to rewrite a bunch of the Typography components to use Layout as they currently use media queries.


jeresig:

Built with commit 51dac6459a62f3ecf9bb646c7bcad5d236174c44

https://deploy-preview-311--wonder-blocks.netlify.com

jeresig:

Built with commit ad9013dd480376301f030db14f8f941fbece8eb7

https://deploy-preview-311--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-dialog.js
```diff
@@ -19,39 +19,42 @@ export default class ModalDialog extends React.Component<Props> {
         const {style, children} = this.props;
         return (
             <MediaLayout spec={MEDIA_MODAL_SPEC}>
```
kevinbarabash: We may actually want to move this up closer to the portal since eventually we'll want to render left/right chevrons on the modal backdrop itself and those may want access to the current media spec.

packages/wonder-blocks-modal/components/one-column-modal.js
```diff
@@ -61,57 +87,29 @@ class ContentWrapper extends React.Component<WrappedProps> {
...
-            >
-                <WrappedContentWrapper {...this.props} />
-            </ModalDialog>
+            <React.Fragment>
+                <Layout mdOrLarger={false}>
+                    <SmallOneColumnModal {...this.props} />
+                </Layout>
+                <Layout small={false}>
+                    <LargeOneColumnModal {...this.props} />
+                </Layout>
```
kevinbarabash: Using the callback variant of `Layout` would work quite well here:
```
<Layout>
   {({mediaSize}) => 
        mediaSize === "small"
            ? <SmallOneColumnModal {...this.props} />
            : <LargeOneColumnModal {...this.props} />
    }
</Layout>
```

packages/wonder-blocks-modal/components/standard-modal.js
```diff
@@ -79,64 +80,65 @@ export default class StandardModal extends React.Component<Props> {
...
-    smallPreview: {
-        display: "none",
-    },
-});
+    mdOrLarger: StyleSheet.create({
+        wrapper: {
+            width: "93.75%",
+            maxWidth: 960,
+            height: "81.25%",
+            maxHeight: 624,
```
kevinbarabash: Shouldn't the these styles be in the `all` section?

packages/wonder-blocks-modal/components/two-column-modal.js
```diff
@@ -89,74 +140,30 @@ class ContentWrapper extends React.Component<WrappedProps> {
...
-                //style={(mediaSize) =>
-                //    mediaSize === "small" ? styles.smallDialog : styles.dialog
-                //}
-                style={styles.dialog}
-            >
-                <WrappedContentWrapper {...this.props} />
-            </ModalDialog>
+            <React.Fragment>
+                <Layout mdOrLarger={false}>
+                    <SmallTwoColumnModal {...this.props} />
```
kevinbarabash: Could use the children-as-function style of `Layout` here as well.

packages/wonder-blocks-typography/util/types.js
```diff
@@ -6,10 +6,10 @@ import type {
     StyleType,
 } from "@khanacademy/wonder-blocks-core";
 
-// TODO(kevinb): fix style type after upgrading flow
-export type Props = AriaProps & {
+// eslint-disable-next-line flowtype/require-exact-type
```
kevinbarabash: It's unfortunate there's no way to spell checking the props in AriaProps while allowing other props like event handlers through.  I think eventually we'll want to create a list of all possible event handlers so that we can use exact object types on `View` and `Text` to make them safer.

packages/wonder-blocks-core/util/types.js
```diff
@@ -93,8 +91,7 @@ type IdRefList = IdRef | Array<IdRef>;
 
 // Source: https://www.w3.org/WAI/PF/aria-1.1/states_and_properties
 // TODO(kevinb): convert to disjoint union of exact object types keyed on role
-// eslint-disable-next-line flowtype/require-exact-type
-export type AriaProps = {
+export type AriaProps = {|
```
kevinbarabash: I didn't think `flow` would allow mixing exact objects with inexact types using the intersection operator.

kevinbarabash:
I've merge in master.  I'm going to do some testing to make sure everything's work as expected and then will do a pass over the previous comments.

packages/wonder-blocks-layout/components/layout.js
```diff
@@ -0,0 +1,277 @@
...
+        all?: StyleDeclaration,
+        mdOrLarger?: StyleDeclaration,
+        mdOrSmaller?: StyleDeclaration,
+        [mediaSize: MediaSize]: StyleDeclaration,
+    },
+
+    /** Should the contents be displayed at a media size of "small"? */
+    small: boolean,
+    /** Should the contents be displayed at a media size of "medium"? */
+    medium: boolean,
```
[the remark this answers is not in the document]
kevinbarabash: I'll look at addressing this in a following PR.

packages/wonder-blocks-layout/components/layout.js
```diff
@@ -0,0 +1,277 @@
...
+    };
+
+    state = {
+        size: undefined,
+    };
+
+    componentWillUnmount() {
+        // If we're on the server then we don't have anything to unmount
+        // as we never mounted anything in the first place!
+        if (this.isServerSide()) {
```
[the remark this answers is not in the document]
kevinbarabash: After doing some Google it appears that this method doesn't get called during SSR some I'm going to remove this code.

packages/wonder-blocks-layout/components/layout.js
```diff
@@ -0,0 +1,277 @@
...
+            const {query} = spec[size];
+
+            // Don't watch sizes that don't have an associated query
+            if (!query) {
+                continue;
+            }
+
+            // Create a new matchMedia watcher if one doesn't exist yet
+            if (!Layout.WATCHERS[query]) {
+                Layout.WATCHERS[query] = window.matchMedia(query);
```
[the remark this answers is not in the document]
kevinbarabash: I see, doing it in `componentDidMount` is too late because we need to know the current media size on first render.  We could move this to `componentWillMount` and add a `isServerSide()` check there.  I'll do that in a separate PR.

packages/wonder-blocks-layout/components/layout.js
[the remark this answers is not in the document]
kevinbarabash: I noticed in their use `OneColumnModal` that we end of passing `false` to the ones we want to disable which feels a little awkward as well.  The followup PR will address this.

packages/wonder-blocks-grid/components/flex-cell.js
```diff
@@ -54,35 +45,43 @@ type Props = {|
...
+export default class FlexCell extends React.Component<Props> {
     static defaultProps = {
-        small: false,
-        medium: false,
-        large: false,
+        small: true,
+        medium: true,
+        large: true,
+        mdOrLarger: true,
+        mdOrSmaller: true,
```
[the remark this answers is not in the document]
kevinbarabash: It's also a little weird how `small={true}` and `mdOrSmaller={false}` can both be set.

packages/wonder-blocks-grid/docs.md
```diff
@@ -111,140 +113,136 @@ See the [MediaLayout](#medialayout) component for more details. The layout break
...
-				height: 360,
-				padding: 24,
-				borderTop: `1px solid ${Color.offBlack8}`,
-				borderBottom: `1px solid ${Color.offBlack8}`,
-			}}
-		>
-			<Cell>Angles</Cell>
-		</Row>
-	</MediaLayout>
+    <Row
```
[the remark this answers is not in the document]
kevinbarabash: I see, this example doesn't actually use any of the data from the `MediaSpec`.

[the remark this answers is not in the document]
kevinbarabash: Scratch that... Layout uses MediaLayoutContext.Consumer which if isn't wrapped by a MediaLayout ends up using the default value for the MediaLayoutContext.

packages/wonder-blocks-form/components/choice.js
```diff
@@ -72,8 +72,7 @@ type Props = {|
  *
  * If you wish to use just a single field, use Checkbox or Radio with the
  * optional label and description props.
- */
-export default class Choice extends React.Component<Props> {
+ */ export default class Choice extends React.Component<Props> {
```
somewhatabstract: Missing a newline?

packages/wonder-blocks-form/components/radio.js
```diff
@@ -76,8 +76,7 @@ type ChoiceComponentProps = {|
  *
  * This component should not really be used by itself because radio buttons are
  * often grouped together. See RadioGroup.
- */
-export default class Radio extends React.Component<ChoiceComponentProps> {
+ */ export default class Radio extends React.Component<ChoiceComponentProps> {
```
somewhatabstract: Hmm, also missing a newline

packages/wonder-blocks-grid/components/flex-cell.js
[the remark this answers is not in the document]
somewhatabstract: I wonder if we should drive this a little differently. It does seem like we could get really confusing with these options. Could we not use string constants instead or something else that doesn't allow users to create contradictory prop combinations?

packages/wonder-blocks-grid/components/flex-cell.js
```diff
@@ -54,35 +45,43 @@ type Props = {|
...
-        }
+        return (
+            <Layout>
+                {({mediaSize, mediaSpec}) => {
+                    if (!FlexCell.shouldDisplay(this.props, mediaSize)) {
+                        return null;
+                    }
 
-        let contents = children;
+                    let contents = children;
```
somewhatabstract: `let` feels like a smell to me in many cases.

For example, this can be `const` if the subsequent code is a function, like this.


```
const contents = _getContents(mediaSize, mediaSpec);
```

```
_getContents = (mediaSize, mediaSpec) => {
    const {children} = this.props;
    if (typeof children === "function") {
        const {totalColumns} = mediaSpec[mediaSize];
        return children({mediaSize, totalColumns});
    }

    return children;
};
```

This would make things cleaner, I think. What do you think?

packages/wonder-blocks-grid/components/gutter.js
```diff
@@ -20,41 +15,40 @@ import FixedWidthCell from "./fixed-width-cell.js";
...
-     * The current media layout spec being used. Populated by MediaLayoutWrapper.
-     * @ignore
-     */
-    mediaSpec: MediaSpec,
+    large: boolean,
+    /** Should this gutter be shown at Medium or larger grids? */
+    mdOrLarger: boolean,
+    /** Should this gutter be shown at Medium or smaller grids? */
+    mdOrSmaller: boolean,
 }> {
```
somewhatabstract: Should be a prop type for this.

packages/wonder-blocks-grid/components/gutter.js
```diff
@@ -20,41 +15,40 @@ import FixedWidthCell from "./fixed-width-cell.js";
...
     static defaultProps = {
-        small: false,
-        medium: false,
-        large: false,
+        small: true,
+        medium: true,
+        large: true,
+        mdOrLarger: true,
+        mdOrSmaller: true,
```
somewhatabstract: Again, feels overly complex and easy to get into a contradictory state, but perhaps I am misreading this.

packages/wonder-blocks-grid/components/row.js
```diff
@@ -54,85 +55,88 @@ class Row extends React.Component<{
...
-     * The size of the media layout being used. Populated by MediaLayoutWrapper.
-     * @ignore
-     */
-    mediaSize: MediaSize,
-    /**
-     * The current media layout spec being used. Populated by MediaLayoutWrapper.
-     * @ignore
-     */
-    mediaSpec: MediaSpec,
 }> {
```
somewhatabstract: Props type?

kevinbarabash: Good call.

kevinbarabash:
> This needs tests. It's complex and we already know MediaLayout has had a bug or two from that issue we've seen before when I first started working on the Tooltip. I think adding the tests would help tighten up the API too as it would make it much easier to see what the edge cases were.

Agreed.  I'm going to keep iterating on it in this PR.  If at the end we think it's too unwieldy I can break it up into separate PRs.

packages/wonder-blocks-form/components/choice.js
kevinbarabash: This is a prettier bug. :(

packages/wonder-blocks-form/components/radio.js
kevinbarabash: same

kevinbarabash:
I've addressed the minor nits.  I'm going to work on API revisions next.  After that I'll look at improving test coverage.

kevinbarabash:
Closing this in favour of #372, #372, and additional followup PRs.

PR #312, opened 2018-08-30
Remove default margin from view items

pamelafox:
Addresses WB-426.

This removes the default margin from the items inside the toolbar. I put an 8px padding on the entire toolbar container, however. That means that you can add 8px margin to your toolbar items, and the buttons will end up 16px from the edge and 16px from each other. That seems to be what the Zeplin screenshot showed and also what Elizabeth requested in her design review of my <Toolbar> use. 

That 8px padding is also compatible with the use case of icon buttons on the side, like the close button. In my examples with those, I added no margin, and they seem to match the Zeplin screenshot.

What do you think? I have a hard time judging from the Zeplin what is intended to be built in, and what's intended to be provided. I believe this setup would work great for my use case, but it's hard to know how it will work for all cases.

FYI @scotchfield, original toolbar author.


jeresig:

Built with commit cb7cc2eb692fbcc24adb9bddc8318bfc6be06a8e

https://deploy-preview-312--wonder-blocks.netlify.com

pamelafox:
Hm, adding 16px on the left will mean that the left-hand icon button show up farther over than they do in Zeplin. In Zeplin, the icons start 16px over, but once you wrap those icons in an IconButton, the icon itself starts 32px over. To match Zeplin, you'd have to add -16px margin on the left-hand iconbuttons.

I think perhaps the screenshots did not take into account the idea of icons being wrapped in IconButtons, and thus having a hover state with a border and padding. So I'm on board with adding 16px padding, but I want to point out there'll be a discrepancy with Zeplin there.

kevinbarabash:
The tricky thing is we're using the IconButton's focus ring for positioning when really we should be using the bounds of the Icon itself.

pamelafox:
I've updated the diff to use padding:16px on the left and right sides of the container, and I updated the examples that use multiple buttons to have <Strut size={16}/>.

I'm unsure what to do about the IconButton's on the left hand side of the toolbar. They are currently padded in at 16px, which means they don't match the Zeplin screenshots, but they are nicer when you interact with them, since there's nice whitespace around their focus ring. I could add in a negative margin to push them over, but then there'll be less whitespace around the focus ring.

kevinbarabash:
@pamelafox we talked about this on Friday in our regular Wonder Blocks sync.  I'm going to type up a doc that describes a proposed direction forward which should make it easier to implement zeplin layouts using `IconButton` in general and `Toolbar` in particular.

pamelafox:
OK, after reading Kevin's doc, I believe this PR is what's desired for <Toolbar>: 16px left and right, struts to space out multiple components. (And <IconButton> changes will get this to match Zeplin better) Let me know if you'd like other changes, thanks!

pamelafox:
@kevinbarabash Good catch, I've added the <Strut>. (Not sure if you also wanted icon-button-core changes in this PR, my assumption was that it's a separate change).

kevinbarabash:
I'm okay with making the icon-button-core changes separately, but we should do them before our next release.

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -63,10 +63,7 @@ export default class Toolbar extends React.Component<Props> {
         const contentArray = Array.isArray(content) ? content : [content];
         // Remove empty/null/undefined elements, then map to Views
         return contentArray.filter((c) => c).map((content, i) => (
-            <View
-                style={[sharedStyles.content, sharedStyles.verticalAlign]}
-                key={i.toString()}
-            >
+            <View style={[sharedStyles.verticalAlign]} key={i.toString()}>
```
kevinbarabash: This can be `style={sharedStyles.verticalAlign}` although I don't think this is necessary if you set `alignItems: "center"` on styles.container.  Then you can get rid of the style prop here.  Also, since there's no padding on the wrapper is the `filter` necessary?

packages/wonder-blocks-toolbar/components/toolbar.js
```diff
@@ -61,14 +61,8 @@ export default class Toolbar extends React.Component<Props> {
...
-        // Remove empty/null/undefined elements, then map to Views
-        return contentArray.filter((c) => c).map((content, i) => (
-            <View
-                style={[sharedStyles.content, sharedStyles.verticalAlign]}
-                key={i.toString()}
-            >
-                {content}
-            </View>
+        return contentArray.map((content, i) => (
+            <View key={i.toString()}>{content}</View>
```
kevinbarabash: Requesting one last change: this can be
```
renderContent(content: React.Node) {
    return Array.isArray(content) ? content : [content];
}
```

kevinbarabash:
> I had to add that on the leftColumn/rightColumn, to override alignItems=stretch.

That makes sense since it's the immediate parent of the items in leftContent/rightContent.

pamelafox:
@kevinbarabash Removed the <View> wrapper in renderContent. DivSoup--

pamelafox:
Brought back <View> wrapper, since React wants each component to have a key. Kevin suggests long term that people pass in their own View for each side.

packages/wonder-blocks-toolbar/components/toolbar.js
jeresig: Yeah, agreed with Kevin - I don't think we want the <View> wrapper here as it means that the Flex styling of the individual content items won't come through.

jeresig: Ok! So thinking about this problem more - it looks like you want to avoid having to add keys to the items that you're inserting. I think the easiest way to do this is to pass in a React.Fragment for the value of leftContent/rightContent instead of an array.

In the example below you currently have:

```
<Toolbar size="small" rightContent={[smallButton, strut, secondaryButton]} />
```

We can instead do:

```
<Toolbar size="small" rightContent={<React.Fragment>
	{smallButton}
	{strut}
	{secondaryButton}
</React.Fragment>} />
```

Once we finish upgrading to Babel 7 the `<React.Fragment>`s can just become: `<>...</>`.

So in this case I think we can just get rid of "renderContent" entirely and just put the contents directly into the page, don't bother with the key insertion at all (leave that up to the user - if they actually want an array then it's up to them to have keys on the items).

jeresig: Given what I said above I think we should remove the default values of leftContent and rightContent above (or make them `null` and make the properties required in the Flow type).

pamelafox:
Okay, I've updated Toolbar to render the leftContent and rightContent directly. The default values are null, and their flow types are React.Node (curiously, that's what it was before!) Both leftContent and rightContent are marked as optional. The demos now use React.Fragment when they want to pass multiple nodes.

pamelafox:
Hm, codecov says coverage decreased, but toolbar.js is shown as 100% coverage, so I think its a line number difference change.

PR #313, opened 2018-08-31
Update `isExternalUrl` func to use `indexOf("//") >= 0` instead of Regex

jangmi-jo:


jangmi-jo:
https://khanacademy.atlassian.net/browse/WB-344

jeresig:

Built with commit 36c79774c87e1c3fe975a4dc1186119946efa3fa

https://deploy-preview-313--wonder-blocks.netlify.com

packages/wonder-blocks-core/util/get-clickable-behavior.js
```diff
@@ -14,7 +14,7 @@ import {withRouter} from "react-router-dom";
 import ClickableBehavior from "../components/clickable-behavior.js";
 
 function isExternalUrl(url: string) {
-    return /^https?:\/\//i.test(url);
+    return url.indexOf("//") >= 0;
```
kevinbarabash: We should have tests for this.  Currently `/foo//bar` would be incorrectly identified as being external.  Maybe we can do:
```
/^(https?:)?\/\//i.test(url);
```

jangmi-jo: Sure. I'll update the check and add test as well.

kevinbarabash:
We'll have to re-run the tests once the `npm` outage subsides.

PR #314, opened 2018-08-31
Implementing ADR-93 to ensure NoSSR doesn't cascade

somewhatabstract:
In this change, we add context awareness to the `NoSSR` component so that only the highest level instance in a component tree will actually delay true rendering until the second render.

This change is pre-empting that ADR-93 will get approved.


jeresig:

Built with commit a25b7007010e6d6ac7270741e5bcccb3d9a2a74f

https://deploy-preview-314--wonder-blocks.netlify.com

packages/wonder-blocks-core/components/no-ssr.test.js
```diff
@@ -11,34 +11,60 @@ describe("NoSSR", () => {
...
-            // Therefore, if the placeholder has been called, it must have
-            // been called first.
-            expect(mockPlaceholder).toHaveBeenCalledTimes(1);
-            done();
-        };
+        // Assert
+        // Our promise doesn't resolve until the children render, therefore
+        // we don't get here until that and so if the placeholder has been
+        // called, it must have been called first.
+        expect(mockPlaceholder).toHaveBeenCalledTimes(1);
```
kevinbarabash: Given the title of the test I would expect to see and `expect` for what was rendered into `nodes`.

packages/wonder-blocks-core/components/no-ssr.test.js
```diff
@@ -11,34 +11,60 @@ describe("NoSSR", () => {
...
 
-        arrangeAct(assert);
+        // Assert
+        // Our promise doesn't resolve until the children of our nested NoSSR
+        // are rendered, therefore we don't get here until that and so if the
+        // parent placeholder has been called, it must have been called first.
+        // In addition, if our code is working right, the placeholder of the
+        // nested component should have been skipped.
+        expect(mockPlaceholder).toHaveBeenCalledTimes(1);
+        expect(mockPlaceholderNotCalled).not.toHaveBeenCalled();
```
kevinbarabash: We should also check what's in `nodes` to verify that everything in the tree has been rendered.

kevinbarabash: There isn't a snapshot test that checks this nested case.  Also, what's rendered in the snapshot is
```
This is rendered only by the client for all but the very first render of the component tree.
```
and what I'd actually like to check is what gets rendered after the first render.

somewhatabstract: That is what gets rendered after the first render.

somewhatabstract: @kevinbarabash Not sure what you're asking to be changed here, since that snapshot is already testing what gets rendered after the first render.

kevinbarabash: Sorry for wording this imprecisely.  The snapshots handle the first render, I'm curious about what gets rendered on the second render.  Does `NoSSR` actually render its children like it's supposed to?

somewhatabstract: I am confused. The snapshot is showing the subsequent renders, not the first.

packages/wonder-blocks-core/components/no-ssr.test.js
somewhatabstract: Well, this isn't testing that the render is working, just that the call to render is being called. Snapshot tests etc. cover the actual return result getting used in the output. This is more about ensuring order of events match our expectations.

somewhatabstract: Renamed the test

PR #317, opened 2018-08-31
Ensure we have a single source of obscurity truth

somewhatabstract:
Made me nervous having the same code in two places, especially with one location lacking tests and us having made fixes in this code in the tested one.

So, this puts it all in one location and then has the two places import it from there. Since it's only used internally, I didn't add this to the exports of core, but rather each package using it just grabs it from there. This seems reasonable to me and MUCH better than duplicating this code.


jeresig:

Built with commit 10a6bb872cca8ac64726917708064a21a0829a07

https://deploy-preview-317--wonder-blocks.netlify.com

jeresig:

Built with commit 452ec8a484865754dd8723bee83d85d838b73556

https://deploy-preview-317--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -12,7 +12,7 @@ import Spacing from "@khanacademy/wonder-blocks-spacing";
 import {View} from "@khanacademy/wonder-blocks-core";
 
 import type {StyleType} from "@khanacademy/wonder-blocks-core";
-import visibilityModifierDefaultConfig from "../util/visibility-modifier.js";
+import visibilityModifierDefaultConfig from "../../wonder-blocks-core/util/visibility-modifier.js";
```
kevinbarabash: This results in the code appearing in multiple packages and results in packages being larger than they need to be.  Please export this from the `wonder-blocks-core` package.

kevinbarabash: >  Since it's only used internally, I didn't add this to the exports of core, but rather each package using it just grabs it from there. This seems reasonable to me and MUCH better than duplicating this code.

I should read the summary before requesting changes.  I wonder if it makes sense to have a `wonder-blocks-internal` that's used for this kind of stuff so that we can avoid these anti-patterns.  The reason why I'm against requiring files directly is that it can lead to packages being larger (and in some case much larger) than they need to be.  In a previous PR I cleaned up a number of places where we were doing this.

somewhatabstract: Yes, I understand that. In this case, it felt like the right move, but we could add an internals package or just accept this as an export of core. The issue I have with either of those is that once something is on NPM, it's public and therefore its API is public; I don't know that we want these things to be public at all.

kevinbarabash: I see the concern.  Paying for `npm` to have a single private `npm` module seems like overkill.  Maybe we could just have a `internal` folder that lives outside of `packages` for this reason.  I'm not really sure the best way to address this so let's go with this for now.  Please add a comment to the import to discourage people from using this pattern except when it's absolutely necessary.  I want to avoid bloating our packages at worse and or ending up with circular dependencies at worst.

somewhatabstract: I think we can do two things here and I'll do both.

1) Move the files in question to an internal-only folder
2) Comment on the places they are imported

That way we can eventually lint for this and ensure that only files in an `internal-only` folder are used this way, which will help us identify it during reviews and discourage it from happening.

somewhatabstract: @kevinbarabash Looks like this eslint rule would work. Any issues with adding this library of rules as a dev dependency?

https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/no-restricted-paths.md

kevinbarabash: Go for it.  I wasn't sure how to use it to enforce the constraint we want, but maybe there's a way to have different restricted paths for each package.

.eslintrc
```diff
@@ -22,17 +25,110 @@
...
+                        "target": "./packages/(?!.*test.js).*",
+                        "from": "utils"
+                    },
+                    {
+                        "target": "./packages/wonder-blocks.*",
+                        "from": "./shared-unpackaged"
+                    },
+                    {
+                        "target": "./packages/wonder-blocks-button",
+                        "from": "./packages/wonder-blocks-(?!button).*"
```
kevinbarabash: It's too bad this plugin doesn't provide a way to specify allowed paths.

somewhatabstract:
@kevinbarabash Looks like the monorepo plugin does what we need, mostly. So, I've adjusted things accordingly. It doesn't stop imports from non-package locations, so I've left the restriction zones for those cases. However, I think we're in better shape.

PR #319, opened 2018-09-01
Increase icon package test coverage

somewhatabstract:
Add some more testing and realign so that the test files map to the things they're testing

### DIFF
--- a/packages/wonder-blocks-icon/__snapshots__/generated-snapshot.test.js.snap
+++ b/packages/wonder-blocks-icon/__snapshots__/generated-snapshot.test.js.snap
@@ -834,7 +834,25 @@ exports[`wonder-blocks-icon example 2 1`] = `
 `;
 
 exports[`wonder-blocks-icon example 3 1`] = `
-<div>
+<div
+  className=""
+  style={
+    Object {
+      "alignItems": "stretch",
+      "borderStyle": "solid",
+      "borderWidth": 0,
+      "boxSizing": "border-box",
+      "display": "flex",
+      "flexDirection": "column",
+      "margin": 0,
+      "minHeight": 0,
+      "minWidth": 0,
+      "padding": 0,
+      "position": "relative",
+      "zIndex": 0,
+    }
+  }
+>
   <svg
     className=""
     height={16}
@@ -907,7 +925,25 @@ exports[`wonder-blocks-icon example 3 1`] = `
 `;
 
 exports[`wonder-blocks-icon example 4 1`] = `
-<div>
+<div
+  className=""
+  style={
+    Object {
+      "alignItems": "stretch",
+      "borderStyle": "solid",
+      "borderWidth": 0,
+      "boxSizing": "border-box",
+      "display": "flex",
+      "flexDirection": "column",
+      "margin": 0,
+      "minHeight": 0,
+      "minWidth": 0,
+      "padding": 0,
+      "position": "relative",
+      "zIndex": 0,
+    }
+  }
+>
   Here is an icon inline:
   <svg
     className=""
--- a/packages/wonder-blocks-icon/components/icon.test.js
+++ b/packages/wonder-blocks-icon/components/icon.test.js
@@ -1,0 +1,275 @@
+// @flow
+import * as React from "react";
+import * as Core from "@khanacademy/wonder-blocks-core";
+import icons from "../util/icon-assets.js";
+import {getPathForIcon, viewportPixelsForSize} from "../util/icon-util.js";
+
+import {mount, unmountAll} from "../../../utils/testing/mount.js";
+
+// We mock things out so that we're in control of what really gets rendered.
+// Means we can test that we're using addStyle to generate the component
+// and then verify what that component has going on.
+jest.mock("@khanacademy/wonder-blocks-core", () => {
+    const mockStyledSVGComponent = jest.fn((props: any) => <div>Pretend</div>);
+    return {
+        _mockStyledSVGComponent: mockStyledSVGComponent,
+        addStyle: jest.fn(() => mockStyledSVGComponent),
+    };
+});
+
+// We cannot make the mock outside of the jest.mock call, so we make it inside
+// then attach it to the exports and grab it here.
+const mockStyledSVGComponent: JestMockFn<any, any> = (Core: any)
+    ._mockStyledSVGComponent;
+
+// Also, let's type up a couple of other mocks to shut flow up.
+const mockGetPathForIcon: JestMockFn<any, any> = (getPathForIcon: any);
+const mockViewportPixelsForSize: JestMockFn<
+    any,
+    any,
+> = (viewportPixelsForSize: any);
+
+jest.mock("../util/icon-util.js", () => ({
+    getPathForIcon: jest.fn(() => ({})),
+    viewportPixelsForSize: jest.fn(() => ({})),
+}));
+
+describe("Icon", () => {
+    beforeEach(() => {
+        jest.clearAllMocks();
+        unmountAll();
+    });
+
+    test("creates a styled svg using addStyle", async () => {
+        // Arrange
+        const importModulePromise = import("./icon.js");
+
+        // Act
+        await importModulePromise;
+
+        // Assert
+        expect(Core.addStyle).toHaveBeenCalledWith("svg");
+    });
+
+    test("applies aria-label to svg", async () => {
+        // Arrange
+        const {default: Icon} = await import("./icon.js");
+        const render = new Promise((resolve) => {
+            const nodes = (
+                <div>
+                    <Icon
+                        ref={() => resolve()}
+                        aria-label="ARIA LABEL"
+                        icon={icons.add}
+                    />
+                </div>
+            );
+
+            mount(nodes);
+        });
+
+        // Act
+        await render;
+
+        // Assert
+        expect(mockStyledSVGComponent).toHaveBeenCalledTimes(1);
+        expect(mockStyledSVGComponent.mock.calls[0][0]).toEqual(
+            expect.objectContaining({"aria-label": "ARIA LABEL"}),
+        );
+    });
+
+    test("calls getPathForIcon", async () => {
+        // Arrange
+        const {default: Icon} = await import("./icon.js");
+        const render = new Promise((resolve) => {
+            const nodes = (
+                <div>
+                    <Icon
+                        ref={() => resolve()}
+                        icon={icons.add}
+                        size="medium"
+                    />
+                </div>
+            );
+
+            mount(nodes);
+        });
+
+        // Act
+        await render;
+
+        // Assert
+        expect(mockGetPathForIcon).toHaveBeenCalledTimes(1);
+        expect(mockGetPathForIcon).toHaveBeenCalledWith(icons.add, "medium");
+    });
+
+    test("calls viewportPixelsForSize with size from props and asset size from getPathForIcon", async () => {
+        // Arrange
+        const {default: Icon} = await import("./icon.js");
+        mockGetPathForIcon.mockImplementationOnce(() => ({
+            assetSize: "large",
+            path: "TESTPATH",
+        }));
+        const render = new Promise((resolve) => {
+            const nodes = (
+                <div>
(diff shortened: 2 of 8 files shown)


packages/wonder-blocks-icon/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -834,7 +834,25 @@ exports[`wonder-blocks-icon example 2 1`] = `
 `;
 
 exports[`wonder-blocks-icon example 3 1`] = `
-<div>
+<div
```
somewhatabstract: Changed fixtures to `View`, so that added these snapshot changes.

packages/wonder-blocks-icon/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -907,7 +925,25 @@ exports[`wonder-blocks-icon example 3 1`] = `
 `;
 
 exports[`wonder-blocks-icon example 4 1`] = `
-<div>
+<div
```
somewhatabstract: Changed fixtures to `View`, so that added these snapshot changes.

packages/wonder-blocks-icon/index.test.js
```diff
@@ -1,79 +1,13 @@
...
+        // Arrange
+        const importedModule = import("./index.js");
 
-import type {IconSize} from "@khanacademy/wonder-blocks-icon/util/icon-assets.js";
+        // Act
+        const result = await importedModule;
 
-const SIZES = ["small", "medium", "large", "xlarge"];
-
-const DUMMY_ICON_MEDIUM_ONLY = {
```
somewhatabstract: Moved to `util/icon-util.test.js`

jeresig:

Built with commit 3242ef3bc2cc7f74fd060803441e8e65c5a62fa1

https://deploy-preview-319--wonder-blocks.netlify.com

packages/wonder-blocks-icon/util/icon-util.test.js
```diff
@@ -0,0 +1,89 @@
...
+        const iconAsset: IconAsset = ({}: any);
+
+        // Act
+        const underTest = () => getPathForIcon(iconAsset, "medium");
+
+        // Assert
+        expect(underTest).toThrowErrorMatchingInlineSnapshot(
+            `"Icon does not contain any valid asset sizes!"`,
+        );
+    });
```
somewhatabstract: Everything in this file is as it was when it was in the root dir of the package, except for this last test case which tests the error case that is in the function.

packages/wonder-blocks-icon/util/icon-util.test.js
```diff
@@ -0,0 +1,89 @@
...
+
+    test("no valid asset sizes, throws", () => {
+        // Arrange
+        const iconAsset: IconAsset = ({}: any);
+
+        // Act
+        const underTest = () => getPathForIcon(iconAsset, "medium");
+
+        // Assert
+        expect(underTest).toThrowErrorMatchingInlineSnapshot(
```
somewhatabstract: BTW, love the new inline snapshots for things like this. Just write the test without the string, then run `yarn jest -u` and it automatically updates the test definition with the string. Awesomeness!

packages/wonder-blocks-icon/components/icon.test.js
```diff
@@ -0,0 +1,275 @@
...
+import * as React from "react";
+import * as Core from "@khanacademy/wonder-blocks-core";
+import icons from "../util/icon-assets.js";
+import {getPathForIcon, viewportPixelsForSize} from "../util/icon-util.js";
+
+import {mount, unmountAll} from "../../../utils/testing/mount.js";
+
+// We mock things out so that we're in control of what really gets rendered.
+// Means we can test that we're using addStyle to generate the component
+// and then verify what that component has going on.
```
kevinbarabash: I wonder if eventually we'll have a `wonder-blocks-test-tools` package. 🤔

packages/wonder-blocks-icon/components/icon.test.js
```diff
@@ -0,0 +1,275 @@
...
+    return {
+        _mockStyledSVGComponent: mockStyledSVGComponent,
+        addStyle: jest.fn(() => mockStyledSVGComponent),
+    };
+});
+
+// We cannot make the mock outside of the jest.mock call, so we make it inside
+// then attach it to the exports and grab it here.
+const mockStyledSVGComponent: JestMockFn<any, any> = (Core: any)
+    ._mockStyledSVGComponent;
```
kevinbarabash: Where does `JestMockFn` come from?  Does `flow` just know about it?

packages/wonder-blocks-icon/components/icon.test.js
```diff
@@ -0,0 +1,275 @@
...
+// Also, let's type up a couple of other mocks to shut flow up.
+const mockGetPathForIcon: JestMockFn<any, any> = (getPathForIcon: any);
+const mockViewportPixelsForSize: JestMockFn<
+    any,
+    any,
+> = (viewportPixelsForSize: any);
+
+jest.mock("../util/icon-util.js", () => ({
+    getPathForIcon: jest.fn(() => ({})),
+    viewportPixelsForSize: jest.fn(() => ({})),
```
kevinbarabash: Cool.  I need to learn more about `jest`'s mocks... they're super powerful.

packages/wonder-blocks-icon/components/icon.test.js
```diff
@@ -0,0 +1,275 @@
...
+            mount(nodes);
+        });
+
+        // Act
+        await render;
+
+        // Assert
+        expect(mockStyledSVGComponent).toHaveBeenCalledTimes(1);
+        expect(mockStyledSVGComponent.mock.calls[0][0]).toEqual(
+            expect.objectContaining({"aria-label": "ARIA LABEL"}),
```
kevinbarabash: It seems like this test could be accomplished using a snapshot.  I know the snapshot tests are a bit noisy.  Once we get image comparison tests I'd like to eventually remove all visual CSS props from the snapshots so that we're only checking props like `aria-label` and other things that don't have any impact on the visuals.

packages/wonder-blocks-icon/components/icon.test.js
```diff
@@ -0,0 +1,275 @@
...
+        });
+
+        // Act
+        await render;
+
+        // Assert
+        expect(mockStyledSVGComponent).toHaveBeenCalledTimes(1);
+        expect(mockStyledSVGComponent.mock.calls[0][0]).toEqual(
+            expect.objectContaining({
+                viewBox: `0 0 ${expectedAssetSize} ${expectedAssetSize}`,
```
kevinbarabash: You should be able to find the `svg` element and check the prop directly.  Why not check these properties of the markup that way?

packages/wonder-blocks-icon/util/icon-util.test.js
kevinbarabash: Cool!

packages/wonder-blocks-icon/components/icon.test.js
somewhatabstract: It's in the `jest` flow-typed definition.

packages/wonder-blocks-icon/components/icon.test.js
somewhatabstract: It seems likely.

packages/wonder-blocks-icon/components/icon.test.js
somewhatabstract: Enzyme didn't like me trying that; kept complaining that the element wasn't rendered with enzyme, even though it is. I suspect it's a side-effect of test isolation, but it's important to ensure that we isolate `addStyle` side-effects and test the boundary, hence why it's this way.

packages/wonder-blocks-icon/components/icon.test.js
somewhatabstract: The snapshot tests test this with `addStyle`. These are unit tests, isolating the component from externals like `addStyle` to ensure the component is doing the right thing.

packages/wonder-blocks-icon/components/icon.test.js
kevinbarabash: I'm not sold on the value of test isolation.  I've found that in general it results in test code that's brittle to even non-functional changes.  I also like tests that exercise more of the stack because it exercises those boundaries in between layers which are hard to get at when everything's isolated.  What benefit do we get from isolating these tests from using the real `addStyle`?

somewhatabstract: When bugs occur it helps us identify where. We totally should also have integration tests that exercise more portions, but that's not the role of unit tests. Unit tests are supposed to test a component at its boundaries so we can be confident that as a component, given the inputs it is designed to accept, it works. That's what isolation gets us.

We also need integration tests that remove that isolation. I wonder where they should live or how we should differentiate them?

kevinbarabash: I feel like we currently have a mix of both unit tests and integration tests.  I think we should do the following (in a separate diff):
- split all current .test.js files into .unit_test.js and .int_test.js
- generate separate codecov reports for each set of tests
- have codecov merge the results when reporting the coverage in PRs (we do this in KaTeX since we have unit and image comparison tests there)

It's probably worth writing an ADR about this so I'll do that.

I'm glad that you're on-board with integration tests.

I do worry about the cost of isolation needed for writing some unit tests.  My take on it, is that if a unit fulfills its duties to its callers then I don't really care how it does it's working.  The issue with isolation is that it requires us to know how it does its work.  This means that if I want to refactor something I  now have to refactor both the unit and its tests.

If I have a set of integration tests for the same unit and they have 100% coverage, if I make changes and they still pass then I should be good to go, plus they have the benefit of testing the seams.  I would advocate for writing integration tests first b/c in general they seems to offer more benefit at less cost.

While an integration test will identify that an issue occur it won't be as precise about what is causing the issue.  So the trade off is really between how long it'll take to identify what's causing the bug vs. the cost of writing the tests and the potential cast to refactor the tests if needed.

It's hard to judge these costs b/c many of them are hypothetical.  Maybe there won't be any bugs in this part of the code or maybe they'll be easy to identify without a unit test or maybe we won't need to refactor this part of the code.  It's hard to say.

I think where unit tests really shine are those situations where there are no dependencies or if there are they're being passed to the function making mocking trivial.

For context, at my previous job we had lots of unit tests but very few integration tests resulting in a very brittle codebase that nobody wanted to refactor b/c doing so would break all the tests.  Our goal at the time was to both have 100% unit test coverage and also completely isolate the units we were testing.  This approach felt very dogmatic an un-pragmatic.

I don't think we talk very often about different ways of testing code and the trade-offs between them.  This is mainly b/c there are two ways to test things in webapp: unit tests and fixtures.

Thank you for writing these unit tests.  I hope I didn't come off as sounding ungrateful.  In order for wonder-blocks to be successfully it needs to be well tested.

kevinbarabash:
@somewhatabstract can you rebase this?

packages/wonder-blocks-icon/components/icon.test.js
somewhatabstract: In the end both are super valuable, we often (almost always) undervalue the cost of our future time, yet the data comes back repeatedly that the better test coverage (that means unit tests to cover interfaces, integration tests to cover integrated components, end-to-end tests to cover behaviours, etc.), better up front design, and code review all help reduce the lifetime cost of software because they not only avoid bugs in the first place, but simplify their detection and mitigation further down the line.

Cutting corners on unit testing will undoubtedly cost down the line. Also, anything that encourages devs to cut corners like is likely to lead to code that is not as well written as it would have been if the developer were having to ensure that a piece of code's interfaces were well tested.

Therefore, I don't think this is an either/or discussion. We need to have both in order to have confidence in the code we ship.

kevinbarabash: I agree that it isn't an either/or discussion.  I just wanted to call out that in certain circumstances unit tests have a higher cost than the benefit they provide.

PR #320, opened 2018-09-01
Ignore all the dist files

somewhatabstract:
Before this change, the dist/es folders were appearing in the output for `yarn coverage`. We don't want that, so this ensures it doesn't happen.


jeresig:

Built with commit c8cb3a5bf6a63935185b6ddb71b03c12e4878d66

https://deploy-preview-320--wonder-blocks.netlify.com

PR #321, opened 2018-09-11
Remove the 4px of padding from tertiary kind buttons

pamelafox:
Addresses https://khanacademy.atlassian.net/browse/WB-447

Removes 4px of padding from tertiary buttons and associated 4px in the hover lines.

@chenyejia Was there a reason for the 4px originally? Kevin shared a design diagram where it looks like there's 0px of padding, but I'm wondering if there's some reason for the 4px.


jeresig:

Built with commit 8f6742bde3cd35d8d43ebf53dc1dee5d4b11aebc

https://deploy-preview-321--wonder-blocks.netlify.com

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -276,16 +276,16 @@ const _generateStyles = (color, kind, light, iconWidth) => {
...
+                paddingRight: 0,
             },
             focus: {
                 ":after": {
                     content: "''",
                     position: "absolute",
                     height: 2,
-                    width: `calc(100% - 8px - ${iconWidth}px)`,
-                    right: 4,
+                    width: `calc(100% - ${iconWidth}px)`,
```
kevinbarabash: This could probably be simplified further but would require some restructuring so we'll leave it as is for now.

chenyejia:
@pamelafox The old button designs for focused tertiary buttons had a ring around it (https://zpl.io/V4AekyN), and I think I forgot to remove the padding when it was changed to an underline 😅

PR #322, opened 2018-09-15
Fix all console errors emitted during tests.

jeresig:
I was bothered by the console errors that were coming up during testing, I was worried that they might be obscuring real errors so I went and fixed them all.

- IconButton: This was fixed by creating an ErrorBoundary around the component that intentionally swallows the error. This was discussed in https://github.com/facebook/react/issues/11098 and required an upgrade to React 16.4.3 - which, in turn, required an upgrade of the Enzyme React Adapter. I pulled in the utility for testing these in the future into `utils/testing/expect-render-error.js`.
- Dropdown: We needed to not pre-mount the button before passing it in. The tests worked fine even without doing that!
- ModalBackdrop & ModalLauncher: Both needed to reference real Modals, not mock divs.

I also modified test-setup.js to throw an error if a `console.error` is ever logged in our tests -- these should be failing as they might indicate a problem with our code! We may want to consider doing this with `console.log`, as well (but we don't currently have any of those).

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,9 +32,6 @@
   },
   "author": "",
   "license": "MIT",
-  "//": [
-    "enzyme-adapter-react-16: TODO(sophie): We should revert back to the latest release of this when the issue at https://github.com/airbnb/enzyme/issues/1509 is fixed"
-  ],
   "devDependencies": {
     "babel-core": "6.26.3",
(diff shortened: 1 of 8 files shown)


jeresig:

Built with commit 7f0d89beb639c1de84071234184292f564c86387

https://deploy-preview-322--wonder-blocks.netlify.com

PR #323, opened 2018-09-16
Add `visitable` prop on Link

jangmi-jo:
https://khanacademy.atlassian.net/browse/WB-391
In this commit, I add a new prop called `visitable` which will default
to `false`. Since only primary links have a visited state we shouldn't
allow it on other link types.


jeresig:

Built with commit 090950bc097697decb1bb1cb1fd1b3eb0f45eb62

https://deploy-preview-323--wonder-blocks.netlify.com

PR #324, opened 2018-09-19
Fixes #WB-459: Don't include focus ring in the size of the IconButton

kevinbarabash:
Since the focus rings aren't shown in zeplin mocks, not including them in IconButton will make it easier to implement designs.  This also brings IconButton more inline with Button which doesn't include the focus ring in its size either.


jeresig:

Built with commit 175899edd3d2311d347f5758b5be04069de0bbcc

https://deploy-preview-324--wonder-blocks.netlify.com

packages/wonder-blocks-icon-button/components/icon-button.js
```diff
@@ -116,6 +116,10 @@ export type SharedProps = {|
  * light (whether the IconButton will be rendered on a dark background),
  * disabled , test ID, and custom styling.
  *
+ * Note: the size of an IconButton match the size of icon it wraps which is
+ * 24x24 pixels.  The focus ring which is display on hover and focus is much
+ * large but does not affect its size.  This matches the behavior of Button.
```
chenyejia: Grammar nits: "match**es** the size of **the** icon"; "which is display**ed** on hover", "much large**r** but"

kevinbarabash: Thanks.  I think I need grammarly. 😞

packages/wonder-blocks-icon-button/components/icon-button.js
```diff
@@ -116,6 +116,11 @@ export type SharedProps = {|
  * light (whether the IconButton will be rendered on a dark background),
  * disabled , test ID, and custom styling.
  *
+ * Note: the size of an IconButton matches the size of the icon it wraps 
+ * which is 24x24 pixels.  The focus ring, which is displayed on hover and 
+ * focus, is much larger but does not affect its size.  This matches the 
+ * behavior of Button.
```
jeresig: Sort of - the focus ring on button is much less pronounced - it's easier to put two buttons next to each other, but you can't put two iconbuttons next to each other (without spacing).

packages/wonder-blocks-icon-button/components/icon-button.md
```diff
@@ -9,7 +9,7 @@ const styles = StyleSheet.create({
         flexDirection: "row",
     },
     sideMargins: {
-        marginRight: 10,
+        marginRight: 16,
```
jeresig: I think we should update the toolbar demos and all the places where iconbuttons are used to use Strut spacing instead of CSS margins. We should have our demo code represent how people should be using the components.

Right now I'm worried that the default state for IconButton is going to end up with people putting iconbuttons next to each other and having the focus rings overlap - so we should be putting the defaults in for whatever the designers want with some Struts. Same goes for buttons too - we should update all the demos to represent this (and it should be part of this PR as the problem is going to become much more pronounced).

pamelafox:
@kevinbarabash Can you merge from master, now that my change is in? I'd like to see what the toolbar examples look like. I'm trying it locally and see them a few pixels off, but I think that may be due to the SVG icon internal placement.
Also, does Github let you fix the typo in the PR title?

packages/wonder-blocks-icon-button/components/icon-button.md
kevinbarabash: I'll update the examples.  I doubt people will end up with overlapping focus rings since the focus ring is never shown in any of the mocks, only the icon itself is.

kevinbarabash:
@pamelafox I will update.  some of the icons have internal padding and the zeplin mocks nudge them a bit.  I'll make sure the examples are nudge appropriate and include comments that nudge is sometimes necessary.

packages/wonder-blocks-icon-button/components/icon-button.js
kevinbarabash: A person shouldn't be putting buttons next to each other either.  If they did, the focus ring of one button would (and does) overlap the other, obviously not to the extent that would occur with IconButtons that adjacent without spacing.

jeresig: True! I think we should provide clarity in the documentation around this as I'm not sure the best practices are apparent from what we've presented (especially in the context of a toolbar).

kevinbarabash:
Screenshot from the most recent netlify deploy for this PR.
screen shot 2018-09-20 at 9 47 41 pm

kevinbarabash:
I'm going to add some more comments in the IconButton docs.

packages/wonder-blocks-toolbar/components/toolbar.md
```diff
@@ -31,13 +31,15 @@ const rightContent = <React.Fragment>
...
+**Note:** The `IconButton`s are nudged to the left by 1px to match designs.
+
 ```js
 const {icons} = require("@khanacademy/wonder-blocks-icon");
 const Button = require("@khanacademy/wonder-blocks-button").default;
 const IconButton = require("@khanacademy/wonder-blocks-icon-button").default;
 const {Strut} = require("@khanacademy/wonder-blocks-layout");
 
-const zoomOutButton = <IconButton icon={icons.zoomOut} kind="primary"/>;
+const zoomOutButton = <IconButton icon={icons.zoomOut} kind="primary" style={{marginLeft:-1}}/>;
```
jeresig: Doesn't this feel wrong to you? I feel like either the designs need to be changed or the icons themselves need to be changed. Are we not centering the icons correctly or something? I think if we can't just drop a button or iconbutton into a toolbar and have it "work correctly" then we're failing somewhere! Maybe we can discuss this in the sync tomorrow?

kevinbarabash: This is correct.  All of the icons are 24x24 but many of the visuals occupy some subset of that space.  In this case the `zoomOut` icon there's 1px padding on the left side.  In the case of the `dismiss` icon, the icon itself is only 12x12 so there's 6px of padding around it.  If we want IconButton to "just work" then we need to include additional information about it in the `IconAsset` data structure.  We should probably do that, but it feels like it's outside the scope of this PR.

kevinbarabash: I'm not actually sure how we'd get this to "just work" since there are some situation where we want to use the visual size of the icon (aligning to the edge of the Toolbar) and other times we want to use the 24x24 bounding box (spacing between icons).  We should talk to design though about the spacing.  In the mocks we happen to use 24x24 bounding box between the zoom-in and zoom-out icons, but those icons are close enough to 24x24 that maybe the intention was to actually use the visual bounding box for that spacing.

pamelafox: I'm interested in this too. I guess I'd feel more confident if the Zeplin mocks also included the focus state, since I'd know that the designer understood the visual implications of the focus rings. Right now I'd have to do a similar margin tweaking on my use of IconButton, and it feels a bit funny, as John said. Did you get a consensus on this in the design sync?

kevinbarabash: There were no designers in the sync today.  I'll try to get an answer to this next week from design.  I think that this is a step in the right direction though.

PR #325, opened 2018-09-19
Fixes WB-450 and WB-452: Replace onClickCloseButton with Context API and add custom modal example

kevinbarabash:
Using the Context API simplifies things quite a bit.  I was actually investigating WB-449 (OneColumnModal focus trap isn't work) b/c I had a hunch it was a result of creating a custom modal by wrapping OneColumnModal which is what I was doing when I ran into the issue.  It appears though that it's webapp specific.  I'll continue to investigate that issue tomorrow.  In the meantime I figured I may as well introduce the Context API so we don't have to pass `onClickCloseButton` manually like we had to do before when creating a custom modal.


jeresig:

Built with commit 5a6d632bf73fa1dbe2da29dea286b338cdac7402

https://deploy-preview-325--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-launcher.js
```diff
@@ -35,7 +36,7 @@ type Props = {|
 
     /**
      * If the parent needs to be notified when the modal is closed, use
-     * this prop. You probably want to use this instead of `onClickCloseButton`
+     * this prop. You probably want to use this instead of `onClose`
```
matchu: nit: re-wrap text?

packages/wonder-blocks-modal/components/modal-context.js
```diff
@@ -0,0 +1,6 @@
+// @flow
+import * as React from "react";
+
+// babel-plugin-strip-flow-types doesn't know how to deal with
+// React.createContext<Context>(); where Context is a flow type.
```
matchu: I think maybe we can solve this with other idioms to help Flow understand the type? In Test Prep's `global-exam-data-context.js`, we make a type assertion about the value that we pass to `React.createContext`:

```
const GlobalExamDataContext = React.createContext((null: ?Exam));
```

I think this gets Flow to realize that it's a `React.Context<?Exam>`.

packages/wonder-blocks-modal/components/modal-panel.js
```diff
@@ -85,16 +81,22 @@ export default class ModalPanel extends React.Component<Props> {
...
-                    }
-                    light={topBackgroundColor === "dark"}
-                />
+                <Context.Consumer>
+                    {({closeModal}) => (
+                        <IconButton
+                            icon={icons.dismiss}
+                            // TODO(mdr): Translate this string for i18n.
+                            aria-label="Close modal"
+                            onClick={closeModal || onClose}
```
matchu: I wonder if this should have similar behavior to the old version: if both callbacks are provided, maybe they should both be called?

packages/wonder-blocks-modal/components/modal-context.js
kevinbarabash: Cool.  I'll give this a try.  I hope someone updates the `strip-flow-types` plugin eventually though.

packages/wonder-blocks-modal/components/modal-panel.js
kevinbarabash: I can give that a try and see what happens. I'm curious to see if any of the callback get double called.

packages/wonder-blocks-modal/components/modal-context.js
jeresig: Oh yeah, this definitely works - I had it work in my `<Layout>` PR. I had luck in doing:

```
type ContextType = {closeModal: ....type....};

export default React.createContext({closeModal: undefined}: ContextType);

```

You can see an example of how I did it here:
https://github.com/Khan/wonder-blocks/pull/311/files#diff-545f77fd7a4db77f9e1d373132cca3f9

packages/wonder-blocks-modal/components/modal-launcher.js
```diff
@@ -6,6 +6,7 @@ import FocusTrap from "./focus-trap.js";
 import ModalBackdrop from "./modal-backdrop.js";
 import ScrollDisabler from "./scroll-disabler.js";
 import type {ModalElement} from "../util/types.js";
+import Context from "./modal-context.js";
```
jeresig: I think it'd be better to call this "ModalContext".

packages/wonder-blocks-modal/components/one-column-modal.js
```diff
@@ -30,7 +30,7 @@ type BaseProps = {|
      * clicks too.)
      * @ignore
      */
-    onClickCloseButton?: () => void,
+    onClose?: () => void,
```
jeresig: It looks like this can be made mandatory now.

Also, should this property still be marked with "@ignore"? (That made sense before when we were passing it through but less so now if you're expecting people to use this property directly.)

packages/wonder-blocks-modal/components/standard-modal.js
```diff
@@ -57,19 +57,19 @@ type Props = {|
      * clicks too.)
      * @ignore
```
jeresig: Here too.

packages/wonder-blocks-modal/components/two-column-modal.js
```diff
@@ -37,7 +37,7 @@ type BaseProps = {|
      * clicks too.)
      * @ignore
      */
-    onClickCloseButton?: () => void,
+    onClose?: () => void,
```
jeresig: Here too, also remove the optional.

packages/wonder-blocks-modal/docs.md
```diff
@@ -0,0 +1,35 @@
+# Custom modals
+
+```js
+const React = require("react");
+
+const {Title} = require("@khanacademy/wonder-blocks-typography");
+const {View} = require("@khanacademy/wonder-blocks-core");
+const Button = require("@khanacademy/wonder-blocks-button").default;
+
+class CustomModal extends React.Component {
```
jeresig: Hmm - I think I have a different definition of a "Custom Modal" than you. I was imagining that a custom one would be entirely built up from the primitive components. I think calling this a "Wrapped Modal" might be more accurate.

packages/wonder-blocks-modal/docs.md
```diff
@@ -0,0 +1,35 @@
+# Custom modals
+
+```js
```
jeresig: This section would benefit from a sentence or two explaining what this code is doing and why people might want to do this.

packages/wonder-blocks-modal/components/modal-panel.js
kevinbarabash: I thought about it some more and I think we should disallow using both.  I can't think of a situation where we'd want to be able to close a modal using the backdrop, default close button, and possible a custom close button but do something special for only the default close button.  I added a `console.warn` if both `closeModal` and `onClose` are set to help people do the right thing in that situation.

packages/wonder-blocks-modal/components/two-column-modal.js
kevinbarabash: I ended up getting rid of the default prop since we pass this through to `ModalPanel` and if we define a default for this then it throws when we're using the `ModalLauncher` which defines a `closeModal` function on the `ModalContext` b/c we want to avoid situations where both are set.

packages/wonder-blocks-modal/docs.md
kevinbarabash: I don't know this is still showing up.  I've changed this to `# Wrapping Modals` and added a comment.

kevinbarabash:
I'm going to change the `console.warn` to be an exception instead.

PR #326, opened 2018-09-20
Fixes WB-449: Allow modals to be launched programmatically

kevinbarabash:
The original bug as report was that the focus trap was broken in OneColumnModal.  After some investigation it turned out that modals launched from a menu item were disappearing as soon as the focus changed.  This is because the dropdown menu was closing when this happened which meant that the menu item and the ModalLauncher that was wrapping it were being unmounted.

This PR allows ModalLauncher to be rendered without children and adds an `opened` prop which controls whether the modal is being shown or not.  The reason for going with `opened` and not `open` like some of our other components it to follow the HTML pattern for naming booleans.  I'll be updating the other boolean props to match eventually.

This feature is necessary for some SAT/TestPrep cases where modals are launched on a timer as well as launching modals in response to client-side navigation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,6 @@
     "concurrently": "^3.6.0",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.5.0",
-    "enzyme-matchers": "^6.0.4",
     "eslint": "^4.19.1",
     "eslint-config-prettier": "2.6.0",
     "eslint-plugin-flowtype": "^2.25.0",
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "concurrently": "^3.6.0",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.5.0",
+    "enzyme-matchers": "^6.0.4",
     "eslint": "^4.19.1",
     "eslint-config-prettier": "2.6.0",
     "eslint-plugin-flowtype": "^2.25.0",
--- a/packages/wonder-blocks-modal/components/modal-launcher.js
+++ b/packages/wonder-blocks-modal/components/modal-launcher.js
@@ -6,7 +6,7 @@ import FocusTrap from "./focus-trap.js";
 import ModalBackdrop from "./modal-backdrop.js";
 import ScrollDisabler from "./scroll-disabler.js";
 import type {ModalElement} from "../util/types.js";
-import Context from "./modal-context.js";
+import ModalContext from "./modal-context.js";
 
 type Props = {|
     /**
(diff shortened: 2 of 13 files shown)


jeresig:

Built with commit 1f90a6055b946331df06d6219e008d23533007bd

https://deploy-preview-326--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-launcher.js
```diff
@@ -41,6 +41,8 @@ type Props = {|
      * close events.
      */
     onClose?: () => void,
+
+    opened?: boolean,
```
jeresig: Documentation for this, please!

packages/wonder-blocks-modal/components/modal-launcher.js
```diff
@@ -65,6 +67,16 @@ type State = {|
  * the `modal` prop.
  */
 export default class ModalLauncher extends React.Component<Props, State> {
+    static getDerivedStateFromProps(props: Props, state: State) {
+        return {
+            opened: props.opened != null ? props.opened : state.opened,
```
jeresig: I think a slightly better check here would be: `typeof props.opened === "boolean" ? props.opened : state.opened;`

package.json
```diff
@@ -48,6 +48,7 @@
     "concurrently": "^3.6.0",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.5.0",
+    "enzyme-matchers": "^6.0.4",
```
jeresig: Huh - it looks like this is used but it's confusing that Enzyme's API obscures this...

kevinbarabash: Yeah, I wasn't sure how to set this up.  I couldn't find any docs but it seems to be working after installing it.  I should try uninstalling it to see if the `toExist()` matcher still works.

packages/wonder-blocks-modal/components/modal-launcher.md
```diff
@@ -152,3 +152,79 @@ const oneColumnModal = ({closeModal}) => <OneColumnModal
...
+                )}
+            />
+        </View>;
+    }
+}
+
+<Example />
+```
+
+**Warning:** Do not wrap items within an dropdown in a `ModalLauncher`.  It causes the
```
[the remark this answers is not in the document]
kevinbarabash: I was going to limit what the `children` function can return to only be things that work well with that pattern like `Button`.

packages/wonder-blocks-modal/components/modal-launcher.md
```diff
@@ -152,3 +152,79 @@ const oneColumnModal = ({closeModal}) => <OneColumnModal
...
+modal to disappear as soon as the focus changes.  The reason is that the change in
+focus results in the item that in the dropdown that was clicked to be blur which
+closes the dropdown.  This results in all of its children to unmount including the
+ModalLauncher which was wrapping the menu item.
+
+**TODO:** use flow to prevent this.
```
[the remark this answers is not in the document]
kevinbarabash: I've created 
https://khanacademy.atlassian.net/browse/WB-462 to track this.

packages/wonder-blocks-modal/components/modal-launcher.md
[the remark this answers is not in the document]
kevinbarabash: I'll do this in a separate PR.

PR #328, opened 2018-09-23
WB-448: Fix pressing down and holding and then dragging bug

jangmi-jo:
https://khanacademy.atlassian.net/browse/WB-448


jeresig:

Built with commit eaf7e21e709caf8c9bea67d774c371ae4fd5bcfa

https://deploy-preview-328--wonder-blocks.netlify.com

kevinbarabash:
I should've added more detail in https://khanacademy.atlassian.net/browse/WB-390.  In addition to not having a height, now that both width and height are being set, it can cause other issues with the layout.  For instance, if you have a horizontal layout that's a single line but you add a `<Strut size={48}/>` it will cause the height of the layout to be `48` which isn't what we want.  I was hoping that `visibility: collapse` would prevent this from happening but it does not.  

I think we'll want to replace `Strut` with `VStrut` and `HStrut`.  @somewhatabstract you've been using `Strut` a bit.  How do you feel about having separate struts for vertical/column vs horizontal/row layouts.  Also, we may want to call them `ColumnStrut` and `RowStrut`.

@jangmi-jo can you remove the Strut changes from this PR?  I'll update the ticket and we can again later in a separate PR.

jangmi-jo:
@kevinbarabash I tried to do what you asked for. Not sure if this is the best practice though. Let me know your thoughts. I'll add more tests cases once the feature is good to go.

kevinbarabash:
> Should the click event happen if I release the mouse on other item?

Yes.  I was using https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select to check the behavior of the native HTML select element.

khirasaki:
@kevinbarabash the behavior you're describing sounds good to me, thank you!

kevinbarabash:
@jangmi-jo the behaviour's looking really good.  I ran into one bug though.  

Here are the repro steps:
- open the "Betsy Appleseed" action menu
- mouse down on one of the items
- drag up until the mouse is over the opener and mouseup 
- click on the opener again

Expected behaviour: the action menu should open.

Observed behaviour: the action menu doesn't open.

kevinbarabash:
@jangmi-jo I want to merge #336 first since it fixes a bug that was recently introduced.  After that I'll do another review of this PR.

packages/wonder-blocks-core/components/clickable-behavior.test.js
```diff
@@ -511,6 +533,23 @@ describe("ClickableBehavior", () => {
         expect(onClick).toHaveBeenCalledTimes(expectedNumberTimesCalled);
     });
 
+    it("calls onClick on mouseup when the mouse was dragging", () => {
```
kevinbarabash: I think this test needs a `mousedown` to initiate the dragging.  I'm concerned that the way we're simulating events is actually triggering the drag handler.  It would be good to verify that the drag handler is being called as part of this test.  We may end of having to simulate a `dragStart` even ourselves as part of this and the other test(s) that check the drag behavior.

packages/wonder-blocks-core/components/clickable-behavior.test.js
```diff
@@ -47,12 +47,38 @@ describe("ClickableBehavior", () => {
...
+        button.simulate("dragstart", {preventDefault: jest.fn()});
+        expect(button.state("pressed")).toEqual(true);
+
+        button.simulate("mouseleave");
+        expect(button.state("pressed")).toEqual(false);
+
+        button.simulate("mouseenter", {
+            buttons: 1,
+        });
+        expect(button.state("pressed")).toEqual(true);
```
kevinbarabash: Nice!  I like this test.  I also like how there aren't any changes specific to `dropdown.js`.  I think we may run into some edge cases where a user could drag from a dropdown and release on a button.  I don't know what behavior we want into those scenarios but I think it's sufficiently edge case-y (is that a word?) that we can ignore it for now and follow up with someone the interaction guild later.  We'll probably want to do whatever browsers do by default when using built-in `<select>` and `<button>` components.

packages/wonder-blocks-core/components/clickable-behavior.test.js
```diff
@@ -511,6 +539,34 @@ describe("ClickableBehavior", () => {
...
+        button.simulate("mousedown");
+        button.simulate("dragstart", {preventDefault: jest.fn()});
+        button.simulate("mouseup");
+        expect(onClick).toHaveBeenCalledTimes(1);
+
+        button.simulate("mouseenter", {
+            buttons: 1,
+        });
+        button.simulate("mouseup");
+        expect(onClick).toHaveBeenCalledTimes(2);
```
kevinbarabash: Cool!

PR #333, opened 2018-10-03
Fixes WB-449: Allow modals to be launched programmatically (take 2)

kevinbarabash:
I forgot that the original PR was stacked and merged it by accident before the parent branch was reviewed and merged into `master`.

The original bug as report was that the focus trap was broken in OneColumnModal. After some investigation it turned out that modals launched from a menu item were disappearing as soon as the focus changed. This is because the dropdown menu was closing when this happened which meant that the menu item and the ModalLauncher that was wrapping it were being unmounted.

This PR allows ModalLauncher to be rendered without children and adds an opened prop which controls whether the modal is being shown or not. The reason for going with opened and not open like some of our other components it to follow the HTML pattern for naming booleans. I'll be updating the other boolean props to match eventually.

This feature is necessary for some SAT/TestPrep cases where modals are launched on a timer as well as launching modals in response to client-side navigation.


jeresig:

Built with commit 83d158c76e708c8d6d8904d1fa80a1f65dd1dfa4

https://deploy-preview-333--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-launcher.md
```diff
@@ -152,3 +151,77 @@ const oneColumnModal = ({closeModal}) => <OneColumnModal
     </ModalLauncher>
 </View>;
 ```
+
+## Triggering programmatically
+
+Sometimes you'll want to trigger a modal programmitically.  This can be done
```
matchu: nit: programmatically

packages/wonder-blocks-modal/components/modal-launcher.js
```diff
@@ -65,6 +70,17 @@ type State = {|
  * the `modal` prop.
  */
 export default class ModalLauncher extends React.Component<Props, State> {
+    static getDerivedStateFromProps(props: Props, state: State) {
+        return {
+            opened:
+                typeof props.opened === "boolean" ? props.opened : state.opened,
```
somewhatabstract: Doesn't this mean that if any prop changes, the modal might suddenly reappear?

kevinbarabash: The `opened` and `children` props shouldn't be used together.  I should probably throw an error if we detect that situation and provide a helpful warning message.  I'm not sure how changing any prop would cause the modal to reappear.  Can you elaborate?

somewhatabstract: Well, `getDerivedStateFromProps` gets called whenever props change, so this would then set the opened state back to `true` when props change.

kevinbarabash: We discussed this and decided it should be fine except in the case where devs pass both children and opened so I'm going to add an exception in that case.

packages/wonder-blocks-modal/components/modal-launcher.test.js
```diff
@@ -149,4 +153,30 @@ describe("ModalLauncher", () => {
...
+            shallow(
+                <ModalLauncher
+                    modal={exampleModal}
+                    opened={false}
+                    onClose={() => {}}
+                >
+                    {({openModal}) => <button onClick={openModal} />}
+                </ModalLauncher>,
+            ),
+        ).toThrow("'children' and 'opened' can't be used together");
```
jeresig: I don't think this is going to work, you'll need to use the component that I created that catches React errors and hides the console.errors.

kevinbarabash:
> Looks like the tests might be failing but thank you for catching these edge cases and throwing errors! I have a suggestion for how to fix the failing tests inline.

The tests were running out of memory and now they're timing out.  Not sure what's up with circleci.  I'll give them another try later tonight.

packages/wonder-blocks-modal/components/modal-launcher.test.js
kevinbarabash: @jeresig which tests are those being used in?

kevinbarabash: Found it.

packages/wonder-blocks-dropdown/util/types.js
```diff
@@ -6,7 +6,9 @@ import typeof OptionItem from "../components/option-item.js";
 import typeof SeparatorItem from "../components/separator-item.js";
 
 //TODO: rename into something more descriptive
-export type Item = false | React.Element<ActionItem | OptionItem | SeparatorItem>;
+export type Item =
```
jeresig: Huh - why did all these files change? Did prettier update or something?

packages/wonder-blocks-modal/components/modal-launcher.md
```diff
@@ -152,3 +151,77 @@ const oneColumnModal = ({closeModal}) => <OneColumnModal
...
+                )}
+            />
+        </View>;
+    }
+}
+
+<Example />
+```
+
+**Warning:** Do not wrap items within an dropdown in a `ModalLauncher`.  It causes the
```
jeresig: "within a dropdown"

It'd also be good to mention what devs should do instead (something like the above example).

packages/wonder-blocks-modal/components/modal-launcher.md
```diff
@@ -152,3 +151,77 @@ const oneColumnModal = ({closeModal}) => <OneColumnModal
...
+    render() {
+        return <View>
+            <ActionMenu menuText="actions">
+                <ActionItem
+                    label="Open modal"
+                    onClick={() => this.handleOpen()}
+                />
+            </ActionMenu>
+            <ModalLauncher
+                key="clearFlags"
```
jeresig: Do we need this ref?

rollup.config.js
```diff
@@ -15,7 +15,7 @@ const createConfig = (pkgName) => ({
             presets,
             plugins,
             exclude: "node_modules/**",
-        }),
```
jeresig: ??? is prettier being run differently somehow?

packages/wonder-blocks-dropdown/util/types.js
kevinbarabash: Not sure why.  This file also changed for J in one of her PRs.

rollup.config.js
kevinbarabash: Not sure.  The extra whitespace was added in https://github.com/Khan/wonder-blocks/commit/c4aa5a633effcf32b86c83a669c182127afec80a.

packages/wonder-blocks-modal/components/modal-launcher.md
kevinbarabash: Nope.  Must've been copy/pasta from somewhere.  I'll remove this.

PR #334, opened 2018-10-04
Fixes WB-457: Fix the dropdown menu for mobile.

jaredly:
https://khanacademy.atlassian.net/browse/WB-457

Pretty easy fix. the bug was that the `touchend` event would happen on
the document, but we weren't checking the menu div (which in rendered
in a portal).
It didn't manifest on desktop, because the "click" event would be
handled by the OptionItem before it got to the document.

Test Plan:
View the wonderblocks documentation with mobile emulation turned on, see
that the <SingleSelect> and <MultiSelect> examples work.


jeresig:

Built with commit f357796532826fb15bc1649d331c2342a18979f1

https://deploy-preview-334--wonder-blocks.netlify.com

jaredly:
The `click` event never happens because we remove the nodes on the `ontouchend` handler

kevinbarabash:
> The click event never happens because we remove the nodes on the ontouchend handler

I see... sneaky.

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -87,6 +87,27 @@ describe("Dropdown", () => {
         expect(dropdown.instance().focusedIndex).toBe(2);
     });
 
+    it("doesn't close on touch interaction with option", () => {
```
kevinbarabash: Good call putting this inside dropdown.test.js instead of the files I was suggesting earlier.

PR #335, opened 2018-10-04
Fixes WB-467 modal should appear above global nav in webapp

kevinbarabash:
The issue was that when we upgrade modals to use React portals we restructured some things resulting in the View that had `zIndex: 1080` set being wrapped in a View with the default `zIndex: 0` which means everything within that part of the tree is `zIndex: 0` b/c `zIndex` within Views only affects siblings instead of the global z-index.  This PR the `zIndex: 1080` up a level to be on a sibling of the global nav in webapp.  I tested this by manually changing the z-index on this div in web app's react-sandbox.


jeresig:

Built with commit 7658371d2f560932996a1ddb9be8bc41619494bd

https://deploy-preview-335--wonder-blocks.netlify.com

kevinbarabash:
> maybe passing it in via <FocusTrap style={styles.modalRoot}> would be one way to make the dependency clearer?

Good call.  I'll make that change.

PR #336, opened 2018-10-04
Fix dropdown not dismissing on click

jaredly:
fixed a bug in #334. There was a test that should have caught this bug, but it is giving us a false-positive 😭 

Unfortunately the "modalHost" I was using was document.body, so checking
if the event was inside the modalHost always returned true! This uses a
ref on the actual popup element, so it works.

Test plan:
desktop & phone:
- open a SingleSelect
- select an item, it should work
- open it again
- tap/click outside of it, and it should dismiss

### DIFF
--- a/packages/wonder-blocks-dropdown/components/dropdown.js
+++ b/packages/wonder-blocks-dropdown/components/dropdown.js
@@ -124,6 +124,7 @@ export default class Dropdown extends React.Component<DropdownProps, State> {
     keyboardNavOn: boolean;
     // Whether any items have been selected since the menu was opened
     itemsClicked: boolean;
+    popperElement: ?HTMLElement;
 
     static defaultProps = {
         alignment: "left",
@@ -290,13 +291,12 @@ export default class Dropdown extends React.Component<DropdownProps, State> {
         const {open, onOpenChanged} = this.props;
         const target: Node = event.target;
         const thisElement = ReactDOM.findDOMNode(this);
-        const modalHost = this.getModalHost();
         if (
             open &&
             thisElement &&
             !thisElement.contains(target) &&
-            modalHost &&
-            !modalHost.contains(target)
+            this.popperElement &&
+            !this.popperElement.contains(target)
         ) {
             onOpenChanged(false);
         }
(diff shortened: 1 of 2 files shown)


jeresig:

Built with commit 77076134c6b2c657da07003fdec8c1096fc66f52

https://deploy-preview-336--wonder-blocks.netlify.com

jaredly:
@jeresig there is a test that would have failed previously, but it's a false-positive, I'm guessing because of breakdowns in event propagation in our `simulate()` function?

kevinbarabash:
@jaredly if it's a false positive we should disable it and add a TODO to migrate it to selenium once we have a test harness set up for that.

kevinbarabash:
I'm going to try to debug the test on `master` where it's broken to see why it's passing when it shouldn't.

jeresig:
@kevinbarabash I'm not sure we should wait for selenium test running to be in place before we have an adjusted test in place. I think it should be possible to generate a test that would've failed previously and now passes - to ensure that we don't revert the changes here and ensure that it continues to work as expected.

kevinbarabash:
I found what the issue was with the previous test.  We were simulating an event directly on `document` which allowed the old logic to succeed even though it failed when clicking on an element within body.  We can add a test that simulates an event on another element within body that's outside of the dropdown and that should result in a failing test on master.  I'll give that a try and then copy the test over to this PR once I have a failure on master.  It should pass on this branch. 🤞

kevinbarabash:
So this is hard b/c:
> Enzyme uses renderIntoDocument from React.TestUtils.
>
> https://github.com/facebook/react/blob/510155e027d56ce3cf5c890c9939d894528cf007/src/test/ReactTestUtils.js#L85
```
{
  renderIntoDocument: function(instance) {
    var div = document.createElement('div');
    // None of our tests actually require attaching the container to the
    // DOM, and doing so creates a mess that we rely on test isolation to
    // clean up, so we're going to stop honoring the name of this method
    // (and probably rename it eventually) if no problems arise.
    // document.documentElement.appendChild(div);
    return ReactDOM.render(instance, div);
  },
// ...
}
```

kevinbarabash:
This is ready for review again.  I was able to get a test working that relies on bubbling by using `ReactDOM.render` directly.

PR #337, opened 2018-10-04
 Fix mobile click delay in clickable-behavior

jaredly:
Test Plan:
using mobile emulation:
- tap the dropdowns, see that they are *much* more responsive (no 300ms
  delay)
- tap the checkboxes, it should also respond quickly!
- same with radios
- and modal buttons

### DIFF
--- a/packages/wonder-blocks-core/components/clickable-behavior.js
+++ b/packages/wonder-blocks-core/components/clickable-behavior.js
@@ -119,8 +119,8 @@ export type ClickableHandlers = {|
     onMouseDown: () => void,
     onMouseUp: (e: SyntheticMouseEvent<>) => void,
     onTouchStart: () => void,
-    onTouchEnd: (e: SyntheticMouseEvent<*>) => void,
-    onTouchCancel: (e: SyntheticMouseEvent<*>) => void,
+    onTouchEnd: () => void,
+    onTouchCancel: () => void,
     onKeyDown: (e: SyntheticKeyboardEvent<*>) => void,
     onKeyUp: (e: SyntheticKeyboardEvent<*>) => void,
     onFocus: (e: SyntheticFocusEvent<*>) => void,
@@ -234,7 +234,7 @@ const startState = {
  * See https://reacttraining.com/react-router/web/guides/basic-components.
  */
 export default class ClickableBehavior extends React.Component<Props, State> {
-    handlingTouch: boolean;
+    waitingForClick: boolean;
     enterClick: boolean;
 
     static defaultProps = {
@@ -255,7 +255,7 @@ export default class ClickableBehavior extends React.Component<Props, State> {
         super(props);
 
         this.state = startState;
-        this.handlingTouch = false;
+        this.waitingForClick = false;
         this.enterClick = false;
     }
 
@@ -263,16 +263,21 @@ export default class ClickableBehavior extends React.Component<Props, State> {
         if (this.enterClick) {
             return;
         } else if (this.props.onClick) {
+            this.waitingForClick = false;
             this.props.onClick(e);
         }
     };
 
     handleMouseEnter = () => {
-        this.setState({hovered: true});
+        if (!this.waitingForClick) {
+            this.setState({hovered: true});
+        }
     };
 
     handleMouseLeave = () => {
-        this.setState({hovered: false, pressed: false});
+        if (!this.waitingForClick) {
+            this.setState({hovered: false, pressed: false});
+        }
     };
 
     handleMouseDown = () => {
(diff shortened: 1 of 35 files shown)


jeresig:

Built with commit 33568123a630c5d6c2996342aa23bd68d797ad3a

https://deploy-preview-337--wonder-blocks.netlify.com

jeresig:

Built with commit be9b1226d654fb85e6f27ce78751f0f9877f3127

https://deploy-preview-337--wonder-blocks.netlify.com

packages/wonder-blocks-core/components/clickable-behavior.test.js
```diff
@@ -337,9 +343,11 @@ describe("ClickableBehavior", () => {
         button.simulate("keyup", {keyCode: keyCodes.enter});
         expect(onClick).toHaveBeenCalledTimes(3);
 
-        button.simulate("touchstart", {keyCode: keyCodes.space});
-        button.simulate("touchend", {keyCode: keyCodes.space});
-        button.simulate("click");
```
jeresig: I'm curious as to why you removed this click? Wouldn't we expect this to still fire?

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -288,16 +283,23 @@ export default class ClickableBehavior extends React.Component<Props, State> {
...
+        this.handlingTouch = true;
     };
 
-    handleTouchEnd = () => {
+    handleTouchEnd = (evt: SyntheticMouseEvent<*>) => {
+        evt.stopPropagation();
+        evt.preventDefault();
         this.setState({pressed: false});
-        this.waitingForClick = true;
+        this.handleClick(evt);
```
jeresig: I think this may mean that we end up calling `handleClick` twice on mobile, correct? It seems like both touchend/touchcancel and click are fired: https://www.html5rocks.com/en/mobile/touchandmouse/

I think your changed test may indicate that issue, as well (since the same number of events are firing even though a click event hasn't been triggered).

packages/wonder-blocks-core/components/clickable-behavior.test.js
jaredly: nope, because `preventDefault()` on touchend means that none of the `mouse` events fire.

packages/wonder-blocks-core/components/clickable-behavior.js
jaredly: see https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Supporting_both_TouchEvent_and_MouseEvent#Event_order

> If the touchstart, touchmove or touchend event is canceled during an interaction, no mouse or click events will be fired, and the resulting sequence of events would just be:
> 
> - touchstart
> - Zero or more touchmove events, depending on movement of the finger(s)
> - touchend

kevinbarabash:
In testing this change inside the iOS simulator I ran into some issues:
- primary and secondary buttons first their onClick callback twice but the tertiary button only fires once
- there seems to be a delay in showing the alert dialog when clicking on buttons
- multiple dropdowns can be open at the same time on both mobile and desktop, this is a regression (open a dropdown should close any other dropdown that's already open)
- tapping anywhere outside a dropdown should dismiss, this is a regression

The last two issues are on master.  I'm guessing they were introduced in https://github.com/Khan/wonder-blocks/pull/334.

jaredly:
gahhh why is mobile safari so weird

jaredly:
Okkk, so I've determined this is the wrong way to do it. See this article: https://developers.google.com/web/updates/2013/12/300ms-tap-delay-gone-away
What we need, is either to have the meta tag: `<meta name="viewport" content="width=device-width">`,
or to use the css attribute `touch-action: manipulation`. I'll do the latter, so that buttons and dropdowns (at least) will be immediately-tappable.

One thing that threw me off is that mobile safari "automatically assumes" that you want to tap buttons, *unless you have zoomed this page before*. So I went to wonderblocks master & everything was fast on ios safari & not on android, until I tried zooming in safari, and then all the click events were delayed.

jaredly:
Ok, this is now ready to go! Rather simpler, by using the `touch-action` css attribute. Safari still has some weird behavior (looks like the first time you tap a button, you get a delay, but subsequent times, you don't 🤷‍♂️ ) but that's probably the best we can get.

kevinbarabash:
> looks like the first time you tap a button, you get a delay

This might actually be a delay in styles being applied.

kevinbarabash:
I think we might also want `touch-action` on `ChoiceInternal` and `OptionItem` as well.

PR #338, opened 2018-10-08
Fix close icon on modals and update toolbar to match updated Toolbar spec

kevinbarabash:
The updated Toolbar spec aligns the 24x24 bounding box of an IconButton's icon to be 16px in from the left side of the toolbar.  This PR updates the Toolbar examples to remove all pixel nudging that was there previously.  It also updates Modal to put the close icon inside the Toolbar when there is a toolbar instead of having it float above the toolbar.  The close icons on modals without a toolbar are positioned to match the position of the close icon on modals that do have a toolbar.

Test Plan:
- go to the Modal section of the netlify link
- click on some of the close buttons to make sure they're still working for both ModalLauncher examples and the other examples
- see that Modals without the toolbar header have the close buttons with the same top padding as left padding


packages/wonder-blocks-modal/components/standard-modal.js
```diff
@@ -58,6 +61,32 @@ type Props = {|
  * The "standard" modal layout: a titlebar, a content area, and a footer.
  */
 export default class StandardModal extends React.Component<Props> {
+    renderCloseButton() {
```
jeresig: Why is this duplicated from ModalPanel? Shouldn't we just handle this rendering in one place? And if we're doing it here why aren't we also doing it in OneColumnModal and TwoColumnModal?

kevinbarabash: Whoops, that's a pretty big oversight on my part.  You are correct, we need to do this in the other modals as well.  Thanks for catching this.

kevinbarabash: I double checked an we aren't using Toolbar for either `OneCloumnModal` or `TwoColumnModal` so this should be fine as is.

kevinbarabash: We should also be using Toolbar for the footer, but I'll do that as a separate PR.

kevinbarabash:
> The duplication of the logic here between ModalPanel and StandardModal isn't great - they're both rendering close buttons and in very similar ways. Would be good to, at least, add a comment explaining how these things could be merged, potentially.

I can extract this out into a `CloseButton` component that's only used in `wonder-blocks-modal`.

PR #339, opened 2018-10-08
fix issue with UniqueIdProvider inside NoSSR

kevinbarabash:
I ran into a situation where having a `NoSSR` wrapping a `UniqueIdProvider` resulted in `this._performRender(true)` not getting called which meant that `this._idFactory` never gets initialized.  This PR fixes that by always check if `this._idFactory` is initialized before calling it and initializing it if it hasn't already been done.

### DIFF
--- a/packages/wonder-blocks-core/components/unique-id-provider.js
+++ b/packages/wonder-blocks-core/components/unique-id-provider.js
@@ -66,16 +66,20 @@ export default class UniqueIDProvider extends React.Component<Props> {
         const {children, mockOnFirstRender, scope} = this.props;
 
         // If this is our first render, we're going to stop right here.
+        // Note: `firstRender` will be `false` on the first render if this
+        // component is a descendant of a `NoSSR`.
         if (firstRender) {
-            // We'll be needing this on the next render, so let's set it up.
-            this._idFactory = new UniqueIDFactory(scope);
-
             if (mockOnFirstRender) {
                 // We're allowing an initial render, so let's pass our mock
                 // identifier factory to support SSR.
                 return children(SsrIDFactory);
             }
             return null;
+        }
+
+        // Create an identifier factory if we don't already have one
+        if (!this._idFactory) {
+            this._idFactory = new UniqueIDFactory(scope);
         }
 
         // It's a regular render, so let's use our identifier factory.
--- a/packages/wonder-blocks-core/components/unique-id-provider.test.js
+++ b/packages/wonder-blocks-core/components/unique-id-provider.test.js
@@ -9,6 +9,7 @@ import View from "./view.js";
 import SsrIDFactory from "../util/ssr-id-factory.js";
 import UniqueIDFactory from "../util/unique-id-factory.js";
 import UniqueIDProvider from "./unique-id-provider.js";
+import NoSSR from "./no-ssr.js";
 
 describe("UniqueIDProvider", () => {
     beforeEach(() => {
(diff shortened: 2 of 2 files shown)


jeresig:

Built with commit 28cfab5ac501eeb94ed67a1fc846216c37f15aa5

https://deploy-preview-339--wonder-blocks.netlify.com

packages/wonder-blocks-core/components/unique-id-provider.js
```diff
@@ -66,10 +66,9 @@ export default class UniqueIDProvider extends React.Component<Props> {
         const {children, mockOnFirstRender, scope} = this.props;
 
         // If this is our first render, we're going to stop right here.
+        // Note: `firstRender` will be `false` on the first render if this
```
somewhatabstract: Should we just rename this variable? Suggestions:

- `initialSsrLevelRender`
- `placeholderRender`
- something similar

packages/wonder-blocks-core/components/unique-id-provider.js
```diff
@@ -66,10 +66,9 @@ export default class UniqueIDProvider extends React.Component<Props> {
...
+        // component is a descendant of a `NoSSR`.
         if (firstRender) {
-            // We'll be needing this on the next render, so let's set it up.
-            this._idFactory = new UniqueIDFactory(scope);
-
             if (mockOnFirstRender) {
```
somewhatabstract: I never liked this prop name; just went with it since it is how the ADR talked about it and is consistent in like components. Should really be `mockIfInitialSSRCompatibleRender` or something.

packages/wonder-blocks-core/components/unique-id-provider.js
```diff
@@ -78,6 +77,11 @@ export default class UniqueIDProvider extends React.Component<Props> {
             return null;
         }
 
+        // Create an identifier factory if we don't already have one
+        if (!this._idFactory) {
+            this._idFactory = new UniqueIDFactory(scope);
```
somewhatabstract: ASIDE: Should we consider a future update to the provider that allows for `scope` prop to change, and recreates the factory accordingly?

packages/wonder-blocks-core/components/unique-id-provider.test.js
```diff
@@ -124,4 +125,28 @@ describe("UniqueIDProvider", () => {
...
+                        <UniqueIDProvider>
+                            {(ids) => foo(ids.get(""))}
+                        </UniqueIDProvider>
+                    )}
+                </NoSSR>
+            );
+            const wrapper = mount(nodes);
+
+            // Act
+            wrapper.instance().forceUpdate();
```
somewhatabstract: Should we verify that `foo` is not called before the `forceUpdate` just to ensure that we're testing what we think we are?

packages/wonder-blocks-core/components/unique-id-provider.test.js
```diff
@@ -124,4 +125,28 @@ describe("UniqueIDProvider", () => {
...
+
+            // Assert
+            expect(foo).toHaveBeenCalled();
+            expect(foo.mock.calls[0][0]).toBeTruthy();
+        });
+    });
```
somewhatabstract: Great test coverage! Thank you for maintaining the "Arrange/Act/Assert" pattern.

packages/wonder-blocks-core/components/unique-id-provider.test.js
kevinbarabash: I can add that check.

packages/wonder-blocks-core/components/unique-id-provider.test.js
kevinbarabash: Thanks for starting the pattern even if I was resistant at first.

packages/wonder-blocks-core/components/unique-id-provider.test.js
somewhatabstract: No, I think you can drop the `forceUpdate` and change the `// Act` to just be `mount(nodes);`.

Since the provider is wrapped in the `NoSSR` component, my guess is that the `setState` in the `NoSSR` component `componentDidMount` is making the second render occur, and causing the provider to do its render.

kevinbarabash: I'm going to remove this from the test since it isn't necessary.

packages/wonder-blocks-core/components/unique-id-provider.js
kevinbarabash: That makes sense to me although I can't think of a situation where that would be useful.

somewhatabstract: Fair; if we don't support it, we should probably call it out in the docs.

packages/wonder-blocks-core/components/unique-id-provider.js
kevinbarabash: Maybe `ssrRender`?

PR #341, opened 2018-10-12
output ES5 code so that we can run the phantomjs tests in webapp

kevinbarabash:
This will also avoid having to update UglifyJS which also can't handle ES6 code.  Upgrading both the phantomjs test runner and UglifyJS as part of the current build system aren't good uses of time given that we'll soon be replacing our current build system and test runner.

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -5,3 +5,4 @@ last 1 Edge version
 last 1 Firefox version
 last 1 iOS version
 last 1 Safari version
+IE 10
--- a/babel.config.js
+++ b/babel.config.js
@@ -6,4 +6,20 @@ module.exports = {
         "@babel/plugin-proposal-export-default-from",
         "babel-plugin-dynamic-import-node",
     ],
+    env: {
+        test: {
+            presets: [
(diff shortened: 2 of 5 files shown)


jeresig:

Built with commit da082d75b971a39e58300899c153df950d29d3df

https://deploy-preview-341--wonder-blocks.netlify.com

somewhatabstract:
Which change in here actually changes the output to ES5? I don't see it.

.browserslistrc
```diff
@@ -5,3 +5,4 @@ last 1 Edge version
 last 1 Firefox version
 last 1 iOS version
 last 1 Safari version
+IE 10
```
kevinbarabash: @somewhatabstract this is the change that forces ES5 output.  `@babel/preset-env` looks at this config file and uses it to decide which syntax constructs to transform or not.  The old version of this list transformed barely anything, adding IE 10 should guarantee that we'll be able to also run in phantomjs.  I also checked the output in the dist/ folders for any uses of `class` or `=>` and there weren't any in the output code.

PR #342, opened 2018-10-15
create a build-settings package that all other packages depend on

kevinbarabash:
The reason for this is that we want lerna to detect any changes to our build settings and bump the versions of our packages when that happens since the compiled code is different.

It's going to be hard to test this for real b/c I'll need another commit which just changes the `build-settings` package and none of the other packages.  With this PR we've made changes to all packages by either their package.json files which `lerna` will detect allowing us to publish new versions of everything.


build-settings/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "build-settings",
+  "version": "0.0.1",
+  "main": "index.js",
+  "license": "MIT",
+  "private": true
```
jeresig: This isn't going to be published? Do we not having to prefix this with `@khanacademy`? What happens when we publish the Wonder Blocks packages that depend on this?

build-settings/rollup.config.js
```diff
@@ -1,8 +1,9 @@
+/* eslint-disable import/no-commonjs */
 import fs from "fs";
 import autoExternal from "rollup-plugin-auto-external";
 import babel from "rollup-plugin-babel";
 
-const {presets, plugins} = require("./babel.config.js");
+const {presets, plugins} = require("../babel.config.js");
```
jeresig: Why wasn't the babel config moved too?

kevinbarabash: I was worried about it no longer being visible to those things need, but maybe that concern is unfounded since the the things that need are already in this folder.

build-settings/package.json
kevinbarabash: When we go to install one of our `wonder-blocks-*` packages somewhere, this should be ignored b/c we've marked it as a dev dependency.  The `@khanacademy` prefix isn't necessary but we could add it if we wanted.  Since it's not a package we intend on publishing I opted not to include prefix to better differentiate it from those packages we are publishing.

jeresig: In this world if I `npm install wonder-blocks-foo` won't it get confused as it doesn't know about `build-settings`? Wouldn't it try to install it directly from NPM?

jeresig: Hmm - ok! I think a better name would be something like `wonder-blocks-build-settings` or some such, just to make it clear that this isn't some third-party package.

kevinbarabash:
Looks like I need to get the docs building too.

build-settings/webpack.config.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/no-commonjs */
```
somewhatabstract: Thank you for adding these!

build-settings/package.json
somewhatabstract: How do you folks feel about using a naming convention that encapsulates the internal, dev-only nature of this style of package, making that more obvious?

`wb-dev-build-settings` or just `wb-dev-build`?

For some reason, it makes me a little uneasy how similar to our prod packages this looks.

kevinbarabash: I'll change into `wb-dev-build`.

build-settings/webpack.config.js
kevinbarabash: I think eventually we'll want want to have two different eslint configs, one for node and one for browser.  We should be able to configure eslint to use different configs based on a path regex.

PR #343, opened 2018-10-21
initial setup of vite test runner

kevinbarabash:
This runs some test tests, but this PR will eventually be updated to run tests on wonder-blocks components.


jeresig:

Built with commit 12a9bdea1f560a52ade9d4581355056d8ad059f7

https://deploy-preview-343--wonder-blocks.netlify.com

.eslintignore
```diff
@@ -8,3 +8,5 @@ packages/wonder-blocks-core/util/add-style.js
 
 # it probably passes, but we really don't care
 styleguide.prod.config.js
+
+config
```
kevinbarabash: Ideally we'd used separate rules for folders containing node scripts.

.circleci/config.yml
```diff
@@ -33,11 +33,29 @@ jobs:
...
+          keys:
+            - npm-deps-{{ checksum "yarn.lock" }}
+            - npm-deps-
+      - run: yarn install
+      - save_cache:
+          key: npm-deps-{{ checksum "yarn.lock" }}
+          paths:
+            - node_modules
+      - run: yarn run build:cjs
+      - run: yarn run build:es6
```
kevinbarabash: We have to build both b/c the node runner still needs to be able to import `@khanacademy/wonder-blocks-foo` without having to compile which is why we're building the commonjs modules.  The es6 modules are being built because it's easier for `vite-server` to serve those modules to whatever browser selenium is controlling.

config/jest/test.config.js
```diff
@@ -0,0 +1,22 @@
...
+    },
+    testEnvironment: "jest-environment-jsdom",
+    globals: {
+        SNAPSHOT_INLINE_APHRODITE: true,
+    },
+    testMatch: ["<rootDir>/packages/**/*.test.js"],
+    setupTestFrameworkScriptFile: "<rootDir>/config/jest/test-setup.js",
+    moduleNameMapper: {
+        "^@khanacademy/(.*)$":
+            "<rootDir>/node_modules/@khanacademy/$1/index.js",
```
kevinbarabash: This change is what's causing the increase in test coverage.  It's cause the normal `jest` tests to import the source `index.js` file instead of `dist/index.js` which we aren't collecting coverage on.

lerna.json
```diff
@@ -1,5 +1,4 @@
 {
-  "lerna": "2.11.0",
```
kevinbarabash: I don't know why I made this change.  I'll revert this but we should maybe update our version of lerna which no longer requires this.

package.json
```diff
@@ -16,7 +16,8 @@
     "flow-coverage": "flow-coverage-report -t text -t html -f ./node_modules/.bin/flow -i 'packages/wonder-blocks-*/**/*.js' -x 'packages/wonder-blocks-*/dist/*.js' -x '**/*.test.js' -x '**/node_modules/**/*.js'",
     "gen-snapshot-tests": "node utils/gen-snapshot-tests.js && prettier --write packages/*/generated-snapshot.test.js",
     "gen-styleguide-prod-config": "node utils/gen-styleguide-prod-config.js && prettier --write styleguide.prod.config.js",
-    "jest": "jest",
+    "jest": "jest --config config/jest/test.config.js",
+    "test:vite": "jest --config config/jest/vite.config.js",
```
kevinbarabash: `jest` has a `packages` option which allows you to specify multiple configurations in a single file, but I couldn't get it to work so this is what I ended up with.

package.json
```diff
@@ -37,10 +38,14 @@
     "@babel/plugin-proposal-class-properties": "7.0.0",
     "@babel/plugin-proposal-export-default-from": "^7.0.0",
     "@babel/plugin-proposal-object-rest-spread": "7.0.0",
+    "@babel/plugin-syntax-dynamic-import": "^7.0.0",
```
kevinbarabash: This is required by the tests which do `await import()`.

package.json
```diff
@@ -91,28 +96,12 @@
     "react": "^16.4.3",
     "react-dom": "^16.4.3",
     "react-popper": "^1.0.0",
-    "react-router-dom": "^4.2.2"
+    "react-router-dom": "^4.3.1"
```
kevinbarabash: This change isn't strictly necessary.  I'll revert it.

vite-tests/button.test.js
```diff
@@ -0,0 +1,14 @@
...
+import Button from "../packages/wonder-blocks-button/components/button.js";
+
+describe("foo", () => {
+    test("it should render a button", async () => {
+        // $FlowFixMe
+        const element = await render(<Button>Hello, world!</Button>);
+        const text = await element.getText();
+        expect(text).toBe("Hello, world!");
+    });
+});
```
kevinbarabash: I'd like these test files to live in the same hierarchy as the source files.  In order to differentiate them from other test files I was thinking of having a structure like:
```
- wonder-blocks-button
-- components
--- __test__ // contains regular jest tests
---- button.test.js
--- __vite__ // contains vite tests
---- button.test.js
--- button.js
--- button-core.js
```
another option is
```
- wonder-blocks-button
-- components
--- __test__
---- button.test.js
---- button.vite_test.js // or just button.vite.js
--- button.js
--- button-core.js
```

config/jest/test.config.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+    rootDir: "../../",
+    transform: {
+        "^.+\\.jsx?$": "<rootDir>/config/jest/test.transform.js",
+    },
+    testEnvironment: "jest-environment-jsdom",
+    globals: {
+        SNAPSHOT_INLINE_APHRODITE: true,
+    },
+    testMatch: ["<rootDir>/packages/**/*.test.js"],
```
kevinbarabash: I find globs easier to grok than regexes for this kind of stuff.

kevinbarabash:
As part of this PR I also created a `config/jest` folder.  My hope was to move `build-settings` into the `config` folder resulting in the following structure:
```
- config
-- build
-- test
```
I think it'd be better to move `build-settings` to `config/build` in a separate PR though since there's already a lot going on this one.

config/jest/vite.config.js
```diff
@@ -0,0 +1,33 @@
...
+        "^.+\\.jsx?$": "<rootDir>/config/jest/vite.transform.js",
+    },
+    testMatch: ["<rootDir>/vite-tests/*.test.js"],
+    testEnvironment: "@khanacademy/jest-environment-vite",
+    testEnvironmentOptions: {
+        capabilities: {
+            browserName: "chrome",
+            chromeOptions: {
+                args: ["headless", "disable-gpu"],
+            },
```
kevinbarabash: We'll want to make this configurable so that we can both run these tests on more browser on circleci as well as run them in a non-headless browser locally which is useful for debugging things.  Ideally though I'd like to figure out how to forward console.log, console.error, and exception stack traces back from the browser back to jest and report the errors there.

config/jest/test.config.js
somewhatabstract: Good. I wanted to add test coverage of the regular `index.js` (and did for some of our packages), but adding it for all is great. Of course, we should have unit tests for them to verify the public API is what we expect, as well as the vite things, but baby steps.

config/jest/test.config.js
somewhatabstract: ++

kevinbarabash: Good catch.  I'll fix that.

config/jest/test.config.js
```diff
@@ -0,0 +1,25 @@
...
+    transform: {
+        "^.+\\.jsx?$": "<rootDir>/config/jest/test.transform.js",
+    },
+    testEnvironment: "jest-environment-jsdom",
+    globals: {
+        SNAPSHOT_INLINE_APHRODITE: true,
+    },
+    testMatch: [
+        "<rootDir>/packages/**/*.test.js",
+        "<rootDir>/shared-unpackaged/*.test.js",
```
somewhatabstract: This is good, but does it pass the test of "least surprise"? Like if we added tests for some of our tooling or node scripts, would we expect them to just run? I am on the fence about it, but I did prefer when we were guaranteed to pick up all tests in the entire repo rather than each location be opt-in.

kevinbarabash: I can switch it back to a regex or we can do `"<rootDir>/**/*.test.js"` and I can change the `vite` tests to have a different pattern.

somewhatabstract: I think we probably should change it. I am all for the vite tests having a different pattern anyway because I am in love with my `.vist.js` suggestion. 😂

config/jest/vite.transform.js
```diff
@@ -0,0 +1,15 @@
+const babelJest = require("babel-jest");
```
jeresig: Can you add some details here explaining why Vite has a different babel config?

config/jest/vite.config.js
```diff
@@ -0,0 +1,33 @@
+module.exports = {
```
jeresig: I found this to be a bit confusing as it's not apparent at first that vite is run by jest - adding a comment here explaining what vite is and how it's being run would be helpful.

packages/wonder-blocks-button/components/button.vite_test.js
```diff
@@ -0,0 +1,19 @@
+/* global render */
```
jeresig: That seems annoying - is there a way in which you can make that "just work" like it does with describe/test/etc. coming from jest?

packages/wonder-blocks-button/components/button.vite_test.js
```diff
@@ -0,0 +1,19 @@
...
+// @flow
+import * as React from "react";
+
+import Button from "./button.js";
+
+describe("foo", () => {
+    test("it should render a button", async () => {
+        // Arrange
+        // $FlowFixMe
```
jeresig: Why is this FlowFixMe?

packages/wonder-blocks-button/components/button.vite_test.js
kevinbarabash: That's on my list of things to do.

packages/wonder-blocks-button/components/button.vite_test.js
kevinbarabash: I'll add a note.

.circleci/config.yml
```diff
@@ -33,11 +33,29 @@ jobs:
       - run: yarn run gen-snapshot-tests
       - run: yarn run build:cjs
       - run: yarn run build:es6
-      - run: yarn run jest --coverage
+      - run: yarn run jest --coverage --runInBand
```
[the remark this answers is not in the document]
kevinbarabash: It runs in a single thread.  Without it we were running into ENOMEM errors on CircleCI.

packages/wonder-blocks-button/components/button.vite_test.js
kevinbarabash: The problem is that flow can resolve this global.  I hope that we can make it work the same way other jest globals do.

packages/wonder-blocks-button/components/button.vite_test.js
kevinbarabash: I think we'll have to make our own custom eslint environment.  If `vite` gets popular enough then maybe it will become an official eslint environment like `jest`.  Here's the docs on how to do this: https://eslint.org/docs/developer-guide/working-with-plugins#environments-in-plugins

PR #344, opened 2018-10-22
make Icon fixed size, fix truncation behavior in ActionMenu opener

kevinbarabash:
This PR fixes a couple of issues with the ActionMenu opener:
- text wasn't being truncated
- icon was be squished

Icon is now fixed size.  It can't shrink or grow.

Before:
screen shot 2018-10-19 at 9 30 50 am

After:
screen shot 2018-10-22 at 12 15 45 am

Unfortunately we don't have an easy way to test changes in webapp.  We will once the webpack work is done, but I have high confidence that this will fix the issue.  In the before screenshot notice how the down caret is smaller than it should be.  I was able to reproduce this in one of the ActionMenu examples before fixing it with by setting `flexShrink: 0` on the Icon component.  I'll try copy/pasting the built component into webapp to verify things are working as expected.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ jobs:
       - run: yarn run gen-snapshot-tests
       - run: yarn run build:cjs
       - run: yarn run build:es6
-      - run: yarn run jest --coverage
+      - run: yarn run jest --coverage --runInBand
       - run: yarn run codecov
 workflows:
   version: 2
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "test:coverage": "yarn run test:common && yarn run jest --coverage",
     "coverage": "yarn run test:coverage && open coverage/lcov-report/index.html",
     "styleguidist": "styleguidist server",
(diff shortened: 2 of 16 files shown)


kevinbarabash:
There's still an issue with the caret getting squished in webapp.
screen shot 2018-10-23 at 12 15 32 pm

It should look like:
screen shot 2018-10-23 at 12 16 06 pm

Thankfully the fix is really easy, we just need to change the `padding-left: 4px` in between the text and the caret to be `margin-left: 4px`.

jaredly:
image
focus state looks a bit awkward on mobile though 😕

kevinbarabash:
I'm still unable to reproduce this bug locally within wonder-blocks, but I was able to fix it and verify that it's fixed by copying the built file for wonder-blocks-dropdown over to webapp and updating the imports appropriately.

kevinbarabash:
@jaredly thanks for testing on mobile.  I believe there's a special CSS property to control that highlight on mobile.  I'll open a ticket for that and tackle it separately.

PR #348, opened 2018-11-01
ignore more files to avoid unnecessary upgrades while publishing

kevinbarabash:
In order to avoid bumping all modules all the time, I've added some more files to ignore to our lerna.json file.

Test Plan:
- run `node_modules/.bin/lerna updated`
- see the following output

```
lerna info version 2.11.0
lerna info versioning independent
lerna info ignore [ '*.test.js', '*.vite_test.js', '*.md', '*.js.snap' ]
lerna info Checking for updated packages...
lerna info Comparing with @khanacademy/wonder-blocks-button@2.2.5.
lerna info Checking for prereleased packages...
lerna info result 
- @khanacademy/wonder-blocks-dropdown
```

In this case `wonder-blocks-dropdown` has actually seen some code changes since the last publish.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -7,7 +7,9 @@
     "publish": {
       "ignore": [
         "*.test.js",
-        "*.md"
+        "*.vite_test.js",
+        "*.md",
+        "*.js.snap"
       ]
     }
   },


lerna.json
```diff
@@ -7,7 +7,9 @@
     "publish": {
       "ignore": [
         "*.test.js",
-        "*.md"
+        "*.vite_test.js",
+        "*.md",
```
somewhatabstract: Not something to hold up this diff; just musing.

Wouldn't it be really cool if when a release is cut, a version of the docs relating to that specific release were put up so that it doesn't change as we merge more things to master?

kevinbarabash: what we really need is versioned docs... I was thinking of submitting a PR to docusaurus to versioning of independent packages within a single documentation project.  I asked them about it and they're interested in it.  I think that's something we should tackle as part of the frontend infra team.

kevinbarabash: We use docusaurus for the KaTeX docs and versioning single packages works well already.

PR #352, opened 2018-11-02
WB-474: allow dropdowns to work without children

kevinbarabash:
It seems that ActionMenu and SingleSelect already worked without any children.  This PR adds examples to show that they work and update MultiSelect to ensure that it works.  I've also updated the flow types for all of the dropdowns to allow code to type check when not passing children.

TODO: move the examples closer to the exiting disabled examples.


jeresig:

Built with commit 713a2891c81b77de98e4522f61dbdcca08ebddb2

https://deploy-preview-352--wonder-blocks.netlify.com

jangmi-jo:
I wonder if we could only allow no children when the select is disabled or make the select disabled when there's no children.

kevinbarabash:
@jangmi-jo I think automatically disabling it when there's no children probably makes sense.  I'll make that change.

PR #353, opened 2018-11-03
Test(modal) - Add FocusTrap test, fix vscode-jest setup

kevinbarabash:
While this does test the FocusTrap, there are a number of limitations that will require some changes to `vite` to fix.  In particular:
- `import "@khanacademy/wonder-blocks-button"` isn't work hence the relative import of button
- coverage numbers are totally wrong for some reason (need to investigate)

That being said, at least we have some tests now for the FocusTrap.

As I was working on this change I noticed that all of the other tests were showing up as failing inside of VSCode.  After reading a bit, it appears that vscode-jest doesn't support configurations in external files so I had to move the settings into package.json.


jeresig:

Built with commit b8eeae940d6c1b25d01ace4b361dfd63e89fe6b9

https://deploy-preview-353--wonder-blocks.netlify.com

kevinbarabash:
For some reason coverage appears to be working on circleci even if it isn't working locally.
screen shot 2018-11-03 at 7 12 21 pm

packages/wonder-blocks-modal/components/modal.vite_test.js
```diff
@@ -0,0 +1,101 @@
+/* global render, driver */
+// @flow
+import * as React from "react";
+
+// TODO(kevinb): fix vite so that it can handle importing modules
```
somewhatabstract: Is there an issue number/link to include in this TODO?

packages/wonder-blocks-modal/components/modal.vite_test.js
```diff
@@ -0,0 +1,101 @@
...
+                    />
+                }
+                opened={true}
+                onClose={() => {}}
+            />,
+        );
+
+        // Act
+        // $FlowFixMe: flow can't resolve driver
+        const element = await driver.switchTo().activeElement();
```
somewhatabstract: To help out flow, is there a way `render` and `driver` can be imported explicitly in the test file from `vite` like we do `enzyme` methods?

somewhatabstract: And in that vein, is there an issue number for fixing this that we can tag these `$FlowFixMe`'s with?

packages/wonder-blocks-modal/components/modal.vite_test.js
```diff
@@ -0,0 +1,101 @@
...
+
+        // Act
+        // $FlowFixMe: flow can't resolve driver
+        await driver
+            .actions()
+            .sendKeys(TAB)
+            .perform();
+
+        // Assert
+        const element = await driver.switchTo().activeElement();
```
somewhatabstract: Should this be in the `//Act` section?

packages/wonder-blocks-modal/util/needs-hacky-mobile-safari-scroll-disabler.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+
+const needsHackyMobileSafariScrollDisabler = (() => {
+    const userAgent = window.navigator.userAgent;
+    return userAgent.indexOf("iPad") > -1 || userAgent.indexOf("iPhone") > -1;
+})();
+
+export default needsHackyMobileSafariScrollDisabler;
```
somewhatabstract: This needs tests.

packages/wonder-blocks-modal/components/modal.vite_test.js
kevinbarabash: It's hard to import these things explicitly b/c there's some setup involved in creating these objects and functions that are being injected into the environment.  Since jest is multi-threaded it runs the environment code when each thread starts up.  I'm working on this in https://github.com/Khan/vite/pull/21.  I think we should probably just fix this the issue with globals in this PR.

packages/wonder-blocks-modal/components/modal.vite_test.js
kevinbarabash: I'll create one.

packages/wonder-blocks-modal/components/modal.vite_test.js
somewhatabstract: Yeah, shame there isn't some way of having typed placeholders that can be imported but that themselves call the generated functions. I don't now the vite stuff to know if that makes sense or not.

kevinbarabash:
Give that I've disabled `vite` tests in #375 I'm going to close this PR.  I'd like to experiment with BrowserStack for selenium testing in the future and which point we can hopefully try re-adding these tests.

PR #355, opened 2018-11-05
WB-407: prevent spinner's bounding box from change as it rotates

kevinbarabash:
This diff rotates the path inside the svg.  I removed the caching but I can put it back if necessary.


jeresig:

Built with commit 3850b0a3804aa5bc2543c8463b4e295427195128

https://deploy-preview-355--wonder-blocks.netlify.com

PR #356, opened 2018-11-05
WB-489: prevent navigation when using href and disabled props together

kevinbarabash:
The issue is that when using href we end up rendering either an `<a>` or a ` <Link>`.  With `<a>` you can disabled it not giving an `href`, but with `<Link>` its `to` prop is a required.  This PR instead renders a `<button>` whenever the Button, IconButton or ActionItem is disabled and sets `disabled={true}`.  This results in the component being un-clickable when disabled regardless of whether it uses `href` or not.

### DIFF
--- a/packages/wonder-blocks-button/__snapshots__/generated-snapshot.test.js.snap
+++ b/packages/wonder-blocks-button/__snapshots__/generated-snapshot.test.js.snap
@@ -1822,11 +1822,11 @@ exports[`wonder-blocks-button example 8 1`] = `
     }
   }
 >
-  <a
+  <button
     aria-disabled="true"
     aria-label="loading"
     className=""
-    href="/foo"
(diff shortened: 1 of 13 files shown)


jeresig:

Built with commit 40345e1e69179bd4709b7a79fca760a61ba992ea

https://deploy-preview-356--wonder-blocks.netlify.com

kevinbarabash:
@chenyejia good catch.  I'll fix that before merging.

somewhatabstract:
LGTM although is there a missing unit test for this behavior? Feels like there should be a test for this, not just a snapshot.

kevinbarabash:
> Feels like there should be a test for this, not just a snapshot.

Good point.  I'll see if I can write a failing test first.  If I can't access window.location I can try using react-router.

PR #357, opened 2018-11-07
WB-491: update dropdown components to have appropriate aria roles

kevinbarabash:
I'm not sure if I'm using the right roles for the right things.  The openers no-longer have any role.  They use a `<button>` under the hood whose role is `button` by default which I think is okay.  `Dropdown` is a pretty general component so in order to make people think about what role they want I've made `role` mandatory.

### DIFF
--- a/packages/wonder-blocks-dropdown/components/action-item.js
+++ b/packages/wonder-blocks-dropdown/components/action-item.js
@@ -70,6 +70,11 @@ type ActionProps = {|
      * @ignore
      */
     indent: boolean,
+
+    /**
+     * Aria role to use, defaults to "menuitem".
+     */
+    role: "menuitem" | "option",
 |};
 
 const StyledAnchor = addStyle("a");
@@ -85,6 +90,7 @@ export default class ActionItem extends React.Component<ActionProps> {
     static defaultProps = {
         disabled: false,
         indent: false,
+        role: "menuitem",
     };
 
     static contextTypes = {router: PropTypes.any};
@@ -97,6 +103,7 @@ export default class ActionItem extends React.Component<ActionProps> {
             indent,
             label,
             onClick,
+            role,
             testId,
         } = this.props;
         const {router} = this.context;
@@ -112,7 +119,7 @@ export default class ActionItem extends React.Component<ActionProps> {
                 disabled={disabled}
                 onClick={onClick}
                 href={href}
-                role="menuitem"
+                role={role}
             >
                 {(state, handlers) => {
                     const {pressed, hovered, focused} = state;
@@ -129,6 +136,7 @@ export default class ActionItem extends React.Component<ActionProps> {
                     const props = {
                         "data-test-id": testId,
                         disabled,
+                        role,
                         style: [defaultStyle],
                         ...handlers,
                     };
--- a/packages/wonder-blocks-dropdown/components/action-menu-opener-core.js
+++ b/packages/wonder-blocks-dropdown/components/action-menu-opener-core.js
@@ -56,21 +56,22 @@ export default class ActionMenuOpenerCore extends React.Component<Props> {
             !disabled && pressed && buttonStyles.active,
         ];
 
-        const commonProps = {
-            "aria-disabled": disabled ? "true" : undefined,
-            "aria-label": ariaLabel,
-            "data-test-id": testId,
-            role: "button",
-            style: [defaultStyle],
-            ...handlers,
-        };
-
         const label = (
             <LabelLarge style={sharedStyles.text}>{children}</LabelLarge>
         );
 
         return (
-            <StyledButton type="button" {...commonProps} disabled={disabled}>
+            <StyledButton
+                aria-disabled={disabled ? "true" : undefined}
+                aria-expanded={open ? "true" : "false"}
+                aria-haspopup="menu"
+                aria-label={ariaLabel}
+                data-test-id={testId}
+                disabled={disabled}
+                style={defaultStyle}
+                type="button"
+                {...handlers}
+            >
                 <View
                     style={
                         !disabled && (hovered || focused) && buttonStyles.focus
--- a/packages/wonder-blocks-dropdown/components/dropdown.js
+++ b/packages/wonder-blocks-dropdown/components/dropdown.js
@@ -83,6 +83,11 @@ type DropdownProps = {|
      * Optional styling for the entire dropdown component.
      */
     style?: StyleType,
+
+    /**
+     * The aria "role" applied to the dropdown container.
+     */
+    role: "listbox" | "menu",
 |};
 
 type State = {|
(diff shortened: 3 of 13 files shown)


jeresig:

Built with commit 11f3edb67f5d410a7ffbb2c262c8bd7196ce4c61

https://deploy-preview-357--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -450,12 +455,23 @@ export default class Dropdown extends React.Component<DropdownProps, State> {
...
         let focusCounter = 0;
+        let itemRole;
+
+        switch (this.props.role) {
+            case "listbox":
+                itemRole = "option";
+                break;
+            case "menu":
+                itemRole = "menuitem";
+                break;
```
kevinbarabash: There's a hybrid action menu in https://deploy-preview-357--wonder-blocks.netlify.com/#!/ActionMenu which includes menu items and options.  I wonder if it makes sense to embed the role in the ActionItem and OptionItem instead of basing it on the container's `role`.

kevinbarabash: I checked the code at that's what we're doing.  I'm going to get rid of this.

kevinbarabash: I chatted with @somewhatabstract on Slack and it sounds like what we have here is actually what we want so I'm going to leave it.  I just need to pass this down to `ActionItem` and `OptionItem`.

dierat:
Moving the `role="listbox"` to the container of `role="option"` action items looks right to me. I think the only things that it's missing is
- `aria-haspopup="listbox"` on the button that opens the listbox
- `aria-expanded="true"` on the button that opens the listbox when the listbox is open
- an id on the label that describes the dropdown
- `aria-labelledby=` + the id of the label, placed on both the button that opens the listbox and the listbox itself
- `aria-selected="true"` on the currently selected option

They also suggest having `aria-activedescendant=` + the id of the currently focused option, placed on the listbox. I've never heard of that and I don't think we've ever been called out by LevelAccess for not having it. Do you know anything about that @somewhatabstract ?

For reference for the aria attributes, I'm looking at this page, which has a working example and a table showing the recommended roles and attributes for each element https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html

packages/wonder-blocks-dropdown/components/action-item.js
```diff
@@ -70,6 +70,11 @@ type ActionProps = {|
      * @ignore
      */
     indent: boolean,
+
+    /**
+     * Aria role to use, defaults to "menuitem".
+     */
+    role: "menuitem" | "option",
```
somewhatabstract: Should we have a type like `DropdownItemRole` or something? This seems to be repeated often.

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -450,12 +455,23 @@ export default class Dropdown extends React.Component<DropdownProps, State> {
             : 0;
 
         let focusCounter = 0;
+        let itemRole;
```
somewhatabstract: NIT: This could be immutable (`const`) if the switch were in a `getRole` method.

packages/wonder-blocks-dropdown/components/multi-select.test.js
```diff
@@ -56,10 +56,10 @@ describe("MultiSelect", () => {
         opener.simulate("keyup", {keyCode: keyCodes.space});
         expect(select.state("open")).toEqual(false);
 
-        // Shouldn't open with enter
+        // Should open with enter
```
somewhatabstract: Interesting that this test case has flipped. So it was by design that enter didn't work before?

packages/wonder-blocks-dropdown/components/multi-select.test.js
```diff
@@ -56,10 +56,10 @@ describe("MultiSelect", () => {
...
         opener.simulate("keydown", {keyCode: keyCodes.enter});
         opener.simulate("keyup", {keyCode: keyCodes.enter});
-        expect(select.state("open")).toEqual(false);
+        expect(select.state("open")).toEqual(true);
```
somewhatabstract: This will be an exciting test to have in vite too.

packages/wonder-blocks-dropdown/components/multi-select.test.js
kevinbarabash: We want the opener to behave like a button and it wasn't because we were following the listbox behaviour.

packages/wonder-blocks-dropdown/components/multi-select.test.js
kevinbarabash: Are you thinking screenshot tests for this?

packages/wonder-blocks-dropdown/components/action-item.js
kevinbarabash: I'm going to defer this for now since our documentation tool doesn't know how to expand type aliases. 😞

packages/wonder-blocks-dropdown/components/action-menu-opener-core.js
```diff
@@ -56,21 +56,22 @@ export default class ActionMenuOpenerCore extends React.Component<Props> {
...
-        };
-
         const label = (
             <LabelLarge style={sharedStyles.text}>{children}</LabelLarge>
         );
 
         return (
-            <StyledButton type="button" {...commonProps} disabled={disabled}>
+            <StyledButton
+                aria-disabled={disabled ? "true" : undefined}
```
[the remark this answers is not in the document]
kevinbarabash: The docs use strings so I'm going to stick with that for now.

packages/wonder-blocks-dropdown/components/multi-select.test.js
somewhatabstract: Yeah, I think that would be great in the long term - after all, seeing that the dropdown visually appears when we expect it too and looks right is a good thing.

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -445,6 +445,15 @@ export default class Dropdown extends React.Component<DropdownProps, State> {
         }
     };
 
+    getItemRole() {
+        switch (this.props.role) {
+            case "listbox":
+                return "option";
+            case "menu":
+                return "menuitem";
+        }
```
somewhatabstract: NIT: Should have a "default" case to throw an error if the role isn't one we explicitly support. Will make maintenance easier if things change expectedly or otherwise.

kevinbarabash: I'll make that change before merging.

kevinbarabash: While I added a throw here, I did not write a test for it because I would've had to disable flow checking to write a test that actually throws which seemed a little silly.

somewhatabstract: Oh, I just cast to `any` to get around that. :)

PR #358, opened 2018-11-09
WB-427: prevent page scrolling to bottom when opening modal

kevinbarabash:
This diff solves two issues:
- we were focus the button inside the modal too soon and so it was actually at the bottom of the page since styles hadn't been applied yet
- the sentinel divs in `FocusTrap` were positioned at the bottom of the page

Now we wait a bit for styles to be applied before focusing the button.  We also set the style of the sentinel divs to be `position: fixed` so that they're always in view.

### DIFF
--- a/packages/wonder-blocks-modal/components/modal-backdrop.js
+++ b/packages/wonder-blocks-modal/components/modal-backdrop.js
@@ -42,7 +42,10 @@ export default class ModalBackdrop extends React.Component<Props> {
         if (!lastButton) {
             return;
         }
-        lastButton.focus();
+        // wait for styles to applied
+        setTimeout(() => {
+            lastButton.focus();
+        }, 0);
     }
 
     /**
--- a/packages/wonder-blocks-modal/components/modal-backdrop.test.js
+++ b/packages/wonder-blocks-modal/components/modal-backdrop.test.js
@@ -5,6 +5,9 @@ import {mount} from "enzyme";
 import ModalBackdrop from "./modal-backdrop.js";
 import StandardModal from "./standard-modal.js";
 
+const sleep = (duration: number = 0) =>
+    new Promise((resolve, reject) => setTimeout(resolve, duration));
+
 const exampleModal = (
     <StandardModal
         content={<div data-modal-content />}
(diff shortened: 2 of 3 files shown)


jeresig:

Built with commit ed6cb8539e8b34f3cc8f6afee8f9c83ab4c877cd

https://deploy-preview-358--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-backdrop.js
```diff
@@ -42,7 +42,10 @@ export default class ModalBackdrop extends React.Component<Props> {
         if (!lastButton) {
             return;
         }
-        lastButton.focus();
+        // wait for styles to applied
+        setTimeout(() => {
+            lastButton.focus();
+        }, 0);
```
somewhatabstract: I hate the `setTimeout` hack in general but in specific, I think it's made for situations like this. :)

Shame Aphrodite doesn't have some sort of "I've applied styles" callback.

packages/wonder-blocks-modal/components/modal-backdrop.test.js
```diff
@@ -75,15 +75,18 @@ describe("ModalBackdrop", () => {
         expect(onCloseModal).not.toHaveBeenCalled();
     });
 
-    test("On mount, we focus the last button in the modal", () => {
+    test("On mount, we focus the last button in the modal", (done) => {
```
somewhatabstract: Is it good or bad that as soon as I see `done` I wonder "should this by async?"? Doing so would make the arrange/act/assert easier to do.

packages/wonder-blocks-modal/components/modal-backdrop.test.js
```diff
@@ -75,15 +75,18 @@ describe("ModalBackdrop", () => {
...
         const wrapper = mount(
             <ModalBackdrop onCloseModal={() => {}}>
                 {exampleModalWithButtons}
             </ModalBackdrop>,
         );
 
-        const lastButton = wrapper.find("[data-last-button]").getDOMNode();
-        expect(document.activeElement).toBe(lastButton);
+        setTimeout(() => {
```
somewhatabstract: Arrange/Act/Assert

packages/wonder-blocks-modal/components/modal-backdrop.test.js
kevinbarabash: I can update this to use `async`/`await`.

packages/wonder-blocks-modal/components/modal-backdrop.js
kevinbarabash: That'd be cool.  I'm not sure how to wire it up so that tests would have access to it.

packages/wonder-blocks-modal/components/modal-backdrop.test.js
```diff
@@ -75,18 +78,20 @@ describe("ModalBackdrop", () => {
...
-            const lastButton = wrapper.find("[data-last-button]").getDOMNode();
-            expect(document.activeElement).toBe(lastButton);
-            done();
-        }, 0);
+        // Act
+        await sleep(); // wait for styles to be applied
+        const lastButton = wrapper.find("[data-last-button]").getDOMNode();
+
+        // Assert
+        expect(document.activeElement).toBe(lastButton);
```
somewhatabstract: 😻

packages/wonder-blocks-modal/components/modal-backdrop.test.js
```diff
@@ -5,6 +5,9 @@ import {mount} from "enzyme";
 import ModalBackdrop from "./modal-backdrop.js";
 import StandardModal from "./standard-modal.js";
 
+const sleep = (duration: number = 0) =>
```
somewhatabstract: This might be a useful addition to our test utils

PR #359, opened 2018-11-09
WB-488: make IconButton focus work with Tooltip when tabbing

kevinbarabash:
The issue was that IconButton had an extra View wrapping the <button>, <a> or Link.  Since Button doesn't have this I don't think we need it on IconButton either.


jeresig:

Built with commit bf020b17183dbb11b6cf65c856140e3598a51ae3

https://deploy-preview-359--wonder-blocks.netlify.com

kevinbarabash:
A `vite` test for this is hard at the moment since I think we'd want a screenshot to verify that the focus ring is visible.

As for bumping the version, we just bumped and nobody's updated yet.  I think this would only be a breaking change if people set `padding` to something on the component or something like that which they shouldn't be doing anyway.

PR #360, opened 2018-11-14
Breaking(core) - Rename NoSSR to WithSSRPlaceholder

csilvers:
`NoSSR` is a misleading name, because with the `placeholder` param
it's not actually "no" ssr, it's "different" ssr.  (The name is doubly
misleading, since to me "No ssr" means that you can't render this
component during ssr -- it will raise an exception or some such --
when really what it means is that, by default, its children are
ignored when rendered via ssr.)

The new name, I think, makes the intent clearer.  I also made the
placeholder arg required, since I don't think it's right to privilege
"don't show anything" as the default behavior, since it's probably
not the right thing to do most of the time.  (If anything, I'd have
the default be "show a spinner".)

Test Plan:
yarn test

### DIFF
--- a/packages/wonder-blocks-core/components/with-ssr-placeholder.js
+++ b/packages/wonder-blocks-core/components/with-ssr-placeholder.js
@@ -7,19 +7,22 @@ import * as React from "react";
  */
 type Props = {|
     /**
-     * The content that is client-only.
+     * The content that is client-only.  This is what is rendered when
+     * not server-side rendering, or (when server-side rendering) after
+     * the initial rehydration has finished.
      */
     children: () => ?React.Node,
 
     /**
-     * Optionally, instead of rendering nothing the first time around, this
-     * component can render a placeholder. To enable this behavior,
-     * pass a "placeholder" prop.
+     * What to render during server-side rendering, or null not to
+     * render anything.
      *
-     * NOTE: Make sure the placeholder will render the same for both client and
-     * server, or it defeats the purpose of using the NoSSR component.
+     * NOTE: Make sure the placeholder will render the same for both
+     * client and server -- that is, it does the same thing for both
+     * the server-side renderer and the rehydration -- or it defeats
+     * the purpose of using the WithSSRPlaceholder component.
      */
-    placeholder?: () => ?React.Node,
+    placeholder: null | (() => React.Node),
 |};
 
 type State = {|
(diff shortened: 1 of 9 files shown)


somewhatabstract:
Is there an ADR discussing this change? Seems like this not only impacts the `NoSSR` component but also other things that use it like the `UniqueIDProvider`, changing the expectations between their use since now one demands placeholder and one does not.

In addition, the breaking change incurs potential tech debt when upgrading webapp as existing uses need to evaluate what this change means for those uses.

Perhaps these aren't big concerns, but I would like to ensure some level of consistency in our APIs and document the considerations to ensure the impact is fully understood.

somewhatabstract:
@jnetterf Since you wrote the ADR that led to this specific component, can you take a look at this? Thanks!

packages/wonder-blocks-core/components/with-ssr-placeholder.js
```diff
@@ -7,19 +7,22 @@ import * as React from "react";
...
      *
-     * NOTE: Make sure the placeholder will render the same for both client and
-     * server, or it defeats the purpose of using the NoSSR component.
+     * NOTE: Make sure the placeholder will render the same for both
+     * client and server -- that is, it does the same thing for both
+     * the server-side renderer and the rehydration -- or it defeats
+     * the purpose of using the WithSSRPlaceholder component.
      */
-    placeholder?: () => ?React.Node,
+    placeholder: null | (() => React.Node),
```
somewhatabstract: Why change the return type of the placeholder function? I think it's perfectly valid to let the calling code decide of nothing is the right return value.

The `null` can also be dropped and replaced with `?` like `?(() => ?React.Node)`.

csilvers:
> UniqueIDProvider, changing the expectations between their use since now one demands placeholder and one does not

Actually, that code is fine as-is.  UniqueIDProvider uses NoSSR, but provides its own placeholder, it's not expected the client will provide one.  (Instead the client specifies mockOnFirstRender, which I'd like to change to be non-optional as well, but I'll do that in a separate PR.)

packages/wonder-blocks-core/components/with-ssr-placeholder.js
csilvers: Good points!  Will change.

somewhatabstract:
> Is your worry that someone will think that WithSSRPlaceholder will render the placeholder, but it won't because some parent is rendering its placeholder instead, so the code here is never even getting seen? I don't think that's a problem, personally. To me it seems obvious that WithSSRPlaceholder just means "if this code is executed, at ssr time it will use the placeholder." After all, we don't have react method names like renderIfThisComponentIsEverMounted(); it's just render().

Yes, that's my concern because if the top-level thing was, say `UniqueIDProvider` with `mockOnFirstRender` as true, then it will tell all its children to render, including those that are `WithSSRPlaceholder`, which, knowing that a parent already took care of the SSR check will render its main content instead of the placeholder, but that may not be quite evident to those using the SSR-controlling components.

Having each one naively assume first render is SSR and invoke a second causes a big cascade, but having each one assume that if the top level component says it's all good, it isn't SSR is wrong. I need to check the implementation, but I suspect what we need to do is rework the `WithSSRPlaceholder` component to use the context value slightly differently so that child components know they're still rendering in SSR but we don't cascade renders.

This isn't a concern for your review. I'll put up a separate diff for that to ensure that this component works intuitively but is also performant and not cascading renders.

somewhatabstract:
@kevinbarabash You ok with this getting merged now?

PR #361, opened 2018-11-15
Breaking(core) - Make mockOnFirstRender required for UniqueIDProvider.

csilvers:
The (old) default behavior of UniqueIDProvider is that it would not
actually render its children during server-side rendering -- that is,
for the initial render.  That seems like a very surprising default to
me, and I think it would be for others as well.

I could change the default, but then you get a different non-obvious
behavior, that you get two different unique-id factories for the
initial render and subsequent renders.

So instead, I just make clients be explicit about which alternative
they prefer.  This helps ensure that they understand the consequences
of their choice.

Test Plan:
yarn test

### DIFF
--- a/packages/wonder-blocks-core/components/unique-id-provider.js
+++ b/packages/wonder-blocks-core/components/unique-id-provider.js
@@ -14,7 +14,8 @@ type Props = {|
      * the content to be rendered.
      *
      * If mockOnFirstRender is false, this is only called after
-     * the initial render has occurred and will always be called with the same
+     * the initial render has occurred -- it will be blank for the
+     * the first render -- and will always be called with the same
      * IIdentifierFactory instance.
      *
      * If mockOnFirstRender is true, this is called once with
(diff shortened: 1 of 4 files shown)


jeresig:

Built with commit 7a65ec041c17713c80bdf1c2276cad6971822c0c

https://deploy-preview-361--wonder-blocks.netlify.com

packages/wonder-blocks-core/components/unique-id-provider.js
```diff
@@ -34,7 +35,7 @@ type Props = {|
      * a mock IIdentifierFactory for the initial render, and then a unique ID
      * factory thereafter.
      */
-    mockOnFirstRender?: boolean,
+    mockOnFirstRender: boolean,
```
kevinbarabash: I believe this is a breaking change because clients using this component may not have set this in the past.  In practice though it won't break anything b/c we currently aren't using wonder-blocks' flow types yet in webapp, see https://phabricator.khanacademy.org/D49926 for details.

PR #362, opened 2018-11-16
Build files before linting

kevinbarabash:
We need to build all the packages before linting because we have a lint
check that verifies that all imports can be resolved.  The main entry
point for all of our packages is dist/index.js which is a built file.
Thus, in order to pass this check we must build the files before running
lint.  Reording the commands in the "test" script in package.json fixes
this.

Test Plan:
- yarn clean
- yarn install
- yarn test


jeresig:

Built with commit 4af63c88c25b2607c522d2994d9b4dc08184e889

https://deploy-preview-362--wonder-blocks.netlify.com

PR #363, opened 2018-11-20
Ensure WithSSRPlaceholder components nested inside placeholder renders work right

somewhatabstract:
Stacked on #360 

This change fixes a bug introduced when we removed the naive rendering cascade.

Before this change, if a SSR-aware component based off `WithSSRPlaceholder` was rendered inside the placeholder of another SSR-aware component, it would act as though it was not in the first server-side render, and render its full children. This was not necessarily what we want (for example, `UniqueIDProvider` instances nested inside the children of other `UniqueIDProvider` instances would render as if client-side instead of mocking their own ID factories).

This change fixes that by properly differentiating between:

* the root render, where the component is performing its very first render and is the root of an SSR component tree, and therefore responsible for invoking a second render
* the initial render, where the component is performing its very first render, but otherwise doesn't need to worry about things like invoking additional renders
* all subsequent renders, where the component renders normally.


jeresig:

Built with commit d3cfd804f11cda6fd68ca2be00e7bdd01f052ac0

https://deploy-preview-363--wonder-blocks.netlify.com

packages/wonder-blocks-core/components/with-ssr-placeholder.js
```diff
@@ -0,0 +1,158 @@
...
+        if (placeholder) {
+            return (
+                <RenderStateContext.Provider value="initial">
+                    {placeholder()}
+                </RenderStateContext.Provider>
+            );
+        }
+
+        // Otherwise, we return nothing.
+        return null;
```
somewhatabstract: Needs a test case

packages/wonder-blocks-core/components/with-ssr-placeholder.js
```diff
@@ -0,0 +1,158 @@
...
+                }
+                // Otherwise, we render nothing.
+                return null;
+
+            case "standard":
+                // We have covered the SSR render, we're now rendering normally.
+                return children();
+
+            default:
+                throw new Error("We got a render state we don't understand");
```
somewhatabstract: Needs a test case

packages/wonder-blocks-core/components/with-ssr-placeholder.js
```diff
@@ -0,0 +1,158 @@
...
+
+            case "initial":
+                // We're not the root component, so we just have to either
+                // render our placeholder or nothing.
+                // The second render is going to be triggered for us.
+                if (placeholder) {
+                    return placeholder();
+                }
+                // Otherwise, we render nothing.
+                return null;
```
somewhatabstract: Needs a test case

packages/wonder-blocks-core/components/with-ssr-placeholder.js
kevinbarabash: I hadn't considered excluding exceptions from code coverage.  This is one of those things that should in theory never happen, right?

packages/wonder-blocks-core/components/with-ssr-placeholder.js
kevinbarabash: Does this still need a test case?

packages/wonder-blocks-core/components/with-ssr-placeholder.js
```diff
@@ -22,14 +22,27 @@ type Props = {|
      * the server-side renderer and the rehydration -- or it defeats
      * the purpose of using the WithSSRPlaceholder component.
      */
-    placeholder: ?(() => ?React.Node),
+    placeholder: ?() => ?React.Node,
```
kevinbarabash: This is super sketchy, I'm surprised that prettier does this.

packages/wonder-blocks-core/components/with-ssr-placeholder.js
```diff
@@ -22,14 +22,27 @@ type Props = {|
...
 |};
 
-const HasHadFirstRenderContext = React.createContext<boolean>(false);
+type RenderState = "root" | "initial" | "standard";
+/**
+ * This is the context that tracks who is doing what in our SSR component tree.
+ *
+ * root:
+ *   no one has instigated an initial SSR render so the component that sees
+ *   this is responsible for updating the value to 1
```
kevinbarabash: I don't understand this comment.  Which value is being updated to 1?

packages/wonder-blocks-core/components/with-ssr-placeholder.js
```diff
@@ -72,49 +85,74 @@ export default class WithSSRPlaceholder extends React.Component<Props, State> {
...
+                // render our placeholder or nothing.
+                // The second render is going to be triggered for us.
+                if (placeholder) {
+                    return placeholder();
+                }
+                // Otherwise, we render nothing.
+                return null;
+
+            case "standard":
+                // We have covered the SSR render, we're now rendering normally.
```
kevinbarabash: I wonder if `"normal"` is a better constant for this case.

packages/wonder-blocks-core/components/with-ssr-placeholder.js
somewhatabstract: Yes, it's there for us when we're maintaining code to help avoid hidden issues.

packages/wonder-blocks-core/components/with-ssr-placeholder.js
somewhatabstract: Yes, I will be adding it shortly.

packages/wonder-blocks-core/components/with-ssr-placeholder.js
somewhatabstract: Agreed. I preferred what I wrote.

packages/wonder-blocks-core/components/with-ssr-placeholder.js
somewhatabstract: Ah, yeah. I need to clarify this. Great catch. Thanks

packages/wonder-blocks-core/components/with-ssr-placeholder.js
somewhatabstract: I considered `normal` but I think it's an ambiguous term these days (what is normal?); I felt `standard` had less baggage.

kevinbarabash: In that case maybe the comment should be update to match the constant.

PR #364, opened 2018-11-28
Fix jest config so that vscode's jest plugin works again

kevinbarabash:
The issue was that the jest plugin only looks at package.json so all the configuration has to be in there.  I didn't move the vite config in there b/c I think for now we'll want to run those tests manually.  Also, there's an issue with plugin not understanding the "projects" setting.


jeresig:

Built with commit 444abfd38bdbce4c4b444b2cecd2674080f5845d

https://deploy-preview-364--wonder-blocks.netlify.com

somewhatabstract:
@kevinbarabash This is great to get things working, but I suggest you look into `jest.pathToConfig` as a VSCode workspace setting for Wonder Blocks so that the config file can be used instead.

kevinbarabash:
@somewhatabstract `jest.pathToConfig` worked like a charm.  Thanks for the suggestion.

PR #366, opened 2018-11-30
WB-485: prevent disabled IconButtons from interferred with ones that aren't

kevinbarabash:
In practice this shouldn't come up b/c there should be appropriate spacing between IconButtons.  It doesn't hurt to set `point-events: none;` on the disabled IconButtons though so why not.


kevinbarabash:
I'm closing this b/c I got the ticket number wrong on the branch.

PR #368, opened 2018-12-10
[WB-505] Deal with strange context issues for root SSR use

somewhatabstract:
This makes the `default` case in context fall through to the `"root"` context rather than throwing an error. This should allow for the strange context bug seen in webapp.

I tried reproducing the issue in Wonderblocks (even downgrading) but it was not sufficient to reproduce at this time. For now, ensuring we cope gracefully seems fair.

### DIFF
--- a/packages/wonder-blocks-core/components/with-ssr-placeholder.js
+++ b/packages/wonder-blocks-core/components/with-ssr-placeholder.js
@@ -150,9 +150,14 @@ export default class WithSSRPlaceholder extends React.Component<Props, State> {
             // We ignore this from coverage. It's a maintenance case to help
             // us catch code changes that affect the control flow unexpectedly,
             // but it's not something we need to write a test case for.
+            // NOTE: This can occur if a v2 WithSSRPlaceholder-based component
+            // is rendered inside a v1 WithSSRPlaceholder-based component since
+            // the same context object is used.
             /* istanbul ignore next */
             default:
-                throw new Error("We got a render state we don't understand");
+                throw new Error(
+                    `We got a render state we don't understand: "${renderState}"`,
+                );
         }
     }
 
--- a/packages/wonder-blocks-core/components/with-ssr-placeholder.js
+++ b/packages/wonder-blocks-core/components/with-ssr-placeholder.js
@@ -129,6 +129,29 @@ export default class WithSSRPlaceholder extends React.Component<Props, State> {
         const {children, placeholder} = this.props;
 
         switch (renderState) {
+            // There are edge cases where for some reason, we get an unknown
+            // context value here. So far it seems to be when we're nested in a
+            // v1 WithSSRPlaceholder equivalent component, or in some older
+            // React v16 situations where we're nested in the provider of a
+            // different context.
+            //
+            // We ignore this from coverage. It's a maintenance case to help
+            // us catch code changes that affect the control flow unexpectedly,
+            // but it's not something we need to write a test case for.
+            //
+            /* istanbul ignore next */
+            default:
+                // Let's log this case so we can debug it easily.
+                // Then fall through to the root case.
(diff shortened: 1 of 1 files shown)


jeresig:

Built with commit 5bc798a26f13b53406610db69606958e9538a943

https://deploy-preview-368--wonder-blocks.netlify.com

somewhatabstract:
Put back into my queue as I'm going to expand this to address WB-505 issue around context.

kevinbarabash:
We should probably change our react version to match webapp's which is 16.3.0 and see that affects `WithPlaceholderSSR`.

packages/wonder-blocks-core/components/with-ssr-placeholder.js
```diff
@@ -150,9 +150,14 @@ export default class WithSSRPlaceholder extends React.Component<Props, State> {
...
             // us catch code changes that affect the control flow unexpectedly,
             // but it's not something we need to write a test case for.
+            // NOTE: This can occur if a v2 WithSSRPlaceholder-based component
+            // is rendered inside a v1 WithSSRPlaceholder-based component since
+            // the same context object is used.
             /* istanbul ignore next */
             default:
-                throw new Error("We got a render state we don't understand");
+                throw new Error(
+                    `We got a render state we don't understand: "${renderState}"`,
```
csilvers: In our case `renderState` was an object.  Will this serialize the object reasonably, or just show `[Object]`?

somewhatabstract: Oh good catch!

somewhatabstract:
OK, I copied this change into a local version of webapp with @csilvers changes for using core-v2 and I was able to verify that everything seems to be working now bar some unrelated issues with the SSR testing harness. We will have to verify properly once this is released and available in webapp, but it appears to be working.

packages/wonder-blocks-core/components/with-ssr-placeholder.js
```diff
@@ -129,6 +129,31 @@ export default class WithSSRPlaceholder extends React.Component<Props, State> {
         const {children, placeholder} = this.props;
 
         switch (renderState) {
+            // There are edge cases where for some reason, we get an unknown
+            // context value here. So far it seems to be when we're nested in a
+            // v1 WithSSRPlaceholder equivalent component, or in some older
+            // React v16 situations where we're nested in the provider of a
+            // different context.
```
kevinbarabash: I thought it was an issue specifically with apollo and react-router's abuse of Context API internals.  IS that no longer the case?

somewhatabstract: I don't know for sure. That's merely a guess.

somewhatabstract: I have not been able to repro outside of webapp nor identify why webapp is bleeding values across contexts. All thoughts on why it happens are really guesses. I've made this stuff more robust to that circumstance and tested it out with Craig's changes to check that it address the core issues.

Why the context bleeds at all is yet to be fully understood.

packages/wonder-blocks-core/components/with-ssr-placeholder.js
```diff
@@ -129,6 +129,31 @@ export default class WithSSRPlaceholder extends React.Component<Props, State> {
...
+            //
+            // We ignore this from coverage. It's a maintenance case to help
+            // us catch code changes that affect the control flow unexpectedly,
+            // but it's not something we need to write a test case for.
+            //
+            /* istanbul ignore next */
+            default:
```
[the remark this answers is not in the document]
somewhatabstract: Because I felt it was worth highlighting as important given the perhaps unexpected behavior of it and that it falls through to the "root" case directly below it, so it makes for a more readable flow if it sits above it.

PR #369, opened 2018-12-10
WB-504: BREAKING(core) - Add testId (and AriaProps) to all public components

kevinbarabash:
This PR does a little more work than is necessary, but I ran into some issues with the inexact types we were using for `View` and `Text` so I ended up making them exact including `AriaProps`.  Also, now that `AriaProps` is exact it was super easy to add it to all of our other components which had `aria-label?: string` as a prop.

This would be a breaking change if we were using flow typing.  I'm going to try to turn on flow typing for some packages (not typography).  There's another breaking change I'd like to make to switch to adopt ref forwarding on all of our components so that people can easily access the underlying DOM element.  I'd like to group multiple breaking changes together to avoid having more versions than is necessary of components in webapp.

### DIFF
--- a/packages/wonder-blocks-core/__snapshots__/generated-snapshot.test.js.snap
+++ b/packages/wonder-blocks-core/__snapshots__/generated-snapshot.test.js.snap
@@ -138,6 +138,63 @@ exports[`wonder-blocks-core example 3 1`] = `
 >
   <div
     className=""
+    data-test-id="foo"
+    style={
+      Object {
+        "alignItems": "stretch",
+        "borderStyle": "solid",
+        "borderWidth": 0,
+        "boxSizing": "border-box",
+        "display": "flex",
+        "flexDirection": "column",
+        "margin": 0,
+        "minHeight": 0,
+        "minWidth": 0,
+        "padding": 0,
+        "position": "relative",
+        "zIndex": 0,
+      }
+    }
+  >
+    Foo
+  </div>
+  <span
+    className=""
+    data-test-id="bar"
+    style={
+      Object {
+        "MozOsxFontSmoothing": "grayscale",
+        "WebkitFontSmoothing": "antialiased",
+      }
+    }
+  >
+    Bar
+  </span>
+</div>
+`;
+
+exports[`wonder-blocks-core example 4 1`] = `
+<div
+  className=""
+  style={
+    Object {
+      "alignItems": "stretch",
+      "borderStyle": "solid",
+      "borderWidth": 0,
+      "boxSizing": "border-box",
+      "display": "flex",
+      "flexDirection": "column",
+      "margin": 0,
+      "minHeight": 0,
+      "minWidth": 0,
+      "padding": 0,
+      "position": "relative",
+      "zIndex": 0,
+    }
+  }
+>
+  <div
+    className=""
     style={
       Object {
         "alignItems": "stretch",
@@ -333,7 +390,7 @@ exports[`wonder-blocks-core example 3 1`] = `
 </div>
 `;
 
-exports[`wonder-blocks-core example 4 1`] = `
+exports[`wonder-blocks-core example 5 1`] = `
 <div
   className=""
   style={
@@ -526,7 +583,7 @@ exports[`wonder-blocks-core example 4 1`] = `
 </div>
 `;
 
-exports[`wonder-blocks-core example 5 1`] = `
+exports[`wonder-blocks-core example 6 1`] = `
 <div
   className=""
   style={
@@ -659,7 +716,7 @@ exports[`wonder-blocks-core example 5 1`] = `
 </div>
 `;
 
-exports[`wonder-blocks-core example 6 1`] = `
+exports[`wonder-blocks-core example 7 1`] = `
 <div
   className=""
   style={
@@ -776,7 +833,7 @@ exports[`wonder-blocks-core example 6 1`] = `
 </div>
 `;
 
-exports[`wonder-blocks-core example 7 1`] = `
+exports[`wonder-blocks-core example 8 1`] = `
 <div
   className=""
   style={
@@ -835,7 +892,7 @@ exports[`wonder-blocks-core example 7 1`] = `
 </div>
 `;
 
-exports[`wonder-blocks-core example 8 1`] = `
+exports[`wonder-blocks-core example 9 1`] = `
 <div
   className=""
   style={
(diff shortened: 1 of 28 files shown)


somewhatabstract:
Why did the test coverage decrease?

kevinbarabash:
It looks like we're computing test coverage on types.js but we codecov doesn't realize that this file is actually being included by other files b/c it's strictly flow types.  We should probably ignore types.js files when computing code coverage.

packages/wonder-blocks-core/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -117,6 +117,63 @@ exports[`wonder-blocks-core example 2 1`] = `
...
+      "minWidth": 0,
+      "padding": 0,
+      "position": "relative",
+      "zIndex": 0,
+    }
+  }
+>
+  <div
+    className=""
+    data-test-id="foo"
```
kevinbarabash: This is the `testId` from the `View`.

packages/wonder-blocks-core/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -117,6 +117,63 @@ exports[`wonder-blocks-core example 2 1`] = `
...
+        "position": "relative",
+        "zIndex": 0,
+      }
+    }
+  >
+    Foo
+  </div>
+  <span
+    className=""
+    data-test-id="bar"
```
kevinbarabash: This is the `testId` from the `Text`.

packages/wonder-blocks-core/util/types.js
```diff
@@ -132,13 +132,92 @@ export type AriaProps = {
...
+type EventHandlers = {|
+    ...MouseEvents,
+    ...KeyboardEvents,
+    ...InputEvents,
+    ...TouchEvents,
+    ...FocusEvents,
+    // TODO: add remaining supported events https://reactjs.org/docs/events.html#supported-events
+|};
+
+export type CommonProps = {|
```
kevinbarabash: I should maybe renamed this `SharedProps` since we use that in other files.

kevinbarabash:
@somewhatabstract I removed `types.js` files from the codecov report.  I think this is a more accurate measure of our coverage.

somewhatabstract:
Nice coverage gain! :grin:

packages/wonder-blocks-core/util/types.js
somewhatabstract: I think that either way, it's not a great name since it doesn't really say what it is for; (what is it common or shared with?). I think I would like to see a name that conveyed when it should be used rather than that it is shared in any way. In lieu of that (or perhaps in addition to that), it needs some commentary to explain its purpose.

packages/wonder-blocks-core/util/types.js
```diff
@@ -132,13 +132,93 @@ export type AriaProps = {
...
+|};
+
+type EventHandlers = {|
+    ...MouseEvents,
+    ...KeyboardEvents,
+    ...InputEvents,
+    ...TouchEvents,
+    ...FocusEvents,
+    // TODO: add remaining supported events https://reactjs.org/docs/events.html#supported-events
+|};
```
somewhatabstract: NIT: It would be nice if there were comments on each of these props, but I also get why there aren't.

packages/wonder-blocks-core/components/text.md
```diff
@@ -38,3 +38,12 @@ Other props can be passed through `View` or `Text`, as if they were normal tags.
     </Text>
 </View>
 ```
+
+Both `View` and `Text` support a `testId` prop.
+
+```js
+<View>
+    <View testId="foo">Foo</View>
+    <Text testId="bar">Bar</Text>
```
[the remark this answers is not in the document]
kevinbarabash: I'll move the `Text` example into an actual test file before merging this.

kevinbarabash:
Not everything has to be in the docs either.  It just have happen to be convenient and I should've put some more thought into whether it needed to be an example or whether putting it in a separate .test.js file was more appropriate.

kevinbarabash:
I'm going to split up this diff to separate breaking changes from non-breaking changes.

packages/wonder-blocks-core/util/types.js
```diff
@@ -132,13 +132,92 @@ export type AriaProps = {
...
+
+    tabIndex?: number,
+
+    id?: string,
+
+    // TODO(kevinb) remove the need for this
+    "data-modal-launcher-portal"?: boolean,
+
+    // Used by tooltip bubble
+    "data-placement"?: string,
```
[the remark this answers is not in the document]
jeresig: Hmm yeah - this is pretty awkward. Can you file issues to track these data-* props so that we can remove them/change them later?

packages/wonder-blocks-core/util/types.js
```diff
@@ -132,13 +132,93 @@ export type AriaProps = {
...
+    style?: StyleType,
+
+    /**
+     * Test ID used for e2e testing.
+     */
+    testId?: string,
+
+    tabIndex?: number,
+
+    id?: string,
```
jeresig: Might be worth adding comments for these, just to clarify that there's nothing special going on.

packages/wonder-blocks-dropdown/components/option-item.js
```diff
@@ -122,7 +122,7 @@ export default class OptionItem extends React.Component<OptionProps> {
 
                     return (
                         <View
-                            data-test-id={testId}
```
jeresig: Ah, caught a bug!

packages/wonder-blocks-form/components/radio-group.js
```diff
@@ -56,6 +56,11 @@ type RadioGroupProps = {|
      * Value of the selected radio item.
      */
     selectedValue: string,
+
```
jeresig: I wonder if it would be worth it to have a `TestIdProps` Type that we import and apply, like we do with AriaProps. It would make future changes simpler and put the docs all in one place.

packages/wonder-blocks-tooltip/components/tooltip-anchor.test.js
```diff
@@ -138,7 +138,7 @@ describe("TooltipAnchor", () => {
                         anchorRef={resolve}
                         onActiveChanged={() => {}}
                     >
-                        <View tabIndex="-1">This is the anchor</View>
+                        <View tabIndex={-1}>This is the anchor</View>
```
jeresig: Nice!

jeresig:
I'm going to land this!

kevinbarabash:
I can't believe I missed adding `AriaProps` to `Button`.

PR #370, opened 2019-01-12
Button: Prevent user-selection during press on safari mobile

ariabuckles:
When accessing https://wonder-blocks.netlify.com/ on mobile safari on iOS,
pressing and holding on the default coloured buttons results in the
text-selection UI popping up after a couple seconds. This prevents the
user from being able to click the button.

This commit moves the `user-select: none` styling from the text inside
ButtonCore to the ButtonCore button element itself, which prevents this
misbehaviour on iOS.

Test plan:

1. Ran `yarn start`
2. Opened the resulting page on my iPhone from the local network
3. Pressed and held on the default blue buttons on my phone
    * Verified that no selection popup occurred
    * Verified that releasing the press triggered the modals
4. Ran `yarn jest` and verified that the snapshot changes consisted
   entirely of changes to `user-select: none` styles.

### DIFF
--- a/packages/wonder-blocks-button/components/button-core.js
+++ b/packages/wonder-blocks-button/components/button-core.js
@@ -167,6 +167,7 @@ const sharedStyles = StyleSheet.create({
         // This removes the 300ms click delay on mobile browsers by indicating that
         // "double-tap to zoom" shouldn't be used on this element.
         touchAction: "manipulation",
+        userSelect: "none",
     },
     withIcon: {
         // The left padding for the button with icon should have 4px less padding
@@ -182,7 +183,6 @@ const sharedStyles = StyleSheet.create({
         display: "flex",
         alignItems: "center",
         fontWeight: "bold",
-        userSelect: "none",
         whiteSpace: "nowrap",
         overflow: "hidden",
         textOverflow: "ellipsis",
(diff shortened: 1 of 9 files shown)


jeresig:

Built with commit 1081a091bda47b44eecf7e3bee9aab3fe648ab46

https://deploy-preview-370--wonder-blocks.netlify.com

ariabuckles:
Hi! I noticed this while poking around on https://wonder-blocks.netlify.com/ . I'd be happy to try a different approach to fixing this if you'd prefer, or if this is somehow incompatible with KA's internal usage, feel free to not worry about this.

Best,
Aria

somewhatabstract:
Thank you for your contribution, @ariabuckles. We'll take a look.

🎉

ariabuckles:
Yay thanks for taking a look at it @somewhatabstract and @jeresig !!

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -167,6 +167,7 @@ const sharedStyles = StyleSheet.create({
         // This removes the 300ms click delay on mobile browsers by indicating that
         // "double-tap to zoom" shouldn't be used on this element.
         touchAction: "manipulation",
+        userSelect: "none",
```
kevinbarabash: Good call moving this to `shared`.

PR #372, opened 2019-02-25
Implement MediaLayout component in wonder-blocks-layout

kevinbarabash:
This diff is the first part of splitting up #311 in more manageable chunks.  media-layout.js is this diff is equivalent to layout.js from #311.  The media-layout.js from #311 has been deprecated in favor of using `MediaLayoutContext` directly.  I've also moved `MediaLayoutContext` into its own file similar to what we do for `ModalContext`.

I'm going to finish splitting up #311 and will come back to this PR to flesh out docs and tests.


jeresig:

Built with commit d8b34f49161fa890165ed931bfdee70504b7c04e

https://deploy-preview-372--wonder-blocks.netlify.com

kevinbarabash:
As part of this PR I want to remove all MediaLayout stuff from `wonder-blocks-core`.

kevinbarabash:
> As part of this PR I want to remove all MediaLayout stuff from `wonder-blocks-core`.

I actually I'm going to do this as a PR after I've transitioned modals and grid to use `MediaLayout` from `wonder-blocks-layout` since it'll be easier to do that way.

packages/wonder-blocks-layout/components/media-layout-context.js
```diff
@@ -0,0 +1,56 @@
+// @flow
+import * as React from "react";
+
+import {MEDIA_DEFAULT_SPEC} from "../util/specs.js";
+import type {MediaSize, MediaSpec} from "../util/types.js";
+
+type Context = {|
```
somewhatabstract: Export this type?

kevinbarabash: Will do.

packages/wonder-blocks-layout/components/media-layout-context.test.js
```diff
@@ -0,0 +1,164 @@
...
+                                resolve({mediaSize, mediaSpec, styles});
+                                return <View>Hello, world!</View>;
+                            }}
+                        </MediaLayout>
+                    </MediaLayoutContext.Provider>,
+                );
+            });
+
+            // Act
+            const args = await promise;
```
somewhatabstract: What do you think about combining this with the `// Act` line since it really defines what the action is?

packages/wonder-blocks-layout/components/media-layout-context.test.js
```diff
@@ -0,0 +1,164 @@
...
+
+            // Assert
+            expect(args.mediaSize).toEqual("large");
+        });
+    });
+});
```
somewhatabstract: Loving all the tests and yay for async/await.

packages/wonder-blocks-layout/components/media-layout.js
```diff
@@ -0,0 +1,242 @@
...
+                    const mediaSize =
+                        overrideSize ||
+                        (this.isServerSide() && ssrSize) ||
+                        this.getCurrentSize(mediaSpec);
+
+                    // Generate a mock stylesheet
+                    const styles = this.getMockStyleSheet(mediaSize);
+
+                    return children({mediaSize, mediaSpec, styles});
+                }}
```
somewhatabstract: It's only a small method, but I think moving this to a separate method rather than inline would be cleaner. Thoughts?

packages/wonder-blocks-layout/components/media-layout.test.js
```diff
@@ -0,0 +1,343 @@
...
+        <MediaLayout styleSheets={styleSheets}>
+            {({mediaSize, mediaSpec, styles}) => {
+                return <View style={styles.test}>Hello, world!</View>;
+            }}
+        </MediaLayout>,
+    );
+    const style = wrapper.find("div").prop("style");
+
+    return style;
+};
```
somewhatabstract: Whereas in implementations, duplication is often a code smell, I find the opposite is true in testing since test cases are intended to be discrete, isolated scenarios. What do you think about inlining these things in each test instead of sharing implementation across tests? I think it would make each test case easier to grok in place as well as ease maintenance (especially if test cases diverge).

packages/wonder-blocks-layout/components/media-layout.test.js
```diff
@@ -0,0 +1,343 @@
...
+                };
+                resizeWindow(size);
+
+                // Act
+                const style = getStyle(styleSheets);
+                const color = {
+                    small: "blue",
+                    medium: "blue",
+                    large: "orange",
+                }[size];
```
somewhatabstract: This `color` variable should be in the arrange section, right? And probably named `expectation`?

packages/wonder-blocks-layout/components/media-layout.test.js
```diff
@@ -0,0 +1,343 @@
...
+                        test: {
+                            color: "orange",
+                        },
+                    }),
+                };
+                resizeWindow(size);
+
+                // Act
+                const style = getStyle(styleSheets);
+                const color = {
```
somewhatabstract: expectation?

packages/wonder-blocks-layout/components/media-layout.test.js
```diff
@@ -0,0 +1,343 @@
...
+                        return <View style={styles.test}>Hello, world!</View>;
+                    }}
+                </MediaLayout>,
+            );
+
+            resizeWindow("small");
+            wrapper.update();
+
+            // Act
+            const style = wrapper.find("div").prop("style");
```
somewhatabstract: I think some of the things in the `Arrange` part should be here as they are the action that is being tested. This line specifically is really a sampling point after the action (it belongs here, but it is not actually the action under test).

packages/wonder-blocks-layout/components/media-layout.test.js
```diff
@@ -0,0 +1,343 @@
...
+
+            // Assert
+            expect(style.color).toBe("blue");
+        });
+    });
+});
```
somewhatabstract: Love the tests!

I think some of them have their "act" code in the "arrange" section and could do with some tweaks.

packages/wonder-blocks-layout/docs.md
```diff
@@ -32,3 +32,57 @@ const styles = StyleSheet.create({
     <Button style={styles.button}>Accept</Button>
 </View>
 ```
+
+`MediaLayout` is a container component takes in a `styleSheets` object whose keys are
```
somewhatabstract: Should there be a heading here?

packages/wonder-blocks-layout/docs.md
```diff
@@ -32,3 +32,57 @@ const styles = StyleSheet.create({
...
+media sizes.  I listens for changes to the current media size and passes the current
+`mediaSize`, `mediaSpec`, and `styles` to `children` which is a render function taking
+those three values as object.
+
+Valid keys for the `styleSheets` object are (in order of precedence):
+- small, medium, large
+- mdOrSmaller, mdOrLarger
+- all
```
somewhatabstract: Should each of these be `code` highlighted? e.g. `small`, `medium`, `large`

packages/wonder-blocks-layout/docs.md
```diff
@@ -32,3 +32,57 @@ const styles = StyleSheet.create({
...
+
+`MediaLayout` will merge style rules from multiple styles that match the current media
+size.
+
+The `mediaSpec` is an object with one or more of the following keys: small, medium, or
+large.  Each value contains the following data:
+- `query: string` e.g. "(min-width: 1024px)"
```
somewhatabstract: e.g. "`(min-width: 1024px)`"?

packages/wonder-blocks-layout/index.js
```diff
@@ -1,3 +1,19 @@
...
 export {default as Spring} from "./components/spring.js";
 export {default as Strut} from "./components/strut.js";
+export * from "./util/specs.js";
+export {queryMatchesSize} from "./util/util.js";
+
+export type {
+    MediaQuery,
+    MediaSize,
+    MediaSpec,
+    Context as MediaLayoutContextType,
```
somewhatabstract: Does this need the word `Type` on the end?

packages/wonder-blocks-layout/util/test-util.js
```diff
@@ -0,0 +1,62 @@
...
+    if (size === "large") {
+        window.innerWidth = 1200;
+    } else if (size === "medium") {
+        window.innerWidth = 800;
+    } else if (size === "small") {
+        window.innerWidth = 640;
+    }
+
+    window.dispatchEvent(new Event("resize"));
+}
```
somewhatabstract: I think we might want tests for some of these helpers to ensure they work as we expect.

packages/wonder-blocks-layout/util/test-util.js
```diff
@@ -0,0 +1,62 @@
...
+            return checkQuery(query, window.innerWidth);
+        },
+        addListener: (listener: Listener) => {
+            listeners.add(listener);
+        },
+        removeListener: (listener: Listener) => {
+            listeners.delete(listener);
+        },
+    };
+}
```
somewhatabstract: I think we need a test for this to ensure it's working the way we want. Should help with test failures when maintaining the tests that depend on them. What do you think?

packages/wonder-blocks-layout/util/types.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+export type MediaSize = "small" | "medium" | "large";
+
+export type MediaQuery = "all" | "mdOrSmaller" | "mdOrLarger" | MediaSize;
+
+// eslint-disable-next-line flowtype/require-exact-type
```
somewhatabstract: Can you add a comment on why we're suppressing this? I think it will help others when they come to this code.

packages/wonder-blocks-layout/util/util.js
```diff
@@ -0,0 +1,24 @@
+// @flow
+import type {MediaQuery, MediaSize} from "./types.js";
+
+export const queryMatchesSize = (
```
somewhatabstract: Comment to document the method?

packages/wonder-blocks-layout/util/util.js
```diff
@@ -0,0 +1,24 @@
...
+            return mediaSize === "medium" || mediaSize === "small";
+        case "medium":
+            return mediaSize === "medium";
+        case "mdOrLarger":
+            return mediaSize === "medium" || mediaSize === "large";
+        case "large":
+            return mediaSize === "large";
+        default:
+            throw new Error(`Unsupported mediaSize: ${mediaSize}`);
+    }
```
somewhatabstract: NIT: I'm used to having a newline after each case. Not sure if that's a standard style of ours or not.

packages/wonder-blocks-layout/util/util.test.js
```diff
@@ -0,0 +1,41 @@
...
+            }match ${size}`, () => {
+                // Act
+                const actualResult = queryMatchesSize(query, size);
+
+                // Assert
+                expect(actualResult).toEqual(expectedResult);
+            });
+        }
+    });
+};
```
somewhatabstract: I'm really not keen on this kind of test-case generation.

Usually, we don't need to test every single possible combination and I don't know that it really is helpful when tests fail. For example, if only one case fails, it's really hard to investigate that case in isolation.

Instead, could you write each case out itself. It will be clearer what's happening for folks working on this code and though it means a lot of very similar code, it will be easier to maintain.

packages/wonder-blocks-layout/util/util.test.js
```diff
@@ -0,0 +1,41 @@
...
+
+describe.only("queryMatchesSize", () => {
+    testQueryMatchesSizes("all", new Set(["small", "medium", "large"]));
+    testQueryMatchesSizes("mdOrSmaller", new Set(["small", "medium"]));
+    testQueryMatchesSizes("mdOrLarger", new Set(["medium", "large"]));
+    testQueryMatchesSizes("large", new Set(["large"]));
+    testQueryMatchesSizes("medium", new Set(["medium"]));
+    testQueryMatchesSizes("small", new Set(["small"]));
+
+    // TODO(kevinb): figure out how to add $FlowIgnore without breaking things
```
somewhatabstract: Raise an issue for this?

utils/gen-snapshot-tests.js
```diff
@@ -139,6 +139,10 @@ function extractExamplesAndComponentsForFile(sourceFile, componentFileMap) {
         src.match(/export default class \s*(\w+)/) ||
         src.match(/export default \s*\w+\(\s*(\w+)/);
 
+    if (!match) {
```
somewhatabstract: What is this guarding against? Perhaps a comment?

packages/wonder-blocks-layout/docs.md
kevinbarabash: Good call.

packages/wonder-blocks-layout/index.js
kevinbarabash: We're already exporting `MediaLayoutContext`.  I'll change this `MediaLayoutContextValue` which is what this type is describing.

packages/wonder-blocks-layout/util/test-util.js
kevinbarabash: Good idea.

packages/wonder-blocks-layout/util/util.js
kevinbarabash: There isn't anything in our style guides, I can add newlines.

packages/wonder-blocks-layout/components/media-layout.test.js
kevinbarabash: ah... `expectedColor`?

packages/wonder-blocks-layout/components/media-layout-context.test.js
kevinbarabash: I don't follow.  This is the `// Act` line. I can combine this with `// Arrange` and do something like:
```
// Arrange, Act
const args = await new Promise(...
```

utils/gen-snapshot-tests.js
kevinbarabash: There was a file that we were traversing that didn't have a default export.

packages/wonder-blocks-layout/components/media-layout-context.test.js
somewhatabstract: Yeah, that's what I was getting at. When the test reads like a story, it is often easier to understand what happens when.

packages/wonder-blocks-layout/components/media-layout.test.js
```diff
@@ -0,0 +1,403 @@
...
+    });
+
+    describe("styleSheets", () => {
+        const testSizes = {
+            small: [640, 480],
+            medium: [800, 600],
+            large: [1200, 800],
+        };
+
+        for (const size of Object.keys(testSizes)) {
```
somewhatabstract: Not convinced this loops is worth it. Not enough to hold up the review, but I'd prefer distinct test cases than clever test generation tricks.

packages/wonder-blocks-layout/components/media-layout.test.js
somewhatabstract: Still think this is a valid point. There's no actual action happening on this line. I think the second resize and update call should be here too; they're what trigger the code that we're testing, right?

packages/wonder-blocks-layout/components/media-layout.test.js
```diff
@@ -0,0 +1,403 @@
...
+                        return <View style={styles.test}>Hello, world!</View>;
+                    }}
+                </MediaLayout>,
+            );
+
+            resizeWindow("large");
+            wrapper.update();
+
+            // Act
+            const style = wrapper.find("div").prop("style");
```
somewhatabstract: Same comment as above.

packages/wonder-blocks-layout/util/test-util.test.js
```diff
@@ -0,0 +1,275 @@
...
+            // Assert
+            expect(promise).resolves.toHaveProperty("type", "resize");
+            window.removeEventListener("resize", handler);
+        });
+    });
+
+    describe("checkQuery", () => {
+        describe("(max-width: 767px)", () => {
+            // Arrange
+            const queryString = "(max-width: 767px)";
```
somewhatabstract: I don't think this constant is valuable. Rather have the string inline in the test so the test case is easily understood as one thing.

packages/wonder-blocks-layout/util/test-util.test.js
```diff
@@ -0,0 +1,275 @@
...
+                const result = checkQuery(queryString, width);
+
+                // Assert
+                expect(result).toBe(false);
+            });
+        });
+
+        describe("(min-width: 768px) and (max-width: 1023px)", () => {
+            // Arrange
+            const queryString = "(min-width: 768px) and (max-width: 1023px)";
```
somewhatabstract: Prefer this text were inline in each test.

packages/wonder-blocks-layout/components/media-layout-context.test.js
```diff
@@ -0,0 +1,163 @@
...
+        window.matchMedia = matchMedia;
+    });
+
+    describe("overrideSize", () => {
+        it("should override the currentSize", async () => {
+            // Arrange
+            resizeWindow("large");
+
+            // Act
+            const args = await new Promise((resolve, reject) => {
```
somewhatabstract: This is great. Can you do it for the other test cases in this file?

packages/wonder-blocks-layout/util/types.js
somewhatabstract: Since this is now exact, isn't this lint suppression redundant?

packages/wonder-blocks-layout/components/media-layout.test.js
kevinbarabash: I'm going to leave these for now.

packages/wonder-blocks-layout/util/types.js
kevinbarabash: Thanks for catching that.

PR #373, opened 2019-02-26
Update modals to use MediaLayout to implement mobile designs

kevinbarabash:
This PR builds on top of #372 by using MediaLayout to implement mobile modal designs.

TODOs: 
- [x] add some tests that use `overrideSize` to create snapshot tests for the mobile versions of the modals.


jeresig:

Built with commit b460befa461efd6483564954ad22ff48a60d7b74

https://deploy-preview-373--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/close-button.js
```diff
@@ -2,50 +2,39 @@
...
-     * tertiary button title to color.
-     */
-    light: boolean,
-
-    /**
-     * Called when the close button is clicked.
-     */
-    onClick?: () => void,
+    light?: boolean,
+    onClose?: () => void,
```
somewhatabstract: Where are the prop comments and why the change to `onClose` from `onClick`. The `CloseButton` doesn't get closed, so `onClose` doesn't really make sense

packages/wonder-blocks-modal/components/one-column-modal.js
```diff
@@ -51,53 +73,28 @@ class ContentWrapper extends React.Component<WrappedProps> {
...
+                <MediaLayout>
+                    {({mediaSize}) =>
+                        mediaSize === "small" ? (
+                            <SmallOneColumnModal {...this.props} />
+                        ) : (
+                            <LargeOneColumnModal {...this.props} />
+                        )
+                    }
+                </MediaLayout>
+            </React.Fragment>
```
somewhatabstract: Is the React.Fragment needed?

packages/wonder-blocks-modal/components/two-column-modal.js
```diff
@@ -79,73 +126,32 @@ class ContentWrapper extends React.Component<WrappedProps> {
...
+                <MediaLayout>
+                    {({mediaSize}) =>
+                        mediaSize === "small" ? (
+                            <SmallTwoColumnModal {...this.props} />
+                        ) : (
+                            <LargeTwoColumnModal {...this.props} />
+                        )
+                    }
+                </MediaLayout>
+            </React.Fragment>
```
somewhatabstract: Is the React.Fragment necessary?

packages/wonder-blocks-modal/components/close-button.js
kevinbarabash: Good call.

packages/wonder-blocks-modal/components/one-column-modal.js
kevinbarabash: I'll drop it.

kevinbarabash:
I noticed I still have a TODO to add some fixture tests for the mobile versions of the modals so I'm going to add those now.

packages/wonder-blocks-modal/components/close-button.js
```diff
@@ -2,34 +2,20 @@
...
-     * tertiary button title to color.
-     */
-    light: boolean,
-
-    /**
-     * Called when the close button is clicked.
-     */
+    light?: boolean,
     onClick?: () => void,
+    style?: StyleType,
```
somewhatabstract: Shouldn't we still have comments to describe these props?

packages/wonder-blocks-modal/components/close-button.js
```diff
@@ -38,14 +24,17 @@ export default class CloseButton extends React.Component<Props> {
...
                     }
+
                     return (
                         <IconButton
                             icon={icons.dismiss}
                             // TODO(mdr): Translate this string for i18n.
+                            // TODO(kevinb): provide a way to set this label
                             aria-label="Close modal"
-                            onClick={closeModal || onClick}
+                            onClick={onClick || closeModal}
```
somewhatabstract: Why are these reversed?

packages/wonder-blocks-modal/components/modal-content.js
```diff
@@ -26,54 +27,54 @@ export default class ModalContent extends React.Component<Props> {
...
-    small: {
-        padding: "32px 16px",
-    },
-});
+    small: StyleSheet.create({
+        content: {
+            padding: "32px 16px",
+        },
+    }),
+};
```
somewhatabstract: I like this pattern together with the `MediaLayout` component. Nice.

packages/wonder-blocks-modal/components/modal-dialog.js
```diff
@@ -17,41 +17,48 @@ type Props = {|
 export default class ModalDialog extends React.Component<Props> {
     render() {
         const {style, children} = this.props;
+        const contextValue = {
```
somewhatabstract: Add the context type here so we can flow type this nicely?

packages/wonder-blocks-modal/components/modal-panel.js
```diff
@@ -88,100 +85,113 @@ export default class ModalPanel extends React.Component<Props> {
...
-                {!footer ||
-                (typeof footer === "object" && footer.type === ModalFooter) ? (
-                    footer
-                ) : (
-                    <ModalFooter>{footer}</ModalFooter>
-                )}
-            </View>
+            <MediaLayout styleSheets={styleSheets}>
+                {({styles}) => {
+                    let mainContent =
```
somewhatabstract: I find use of `let` to be a code smell that a function is needed. What do you think about using a function so that this can be `const` instead? Might make the code a little less nested and easier to follow.

packages/wonder-blocks-modal/components/one-column-modal.js
```diff
@@ -27,22 +27,44 @@ type BaseProps = {|
...
+                <View style={styles.contentFooterWrapper}>
+                    <View style={styles.smallContentWrapper}>
+                        <ModalPanel
+                            showCloseButton
+                            onClose={onClose}
+                            content={content}
+                            scrollOverflow={false}
+                        />
+                    </View>
+                    {footer && (
```
somewhatabstract: Flow upgrades aren't going to like this. Change to `!!footer`.

packages/wonder-blocks-modal/components/one-column.modal.test.js
```diff
@@ -0,0 +1,78 @@
...
+import {mount, unmountAll} from "../../../utils/testing/mount.js";
+
+describe("OneColumnModal", () => {
+    beforeEach(() => {
+        unmountAll();
+    });
+
+    it("should render SmallOneColumnModal on mobile", () => {
+        // Arrange
+        const wrapper = mount(
```
somewhatabstract: This is the action so it probably should go under `// Act`. Sometimes there isn't anything to arrange, and that's ok. Just have an empty `// Arrange` section.

packages/wonder-blocks-modal/components/one-column.modal.test.js
```diff
@@ -0,0 +1,78 @@
...
+        // Act
+        const modal = wrapper.find(LargeOneColumnModal);
+
+        // Assert
+        expect(modal);
+    });
+
+    // We already capture desktop snapshots from auto generated tests
+    describe("mobile", () => {
+        test("with footer", () => {
```
somewhatabstract: Although there's only one test case for this variation right now, it can be useful to use `describe` to layout general scenarios and then have cases within that.

Like `describe("with footer")` then have a test case like `it("should match snapshot")` or  `test("it matches snapshot")`.

The nice thing about that is that we then document the circumstances and the expectation, rather than just the circumstances as in the code as it stands.

What do you think?

packages/wonder-blocks-modal/components/standard-modal.js
```diff
@@ -71,69 +88,65 @@ export default class StandardModal extends React.Component<Props> {
...
-                            //(mediaSize) =>
-                            //    mediaSize === "small" && styles.smallPreview
-                        ]}
-                        content={
-                            <ModalContent style={styles.previewContent}>
-                                {preview}
-                            </ModalContent>
-                        }
-                    />
+                        {preview && (
```
somewhatabstract: `!!preview`

packages/wonder-blocks-modal/components/standard-modal.test.js
```diff
@@ -2,14 +2,78 @@
...
+import {
+    MediaLayoutContext,
+    MEDIA_DEFAULT_SPEC,
+} from "@khanacademy/wonder-blocks-layout";
 import StandardModal from "./standard-modal.js";
+import {mount, unmountAll} from "../../../utils/testing/mount.js";
 
 describe("StandardModal", () => {
+    beforeEach(() => {
+        unmountAll();
```
somewhatabstract: Thanks for add this!

packages/wonder-blocks-modal/components/standard-modal.test.js
```diff
@@ -2,14 +2,78 @@
...
             <StandardModal title="Title" content="Content" footer="Footer" />,
         );
 
         expect(wrapper.find("ModalCloseButton").exists()).toBeFalsy();
     });
+
+    // We already capture desktop snapshots from auto generated tests
+    test("mobile", () => {
+        // Arrange
+        const wrapper = mount(
```
somewhatabstract: This is really where the action is. What do you think about moving it to `// Act`?

packages/wonder-blocks-modal/components/two-column-modal.js
```diff
@@ -34,29 +34,76 @@ type BaseProps = {|
...
+                            }
+                            scrollOverflow={false}
+                        />
+                        <ModalPanel
+                            style={styles.smallColumn}
+                            content={content}
+                            scrollOverflow={false}
+                        />
+                    </View>
+                    {footer && (
```
somewhatabstract: `!!footer`

somewhatabstract: Or as with all the other cases, consider moving to a `maybeRenderFooter` method and then just have that return null when `footer` is falsy.

packages/wonder-blocks-modal/components/two-column-modal.test.js
```diff
@@ -2,14 +2,104 @@
...
+            const wrapper = shallow(
+                <SmallTwoColumnModal
+                    content="Content"
+                    fullBleedSidebar={true}
+                    sidebar="Sidebar"
+                    footer="Footer"
+                />,
+            );
+
+            expect(wrapper).toMatchSnapshot();
```
somewhatabstract: These should still have the AAA layout, even if the first A is empty.

packages/wonder-blocks-modal/components/close-button.js
kevinbarabash: Not sure why I reversed them.  Only one is possible at a time since we throw an error if both are provided.

packages/wonder-blocks-modal/components/modal-content.js
kevinbarabash: Glad to hear it.  I hope that other devs like it too. 🤞

packages/wonder-blocks-modal/components/one-column.modal.test.js
kevinbarabash: In the case we're testing the rendering so having it under `// Act` makes sense.  I got in the habit from some of the previous tests putting the `mount` under arrange.

packages/wonder-blocks-modal/components/one-column.modal.test.js
kevinbarabash: Sounds good.

packages/wonder-blocks-modal/components/close-button.js
```diff
@@ -7,8 +7,18 @@ import type {StyleType} from "@khanacademy/wonder-blocks-core";
...
+     *
+     * Sets primary button background color to white, and secondary and
+     * tertiary button title to color.
+     */
     light?: boolean,
+
+    /** Optional click handler */
     onClick?: () => void,
+
+    /** Optional custom styles. */
```
somewhatabstract: Thanks for adding these!

packages/wonder-blocks-modal/components/modal-dialog.js
```diff
@@ -17,7 +18,7 @@ type Props = {|
 export default class ModalDialog extends React.Component<Props> {
     render() {
         const {style, children} = this.props;
-        const contextValue = {
+        const contextValue: MediaLayoutContextValue = {
```
somewhatabstract: Nice! (Not convinced we need the word `Value` on the end of that type, but I'm not holding a strong opinion on that).

packages/wonder-blocks-modal/components/modal-panel.js
```diff
@@ -78,49 +79,53 @@ export default class ModalPanel extends React.Component<Props> {
...
+                ((content: any): React.Element<typeof ModalContent>)
+            ) : (
+                <ModalContent>{content}</ModalContent>
+            );
+
+        if (mainContent) {
+            return mainContent;
+        }
+
+        return React.cloneElement(mainContent, {
```
somewhatabstract: If we get here, won't `mainContent` be `undefined` or otherwise falsy?

packages/wonder-blocks-modal/components/modal-panel.js
```diff
@@ -78,49 +79,53 @@ export default class ModalPanel extends React.Component<Props> {
...
+        const mainContent =
+            content &&
+            typeof content === "object" &&
+            content.type === ModalContent ? (
+                ((content: any): React.Element<typeof ModalContent>)
+            ) : (
+                <ModalContent>{content}</ModalContent>
+            );
+
+        if (mainContent) {
```
somewhatabstract: I think this should be `!mainContent`.

packages/wonder-blocks-modal/components/modal-panel.js
```diff
@@ -78,49 +79,53 @@ export default class ModalPanel extends React.Component<Props> {
...
-                                !!titleBar && styles.hasTitleBar,
-                                !!footer && styles.hasFooter,
-                                showCloseButton &&
-                                    !titleBar &&
-                                    styles.withCloseButton,
-                                mainContent.props.style,
-                            ],
-                        });
-                    }
+                    const mainContent = this.renderMainContent(styles);
```
somewhatabstract: Nice.

packages/wonder-blocks-modal/components/one-column.modal.test.js
```diff
@@ -61,18 +61,24 @@ describe("OneColumnModal", () => {
...
-            const wrapper = shallow(<SmallOneColumnModal content="Content" />);
+        describe("without footer", () => {
+            it("should match snapshot", () => {
+                const wrapper = shallow(
+                    <SmallOneColumnModal content="Content" />,
+                );
 
-            expect(wrapper).toMatchSnapshot();
+                expect(wrapper).toMatchSnapshot();
+            });
```
somewhatabstract: Think these should follow the AAA comment format.

packages/wonder-blocks-modal/components/modal-dialog.js
kevinbarabash: I'm not super happy about the name either but `MediaLayoutContext` is already taken. 😞

PR #374, opened 2019-02-26
Update wonder-blocks-grid to use MediaLayout

kevinbarabash:
This PR splits out the changes to wonder-blocks-grid from #311.


jeresig:

Built with commit 3ab4ecb49b6c30e5949461731632210f22a79b5f

https://deploy-preview-374--wonder-blocks.netlify.com

kevinbarabash:
Even though we're up 0.19%, there's some tests I could write.  Also, there's a bug that I noticed when play testing so I'm going to try to write a failing test case for it before trying to fix it.

kevinbarabash:
I fixed the bug from my previous comment in #372.  I'm going merge #372 in #373 and then #373 into this PR before writing the tests I was planning to write for this PR.

kevinbarabash:
I've added tests for this now.

packages/wonder-blocks-button/components/button.md
```diff
@@ -537,7 +537,7 @@ class Example extends React.Component {
     }
 
     componentWillUnmount() {
-        this.timeout.clear();
+        this.timeout && this.timeout.clear();
```
somewhatabstract: Can't wait until something like ADR-77 lands and we can have a standard way to handle all this.

packages/wonder-blocks-grid/components/cell.js
```diff
@@ -104,61 +98,70 @@ class Cell extends React.Component<Props> {
...
+                    // Now that we have the full width we can calculate the width of this
+                    // particular cell by multiplying the full width (allCellWidth) by
+                    // the ratio of this cell (cols / totalColumns). But we then need to
+                    // add back in the missing gutter widths:
+                    // (gutterWidth * (cols - 1)). This gives us to full width of
+                    // this particular cell.
+                    const calcWidth = `calc(${contentWidth} * ${cols /
+                        totalColumns} + ${gutterWidth * (cols - 1)}px)`;
+
+                    let contents = children;
```
somewhatabstract: I don't like `let` in these situations. Usually indicates the need for a method.

What do you think about a `getChildren` method that can then return different values based on this state? That way this becomes `const`.

packages/wonder-blocks-grid/components/cell.md
```diff
@@ -4,76 +4,78 @@ A row inside of a grid containing many Cells. Each cell has a column width of 1
...
+        cell: {
+            background: Color.green,
+        },
+    }),
+    large: StyleSheet.create({
+        cell: {
+            background: Color.gold,
+        },
+    }),
+};
```
somewhatabstract: Yup, I like this approach to stylesheets. Nice.

packages/wonder-blocks-grid/components/cell.md
```diff
@@ -4,76 +4,78 @@ A row inside of a grid containing many Cells. Each cell has a column width of 1
...
-		</Row>
-	</MediaLayout>
-</View>;
-```
+<MediaLayout styleSheets={styleSheets}>
+    {({styles}) => <View style={styles.background}>
+        <Row>
+            <Cell cols={1} style={styles.cell}>
+                1
+            </Cell>
```
somewhatabstract: ASIDE: It's a bit intense that these all have to be split across lines like this. Seems like it might be more readable if they were on one line. What do you think? Is this a prettier/style thing or preference?

packages/wonder-blocks-grid/components/gutter.js
```diff
@@ -1,16 +1,25 @@
...
+     * Which media should this cell be renderer on.  Defaults to all.
+     */
+    mediaQuery: MediaQuery,
+|};
 
 /**
- * Gutter is a form of [FixedWidthCell](#fixedwidthcell) whose width is set based on the size
- * of grid currently being displayed. Used for spacing out cells from each
- * other. The gutter itself doesn't hold any content, it just spaces it out.
+ * Gutter is a private component.
```
somewhatabstract: Should we specify what it is private to? Like "private to the grid package" or "private to the Grid component"?

packages/wonder-blocks-grid/components/gutter.js
```diff
@@ -1,16 +1,25 @@
...
+ *
+ * Gutter is a component whose width is set based on the size of grid currently
```
somewhatabstract: `Gutter`?

packages/wonder-blocks-grid/components/row.js
```diff
@@ -31,108 +53,89 @@ import Gutter from "./gutter.js";
...
-        if (!shouldDisplay) {
-            return null;
-        }
+                    // Don't render the row if it's been disabled at this size
+                    if (!shouldDisplay) {
+                        return null;
+                    }
 
-        let contents = children;
+                    let contents = children;
```
somewhatabstract: Same as before; rework so this can be `const`

packages/wonder-blocks-grid/components/row.js
```diff
@@ -31,108 +53,89 @@ import Gutter from "./gutter.js";
...
-        // Go through all of the contents and pre-emptively remove anything
-        // that shouldn't be rendered and insert Gutters inbetween everything
-        // that is still visible.
-        const filteredContents = [];
-        let hasVisibleCell = false;
+                    // Go through all of the contents and pre-emptively remove anything
+                    // that shouldn't be rendered and insert Gutters inbetween everything
+                    // that is still visible.
+                    const filteredContents = [];
+                    let hasVisibleCell = false;
```
somewhatabstract: Seems like this should be its own method for sure.

somewhatabstract: In fact, isn't this just a placeholder for `filteredContents.length > 0`?

packages/wonder-blocks-grid/components/row.md
```diff
@@ -1,40 +1,40 @@
...
-				<Text>FlexCell</Text>
-			</FlexCell>
-		</Row>
-	</MediaLayout>
+    <Row style={styles.row}>
+        <Cell smallCols={2} mediumCols={4} largeCols={6} style={styles.cell}>
+            <Text>Cell</Text>
+        </Cell>
+        <Cell smallCols={2} mediumCols={4} largeCols={6} style={styles.cell}>
+            <Text>Cell</Text>
```
somewhatabstract: We shouldn't be using `Text` directly, right? Only `Typography` components?

packages/wonder-blocks-grid/components/row.test.js
```diff
@@ -0,0 +1,206 @@
...
+
+describe("Row", () => {
+    beforeEach(() => {
+        unmountAll();
+    });
+
+    describe("large", () => {
+        it("should render Cells with largeCols and cols", () => {
+            // Arrange
+            const wrapper = mount(
```
somewhatabstract: This is where the action is, so it should be in the `// Act` section.

packages/wonder-blocks-grid/components/row.test.js
```diff
@@ -0,0 +1,206 @@
...
+                        </MediaLayoutContext.Provider>
+                    </div>,
+                );
+            };
+
+            // Act, Assert
+            expect(render).toThrow();
+
+            // eslint-disable-next-line no-console
+            console.error.mockRestore();
```
somewhatabstract: This should be in a `afterEach()` so that it runs even if the test fails. If `mockRestore` only exists when this has been mocked, then do a check with `typeof` before calling it.

packages/wonder-blocks-grid/components/row.test.js
```diff
@@ -0,0 +1,206 @@
...
```
somewhatabstract: `afterEach`

packages/wonder-blocks-grid/docs.md
```diff
@@ -4,66 +4,80 @@ It's sometimes easiest to just see an example of how it works, like in the follo
...
-				<View style={{textAlign: "center"}}>
-					<Text>⇠ Gutters ⇢</Text>
-				</View>
+<MediaLayout styleSheets={styleSheets}>
+    {({styles}) => <View style={styles.background}>
+        <Row>
+            <Cell smallCols={2} mediumCols={3} largeCols={4} style={styles.cell}>
+                {({cols}) => {
+                    return <View>
+                        <Text>
```
somewhatabstract: No `Text` in examples.

styleguide.config.js
```diff
@@ -74,6 +74,10 @@ module.exports = {
             name: "Grid",
             content: "packages/wonder-blocks-grid/docs.md",
             components: "packages/wonder-blocks-grid/components/*.js",
+            ignore: [
+                "packages/wonder-blocks-grid/components/fixed-width-cell.js",
+                "packages/wonder-blocks-grid/components/gutter.js",
+            ],
```
somewhatabstract: Put these in a private section for `Internal Components` like in the section above this one.

packages/wonder-blocks-grid/components/cell.md
kevinbarabash: I don't believe prettier does anything to the .md files.  I'll put this on a single line.

packages/wonder-blocks-grid/components/gutter.js
kevinbarabash: Sorry, I'm not sure what this comment is getting at.

packages/wonder-blocks-grid/components/cell.md
somewhatabstract: Ha, didn't even realize this was an `md` file when I wrote that!

packages/wonder-blocks-grid/components/row.js
kevinbarabash: yeah, I think this could be reworked to filter first and then intersperse `<Gutter>` elements in between the `<Cells>` afterwards.

packages/wonder-blocks-grid/components/row.js
```diff
@@ -77,47 +78,37 @@ export default class Row extends React.Component<Props> {
...
-                        }
-                    }
+                    const filteredContents: Array<
+                        React.Node,
+                    > = (React.Children.toArray(contents): Array<React.Node>)
+                        // Go through all of the contents and pre-emptively remove anything
+                        // that shouldn't be rendered.
+                        .filter(
+                            // Flow doesn't let us check .type on a non-null React.Node so
+                            // we have to cast it to any.
```
somewhatabstract: Hmm, seems strange that we have to do this, but not a show stopper.

packages/wonder-blocks-grid/components/row.js
```diff
@@ -77,47 +78,37 @@ export default class Row extends React.Component<Props> {
...
+                        // Intersperse Gutter elements between the cells.
+                        .reduce(
+                            (acc, elem, index) => [
+                                ...acc,
+                                elem,
+                                <Gutter key={`gutter-${index}`} />,
+                            ],
+                            [],
+                        )
+                        // Remove the extra Gutter at the end.
```
somewhatabstract: Can you elaborate on why?

packages/wonder-blocks-grid/components/row.test.js
```diff
@@ -13,11 +13,19 @@ import {mount, unmountAll} from "../../../utils/testing/mount.js";
 describe("Row", () => {
     beforeEach(() => {
         unmountAll();
+
+        // eslint-disable-next-line no-console
+        if (console.error.mockRestore) {
+            // eslint-disable-next-line no-console
+            console.error.mockRestore();
```
somewhatabstract: Could combine to `console.error.mockRestore && console.error.mockRestore()` and just have a single lint disable comment.

styleguide.config.js
```diff
@@ -74,9 +74,14 @@ module.exports = {
             name: "Grid",
             content: "packages/wonder-blocks-grid/docs.md",
             components: "packages/wonder-blocks-grid/components/*.js",
-            ignore: [
-                "packages/wonder-blocks-grid/components/fixed-width-cell.js",
-                "packages/wonder-blocks-grid/components/gutter.js",
+            sections: [
+                {
+                    name: "Internal Components",
+                    private: true,
```
somewhatabstract: 🎉

kevinbarabash:
Not sure how the merge commit worked with all the flow issues.  I'll fix those now.

PR #378, opened 2019-02-27
[feature-link] Add `openInNewTab` property to `Link` WIP

jsatk:
This is a work in progress.

First pass at this idea.

Flow is not happy. Not sure the best way to tell flow that if `openInNewTab` is defined we should not have a `skipClientNav` prop and that `skipClientNav` is therefore implicitly true?

OR

Do we want to force users to pass `skipClientNav=true` any time they pass `openInNewTab=true`?

### DIFF
--- a/packages/wonder-blocks-link/components/link-core.js
+++ b/packages/wonder-blocks-link/components/link-core.js
@@ -36,6 +36,7 @@ export default class LinkCore extends React.Component<Props> {
             kind,
             light,
             visitable,
+            openInNewTab,
             pressed,
             style,
             testId,
@@ -53,9 +54,14 @@ export default class LinkCore extends React.Component<Props> {
                 : (hovered || focused) && linkStyles.focus,
         ];
 
+        const maybeOpenInNewTab = openInNewTab ? {
+            target: "_blank",
+        } : {};
+
         const commonProps = {
             "data-test-id": testId,
             style: [defaultStyles, style],
+            ...maybeOpenInNewTab,
             ...handlers,
         };
 
--- a/packages/wonder-blocks-link/components/link.js
+++ b/packages/wonder-blocks-link/components/link.js
@@ -6,7 +6,7 @@ import {getClickableBehavior} from "@khanacademy/wonder-blocks-core";
 import type {StyleType} from "@khanacademy/wonder-blocks-core";
 import LinkCore from "./link-core.js";
 
-export type SharedProps = {|
+type BaseProps = {|
     /**
      * Text to appear on the link.
      */
@@ -45,20 +45,6 @@ export type SharedProps = {|
     testId?: string,
 
     /**
-     * Whether to avoid using client-side navigation.
-     *
-     * If the URL passed to href is local to the client-side, e.g.
-     * /math/algebra/eval-exprs, then it tries to use react-router-dom's Link
-     * component which handles the client-side navigation. You can set
-     * `skipClientNav` to true avoid using client-side nav entirely.
-     *
-     * NOTE: All URLs containing a protocol are considered external, e.g.
-     * https://khanacademy.org/math/algebra/eval-exprs will trigger a full
-     * page reload.
-     */
-    skipClientNav?: boolean,
-
-    /**
      * Custom styles.
      */
     style?: StyleType,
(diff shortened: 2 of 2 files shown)


jeresig:

Built with commit 95e0e4927de655f766c40408b945ccef091f90e5

https://deploy-preview-378--wonder-blocks.netlify.com

packages/wonder-blocks-link/components/link.js
```diff
@@ -131,6 +144,9 @@ export default class Link extends React.Component<SharedProps> {
             this.context.router,
         );
 
+        // `"_self"` is the default `target` on anchor tags.
+        const target = openInNewTab ? "_blank" : "_self";
```
kevinbarabash: TIL.  This change will result in lots of the snapshots changing.  I wonder if we should maybe do:
```
const target = openInNewTab ? "_blank" : undefined;
```

packages/wonder-blocks-link/components/link.js
```diff
@@ -89,6 +75,32 @@ export type SharedProps = {|
     onClick?: (e: SyntheticEvent<>) => void,
 |};
 
+export type SharedProps =
+    | BaseProps & {
+      /**
```
kevinbarabash: Thanks for making these mutually exclusive.  Give that `BaseProps` is an exact object type, I think we should do the same here, e.g.
```
type SharedProps = {|
   ...BaseProps,
   skipClientNav?: boolean,
|} | {|
   ...BaseProps,
   openInNewTab?: boolean,
|};
```

packages/wonder-blocks-link/components/link-core.js
```diff
@@ -9,6 +9,12 @@ import Color, {mix, fade} from "@khanacademy/wonder-blocks-color";
 import type {ClickableHandlers} from "@khanacademy/wonder-blocks-core";
 import type {SharedProps} from "./link.js";
 
+type Target =
+    | "_self"
+    | "_blank"
+    | "_parent"
+    | "_top";
```
kevinbarabash: `LinkCore` isn't exported and nothing is using some of these options.  We should probably just pass `openInNewTab` to `LinkCore` and have it decided what string to use for `target`.

packages/wonder-blocks-link/components/link.js
jsatk: If that works!

packages/wonder-blocks-link/components/link.js
jsatk: Oo! I like that better. Good call.

packages/wonder-blocks-link/components/link-core.js
jsatk: I'm fine with that. I was mostly doing this for documentation.

somewhatabstract:
Shouldn't this be done so that it also affects how `Button` works too?

jeresig:
@somewhatabstract You're right - or more accurately this should apply to ClickableBehavior so that every place that has link-like things can be optionally opened in a new window (Button, Link, ActionItem, OptionItem, IconButton).

packages/wonder-blocks-link/components/link.js
jsatk: I'm going to drop this all together given your suggestion of just letting `LinkCore` handle things.

packages/wonder-blocks-link/components/link.js
```diff
@@ -120,6 +132,7 @@ export default class Link extends React.Component<SharedProps> {
         const {
             onClick,
             href,
+            openInNewTab,
```
jsatk: Flow is mad about me pulling out both `openInNewTab` and `skipClientNav`. Not sure how to handle this?

jsatk:
> @somewhatabstract You're right - or more accurately this should apply to ClickableBehavior so that every place that has link-like things can be optionally opened in a new window (Button, Link, ActionItem, OptionItem, IconButton).

Should I drop this diff and attempt to add it to `ClickableBehavior`?

kevinbarabash:
@jsatk I think it might be easier to start a new PR.

jsatk:
Closing. Will attempt this in a new separate diff based on feedback received here.

PR #379, opened 2019-03-20
Upgrade Flow and Handle Flow Edge Cases

jeresig:
This upgrades Wonder Blocks to Flow v0.93 and in the process fixes a number of issues that came up in Webapp. There were only a couple issues that were new as a result of the upgrade, thanks to Jeff for helping to identify solutions!

Webapp-specific Flow issues fixed:

1) Text/View/Typography components were made less strict, allowing any type of "tag" and allowing for other props (this allows people to do things like make labels with `htmlFor` props, for example).
2) The return type of all callback functions were turned to `mixed` instead of `void`, allowing devs to write stuff like: `onClick: () => something && onClick()` and have it not complain.
3) A `target` Flow type was added to Link. I know that we'll be adding something to replace it at some point, but for right now there are a lot of people using it and we should probably accommodate them until we switch.
4) I added an export for the IconSize type as it's used in some places.

### DIFF
--- a/packages/wonder-blocks-core/components/text.js
+++ b/packages/wonder-blocks-core/components/text.js
@@ -2,11 +2,13 @@
 import React from "react";
 import {StyleSheet} from "aphrodite";
 
-import type {StyleType} from "@khanacademy/wonder-blocks-core";
 import {processStyleList} from "../util/util.js";
 
 import type {TextViewSharedProps} from "../util/types.js";
 
+// NOTE(jeresig): We want to leave the props for these open so that we can
+// handle uncommon props for elements (e.g. htmlFor for labels).
+// eslint-disable-next-line flowtype/require-exact-type
 type Props = {
     ...TextViewSharedProps,
     tag: string,
--- a/packages/wonder-blocks-core/components/text.js
+++ b/packages/wonder-blocks-core/components/text.js
@@ -5,13 +5,12 @@ import {StyleSheet} from "aphrodite";
 import type {StyleType} from "@khanacademy/wonder-blocks-core";
 import {processStyleList} from "../util/util.js";
 
-import type {TextTag, TextViewSharedProps} from "../util/types.js";
+import type {TextViewSharedProps} from "../util/types.js";
 
-type Props = {|
+type Props = {
     ...TextViewSharedProps,
-    style?: StyleType,
-    tag: TextTag,
-|};
+    tag: string,
+};
 
 const isHeaderRegex = /^h[1-6]$/;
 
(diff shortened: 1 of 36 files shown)


jeresig:

Built with commit 189094f7338a14a8a57a2a6224e9cc49b96a2699

https://deploy-preview-379--wonder-blocks.netlify.com

packages/wonder-blocks-core/components/text.js
```diff
@@ -2,16 +2,17 @@
...
-import type {StyleType} from "@khanacademy/wonder-blocks-core";
 import {processStyleList} from "../util/util.js";
 
-import type {TextTag, TextViewSharedProps} from "../util/types.js";
+import type {TextViewSharedProps} from "../util/types.js";
 
-type Props = {|
+// NOTE(jeresig): We want to leave the props for these open so that we can
+// handle uncommon props for elements (e.g. htmlFor for labels).
+// eslint-disable-next-line flowtype/require-exact-type
```
kevinbarabash: Nice comment!

packages/wonder-blocks-core/util/types.js
```diff
@@ -135,48 +133,48 @@ export type AriaProps = {|
...
-    onDragEnd?: (e: SyntheticMouseEvent<*>) => void,
-    onDragEnter?: (e: SyntheticMouseEvent<*>) => void,
-    onDragExit?: (e: SyntheticMouseEvent<*>) => void,
-    onDragLeave?: (e: SyntheticMouseEvent<*>) => void,
-    onDragOver?: (e: SyntheticMouseEvent<*>) => void,
-    onDragStart?: (e: SyntheticMouseEvent<*>) => void,
-    onDrop?: (e: SyntheticMouseEvent<*>) => void,
+    onMouseDown?: (e: SyntheticMouseEvent<*>) => mixed,
+    onMouseUp?: (e: SyntheticMouseEvent<*>) => mixed,
+    onMouseMove?: (e: SyntheticMouseEvent<*>) => mixed,
```
kevinbarabash: Thanks for changing this.  I'll try to be more thoughtful about dev ergonomics when typing things in the future.

packages/wonder-blocks-icon/index.js
```diff
@@ -1,8 +1,8 @@
 // @flow
 import Icon from "./components/icon.js";
 import icons from "./util/icon-assets.js";
-import type {IconAsset} from "./util/icon-assets.js";
+import type {IconAsset, IconSize} from "./util/icon-assets.js";
 
-export type {IconAsset};
+export type {IconAsset, IconSize};
```
kevinbarabash: There's a couple of places in my PR's where I noticed we need to export more things than we were.  Thankfully it's an easy fix.

jeresig: That's fair! Although if people rely upon the Flow types that means that any change to them will be a breaking change. I think we should be careful about exposing as little as we possibly can 😄

PR #380, opened 2019-03-21
Fix a few more Flow issues.

jeresig:
Following up from the last PR, a couple more issues that we need to resolve to get it working in Webapp!

### DIFF
--- a/packages/wonder-blocks-button/components/button.js
+++ b/packages/wonder-blocks-button/components/button.js
@@ -81,6 +81,11 @@ export type SharedProps = {|
     href?: string,
 
     /**
+     * A target destination window for a link to open in.
+     */
+    target?: string,
+
+    /**
      * Whether to avoid using client-side navigation.
      *
      * If the URL passed to href is local to the client-side, e.g.
--- a/packages/wonder-blocks-button/index.js
+++ b/packages/wonder-blocks-button/index.js
@@ -1,4 +1,6 @@
 // @flow
 import Button from "./components/button.js";
+import type {SharedProps} from "./components/button.js";
 
+export type {SharedProps as ButtonProps};
 export {Button as default};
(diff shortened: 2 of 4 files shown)


packages/wonder-blocks-button/index.js
```diff
@@ -1,4 +1,6 @@
 // @flow
 import Button from "./components/button.js";
+import type {SharedProps} from "./components/button.js";
 
+export type {SharedProps as ButtonProps};
```
jeresig: A number of places are trying to get our Button SharedProps, which don't exist. I think providing this is a good stopgap.

jeresig:

Built with commit a2fcaa387b238a4c5c2550878374d3ae0704ca96

https://deploy-preview-380--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -15,7 +15,7 @@ import type {StyleType} from "@khanacademy/wonder-blocks-core";
 // NOTE(jeff): Here we share some code for use with PopperJS. Long term,
 // we should either contribute this code to the PopperJS component, or its
 // own non-wonder-blocks package.
-// eslint-disable-next-line import/no-restricted-paths
+// eslint-disable-next-line import/no-restricted-paths $FlowIgnoreMe
```
jeresig: lol, apparently this causes eslint to not work - I'm not sure how to cause both to apply to the next line!

packages/wonder-blocks-button/index.js
kevinbarabash: Interesting... I hadn't considered people using Props types to construct other Props types for other components.  In a future release we should try to be more consistent about exporting those and thinking about how we're naming them.

packages/wonder-blocks-button/components/button.js
```diff
@@ -80,6 +80,11 @@ export type SharedProps = {|
      */
     href?: string,
 
+    /**
+     * A target destination window for a link to open in.
+     */
+    target?: string,
```
kevinbarabash: We can change this to what #378 does later.

jeresig: Yep!

PR #382, opened 2019-03-23
Experiment with typescript by converting wonder-blocks-core to use it

kevinbarabash:
I was curious to the benefits but also the challenges of switching to TypeScript.


packages/wonder-blocks-core/util/enumerate-scroll-ancestors.test.tsx
```diff
@@ -29,7 +29,7 @@ describe("enumerateScrollAncestors", () => {
 
     test("if passed an invalid parameter, yields nothing", () => {
         // Arrange
-        const param = (("this is not an element": any): Element);
+        const param = ("this is not an element" as any) as Element;
```
kevinbarabash: Casting feels nicer to read.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.ts
```diff
@@ -31,7 +31,7 @@ const canScroll = function(node: Element) {
 // iterator property itself. $FlowFixMe
 class ScrollAncestorsIterator implements Iterator<Element> {
     done: boolean = false;
-    parentElement: ?Element;
+    parentElement: Element | null;
```
kevinbarabash: This isn't actually necessary.  I didn't realize that `Element` was a nullable type and I had to turn on strict null checks to get the safety.

packages/wonder-blocks-core/util/enumerate-scroll-ancestors.ts
```diff
@@ -46,13 +46,13 @@ class ScrollAncestorsIterator implements Iterator<Element> {
         return this;
     }
 
-    next(): IteratorResult<Element, void> {
+    next(): IteratorResult<Element> {
         if (this.done) {
-            return {done: true};
+            return {done: true, value: undefined};
```
kevinbarabash: In retrospect should probably use `// @ts-ignore` until https://github.com/Microsoft/TypeScript/issues/11375 is addressed.

packages/wonder-blocks-core/util/get-element-intersection.test.tsx
```diff
@@ -42,17 +43,17 @@ describe("getElementIntersection", () => {
                 mount(nodes);
             });
 
-            const element = (ReactDOM.findDOMNode(ref): any);
+            const element = ReactDOM.findDOMNode(ref) as any;
```
kevinbarabash: The cast here is only necessary b/c we don't do a null check.

packages/wonder-blocks-core/util/get-element-intersection.test.tsx
```diff
@@ -1,26 +1,27 @@
...
-import getElementIntersection from "./get-element-intersection.js";
+import {mount, unmountAll} from "../../../utils/testing/mount";
+import enumerateScrollAncestors from "./enumerate-scroll-ancestors";
+import getElementIntersection from "./get-element-intersection";
 
-jest.mock("./enumerate-scroll-ancestors.js", () => jest.fn(() => []));
+jest.mock("./enumerate-scroll-ancestors", () => jest.fn(() => []));
 
-function rect(top, right, bottom, left) {
+const mockedEnumerateScrollAncestors = enumerateScrollAncestors as jest.Mock;
```
kevinbarabash: This is probably something we could do in flow as well.

packages/wonder-blocks-core/util/types.ts
```diff
@@ -87,26 +89,26 @@ type roles =
...
     "aria-busy"?: "true" | "false",
-    "aria-checked"?: "false" | "mixed" | "true" | "undefined",
+    "aria-checked"?: "false" | "mixed" | "true" | boolean,
     "aria-controls"?: IdRefList,
     "aria-describedat"?: string, // URI
     "aria-describedby"?: IdRefList,
     "aria-disabled"?: "false" | "true",
     "aria-dropeffect"?: "copy" | "execute" | "link" | "move" | "none" | "popup",
-    "aria-expanded"?: "false" | "true" | "undefined",
+    "aria-expanded"?: "false" | "true" | boolean,
```
kevinbarabash: These changes were to bring these type definitions into line with TypeScript's aria type definitions.  I wonder if there's a way we just reuse those type definitions instead of creating our own.  🤔

packages/wonder-blocks-core/util/types.ts
```diff
@@ -119,72 +121,68 @@ export type AriaProps = {|
     "aria-orientation"?: "horizontal" | "vertical",
     "aria-owns"?: IdRefList,
     "aria-posinset"?: number, // integer
-    "aria-pressed"?: "false" | "mixed" | "true" | "undefined",
+    "aria-pressed"?: "false" | "mixed" | "true" | boolean,
     "aria-readonly"?: "false" | "true",
-    "aria-relevant"?: void, // Not using aria-relevant is a best practice
+    "aria-relevant"?: "text" | "all" | "additions" | "additions text" | "removals", // Not using aria-relevant is a best practice
```
kevinbarabash: This was `void` before to prevent people from using this prop.  We could instead use https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-component-props.md to forbid the use of this prop.

packages/wonder-blocks-core/util/types.ts
```diff
@@ -215,11 +213,7 @@ export type TextViewSharedProps = {
 
     // Used by tooltip bubble
     "data-placement"?: string,
-
-    ...AriaProps,
-
-    ...EventHandlers,
-};
+} & AriaProps & EventHandlers;
```
kevinbarabash: There's no spread operator for types.  This should be okay in most cases.  The only situation where we'd have to do something different is if we want to replace `x: number` with `x: string`.  Replacing `x: any` with `x: string` is fine as is replacing `x: string` with `x: "foo"`.

packages/wonder-blocks-core/util/util.ts
```diff
@@ -1,8 +1,7 @@
 // @flow
-import {StyleSheet, css} from "aphrodite";
-import type {CSSProperties} from "aphrodite";
+import {StyleSheet, CSSProperties, css} from "aphrodite";
```
kevinbarabash: I like the fact that there isn't separate imports for types.

packages/wonder-blocks-core/util/types.ts
```diff
@@ -1,14 +1,16 @@
...
-import * as React from "react";
+import React, {MouseEvent, KeyboardEvent, TouchEvent, FormEvent, FocusEvent, ReactNode} from "react";
 
-import type {CSSProperties} from "aphrodite";
+import {OpenCSSProperties} from "aphrodite";
 
-type NestedArray<T> = $ReadOnlyArray<T | NestedArray<T>>;
 type Falsy = false | 0 | null | void;
+
+interface NestedArray<T> extends ReadonlyArray<T | NestedArray<T>> { };
```
kevinbarabash: TypeScript requires the use of interfaces when writing recursive types, see https://github.com/Microsoft/TypeScript/issues/3496#issuecomment-128553540.

kevinbarabash:
Closing since this isn't a real PR.  It did provide some useful in about what kind of changes are necessary.

PR #383, opened 2019-03-26
add -layout as a dep of -modal, simplify creation of MediaQueryLists and addition/removal of listeners

kevinbarabash:
This is in response to a bug I noticed yesterday when trying to update wonder-blocks in webapp.  From slack:
> The behavior I’m seeing is bizarre and I have yet to explain it, but when I click anywhere on the modal, it resizes the modal once, and if I click it again, it gets small, then if I click it stops and I can actually click the cancel button.  This only happens when in its mobile size.  I’m unable to reproduce this on wonder-blocks.netlify.com.  It feels like it has something to with MediaLayout re-rendering a child when it shouldn’t though I can’t say for sure and I don’t know how clicking would trigger this.

It turns out there were two copies of wonder-blocks-layout in code.  Adding `wonder-blocks-layout` as a dep of `wonder-blocks-modal` will remove the additional copy b/c our build setup will make it as external and not include it in `wonder-blocks-modal`.  

In debugging the issue I only noticed that one copy of `MediaLayout` was being used so I don't think having two copies is what was causing the issues I observed.  This is why I decided to make some additional changes to simplify the creation of MediaListQueries and addition/removal of listeners on those query list objects.

To make sure this works, I'm going to try using `yarn link` and see if that works otherwise I might have to publish to test this inside of webapp.

### DIFF
--- a/packages/wonder-blocks-layout/components/media-layout.js
+++ b/packages/wonder-blocks-layout/components/media-layout.js
@@ -6,22 +6,19 @@ import type {StyleType} from "@khanacademy/wonder-blocks-core";
 import MediaLayoutContext from "./media-layout-context.js";
 import type {MediaSize, MediaSpec} from "../util/types.js";
 import type {Context} from "./media-layout-context.js";
+import {
+    MEDIA_DEFAULT_SPEC,
+    MEDIA_INTERNAL_SPEC,
+    MEDIA_MODAL_SPEC,
+} from "../util/specs.js";
 
 const queries = [
-    // normal queries
-    "(max-width: 767px)", // small
-    "(min-width: 768px) and (max-width: 1023px)", // medium
-    "(min-width: 1024px)", // large
-
-    // internal tools
-    "(min-width: 1px)", // large
-
-    // modal queries
-    "(max-width: 767px)", // small
-    "(min-width: 768px)", // large
+    ...Object.values(MEDIA_DEFAULT_SPEC).map((spec: any) => spec.query),
+    ...Object.values(MEDIA_INTERNAL_SPEC).map((spec: any) => spec.query),
+    ...Object.values(MEDIA_MODAL_SPEC).map((spec: any) => spec.query),
 ];
 
-const MEDIA_QUERY_LISTS: {[key: string]: MediaQueryList} = {};
+const mediaQueryLists: {[key: string]: MediaQueryList} = {};
 
 // eslint-disable-next-line flowtype/require-exact-type
 export type MockStyleSheet = {
@@ -76,6 +73,8 @@ type CombinedProps = {|...Props, ...Context|};
  * `MEDIA_DEFAULT_SPEC`.  See media-layout-context.js for additiional options.
  */
 class MediaLayoutInternal extends React.Component<CombinedProps, State> {
+    // A collection of thunks that's used to clean up event listeners
+    // when the component is unmounted.
     cleanupThunks: Array<() => void>;
 
     constructor(props: CombinedProps) {
@@ -93,7 +92,7 @@ class MediaLayoutInternal extends React.Component<CombinedProps, State> {
         >);
 
         for (const [size, spec] of entries) {
-            const mql = MEDIA_QUERY_LISTS[spec.query];
+            const mql = mediaQueryLists[spec.query];
             // during SSR there are no MediaQueryLists
             if (!mql) {
                 continue;
@@ -125,7 +124,7 @@ class MediaLayoutInternal extends React.Component<CombinedProps, State> {
             >);
 
             for (const [size, spec] of entries) {
-                const mql = MEDIA_QUERY_LISTS[spec.query];
+                const mql = mediaQueryLists[spec.query];
                 if (mql.matches) {
                     return size;
                 }
@@ -206,9 +205,9 @@ class MediaLayoutInternal extends React.Component<CombinedProps, State> {
         // to query whether any of them match.
         if (!this.isServerSide()) {
             for (const query of queries.filter(
-                (query) => !MEDIA_QUERY_LISTS[query],
+                (query) => !mediaQueryLists[query],
             )) {
-                MEDIA_QUERY_LISTS[query] = window.matchMedia(query);
+                mediaQueryLists[query] = window.matchMedia(query);
             }
         }
 
--- a/packages/wonder-blocks-layout/components/media-layout.js
+++ b/packages/wonder-blocks-layout/components/media-layout.js
@@ -88,9 +88,11 @@ class MediaLayoutInternal extends React.Component<CombinedProps, State> {
 
     componentDidMount() {
         // TODO(WB-534): handle changes to mediaSpec prop
-        for (const [size, spec] of ((Object.entries(
-            this.props.mediaSpec,
-        ): any): Array<[MediaSize, {query: string}]>)) {
+        const entries = ((Object.entries(this.props.mediaSpec): any): Array<
+            [MediaSize, {query: string}],
+        >);
+
+        for (const [size, spec] of entries) {
             const mql = MEDIA_QUERY_LISTS[spec.query];
             // during SSR there are no MediaQueryLists
             if (!mql) {
@@ -107,7 +109,8 @@ class MediaLayoutInternal extends React.Component<CombinedProps, State> {
     }
 
     componentWillUnmount() {
-        this.cleanupThunks.map((thunk) => thunk());
+        // Remove our listeners.
+        this.cleanupThunks.forEach((cleaup) => cleaup());
     }
 
     getCurrentSize(spec: MediaSpec): MediaSize {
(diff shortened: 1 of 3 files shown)


jeresig:

Built with commit 3869caa7c875dd6916bac7f1beaacdb3b45f90c4

https://deploy-preview-383--wonder-blocks.netlify.com

kevinbarabash:
I tested this in webapp using `yarn link`.  Here are screenshots from the two modals I was seeing weird behavior:

*ConfirmRemoveModal* from course-tree-action-menu.jsx:
Screen Shot 2019-03-26 at 5 08 12 PM
Screen Shot 2019-03-26 at 5 08 34 PM

*ClearFlagsModal*
Screen Shot 2019-03-26 at 5 08 47 PM
Screen Shot 2019-03-26 at 5 09 03 PM

packages/wonder-blocks-layout/components/media-layout.js
```diff
@@ -52,104 +67,67 @@ type State = {|
...
-            if (watcher && this.watchHandlers) {
-                const {watchHandlers} = this;
-                const handler = this.watchHandlers[query];
-                watcher.removeListener(handler);
-                delete watchHandlers[query];
+    componentDidMount() {
+        // TODO(WB-534): handle changes to mediaSpec prop
+        for (const [size, spec] of ((Object.entries(
+            this.props.mediaSpec,
+        ): any): Array<[MediaSize, {query: string}]>)) {
```
somewhatabstract: Could you put the result of `Object.entries` into a const before this loop, just to make this more readable. Like:

`const specs: Array<[MediaSize, {query: string}]> = Object.entries(this.props.mediaSpec);`

..should work, though you "might" have to cast the rhs to `any` as you do now.

packages/wonder-blocks-layout/components/media-layout.js
```diff
@@ -52,104 +67,67 @@ type State = {|
...
+            if (!mql) {
+                continue;
             }
+            const listener = (e) => {
+                if (e.matches) {
+                    this.setState({size});
+                }
+            };
+            mql.addListener(listener);
+            this.cleanupThunks.push(() => mql.removeListener(listener));
```
somewhatabstract: Nice.

packages/wonder-blocks-layout/components/media-layout.js
```diff
@@ -52,104 +67,67 @@ type State = {|
...
         }
     }
 
-    getCurrentSize(spec: MediaSpec) {
+    componentWillUnmount() {
+        this.cleanupThunks.map((thunk) => thunk());
```
somewhatabstract: This is great, though I think semantically, you should use `forEach` rather than `map` as the latter implies the results matter and creates an array of the results which we're then discarding.

Also, what do you think about adding a comment that says something like "Remove our listeners"? I wonder if folks reading the code will know what the thunks are.

packages/wonder-blocks-layout/components/media-layout.js
```diff
@@ -52,104 +67,67 @@ type State = {|
...
-
-            // Attach a handler that watches for the change, saving a
-            // references to it so we can remove it later
-            const handler = (this.watchHandlers[query] = (e) => {
-                if (e.matches) {
-                    this.setState({size});
+        } else {
+            for (const [size, spec] of ((Object.entries(
+                this.props.mediaSpec,
+            ): any): Array<[MediaSize, {query: string}]>)) {
```
somewhatabstract: Again, I'd like to see this split into a const and the loop to make the code and typing a little easier on the eye.

packages/wonder-blocks-layout/components/media-layout.js
```diff
@@ -215,8 +194,16 @@ export default class MediaLayout extends React.Component<Props, State> {
...
 
-    renderChildren({overrideSize, ssrSize, mediaSpec}: Context) {
-        const {children} = this.props;
+    render() {
+        const {children, mediaSpec, ssrSize, overrideSize} = this.props;
+
+        // We need to create the MediaQueryLists during the first render in order
+        // to query whether any of them match.
+        for (const query of queries) {
+            if (!MEDIA_QUERY_LISTS[query] && !this.isServerSide()) {
```
somewhatabstract: NIT: Hmmm...would it look neater if this were a `.filter` on the `queries`?

somewhatabstract: Also, is it cheaper to check `isServerSide` first and then check the array?

somewhatabstract: In fact, why run the loop at all if `isServerSide` is `true`?

packages/wonder-blocks-layout/components/media-layout.js
```diff
@@ -117,9 +120,11 @@ class MediaLayoutInternal extends React.Component<CombinedProps, State> {
         if (this.state.size) {
             return this.state.size;
         } else {
-            for (const [size, spec] of ((Object.entries(
-                this.props.mediaSpec,
-            ): any): Array<[MediaSize, {query: string}]>)) {
+            const entries = ((Object.entries(this.props.mediaSpec): any): Array<
+                [MediaSize, {query: string}],
+            >);
```
somewhatabstract: NIT: I think it's better to define the type of `entries` as this is setting an expectation rather than allowing it to be inferred from the assignment. (i.e. move the type off the end of the second cast and use it to type the variable being assigned.

packages/wonder-blocks-layout/components/media-layout.js
```diff
@@ -4,10 +4,25 @@ import type {StyleDeclaration} from "aphrodite";
 import type {StyleType} from "@khanacademy/wonder-blocks-core";
 
 import MediaLayoutContext from "./media-layout-context.js";
-import {VALID_MEDIA_SIZES} from "../util/specs.js";
 import type {MediaSize, MediaSpec} from "../util/types.js";
 import type {Context} from "./media-layout-context.js";
 
+const queries = [
```
jeresig: Hmm - I don't like this duplication of constants from the `specs.js` file. I would prefer that you do something like:

```
const queries: Array<string> = [
	...Object.values(MEDIA_DEFAULT_SPEC),
	...Object.values(MEDIA_INTERNAL_SPEC),
	...Object.values(MEDIA_MODAL_SPEC),
].map(spec => spec.query);
```

packages/wonder-blocks-layout/components/media-layout.js
```diff
@@ -4,10 +4,25 @@ import type {StyleDeclaration} from "aphrodite";
...
+
+    // internal tools
+    "(min-width: 1px)", // large
+
+    // modal queries
+    "(max-width: 767px)", // small
+    "(min-width: 768px)", // large
+];
+
+const MEDIA_QUERY_LISTS: {[key: string]: MediaQueryList} = {};
```
jeresig: Hmm - an all-caps constant implies to me that we expect this to be used elsewhere, but this really instant a constant, it's just a variable, so probably `const mediaQueryLists: ...` is fine.

packages/wonder-blocks-layout/components/media-layout.js
```diff
@@ -52,104 +67,72 @@ type State = {|
...
-export default class MediaLayout extends React.Component<Props, State> {
-    watchHandlers: ?{
-        [query: string]: any,
-    };
-
-    static WATCHERS: {
-        [query: string]: any,
-    } = {};
+class MediaLayoutInternal extends React.Component<CombinedProps, State> {
+    cleanupThunks: Array<() => void>;
```
jeresig: Can you add a comment explaining this variable? It's not super obvious on first glance.

packages/wonder-blocks-layout/components/media-layout.js
kevinbarabash: I usually think about whether a variable exists at the top level (as opposed to whether it's being exported) when deciding whether to use all-caps.

kevinbarabash: That doesn't really explain why `queries` is lowercase above it though.  I'm not very consistent. 😅

packages/wonder-blocks-layout/components/media-layout.js
kevinbarabash: I ended up having to do:
```
const queries = [
    ...((Object.values(MEDIA_DEFAULT_SPEC): any): Array<string>),
	...((Object.values(MEDIA_INTERNAL_SPEC): any): Array<string>),
	...((Object.values(MEDIA_MODAL_SPEC): any): Array<string>),
];
```
😞

kevinbarabash: That wasn't right either.. but this is:
```
const queries = [
    ...Object.values(MEDIA_DEFAULT_SPEC).map((spec: any) => spec.query),
    ...Object.values(MEDIA_INTERNAL_SPEC).map((spec: any) => spec.query),
    ...Object.values(MEDIA_MODAL_SPEC).map((spec: any) => spec.query),
];
```

PR #384, opened 2019-04-01
[feature-modal]: Added style prop to modals

erikrestificar:
….  This allows users to customize the style of the modal's ModalDialog

### DIFF
--- a/packages/wonder-blocks-modal/components/two-column-modal/__snapshots__/two-column-modal.test.js.snap
+++ b/packages/wonder-blocks-modal/components/two-column-modal/__snapshots__/two-column-modal.test.js.snap
@@ -3,13 +3,16 @@
 exports[`TwoColumnModal mobile with footer should match snapshot 1`] = `
 <ModalDialog
   style={
-    Object {
-      "_definition": Object {
-        "background": "#ffffff",
+    Array [
+      Object {
+        "_definition": Object {
+          "background": "#ffffff",
+        },
+        "_len": 24,
+        "_name": "smallDialog_144oaus",
       },
-      "_len": 24,
-      "_name": "smallDialog_144oaus",
-    }
+      undefined,
+    ]
   }
 >
   <View
@@ -117,13 +120,16 @@ exports[`TwoColumnModal mobile with footer should match snapshot 1`] = `
 exports[`TwoColumnModal mobile without footer should match snapshot 1`] = `
 <ModalDialog
   style={
-    Object {
-      "_definition": Object {
-        "background": "#ffffff",
+    Array [
+      Object {
+        "_definition": Object {
+          "background": "#ffffff",
+        },
+        "_len": 24,
+        "_name": "smallDialog_144oaus",
       },
-      "_len": 24,
-      "_name": "smallDialog_144oaus",
-    }
+      undefined,
+    ]
   }
 >
   <View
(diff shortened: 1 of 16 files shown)


jeresig:

Built with commit afb618ec3a722040913475d2f44b81f331cb43ce

https://deploy-preview-384--wonder-blocks.netlify.com

kevinbarabash:
I re-ran the circleci jobs.  I'm not sure why the lint job failed but everything's passing now.  I've update the title of this PR so that it's easy to see what kind of change this (bug fix, feature, breaking) is and which package it's changing without having to look at changes in the PR itself.  This isn't actually necessary since you know what kind of change you're making, but I'm trying to get in the habit.  I'll review the PR itself tomorrow.

packages/wonder-blocks-modal/docs.md
```diff
@@ -37,3 +37,70 @@ class ModalWrapper extends React.Component {
...
+        position: "absolute",
+        left: 0,
+        right: 0,
+        top: 0,
+        bottom: 0,
+    },
+});
+
+const styleSheets = {
+    all: StyleSheet.create({
```
kevinbarabash: I think this example should use `mdOrLarger` since currently our modals on mobile go full-screen and my guess is that people will probably want to maintain that behavior.

packages/wonder-blocks-modal/docs.md
```diff
@@ -37,3 +37,70 @@ class ModalWrapper extends React.Component {
     </ModalLauncher>
 </View>
 ```
+
+### Example: Modals with custom styles 
+
+Sometimes you'll want to customize the styling of the modal .e.g., custom width or height.  You can pass in `style` which will customize the styling of the `ModalDialog`.
```
kevinbarabash: Specifying the component name `ModalDialog` is an implementation detail that probably isn't necessary.

packages/wonder-blocks-modal/components/two-column-modal/two-column-modal.test.js
```diff
@@ -74,6 +74,35 @@ describe("TwoColumnModal", () => {
...
+                    }}
+                />
+            </MediaLayoutContext.Provider>,
+        );
+
+        // Act
+        const modal = wrapper.find(LargeTwoColumnModal);
+
+        // Assert
+        expect(modal);
```
kevinbarabash: This should also be `.toMatchSnapshot()` so that we can verify the width/height.

packages/wonder-blocks-modal/components/two-column-modal/__snapshots__/two-column-modal.test.js.snap
```diff
@@ -3,13 +3,16 @@
...
+          "background": "#ffffff",
+        },
+        "_len": 24,
+        "_name": "smallDialog_144oaus",
       },
-      "_len": 24,
-      "_name": "smallDialog_144oaus",
-    }
+      undefined,
+    ]
```
kevinbarabash: I'm not sure why we aren't printing out the style objects for these tests but that's something the Frontend Infra team can investigate independently of this PR.

packages/wonder-blocks-modal/components/two-column-modal/two-column-modal.test.js
erikrestificar: added

packages/wonder-blocks-modal/docs.md
erikrestificar: updated to modal wrapper

PR #386, opened 2019-04-10
Add support for Noto in wonder-blocks

kevinbarabash:
This PR add `Noto Sans` and `Noto Serif` font faces which are used by our typography components.  It also adds a few fonts for different character ranges to the styleguidist docs for testing purposes.


jeresig:

Built with commit 16a793bcfca0036624f916ca3b2e4430a85b9f1e

https://deploy-preview-386--wonder-blocks.netlify.com

BrianGenisio:
Some of those fonts are very large in size... like 300KB each.  How many of those will get downloaded?  I'm worried that we could be asking our RTL users to download over 1MB worth of font files.

kevinbarabash:
@briangenisio I have a TODO to convert them to .woff/.woff2.  I'm going to write a script to do this tomorrow.

packages/wonder-blocks-typography/util/styles.js
```diff
@@ -12,13 +12,15 @@ const common = {
     display: "block",
 };
 
-const LatoFamily = "Lato, sans-serif";
+const SansFamily = 'Lato, "Noto Sans", sans-serif';
+// TODO(kevinb): Use Minion Pro here
```
jeresig: Can you file a task to track this and link to it here?

static/noto.css
```diff
@@ -0,0 +1,93 @@
...
+}
+
+@font-face {
+    font-family: "Noto Sans";
+    font-weight: 900;
+    src: url("fonts/NotoSans-Black.ttf");
+    unicode-range: U+0400-04FF;
+}
+
+/* Cree */
```
jeresig: How were these languages selected? We don't have any Cree on our site (I don't think) but we do have Japanese, for example (there is also a Noto Sans for Chinese but it appears to be massive).

kevinbarabash: Cree was for fun.  I'll replace it with something else. I'm deferring CJK languages for now due to the size of the font files.

kevinbarabash:
@jeresig both of the sites you linked to don't have the Noto Sans/Serif Arabic so I'm going to have to build a tool to convert .tt to .woff for that.  I was thinking of using fonttools or maybe opentype.js for that.

kevinbarabash:
I've switched to .woff2 to but Greek and Cyrillic share the same file so I'm going to split it contain font files that only contain the correct unicode range.

jandrade:
Closing this pull request, but I will continue with the task based on this progress.

PR #387, opened 2019-04-19
[core:feature] Add Clickable component to wrap ClickableBehavior (wip)

kevinbarabash:
I've wanted to create a `Clickable` component for some time now to make `ClickableBehavior` easier to use.  This needs comments, better docs, and tests, but I thought I put it up early so that people could see what I was thinking with this wrapper.  I'd like to eventually update the existing component that use `ClickableBehavior` to use `Clickable` instead.


jeresig:

Built with commit c48efde4de07616ff1c60012062d8b745349e7b5

https://deploy-preview-387--wonder-blocks.netlify.com

packages/wonder-blocks-core/components/clickable.js
```diff
@@ -0,0 +1,141 @@
+// @flow
+import * as React from "react";
+import {StyleSheet} from "aphrodite";
```
jeresig: Before we add this to core - we should ask if we want this in core (since otherwise will require a version change later to move it). I guess if we ever wanted to move `getClickableBehavior` out of core and this into a "clickable" package we'd have to bump the version anyway.

packages/wonder-blocks-core/components/clickable.js
```diff
@@ -0,0 +1,141 @@
...
+    skipClientNav?: boolean,
+    style?: StyleType,
+    testId?: string,
+|};
+
+const StyledAnchor = addStyle<"a">("a");
+const StyledButton = addStyle<"button">("button");
+const StyledLink = addStyle<typeof Link>(Link);
+
+export default class Clickable extends React.Component<Props> {
```
jeresig: Can you add a detailed comment here explaining how this component should be used? I think you can move the comment that you currently have at the top of `clickable.md` into here and maybe expand it a bit explaining the usage and especially explaining when you should be using it and not Link or Button.

packages/wonder-blocks-core/components/clickable.js
```diff
@@ -0,0 +1,141 @@
...
+import {StyleSheet} from "aphrodite";
+
+import {Link} from "react-router-dom";
+import addStyle from "../util/add-style.js";
+import getClickableBehavior from "../util/get-clickable-behavior.js";
+import type {ClickableRole, State} from "./clickable-behavior.js";
+import type {StyleType} from "../util/types.js";
+
+type Props = {|
+    "aria-label": string,
```
jeresig: Can you add comments to these props so that they show up in the documentation? You can probably copy most of them from Link or Button.

packages/wonder-blocks-core/components/clickable.js
```diff
@@ -0,0 +1,141 @@
...
+                    } else {
+                        return (
+                            <StyledButton
+                                type="button"
+                                {...commonProps}
+                                disabled={disabled}
+                            >
+                                {content}
+                            </StyledButton>
+                        );
```
somewhatabstract: Tempted to tidy this up by having some seperate methods like `maybeRenderStyledButton`, etc. and then have them control the conditions of if they render. Or at least have `renderStyledButton` and have it extract the props it needs. I think it would make the code a little easier to follow.

packages/wonder-blocks-core/components/clickable.js
kevinbarabash: Good call.  Maybe we could create a `wonder-blocks-clickable` package.  We can move `ClickableBehavior` over that package but still have `wonder-blocks-core` export `ClickableBehavior` though I think we'd want to eventually drop that export and have webapp just use `Clickable` and have `ClickableBehavior` be completely private.

packages/wonder-blocks-core/components/clickable.js
```diff
@@ -0,0 +1,141 @@
+// @flow
+import * as React from "react";
+import {StyleSheet} from "aphrodite";
+
+import {Link} from "react-router-dom";
```
jandrade: maybe arriving a bit late to the party.... just wondering, is there a strong reason for including `react-router-dom` as a WB dependency? instead of letting `webapp` handling directly the routing? Asking these questions b/c I would like to understand the context 😃

packages/wonder-blocks-core/components/clickable.md
```diff
@@ -0,0 +1,68 @@
...
+            borderRadius: 10,
+            border: "dotted 4px black",
+        },
+    },
+});
+
+<View>
+    <Clickable>
+        {(state) => {
+            console.log(state);
```
jandrade: maybe this line could be removed?

packages/wonder-blocks-core/components/clickable.js
kevinbarabash: I wish we didn't, but we wanted to be able to use client-side routing with wonder-blocks components and being able to set `href="/foo/bar"` on components seemed a convenient way to specify routes.  I couldn't figure how to make this work without importing `Link` from `react-router-dom` within wonder-blocks. 😅

packages/wonder-blocks-clickable/components/clickable.js
```diff
@@ -0,0 +1,147 @@
...
+import {Link} from "react-router-dom";
+import type {
+    ClickableRole,
+    ClickableState,
+} from "@khanacademy/wonder-blocks-core";
+import {addStyle, getClickableBehavior} from "@khanacademy/wonder-blocks-core";
+import type {StyleType} from "../util/types.js";
+
+type Props = {|
+    children: (eventState: ClickableState) => React.Node,
```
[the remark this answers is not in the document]
kumailn: done!

packages/wonder-blocks-clickable/components/clickable.md
```diff
@@ -0,0 +1,46 @@
+### Client-Side routing with Clickable
```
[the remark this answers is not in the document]
kumailn: sure thing, will do

packages/wonder-blocks-clickable/index.js
```diff
@@ -0,0 +1,4 @@
+// @flow
+import Clickable from "./components/clickable.js";
+
+export {Clickable};
```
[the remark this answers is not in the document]
kumailn: Yeah I think it’s better if it’s default

packages/wonder-blocks-core/components/clickable.js
```diff
@@ -0,0 +1,141 @@
+// @flow
```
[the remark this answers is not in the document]
kumailn: done :)

packages/wonder-blocks-clickable/util/types.js
```diff
@@ -0,0 +1,229 @@
+// @flow
```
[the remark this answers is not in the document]
kumailn: Yes I was thinking of a way, maybe we can export these types in Core?

packages/wonder-blocks-clickable/package.json
```diff
@@ -0,0 +1,39 @@
...
+    },
+    "author": "",
+    "license": "MIT",
+    "publishConfig": {
+      "access": "public"
+    },
+    "dependencies": {
+      "@khanacademy/wonder-blocks-button": "^2.3.3",
+      "@khanacademy/wonder-blocks-color": "^1.1.2",
+      "@khanacademy/wonder-blocks-core": "^2.2.0",
```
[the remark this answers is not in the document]
kumailn: thanks for catching this, fixed

packages/wonder-blocks-clickable/docs.md
```diff
@@ -0,0 +1,6 @@
+The Clickable component allows you to easily turn any custom compoent into a clickable one with consistent behavior and accesibility features.
+
+**Clickable allows your components to:**
+- Handle mouse / touch / keyboard events
+- Match the standard behavior of the given role
+- Perform Client Side Navigation when `href` is passed and the component is a descendent of a react-router `Router`
```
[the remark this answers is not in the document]
kumailn: good idea, done :)

packages/wonder-blocks-clickable/util/types.js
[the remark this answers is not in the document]
kevinbarabash: I think some of the types are already being exported from `core`.  Export any others that you need.

packages/wonder-blocks-clickable/components/clickable.js
```diff
@@ -0,0 +1,174 @@
...
+                        role: role,
+                        style: [styles.reset, style],
+                        ...handlers,
+                    };
+
+                    const content = children(state);
+                    const activeHref = href && !disabled;
+                    const useClientRouter = router && !skipClientNav;
+
+                    if (activeHref && useClientRouter) {
```
somewhatabstract: What do you think about putting the main bulk of this inline function into a separate method or set of methods? I think it might make the code a little easier to read by reducing indentation?

packages/wonder-blocks-clickable/components/clickable.js
```diff
@@ -0,0 +1,174 @@
...
+                            <StyledLink
+                                {...commonProps}
+                                to={href}
+                                onClick={onClick}
+                                aria-disabled={disabled ? "true" : undefined}
+                            >
+                                {content}
+                            </StyledLink>
+                        );
+                    } else if (activeHref && !useClientRouter) {
```
somewhatabstract: This can just be an `if` since the previous condition returns.

packages/wonder-blocks-clickable/components/clickable.js
```diff
@@ -0,0 +1,174 @@
...
+                            <StyledAnchor
+                                {...commonProps}
+                                href={href}
+                                onClick={onClick}
+                                aria-disabled={disabled ? "true" : undefined}
+                            >
+                                {content}
+                            </StyledAnchor>
+                        );
+                    } else {
```
somewhatabstract: This else is unnecessary since both previous conditions return anyway.

packages/wonder-blocks-clickable/components/clickable.md
```diff
@@ -0,0 +1,85 @@
...
+const {View} = require("@khanacademy/wonder-blocks-core");
+const {MemoryRouter, Route, Switch} = require("react-router-dom");
+
+const styles = StyleSheet.create({
+    row: {
+        flexDirection: "row",
+        alignItems: "center",
+    },
+    h1: {
+        marginRight: "25px",
```
somewhatabstract: What do you think about using Wonder Blocks Spacing constants in our docs? Seems like this is an opportunity to encourage best practices.

packages/wonder-blocks-clickable/package.json
```diff
@@ -0,0 +1,39 @@
+{
+    "name": "@khanacademy/wonder-blocks-clickable",
+    "version": "1.0.0",
```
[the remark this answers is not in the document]
somewhatabstract: I'd err on the side of staying `private: true` and version `0.1.0`. It's easier to go from this to public release than to discover a problem at public release and have to back pedal. I prefer to plan for the worst rather than hope for the best. What do you think?

packages/wonder-blocks-clickable/package.json
```diff
@@ -0,0 +1,39 @@
...
+      "@khanacademy/wonder-blocks-typography": "^1.1.2"
+    },
+    "peerDependencies": {
+      "aphrodite": "^1.2.5",
+      "popper.js": "^1.14.1",
+      "prop-types": "^15.6.2",
+      "react": "^16.4.1",
+      "react-dom": "^16.4.1",
+      "react-popper": "^1.0.0",
+      "react-router-dom": "^4.2.2"
```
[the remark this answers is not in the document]
kumailn: thanks, done

packages/wonder-blocks-clickable/components/clickable.js
kumailn: I feel like it’s better this way for clarity, what do you think?

packages/wonder-blocks-clickable/components/clickable.md
kumailn: good idea!

packages/wonder-blocks-clickable/components/clickable.js
kumailn: I agree, I think that’d be better

packages/wonder-blocks-clickable/components/clickable.js
```diff
@@ -0,0 +1,147 @@
...
+                        );
+                    }
+                }}
+            </ClickableBehavior>
+        );
+    }
+}
+
+const styles = StyleSheet.create({
+    reset: {
```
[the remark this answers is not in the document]
kumailn: this was a part of the old Clickable component i took over

packages/wonder-blocks-clickable/components/clickable.js
kumailn: same as above

packages/wonder-blocks-clickable/components/clickable.md
kumailn: I can’t get spacing to work, when I apply it here it has to effect :/

kevinbarabash: Have a look at the uses of `wonder-blocks-spacing` in the other markdown files.

packages/wonder-blocks-clickable/util/types.js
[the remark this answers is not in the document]
kevinbarabash: This still needs to be addressed.

packages/wonder-blocks-clickable/package.json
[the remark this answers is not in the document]
kevinbarabash: This looks good.  Thanks!

packages/wonder-blocks-clickable/components/clickable.js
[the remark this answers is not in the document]
kevinbarabash: It looks like it was based on https://gist.github.com/MoOx/9137295.  Can you add a link to that file in a comment?

packages/wonder-blocks-clickable/components/clickable.md
```diff
@@ -0,0 +1,86 @@
...
+```jsx
+const {StyleSheet} = require("aphrodite");
+const {View} = require("@khanacademy/wonder-blocks-core");
+const {MemoryRouter, Route, Switch} = require("react-router-dom");
+const Color = require("@khanacademy/wonder-blocks-color").default;
+const {Body} = require("@khanacademy/wonder-blocks-typography");
+
+const styles = StyleSheet.create({
+    focused: {
+        border: "none",
```
kevinbarabash: It's hard to see this style since the unfocused state has `border: "none"` as well.

packages/wonder-blocks-clickable/components/clickable.md
```diff
@@ -0,0 +1,86 @@
...
+    pressed: {
+        color: Color.teal,
+    },
+});
+
+<View>
+    <Clickable onClick={() => alert("You clicked some text!")}>
+        {
+            eventState =>
+                <Body style={[
```
kevinbarabash: Can you wrap this in a View to make it look more like a card and change the background color of it as the eventState changes?

packages/wonder-blocks-clickable/components/clickable.md
kumailn: i’ll just remove it

packages/wonder-blocks-clickable/components/clickable.js
```diff
@@ -113,23 +116,22 @@ export default class Clickable extends React.Component<Props> {
         );
 
         return (
-            <ClickableBehavior>
-                {(state, handlers) =>
+            <ClickableBehavior onClick={this.props.onClick}>
+                {(state) =>
```
kevinbarabash: Not passing the handlers means that the state won't get update correctly.

packages/wonder-blocks-clickable/components/clickable.js
```diff
@@ -18,12 +18,12 @@ type Props = {|
     children: (eventState: ClickableState) => React.Node,
 
     /**
-     * An onClick function which Clickable can execure
+     * An onClick function which Clickable can execute when clicked
```
kevinbarabash: Thanks for fixing my typos.

packages/wonder-blocks-clickable/package.json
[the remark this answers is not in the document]
jandrade: +1 for `private: true` and adding `0.1.0`. By adding this initial version, lerna will try to publish `1.0.0` instead of `1.0.1` (or something similar). Nothing that impacts at all, but good for keeping semver consistency.

packages/wonder-blocks-clickable/util/types.js
```diff
@@ -0,0 +1,229 @@
...
+    // abstract role
+    | "window";
+
+type IdRef = string;
+type IdRefList = IdRef | Array<IdRef>;
+
+// Source: https://www.w3.org/WAI/PF/aria-1.1/states_and_properties
+// TODO(kevinb): convert to disjoint union of exact object types keyed on role
+// eslint-disable-next-line flowtype/require-exact-type
+export type AriaProps = {|
```
jandrade: e.g. If you needed it, you shouldn't need to export `AriaProps` here. Instead, you could import directly from `wonderblocks-core`.

packages/wonder-blocks-clickable/components/clickable.js
```diff
@@ -0,0 +1,168 @@
...
+     * Test ID used for e2e testing.
+     */
+    testId?: string,
+|};
+
+const StyledAnchor = addStyle<"a">("a");
+const StyledButton = addStyle<"button">("button");
+const StyledLink = addStyle<typeof Link>(Link);
+
+export default class Clickable extends React.Component<Props> {
```
jandrade: It would be great if you could add a comment here explaining this component.

packages/wonder-blocks-clickable/components/clickable.js
```diff
@@ -0,0 +1,168 @@
...
+        );
+
+        return (
+            <ClickableBehavior onClick={this.props.onClick}>
+                {(state, handlers) =>
+                    this.getCorrectTag(state, {
+                        // eslint-disable-next-line react/prop-types
+                        "aria-label": this.props["aria-label"],
+                        "data-test-id": this.props.testId,
+                        role: this.props.role,
```
jandrade: This `role` prop shouldn't be part of `commonProps` b/c `<button>` doesn't need to use this role.

More info here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#Best_practices

packages/wonder-blocks-clickable/util/types.js
kumailn: it actually wasn’t being used at all, I removed it

packages/wonder-blocks-clickable/components/clickable.js
kumailn: thanks for catching that!

packages/wonder-blocks-clickable/components/clickable.test.js
```diff
@@ -0,0 +1,114 @@
...
+        // Act
+        const buttonWrapper = wrapper.find(`[data-test-id="button"]`).first();
+        buttonWrapper.simulate("click", {button: 0});
+
+        // Assert
+        expect(wrapper.find("#foo").exists()).toBe(false);
+    });
+
+    test("disallow navigation when href and disabled are both set", () => {
+        const wrapper = mount(
```
somewhatabstract: NIT: Missing `// Arrange`

kumailn: good catch :)

PR #388, opened 2019-04-23
Add onOpen and onClose callbacks to dropdowns

matchu:
I'm working on a feature that I think requires knowing whether the dropdown is open or closed. So, I tried adding these `onOpen` and `onClose` callbacks. What do you think?

(Specifically, I'm building filters for the district dashboards, and we're tracking which filter values are selected in the URL's query string. If the user makes multiple changes to the dropdown in a row, i.e. they check a bunch of boxes at once, I want to treat that as an atomic action and use `replaceState`, to avoid creating a new history entry for each individual box they checked. But, once the user closes the dropdown, I want to "finalize" their input and make a permanent history entry for it. I can talk through this in more detail as needed! This isn't super-urgent though; for now, I'll just use `pushState` and accept the worser history experience.)

### DIFF
--- a/packages/wonder-blocks-dropdown/components/multi-select.test.js
+++ b/packages/wonder-blocks-dropdown/components/multi-select.test.js
@@ -15,8 +15,13 @@ describe("MultiSelect", () => {
         allChanges.push(update);
     };
     const onChange = jest.fn();
+    const onOpen = jest.fn();
+    const onClose = jest.fn();
 
     beforeEach(() => {
+        onChange.mockClear();
+        onOpen.mockClear();
+        onClose.mockClear();
         window.scrollTo = jest.fn();
         select = mount(
             <MultiSelect
@@ -24,6 +29,8 @@ describe("MultiSelect", () => {
                     saveUpdate(selectedValues);
                     onChange();
                 }}
+                onOpen={onOpen}
+                onClose={onClose}
                 placeholder="Choose"
                 selectItemType="students"
                 selectedValues={["2"]}
@@ -41,7 +48,7 @@ describe("MultiSelect", () => {
         unmountAll();
     });
 
-    it("closes/opens the select on mouse click, space, and enter", () => {
+    it("closes/opens the select on mouse click", () => {
         const opener = select.find(SelectOpener);
         expect(select.state("open")).toEqual(false);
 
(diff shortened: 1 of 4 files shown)


matchu:
^ hm, looks like the CI failure is just the Flow server being missing? 🤔

jeresig:

Built with commit d741048cda3e5f5142932a2e696056674ab6d0c2

https://deploy-preview-388--wonder-blocks.netlify.com

kevinbarabash:
@matchu thanks for the PR.  This seems like a reasonable addition to the API.  I'm not sure why flow isn't working on CirceCI.  I re-ran the job but that didn't help. :(

packages/wonder-blocks-dropdown/components/multi-select.test.js
```diff
@@ -50,16 +57,76 @@ describe("MultiSelect", () => {
         opener.simulate("mouseup");
         opener.simulate("click");
         expect(select.state("open")).toEqual(true);
+        expect(onOpen).toHaveBeenCalledTimes(1);
+        expect(onClose).toHaveBeenCalledTimes(0);
+
+        // Close select with mouse
```
kevinbarabash: This feels like a separate test case.

packages/wonder-blocks-dropdown/components/multi-select.test.js
```diff
@@ -50,16 +57,76 @@ describe("MultiSelect", () => {
...
+        opener.simulate("mousedown");
+        opener.simulate("mouseup");
+        opener.simulate("click");
+        expect(select.state("open")).toEqual(false);
+        expect(onOpen).toHaveBeenCalledTimes(1);
+        expect(onClose).toHaveBeenCalledTimes(1);
+
+        // Open select with mouse
```
kevinbarabash: This should probably be a separate test case, something along the line of `it("should re-open when clicking after closing", () => {})` or something like that.

packages/wonder-blocks-dropdown/components/multi-select.test.js
somewhatabstract: Try to keep each test case to a single expectation.

For example,

`opens the select on mouse click`
`opens the select on space`
`opens the select on enter`
`calls onOpen when opened`
`calls OnClose when closed`

It makes debugging test failures and correcting behaviour much easier that way.

somewhatabstract: I don't think it's necessary to check the callbacks for al the different ways the component can be opened and closed.

packages/wonder-blocks-dropdown/components/single-select.test.js
```diff
@@ -36,16 +46,76 @@ describe("SingleSelect", () => {
...
+        expect(onClose).toHaveBeenCalledTimes(1);
 
-        // Should open with enter
+        // Open select with enter
         opener.simulate("keydown", {keyCode: keyCodes.enter});
         opener.simulate("keyup", {keyCode: keyCodes.enter});
         expect(select.state("open")).toEqual(true);
+        expect(onOpen).toHaveBeenCalledTimes(2);
+        expect(onClose).toHaveBeenCalledTimes(1);
     });
```
somewhatabstract: Please use Arrange/Act/Assert layout for the tests.

I know these weren't before, but since you're touching them, that would be a fantastic improvement in line with other tests in Wonder Blocks.

packages/wonder-blocks-dropdown/components/single-select.js
```diff
@@ -24,6 +24,16 @@ type Props = {|
...
 
+    /**
+     * Callback for when the dropdown is opened.
+     */
+    onOpen?: () => mixed,
+
+    /**
+     * Callback for when the dropdown is closed.
+     */
+    onClose?: () => mixed,
```
somewhatabstract: Rather than optional, what do you think about providing a default noop function?

packages/wonder-blocks-dropdown/components/single-select.js
```diff
@@ -111,6 +121,12 @@ export default class SingleSelect extends React.Component<Props, State> {
             open,
             keyboard,
         });
+
+        if (open) {
+            this.props.onOpen && this.props.onOpen();
```
somewhatabstract: NIT: If these defaulted to a noop function, this code would be simpler to read.

packages/wonder-blocks-dropdown/components/single-select.js
```diff
@@ -111,6 +121,12 @@ export default class SingleSelect extends React.Component<Props, State> {
             open,
             keyboard,
         });
+
+        if (open) {
```
somewhatabstract: This is a little confusing (my mind things, if we're open, we're about to close). Can you add a comment explaining that we just opened, so we need to tell folks we did (and vice versa for the close?

kevinbarabash:
I re-ran the CircleCI process and lint (and flow) is passing now.

matchu:
(btw the reqs on this aren't actually what I thought, so this PR is no
longer necessary for my webapp change 😅 I'll backburner it for a bit, if
that's okay.)

kevinbarabash:
@matchu sounds good.  I'm going to close this for now.  Please re-open if things change.

PR #390, opened 2019-04-26
WB-554: update components for use with react-hot-loader

kevinbarabash:
The issue is that react-hot-loader wraps components in ReactProxy classes which breaks direct component class checks.

Test Plan:
- use yarn link to test these components in webapp
- load demos.jsx.fixture.js in the react sandbox
- verify that all of the dropdowns and the modals work (the demos fixture doesn't have a grid example yet)

Aside: I really wanted to use `Symbol` so I could copy/paste the `isClassOf` snippet without having to modify it each time, but flow doesn't support computed static vars. 😞

### DIFF
--- a/packages/wonder-blocks-dropdown/components/action-item.js
+++ b/packages/wonder-blocks-dropdown/components/action-item.js
@@ -94,6 +94,10 @@ export default class ActionItem extends React.Component<ActionProps> {
     };
 
     static contextTypes = {router: PropTypes.any};
+    static __IS_ACTION_ITEM__ = true;
+    static isClassOf(instance: React.Element<any>) {
+        return instance && instance.type && instance.type.__IS_ACTION_ITEM__;
+    }
 
     render() {
         const {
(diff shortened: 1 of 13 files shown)


jeresig:

Built with commit eb5e8061df21d33dc663d1d89632e19e47d26e1f

https://deploy-preview-390--wonder-blocks.netlify.com

somewhatabstract:
> Aside: I really wanted to use Symbol so I could copy/paste the isClassOf snippet without having to modify it each time, but flow doesn't support computed static vars. 😞

I wonder if they plan to or if this is a WON'T FIX.

somewhatabstract:
I found this and added a comment (mainly to revive it a bit - it hasn't been commented on since 2017 until today).

https://github.com/facebook/flow/issues/2286

jandrade:
Are these changes going to result in a new patch version for each component (e.g. v1.1.0 -> v1.1.1)? If that's the case, is that a manual change? or do we have a script that does that?

kevinbarabash:
@jandrade running `yarn run publish` runs `lerna` which provides a prompt to bump things.  I'll update the title of this diff to indicate which package should be bumped.

kevinbarabash:
I should've changed the title before merging, d'oh.

PR #391, opened 2019-04-29
WB-553: Implement Breadcrumb v1

jandrade:
Implemented Breadcrumb v1:
- Added unit tests.
- Included docs and examples.

breadcrumb-docs


jeresig:

Built with commit 12dbb91873dac26a74845a362b8a06d240259ec9

https://deploy-preview-391--wonder-blocks.netlify.com

flow-typed/aphrodite.flow.js
```diff
@@ -334,6 +334,17 @@ type _CSSProperties = {|
     ":disabled"?: _CSSProperties,
     "::-moz-focus-inner"?: _CSSProperties,
 
+    // Filters
```
jandrade: Any thoughts about including this into the valid style properties?

The main reason why I did this was to handle the separators using `:last-child` and prevent any extra logic in JS.

packages/wonder-blocks-breadcrumb/components/breadcrumb.test.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+import * as React from "react";
+import {shallow} from "enzyme";
+
+import Breadcrumb from "./breadcrumb.js";
+import BreadcrumbItem from "./breadcrumb-item.js";
+
+describe("Breadcrumb", () => {
+    it("should render a Breadcrumb", () => {
```
jandrade: Not sure about adding this test here. Might be a bit unnecessary and it's covered by the auto-generated snapshot test.

flow-typed/aphrodite.flow.js
jeresig: I think I'd prefer that we not do this. Will comment more below...

packages/wonder-blocks-breadcrumb/components/breadcrumb-item.js
```diff
@@ -0,0 +1,49 @@
...
+            </StyledListItem>
+        );
+    }
+}
+
+const styles = StyleSheet.create({
+    item: {
+        display: "inline-block",
+        margin: `0 ${Spacing.xxSmall}`,
+        "::after": {
```
jeresig: Hmm... so... I think I'd prefer that we insert the spacers in-between the BreadcrumbItems inside the Breadcrumb component itself, rather than using CSS. It allows us to have greater control over the display of the spacers and also choose to show/hide them from screenreaders (probably hide them - by default Chrome and Firefox don't hide CSS content from screenreaders so having these spacer characters seems not ideal).

packages/wonder-blocks-breadcrumb/components/breadcrumb-item.md
```diff
@@ -0,0 +1,17 @@
+### breadcrumb item (with links):
```
jeresig: I'm not sure that these are examples that we actually want to show - we don't really want people using BreadcrumbItem outside of a Breadcrumb. I think we should only show it in examples for Breadcrumb itself.

packages/wonder-blocks-breadcrumb/components/breadcrumb.js
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * This is the content for the collection of Breadcrumbs
+     */
+    children: React.ChildrenArray<React.Element<typeof BreadcrumbItem>>,
+|};
+
+const StyledList = addStyle("ol");
+
+/**
```
jeresig: Can you expand this comment and start with an explanation of what a Breadcrumb is and where we expect people to use it? That way when people read the docs they'll have a bit more context.

packages/wonder-blocks-breadcrumb/components/breadcrumb.md
```diff
@@ -0,0 +1,10 @@
+### breadcrumb example (with links):
```
jeresig: Would be nice to show one here that's a mix of links and plaintext, too.

packages/wonder-blocks-breadcrumb/docs.md
```diff
@@ -0,0 +1 @@
+A breadcrumb trail consists of a list of links to the parent pages of the current page in hierarchical order. It helps users find their place within a website or web application.
```
jeresig: I think just moving this into the comment for Breadcrumb will achieve what I mentioned above.

packages/wonder-blocks-breadcrumb/index.js
```diff
@@ -0,0 +1,6 @@
+// @flow
+import Breadcrumb from "./components/breadcrumb.js";
+
```
jeresig: Can remove this empty line.

packages/wonder-blocks-breadcrumb/index.js
```diff
@@ -0,0 +1,6 @@
...
+import BreadcrumbItem from "./components/breadcrumb-item.js";
+
+export {Breadcrumb as default, BreadcrumbItem};
```
jeresig: I'm not actually sure we want Breadcrumb to be exported as default here. If we do this then devs will have to do: `import Breadcrumb, {BreadcrumbItem} from "...";` every time they use it. I think we should instead just export them both by name.

packages/wonder-blocks-breadcrumb/components/breadcrumb.md
jeresig: I think this example can be moved into breadcrumb/docs.md and you can get rid of these per-component examples as people will mostly just be interested in how to use this at a full component level.

packages/wonder-blocks-breadcrumb/components/breadcrumb.js
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * 1. Breadcrumb Item: Represents a section within the page.
+ * 2. Separator: Adds a separator between each item.
+ */
+export default class Breadcrumb extends React.Component<Props> {
+    render() {
+        return (
+            <nav aria-label="Breadcrumb">
+                <StyledList style={styles.container}>
+                    {this.props.children}
```
jeresig: Unfortunately what I proposed will make this logic more challenging. I think you'll have to do something where you intersperse spacers between the children (adding `key`s to each of them). It's probably fine to add a component in this file that's nothing but the spacer and is only used internally.

packages/wonder-blocks-breadcrumb/components/breadcrumb-item.js
jeresig: More details on this: https://tink.uk/accessibility-support-for-css-generated-content/

kevinbarabash: I was going to suggest this as well, but I couldn't think of a good rationale for it.  Thanks for providing some rationale @jeresig.

packages/wonder-blocks-breadcrumb/components/breadcrumb.js
```diff
@@ -0,0 +1,45 @@
...
+
+const StyledList = addStyle("ol");
+
+/**
+ * The Breadcrumb component will have the following structure:
+ *
+ * 1. Breadcrumb Item: Represents a section within the page.
+ * 2. Separator: Adds a separator between each item.
+ */
+export default class Breadcrumb extends React.Component<Props> {
```
kevinbarabash: I think we may want to change the name of the component to `Breadcrumbs` and the package to `wonder-blocks-breadcrumbs` since people usually use the plural form.

packages/wonder-blocks-breadcrumb/index.js
kevinbarabash: In general I think most of our packages are set up such that there's either a default export or there's multiple named exports but not both.  I believe the icon package is the only exception to this and we should probably rectify that at some point.

packages/wonder-blocks-breadcrumb/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "dependencies": {
+      "@khanacademy/wonder-blocks-color": "^1.1.2",
+      "@khanacademy/wonder-blocks-core": "^2.2.0",
+      "@khanacademy/wonder-blocks-spacing": "^2.1.2",
+      "@khanacademy/wonder-blocks-typography": "^1.1.2"
+    },
+    "peerDependencies": {
+      "aphrodite": "^1.2.5",
+      "react": "^16.4.1"
+    },
```
kevinbarabash: Thanks for adding peer deps to the package.

packages/wonder-blocks-breadcrumb/components/breadcrumb-item.js
```diff
@@ -0,0 +1,49 @@
...
+            display: "inline-block",
+            margin: `0 ${Spacing.xxSmall}`,
```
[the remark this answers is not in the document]
jandrade: Oops, my bad :). I think I used the wrong var, as per figma's specs, it should be `xSmall` (8px). But I'll double check with Design if they want a different measure.

packages/wonder-blocks-breadcrumb/components/breadcrumb-item.js
```diff
@@ -0,0 +1,49 @@
...
+import Link from "@khanacademy/wonder-blocks-link";
+import Spacing from "@khanacademy/wonder-blocks-spacing";
+
+type Props = {|
+    ...AriaProps,
+
+    /**
+     * This is the content for the collection of Breadcrumbs
+     */
+    children: string | React.Element<typeof Link>,
```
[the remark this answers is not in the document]
jandrade: I'm really curious about that... What are you using for e2e tests? and is there a particular script that runs those tests at some point in the publishing process?

packages/wonder-blocks-breadcrumb/components/breadcrumb.js
jandrade: @kevinbarabash for a single item, do you think it should be `BreadcrumbsItem` or `BreadcrumbItem?

packages/wonder-blocks-breadcrumb/components/breadcrumb.js
```diff
@@ -0,0 +1,45 @@
...
+import {addStyle} from "@khanacademy/wonder-blocks-core";
+import BreadcrumbItem from "./breadcrumb-item.js";
+
+type Props = {|
+    ...AriaProps,
+
+    /**
+     * This is the content for the collection of Breadcrumbs
+     */
+    children: React.ChildrenArray<React.Element<typeof BreadcrumbItem>>,
```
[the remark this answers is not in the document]
jandrade: The difference is that `ChildrenArray` can be a single value or an array nested to any leve (e.g. `[val, val]` or just `val`):

https://github.com/facebook/flow/blob/master/lib/react.js#L274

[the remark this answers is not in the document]
jandrade: BTW I'm new to Flow, but have worked before with TypeScript. So happy to learn about any best practices/recommendations for Flow :)

packages/wonder-blocks-breadcrumb/components/breadcrumb.js
```diff
@@ -0,0 +1,45 @@
...
+|};
+
+const StyledList = addStyle("ol");
```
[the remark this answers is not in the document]
jandrade: IMO `ol` would be semantically appropriate, mostly b/c these links are represented in a hierarchical structure.
Also, I found this link about people's interpretation: https://webaim.org/projects/screenreadersurvey5/#breadcrumb

packages/wonder-blocks-breadcrumb/components/breadcrumb-item.js
jandrade: That's a really good point.

I have a possible solution for that:

Separator as a custom <svg> path:
```
<BreadcrumbItem>
  <Link or string goes here />
  <svg/>
<BreadcrumbItem/>
```
What would be the ideal way to build those spacers in WB?

Reading the documentation, it seems I can rely on a custom icon?
```import Icon from "@khanacademy/wonder-blocks-icon";
import type {IconAsset} from "@khanacademy/wonder-blocks-icon";

// Easter egg: what shape am I?
const customIcon: IconAsset = {
    small: "M6.92820 0L13.85640 4L13.85640 12L6.92820 16L0 12L0 4Z",
};
```

jeresig: @jandrade Hmm - so I think this would be better managed inside the Breadcrumbs component itself. The BreadcrumbItem would be left alone but when it comes to rendering the children it would insert the spacer in-between each of the BreadcrumbItems. So the children would essentially be: `[<BreadcrumbItem>, <Spacer>, <BreadcrumbItem>]`. So this wouldn't be something that the user would ever have to specify - although maybe eventually we could provide a way to specify custom ones, maybe something like: `<Breadcrumbs spacer={<Icon/>}> ... </Breadcrumbs>`.

Also: it's a hexagon! 😆

jandrade: hahah oops I didn't check that shape 😊. I changed the that logic to insert the spacer inside the `<BreadcrumbItem />` component (not sure why you don't see these changes here). The main reason for including the spacer inside there is to add the icon inside each `<li>`.

You can see the new implementation here: https://github.com/Khan/wonder-blocks/pull/391/files#diff-741ab7a0a4aa7032d534457c6b866936R60

The output for this new version would be something like this:

```
<ol>
  <li>
    <Link or string />
    <separator />
  </li>
  ...
</ol>
```

If you still think it's better to add the spacer outside of `BreadcrumbsItem`, we would need to move the `<li />` up to the `Breadcrumb` component ([link](https://github.com/Khan/wonder-blocks/pull/391/files#diff-b2b53ffb4b23955fb0995a5c5ac79be2R48)) to preserve this markup/structure.

What do you think?

It's a good idea to add the `spacer` prop to `<Breadcrubs />` :)

packages/wonder-blocks-breadcrumb/components/breadcrumb-item.js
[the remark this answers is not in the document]
kevinbarabash: They're used by the e2e tests in webapp.

packages/wonder-blocks-breadcrumb/components/breadcrumb.js
kevinbarabash: I think probably `BreadcrumbItem`.

packages/wonder-blocks-breadcrumbs/docs.md
```diff
@@ -0,0 +1,25 @@
+```js
+const Link = require("@khanacademy/wonder-blocks-link").default;
+const {BreadcrumbsItem} = require("@khanacademy/wonder-blocks-breadcrumbs");
+
+<Breadcrumbs>
+  <BreadcrumbsItem>
+    <Link href="https://khanacademy.org/">Home</Link>
+  </BreadcrumbsItem>
+    <BreadcrumbsItem>
```
kevinbarabash: nit: this is indented too far

packages/wonder-blocks-breadcrumbs/docs.md
```diff
@@ -0,0 +1,25 @@
+```js
+const Link = require("@khanacademy/wonder-blocks-link").default;
+const {BreadcrumbsItem} = require("@khanacademy/wonder-blocks-breadcrumbs");
+
+<Breadcrumbs>
+  <BreadcrumbsItem>
```
kevinbarabash: nit: most of our code use 4 spaces tabs.

packages/wonder-blocks-breadcrumbs/docs.md
```diff
@@ -0,0 +1,25 @@
...
+```
+
+**NOTE**: `<BreadcrumbsItem />` only accepts two element types:
+
+1. string
+2. `<Link />`
+
+## Accessibility
+
+- It should follow guidelines from [W3C](https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html).
```
kevinbarabash: I love that you're linking to an a11y guideline.

packages/wonder-blocks-breadcrumbs/components/breadcrumbs-item.js
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * This is the content for the collection of Breadcrumbs
+     */
+    children: string | React.Element<typeof Link>,
+
+    /**
+     * When true, the separator is shown; otherwise, the separator is not shown.
+     */
+    showSeparator: boolean,
```
kevinbarabash: Can you mark this prop as private?

packages/wonder-blocks-breadcrumbs/components/breadcrumbs.js
```diff
@@ -0,0 +1,67 @@
...
+import type {AriaProps} from "@khanacademy/wonder-blocks-core";
+import {addStyle} from "@khanacademy/wonder-blocks-core";
+import BreadcrumbsItem from "./breadcrumbs-item.js";
+
+type Props = {|
+    ...AriaProps,
+
+    /**
+     * This is the content for the collection of Breadcrumbs
+     *
```
kevinbarabash: nit: blank line

packages/wonder-blocks-breadcrumbs/components/breadcrumbs.js
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * 1. Breadcrumbs Item: Represents a section within the page.
+ * 2. Separator: Adds a separator between each item.
+ */
+export default class Breadcrumbs extends React.Component<Props> {
+    render() {
+        const {children} = this.props;
+        // using React.Children allows to deal with opaque data structures
+        // e.g. children = 'string' vs children = []
+        const numChildren = React.Children.count(children) - 1;
```
kevinbarabash: This should probably be renamed to `numSeparators` or `lastChildIndex`.

packages/wonder-blocks-breadcrumbs/docs.md
jandrade: That's a good observation, I think I'm relying to much in Prettier 😄. Do you know if there's a setting in VS Code to format code snippets inside markdown files?

packages/wonder-blocks-breadcrumbs/components/breadcrumbs.js
jandrade: 👍  I'll go with `lastChildIndex` (b/c of the variable intention)

PR #393, opened 2019-05-03
[breaking: modal] WB-545: Implement Modal v2 - Remove previous variations (2/3) [WIP]

jandrade:
This PR builds on #392 and it's part of [WB-545](https://khanacademy.atlassian.net/browse/WB-545).

It contains the following changes:

# Refactor
- Removes `StandardModal, OneColumnModal and TwoColumnModal` components from the `@khanacademy/wonder-blocks-modal` package.
- Updates modal documentation to only show `OnePaneDialog` examples.
- Updates documentation that points to Modal v1 inside other packages: `wb-dropdown`, `wb-tooltip`.

PR #394, opened 2019-05-09
[breaking: modal] WB-545: Implement Modal v2 - Modify building blocks (3/3)

jandrade:
This PR builds on #393 and it intends to complete the remaining work for [WB-545](https://khanacademy.atlassian.net/browse/WB-545).

It contains the following changes:

## General
- Adds/updates examples to showcase v2.

### ModalContent
- Removes `header` from this component (not used anymore).
- Updates styles to match design specs.

### ModalDialog
- Updates the dialog wrapper to match design specs.

### ModalFooter
- Removes `style` prop.

### ModalHeader
- Adds new props (`title`, `subtitle`, `breadcrumbs`) to match new implementation spec.
- Updates styles to match design specs.
- Adds unit tests.

### ModalPanel
- Uses `<ModalHeader />` instead of `<Toolbar />` to match new impl. spec.

## TODOs
- [ ] Design review with Nefaur
- [ ] Give modal titles unique ids ([WB-180](https://khanacademy.atlassian.net/browse/WB-180))
- [x] Add `<Breadcrumbs />` as dependency. We first need to publish `@khanacademy/wonder-blocks-breadcrumbs`.


jeresig:

Built with commit 8b19f514f86e9f71e908f99c9b43942f95420032

https://deploy-preview-394--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-header.js
```diff
@@ -4,38 +4,67 @@ import {StyleSheet} from "aphrodite";
...
-    static __IS_MODAL_HEADER__ = true;
-    static isClassOf(instance: any) {
-        return instance && instance.type && instance.type.__IS_MODAL_HEADER__;
-    }
-
     render() {
-        const {style, color, children} = this.props;
+        const {light, title, subtitle, breadcrumbs} = this.props;
+
+        if (subtitle && breadcrumbs) {
```
jandrade: @kevinbarabash @jeresig At the end I decided to go with this approach b/c Flow threw an error when trying to render these props inside the `render` method. You can see the example here:

https://flow.org/try/#0PTAEAEDMBsHsHcBQBLAtgB1gJwC6gFSgCGAzqAEoCmRAxnpFrKqAERbV0sDcivOAnukqgAQqUoAFRujIBeUAG8APolBrQNWADsclHQC4KHHADoAcrAAmlADSr1OZDmiVDJHFmRaA5nfWhIWFhdLAB+QypaUwtrP3VtAGE4Ekpw0AAKAEpQWQA+UFRkAA9KSzi1IgAjWAA3VIjjcytbezVKyjh4NMi6JtjW0F13AElLNPdPHzslAF8eRAEhUAB1JwALAGUAV0rHZ2F5ZQGTE7EUqVgZctASHb2XNw8vX0RZ+cXhVZw1kXYiSxoWC2qEqckUKn8JxMZ0k0hI10qfwBQJBJAaUT6LTefEEwguMhyonE+LIShW622uycLlAZK+PyRgOBoPmiBo0FIZAAslYiNAABLUaxYUCUIq6LSWMg9UwJJiYLR6HAAHhJ+QU9nYksoWCyag1-nYOC2WC0oGVawAjLkFN9kCQTOg4SZ7pQZoo7Q6nZcHYihUzUTNlcArbkeGoZohI2ztO5QGshTrLYTlTzLHzBf8dRptBKcLIWEMcCxBtTKAXArASyRLJAC7XLCXgGGY1o4wms1gAEwptMZxMizQ6JUFosl10VoLVu5lguVIhYJstoftgcAZl7vIFA5zw50o8o7nHs5YlZLfv+AdBsgUAG0WPOAF4sGysACOWyKLAAuu7mzwVzwDthQAFk3dNt07Xc8wPI9S32Scq1AC9kWZEgb3vJ8X3fT8f3dW4qQQh8FyXHggA

packages/wonder-blocks-modal/components/one-pane-dialog/one-pane-dialog.js
```diff
@@ -109,14 +117,15 @@ export default class OnePaneDialog extends React.Component<Props> {
                         <ModalPanel
                             onClose={onClose}
                             header={
-                                <ModalHeader>
-                                    <h1>{title}</h1>
-                                    <h2>{subtitle}</h2>
-                                </ModalHeader>
+                                <ModalHeader
```
jandrade: As I previously mentioned, this is the final version of the Header.

kevinbarabash:
> Give modal titles unique ids

This could even be a separate PR.

packages/wonder-blocks-modal/components/modal-header.js
```diff
@@ -4,38 +4,67 @@ import {StyleSheet} from "aphrodite";
...
-                        {children}
+                    <View style={[styles.header, !light && styles.dark]}>
+                        {breadcrumbs &&
+                            !subtitle && (
+                                <View style={styles.breadcrumbs}>
+                                    {breadcrumbs}
+                                </View>
+                            )}
+                        <HeadingMedium id="wb-toolbar-title">
+                            {title}
```
kevinbarabash: I'm glad we don't have to ellipsize things. 🙂

packages/wonder-blocks-modal/components/modal-header.test.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+import React from "react";
+
+import expectRenderError from "../../../utils/testing/expect-render-error.js";
+import ModalHeader from "./modal-header.js";
+
+describe("ModalHeader", () => {
+    test("using `subtitle` and `breadcrumbs` should throw", () => {
```
kevinbarabash: Can you add test cases for the valid cases?

packages/wonder-blocks-modal/docs.md
```diff
@@ -41,9 +60,9 @@ class ModalWrapper extends React.Component {
 </View>
 ```
 
-### Example: Modals with custom styles
+### Example: Dialogs with custom styles
```
kevinbarabash: I think we should also have an example that shows us using the building blocks.  building-blocks.md would be a good place for this.

packages/wonder-blocks-modal/docs.md
```diff
@@ -91,16 +130,17 @@ const styleSheets = {
...
                         <View>
-                            <Title style={styles.title}>Title</Title>
                             <Body>
                                 Hello World!
                             </Body>
                         </View>
                     }
                     onClose={() => alert("This would close the modal.")}
+                    below={<View style={styles.below} />}
+                    above={<View style={styles.above} />}
```
kevinbarabash: nice

packages/wonder-blocks-modal/components/one-pane-dialog/one-pane-dialog.md
```diff
@@ -64,13 +65,15 @@ const styles = StyleSheet.create({
 </View>;
 ```
 
-### Example: Custom OnePaneDialog with long content (overflows)
+### Example: Custom Footer (3 actions)
```
kevinbarabash: This example is awesome.  A comment here to encourage people to resize the page so that they can see what the mobile version looks like would be helpful.  One minor issue is the tertiary buttons look weird at 100%. Can you file a bug for that?
Screen Shot 2019-05-09 at 7 09 43 PM

packages/wonder-blocks-modal/components/one-pane-dialog/one-pane-dialog.md
```diff
@@ -93,46 +96,165 @@ const styles = StyleSheet.create({
...
+});
+
+const styleSheets = {
+    all: defaultStyles,
+    small: smallStyles
+};
+
+<View style={styles.previewSizer}>
+    <View style={styles.modalPositioner}>
+        <MediaLayout styleSheets={styleSheets}>
```
kevinbarabash: Love the use of `MediaLayout` in these examples.  I want push the use of `MediaLayout` more heavily in webapp.  It makes composition of styles easier when dealing with media queries.

packages/wonder-blocks-modal/components/one-pane-dialog/one-pane-dialog.md
```diff
@@ -93,46 +96,165 @@ const styles = StyleSheet.create({
...
-            title="Custom modal"
-            subtitle="With long content"
+            title="Multi-line title that wraps to show how this component adjusts with longer content"
+            breadcrumbs={
+                <Breadcrumbs>
+                    <BreadcrumbsItem>
+                        Bread
+                    </BreadcrumbsItem>
+                    <BreadcrumbsItem>
+                        Crumb
```
kevinbarabash: lol

packages/wonder-blocks-modal/components/one-pane-dialog/one-pane-dialog.md
```diff
@@ -93,46 +96,165 @@ const styles = StyleSheet.create({
...
+                    </BreadcrumbsItem>
+                    <BreadcrumbsItem aria-current="page">Trail</BreadcrumbsItem>
```
kevinbarabash: We should probably have `Breadcrumbs` automatically add `aria-current` to the last `BreadcrumbsItem` automatically.

packages/wonder-blocks-modal/components/one-pane-dialog/one-pane-dialog.md
```diff
@@ -93,46 +96,165 @@ const styles = StyleSheet.create({
...
+    small: smallStyles
+};
+
+const Footer = () => (
+    <MediaLayout styleSheets={styleSheets}>
+        {({ styles }) => (
+            <View style={styles.row}>
+                <Button style={styles.button} kind="tertiary">
+                    Tertiary action
+                </Button>
```
kevinbarabash: Another option is placing `<Strut size={16}/>` in between the buttons.

packages/wonder-blocks-modal/components/one-pane-dialog/one-pane-dialog.md
```diff
@@ -93,46 +96,165 @@ const styles = StyleSheet.create({
...
-        position: "absolute",
-        top: 40,
-        left: -120
+const smallStyles = StyleSheet.create({
+    row: {
+        flexDirection: "column-reverse",
+        width: "100%"
     },
+    button: {
+        marginBottom: 16
```
kevinbarabash: Use wonder-blocks-spacing for numbers like this.

packages/wonder-blocks-modal/components/modal-panel.js
```diff
@@ -176,18 +181,14 @@ const styleSheets = {
...
             paddingBottom: 32,
         },
     }),
 
     small: StyleSheet.create({
         closeButton: {
-            left: 16,
+            right: 16,
             top: 16,
         },
```
kevinbarabash: I thought that we merged `small` styles will `all` styles... won't this result in the close button having `position: "absolute"` regardless of the size?

packages/wonder-blocks-modal/components/modal-panel.js
```diff
@@ -19,22 +18,37 @@ type Props = {|
...
     scrollOverflow: boolean,
-    /** The color of the panel (defaults to light). */
-    color: "light" | "dark",
-    /** Any optional styling to apply to the panel. */
+
+    /**
+     * Whether to display the "light" version of this component instead, for
+     * use when the item is used on a dark background.
+     */
+    light: boolean,
```
kevinbarabash: Yay!  Thank you.

packages/wonder-blocks-modal/docs.md
kevinbarabash: Maybe build the two-column modal from https://www.figma.com/file/CRryTHqyr1l6aOOltD32KMh5/Modal-v2?node-id=132%3A0.

packages/wonder-blocks-modal/docs.md
```diff
@@ -1,10 +1,29 @@
 Looking for docs for StandardModal, OneColumnModal, or TwoColumnModal click [here](https://deploy-preview-389--wonder-blocks.netlify.com/#modal)
 
-## Wrapping Modals
+## Accessibility
```
kevinbarabash: Thanks for adding all of this documentation.

packages/wonder-blocks-modal/docs.md
```diff
@@ -1,10 +1,29 @@
...
-modal component.  These wrapped components will also work with ModalLauncher
-in the same way the default ones do.
+It should follow guidelines from [W3C](https://www.w3.org/TR/wai-aria-practices/#dialog_modal).
+
+### Keyboard Interaction
+When a dialog opens, focus moves to an element inside the dialog. See notes below regarding initial focus placement.
+
+- Tab:
+    - Moves focus to the next tabbable element inside the dialog.
+    - If focus is on the last tabbable element inside the dialog, moves focus to the first tabbable element inside the dialog.
```
kevinbarabash: Which element is the first item?

packages/wonder-blocks-modal/docs.md
```diff
@@ -1,10 +1,29 @@
...
-## Wrapping Modals
+## Accessibility
 
-Often you'll want to define a new modal component by wrapping an existing
-modal component.  These wrapped components will also work with ModalLauncher
-in the same way the default ones do.
+It should follow guidelines from [W3C](https://www.w3.org/TR/wai-aria-practices/#dialog_modal).
+
+### Keyboard Interaction
+When a dialog opens, focus moves to an element inside the dialog. See notes below regarding initial focus placement.
```
kevinbarabash: It's probably worth mentioning the behavior of focus when closing the modal.

packages/wonder-blocks-modal/docs.md
```diff
@@ -1,10 +1,29 @@
...
+- Shift + Tab:
+    - Moves focus to the previous tabbable element inside the dialog.
+    - If focus is on the first tabbable element inside the dialog, moves focus to the last tabbable element inside the dialog.
+- Escape: Closes the dialog.
+
+### WAI-ARIA Roles
+- The element that serves as the **dialog container** has `aria-role` defined as `dialog`.
+- The dialog has a value set for the `aria-labelledby` property that refers to a **visible dialog title**.
+
+## Wrapping Dialogs
```
kevinbarabash: I got some feedback that this example is too complex to start with and should be moved down.  Unfortunately, styleguidist makes a lot of decisions about where things live.  I would like to do something about it in future but don't worry about it right now.

packages/wonder-blocks-modal/components/modal-panel.js
```diff
@@ -49,28 +63,23 @@ type Props = {|
...
 
-        // Figure out the background color that'll be behind the close button
-        // If a titlebar is specified then we use that, otherwise we use the
-        // default background color.
-        const topBackgroundColor = (titleBar && titleBar.props.color) || color;
-
         return (
             <MediaLayout styleSheets={styleSheets}>
                 {({styles}) => (
                     <CloseButton
```
kevinbarabash: This could maybe be inlined.  That would avoid an extra `MediaLayout`... also there's less logic now.

packages/wonder-blocks-modal/components/modal-panel.js
```diff
@@ -19,22 +18,37 @@ type Props = {|
      * are positioned around it.
      */
     content: React.Element<typeof ModalContent> | React.Node,
-    /** A title bar (with optional subtitle) to show at the top of the panel. */
-    titleBar?: React.Element<typeof Toolbar>,
-    /** A header to show above the contents, but below the title bar. */
+
+    /**
+     * The modal header to show at the top of the panel.
+     */
```
kevinbarabash: Thanks for expanding these comments.  This is the more prevalent style.

packages/wonder-blocks-modal/components/modal-header.js
```diff
@@ -46,22 +75,29 @@ export default class ModalHeader extends React.Component<Props> {
...
             color: Color.white,
         },
+
+        breadcrumbs: {
+            color: "rgba(33, 36, 44, 0.64)",
+            marginBottom: 8,
+        },
+
+        subtitle: {
+            color: "rgba(33, 36, 44, 0.64)",
```
kevinbarabash: Same, these could be `Colors.black64`.

packages/wonder-blocks-modal/docs.md
jandrade: Good call. What's the current behavior? Are we currently returning the focus to the element that opened the modal in the fist place?

packages/wonder-blocks-modal/components/one-pane-dialog/one-pane-dialog.md
jandrade: Nice catch, I'll create that ticket :)

packages/wonder-blocks-modal/components/one-pane-dialog/one-pane-dialog.md
jandrade: That makes sense. I'll remove it from this example and will file a ticket for implementing this inside Breadcrumbs as well. Unless you think I can include that functionality as part of this PR. What do you think?

packages/wonder-blocks-modal/components/modal-header.test.js
jandrade: Added a new test on line 9.

packages/wonder-blocks-modal/components/modal-header.test.js
```diff
@@ -0,0 +1,28 @@
...
+import expectRenderError from "../../../utils/testing/expect-render-error.js";
+import ModalHeader from "./modal-header.js";
+
+describe("ModalHeader", () => {
+    test("using only `subtitle` should render the header", () => {
+        const wrapper = shallow(
+            <ModalHeader title="Title" subtitle="Subtitle" />,
+        );
+
+        // we make sure the component is rendered
```
jandrade: I used this approach b/c we are already generating snapshots automatically. Or, do you think it would be better to use `toMatchSnapshot()` instead?

packages/wonder-blocks-modal/components/one-pane-dialog/one-pane-dialog.md
kevinbarabash: A separate PR is probably best.

packages/wonder-blocks-modal/docs.md
kevinbarabash: I thought we were, but I just tested and we aren't.  We should though.  Can you file a ticket for this?

packages/wonder-blocks-modal/building-blocks.md
```diff
@@ -1 +1,134 @@
 Use these low-level building blocks to build your own modal dialog, instead of using OnePaneDialog. This should happen very rarely and only when a specific exception is required.
+
```
jandrade: @kevinbarabash Turns out adding a `TwoPaneDialog` example was quite easy to build!.

Also I have to mention that `MediaLayout` is 🔥⚡️, it helped me to handle the logic to show the close button in different panels for different screen resolutions (using `mediaSize`).

packages/wonder-blocks-modal/docs.md
jandrade: Added to `building-blocks.md` as you suggested 😃

packages/wonder-blocks-modal/docs.md
jandrade: https://khanacademy.atlassian.net/browse/WB-564

packages/wonder-blocks-modal/docs.md
jandrade: I'm not sure what you mean here. element in terms of element type (e.g. button, input, etc)? or describe which element should be focused according to the current implementation?

packages/wonder-blocks-modal/components/modal-header.test.js
kevinbarabash: I think this is fine for now.  We talked a while back about changing the generated snapshots to match shallow snapshots at which point we'd probably want to snapshot tests to this (and other test suites) that compare full snapshots.

packages/wonder-blocks-modal/building-blocks.md
```diff
@@ -1 +1,134 @@
...
+    small: StyleSheet.create({
+        dialog: {
+            width: "100%",
+            height: "100%",
+            overflow: "hidden",
+        },
+
+        row: {
+            flexDirection: "column",
```
kevinbarabash: Nit: this is a little weird.  It might make sense to call this style `column` and then use `mediaSize` to decide whether to use `style.row` or `style.column`.

packages/wonder-blocks-modal/components/modal-header.test.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+import React from "react";
+import {shallow} from "enzyme";
+
+import expectRenderError from "../../../utils/testing/expect-render-error.js";
+import ModalHeader from "./modal-header.js";
+
+describe("ModalHeader", () => {
+    test("using only `subtitle` should render the header", () => {
```
kevinbarabash: We should probably also have test cases for `header` only and `header and breadcrumbs`.

packages/wonder-blocks-modal/components/modal-header.test.js
```diff
@@ -0,0 +1,28 @@
...
+        const wrapper = shallow(
+            <ModalHeader title="Title" subtitle="Subtitle" />,
+        );
+
+        // we make sure the component is rendered
+        expect(wrapper.exists()).toBe(true);
```
kevinbarabash: Nit: we try to use AAA (Arrange, Act, Assert) style testing.  Search for `// Arrange` to see examples.  In this case the "Arrange" and "Act" are one in the same so you could `// Arrange, Act` and then `// Assert` before the `expect`.

packages/wonder-blocks-modal/docs.md
kevinbarabash: I guess it doesn't matter a whole lot which is the first item in terms of wrapping around when tabbing.  We can leave this as is for now seeing as have a separate task to improve initial focus behavior.

packages/wonder-blocks-modal/building-blocks.md
jandrade: To avoid adding extra logic in the examples, I'm going to use a more generic name: `panelGroup`.

PR #395, opened 2019-05-11
Add wallaby.js config, update enzyme, add more tests for ActionMenu and ActionItem

kevinbarabash:
I wanted to evaluate wallaby so I wrote some more tests for ActionMenu and ActionItem.  I wanted to use `shallow` rendering when testing `ActionItem` so that I could verify when `Link` vs. `a` tags were being rendered.  `ActionItem` uses `ClickableBehavior` which uses a `children` as a render prop.  In order to test this correctly I needed to upgrade `enzyme` (and the adapter) which added a `renderProp` method which provides a way call the `children` method that get's passed to `ClickableBehavior`.  I'm planning to use this testing approach when I get around to revising #387.


jeresig:

Built with commit 45a454a877f44e38a171dee29c971d92b5e4197b

https://deploy-preview-395--wonder-blocks.netlify.com

jeresig:

Built with commit f7ccaaa0fdd32465041786509090730b925f26c1

https://deploy-preview-395--wonder-blocks.netlify.com

wallaby.js
```diff
@@ -0,0 +1,64 @@
...
+
+        tests: [
+            'packages/wonder-blocks-*/**/*.test.js',
+            'shared-unpackaged/**/*.test.js',
+        ],
+
+        compilers: {
+            "packages/wonder-blocks-*/**/*.js": wallaby.compilers.babel(babelConfig),
+            "shared-unpackaged/**/*.js": wallaby.compilers.babel(babelConfig),
+            "utils/**/*.js": wallaby.compilers.babel(babelConfig),
```
kevinbarabash: Even though we're using jest normally jest would be responsible for compiling files, wallaby compiles files itself before passing them for testing.  The reason why wallaby handles the compilation is that some its features like advanced logging and live comments requires instrumenting the code. https://wallabyjs.com/docs/intro/advanced-logging.html

wallaby.js
```diff
@@ -0,0 +1,64 @@
...
+            const jestConfigPath = path.join(
+                wallaby.localProjectDir,
+                "./config/jest/test.config.js",
+            );
+            const jestConfig = require(jestConfigPath);
+            delete jestConfig.transform;
+            jestConfig.moduleNameMapper = {
+                "^@khanacademy/(.*)$":
+                    wallaby.projectCacheDir + "/packages/$1",
+            };
```
kevinbarabash: This is so that the tests use the source version of imported packages such as `@khanacademy/wonder-block-core`, etc.

utils/testing/mount.js
```diff
@@ -15,7 +15,9 @@ const unmountAll = () => {
     for (const key of wrappersToUnmount) {
         const wrapper = ACTIVE_WRAPPERS[key];
         delete ACTIVE_WRAPPERS[key];
-        wrapper.unmount();
+        if (wrapper.first().length) {
+            wrapper.unmount();
+        }
```
kevinbarabash: After upgrading `enzyme` calling `.unmount()` failed b/c a couple of our tests do this manually (b/c we're actually testing the what happens when unmounting occurs).  When a component is unmounted `wrapper.first().length` will be `0`.

packages/wonder-blocks-dropdown/components/action-menu.test.js
```diff
@@ -80,4 +92,107 @@ describe("ActionMenu", () => {
...
+            <ActionMenu menuText={"Action menu!"}>
+                <ActionItem label="Create" />
+                {showDeleteAction && <ActionItem label="Delete" />}
+            </ActionMenu>,
+        );
+
+        menu.setState({open: true});
+
+        expect(menu.find(ActionItem)).toHaveLength(1);
+        expect(menu.find(ActionItem).at(0)).toHaveProp("label", "Create");
```
kevinbarabash: In terms of providing best practices around jest and enzyme, I think we should encourage people to use more specific matchers as opposed to:
```
expect(menu.find(ActionItem).length).toEqual(1);
expect(menu.find(ActionItem).at(0).props().label).toEqual("Create");
```

packages/wonder-blocks-dropdown/components/action-item.test.js
```diff
@@ -0,0 +1,47 @@
...
+    it("should render an anchor if there's no router", () => {
+        const wrapper = shallow(<ActionItem href="/foo" />);
+
+        expect(
+            wrapper
+                // call the props.children(state, handlers) with empty objects for both params
+                .renderProp("children")({}, {})
+                // shallow render the first child of that result to unwrap StyledAnchor HOC
+                .first()
+                .shallow()
```
kevinbarabash: There's some discussion in https://github.com/airbnb/enzyme/issues/539 around possible API improvements to make this easier.

packages/wonder-blocks-dropdown/components/action-item.test.js
```diff
@@ -0,0 +1,47 @@
+import * as React from "react";
+import renderer from "react-test-renderer";
+import {Link} from "react-router-dom";
+import {shallow} from "enzyme";
+import ReactRouterEnzymeContext from "react-router-enzyme-context";
```
kevinbarabash: Needed for using `shallow` when testing components that use `react-router`.

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -16,7 +16,7 @@ type MenuProps = {|
     /**
      * The items in this dropdown.
      */
-    children?: Array<Item>,
+    children?: Array<Item> | Item,
```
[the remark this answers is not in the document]
kevinbarabash: I gave this a try but then a whole bunch of errors cropped up in `getMenuItems` in the `map` callback. 😞

packages/wonder-blocks-dropdown/components/action-menu.test.js
kevinbarabash: https://www.npmjs.com/package/eslint-plugin-jest has a `prefer-to-have-length` rule that we could adopt.

jandrade: That makes sense to me :).

packages/wonder-blocks-dropdown/components/action-item.test.js
```diff
@@ -0,0 +1,47 @@
...
+        expect(actionItem).toMatchSnapshot();
+    });
+
+    it("should render an anchor if there's no router", () => {
+        const wrapper = shallow(<ActionItem href="/foo" />);
+
+        expect(
+            wrapper
+                // call the props.children(state, handlers) with empty objects for both params
+                .renderProp("children")({}, {})
```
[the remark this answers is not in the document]
kevinbarabash: This was new to me too. :)

wallaby.js
somewhatabstract: Make sense although I wonder how much time we'd end up in a situation where a test fails because of a wallaby compilation issue?

somewhatabstract: Does this mean wallaby will run on all of wonderblocks?

somewhatabstract:
It looks good to me from a changes perspective. Nothing that I feel strongly about.

wallaby.js
kevinbarabash: If things are wonky we can always redirect imports to the dist files instead of the source files.

Yes, it runs all of the tests and produces nice coverage reports with rollups in the browser app.  wallabyjs.com/app redirects to localhost:54321 (I forget the exact port).

Screen Shot 2019-05-15 at 5 04 18 PM

PR #396, opened 2019-05-15
WB-180 Give modal titles unique ids

jandrade:
- Added unique ID to modal titles
- Created `<UniqueDialog />` component to either return the unique ID or return the custom ID already provided by the parent component.
- Modified both `ModalHeader` and `ModalDialog` to associate the the modal title with the Dialog using the `aria-labelledby` attribute.
- Added unit tests.


jeresig:

Built with commit b9c506058eb08d0f6d6cf4503948860ccd6536c3

https://deploy-preview-396--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/unique-dialog.js
```diff
@@ -0,0 +1,63 @@
...
+|};
+
+/**
+ * This is a Dialog wrapper that returns a unique Modal Title for A11Y purposes.
+ * This way, the wrapped Dialog will receive this custom ID and will use it to refer to a
+ * visible dialog title.
+ *
+ * @see https://www.w3.org/TR/wai-aria-practices/#dialog_roles_states_props
+ */
+export default class UniqueDialog extends React.Component<Props> {
```
jandrade: Added this functionality in a new component for reusability/composition purposes. This way, if a dev needs to create his own Modal/Dialog variation (e.g. TwoPanelDialog), he can rely on this component to add this unique title to the dialog.

e.g.
```
<UniqueDialog>
    {(id) => <CustomTwoPaneDialog id={id}> .... </CustomTwoPaneDialog>}
</UniqueDialog>
```

somewhatabstract: I wonder if we can make a more generic variation of this component. I mean, there's nothing `modal`-specific here except the scope. Perhaps we can elevate scope to props and then make this be a drop in replacement for `UniqueIDProvider`?

packages/wonder-blocks-modal/components/modal-dialog.js
```diff
@@ -35,15 +35,22 @@ type Props = {|
      * Custom styles
      */
     style?: StyleType,
+
+    /**
+     * An optional id to provide a selector for the title element.
+     */
+    id?: string,
```
kevinbarabash: Maybe change this to `"aria-labelledby"": string`.

packages/wonder-blocks-modal/components/modal-header.js
```diff
@@ -32,6 +32,11 @@ type Props = {|
      * use when the item is used on a dark background.
      */
     light: boolean,
+
+    /**
+     * An optional id to provide a selector for the title element.
+     */
+    id?: string,
```
kevinbarabash: Maybe change this to `titleId?: string`.  Should this be mandatory?

packages/wonder-blocks-modal/components/one-pane-dialog/one-pane-dialog.js
```diff
@@ -80,6 +81,12 @@ type Props = {|
      * Test ID used for e2e testing.
      */
     testId?: string,
+
+    /**
+     * An optional id parameter for the title. If one is
+     * not provided, a unique id will be generated.
+     */
+    id?: string,
```
kevinbarabash: Maybe `"titleId"?: string`

packages/wonder-blocks-modal/components/modal-header.js
jandrade: Agreed, this should be mandatory if we want to enforce accessibility.

packages/wonder-blocks-modal/components/unique-dialog.js
jandrade: That makes total sense!. This task will be implemented here: https://khanacademy.atlassian.net/browse/WB-576

PR #399, opened 2019-05-18
Update .vimrc to actually work with ALE

jsatk:
Previously this was flat-out incorrect and broken.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -14,15 +14,17 @@
 " Asynchronous Lint Engine (ALE) {{{
 " Github link for ALE: https://github.com/w0rp/ale
 
-" Use Khan Academy Linter's eslint executable instead of your local or global.
-let b:ale_javascript_eslint_executable = $HOME."/khan/devtools/khan-linter/node_modules/.bin/eslint"
-let b:ale_javascript_eslint_use_global = 1
-
-" Linters we want to run.
-let b:ale_linters = {
+let g:ale_linters = {
 \   'javascript': [
 \       'flow',
 \       'eslint',
+\   ],
+\}
+
+let g:ale_fixers = {
+\   'javascript': [
+\       'eslint',
+\       'prettier_eslint',
 \       'trim_whitespace',
 \       'remove_trailing_lines',
 \   ],


jeresig:

Built with commit 7483c37054c348ad558f92241d1a84084a09885d

https://deploy-preview-399--wonder-blocks.netlify.com

.vimrc
```diff
@@ -15,14 +15,20 @@
 " Github link for ALE: https://github.com/w0rp/ale
 
 " Use Khan Academy Linter's eslint executable instead of your local or global.
-let b:ale_javascript_eslint_executable = $HOME."/khan/devtools/khan-linter/node_modules/.bin/eslint"
-let b:ale_javascript_eslint_use_global = 1
+let g:ale_javascript_eslint_executable = $HOME."/khan/devtools/khan-linter/node_modules/.bin/eslint"
+let g:ale_javascript_eslint_use_global = 1
```
kevinbarabash: We should use the version of eslint that's local to wonder-blocks.

.vimrc
```diff
@@ -15,14 +15,20 @@
...
-" Linters we want to run.
-let b:ale_linters = {
+let g:ale_linters = {
 \   'javascript': [
 \       'flow',
 \       'eslint',
+\   ],
+\}
+
+let g:ale_fixers = {
```
kevinbarabash: Does this provide a way to autofix issues that eslint finds from within vim?

.vimrc
```diff
@@ -15,14 +15,20 @@
...
+\   'javascript': [
+\       'eslint',
+\       'prettier_eslint',
```
kevinbarabash: Why does `prettier_eslint` need to be specified separately.

.vimrc
jsatk: Yep!

.vimrc
jsatk: Because, AFAIK, we don't use prettier by itself. We use a prettier integration with eslint. This specifies that this isn't default prettier, but the eslint integrated one. 

I copied this over from webapp and didn't verify it here :(

jsatk: Confirmed. We need `prettier_eslint`.

PR #401, opened 2019-05-20
[fix: grid] Drop unnecessary nested Views from Grid's Row

jsatk:
The nested `View` caused issues during implementation.  As far as I can tell there isn't a reason for it.  Examples still look good.


jeresig:

Built with commit 731eefa6c726c4ce835c6be4a5294eddb1df3bbb

https://deploy-preview-401--wonder-blocks.netlify.com

somewhatabstract:
Can you give more details regarding what issues were caused during implementation and the scenarios tested to ensure the specification isn't broken? Consider things like viewport resizing and such.

I think having this additional information will help if this gets revisited in future to ensure we don't have regressions. Also, since this is a change to address an encountered problem, a test case that verifies that problem is addressed would be very helpful.

kevinbarabash:
Having an example in the docs that shows off the desired behavior would be good.  A test case would be even better.  Additionally, it should be possible to design a test case which renders row with `height: 100%` into a container of a certain height and then check that the height of the row matches the height of the container.  Having an automated test case is a better guard against regressions.

jsatk:
> Can you give more details regarding what issues were caused during implementation and the scenarios tested to ensure the specification isn't broken? Consider things like viewport resizing and such.
> 
> I think having this additional information will help if this gets revisited in future to ensure we don't have regressions. Also, since this is a change to address an encountered problem, a test case that verifies that problem is addressed would be very helpful.

Will do shortly!

jsatk:
> Having an example in the docs that shows off the desired behavior would be good. A test case would be even better. Additionally, it should be possible to design a test case which renders row with `height: 100%` into a container of a certain height and then check that the height of the row matches the height of the container. Having an automated test case is a better guard against regressions.

I will do my best to update tests. I might need your help here. As we've chatted quite a bit in Slack about this already, the docs around how to update things are missing. I might need to lean on you for the best approach to doing this right.

kevinbarabash:
I can help out with the test case.

kevinbarabash:
I don’t think we’ll be able write a test case for this.  JSDOM doesn’t actually render things so we can’t get the size/position of elements on the page.  We’d need selenium for that and we don’t have that set up in wonder-blocks yet.  I think add a TODO to add a test for this later once have selenium set up is the way to go.

packages/wonder-blocks-grid/components/row.md
```diff
@@ -38,3 +38,64 @@ const styles = StyleSheet.create({
...
+Another example – If the height of the [Cell's](#cell) contents is taller than
+the [Row](#row) it will allow vertical scrolling. 
+
+```jsx
+const Color = require("@khanacademy/wonder-blocks-color").default;
+const {Body} = require("@khanacademy/wonder-blocks-typography");
+const {View, Text} = require("@khanacademy/wonder-blocks-core");
+const {StyleSheet} = require("aphrodite");
+
+const styles = StyleSheet.create({
```
kevinbarabash: I think something more like:
```
    view: {
        background: Color.offBlack,
        height: "400px",
    },

    row: {
        padding: "16px 0",
        border: `1px solid ${Color.gold}`,
        background: Color.white,
        height: "100%",
        display: "flex",
        alignItems: "stretch",
    },

    cell: {
        background: Color.gold,
        overflowY: "auto",
        padding: 5,
    },
});
```
would better illustrate that the rows can be styled so that their height fills the parent container.

PR #402, opened 2019-05-21
[feature: modal] WB-405 Figure out what to focus when modal opens

jandrade:
Modified logic to change focus placement:

a) By default, try to focus on the first focusable element inside the Dialog container.
b) In a given scenario, no focusable elements could be found, so we will focus on the Dialog element itself.

For reference, see:
https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal

## Test plan
`yarn start`

### a) Default scenario
- Open http://localhost:6060/#!/Modal/1
- Open modal and make sure the focus is set on the `<input />` element.

### b) Alternative scenario
- Open http://localhost:6060/#!/ModalLauncher/3
- Open modal and make sure the focus is set on the `<div role="dialog"/>` element.


jeresig:

Built with commit a3ecea81d592854847043943485849cea9d61f75

https://deploy-preview-402--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-backdrop.js
```diff
@@ -26,25 +33,30 @@ type Props = {|
...
+                node.querySelector('[role="dialog"]'),
+            ): any);
+            // add tabIndex to make the Dialog focusable
+            firstFocusableElement.tabIndex = -1;
         }
+
         // wait for styles to applied
         setTimeout(() => {
-            lastButton.focus();
+            firstFocusableElement.focus();
```
kevinbarabash: It seems like `firstFocusableElement` could be `null`.

packages/wonder-blocks-modal/components/modal-backdrop.test.js
```diff
@@ -88,10 +92,30 @@ describe("ModalBackdrop", () => {
 
         // Act
         await sleep(); // wait for styles to be applied
-        const lastButton = wrapper.find("[data-last-button]").getDOMNode();
+        const focusableElement = wrapper
+            .find("[data-first-button]")
```
kevinbarabash: This `data` attribute is just for testing purposes right?

packages/wonder-blocks-modal/components/modal-backdrop.test.js
```diff
@@ -88,10 +92,30 @@ describe("ModalBackdrop", () => {
...
+            .getDOMNode();
+
+        // Assert
+        expect(document.activeElement).toBe(focusableElement);
+    });
+
+    test("If there are no buttons, we focus the Dialog instead", async () => {
```
kevinbarabash: Cool!

kevinbarabash: Can you document this behavior in the modal docs somewhere?

packages/wonder-blocks-modal/components/modal-backdrop.js
kevinbarabash: I see, you set it to `[role="dialog"]` if that's the case.

packages/wonder-blocks-modal/components/modal-backdrop.test.js
jandrade: yes, just for testing

packages/wonder-blocks-modal/docs.md
```diff
@@ -15,6 +16,64 @@ When a dialog opens, focus moves to an element inside the dialog. See notes belo
     - If focus is on the first tabbable element inside the dialog, moves focus to the last tabbable element inside the dialog.
 - Escape: Closes the dialog.
 
+#### Initial focus placement:
+The initial focus placement depends on the following scenarios:
+
+1. `initialFocusId` (default): `ModalLauncher` exposes this prop as a string. The dialog will try to find this element into the DOM. If it's found, focus is initially set on this element.
+2. focusable elements: This is the second scenario, where the dialog tries to find the first ocurrence of possible focusable elements.
+3. Dialog: If the first two conditions are not met, then focus is initially set to the Dialog element itself.
```
kevinbarabash: Nice!

packages/wonder-blocks-modal/docs.md
```diff
@@ -15,6 +16,64 @@ When a dialog opens, focus moves to an element inside the dialog. See notes belo
...
+                </Button>
+            </React.Fragment>
+        }
+    />
+);
+
+<View style={{flexDirection: "row"}}>
+    <ModalLauncher
+        onClose={() => window.alert("you closed the modal")}
+        initialFocusId="[data-initial-focus]"
```
kevinbarabash: This is a little confusing.  I was expecting to specify an id not a selector.

packages/wonder-blocks-modal/components/modal-backdrop.test.js
```diff
@@ -16,6 +16,19 @@ const exampleModal = (
...
+                <input type="text" />
+                <button data-initial-focus />
+            </div>
+        }
+        title="Title"
+        footer={<div data-modal-footer />}
+    />
+);
+
 const exampleModalWithButtons = (
```
kevinbarabash: It think it's preferable to inline these snippets even if it adds some duplicate code.  It makes the tests easier to read.  Also, we try to avoid DRY-ing out test code.

packages/wonder-blocks-modal/components/modal-backdrop.js
```diff
@@ -26,28 +38,68 @@ type Props = {|
...
-        if (!lastButton) {
-            return;
-        }
+        const firstFocusableElement =
+            // 1. try to get element specified by the user
+            this._getInitialFocusElement(node) ||
+            // 2. get first occurence from list of focusable elements
+            this._getFirstFocusableElement(node) ||
+            // 3. get the dialog itself
+            this._getDialogElement(node);
```
kevinbarabash: Nice!  This is really clean and documents the behavior, thank you.

packages/wonder-blocks-modal/components/modal-backdrop.test.js
jandrade: Good point, thanks for clarifying on this :)

packages/wonder-blocks-modal/docs.md
jandrade: That's a valid point. Using IDs makes total sense b/c that way we should only be looking for one single node.

On the other side, we could use selectors, and add some documentation to recommend devs to use custom data attributes tight to WB (?). 

What do you think it would be the best approach?

1. initialFocusId="some-specific-id", then inside ModalBackdrop using `node.querySelector(`#${initialFocusId}`)` b/c for some weird reason `getElementId` is not included in `HTMLElement`.
vs
2. initialFocusId="[data-wb-modal-initial-focus]" and adding some extra documentation to recommend devs following these kind of naming conventions. For reference you can see what cypress does (https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements).

kevinbarabash: My vote is for the option 1.

jandrade: Here it is :) 
https://github.com/Khan/wonder-blocks/pull/402/files#diff-0a507f47a429306bd08c332dda512cb6R70

PR #403, opened 2019-05-30
[patch:modal] WB-566 Modals and Portals example

jandrade:
## Changes
- Adds an example verifying the parent of a modal be able to update the contents within the modal.
- Small styling tweaks in documentation.

## Test plan
- Open:
    - Staging: https://deploy-preview-403--wonder-blocks.netlify.com/#!/Modal/3
    - Local Dev: http://localhost:6060/#!/Modal/3
- Click on `Open Modal` and close it multiple times.
- Verify the modal title and content are updated every time the modal is reopened.


packages/wonder-blocks-modal/components/one-pane-dialog/one-pane-dialog.md
```diff
@@ -224,46 +224,28 @@ const defaultStyles = StyleSheet.create({
...
-const smallStyles = StyleSheet.create({
-    footer: {
-        justifyContent: "flex-end"
-    },
-    label: {
-        display: "none"
-    }
-});
-
-const styleSheets = {
```
jandrade: Small tweak here: removed mobile styles (by Nefaur's request).

jeresig:

Built with commit 9ed3a00dbe79f59eced3b371daeacf413223c3e9

https://deploy-preview-403--wonder-blocks.netlify.com

kevinbarabash:
The example should test whether the modal can be updated from the parent while the modal is open.  There are a variety of uses cases where this is useful.  Some examples: a multi-step "wizard" dialog; the exercises modal where you have a start card, one or more exercises, and a summary card at the end.

I think something like the following would be an accurate test to verify that modal v2 can meet these use cases:
```
type Props = {
   questions: Array<React.Node>,
};

type State = {
  currentQuestion: number,
};

class ExerciseModal extends React.Component<Props, State> {
   state = {currentQuestion: 0};

   handleNextButton = () => {
      this.setState({
          currentQuestion: Math.max(state.currentQuestion + 1, props.questions.length - 1),
      });
   };
   
   handlePreviousButton = () => { ... }

   render() {
       return <OnePaneDialog
          title="Exercises"
          content={this.props.questions[this.state.currentQuestion]}
          footer={<Toolbar>
              <Button onClick={this.handlePreviousButton}>Previous</Button>
              <Button onClick={this.handleNextButton}>Next</Button>
          </Toolbar>}
      />;
  }
}

<ModalLauncher
    modal={<ExerciseModal questions={["question 1", "question 2", "question 3"]}/>}
    onClose={...}
>
   {({openModal}) => <Button onClick={openModal}>Start Exercises</Button>
</ModalLauncher>
```
Note: I've omitted some of the implementation details.

jandrade:
That makes sense. I initially tried something similar, but I wanted to create that example that external state modifies contents inside the portal. This is the current way Modals work:

`Parent > ModalLauncher > Portal > OnePaneDialog > dialog contents`

In your example, it seems that everything is changing inside the portal:

`ModalLauncher > Portal > Parent > OnePaneDialog > dialog contents`

Just for better understanding, Is that what we want to test? I'm asking this mostly because I'm not completely aware of the use cases inside Webapp. If your example emulates the way we use modals inside webapp, then it's perfect for that. If not, I can reuse your example but modifying the structure to make sure everything is changing inside the portal (?). I hope my point is clear enough :)

What do you think?

kevinbarabash:
Good point.  I think the
```
Parent > ModalLauncher > Portal > OnePaneDialog > dialog contents
```
case is the one to worry about.  Keeping the modal open while the parent's state is changing to update the contents of the modal is important.

jandrade:
Added updated example here: https://github.com/Khan/wonder-blocks/pull/403/files#diff-ce3d90a10495ee44fc17d86b7e754a0eR120

PR #404, opened 2019-05-31
[feat:core,modal] WB-576: Create IDprovider component

jandrade:
## Core - IDProvider
- Moved `UniqueDialog` from `modal` to `core` and renamed it to `IDProvider`.
- Refactored `IDProvider` to be used by different components/packages.
- Added documentation example.
- Updated unit tests.

## Test plan
1. Run `yarn test`.
2. Open `http://localhost:6060/#!/IDProvider/1` and make sure the example works as expected.

### DIFF
--- a/packages/wonder-blocks-core/components/id-provider.js
+++ b/packages/wonder-blocks-core/components/id-provider.js
@@ -1,14 +1,15 @@
 // @flow
 import * as React from "react";
 
-import {UniqueIDProvider} from "@khanacademy/wonder-blocks-core";
-import type {IIdentifierFactory} from "@khanacademy/wonder-blocks-core";
+import UniqueIDProvider from "./unique-id-provider.js";
+
+import type {IIdentifierFactory} from "../util/types.js";
 
 type Props = {|
     /**
      * Use the children-as-function pattern to pass a uniqueId string for
-     * use anywhere within children. This provides a way of adding a unique title
-     * to an existant modal variation (OnePaneDialog) or a custom modal implementation.
+     * use anywhere within children. This provides a way of adding a unique identifier
+     * to a given component for a11y purposes.
      */
     children: (uniqueId: string) => React.Node,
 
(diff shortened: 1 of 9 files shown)


jeresig:

Built with commit f2cd3c1c5d7d673aa85b157c2c919154358f14e8

https://deploy-preview-404--wonder-blocks.netlify.com

packages/wonder-blocks-core/components/id-provider.md
```diff
@@ -0,0 +1,18 @@
+### Example: Usage with form fields
+
+This example allows you to generate an unique ID and make it available to associate the `<label>` and `<input>` elements. To see this example in action, check that `label[for]` and `input[id]` are using the same id.
+
+```jsx
+const children = (uniqueId) => (
```
jeresig: It might be slightly more idiomatic to put this inline below, rather than having `children` be a separate variable. So something like:

```
<IDProvider scope="field">
    {(uniqueId) => (
        <label htmlFor={uniqueId}>
            ...
        </label>
    )}
</IDProvider>
```

packages/wonder-blocks-core/components/id-provider.js
```diff
@@ -30,24 +31,25 @@ type Props = {|
 |};
 
 /**
- * This is a Dialog wrapper that returns a unique Modal Title for A11Y purposes.
- * This way, the wrapped Dialog will receive this custom ID and will use it to refer to a
- * visible dialog title.
+ * This is a wrapper that returns a unique Identifier for A11Y purposes.
```
jeresig: Now that this is public it might be good to also explain when this should be used contrasted with UniqueIDProvider.

packages/wonder-blocks-core/components/id-provider.md
jandrade: Makes total sense, I will do it right away

packages/wonder-blocks-core/components/id-provider.js
somewhatabstract: Agreed. I think it is likely that IDProvider is the preferred API in most cases, with UniqueIDProvider being a "break the glass" option when implementing more complex scenarios.

somewhatabstract: Also, I don't think `for A11Y purposes` is necessary here. There are many reasons one may need this behavior; a11y is just one of them.

packages/wonder-blocks-core/components/id-provider.md
```diff
@@ -0,0 +1,16 @@
...
+    <IDProvider scope="field">
+        {(uniqueId) => (
+            <label htmlFor={uniqueId}>
+                Label with ID {uniqueId}:
+                <input type="text" id={uniqueId} />
+            </label>
+        )}
+    </IDProvider>
+</View>
+```
```
somewhatabstract: I think another example where the id prop is used would also be helpful. Perhaps a scenario where we have a component that takes an optional `id` prop, which then gets passed internally to the `IDProvider`, which then either uses that `id` prop if it was set, but otherwise defaults to a unique ID. This would show why this component exists; the idea that it's for cases where you need an ID, so if one isn't given, it will enforce one.

PR #405, opened 2019-05-31
[feat:dropdown] WB-577: Add id and aria-labelledby as props to SingleSelect & MultiSelect

jandrade:
## Dropdown

Added `id` and `aria-labelledby` as props to **SingleSelect** & **MultiSelect** components (examples included).
- `id`: It's used in case an external component/element needs to reference to the current `Select` component. e.g. `<label for="some-id">...</label><SingleSelect id="some-id"/>`
- `aria-labelledby`: It's used in the opposite way. In case the current `Select` component needs to reference to an external component/element  e.g. `<label id="some-label-id">...</label><SingleSelect aria-labelledby="some-label-id"/>`

#### Local env:
1. `yarn`
2. `yarn start`
3. Open the following URLs:
    - SingleSelect: http://localhost:6060/#!/SingleSelect/13
    - MultiSelect: http://localhost:6060/#!/MultiSelect/11
4. Make sure both examples are rendered properly.

#### Staging env:
Open the following URLs:
- SingleSelect: https://deploy-preview-405--wonder-blocks.netlify.com/#!/SingleSelect/13
- MultiSelect: https://deploy-preview-405--wonder-blocks.netlify.com/#!/MultiSelect/11


jeresig:

Built with commit ef5bedc9f0d037c16ecb0e1c620f4fcc17a82835

https://deploy-preview-405--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -262,7 +271,9 @@ export default class MultiSelect extends React.Component<Props, State> {
     render() {
         const {
             alignment,
+            "aria-labelledby": ariaLabelledBy,
             disabled,
+            id,
```
kevinbarabash: I think we should use `...restProps` or `...otherProps` to pass through props otherwise we'll continue adding additional things to pass through.  We can control what's accepted via `AriaProps` and/or some other shared props type.

packages/wonder-blocks-dropdown/components/single-select.md
```diff
@@ -301,3 +301,31 @@ const styles = StyleSheet.create({
     <SingleSelect menuText="Empty" placeholder="empty" />
 </View>
 ```
+
+### Accessibility
+
+If you need to associate this component with another element (e.g. `<label>`),
+make sure to pass the `aria-labelledby` and/or `id` props to the `SingleSelect` component.
```
kevinbarabash: Given that we'll want to add these props to every component eventually I think we should probably just do that.  Having examples for each component to show that they handle common props seems like a lot of overhead.  Having some tests that we can automatically determine if `id`, `aria-labelledby`, etc. got set on a particular element within a component may be a more sustainable approach.

It would be good to have a top-level .md file for the whole repo that documents what the common props are on our components.  Not sure how to make that work with styleguidist though.  This is something for a separate PR though.

jandrade: That's a great point, I'm going to create a separate ticket to work on that.

packages/wonder-blocks-dropdown/components/multi-select.js
jandrade: Great point, I'm going to work on that as part of this ticket: https://khanacademy.atlassian.net/browse/WB-580

That way I'll be revisiting all the cases where we need to use this approach!

packages/wonder-blocks-dropdown/components/single-select.md
jandrade: https://khanacademy.atlassian.net/browse/WB-583

MrNickBreen: Follow up diff where ariaProps are added to Wonderblocks components: https://github.com/Khan/wonder-blocks/pull/411#pullrequestreview-245110722

packages/wonder-blocks-dropdown/components/multi-select.js
jandrade: @kevinbarabash As you suggested, I added this functionality as part of WB-180 (#411)

kevinbarabash: Awesome, thanks!

PR #407, opened 2019-06-03
WB-569 Implement Timeout class

somewhatabstract:
This is the first diff in the implementation of ADR-77.

Here, I have added the new `wonder-blocks-timing` package and implemented the `setTimeout` wrapper class, `Timeout`. Eventually, there will be things exported from this new package, but this class is not directly made available per the API defined in ADR-77.

Once the other pieces are in place, we will have a package we can release.


jeresig:

Built with commit cfed2e86ada295be34b14930817c2dfa342fafba

https://deploy-preview-407--wonder-blocks.netlify.com

packages/wonder-blocks-timing/docs.md
```diff
@@ -0,0 +1 @@
+Abstractions for common timing APIs like `setTimeout`, `setInterval` and `requestAnimationFrame`.
```
somewhatabstract: Docs will be fleshed out once the actual package is finished and there are exported components.

packages/wonder-blocks-timing/index.js
```diff
@@ -0,0 +1,4 @@
+// @flow
+import type {ITimeout} from "./util/types.js";
+
+export type {ITimeout};
```
somewhatabstract: There will be more here once more of the package is implemented.

CODEOWNERS
```diff
@@ -6,5 +6,6 @@ packages/wonder-blocks-core/*   @somewhatabstract @kevinbarabash @jeresig
 packages/wonder-blocks-icon-button/*   @kevinbarabash @jeresig
 packages/wonder-blocks-link/*   @kevinbarabash @jeresig
 packages/wonder-blocks-modal/*   @jandrade @kevinbarabash @jeresig
+packages/wonder-blocks-timers/*   @somewhatabstract @jeresig
```
jeresig: Err, timing right?

packages/wonder-blocks-timing/package.json
```diff
@@ -0,0 +1,23 @@
...
+    },
+    "description": "",
+    "main": "dist/index.js",
+    "module": "dist/es/index.js",
+    "source": "index.js",
+    "scripts": {
+      "test": "echo \"Error: no test specified\" && exit 1"
+    },
+    "peerDependencies": {
+      "react": "^16.4.1"
```
jeresig: Doesn't seem like this is needed?

styleguide.config.js
```diff
@@ -154,6 +154,11 @@ module.exports = {
             name: "Spacing",
             content: "packages/wonder-blocks-spacing/docs.md",
         },
+        {
+            private: true,
+            name: "Timing",
```
jeresig: IMO, I think this might make sense as an external package - but having this be in WB for now is totally fine by me!

packages/wonder-blocks-timing/package.json
somewhatabstract: It will be once this package is finished.

styleguide.config.js
somewhatabstract: Yes, this private should be in the package.json. I'll make a change.

Once the package is finished, we can make it public.

PR #410, opened 2019-06-03
WEB-571: Added AnimationFrame, IAnimationFrame

somewhatabstract:
More implementation for ADR-77.

### DIFF
--- a/packages/wonder-blocks-timing/util/animation-frame.js
+++ b/packages/wonder-blocks-timing/util/animation-frame.js
@@ -1,0 +1,86 @@
+// @flow
+import type {IAnimationFrame} from "./types.js";
+
+/**
+ * Encapsulates everything associated with calling requestAnimationFrame/
+ * cancelAnimationFrame, and managing the lifecycle of that request, including
+ * the ability to resolve or cancel a pending request action.
+ *
+ * @export
+ * @class AnimationFrame
+ * @implements {IAnimationFrame}
+ */
+export default class AnimationFrame implements IAnimationFrame {
+    _animationFrameId: ?AnimationFrameID;
+    _action: () => mixed;
+
+    /**
+     * Creates an animation frame request that will invoke the given action.
+     * The request is not made until set is called.
+     *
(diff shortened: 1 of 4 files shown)


jeresig:

Built with commit 4d5d1147430295923fdca1b5c3c9960d1736ccdf

https://deploy-preview-410--wonder-blocks.netlify.com

packages/wonder-blocks-timing/util/animation-frame.js
```diff
@@ -0,0 +1,86 @@
...
+     * invoking the action. If no request is pending, this does nothing.
+     *
+     * @param {boolean} [resolve] When true, if the request was pending when
+     * called, the request action is invoked after cancelling the timeout.
+     * Defaults to false.
+     *
+     * @returns {void}
+     * @memberof AnimationFrame
+     */
+    clear(resolve?: boolean): void {
```
jandrade: I'm not fully aware of the detailed implementation, but I'm seeing a lot of similitudes when defining this `clear` method between all the classes (AnimFrame, Interval and Timeout). The only difference I see is the Web API function called (cancelAnimationFrame) in this case.

Do you think it would make sense to create an abstract base class that could be used by these classes/APIs and internally handle this logic?

somewhatabstract: So, I have thought about this and I decided that coupling the implementations would mostly be a negative. It would make the code harder to understand and would potentially incur potential for some nasty bugs due to the subtle differences between each of the three APIs.

For example, timeouts and animation frames fire just once, intervals do not; timeouts and intervals have an associated time value, frames do not. Just from writing tests I found some subtle implementation differences that this incurs and otherwise introduces bugs. Having all the pertinent code in one place feels right.

I also considered having them all implement the same interface, since the method signatures are the same for each one. However, an interface is a contract not only on what things are called and their signatures, but also what they do, and again, in each case there are subtle differences that I felt it was important to keep separate.

Do you see a specific win we would get by sharing implementation details?

PR #413, opened 2019-06-05
[feat:modal]WB-585: Add building blocks to public API

jandrade:
## Modals
- Added `ModalDialog, ModalHeader, ModalFooter, ModalPanel` to the public API.
- Included `Modals > Building blocks` section into the documentation website.
- Added descriptions to: `ModalDialog, ModalHeader, ModalFooter, ModalPanel`

### Test plan
- Open https://deploy-preview-413--wonder-blocks.netlify.com/#!/Building%20Blocks

### DIFF
--- a/packages/wonder-blocks-modal/components/modal-header.js
+++ b/packages/wonder-blocks-modal/components/modal-header.js
@@ -48,7 +48,7 @@ type Props = {|
  *
  * **Accessibility notes:**
  *
- * - By default (e.g. using [OnePaneDialog](/#onepanedialog)), `titleId` is populated automaticallyby the parent container.
+ * - By default (e.g. using [OnePaneDialog](/#onepanedialog)), `titleId` is populated automatically by the parent container.
  * - If there is a custom Dialog implementation (e.g. `TwoPaneDialog`), the ModalHeader doesn’t have to have
  * the `titleId` prop however this is recommended. It should match the `aria-labelledby` prop of the [ModalDialog](/#modaldialog) component.
  * If you want to see an example of how to generate this ID, check [IDProvider](/#idprovider).
--- a/packages/wonder-blocks-modal/components/modal-header.js
+++ b/packages/wonder-blocks-modal/components/modal-header.js
@@ -39,6 +39,39 @@ type Props = {|
     titleId: string,
 |};
 
+/**
+ * This is a helper component that is never rendered by itself.
+ * It is always pinned to the top of the dialog, is responsive using the same behavior as its parent dialog,
+ * and has the following properties:
+ * - title
+ * - breadcrumb OR 2B subtitle, but not both.
+ *
+ * **Accessibility notes:**
+ *
+ * - By default (e.g. using `OnePaneDialog`), `titleId` is populated automaticallyby the parent container.
+ * - If there is a custom Dialog implementation (e.g. `TwoPaneDialog`), the ModalHeader doesn’t have to have
+ * the `titleId` prop however this is recommended. It should match the `aria-labelledby` prop of the ModalDialog component.
+ *
+ * **Implementation notes:**
+ *
+ * If you are creating a custom Dialog, make sure to follow these guidelines:
+ * - Make sure to include it as part of `<ModalPanel />` by using the `header` prop.
+ * - Add a title (required).
+ * -  Optionally add a subtitle or breadcrumbs.
+ * - We encourage you to add `titleId` (see Accessibility notes).
+ * - If the `ModalPanel` has a dark background, make sure to set `light` to `false`.
+ *
+ * Example:
+ *
+ * ```js
+ * <ModalHeader
+ *      title="Sidebar using ModalHeader"
+ *      subtitle="subtitle"
+ *      titleId="uniqueTitleId"
+ *      light={false}
+ *  />
+ * ```
+ */
 export default class ModalHeader extends React.Component<Props> {
     static defaultProps = {
         light: true,
--- a/packages/wonder-blocks-modal/components/modal-header.js
+++ b/packages/wonder-blocks-modal/components/modal-header.js
@@ -44,20 +44,21 @@ type Props = {|
  * It is always pinned to the top of the dialog, is responsive using the same behavior as its parent dialog,
  * and has the following properties:
  * - title
- * - breadcrumb OR 2B subtitle, but not both.
+ * - breadcrumb OR subtitle, but not both.
  *
  * **Accessibility notes:**
  *
- * - By default (e.g. using `OnePaneDialog`), `titleId` is populated automaticallyby the parent container.
+ * - By default (e.g. using [OnePaneDialog](/#onepanedialog)), `titleId` is populated automaticallyby the parent container.
  * - If there is a custom Dialog implementation (e.g. `TwoPaneDialog`), the ModalHeader doesn’t have to have
- * the `titleId` prop however this is recommended. It should match the `aria-labelledby` prop of the ModalDialog component.
+ * the `titleId` prop however this is recommended. It should match the `aria-labelledby` prop of the [ModalDialog](/#modaldialog) component.
+ * If you want to see an example of how to generate this ID, check [IDProvider](/#idprovider).
  *
  * **Implementation notes:**
  *
  * If you are creating a custom Dialog, make sure to follow these guidelines:
- * - Make sure to include it as part of `<ModalPanel />` by using the `header` prop.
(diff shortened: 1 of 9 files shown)


jeresig:

Built with commit cd2135e5af540e41799a2aa06e6b3a9912ebaf6b

https://deploy-preview-413--wonder-blocks.netlify.com

jandrade:
I was just thinking about how devs can use these Building Blocks, mostly because they are not part of  the `wb-modals` public API.

Do you think is ok to add a message about how to import these files? and if this would be the correct way of importing them?

e.g.
If you need to import these Building blocks, make sure to import them directly:
```
import ModalHeader from `@khanacademy/wonder-blocks-modal/components/modal-header`;
...
```

kevinbarabash:
I don't think
```
import ModalHeader from `@khanacademy/wonder-blocks-modal/components/modal-header`;
```
will actually work, because that code is uncompiled and webapp isn't set up to compile anything in `node_modules`.  I think any of the components that are used in the flexible modal example should be exported from `index.js` and anything that's exported should have styleguidist docs.  Things that aren't exported shouldn't have styleguidist docs.

kevinbarabash:
Actually, we may as well export `ModalHeader` and `ModalFooter` since designers will be expected to be able to use the building blocks and we want to avoid a situation where we have to keep updating wonder-blocks-modal to export things that people are expecting to be usable.

styleguide.config.js
```diff
@@ -130,14 +130,11 @@ module.exports = {
                     ],
                 },
                 {
-                    // NOTE(jeresig): Not ready to be used yet, need docs.
-                    private: true,
```
somewhatabstract: If this is internal only stuff that isn't part of the public API, then it should remain marked private so that it only appears in the dev server version of docs that goes up for PRs and local dev; rather than in the published public docs on the main wonder-blocks netlify.

packages/wonder-blocks-modal/components/modal-header.js
```diff
@@ -39,6 +39,39 @@ type Props = {|
...
+ * - By default (e.g. using `OnePaneDialog`), `titleId` is populated automaticallyby the parent container.
+ * - If there is a custom Dialog implementation (e.g. `TwoPaneDialog`), the ModalHeader doesn’t have to have
+ * the `titleId` prop however this is recommended. It should match the `aria-labelledby` prop of the ModalDialog component.
+ *
+ * **Implementation notes:**
+ *
+ * If you are creating a custom Dialog, make sure to follow these guidelines:
+ * - Make sure to include it as part of `<ModalPanel />` by using the `header` prop.
+ * - Add a title (required).
+ * -  Optionally add a subtitle or breadcrumbs.
```
kevinbarabash: nit: remove extra space

packages/wonder-blocks-modal/components/modal-header.js
```diff
@@ -39,6 +39,39 @@ type Props = {|
...
+ * It is always pinned to the top of the dialog, is responsive using the same behavior as its parent dialog,
+ * and has the following properties:
+ * - title
+ * - breadcrumb OR 2B subtitle, but not both.
+ *
+ * **Accessibility notes:**
+ *
+ * - By default (e.g. using `OnePaneDialog`), `titleId` is populated automaticallyby the parent container.
+ * - If there is a custom Dialog implementation (e.g. `TwoPaneDialog`), the ModalHeader doesn’t have to have
+ * the `titleId` prop however this is recommended. It should match the `aria-labelledby` prop of the ModalDialog component.
```
kevinbarabash: Maybe link to the `UniqueIdProvider` (or is it `IdProvider` now).

packages/wonder-blocks-modal/components/modal-panel.js
```diff
@@ -60,6 +60,21 @@ type Props = {|
     onClose?: () => mixed,
 |};
 
+/**
+ * ModalPanel is  the content container.
+ *
+ * **Implementation notes:**
+ *
+ * If you are creating a custom Dialog, make sure to follow these guidelines:
+ * - Make sure to add this component inside the **ModalDialog**.
```
kevinbarabash: It would be nice if there was a way to cross link to `ModalDialog`.  No action necessary for this PR, but maybe something to look at for the docs in general.

styleguide.config.js
jandrade: That's one of the big changes we are doing with v2. Devs now should be able (under certain conditions) to create their own modal variations. That's the main reason why we want to expose these building blocks. You can take a look at the source code for this example: https://deploy-preview-413--wonder-blocks.netlify.com/#!/Building%20Blocks/1

That said, I'm thinking there could be a risk of being so flexible with modal layouts, as this could cause some fragmentation in WB Modals created inside Webapp. Nefaur expressed this concern as well. Maybe we would need to revisit this.

What do you think?

kevinbarabash: To avoid fragmentation I think some patterns could be codified such as the two-column layout which you prototyped.  That layout is used in a number of places in webapp.  One question though is whether it makes sense to update the existing two column layouts to use the new version you prototype.  That's a question for design.

styleguide.config.js
```diff
@@ -130,14 +130,11 @@ module.exports = {
...
                     name: "Building Blocks",
                     content: "packages/wonder-blocks-modal/building-blocks.md",
```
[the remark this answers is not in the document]
jandrade: No worries at all, same happened to me. I noticed that after finishing the implementation and adding some examples 😄

packages/wonder-blocks-modal/generated-snapshot.test.js
```diff
@@ -12,7 +12,6 @@ import ModalLauncher from "./components/modal-launcher.js";
 import OnePaneDialog from "./components/one-pane-dialog/one-pane-dialog.js";
 import ModalDialog from "./components/modal-dialog.js";
 import ModalPanel from "./components/modal-panel.js";
-import ModalContent from "./components/modal-content.js";
```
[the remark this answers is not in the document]
jandrade: I suspect that is b/c I removed that from `styleguide.config.js`. I noticed I didn't need to add it in the TwoPaneDialog example.

Then, It would make sense(?) if we don't expose this block, mostly because it's only used/consumed internally by `ModalPanel`. Any thoughts?

packages/wonder-blocks-modal/components/modal-header.js
```diff
@@ -39,6 +39,39 @@ type Props = {|
     titleId: string,
 |};
 
+/**
+ * This is a helper component that is never rendered by itself.
+ * It is always pinned to the top of the dialog, is responsive using the same behavior as its parent dialog,
+ * and has the following properties:
+ * - title
+ * - breadcrumb OR 2B subtitle, but not both.
```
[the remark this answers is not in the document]
jandrade: Opps copypasta 🤦‍♂. Nefaur added that in the design specs. I'll remove it right away.

packages/wonder-blocks-modal/components/modal-panel.js
jandrade: Good point! I'm going to add a quick fix in the meantime. But it would be great to revisit that in general for sure!

packages/wonder-blocks-modal/components/modal-header.js
```diff
@@ -39,6 +39,40 @@ type Props = {|
...
+/**
+ * This is a helper component that is never rendered by itself.
+ * It is always pinned to the top of the dialog, is responsive using the same behavior as its parent dialog,
+ * and has the following properties:
+ * - title
+ * - breadcrumb OR subtitle, but not both.
+ *
+ * **Accessibility notes:**
+ *
+ * - By default (e.g. using [OnePaneDialog](/#onepanedialog)), `titleId` is populated automaticallyby the parent container.
```
jeresig: *automatically by

packages/wonder-blocks-modal/index.js
```diff
@@ -1,6 +1,18 @@
...
+import ModalDialog from "./components/modal-dialog.js";
+import ModalFooter from "./components/modal-footer.js";
+import ModalHeader from "./components/modal-header.js";
 import ModalLauncher from "./components/modal-launcher.js";
+import ModalPanel from "./components/modal-panel.js";
 import OnePaneDialog from "./components/one-pane-dialog/one-pane-dialog.js";
 import maybeGetPortalMountedModalHostElement from "./util/maybe-get-portal-mounted-modal-host-element.js";
 
-export {ModalLauncher, OnePaneDialog, maybeGetPortalMountedModalHostElement};
+export {
```
jeresig: Yayy

PR #414, opened 2019-06-05
[fix:dropdown] WB-443: Change ActionMenu from bold to regular

kumailn:


jeresig:

Built with commit 182cb7b01c2e04484e07843188ba54c8c9f1f98a

https://deploy-preview-414--wonder-blocks.netlify.com

jandrade:
It looks great to me!

Just one little comment that doesn't affect these changes:
I changed the PR title to know which component and what type of change you are including. This is helpful for when we have to publish a new WB version. Not sure if this is documented somewhere, but if not I will add this info into our confluence Space :).

kevinbarabash:
@kumailn even though the title of this change is pretty self explanatory it's good to get in the habit of including a short summary of the changes.  In this case linking to the Jira ticket and the design spec would be good.  A screenshot of before and after would be even better.

PR #415, opened 2019-06-05
WB-529: Add support for tag prop in View component

kumailn:
Added support for the 'tag' prop in view components to override the default div tag if needed. Changed StyledDiv declaration from const to let because I reassign it later if the tag prop is present. Pulled out tag prop from restProps for child safety once the StyledDiv is adjusted (although I don't know if this is necessary).

### DIFF
--- a/packages/wonder-blocks-core/components/view.js
+++ b/packages/wonder-blocks-core/components/view.js
@@ -29,7 +29,7 @@ type ValidViewTags = "div" | "article" | "aside" | "nav" | "section";
 // eslint-disable-next-line flowtype/require-exact-type
 type Props = {
     ...TextViewSharedProps,
-    tag?: ValidViewTags,
+    tag: ValidViewTags,
 };
 
 const StyledDiv = addStyle<"div">("div", styles.default);
@@ -52,6 +52,10 @@ const StyledSection = addStyle<"section">("section", styles.default);
  * - An array combining the above
  */
 export default class View extends React.Component<Props> {
+    static defaultProps = {
+        tag: "div",
+    };
+
     render() {
         const {testId, tag, ...restProps} = this.props;
         switch (tag) {
@@ -63,6 +67,8 @@ export default class View extends React.Component<Props> {
                 return <StyledNav data-test-id={testId} {...restProps} />;
             case "section":
                 return <StyledSection data-test-id={testId} {...restProps} />;
+            case "div":
+                return <StyledDiv data-test-id={testId} {...restProps} />;
             default:
                 return <StyledDiv data-test-id={testId} {...restProps} />;
         }
--- a/packages/wonder-blocks-core/components/view.js
+++ b/packages/wonder-blocks-core/components/view.js
@@ -25,7 +25,18 @@ const styles = StyleSheet.create({
     },
 });
 
-let StyledDiv = addStyle<"div">("div", styles.default);
+type ValidViewTags = "div" | "article" | "aside" | "nav" | "section";
+// eslint-disable-next-line flowtype/require-exact-type
+type Props = {
+    ...TextViewSharedProps,
+    tag?: ValidViewTags,
+};
+
+const StyledDiv = addStyle<"div">("div", styles.default);
+const StyledArticle = addStyle<"article">("article", styles.default);
+const StyledAside = addStyle<"aside">("aside", styles.default);
+const StyledNav = addStyle<"nav">("nav", styles.default);
+const StyledSection = addStyle<"section">("section", styles.default);
 
 /**
  * View is a building block for constructing other components. `View` roughly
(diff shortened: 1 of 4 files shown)


jeresig:

Built with commit a493ace25e37e9b871c5bce202b857af3ac1e62f

https://deploy-preview-415--wonder-blocks.netlify.com

packages/wonder-blocks-core/util/types.js
```diff
@@ -210,6 +210,8 @@ export type TextViewSharedProps = {
 
     id?: string,
 
+    tag?: string,
```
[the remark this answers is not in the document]
kevinbarabash: Good call.  There's definitely some tags that don't make sense.  This means we won't be able to share the type for `tag` between `View` and `Text` since different tags make sense for these components.

kevinbarabash:
There's a new uncovered line in the coverage report.  Can you add a test for this new line?
Screen Shot 2019-06-06 at 11 58 08 AM

packages/wonder-blocks-core/components/view.js
```diff
@@ -25,7 +25,7 @@ const styles = StyleSheet.create({
     },
 });
 
-const StyledDiv = addStyle<"div">("div", styles.default);
+let StyledDiv = addStyle<"div">("div", styles.default);
```
somewhatabstract: I'd like to keep this const and use a different variable than reuse this one.

packages/wonder-blocks-core/components/view.js
```diff
@@ -42,7 +42,11 @@ const StyledDiv = addStyle<"div">("div", styles.default);
  */
 export default class View extends React.Component<TextViewSharedProps> {
     render() {
-        const {testId, ...restProps} = this.props;
+        if (this.props.tag) {
+            StyledDiv = addStyle<string>(this.props.tag, styles.default);
```
somewhatabstract: Instead of overriding the variable, I think we should use a different one and do something like.

```
const StyledTag = tag ? addStyle... : StyledDiv;
```
(`addStyle...` being a placeholder in this example for whatever the call is with the appropriate type for allowed tags)
 
Then render with `StyledTag`.

What do you think?

packages/wonder-blocks-core/components/view.js
```diff
@@ -40,9 +51,20 @@ const StyledDiv = addStyle<"div">("div", styles.default);
...
  * - An array combining the above
  */
-export default class View extends React.Component<TextViewSharedProps> {
+export default class View extends React.Component<Props> {
     render() {
-        const {testId, ...restProps} = this.props;
-        return <StyledDiv data-test-id={testId} {...restProps} />;
+        const {testId, tag, ...restProps} = this.props;
+        switch (tag) {
+            case "article":
```
kevinbarabash: Now that we have explicit cases we need to test each of them separately in view.test.js.

packages/wonder-blocks-core/components/view.js
```diff
@@ -25,7 +25,18 @@ const styles = StyleSheet.create({
     },
 });
 
+type ValidViewTags = "div" | "article" | "aside" | "nav" | "section";
+// eslint-disable-next-line flowtype/require-exact-type
+type Props = {
+    ...TextViewSharedProps,
+    tag?: ValidViewTags,
```
kevinbarabash: We should make this required here and the set `"div"` to be the default value using `defaultProps`.

packages/wonder-blocks-core/components/view.js
```diff
@@ -40,9 +51,20 @@ const StyledDiv = addStyle<"div">("div", styles.default);
...
+            case "article":
+                return <StyledArticle data-test-id={testId} {...restProps} />;
+            case "aside":
+                return <StyledAside data-test-id={testId} {...restProps} />;
+            case "nav":
+                return <StyledNav data-test-id={testId} {...restProps} />;
+            case "section":
+                return <StyledSection data-test-id={testId} {...restProps} />;
+            default:
+                return <StyledDiv data-test-id={testId} {...restProps} />;
```
kevinbarabash: Once `tag` is required can you add a `case "div": ...` and then follow https://medium.com/@ibosz/advance-flow-type-1-exhaustive-checking-with-empty-type-a02e503cd3a0 so that we can use `flow` to ensure this switch statement is exhaustive.  The benefit of this is that if we update `ValidViewTags`, flow will let us know to update the switch/case statement here.

packages/wonder-blocks-core/components/view.js
jandrade: Is it possible to avoid using a switch statement here?

```
const StyledTag = addStyle<string>(tag, styles.default);
return <StyledTag data-test-id={testId} {...restProps} />;
```

Not sure if this is one of flow's limitations about type checking. Mostly b/c the tests won't fail unless we use `// @flow`.
cc @kevinbarabash

kevinbarabash: Unfortunately that won't work b/c `tag` is typed as `"article" | "section" | ...` which incompatible with `string`.  Also, I think that calling `addStyle` in `render()` is probably bad practice since it creates a new class each time which feel expensive.

packages/wonder-blocks-core/components/view.js
```diff
@@ -40,9 +51,26 @@ const StyledDiv = addStyle<"div">("div", styles.default);
...
+            case "aside":
+                return <StyledAside data-test-id={testId} {...restProps} />;
+            case "nav":
+                return <StyledNav data-test-id={testId} {...restProps} />;
+            case "section":
+                return <StyledSection data-test-id={testId} {...restProps} />;
+            case "div":
+                return <StyledDiv data-test-id={testId} {...restProps} />;
+            default:
+                return <StyledDiv data-test-id={testId} {...restProps} />;
```
kevinbarabash: The `default` case should look something like:
```
default:
    (tag: empty);
    throw Error(`${tag} is not an allowed value for the 'tag' prop`);
```
The `(tag: empty);` line tries to case `tag` to `empty` which is never valid.  If the switch/case is exhaustive flow nows that this will never happen and does not raise an error.  Try removing one of the other cases to see what happens.

The `throw` isn't really necessary since all of our code is typed and that should never happen but since there's an expectation that `render()` return a React.Node it's better to throw to indicate that this path doesn't return a value and that's okay.  It would be cool if the tools were smart enough to know that it isn't necessary to return a value from this path, but they're not there yet.

kevinbarabash:
I don't think we need to test the uncovered lines in the screenshot since flow has assured us that those lines never get called.
Screen Shot 2019-06-07 at 2 02 47 PM

PR #416, opened 2019-06-06
[fix:button,dropdown,icon-button] WB-535: Remove aria-disabled when disabled is present

jandrade:
## Button, Dropdown, IconButton
Removed `aria-disabled` as prop when the `disabled` attribute is present.

### Test plan
`yarn test`

### DIFF
--- a/packages/wonder-blocks-button/components/button-core.js
+++ b/packages/wonder-blocks-button/components/button-core.js
@@ -90,13 +90,10 @@ export default class ButtonCore extends React.Component<Props> {
         };
 
         if (commonProps["aria-disabled"]) {
-            // eslint-disable-next-line no-console
-            console.warn(
-                "wb-button: <button> supports the 'disabled' attribute. You don't need to use the 'aria-disabled' attribute.",
-            );
-
             // WB-535: Allows only to use `disabled` (even if is set from the parent component)
-            commonProps["aria-disabled"] = undefined;
+            throw new Error(
+                "wb-button: <button> supports the 'disabled' attribute. You should not use the 'aria-disabled' attribute.",
+            );
         }
 
         const Label = size === "small" ? LabelSmall : LabelLarge;
--- a/packages/wonder-blocks-button/components/button-core.js
+++ b/packages/wonder-blocks-button/components/button-core.js
@@ -83,11 +83,22 @@ export default class ButtonCore extends React.Component<Props> {
         ];
 
         const commonProps = {
+            "aria-disabled": undefined,
             "data-test-id": testId,
             role: "button",
             style: [defaultStyle, style],
             ...handlers,
         };
+
+        if (commonProps["aria-disabled"]) {
+            // eslint-disable-next-line no-console
+            console.warn(
+                "wb-button: <button> supports the 'disabled' attribute. You don't need to use the 'aria-disabled' attribute.",
+            );
+
+            // WB-535: Allows only to use `disabled` (even if is set from the parent component)
+            commonProps["aria-disabled"] = undefined;
+        }
 
         const Label = size === "small" ? LabelSmall : LabelLarge;
 
@@ -136,8 +147,6 @@ export default class ButtonCore extends React.Component<Props> {
                     type="button"
                     {...commonProps}
                     disabled={disabled}
-                    // WB-535: Allows only to use `disabled`
-                    aria-disabled={undefined}
                 >
                     {contents}
                 </StyledButton>
--- a/packages/wonder-blocks-button/components/button-core.js
+++ b/packages/wonder-blocks-button/components/button-core.js
@@ -83,7 +83,6 @@ export default class ButtonCore extends React.Component<Props> {
         ];
 
         const commonProps = {
-            "aria-disabled": undefined,
             "data-test-id": testId,
             role: "button",
             style: [defaultStyle, style],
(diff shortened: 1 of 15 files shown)


jeresig:

Built with commit 589f78794fd3346b70ffd0c7f9e3979a9e19902b

https://deploy-preview-416--wonder-blocks.netlify.com

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -84,14 +83,22 @@ export default class ButtonCore extends React.Component<Props> {
...
-            "aria-label": ariaLabel,
             "data-test-id": testId,
             role: "button",
             style: [defaultStyle, style],
             ...handlers,
         };
 
+        if (commonProps["aria-disabled"]) {
+            // eslint-disable-next-line no-console
+            console.warn(
```
jeresig: I wonder if we should make this a lot angrier? Since this is a problem if it gets to the client. I'd be open to making this: `throw new Error("wb-button: ...});` and removing the logic after it. We want to make sure people don't do this!

jeresig: I think we should either complain (in which case we throw an exception) or we make `aria-disabled` do the right thing (deleting the original value and setting the `disabled` prop to be the right value).

jeresig: I think I'd err on the side of making it do the right thing and complaining if the two values don't match up for some reason.

packages/wonder-blocks-button/components/button.md
```diff
@@ -120,6 +120,8 @@ const styles = StyleSheet.create({
 </View>
 ```
 
+**Note:** If you are adding `disabled` you don't need to add `aria-disabled`. Doing so will result in a console.warn().
```
jeresig: It might be worth adding a bit more context here to help people understand *why* this is undesirable.

packages/wonder-blocks-icon-button/components/icon-button-core.js
```diff
@@ -98,14 +97,21 @@ export default class IconButtonCore extends React.Component<Props> {
...
         const commonProps = {
-            // TODO(kevinb): figure out a better way of forward ARIA props
-            "aria-disabled": disabled ? "true" : undefined,
-            "aria-label": ariaLabel,
             "data-test-id": testId,
             style: [defaultStyle, style],
             ...handlers,
         };
 
+        if (commonProps["aria-disabled"]) {
```
jeresig: Same here, too.

packages/wonder-blocks-button/components/button.md
jeresig: It's not just that you don't need it, it's that it can confuse the screen readers, I believe?

packages/wonder-blocks-button/components/button-core.js
kevinbarabash: I think we might be able to use flow's `$Rest` utility type to construct `Prop`s that omit `aria-disabled` for components have a `disabled` prop, e.g.
```
type Props = {|
   ...$Rest<AriaProps, {| "aria-disabled": boolean |},
   disabled: boolean,
   ...
|}
```

jandrade: @kevinbarabash Tried with `$Rest` and it works only if the prop is required, but the problem is that all `AriaProps` attributes are optional. I have to use `"aria-disabled"?` to map this key to the original `AriaProps` object. Not sure if I'm missing something on the implementation. 

`...$Rest<AriaProps, {| "aria-disabled"?: "false"|"true" |}>,`

kevinbarabash: That makes sense.  I forgot about the`"false"|"true"` that used in `AriaProps`.

kevinbarabash: There's a note in the [docs](https://flow.org/en/docs/types/utilities/#toc-diff) for `$Diff` (which is the inverse? of `$Rest`) on how to get around the issue of optionality:
```
type A = $Diff<{}, {nope: number}>; // Error
type B = $Diff<{}, {nope: number \| void}>; // OK
```
I tried the following and flow accepts it and finds the error:
```
type Props = {|
    ...$Rest<AriaProps, {| "aria-disabled": "true" | "false" | void |}>,
    disabled: boolean,
|};

const props: Props = {
    "aria-disabled": "true",
    disabled: true,
};
```

kevinbarabash: I wish Flow had TypeScript's `Omit` helper which lets you do:
```
type Props = Omit<AriaProps, "aria-disable"> & { disabled: boolean, ... };
```
https://devblogs.microsoft.com/typescript/announcing-typescript-3-5/#the-omit-helper-type

One thing `$Rest` has going for it is that you can remove multiple properties.

packages/wonder-blocks-button/components/button.md
```diff
@@ -79,7 +83,19 @@ const styles = StyleSheet.create({
 </View>
 ```
 
-Buttons can be `disabled`.
+#### Example: disabled
+
+Buttons can be `disabled`:
+
+⚠️ Buttons do not need an `aria-disabled` attribute, if it also has a `disabled` attribute.
```
jeresig: Yay!

packages/wonder-blocks-button/components/button.md
```diff
@@ -336,6 +362,8 @@ const styles = StyleSheet.create({
 </View>
 ```
 
+#### Example: style
```
jeresig: Thank you for adding all of these headings.

jandrade:
Good call. I added that to `action-menu-opener` as well.

https://github.com/Khan/wonder-blocks/pull/416/files#diff-82d1803e8e64d92ad5a5e859c45b9d8cR9

PR #419, opened 2019-06-09
[WB-572] ActionScheduler implementation completion

somewhatabstract:
- Added error cases for the timeout, interval, and animation frame types
- Implemented `ActionScheduler` class that provides the API we'll expose via the `ActionSchedulerComponent` (to be implemented as part of WB-573)
- Added unit tests for functionality

### DIFF
--- a/packages/wonder-blocks-timing/util/action-scheduler.js
+++ b/packages/wonder-blocks-timing/util/action-scheduler.js
@@ -1,0 +1,59 @@
+// @flow
+import {
+    IAnimationFrame,
+    IInterval,
+    ITimeout,
+    IScheduleActions,
+} from "./types.js";
+
+import Timeout from "./timeout.js";
+import Interval from "./interval.js";
+import AnimationFrame from "./animation-frame.js";
+
+/**
+ * Implements the `IScheduleActions` API to provide timeout, interval, and
+ * animation frame support.
+ */
+export default class ActionScheduler implements IScheduleActions {
+    _registeredActions: Array<() => void> = [];
+
+    timeout(
+        action: () => mixed,
+        period: number,
+        autoSet?: boolean,
+        resolveOnClear?: boolean,
+    ): ITimeout {
+        const timeout = new Timeout(action, period, autoSet);
+        this._registeredActions.push(() => timeout.clear(resolveOnClear));
+        return timeout;
+    }
+
+    interval(
+        action: () => mixed,
+        period: number,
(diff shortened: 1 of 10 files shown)


jeresig:

Built with commit c24b8b1f8845951b17dc7c4611c0abfff67779de

https://deploy-preview-419--wonder-blocks.netlify.com

packages/wonder-blocks-timing/util/animation-frame.js
```diff
@@ -25,6 +25,10 @@ export default class AnimationFrame implements IAnimationFrame {
      * @memberof AnimationFrame
      */
     constructor(action: () => mixed, autoSet?: boolean) {
+        if (typeof action !== "function") {
```
jandrade: For learning purposes, is there a way in `flow` to define that `action` has a function signature and avoid using `typeof` to check this?

somewhatabstract: That is already done, but `flow` is only a code-time check, not a runtime check. Not everyone that consumes our code is necessarily going to be using flow or type checking. Since this is a public facing API, I want it to be robust.

jandrade: That's an excellent point! thanks for clarifying.

packages/wonder-blocks-timing/util/action-scheduler.js
```diff
@@ -0,0 +1,59 @@
+// @flow
```
jeresig: Might be helpful to have a comment here to explain the context in which this might be used.

packages/wonder-blocks-timing/util/types.js
```diff
@@ -129,3 +129,102 @@ export interface IAnimationFrame {
      */
     clear(resolve?: boolean): void;
 }
+
+/**
```
jeresig: Maybe this comment would be useful in that file?

packages/wonder-blocks-timing/util/action-scheduler.js
somewhatabstract: Well, no one should be using this directly, but yeah, I can make that explicit.

packages/wonder-blocks-timing/util/types.js
somewhatabstract: Not sure. The interface is public API facing, but the implementation is not.

PR #420, opened 2019-06-09
[tooling] Update to flow 0.100.0

somewhatabstract:
- Add flow-bin 0.100.0

- Update `.flowconfig` to match

- Add flow suppressions now that flow is spotting we're using mocks off things that it doesn't know we have mocks of

### DIFF
--- a/packages/wonder-blocks-grid/components/row.test.js
+++ b/packages/wonder-blocks-grid/components/row.test.js
@@ -14,8 +14,10 @@ describe("Row", () => {
     beforeEach(() => {
         unmountAll();
 
-        // eslint-disable-next-line no-console
+        /* eslint-disable no-console */
+        // Flow doesn't know we added jest mocks to this $FlowFixMe
         console.error.mockRestore && console.error.mockRestore();
+        /* eslint-enable no-console */
     });
 
     describe("large", () => {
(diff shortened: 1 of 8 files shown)


jeresig:

Built with commit 5034036f7b7a8ba6cce797d7385a13ea3d791f9d

https://deploy-preview-420--wonder-blocks.netlify.com

packages/wonder-blocks-grid/components/row.test.js
```diff
@@ -112,8 +116,10 @@ describe("Row", () => {
 
             // Prevent jest complaining when JSDOM console.error's an exception
             jest.spyOn(console, "error");
-            // eslint-disable-next-line no-console
+            /* eslint-disable no-console */
```
jandrade: Do you think is a good idea removing these lines by chaining `mockImplementation` to `jest.spyOn`?

e.g `jest.spyOn(console, "error").mockImplementationOnce(() => {});`

somewhatabstract:
> LGTM. I think we could avoid the flow errors by using the return value of `jest.spyOn` when restoring the mock instead, but that can be a separate PR. Thanks for updating flow.

Yeah, but I think that is a worse solution as this is just a change to tell flow it's ok, rather than changing how we code the test to avoid its shortcomings. I'd rather avoid anything that has us sharing variables between test cases and things if we can help it, but you are, of course, correct.

Mind you, maybe we nest the `afterEach` inside the `beforeEach` 😈

PR #421, opened 2019-06-10
 WB-302: Implement Hoverable tooltip bubbles

kumailn:
Implemented hoverable tooltip bubbles by enabling pointerEvents in tooltip-bubble.js styles, the hover ability works by passing the reference of the timeout to the bubble, which occurs when hovering outside of the anchor element. If the bubble is hovered on in time, it cancels the timeout thus keeping the tooltip-bubble on screen indefinitely until the user engages in the `mouseleave` event.


jeresig:

Built with commit cf9c17135a9fd995e4fd9262918f240bd79e8a28

https://deploy-preview-421--wonder-blocks.netlify.com

somewhatabstract:
Haven't looked at the code in-depth yet, but I just spent a few minutes testing out the netlify site, and things look great! The <kbd>Esc</kbd> key is working right to dismiss the tooltip still, even when hovered. Everything feels natural. Nice.

Behaviourally, I think this is great. I'll try to look at the code later, but if others say the code is good in the meantime, I'm happy.

somewhatabstract:
That said, looks like there's perhaps some additional testing we could implement for this. I see a small drop in coverage which suggests this new behavior may not have enzyme tests. Recommend implementing some.

kumailn:
Yes I was planning to implement tests to increase coverage but wanted to get this PR up as soon as I could so I could get feedback. As for functionality, from what I've tested it retains pretty much everything from the original version.

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
```diff
@@ -37,12 +38,61 @@ export type Props = {|
 
     /** The `TooltipContent` element that will be rendered in the bubble. */
     children: React.Element<typeof TooltipContent>,
+
+    onActiveChanged: (active: boolean) => mixed,
+
+    onTimeoutChanged: (timeoutID: ?TimeoutID) => mixed,
+
+    anchorTimeoutID: ?TimeoutID,
```
kevinbarabash: These props should have comments.

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
```diff
@@ -37,12 +38,61 @@ export type Props = {|
...
+    componentDidMount() {
+        const bubbleNode = ReactDOM.findDOMNode(this);
+
+        if (!bubbleNode) {
+            return;
+        }
+
+        bubbleNode.addEventListener("mouseenter", this._handleMouseEnter);
+        bubbleNode.addEventListener("mouseleave", this._handleMouseLeave);
+        //console.log("Bubbble was mounted");
```
kevinbarabash: remove

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -252,7 +260,9 @@ export default class TooltipAnchor extends React.Component<Props, State>
             this._timeoutID = setTimeout(() => {
                 this._timeoutID = null;
                 this._setActiveState(active, true);
+                this.props.onTimeoutChanged(null);
             }, delay);
+            this.props.onTimeoutChanged(this._timeoutID);
```
kevinbarabash: What happens if `instant` is `true`?  Do we need to call `this.props.onTimeoutChanged` in that case as well?

kumailn: Thanks for the catch, I don't believe we would need to actually, but I fixed it up in the latest commit regardless.

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
```diff
@@ -37,12 +38,61 @@ export type Props = {|
...
+    };
+
+    componentDidMount() {
+        const bubbleNode = ReactDOM.findDOMNode(this);
+
+        if (!bubbleNode) {
+            return;
+        }
+
+        bubbleNode.addEventListener("mouseenter", this._handleMouseEnter);
```
jandrade: Maybe I don't have the full context here... I'd like to understand what is the reasoning behind using `addEventListener` vs using `onMouseEnter` as a prop attached to the main `View`.

kumailn: It's mainly because I was keeping the code consistent with tooltip-anchor.js, which uses `addEventListener`. However `onMouseEnter` should work perfectly fine too, I can change it if needed.

jandrade: Oh ok, I checked `tooltip-anchor.js` and for your case I think it's easier/better to use `onMouseEnter`. That way, you don't need to check if the element is already present in the DOM, and also don't need to add `removeEventListener` or any extra lifecycle events (`componentDidMount` and `componentWillUnmount` for the `removeEventListener` case)  b/c React handles that internally.

packages/wonder-blocks-tooltip/components/tooltip.integration.test.js
```diff
@@ -42,14 +45,50 @@ describe("tooltip integration tests", () => {
         const anchor = anchorWrapper.getDOMNode();
         // start hovering
         anchor && anchor.dispatchEvent(new FocusEvent("mouseenter"));
+        anchor && anchor.dispatchEvent(new FocusEvent("mouseleave"));
         expect(wrapper.state("timeoutID")).toEqual(expect.any(Number));
 
+        const bubbleTimeoutCheck = (callback) =>
+            setTimeout(() => {
+                expect(wrapper.state("timeoutID")).toEqual(expect.any(Number));
```
kevinbarabash: Instead of doing the `expect` inside of the timeout, a more reliable approach would be to add a local variable to the test and set the value inside the timeout and move the `expect` to the bottom of the the test.

packages/wonder-blocks-tooltip/components/tooltip.integration.test.js
```diff
@@ -42,14 +45,50 @@ describe("tooltip integration tests", () => {
...
+                callback && callback();
+            }, TooltipDisappearanceDelay / 2);
+
+        bubbleTimeoutCheck(callback);
```
kevinbarabash: This timeout check seems completely separate from `Tooltip` and `TooltipBubble`.  Am I missing something?

kevinbarabash: It looks like it waits a bit of time before doing the `expect`.  Since the code is based on timers, calling `jest.advanceTimersByTime(TooltipDisappearanceDelay / 2);` should be sufficient.

packages/wonder-blocks-tooltip/components/tooltip.integration.test.js
```diff
@@ -42,14 +45,50 @@ describe("tooltip integration tests", () => {
...
+    it("should eventually disable the TooltipBubble after mouseleave", () => {
+        const callback = jest.fn();
+        const wrapper = mount(
+            <Tooltip content="hello, world">an anchor</Tooltip>,
+        );
+
+        wrapper.setState({active: true});
+
+        const bubbleWrapper = wrapper.find(TooltipBubble);
+        expect(bubbleWrapper.length).toEqual(1);
```
kevinbarabash: It's safe to drop these `expects`.

packages/wonder-blocks-tooltip/components/tooltip.integration.test.js
kevinbarabash:
```
expect(wrapper.state("timeoutID")).toEqual(expect.any(Number));
```
was asserted on line 49, shouldn't that be sufficient?

packages/wonder-blocks-tooltip/components/tooltip.integration.test.js
```diff
@@ -42,14 +45,50 @@ describe("tooltip integration tests", () => {
...
+
+        const anchorWrapper = wrapper.find(TooltipAnchor);
+        expect(anchorWrapper.length).toEqual(1);
+
+        const anchor = anchorWrapper.getDOMNode();
+        // start hovering
+        anchor && anchor.dispatchEvent(new FocusEvent("mouseenter"));
         anchor && anchor.dispatchEvent(new FocusEvent("mouseleave"));
+        expect(wrapper.state("timeoutID")).toEqual(expect.any(Number));
```
kevinbarabash: This was already tested In the previous test.

packages/wonder-blocks-tooltip/components/tooltip.integration.test.js
```diff
@@ -42,14 +45,50 @@ describe("tooltip integration tests", () => {
...
+        const bubbleTimeoutCheck = (callback) =>
+            setTimeout(() => {
+                expect(wrapper.state("timeoutID")).toEqual(null);
+                callback && callback();
+            }, TooltipDisappearanceDelay + 10);
 
-        expect(wrapper).toHaveState("active", true);
-        expect(anchorWrapper).toHaveState("active", false);
+        bubbleTimeoutCheck(callback);
+        jest.advanceTimersByTime(TooltipDisappearanceDelay + 10);
```
kevinbarabash: Same thing here: calling `jest.advanceTimersByTime` should be sufficient.

packages/wonder-blocks-tooltip/components/tooltip.integration.test.js
```diff
@@ -0,0 +1,100 @@
...
+            <Tooltip content="hello, world">an anchor</Tooltip>,
+        );
+
+        wrapper.setState({active: true});
+
+        const bubbleWrapper = wrapper.find(TooltipBubble);
+        expect(bubbleWrapper.length).toEqual(1);
+
+        const anchorWrapper = wrapper.find(TooltipAnchor);
+        expect(anchorWrapper.length).toEqual(1);
```
kevinbarabash: I'd remove these expects for `bubbleWrapper` and `anchorWrapper` from this test since this test is specifically looking for `timeoutID`.

packages/wonder-blocks-tooltip/components/tooltip-anchor.js
```diff
@@ -269,6 +280,7 @@ export default class TooltipAnchor extends React.Component<Props, State>
     };
 
     _handleMouseLeave = () => {
+        // debugger;
```
jandrade: maybe you can remove this line

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
```diff
@@ -60,6 +98,8 @@ export default class TooltipBubble extends React.Component<Props> {
                 id={id}
                 role="tooltip"
                 data-placement={placement}
+                onMouseEnter={() => this._handleMouseEnter()}
```
jandrade: you might don't need to wrap this with an arrow function.

This should work: `onMouseEnter={this._handleMouseEnter}`

package.json
```diff
@@ -71,7 +71,7 @@
     "flow-coverage-report": "^0.5.0",
     "glob": "^7.1.2",
     "jest": "^23.5.0",
-    "jest-enzyme": "^6.0.2",
+    "jest-enzyme": "^7.0.2",
```
kevinbarabash: The old version didn't support `toContainMatchingElement`.

flow-typed/npm/jest_v23.x.x.js
```diff
@@ -227,6 +228,11 @@ type JestJQueryMatchersType = {
     toHaveDescendantWithText(sel: string, text: string | RegExp): void,
 };
 
+// TODO(WEB-902): create flow types for jest-enzyme
```
kevinbarabash: I brought this task into this sprint b/c we want people to use jest-enzyme matchers when writing tests and they can't do that if flow is going to complain about them.

packages/wonder-blocks-tooltip/components/tooltip.integration.test.js
```diff
@@ -0,0 +1,108 @@
...
+        anchor && anchor.dispatchEvent(new FocusEvent("mouseenter"));
+        // There's a 100ms delay before TooltipAnchor calls _setActiveState with
+        // instant set to true.  This second call is what actually triggers the
+        // call to this.props.onActiveChanged() which updates Tooltip's active
+        // state.
+        jest.runAllTimers();
+        // Since the call to update Tooltip's active state happens in a timeout
+        // we need to call update b/c it happens outside of the normal React lifecycle
+        // methods.
+        wrapper.update();
```
kevinbarabash: Without this call to `.update()` the test will fail.  Normally we'd wouldn't need all of this when simulating events but `TooltipAnchor` event handling is more complicated than the usually event handling.

packages/wonder-blocks-tooltip/components/tooltip.integration.test.js
```diff
@@ -0,0 +1,108 @@
...
+
+    it("should set timeoutId be null when TooltipBubble is active", () => {
+        // Arrange
+        const wrapper = mount(
+            <Tooltip content="hello, world">an anchor</Tooltip>,
+        );
+        const anchor = wrapper.find(TooltipAnchor).getDOMNode();
+
+        // Act
+        anchor && anchor.dispatchEvent(new FocusEvent("mouseenter"));
```
kevinbarabash: Simulating the events is better than for integration tests than setting state since it exercises more parts of the system.  In this case we're verifying that the event handlers are behaving as we expect.

packages/wonder-blocks-tooltip/components/tooltip.integration.test.js
```diff
@@ -0,0 +1,108 @@
...
+        // we need to call update b/c it happens outside of the normal React lifecycle
+        // methods.
+        wrapper.update();
+
+        // Assert
+        expect(wrapper).toContainMatchingElement(TooltipBubble);
+        expect(wrapper).toHaveState("timeoutID", null);
+    });
+
+    it("should set a timeout on mouseleave on TooltipAnchor", () => {
```
kevinbarabash: I re-enabled this test and update it to use similar techniques as the first test.

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
```diff
@@ -37,12 +37,50 @@ export type Props = {|
...
+type State = {|
+    active: boolean,
+|};
+
+export default class TooltipBubble extends React.Component<Props, State> {
+    state = {
+        active: false,
+    };
+
+    _updateActiveState(hovered: boolean) {
```
jandrade: I think this function can be removed or merged with `_setActiveState` b/c both functions are being called in one place, right?

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
```diff
@@ -37,12 +37,50 @@ export type Props = {|
...
+        this._setActiveState(hovered);
+    }
+
+    _setActiveState(active: boolean, instant?: boolean) {
```
jandrade: I see you have the `instant` parameter defined, but I don't see where is it called inside the function body.

Is ok to remove that parameter?

packages/wonder-blocks-tooltip/components/tooltip.integration.test.js
```diff
@@ -0,0 +1,106 @@
...
+        wrapper.update();
+
+        const bubbleWrapper = wrapper.find(TooltipBubble);
+
+        // Act
+        bubbleWrapper.simulate("mouseleave");
+
+        // Assert
+        expect(wrapper.state("timeoutID")).toEqual(null);
+        expect(wrapper.state("active")).toEqual(false);
```
kevinbarabash: Could you add a `expect` here to verify that `TooltipBubble` is not longer being displayed?

packages/wonder-blocks-tooltip/components/tooltip.integration.test.js
```diff
@@ -0,0 +1,106 @@
...
+        anchor && anchor.dispatchEvent(new FocusEvent("mouseleave"));
+        jest.advanceTimersByTime(TooltipDisappearanceDelay / 2);
+        wrapper.update();
+
+        bubbleWrapper.simulate("mouseenter");
+        wrapper.update();
+
+        // Assert
+        expect(wrapper.state("timeoutID")).toEqual(null);
+        expect(wrapper.state("active")).toEqual(true);
```
kevinbarabash: Could you advance the timers by `TooltipDisappearanceDelay + 100` or something like that and check that the `TooltipButton` is still there?

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
```diff
@@ -98,8 +94,10 @@ export default class TooltipBubble extends React.Component<Props, State> {
                 id={id}
                 role="tooltip"
                 data-placement={placement}
-                onMouseEnter={() => this._handleMouseEnter()}
-                onMouseLeave={() => this._handleMouseLeave()}
+                // eslint-disable-next-line react/jsx-handler-names
```
jandrade: I'd suggest to remove this comment and rename that function to something that eslint could accept. Maybe something like `this.handleMouseEnter`?

packages/wonder-blocks-tooltip/components/tooltip.integration.test.js
```diff
@@ -0,0 +1,109 @@
...
+        jest.advanceTimersByTime(TooltipDisappearanceDelay / 2);
+        wrapper.update();
+
+        bubbleWrapper.simulate("mouseenter");
+        jest.advanceTimersByTime(TooltipDisappearanceDelay + 100);
+        wrapper.update();
+
+        // Assert
+        expect(wrapper.state("timeoutID")).toEqual(null);
+        expect(wrapper.find(TooltipBubble).exists()).toBeTruthy();
```
kevinbarabash: This could be:
```
expect(wrapper).toContainMatchingElement(TooltipBubble);
```

packages/wonder-blocks-tooltip/components/tooltip.integration.test.js
```diff
@@ -0,0 +1,109 @@
...
+        wrapper.update();
+
+        const bubbleWrapper = wrapper.find(TooltipBubble);
+
+        // Act
+        bubbleWrapper.simulate("mouseleave");
+
+        // Assert
+        expect(wrapper.state("timeoutID")).toEqual(null);
+        expect(wrapper.find(TooltipBubble).exists()).toBeFalsy();
```
kevinbarabash: This could be:
```
expect(wrapper).not.toContainMatchingElement(TooltipBubble);
```

flow-typed/npm/jest_v23.x.x.js
```diff
@@ -1,5 +1,6 @@
 // flow-typed signature: f5a484315a3dea13d273645306e4076a
 // flow-typed version: 7c5d14b3d4/jest_v23.x.x/flow_>=v0.39.x
```
kevinbarabash: #426 updates jest to 24 which includes a new version of this file with the updated jest-enzyme matchers.

PR #422, opened 2019-06-10
Allow for tests to mock console.error, call console.error and the like, but only output log stuff on test failure

somewhatabstract:
Our global mock and throw of `console.error` was causing some test failures to not report properly, which made figuring them out really hard. Also, we had some strange test patterns where we mocked console.error just to hide it from the test output.

This adds a new reporter (based on the one in webapp but modified for the older version of jest we currently use) that only outputs console stuff if a test fails. It also removes the global mock. If we want to fail if anything calls console.error, I think we need to think it through differently, since it may be entirely legitimate for that to occur in tests and the global mock is really fragile since any test could rest that mock without even knowing it existed.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,5 @@
 [version]
-0.93.0
+0.100.0
 
 [ignore]
 <PROJECT_ROOT>/node_modules/react-element-to-jsx-string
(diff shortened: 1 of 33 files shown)


jeresig:

Built with commit 4bf1c9508ff5f75ff3db917932b6b6656313d02a

https://deploy-preview-422--wonder-blocks.netlify.com

somewhatabstract:
> Maybe we could consider later if is a good idea using `restoreMocks`. I'm not sure right now about the implications of using this config option in our environment. https://jestjs.io/docs/en/configuration.html#restoremocks-boolean.

Yes, I agree. We should definitely be doing that.

For context, my preferred jest config is to have `timers` as `fake` and `restoreMocks` set `true`. I feel it's a path of least surprise and most versatility in testing.

PR #424, opened 2019-06-11
[dropdown:fix] use props.open in action-menu-opener-core.js instead of global open

kevinbarabash:
Jira: https://khanacademy.atlassian.net/browse/LP-5535

We were using the global `open` which is always truthy so `aria-expanded` was always being set to `"true"`.  This PR fixes this by using `props.open` instead.  The fixtures have been updated which should be sufficient to show that the issue has been fixed.

### DIFF
--- a/packages/wonder-blocks-dropdown/__snapshots__/generated-snapshot.test.js.snap
+++ b/packages/wonder-blocks-dropdown/__snapshots__/generated-snapshot.test.js.snap
@@ -44,7 +44,7 @@ exports[`wonder-blocks-dropdown example 1 1`] = `
     }
   >
     <button
-      aria-expanded="true"
+      aria-expanded="false"
       aria-haspopup="menu"
       className=""
       data-test-id="teacher-menu"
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-dropdown/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -62,7 +62,6 @@ exports[`wonder-blocks-dropdown example 1 1`] = `
       onTouchCancel={[Function]}
       onTouchEnd={[Function]}
       onTouchStart={[Function]}
-      open={false}
```
kevinbarabash: The `open` prop wasn't being destructured so it was being included when the sharedProps were spread into this button.  Note: each `aria-expanded` should not correspond to whatever the value `open` was in the same button.

jeresig:

Built with commit c9519b3ed0d5d488eea082c68f4c1b5fc0cf9f3b

https://deploy-preview-424--wonder-blocks.netlify.com

PR #426, opened 2019-06-13
upgrade jest and jest-enzyme

kevinbarabash:
I upgraded jest b/c the flow definitions for jest 24 include definitions for jest-enzyme which is better than us adding definitions for jest-enzyme matchers ourselves.

As part of this change I had to tweak some of our config files b/c some of the options changed.  Also, was able to use an unmodified version of log-on-fail-reporter.js from webapp.


jeresig:

Built with commit 1ad40586cd83b4915e6c98f2796c52fd16f1a7af

https://deploy-preview-426--wonder-blocks.netlify.com

PR #427, opened 2019-06-14
Enforce ModalHeader constraints using flow

kevinbarabash:


packages/wonder-blocks-modal/components/one-pane-dialog/one-pane-dialog.js
```diff
@@ -100,12 +115,35 @@ export default class OnePaneDialog extends React.Component<Props> {
         closeButtonVisible: true,
     };
 
+    renderHeader(uniqueId: string): React.Element<typeof ModalHeader> {
+        const {title, subtitle, breadcrumbs} = this.props;
+
+        if (breadcrumbs) {
```
kevinbarabash: `breadcrumbs` is `void | React.Element<typeof Breadcrumbs>` so the type refinement should work but for some reason I had to add a cast below.

packages/wonder-blocks-modal/components/one-pane-dialog/one-pane-dialog.js
```diff
@@ -100,12 +115,35 @@ export default class OnePaneDialog extends React.Component<Props> {
...
+            return (
+                <ModalHeader
+                    title={title}
+                    breadcrumbs={
+                        (breadcrumbs: React.Element<typeof Breadcrumbs>)
+                    }
+                    titleId={uniqueId}
+                />
+            );
+        } else if (subtitle) {
```
kevinbarabash: same

jeresig:

Built with commit bde0986d4feab6acea79a6bcfcbc589329e7c50f

https://deploy-preview-427--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-header.js
```diff
@@ -11,34 +11,48 @@ import {
...
     /**
-     * Adds a breadcrumb-trail, appearing in the ModalHeader, above the title.
+     * An id to provide a selector for the title element.
      */
-    breadcrumbs?: React.Element<typeof Breadcrumbs>,
+    titleId: string,
+
+    // Keep flow happy
+    subtitle?: void,
+    breadcrumbs?: void,
```
kevinbarabash: Without these, flow complains about `subtitle` and `breadcrumbs` not being available on props at all b/c these are exact object types, flow looks for `subtitle` in each of `Common`, `WithSubtitle` and `WithBreadcrumbs`.  I also tried making the types inexact and flow still complains when destructuring which is a little odd.

jeresig: It might be worth putting this exact comment inline here in the code.

kevinbarabash:
@jeresig I search webapp for all uses of `OnePaneDialog` and it's being used correctly in both places: https://github.com/Khan/webapp/search?q=OnePaneDialog&unscoped_q=OnePaneDialog.

PR #428, opened 2019-06-14
WB-587: [feature:dropdown] Custom Dropdown use case prototype 

kumailn:
This is a prototype of the `Dropdown` component which allows custom dropdowns to be implemented. The implementation of the `Dropdown` is consistent with other components which implement `DropdownCore`. Thoughts / feedback?


jeresig:

Built with commit ee0435ff1daf3e05e629c1dc5ec225533597fd8b

https://deploy-preview-428--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
```diff
@@ -0,0 +1,244 @@
...
+            <span
+                ref={this.handleOpenerRef}
+                disabled={!dropdownItems.length || disabled}
+            >
+                {children(this._openDropdown)}
+            </span>
+        );
+
+        return (
+            <Dropdown
```
kevinbarabash: `Dropdown` is really starting to feel like it should be called `DropdownCore` and `DropdownLauncher` should probably be renamed to `DropdownCore` or simple `Dropdown`.  Don't worry about doing this in this PR but something to consider when writing up the implementation spec.

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
```diff
@@ -0,0 +1,244 @@
...
+
+type Props = {|
+    ...AriaProps,
+
+    /*
+     * The callback function used to open the Dropdown.
+     */
+    children: (openDropdown: () => void) => React.Element<any>,
+
+    opened?: boolean,
```
jandrade: Could you please add a comment explaining this prop?

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
```diff
@@ -0,0 +1,244 @@
...
+
+    /*
+     * Text for the opener of this menu.
+     */
+    menuText: string,
```
kevinbarabash: `menuText` doesn't make sense with this component and can be removed.

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
```diff
@@ -0,0 +1,244 @@
...
+import Dropdown from "./dropdown.js";
+import type {Item, DropdownItem} from "../util/types.js";
+import ActionItem from "./action-item.js";
+import OptionItem from "./option-item.js";
+
+type Props = {|
+    ...AriaProps,
+
+    /*
+     * The callback function used to open the Dropdown.
```
kevinbarabash: This isn't correct.  It's a render callback which we pass an `openDropdown` function to.  It is responsible for rendering the opener.

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
```diff
@@ -0,0 +1,244 @@
...
+     * Whether this menu should be left-aligned or right-aligned with the
+     * opener component. Defaults to left-aligned.
+     */
+    alignment: "left" | "right",
+
+    /**
+     * Whether this component is disabled. A disabled dropdown may not be opened
+     * and does not support interaction. Defaults to false.
+     */
+    disabled: boolean,
```
kevinbarabash: How will this be forwarded to the opener?

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
```diff
@@ -0,0 +1,244 @@
...
+                alignment={alignment}
+                items={this._getMenuItems()}
+                keyboard={this.state.keyboard}
+                dropdownStyle={[styles.menuTopSpace, dropdownStyle]}
+                onOpenChanged={this.handleOpenChanged}
+                open={this.state.open}
+                opener={opener}
+                openerElement={this.openerElement}
+                role="menu"
+                style={style}
```
kevinbarabash: Does the `style` prop work in the example?  Can you set `marginLeft: 100` to see if applies the style correctly?

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
```diff
@@ -0,0 +1,244 @@
...
+    };
+
+    handleOpenChanged = (open: boolean, keyboard?: boolean) => {
+        this.setState({
+            open,
+            keyboard,
+        });
+    };
+
+    handleOptionSelected = (selectedValue: string) => {
```
kevinbarabash: Can you ask Nefaur which menu items need to be available in custom dropdown (we might not need this)?

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
jandrade: Also, I'm not sure if you are using this prop somewhere inside this component.

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
```diff
@@ -0,0 +1,244 @@
...
+
+    static defaultProps = {
+        alignment: "left",
+        disabled: false,
+    };
+
+    constructor(props: Props) {
+        super(props);
+        this.state = {
+            open: false,
```
jandrade: Should `open` be synchronized with the `opened` prop? e.g. If `opened` is set to `true`, then `state.open` should be `true` as well?

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
```diff
@@ -0,0 +1,244 @@
...
+    }
+}
+
+const styles = StyleSheet.create({
+    caret: {
+        marginLeft: 4,
+    },
+
+    // The design calls for additional offset around the opener.
+    opener: {
```
jandrade: Not sure if you are using this style inside the component, could you please confirm?

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
```diff
@@ -0,0 +1,244 @@
...
+                openerElement={this.openerElement}
+                role="menu"
+                style={style}
+            />
+        );
+    }
+}
+
+const styles = StyleSheet.create({
+    caret: {
```
jandrade: Same as the `opener` style.

packages/wonder-blocks-dropdown/components/dropdown-launcher.md
```diff
@@ -0,0 +1,32 @@
+A DropdownLauncher allows various elements to be used as dropdown menus. In this example we use the IconButton to open a Dropown of items to select from.
+
+```js
+const React = require("react");
+const {View} = require("@khanacademy/wonder-blocks-core");
+const IconButton = require("@khanacademy/wonder-blocks-icon-button").default;
+const {icons} = require("@khanacademy/wonder-blocks-icon");
+const {StyleSheet} = require("aphrodite");
+
+const dropdownItems = [<ActionItem label="Profile" href="http://khanacademy.org/profile" testId="profile" />,
```
jandrade: Nit: It seems this code is incorrectly indented :)

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
jandrade: @kevin Agree on this, we need to provide an API that feels more natural for devs that use these components. I'm also thinking a lot about this for the `Popover`. As these components (dropdown and popover) have some similarities, maybe this is a good point to agree on this API for these components for consistency purposes (including Modal as well). What do you think about this?

### Dropdown
- `Dropdown` becomes `DropdownCore`.
- `DropdownLauncher` becomes `Dropdown`.

### Modal
- `ModalLauncher` becomes `Modal`.

### Popover
- `Popover` could be the launcher as well.
- `PopoverCore`: Base popover that is used by all the variations. Similar to what `Dropdown` is doing as of today.

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
kumailn: at the moment I'm not, but its there incase a parent component needs to override the open state.

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
kumailn: It works in some cases, marginLeft isnt working, but menuTopSpace is working

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
kumailn: its not being used, removing it makes no difference

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
kevinbarabash: The modal launcher change kind of makes sense since we're calling things dialogs now in v2.  Thanks for think about how pattern might work across similar components.

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
kevinbarabash: I don't think we need this anymore after the addition of `DropdownAnchor`.  People can set `style` on the anchor itself.

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
kevinbarabash: Even though this is a prototype removing extraneous details will help people understand what the prototype more easily.

packages/wonder-blocks-dropdown/components/dropdown-anchor.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    render() {
+        const ClickableBehavior = getClickableBehavior();
+        return (
+            <ClickableBehavior
+                disabled={false}
+                onClick={this.props.handleOpenDropdown}
+            >
+                {(state, handlers) =>
```
kevinbarabash: We'll want to move `ClickableBehavior` into `dropdown-launcher.js` so that we can pass `state` to `DropdownLauncher`'s `children` callback prop.

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
```diff
@@ -0,0 +1,229 @@
...
+import DropdownAnchor from "./dropdown-anchor.js";
+
+type Props = {|
+    ...AriaProps,
+
+    /*
+     * The callback function that's passed the openDropdown function, used to rerender
+     the opener
+     */
+    children: (openDropdown: () => void) => React.Element<any>,
```
kevinbarabash: This should look something like:
```
children: (openDropdown: () => void, state: ClickableState) => React.Element<any>,
```
where `ClickableState` is defined as:
```
type ClickableState = {
    pressed: boolean,
    hovered: boolean,
    focused: boolean,
};
```
We don't have a `ClickableState` type defined anywhere but we probably should.

We may also want to pass the handlers too.  If we don't then we'll need to prevent `children` from returning a `React.Fragment`.  It's possible that `React.Element<any>` already does this.

packages/wonder-blocks-dropdown/components/dropdown-launcher.md
```diff
@@ -0,0 +1,122 @@
...
+];
+
+<View>
+    <DropdownLauncher
+        disabled={false}
+        menuText="Betsy Appleseed"
+        testId="teacher-menu"
+        menuItems={dropdownItems}
+    >
+        {(openDropdown) => (
```
kevinbarabash: Once `DropdownLauncher` has been updated to include `state` in the callback params, this should be updated to make use of that `state` to modify the style of the title to show that it works, e.g.
```
{(openDropdown, {focused, hovered, pressed}) => (
     <Title
          aria-role="menu"
          style={[
               focused && styles.focused,
               hovered && styles.hovered,
               pressed && styles.pressed,
          ]}
     >
          Hello, world!
     </Title>
)}
```

packages/wonder-blocks-icon-button/components/icon-button-core.js
```diff
@@ -207,7 +207,7 @@ const _generateStyles = (color, kind, light) => {
             color: offBlack64,
         };
     } else {
-        throw new Error("IconButton kind not recognized");
+        throw new Error(" kind not recognized");
```
kevinbarabash: Can you change this message to include the value of `kind`?

packages/wonder-blocks-dropdown/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@khanacademy/wonder-blocks-dropdown",
   "version": "1.3.0",
   "design": "v1",
-  "description": "Dropdown variants for Wonder Blocks.",
+  "description": "DropdownCore variants for Wonder Blocks.",
```
kevinbarabash: This should be reverted or changed to something that makes more sense.

packages/wonder-blocks-dropdown/components/dropdown.md
```diff
@@ -0,0 +1,140 @@
...
+        testId="teacher-menu"
+        menuItems={dropdownItems}
+    >
+        {(openDropdown, state) => (
+            <div
+                // style={[
+                //     state.focused && styles.focused,
+                //     state.hovered && styles.hovered,
+                //     state.pressed && styles.pressed,
+                // ]}
```
kevinbarabash: These should be re-enabled.

packages/wonder-blocks-dropdown/components/dropdown.md
```diff
@@ -0,0 +1,140 @@
...
+        background:"#42f4a1"
+    }
+});
+
+
+const dropdownItems = [
+    <ActionItem
+        label="Profile"
+        href="http://khanacademy.org/profile"
+        testId="profile"
```
kevinbarabash: This `testId` doesn't seem very important to this example and could be dropped from each of the items.

packages/wonder-blocks-dropdown/components/dropdown.md
```diff
@@ -0,0 +1,140 @@
...
+                //     state.pressed && styles.pressed,
+                // ]}
+            >
+            Hello, World!
+            </div>
+        )}
+    </Dropdown>
+</View>;
+```
+
```
kevinbarabash: Can you add an example showing `Dropdown` being used as a controlled component?

packages/wonder-blocks-dropdown/components/dropdown.md
```diff
@@ -0,0 +1,140 @@
...
+</Dropdown>
+```
+
+In this example we use a Title component with a Dropdown. You can define custom styles that will take place on different mouse events that ClickableBehavior supports.
+
+```js
+const React = require("react");
+const {View} = require("@khanacademy/wonder-blocks-core");
+const IconButton = require("@khanacademy/wonder-blocks-icon-button").default;
+const {icons} = require("@khanacademy/wonder-blocks-icon");
```
kevinbarabash: IconButton and icons aren't used in this example.

packages/wonder-blocks-dropdown/components/dropdown.md
```diff
@@ -0,0 +1,140 @@
...
+
+const styles = StyleSheet.create({
+    focused: {
+        border: "solid 2px blue"
+    },
+    hovered: {
+        background:"#89c8f9"
+    },
+    pressed: {
+        background:"#42f4a1"
```
kevinbarabash: Please use colors from `wonder-blocks-colors`.

packages/wonder-blocks-dropdown/components/dropdown.md
```diff
@@ -0,0 +1,140 @@
...
+const {View} = require("@khanacademy/wonder-blocks-core");
+const IconButton = require("@khanacademy/wonder-blocks-icon-button").default;
+const {icons} = require("@khanacademy/wonder-blocks-icon");
+const {StyleSheet} = require("aphrodite");
+
+const dropdownItems = [
+    <ActionItem
+        label="Profile"
+        href="http://khanacademy.org/profile"
+        testId="profile"
```
kevinbarabash: The `testId`s probably aren't necessary in the examples.

packages/wonder-blocks-dropdown/components/dropdown-anchor.js
```diff
@@ -0,0 +1,32 @@
...
+import * as React from "react";
+import * as ReactDOM from "react-dom";
+
+type Props = {|
+    /**
+     * The content for anchoring the tooltip.
+     * This element will be used to position the tooltip.
+     * If a string is passed as children we wrap it in a Text element.
+     * We allow children to be a string so that we can add tooltips to
+     * words within a large block of text easily.
```
kevinbarabash: copy/pasta

packages/wonder-blocks-dropdown/components/dropdown-anchor.js
```diff
@@ -0,0 +1,32 @@
...
+     */
+    children: React.Element<any>,
+
+    /**
+     * Callback to be invoked when the anchored content is mounted.
+     * This provides a reference to the anchored content, which can then be
+     * used for calculating tooltip bubble positioning.
+     */
```
kevinbarabash: copy/pasta

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -1,488 +1,158 @@
 //@flow
```
kevinbarabash: I'll review these tests after you rebase this PR over a new PR that does just the renaming.

kevinbarabash:
I updated the PR to use `refactor-dropdown` has the parent, but that still hasn't cleaned up the PR.  I think you'll have to fix the conflicts in dropdown.js and dropdown.test.js before proceeding.  It might be easier to squash your commits in this branch before resolving those conflicts.

kevinbarabash:
I changed the base back to `master` since the other PR's been merged.

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -0,0 +1,234 @@
...
+    pressed: boolean,
+    hovered: boolean,
+    focused: boolean,
+|};
+
+type State = {|
+    /**
+     * Whether or not the dropdown is currently open.
+     */
+    open: boolean,
```
kevinbarabash: I think we should use `opened` here as well.  I know we use `open` in other parts of the codebase but it's kind of ambiguous as to whether it's a verb or an adjective which is confusing.  Eventually we'd like to transition everything to using `opened`.

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -0,0 +1,234 @@
...
+
+    /**
+     * Whether or not last open state change was triggered by a keyboard click.
+     */
+    keyboard?: boolean,
+|};
+
+export default class Dropdown extends React.Component<Props, State> {
```
kevinbarabash: Since a lot of this logic has been lifted from `action-menu.js`, I think a good follow up task would be to refactor `action-menu.js` to use `Dropdown` instead of `DropdownCore`.  Can you create a Jira ticket for this and add a TODO with the ticket number in `action-menu.js`?

packages/wonder-blocks-icon-button/components/icon-button-core.js
kevinbarabash: I think this change can be reverted completely.

packages/wonder-blocks-dropdown/components/dropdown-anchor.js
```diff
@@ -0,0 +1,26 @@
...
+type Props = {|
+    /**
+     * The child element which will act as the opener for the dropdown
+     */
+    children: React.Element<any>,
+
+    /**
+     * Callback used to pass the childs Ref back up to the parent element
+     */
+    anchorRef: (?Element | ?Text | ?HTMLElement) => mixed,
```
kevinbarabash: The arg type of this callback should match `openerElement: ?HTMLElement;` in dropdown.js.  I think `HTMLElement` should be sufficient.  Is there a reason you added `Element` and `Text` here?

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -0,0 +1,234 @@
...
+        const ClickableBehavior = getClickableBehavior();
+        const isOpen = this.state.open;
+        const opener = (
+            <ClickableBehavior onClick={this.handleClick}>
+                {(state, handlers) => (
+                    <DropdownAnchor
+                        anchorRef={(ref) =>
+                            //TODO: Fix this flow error
+                            //$FlowFixMe
+                            (this.openerElement = ReactDOM.findDOMNode(ref))
```
kevinbarabash: Nit: leave a space after the `//`

I think it's possible to resolve the flow error here.  See my comment in dropdown-anchor.js.

packages/wonder-blocks-dropdown/components/dropdown.md
```diff
@@ -0,0 +1,233 @@
...
+
+// Custom styles for mouse events can be defined like so
+const styles = StyleSheet.create({
+    focused: {
+        border: "none",
+    },
+    hovered: {
+        textDecoration: "underline",
+    },
+    pressed: {},
```
kevinbarabash: Can you modify the `focused` and `pressed` states to be more visually different from the default and `hovered` states?  Maybe the `pressed` state changes the color of the the text and maybe focused gives it a background color.

packages/wonder-blocks-dropdown/components/dropdown.md
kevinbarabash: Friendly ping.

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,158 @@
...
+import {icons} from "@khanacademy/wonder-blocks-icon";
+import {mount, unmountAll} from "../../../utils/testing/mount.js";
+import ActionItem from "./action-item.js";
+import OptionItem from "./option-item.js";
+import SeparatorItem from "./separator-item.js";
+import DropdownCore from "./dropdown-core.js";
+import Dropdown from "./dropdown.js";
+import {keyCodes} from "../util/constants.js";
+
+describe("Dropdown Launcher", () => {
```
kevinbarabash: This should be `"Dropdown"`.

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,158 @@
...
+        anchor.simulate("click");
+
+        // Act
+        anchor.simulate("keydown", {keyCode: keyCodes.down});
+        anchor.simulate("keyup", {keyCode: keyCodes.down});
+
+        // Assert
+        expect(dropdown.state("open")).toBe(true);
+    });
+});
```
kevinbarabash: The tests look good.  There are a few hole in the the coverage though: https://codecov.io/gh/Khan/wonder-blocks/compare/b395cbd861e5211cd2f8133773ffe31f720991db...9311baab3540b1c865786c93848bf74ff0f6ba3a/diff.

packages/wonder-blocks-dropdown/components/dropdown-anchor.js
kumailn: Yes its because I get a flow error when I pass the anchorNode to the anchorRef prop, flow believes its a Text type.

kumailn: https://github.com/facebook/flow/issues/3548

kevinbarabash: I see.  I think it'd probably be better to make this `openerElement: (?HTMLElement) => mixed` then change
```
const anchorNode = ReactDOM.findDOMNode(this);
```
to be
```
const anchorNode = ((ReactDOM.findDOMNode(this): any): ?HTMLElement);
```
Also, it'd be good to link to that flow ticket directly in the code so people know why we're casting.  Once you cast here the rest of the downstream type-checking should work fine.

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -0,0 +1,223 @@
...
+     * Optional styling to add to the dropdown wrapper.
+     */
+    dropdownStyle?: StyleType,
+|};
+
+type ClickableState = {|
+    pressed: boolean,
+    hovered: boolean,
+    focused: boolean,
+|};
```
kevinbarabash: It would be nice to create a shared type for this in a follow up PR.  Can you create a ticket in Jira for this and then add a `// TODO(WB-???): create a shared type for this` comment?

packages/wonder-blocks-dropdown/components/dropdown.js
kevinbarabash: Friendly ping.

packages/wonder-blocks-dropdown/components/dropdown.md
kevinbarabash: I see `MixedDropdownExample` is using `Dropdown` as a controlled component.

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,241 @@
...
+
+        // Act
+        anchor.simulate("keydown", {keyCode: keyCodes.down});
+        anchor.simulate("keyup", {keyCode: keyCodes.down});
+
+        // Assert
+        expect(dropdown.state("opened")).toBe(true);
+    });
+
+    it.only("updates selectedValues when OptionItem clicked", () => {
```
kevinbarabash: Forgot to remove the `.only`.

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,241 @@
...
+        // Arrange
+        const controlledComponent = mount(<ControlledComponent />);
```
kevinbarabash: All of these `ControlledComponents` could live inside a `describe("used a controlled component", () => { ... }` block.

packages/wonder-blocks-dropdown/components/dropdown.test.js
kevinbarabash: Friendly ping.

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,241 @@
...
+        expect(dropdown.state("opened")).toBe(true);
+    });
+
+    it("can handle keyboard navigation", () => {
+        // Arrange
+        const anchor = dropdown.find(IconButton);
+        anchor.simulate("click");
+        const dropdownCore = dropdown.find(DropdownCore);
+
+        // Act & Assert
```
kevinbarabash: This should really be split into separate test cases, e.g.
- `it("should navigate to the first item when pressing down", () => {...})`
- `it("should navigate from one item to the next when pressing down", () => {...})`

By splitting these up, you'll be able to have separate `Act` and `Assert` blocks.

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,241 @@
...
+    it("removes selected values from selectedValues", () => {
+        // Arrange
+        const controlledComponent = mount(<ControlledComponent />);
+        controlledComponent.simulate("click");
+        dropdown = controlledComponent.find(DropdownCore);
+
+        // Act
+        dropdown.simulate("keydown", {keyCode: keyCodes.down});
+        dropdown.simulate("keyup", {keyCode: keyCodes.down});
+        expect(dropdown.instance().focusedIndex).toBe(0);
```
kevinbarabash: Keyboard navigation has already been tested in another test case so this assertion isn't necessary and can be removed.  Also, `Act` sections shouldn't have any assertions.  It's a sign that your test is trying to do too much.  Also, these "actions" should probably be part of the `Arrange` block b/c they're not the action we're trying to test as indicated by the title of the test.  I know some of our existing tests don't get this right all of the time

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -0,0 +1,241 @@
...
+        // Arrange
+        const controlledComponent = mount(<ControlledComponent />);
+        controlledComponent.simulate("click");
+        controlledComponent.setState({selectedValues: null});
+        dropdown = controlledComponent.find(DropdownCore);
+
+        // Act
+        dropdown.simulate("keydown", {keyCode: keyCodes.down});
+        dropdown.simulate("keyup", {keyCode: keyCodes.down});
+        expect(dropdown.instance().focusedIndex).toBe(0);
```
kevinbarabash: Same: remove the `expect` and move the `simulate` calls into the `Arrange` block.

packages/wonder-blocks-dropdown/components/dropdown.md
```diff
@@ -0,0 +1,232 @@
...
+            />,
+            <SeparatorItem />,
+            <OptionItem
+                label="Alex"
+                value="alex"
+            />,
+            <OptionItem
+                label="Cathy"
+                value="cathy"
+            />,
```
kevinbarabash: Could you format this so each item is on a single line?

packages/wonder-blocks-dropdown/components/dropdown.md
```diff
@@ -0,0 +1,232 @@
...
+const dropdownItems = [
+    <ActionItem
+        label="Change password"
+    />,
+    <ActionItem
+        label="Manage email"
+    />,
+    <ActionItem
+        label="Set up 2FA"
+    />,
```
kevinbarabash: Can you reformat each item to be on a single line?

packages/wonder-blocks-dropdown/components/dropdown.js
kevinbarabash: Can you cast this one as well just like the one in dropdown-anchor.js.

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -0,0 +1,224 @@
...
+            dropdownStyle,
+            /* eslint-disable no-unused-vars */
+            onChange,
+            selectedValues,
+            "aria-disabled": ariaDisabled,
+            ...sharedProps
+        } = this.props;
+
+        const ClickableBehavior = getClickableBehavior();
+        const isOpen = this.state.opened;
```
jandrade: It seems you don't need this variable here.

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -0,0 +1,224 @@
...
+import DropdownAnchor from "./dropdown-anchor.js";
+
+type Props = {|
+    ...AriaProps,
+
+    /**
+     * The child function, that takes in state the state of the element and
+     * returns an Element which will become the opener element for the dropdown.
+     */
+    children: (state: ClickableState) => React.Element<any>,
```
jandrade: Not sure about naming this argument as `state`. When I read the example, I got a bit confused b/c I thought you were sharing React's `state`, but you're sharing the "DOM" state of the element instead, is that the case?.

kevinbarabash: That's on me.  I think I called it `state` when I originally wrote `ClickableBehavior`.  Agree that the name could be be improved.

PR #429, opened 2019-06-14
WB-588: DropdownLauncher with Title component use case prototype

kumailn:
This is a prototype of a dropdown attached to a Title using the `DropdownLauncher` element. The `dropdown-launcher.md` file contains an example. The `DropdownLauncher` element detects the type of the child (anchor) element and then acts accordingly.

### DIFF
--- a/packages/wonder-blocks-dropdown/components/dropdown-launcher.js
+++ b/packages/wonder-blocks-dropdown/components/dropdown-launcher.js
@@ -4,6 +4,10 @@ import * as React from "react";
 import ReactDOM from "react-dom";
 import {StyleSheet} from "aphrodite";
 import type {AriaProps, StyleType} from "@khanacademy/wonder-blocks-core";
+import {Title} from "@khanacademy/wonder-blocks-typography";
+import {View} from "@khanacademy/wonder-blocks-core";
+import Icon, {icons} from "@khanacademy/wonder-blocks-icon";
+import IconButton from "@khanacademy/wonder-blocks-icon-button";
 import Dropdown from "./dropdown.js";
 import type {Item, DropdownItem} from "../util/types.js";
 import ActionItem from "./action-item.js";
@@ -198,7 +202,8 @@ export default class DropdownLauncher extends React.Component<Props, State> {
             ...sharedProps
         } = this.props;
 
-        const opener = (
+        const childAnchor = children(this._openDropdown);
+        let opener = (
             <span
                 ref={this.handleOpenerRef}
                 disabled={!dropdownItems.length || disabled}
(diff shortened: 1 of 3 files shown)


jeresig:

Built with commit e091ca4c71cc3de252719b3600a44a4574d562d0

https://deploy-preview-429--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
```diff
@@ -207,6 +212,45 @@ export default class DropdownLauncher extends React.Component<Props, State> {
...
+                <span
+                    ref={this.handleOpenerRef}
+                    disabled={!dropdownItems.length || disabled}
+                >
+                    {children(this._openDropdown)}
+                </span>
+            );
+        } else {
+            opener = (
+                <View
```
kevinbarabash: This should be moved into the example.  Also, we want to be able to provide information about the state of the opener to the `children` callback, i.e. the pressed, focused, and hovered state that `ClickableBehavior`.  This prototype need to integrate `ClickableBehavior` into `DropdownLauncher` some how.

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
```diff
@@ -207,6 +212,45 @@ export default class DropdownLauncher extends React.Component<Props, State> {
             </span>
         );
 
+        /*         TODO: Enforce better type check here
+        Currently this only looks for the 'icon' prop to decide
+        if an element is a IconButton or not as the 'isClassOf' method isnt present
+        */
+        if (childAnchor.props.icon) {
```
kevinbarabash: We shouldn't have to check for specific child types.

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
```diff
@@ -207,6 +212,45 @@ export default class DropdownLauncher extends React.Component<Props, State> {
...
+                    disabled={!dropdownItems.length || disabled}
+                    style={{display: "inline-block"}}
+                >
+                    <span style={{float: "left", cursor: "pointer"}}>
+                        {children(this._openDropdown)}
+                    </span>
+                    <Icon
+                        icon={icons.caretDown}
+                        aria-label="Close modal"
+                        onClick={this._openDropdown}
```
jandrade: Flow threw one error here, could you check it please?

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
```diff
@@ -207,6 +212,45 @@ export default class DropdownLauncher extends React.Component<Props, State> {
             </span>
         );
 
+        /*         TODO: Enforce better type check here
+        Currently this only looks for the 'icon' prop to decide
+        if an element is a IconButton or not as the 'isClassOf' method isnt present
+        */
+        if (childAnchor.props.icon) {
+            opener = (
```
jandrade: I'm not completely sure, but it seems this opener and the opener on line 206 use the same elements. If that's the case, this line can be removed.

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
kumailn: its because Icon doesn't implement onClick

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
kumailn: they do, this was in order to fix a flow error I was getting at the time

packages/wonder-blocks-dropdown/components/dropdown-launcher.js
kevinbarabash: We need to rework this anyways to use `ClickableBehavior`.

PR #430, opened 2019-06-14
WB-598: fix TooltipBubble still disappearing in some cases

kumailn:
I noticed the TooltipBubble was still disappearing in some cases because you could mouse over it and cancel the timer. I believe this was because the timer wasn't getting set in some cases, so I updated `instant` to be false in the `_setActiveState` method, I also updated the disappearance delay to allow for a little more time to mouse over.


jeresig:

Built with commit f55ecb47316f73f8ca686f25a524441fb20fabc6

https://deploy-preview-430--wonder-blocks.netlify.com

jandrade:
I think this is a stacked PR(?), could you please check this? It would b easier to review the changes that apply to this PR :). Thanks!

PR #433, opened 2019-06-21
1/4 [feature:popover] WB-603: Implement PopoverContentCore

jandrade:
## Summary

Renders popover content and allows to create custom content.

### NOTE: Stacked PRs

This is part of [WB-579: Implement Popovers](https://khanacademy.atlassian.net/browse/WB-579)

These PRs are created to cover the whole Popover implementation:

**1. Implement PopoverContentCore (current PR): Renders popover content.**
2. Implement PopoverContent: Add text-only, icon and illustration variations.
3. Implement Popover: Adds Popover component to handle popover triggering and positioning. Also, integrates TooltipTail and TooltipPopper.
4. Popover: Add Focus management

### Popover
[Implementation Spec](https://khanacademy.atlassian.net/wiki/spaces/FRONTEND/pages/115310606/Popover+v1+Implementation+Spec#PopoverContentCore-public-)
- Created `wonder-blocks-popover` package
- Added `PopoverContentCore` component to be used as the base window to display content.

### Tooltip
- Added `Placement` type as part of the public API.

## Testplan
1. Open https://deploy-preview-433--wonder-blocks.netlify.com/#!/PopoverContentCore/0
2. Check the component is displaying properly.
popover-content-core


jeresig:

Built with commit a66c223632f3abe9d4fab29f025db0772817fb2e

https://deploy-preview-433--wonder-blocks.netlify.com

packages/wonder-blocks-popover/components/popover-content-core.js
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * This is the base popover container. It’s used internally by all the variants.
+ * Also, It can be used to create flexible popovers.
+ */
+export default class PopoverContentCore extends React.Component<Props> {
+    static defaultProps = {
+        color: "light",
+        placement: "bottom",
+        closeButtonVisible: true,
+        "aria-label": "Close Popover",
```
kevinbarabash: This is the label for the close button not the whole popover content.  I wonder if we should have a separate prop that's "close-button" specific.

packages/wonder-blocks-popover/components/popover-content-core.js
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * When true, the close button is shown; otherwise, the close button is not shown.
+     */
+    closeButtonVisible?: boolean,
+
+    /**
+     * Whether we should use the default light color scheme or switch to a
+     * different color scheme.
+     */
+    color: "blue" | "dark" | "light",
```
kevinbarabash: Is `dark` the only custom color we want to allow?

packages/wonder-blocks-popover/components/popover-content-core.js
```diff
@@ -0,0 +1,125 @@
...
+     */
+    placement: Placement,
+
+    /**
+     * Called when the popover closes
+     */
+    onClose?: () => mixed,
+
+    style?: StyleType,
+|};
```
kevinbarabash: I think we'll need a `testId` prop on this and `PopoverContent`.

packages/wonder-blocks-popover/components/popover-content-core.js
jandrade: Agreed, `aria-label` in this case is too open to interpretation, so it would be better to make this prop name more related to the close element.

packages/wonder-blocks-popover/components/popover-content-core.js
jandrade: Yes for now. I checked in Webapp and so far we are only using white or darkBlue backgrounds for popovers. Maybe we can discuss this with designers in the next sync meeting.

packages/wonder-blocks-popover/components/close-button.js
```diff
@@ -0,0 +1,54 @@
...
+     */
+    light?: boolean,
+
+    /**
+     * Called when the popover closes
+     */
+    onClose?: () => mixed,
+|};
+
+export default class CloseButton extends React.Component<Props> {
```
somewhatabstract: Could you add a comment on the component class explaining it's purpose?

packages/wonder-blocks-popover/components/popover-content-core.js
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * Test ID used for e2e testing.
+     */
+    testId?: string,
+|};
+
+/**
+ * This is the base popover container. It’s used internally by all the variants.
+ * Also, It can be used to create flexible popovers.
```
somewhatabstract: NIT: "It" -> "it"

packages/wonder-blocks-popover/components/popover-content-core.js
```diff
@@ -0,0 +1,120 @@
...
+            Colors.offBlack8
+        }`,
+        margin: 0,
+        maxWidth: 280,
+        padding: Spacing.large,
+        overflow: "hidden",
+        justifyContent: "center",
+    },
+    /**
+     * Theming
```
somewhatabstract: Can't wait until we have proper theming

packages/wonder-blocks-popover/components/popover-content-core.md
```diff
@@ -0,0 +1,75 @@
+```js
```
somewhatabstract: Should there be additional docs here to describe the component?

packages/wonder-blocks-tooltip/index.js
```diff
@@ -1,5 +1,9 @@
 // @flow
+import type {Placement} from "./util/types.js";
+
 import Tooltip from "./components/tooltip.js";
 import TooltipContent from "./components/tooltip-content.js";
 
 export {Tooltip as default, TooltipContent};
+
+export type {Placement};
```
somewhatabstract: Feels weird to make this part of the exported Tooltip package API. It's a commitment to support it as part of the API even if popover eventually doesn't need it. That said, this type is part of the API anyway, given that it's used by the exported components.

However, I think if we start exporting anything else from there for use in popover, we should really reconsider how we share code as we're committing to maintaining that API surface long term.

styleguide.config.js
```diff
@@ -141,6 +141,18 @@ module.exports = {
                 },
             ],
         },
+        {
+            name: "Popover",
+            content: "packages/wonder-blocks-popover/docs.md",
+            sections: [
+                {
+                    name: "Building Blocks",
```
somewhatabstract: This section should have `private: true`, I believe.

jandrade:
I couldn't reproduce that behavior. Did you have to change something in the code snippet? Popover dismissing shouldn't work yet 😟, but my plan is to put it working as part of PR 3/4.

somewhatabstract:
Hmm, just noticed that the modal close button has the same issue on the netlify. Perhaps a deeper rooted bug?

somewhatabstract:
So, I've seen it twice this morning and now I can't repro. Not sure what's going on, but it's a curious bug.

somewhatabstract:
OK, just reproed again. Let me see if I can get a video.

styleguide.config.js
jandrade: These components (PopoverContentCore and PopoverContent) are going to be exported as part of the public API. For now I'm going to leave it this way. But this might change in one of the upcoming PRs. Thanks for bringing this up!

kevinbarabash:
fwiw, I wasn't able to repro the alert issue.

packages/wonder-blocks-popover/components/close-button.js
```diff
@@ -0,0 +1,59 @@
...
+}
+
+const styles = StyleSheet.create({
+    closeButton: {
+        position: "absolute",
+        right: Spacing.xSmall,
+        top: Spacing.xSmall,
+        // Allows the button to be above the title and/or custom content
+        zIndex: 1,
+    },
```
kevinbarabash: It feels weird to specify this style in the component itself.  I know we're not reusing this eleswhere, but I do think it's a good habit to not bake positioning styles into components themselves.  Can you add a `style` prop to `CloseButton` and move this style to the component using `CloseButton`?

packages/wonder-blocks-popover/components/popover-content-core.md
```diff
@@ -0,0 +1,79 @@
...
+            </View>
+        </View>
+    </PopoverContentCore>
+</View>
+```
+
+#### Accessibility notes
+The popover component will populate the `aria-describedby` attribute
+automatically, unless the user sets an `id` prop inside the Popover instance.
+Internally, it will be set on the trigger element.
```
kevinbarabash: I can't find `aria-describedby` used anywhere.  Were you planning to add this behavior of part of this PR or a future one?

jandrade: Nice catch, I'll add it as part of 3/4.

packages/wonder-blocks-popover/components/popover-content-core.js
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Called when the popover closes
+     */
+    onClose?: () => mixed,
+
+    /**
+     * Custom styles
+     */
+    style?: StyleType,
+
```
kumailn: maybe we should elaborate where exactly these styles will be applied

packages/wonder-blocks-popover/index.test.js
```diff
@@ -0,0 +1,15 @@
...
+    test("package exports default", async () => {
+        // Arrange
+        const importedModule = import("./index.js");
+
+        // Act
+        const result = await importedModule;
+
+        // Assert
+        expect(Object.keys(result).sort()).toEqual(
+            ["PopoverContentCore"].sort(),
```
kumailn: we can get rid of the second `.sort()`

packages/wonder-blocks-popover/components/popover-content-core.js
jandrade: Good call, I'll update this in PR 3/4.

PR #435, opened 2019-06-22
Add some more lint rules for react

kevinbarabash:
I was trying out some lint rules that we may want to adopt in webapp as well.  I really wanted `"react/default-props-match-prop-types": ["error", { "allowRequiredDefaults": true }]` to work, but there are a number of issues which make it unviable at the moment for our use cases.  See https://github.com/yannickcr/eslint-plugin-react/issues/1593 for details.


jeresig:

Built with commit 07baeed7b94ec877350081b0c0c17f6024194cc9

https://deploy-preview-435--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/action-menu-opener.js
```diff
@@ -16,7 +16,7 @@ export type SharedProps = {|
     /**
      * Whether the opener is disabled. If disabled, disallows interaction.
      */
-    disabled?: boolean,
+    disabled: boolean,
```
kevinbarabash: `"react/default-props-match-prop-types": ["error", { "allowRequiredDefaults": true }]` found this, but I ended up not including this rule b/c there was a false positive.

.eslintrc
```diff
@@ -51,6 +51,13 @@
         "promise/no-new-statics": "error",
         "promise/no-return-in-finally": "error",
         "react/jsx-handler-names": "error",
+        "react/no-children-prop": "error",
+        "react/no-typos": "error",
+        "react/no-string-refs": "error",
+        "react/no-this-in-sfc": "error",
+        "react/no-unescaped-entities": "error",
+        "react/react-in-jsx-scope": "error",
+        "react/require-render-return": "error",
```
kumailn: think we should also add `react/no-direct-mutation-state`. Also, `react/sort-prop-types` may be helpful

kevinbarabash: `react/no-direct-mutation-state` would be a solid addition.  In the past we tried `react/sort-prop-types` in webapp, but it turned out to be a lot of work for not much benefit.

somewhatabstract:
Looks ok to me.

PR #439, opened 2019-06-26
[core:feature] Implement ClickableState type

kumailn:
I've implemented a new exported type `ClickableState` which can be used in props.

### DIFF
--- a/packages/wonder-blocks-core/components/clickable-behavior.js
+++ b/packages/wonder-blocks-core/components/clickable-behavior.js
@@ -89,9 +89,7 @@ type Props = {|
     role?: ClickableRole,
 |};
 
-type State = {|
-    ...ClickableState,
-|};
+type State = ClickableState;
 
 export type ClickableState = {|
     /**
--- a/packages/wonder-blocks-core/components/clickable-behavior.js
+++ b/packages/wonder-blocks-core/components/clickable-behavior.js
@@ -90,6 +90,10 @@ type Props = {|
 |};
 
 type State = {|
+    ...ClickableState,
+|};
+
+export type ClickableState = {|
     /**
      * Whether the component is hovered.
      *
--- a/packages/wonder-blocks-core/components/clickable-behavior.js
+++ b/packages/wonder-blocks-core/components/clickable-behavior.js
@@ -48,7 +48,10 @@ type Props = {|
      * The React `Element` returned should take in this component's state
      * (`{hovered, focused, pressed}`) as props.
      */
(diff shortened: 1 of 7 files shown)


jeresig:

Built with commit 89cd4c82f060b6212a64999d03f67cc4a51e906b

https://deploy-preview-439--wonder-blocks.netlify.com

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -90,6 +90,10 @@ type Props = {|
 |};
 
 type State = {|
+    ...ClickableState,
```
jandrade: As you are not adding more keys here, I would suggest to remove this `State` type and only use `ClickableState` instead.

kumailn: good suggestion, I included it in the latest commit

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -89,7 +89,9 @@ type Props = {|
     role?: ClickableRole,
 |};
 
-type State = {|
+type State = ClickableState;
```
jandrade: oh, my point here is that you don't need this line at all. Instead of it, you can use `ClickableState` in the component definition. More exactly here: https://github.com/Khan/wonder-blocks/blob/a250b383272bbb94070f2dff224ebfcc4151a449/packages/wonder-blocks-core/components/clickable-behavior.js#L240

kumailn: I see, I fixed it in the latest commit

packages/wonder-blocks-icon-button/components/icon-button-core.js
```diff
@@ -10,37 +10,17 @@ import Color, {
...
-    focused: boolean,
-
-    /**
-     * Whether the IconButton is pressed.
-     *
-     * Refer to `ClickableBehavior` for more information on when this prop
-     * should be `true`.
-     */
-    pressed: boolean,
+    ...ClickableState,
```
kevinbarabash: Nice!

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -60,11 +64,6 @@ type Props = {|
 |};
 
 // TODO: WB-609 create a shared type for this
```
kevinbarabash: The TODO should be removed.

kumailn: right! sorry, missed this

PR #440, opened 2019-06-27
3/4 [feature:popover] WB-605: Implement Popover

jandrade:
## Summary

Renders a fully integrated Popover component.

### TODOs
This will go in the last PR (4/4)
- Add focus management
- Add keyboard support
- Add ARIA attributes

### NOTE: Stacked PRs

This is part of [WB-579: Implement Popovers](https://khanacademy.atlassian.net/browse/WB-579)

These PRs are created to cover the whole Popover implementation:

1. Implement PopoverContentCore (current PR): Renders popover content.
2. Implement PopoverContent: Add text-only, icon and illustration variations.
3. **Implement Popover: Adds Popover component to handle popover triggering and positioning. Also, integrates TooltipTail and TooltipPopper.**
4. Popover: Add Focus management and accessibility (keyboard support, ARIA attributes)

### Popover
[Implementation Spec](https://khanacademy.atlassian.net/wiki/spaces/FRONTEND/pages/115310606/Popover+v1+Implementation+Spec#Popover-public-)

#### Added
- `Popover`: It's used as the popover launcher.
    - Included TooltipPopper and TooltipTail from `wonder-blocks-tooltip`.
- `PopoverAnchor`: Works as trigger element and/or placement ref.
- `PopoverDialog`: Wraps the popover content and `TooltipTail`.
- `PopoverContext`: Handles `onClose` and `placement` behavior across the component.

#### Changed
- Updated `PopoverContent` and `PopoverContentCore` to make them work with the `Popover`.
- Improved documentation.
- Changed `color` prop, to use `wonder-blocks-color` palette instead of just 3 different colors.

### Tooltip
- Renamed `TooltipBubbleProps` to `PopperElementProps` to make the name more generic.
- Added `PopperElementProps` type as part of the public API.
- Added `TooltipPopper` and `TooltipTail` to the public API.
- `TooltipTail`: Added `color` prop to change the tail's bg color.

## Testplan
1. Open https://deploy-preview-440--wonder-blocks.netlify.com/#popover
2. Make sure all the examples are displaying properly.
popover-example


jeresig:

Built with commit 8cb0fe454a48ba403f62d97968ca71af09137b94

https://deploy-preview-440--wonder-blocks.netlify.com

packages/wonder-blocks-popover/components/close-button.js
```diff
@@ -20,6 +20,11 @@ type Props = {|
      * Called when the popover closes
      */
     onClose?: () => mixed,
+
+    /**
+     * Custom styles
+     */
+    style?: StyleType,
```
jandrade: @kevinbarabash As your suggested in PR 1/4, I moved the styles up to the parent container :)

packages/wonder-blocks-popover/components/popover-anchor.js
```diff
@@ -0,0 +1,63 @@
...
+            return children({
+                open: onClick,
+            });
+        } else {
+            // add onClick handler to automatically open the dialog after
+            // clicking on this anchor element
+            return React.cloneElement(children, {
+                ...children.props,
+                onClick: children.props.onClick
+                    ? () => {
```
jandrade: Added this condition here in case the current Popover instance is adding another `onClick` handler to the anchor element. This happens b/c for this particular condition, I'm automatically adding an `onClick` handler to the `PopoverAnchor` component in order to open the `PopoverDialog` by clicking on this button.

e.g. This works for this condition:
```
<Popover
    {...props}
>
    <Button onClick={customClickHandler}>click to open PopoverDialog</Button>
</Popover>
```
In this case, if you click on the button, Popover will open the Dialog and will execute `customClickHandler` as well.

packages/wonder-blocks-popover/components/popover-content-core.js
```diff
@@ -31,7 +36,7 @@ type Props = {|
      * Whether we should use the default light color scheme or switch to a
      * different color scheme.
      */
-    color: "blue" | "dark" | "light",
+    color: "blue" | "darkBlue" | "white",
```
jandrade: Changed colors here: from `dark` to `darkBlue` and from `light` to `white` to match with the tokens we have defined in `wonder-block-colors`. Let me know what you think and if this makes sense, we could start following this pattern in other packages.

packages/wonder-blocks-popover/components/popover-context.js
```diff
@@ -0,0 +1,16 @@
...
+import type {Placement} from "@khanacademy/wonder-blocks-tooltip";
+
+type ContextType = {|
+    close?: () => mixed,
+    placement?: Placement,
+|};
+
+const defaultContext: ContextType = {
+    close: undefined,
+    placement: "top",
```
jandrade: Added `placement` to the context to facilitate passing this value from `Popover` (via `TooltipPopper`) down to `PopoverContent`. This is needed here to reposition the illustration to the start or the end of the content, in case the popper changes its `placement`.

packages/wonder-blocks-popover/components/popover-dialog.js
```diff
@@ -0,0 +1,124 @@
+// @flow
```
jandrade: This component is a lite version of `TooltipBubble`. I had to create a new component b/c of the different constraints between both packages. 

Main difference is that `TooltipBubble` handles hover states and `PopoverDialog` doesn't need to handle any states at all (for now). Also, `PopoverDialog` needs to coordinate different background colors for the content and tail components.

packages/wonder-blocks-popover/components/popover.js
```diff
@@ -0,0 +1,222 @@
...
+    /**
+     * Where the popover should try to appear in relation to the trigger element.
+     */
+    placement: Placement,
+
+    /**
+     * When enabled, user can hide the popover content by pressing the `esc` key
+     * or clicking/tapping outside of it.
+     */
+    dismissEnabled?: boolean,
```
jandrade: This is not working yet. This will be part of PR 4/4. Same for the `id` and `testId` props.

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
```diff
@@ -11,7 +11,7 @@ import TooltipTail from "./tooltip-tail.js";
 
 import type {getRefFn, Offset, Placement} from "../util/types.js";
 
-export type TooltipBubbleProps = {|
```
jandrade: Changed the name here to make it more generic (now that it's also being used by Popover). What do you think @somewhatabstract?

packages/wonder-blocks-tooltip/components/tooltip-popper.js
```diff
@@ -7,23 +7,21 @@ import * as React from "react";
...
+import type {PopperElementProps} from "./tooltip-bubble.js";
 
 type Props = {|
     /**
      * This uses the children-as-a-function approach, mirroring react-popper's
-     * implementation, except we enforce the return type to be our
-     * `TooltipBubble` component.
+     * implementation.
      */
-    children: (TooltipBubbleProps) => React.Element<typeof TooltipBubble>,
```
jandrade: I removed the `typeof TooltipBubble` validation as we need to make it work for `PopoverDialog` as well. But I didn't want to import popover files here, so that's way I used `any` instead. Is there a better way to approach this?

packages/wonder-blocks-tooltip/components/tooltip-tail.js
```diff
@@ -9,6 +9,15 @@ import Spacing from "@khanacademy/wonder-blocks-spacing";
...
 
 export type Props = {|
+    /**
+     * Whether we should use the default white background color or switch to a
+     * different bg color.
+     *
+     * NOTE: Added to support custom popovers
+     * @ignore
+     */
+    color: $Keys<typeof Colors>,
```
jandrade: Used this here to only allow valid WB colors.

packages/wonder-blocks-popover/components/popover.test.js
```diff
@@ -0,0 +1,115 @@
...
+            </Popover>,
+        );
+
+        // Act
+        const anchorElement = wrapper.find("[data-anchor]").getDOMNode();
+
+        // Assert
+        expect(wrapper.state("anchorElement")).toBe(anchorElement);
+
+        expect(wrapper.find(PopoverContent)).toHaveLength(0);
```
kumailn: Maybe we can change these expect statements to be something like `expect(wrapper.find(PopoverContent).exists()).toBeFalsy();` (or `toBeTruthy()`). I think it’s a little clearer than checking the length to be 0 or 1 to see if it exists.

packages/wonder-blocks-popover/components/popover.test.js
```diff
@@ -0,0 +1,115 @@
...
+                    </button>
+                )}
+            </Popover>,
+        );
+
+        // Act
+        wrapper.find("[data-anchor]").simulate("click");
+
+        // Assert
+        expect(wrapper.find(PopoverContent)).toHaveLength(1);
```
kumailn: same as above

packages/wonder-blocks-tooltip/index.test.js
```diff
@@ -9,7 +9,12 @@ describe("@khanacademy/wonder-blocks-tooltip", () => {
 
         // Assert
         expect(Object.keys(result).sort()).toEqual(
-            ["default", "TooltipContent"].sort(),
+            [
+                "default",
+                "TooltipContent",
+                "TooltipPopper",
+                "TooltipTail",
+            ].sort(),
```
kumailn: I think we can replace this with `[
                "TooltipContent",
                "TooltipPopper",
                "TooltipTail",
                "default",
            ]` and get rid of the call to `sort()` here

packages/wonder-blocks-popover/components/popover.test.js
jandrade: Good call, we can also use a jest-enzyme matcher for that :). I'll update that.

https://www.npmjs.com/package/jest-enzyme#toexist

packages/wonder-blocks-tooltip/index.test.js
jandrade: This is a pattern we are using on all packages to avoid injecting manual ordering issues, so I'd recommend to leave it for consistency. What do you think?

kumailn: I see, If it’s to ensure consistency I think it’s fine to keep

packages/wonder-blocks-popover/components/close-button.js
```diff
@@ -34,26 +39,27 @@ export default class CloseButton extends React.Component<Props> {
...
-                kind={light ? "primary" : "tertiary"}
-                light={light}
-                style={styles.closeButton}
-            />
+            <PopoverContext.Consumer>
+                {({close}) => {
+                    if (close && onClose) {
+                        throw new Error(
+                            "You've specified 'onClose' on the content when using Popover. Please specify 'onClose' on the Popover instead",
+                        );
```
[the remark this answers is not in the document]
jandrade: Yes, there is a case when `onClose` is being passed down from `PopoverContent` and/or `PopoverContentCore`.
But now that you mention it, I think that it could be unnecessary to add the `onClose` prop to these components. At the end we are delegating the close behavior to the `Popover` component itself. 

Do you think is it better to remove this `onClose` prop from `PopoverContent` and/or `PopoverContentCore` and?

packages/wonder-blocks-popover/components/popover.md
```diff
@@ -0,0 +1,265 @@
+#### Example: Opening a popover using a trigger element
```
kevinbarabash: It may be worth mentioning at the top before the examples what combinations of props aren't allowed.

packages/wonder-blocks-popover/components/popover.test.js
```diff
@@ -0,0 +1,113 @@
...
+                    </button>
+                )}
+            </Popover>,
+        );
+
+        // Assert
+        expect(wrapper.find(PopoverContent)).not.toExist();
+    });
+
+    it("should render a text-only variant", () => {
```
kevinbarabash: The test title could be a bit more descriptive, maybe `it("should render the popover dialog after clicking the trigger", () => {...});` or something like that.

packages/wonder-blocks-popover/components/popover.test.js
```diff
@@ -0,0 +1,113 @@
...
+                        Open default popover
+                    </button>
+                )}
+            </Popover>,
+        );
+
+        // Act
+        // first open the popover
+        wrapper.find("[data-anchor]").simulate("click");
+        expect(wrapper.find(PopoverContentCore)).toExist();
```
kevinbarabash: These two lines should be part of `Arrange` since the thing we're testing is the closing.  Also, the `expect` isn't necessary since there's another test that checks that popover can be opened from a trigger element.

packages/wonder-blocks-popover/components/popover.test.js
```diff
@@ -0,0 +1,113 @@
...
+        // then we try to close it from inside the content
+        wrapper.find("[data-close-button]").simulate("click");
+
+        // Assert
+        expect(onCloseMock).toBeCalled();
```
kevinbarabash: Since this is an uncontrolled component the popover dialog should also be hidden now.  We should probably check for that as well.

packages/wonder-blocks-tooltip/components/tooltip-tail.js
kevinbarabash: Nice!

packages/wonder-blocks-popover/components/close-button.js
```diff
@@ -34,26 +39,27 @@ export default class CloseButton extends React.Component<Props> {
     };
```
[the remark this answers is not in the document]
jandrade: https://khanacademy.atlassian.net/browse/WB-612

packages/wonder-blocks-tooltip/components/tooltip-bubble.js
somewhatabstract: I'm still uncomfortable making some of these things public API elements and the increased support surface area that implies. But as far as naming goes, it seems reasonable to me.

packages/wonder-blocks-tooltip/components/tooltip-popper.js
somewhatabstract: Templatize the props and pass the element type in?

packages/wonder-blocks-popover/components/close-button.test.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+import * as React from "react";
+
+import expectRenderError from "../../../utils/testing/expect-render-error.js";
+import CloseButton from "./close-button.js";
+import PopoverContext from "./popover-context.js";
+
+describe("CloseButton", () => {
+    test("PopoverContext.Provider and onClose should warn", () => {
+        expectRenderError(
```
somewhatabstract: Hmmm, I don't like test cases that don't follow the AAA layout. It tends to obfuscate what's being tested from what is being asserted.

Could you put the JSX as a variable in the arrange or act, then have the expectation in the assert.

Using shared helper asserts like `expectRenderError` also makes debugging tests hard because the error is thrown in a place unrelated to the test case that fails, which can make the test costly to fix if it starts failing. Although shared code in non-test code is usually a good thing, in tests I believe the opposite is the case as coupling tests together can make it really hard to maintain those tests. Tests should be as isolated as possible.

What do you think about copy pasting the work done by expectRenderError and following AAA layout for consistency and ease of maintenance, rather than sharing code across test cases?

packages/wonder-blocks-popover/components/popover-anchor.test.js
```diff
@@ -0,0 +1,70 @@
...
+            </PopoverAnchor>,
+        );
+
+        // Act
+        wrapper.find("[data-anchor]").simulate("click");
+
+        // Assert
+        // both custom and internal click should be called
+        expect(onClickInnerMock).toBeCalled();
+        expect(onClickMock).toBeCalled();
```
somewhatabstract: I know it's repetitive, but I'd rather each expect had it's own test case. They are verifying different things and if the first expectation fails, we never test the second, which means that we lose fidelity in our test runs. If they are separate test cases then we get more information from our test runs.

Could you split these into separate test cases please?

packages/wonder-blocks-popover/components/popover-content-core.js
```diff
@@ -101,7 +109,7 @@ const styles = StyleSheet.create({
             Colors.offBlack8
         }`,
         margin: 0,
-        maxWidth: 280,
+        maxWidth: 288,
```
somewhatabstract: Should we be using our `Spacing` constants to build these values rather than magic numbers?

packages/wonder-blocks-popover/components/popover-content.js
```diff
@@ -53,20 +59,25 @@ type CommonProps = {|
...
+     * Decorate the popover with a full-bleed illustration. It cannot be used at
+     * the same time with icon.
+     */
+    image?: React.Element<"img"> | React.Element<"svg">,
+
+    /**
+     * Without these, flow complains about icon and emphasized not being
      * available on props at all b/c these are exact object types.
      */
     icon?: void,
```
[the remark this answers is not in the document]
jandrade: My only concern would be that the Icon must have a predefined size (40 x 40). So, not sure about the impact on that on `img` and `Icon`. If you think it's ok to add them, Do you think adding `string` as well would be unncecessary?

packages/wonder-blocks-popover/components/popover-content.js
```diff
@@ -122,29 +135,75 @@ export default class PopoverContent extends React.Component<Props> {
...
+
+                                <View style={styles.text}>
+                                    <HeadingSmall style={styles.title}>
+                                        {title}
+                                    </HeadingSmall>
+                                    <Body>{content}</Body>
+                                </View>
+                            </View>
+
+                            {actions && (
```
somewhatabstract: I often see this sort of inline optional rendering as a refactor smell and prefer the use of `maybe...` functions for rendering optional pieces like this. It tends to make for easier to read code. What do you think?

```
maybeRenderActions() {
    const {actions} = this.props;
    if (!actions) {
       return null;
    }
    return (<ComponentStuffToRender />);
}
```

then the main render just has:

```
...
{this.maybeRenderActions()}
...
```

somewhatabstract: This goes for the icon and image stuff too.

packages/wonder-blocks-popover/components/popover-content.js
```diff
@@ -122,29 +135,75 @@ export default class PopoverContent extends React.Component<Props> {
...
+
+                    // illustration popover can't be placed horizontally
+                    if (
+                        image &&
+                        (placement === "left" || placement === "right")
+                    ) {
+                        throw new Error(
+                            "'image' can only be vertically placed. You can fix this by either changing `placement` to `top` or `bottom` or removing the `image` prop inside `content`.",
+                        );
+                    }
```
somewhatabstract: I wonder if we should encapsulate these error checks in a `validateProps` call or something to make this inline function a little easier to follow.

packages/wonder-blocks-popover/components/popover-content.test.js
```diff
@@ -0,0 +1,75 @@
...
+            <PopoverContext.Provider value={{close: () => {}}}>
+                <PopoverContent
+                    title="title"
+                    content="content"
+                    onClose={() => {}}
+                />
+                ,
+            </PopoverContext.Provider>,
+            "You've specified 'onClose' on the content when using Popover. Please specify 'onClose' on the Popover instead",
+        );
```
somewhatabstract: I really don't like this.

packages/wonder-blocks-popover/components/popover-context.js
```diff
@@ -0,0 +1,16 @@
...
+};
+
+export default React.createContext<ContextType>(defaultContext);
```
somewhatabstract: Can you add some commentary here on what this is? Should help future maintenance and code spelunking.

packages/wonder-blocks-popover/components/popover-dialog.js
```diff
@@ -0,0 +1,124 @@
...
+ * [PopoverContentCore](#PopoverContentCore), and the
+ * [TooltipTail](#TooltipTail).
+ *
+ * Note that without explicit positioning, the tail will not be centered.
+ */
+export default class PopoverDialog extends React.Component<Props> {
+    componentDidUpdate(prevProps: Props) {
+        if (prevProps.placement !== this.props.placement) {
+            this.props.onUpdate(this.props.placement);
+        }
```
somewhatabstract: Can you add a comment on why we're doing this? I think I get it, but it will be useful for future code viewers to understand why we have this.

packages/wonder-blocks-popover/components/popover-dialog.js
```diff
@@ -0,0 +1,124 @@
...
+                    ref={updateBubbleRef}
+                    data-placement={placement}
+                    style={[
+                        outOfBoundaries && styles.hide,
+                        styles[`content-${placement}`],
+                        style,
+                    ]}
+                >
+                    {children}
+                    <TooltipTail
```
somewhatabstract: I think we should have a single package for Popover and Tooltip, and this internal component should just be called `Tail`.

packages/wonder-blocks-popover/components/popover-dialog.test.js
```diff
@@ -0,0 +1,67 @@
...
+import {mount, unmountAll} from "../../../utils/testing/mount.js";
+
+import PopoverDialog from "./popover-dialog.js";
+import PopoverContentCore from "./popover-content-core.js";
+
+jest.mock("@khanacademy/wonder-blocks-tooltip");
+
+describe("PopoverDialog", () => {
+    afterEach(() => {
+        unmountAll();
```
somewhatabstract: @kevinbarabash What do you think about adding your cool hack so that this just happens and our use of our custom mount/unmount is hidden away to reduce boilerplate?

No action for this diff.

packages/wonder-blocks-popover/components/popover.js
```diff
@@ -0,0 +1,220 @@
...
+    id?: string,
+
+    /**
+     * Renders the popover when true, renders nothing when false.
+     *
+     * Using this prop makes the component behave as a controlled component. The
+     * parent is responsible for managing the opening/closing of the popover
+     * when using this prop.
+     */
+    opened?: boolean,
```
somewhatabstract: I've found in a few places before that often, with optional `boolean` values, folks want to assign things that evaluate to `undefined` or other falsy values. I wonder if for ease of API use, we should type these as `?boolean` instead?

It's a philosphical thing, I guess. Not asking for change here, just sparking discussion.

packages/wonder-blocks-popover/components/popover.js
```diff
@@ -0,0 +1,220 @@
...
+
+    /**
+     * Test ID used for e2e testing.
+     */
+    testId?: string,
+|};
+
+type State = {|
+    opened: boolean,
+    anchorElement: ?HTMLElement,
```
somewhatabstract: Does Tooltip also specify `HTMLElement`? I don't remember how I made it anchor on text (I know it does).

packages/wonder-blocks-popover/components/popover.js
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    renderContent() {
+        const {content} = this.props;
+
+        if (typeof content === "function") {
+            return content({
+                close: this.handleClose,
+            });
+        } else {
```
somewhatabstract: Else is unnecessary.

packages/wonder-blocks-popover/components/popover.md
```diff
@@ -0,0 +1,265 @@
...
+const Spacing = require("@khanacademy/wonder-blocks-spacing").default;
+const {HeadingSmall, LabelLarge} = require("@khanacademy/wonder-blocks-typography");
+
+const customIcon = {
+    small: "M6.92820 0L13.85640 4L13.85640 12L6.92820 16L0 12L0 4Z",
+};
+
+const styles = StyleSheet.create({
+    example: {
+        paddingTop: 300
```
somewhatabstract: SUPER NIT: Would be nice to use `Spacing` constants in doc examples (and multiples thereof).

packages/wonder-blocks-popover/components/popover.md
```diff
@@ -0,0 +1,265 @@
...
+                                </View>
+                            </View>
+                        </PopoverContentCore>
+                    }
+                >
+                    <Button onClick={() => console.log('This is a controlled popover.')}>
+                        Anchor element (it doesn't open the popover)
+                    </Button>
+                </Popover>
+                <Strut size={32} />
```
somewhatabstract: Spacing contants, perhaps?

packages/wonder-blocks-tooltip/index.js
```diff
@@ -1,9 +1,12 @@
...
 import type {Placement} from "./util/types.js";
+import type {PopperElementProps} from "./components/tooltip-bubble.js";
 
 import Tooltip from "./components/tooltip.js";
 import TooltipContent from "./components/tooltip-content.js";
+import TooltipPopper from "./components/tooltip-popper.js";
+import TooltipTail from "./components/tooltip-tail.js";
 
-export {Tooltip as default, TooltipContent};
+export {Tooltip as default, TooltipContent, TooltipPopper, TooltipTail};
```
somewhatabstract: This is really starting to feel like these should be in the same package.

packages/wonder-blocks-tooltip/index.test.js
somewhatabstract: Yeah, this was here to help us maintain control over what we export without being overly picky about the ordering so that we dodge busywork if the order changed.

packages/wonder-blocks-popover/components/popover-content.js
```diff
@@ -122,29 +135,75 @@ export default class PopoverContent extends React.Component<Props> {
...
+                                    </HeadingSmall>
+                                    <Body>{content}</Body>
+                                </View>
+                            </View>
+
+                            {actions && (
+                                <View style={styles.actions}>
+                                    {typeof actions === "function"
+                                        ? actions({
+                                              close: (close || onClose: any),
```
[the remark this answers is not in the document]
jandrade: Good call, I had an issue with Flow, but now I'm doing a small refactor to fix this.

styleguide.config.js
```diff
@@ -145,6 +145,11 @@ module.exports = {
             name: "Popover",
             content: "packages/wonder-blocks-popover/docs.md",
             sections: [
+                {
+                    components: [
+                        "packages/wonder-blocks-popover/components/popover.js",
+                    ],
+                },
                 {
                     name: "Building Blocks",
```
somewhatabstract: This section should have `private: true`

packages/wonder-blocks-popover/components/popover-content.js
```diff
@@ -190,4 +249,10 @@ const styles = StyleSheet.create({
         marginTop: -Spacing.large,
         width: `calc(100% + ${Spacing.large * 2}px)`,
     },
+
+    imageToBottom: {
+        marginBottom: -Spacing.large,
+        marginTop: Spacing.large,
+        order: 1,
```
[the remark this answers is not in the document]
jandrade: exactly!

packages/wonder-blocks-popover/components/popover.js
```diff
@@ -0,0 +1,220 @@
...
+ */
+export default class Popover extends React.Component<Props, State> {
+    static defaultProps = {
+        placement: "top",
+    };
+
+    static getDerivedStateFromProps(props: Props, state: State) {
+        return {
+            opened:
+                typeof props.opened === "boolean" ? props.opened : state.opened,
```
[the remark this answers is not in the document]
jandrade: yes, I'll add the comment explaining it.

packages/wonder-blocks-popover/components/popover.js
```diff
@@ -0,0 +1,220 @@
...
+
+        return (
+            <PopoverContext.Provider
+                value={{
+                    close: this.handleClose,
+                    placement: placement,
+                }}
+            >
+                <PopoverAnchor
+                    anchorRef={(ref) => this.updateRef(ref)}
```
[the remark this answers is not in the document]
jandrade: Good call!

packages/wonder-blocks-popover/components/popover.md
```diff
@@ -0,0 +1,265 @@
...
+const Color = require("@khanacademy/wonder-blocks-color").default;
+const {View} = require("@khanacademy/wonder-blocks-core");
+const {default: Icon, icons} = require("@khanacademy/wonder-blocks-icon");
+const {Strut} = require("@khanacademy/wonder-blocks-layout");
+const Spacing = require("@khanacademy/wonder-blocks-spacing").default;
+const {HeadingSmall, LabelLarge} = require("@khanacademy/wonder-blocks-typography");
+
+const customIcon = {
+    small: "M6.92820 0L13.85640 4L13.85640 12L6.92820 16L0 12L0 4Z",
+};
```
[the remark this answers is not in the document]
jandrade: yes!

packages/wonder-blocks-popover/components/popover-anchor.test.js
jandrade: Agreed that it's ideally better to use one `expect` per test, but I think this is a special case where we want to test that the inner `onClick` is preserved and called at the same time with the PopoverAchor's `onClick` handler.

I separated this into two different tests, but maybe the second test feels kind of redundant (?), as we would be testing implementation details in this case? We would be testing that the onClick handler gets called if the button is clicked.

This would be result of creating the second test:
```
it("should also call the internal onClick handler if child is a Node", () => {
    // Arrange
    const onClickInnerMock = jest.fn();

    const wrapper = mount(
        <PopoverAnchor anchorRef={jest.fn()} onClick={jest.fn()}>
            <button onClick={onClickInnerMock} data-anchor>
                test
            </button>
        </PopoverAnchor>,
    );

    // Act
    wrapper.find("[data-anchor]").simulate("click");

    // Assert
    expect(onClickInnerMock).toBeCalled();
});
```

What do you think? Do you find a better way to approach this test?

packages/wonder-blocks-popover/components/popover-content-core.js
jandrade: Good idea! it's great to see the design spec is following the spacing constraints :)

packages/wonder-blocks-popover/components/popover-content.js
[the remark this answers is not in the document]
kevinbarabash: Good question.  We could set the width/height + `overflow: hidden` on a container.  I know we use inline SVG in a number of places on the frontend so not being able to reuse those will be a bit awkward.  In terms of keeping `string` it feels inconsistent to allow `string` here but not for `image`.  Thoughts?

packages/wonder-blocks-popover/components/close-button.test.js
jandrade: Fair enough.

What do you think of this approach?

```
// Arrange
const nodes = (
    <PopoverContext.Provider value={{close: () => {}}}>
        <PopoverContent
            title="title"
            content="content"
            onClose={() => {}}
        />
        ,
    </PopoverContext.Provider>
);

// Act
const underTest = () => mount(nodes);

// Assert
expect(underTest).toThrowError(
    "You've specified 'onClose' on the content when using Popover. Please specify 'onClose' on the Popover instead",
);
```

packages/wonder-blocks-popover/components/popover-dialog.js
jandrade: Just an idea.... What do you think about decoupling `Tail` and `TooltipPopper` (`Popper`) to a new private package? e.g. `wonder-blocks-floating`.

packages/wonder-blocks-popover/components/popover.js
jandrade: Yes, it does the same.

styleguide.config.js
jandrade: Not necessarily, as this section is part of the public API. Maybe we could include `Internal Components` (PopoverDialog, CloseButton, PopoverContext, etc) and make it private. The only thing I see is that I find it a bit unnecessary to expose this information inside React Styleguidist. Personally, I think that consumers wouldn't need to know the details of the private API.

Are we using these private sections only for development purposes (localhost), right? Asking this mostly because I'm not sure how needed is to include this information into Styleguidist. What are your thoughts?

packages/wonder-blocks-tooltip/index.js
jandrade: About this, I keep thinking about this. Even that Tooltip and Popover share some logic, they serve different purposes. Also, merging both into the same package could lead to mixed API.

e.g.
`import {Tooltip, TooltipContent, Popover, PopoverContent, PopoverContentCore} from "wb-tooltip"`

packages/wonder-blocks-popover/components/popover-content.js
jandrade: Sounds good to me :). Thanks for the suggestion!

packages/wonder-blocks-popover/components/popover-anchor.js
jandrade: Added a comment in line # 56

packages/wonder-blocks-popover/components/popover-content.test.js
jandrade: What do you think of this approach?

packages/wonder-blocks-popover/components/popover.js
jandrade: Didn't add this, b/c I already added this functionality in PR 4/4 but next time I'll make sure to do it. Thanks for the suggestion!

packages/wonder-blocks-popover/components/popover.js
jandrade: 🤔  That's a good point. I personally prefer to enforce strict types for API consistency, but you're right that folks might want to do that as well. The good thing is that now we are going to enforce using `// @flow` on every file, so this could help to enforce these types when creating the instances. Pretty valid point tho.

packages/wonder-blocks-popover/components/close-button.js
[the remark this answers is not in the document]
kevinbarabash: If we can remove it we should.  It avoid what could be a common mistake completely.

packages/wonder-blocks-popover/components/popover-dialog.js
kevinbarabash: I see, it's been added below.  Thank you.

packages/wonder-blocks-popover/components/popover-dialog.test.js
kevinbarabash: I think we should.

packages/wonder-blocks-popover/components/close-button.js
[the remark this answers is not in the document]
jandrade: I'm removing it, I noticed that it reduces the API surface area: one prop less to deal with for PopoverContent, PopoverContentCore and CloseButton. Now, `onClose` is completely centralized in `Popover`.

Please correct me if I'm wrong... I think that we originally used this technique in `Modals` b/c we needed to cover both scenarios: Opening a modal using the `ModalLauncher` and opening a modal by itself (e.g. `StandardModal` or `OnePaneDialog`).

The difference with Popover is that the `PopoverDialog` depends on `Popover` to be opened, where as in Modals, `OnePaneDialog` can be independent from `ModalLauncher` in some situations.

packages/wonder-blocks-popover/components/popover-dialog.test.js
jandrade: Interesting! How could we make that work?

packages/wonder-blocks-popover/components/popover.md
jandrade: Mentioned that on each variant example. Do you think is it easier to read at the top instead?

packages/wonder-blocks-popover/components/popover-content.js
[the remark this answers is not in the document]
jandrade: After our last discussion, added `img` and `svg` but left out `Icon` for now. As current Icon sizes don't support the design spec (64px), we need to address this with the Design team. So, I'm going to leave it that way now, and wait for their feedback.

styleguide.config.js
jandrade: As per our last conversation, I changed the structure here to keep it consistent with the rest of the packages. Thanks for the context!

packages/wonder-blocks-popover/components/close-button.js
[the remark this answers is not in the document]
kevinbarabash: That sounds right to me.  Modals and popovers are similar but there's some differences too.  Thanks for removing the prop.

packages/wonder-blocks-popover/components/popover.md
kevinbarabash: Not sure.  Let's give it a try and see if people are confused or not.

packages/wonder-blocks-popover/components/popover-dialog.test.js
kevinbarabash: In jest.config.js in webapp we do:
```
    moduleNameMapper: {
        "^enzyme$": "<rootDir>/dev/tools/jest-runner/enzyme-shim.js",
        // Load a .js file with no exports whenever a .css or .less file is requested.
        "\\.(css|less)$": "<rootDir>/dev/tools/jest-runner/style-mock.js",
    },
```
The shim there is very similar to shim here, see https://github.com/Khan/webapp/blob/master/dev/tools/jest-runner/enzyme-shim.js for details.

packages/wonder-blocks-popover/components/popover-content-core.js
somewhatabstract: No change, just a question.

What's the rationale behind using color names instead of semantic names here? Seems like using color names is going to really hamstring theming work.

packages/wonder-blocks-popover/components/popover-content-core.md
```diff
@@ -16,27 +16,27 @@ const customIcon = {
 
 const styles = StyleSheet.create({
     customPopover: {
-        maxWidth: 420,
-        width: 420,
-        textAlign: "center"
+        maxWidth: Spacing.medium * 25,
+        width: Spacing.medium * 25,
```
somewhatabstract: Nice. Musing: I wonder if we should have an API for these constants like `Spacing.medium(25)`. No change requested, obviously.

packages/wonder-blocks-popover/components/popover-content.js
```diff
@@ -119,16 +111,84 @@ export default class PopoverContent extends React.Component<Props> {
...
+            this.props.image &&
+            (placement === "left" || placement === "right")
+        ) {
+            throw new Error(
+                "'image' can only be vertically placed. You can fix this by either changing `placement` to `top` or `bottom` or removing the `image` prop inside `content`.",
+            );
+        }
+    }
+
+    maybeRenderImage = ({placement}: PopoverContextType) => {
```
somewhatabstract: These small refactorings are great! Thank you.

packages/wonder-blocks-popover/components/popover-content.js
```diff
@@ -119,16 +111,84 @@ export default class PopoverContent extends React.Component<Props> {
         }
     }
 
+    /**
+     * Runtime validation in case we try to use an invalid shape
+     */
+    validateProps({placement}: PopoverContextType) {
```
somewhatabstract: Awesome!

packages/wonder-blocks-popover/components/popover-content.js
```diff
@@ -230,9 +253,18 @@ const styles = StyleSheet.create({
...
      * Icon styles
      */
-    icon: {
-        marginRight: Spacing.medium,
+    iconContainer: {
+        alignItems: "center",
+        justifyContent: "center",
+        height: 64,
         width: 64,
+        minWidth: 64,
```
somewhatabstract: Spacing constants?

packages/wonder-blocks-popover/components/popover-content.test.js
```diff
@@ -32,34 +34,29 @@ describe("PopoverContent", () => {
...
+                image={<img src="/dummy-image.png" alt="popover image" />}
+                icon={<img src="/dummy-icon.png" alt="popover icon" />}
+            />
+        );
+
+        // Act
+        const underTest = () => mount(nodes);
+
+        // Assert
+        expect(underTest).toThrowError(
```
somewhatabstract: Why not toThrowErrorMatchingInlineSnapshot? That way it will auto update for us too.

packages/wonder-blocks-popover/components/popover-content.test.js
```diff
@@ -68,7 +65,14 @@ describe("PopoverContent", () => {
...
                 />
-            </PopoverContext.Provider>,
+            </PopoverContext.Provider>
+        );
+
+        // Act
+        const underTest = () => mount(nodes);
+
+        // Assert
+        expect(underTest).toThrowError(
```
somewhatabstract: Snapshot?

packages/wonder-blocks-tooltip/components/tooltip-popper.js
```diff
@@ -20,6 +20,8 @@ type Props = {|
     /**
      * This uses the children-as-a-function approach, mirroring react-popper's
      * implementation.
+     *
+     * TODO(juan): figure out to only allow TooltipBubble and PopoverDialog
```
somewhatabstract: Should this have a WB issue number tagged in it?

packages/wonder-blocks-popover/components/popover-content.test.js
jandrade: 👍 TIL. This matcher is awesome! thanks for the suggestion :)

PR #441, opened 2019-06-28
[fix:breadcrumbs] WB-562 WB-600: Breadcrumbs fixes > aria-current and testId

jandrade:
## Breadcrumbs
- Changed `aria-current="page"` to be added automatically to the last breadcrumb item [WB-562]
- Added/fixed `testId` prop to work both Breadcrumb and BreadcrumbItem. [WB-600]

### Test plan
1. Open https://deploy-preview-441--wonder-blocks.netlify.com/#breadcrumbs
2. Make sure the example includes `aria-current="page"` automatically.
3. Add a `testId` in the source code example and check the HTML now includes `data-test-id="the-testId-value"`

### DIFF
--- a/packages/wonder-blocks-breadcrumbs/components/breadcrumbs.js
+++ b/packages/wonder-blocks-breadcrumbs/components/breadcrumbs.js
@@ -37,7 +37,7 @@ export default class Breadcrumbs extends React.Component<Props> {
     // Moved it here, in case we need to override the label for a different
     // language
     static defaultProps = {
-        "aria-label": "Breadcrumb",
+        "aria-label": "Breadcrumbs",
     };
 
     render() {
--- a/packages/wonder-blocks-breadcrumbs/components/breadcrumbs.js
+++ b/packages/wonder-blocks-breadcrumbs/components/breadcrumbs.js
@@ -34,21 +34,30 @@ const StyledList = addStyle("ol");
  * 2. Separator: Adds a separator between each item.
  */
 export default class Breadcrumbs extends React.Component<Props> {
+    // Moved it here, in case we need to override the label for a different
+    // language
+    static defaultProps = {
+        "aria-label": "Breadcrumb",
+    };
+
     render() {
-        const {children} = this.props;
+        const {children, testId, ...otherProps} = this.props;
         // using React.Children allows to deal with opaque data structures
         // e.g. children = 'string' vs children = []
         const lastChildIndex = React.Children.count(children) - 1;
 
         return (
-            <nav aria-label="Breadcrumb">
+            <nav {...otherProps} data-test-id={testId}>
                 <StyledList style={styles.container}>
-                    {React.Children.map(children, (item, index) =>
-                        React.cloneElement(item, {
+                    {React.Children.map(children, (item, index) => {
+                        const isLastChild = index === lastChildIndex;
+
+                        return React.cloneElement(item, {
                             ...item.props,
-                            showSeparator: index !== lastChildIndex,
-                        }),
-                    )}
+                            showSeparator: !isLastChild,
+                            ["aria-current"]: isLastChild ? "page" : undefined,
+                        });
+                    })}
                 </StyledList>
             </nav>
         );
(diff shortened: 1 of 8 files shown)


jeresig:

Built with commit 41f4628e52fb82681563a4ea8df0ac702d881d74

https://deploy-preview-441--wonder-blocks.netlify.com

packages/wonder-blocks-breadcrumbs/components/breadcrumbs.js
```diff
@@ -34,21 +34,30 @@ const StyledList = addStyle("ol");
  * 2. Separator: Adds a separator between each item.
  */
 export default class Breadcrumbs extends React.Component<Props> {
+    // Moved it here, in case we need to override the label for a different
+    // language
+    static defaultProps = {
+        "aria-label": "Breadcrumb",
```
kevinbarabash: Should this be `"Breadcrumbs"` maybe?

packages/wonder-blocks-breadcrumbs/components/breadcrumbs.test.js
```diff
@@ -0,0 +1,27 @@
...
+                <BreadcrumbsItem>First</BreadcrumbsItem>
+                <BreadcrumbsItem>Last</BreadcrumbsItem>
+            </Breadcrumbs>,
+        );
+
+        // Act
+        const lastItem = wrapper.find(BreadcrumbsItem).last();
+
+        // Assert
+        expect(lastItem.prop("aria-current")).toBe("page");
```
kevinbarabash: Nice!

packages/wonder-blocks-breadcrumbs/components/breadcrumbs.test.js
```diff
@@ -0,0 +1,27 @@
...
+
+describe("Breadcrumbs", () => {
+    afterEach(() => {
+        unmountAll();
+    });
+
+    it("should set aria-current to the last item", () => {
+        // Arrange
+        const wrapper = mount(
+            <Breadcrumbs>
```
kevinbarabash: Can you add a test to verify `data-test-id` gets set when `testId` is present on `Breadcrumbs`?

packages/wonder-blocks-core/util/types.js
```diff
@@ -100,6 +100,15 @@ export type AriaProps = {|
     "aria-busy"?: "true" | "false",
     "aria-checked"?: "false" | "mixed" | "true" | "undefined",
     "aria-controls"?: IdRefList,
+    // https://www.w3.org/TR/wai-aria-1.1/#aria-current
+    "aria-current"?:
+        | "false"
```
kevinbarabash: TIL, thanks for linking to the spec.

packages/wonder-blocks-breadcrumbs/components/breadcrumbs.js
kevinbarabash: Also, can you create ticket to translate this eventually.

packages/wonder-blocks-breadcrumbs/components/breadcrumbs.test.js
jandrade: Added a second unit test on line # 28

packages/wonder-blocks-core/util/types.js
kevinbarabash: Did I completely miss `"aria-current"`?  That's embarrassing. 🙈

packages/wonder-blocks-breadcrumbs/components/breadcrumbs.test.js
kevinbarabash: Thanks!

packages/wonder-blocks-core/util/types.js
jandrade: You don't have to 😃. I checked the original link used for reference (https://www.w3.org/WAI/PF/aria-1.1/roles#role_definitions) and they don't include `aria-current` there.

I'm going to create a ticket to check if there are more attributes that would be good to add!

jandrade: https://khanacademy.atlassian.net/browse/WB-611

PR #442, opened 2019-06-28
WB-409: add aria-invalid, remove aria-checked

kumailn:
Added aria-invalid to Checkboxes and Radio buttons, also removed aria-checked from Checkboxes as per WB-584.

### DIFF
--- a/packages/wonder-blocks-form/components/radio-core.js
+++ b/packages/wonder-blocks-form/components/radio-core.js
@@ -59,6 +59,7 @@ const StyledInput = addStyle("input");
                 <StyledInput
                     {...sharedProps}
                     type="radio"
+                    aria-invalid={error}
                     aria-checked={checked}
                     checked={checked}
                     disabled={disabled}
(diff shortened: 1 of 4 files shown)


jeresig:

Built with commit 4e081be8ec736e21502f99394a653ae2052262dd

https://deploy-preview-442--wonder-blocks.netlify.com

packages/wonder-blocks-form/components/radio-core.js
```diff
@@ -59,6 +59,7 @@ const StyledInput = addStyle("input");
                 <StyledInput
                     {...sharedProps}
                     type="radio"
+                    aria-invalid={error}
                     aria-checked={checked}
```
jandrade: Could you please remove `aria-checked` from RadioButtons as well?

kevinbarabash: This reminds me of that other PR a while back.  It seems like if there's already an attribute for checked or disabled or whatever we don't need to add the `aria-` version of it.  It's too bad there isn't way to check for this automatically.

PR #444, opened 2019-07-02
Refactor Icon-Button to implement Clickable

kumailn:
Work on refactoring existing components to implement the new Clickable component, this is work in progress.

PR #445, opened 2019-07-08
Refactor Button to implement Clickable

kumailn:
Refactored Button/ButtonCore to implement the Clickable component, the refactor is similar to https://github.com/Khan/wonder-blocks/pull/444

### DIFF
--- a/packages/wonder-blocks-button/components/button-core.js
+++ b/packages/wonder-blocks-button/components/button-core.js
@@ -1,29 +1,22 @@
 // @flow
 import * as React from "react";
-import {StyleSheet} from "aphrodite";
-import {Link} from "react-router-dom";
 import PropTypes from "prop-types";
-
-import {LabelLarge, LabelSmall} from "@khanacademy/wonder-blocks-typography";
+import {StyleSheet} from "aphrodite";
+import Icon from "@khanacademy/wonder-blocks-icon";
+import {View} from "@khanacademy/wonder-blocks-core";
+import Spacing from "@khanacademy/wonder-blocks-spacing";
 import Color, {
     SemanticColor,
     mix,
     fade,
 } from "@khanacademy/wonder-blocks-color";
-import {addStyle} from "@khanacademy/wonder-blocks-core";
+import type {ClickableState} from "@khanacademy/wonder-blocks-core";
 import {CircularSpinner} from "@khanacademy/wonder-blocks-progress-spinner";
-import Icon from "@khanacademy/wonder-blocks-icon";
-import Spacing from "@khanacademy/wonder-blocks-spacing";
-
-import type {
-    ClickableHandlers,
-    ClickableState,
-} from "@khanacademy/wonder-blocks-core";
+import {LabelLarge, LabelSmall} from "@khanacademy/wonder-blocks-typography";
 import type {SharedProps} from "./button.js";
 
 type Props = {|
     ...SharedProps,
-    ...ClickableHandlers,
     ...ClickableState,
 |};
 
@@ -32,20 +25,18 @@ export default class ButtonCore extends React.Component<Props> {
 
     render() {
         const {
-            children,
-            color,
-            disabled: disabledProp,
-            focused,
-            hovered,
             kind,
-            light,
-            pressed,
             size,
-            style,
-            testId,
-            spinner,
             icon,
-            ...handlers
+            style,
+            color,
+            light,
+            spinner,
+            hovered,
+            focused,
+            pressed,
+            children,
+            disabled: disabledProp,
         } = this.props;
 
         const buttonColor =
@@ -76,13 +67,6 @@ export default class ButtonCore extends React.Component<Props> {
             size === "small" && sharedStyles.small,
         ];
 
-        const commonProps = {
-            "data-test-id": testId,
-            role: "button",
-            style: [defaultStyle, style],
-            ...handlers,
-        };
-
         const Label = size === "small" ? LabelSmall : LabelLarge;
 
         const label = (
@@ -102,7 +86,7 @@ export default class ButtonCore extends React.Component<Props> {
         );
 
         return (
-            <React.Fragment>
+            <View style={[defaultStyle, style]}>
                 {label}
                 {spinner && (
                     <CircularSpinner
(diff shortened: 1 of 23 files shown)


jeresig:

Built with commit f1abb178e1daba28e90272e5b52d83a83f9aad64

https://deploy-preview-445--wonder-blocks.netlify.com

packages/wonder-blocks-button/custom-snapshot.test.js
```diff
@@ -48,11 +49,9 @@ describe("ButtonCore", () => {
                                         size={size}
                                         color={color}
                                         light={light}
-                                        tabIndex={disabled ? -1 : 0}
```
kevinbarabash: I see, `disabled` is part of `stateProps`.

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -104,48 +86,29 @@ export default class ButtonCore extends React.Component<Props> {
                         style={sharedStyles.icon}
                     />
                 )}
-                {children}
+                <View style={sharedStyles.parentWidth}>{children}</View>
```
jandrade: This would translate to: `button > div > children`, right?

If that's the case, I'd suggest to remove `View` here and try to use a different alternative.

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -104,48 +86,29 @@ export default class ButtonCore extends React.Component<Props> {
...
             </Label>
         );
-
-        const contents = (
-            <React.Fragment>
+        return (
+            <View style={[defaultStyle]}>
```
jandrade: Same here with using `View` (div).

packages/wonder-blocks-button/components/button.js
```diff
@@ -102,6 +101,8 @@ export type SharedProps = {|
     /**
      * Optional custom styles.
      */
+    innerStyle?: StyleType,
+
```
jandrade: Could you add a comment explaining the difference between `innerStyle` and `style`? I think adding a comment to `style` would help.

packages/wonder-blocks-button/components/button-core.js
kumailn: It would unfortunately, but how can I apply this style without View? The button can't be directly styled anymore because the StyledButton is being generated by Clickable and not ButtonCore anymore

packages/wonder-blocks-button/components/button-core.js
kumailn: I would need a way to apply the style as above

packages/wonder-blocks-button/components/button.js
kumailn: yes will do, I had to implement this prop due to the new View which wraps the buttons contents

packages/wonder-blocks-button/components/button-core.js
kevinbarabash: You could create a `StyledSpan` with `addStyle`.

packages/wonder-blocks-button/components/button.js
kevinbarabash: Agree that comments will be super helpful.  It might be worth adding a small diagram showing the nesting of elements showing which styles are being applied where.

kumailn:
> Also, are all the `wb-dropdown` changes coming from of a different PR/branch? I remember seeing those changes before, but not completely sure about it.

Yes the dropdown changes are from a different branch

packages/wonder-blocks-button/components/button-core.js
kumailn: I’ve changed both Views to Spans in the latest commit

jandrade: Thanks for the change, but now that I see the working example it looks that you are adding an extra element here. What's the reason for adding `parentWidth` here?

kumailn: Without the parentWidth here the content can’t be centered 

image

jandrade: Ohh, so now we are using 3 spans inside the `<button>`.
https://github.com/Khan/wonder-blocks/pull/445/files#diff-c74745c6ff6054372236d55d0a92bf2bR97

I'm starting to feel this refactor needs to take a different approach:

Before this refactor: `button > span > label`

After this refactor: `button > span > span > span > label`

kumailn: Yeah that’s correct, I agree this isn’t ideal and I’m open to any other ideas, but this is the only way I was able to get the styling consistent with the original

jandrade: Got it, thanks for sharing! let me think about any possible alternatives and will comment back here.

jandrade:
I also noticed that the spinner example is not working. Could you please check this?

https://deploy-preview-445--wonder-blocks.netlify.com/#!/Button/27

kumailn:
> I also noticed that the spinner example is not working. Could you please check this?
> 
> https://deploy-preview-445--wonder-blocks.netlify.com/#!/Button/27
> 
> There seems to be a problem with the `onClick` prop inside `Button.js`. It could be related to the logic inside `Clickable.js` maybe?

This was working before, I think this happened after I changed the View tags to Spans

packages/wonder-blocks-button/components/button-core.js
kevinbarabash: We may decide that it doesn't actually make sense to use `Clickable` internally.  I thought it would be nice too dedupe things, but if it has adverse side-effects then it may not be worth it.

kevinbarabash:
I'm still seeing the issue with buttons with lots of content extending past the bounds of their container.  Since updating existing components is a nice-to-have I think it's fine to let this sit for a while.  I think if we give it some breathing room we'll either come up with an idea that solves the issues or we'll decide that things are fine the way they are.  Clickable will still be useful to people wanting to making non-button/link things clickable in webapp.

packages/wonder-blocks-button/components/button-core.js
jandrade: Yes, it's great to abstract that logic away from Button, but I think that for this particular component, this is causing (for now) adding some non-semantic HTML.

jandrade: To give some more context to this, this is what the W3 validator throws:
```
Error: Element div not allowed as child of element button in this context.
...
Content model for element button:
Phrasing content, but there must be no interactive content descendant.
```

This means that only [phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content) is allowed inside buttons. Hope this helps :)

kevinbarabash:
I'm going to close this for now.  If we have some time in the future and come up with a good way to address some of the styling issues we can re-open this.

PR #446, opened 2019-07-08
Refactor link to implement Clickable

kumailn:
Refactored Link to use Clickable, also introduced a new optional `isLink` prop in Clickable to handle Link children.

### DIFF
--- a/packages/wonder-blocks-breadcrumbs/__snapshots__/generated-snapshot.test.js.snap
+++ b/packages/wonder-blocks-breadcrumbs/__snapshots__/generated-snapshot.test.js.snap
@@ -28,6 +28,7 @@ exports[`wonder-blocks-breadcrumbs example 1 1`] = `
       }
     >
       <a
+        aria-label=""
         className=""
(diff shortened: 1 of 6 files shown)


jeresig:

Built with commit ffd8e175cc8f29386fe3b944b009cda8c9c1e52b

https://deploy-preview-446--wonder-blocks.netlify.com

kevinbarabash:
I'm going to hold off on reviewing this until the other Clickable related reviews complete.  It think much of the feedback will be similar and there's no need to hash the same thing out three times.

kumailn:
> I'm going to hold off on reviewing this until the other Clickable related reviews complete. It think much of the feedback will be similar and there's no need to hash the same thing out three times.

I think this can actually be reviews before the others because it doesn’t have the same problem of using nested Divs to apply style. This refactor should keep the generated HTML the same as currently in master for the most part

kevinbarabash:
I'm going to close this for now.  If we have some time in the future and come up with a good way to address some of the styling issues we can re-open this.

PR #447, opened 2019-07-10
WB-558 Upgrade to the most recent version of Styleguidist

jandrade:
## Internal
- Upgraded React Storybook to its latest version (9.1.11).
- Updated code examples to match with RS requirements for this new version: Each example needed to import (`require`) all of its dependencies used in each snippet/block.

### Test Plan
1. Open https://deploy-preview-447--wonder-blocks.netlify.com/
2. Check that all the examples are being rendered.

### NOTE:
- Added modules using `require` instead of `import` because this will be done in a separate ticket: WB-559.
- In order to make `import` works correctly, we will need to update `utils/gen-snapshot-tests.js` to change the way we are parsing all the examples.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "pretty-quick": "^1.4.1",
     "react-docgen": "^2.20.1",
     "react-router-enzyme-context": "^1.2.0",
-    "react-styleguidist": "^7.0.20",
+    "react-styleguidist": "^9.1.11",
     "react-test-renderer": "^16.4.1",
     "rollup": "^0.64.1",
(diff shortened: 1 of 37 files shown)


jeresig:

Built with commit f5d841b0ec18e8ad1fa23fcd7b11e7d9bd9e79f4

https://deploy-preview-447--wonder-blocks.netlify.com

kevinbarabash:
I wonder how many of the security alerts will go away with this upgrade.

jandrade:
That was a Styleguidist requirement.

And now all the security alerts are gone 🎉

PR #448, opened 2019-07-11
WB-623: Implement single select OptionItems in Dropdown 

kumailn:
Added the ability to override the opened state of Dropdown and implemented a new prop to treat OptionItems as single selects (select only one & close dropdown when clicked)

### DIFF
--- a/packages/wonder-blocks-clickable/docs.md
+++ b/packages/wonder-blocks-clickable/docs.md
@@ -1,4 +1,4 @@
-The Clickable component allows you to easily turn any custom compoent into a clickable one with consistent behavior and accesibility features.
+The Clickable component allows you to easily turn any custom component into a clickable one with consistent behavior and accesibility features.
 
 **Clickable allows your components to:**
 - Handle mouse / touch / keyboard events
--- a/packages/wonder-blocks-dropdown/components/dropdown.js
+++ b/packages/wonder-blocks-dropdown/components/dropdown.js
@@ -38,7 +38,7 @@ type Props = {|
      * Closes the Dropdown when an OptionItem is selected, use this
      * prop if you want single-select OptionItems
      */
-    closeOnOptionSelect?: boolean,
+    singleSelectOption?: boolean,
 
     /**
      * A callback that returns items that are newly selected. Use only if this
@@ -128,7 +128,7 @@ export default class Dropdown extends React.Component<Props, State> {
     };
 
     handleOptionSelected = (selectedValue: string) => {
-        const {onChange, selectedValues} = this.props;
+        const {onChange, selectedValues, singleSelectOption} = this.props;
 
         if (!onChange || !selectedValues) {
             return;
@@ -140,13 +140,17 @@ export default class Dropdown extends React.Component<Props, State> {
                 ...selectedValues.slice(0, index),
                 ...selectedValues.slice(index + 1),
             ];
-            onChange(updatedSelection);
+            onChange(singleSelectOption ? [selectedValue] : updatedSelection);
         } else {
             // Item was newly selected
-            onChange([...selectedValues, selectedValue]);
+            onChange(
+                singleSelectOption
+                    ? [selectedValue]
+                    : [...selectedValues, selectedValue],
+            );
         }
 
-        this.props.closeOnOptionSelect
+        singleSelectOption
             ? this.handleOpenChanged(false)
             : this.handleItemSelected();
     };
--- a/packages/wonder-blocks-dropdown/components/dropdown.js
+++ b/packages/wonder-blocks-dropdown/components/dropdown.js
@@ -9,12 +9,13 @@ import type {
     ClickableState,
 } from "@khanacademy/wonder-blocks-core";
 import {getClickableBehavior} from "@khanacademy/wonder-blocks-core";
-import DropdownCore from "./dropdown-core.js";
-import type {Item, DropdownItem} from "../util/types.js";
 import ActionItem from "./action-item.js";
 import OptionItem from "./option-item.js";
+import DropdownCore from "./dropdown-core.js";
 import DropdownAnchor from "./dropdown-anchor.js";
+import type {Item, DropdownItem} from "../util/types.js";
 
+type SelectType = "single" | "multi";
 type Props = {|
     ...AriaProps,
 
@@ -38,7 +39,7 @@ type Props = {|
      * Closes the Dropdown when an OptionItem is selected, use this
      * prop if you want single-select OptionItems
      */
-    singleSelectOption?: boolean,
+    selectionType?: SelectType,
 
     /**
      * A callback that returns items that are newly selected. Use only if this
@@ -97,6 +98,7 @@ export default class Dropdown extends React.Component<Props, State> {
 
     static defaultProps = {
         alignment: "left",
+        selectionType: "multi",
     };
 
     constructor(props: Props) {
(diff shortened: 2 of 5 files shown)


jeresig:

Built with commit 17cf8ae872310ca2fd9e05e3ce7ea8a77ec59829

https://deploy-preview-448--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -34,6 +34,12 @@ type Props = {|
      */
     menuItems: Array<Item> | Item,
 
+    /**
+     * Closes the Dropdown when an OptionItem is selected, use this
+     * prop if you want single-select OptionItems
+     */
+    singleSelectOption?: boolean,
```
jandrade: What do you think about using `singleSelectEnabled` or `singleSelectionEnabled` for consistency with booleans.

To give you some context, there are a lot of props that don't use this naming, but we are trying to get a more consistent API naming, so it would be helpful to name new props this way :)

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -101,6 +107,12 @@ export default class Dropdown extends React.Component<Props, State> {
         };
     }
 
+    componentDidMount() {
+        if (this.props.opened === true || this.props.opened === false) {
```
jandrade: Is there any particular reason to evaluate opened here? Based on my initial observation, it seems you could remove this condition, but I want to double check with you first.

jandrade:
It's great that you added an example showcasing this new prop. Could you also add a unit test checking this new prop is behaving properly?

packages/wonder-blocks-clickable/docs.md
```diff
@@ -1,4 +1,4 @@
-The Clickable component allows you to easily turn any custom compoent into a clickable one with consistent behavior and accesibility features.
+The Clickable component allows you to easily turn any custom component into a clickable one with consistent behavior and accesibility features.
```
somewhatabstract: Nice find! I wonder how hard it would be to get some sort of spellchecker thing to run on our docs. Not something for you to tackle, just musing aloud.

packages/wonder-blocks-dropdown/components/dropdown.js
somewhatabstract: Could this be even more descriptive of its purpose? Like `closeOnSingleSelect` or `closeOnSingleSelectEnabled`?

While the docs here explain what it does, those docs aren't present at places where this actually gets set and having a more descriptive name will then mean that uses are immediately easier to grok.

packages/wonder-blocks-dropdown/components/dropdown.js
somewhatabstract: Could this just be `this.props.opened != null`?

somewhatabstract:
This needs unit tests to validate the updated behavior.

packages/wonder-blocks-dropdown/components/dropdown.js
kumailn: @jandrade I wanted to override the opened state only if opened was passed in, I think @somewhatabstract’s suggestion works the same

packages/wonder-blocks-dropdown/components/dropdown.js
kumailn: @jandrade I included `Option` in the name because I wanted to convey that this prop only affected the behavior of OptionItems. @somewhatabstract It doesn’t only close the Dropdown however, it also defines the single select behavior by updating the state of the parent component accordingly

jandrade: If that's the case, I'd suggest renaming this prop to something more generic and familiar for devs that previously have worked with `<SingleSelect/>` and `<MultiSelect/>` components. I understand your point about restricting the behavior with the prop name, but I think that maybe it could be better to define that behavior in the value. This way, there would be a mapping between the Dropdown component and the other variants (SingleSelect, MultiSelect and ActionMenu if needed).

Instead of using a boolean here, you could use a union type with values that could help to describe the behavior you are trying to apply.

Here are some examples to try to illustrate my point a bit better:
e.g. 
a. `selectAs: "singleSelect" | "multiSelect"`
b. `type: "single" | "multi"`
c. `selectionType: "single" | "multi"`

..and making one of those options as default.

This way, this prop could be used later without needing to add a new prop for a given behavior.

Any thoughts?

kevinbarabash: I like an enum for this kind of thing.

packages/wonder-blocks-dropdown/components/dropdown.md
```diff
@@ -176,6 +176,7 @@ class MixedDropdownExample extends React.Component {
 
         return (
             <Dropdown
+                singleSelectOption
```
kevinbarabash: I think we should use `singleSelectOption={true}` since it's less confusing.  Some people who are new to React may confuse `singleSelectOption` with `singleSelectOption={singleSelectOption}`.

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -257,5 +257,63 @@ describe("Dropdown", () => {
...
+            const controlledComponent = mount(
+                <ControlledComponent singleSelectOption />,
+            );
+            controlledComponent.simulate("click");
+            controlledComponent.setState({selectedValues: ["B"]});
+            const dropdownCore = controlledComponent.find(DropdownCore);
+
+            // Act
+            dropdownCore.simulate("keydown", {keyCode: keyCodes.down});
+            dropdownCore.simulate("keyup", {keyCode: keyCodes.down});
```
kevinbarabash: This should probably be part of `// Arrange` since the thing that's being tested here is what happens when an option is selected.

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -257,5 +257,63 @@ describe("Dropdown", () => {
...
+
+            const optionItem = controlledComponent.find(OptionItem).at(0);
+            optionItem.simulate("click");
+
+            // Assert
+            expect(controlledComponent.state("selectedValues")).toHaveLength(1);
```
kevinbarabash: This indicates a slight inconsistency in the API between `SingleSelect` and `Dropdown` in select-select mode.  `SingleSelect`'s `onChange` passes back a single value where as `Dropdown`'s `onChange` passes an array even in single select mode.  I don't think this is something we need to solve in this PR.  Maybe this inconsistency is fine, but it is something to think about.

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -101,6 +107,12 @@ export default class Dropdown extends React.Component<Props, State> {
         };
     }
 
+    componentDidMount() {
+        if (this.props.opened != null) {
+            this.handleOpenChanged(this.props.opened);
+        }
```
kevinbarabash: This change feels pretty orthogonal to the single-select change and could've gone in a separate PR.

packages/wonder-blocks-clickable/docs.md
kevinbarabash: I wonder if it could be configured to only spell-check non-code parts of files and ignore words in the comments that are also in the code.

packages/wonder-blocks-dropdown/components/dropdown.md
kumailn: Sure, we also do something similar with `skipClientNav` which is why I did it here

packages/wonder-blocks-dropdown/components/dropdown.js
kumailn: @jandrade I agree, I’ve updated it in my latest commit to reflect your suggesiton

PR #453, opened 2019-07-19
[fix:popover, modal] WB-628 fix(internal): Render Modal and Popover examples correctly

jandrade:
## Summary
Moved `Modal` and `Popover` examples to a different MD file due to a React Styleguidist v9 (**RS**) issue that led to some previewing inconsistencies when trying to render examples that included React Context.

The problem is that **RS** is having some issues when trying to render components that are located inside the `components` folder and use React Context in their logic.

e.g. Any modal examples inside `wb-modal/components/modal-launcher.md` won't close when clicking on the Close Button due to the issue described below.

This is temporarily fixed by moving these examples to the main MD file: `wb-modal/docs.md`.

This issue has been already reported to the RS GitHub project: https://github.com/styleguidist/react-styleguidist/issues/1374

### Note
 I still don't know the exact reasons why this is happening in some isolated scenarios. Definitively, I strongly suggest to keep an eye about the React Storybooks progress on the Docs feature (currently on Beta).

### Test plan
1. Open https://deploy-preview-453--wonder-blocks.netlify.com/#!/Modal
2. Make sure the Modal examples open, then close correctly after clicking on the close button.
3. Open https://deploy-preview-453--wonder-blocks.netlify.com/#section-popover
4. Make sure the Popover examples close correctly after clicking on the close button.


jeresig:

Built with commit ca9f05b1d3acd25a795d9c5f38f58b172b45b928

https://deploy-preview-453--wonder-blocks.netlify.com

kevinbarabash:
That's bizarre that just moving the location of the examples fixes the issue.

packages/wonder-blocks-modal/docs.md
```diff
@@ -1,4 +1,209 @@
...
+[here](https://deploy-preview-389--wonder-blocks.netlify.com/#modal)
+
+## Examples
+
+### Example: Default modal
+
+Once the modal is launched, tab focus wraps inside the modal content. Pressing Tab at the end of the modal will focus the modal's first element, and pressing Shift-Tab at the start of the modal will focus the modal's last element.
+
+```js
+const {StyleSheet} = require("aphrodite");
```
jeresig: I assume there will be some merge conflicts with your import diff here.

packages/wonder-blocks-popover/package.json
```diff
@@ -18,6 +18,8 @@
   "dependencies": {
     "@khanacademy/wonder-blocks-color": "^1.1.4",
     "@khanacademy/wonder-blocks-core": "^2.3.0",
+    "@khanacademy/wonder-blocks-icon": "^1.2.2",
```
jeresig: All these dependencies were missing - or were they only used in the docs? If they're only in the docs then adding them to `devDependencies` would probably be better.

jandrade: They were missing dependencies. I'm using them to create the CloseButton component.

packages/wonder-blocks-modal/docs.md
jandrade: I created the import diff based on this PR ....and I forgot to mention that in the other PR :|

jandrade:
Yes, it's too odd, I'm still trying to check the RS source code to understand how they build that, but no clues so far. I'll try to take some time later to see if I can find an alternative/better solution for this.

PR #456, opened 2019-07-26
[feat:dropdown] WB-633: Fix controlled mode in Dropdown

jandrade:
## Summary
Fixes the `Dropdown` component to make it fully controllable.

### Dropdown
- Added `getDerivedStateFromProps` to sync the `opened` prop with the parent's state.
- Added `onClose` prop to sync up the `opened` prop within the parent component (similar to `ModalLauncher` and `Popover`).
- Added `onClick` prop to the Dropdown anchor to keep it in sync with the parent state.
- Added an example explaining this behavior.
- Updated unit tests.

### Test plan
1. Open https://deploy-preview-456--wonder-blocks.netlify.com/#dropdown-1
2. Go to the example: **Opening a dropdown programmatically** and check that works as expected.
3. Also, check the explanation is clear enough.

dropdown-controlled

### DIFF
--- a/packages/wonder-blocks-dropdown/components/dropdown.js
+++ b/packages/wonder-blocks-dropdown/components/dropdown.js
@@ -6,6 +6,7 @@ import {StyleSheet} from "aphrodite";
 import type {
     AriaProps,
     StyleType,
+    ClickableHandlers,
     ClickableState,
 } from "@khanacademy/wonder-blocks-core";
 import {getClickableBehavior} from "@khanacademy/wonder-blocks-core";
@@ -26,9 +27,14 @@ type Props = {|
     children: (eventState: ClickableState) => React.Element<any>,
 
     /**
-     * Can be used to override the state of the Dropdown by parent elemnents
+     * Can be used to override the state of the Dropdown by parent elements
      */
     opened?: boolean,
+
+    /**
+     * Called when the menu closes
+     */
+    onClose?: () => mixed,
 
     /**
      * The items present in the Dropdown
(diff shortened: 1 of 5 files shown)


jeresig:

Built with commit c69d6e0415f9a44f9260ba4aa999842713aad272

https://deploy-preview-456--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -26,10 +27,15 @@ type Props = {|
...
     /**
-     * Can be used to override the state of the Dropdown by parent elemnents
+     * Can be used to override the state of the Dropdown by parent elements
      */
     opened?: boolean,
 
+    /**
+     * Called when the menu closes
+     */
+    onClose?: () => mixed,
```
jandrade: Added this new `onClose` prop here to update the controlled component from its parent. Initially I tried to use the existing `onChange` prop, but found some issues when trying to use the some keyboard keys (`esc`, `enter`) and when clicking outside the menu (`document.mouseout`). Also, this feels more appropriate b/c if we close the menu by clicking outside, that means the dropdown hasn't changed any selection.

packages/wonder-blocks-dropdown/components/dropdown.md
```diff
@@ -203,3 +203,113 @@ class MixedDropdownExample extends React.Component {
...
+                                eventState.focused && styles.focused,
+                                eventState.hovered && styles.hovered,
+                            ]}
+                        >
+                            Open custom dropdown
+                        </HeadingSmall>
+                    )}
+                </Dropdown>
+                <Strut size={Spacing.medium} />
+                <Button onClick={this.openDropdown}>Open dropdown programatically</Button>
```
jeresig: After playing around with this I think it'd be nice if this toggled the dropdown open/closed.

Also, I was surprised that I was able to close the dropdown by clicking out of it - if it's a controlled component wouldn't that be entirely up to the `ControlledPopoverExample` to manage?

jandrade: Good call about the toggle fn.

And about the dropdown being closed when clicking outside, it's being managed inside the `ControlledPopoverExample` by using the `onClose` prop. If you take a look into the `handleClose` function, there's where I'm controlling the Dropdown to be hidden. Maybe the documentation needs to be more explicit about the `onClose` usage?

jeresig: Ohh, you're right - I missed that, sorry!

jandrade: But you're right, if you missed that, it could happen that the usage might be a bit confusing :). I'm going to improve this documentation in my upcoming PR (adding the controlled mode to the other dropdowns). Thanks for bringing that up!

PR #457, opened 2019-07-26
[feature:dropdown] WB-632 - Add controlled mode to ActionMenu, SingleSelect and MultiSelect

jandrade:
## Summary
Adds a "controlled" mode to all the dropdown variants. To make this mode work, you need to add the `opened` and `onToggle` props to the component instance. The dropdown openers still work on this mode due to the dropdown/select nature.

### Dropdown
- Changed `onClose` to `onToggle` for better understanding on the controlled mode. This means that `onToggle` will be triggered every time the menu is open or closed.

### ActionMenu, SingleSelect, MultiSelect
- Added `getDerivedStateFromProps` to sync the `opened` prop with the parent's state.
- Added `onToggle` prop to sync up the `opened` prop within the parent component.
- Updated unit tests.
- Added examples.

### Test plan
1. Open https://deploy-preview-457--wonder-blocks.netlify.com/#section-dropdown
2. Check the documentation is clear enough.
3. Go to the examples related to controlled components and check that everything works as expected:
    - ActionMenu: https://deploy-preview-457--wonder-blocks.netlify.com/#!/ActionMenu/11
    - SingleSelect: https://deploy-preview-457--wonder-blocks.netlify.com/#!/SingleSelect/15
    - MultiSelect: https://deploy-preview-457--wonder-blocks.netlify.com/#!/MultiSelect/15

### DIFF
--- a/packages/wonder-blocks-dropdown/components/action-menu.js
+++ b/packages/wonder-blocks-dropdown/components/action-menu.js
@@ -7,7 +7,7 @@ import Dropdown from "./dropdown.js";
 import ActionMenuOpenerCore from "./action-menu-opener-core.js";
 import type {Item} from "../util/types.js";
 
-type MenuProps = {|
+type Props = {|
     ...AriaProps,
 
     /**
@@ -19,6 +19,17 @@ type MenuProps = {|
      * Text for the opener of this menu.
      */
     menuText: string,
+
+    /**
+     * Can be used to override the state of the ActionMenu by parent elements
+     */
+    opened?: boolean,
+
+    /**
+     * In controlled mode, use this prop in case the parent needs to be notified
+     * when the menu opens/closes.
+     */
+    onToggle?: (opened: boolean) => mixed,
 
     /**
      * A callback that returns items that are newly selected. Use only if this
@@ -75,7 +86,7 @@ type State = {|
 /**
  * A menu that consists of various types of items.
  */
-export default class ActionMenu extends React.Component<MenuProps, State> {
+export default class ActionMenu extends React.Component<Props, State> {
     openerElement: ?HTMLElement;
 
     static defaultProps = {
@@ -85,6 +96,31 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
 
     state = {
         opened: false,
+    };
+
+    /**
+     * Used to sync the `opened` state when this component acts as a controlled
+     * component
+     */
+    static getDerivedStateFromProps(props: Props, state: State) {
+        return {
+            opened:
+                typeof props.opened === "boolean" ? props.opened : state.opened,
+        };
+    }
+
+    /**
+     * Update the internal state and notify the parent component when menu is
+     * toggled
+     */
+    handleToggleMenu = (opened: boolean) => {
+        this.setState({
+            opened: opened,
+        });
+
+        if (this.props.onToggle) {
+            this.props.onToggle(opened);
+        }
     };
 
     render() {
@@ -99,18 +135,22 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
             /* eslint-disable no-unused-vars */
             children,
             onChange,
+            onToggle,
+            opened,
             selectedValues,
             style,
             "aria-disabled": ariaDisabled, // WB-535 avoids passing this prop to the opener
             /* eslint-enable no-unused-vars */
             ...sharedProps
         } = this.props;
-
         const menuItems = React.Children.toArray(this.props.children);
+
         return (
             <Dropdown
                 style={style}
                 onChange={onChange}
+                onToggle={this.handleToggleMenu}
+                opened={this.state.opened}
                 alignment={alignment}
                 menuItems={menuItems}
                 selectedValues={selectedValues}
@@ -132,12 +172,10 @@ export default class ActionMenu extends React.Component<MenuProps, State> {
         );
     }
 }
-
 const styles = StyleSheet.create({
     caret: {
         marginLeft: 4,
     },
-
     // The design calls for additional offset around the opener.
     opener: {
         whiteSpace: "nowrap",
@@ -145,7 +183,6 @@ const styles = StyleSheet.create({
         overflow: "hidden",
         textOverflow: "ellipsis",
     },
-
     // This is to adjust the space between the menu and the opener.
     menuTopSpace: {
         top: -4,
--- a/packages/wonder-blocks-dropdown/components/action-menu.md
+++ b/packages/wonder-blocks-dropdown/components/action-menu.md
@@ -187,3 +187,86 @@ const styles = StyleSheet.create({
     </ActionMenu>
 </View>
 ```
+
+### Example: Opening an ActionMenu programmatically
+
+Sometimes you'll want to trigger a dropdown programmatically. This can be done by
+setting the `opened` prop to `true`. In this situation the `ActionMenu` is a
+controlled component. The parent is responsible for managing the opening/closing
+of the dropdown when using this prop.
+
+This means that you'll also have to update `opened` for these situations:
+-  to `true`: inside the Dropdown children instance (dropdown anchor). See example
+   below.
+-  to `false`: in response to the `onClose` callback being triggered.
+
+```js
+import {Dropdown, ActionItem, OptionItem, SeparatorItem} from "@khanacademy/wonder-blocks-dropdown";
+import Button from "@khanacademy/wonder-blocks-button";
+import {View} from "@khanacademy/wonder-blocks-core";
+import {Strut} from "@khanacademy/wonder-blocks-layout";
+import Spacing from "@khanacademy/wonder-blocks-spacing";
+import {StyleSheet} from "aphrodite";
+
+const styles = StyleSheet.create({
+    row: {
+        flexDirection: "row",
+    }
+});
+
+class ControlledActionMenuExample extends React.Component {
+    constructor() {
+        super();
+        this.state = {
+            opened: false,
+            selectedValues: ["kumail"],
+        };
+        this.handleChange = this.handleChange.bind(this);
+        this.toggleMenu = this.toggleMenu.bind(this);
+    }
+
+    handleChange(update) {
+        this.setState({
+            selectedValues: update,
+        });
+    }
+
+    toggleMenu(opened) {
+        console.log("🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥");
+        console.log('should toggle!!!', opened);
+        this.setState({
+            opened: opened,
+        });
+    }
+
+    render() {
+        return (
+            <View style={styles.row}>
+                <ActionMenu
+                    menuText="Betsy Appleseed"
+                    onChange={this.handleChange}
+                    onToggle={(opened) => {
+                        console.log('toggle called!!!! ', opened);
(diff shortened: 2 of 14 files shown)


jeresig:

Built with commit 191020426bc98a331a58ab26bc77a65440bec3a1

https://deploy-preview-457--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -32,9 +32,10 @@ type Props = {|
     opened?: boolean,
 
     /**
-     * Called when the menu closes
+     * In controlled mode, use this prop in case the parent needs to be notified
+     * when the menu opens/closes.
      */
-    onClose?: () => mixed,
+    onToggle?: (opened: boolean) => mixed,
```
jeresig: Technically this would be a breaking change, but since we haven't released it yet this is ok!

packages/wonder-blocks-dropdown/components/dropdown.md
```diff
@@ -258,15 +256,18 @@ class ControlledPopoverExample extends React.Component {
         });
     }
 
-    toggleDropdown() {
+    toggleDropdown(opened) {
         this.setState({
-            opened: !this.state.opened,
+            opened: opened,
```
jeresig: Can be just `opened,`

packages/wonder-blocks-dropdown/components/action-menu.md
```diff
@@ -187,3 +187,82 @@ const styles = StyleSheet.create({
     </ActionMenu>
 </View>
 ```
+
+### Example: Opening an ActionMenu programmatically
+
+Sometimes you'll want to trigger a dropdown programmatically. This can be done by
+setting the `opened` prop to `true`. In this situation the `ActionMenu` is a
```
jeresig: It can also be the case that `false` works, too - maybe be good to make clear in this paragraph that any boolean passed in to `opened` will cause it to be controlled. (I missed this the first time I saw this!)

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -184,12 +184,12 @@ describe("Dropdown", () => {
                     selectedValues: update,
                 });
 
-            handleClose = () => {
+            handleToggle = (opened) => {
                 this.setState({
-                    opened: false,
+                    opened: opened,
```
jeresig: Here too.

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -20,6 +20,17 @@ type MenuProps = {|
...
+    /**
+     * Can be used to override the state of the ActionMenu by parent elements
+     */
+    opened?: boolean,
+
+    /**
+     * In controlled mode, use this prop in case the parent needs to be notified
+     * when the menu opens/closes.
+     */
+    onToggle?: (opened: boolean) => mixed,
```
kevinbarabash: We use `onToggle` on `OptionItem`.  Since the name of the `prop` that controls whether the action menu is opened or closed I think we should use either:
```
onOpenedChanged?: (opened: boolean) => mixed,
```
or
```
onClose?: () => mixed,
```
The latter is more common in our API.  Is there a need to know when `opened` is transitioning from `false => true`?  I wouldn't expect there to be since if the component is controlled then the parent is triggering this transition.

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -41,6 +41,17 @@ type Props = {|
...
+    /**
+     * Can be used to override the state of the ActionMenu by parent elements
+     */
+    opened?: boolean,
+
+    /**
+     * In controlled mode, use this prop in case the parent needs to be notified
+     * when the menu opens/closes.
+     */
+    onToggle?: (opened: boolean) => mixed,
```
kevinbarabash: Same concern as for `action-menu.js`.

packages/wonder-blocks-dropdown/components/multi-select.md
```diff
@@ -371,3 +371,77 @@ class ImplicitAllEnabledExample extends React.Component {
...
+
+class ControlledMultiSelectExample extends React.Component {
+    constructor() {
+        super();
+        this.state = {
+            opened: false,
+            selectedValues: [],
+        };
+        this.handleChange = this.handleChange.bind(this);
+        this.toggleMenu = this.toggleMenu.bind(this);
```
kevinbarabash: I'm assuming that styleguidist doesn't allow class properties.

packages/wonder-blocks-dropdown/components/multi-select.md
```diff
@@ -371,3 +371,77 @@ class ImplicitAllEnabledExample extends React.Component {
...
+    }
+
+    render() {
+        return (
+            <View style={styles.row}>
+                <MultiSelect
+                    selectItemType="fruits"
+                    onChange={this.handleChange}
+                    opened={this.state.opened}
+                    onToggle={this.toggleMenu}
```
kevinbarabash: What happens when a user clicks on the `MultiSelect` itself?

packages/wonder-blocks-dropdown/components/action-menu.js
jandrade: I originally used `onClose` but found some issues when trying to completely control this component from its parent. The main reason to change it to `onToggle` was because the trigger element (`Dropdown menu opener`) wasn't aware when the dropdown menu should open. 

`onClose` is good enough for the `Dropdown` component. Because that component works similar to the `ModalLauncher). In Modals, as we are passing the `openModal` as a renderProp, then we let the developer to implement any modal opener from the parent, then they don't need to be aware about when `opened` changes internally.

In `wb-dropdown`,  `Dropdown` works similarly, but the other variants (`ActionMenu`, `SingleSelect` and `MultiSelect`) need to know when to open, as we are hiding the opener implementation details to the devs.

And answering you last question, in dropdowns we know when `opened` transitions from `false` to `true`, and it's when the user clicks on the opener button.

What do you think?

packages/wonder-blocks-dropdown/components/multi-select.md
jandrade: yes, sadly they don't support them :(

packages/wonder-blocks-dropdown/components/multi-select.md
jandrade: It opens or hides the menu, depending on the internal `opened` state value.

packages/wonder-blocks-dropdown/components/action-menu.js
jandrade: That said, we could change the prop name. My intention about naming that prop that way was to use something short that devs could be familiar with and that it could be used (if needed) in another packages. It would be great if we could generalize that behavior to cover more scenarios.

I feel if we restrict that to `onOpenedChange`, then we couldn't reuse the same pattern for other package that don't necessarily use `open/opened` for X reasons.

packages/wonder-blocks-dropdown/components/single-select.test.js
```diff
@@ -81,4 +81,97 @@ describe("SingleSelect", () => {
...
+                <ControlledComponent onToggle={onToggleMock} />,
+            );
+
+            // Act
+            wrapper.find(SelectOpener).simulate("click");
+
+            // Assert
+            expect(onToggleMock).toHaveBeenCalledWith(true);
+        });
+    });
```
kevinbarabash: Can you add some tests to check that changing the `opened` prop from `true => false` and `false => true` does the right thing w.r.t the menu hiding/showing?

packages/wonder-blocks-dropdown/generated-snapshot.test.js
```diff
@@ -251,15 +251,18 @@ describe("wonder-blocks-dropdown", () => {
                 });
             }
 
-            toggleDropdown() {
+            toggleDropdown(opened) {
```
kevinbarabash: nit: `handleToggle`

packages/wonder-blocks-dropdown/components/dropdown.md
```diff
@@ -258,15 +256,18 @@ class ControlledPopoverExample extends React.Component {
         });
     }
 
-    toggleDropdown() {
+    toggleDropdown(opened) {
```
kevinbarabash: nit: handleToggle

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -136,11 +147,25 @@ export default class MultiSelect extends React.Component<Props, State> {
...
+            open: opened,
             keyboard,
         });
+
+        if (this.props.onToggle) {
+            this.props.onToggle(opened);
+        }
     };
 
     handleToggle = (selectedValue: string) => {
```
kevinbarabash: ~It might be a little confusing to someone looking at this latter why we don't call `this.props.onToggle`.  Can you add some comments here.  Maybe it would be good to name the prop `onToggleMenu` and then this event handler could be called `handleToggleSelection`.~

kevinbarabash: I think a more drastic refactoring though might be in order in a future PR.  I noticed that action-menu.js uses `Dropdown` but here we use `DropdownCore` directly.  I think we should be able to use `Dropdown` here since it has a `selectionType` prop now.

packages/wonder-blocks-dropdown/components/single-select.js
```diff
@@ -225,7 +258,7 @@ export default class SingleSelect extends React.Component<Props, State> {
                 keyboard={this.state.keyboard}
```
kevinbarabash: I noticed that we're using `DropdownCore` here as well.  Maybe in a future PR we can update this to use `Dropdown` instead similar to how `ActionMenu` works.

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -136,11 +147,25 @@ export default class MultiSelect extends React.Component<Props, State> {
...
+    static getDerivedStateFromProps(props: Props, state: State) {
+        return {
+            open: typeof props.opened === "boolean" ? props.opened : state.open,
+        };
+    }
+
+    handleOpenChanged = (opened: boolean, keyboard?: boolean) => {
         this.setState({
-            open,
+            open: opened,
```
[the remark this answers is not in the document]
jandrade: Created a new JIRA ticket to work on this soon: https://khanacademy.atlassian.net/browse/WB-634

packages/wonder-blocks-dropdown/components/multi-select.js
jandrade: This work will be also included in this ticket: https://khanacademy.atlassian.net/browse/WB-634

packages/wonder-blocks-dropdown/components/single-select.js
```diff
@@ -207,7 +240,7 @@ export default class SingleSelect extends React.Component<Props, State> {
                 isPlaceholder={!selectedItem}
                 light={light}
                 onOpenChanged={this.handleOpenChanged}
-                open={open}
+                open={this.state.open}
```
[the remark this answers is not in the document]
jandrade: Not necessarily a breaking change as we are not changing the public API (?) What do you think about it?

PR #458, opened 2019-07-30
WB-591: Add detailed documentation for Timing package

somewhatabstract:
Closes WB-591.

This makes the following changes:

1. Makes the Wonder Blocks Timing package public
2. Adds detailed documentation for the Timing API, including migration example
3. Updates the snapshot generation to cope with static code examples (so that they aren't compiled for snapshot testing)

### DIFF
--- a/packages/wonder-blocks-timing/components/with-action-scheduler.md
+++ b/packages/wonder-blocks-timing/components/with-action-scheduler.md
@@ -1,150 +1,19 @@
-This is a higher order component that attaches the given component to an
-`IScheduleActions` instance. Any actions scheduled will automatically be
+This is a higher order component (HOC) that attaches the given component to an
+[`IScheduleActions`](#ischeduleactions) instance. Any actions scheduled will automatically be
 cleared on unmount. This allows for "set it and forget it" behavior that won't
 leave timers dangling when the component's lifecycle ends.
 
-For example, the following component, `MyNaughtyComponent`, will keep spamming
-our pretend log even after it was unmounted.
+For more details on using this component and the [`IScheduleActions`](#ischeduleactions) interface,
+see the [API overview](#timing-api-overview).
 
-```jsx
-const Button = require("@khanacademy/wonder-blocks-button").default;
-const {IDProvider, View} = require("@khanacademy/wonder-blocks-core");
+### Flow Types
+If you are using Flow typing, you can use the `WithActionScheduler<TProps>`
+generic type to build the props for the component that you will pass to the
+`withActionScheduler` function, where `TProps` represents all the props your
+component uses other than the `schedule` prop that will get added.
 
+The added `schedule` prop is of type [`IScheduleActions`](#ischeduleactions). This is what the
+`withActionScheduler` function injects to your component.
 
-class Unmounter extends React.Component {
-    constructor() {
-        super();
-        this.state = {
-            mountKids: true,
-        };
-    }
-
-    maybeRenderKids() {
-        if (this.state.mountKids) {
-            return (
-                <React.Fragment>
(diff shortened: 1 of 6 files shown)


jeresig:

Built with commit 9190d4afdfea15f9ba0fcac37de90082981b406e

https://deploy-preview-458--wonder-blocks.netlify.com

packages/wonder-blocks-timing/components/with-action-scheduler.md
```diff
@@ -1,150 +1,19 @@
...
+### Flow Types
+If you are using Flow typing, you can use the `WithActionScheduler<TProps>`
+generic type to build the props for the component that you will pass to the
+`withActionScheduler` function, where `TProps` represents all the props your
+component uses other than the `schedule` prop that will get added.
 
+The added `schedule` prop is of type [`IScheduleActions`](#ischeduleactions). This is what the
+`withActionScheduler` function injects to your component.
 
-class Unmounter extends React.Component {
```
somewhatabstract: These have moved to the main package docs.

packages/wonder-blocks-timing/docs.md
```diff
@@ -1,11 +1,398 @@
...
+Access to the timing API is provided via the `withActionScheduler` higher order
+component.
+
+### Timing API Usage Example
+
+The following component, `MyNaughtyComponent`, will keep spamming our pretend
+log even after it was unmounted.
+
+```jsx
+const Button = require("@khanacademy/wonder-blocks-button").default;
```
jandrade: Not required but nice to have..... Now that we've updated Styleguidist, it would be great if we could use `import` here instead of `require`.

packages/wonder-blocks-timing/docs.md
```diff
@@ -1,11 +1,398 @@
...
+        resolveOnClear?: boolean,
+    ): ITimeout;
+```
+
+The `timeout` function replaces the `setTimeout` and `clearTimeout` functions
+in the standard timer API.
+
+Unless indicated otherwise, all timeouts scheduled this way will be cleared on component unmount.
+
+| | Flow&nbsp;Type | Default | Description |
```
jandrade: cool usage of md tables! curious to know if you use a tool to write them down or did it manually.

packages/wonder-blocks-timing/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@khanacademy/wonder-blocks-timing",
-  "private": true,
+  "private": false,
   "version": "0.0.3",
   "design": "v1",
   "publishConfig": {
```
jandrade: Could you please change this "access" key to "public"?

somewhatabstract: Oh darn. Thanks! I will.

somewhatabstract:
OK, @jandrade. I think that covers it.

packages/wonder-blocks-timing/docs.md
somewhatabstract: I did it manually.

packages/wonder-blocks-timing/docs.md
somewhatabstract: Done

packages/wonder-blocks-timing/docs.md
```diff
@@ -181,8 +181,6 @@ The `IScheduleActions` interface provides 4 (four) different functions:
 The `timeout` function replaces the `setTimeout` and `clearTimeout` functions
 in the standard timer API.
 
-Unless indicated otherwise, all timeouts scheduled this way will be cleared on component unmount.
```
somewhatabstract: This line was untrue in that there is no "indicated otherwise".

PR #460, opened 2019-07-30
[feature:modal] WB-636 Add testId prop to modal elements

jandrade:
## Summary
Adds the missing `testId` prop to the modal package.

### Public API
- **ModalLauncher**: Added `testId` prop to be passed down to the `ModalBackdrop` component.
- **OnePaneDialog**: Added `testId` prop to be passed down to the `ModalDialog` component.
- **ModalHeader**: Added `testId` prop to be used inside the main wrapper. Also, used for both title and subtitle elements.
    - **title**: For `<ModalHeader testId="some-id" />`, then title => e.g. `<h3 data-test-id="some-id-title" />`
    - **subtitle**: For `<ModalHeader testId="some-id" />`, then subtitle => e.g. `<h4 data-test-id="some-id-subtitle" />`
- **ModalPanel**: Added `testId` prop to be used inside the main wrapper. e.g `<div class="modal-panel" data-test-id"some-id-ModalPanel" />`

## Internal components
- CloseButton: Added `testId` prop combining the `testId` defined in the parent element + adding 
the `CloseButton` suffix => `some-id-CloseButton`.
- ModalBackdrop: Added `testId` prop combining the `testId` defined in the `ModalLauncher` + adding the `ModalBackdrop` suffix => `some-id-ModalBackdrop`.

### Test plan
1. Open https://deploy-preview-460--wonder-blocks.netlify.com/#!/Modal/1
2. Open the example source code.
3. Make sure the first example includes all the associated `testId` values.

### DIFF
--- a/packages/wonder-blocks-modal/components/modal-header.js
+++ b/packages/wonder-blocks-modal/components/modal-header.js
@@ -29,6 +29,19 @@ type Common = {|
     titleId: string,
 
     /**
+     * Test ID used for e2e testing.
+     *
+     * In this case, this component is internal, so `testId` is composed with
+     * the `testId` passed down from the Dialog variant + a suffix to scope it
+     * to this component.
+     *
+     * @example
+     * For testId="some-random-id"
+     * The result will be: `some-random-id-modal-header`
+     */
+    testId?: string,
+
+    /**
      * Without these, flow complains about subtitle and breadcrumbs not being
      * available on props at all b/c these are exact object types, flow looks
      * for subtitle in each of Common, WithSubtitle and WithBreadcrumbs. I also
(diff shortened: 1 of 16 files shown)


jeresig:

Built with commit dba972c427981978dc0ce6cb3e5295db9d4d8552

https://deploy-preview-460--wonder-blocks.netlify.com

packages/wonder-blocks-modal/components/modal-header.js
```diff
@@ -110,15 +139,26 @@ export default class ModalHeader extends React.Component<Props> {
...
+                    >
                         {breadcrumbs && (
                             <View style={styles.breadcrumbs}>
                                 {breadcrumbs}
                             </View>
                         )}
-                        <HeadingMedium id={titleId}>{title}</HeadingMedium>
+                        <HeadingMedium
+                            id={titleId}
+                            testId={testId && `${testId}-title`}
```
kevinbarabash: That's pretty cool.

packages/wonder-blocks-modal/components/modal-panel.js
```diff
@@ -114,17 +127,22 @@ export default class ModalPanel extends React.Component<Props> {
...
+            testId,
         } = this.props;
 
         const mainContent = this.renderMainContent();
 
         return (
-            <View style={[styles.wrapper, !light && styles.dark, style]}>
+            <View
+                style={[styles.wrapper, !light && styles.dark, style]}
+                testId={testId && `${testId}-ModalPanel`}
```
kevinbarabash: nit: the suffix here is probably overkill.

packages/wonder-blocks-modal/components/modal-panel.js
```diff
@@ -114,17 +127,22 @@ export default class ModalPanel extends React.Component<Props> {
...
+            >
                 {closeButtonVisible && (
                     <CloseButton
                         light={!light}
                         onClick={onClose}
                         style={styles.closeButton}
+                        testId={testId && `${testId}-CloseButton`}
```
kevinbarabash: nit: It's a little weird seeing a mix of kebab-case and PascalCase.  Maybe this should be `${testId}-close`.

packages/wonder-blocks-modal/components/one-pane-dialog/one-pane-dialog.js
```diff
@@ -139,6 +139,7 @@ export default class OnePaneDialog extends React.Component<Props> {
                     title={title}
                     subtitle={(subtitle: string)}
                     titleId={uniqueId}
+                    testId={testId && `${testId}-ModalHeader`}
```
kevinbarabash: Including `-ModalHeader` makes this a bit more awkward to use and embeds an implementation detail in the testing interface.  I think passing `testId` on its own should be fine.

packages/wonder-blocks-modal/components/modal-panel.js
jandrade: Added that suffix here to avoid duplication with the `testId` set in `ModalDialog`. Maybe `-panel` would be a better name?

packages/wonder-blocks-modal/components/one-pane-dialog/one-pane-dialog.js
jandrade: Same situation as described before. Added this suffix here to avoid duplication with the testId set in ModalDialog. Maybe `-header` would be a better name?

jandrade:
That makes sense to me. Added tests to cover those cases. One thing to notice is that I mostly had to use `mount` to test that some `testId`s were correctly passed down from the public components.

packages/wonder-blocks-modal/components/modal-header.test.js
```diff
@@ -52,4 +58,28 @@ describe("ModalHeader", () => {
         // Assert
         expect(wrapper.exists()).toBe(true);
     });
+
+    test("renders the title by default", () => {
```
jeresig: It would be more accurate for this to refer to the testId

packages/wonder-blocks-modal/components/modal-header.test.js
```diff
@@ -52,4 +58,28 @@ describe("ModalHeader", () => {
...
+
+        // Act
+        const title = wrapper.find(
+            `[data-test-id="test-example-header-title"]`,
+        );
+        const subtitle = wrapper.find(
+            `[data-test-id="test-example-header-subtitle"]`,
+        );
+
+        // Assert
```
jeresig: It's generally good practice to have your assert section only assert for one thing - I think this test can be split into two - one to check the title and one to check the subtitle.

packages/wonder-blocks-modal/components/modal-panel.test.js
```diff
@@ -18,4 +26,19 @@ describe("ModalPanel", () => {
...
+        // Arrange
+        const wrapper = mount(
+            <ModalPanel content="dummy content" testId="test-id" />,
+        );
+
+        // Act
+        const mainElement = wrapper.find(View).first();
+        const closeButton = wrapper.find(CloseButton);
+
+        // Assert
```
jeresig: Same here, too.

PR #462, opened 2019-08-01
Correct the type for WithActionScheduler

somewhatabstract:


jeresig:

Built with commit 99dacc9b6c82b1d341b83d6e25260243160879f3

https://deploy-preview-462--wonder-blocks.netlify.com

somewhatabstract:
I haven't worked out a test yet, but I think we could make one similar to how `flow-typed` tests their flow stuff with error expectation suppressions in unit tests (then run flow on unit tests with "warnings as errors" to generate test failures).

jandrade:
That could be a good solution, and that could be added on a different PR later :).

PR #463, opened 2019-08-01
[docs:core] WB-394: Add docs for addStyle

jandrade:
## Summary
Adds documentation and examples to the `addStyle` util.

### Core
- Added docs and examples to `addStyle`.

### Internal
- Fixed `gen-snapshot-tests.js` script to allow adding flow typed functions to the examples.
e.g. `export default function addStyle<T: React.AbstractComponent<*> | string>(`

### Test plan
1. Open https://deploy-preview-463--wonder-blocks.netlify.com/#!/AddStyle
2. Check the documentation and examples are clear and concise.

core-addStyle

### DIFF
--- a/packages/wonder-blocks-core/util/add-style.md
+++ b/packages/wonder-blocks-core/util/add-style.md
@@ -14,10 +14,10 @@ addStyle(
 ): React.Element;
 ```
 
-The `addStyle` function is a HOC returns a **React Component** or a **DOM**
+The `addStyle` function is a HOC that accepts a **React Component** or a **DOM**
 **intrinsic** ("div", "span", etc.) as its first argument and optional default
-styles as its second argument. This HOC returns a React Element with a `style`
-prop included ready to be rendered.
+styles as its second argument. This HOC returns a **React Component** with a
+`style` prop included ready to be rendered.
 
 #### Function arguments
 
@@ -39,6 +39,9 @@ export type StyleType =
     | Falsy
     | NestedArray<CSSProperties | Falsy>;
 ```
+
+**Note:** `StyleType` can contain a combination of style rules from an Aphrodite
+StyleSheet as well inline style objects (see example 4).
 
 #### CSSProperties
 
@@ -220,50 +223,28 @@ class CustomComponent extends React.Component {
 const styles = StyleSheet.create({
     // default styles
     default: {
-        background: Color.lightBlue,
+        background: Color.white,
         color: Color.white,
-        padding: Spacing.xxLarge,
+        padding: Spacing.medium,
     },
     // style to be passed as a prop
     customStyle: {
-        background: Color.darkBlue,
+        background: Color.lightBlue,
     },
 });
 
-class UsingStyleExample extends React.Component {
-    constructor() {
-        super();
-        this.state = {
-            showCustomStyle: false,
-        };
-        this.handleChange = this.handleChange.bind(this);
-    }
+<CustomComponent
+    style={[
+        // you can pass style rules from an Aphrodite StyleSheet
+        styles.customStyle,
+        // and pass inline styles as well
+        {
+            border: `1px solid ${Color.darkBlue}`,
+            padding: Spacing.xxLarge,
+        }
+    ]}
+/>
 
-    handleChange(checked) {
-        this.setState({
-            showCustomStyle: checked,
-        });
-    }
-
-    render() {
-        return (
-            <View>
-                <Checkbox
-                    label="Click here to toggle the custom style inside the custom component"
-                    checked={this.state.showCustomStyle}
-                    onChange={this.handleChange}
-                    style={{
-                        marginBottom: Spacing.medium,
-                    }}
-                />
-                <CustomComponent
-                    style={this.state.showCustomStyle && styles.customStyle}
-                />
-            </View>
-        );
-    }
-}
-<UsingStyleExample />
 ```
 
 **Warning:** In the case of React components from other packages, they may not
--- a/packages/wonder-blocks-core/util/add-style.md
+++ b/packages/wonder-blocks-core/util/add-style.md
@@ -1,0 +1,130 @@
+Adds a `style` property to components. A component can be either a custom React
+Component (e.g. Link, Button) or a JSX intrinsic (e.g. "div", "span").
+This helps you to manipulate and extend styles more easily by using the `styles`
+prop as an array of different style objects.
+
+### Example: Adding default styles
+
+You can create a new styled component by using the `addStyle` function. Note
+here that you can also define default styles for this component by passing an
+initial style object to this function.
+
+```jsx
+import {StyleSheet} from "aphrodite";
+import Color from "@khanacademy/wonder-blocks-color";
+import Spacing from "@khanacademy/wonder-blocks-spacing";
+import {addStyle} from "@khanacademy/wonder-blocks-core";
+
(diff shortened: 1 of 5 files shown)


packages/wonder-blocks-core/util/add-style.md
```diff
@@ -0,0 +1,130 @@
+Adds a `style` property to components. A component can be either a custom React
```
jeresig: Maybe "...to a component."?

packages/wonder-blocks-core/util/add-style.md
```diff
@@ -0,0 +1,130 @@
...
+        border: `1px solid ${Color.offBlack16}`,
+        borderRadius: Spacing.xxxSmall,
+        fontSize: Spacing.medium,
+        padding: Spacing.xSmall,
+    }
+});
+
+const StyledInput = addStyle("input", styles.input);
+
+<StyledInput type="text" placeholder="Lorem ipsum"/>;
```
jeresig: Haha, I guess I didn't realize that we had, essentially, replicated the `styled-components` API, that's pretty great!

jeresig:

Built with commit 0810272bfe7e94a4bc38144a1277c7f614612c9d

https://deploy-preview-463--wonder-blocks.netlify.com

utils/gen-snapshot-tests.js
```diff
@@ -321,20 +321,30 @@ function extractExamplesAndComponentsForFile(sourceFile, componentFileMap) {
     componentFileMap = componentFileMap || {};
 
     const src = fs.readFileSync(sourceFile, "utf8");
-    const match =
-        src.match(/export default class \s*(\w+)/) ||
-        src.match(/export default \s*\w+\(\s*(\w+)/);
+
+    const ast = parse(src, options);
```
jeresig: Yayy - thank you for changing this to a proper ast parser!

packages/wonder-blocks-core/util/add-style.md
```diff
@@ -0,0 +1,130 @@
+Adds a `style` property to components. A component can be either a custom React
+Component (e.g. Link, Button) or a JSX intrinsic (e.g. "div", "span").
+This helps you to manipulate and extend styles more easily by using the `styles`
+prop as an array of different style objects.
+
```
jeresig: Since we don't have a proper types inside Styleguidist it might be worth writing up exactly what the `style` property can take (for example `style={object|false|null|undefined|array<object|...>}`. I dunno, something that makes it clear what all the types of valid inputs are?

jandrade: Thanks for the suggestion!. I made some changes on that part and borrowed some ideas from @somewhatabstract's Timing documentation :)
https://deploy-preview-463--wonder-blocks.netlify.com/#!/AddStyle

packages/wonder-blocks-core/util/add-style.md
```diff
@@ -0,0 +1,167 @@
+Adds a `style` property to a component. A component can be either a custom React
+Component (e.g. Link, Button) or a DOM intrinsic (e.g. "div", "span"). This
```
kevinbarabash: Given that we mention costume React Components here, I think it's worth highlighting when `addStyle` should not be used.  In particular, if we've defined a component ourselves, it's preferable to simply add a `style: StyleType` prop to the `Props` definition ourselves and avoid the `addStyle` property.  Also, in the case of React components from other packages, they may not handle `className` or `style` in which case wrapping them in `addStyle` won't do anything.

kevinbarabash: Can you add an example of adding `style: StyleType` to the `Props` of a custom component and then show how we can pass the incoming `style` into an array of other styles we pass to a child component?

packages/wonder-blocks-core/util/add-style.md
kevinbarabash: I assume this comment is tongue-in-cheek since the style-components is quite different: https://www.styled-components.com/.

packages/wonder-blocks-core/util/add-style.md
```diff
@@ -0,0 +1,167 @@
...
+```js static
+addStyle(
+    Component: React.Element | "string",
+    defaultStyle?: StyleType
+): React.Element;
+```
+
+The `addStyle` function is a HOC that accepts a **React Component** or a **DOM**
+**intrinsic** ("div", "span", etc.) as its first argument and optional default
+styles as its second argument. This HOC returns a React Element with a `style`
```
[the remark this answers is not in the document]
kevinbarabash: Line 16 was correct `The addStyle function is an HOC that accepts a **React Component** ...`.  The issue was line 18.  The HOC does not return a "React Element", it returns a "React Component".

packages/wonder-blocks-core/util/add-style.md
```diff
@@ -164,4 +172,100 @@ const styles = StyleSheet.create({
...
+        color: Color.white,
+        padding: Spacing.xxLarge,
+    },
+    // style to be passed as a prop
+    customStyle: {
+        background: Color.darkBlue,
+    },
+});
+
+class UsingStyleExample extends React.Component {
```
kevinbarabash: This example feels overly complex.  I'd drop `UsingStyleExample` and just do:
```
<CustomComponent style={styles.customStyle} />
```

packages/wonder-blocks-core/util/add-style.md
```diff
@@ -30,7 +31,13 @@ prop included ready to be rendered.
 #### StyleType
```
kevinbarabash: One important detail that's missing from this is that `StyleType` can contain a combination of style rules from an Aphrodite StyleSheet as well inline style objects.  It think it'd be good to include the later in one of the examples.

packages/wonder-blocks-core/util/add-style.md
```diff
@@ -164,4 +172,100 @@ const styles = StyleSheet.create({
 });
 
 <DynamicStyles />
-```
\ No newline at end of file
+```
+
+#### 4. Using the `style` prop
```
kevinbarabash: Thanks for adding this example.

packages/wonder-blocks-core/util/add-style.md
```diff
@@ -164,4 +172,100 @@ const styles = StyleSheet.create({
...
+            </View>
+        );
+    }
+}
+<UsingStyleExample />
+```
+
+**Warning:** In the case of React components from other packages, they may not
+handle `className` or `style` in which case wrapping them in `addStyle` won't do
+anything.
```
kevinbarabash: Thanks for adding this caveat.

packages/wonder-blocks-core/util/add-style.md
jandrade: I initially created the example that way, but as I also wanted to illustrate how the styles are injected in runtime, I decided to add that extra part. If you still feel it's too complex, I definitively can drop that extra component.

packages/wonder-blocks-core/util/add-style.md
jandrade: That's an excellent point, I'll add that.

packages/wonder-blocks-core/util/add-style.md
jandrade: re🤔 about this, you're right, it feels more complex than it needs to be. I'm going to drop that extra component.

PR #464, opened 2019-08-07
Fix type issue with withActionScheduler

somewhatabstract:
This fixes an issue with the type expected by `withActionScheduler` and also adds some flow-only test cases that setup the expectations we have around how the types should work.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -11,3 +11,17 @@
 
 [options]
 server.max_workers=1
+
+# $FlowFixMe
+# This suppression hould be used to suppress an issue that requires additional
+# effort to be resolved. That effort should be recorded in our issue tracking
+# system for follow-up. Usually, this suppression is added by tooling during
+# flow upgrades and should not be used directly by developers. If it is
+# necessary, make sure to raise the corresponding issue and add the issue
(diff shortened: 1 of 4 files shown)


jeresig:

Built with commit 1d6bb5061f513f23b7d68f81453fe2c3d8bf4e09

https://deploy-preview-464--wonder-blocks.netlify.com

.flowconfig
```diff
@@ -10,4 +10,18 @@
...
+# $FlowFixMe
+# This suppression hould be used to suppress an issue that requires additional
+# effort to be resolved. That effort should be recorded in our issue tracking
+# system for follow-up. Usually, this suppression is added by tooling during
+# flow upgrades and should not be used directly by developers. If it is
+# necessary, make sure to raise the corresponding issue and add the issue
+# number to a TODO comment associated with the suppression.
+suppress_comment=\\(.\\|\n\\)*\\$FlowFixMe
+
+# $FlowExpectError
```
jandrade: TIL: This is a good way to solve that

somewhatabstract: Yeah, if we also run flow where warnings are errors, then this will work even better. I shied away from that just yet, but I think it's not a bad idea.

PR #466, opened 2019-08-14
WB-641: Add flow type support for falsy elements in Dropdowns

kumailn:
Added flow type support for falsy elements in Dropdowns, SingleSelects and MultiSelects. 

Flow will now recognize children such as `{false && <OptionItem />}`

### DIFF
--- a/packages/wonder-blocks-dropdown/components/dropdown.js
+++ b/packages/wonder-blocks-dropdown/components/dropdown.js
@@ -40,7 +40,7 @@ type Props = {|
     /**
      * The items present in the Dropdown
      */
-    menuItems: Array<Item> | Item,
+    menuItems: Array<Item | false | typeof undefined | null> | Item,
 
     /**
      * Closes the Dropdown when an OptionItem is selected, use this
--- a/packages/wonder-blocks-dropdown/components/dropdown.js
+++ b/packages/wonder-blocks-dropdown/components/dropdown.js
@@ -40,7 +40,7 @@ type Props = {|
     /**
      * The items present in the Dropdown
      */
-    menuItems: Array<Item | false | typeof undefined | null> | Item,
+    menuItems: Array<?(Item | false)> | Item,
 
     /**
      * Closes the Dropdown when an OptionItem is selected, use this
--- a/packages/wonder-blocks-dropdown/components/multi-select.js
+++ b/packages/wonder-blocks-dropdown/components/multi-select.js
@@ -20,9 +20,7 @@ type Props = {|
     /**
      * The items in this select.
      */
-    children?: Array<
-        React.Element<OptionItem> | false | typeof undefined | null,
-    >,
+    children?: Array<?(React.Element<OptionItem> | false)>,
 
     /**
      * Whether this component is disabled. A disabled dropdown may not be opened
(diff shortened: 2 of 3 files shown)


jeresig:

Built with commit f576865965a1104ce21b00c8c84c193c7dfe601c

https://deploy-preview-466--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/dropdown.js
```diff
@@ -40,7 +40,7 @@ type Props = {|
     /**
      * The items present in the Dropdown
      */
-    menuItems: Array<Item> | Item,
+    menuItems: Array<Item | false | typeof undefined | null> | Item,
```
somewhatabstract: I think this can just be `Array<?(Item | false)> | Item`.

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -20,7 +20,9 @@ type Props = {|
     /**
      * The items in this select.
      */
-    children?: Array<React.Element<OptionItem>>,
+    children?: Array<
+        React.Element<OptionItem> | false | typeof undefined | null,
```
somewhatabstract: `Array<?(React.Element<OptionItem> | false)>`?

packages/wonder-blocks-dropdown/components/single-select.js
```diff
@@ -18,7 +18,9 @@ type Props = {|
     /**
      * The items in this select.
      */
-    children?: Array<React.Element<OptionItem>>,
+    children?: Array<
```
somewhatabstract: Array<?(React.Element<OptionItem> | false)>?

somewhatabstract:
Is there an request/Jira task driving this change?

kumailn:
> Is there an request/Jira task driving this change?

Yes, it's WB344

packages/wonder-blocks-dropdown/components/dropdown.js
kumailn: I spoke to at @kevinbarabash about this and he recommended including the null type too

somewhatabstract: THe `?` modify includes allowing `null`.

kumailn: oh didn't catch the `?` there, yup that looks good to me! Thanks for the suggestion, I'll update the PR

kevinbarabash: `Array<?Item | false>` should also work.  Also, if you ever need `typeof undefined` you should use `void` instead.

packages/wonder-blocks-dropdown/components/multi-select.js
kevinbarabash: Shouldn't there be a `typeof` in there, e.g.
```
Array<?React.Element<typeof OptionItem> | false>
```

kevinbarabash:
https://khanacademy.atlassian.net/browse/WB-344 is the Incoming Issues epic.  There should be a ticket inside of it though for this task.

packages/wonder-blocks-dropdown/components/multi-select.js
kevinbarabash: My bad, I did realize that `OptionItem` was using a `typeof` import at the top of the file.

kumailn:
> The ticket number in the title needs to be correct before merging this so that the Jira-GitHub integration we're using doesn't mark the epic as done.

Fixed, thanks for catching that :)

PR #468, opened 2019-08-16
lint against deprecated flow types

kevinbarabash:
I thought I'd try out some of the lint rules that flow has to offer.  This commit originally also included sketchy null checks as well but it turns out those are easier to fix if you have null coalescing which was easy to add but fixing them was still a bit of a pain so I scaled this back to just fix up deprecated types, namely `*`, `$Supertype`, and `$Subtype`.  I also decided to add flow types for enzyme which brought out batch coverage up from about 85% to 95%.

### DIFF
--- a/packages/wonder-blocks-core/util/add-style.js
+++ b/packages/wonder-blocks-core/util/add-style.js
@@ -7,7 +7,7 @@ import {processStyleList} from "./util.js";
 import type {StyleType} from "./types.js";
 
 // TODO(kevinb): have an a version which uses exact object types
-export default function addStyle<T: React.AbstractComponent<*> | string>(
+export default function addStyle<T: React.AbstractComponent<any> | string>(
     Component: T,
     defaultStyle?: StyleType,
 ): React.AbstractComponent<React.ElementConfig<T> & {style: StyleType}> {
--- a/packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.test.js
+++ b/packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.test.js
@@ -30,10 +30,12 @@ describe("maybeGetPortalMountedModalHostElement", () => {
             </div>
         );
         const wrapper = mount(nodes);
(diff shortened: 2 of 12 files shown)


jeresig:

Built with commit 17b927f1b9b9176849dcd428706322347734be4c

https://deploy-preview-468--wonder-blocks.netlify.com

packages/wonder-blocks-core/util/add-style.js
```diff
@@ -7,7 +7,7 @@ import {processStyleList} from "./util.js";
 import type {StyleType} from "./types.js";
 
 // TODO(kevinb): have an a version which uses exact object types
-export default function addStyle<T: React.AbstractComponent<*> | string>(
+export default function addStyle<T: React.AbstractComponent<any> | string>(
```
jandrade: Is it strictly necessary to use `any` here? what would happen if you use `React.Node` instead?

packages/wonder-blocks-modal/util/maybe-get-portal-mounted-modal-host-element.test.js
```diff
@@ -30,10 +30,12 @@ describe("maybeGetPortalMountedModalHostElement", () => {
             </div>
         );
         const wrapper = mount(nodes);
-        const candidateElement = wrapper.find("button")[0];
+        const candidateElement = wrapper.find("button").at(0);
```
jandrade: 🎉 this is way better!

packages/wonder-blocks-core/util/add-style.js
kevinbarabash: `React.Node` isn't a component.

kevinbarabash: I wonder if this should be `mixed` and not `any`.

kevinbarabash: That didn't work.  I need to figure out how to make this be a subtype of `{}`.  In the meantime though `any` isn't any worse than `*`.

PR #470, opened 2019-08-16
Experiment: try out optional chaining

kevinbarabash:
I wanted to try it out to see how it works and how much work it would be to adopt in a webapp.  The `unnecessary-optional-chain` flow lint was helpful.  It's too bad flow doesn't support method calls with optional chaining yet.  There's a couple of places in wonder-blocks where that would be helpful.

### DIFF
--- a/packages/wonder-blocks-dropdown/components/action-item.js
+++ b/packages/wonder-blocks-dropdown/components/action-item.js
@@ -96,7 +96,7 @@ export default class ActionItem extends React.Component<ActionProps> {
     static contextTypes = {router: PropTypes.any};
     static __IS_ACTION_ITEM__ = true;
     static isClassOf(instance: React.Element<any>) {
-        return instance && instance.type && instance.type.__IS_ACTION_ITEM__;
+        return instance.type?.__IS_ACTION_ITEM__;
     }
 
     render() {
--- a/packages/wonder-blocks-modal/components/modal-content.js
+++ b/packages/wonder-blocks-modal/components/modal-content.js
@@ -26,7 +26,7 @@ export default class ModalContent extends React.Component<Props> {
 
     static __IS_MODAL_CONTENT__ = true;
     static isClassOf(instance: any) {
-        return instance && instance.type && instance.type.__IS_MODAL_CONTENT__;
+        return instance?.type?.__IS_MODAL_CONTENT__;
     }
 
     render() {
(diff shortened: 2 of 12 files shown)


jeresig:

Built with commit eee70883d391f4d8d2f8197b555841e4317cc141

https://deploy-preview-470--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/action-item.js
```diff
@@ -96,7 +96,7 @@ export default class ActionItem extends React.Component<ActionProps> {
     static contextTypes = {router: PropTypes.any};
     static __IS_ACTION_ITEM__ = true;
     static isClassOf(instance: React.Element<any>) {
-        return instance && instance.type && instance.type.__IS_ACTION_ITEM__;
```
kevinbarabash: I turns out the first `&&` wasn't necessary.

packages/wonder-blocks-modal/components/modal-content.js
```diff
@@ -26,7 +26,7 @@ export default class ModalContent extends React.Component<Props> {
 
     static __IS_MODAL_CONTENT__ = true;
     static isClassOf(instance: any) {
-        return instance && instance.type && instance.type.__IS_MODAL_CONTENT__;
+        return instance?.type?.__IS_MODAL_CONTENT__;
```
jandrade: What would be the reason to need the first `?` here?

kevinbarabash: Because `instance` should be typed as `React.Element<any>` like the other `isClassOf` functions.  Nice catch.  I'll make that change.

kevinbarabash: I can't use `React.Element<any>` here because another part of the code tries to pass a `React.Node` here which can be `null`.

kevinbarabash: I'll add a comment.

kevinbarabash:
Even though this is an "experiment", I'm going to merge this.  There are a couple of limitations to flow's current implementation of optional chaining: doesn't work with method calls, doesn't cause type refinement.  If either of these issue turn out to be too much of a headache we can revert this diff.

kevinbarabash:
I guess I can't merge this yet. 😅

kevinbarabash:
I'll leave this up as an experiment and write an ADR first. :)

PR #471, opened 2019-08-19
WB-98: docs(): Update README.md

jandrade:
## Summary

Updated README.md to show a `getting started` section that includes short installation instructions and adds the URL to see the docs locally. Also, added a link to the confluence page for a more detailed explanation on the process.

Output: https://github.com/Khan/wonder-blocks/blob/internal/readme-docs/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Installs project dependencies
 
 Runs the docs in dev mode
 
-Now you can open `http://localhost:6060` to view the docs. This page will
+Now you can open http://localhost:6060 to view the docs. This page will
 automatically update as you make changes to components.
 
 ## Contributing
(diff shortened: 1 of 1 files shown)


jeresig:

Built with commit 640ba6ff76445859c06fe4fd676c04ad2bf8a489

https://deploy-preview-471--wonder-blocks.netlify.com

README.md
```diff
@@ -2,10 +2,42 @@
...
+
+#### `yarn install`
+
+Installs project dependencies
+
+#### `yarn start`
+
+Runs the docs in dev mode
+
+Now you can open http://localhost:6060 to view the docs. This page will
```
somewhatabstract: Yes! Thank you!

README.md
```diff
@@ -2,10 +2,42 @@
...
+automatically update as you make changes to components.
+
 ## Contributing
```
somewhatabstract: I think we should add an explicit statement about accepting external contributions that makes it clear if we do or do not, and if we do, under what circumstances (i.e. what types of changes).

jandrade: That's a great suggestion, I'll add that shortly.

PR #472, opened 2019-08-19
Create .flowconfig-ci with server.max_workers=1 for CI operation only

kevinbarabash:
Flow is slow running locally because it was only using a single worker.  This fixes that.

The reason why we only use one worker when running on CI to avoid situations where flow has trouble launching.  This can happen on CI because the free version of CircleCI using single core machines.  See https://circleci.com/gh/Khan/wonder-blocks/2076?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link for an example of the error.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ jobs:
           paths:
             - node_modules
       - run: yarn run lint
-      - run: yarn run flow
+      - run: yarn run flow-ci
   test:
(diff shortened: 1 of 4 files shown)


jeresig:

Built with commit 22cd4c716de05243f590ab5b50e194fd84a00014

https://deploy-preview-472--wonder-blocks.netlify.com

PR #478, opened 2019-09-08
Use flow to enforce invalid props in ModalLauncher

kevinbarabash:
This change was motivated by https://phabricator.khanacademy.org/D57666.  We can use flow to enforce the prop constraints we were previously throw exceptions for.  This change though won't fix the issue fixed in D57666 so we should look at updating all uses of -v1 modals to -v2 modals.

While making this change I found myself a little confused about the various usages of ModalLauncher.  I think we should consider splitting ModalLauncher into two components:
- one that requires children and is always uncontrolled
- one that doesn't use children and is always controlled

I think that should cover all of our use cases and be less confusing both in terms of implementation but also in terms of use.

### DIFF
--- a/packages/wonder-blocks-modal/components/modal-launcher.js
+++ b/packages/wonder-blocks-modal/components/modal-launcher.js
@@ -112,6 +112,18 @@ export default class ModalLauncher extends React.Component<Props, State> {
     };
 
     static getDerivedStateFromProps(props: Props, state: State) {
+        if (typeof props.opened === "boolean" && props.children) {
+            // eslint-disable-next-line no-console
+            console.warn("'children' and 'opened' can't be used together");
+        }
+        if (typeof props.opened === "boolean" && !props.onClose) {
+            // eslint-disable-next-line no-console
+            console.warn("'onClose' should be used with 'opened'");
+        }
+        if (typeof props.opened !== "boolean" && !props.children) {
+            // eslint-disable-next-line no-console
+            console.warn("either 'children' or 'opened' must be set");
+        }
         return {
             opened:
                 typeof props.opened === "boolean" ? props.opened : state.opened,
--- a/packages/wonder-blocks-modal/components/modal-launcher.js
+++ b/packages/wonder-blocks-modal/components/modal-launcher.js
@@ -112,9 +112,6 @@ export default class ModalLauncher extends React.Component<Props, State> {
     };
 
     static getDerivedStateFromProps(props: Props, state: State) {
-        if (typeof props.opened === "boolean" && !props.onClose) {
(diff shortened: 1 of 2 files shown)


jeresig:

Built with commit 67c9d1bffc755d15746c9fd1ebb3f1963cf72f3a

https://deploy-preview-478--wonder-blocks.netlify.com

kevinbarabash:
The downside of this change (and other components that use disjoint unions for their props) is that styleguidist (really react-docgen) doesn't understand them:

**BEFORE**
Screen Shot 2019-09-08 at 2 08 59 PM

**AFTER**
Screen Shot 2019-09-08 at 2 10 10 PM

packages/wonder-blocks-modal/components/modal-launcher.js
```diff
@@ -70,6 +49,36 @@ type Props = {|
...
+     * Using this prop makes the component behave as a controlled component.
+     * The parent is responsible for managing the opening/closing of the modal
+     * when using this prop.  `onClose` should always be used and `children`
+     * should never be used with this prop.  Not doing so will result in an
+     * error being thrown.
+     */
+    opened: boolean,
+
+    /**
+     * Called when the modal needs notifies the parent component that it should
```
jandrade: super NIT: `needs to notify`

kevinbarabash:
> This is something else we would need to really check if we would want to migrate to a different tool in the future.

I'm becoming more and more convinced that writing our documentation tool makes sense especially if make it open source for the start.  I think splitting it into two tools: one that generates a JSON blob containing all of the data and one that renders that JSON blob.  That way people can make other frontends.

packages/wonder-blocks-modal/components/modal-launcher.js
```diff
@@ -103,15 +112,6 @@ export default class ModalLauncher extends React.Component<Props, State> {
...
     static getDerivedStateFromProps(props: Props, state: State) {
-        if (typeof props.opened === "boolean" && props.children) {
-            throw new Error("'children' and 'opened' can't be used together");
-        }
-        if (typeof props.opened === "boolean" && !props.onClose) {
-            throw new Error("'onClose' should be used with 'opened'");
-        }
-        if (typeof props.opened !== "boolean" && !props.children) {
-            throw new Error("either 'children' or 'opened' must be set");
-        }
```
somewhatabstract: Should we at least log a warning? Flow doesn't preclude having runtime checks for bad data.

kevinbarabash: I've added warnings and tests for the warnings.  This is probably good practice for any constraints like this that depend on flow.

PR #479, opened 2019-09-10
WB-642: MultiSelect with search filter

jangmi-jo:
Mockup link: https://www.figma.com/file/4l5Y4zG14xaj2d8wqjCCA8/Admin-Reports-(1.3-Release)?node-id=5234%3A1850&viewport=1160%2C306%2C0.6481390595436096

**Context**
We want to have a search field text input within the MultiSelect
component's dropdown body that allows filtering the option items by the
"label" of the options.

**This PR has the following changes**
In MultiSelect:
- Fix a bug related to the item count in MultiSelect (when the children has `false`, it was counted as an item and led incorrect menu text, selected count) 
- Add `searchText` in state to filter the option items by
- Add `lastSelectedValues` in state to obtain the selectedValues when the dropdown is opened to maintain the order of previously selected items at the top.
- Add an optional `isFilterable` in prop
- If `isFilterable` is true
  - Filter (case insensitive) the option items by searchText
  - Put the "last" selected items at the top (add a separator item) 
  - Hide shortcuts when search text input has been entered
  - Pass the searchText and searchText change handler to DropdownCore
  - Apply filterable dropdown style (min/max height) 

In DropdownCore:
- Add `onSearchTextChanged` and `searchText` as an optional prop
- When `onSearchTextChanged` is provided,
  - Display a SearchTextInput and add it into itemRefs to adjust focus.
  - Shift the focus index by 1 in all itemRefs to adjust focus for the rest of items
  - Display "No results" when there's no items
  - Enable space key up/down when the SearchTextInput is focused.
  - Enable tab key to move focus to dismiss button when the SearchTextInput is focused and text is entered.

SearchTextInput:
- Takes required props searchText, onChange, onClick, itemRef
- Takes optional style prop
- Displays a text input with search icon on its left side
- When there's text in the input, displays dismiss icon on its right side

Test plan:
Visit http://localhost:6060/#multiselect
Open the selector for "Multi select with search filter section"
Try entering some text to filter the results.

### DIFF
--- a/packages/wonder-blocks-dropdown/__snapshots__/generated-snapshot.test.js.snap
+++ b/packages/wonder-blocks-dropdown/__snapshots__/generated-snapshot.test.js.snap
@@ -4048,3 +4048,139 @@ exports[`wonder-blocks-dropdown example 26 1`] = `
   </button>
 </div>
 `;
+
+exports[`wonder-blocks-dropdown example 27 1`] = `
+<div
+  className=""
+  style={
+    Object {
+      "alignItems": "stretch",
+      "borderStyle": "solid",
+      "borderWidth": 0,
+      "boxSizing": "border-box",
+      "display": "flex",
+      "flexDirection": "row",
+      "margin": 0,
+      "minHeight": 0,
+      "minWidth": 0,
+      "padding": 0,
+      "position": "relative",
+      "zIndex": 0,
+    }
+  }
+>
+  <div
+    className=""
+    onKeyDown={[Function]}
+    onKeyUp={[Function]}
+    style={
+      Object {
+        "alignItems": "stretch",
+        "borderStyle": "solid",
+        "borderWidth": 0,
+        "boxSizing": "border-box",
+        "display": "flex",
+        "flexDirection": "column",
+        "margin": 0,
+        "minHeight": 0,
+        "minWidth": 0,
+        "padding": 0,
+        "position": "relative",
+        "width": "fit-content",
+        "zIndex": 0,
+      }
+    }
+  >
+    <button
+      aria-expanded="false"
+      aria-haspopup="listbox"
+      className=""
+      disabled={false}
+      onBlur={[Function]}
+      onClick={[Function]}
+      onDragStart={[Function]}
+      onFocus={[Function]}
+      onKeyDown={[Function]}
+      onKeyUp={[Function]}
+      onMouseDown={[Function]}
+      onMouseEnter={[Function]}
+      onMouseLeave={[Function]}
+      onMouseUp={[Function]}
+      onTouchCancel={[Function]}
+      onTouchEnd={[Function]}
+      onTouchStart={[Function]}
+      style={
+        Object {
+          "::MozFocusInner": Object {
+            "border": 0,
+          },
+          "alignItems": "center",
+          "backgroundColor": "#ffffff",
+          "borderColor": "rgba(33,36,44,0.16)",
+          "borderRadius": 4,
+          "borderStyle": "solid",
+          "borderWidth": 1,
+          "boxSizing": "border-box",
+          "color": "#21242c",
+          "display": "inline-flex",
+          "height": 40,
+          "justifyContent": "space-between",
+          "margin": 0,
+          "outline": "none",
+          "paddingLeft": 16,
+          "paddingRight": 12,
+          "position": "relative",
+          "textDecoration": "none",
+          "touchAction": "manipulation",
+          "whiteSpace": "nowrap",
+        }
+      }
+      tabIndex={0}
+      type="button"
+    >
+      <span
+        className=""
+        style={
+          Object {
+            "MozOsxFontSmoothing": "grayscale",
+            "WebkitFontSmoothing": "antialiased",
+            "display": "block",
+            "fontFamily": "Lato, \\"Noto Sans\\", sans-serif",
+            "fontSize": 16,
+            "fontWeight": 400,
+            "lineHeight": "20px",
+            "marginRight": 8,
+            "overflow": "hidden",
+            "textOverflow": "ellipsis",
+            "userSelect": "none",
+            "whiteSpace": "nowrap",
+          }
+        }
+      >
+        0 schools
+      </span>
+      <svg
+        aria-hidden="true"
+        className=""
+        height={16}
+        style={
+          Object {
+            "display": "inline-block",
+            "flexGrow": 0,
+            "flexShrink": 0,
+            "minWidth": 16,
+            "verticalAlign": "text-bottom",
+          }
+        }
+        viewBox="0 0 16 16"
+        width={16}
+      >
+        <path
+          d="M8 8.586l3.293-3.293a1 1 0 0 1 1.414 1.414l-4 4a1 1 0 0 1-1.414 0l-4-4a1 1 0 0 1 1.414-1.414L8 8.586z"
+          fill="rgba(33,36,44,0.64)"
+        />
+      </svg>
+    </button>
+  </div>
+</div>
+`;
--- a/packages/wonder-blocks-dropdown/components/dropdown-core.js
+++ b/packages/wonder-blocks-dropdown/components/dropdown-core.js
@@ -31,17 +31,17 @@ type DropdownProps = {|
 
     /**
      * An optional handler to set the searchText of the parent. When this and
-     * the searchText are provided, SearchTextInput will be displayed at the
-     * top of the dropdown body.
+     * the searchText exist, SearchTextInput will be displayed at the top of
+     * the dropdown body.
      */
-    onSearchTextChanged?: (searchText: string) => mixed,
+    onSearchTextChanged?: ?(searchText: string) => mixed,
 
     /**
      * An optional string that the user entered to search the items. When this
-     * and the onSearchTextChanged are provided, SearchTextInput will be
-     * displayed at the top of the dropdown body.
+     * and the onSearchTextChanged exist, SearchTextInput will be displayed at
+     * the top of the dropdown body.
      */
-    searchText?: string,
+    searchText?: ?string,
 
     /**
      * An index that represents the index of the focused element when the menu
@@ -522,15 +522,23 @@ export default class DropdownCore extends React.Component<
         }
     }
 
+    maybeRenderNoResults() {
+        const {items} = this.props;
+        // TODO(jangmi): Use translated string for "No results"
+        if (items.length === 0) {
+            return (
+                <LabelMedium style={styles.noResult}>No results</LabelMedium>
+            );
+        }
+    }
+
     maybeRenderSearchTextInput() {
-        const {items, onSearchTextChanged, searchText} = this.props;
+        const {onSearchTextChanged, searchText} = this.props;
 
         if (!onSearchTextChanged || typeof searchText !== "string") {
             return null;
         }
 
-        const noResult = onSearchTextChanged && items.length === 0;
-        // TODO(jangmi): Use translated string for "No results"
         return (
             <React.Fragment>
                 <SearchTextInput
@@ -543,11 +551,7 @@ export default class DropdownCore extends React.Component<
                     searchText={searchText}
                     style={styles.searchInput}
                 />
-                {noResult && (
-                    <LabelMedium style={styles.noResult}>
-                        No results
-                    </LabelMedium>
-                )}
+                {this.maybeRenderNoResults()}
             </React.Fragment>
         );
     }
--- a/packages/wonder-blocks-dropdown/components/dropdown-core.js
+++ b/packages/wonder-blocks-dropdown/components/dropdown-core.js
@@ -30,11 +30,17 @@ type DropdownProps = {|
     items: Array<DropdownItem>,
 
     /**
-     * An optional handler to set the searchText of the parent. When this is
-     * provided, SearchTextInput will be displayed at the top of the dropdown
-     * body.
+     * An optional handler to set the searchText of the parent. When this and
+     * the searchText are provided, SearchTextInput will be displayed at the
+     * top of the dropdown body.
      */
     onSearchTextChanged?: (searchText: string) => mixed,
+
+    /**
+     * An optional string that the user entered to search the items. When this
+     * and the onSearchTextChanged are provided, SearchTextInput will be
+     * displayed at the top of the dropdown body.
+     */
     searchText?: string,
 
     /**
@@ -422,7 +428,7 @@ export default class DropdownCore extends React.Component<
                 if (
                     showSearchTextInput &&
                     this.focusedIndex === 0 &&
-                    this.props.searchText
+                    searchText
                 ) {
                     return;
                 }
@@ -524,7 +530,7 @@ export default class DropdownCore extends React.Component<
         }
 
         const noResult = onSearchTextChanged && items.length === 0;
-        // TODO(jangmi): Use translated strings for "Filter", "No results"
+        // TODO(jangmi): Use translated string for "No results"
         return (
             <React.Fragment>
                 <SearchTextInput
--- a/packages/wonder-blocks-dropdown/components/dropdown-core.js
+++ b/packages/wonder-blocks-dropdown/components/dropdown-core.js
@@ -692,6 +692,7 @@ const styles = StyleSheet.create({
     noResult: {
         color: Color.offBlack64,
         alignSelf: "center",
+        marginTop: Spacing.xxSmall,
     },
     searchInput: {
         margin: Spacing.xSmall,
--- a/packages/wonder-blocks-dropdown/components/dropdown-core.js
+++ b/packages/wonder-blocks-dropdown/components/dropdown-core.js
@@ -29,10 +29,6 @@ type DropdownProps = {|
     items: Array<DropdownItem>,
 
     /**
-     * The search text that filters the item labels by.
-     */
-    searchText: string,
-    /**
      * A handler to set the searchText of the parent. When this is provided,
      * a search text input will be displayed at the top of the dropdown body.
      */
@@ -149,7 +145,6 @@ export default class DropdownCore extends React.Component<
         alignment: "left",
         initialFocusedIndex: 0,
         light: false,
-        searchText: "",
         handleSearchTextChanged: null,
     };
 
@@ -505,9 +500,8 @@ export default class DropdownCore extends React.Component<
     }
 
     renderSearchTextInput() {
-        const {items, searchText, handleSearchTextChanged} = this.props;
-        const noResult =
-            handleSearchTextChanged && searchText && items.length === 0;
+        const {items, handleSearchTextChanged} = this.props;
+        const noResult = handleSearchTextChanged && items.length === 0;
         // TODO(jangmi): Use TextField once we have it.
         // https://khanacademy.atlassian.net/browse/WB-578
         // TODO(jangmi): Use translated strings for "Filter", "No results"
@@ -515,7 +509,6 @@ export default class DropdownCore extends React.Component<
             <React.Fragment>
                 <input
                     type="text"
-                    value={searchText}
                     onChange={handleSearchTextChanged}
                     onClick={() => this.handleClickFocus(0)}
                     ref={this.state.itemRefs[0].ref}
--- a/packages/wonder-blocks-dropdown/components/dropdown-core.js
+++ b/packages/wonder-blocks-dropdown/components/dropdown-core.js
@@ -28,6 +28,16 @@ type DropdownProps = {|
     items: Array<DropdownItem>,
 
     /**
+     * The search text that filters the item labels by.
+     */
+    searchText: string,
+    /**
+     * A handler to set the searchText of the parent. When this is provided,
+     * a search text input will be displayed at the top of the dropdown body.
+     */
+    handleSearchTextChanged: ?(e: SyntheticInputEvent<>) => void,
+
+    /**
      * An index that represents the index of the focused element when the menu
      * is opened.
      */
@@ -138,6 +148,8 @@ export default class DropdownCore extends React.Component<
         alignment: "left",
         initialFocusedIndex: 0,
         light: false,
+        searchText: "",
+        handleSearchTextChanged: null,
     };
 
     // Figure out if the same items are focusable. If an item has been added or
@@ -166,10 +178,19 @@ export default class DropdownCore extends React.Component<
             !DropdownCore.sameItemsFocusable(state.prevItems, props.items)
         ) {
             const itemRefs = [];
+            // If the searchTextChange handler is provided, the first focusable
+            // item is the search text input element.
+            if (props.handleSearchTextChanged) {
+                const ref = React.createRef<null | HTMLDivElement>();
+                itemRefs.push({ref, originalIndex: 0});
+            }
+
+            const shiftItemIndexBy = props.handleSearchTextChanged ? 1 : 0;
+
             for (let i = 0; i < props.items.length; i++) {
                 if (props.items[i].focusable) {
                     const ref = React.createRef<null | HTMLDivElement>();
-                    itemRefs.push({ref, originalIndex: i});
+                    itemRefs.push({ref, originalIndex: i + shiftItemIndexBy});
                 }
             }
             return {
@@ -358,7 +379,12 @@ export default class DropdownCore extends React.Component<
     }
 
     handleKeyDown = (event: SyntheticKeyboardEvent<>) => {
-        const {initialFocusedIndex, onOpenChanged, open} = this.props;
+        const {
+            initialFocusedIndex,
+            onOpenChanged,
+            open,
+            handleSearchTextChanged: displaySearchTextInput,
+        } = this.props;
         const keyCode = event.which || event.keyCode;
 
         // If menu isn't open and user presses down, open the menu
@@ -393,6 +419,11 @@ export default class DropdownCore extends React.Component<
                 onOpenChanged(false, true);
                 return;
             case keyCodes.space:
+                // When we display search text input and the focus is on that
+                // input, we should let the user type space.
+                if (displaySearchTextInput && this.focusedIndex === 0) {
+                    return;
+                }
                 // Prevent space from scrolling down the page
                 event.preventDefault();
                 return;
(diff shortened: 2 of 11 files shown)


jeresig:

Built with commit 3de826b022f13ab598e50135ca0e9a2dae369b82

https://deploy-preview-479--wonder-blocks.netlify.com

jangmi-jo:
Hi team! I think the functionality is ready for review. The only two items left in TODO  are adding unit tests and using TextField. I'd like to add unit tests once I get some feedback (in case the approach I went is not ideal). Will look forward to hearing from you all! Thanks.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -463,8 +499,38 @@ export default class DropdownCore extends React.Component<
         }
     }
 
+    renderSearchTextInput() {
+        const {items, handleSearchTextChanged} = this.props;
+        const noResult = handleSearchTextChanged && items.length === 0;
+        // TODO(jangmi): Use TextField once we have it.
```
jandrade: As of today, the `TextField` component won't include icons or any other variants, so I'd recommend for now to abstract the contents of this function into a different component and style it based off the design spec you are using for reference.

We can revisit later when building the WB TextField component.

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -326,20 +367,23 @@ export default class MultiSelect extends React.Component<Props, State> {
...
             /* eslint-enable no-unused-vars */
             ...sharedProps
         } = this.props;
         const {open} = this.state;
 
-        const menuText = this.getMenuText();
-
-        const items = [...this.getShortcuts(), ...this.getMenuItems()];
+        const allChildren = React.Children.toArray(children).filter((child) =>
+            Boolean(child),
```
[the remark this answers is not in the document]
kevinbarabash: style nit: `.filter(Boolean)` also works if you do need to filter out falsey values.

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -76,6 +76,14 @@ type Props = {|
      */
     implicitAllEnabled?: boolean,
 
+    /**
+     * When this is true, the dropdown body shows a search text input at the
+     * top. The items will be filtered by the input.
+     * Selected items will be moved to the top and shortcuts won't be added
+     * even if shortcuts prop is set to true.
+     */
+    isFilterableByLabel?: boolean,
```
kevinbarabash: The `ByLabel` part feels excessive, maybe just `isFilterable`?

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -284,28 +296,57 @@ export default class MultiSelect extends React.Component<Props, State> {
...
+        });
+
+        if (isFilterableByLabel) {
+            filteredChildren.sort((a, b) => {
+                const aIsSelected = a.populatedProps.selected;
+                const bIsSelected = b.populatedProps.selected;
+                if (aIsSelected === bIsSelected) {
+                    return 0;
+                } else if (aIsSelected) {
+                    return -1;
```
kevinbarabash: Is this so that selected items appear a the top?

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -352,12 +396,22 @@ export default class MultiSelect extends React.Component<Props, State> {
...
                 alignment={alignment}
                 dropdownStyle={[selectDropdownStyle, dropdownStyle]}
-                items={items}
+                handleSearchTextChanged={
+                    isFilterableByLabel ? this.handleSearchTextChanged : null
+                }
+                items={
+                    isFilterableByLabel
+                        ? filteredMenuItems
+                        : [...shortcutItems, ...filteredMenuItems]
```
kevinbarabash: We could inline the call to `this.getShortcuts(numOptions)` here to avoid calling it when we don't need to.

packages/wonder-blocks-dropdown/components/multi-select.md
```diff
@@ -444,4 +444,60 @@ class ControlledMultiSelectExample extends React.Component {
...
+    handleChange(update) {
+        this.setState({
+           selectedValues: update,
+        });
+    }
+
+    render() {
+        return <MultiSelect
+            shortcuts={true}
+            isFilterableByLabel={true}
```
kevinbarabash: I thought that shortcuts aren't shown when `isFilterableByLabel` is true.  It would be nice to enforce this constraint using flow if possible.

packages/wonder-blocks-dropdown/components/multi-select.md
```diff
@@ -444,4 +444,60 @@ class ControlledMultiSelectExample extends React.Component {
...
+    constructor() {
+        super();
+        this.state = {
+            selectedValues: [],
+        };
+        // Styleguidist doesn't support arrow functions in class field properties
+        this.handleChange = this.handleChange.bind(this);
+    }
+
+    handleChange(update) {
```
kevinbarabash: nit: I think `handleChange(selectedValues) {` might be clearer.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -358,7 +375,12 @@ export default class DropdownCore extends React.Component<
     }
 
     handleKeyDown = (event: SyntheticKeyboardEvent<>) => {
-        const {initialFocusedIndex, onOpenChanged, open} = this.props;
+        const {
+            initialFocusedIndex,
+            onOpenChanged,
+            open,
+            handleSearchTextChanged: displaySearchTextInput,
```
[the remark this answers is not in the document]
kevinbarabash: style nit: maybe use `showFoo` instead of `displayFoo` since we use `showFoo` elsewhere in the code base.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -27,6 +28,12 @@ type DropdownProps = {|
      */
     items: Array<DropdownItem>,
 
+    /**
+     * A handler to set the searchText of the parent. When this is provided,
+     * a search text input will be displayed at the top of the dropdown body.
+     */
+    handleSearchTextChanged: ?(e: SyntheticInputEvent<>) => void,
```
kevinbarabash: This should be `onSearchTextChanged` or preferably `onSearchTextChanged?: (searchText: string) => mixed`.  I think we should avoid passing the raw event back.  We're always going to want to extract the text from the input value so DropdownCore may as well do that before passing it back to.

kevinbarabash: This got me thinking about whether we should change the `onChange: (selectedValues: Array<string>) => mixed,` in multi-select.js in the future to be `onSelectionChange` or `onSelectionChanged`. 🤔
No action necessary for this PR.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -490,6 +558,7 @@ export default class DropdownCore extends React.Component<
                     dropdownStyle,
                 ]}
             >
+                {displaySearchTextInput && this.renderSearchTextInput()}
```
kevinbarabash: Don't we want the filter to appear in the opener?

kevinbarabash: I had another look at the figma and you're right.  I posted a comment requesting mocks of the dropdown with filter search field and the opener.  I think that help give us a better idea of what this thing should look like when it's finished. 

I posted another comment ini the figma asking about scroll behavior within the dropdown.  Right now it looks like the search field will scroll out of the view as soon as someone starts scrolling down.  I'm not sure that's the behavior we want because then people are forced to scroll up to the top of the list if they want to filter the list.

kevinbarabash:
The new SearchTextField looks good.  If there's too many items it gets squished though.  I set the number of items in the example to 100 and it messed up the search field.

Screen Shot 2019-09-13 at 4 47 15 PM

kevinbarabash:
The hover state for the clear button looks weird.  Can you check this with Design?
Screen Shot 2019-09-13 at 4 49 59 PM

jangmi-jo:
@kevinbarabash I added minHeight for the search text input and will update the IconButton hover shortly!

packages/wonder-blocks-dropdown/util/constants.js
```diff
@@ -14,3 +14,8 @@ export const selectDropdownStyle = {
     marginTop: Spacing.xSmall,
     marginBottom: Spacing.xSmall,
 };
+
+export const filterableDropdownStyle = {
+    minHeight: 100,
+    maxHeight: 384,
```
somewhatabstract: Where do these constants come from? How were they determined? Can you add a comment, perhaps?

jangmi-jo: Hi Jeff! The constants are from the design spec, "Dimensions" in the Figma. I wonder if adding "These numbers are from design" or "See dimensions from Figma <link>" would be helpful comment.

jangmi-jo:
@somewhatabstract Absolutely! That's what I will do today :) Adding tests for this new feature. Stay tuned!

packages/wonder-blocks-dropdown/components/search-text-input.js
```diff
@@ -0,0 +1,130 @@
...
+                    placeholder="Filter"
+                    value={searchText}
+                    className={css(
+                        styles.inputStyleReset,
+                        typographyStyles.LabelMedium,
+                    )}
+                />
+                {searchText.length > 0 && (
+                    <IconButton
+                        icon={icons.dismiss}
```
jandrade: It would be good to include a label to let screen readers know the purpose of this button (e.g `aria-label="Clear search"`).

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -386,13 +410,31 @@ export default class DropdownCore extends React.Component<
             }
         }
 
+        const showSearchTextInput =
+            !!onSearchTextChanged && typeof searchText === "string";
```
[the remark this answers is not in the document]
jangmi-jo: Mmm... Not particularly. Just thought both should exist to work. I could remove the second condition if that feels better.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -409,11 +451,22 @@ export default class DropdownCore extends React.Component<
...
+        const {
+            onOpenChanged,
+            open,
+            onSearchTextChanged,
+            searchText,
+        } = this.props;
         const keyCode = event.which || event.keyCode;
-
+        const showSearchTextInput =
+            !!onSearchTextChanged && typeof searchText === "string";
```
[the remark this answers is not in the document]
jangmi-jo: I'll revise this when I resolve the comment above :)

packages/wonder-blocks-dropdown/components/dropdown-core.js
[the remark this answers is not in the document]
jandrade: makes sense, I think is fine to enforce it :)

packages/wonder-blocks-dropdown/components/multi-select.test.js
```diff
@@ -32,6 +35,7 @@ describe("MultiSelect", () => {
                 <OptionItem label="item 1" value="1" />
                 <OptionItem label="item 2" value="2" />
                 <OptionItem label="item 3" value="3" />
+                {false && <OptionItem label="item 4" value="4" />}
```
jandrade: You might not need this line, unless I'm missing something.

jangmi-jo: Oh this was intentional to test the "falsey" items shouldn't get considered in total item count.

kevinbarabash:
In play testing I noticed kind of weird behavior with the clear button.  There's no hover state.

packages/wonder-blocks-dropdown/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -4048,3 +4048,139 @@ exports[`wonder-blocks-dropdown example 26 1`] = `
...
+      >
+        <path
+          d="M8 8.586l3.293-3.293a1 1 0 0 1 1.414 1.414l-4 4a1 1 0 0 1-1.414 0l-4-4a1 1 0 0 1 1.414-1.414L8 8.586z"
+          fill="rgba(33,36,44,0.64)"
+        />
+      </svg>
+    </button>
+  </div>
+</div>
+`;
```
somewhatabstract: ASIDE: @kevinbarabash @jandrade  how do you feel about introducing a new snapshot serializer that annotates the output with comments indicating what components are rendering? I think it would help me when I'm reviewing snapshots and snapshot changes to see both the real DOM structure and the component structure in the snapshots.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -27,6 +29,20 @@ type DropdownProps = {|
      */
     items: Array<DropdownItem>,
 
+    /**
+     * An optional handler to set the searchText of the parent. When this and
+     * the searchText are provided, SearchTextInput will be displayed at the
+     * top of the dropdown body.
+     */
+    onSearchTextChanged?: (searchText: string) => mixed,
```
somewhatabstract: Nice with the `=> mixed`. Makes using these props much easier.

To that end, do we ever envisage a case where we want to optionally set this, like:

```
<Component
   onSearchTextChanged={this.someFlag ? this.handler : null}
/>
```

What do you think about typing this as:

`onSearchTextChanged?: ?((searchText:string) => mixed)`

So that it enables that inline syntax.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -27,6 +29,20 @@ type DropdownProps = {|
...
+
+    /**
+     * An optional string that the user entered to search the items. When this
+     * and the onSearchTextChanged are provided, SearchTextInput will be
+     * displayed at the top of the dropdown body.
+     */
+    searchText?: string,
```
somewhatabstract: Same here? Should this be `searchText?: ?string`?

packages/wonder-blocks-dropdown/__snapshots__/generated-snapshot.test.js.snap
kevinbarabash: That's a great idea!  I've always struggled with knowing what code results in a particular fixture.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -463,8 +522,45 @@ export default class DropdownCore extends React.Component<
         }
     }
 
+    maybeRenderSearchTextInput() {
```
somewhatabstract: Love the use of `maybe` on these types of methods! <3

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -463,8 +522,45 @@ export default class DropdownCore extends React.Component<
...
+                    }}
+                    itemRef={this.state.itemRefs[0].ref}
+                    searchText={searchText}
+                    style={styles.searchInput}
+                />
+                {noResult && (
+                    <LabelMedium style={styles.noResult}>
+                        No results
+                    </LabelMedium>
+                )}
```
somewhatabstract: What do you think about making this a `maybe...` function too rather than inlining it?

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -474,7 +570,11 @@ export default class DropdownCore extends React.Component<
 
         const itemRole = this.getItemRole();
 
-        let focusCounter = 0;
+        // If SearchTextInput is visible, we consider that as the first
+        // focusable item.
+        const showSearchTextInput =
+            !!onSearchTextChanged && typeof searchText === "string";
+        let focusCounter = showSearchTextInput ? 1 : 0;
```
somewhatabstract: Feels like a smell that this is a `let` rather than `const`. No change requested, but I suspect we can tidy this code up in a future change.

jangmi-jo:
@kevinbarabash that's the intended behavior from Design. The Icon button's border hover looked strange so we removed it.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -490,6 +590,7 @@ export default class DropdownCore extends React.Component<
                     dropdownStyle,
                 ]}
             >
+                {this.maybeRenderSearchTextInput()}
                 {items.map((item, index) => {
```
somewhatabstract: I think this map function should be a refactored into a member call and we could use reduce to build the focusable count. No change requested. Just a consideration for a future refactor.

packages/wonder-blocks-dropdown/components/dropdown-core.test.js
```diff
@@ -485,4 +486,83 @@ describe("DropdownCore", () => {
         option1.simulate("click");
         expect(onClick1).toHaveBeenCalledTimes(1);
     });
+
+    it("shows SearchTextInput when onSearchTextChanged and searchText is provided", () => {
+        // Arrange, Act
```
somewhatabstract: Can you change this to distinct sections for `// Arrange` and `// Act` please.

The structure is helpful for future maintenance as it encourages folks to put things in the right place. In this example, the `handleSearchTextChanged` is clearly the `// Arrange` and the `setProps` is the `// Act`.

Same for other test cases.

packages/wonder-blocks-dropdown/components/dropdown-core.test.js
```diff
@@ -485,4 +486,83 @@ describe("DropdownCore", () => {
...
+        const handleSearchTextChanged = jest.fn();
+        dropdown.setProps({
+            onSearchTextChanged: (text) => handleSearchTextChanged(text),
+            searchText: "",
+            open: true,
+        });
+        // Assert
```
somewhatabstract: NIT: Missing newline before `// Assert`

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -76,6 +79,13 @@ type Props = {|
      */
     implicitAllEnabled?: boolean,
 
+    /**
+     * When this is true, the dropdown body shows a search text input at the
+     * top. The items will be filtered by the input.
+     * Selected items will be moved to the top when the dropdown is re-opened.
+     */
+    isFilterable?: boolean,
```
somewhatabstract: ASIDE: I know we haven't done this for other, similar props, so no change requested. 

I wonder if the API would be easier if these were all maybe types; `isFilterable?: ?boolean`

@jandrade?

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -118,6 +128,17 @@ type State = {|
      * Whether or not last open state change was triggered by a keyboard click.
      */
     keyboard?: boolean,
+    /**
```
somewhatabstract: NIT: Missing newline before comment

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -118,6 +128,17 @@ type State = {|
...
+     * The text input to filter the items by their label. Defaults to an empty
+     * string.
+     */
+    searchText: string,
+    /**
```
somewhatabstract: NIT: Missing newline before comment.

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -284,28 +306,84 @@ export default class MultiSelect extends React.Component<Props, State> {
...
+        const {isFilterable} = this.props;
+        // If it's not filterable, no need to do any extra besides mapping the
+        // option items to dropdown items.
+        if (!isFilterable) {
+            return children.map(this.mapOptionItemToDropdownItem);
+        }
+
+        const {searchText, lastSelectedValues} = this.state;
+
+        let filteredChildren = children;
```
somewhatabstract: `let` is a code smell to me. I think this could be refactored by moving the creation of `filteredChildren` into a function so that `filteredChildren` is `const.

packages/wonder-blocks-dropdown/components/multi-select.md
```diff
@@ -444,4 +444,57 @@ class ControlledMultiSelectExample extends React.Component {
...
+        >
+            {optionItems}
+        </MultiSelect>;
+    }
+}
+
+<View style={styles.row}>
+    <ExampleWithShortcuts />
+</View>
+```
```
somewhatabstract: Nice!

packages/wonder-blocks-dropdown/components/multi-select.test.js
somewhatabstract: Can you add a comment to that effect so that maintainers understand why this is here?

packages/wonder-blocks-dropdown/components/multi-select.test.js
```diff
@@ -281,4 +285,194 @@ describe("MultiSelect", () => {
...
+
+        // one more up should move focus to the input again.
+        select.simulate("keydown", {keyCode: keyCodes.up});
+        select.simulate("keyup", {keyCode: keyCodes.up});
+        expect(searchInput.state("focused")).toBe(true);
+
+        // Press keyboard down arrow should move the focus down
+        select.simulate("keydown", {keyCode: keyCodes.down});
+        select.simulate("keyup", {keyCode: keyCodes.down});
+        expect(selectAll.state("focused")).toBe(true);
```
somewhatabstract: Multiple expects and acts in the same test is really hard to maintain.

I think that these should be separate test cases to keep things simple and clear. As it stands, if one of the early expectations fails, we never test the remainder, which can make tracking down why something fails much harder.

What do you think about splitting this into multiple, specific cases either as separate `it` calls or a combination of `it.each` uses?

packages/wonder-blocks-dropdown/components/search-text-input.js
```diff
@@ -0,0 +1,133 @@
...
+                    onBlur={this.handleBlur}
+                    ref={itemRef}
+                    placeholder="Filter"
+                    value={searchText}
+                    className={css(
+                        styles.inputStyleReset,
+                        typographyStyles.LabelMedium,
+                    )}
+                />
+                {searchText.length > 0 && (
```
somewhatabstract: This feels like it should be a separate `maybe...` function. What do you think?

packages/wonder-blocks-dropdown/components/search-text-input.js
```diff
@@ -0,0 +1,133 @@
...
+}
+
+const styles = StyleSheet.create({
+    inputContainer: {
+        flexDirection: "row",
+        border: `1px solid ${Color.offBlack16}`,
+        borderRadius: 4,
+        alignItems: "center",
+        height: 40,
+        minHeight: 40,
```
somewhatabstract: I wonder if these should be commented to indicate why it is `40`?

packages/wonder-blocks-dropdown/components/search-text-input.js
```diff
@@ -0,0 +1,133 @@
...
+            </View>
+        );
+    }
+}
+
+const styles = StyleSheet.create({
+    inputContainer: {
+        flexDirection: "row",
+        border: `1px solid ${Color.offBlack16}`,
+        borderRadius: 4,
```
somewhatabstract: Perhaps use a Wonder Blocks Spacing constant?

packages/wonder-blocks-dropdown/util/constants.js
somewhatabstract: I think commenting on the origins of these numbers would be helpful; just something that helps others know where they come from so that if we are considering changing them in future, we have context.

kevinbarabash:
@jangmi-jo the hover ring shows up when the close button is focused when using the keyboard control.  It feels weird to not have any feedback when using the mouse.  It makes it feel like it's disabled even though isn't.  Maybe there could be a different hover state for this button, but just not having one feels wrong.

packages/wonder-blocks-dropdown/components/dropdown-core.js
somewhatabstract: /cc @kevinbarabash @jandrade

packages/wonder-blocks-dropdown/components/dropdown-core.js
somewhatabstract: /cc @kevinbarabash @jandrade

packages/wonder-blocks-dropdown/components/dropdown-core.js
somewhatabstract: /cc @kevinbarabash @jandrade

jangmi-jo:
@kevinbarabash Ahhh I see what you mean. I'll talk to design and come up with some solution for that.

packages/wonder-blocks-dropdown/components/search-text-input.js
jangmi-jo: Hmm I generally don't use WB Spacing for borderRadius because border radius 4px means different than 4px spacing. Is that recommended use case for WB Spacing?

jangmi-jo:
@kevinbarabash the hover state seems better now! Please give another look when you have a moment 😄

kevinbarabash:
@jangmi-jo awesome!  Thanks for adding that. :hearthands:

packages/wonder-blocks-dropdown/__snapshots__/generated-snapshot.test.js.snap
jandrade: +1 That would be awesome :). Sometimes is hard to understand that in a non-visual way.

packages/wonder-blocks-dropdown/components/dropdown-core.js
jandrade: Good call! I might do this refactor in a new PR that integrates `react-window` into the `Dropdown` component.

packages/wonder-blocks-dropdown/components/multi-select.js
jandrade: I understand your point. For these cases, where this prop value is statically added (e.g. `isFilterable={true}`, I'd prefer to enforce valid states instead of expecting some truthy or falsy values, but I'm not sure what have been the case with the previous uses of other components. Do you think that we have had less bugs by using maybe types for these flags in WB components?. If that's the case, I think it would be ok to change this to avoid future issues.

PR #483, opened 2019-09-27
WB-643 feat(dropdown): Add react-window

jandrade:
## Dropdown
- Added `react-window` to optionally to render MultiSelect dropdown items (100+ items)
- Created `DropdownCoreVirtualized` to use the `react-window`'s `VariableSizeList` component.
- Created `DropdownVirtualizedItem` class component in order to use `ref` inside each item.
- Updated the **MultiSelect with search filter**  example to use this version.

### Test Plan
- Open https://deploy-preview-483--wonder-blocks.netlify.com/#!/MultiSelect/17 and check that the example is using the virtualized version.

multi-select-windowing


jeresig:

Built with commit 0d680c838571c8d8ee85ec481faad900c8fbd21b

https://deploy-preview-483--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.js
```diff
@@ -0,0 +1,113 @@
...
+ */
+class DropdownCoreVirtualized extends React.Component<Props> {
+    /**
+     * Calculates item height
+     */
+    getItemSize = (index: number) => {
+        // get the current item in the list
+        const item = this.props.data[index];
+
+        return SeparatorItem.isClassOf(item.component) ? 9 : 40;
```
jangmi-jo: could be worth commenting these numbers.

jandrade:
Something I forgot to mention here (related to the dropdown width):

There is a problem of using `react-window` and any other virtualization libraries. As the algorithm needs to calculate which items will be visible, the library adds `position: absolute` to each list item and also adds a vertical position (e.g. `top: 120px`), leading to create a new stacking context on each list item. Sadly, as these items are in a different stacking context, then the list container doesn't have a reference to automatically calculate its width from their children contents. To solve that, `react-window` expects a fixed `width` that will be used by the container.

kevinbarabash:
I tried setting the `dropdownStyle` on `MultiSelect` to make the dropdown wider but wasn't able to work.  I think we should automatically ellipsize long string when using `react-window`.  Also, are there anyway to control how many items it loads around the visible "window"?  I noticed that sometimes you can see empty space when scrolling quickly.

jandrade:
good points... I'm going to check the `dropdownStyle` issue and truncating strings as well. And yes, there's one prop to control how many items to load, I'll modify it.

packages/wonder-blocks-dropdown/components/separator-item.js
```diff
@@ -9,17 +9,31 @@ import Color from "@khanacademy/wonder-blocks-color";
...
     static __IS_SEPARATOR_ITEM__ = true;
     static isClassOf(instance: React.Element<any>) {
         return instance && instance.type && instance.type.__IS_SEPARATOR_ITEM__;
     }
 
     render() {
-        return <View style={styles.separator} />;
+        return (
+            //  wrapper to use styles from react-window (if applies)
+            <View style={this.props.style}>
```
jeresig: Do we want this to be a wrapper or is it enough to do: `<View style={[this.props.style, styles.separator]} />`

packages/wonder-blocks-dropdown/components/multi-select.md
```diff
@@ -448,7 +448,12 @@ class ControlledMultiSelectExample extends React.Component {
 
 ### Multi select with search filter
 
-When there are many options, you could use a search filter in the MultiSelect. The search filter will be performed toward the labels of the option items.
+When there are many options, you could use a search filter in the MultiSelect.
+The search filter will be performed toward the labels of the option items.
+
+*NOTE:* If there are more than 100 options, the component will automatically use
```
jeresig: I'd frame this as "If there are 100+ options, ...`

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -595,34 +642,17 @@ export default class DropdownCore extends React.Component<
...
-                                }
-                                if (item.populatedProps.onClick) {
-                                    item.populatedProps.onClick();
-                                }
-                            },
-                        });
-                    }
-                })}
+
+                {itemsList.length < 100 ? (
```
jeresig: I feel like this should be a constant at the top of this file (and potentially something that should be configurable - but we can explore that later).

packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.js
jeresig: Agreed! And do these numbers need to be kept in-sync with any others?

packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.js
```diff
@@ -0,0 +1,113 @@
+// @flow
+import * as React from "react";
+import {VariableSizeList as List} from "react-window";
```
jeresig: I was a little surprised by the rename here, I think it's ok to leave the name as-is.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -595,34 +642,17 @@ export default class DropdownCore extends React.Component<
...
+                    this.renderAllItems(itemsList)
+                ) : (
+                    <DropdownCoreVirtualized
```
jeresig: How bad would it be to just always have it be virtualized? I'm really not excited about having two code paths for everything and having the potential for stuff to break in strange ways as we can't anticipate the change. If we do want this separated what can we do to ensure that we're able to keep these two modes in sync?

jandrade: Yeah, that's an excellent point. I'm going to change that to always use virtualized lists. At the end, if there are a few items, then the library should be able to detect when to virtualize items.

packages/wonder-blocks-dropdown/components/separator-item.js
jandrade: We can merge both styles if we use `box-shadow` instead of `background`. I'll make that change.

packages/wonder-blocks-dropdown/components/multi-select.md
jandrade: Changed the copy here to point out we are using `react-window` for all the cases.

packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.js
```diff
@@ -1,113 +1,206 @@
...
+class DropdownCoreVirtualized extends React.Component<Props, State> {
+    state = {
+        height: this.getHeight(),
+        width: null,
+    };
+
+    componentDidMount() {
+        // Wait for styles to be applied. This way, we can get a more precise
+        // value of the container dimensions.
+        setTimeout(() => {
```
jeresig: I think we definitely want to use the timing API for this! And I think we should be using `animationFrame` specifically as the styles should be laid out by that point.

packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.js
```diff
@@ -1,113 +1,206 @@
...
-    width: number,
-|};
+    getHeight() {
+        // calculate using the first 10 items on the array as we want to display
+        // this number of elements in the visible area
+        return this.props.data
+            .slice(0, MAX_VISIBLE_ITEMS)
+            .reduce((sum, item) => {
+                if (SeparatorItem.isClassOf(item.component)) {
+                    return sum + 9;
```
jeresig: Hmm - this should probably be a constant like `SEPARATOR_ITEM_HEIGHT`?

packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.js
```diff
@@ -1,113 +1,206 @@
...
+                } else if (SearchTextInput.isClassOf(item.component)) {
+                    // search text input height
+                    return sum + SEARCH_ITEM_HEIGHT;
+                } else {
+                    return sum + DROPDOWN_ITEM_HEIGHT;
+                }
+            }, 5);
```
jeresig: Starting with 5? Can we make this a constant and document it?

packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.js
```diff
@@ -1,113 +1,206 @@
...
      * Calculates item height
      */
     getItemSize = (index: number) => {
         // get the current item in the list
         const item = this.props.data[index];
 
-        return SeparatorItem.isClassOf(item.component) ? 9 : 40;
+        if (SeparatorItem.isClassOf(item.component)) {
+            // this is the separator's height (1px) + vertical margin (8px)
+            return 9;
```
jeresig: Same here, too.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -361,7 +358,13 @@ export default class DropdownCore extends React.Component<
...
             this.focusedIndex -= 1;
         }
-        this.focusCurrentItem();
+
+        // force react-window to scroll on the correct position
+        this.listRef.current &&
+            this.listRef.current.scrollToItem(this.focusedIndex);
+
+        // wait for windowed items to be recalculated
+        setTimeout(() => this.focusCurrentItem(), 0);
```
jeresig: Need to use the same `animationFrame` thing here, too.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -361,7 +358,13 @@ export default class DropdownCore extends React.Component<
         } else {
             this.focusedIndex -= 1;
         }
-        this.focusCurrentItem();
+
+        // force react-window to scroll on the correct position
+        this.listRef.current &&
```
jeresig: Minor nit, could you use an explicit if statement for this check?

```
if (this.listRef.current) {
    this.listRef.current.scrollToItem(this.focusedIndex);
}
```

I find these to be easier to understand (especially when we're not returning a value)

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -370,7 +373,13 @@ export default class DropdownCore extends React.Component<
         } else {
             this.focusedIndex += 1;
         }
-        this.focusCurrentItem();
+
+        // force react-window to scroll on the correct position
+        this.listRef.current &&
```
jeresig: Same here.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -370,7 +373,13 @@ export default class DropdownCore extends React.Component<
...
+            this.listRef.current.scrollToItem(this.focusedIndex);
+
+        // wait for windowed items to be recalculated
+        setTimeout(() => this.focusCurrentItem(), 0);
```
jeresig: Timing API here too.

packages/wonder-blocks-dropdown/components/multi-select.md
```diff
@@ -451,7 +451,7 @@ class ControlledMultiSelectExample extends React.Component {
 When there are many options, you could use a search filter in the MultiSelect.
 The search filter will be performed toward the labels of the option items.
 
-*NOTE:* If there are more than 100 options, the component will automatically use
+*NOTE:* The component automatically uses
 [react-window](https://github.com/bvaughn/react-window) to improve performance
 when rendering these elements.
```
jeresig: Might be worth mentioning something like "...when rendering these elements and is capable of handling many hundreds of items without performance problems."

packages/wonder-blocks-dropdown/components/search-text-input.js
```diff
@@ -38,7 +70,7 @@ export default class SearchTextInput extends React.Component<Props, State> {
         const {onClick, onChange} = this.props;
         // Empty the search text and focus the SearchTextInput
         onChange("");
-        onClick();
+        onClick && onClick();
```
jeresig: `if` nit here too.

packages/wonder-blocks-dropdown/components/separator-item.js
```diff
@@ -22,25 +22,23 @@ export default class SeparatorItem extends React.Component<{|
...
+            // pass optional styles from react-window (if applies)
+            <View style={[styles.separator, this.props.style]} />
         );
     }
 }
 
 const styles = StyleSheet.create({
     separator: {
-        backgroundColor: Color.offBlack16,
+        boxShadow: `0 -1px ${Color.offBlack16}`,
```
jeresig: Oh, nice! What's the effect of this?

packages/wonder-blocks-dropdown/components/option-item.js
```diff
@@ -184,6 +196,9 @@ const styles = StyleSheet.create({
         whiteSpace: "nowrap",
         userSelect: "none",
         marginLeft: Spacing.xSmall,
+        // added to truncate strings that are longer than expected
+        overflow: "hidden",
+        textOverflow: "ellipsis",
```
jangmi-jo: I recall we chatted about using `title`. Did you get a chance to test with screen reader? Is there any way to add them but not for screen readers so that we don't have redundant labels?

packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.js
```diff
@@ -0,0 +1,206 @@
...
+     * Update container width
+     */
+    setWidth() {
+        const rootNode = ((ReactDOM.findDOMNode(this): any): ?HTMLElement);
+
+        // after the non-virtualized items are rendered, we get the container
+        //  width to pass it to react-window's List
+        if (rootNode) {
+            this.setState({
+                width: rootNode.getBoundingClientRect().width,
```
jangmi-jo: Hmm so, I see that we are sorting the items and get the top 10 longest labels to render but somehow I still see the ellipsis 😭 
The only change I made was:
```

const optionItems = new Array(1000).fill(null).map((_, i) => (<OptionItem
    key={i}
    value={(i + 1).toString()}
    label={`School ${i + 1} in Wizarding World Some more really long labels?`}
/>));
```
Screen Shot 2019-10-10 at 12 43 23 PM

packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.js
jandrade: Oh good catch, I think it was me playing around when defining this function. I forgot to remove that :|

packages/wonder-blocks-dropdown/components/separator-item.js
jandrade: Before that we used `height: 1px` to fill the space with the `background` color,  but `react-window` was overriding that value to `9px`, b/c of the SEPARATOR_ITEM_HEIGHT that is 1px (border) + 8px (vertical margin).

This change prevents that behavior.

packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.js
```diff
@@ -194,13 +226,14 @@ class DropdownCoreVirtualized extends React.Component<Props, State> {
...
         } else {
-            // then render the virtualized list
+            // width has been provided, then render the virtualized list
             return this.renderVirtualizedList();
         }
     }
 }
 
-export default DropdownCoreVirtualized;
+export default withActionScheduler/*:: <OwnProps> */(DropdownCoreVirtualized);
```
jeresig: Hmm - this is a bit unfortunate. I think we may need to look in to updating `withActionScheduler` with the better Flow types that @kevinbarabash figured out.

I noticed that you were able to use the correct (non-comment) Flow types for `withActionScheduler` below - why didn't it work here?

packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.test.js
```diff
@@ -1,16 +1,33 @@
 //@flow
 import * as React from "react";
 import {VariableSizeList as List} from "react-window";
+// import {shallow} from "enzyme";
```
jeresig: Oops - you can remove this!

packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.test.js
```diff
@@ -1,16 +1,33 @@
...
 import DropdownCoreVirtualized from "./dropdown-core-virtualized.js";
 import SearchTextInput from "./search-text-input.js";
 
-jest.useFakeTimers();
-
 describe("DropdownCoreVirtualized", () => {
+    beforeEach(() => {
+        jest.useFakeTimers();
+
+        // Jest doesn't fake out the animation frame API, so we're going to do
```
jeresig: Ahh - that's annoying!

packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.test.js
```diff
@@ -1,16 +1,33 @@
...
+        // it here and map it to timeouts, that way we can use the fake timer
+        // API to test our animation frame things.
+        jest.spyOn(global, "requestAnimationFrame").mockImplementation(
+            (fn, ...args) => setTimeout(fn, 0),
```
jeresig: Is the `...args` necessary? It seems like you could just remove it - same for the mock below, too.

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -333,11 +340,12 @@ export default class DropdownCore extends React.Component<
         }
     };
 
+    scheduleToFocusCurrentItem() {
+        // wait for windowed items to be recalculated
+        this.props.schedule.animationFrame(() => this.focusCurrentItem());
+    }
+
     focusCurrentItem() {
-        // Because the dropdown menu is portalled, focusing this element
```
jeresig: Ah - this isn't necessary anymore?

packages/wonder-blocks-dropdown/components/dropdown.test.js
```diff
@@ -23,11 +23,21 @@ type State = {|
 jest.mock("./dropdown-core-virtualized.js");
 
 describe("Dropdown", () => {
-    window.scrollTo = jest.fn();
     window.getComputedStyle = jest.fn();
     let dropdown;
 
     beforeEach(() => {
+        jest.useFakeTimers();
+        // Jest doesn't fake out the animation frame API, so we're going to do
```
jeresig: Same comments as above.

packages/wonder-blocks-dropdown/components/multi-select.test.js
```diff
@@ -24,7 +24,18 @@ describe("MultiSelect", () => {
     const onChange = jest.fn();
 
     beforeEach(() => {
-        window.scrollTo = jest.fn();
+        jest.useFakeTimers();
+
+        // Jest doesn't fake out the animation frame API, so we're going to do
```
jeresig: Here too.

packages/wonder-blocks-dropdown/components/search-text-input.js
```diff
@@ -42,6 +42,11 @@ type Props = {|
      * Custom styles for the main wrapper
      */
     style?: StyleType,
+
+    /**
+     * Test ID used for e2e testing.
+     */
+    testId?: string,
```
jeresig: Ah - good catch!

packages/wonder-blocks-dropdown/util/constants.js
```diff
@@ -30,5 +30,7 @@ export const searchInputStyle = {
 // The default item height
 export const DROPDOWN_ITEM_HEIGHT = 40;
 
+export const SEPARATOR_ITEM_HEIGHT = 9;
```
jeresig: Thank you!

packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.js
jandrade: 🤦‍♂ you're right, I implemented this one before and I think I just got used to add those comments without testing this way first.

packages/wonder-blocks-dropdown/components/option-item.js
jandrade: Not yet :(. Also, I noticed that we don't support `aria-label` and or `title` in option items, so I'm going to work on that functionality next sprint as part of this ticket: https://khanacademy.atlassian.net/browse/WB-615

packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.js
jandrade: I used your snippet and cannot see the ellipsis. Btw I added the 512 max rule to the width :).

jandrade:
@jangmi-jo Answering to your comments here:

> 1. When I open the dropdown with "Enter" and press down arrow, the focus starts from index 1, not from 0.

> 2. The "No results" doesn't seem centered.
Screen Shot 2019-10-10 at 12 49 08 PM

> 3. When I have these props for more than 100 items,I see this: "pressing arrow up from last item" scrolls up one item. Not a huge deal but just one nit I see :)

1 are 2 should be fixed :).

For 3, I tested it and yeah, it's a weird behavior. I created this ticket to work on that ASAP: https://khanacademy.atlassian.net/browse/WB-649

PR #541, opened 2019-10-10
Github Action to Auto-approve Dependabot diffs

jeresig:
I was reading about how people manage dependabot with a Code Owners file and it seems like some recommend auto-approving them with a Github Action, so I thought I'd give it a try! Let's see what happens.

### DIFF
--- a/.github/workflows/auto-approve.yml
+++ b/.github/workflows/auto-approve.yml
@@ -1,0 +1,11 @@
+name: Auto approve
+on: pull_request
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: hmarr/auto-approve-action@v2.0.0
+      if: github.actor == 'dependabot[bot]' || github.actor == 'dependabot-preview[bot]'
+      with:
+        github-token: "${{ secrets.GITHUB_TOKEN }}"
--- a/.github/workflows/auto-approve.yml
+++ b/.github/workflows/auto-approve.yml
@@ -2,10 +2,11 @@ name: Auto approve
 on: pull_request
 
 jobs:
-  build:
+  approve:
+    name: Auto-approve dependabot PRs
+    if: github.actor == 'dependabot[bot]' || github.actor == 'dependabot-preview[bot]'
     runs-on: ubuntu-latest
     steps:
     - uses: hmarr/auto-approve-action@v2.0.0
-      if: github.actor == 'dependabot[bot]' || github.actor == 'dependabot-preview[bot]'
       with:
         github-token: "${{ secrets.GITHUB_TOKEN }}"


jeresig:

Built with commit c4c631ecf8b131174cee9b13b44f8d0ad445290a

https://deploy-preview-541--wonder-blocks.netlify.com

somewhatabstract:
Does this mean we never even review them?

jeresig:
@somewhatabstract Correct - for anything that is submitted and the build is passing it'll get auto-approved and merged. If we want to prevent it from upgrading certain things or only upgrading to certain versions we can configure that.

.github/workflows/auto-approve.yml
```diff
@@ -0,0 +1,11 @@
...
+on: pull_request
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: hmarr/auto-approve-action@v2.0.0
+      if: github.actor == 'dependabot[bot]' || github.actor == 'dependabot-preview[bot]'
+      with:
+        github-token: "${{ secrets.GITHUB_TOKEN }}"
```
jandrade: How secrets would work in this case? is it a personal token? or shared token for this project (via settings)?

jeresig: @jandrade It seems like it's auto-configured! More details here: https://help.github.com/en/articles/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables

somewhatabstract: As for secrets in general for actions, they are specified on the repo. Once set, you cannot see their values.

somewhatabstract:
> @somewhatabstract Correct - for anything that is submitted and the build is passing it'll get auto-approved and merged. If we want to prevent it from upgrading certain things or only upgrading to certain versions we can configure that.

Good idea to make sure our checks cover all aspects of release then.

I recommend taking Lerna out of the auto-update loop until we have a good test for that.

.github/workflows/auto-approve.yml
```diff
@@ -0,0 +1,11 @@
+name: Auto approve
+on: pull_request
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: hmarr/auto-approve-action@v2.0.0
+      if: github.actor == 'dependabot[bot]' || github.actor == 'dependabot-preview[bot]'
```
somewhatabstract: I think it will save time and resources if you move or duplicate this condition on the job itself. That way the job won't even run if this doesn't match.

.github/workflows/auto-approve.yml
```diff
@@ -0,0 +1,11 @@
+name: Auto approve
+on: pull_request
+
+jobs:
+  build:
```
somewhatabstract: This name is weird for this job. Perhaps "approve" would be better?

Also, consider giving the job a `name:` for the logs that's clearer, like "Auto-approve dependabot diff"

somewhatabstract:
I made my recommended changes and added this to `ancesdir`. Seems to be working.

https://github.com/somewhatabstract/ancesdir/pull/8

somewhatabstract:
Hmm, looks like we probably need to change these settings too:

image

For the dev deps, might want to consider not doing "All updates" to cater to the lerna issue, perhaps?

somewhatabstract:
With this learned, I'm going to try updating this on `checksync` and try it again, to see what works.

somewhatabstract:
OK, that seems to be the trick. GitHub and dependabot are working together on my `checksync` project to update dependencies.

somewhatabstract:
Ooof, watch out. Required Checks is broken when you have multiple workflow files.

jeresig:
@somewhatabstract Good call on the changes! I've fixed it up to match what you did in ancesdir.

PR #548, opened 2019-10-15
Add Alexjs to detect insensitive language

jeresig:
This was added to React Native recently and I thought it was a good idea!
https://github.com/facebook/react-native-website/pull/1337

More information: https://alexjs.com/

I think the one piece of our API we could consider changing would be "disabled", but I have deferred that for now.

This feels like a pretty harmless linter and is easy to configure as well, may be worth a shot giving it a try!

Test plan:
I ran `yarn alex` and it passed!


jeresig:

Built with commit 8990fdaf858d354e4a607eaf292248452a311cb1

https://deploy-preview-548--wonder-blocks.netlify.com

jeresig:
@jandrade Ahh - good call! Let's see how this tool goes and then we can start adding it to the vscode config too!

jandrade:
Sounds good to me :)

kevinbarabash:
Cool!  There's a vscode plugin too: https://marketplace.visualstudio.com/items?itemName=shinnn.alex

kevinbarabash:
Why are `easily` and `simply` inconsiderate?

jeresig:
@kevinbarabash I think it's because it's considered potentially demeaning - assuming that people will find something "simple", "trivial", etc. when it may not actually be that - can make people feel excluded from using the tools.

kevinbarabash:
@jeresig that makes sense.  I'll try to avoid those words.  I'm glad we have `alexjs` to help with that.

kevinbarabash:
I just have to remember all of those times my uni profs would say something was simple or easy and I was there scratching my head thinking to myself this is the hardest thing I've ever done in my life.

jeresig:
@kevinbarabash Yes, exactly!!

PR #555, opened 2019-10-17
Allow buttons to truncate if parent container is too long

dierat:
Whilest working on a banner in webapp, I noticed that WB buttons don't correctly truncate when the parent container is too narrow. The issue is the `display: "flex"` styling. Just removing it fixes it, but I chose to instead change it to `display: "inline-block"` so it's clear that it's not set to flex for a reason. I'm happy to remove that and leave the comment if you all feel that would be better.

I also updated the docs so now there's an example of that truncation and we can easily see if it breaks again!

Before:
before

After:
after

Test plan:
Navigate to `/#button-1` in the WB docs,
verify there's now a section under "Best practices" with an example where the parent is cutting off one of the buttons,
and verify that the button that's cut off is truncating properly with ellipsis.


jeresig:

Built with commit e6b4bdb78914179f5801a8bf5b0ad1640acae7c8

https://deploy-preview-555--wonder-blocks.netlify.com

kevinbarabash:
Thanks for the PR.  The ellispis look good, but the proposed change has changed the vertical alignment of the label when there's an icon in the button.
Screen Shot 2019-10-17 at 11 37 10 PM

kevinbarabash:
[clabot:check]

dierat:
Oh I see! That's the reason the span had flex on it, to allow it the icon and text to sit nicely together.

dierat:
Updated so it shouldn't negatively affect the icon anymore. I don't see anything weird in any of the other buttons either, but I'm not as familiar with the buttons and their different permutations.

PR #565, opened 2019-10-25
[feat:dropdown] WB-615: OptionItem -> Add ARIA props and switch to LabelMedium

jandrade:
## Summary
**OptionItem:**
- Added `aria-label` prop to the `OptionItem` component. This change is introduced to add the ability to pass any relevant information on every option item that screen readers can understand.
- Changed `font-size` from `bold` to `normal` to match design specs.

## Test plan
- Open https://deploy-preview-565--wonder-blocks.netlify.com/#multiselect.
- Check the `accessibility` example and check that `aria-label` is added to the option items.

option-item-label

cc @aag


jeresig:

Built with commit a55ee576d6e5dd93e4c7630393c7bf10b377951e

https://deploy-preview-565--wonder-blocks.netlify.com

packages/wonder-blocks-dropdown/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -3615,7 +3615,7 @@ exports[`wonder-blocks-dropdown example 24 1`] = `
           "borderStyle": "solid",
           "borderWidth": 1,
           "boxSizing": "border-box",
-          "color": "rgba(33,36,44,0.64)",
+          "color": "#21242c",
```
kevinbarabash: Why did the color change here?

packages/wonder-blocks-dropdown/components/option-item.js
```diff
@@ -6,16 +6,18 @@ import PropTypes from "prop-types";
...
 
-import type {StyleType} from "@khanacademy/wonder-blocks-core";
+import type {AriaProps, StyleType} from "@khanacademy/wonder-blocks-core";
 
 import {DROPDOWN_ITEM_HEIGHT} from "../util/constants.js";
 import Check from "./check.js";
 import Checkbox from "./checkbox.js";
 
 type OptionProps = {|
+    ...AriaProps,
```
kevinbarabash: If we accept all `AriaProps` we should be passing all them to the underlying `View`, not just `aria-label`.

packages/wonder-blocks-dropdown/components/option-item.js
```diff
@@ -148,9 +159,9 @@ export default class OptionItem extends React.Component<OptionProps> {
                                 selected={selected}
                                 {...state}
                             />
-                            <LabelLarge style={styles.label}>
+                            <LabelMedium style={styles.label}>
                                 {label}
-                            </LabelLarge>
+                            </LabelMedium>
```
kevinbarabash: Why did the size of the label change?  Is there a JIRA ticket for this?

packages/wonder-blocks-dropdown/components/separator-item.js
```diff
@@ -31,7 +31,10 @@ export default class SeparatorItem extends React.Component<{|
     render() {
         return (
             // pass optional styles from react-window (if applies)
-            <View style={[styles.separator, this.props.style]} />
+            <View
+                style={[styles.separator, this.props.style]}
+                aria-hidden="true"
```
kevinbarabash: Good call!

packages/wonder-blocks-dropdown/__snapshots__/generated-snapshot.test.js.snap
jandrade: I think that it's because I changed that label to use `LabelMedium`.

packages/wonder-blocks-dropdown/components/option-item.js
jandrade: Ohh good call, I didn't mention that here. We found a visual bug with the option items using `font-weight: bold` instead of `normal`. As the change was too small, I decided to make it as part of this PR.

Something I didn't think before, is that `LabelLarge` and `LabelMedium` have the same `font-size` but different `font-weight`. Maybe we should revisit this as the names could be suggesting different sizes between them.

PR #573, opened 2019-10-31
Add import to Tooltip example

jeremywiebe:
The first example for Tooltip didn't have an import. It's maybe obvious for folks who've used WB for a while, but I had to hunt a bit for the right syntax.  So this PR adds the import.

*Before*

image


jeresig:

Built with commit 6b90e39e2dca0d22cb06c636886437605bc723cd

https://deploy-preview-573--wonder-blocks.netlify.com

PR #615, opened 2019-12-01
Delete .npmignore files

somewhatabstract:
They aren't useful anymore and `.npmignore` is generally a dangerous file to have around:

https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d

There's also a snapshot file change here. I think it's down to the lint rules ordering imports.

### DIFF
--- a/packages/wonder-blocks-button/.npmignore
+++ b/packages/wonder-blocks-button/.npmignore
@@ -1,2 +1,0 @@
-index_stories.js
-index_test.js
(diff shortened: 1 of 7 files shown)


jeresig:

Built with commit d1d2eb6599eb8146080b893b95cc321b2b9aea07

https://deploy-preview-615--wonder-blocks.netlify.com

kevinbarabash:
It sounds like we should be using `"files"` in our package.jsons but we can't do that until we figure out how not to include all of the source .js for flow purposes. 😭

PR #616, opened 2019-12-01
Upgrade to flow-bin 0.111.0

somewhatabstract:
We can't go to 0.112.0 just yet. There are more things to tackle. But this gets us to 0.111.0


packages/wonder-blocks-tooltip/generated-snapshot.test.js
```diff
@@ -8,8 +8,8 @@ import renderer from "react-test-renderer";
 
 // Mock react-dom as jest doesn't like findDOMNode.
 jest.mock("react-dom");
-import {View, Text} from "@khanacademy/wonder-blocks-core";
 import Tooltip, {TooltipContent} from "@khanacademy/wonder-blocks-tooltip";
+import {View, Text} from "@khanacademy/wonder-blocks-core";
```
somewhatabstract: Pretty sure this is our linter settings.

jeresig:

Built with commit f0092c8b53c96d01a6ec684519bf3ded2b456b8c

https://deploy-preview-616--wonder-blocks.netlify.com

somewhatabstract:
That's odd. The linter is complaining about code that is actually fixed in this diff.

jandrade:
It seems that it's already solved in the `react-window` repo, but they haven't published a new release yet.

https://github.com/bvaughn/react-window/blob/master/src/createGridComponent.js#L613

kevinbarabash:
I was reading somewhere that people are using `[declarations]` in place of `[ignore]` and in some cases putting their whole `node_modules` in `[declarations]` when upgrading to newer versions of flow.

PR #619, opened 2019-12-02
WEB-1629.3 Implement the RequestHandler base class

somewhatabstract:
This implements the base class and interface for request handlers as per ADR 286. See that and subsequent PRs for more.

- [x] Write tests
- [x] Update docs


jeresig:

Built with commit c2b77531826ee445f2c46ad49f5ffc69ae94aabf

https://deploy-preview-619--wonder-blocks.netlify.com

packages/wonder-blocks-data/util/request-handler.js
```diff
@@ -0,0 +1,40 @@
...
+    invalidateCache(options: TOptions): boolean {
+        /**
+         * By default, the cache is always valid.
+         */
+        return false;
+    }
+
+    getKey(options: TOptions): string {
+        try {
+            return String(JSON.stringify(options));
```
jeresig: Why are you wrapping this with a `String(...)` - `JSON.stringify` already returns a string. If this is a Flow thing then just casting the value should be fine.

packages/wonder-blocks-data/util/request-handler.md
```diff
@@ -0,0 +1,13 @@
...
+base class to implement your own request handler.
+
+The constructor requires a `type` to identify your handler. This should be unique
+among the handlers that are used across your application, otherwise, requests
+may be fulfilled by the wrong handler.
+
+The `fulfillRequest` method of this class is not implemented and will throw if
+called.
+
+A default implemnetation of `getKey` is provided that serializes the options of
```
jeresig: *implementation

packages/wonder-blocks-data/util/request-handler.js
somewhatabstract: `JSON.stringify` can return `undefined` which is not a valid string. We want an actual string here, always.

jeresig: It seems like this only happens if a string isn't passed in, like: `JSON.stringify(undefined)` or `JSON.stringify(function(){})`. I think in any case where `undefined` is returned we definitely want to treat that like an error and not just swallow it, converting it to a string.

packages/wonder-blocks-data/util/types.js
```diff
@@ -0,0 +1,33 @@
...
+     * If this returns true, the framework will fulfill a new request by
+     * calling `fulfillRequest`.
+     */
+    invalidateCache(options: TOptions): boolean;
+
+    /**
+     * Get the key to use for a given request. This should be idempotent for a
+     * given options set if you want caching to work across requests.
+     */
+    getKey(options: TOptions): string;
```
kevinbarabash: Nice when you can copy/paste straight from the ADR. :)

packages/wonder-blocks-data/util/request-handler.js
kevinbarabash: I changed the flow types for `JSON.stringify` in https://github.com/Khan/webapp/blob/master/flow-typed/json.js to get around this.  We could do the same here.

packages/wonder-blocks-data/util/request-handler.test.js
```diff
@@ -0,0 +1,99 @@
...
+    afterEach(() => {
+        /**
+         * This is needed or the JSON.stringify mocks need to be
+         * mockImplementationOnce. This is because if the snapshots need
+         * to update, they write the inline snapshot and that appears to invoke
+         * prettier which in turn, calls JSON.stringify. And if that mock
+         * throws, then boom. No snapshot update and a big old confusing test
+         * failure.
+         */
+        jest.restoreAllMocks();
```
kevinbarabash: That's pretty scary.  Thanks for writing explanation of what goes on without this call to `restoreAllMocks()`

packages/wonder-blocks-data/util/request-handler.js
```diff
@@ -0,0 +1,40 @@
...
+    get type() {
+        return this._type;
+    }
+
+    invalidateCache(options: TOptions): boolean {
+        /**
+         * By default, the cache is always valid.
+         */
+        return false;
+    }
```
kevinbarabash: Is there anytime a sub-claass of RequestHandler would also want to return `false`?

packages/wonder-blocks-data/util/request-handler.js
somewhatabstract: I don't want to block `undefined`. It's a totally valid thing for options to be if, say, a handler doesn't even have options.

I could handle it differently though.

packages/wonder-blocks-data/util/types.js
somewhatabstract: Well, technically, some of this was copied from the ADR, then changed, then copied back. :grin:

packages/wonder-blocks-data/util/request-handler.js
somewhatabstract: Not sure I understand the question. Can you rephrase it?

somewhatabstract: I've updated the `request-handler.md` to cover this method and its uses.

packages/wonder-blocks-data/util/request-handler.js
somewhatabstract: I've replaced this use of `String` with a check for undefined and a cast.

packages/wonder-blocks-data/util/request-handler.md
somewhatabstract: Fixed

PR #620, opened 2019-12-02
WEB-1629.4 Implement TrackData component and tracking cache

somewhatabstract:
This implements the request tracking framework. It perhaps isn't clear on its own how this all fits together. The next PR will implement the fulfillment of these tracked requests, which should make things clearer.

Clearer still when we actually have the main `Data` component.

See ADR 286 for more info.


jeresig:

Built with commit 86640536097ae4b0783883264fc445338d5817e7

https://deploy-preview-620--wonder-blocks.netlify.com

packages/wonder-blocks-data/components/track-data.md
```diff
@@ -0,0 +1,3 @@
+The `TrackData` component is a server-side only component. It should be used as
+a parent to the components whose data requests you want to fulfill during
+server-side rendering.
```
somewhatabstract: I'll be improving docs around this in a final PR.

packages/wonder-blocks-data/components/track-data.test.js
```diff
@@ -0,0 +1,58 @@
...
+                <TrackData>
+                    <TrackerContext.Consumer>
+                        {(fn) => resolve(fn)}
+                    </TrackerContext.Consumer>
+                </TrackData>,
+            );
+        });
+
+        // Assert
+        expect(result).toBe(requestTracker.trackDataRequest);
```
kevinbarabash: Nice test.

packages/wonder-blocks-data/util/request-tracking.js
```diff
@@ -0,0 +1,81 @@
...
+        const key = handler.getKey(options);
+        const type = handler.type;
+
+        /**
+         * Make sure we have stored the handler for use when fulfilling requests.
+         */
+        if (this._trackedHandlers[type] == null) {
+            this._trackedHandlers[type] = handler;
+            this._trackedRequests[type] = {};
+        }
```
kevinbarabash: The assumption here is that both `_trackedHandlers[type]` and `this._trackedRequests[type]` are either both undefined or both defined at the same time, right?

packages/wonder-blocks-data/util/request-tracking.js
```diff
@@ -0,0 +1,81 @@
...
+        };
+    }
+}
+
+/**
+ * The default export is the instance of the request tracker that each of our
+ * other pieces use. This way everything uses the same interface but we have the
+ * class with which to perform testing.
+ */
+export default new RequestTracker();
```
kevinbarabash: nit: the file exports aa `RequestTracker` but is named request-tracking.js, could we call it request-tracker.js to match the default export's name?

packages/wonder-blocks-data/util/request-tracking.js
```diff
@@ -0,0 +1,81 @@
+// @flow
+import * as React from "react";
+
+import type {IRequestHandler} from "./types.js";
+
+type TrackerFn = (handler: IRequestHandler<any, any>, options: any) => void;
+
+type HandlerCache = {
+    [type: string]: IRequestHandler<any, any>,
+    ...,
```
jandrade: TIL: This syntax is completely new to me. What's the purpose of it?

kevinbarabash: This is marking this type as being explicitly inexact.  Flow is moving towards exact by default so they needed to introduce this syntax to allow for inexact objects in the future.

packages/wonder-blocks-data/util/request-tracking.js
somewhatabstract: Yes, that's correct.

packages/wonder-blocks-data/util/request-tracking.js
somewhatabstract: Yes, but I'll make that change after landing everything to avoid annoying merge issues that might crop up.

PR #621, opened 2019-12-02
WEB-1629.5 Add response cache

somewhatabstract:
Here we add the response cache that will store data. This provides means to initialize the cache, update the cache, and retrieve data from the cache per ADR 286.

With this in place and the other pieces in place, we can implement the fulfillment method for server-side use and the client-side `Data` component.


jeresig:

Built with commit f1cf9959a4694b5fbc1df3861445e4bda0b4194a

https://deploy-preview-621--wonder-blocks.netlify.com

packages/wonder-blocks-data/util/response-cache.js
```diff
@@ -0,0 +1,149 @@
...
+    static get Default() {
+        if (!_default) {
+            _default = new ResponseCache();
+        }
+        return _default;
+    }
+
+    _cache: Cache = {};
+
+    constructor(source?: ?Cache = undefined) {
```
kevinbarabash: I think this can be `source: ?Cache = undefined` since default params imply optionality.

packages/wonder-blocks-data/util/response-cache.js
```diff
@@ -0,0 +1,149 @@
...
+        entry: CacheEntry<TData>,
+    ): CacheEntry<TData> {
+        const requestType = handler.type;
+
+        // Ensure we have a cache location for this handler type.
+        this._cache[requestType] = this._cache[requestType] || {};
+
+        // Cache the data.
+        const key = handler.getKey(options);
+        this._cache[requestType][key] = Object.freeze(entry);
```
kevinbarabash: Since we're freezing the `entry` should we make it's param type read-only?

packages/wonder-blocks-data/util/types.js
```diff
@@ -1,4 +1,24 @@
 // @flow
+export type CacheEntry<TData> =
+    | {|
+          error: string,
+          data?: ?void,
+      |}
+    | {|
+          data: TData,
+          error?: ?void,
+      |};
```
kevinbarabash: This type is a little confusing.  For disjoint unions it's better to have a discriminator field, something like:
```
type CacheEntry<TData> =
  | {| type: "data", data: TData |}
  | {| type: "error", error: string |}
```

packages/wonder-blocks-data/util/response-cache.js
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     * Cache an error for a specific response.
+     */
+    cacheError = <TOptions, TData>(
+        handler: IRequestHandler<TOptions, TData>,
+        options: TOptions,
+        error: Error | string,
+    ): CacheEntry<TData> => {
+        const errorMessage = typeof error === "string" ? error : error.message;
```
kevinbarabash: Why not cache the entire `Error`?  It seems like the stacktrace could come in handy.

packages/wonder-blocks-data/util/response-cache.js
```diff
@@ -0,0 +1,149 @@
...
+        // Get the subcache for the handler.
+        const handlerCache = this._cache[requestType];
+        if (!handlerCache) {
+            return null;
+        }
+
+        const key = handler.getKey(options);
+        if (handler.invalidateCache(options)) {
+            delete handlerCache[key];
+        }
```
kevinbarabash: When would we want `getEntry()` to invalidate the cache?

packages/wonder-blocks-data/util/response-cache.js
```diff
@@ -0,0 +1,149 @@
...
+     */
+    clone = (): $ReadOnly<Cache> => {
+        try {
+            return deepClone(this._cache);
+        } catch (e) {
+            throw new Error(
+                `An error occurred while trying to clone the cache: ${e}`,
+            );
+        }
+    };
```
kevinbarabash: Why don't we wrap the `_cache` in a `ResponseCache` before returning it?

packages/wonder-blocks-data/util/response-cache.test.js
```diff
@@ -0,0 +1,346 @@
...
+    afterEach(() => {
+        /**
+         * This is needed or the JSON.stringify mocks need to be
+         * mockImplementationOnce. This is because if the snapshots need
+         * to update, they write the inline snapshot and that appears to invoke
+         * prettier which in turn, calls JSON.stringify. And if that mock
+         * throws, then boom. No snapshot update and a big old confusing test
+         * failure.
+         */
+        jest.restoreAllMocks();
```
kevinbarabash: Is this something worth adding to the global `afterEach()`?

packages/wonder-blocks-data/util/response-cache.test.js
```diff
@@ -0,0 +1,346 @@
...
+            const sourceData = {
+                MY_HANDLER: {
+                    MY_KEY: {data: "THE_DATA"},
+                },
+            };
+
+            // Act
+            cache.initialize(sourceData);
+            // Try to mutate the cache.
+            sourceData["MY_HANDLER"]["MY_KEY"] = {data: "SOME_NEW_DATA"};
```
kevinbarabash: I guess node isn't in strict mode.

packages/wonder-blocks-data/util/response-cache.js
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        const key = handler.getKey(options);
+        if (handler.invalidateCache(options)) {
+            delete handlerCache[key];
+        }
+
+        // Get the response.
+        const entry = handlerCache[key];
+        return entry == null ? null : entry;
```
jandrade: I wonder if you could just use `return entry;`.... Is the idea here to avoid returning `undefined` or something else?

packages/wonder-blocks-data/util/response-cache.js
```diff
@@ -0,0 +1,149 @@
...
+ */
+let _default: ResponseCache;
+
+/**
+ * Implements the response cache.
+ *
+ * INTERNAL USE ONLY
+ */
+export class ResponseCache {
+    static get Default() {
```
jandrade: nit: I wonder if it's worth to rename this method to `static getInstance()` or `static get instance()`. I know this won't affect the implementation at all and its consumers, b/c it's an internal class. So, please disregard this comment if it doesn't add any value.

packages/wonder-blocks-data/docs.md
```diff
@@ -1 +1,60 @@
...
+
+### Usage
+
+```js static
+initializeCache(sourceCache: ResponseCache): void;
+```
+
+#### Function arguments
+
+| Argument | Flow&nbsp;Type | Default | Description |
```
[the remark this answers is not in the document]
somewhatabstract: I copied this from elsewhere. I presumed past me added it for a reason.

packages/wonder-blocks-data/docs.md
```diff
@@ -1 +1,60 @@
...
+A response cache is divided into subcaches by handler type. An example is
+shown below.
+
+```js static
+const responseCache = {
+    HANDLER_TYPE_A: {
+        DATA_ID_1: "DATA STRING",
+        DATA_ID_2: ["array", "of", "data"],
+    },
+    HANDLER_TYPE_B: {
```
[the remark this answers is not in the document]
somewhatabstract: They'll share the same cache and potentially fulfill each other's requests. That's by design. It makes for a simpler declarative usage rather than having to register handlers.

packages/wonder-blocks-data/util/response-cache.js
somewhatabstract: I tried that and it gives an error.

packages/wonder-blocks-data/util/response-cache.js
somewhatabstract: Because `Error` is not cloneable with our `deepClone` function. We could cache the stacktrace with the message if we want, but caching the `Error` object directly was not going to work.

packages/wonder-blocks-data/util/response-cache.js
somewhatabstract: When a handler says it should. For example, perhaps it's been 1 day since the page was loaded and we do a client-side nav. Maybe for a given piece of data, that means the handler wants to re-request. This gives us the ability to do that.

packages/wonder-blocks-data/util/response-cache.js
somewhatabstract: It is a `ResponseCache`. It's just that we can't have both the `ResponseCache` type and the `ResponseCache` class co-exist with the same name.

We don't need this to be a `ResponseCache` class because that's not the purpose of this method.

packages/wonder-blocks-data/util/response-cache.test.js
somewhatabstract: I don't know and I don't want to do that just in case it causes other problems.

packages/wonder-blocks-data/util/response-cache.test.js
somewhatabstract: Not sure I understand. This is modifying this object:

```
const sourceData = {
                MY_HANDLER: {
                    MY_KEY: {data: "THE_DATA"},
                },
            };
```

Not the actual cache. It's showing that we're are indeed cloning rather than using it directly.

Would strict mode complain about that?

packages/wonder-blocks-data/util/response-cache.js
somewhatabstract: I wanted to be consistent that this returns `null` always.

packages/wonder-blocks-data/util/response-cache.js
somewhatabstract: I wanted this to be clear that it is returning the same instance each time, not a new one. I felt `Default` conveyed that better than `Instance`.

packages/wonder-blocks-data/util/types.js
somewhatabstract: Prettier did that, not me.

packages/wonder-blocks-data/docs.md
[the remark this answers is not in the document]
somewhatabstract: To answer your question, nothing bad will happen.

On the clientside, they would share cache; on the serverside, the first one encountered would fulfill requests for any others of a matching type.

packages/wonder-blocks-data/util/response-cache.js
kevinbarabash: It's a little unexpected that's all.  Usually a `clone()` method on a class would return a new instance of the class with a copy of the data from the original.  That isn't what's happening in this case.  It's fine if that's what we want, but I do think it's worth making note of it in the docstring.

packages/wonder-blocks-data/util/response-cache.test.js
kevinbarabash: For some reason I thought that `sourceData` was frozen.  Maybe it isn't.

PR #626, opened 2019-12-04
WEB-1629.7 Add Data component

somewhatabstract:
This adds the `Data` component that will be used by frontend code wishing to request data.

- [x] Tests need implementing
- [x] Docs to be written


jeresig:

Built with commit 8b30366936f261e4101fe081aa417d1eb3d5bd64

https://deploy-preview-626--wonder-blocks.netlify.com

packages/wonder-blocks-data/util/types.js
```diff
@@ -1,4 +1,16 @@
...
+export type Result<TData> =
+    | {|
+          loading: true,
+          data?: void,
+          error?: void,
+      |}
+    | {|
+          loading: false,
+          data?: TData,
+          error?: string,
```
jeresig: There won't be a case where both error and data are defined, correct?

packages/wonder-blocks-data/components/data.js
```diff
@@ -0,0 +1,187 @@
...
+    _buildStateAndfulfillNeeds(): State<TData> {
+        const propsAtFulfillment = this.props;
+        const {handler, options} = propsAtFulfillment;
+        const {getEntry} = ResponseCache.Default;
+
+        const cachedData = getEntry(handler, options);
+        if (!Server.isServerSide() && cachedData == null) {
+            /**
+             * We're not on the server and the cache missed, or the cache is
+             * invalid.
```
jeresig: If the cache is invalid it feels like the sort of thing we should have logging for - I feel like if we hit this in production it'd be really bad and we'd want to know about it!

packages/wonder-blocks-data/components/data.js
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    render() {
+        const result = this._resultFromState();
+        if (result.loading && Server.isServerSide()) {
+            return this._renderWithTrackingContext(result);
+        }
+
+        return this._renderContent(result);
+    }
```
jeresig: I like how all the pieces come together!

packages/wonder-blocks-data/components/data.js
somewhatabstract: That isn't something we would track here though. This code has no idea which way that is. That's something the `RequestFulfillment` manages. I'll add something there.

packages/wonder-blocks-data/components/data.js
```diff
@@ -0,0 +1,187 @@
...
+                     * We should never get here, but if we do.
+                     */
+                    if (this._mounted && this._propsMatch(propsAtFulfillment)) {
+                        this.setState({
+                            loading: false,
+                            data: null,
+                            error: typeof e === "string" ? e : e.message,
+                        });
+                    }
+                    return null;
```
somewhatabstract: This is where we should probably log something since we really should never see this.

packages/wonder-blocks-data/util/types.js
somewhatabstract: That's correct. I tried typing it to make that clear, but hit confusing flow errors.

packages/wonder-blocks-data/components/track-data.md
```diff
@@ -1,3 +1,208 @@
...
+        super(props);
+        this.state = {};
+    }
+
+    static getDerivedStateFromError(error) {
+        return {error: error.message};
+    }
+
+    render() {
+        if (jest) {
```
jeresig: This fails in the prod site, we'll probably need to do: `typeof jest !== "undefined"`:

https://deploy-preview-626--wonder-blocks.netlify.com/#section-data

packages/wonder-blocks-data/components/track-data.md
```diff
@@ -1,3 +1,208 @@
...
+        window.location.reload();
+    }
+
+    setServerMode() {
+        Server.setServerSide();
+        this.setState({refresh: Date.now(), error: null});
+    };
+
+    _renderErrorOrContent() {
+        if (jest) {
```
jeresig: This also fails.

packages/wonder-blocks-data/components/data.js
```diff
@@ -0,0 +1,187 @@
...
+import {Server} from "@khanacademy/wonder-blocks-core";
+
+import {RequestFulfillment} from "../util/request-fulfillment.js";
+import {ResponseCache} from "../util/response-cache.js";
+import {TrackerContext} from "../util/request-tracking.js";
+
+import type {Result, IRequestHandler} from "../util/types.js";
+
+type Props<TOptions, TData> = {|
+    handler: IRequestHandler<TOptions, TData>,
```
jeresig: Can you add comments for these props? Right now they show up in the documentation but have no description.

packages/wonder-blocks-data/components/track-data.md
```diff
@@ -1,3 +1,208 @@
...
+import {Body, BodyMonospace} from "@khanacademy/wonder-blocks-typography";
+import {Strut} from "@khanacademy/wonder-blocks-layout";
+import Spacing from "@khanacademy/wonder-blocks-spacing";
+import Button from "@khanacademy/wonder-blocks-button";
+import {Server, View} from "@khanacademy/wonder-blocks-core";
+import {Data, TrackData, RequestHandler, fulfillAllDataRequests} from "@khanacademy/wonder-blocks-data";
+
+class MyPretendHandler extends RequestHandler {
+    constructor() {
+        super("MY_PRETEND_HANDLER");
```
jandrade: Is there a mechanism to verify the handlers are unique? what would happen if two different handlers happen to have the same type?

packages/wonder-blocks-data/components/track-data.md
somewhatabstract: Oh darn! Thanks. This was the very last thing I added too, and I neglected to check that. Thanks for taking a look. I'll update and test it.

packages/wonder-blocks-data/components/data.js
somewhatabstract: Yes

packages/wonder-blocks-data/components/track-data.js
```diff
@@ -12,17 +12,13 @@ type TrackDataProps = {|
  * Component to enable data request tracking when server-side rendering.
  */
 export default class TrackData extends React.Component<TrackDataProps> {
-    constructor(props: TrackDataProps) {
-        super(props);
-
+    render() {
         if (!Server.isServerSide()) {
```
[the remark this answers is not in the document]
somewhatabstract: To make it possible to have the example docs.

packages/wonder-blocks-data/components/track-data.md
somewhatabstract: This is mentioned in the ADR. If there are two with the same type, they will:

1. Share cache
2. When resolving on the server, the first one encountered will handle the requests.

We don't do anything to verify uniqueness. That's on the consumers of the API. In reality I don't expect any issues and this makes a much nicer API as we don't have to have some sort of "register" step for handlers.

packages/wonder-blocks-data/components/data.js
somewhatabstract: I've clarified what this comment means.

packages/wonder-blocks-data/docs.md
```diff
@@ -39,7 +54,7 @@ type CacheEntry =
         data: any,
       |}
     | {|
-        error: string,
+        error: Error,
```
somewhatabstract: Darn it, accidentally undid this fix.

packages/wonder-blocks-data/docs.md
```diff
@@ -59,7 +74,7 @@ shown below.
 ```js static
 const responseCache = {
     HANDLER_TYPE_A: {
-        DATA_ID_1: {error: "It go 💥boom 😢"},
+        DATA_ID_1: {error: new Error("It go 💥boom 😢")},
```
somewhatabstract: Darn it, accidentally undid this one too!

somewhatabstract: OK, fixed it here: https://github.com/Khan/wonder-blocks/commit/8fc15f7964b96cccd35fc36c2a59f4f17dc72863

packages/wonder-blocks-data/docs.md
somewhatabstract: OK, fixed it here: https://github.com/Khan/wonder-blocks/commit/8fc15f7964b96cccd35fc36c2a59f4f17dc72863

PR #631, opened 2019-12-06
Add flow lint to require explicit exactness on all object types

kevinbarabash:
Flow is moving towards a model where all object types are exact by default.  Before we can move there though, we need to make all of our object types either explicitly exact or explicitly inexact.  This provides a stepping stone towards exact by default.  Before we can make that jump though we'll also need to make all object types in webapp explicitly exact/inexact as well.

A couple of things to note about this diff:
- anything with an indexer must be inexact
- most of the time objects without properties should be inexact except in the case where we want to disallow any props such as in the case of `Spring`.

I made most of the `children` render props inexact.  In retrospect they should be exact.  I'll go back and fix that.


jeresig:

Built with commit d454c78dd924f05e1b7f03520a712477acdb41d7

https://deploy-preview-631--wonder-blocks.netlify.com

packages/wonder-blocks-layout/components/media-layout.js
```diff
@@ -37,11 +38,11 @@ type Props = {|
      *                  stylesheet for this mediaSize (as specified in the
      *                  styleSheets prop).
      */
-    children: ({
+    children: ({|
```
jeresig: You moved `ChildrenArgs` out above, do you want to do that here, too? I guess I like leaving them inline as then it's more obvious what the args are inside the Styleguidist docs.

packages/wonder-blocks-modal/components/modal-launcher.js
```diff
@@ -74,7 +74,7 @@ type ControlledProps = {|
 
 type UncontrolledProps = {|
     ...CommonProps,
-    children: ({openModal: () => mixed}) => React.Node,
+    children: ({openModal: () => mixed, ...}) => React.Node,
```
jeresig: It feels like all these in this file should be strict, right? We're only passing in one argument. I see you're doing this in a number of files, any particular reason?

packages/wonder-blocks-layout/components/media-layout.js
kevinbarabash: VSCode got confused.  I've reverted this change.

packages/wonder-blocks-modal/components/modal-launcher.js
kevinbarabash: Initially I thought that I'd keep the current types and just make things explicit.  I've had a change of heart and decided to make everything exact that could while making non-exact things explicit.

PR #648, opened 2019-12-13
WB-660.1: Add custom opener to ActionMenu

jandrade:
## Summary

- Added a custom opener to the `ActionMenu` component.
- Created `DropdownOpener` to be reused across the multiple components.
- Added new example.

Screen Shot 2019-12-13 at 4 00 25 PM

**Note:** This is the first part of this work:

1. Add custom opener to ActionMenu **[current]**.
2. Add custom opener to SingleSelect.
3. Add custom opener to MultiSelect.

### DIFF
--- a/packages/wonder-blocks-dropdown/components/action-menu.js
+++ b/packages/wonder-blocks-dropdown/components/action-menu.js
@@ -230,7 +230,7 @@ export default class ActionMenu extends React.Component<Props, State> {
                 onClick={this.handleClick}
                 disabled={numItems === 0 || disabled}
                 anchorRef={this.onHandleOpenerRef}
-                testId={testId}
+                testId={opener ? undefined : testId}
             >
                 {opener
                     ? opener
--- a/packages/wonder-blocks-dropdown/components/action-menu.js
+++ b/packages/wonder-blocks-dropdown/components/action-menu.js
@@ -1,11 +1,20 @@
 // @flow
 
 import * as React from "react";
+import ReactDOM from "react-dom";
 import {StyleSheet} from "aphrodite";
-import type {AriaProps, StyleType} from "@khanacademy/wonder-blocks-core";
-import Dropdown from "./dropdown.js";
+import type {
+    AriaProps,
+    ClickableState,
+    StyleType,
+} from "@khanacademy/wonder-blocks-core";
+import DropdownOpener from "./dropdown-opener.js";
+import ActionItem from "./action-item.js";
+import OptionItem from "./option-item.js";
+import DropdownCore from "./dropdown-core.js";
+
 import ActionMenuOpenerCore from "./action-menu-opener-core.js";
-import type {Item} from "../util/types.js";
+import type {Item, DropdownItem} from "../util/types.js";
 
 type Props = {|
     ...AriaProps,
@@ -70,6 +79,13 @@ type Props = {|
      * Optional styling for the entire dropdown component.
      */
     style?: StyleType,
+
+    /**
+     * The child function that returns the anchor the ActionMenu will be
+     * activated by. This function takes eventState, which allows the opener
+     * element to access pointer event state.
+     */
+    opener?: (eventState: ClickableState) => React.Element<any>,
 |};
 
 type State = {|
@@ -95,6 +111,7 @@ export default class ActionMenu extends React.Component<Props, State> {
     };
 
     state = {
+        keyboard: false,
         opened: false,
     };
 
@@ -109,13 +126,20 @@ export default class ActionMenu extends React.Component<Props, State> {
         };
     }
 
-    /**
-     * Update the internal state and notify the parent component when menu is
-     * toggled
-     */
-    handleToggleMenu = (opened: boolean) => {
+    handleItemSelected = () => {
+        // close menu
+        this.handleOpenChanged(false);
+
+        // Bring focus back to the opener element.
+        if (this.openerElement) {
+            this.openerElement.focus();
+        }
+    };
+
+    handleOpenChanged = (opened: boolean, keyboard?: boolean) => {
         this.setState({
-            opened: opened,
+            opened,
+            keyboard,
         });
 
         if (this.props.onToggle) {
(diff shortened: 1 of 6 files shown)


packages/wonder-blocks-dropdown/components/action-menu.md
```diff
@@ -267,4 +267,41 @@ class ControlledActionMenuExample extends React.Component {
...
+
+<ActionMenu
+    disabled={false}
+    menuText="Betsy Appleseed"
+    testId="teacher-menu"
+    opener={(eventState) => (
+        <LabelMedium
+            onClick={()=>{console.log('custom click!!!!!')}}
+        >This is a label</LabelMedium>
+    )}
```
kevinbarabash: I think it would be good too document the `eventState` or cross-link to other docs if they exist.  Also, it would be good to make use of it here to show different states in this example.

packages/wonder-blocks-dropdown/components/action-menu.test.js
```diff
@@ -39,22 +34,116 @@ describe("ActionMenu", () => {
...
-        // Close menu with space
-        opener.simulate("keydown", {keyCode: keyCodes.space});
-        opener.simulate("keyup", {keyCode: keyCodes.space});
-        expect(menu.find(Dropdown).state("opened")).toEqual(false);
+        // Assert
+        expect(menu.state("opened")).toBe(true);
+    });
 
-        // Open menu again with enter
+    it("opens the menu on enter", () => {
```
kevinbarabash: Thanks for splitting these tests up.

packages/wonder-blocks-dropdown/components/action-menu.test.js
```diff
@@ -39,22 +34,116 @@ describe("ActionMenu", () => {
...
+                <OptionItem label="Toggle" value="toggle" onClick={onToggle} />
+            </ActionMenu>,
+        );
+
+        // Act
+        const opener = menu.find(DropdownOpener);
+        // open using the mouse
+        opener.simulate("click");
+        // trigger body click
+        document.dispatchEvent(new MouseEvent("mouseup"));
```
kevinbarabash: I'm surprised we didn't have a test for this already.  Thanks for adding one.

packages/wonder-blocks-dropdown/components/action-menu.md
```diff
@@ -267,4 +267,41 @@ class ControlledActionMenuExample extends React.Component {
...
+    <SeparatorItem />
+    <ActionItem label="Log out" href="http://khanacademy.org/logout" testId="logout" />
+    <OptionItem
+        label="Show homework assignments" value="homework"
+        onClick={() => console.log(`Show homework assignments toggled`)}
+    />
+    <OptionItem
+        label="Show in-class assignments" value="in-class"
+        onClick={() => console.log(`Show in-class assignments toggled`)}
+    />
```
kevinbarabash: Is there a way to adjust the position of the dropdown relative to the opener?

packages/wonder-blocks-dropdown/components/action-menu.test.js
```diff
@@ -331,4 +420,56 @@ describe("ActionMenu", () => {
...
+            // Arrange
+            const menu = mount(
+                <ActionMenu
+                    menuText={"Action menu!"}
+                    testId="openTest"
+                    onChange={onChange}
+                    selectedValues={[]}
+                    opener={(eventState) => (
+                        <button aria-label="Search" onClick={jest.fn()} />
+                    )}
```
kevinbarabash: We probably don't need to test the values of `eventState` because they're being passed directly from `ClickableBehavior`, right?

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -109,69 +126,153 @@ export default class ActionMenu extends React.Component<Props, State> {
...
+    renderOpener(numItems: number) {
+        const {disabled, menuText, opened, opener, testId} = this.props;
+
+        return (
+            <DropdownOpener
+                onClick={this.handleClick}
+                disabled={numItems === 0 || disabled}
+                anchorRef={this.onHandleOpenerRef}
+                testId={testId}
+            >
```
kevinbarabash: It's nice to see we can implement the existing dropdown components using this.

packages/wonder-blocks-dropdown/components/dropdown-opener.js
```diff
@@ -0,0 +1,90 @@
...
+     */
+    testId?: string,
+
+    /**
+     * Callback used to pass the childs Ref back up to the parent element
+     */
+    anchorRef: (?HTMLElement) => mixed,
+|};
+
+class DropdownOpener extends React.Component<Props> {
```
kevinbarabash: Would we ever want to make this public?

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -1,11 +1,20 @@
 // @flow
 
 import * as React from "react";
+import ReactDOM from "react-dom";
 import {StyleSheet} from "aphrodite";
-import type {AriaProps, StyleType} from "@khanacademy/wonder-blocks-core";
-import Dropdown from "./dropdown.js";
```
kevinbarabash: Now that we can customize existing ActionMenu, SingleSelect, and MultiSelect, does this negate the need for Dropdown?  Can we deprecate it?

packages/wonder-blocks-dropdown/components/dropdown-opener.js
```diff
@@ -0,0 +1,90 @@
...
+        this.props.anchorRef(anchorNode);
+    }
+
+    renderAnchorChildren(
+        eventState: ClickableState,
+        handlers: ClickableHandlers,
+    ) {
+        const renderedChildren = this.props.children(eventState);
+
+        return React.cloneElement(renderedChildren, {
```
somewhatabstract: Why do we clone the children?

somewhatabstract: Oh, it's so we can add these extra attributes. Interesting.

jandrade: exactly! this is to make sure we pass the right props to the child component.

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -109,69 +126,153 @@ export default class ActionMenu extends React.Component<Props, State> {
...
+            ];
+            onChange(updatedSelection);
+        } else {
+            // Item was newly selected
+            onChange([...selectedValues, selectedValue]);
+        }
+        this.handleItemSelected();
+    };
+
+    getMenuItems(): Array<DropdownItem> {
```
somewhatabstract: Is this method copied over from the `Dropdown` component?

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -109,69 +126,153 @@ export default class ActionMenu extends React.Component<Props, State> {
...
+            opened,
+            keyboard,
         });
 
         if (this.props.onToggle) {
             this.props.onToggle(opened);
         }
     };
 
+    handleOptionSelected = (selectedValue: string) => {
```
kevinbarabash: It looks like most of the changes in this file were from copy/pasting from dropdown.js, is that correct?

packages/wonder-blocks-dropdown/components/dropdown-opener.js
```diff
@@ -0,0 +1,90 @@
...
+     */
+    onClick: (e: SyntheticEvent<>) => mixed,
+
+    /**
+     * Test ID used for e2e testing.
+     */
+    testId?: string,
+
+    /**
+     * Callback used to pass the childs Ref back up to the parent element
```
kevinbarabash: typo: `the child's Ref`

packages/wonder-blocks-dropdown/components/dropdown-opener.js
```diff
@@ -0,0 +1,90 @@
...
+|};
+
+class DropdownOpener extends React.Component<Props> {
+    static defaultProps = {
+        disabled: false,
+    };
+
+    componentDidMount() {
+        const anchorNode = ((ReactDOM.findDOMNode(this): any): ?HTMLElement);
+        this.props.anchorRef(anchorNode);
```
kevinbarabash: We have to do this because `ClickableBehavior` doesn't support ref forwarding, right?  It might be good to point a comment here.  Also, sometime next year it would good to add ref forwarding to all of the wonder-blocks components.

packages/wonder-blocks-dropdown/components/action-menu.js
jandrade: Yes, now `ActionMenu` does with `Dropdown` was doing, so I decided to copy this logic from `dropdown.js`.

packages/wonder-blocks-dropdown/components/dropdown-opener.js
```diff
@@ -0,0 +1,90 @@
...
+    renderAnchorChildren(
+        eventState: ClickableState,
+        handlers: ClickableHandlers,
+    ) {
+        const renderedChildren = this.props.children(eventState);
+
+        return React.cloneElement(renderedChildren, {
+            ...handlers,
+            disabled: this.props.disabled,
+            "data-test-id": this.props.testId,
```
kevinbarabash: What happens if `children` returns an element that doesn't respond to `data-test-id`?  I feel like `ActionMenu`, `SingleSelect`, and `MultiSelect` should each pass their `testId` prop down to their own openers and we should probably remove the `testId` from `DropdownOpener`.  If people need a `testId` on their custom openers they'll need to add it themselves.

packages/wonder-blocks-dropdown/components/action-menu.js
jandrade: Yes, will do :). This is going to be done in a separate PR/ticket: https://khanacademy.atlassian.net/browse/WB-662

packages/wonder-blocks-dropdown/components/action-menu.test.js
jandrade: Right, and this should be already tested in `ClickableBehavior`.

packages/wonder-blocks-dropdown/components/dropdown-opener.js
jandrade: Maybe in the future, but for now I think it's better to keep this isolated from Webapp and encourage folks to use our current variants (ActionMenu, SingleSelect and MultiSelect).

packages/wonder-blocks-dropdown/components/dropdown-opener.js
jandrade: Exactly, I'm going to add the comment in there. 1000% to add ref forwarding next year.

packages/wonder-blocks-dropdown/components/dropdown-opener.js
jandrade: Please correct me if I'm wrong, but I think any children should respond to `data-test-id` as we are directly setting the DOM attribute instead of a React prop. Do you have an example in mind?

Also, in that case, we could have an issue when we use the default internal opener. We wouldn't have a way to pass the `testId` to the opener DOM element.

kevinbarabash: I don't think that we known that `renderedChildren` is a DOM node, do we?  If we use the default opener, `ActionMenu` can pass `testId` as a prop to `ActionMenuOpener` which has a `testId` prop.

packages/wonder-blocks-dropdown/components/action-menu.md
jandrade: Not sure if I follow, do you mean something like adding `padding` or `margin` to the opener?

packages/wonder-blocks-dropdown/components/action-menu.js
jandrade: Yes, I'm trying to transfer the good parts from dropdown to action-menu. Then, I'll delete dropdown.js in an upcoming PR (different ticket, tho).

packages/wonder-blocks-dropdown/components/dropdown-opener.js
jandrade: My only concern there is that people might get confused now that we have two different entrypoints for the same `testId` element (dropdown opener element). But your point is completely valid. I'm going to implement the change and we'll see how it will work when using it in different scenarios :).

packages/wonder-blocks-dropdown/components/action-menu.md
kevinbarabash: Or on the dropdown itself since vertical padding/margin might affect the position of the opener itself.

packages/wonder-blocks-dropdown/components/action-menu.md
```diff
@@ -267,4 +267,70 @@ class ControlledActionMenuExample extends React.Component {
...
+    menuText="Betsy Appleseed"
+    opener={(eventState) => (
+        <LabelMedium
+            onClick={()=>{console.log('custom click!!!!!')}}
+            testId="teacher-menu-custom-opener"
+            style={[
+                eventState.focused && styles.focused,
+                eventState.hovered && styles.hovered,
+                eventState.pressed && styles.pressed
+            ]}
```
kevinbarabash: Thanks for using `eventState`.

packages/wonder-blocks-dropdown/components/dropdown-opener.js
kevinbarabash: That's a valid concern.  We could require that openers have a `testId` prop.

PR #657, opened 2019-12-16
WB-660.4: Add text argument to the new opener prop

jandrade:
## Summary

- Fixed `testId` avoid using it from the parent component (`ActionMenu, SingleSelect and MultiSelect`).

- Added a new `text` argument as part of the `opener` prop (`ActionMenu, SingleSelect and MultiSelect`). This will allow people to optionally re-use the dropdown's current label in case they need it.

### Before
```
<MultiSelect
  opener={(eventState) => (
    <LabelLarge styles={[eventState.hovered && styles.customHoveredStyles]}>
        Hard-coded label for opener
    </LabelLarge>
  )}
/>
```
custom-opener-before

### After
```
<MultiSelect
  opener={(eventState, text) => (
    <LabelLarge styles={[eventState.hovered && styles.customHoveredStyles]}>
        {text}
    </LabelLarge>
  )}
/>
```
custom-opener-after

### DIFF
--- a/packages/wonder-blocks-dropdown/components/dropdown-opener.js
+++ b/packages/wonder-blocks-dropdown/components/dropdown-opener.js
@@ -1,0 +1,90 @@
+// @flow
+import * as React from "react";
+import * as ReactDOM from "react-dom";
+
+import {ClickableBehavior} from "@khanacademy/wonder-blocks-core";
+
+import type {
+    AriaProps,
+    ClickableHandlers,
+    ClickableState,
+} from "@khanacademy/wonder-blocks-core";
+
+type Props = {|
+    ...$Rest<AriaProps, {|"aria-disabled": "true" | "false" | void|}>,
+
+    /**
+     * The child function that returns the anchor the Dropdown will be activated
+     * by. This function takes `eventState`, which allows the opener element to
+     * access pointer event state.
+     */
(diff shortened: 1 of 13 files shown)


packages/wonder-blocks-dropdown/components/dropdown-opener.js
```diff
@@ -33,32 +35,46 @@ type Props = {|
...
                       // custom onClick handler inside the
                       // children node
-                      renderedChildren.props.onClick(e);
+                      childrenProps.onClick(e);
                       handlers.onClick(e);
                   }
                 : handlers.onClick,
+            // try to get the testId from the child element
+            // If it's not set, try to fallback to the parent's testId
+            "data-test-id": childrenTestId || testId,
```
jandrade: @kevinbarabash I've added this extra check to get the `testId` or the `data-test-id` value from the opener element. There's an special case (`ActionMenu, default opener`) where we need to pass the `testId` from the parent element. In the other cases, if we don't set the ID from the custom one, we won't be passing any `testId` to the cloned element at all.

jandrade:
@jangmi-jo We disabled that check temporarily because netlify started charging for that (Due to all the dependabot PRs). We need to come with a better strategy to only deploy certain PRs.

PR #659, opened 2019-12-17
WB-483: Add custom labels to MultiSelect

jandrade:
### MultiSelect
Added the `labels` prop to the `MultiSelect` component that allows us to pass a set of custom labels to the component.

dropdown-translated

This is the set of keys allowed by the component:

```
labels={
    clearSearch: string,
    filter: string,
    noResults: string,
    selectAllLabel: (numOptions: number) => string,
    selectNoneLabel: string,
    noneSelected: string,
    someSelected: (numSelectedValues: number) => string,
    allSelected: string,
}
```

## Test plan

1. Run tests: `yarn test`.
2. Open the example that uses a set of translated strings: http://localhost:6060/#!/MultiSelect/21.

cc @annr


packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -23,6 +23,17 @@ import {
 import typeof OptionItem from "./option-item.js";
 import type {DropdownItem} from "../util/types.js";
 
+type TranslatedLabels = {|
```
jeresig: I'd like to propose an alternative to what you've done here. What about if we just call this "labels" and make it so that people can use it for translations or just changing the text to something else.

Then it'd make the code a lot cleaner if we put all the strings in one location. For example in the constructor you can do something like: `Object.assign(labels, {clearSearch: "Clear search", ...}, props.labels)` - then later on you can just use the label directly.

Finally, if you do the above, you can make the type for the labels not only strict but also have all the labels be required: `type Labels = {| clearSearch: string, ... |};` In order to provide a property for the user to specify labels you can do: `$Shape<Labels>`, which means that they don't have to provide them all, just some of them.

jandrade: That makes a lot of sense to me. It feels more generic. Thanks for the suggestion!

jandrade: @jeresig As I'm making these changes, there's only one thing to have noticed. In the following scenario, we would need to keep in sync the default values:

### No custom `labels` provided

There will be cases where folks won't pass any custom labels to the `MultiSelect` component. When this happens, and due that some keys use dynamic values (number of total items, number of selected items, the selected item type has changed), we will need to keep them in sync.

These are the cases:

```
selectAllLabel: `Select all (${numOptions})`,
noneSelected: `0 ${props.selectItemType}`,
someSelected: `${props.selectedValues.length} ${props.selectItemType}`,
allSelected: `All ${props.selectItemType}`,
```

Maybe `selectItemType` won't change, but it can be a possibility. For the other cases

### Possible solution
We can achieve it by getting the computed values in `render`, `getDerivedStateFromProps` or `componentDidUpdate` (by checking against the prev and current values of each key):

e.g.
```
// if the user didn't provide any custom labels, then we need to compute the
current value internally, causing some extra re-renders
if (
    numCurrentOptions !== numPrevOptions &&
    !this.props.labels.selectAllLabel
) {
    this.setState({
        labels: {
            ...this.state.labels,
            selectAllLabel: `Select all (${numOptions})`,
        },
    });
}
```

What do you think about this case? any suggestions? Maybe we could still set the static labels in the constructor?

jeresig: @jandrade So this is highlighting a problem with the implementation of this, as it stands - if someone wants to implement a replacement for the selectAll label - how would that work? They would need access to the `numOptions` as well. I think the label definitions would have to look like this:

```
selectAll: ({numOptions: number, itemType: string}) => string,
allSelected: ({numOptions: number, itemType: string}) => string,
...
```

We could pass in any other properties that we'd expect people to use, as well. But this way they will have full control over the strings and how they are rendered and what is included. It also means that we can safely move the default values outside of the inline place where they're located.

jandrade: Got it. I was checking the translations and we are using a similar shape `(numOptions: number) => string`. I think that we could even get rid of the `itemType` prop.

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -16,72 +16,112 @@ import SearchTextInput from "./search-text-input.js";
...
+
+type Props = {|
+    ...AriaProps,
+
+    ...DefaultProps,
+
+    /**
+     * The items in this select.
      */
-    selectItemType: string,
```
jandrade: I'm deprecating these two props in favor of `labels`.

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -184,6 +225,8 @@ export default class MultiSelect extends React.Component<Props, State> {
             searchText: "",
             lastSelectedValues: [],
         };
+        // merge custom labels with the default ones
+        this.labels = {...defaultLabels, ...props.labels};
```
jandrade: I ended up creating this `labels` instance variable to merge the custom labels with the default labels defined in `constants.js`.

packages/wonder-blocks-dropdown/util/constants.js
```diff
@@ -34,3 +34,15 @@ export const SEPARATOR_ITEM_HEIGHT = 9;
 
 export const SEARCH_ITEM_HEIGHT =
     DROPDOWN_ITEM_HEIGHT + searchInputStyle.margin + searchInputStyle.marginTop;
+
+// The default labels that will be used by different components
+export const defaultLabels = {
```
jandrade: Now all the default labels are centralized in this file.

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -16,72 +16,112 @@ import SearchTextInput from "./search-text-input.js";
...
+     * guarantee that the ID is unique within everything rendered on a page.
+     * Used to match `<label>` with `<button>` elements for screenreaders.
+     */
+    id?: string,
 
     /**
-     * When this is true, the menu text shows "All {selectItemType}" when no
-     * item is selected.
+     * When this is true, the menu text shows "All items" when no item is
+     * selected.
```
jeresig: Might be worth mentioning which label this is, as well

packages/wonder-blocks-dropdown/components/multi-select.js
jeresig: Will this update if the props change later on?

packages/wonder-blocks-dropdown/components/multi-select.js
```diff
@@ -238,22 +281,13 @@ export default class MultiSelect extends React.Component<Props, State> {
...
-            implicitAllEnabled,
-            selectItemType,
-            selectedValues,
-        } = this.props;
-        const allSelectedText = `All ${selectItemType}`;
+        const {implicitAllEnabled, selectedValues} = this.props;
+        const {noneSelected, someSelected, allSelected} = this.labels;
 
-        // When implicit all enabled, use the allSelectedText when no selection
+        // When implicit all enabled, use the allSelected when no selection
```
jeresig: *use allSelected

packages/wonder-blocks-dropdown/components/search-text-input.js
```diff
@@ -60,6 +70,13 @@ export default class SearchTextInput extends React.Component<Props, State> {
         );
     }
 
+    static defaultProps = {
+        labels: {
+            clearSearch: defaultLabels.clearSearch,
+            filter: defaultLabels.filter,
+        },
+    };
```
jeresig: I notice you're not doing this `this.labels` here, any reason?

packages/wonder-blocks-dropdown/util/constants.js
jeresig: Very nice!

packages/wonder-blocks-dropdown/components/multi-select.js
jandrade: mmm It won't. I'll make the pertinent changes.

packages/wonder-blocks-dropdown/components/search-text-input.js
jandrade: Yes, this component is only accessed by `MultiSelect` and we are doing the corresponding `this.labels` check in that component. In this case, we have two options:

- No custom labels set: It will use the values from `defaultValues`.
- Custom labels set: It will override the values of `this.props.labels` with the values passed from `MultiSelect`.

jandrade: Disregard my previous comment. We still need to sync the `labels` sent by the parent component, so I might to reuse the same logic to get the current `labels` from `MultiSelect`.

PR #686, opened 2020-01-11
Remove last vestiges of vite

kevinbarabash:
vite was an experimental visual regression testing tool that I was working on.  I've been experimenting with chromatic in a different project and I'm pretty happy with it.  I think we should give it a try in wonder-blocks so that we can avoid regressions like the one we had for ellipsized labels in buttons a while ago.  Adopting chromatic will happen in a future PR.

Test Plan: n/a, the fact that the tests run on C/I successfully should be sufficient testing


package.json
```diff
@@ -76,7 +72,7 @@
     "jest": "^24.8.0",
     "jest-enzyme": "^7.0.2",
     "lerna": "^3.16.5",
-    "marked": "^0.8.0",
+    "marked": "0.7.0",
```
kevinbarabash: I pinned this to the previous version since the new version has engines changed to `">= 8.16.2"`.  There doesn't seem to be a good reason this deep specify this value and I didn't want to change my version of node just to make this dep happy.

PR #687, opened 2020-01-12
Set up Storybook and write some button stories

kevinbarabash:
This is in preparation for testing out chromatic for visual regression testing of our components.  The more stories we have the more snapshots chromatic will take which will take longer during builds.  This is why some of the stories group buttons together.

Test Plan:
- yarn start:storybook
- open http://localhost:53724/
- see the button stories
- yarn build:storybook
- see that storybook-static has been created
- http-server storybook-static
- open http://localhost:8080/, see that this storybook instance serves the same content

Screen Shot 2020-01-11 at 7 37 19 PM
Screen Shot 2020-01-11 at 7 37 29 PM


packages/wonder-blocks-button/components/button.stories.js
```diff
@@ -0,0 +1,121 @@
...
+            <Button light={true} kind="tertiary" color="destructive">
+                Hello, world!
+            </Button>
+        </View>
+    </View>
+);
+
+darkBackgroundButtons.story = {
+    parameters: {
+        backgrounds: [{name: "red", value: Color.darkBlue, default: true}],
```
jandrade: I'm curious, why the name "red"? Does Storybook require a key name for that object?

kevinbarabash: I copy/pasted this code from an example without changing all of the things I needed to.

PR #688, opened 2020-01-13
Set up Chromatic visual regression testing too

kevinbarabash:
This diff sets up Chromatic using Github Actions.  Although we're using CircleCI for most of our CI related needs, I think that we want to use Github Actions for more things and this is a move in that direction.

In addition to Chromatic taking image snapshots of the components, it also deploys a version of the components and their stories to its library.  You can view the library for each build so this could potentially replace the netlify builds, but we'd have to have more stories in storybook, not just the ones where we want to avoid a regression.

Test Plan: 
- See that Chromatic check appears and that the screenshots match the current reference images


package.json
```diff
@@ -9,7 +9,7 @@
     "build:all": "yarn build:cjs && yarn build:es6",
     "build:cjs": "webpack --config build-settings/webpack.config.js --display-modules && node utils/package-flow-types.js",
     "build:es6": "rollup -c build-settings/rollup.config.js",
-    "build:storybook": "build-storybook",
+    "build-storybook": "build-storybook",
```
kevinbarabash: Chromatic expects this to be the name of the build script.  I'm sure there's a way to configure this.  I just wanted to get things working first before making that change.

kevinbarabash:
There doesn't seem to be a way to review the original snapshot.  If there are changes to the images then you must review them in Chromatic before the chromatic check in the GitHub PR will turn green.  I'll make a change to one of the stories to illustrate this workflow.

kevinbarabash:
Here's a screenshot of what the review workflow looks like in Chromatic.
Screen Shot 2020-01-13 at 10 53 52 AM

packages/wonder-blocks-button/components/button.stories.js
```diff
@@ -45,7 +45,7 @@ export const basicButtons = () => (
 );
 
 export const darkBackgroundButtons = () => (
-    <View>
+    <View style={{backgroundColor: Color.darkBlue}}>
```
dierat: Just curious - why this change? Just happened to notice it looked better?

kevinbarabash: The screenshots don't include the background color from the background addon so adding the background color here is necessary for it show up in the screenshot.  Also, making this change also shows off how the review workflow in the Chromatic app when there are changes to the images snapshots.

PR #699, opened 2020-01-14
Add stories for modals

kevinbarabash:
There are no changes to the components themselves.  This PR only adds some storybook stories for modals so that we can some screenshots to avoid visual regressions.  It uses chromatic's `viewport` option to specify a three different browser widths to take screenshots at.  This PR also experiments with the viewport-addon for storybook which allows you to interactively select between different browser dimensions when running storybook locally which is more convenient then using the browsers built-in viewport emulation with styleguidist.

Test Plan:
- yarn start:storybook
- open one of the modal stories
- see that the viewport icon appears and that you can switch between viewports
- see that 3 snapshots were take per modal story for a total of 6 snapshots in Chromatic

**"simple" story on local storybook instance**
Screen Shot 2020-01-13 at 11 25 12 PM

**chromatic snapshot of "kitchenSink" story"**
Screen Shot 2020-01-13 at 11 57 35 PM


packages/wonder-blocks-modal/components/one-pane-dialog.stories.js
```diff
@@ -0,0 +1,116 @@
...
+        <ModalLauncher
+            modal={modal}
+            testId="modal-launcher-default-example"
+            opened={true}
+            onClose={() => alert("This would close the modal.")}
+        />
+    );
+};
+
+export const headerAndFooter = () => {
```
kevinbarabash: I'm going to rename this to `kitchenSink` once I figure out how to load the images for `above` and `below` that we have on one of the styleguidist examples.

package.json
```diff
@@ -25,7 +25,7 @@
     "pretty-quick": "pretty-quick --staged",
     "publish": "node utils/pre-publish-check.js && git diff --stat --exit-code HEAD && yarn build:cjs && yarn build:es6 && lerna publish",
     "start:prod": "yarn run build:styleguidist && yarn run styleguidist --config styleguide.prod.config.js",
-    "start:storybook": "start-storybook",
+    "start-storybook": "start-storybook -s ./static",
```
kevinbarabash: I ended up changing this command to match `build-storybook`.  I was getting confused when I was running `yarn start-storybook` before this change since it was running `start-storybook` with out the `-s` option.

packages/wonder-blocks-modal/components/one-pane-dialog.stories.js
```diff
@@ -0,0 +1,203 @@
...
+            modal={modal}
+            testId="modal-launcher-default-example"
+            opened={true}
+            onClose={() => alert("This would close the modal.")}
+        />
+    );
+};
+
+export const withOpener = () => {
+    type MyModalProps = {|
```
jandrade: This is a major selling point. Having the ability of adding types to our code examples!

packages/wonder-blocks-modal/components/one-pane-dialog.stories.js
```diff
@@ -0,0 +1,203 @@
...
+};
+
+withOpener.story = {
+    parameters: {
+        viewport: {
+            defaultViewport: null,
+        },
+        chromatic: {
+            // Don't take screenshots of this story
+            disable: true,
```
jandrade: Nice! I'm curious to know why we don't want to take screenshots here?

kevinbarabash: Because we'd only see a button which isn't very useful.  I'll update the comment to explain why.

kevinbarabash: Because we'd only see a button which isn't very useful.  I'll update the comment to explain why.

PR #719, opened 2020-01-16
Add actions, knobs, and jsx storybook addons

kevinbarabash:
This PR adds some more addons to our storybook setup.  In particular we use the actions, knobs, and jsx addons in the new "Buttons With Knobs" story.  The addon panel is hidden for other stories because it's not super useful for those stories.  In the future we may want to make a "With Knobs" story for OnePaneDialog.

The JSX addon  generates JSX code based on the current state of the component in the canvas.  This means that changes to the component made by knobs will be reflected in the JSX tab.  It even shows values for props that haven't been explicitly set such as `spinner`.  It also shows a whole bunch of code for the `onClick` handled that was added for the Actions tab.

Screen Shot 2020-01-15 at 11 38 25 PM

Screen Shot 2020-01-15 at 11 37 35 PM


packages/wonder-blocks-button/components/button.stories.js
```diff
@@ -1,12 +1,80 @@
...
 import {View} from "@khanacademy/wonder-blocks-core";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
 import Color from "@khanacademy/wonder-blocks-color";
 
 import Button from "./button.js";
 
-export default {title: "Button"};
+export default {
+    title: "Button",
+    decorators: [withKnobs, jsxDecorator],
```
jandrade: These are candidates to be used globally. Maybe you could add these decorators to the global config instead.

See: https://storybook.js.org/docs/addons/introduction/#storybook-decorators

packages/wonder-blocks-button/components/button.stories.js
```diff
@@ -1,12 +1,80 @@
...
+
+    parameters: {
+        options: {
+            showAddonPanel: false,
```
jandrade: Same with this option, it could be defined globally instead of defined on every file.

PR #726, opened 2020-01-16
WB-835: Update withActionScheduler to match webapp's HOC pattern

kevinbarabash:
Using this pattern adds support for default props and ref forwarding.  This will allow us to compose `withActionScheduler` with the HOCs defined in webapp.  It also means that all of our HOCs from wonder-blocks and webapp will have a consistent API as well as consistent behavior.

### DIFF
--- a/packages/wonder-blocks-timing/components/with-action-scheduler.test.js
+++ b/packages/wonder-blocks-timing/components/with-action-scheduler.test.js
@@ -5,6 +5,8 @@ import {mount, unmountAll} from "../../../utils/testing/mount.js";
 
 import withActionScheduler from "./with-action-scheduler.js";
 
+import type {WithActionSchedulerProps} from "../util/types.js";
+
 describe("withActionScheduler", () => {
     afterEach(() => {
         unmountAll();
@@ -21,5 +23,22 @@ describe("withActionScheduler", () => {
 
         // Assert
         expect(wrapper.text()).toBe("true");
+    });
+
+    it("should forward a ref", () => {
+        // Arrange
+        class Component extends React.Component<WithActionSchedulerProps> {
+            render() {
+                return <div>Hello, world!</div>;
+            }
+        }
+        const TestComponent = withActionScheduler(Component);
+        let ref: mixed = null;
+
+        // Act
+        mount(<TestComponent ref={(node) => (ref = node)} />);
+
+        // Assert
+        expect(ref).toBeInstanceOf(Component);
     });
 });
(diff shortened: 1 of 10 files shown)


packages/wonder-blocks-timing/components/with-action-scheduler.test.js
```diff
@@ -22,4 +24,21 @@ describe("withActionScheduler", () => {
...
+            }
+        }
+        const TestComponent = withActionScheduler(Component);
+        let ref: mixed = null;
+
+        // Act
+        mount(<TestComponent ref={(node) => (ref = node)} />);
+
+        // Assert
+        expect(ref).toBeInstanceOf(Component);
```
somewhatabstract: Nice test!

packages/wonder-blocks-timing/util/types.js
```diff
@@ -230,15 +230,22 @@ export interface IScheduleActions {
...
      * intervals, timeouts, and animation frame requests.
      */
     schedule: IScheduleActions,
 |};
+
+/**
+ * Extends the given props with props that the `withActionScheduler` higher
+ * order component will inject.
+ */
+export type WithActionScheduler<TOwnProps: {...}> = {|
```
somewhatabstract: Is this just to make sure this isn't a breaking change?

kevinbarabash: Yes, exactly.  🙂

kevinbarabash:
@jeresig this shouldn't be breaking at least I intended it not to be since we're still exporting the same types we were before.  I can try using `yarn link` to verify whether or not it is actually breaking before publishing.

kevinbarabash:
I guess it is technically breaking since it'll fix a couple of `$FlowFixMe`s in webapp.  I think I should still give it only a minor bump because it's more like a bug fix, it's not really changing the API of wonder-blocks-timing.

jeresig:
@kevinbarabash ok, that's fine - thank you for checking! A minor version bump seems ok to me.

kevinbarabash:
Sorry if I made it sound like I had already checked with `yarn link` last night.  I knew that there were going to be issues with `$FlowFixMe`s.

I have now checked with `yarn link` and there are some additional issues:
- `withActionScheduler` now takes two generic params, but flow can infer both so I'll have to update all calls to `withActionScheduler` to remove the type params.
- flow gets confused with `yarn link` so I had to restart flow after linking.  Even then it still couldn't resolve "../.../../utils/testing/mount.js" even though it was on disk at the absolute path it said it couldn't find it at.  I think maybe the issue is that that path is above the folder that was symlinked by yarn link.

I still think a minor version bump is appropriate.  I'll be publishing today and will update wonder-blocks in webapp as part of the work I'm doing off the `no-ka` branch.

PR #732, opened 2020-01-17
Use husky and lint-staged for pre-commit check

kevinbarabash:
In #720 @jeresig suggested using `lint-staged` and `husky` for our git hooks infrastructure.  I'm implementing that suggestion now in order to address an issue where the previous `test:fast` script would fail if no .js files had changed.  Thankfully `lint-staged` handles this case for us automatically.

I've also updated the circle-ci config to use node 10 since some of the deps (husky & marked) require v10.

Screen Shot 2020-01-17 at 2 57 56 PM


package.json
```diff
@@ -109,10 +107,20 @@
...
-    "test:fast",
-    "pretty-quick"
-  ]
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.md": [
```
jeresig: Ohh - this is really nice!

jeresig: Err - will this run `gen-shapshot-tests` once for all files or once for every file that was changed?

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   lint:
     docker:
-      - image: circleci/node:8-browsers
+      - image: circleci/node:10-browsers
```
jeresig: May be worth updating the commit message to make it clear that it includes this as well.

package.json
kevinbarabash: This should run over all files since I'll need to make changes to the script being run to accept a list of files.

PR #733, opened 2020-01-19
Update to eslint 6

somewhatabstract:
This allows us to use the update flowtype rules that fixes the no-dupe-keys rule understand get/set.

NOTE: This is part of my update to Wonder Blocks Data to support custom client-side cache strategies, so it's on the feature branch, https://github.com/Khan/wonder-blocks/tree/feature/wbdatacache.


package.json
```diff
@@ -58,15 +58,15 @@
...
+    "eslint-plugin-jsx-a11y": "^6.2.3",
     "eslint-plugin-monorepo": "^0.2.1",
-    "eslint-plugin-prettier": "^3.1.0",
-    "eslint-plugin-promise": "^4.1.1",
-    "eslint-plugin-react": "^7.13.0",
+    "eslint-plugin-prettier": "^3.1.2",
+    "eslint-plugin-promise": "^4.2.1",
+    "eslint-plugin-react": "^7.18.0",
     "eslint-plugin-react-native-animation-linter": "^0.1.2",
     "eslint-watch": "^6.0.1",
```
somewhatabstract: I ran `yarn add -W` for all these.

packages/wonder-blocks-layout/components/media-layout.js
```diff
@@ -168,7 +168,9 @@ class MediaLayoutInternal extends React.Component<CombinedProps, State> {
 
             // And then through each key of each stylesheet
             for (const name of Object.keys(styleSheet)) {
-                if (mockStyleSheet.hasOwnProperty(name)) {
+                if (
+                    Object.prototype.hasOwnProperty.call(mockStyleSheet, name)
```
somewhatabstract: This was the only new violation.

https://eslint.org/docs/rules/no-prototype-builtins

PR #746, opened 2020-01-21
Part 2: Fill some testing gaps

somewhatabstract:
For WB-833.

This adds some test gap filling and a couple of related code tweaks.

### DIFF
--- a/packages/wonder-blocks-data/components/data.test.js
+++ b/packages/wonder-blocks-data/components/data.test.js
@@ -457,6 +457,36 @@ describe("Data", () => {
                     expect(fulfillRequestFn).toHaveBeenCalledTimes(1);
                     expect(fakeHandler.fulfillRequest).not.toHaveBeenCalled();
                 });
+
+                it("should invoke handler if interceptor returns null", () => {
+                    // Arrange
+                    const fakeHandler: IRequestHandler<string, string> = {
+                        fulfillRequest: jest.fn(() => Promise.resolve("data")),
+                        getKey: (o) => o,
+                        shouldRefreshCache: () => false,
+                        type: "MY_HANDLER",
+                        cache: null,
+                    };
+                    const fakeChildrenFn = jest.fn(() => null);
+                    const fulfillRequestFn = jest.fn(() => null);
+
+                    // Act
+                    mount(
+                        <InterceptData
+                            handler={fakeHandler}
+                            fulfillRequest={fulfillRequestFn}
+                        >
+                            <Data handler={fakeHandler} options={"options"}>
+                                {fakeChildrenFn}
+                            </Data>
+                        </InterceptData>,
+                    );
+
+                    // Assert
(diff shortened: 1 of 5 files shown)


packages/wonder-blocks-data/components/data.test.js
```diff
@@ -457,6 +457,36 @@ describe("Data", () => {
...
+                    // Arrange
+                    const fakeHandler: IRequestHandler<string, string> = {
+                        fulfillRequest: jest.fn(() => Promise.resolve("data")),
+                        getKey: (o) => o,
+                        shouldRefreshCache: () => false,
+                        type: "MY_HANDLER",
+                        cache: null,
+                    };
+                    const fakeChildrenFn = jest.fn(() => null);
+                    const fulfillRequestFn = jest.fn(() => null);
```
kevinbarabash: Is `fulfillRequestIFn` the interceptor?  If so, rewording the test title to something like the following may be clearer:
> should invoke handler's fulfillRequest if InterceptData's returns null

Thoughts?

packages/wonder-blocks-data/components/data.test.js
```diff
@@ -748,6 +778,37 @@ describe("Data", () => {
...
+                    // Arrange
+                    const fakeHandler: IRequestHandler<string, string> = {
+                        fulfillRequest: jest.fn(() => Promise.resolve("data")),
+                        getKey: (o) => o,
+                        shouldRefreshCache: jest.fn(() => false),
+                        type: "MY_HANDLER",
+                        cache: null,
+                    };
+                    const fakeChildrenFn = jest.fn(() => null);
+                    const shouldRefreshCacheFn = jest.fn(() => null);
```
kevinbarabash: I was like why would we want to return `?boolean` but it's so that we can defer to Data's `shouldRefreshCacheFn`, neat!

packages/wonder-blocks-data/components/internal-data.js
```diff
@@ -153,22 +153,29 @@ export default class InternalData<TOptions, TData> extends React.Component<
...
+                data,
+            };
+        }
+
+        if (error == null) {
+            // We should never get here ever.
+            throw new Error(
+                "Loaded result has invalid state where data and error are missing",
+            );
         }
```
kevinbarabash: Nice addition.

packages/wonder-blocks-data/util/request-tracking.test.js
```diff
@@ -284,6 +284,29 @@ describe("./request-tracking.js", () => {
...
+                    cache: null,
+                };
+                requestTracker.trackDataRequest(fakeValidHandler, "OPTIONS1");
+
+                // Act
+                const result = await requestTracker.fulfillTrackedRequests();
+
+                // Assert
+                expect(result).toStrictEqual({});
+            });
```
kevinbarabash: Thanks for adding a test for this.

PR #747, opened 2020-01-21
Part 3: Implement deferring to custom cache

somewhatabstract:
See WB-832 and WB-833

This is the main architectural change in how caching works. Here we weave the custom cache that can be optionally provided by handler into our `ResponseCache` behavior.

NOTE: This code and especially the tests will be nicer once the framework cache has a separate implementation.

There are four places:

1. `getEntry`
- If server-side, we ignore the custom cache
- If client-side,
    - We defer to any value in the custom cache
    - If custom cache is a miss, but framework cache is not, we add the framework cache entry to the custom cache, delete the framework cache version, and then return that value

2. `cacheError` and `cacheData`
- These both call the same function which, when client-side, will write to the custom cache rather than the framework cache

3. `remove`
- A new method that will remove matching entries from both the handler's custom cache and from the framework cache. This differs from `removeAll` in that it targets a single record rather than those matching a predicate.

4. `removeAll`
- A new method that will remove matching entries from both the handler's custom cache and from the framework cache. This uses a predicate to find records to delete.

### DIFF
--- a/packages/wonder-blocks-data/index.js
+++ b/packages/wonder-blocks-data/index.js
@@ -2,6 +2,8 @@
 import {ResponseCache as ResCache} from "./util/response-cache.js";
 import {RequestTracker} from "./util/request-tracking.js";
 
+import type {CacheEntry, IRequestHandler} from "./util/types.js";
+
 export type {Cache, CacheEntry, Result, IRequestHandler} from "./util/types.js";
 
 export type ResponseCache = $ReadOnly<Cache>;
@@ -11,6 +13,19 @@ export const initializeCache = (source: ResponseCache): void =>
 
 export const fulfillAllDataRequests = (): Promise<ResponseCache> =>
     RequestTracker.Default.fulfillTrackedRequests();
+
+export const removeFromCache = <TOptions, TData>(
+    handler: IRequestHandler<TOptions, TData>,
+    options: TOptions,
+): boolean => ResCache.Default.remove<TOptions, TData>(handler, options);
+
+export const removeAllFromCache = <TOptions, TData>(
+    handler: IRequestHandler<TOptions, TData>,
+    predicate?: (
+        key: string,
+        cacheEntry: ?$ReadOnly<CacheEntry<TData>>,
+    ) => boolean,
+): number => ResCache.Default.removeAll<TOptions, TData>(handler, predicate);
 
 /**
  * TODO(somewhatabstract): Export each cache type we implement.
--- a/packages/wonder-blocks-data/util/response-cache.js
+++ b/packages/wonder-blocks-data/util/response-cache.js
@@ -1,4 +1,5 @@
 // @flow
+import {Server} from "@khanacademy/wonder-blocks-core";
 import type {CacheEntry, Cache, IRequestHandler} from "./types.js";
 
 function deepClone<T: {...}>(source: T | $ReadOnly<T>): $ReadOnly<T> {
@@ -47,6 +48,16 @@ export class ResponseCache {
         entry: CacheEntry<TData>,
     ): CacheEntry<TData> {
         const requestType = handler.type;
+
+        // We don't support custom caches during SSR.
+        const customCache = Server.isServerSide() ? null : handler.cache;
+
+        // If we have a custom cache, use that and skip our own.
+        if (customCache != null) {
+            const frozenEntry = Object.freeze(entry);
+            customCache.store(handler, options, frozenEntry);
+            return frozenEntry;
+        }
 
         // Ensure we have a cache location for this handler type.
         this._cache[requestType] = this._cache[requestType] || {};
@@ -116,17 +127,33 @@ export class ResponseCache {
     ): ?$ReadOnly<CacheEntry<TData>> => {
         const requestType = handler.type;
 
-        // Get the subcache for the handler.
+        const key = handler.getKey(options);
+
+        // We don't use custom caches during SSR.
+        const customCache = Server.isServerSide() ? null : handler.cache;
+        const entry = customCache && customCache.retrieve(handler, options);
+        if (entry != null) {
+            // Custom cache has an entry, so use it.
+            return entry;
+        }
+
+        // Get the internal subcache for the handler.
         const handlerCache = this._cache[requestType];
         if (!handlerCache) {
             return null;
         }
 
-        const key = handler.getKey(options);
-
         // Get the response.
-        const entry = handlerCache[key];
-        return entry == null ? null : entry;
+        const internalEntry = handlerCache[key];
+        if (internalEntry == null) {
+            return null;
+        }
+
+        // If we have a custom cache on the handler, make sure it has the entry.
+        if (customCache != null) {
+            customCache.store(handler, options, internalEntry);
+        }
+        return internalEntry;
     };
 
     /**
--- a/packages/wonder-blocks-data/util/response-cache.js
+++ b/packages/wonder-blocks-data/util/response-cache.js
@@ -42,7 +42,7 @@ export class ResponseCache {
         }
     }
 
-    // TODO(somewhatabstract): remove/removeAll and tests
+    // TODO(somewhatabstract): tests for remove/removeAll
     // TODO(somewhatabstract): tests for store to custom on retrieve
     // TODO(somewhatabstract): tests for store to custom on store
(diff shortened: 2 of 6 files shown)


packages/wonder-blocks-data/index.js
```diff
@@ -12,6 +14,19 @@ export const initializeCache = (source: ResponseCache): void =>
...
+    options: TOptions,
+): boolean => ResCache.Default.remove<TOptions, TData>(handler, options);
+
+export const removeAllFromCache = <TOptions, TData>(
+    handler: IRequestHandler<TOptions, TData>,
+    predicate?: (
+        key: string,
+        cacheEntry: ?$ReadOnly<CacheEntry<TData>>,
+    ) => boolean,
+): number => ResCache.Default.removeAll<TOptions, TData>(handler, predicate);
```
kevinbarabash: `TData` could be structured to include both the data being cached as well as metadata such as the last used timestamp right?

packages/wonder-blocks-data/util/response-cache.js
```diff
@@ -116,21 +126,132 @@ export class ResponseCache {
...
+        if (!handlerCache) {
+            return removedCountCustom;
+        }
+
+        // Apply the predicate to what we have cached.
+        let removedCount = 0;
+        for (const [key, entry] of Object.entries(handlerCache)) {
+            if (
+                typeof predicate !== "function" ||
+                predicate(key, (entry: any))
```
kevinbarabash: I assume we're casting to `any` here because Flow's definition for `Object.entries` is getting in the way of nice typing.

packages/wonder-blocks-data/util/response-cache.test.js
```diff
@@ -111,117 +112,597 @@ describe("./response-cache.js", () => {
...
+                 *
+                 * TODO(somewhatabstract): Isolate framework cache.
+                 * We should probably isolate the framework cache itself better
+                 * to fix this.
+                 */
+                const result = cache._cache["MY_HANDLER"]["MY_KEY"];
+
+                // Assert
+                expect(customCache.store).not.toHaveBeenCalled();
+                expect(result).toStrictEqual({data: "data"});
```
kevinbarabash: I didn't realize there was a `toStrictEqual`, cool.

packages/wonder-blocks-data/util/response-cache.test.js
```diff
@@ -111,117 +112,597 @@ describe("./response-cache.js", () => {
...
+
+                // Act
+                cache.cacheData(fakeHandler, "options", "data");
+                /**
+                 * We want to bypass the custom stuff and grab from the real
+                 * storage so we can validate without side-effects.
+                 *
+                 * TODO(somewhatabstract): Isolate framework cache.
+                 * We should probably isolate the framework cache itself better
+                 * to fix this.
```
kevinbarabash: Is this handled in a following PR?

packages/wonder-blocks-data/util/response-cache.test.js
```diff
@@ -111,117 +112,597 @@ describe("./response-cache.js", () => {
...
+        });
+
+        describe("with custom cache", () => {
+            it("SSR should use the framework cache, not the custom cache", () => {
+                // Arrange
+                jest.spyOn(Server, "isServerSide").mockReturnValue(true);
+                const cache = new ResponseCache();
+                const customCache = {
+                    store: jest.fn(),
+                    retrieve: jest.fn(),
```
kevinbarabash: `get`/`set` seem like more common names.  These are fine too.  No action required.

packages/wonder-blocks-data/util/response-cache.test.js
```diff
@@ -111,117 +112,597 @@ describe("./response-cache.js", () => {
...
                 shouldRefreshCache: () => false,
                 fulfillRequest: jest.fn(),
-                cache: null,
+                cache: customCache,
             };
 
             // Act
-            cache.cacheData(fakeHandler, "options", "data");
-            const result = cache.getEntry(fakeHandler, "options");
+            const result = cache.remove(fakeHandler, "optionsA");
```
kevinbarabash: Is the use of `"optionsA"` vs `"options"` meaningful?

packages/wonder-blocks-data/util/response-cache.test.js
```diff
@@ -111,117 +112,597 @@ describe("./response-cache.js", () => {
...
-    });
 
-    describe("#cacheError", () => {
-        it("should add the entry to the cache", () => {
+        it("should return true if something was removed from custom cache", () => {
             // Arrange
-            const cache = new ResponseCache();
+            const cache = new ResponseCache({
+                MY_HANDLER: {
+                    MY_OTHER_KEY: {data: "data!"},
```
kevinbarabash: The reason why this isn't `MY_KEY` is because there's no way to get `MY_KEY` into a custom cache without it first being removed from the framework cache?

packages/wonder-blocks-data/util/response-cache.test.js
```diff
@@ -233,17 +714,33 @@ describe("./response-cache.js", () => {
...
-            expect(result).toBeNull();
+            expect(result).toBe(2);
+            expect(after).toStrictEqual({
+                MY_HANDLER: {
+                    MY_KEY2: {data: "1"},
+                },
+                OTHER_HANDLER: {
+                    MY_KEY: {data: "1"},
+                },
+            });
```
kevinbarabash: I sometimes struggle a bit reviewing tests where everything's mocked out (my fault not yours), so it's nice to see some tests with data as well. 🙂

packages/wonder-blocks-data/index.js
somewhatabstract: What is in `TData` is entirely upto the IRequestHandler implementation, so yes, it can include whatever is needed by the use case being handled.

packages/wonder-blocks-data/util/response-cache.js
somewhatabstract: Yup

packages/wonder-blocks-data/util/response-cache.test.js
somewhatabstract: Probably. I cannot remember.

packages/wonder-blocks-data/util/response-cache.test.js
somewhatabstract: I don't feel that `get`/`set` really corresponds to caching. It sounds much more like just retrieving properties. I wanted to codify that it was doing work of some kind.

packages/wonder-blocks-data/util/response-cache.test.js
somewhatabstract: Meaningful? The options (via the key generation method in the request handler) are how requests are differentiated. If two things shared the same key, they would share the same cache. I likely made a change here to ensure that was taken into account.

packages/wonder-blocks-data/util/response-cache.test.js
somewhatabstract: I don't understand the question.

kevinbarabash: In `getEntry` we do `delete handlerCache[key];` if there's a custom cache.  As a result we'd never be in a state where we'd have:
```
MY_HANDLER: {
  MY_KEY: {data: "data!"},
}
```
in both the framework cache and the custom cache since the only way for something to get into the custom cache is by calling `getEntry()` which transfers the entry from the framework cache to the custom cache.  At least that's my understanding of things.  I wanted to confirm that that's actually the case and that the data in the test cases reflects that invariant.

somewhatabstract: There are two ways things go into custom cache.

1. When hydrating on the client from SSR result, such that the framework cache has something in it that should be in the custom cache (done on `getEntry`)
2. When a client-side request is handled and the result is cached

Framework cache is only used client-side when there is no custom cache, or during SSR.

As for this test case, all that data is a red herring.

The custom cache implementation returns `true` and so our framework should be doing so. The data is just present for a complete test case. I'll see if I can make that clearer.

packages/wonder-blocks-data/util/response-cache.test.js
somewhatabstract: https://github.com/Khan/wonder-blocks/pull/748#discussion_r377728479

PR #750, opened 2020-01-22
Part 5: The NoCache cache

somewhatabstract:
For WB-833.

The first of our exported cache strategies; the `NoCache`. A simple example where we can configure a handler to never cache data. This allows us to use the framework cache to support SSR but then once hydrated, never cache, thus avoiding using up memory and just have the client always request new data after the initial hydration.

### DIFF
--- a/packages/wonder-blocks-data/util/no-cache.js
+++ b/packages/wonder-blocks-data/util/no-cache.js
@@ -3,6 +3,14 @@ import type {ValidData, ICache, CacheEntry, IRequestHandler} from "./types.js";
 
 /**
  * A cache implementation to use when no caching is wanted.
+ *
+ * Use this with your request handler if you want to support server-side
+ * rendering of your data requests, but want to ensure data is never cached
+ * on the client-side.
+ *
+ * This is better than having `shouldRefreshCache` always return `true` in the
+ * handler as this ensures that cache space and memory are never used for the
+ * requested data after hydration has finished.
  */
 export default class NoCache<TOptions, TData: ValidData>
     implements ICache<TOptions, TData> {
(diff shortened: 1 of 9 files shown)


packages/wonder-blocks-data/util/no-cache.js
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * A cache implementation to use when no caching is wanted.
+ *
+ * Use this with your request handler if you want to support server-side
+ * rendering of your data requests, but want to ensure data is never cached
+ * on the client-side.
+ *
+ * This is better than having `shouldRefreshCache` always return `true` in the
+ * handler as this ensures that cache space and memory are never used for the
+ * requested data after hydration has finished.
```
kevinbarabash: I hadn't even thought of that being a benefit o the no-cache cache.  Cool!

packages/wonder-blocks-data/util/no-cache.test.js
```diff
@@ -0,0 +1,105 @@
+// @flow
+import NoCache from "./no-cache.js";
+
+import type {IRequestHandler} from "./types.js";
+
+describe("NoCache", () => {
+    describe("#store", () => {
+        it("should not throw", () => {
+            // Arrange
```
kevinbarabash: I wonder if there might be value in calling clone on the data being cached even if we don't cache it so that there cloning errors are consistent between caches.  wdyt?

somewhatabstract: Hmmm, not sure. Seems like it would be allocating memory unnecessarily and adding to the memory burden that has to be garbage collected.

PR #761, opened 2020-01-22
Add `aria-modal=true` to wonder-blocks-modal. This helps with trapping.

annr:
Summary:
A relatively new ARIA attribute -- ”aria-modal” -- gives assistive technoloies more hints as to how to deal with modals, especially around trapping focus.

Here is some background about it: https://developer.paciellogroup.com/blog/2018/06/the-current-state-of-modal-dialog-accessibility/

That post says:
“This attribute is a pretty big deal and a very welcome addition to the specification. The aria-modal attribute will help in taking care of one of the biggest hurdles with custom modal dialogs; keeping a screen reader within the active dialog, as creating a JavaScript trap for standard keyboard focus is not enough to curtail a wandering virtual cursor.”

Adding the aria-modal attribute improved VoiceOver iOS issues for this webapp ticket:
https://khanacademy.atlassian.net/browse/LP-6057

Here’s a comment within that ticket that talks about how I tested this attribute with Wonder Blocks modal and VoiceOver iOS and noticed that the attribute improves trapping there as well. https://khanacademy.atlassian.net/browse/LP-6057?focusedCommentId=109087&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-109087

This discussion includes a video showing Wonder Blocks Modal without `aria-modal=true` reading the background. When we include the attribute, trapping works well.
https://khanacademy.slack.com/archives/CEA6W0F6F/p1576592992059800?thread_ts=1576536426.048600&cid=CEA6W0F6F


packages/wonder-blocks-modal/components/modal-dialog.js
```diff
@@ -77,6 +77,7 @@ export default class ModalDialog extends React.Component<Props> {
                             {below && <View style={styles.below}>{below}</View>}
                             <View
                                 role="dialog"
+                                aria-modal={true}
```
kevinbarabash: I think this should be `aria-model="true"` to match other boolean aria props as defined in https://github.com/Khan/wonder-blocks/blob/master/packages/wonder-blocks-core/util/types.js#L95-L142.  Also, please add an `aria-modal` entry to `AriaProps` in that file.

annr:
Yes, my github user is a member of the Khan organization. I've also signed the "user contrib" twice :) I'll make the comment it suggests to see if the issue goes away.

annr:
[clabot:check]

PR #767, opened 2020-01-24
Try again to prevent dependabot PRs from running chromatic

kevinbarabash:
I'm not sure why the previous check I had for `github.actor != 'dependabot-preview'` wasn't working.  It's possible that the actor is actually something different since `dependabot-preview` is an app not a user.

This solution instead checks the name of the PR and doesn't run the `chromatic` job if the ref for the PR contains `dependabot/`.

It took me a long time to figure out the syntax.  It turns out you have to wrap things in double quotes that have special characters like `!`.  I found the answer on https://github.community/t5/GitHub-Actions/Expression-syntax-for-not-startswith/td-p/39571. 

Test Plan: merge and wait for the next dependabot PR

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -1,13 +1,11 @@
 name: Node CI
 
-on:
-  push:
-    branches:
-      - '!dependabot/**'
+on: [push]
 
 jobs:
   chromatic:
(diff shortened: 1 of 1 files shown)


kevinbarabash:
@jeresig thank you.  I've update the PR and the test plan.

PR #781, opened 2020-01-29
Add optional `required` prop to `Checkbox` and `Radio` components

dierat:
# SUMMARY
Form inputs, like checkboxes, should be able to handle a prop called `required`, which helps screen readers communicate to users that the form element needs to be filled, checked, or selected in order for the user to continue. We don't currently have that prop listed for the Wonder Blocks `Checkbox` and `Radio` components, so it causes a `flow` error if you pass it in. (It does get applied to the HTML element correctly, but you have to suppress the flow error.) We currently have one place where this prop is already being used and its flow error is being suppressed: https://github.com/Khan/webapp/blob/d31d0a873f0d31a4733ade4b5fb6ad04a031602a/javascript/notifications-package/accept-tos-notification.jsx#L205
# TEST PLAN
Run `yarn test` and verify all the tests pass.

### DIFF
--- a/packages/wonder-blocks-form/util/types.js
+++ b/packages/wonder-blocks-form/util/types.js
@@ -22,6 +22,8 @@ export type ChoiceCoreProps = {|
      * guarantee that the ID is unique within everything rendered on a page.
      * Used to match <label> with <input> elements for screenreaders. */
     id?: string,
+    /** Optionally state that filling in or checking the input is required */
+    required?: boolean,
     /** Optional test ID for e2e testing */
     testId?: string,
 |};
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-form/util/types.js
```diff
@@ -22,6 +22,8 @@ export type ChoiceCoreProps = {|
      * guarantee that the ID is unique within everything rendered on a page.
      * Used to match <label> with <input> elements for screenreaders. */
     id?: string,
+    /** Optionally state that filling in or checking the input is required */
+    required?: boolean,
```
jandrade: For some reason, we are (re)defining these props in different locations (maybe for documentation purposes). e.g. https://github.com/Khan/wonder-blocks/blob/checkbox-required/packages/wonder-blocks-form/components/checkbox.js#L8

@kevinbarabash @jeresig Do you know why we are doing this?

kevinbarabash: It was for documentation purposes because react-docgen doesn't understand spreads. :(

kevinbarabash: `ChoiceCoreProps` is used in `CheckboxCore`, but not by `Checkbox`.  Since `Checkbox` and `Radio` are the publicly expose components their props will need to be updated as well.

PR #818, opened 2020-02-13
Fix docs and export ICache type

somewhatabstract:
This relates to WB-832 and WB-833.

I forgot to export the `ICache` type and also the docs needed some changes.

### DIFF
--- a/packages/wonder-blocks-data/util/no-cache.md
+++ b/packages/wonder-blocks-data/util/no-cache.md
@@ -1,0 +1,55 @@
+The `ICache` interface is included below for reference in case you would like
+to implement your own caching strategy.
+
+```js static
+interface ICache<TOptions, TData: ValidData> {
+    /**
+     * Stores a value in the cache for the given handler and options.
+     */
+    store(
+        handler: IRequestHandler<TOptions, TData>,
+        options: TOptions,
+        entry: CacheEntry<TData>,
+    ): void;
+
+    /**
+     * Retrieves a value from the cache for the given handler and options.
+     */
+    retrieve(
+        handler: IRequestHandler<TOptions, TData>,
+        options: TOptions,
+    ): ?$ReadOnly<CacheEntry<TData>>;
+
+    /**
+     * Remove the cached entry for the given handler and options.
+     *
+     * If the item exists in the cache, the cached entry is deleted and true
+     * is returned. Otherwise, this returns false.
+     */
+    remove(
+        handler: IRequestHandler<TOptions, TData>,
+        options: TOptions,
+    ): boolean;
+
+    /**
+     * Remove all cached entries for the given handler that, optionally, match
+     * a given predicate.
+     *
+     * Returns the number of entries that were cleared from the cache.
+     */
+    removeAll(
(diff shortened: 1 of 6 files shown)


packages/wonder-blocks-data/util/no-cache.md
```diff
@@ -0,0 +1,55 @@
...
+```js static
+interface ICache<TOptions, TData: ValidData> {
+    /**
+     * Stores a value in the cache for the given handler and options.
+     */
+    store(
+        handler: IRequestHandler<TOptions, TData>,
+        options: TOptions,
+        entry: CacheEntry<TData>,
+    ): void;
```
kevinbarabash: I wish our documentation tools could extract this from the code. 😞 Maybe someday. 🤞

packages/wonder-blocks-data/util/request-handler.md
```diff
@@ -45,7 +45,10 @@ may be fulfilled by the wrong handler.
 
 There is also an optional constructor argument, `cache`, which can be used to
 provide a custom cache for use with data the handler fulfills. Custom caches
-must implement the `ICache<TOptions, TData>` interface.
+must implement the `ICache<TOptions, TData>` interface. If this is omitted, the
+core Wonder Blocks Data in-memory cache will be used. If you want to avoid
+caching in memory, see `NoCache`, whith is a caching strategy that eliminates
+the use of caching entirely.
```
kevinbarabash: typo: `whith`

packages/wonder-blocks-data/util/no-cache.md
```diff
@@ -0,0 +1,55 @@
+The `ICache` interface is included below for reference in case you would like
+to implement your own caching strategy.
```
kevinbarabash: The filename is `no-cache.md` but this file documents `ICache`, is this intentional?

packages/wonder-blocks-data/util/request-handler.md
somewhatabstract: OOooo, good catch

packages/wonder-blocks-data/util/no-cache.md
somewhatabstract: In short, yes, it is intentional.

This really documents `NoCache` mostly. It includes `ICache` so that it's in the docs. Just like the `RequestHandler` base class docs, which include the `IRequestHandler` interface.

packages/wonder-blocks-data/util/request-handler.md
somewhatabstract: https://github.com/Khan/wonder-blocks/commit/e4b685108879fe29512af3b0674c8fb75c3a21a1

packages/wonder-blocks-data/util/no-cache.md
somewhatabstract: I see the confusion though. I forgot that code comments are only exported to docs for react components. I've updated the NoCache docs to accommodate this oversight. Thanks for catching it.

https://github.com/Khan/wonder-blocks/commit/259898380dd7222f26d100fe0af8131e2b559ae8

PR #819, opened 2020-02-13
Configure netlify to ignore dependabot PRs

kevinbarabash:
I found this recipe on https://community.netlify.com/t/common-issue-how-can-i-optimize-my-netlify-build-time/3907/10.

After this is landed we'll have to re-enable builds on PRs to check that it's working as expected.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,2 @@
+[build]
+  ignore = git log -1 --pretty=%B | grep dependabot


kevinbarabash:
I've updated the project to deploy PRs again.  Now all we have to do is wait for the next dependabot PR to see if this did the trick.

PR #823, opened 2020-02-21
Fix netlify.toml

kevinbarabash:
The `ignore` directive in netlify.toml files only needs the exit code.  If any text is returned it causes the github check to fail.  We'd like it not to build but the github check should still succeed.  Hopefully this will fix it.  

Netlify docs: https://docs.netlify.com/configure-builds/file-based-configuration/#ignore-builds.

I tested this by changing the author name to my own and checking the exit code and then switch it back to dependabot and say that is was `1` when the author matched the author of the most recent commit and `0` when it did not.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,2 +1,2 @@
 [build]
-  ignore = ./netlify_check.sh
+  ignore = "./netlify_check.sh"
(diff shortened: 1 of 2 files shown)


kevinbarabash:
It's trying to parse the command as a number.  Time to read some more docs.

jeresig:

Built with commit 497290468a422a0cd67c0e6093f7cfdc62fcc5c3

https://deploy-preview-823--wonder-blocks.netlify.com

kevinbarabash:
It's working now.  I think it's probably also worth adding a check to see which paths have been modified since changes to things like README.md don't require running tests or the netlify build.

PR #833, opened 2020-02-26
Fix netlify_check.sh to do the right thing.

kevinbarabash:
From the docs:
> An exit-code of 1 indicates the contents have changed, so if your command returns that code the build process will continue per usual. An exit-code of 0 indicates that the build should return early.


jeresig:

Built with commit 89df68ccb471b85d14ab35e8d96878db995e7edb

https://deploy-preview-833--wonder-blocks.netlify.com

PR #850, opened 2020-03-10
In Clickable add event source information

nixterrimus:
Adds event source information in the Clickable component to differentiate between a keyboard, a touch device, and a mouse. This is needed for the math-input which has different states for touch down vs. mouse down.

NOTE:
- Snapshot tests aren't passing and I'm not sure how to update them, @kevinbarabash

### DIFF
--- a/packages/wonder-blocks-core/components/clickable-behavior.js
+++ b/packages/wonder-blocks-core/components/clickable-behavior.js
@@ -113,6 +113,11 @@ export type ClickableState = {|
      * See component documentation for more details.
      */
     pressed: boolean,
+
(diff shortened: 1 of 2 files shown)


kevinbarabash:
@nixterrimus I don't think we should update the snapshots.  We're setting `inputType` on some of the HTML tags and we don't want to.  The problem is we're spreading `stateProps` as props but it includes `inputType` now but we don't want to set `inputType` as a prop on any tag/element.

jeresig:
Closing this out - feel free to re-open if you're working on this again!

PR #864, opened 2020-05-04
fix(clickable) WB-860:  href prop not working with keyboard navigation

jandrade:
Fixed an issue that prevented `Clickable` to work with keyboard navigation.

The specific case happens when we set `href` and `skipClientNav` and try to open a link using the keyboard. The issue was that `href` wasn't passed down to `ClickableBehavior` and this caused the component to avoid navigating because `href` was undefined for that case.

To fix this, we are adding `href` to the internal `ClickableBehavior` instance.

### Test plan

1. Open https://5e1bf4b385e3fb0020b7073c-cwqfgsekob.chromatic.com/?path=/story/clickable--keyboard-navigation
2. Using the keyboard, hover on the clickable link, then press `Enter` and verify the browser opens the URL set in `href` (https://khanacademy.org in this case).

### DIFF
--- a/config/jest/log-on-fail-reporter.js
+++ b/config/jest/log-on-fail-reporter.js
@@ -3,7 +3,7 @@
  * From: https://gist.github.com/GeeWee/71db0d9911b4a087e4b2486386168b05
  */
 const chalk = require("chalk");
-const {getConsoleOutput} = require("jest-util");
+const {getConsoleOutput} = require("@jest/console");
 const DefaultReporter = require("@jest/reporters/build/default_reporter")
     .default;
 const getResultHeader = require("@jest/reporters/build/get_result_header")
--- a/packages/wonder-blocks-clickable/components/clickable.md
+++ b/packages/wonder-blocks-clickable/components/clickable.md
@@ -88,3 +88,49 @@ const styles = StyleSheet.create({
     </View>
 </MemoryRouter>
 ```
+
+### Navigating with the Keyboard
+
+Clickable adds support to keyboard navigation. This way, your components are
+accessible and emulate better the browser's behavior.
+
+*NOTE:* If you want to navigate to an external URL and/or reload the window, make
+sure to use `href` and `skipClientNav={true}`.
+
+```jsx
+import {StyleSheet} from "aphrodite";
+import Clickable from "@khanacademy/wonder-blocks-clickable";
+import {View} from "@khanacademy/wonder-blocks-core";
+import Color from "@khanacademy/wonder-blocks-color";
+import {Body} from "@khanacademy/wonder-blocks-typography";
+
+const styles = StyleSheet.create({
+    hovered: {
+        textDecoration: "underline",
+        backgroundColor: Color.teal,
+    },
+    pressed: {
+        color: Color.blue,
+    },
+    focused: {
+        outline: `solid 4px ${Color.lightBlue}`,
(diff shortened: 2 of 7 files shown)


config/jest/log-on-fail-reporter.js
```diff
@@ -3,7 +3,7 @@
  * From: https://gist.github.com/GeeWee/71db0d9911b4a087e4b2486386168b05
  */
 const chalk = require("chalk");
-const {getConsoleOutput} = require("jest-util");
+const {getConsoleOutput} = require("@jest/console");
```
jandrade, line 6: Changed this line because Jest v25 moved this function to the `console` package.

packages/wonder-blocks-clickable/components/clickable.md
```diff
@@ -88,3 +88,49 @@ const styles = StyleSheet.create({
...
+    pressed: {
+        color: Color.blue,
+    },
+    focused: {
+        outline: `solid 4px ${Color.lightBlue}`,
+    },
+});
+
+<View>
+    <Clickable href="https://khanacademy.org" skipClientNav={true}>
```
jeresig, line 121: You can use `https://www.khanacademy.org/about/tos` as that URL is allowed to load inside an iframe.

packages/wonder-blocks-clickable/components/clickable.stories.js
```diff
@@ -0,0 +1,43 @@
...
+
+import {StyleSheet} from "aphrodite";
+import Clickable from "@khanacademy/wonder-blocks-clickable";
+import {View} from "@khanacademy/wonder-blocks-core";
+import Color from "@khanacademy/wonder-blocks-color";
+import {Body} from "@khanacademy/wonder-blocks-typography";
+
+export default {
+    title: "Clickable",
+};
```
kevinbarabash, line 12: I don't think this warrants a screenshot test.  Please add:
```
    parameters: {
        chromatic: {
            // We already have screenshots of other stories that cover more of the button states
            disable: true,
        },
    },
```

packages/wonder-blocks-clickable/components/clickable.stories.js
```diff
@@ -0,0 +1,43 @@
...
+
+export const keyboardNavigation = () => (
+    <View>
+        <Clickable href="https://khanacademy.org" skipClientNav={true}>
```
kevinbarabash, line 16: This should use https://www.khanacademy.org/about/tos as well.

packages/wonder-blocks-clickable/components/clickable.test.js
```diff
@@ -112,4 +112,35 @@ describe("Clickable", () => {
...
+            </Clickable>,
+        );
+
+        // Act
+        const buttonWrapper = wrapper.find(`[data-test-id="button"]`).first();
+        // simulate Enter
+        buttonWrapper.simulate("keyup", {keyCode: 13});
+
+        // Assert
+        expect(window.location.assign).toHaveBeenCalledWith("/foo");
```
kevinbarabash, line 144: nice

kevinbarabash:
@jandrade I denied the UI Test changes because I don't think we need a screenshot test for the change being made.  I was hoping there would be a place to give a rationale in the chromatic UI, but there wasn't one.  Maybe that's coming in v2 of their UI.

PR #874, opened 2020-05-05
feat(link) WB-859: Add id to Link

jandrade:
- Adds the `id` prop to the `<Link />` component.
- Fixes some remaining changes from the `prettier` upgrade.

### Test plan
Check that the `Link` snapshot includes the `id` when set in the example.

### DIFF
--- a/packages/wonder-blocks-data/generated-snapshot.test.js
+++ b/packages/wonder-blocks-data/generated-snapshot.test.js
@@ -133,7 +133,7 @@ describe("wonder-blocks-data", () => {
 
         const handler = new MyHandler();
 
-        const getEntryInterceptor = function(options) {
+        const getEntryInterceptor = function (options) {
             if (options === "DATA") {
                 return {
                     data: "I'm DATA from the cache",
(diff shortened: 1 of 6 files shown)


packages/wonder-blocks-data/generated-snapshot.test.js
```diff
@@ -133,7 +133,7 @@ describe("wonder-blocks-data", () => {
 
         const handler = new MyHandler();
 
-        const getEntryInterceptor = function(options) {
+        const getEntryInterceptor = function (options) {
```
jandrade, line 136: These are some changes `prettier` didn't catch on the previous PR.

packages/wonder-blocks-dropdown/generated-snapshot.test.js
```diff
@@ -1397,8 +1397,9 @@ describe("wonder-blocks-dropdown", () => {
                 <OptionItem
                     key={i}
                     value={(i + 1).toString()}
-                    label={`School ${i +
-                        1} in Wizarding World Some more really long labels?`}
+                    label={`School ${
```
jandrade, line 1400: Same here.

PR #880, opened 2020-05-10
Experiment with having tests live in a __tests__ folder

kevinbarabash:
Given that we rely on code coverage reports in this repo to inform testing efforts requiring _test.js files to be siblings of their corresponding implementation file is less valuable as a signal that tests should be written for something.  Moving _test.js files into their own `__tests__` folder makes it easier to scan for a particular file because the differentiating prefix is shorter, i.e. I don't have to read the whole file name to know which file is which.

This diff changes a large number of files.  In order to make dealing with history easier, after this change I will add a .git-blame-ignore-revs file with the commit SHA of this change after the PR squash-merged.  If, after a while we determine that having _test.js files in `__tests__` folders is a net-negative we can move files back to where they were and add that commit SHA to .git-blame-ignore-revs as well.

For more details on .git-blame-ignore-revs see https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame.


kevinbarabash:
I'm not sure how lint made it into the first commit.  If it happens again we should investigate further.

kevinbarabash:
@jeresig are you suggesting that I change the fixture test generator to put the generated tests inside of `wonder-blocks-<pkg>/__tests__`?  I want to make sure before making the changes.

jeresig:
@kevinbarabash Yes that way it, and the `index.test.js` file, can both live in `__tests__` inside that directory. Probably better for us to be consistent about this practice!

PR #881, opened 2020-05-11
Remove unsafe componentWillMount from ScrollDisabler

kevinbarabash:
When I was updating webapp to use 16.9, there was a warning about using `componentWillMount` that was causing a number of tests to fail.  This PR moves this logic into the `ScrollDisabler`'s constructor.  I've also added a new lint rule: `no-deprecated` which flags deprecated methods based on the current version of react we're using.

Test Plan: 
- yarn start
- open http://localhost:6060/#section-modal
- click on one of the buttons to open a modal
- try scroll with the modal open, see that the page doesn't scroll
- close the modal, see that the page doesn't jump

### DIFF
--- a/packages/wonder-blocks-modal/components/scroll-disabler.js
+++ b/packages/wonder-blocks-modal/components/scroll-disabler.js
@@ -17,14 +17,18 @@ const needsHackyMobileSafariScrollDisabler = (() => {
     return userAgent.indexOf("iPad") > -1 || userAgent.indexOf("iPhone") > -1;
 })();
 
-class ScrollDisabler extends Component<{||}> {
+type Props = {||};
+
+class ScrollDisabler extends Component<Props> {
     static oldOverflow: string;
     static oldPosition: string;
     static oldScrollY: number;
     static oldWidth: string;
     static oldTop: string;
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-modal/components/scroll-disabler.js
```diff
@@ -17,14 +17,18 @@ const needsHackyMobileSafariScrollDisabler = (() => {
...
+
+class ScrollDisabler extends Component<Props> {
     static oldOverflow: string;
     static oldPosition: string;
     static oldScrollY: number;
     static oldWidth: string;
     static oldTop: string;
 
-    componentWillMount() {
+    constructor(props: Props) {
```
jeresig, line 29: I don't think this change is equivalent (and I also think it wasn't right before). The `constructor` shouldn't have side effects so I think this actually should become `componentDidMount` instead.

kevinbarabash: Will do.  I was thinking about side effects in terms of changes the component was making to things outside of itself instead of things outside of itself that the component was using. 😅

PR #887, opened 2020-05-17
Upgrade to flow 0.115

kevinbarabash:
This will help us upgrade flow in webapp.

### DIFF
--- a/packages/wonder-blocks-core/components/view.js
+++ b/packages/wonder-blocks-core/components/view.js
@@ -59,17 +59,22 @@ export default class View extends React.Component<Props> {
 
     render() {
         const {testId, tag, ...restProps} = this.props;
+        const props = {
+            ...restProps,
+            "data-test-id": testId,
+        };
+
         switch (tag) {
             case "article":
-                return <StyledArticle data-test-id={testId} {...restProps} />;
+                return <StyledArticle {...props} />;
             case "aside":
-                return <StyledAside data-test-id={testId} {...restProps} />;
+                return <StyledAside {...props} />;
             case "nav":
-                return <StyledNav data-test-id={testId} {...restProps} />;
+                return <StyledNav {...props} />;
             case "section":
-                return <StyledSection data-test-id={testId} {...restProps} />;
+                return <StyledSection {...props} />;
             case "div":
-                return <StyledDiv data-test-id={testId} {...restProps} />;
+                return <StyledDiv {...props} />;
             default:
                 // eslint-disable-next-line no-unused-expressions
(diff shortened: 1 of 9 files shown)


packages/wonder-blocks-core/components/view.js
```diff
@@ -59,17 +59,22 @@ export default class View extends React.Component<Props> {
 
     render() {
         const {testId, tag, ...restProps} = this.props;
+        const props = {
+            ...restProps,
+            "data-test-id": testId,
+        };
```
kevinbarabash, line 65: Flow worries that spreading an object with an index will overwrite props that come before the spread which is a valid concern.

packages/wonder-blocks-data/util/__tests__/memory-cache.test.js
```diff
@@ -76,7 +76,7 @@ describe("MemoryCache", () => {
             };
 
             // Act
-            cache.store(fakeHandler, "options", {data: "data"});
+            cache.store<string, string>(fakeHandler, "options", {data: "data"});
```
kevinbarabash, line 79: New flow isn't able to infer the type of the store so we have to give it a little help.

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -241,7 +241,7 @@ export default class ActionMenu extends React.Component<Props, State> {
                               <ActionMenuOpenerCore
                                   {...eventState}
                                   disabled={disabled}
-                                  opened={opened}
+                                  opened={!!opened}
```
kevinbarabash, line 244: `boolean | void` is no longer equivalent to `boolean`.

packages/wonder-blocks-form/components/choice.js
```diff
@@ -19,7 +19,7 @@ type Props = {|
     value: string,
 
     /** User-defined. Whether this choice option is disabled. Default false. */
-    disabled?: boolean,
```
kevinbarabash, line 22: I dropped the `?` because this and `checked` both have default props.

packages/wonder-blocks-dropdown/components/action-menu.js
jandrade: Oh, that Flow change makes sense. No need to change anything here, just I think another way to fix this would be changing `opened` to required setting its default value to `false`.

packages/wonder-blocks-form/components/choice.js
jandrade: Good catch!

PR #891, opened 2020-05-19
fix(dropdown) WB-858: Fix dropdown container width

jandrade:
## Summary

Fixes an issue introduced in `DropdownVirtualizedCore`. We have removed the maximum container width, allowing the dropdown container to take the width from its parent container.

### Test plan

1. Open https://5e1bf4b385e3fb0020b7073c-eneopdgtcf.chromatic.com/?path=/story/multiselect--custom-styles
2. Verify that when the dropdown is opened, the options container takes the parent's width.

#### BEFORE
dropdown-width-before

#### AFTER
dropdown-width-after

### DIFF
--- a/packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.js
+++ b/packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.js
@@ -60,11 +60,6 @@ type State = {|
 const MAX_VISIBLE_ITEMS = 10;
 
 /**
- * Maximum horizontal size allowed for the container
- */
-const MAX_ALLOWED_WIDTH = 512;
-
-/**
  * A react-window's List wrapper that instantiates the virtualized list and
  * dynamically calculates the item height depending on the type
  */
@@ -108,11 +103,7 @@ class DropdownCoreVirtualized extends React.Component<Props, State> {
         // after the non-virtualized items are rendered, we get the container
         //  width to pass it to react-window's List
         if (rootNode) {
-            const clientWidth = rootNode.getBoundingClientRect().width;
-            const width =
-                clientWidth < MAX_ALLOWED_WIDTH
-                    ? clientWidth
-                    : MAX_ALLOWED_WIDTH;
+            const width = rootNode.getBoundingClientRect().width;
 
             this.setState({
                 width,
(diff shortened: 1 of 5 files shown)


packages/wonder-blocks-dropdown/components/dropdown-core-virtualized.js
```diff
@@ -108,11 +103,7 @@ class DropdownCoreVirtualized extends React.Component<Props, State> {
         // after the non-virtualized items are rendered, we get the container
         //  width to pass it to react-window's List
         if (rootNode) {
-            const clientWidth = rootNode.getBoundingClientRect().width;
-            const width =
-                clientWidth < MAX_ALLOWED_WIDTH
-                    ? clientWidth
-                    : MAX_ALLOWED_WIDTH;
+            const width = rootNode.getBoundingClientRect().width;
```
kevinbarabash, line 106: If we wanted to avoid a call to `getBoundingclientRect()` for perf reasons we could set the width to be the width of the window since it's getting clipped by the dropdown anyways.  This is fine too.

packages/wonder-blocks-dropdown/components/multi-select.md
```diff
@@ -329,7 +330,10 @@ import {LabelLarge} from "@khanacademy/wonder-blocks-typography";
     <MultiSelect
         aria-labelledby="label-for-multi-select"
         id="unique-multi-select"
-        placeholder="Accessible MultiSelect"
+        labels={{
+            noneSelected: "Accessible MultiSelect",
+            someSelected: (numSelectedValues) => `${numSelectedValues} planets`,
+        }}
```
kevinbarabash, line 336: Do we no longer support `placeholder`?  If so, do the flow types need to change?

packages/wonder-blocks-dropdown/components/multi-select.stories.js
```diff
@@ -0,0 +1,75 @@
+// @flow
```
kevinbarabash, line 1: yay!

packages/wonder-blocks-dropdown/components/multi-select.md
jandrade: I removed the `placeholder` prop as part of this PR: https://github.com/Khan/wonder-blocks/pull/659, and replaced it with `labels.noneSelected`, but it seems that I forgot to update all the code examples written in the markdown file.

PR #892, opened 2020-05-19
WB-610: Wonder Blocks Modal package assumes a DOM

jandrade:
Adds a validation to verify that `window` exists and avoid any issues trying to run the `Modal` component on a SSR environment.


jandrade:
@kevinbarabash Good question. I noticed that we also use `window.scrollTo` inside this file. I'll add an extra check there just to be cautious.
 
Besides that, there are other few places that use `window` but all of them are accessed inside lifecycle methods that should be safe to use during SSR.

PR #903, opened 2020-05-21
feat(dropdown): SingleSelect with search filter

jandrade:
## Summary

Similar to the MultiSelect component (https://github.com/Khan/wonder-blocks/pull/479), these changes introduce the `isFilterable` prop to `<SingleSelect />`.

This allows us to filter options inside the single select component.

### Test plan
1. Open https://5e1bf4b385e3fb0020b7073c-vojuppvbfz.chromatic.com/?path=/story/dropdown-singleselect--custom-styles
2. Verify that the search input is present and that we can filter options by query.

single-filter


packages/wonder-blocks-dropdown/components/single-select.js
```diff
@@ -271,24 +345,41 @@ export default class SingleSelect extends React.Component<Props, State> {
...
                 alignment={alignment}
-                dropdownStyle={[selectDropdownStyle, dropdownStyle]}
+                dropdownStyle={[
+                    isFilterable && filterableDropdownStyle,
+                    selectDropdownStyle,
+                    dropdownStyle,
+                ]}
                 initialFocusedIndex={this.selectedIndex}
-                items={items}
+                items={[...this.getSearchField(), ...filteredItems]}
```
kevinbarabash, line 371: style nit: It feels a little weird to have `getSearchField` return an array.  Maybe we could do:
```
const searchField = this.getSearchField();
const items = searchField ? [searchField, ...filteredItems] : filteredItems;
...
items={items}
```

packages/wonder-blocks-dropdown/components/single-select.js
```diff
@@ -271,24 +345,41 @@ export default class SingleSelect extends React.Component<Props, State> {
...
+            alignment,
+            children,
+            dropdownStyle,
+            isFilterable,
+            light,
+            style,
+        } = this.props;
+        const {searchText} = this.state;
+        const allChildren = React.Children.toArray(children).filter(Boolean);
+        const filteredItems = this.getMenuItems(allChildren);
```
kevinbarabash, line 358: style nit: I'd probably move `React.Children.toArray(children).filter(Boolean)` inside of `this.getMenuItems()`.  It groups all of the children logic together instead of spreading it across multiple methods.

packages/wonder-blocks-dropdown/components/single-select.js
```diff
@@ -175,36 +194,90 @@ export default class SingleSelect extends React.Component<Props, State> {
...
+    getMenuItems(
+        children: Array<React.Element<OptionItem>>,
+    ): Array<DropdownItem> {
+        const {isFilterable} = this.props;
+
+        // If it's not filterable, no need to do any extra besides mapping the
+        // option items to dropdown items.
+        if (!isFilterable) {
+            this.mapOptionItemsToDropdownItems(children);
+        }
```
kevinbarabash, line 237: Should this be a return?

packages/wonder-blocks-dropdown/components/single-select.js
```diff
@@ -175,36 +194,90 @@ export default class SingleSelect extends React.Component<Props, State> {
...
+        const lowercasedSearchText = searchText.toLowerCase();
+
+        // Filter the children with the searchText if any.
+        const filteredChildren = children.filter(
+            ({props}) =>
+                !searchText ||
+                props.label.toLowerCase().indexOf(lowercasedSearchText) > -1,
+        );
+
+        return this.mapOptionItemsToDropdownItems(filteredChildren);
```
kevinbarabash, line 250: style nit: I think this function could be simplified a bit like so:
```
return this.mapOptionItemsToDropdownItems(
  isFilterable ? this.filterChildren(children) : children
);
```
with the filtering logic living in a separate method.

packages/wonder-blocks-dropdown/components/single-select.js
```diff
@@ -148,6 +166,7 @@ export default class SingleSelect extends React.Component<Props, State> {
         this.setState({
             open: opened,
             keyboard,
+            searchText: "",
```
kevinbarabash, line 169: I wonder if there's any situations where we'd want to maintain the search text. 🤔 No action req'd.

packages/wonder-blocks-dropdown/components/__tests__/single-select.test.js
```diff
@@ -1,12 +1,16 @@
...
 import {mount, unmountAll} from "../../../../utils/testing/mount.js";
 
 import DropdownOpener from "../dropdown-opener.js";
 import SelectOpener from "../select-opener.js";
 import OptionItem from "../option-item.js";
 import SingleSelect from "../single-select.js";
 import {keyCodes} from "../../util/constants.js";
+import SearchTextInput from "../search-text-input.js";
+
+jest.useFakeTimers();
```
kevinbarabash, line 13: You should probably only call this in a `beforeEach()` since if another test calls `jest.useRealTimers()` then this wouldn't do anything.

packages/wonder-blocks-dropdown/components/__tests__/search-text-input.test.js
```diff
@@ -0,0 +1,123 @@
+// @flow
+import * as React from "react";
+import {shallow} from "enzyme";
+
+import SearchTextInput from "../search-text-input.js";
```
kevinbarabash, line 5: naming suggestion: `SearchField`.  You using this terminology elsewhere in this PR, but beyond that I think it generalizes better, e.g.
- `TextField`
- `PasswordField`
- `SearchField`

packages/wonder-blocks-dropdown/components/__tests__/search-text-input.test.js
```diff
@@ -0,0 +1,123 @@
...
+    });
+
+    it("search should be dismissed if the close icon is clicked", () => {
+        // Arrange
+        const onClickMock = jest.fn();
+
+        const wrapper = shallow(
+            <SearchTextInput
+                searchText="query"
+                testId="search-text-input"
```
kevinbarabash, line 82: nit: is the `testId` necessary?

packages/wonder-blocks-dropdown/components/__tests__/search-text-input.test.js
```diff
@@ -0,0 +1,123 @@
+// @flow
+import * as React from "react";
+import {shallow} from "enzyme";
+
+import SearchTextInput from "../search-text-input.js";
+
+describe("SearchTextInput", () => {
+    it("text input container should be focused when focusing on the input", () => {
+        // Arrange
+        const wrapper = shallow(
```
kevinbarabash, line 10: Nice use of `shallow`. 🎉

packages/wonder-blocks-dropdown/__tests__/generated-snapshot.test.js
```diff
@@ -669,7 +669,7 @@ describe("wonder-blocks-dropdown", () => {
             darkBackgroundWrapper: {
                 flexDirection: "row",
                 justifyContent: "flex-end",
-                backgroundColor: Color.darkBlue,
+                background: Color.darkBlue,
```
kevinbarabash, line 672: Why the switch to `background`?

packages/wonder-blocks-dropdown/components/single-select.js
```diff
@@ -105,6 +116,12 @@ type State = {|
      * Whether or not last opened state change was triggered by a keyboard click.
      */
     keyboard?: boolean,
+
+    /**
+     * The text input to filter the items by their label. Defaults to an empty
+     * string.
+     */
+    searchText: string,
```
kevinbarabash, line 124: Is this something that we'll want to have on `MultiSelect` and `ActionMenu` as well?  If so, it might make sense to push some of the logic you added to this component into dropdown-core.js.

packages/wonder-blocks-dropdown/components/single-select.js
```diff
@@ -271,24 +345,41 @@ export default class SingleSelect extends React.Component<Props, State> {
...
                 light={light}
                 onOpenChanged={this.handleOpenChanged}
                 open={this.state.open}
                 opener={opener}
                 openerElement={this.openerElement}
                 style={style}
+                onSearchTextChanged={
+                    isFilterable ? this.handleSearchTextChanged : null
+                }
+                searchText={isFilterable ? searchText : ""}
```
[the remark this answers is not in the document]
jandrade: This was already part of `dropdown-core.js` because J added this when she created the same functionality for the `MultiSelect` functionality. 
You can see the PR here: https://github.com/Khan/wonder-blocks/pull/479

packages/wonder-blocks-dropdown/components/single-select.stories.js
```diff
@@ -0,0 +1,72 @@
...
+const styles = StyleSheet.create({
+    row: {
+        flexDirection: "row",
+    },
+    fullBleed: {
+        width: "100%",
+    },
+});
+
+export const customStyles = () => <SingleSelectWithCustomStyles />;
```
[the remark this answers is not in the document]
jandrade: Good call, I'll change that. About hooks, I thought about that, but I'd prefer to keep using classes as examples here in case folks using Webapp can refer to these snippets in a way they can directly use in Webapp.

I'll gladly replace these examples once we fully enable hooks in Webapp.

Re: aside: I'll be reusing those stories in my next PR (Storybook Docs)... wait for it :)

packages/wonder-blocks-dropdown/components/__tests__/search-text-input.test.js
jandrade: I agree, but I'd prefer to keep it that way to avoid modifying the functionality we already have in place for `MultiSelect`. I wanted to touch as less as I could here to avoid doing any unnecessary refactors on the original implementation. As you mentioned I'll add a new ticket/TODO to work on this in the future.

packages/wonder-blocks-dropdown/components/single-select.js
```diff
@@ -94,6 +99,12 @@ type Props = {|
      * element to access pointer event state.
      */
     opener?: (openerProps: OpenerProps) => React.Element<any>,
+
+    /**
+     * When this is true, the dropdown body shows a search text input at the
+     * top. The items will be filtered by the input.
+     */
+    isFilterable?: boolean,
```
somewhatabstract, line 107: I wonder if we should allow the filter algorithm to be overridden?

There are some really handy filtering rules besides "contains case insensitive match" such as the filtering rules that VSCode has for finding files, like matching starts of words. For example, ABC would match "Always Be Closing".

packages/wonder-blocks-dropdown/components/single-select.js
jandrade: I added it here because we are using allChildren in two different methods (`getMenuItems` and `renderOpener`). What do you think if we keep it that way to prevent calling `React.Children.toArrray` in two different places?

packages/wonder-blocks-dropdown/components/single-select.md
```diff
@@ -505,4 +505,74 @@ class SingleSelectWithCustomOpener extends React.Component {
...
+        width: "100%",
+    },
+});
+
+
+const optionItems = new Array(1000).fill(null).map((_, i) => (<OptionItem
+    key={i}
+    value={(i + 1).toString()}
+    label={`School ${i + 1} in Wizarding World Some more really long labels?`}
+/>));
```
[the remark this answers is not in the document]
jandrade: Oh good catch, I'll fix the banana part and will update the example to iterate over a list of fruits.

packages/wonder-blocks-dropdown/__tests__/generated-snapshot.test.js
jandrade: For some reason, the tests were throwing a warning about using `background` and `backgroundColor`, but I don't remember what was, and now I don't see it, so I 'll revert back to `bgColor`.

packages/wonder-blocks-dropdown/components/__tests__/single-select.test.js
```diff
@@ -388,4 +394,147 @@ describe("SingleSelect", () => {
...
+                    <OptionItem label="item 2" value="2" />
+                    <OptionItem label="item 3" value="3" />
+                </SingleSelect>,
+            );
+
+            // Act
+            wrapper.setState({open: true, searchText: "text"});
+
+            // Assert
+            expect(wrapper.find(IconButton).exists()).toBe(true);
```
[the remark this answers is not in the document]
jandrade: Now that you mention it, I think it makes more sense to move this test inside `search-text-input.test.js`. The `SingleSelect` shouldn't worry about these specific details.

packages/wonder-blocks-dropdown/components/single-select.js
kevinbarabash: I didn't realize that.  Let's keep it as is.

packages/wonder-blocks-dropdown/components/single-select.js
kevinbarabash: @lillialexis wrote a pretty awesome smart matching algorithm for our github CLI tools would be nice to use.

lillialexis: I did write a filter function that does exactly that @somewhatabstract! https://github.com/Khan/our-lovely-cli/blob/develop/cli/pull-request/smart-suggest.js

.github/workflows/node-ci.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - name: Use Node.js 12.x
         uses: actions/setup-node@v1
         with:
-          node-version: 12.x
+          node-version: 12.16.x
```
jandrade, line 16: I changed this b/c storybook/chromatic was failing to build: https://github.com/storybookjs/storybook/issues/10477#issuecomment-635075061

.github/workflows/node-ci.yml
```diff
@@ -35,5 +35,5 @@ jobs:
       - uses: chromaui/action@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          appCode: ${{ secrets.CHROMATIC_APP_CODE }}
+          projectToken: ${{ secrets.CHROMATIC_APP_CODE }}
```
jandrade, line 38: Changed it to remove a Chromatic warning.

.github/workflows/node-ci.yml
kevinbarabash: Hopefully they'll fix the underlying issue in the future.

packages/wonder-blocks-dropdown/components/__tests__/single-select.test.js
```diff
@@ -388,4 +394,147 @@ describe("SingleSelect", () => {
...
+            wrapper.setState({searchText: "some text"});
+            const opener = wrapper.find(SelectOpener);
+
+            // Act
+            opener.simulate("click");
+
+            // Assert
+            expect(wrapper).toHaveState({searchText: ""});
+        });
+    });
```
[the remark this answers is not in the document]
kevinbarabash: @jandrade friendly ping.

packages/wonder-blocks-dropdown/components/single-select.js
kevinbarabash: I see now that `searchText` is already part of `DropdownCore`.

packages/wonder-blocks-dropdown/components/__tests__/single-select.test.js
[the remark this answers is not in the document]
jandrade: Oh I forgot this comment. Thanks for the reminder.

PR #951, opened 2020-08-12
WB-975: Propagate testId to CloseButton within PopoverContentCore

shahhyash:
## Summary

The close button within the `PopoverContentCore` component accepts a `testId` property but did not propagate this to the close button for the popover. This fixes that and makes writing e2e tests easier where the popover needs to be dismissed and cannot click the action item available.

In scenarios where the `testId` is not provided, we can simply set the testId of the closeButton to "popover-close-btn" as a default value.

Issue: https://khanacademy.atlassian.net/browse/WB-975

## Test Plan

- Add a testId to the `PopoverContentCore` component and ensure the `closeButtonVisible` prop is set to true.
- Inspect the close button and ensure that the `data-test-id` attribute of the close button is the provided id appended with "-close-btn".

### DIFF
--- a/packages/wonder-blocks-popover/__tests__/__snapshots__/generated-snapshot.test.js.snap
+++ b/packages/wonder-blocks-popover/__tests__/__snapshots__/generated-snapshot.test.js.snap
@@ -660,6 +660,7 @@ exports[`wonder-blocks-popover example 6 1`] = `
     <button
       aria-label="Close Popover"
       className=""
+      data-test-id="popover-close-btn"
       disabled={false}
       onBlur={[Function]}
       onClick={[Function]}
@@ -1350,6 +1351,7 @@ exports[`wonder-blocks-popover example 9 1`] = `
     <button
       aria-label="Close Popover"
       className=""
+      data-test-id="popover-close-btn"
       disabled={false}
       onBlur={[Function]}
       onClick={[Function]}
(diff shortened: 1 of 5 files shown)


shahhyash:
[clabot:check]

packages/wonder-blocks-popover/components/popover-content-core.js
```diff
@@ -85,6 +85,7 @@ export default class PopoverContentCore extends React.Component<Props> {
                         aria-label={closeButtonLabel}
                         light={closeButtonLight || color !== "white"}
                         style={styles.closeButton}
+                        testId={`${testId || "popover"}-close-btn`}
```
jandrade, line 88: Nice approach!

packages/wonder-blocks-popover/__tests__/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -1350,6 +1351,7 @@ exports[`wonder-blocks-popover example 9 1`] = `
     <button
       aria-label="Close Popover"
       className=""
+      data-test-id="popover-close-btn"
```
jandrade, line 1354: Just one minor request here. Could you please add a `testId` prop to this example to verify that we are overriding the `data-test-id` value properly?

You can modify that example in: https://github.com/Khan/wonder-blocks/blob/master/packages/wonder-blocks-popover/components/popover-content-core.md#L41

packages/wonder-blocks-popover/__tests__/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -1575,6 +1577,76 @@ exports[`wonder-blocks-popover example 10 1`] = `
       }
     }
   >
+    <button
+      aria-label="Close Popover"
+      className=""
+      data-test-id="example-popover-close-btn"
```
shahhyash, line 1583: @jandrade modifying the example in `popover-content-core.md` modified the snapshot here. I think that the snapshot you noted above is for an example in `docs.md`.

jandrade: oh ok. Thanks for making this change!

PR #1011, opened 2020-09-11
Make the 'style' prop that addStyle adds optional

kevinbarabash:
## Summary:
If you're doing 'const StyledList = addStyle("ul", styles.list)' and there's
no custom style for an invidual 'StyledList' on the page then you shouldn't
be required to provide a 'style' prop to 'StyledList'.  This usage is more like
how the styled-components library functions.

Issue: none

## Test plan:
- yarn test


jeresig (approved):
Great! This seems like a safe change to me... can you think of anything that might trip folks up with this?

kevinbarabash:
> can you think of anything that might trip folks up with this?

It shouldn't since it's a non-breaking change as verified by the flow tests I added.

PR #1047, opened 2020-10-06
WEB-2990.1: Add beforeNav and safeWithNav to Button

kevinbarabash:
## Summary:
This PR add the new props `beforeNav` and `safeWithNav` to `Button` and `ClickableBehavior`.  These props were proposed in [ADR #390](https://docs.google.com/document/d/1p4UUOHEqhaXG7rVg7J4j6S0a83wF76QcPqnEJvrtqyA/edit) with one minor change, they must always return a promise.  This provides a clear separation of responsibility: synchronous actions should be handled by `onClick` (which can be used with or without an `href`) and asynchronous actions should be handled by `beforeNav` or `safeWithNav` depending on the desired behavior.

Since this PR modifies `ClickableBehavior` a number of other components required minor changes in order for flow type-checking to succeed.  Future PRs will finish updating these components to have `beforeNav` and `safeWithNav` props just like `Button`.

This PR also updates the build configuration so that async functions aren't transformed.

TODO:
- [ ] ~touch events~ this is already handled since we're relying on the mobile browser to synthesize a "click" event for us.
- [x] keyboard events
- [x] dedupe logic between mouse and keyboard event handling of `beforeNav` and `safeWithNav`
- [x] documentation
- [x] add tests for `beforeNav` and `safeWithNav` to clickable-behavior.test.js
- [x] add tests for `spinner` behavior

Issue: WEB-2990

## Test plan:
- yarn jest packages/wonder-blocks-button/components/__tests__/button.test.js

Reviewers: #fe-infra


packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -269,12 +283,43 @@ export default class ClickableBehavior extends React.Component<
...
+        } = this.props;
+
         if (this.enterClick) {
             return;
-        } else if (this.props.onClick) {
+        } else if (onClick) {
             this.waitingForClick = false;
-            this.props.onClick(e);
+            onClick(e);
+        } else if (beforeNav) {
```
aag, line 300: I think it would be better if `onClick`, `beforeNav`, and `safeWithNav` weren't all mutually exclusive. Instead I think it would be useful if they would all be run, in that order. As the ADR says, "If both beforeNav and safeWithNav functions are passed into the component, then beforeNav will be run first, then safeWithNav."

I think supporting all the props gives users a clear way to divide up their code. For example, if I want to update local component state, do a GraphQL mutation, and start an analytics request, I could do `setState` in the `onClick` function, do the mutation in `beforeNav`, and do the analytics request in `safeWithNav`. Then, if I ever don't need one piece of that in the future, I can just remove one of the handlers, and everything else stays the same. If I have to combine them into one handler, then it could be more complicated to remove or add things in the future, since I might have to move the handler to a different prop. Supporting multiple props could also make the user experience better, since client-side navigation could wait for some server requests to finish, but start while others are still in flight. That wouldn't be possible if the user has to manage all requests themselves in a single handler.

The only complicated thing is what to do if both `beforeNav` and `safeWithNav` return promises. I think the best thing would be to run both functions, then wait for the `beforeNav` promise to resolve, then navigate (unless a server-side navigation is going to happen, when we'd have to wait for both promises to resolve).

kevinbarabash: I can see the utility of `beforeNav` and `safeWithNav` being allowed at the same time.  Allowing `onClick` I think will lead to confusion about when to use `onClick` vs `safeWithNav` since they're almost the same thing.

> The only complicated thing is what to do if both beforeNav and safeWithNav return promises. I think the best thing would be to run both functions, then wait for the beforeNav promise to resolve, then navigate (unless a server-side navigation is going to happen, when we'd have to wait for both promises to resolve).

That makes sense.

kevinbarabash: We chatted on Slack and agreed to allow all three props at once with `onClick` being used for synchronous code and the other two being used for aysnc code.

kevinbarabash:
Sorry this initial PR is so big.  The next ones should be much smaller.

build-settings/babel.config.js
```diff
@@ -1,11 +1,22 @@
 /* eslint-disable import/no-commonjs */
 module.exports = {
-    presets: ["@babel/preset-flow", "@babel/preset-react", "@babel/preset-env"],
+    presets: [
+        "@babel/preset-flow",
+        "@babel/preset-react",
+        [
+            "@babel/preset-env",
+            {
+                useBuiltIns: "usage",
```
jeresig, line 9: What are the file size implications of this change? I'm worried that we're going to be shipping many more poly folks for every WB component that we have.

packages/wonder-blocks-button/components/button.js
```diff
@@ -136,6 +134,23 @@ export type SharedProps = {|
      * href is not
      */
     onClick?: (e: SyntheticEvent<>) => mixed,
+
+    /**
+     * Run async code before navigating. If the promise returned rejects then
+     * navigation will not occur.
+     *
+     * If both safeWithNav and beforeNav are provided, beforeNav will be run
+     * first and safeWithNav will only be run if beforeNav doesn't not reject.
```
jeresig, line 143: *does not reject

packages/wonder-blocks-button/components/button.js
```diff
@@ -136,6 +134,23 @@ export type SharedProps = {|
...
+     */
+    beforeNav?: (e: SyntheticEvent<>) => Promise<mixed>,
+
+    /**
+     * Run async code in the background while client-side navigating. If the
+     * navigation is server-side, the callback must either be resolved or
```
jeresig, line 149: You could say: the promise must be settled...

(Which is a more accurate way to say what's happening)

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -269,13 +298,71 @@ export default class ClickableBehavior extends React.Component<
...
+
+        try {
+            if (beforeNav) {
+                this.setState({waiting: true});
+                await beforeNav(e);
+            }
+
+            if (safeWithNav) {
+                if (!this.state.waiting) {
+                    this.setState({waiting: true});
```
jeresig, line 323: I think we only want to set this if we're doing a server-side bag, as otherwise the user won't see any impact on clicking the link.

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -269,13 +298,71 @@ export default class ClickableBehavior extends React.Component<
...
+        const {
+            onClick = undefined,
+            beforeNav = undefined,
+            safeWithNav = undefined,
+            skipClientNav,
+            history,
+        } = this.props;
+        let shouldNavigate = true;
+
+        if (onClick) {
```
jeresig, line 311: It feels like the onClick should happen after the other two callbacks. This way we could do things like track when modals are opened, etc.

jeresig (changes requested):
Very cool! Thank you for working on this! A few issues online.

build-settings/babel.config.js
kevinbarabash: Good call.  I was copying the webapp config without thinking.  After reading the docs, I think the default value of `false` is what we want.

packages/wonder-blocks-core/components/clickable-behavior.js
kevinbarabash: I think it depends.  If we're navigating from `/foo` to `/foo/bar` and the button is visible on both routes then the users will see a spinner in that situation.

jeresig: But we only want that for beforeNav, now safeWithNav. safeWithNav should only show a spinner if it's a server-side link as it doesn't block navigation when on the client.

packages/wonder-blocks-core/components/clickable-behavior.js
kevinbarabash: Sorry, I don't follow.  Could you explain a bit more about how having `onClick` come last would help with this?

jeresig: Right now clickable things can have an onClick or a href. Right now the href is visited after the tracking, whereas the onClick is fired before the tracking. We should ensure this is consistent so that we don't end up with a case where someone puts a redirect in an onClick and is now not being tracked, due to execution order.

kevinbarabash:
One thing that's a little strange is that you can use `beforeNav` and `safeWithNav` even if `href` isn't set.  In that situation the navigation never happens.

packages/wonder-blocks-core/components/clickable-behavior.js
kevinbarabash: I thought the spinner was to indicate that stuff we were doing stuff in the background.  If we want it to indicate that we're navigating then not showing it for client-side nav with `safeWithNav` makes sense.  I'll make that change and add some comments as to why it isn't show for that case.

packages/wonder-blocks-core/components/clickable-behavior.js
kevinbarabash: I see what you're saying.  It feels a bit weird to artificially delay a `onClick` handler.  I thought the purpose of these new props was to move any tracking out of `onClick` so that we didn't have to manually do the `redirect` ourselves.

build-settings/babel.config.js
kevinbarabash: Doing so didn't seem to have an impact on the size of the dist file produced for wonder-blocks-button.

jeresig: Ok - so it's possible that it's already getting bundled then!

packages/wonder-blocks-core/components/clickable-behavior.js
kevinbarabash: Maybe we could track whether the consumer provided `onClick` handler calls `e.preventDefault()` and use that to not redirect.  That way we'd be preserving more of the default behavior.

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -308,25 +308,42 @@ export default class ClickableBehavior extends React.Component<
         } = this.props;
         let shouldNavigate = true;
 
+        // While defaultPrevented is read-only on DOM events, that isn't the
+        // case for SynethicEvents.
+        e.defaultPrevented = false;
```
jeresig, line 313: Why do we need to explicitly set this to `false`? Shouldn't the browser be doing that for us?

styleguide.setup.js
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line import/no-unassigned-import
+import "@babel/polyfill";
```
jeresig, line 2: What's this change doing?

jeresig (approved):
Ok, thank you! Will be curious to see what happens when we start shipping this in production - hopefully the async/await stuff won't break things!

packages/wonder-blocks-core/components/clickable-behavior.js
kevinbarabash: I'll expand on the comment.  The callers of this function call `e.preventDefault()` before we get here so `e.defaultPrevent` is set to `true` by this point.  We want to determine if `onClick(e)` is also calling `e.preventDefault()` so we have to reset the value of `e.defaultPrevent` ourselves before calling `onClick(e)`.

styleguide.setup.js
kevinbarabash: I'll add a comment to this file as well, but basically styleguidist doesn't include `regeneratorRuntime` when it builds its files which means that nothing loads when using `async` functions.  The built files should just work in webapp since we're already including `regeneratorRuntime` there.  https://github.com/styleguidist/react-styleguidist/issues/162

packages/wonder-blocks-core/components/clickable-behavior.js
kevinbarabash: I ended up simplify things by not calling `e.preventDefault()` before calling this function.

jeresig (approved):

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -357,18 +454,9 @@ export default class ClickableBehavior extends React.Component<
             (triggerOnEnter && keyCode === keyCodes.enter) ||
             (triggerOnSpace && keyCode === keyCodes.space)
         ) {
-            // NOTE: In Firefox, the keyup event needs to be preventDefault-ed
-            // for space.
-            // https://stackoverflow.com/questions/24383722/unable-to-prevent-spacebar-default-action-in-firefox
-            if (keyCode === keyCodes.space) {
-                e.preventDefault();
-            }
```
kevinbarabash, line 365: I don't see any harm in always calling `e.preventDefault()` since we call it inside of `runCallbackAndMaybeNavigate` and if we always called it here then we may as well just move the call inside of `runCallbackAndMaybeNavigate` and be done with it.

kevinbarabash:
> hopefully the async/await stuff won't break things!

It's using `regeneratorRuntime` so as long as we're loading that in webapp it should be fine.

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -269,13 +298,81 @@ export default class ClickableBehavior extends React.Component<
...
+        e.preventDefault();
+
+        try {
+            if (beforeNav) {
+                this.setState({waiting: true});
+                await beforeNav();
+            }
+
+            if (safeWithNav) {
+                if (history && !skipClientNav) {
```
aag, line 331: Shouldn't this be `history && skipClientNav`? It seems like we're trying to cover the case where we don't do client-side navigation, which I think happens when `skipClientNav` is true.

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -269,13 +298,81 @@ export default class ClickableBehavior extends React.Component<
...
+                    if (!this.state.waiting) {
+                        // We only show the spinner for safeWithNav when doing
+                        // a server-side navigation since since the spinner is
+                        // indicating that we're waiting for navigation to occur.
+                        this.setState({waiting: true});
+                    }
```
aag, line 337: I think this `if` block is in the wrong branch of `if (history && !skipClientNav) {`, is that possible? If not, I think the comment is confusing.

kevinbarabash: Good catch.  Yes, the `setState` should go down below with the `await safeWithNav()`.

PR #1052, opened 2020-10-08
WEB-2990.2: Add beforeNav and safeWithNav props to Link

kevinbarabash:
## Summary:
This builds on the initial work from #1047 to add the same props to Link.  I didn't both creating navigation examples like I did for Button since the docs for Button are quite a bit more extensive than they are for Link.  The props show up in the docs and that should be sufficient.

Issue: WEB-2990

## Test plan:
- yarn test

### DIFF
--- a/packages/wonder-blocks-link/components/__tests__/link.test.js
+++ b/packages/wonder-blocks-link/components/__tests__/link.test.js
@@ -13,7 +13,14 @@ const wait = (delay: number = 0) =>
 
 describe("Link", () => {
     beforeEach(() => {
+        // Note: window.location.assign needs a mock function in the testing
+        // environment.
+        window.location.assign = jest.fn();
         unmountAll();
+    });
+
+    afterEach(() => {
+        window.location.assign.mockClear();
     });
 
     describe("client-side navigation", () => {
@@ -42,7 +49,7 @@ describe("Link", () => {
             expect(wrapper.find("#foo").exists()).toBe(true);
         });
 
-        test("navigation with unknown URL fails", () => {
+        test("navigation to without route does not render", () => {
             // Arrange
             const wrapper = mount(
                 <MemoryRouter>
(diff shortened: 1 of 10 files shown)


packages/wonder-blocks-link/components/__tests__/link.test.js
```diff
@@ -5,83 +5,310 @@ import {MemoryRouter, Route, Switch} from "react-router-dom";
...
+
+            // Act
+            const buttonWrapper = wrapper.find(`[data-test-id="link"]`).first();
+            buttonWrapper.simulate("click", {button: 0});
+
+            // Assert
+            expect(wrapper.find("#foo").exists()).toBe(true);
+        });
+
+        test("navigation with unknown URL fails", () => {
```
aag, line 45: Nitpick: I find this test name a bit confusing, since it's really about route-based rendering and not navigation (which I associate more with the URL in the URL bar). Suggestion: "navigation to URL without route does not render".

packages/wonder-blocks-link/components/__tests__/link.test.js
```diff
@@ -5,83 +5,310 @@ import {MemoryRouter, Route, Switch} from "react-router-dom";
...
+            const buttonWrapper = wrapper.find(`[data-test-id="link"]`).first();
+            buttonWrapper.simulate("click", {
+                button: 0,
+            });
+            await wait(0);
+            wrapper.update();
+
+            // Assert
+            expect(window.location.assign).toHaveBeenCalledWith("/foo");
+        });
```
aag, line 312: I'd like to see a test for `doesn't redirect before beforeNav resolves when skipClientNav=true` in addition to the ones that already exist.

packages/wonder-blocks-link/components/link.js
```diff
@@ -100,6 +100,23 @@ export type SharedProps = {|
      * stubbed out.
      */
     onClick?: (e: SyntheticEvent<>) => mixed,
+
+    /**
+     * Run async code before navigating. If the promise returned rejects then
```
aag, line 105: I think it would be useful to be explicit about "navigating" here referring to navigating to the URL passed in to the `href` prop.

aag (commented):
I left a few minor comments, but this looks pretty solid to me. I like how little code is needed in the `Link` component, since it just proxies to `ClickableBehavior`. I'd like to see the additional test I mention inline before I approve.

packages/wonder-blocks-link/components/__tests__/link.test.js
kevinbarabash: That's fair.  I'll change this to be more accurate.

packages/wonder-blocks-link/components/link.js
```diff
@@ -100,6 +100,23 @@ export type SharedProps = {|
...
+     * promise returned rejects then navigation will not occur.
+     *
+     * If both safeWithNav and beforeNav are provided, beforeNav will be run
+     * first and safeWithNav will only be run if beforeNav does not reject.
+     */
+    beforeNav?: () => Promise<mixed>,
+
+    /**
+     * Run async code in the background while client-side navigating. If the
+     * navigation is server-side, the callback must be settled before the
```
somewhatabstract, line 115: NIT: "the callback promise must be settled"

somewhatabstract (approved):
LGTM

One inline clarification requested.

aag (approved):
LGTM! Thanks for adding the additional test.

PR #1053, opened 2020-10-08
Fix issues from #1047

kevinbarabash:
## Summary:
I didn't put the 'waiting' logic for the server-side nav case in the right location.  This PR fixes that.

One of the things from the ADR was wanted to make sure people are using 'beforeNav' and 'safeWithNav' only when 'href' is set.  This PR also fixes that by changing the Button props to be a disjoint union.

Issue: WEB-2990

## Test plan:
- yarn jest

### DIFF
--- a/packages/wonder-blocks-button/components/__tests__/button.flowtest.js
+++ b/packages/wonder-blocks-button/components/__tests__/button.flowtest.js
@@ -1,0 +1,34 @@
+// @flow
+/* eslint-disable flowtype/no-unused-expressions */
+import * as React from "react";
+
+import Button from "../button.js";
+
+// $ExpectError: href must be used with beforeNav
+<Button beforeNav={() => Promise.resolve()}>Hello, world!</Button>;
+
+// $ExpectError: href must be used with safeWithNav
+<Button safeWithNav={() => Promise.resolve()}>Hello, world!</Button>;
+
+// It's okay to use onClick with href
+<Button href="/foo" onClick={() => {}}>
+    Hello, world!
+</Button>;
+
+<Button href="/foo" beforeNav={() => Promise.resolve()}>
+    Hello, world!
+</Button>;
+
+<Button href="/foo" safeWithNav={() => Promise.resolve()}>
+    Hello, world!
+</Button>;
+
+// All three of these props can be used together
+<Button
+    href="/foo"
+    beforeNav={() => Promise.resolve()}
+    safeWithNav={() => Promise.resolve()}
+    onClick={() => {}}
+>
+    Hello, world!
+</Button>;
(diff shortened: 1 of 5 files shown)


jeresig (approved):

packages/wonder-blocks-button/components/__tests__/button.flowtest.js
```diff
@@ -0,0 +1,34 @@
...
+import * as React from "react";
+
+import Button from "../button.js";
+
+// $ExpectError: href must be used with beforeNav
+<Button beforeNav={() => Promise.resolve()}>Hello, world!</Button>;
+
+// $ExpectError: href must be used with safeWithNav
+<Button safeWithNav={() => Promise.resolve()}>Hello, world!</Button>;
+
```
aag, line 12: Maybe add one more test to verify that `href` can be used by itself?

```
// It's okay to use href by itself
<Button href="/foo">
    Hello, world!
</Button>;
```

aag (approved):
Looks good to me. I suggested 1 additional test, but I'll leave it up to you to add it or not.

packages/wonder-blocks-button/components/button.js
```diff
@@ -134,25 +134,35 @@ export type SharedProps = {|
...
+
+          /**
+           * Run async code in the background while client-side navigating. If the
+           * navigation is server-side, the callback must be settled before the
+           * navigation will occur. Errors are ignored so that navigation is
+           * guaranteed to succeed.
+           */
+          safeWithNav?: () => Promise<mixed>,
+      |};
+
```
somewhatabstract, line 165: Thanks for doing this. Be prepared for webapp issues though if folks spread these props for some reason. Our version of flow in webapp is not good to spread unions.

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -329,15 +329,16 @@ export default class ClickableBehavior extends React.Component<
...
-                        // indicating that we're waiting for navigation to occur.
-                        this.setState({waiting: true});
-                    }
+                    // client-side nav
                     safeWithNav();
                 } else {
                     try {
+                        if (!this.state.waiting) {
+                            // We only show the spinner for safeWithNav when doing
+                            // a server-side navigation since since the spinner is
```
somewhatabstract, line 338: This terminology is a bit confusing (at least for me) as I immediately think of "server-side rendering". Perhaps consider "full page load navigation"?

somewhatabstract (approved):
LGTM

Some inline thoughts. Nothing serious.

packages/wonder-blocks-button/components/button.js
kevinbarabash: Thanks for the heads up.

packages/wonder-blocks-core/components/clickable-behavior.js
kevinbarabash: "full page load navigation" is better.  I'll update this before landing.

PR #1054, opened 2020-10-09
Don't set boolean attributes that don't exist on HTML elements

kevinbarabash:
## Summary:
React (dev-only build I believe) complains about setting boolean
attributes that don't exist on HTML elements.  This PR fixes the
issue.  While this should be an issue in prod, it will lead to
noisy console logs during dev so we should fix it.

Issue: none

## Test plan:
- yarn start
- load localhost:6060
- see that there are no React errors complaining about non-existent boolean attributes

Reviewers: #fe-infra


jeresig (approved):
Ah, tricky! Thank you.

PR #1055, opened 2020-10-09
WEB-2990.?: Use promises instead of async/await

kevinbarabash:
## Summary:
Even though we're using async/await in some places in webapp, the code we generator
here doesn't see webapp's version of regeneratorRuntime.  This PR changes the use
of async/await in clickable-behavior.js to Promises instead.

This PR also removes the changes I made to the styleguidist config to drop the import
of @babel/polyfill which was providing regeneratorRuntime locally.

Issue: WEB-2990

## Test plan:
- yarn start
- load localhost:6060 and see that it loads fine without regeneratorRuntime errors
- yarn test

Reviewers: #fe-infra

### DIFF
--- a/packages/wonder-blocks-core/components/clickable-behavior.js
+++ b/packages/wonder-blocks-core/components/clickable-behavior.js
@@ -298,13 +298,70 @@ export default class ClickableBehavior extends React.Component<
         this.dragging = false;
     }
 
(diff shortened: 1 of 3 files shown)


jeresig (approved):
I'm sorry :( This ends up being a lot more complicated... but glad you had the test suite in place!

PR #1056, opened 2020-10-09
Fix the waiting state

kevinbarabash:
## Summary:
We were showing the spinner but not hiding the text or changing
the background color to grey.  This PR fixes that by setting the
'spinner' prop on ButtonCore to 'true' if 'waiting' is 'true'.

Issue: none

## Test plan:
- yarn start
- load localhost:6060
- click on the "Aysnc action, client-side nav" button
- see that the button looks 'spinner={true}' is set

Reviewers: #fe-infra


jandrade (approved):
LGTM!

kevinbarabash:
Not sure why the automated checks aren't running.  I'll run `yarn test` locally before landing this.

PR #1057, opened 2020-10-09
Avoid ambiguous cases in Button's prop types

kevinbarabash:
## Summary:
After deploying to webapp, a bunch of flow errors were triggered because the flow types for Button's props are ambiguous when only 'href' is being passed.

The error only triggers when passing the result of a function call instead of just passing a string.  I've added a flow test that reproduces the issue.

This PR fixes it by splitting up the possible cases to be more explicit to avoid ambiguous cases.

Issue: none

## Test plan:
- yarn flow

Reviewers: #fe-infra

### DIFF
--- a/packages/wonder-blocks-button/components/button.js
+++ b/packages/wonder-blocks-button/components/button.js
@@ -139,7 +139,7 @@ export type SharedProps = {|
 // We structure the props in this way to ensure that whenever we're using
 // beforeNav or safeWithNav that we're also using href.  We also need to specify
 // a number of different variations to avoid ambigious situations where flow
-// finds more than one valide object type in the disjoint union.
+// finds more than one valid object type in the disjoint union.
 type Props =
     | {|
           ...SharedProps,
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-button/components/button.js
```diff
@@ -136,6 +136,10 @@ export type SharedProps = {|
     onClick?: (e: SyntheticEvent<>) => mixed,
 |};
 
+// We structure the props in this way to ensure that whenever we're using
+// beforeNav or safeWithNav that we're also using href.  We also need to specify
+// a number of different variations to avoid ambigious situations where flow
+// finds more than one valide object type in the disjoint union.
```
somewhatabstract, line 142: SUPER NIT: `valide` -> `valid`

somewhatabstract (approved):
LGTM

aag (approved):
The duplication is unfortunate, but I think this version documents the intended usage better than the previous version, so I would call this a net positive change (even independent of it being necessary for Flow).

PR #1062, opened 2020-10-12
Add 'large' size variant to Button

kevinbarabash:
The reason for adding this variant is that it's currently impossible to create type-safe wrappers around button with the current version of flow that we're using in webapp.  Until we can update flow to a version that fixes this issue we'll have to make do with a "large" size variant in wonder-blocks itself.

Test Plan:
- look at the buttons on the netlify preview page see that they look decent
- hover over all button sizes with and without icons to see that the width doesn't change

### DIFF
--- a/packages/wonder-blocks-button/components/button-core.js
+++ b/packages/wonder-blocks-button/components/button-core.js
@@ -68,6 +68,7 @@ export default class ButtonCore extends React.Component<Props> {
             kind,
             light,
             iconWidth,
+            size,
         );
 
         const disabled = spinner || disabledProp;
@@ -85,6 +86,7 @@ export default class ButtonCore extends React.Component<Props> {
                     ? buttonStyles.active
                     : (hovered || focused) && buttonStyles.focus),
             size === "small" && sharedStyles.small,
+            size === "large" && sharedStyles.large,
         ];
 
         const commonProps = {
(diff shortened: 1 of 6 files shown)


kevinbarabash:
Tomorrow I'll update things to match the [figma](https://www.figma.com/file/OEZEJO0TsWKlqXzZgFeWMm/Button-Size-Classes?node-id=21%3A12).

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -225,22 +241,26 @@ const sharedStyles = StyleSheet.create({
...
+    const buttonType =
+        color + kind + light.toString() + iconWidth.toString() + size;
     if (styles[buttonType]) {
         return styles[buttonType];
     }
 
     const {white, white50, white64, offBlack32, offBlack50, darkBlue} = Color;
     const fadedColor = mix(fade(color, 0.32), white);
     const activeColor = mix(offBlack32, color);
+    const padding = size === "xlarge" ? 32 : 16;
```
jeresig, line 254: We should probably use values from Spacing here

jeresig (approved):
Nice! Looks good!

kevinbarabash:
@jeresig could you also review the changes to the screenshots in chromatic?

jeresig:
@kevinbarabash Done - although I do wonder if this is something that you (the author) should do? Since it's sort of like having the snapshot tests update, you want to confirm that they are producing the right result.

jandrade (approved):
Thanks for adding the stories!

kevinbarabash:
@jeresig I do look at the Chromatic snapshots when there are changes.  It's possible I missed something so I do think it's worth having reviewers look at them as well.

PR #1064, opened 2020-10-14
Add missing props to a few components

kevinbarabash:
## Summary:
In some places we pass "rel" to links in webapp.  I've added this optional prop
to Button, Link, and Clickable as those are most likely to need that prop.  I've
also added "target", and all of our ARIA props to Clickable.

Issue: none

## Test plan:
- yarn test

Reivewers: #fe-infra

### DIFF
--- a/packages/wonder-blocks-clickable/__tests__/__snapshots__/generated-snapshot.test.js.snap
+++ b/packages/wonder-blocks-clickable/__tests__/__snapshots__/generated-snapshot.test.js.snap
@@ -126,6 +126,7 @@ exports[`wonder-blocks-clickable example 2 1`] = `
   <a
     aria-label=""
     className=""
+    disabled={false}
     href="/foo"
     onBlur={[Function]}
     onClick={[Function]}
@@ -170,6 +171,7 @@ exports[`wonder-blocks-clickable example 2 1`] = `
   <a
     aria-label=""
     className=""
+    disabled={false}
     href="/foo"
     onBlur={[Function]}
     onClick={[Function]}
@@ -237,6 +239,7 @@ exports[`wonder-blocks-clickable example 3 1`] = `
   <a
     aria-label=""
     className=""
+    disabled={false}
     href="https://www.khanacademy.org/about/tos"
     onBlur={[Function]}
     onClick={[Function]}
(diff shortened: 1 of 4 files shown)


packages/wonder-blocks-clickable/components/clickable.js
```diff
@@ -217,10 +252,9 @@ export default class Clickable extends React.Component<Props> {
             >
                 {(state, handlers) =>
                     this.getCorrectTag(state, {
-                        // eslint-disable-next-line react/prop-types
-                        "aria-label": this.props["aria-label"],
```
jeresig, line 221: We can remove this as `...restProps` takes care of this?

packages/wonder-blocks-clickable/__tests__/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -126,6 +126,7 @@ exports[`wonder-blocks-clickable example 2 1`] = `
   <a
     aria-label=""
     className=""
+    disabled={false}
```
jeresig, line 129: Is it ok that these are now coming through?

jeresig (approved):
Looks good! A question inline.

packages/wonder-blocks-clickable/components/clickable.js
kevinbarabash: Remove the `data-test-id`?

jeresig: Why did you remove the aria-label? Is it safe to do so?

packages/wonder-blocks-clickable/__tests__/__snapshots__/generated-snapshot.test.js.snap
kevinbarabash: This follows what we're doing for `button`.  We probably shouldn't be setting `disabled={false}` on anything.  I'll change that in later in a separate PR.

packages/wonder-blocks-clickable/components/clickable.js
kevinbarabash: Sorry... misparsed your question.  `AriaProps` includes `aria-label` so now that we're including `AriaProp` in the props for this component spreading `restProps` handles `aria-label` for us.

PR #1065, opened 2020-10-14
Implement 'caret' prop for Link

kevinbarabash:
## Summary:
This was work we never got to.  I'm doing this now because I want to replace a ClientLink in webapp with a wonder-blocks-link Link but it has a one-off caret and I want to do it the right way.  This will also make updating other links with carets in webapp easier in the future.

As part of this PR I had to stop processing .md files with lint-staged becuase it was writing to test files at the same time jest was trying to write them and this resulted in test failures.

Issue: none

## Test plan:
- yarn test
- look at the examples in styleguidist and storybook

Reviewers: #fe-infra

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,10 +117,6 @@
     }
   },
   "lint-staged": {
-    "*.md": [
-      "yarn gen-snapshot-tests",
-      "git add packages/*/__tests__/generated-snapshot.test.js"
-    ],
     "*.js": [
       "eslint --config .eslintrc.js",
       "yarn jest --findRelatedTests",
--- a/packages/wonder-blocks-link/components/link-core.js
+++ b/packages/wonder-blocks-link/components/link-core.js
@@ -10,6 +10,8 @@ import type {
     ClickableHandlers,
     ClickableState,
 } from "@khanacademy/wonder-blocks-core";
+import {icons} from "@khanacademy/wonder-blocks-icon";
+
 import type {StyleDeclaration} from "aphrodite";
 import type {SharedProps} from "./link.js";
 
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -117,10 +117,6 @@
     }
   },
   "lint-staged": {
-    "*.md": [
-      "yarn gen-snapshot-tests",
-      "git add packages/*/__tests__/generated-snapshot.test.js"
-    ],
```
jeresig, line 123: Is there anything else that will generate these snapshot tests? or are they just turned off now?

jeresig (approved):
Glad you were able to add this!

packages/wonder-blocks-link/components/link-core.js
```diff
@@ -61,13 +64,25 @@ export default class LinkCore extends React.Component<Props> {
             ...handlers,
         };
 
+        const caretIcon = (
+            <StyledSvg
+                height="1em" // set the height to match the current font size
```
jandrade, line 69: Neat!

packages/wonder-blocks-link/link.stories.js
```diff
@@ -0,0 +1,54 @@
...
+    <View>
+        <Body>
+            <Link href="#nonexistent-link" caret={true}>
+                Link with caret.
+            </Link>
+            <View style={{fontSize: 30}}>
+                <Link href="#nonexistent-link" caret={true}>
+                    Large link with caret.
+                </Link>
+            </View>
```
jandrade, line 51: Nice to see this working example.

jandrade (approved):
LGTM!

package.json
kevinbarabash: I think we could use husky to run our scripts on all files.  It would take a bit longer to run, but they would run in series and wouldn't affect each other.

kevinbarabash:
@nefaurk could you take a look at how the rendering of the caret with links looks?

kevinbarabash:
It turns out that this wasn't needed.  `caret` isn't part of the design spec and there's only one place in webapp that has a link with a caret.  I'm going to change that usage in webapp to have the caret live outside of the Link.  There were more carets in the past but it seems we've moved away from that design pattern.

PR #1066, opened 2020-10-15
Fix Link props

kevinbarabash:
## Summary:
- remove 'caret' since we don't use this design pattern in webapp anymore
- add 'tabIndex' and snapshot tests to make sure it's working

Issue: none

## Test plan:
- check the new snapshots to see that 'tabindex' was set correctly

Reviewers: #fe-infra

### DIFF
--- a/packages/wonder-blocks-button/__tests__/__snapshots__/custom-snapshot.test.js.snap
+++ b/packages/wonder-blocks-button/__tests__/__snapshots__/custom-snapshot.test.js.snap
@@ -1,5 +1,230 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Button <Link tabIndex={-1}> 1`] = `
+<button
+  className=""
+  disabled={false}
+  onBlur={[Function]}
+  onClick={[Function]}
+  onDragStart={[Function]}
+  onFocus={[Function]}
+  onKeyDown={[Function]}
+  onKeyUp={[Function]}
+  onMouseDown={[Function]}
+  onMouseEnter={[Function]}
+  onMouseLeave={[Function]}
+  onMouseUp={[Function]}
+  onTouchCancel={[Function]}
+  onTouchEnd={[Function]}
+  onTouchStart={[Function]}
+  role="button"
+  style={
+    Object {
+      "::MozFocusInner": Object {
+        "border": 0,
+      },
+      ":focus": Object {
+        "WebkitTapHighlightColor": "rgba(0,0,0,0)",
+      },
+      "alignItems": "center",
+      "background": "#1865f2",
+      "border": "none",
+      "borderRadius": 4,
+      "boxSizing": "border-box",
+      "color": "#ffffff",
+      "cursor": "pointer",
+      "display": "inline-flex",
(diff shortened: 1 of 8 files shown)


packages/wonder-blocks-link/__tests__/__snapshots__/custom-snapshot.test.js.snap
```diff
@@ -761,3 +761,105 @@ exports[`LinkCore kind:secondary href:#non-existent-link light:false visitable:f
...
+        "color": "#8755ee",
+      },
+      "color": "#1865f2",
+      "cursor": "pointer",
+      "outline": "none",
+      "textDecoration": "none",
+    }
+  }
+  tabIndex={0}
+>
```
kevinbarabash, line 794: I don't understand why `tabIndex` is always showing up as `0`.

packages/wonder-blocks-link/components/link.js
```diff
@@ -20,12 +20,6 @@ export type SharedProps = {|
      */
     href: string,
 
-    /**
-     * Whether to display a caret after the text.
-     */
-    // TODO(yejia): Add once we have an Icon component
-    caret: boolean,
```
jeresig, line 27: I'm confused - didn't you just add this in an other PR?

jeresig (approved):

packages/wonder-blocks-link/components/link.js
kevinbarabash: I thought this prop was missing because of the `TODO`.  It turns out `caret` isn't actually part of the spec.

packages/wonder-blocks-button/__tests__/__snapshots__/custom-snapshot.test.js.snap
```diff
@@ -1,5 +1,230 @@
...
+      "paddingRight": 16,
+      "paddingTop": 0,
+      "position": "relative",
+      "textDecoration": "none",
+      "touchAction": "manipulation",
+      "userSelect": "none",
+    }
+  }
+  tabIndex={-1}
+  type="button"
```
kevinbarabash, line 52: It's working correctly now.

packages/wonder-blocks-button/components/button.js
```diff
@@ -273,6 +282,10 @@ export default class Button extends React.Component<Props> {
                     spinner={spinner || state.waiting}
                     skipClientNav={skipClientNav}
                     href={href}
+                    // If tabIndex is provide to the component we allow
+                    // it to override the tabIndex provide to use by
+                    // ClickableBehavior.
+                    tabIndex={tabIndex || clickableTabIndex}
```
kevinbarabash, line 288: Actually fix the problem.

kevinbarabash (commented):
I fixed the issue in `Link`, `Button`, and `IconButton`.  This should probably be fixed in other clickable components as well, but it seems less pressing.  I've filed https://khanacademy.atlassian.net/browse/WB-1000 to track the remaining work.

jeresig (approved):
Tricky, thanks for resolving the tabIndex issue!

PR #1068, opened 2020-10-16
Remove unnecessary dev dependencies

kevinbarabash:
## Summary:
lerna looks at changes to local packages that appear in both dependencies and devDependencies when deciding which modules' versions to bump.  Because of how yarn workspaces works and how we've configured jest and flow all of our dev-tooling just works without listing local modules as devDependencies in local packages.  By removing them from devDependencies we avoid unnecessary version bumping from lerna.

Issue: none

## Test plan:
- yarn test
- yarn start
- yarn start:storybook

Reviewers: #fe-infra

### DIFF
--- a/packages/wonder-blocks-clickable/package.json
+++ b/packages/wonder-blocks-clickable/package.json
@@ -19,10 +19,8 @@
   },
   "peerDependencies": {
     "aphrodite": "^1.2.5",
-    "popper.js": "^1.14.1",
     "prop-types": "^15.6.2",
     "react-dom": "^16.4.1",
-    "react-popper": "^1.0.0",
     "react-router-dom": "^4.2.2",
     "react": "^16.4.1"
   },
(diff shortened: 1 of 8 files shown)


packages/wonder-blocks-clickable/package.json
```diff
@@ -21,12 +21,12 @@
     "aphrodite": "^1.2.5",
     "popper.js": "^1.14.1",
```
jandrade, line 22: Nit: This is out of scope for this PR, but I just noticed that we are including `popper.js` here even that Clickable don't use it.

yarn.lock
```diff
@@ -14434,7 +14434,7 @@ react-docgen@^5.0.0:
     node-dir "^0.1.10"
     strip-indent "^3.0.0"
 
-react-dom@^16.4.1, react-dom@^16.4.3, react-dom@^16.8.3:
+react-dom@^16.4.3, react-dom@^16.8.3:
```
jandrade, line 14437: Nice!

jandrade (approved):
LGTM! Thanks for cleaning up the repo.

jeresig (approved):

PR #1072, opened 2020-10-20
Add support for className to addStyle

kevinbarabash:
## Summary:
I'm doing some work on the Google search results component in webapp and we're doing some pretty hacky stuff (replacing Google <a> tags with our <Link> component).  This is so that we can track conversions without having to use /bigbingo_redirect.  In order to do this properly though Link needs to support the "className" prop.  This PR updates 'addStyle' to handle className properly and since the underlying components Link renders are wrapped in 'addStyle' it too will handle className properly.

Issue: none

## Test plan:
- yarn test

Reviewers: #fe-infra

### DIFF
--- a/packages/wonder-blocks-core/util/add-style.js
+++ b/packages/wonder-blocks-core/util/add-style.js
@@ -20,23 +20,24 @@ export default function addStyle<T: React.AbstractComponent<any> | string>(
         style: StyleType,
         ...
     }) {
-        const {style, ...tmpOtherProps} = props;
+        const {className, style, ...tmpOtherProps} = props;
         // NOTE(jeresig): We need to cast the remaining props to be the right
         // value to ensure that they're typed properly.
         const otherProps: React.ElementConfig<T> = (tmpOtherProps: any);
         const reset =
             typeof Component === "string" ? overrides[Component] : null;
 
-        const {className, style: inlineStyles} = processStyleList([
-            reset,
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-core/util/add-style.js
```diff
@@ -20,23 +20,24 @@ export default function addStyle<T: React.AbstractComponent<any> | string>(
...
+        const {
+            className: aphroditeClassName,
+            style: inlineStyles,
+        } = processStyleList([reset, defaultStyle, style]);
 
         return (
             <Component
                 {...otherProps}
-                className={className}
+                className={[aphroditeClassName, className]
```
jandrade, line 38: How precedence would work in this case? I mean, what prop would get more importance when rendering? `style` or `className`?

I'd like to understand how the result would be impacted if the user sets both props.

e.g. what happens in this case?

```
style.inlineStyle = { color: "red"};
.cssDefinedStyle {
  color: blue;
}

<Link style={styles.inlineStyle} className="cssDefinedStyle" />
```

kevinbarabash: Precedence for CSS classes is based on the order the classes are defined in, not the order they appear in the `className`.

jeresig (approved):

jandrade (approved):
LGTM! Thanks for the answer.

PR #1073, opened 2020-10-20
Add 'className' prop type to commonly used components

kevinbarabash:
## Summary:
In #1072 I added support for 'className' to 'addStyle', but in order to use 'className' on components without flow complaining we also have to add it to the prop types.  For some of the components, the new prop needed to be piped through manually to a child component.  This PR doesn't add 'className' to all components though.  I've filed a followup task to complete this work: https://khanacademy.atlassian.net/browse/WB-1002.

Issue: none

## Test plan:
- yarn start
- in one of the examples for Link, give one of the Links a className, see that it shows up in the inspector
- yarn flow

Screen Shot 2020-10-20 at 2 25 52 PM

Reviewers: #fe-infra

### DIFF
--- a/packages/wonder-blocks-button/components/button.js
+++ b/packages/wonder-blocks-button/components/button.js
@@ -142,6 +142,11 @@ export type SharedProps = {|
     }>>,
     */
 
+    /**
+     * Adds CSS classes to the Button.
+     */
+    className?: string,
+
     // NOTE(jeresig): Currently React Docgen (used by Styleguidist) doesn't
     // support ... inside of an exact object type. Thus we had to move the
     // following propers into this SharedProps, even though they should be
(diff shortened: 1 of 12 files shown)


packages/wonder-blocks-button/components/button.js
```diff
@@ -142,6 +142,11 @@ export type SharedProps = {|
     }>>,
     */
 
+    /**
+     * Adds CSS classes to the Button.
+     */
+    className?: string,
```
jeresig, line 148: I'd love for us to be able to build-in `Array<string>` support at some point!

jeresig (approved):

jeresig (approved):

packages/wonder-blocks-button/components/button.js
kevinbarabash: Yeah, have a `classes` prop would be pretty sweet.

PR #1074, opened 2020-10-22
WB-1004: Update Button to work with <form>s

kevinbarabash:
## Summary:
In the process of adding beforeNav and safeWithNav props, I added a call to e.preventDefault() in response to click, keyboard, and touch events. This unfortunately cause forms with submit buttons to stop working.

To fix the issue, if we're handling an event from a `Button` with `type="submit"` we'll look for ancestor element that's a `<form>` and if we find one we'll dispatch a `submit` event to it.

Issue: WB-1004

## Test plan:
- yarn test
- in styleguidist scroll to the "submit" button example
- click on "submit", see the alert message
- use the keyboard to nav to the "submit" button and press "enter", see the alert message, press "spacebar" and see the alert message

Reviewers: #fe-infra

### DIFF
--- a/packages/wonder-blocks-core/components/clickable-behavior.js
+++ b/packages/wonder-blocks-core/components/clickable-behavior.js
@@ -372,6 +372,7 @@ export default class ClickableBehavior extends React.Component<
             onClick = undefined,
             beforeNav = undefined,
             safeWithNav = undefined,
+            href,
         } = this.props;
         let shouldNavigate = true;
 
@@ -385,8 +386,11 @@ export default class ClickableBehavior extends React.Component<
             shouldNavigate = false;
         }
 
-        // Prevent navigation.
-        e.preventDefault();
+        // Prevent navigation, but only if `href` if set.  This is so that forms
+        // containing a Button with `type="submit"` will still work.
+        if (href) {
+            e.preventDefault();
+        }
 
         if (beforeNav) {
             this.setState({waiting: true});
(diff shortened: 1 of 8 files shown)


kevinbarabash:
There are some issues with the keyboard activation of submit buttons that I have to work through.

jeresig (approved):
Looks good to me!

packages/wonder-blocks-core/components/clickable-behavior.js
```diff
@@ -383,11 +391,26 @@ export default class ClickableBehavior extends React.Component<
...
 
-        // Prevent navigation.
         e.preventDefault();
 
+        if (!href && type === "submit" && canSubmit) {
+            let target = e.currentTarget;
+            while (target) {
+                if (target instanceof window.HTMLFormElement) {
+                    const event = new window.Event("submit");
+                    target.dispatchEvent(event);
```
jandrade, line 404: Maybe it would be good adding a test with nested forms.

jandrade (approved):
LGTM! just added a comment, but nothing blocking.

packages/wonder-blocks-core/components/clickable-behavior.js
kevinbarabash: That's a good idea.  I assume it's only supposed to trigger the innermost form.

kevinbarabash: Actually, according to https://stackoverflow.com/questions/379610/can-you-nest-html-forms nested forms is not allowed.

PR #1079, opened 2020-11-06
Move ClickableBehavior and getClickable to wonder-blocks-clickable [BREAKING]

kevinbarabash:
## Summary:
Recent changes to ClickableBehavior have resulted in having to bump all packages that use wonder-blocks-core regardless of whether or not they actually make use of ClickableBehavior.  Moving ClickableBehavior and getClickable into the wonder-blocks-clickable package should reduce the frequency of version bumps to wonder-blocks-core.

Issue: none

## Test plan:
- yarn test

Reviewers: #fe-infra


jeresig (approved):
Looks good! Having this be in one place makes sense...

PR #1080, opened 2020-11-09
Set target on underlying element

kevinbarabash:
## Summary:

Issue: none

## Test plan:
- yarn test

Reviewers: #fe-infra

### DIFF
--- a/packages/wonder-blocks-button/components/button.js
+++ b/packages/wonder-blocks-button/components/button.js
@@ -297,6 +297,7 @@ export default class Button extends React.Component<Props> {
                     spinner={spinner || state.waiting}
                     skipClientNav={skipClientNav}
                     href={href}
+                    target={target}
                     type={type}
                     // If tabIndex is provide to the component we allow
                     // it to override the tabIndex provide to use by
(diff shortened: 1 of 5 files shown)


packages/wonder-blocks-button/components/button.js
```diff
@@ -297,6 +297,7 @@ export default class Button extends React.Component<Props> {
                     spinner={spinner || state.waiting}
                     skipClientNav={skipClientNav}
                     href={href}
+                    target={target}
```
jandrade, line 300: What's the reason for adding `target` to button? I think we should make the distinction whether we use `Link` vs `<button />`.

jandrade: To expand a bit, I think that button should use `formtarget` internally.

kevinbarabash: It's only mean to be used when `href` is set.  In that situation `<Button>` will render as either a `<a>` or a `<Link>`.  I haven't enforced this constraint though.

kevinbarabash: Here's the comment from the prop:
```
    /**
     * A target destination window for a link to open in. Should only be used
     * when `href` is specified.
     */
    target?: string,
```

jandrade: Ohh that makes sense, thanks!

jandrade (approved):
LGTM!

PR #1081, opened 2020-12-01
Spacing: Add number suffixes to keys

bryanjclark:
This allows developers to use autocomplete, which dramatically simplifies the ergonomics of working with the Spacing component!

This is the same thing we did in mobile back in this PR, and it has been super nice to have:
https://github.com/Khan/mobile/pull/599

This is going to break SO MANY THINGS - so I'd recommend making a second PR that greps for all the Spacing usage and replaces it with these values!

(Another option could be to just add these values to the enum, rather than replacing existing ones?)


kevinbarabash:
This PR should probably do the renaming as well so that we don't end up with a commit that leaves things in a broken state.  There will need to be a separate diff/PR to update all the places in webapp as well, but that's something I can tackle.

jandrade (changes requested):
Thanks for tackling this. +1 for what Kevin said about renaming all the instances used inside this repo. 

I think it's definitively better to cover that here, so the website is not broken after merging into `master`. This would include updating Markdown files (used by React Styleguidist) and probably updating snapshot/unit tests.

For reference see the deployed site for this PR: https://deploy-preview-1081--wonder-blocks.netlify.app/

packages/wonder-blocks-spacing/index.js
```diff
@@ -1,17 +1,20 @@
...
+    xSmall_8: 8,
+    small_12: 12,
+    medium_16: 16,
+    large_24: 24,
+    xLarge_32: 32,
+    xxLarge_48: 48,
+    xxxLarge_64: 64,
 };
 
+
```
jandrade, line 16: I think prettier will complain about this.

packages/wonder-blocks-clickable/__tests__/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -152,7 +152,7 @@ exports[`wonder-blocks-clickable example 2 1`] = `
         "font": "inherit",
         "lineHeight": "normal",
         "margin": 0,
-        "marginRight": 24,
+        "marginRight": undefined,
```
kevinbarabash, line 155: Well this is definitely wrong.

jandrade (approved):
LGTM. Thanks for replacing all the instances!

lillialexis:
Haha, I was just thinking about this yesterday!

bryanjclark:
@kevinbarabash would you mind commandeering this one? You've done all the replacements here and I'm happy to resign as driver of this diff!

kevinbarabash:
@bryanjclark no need for me to commandeer.  It was a collaborative PR. 🙂

PR #1082, opened 2020-12-07
WB-1003, WB-857: Update Lato for extended support of cyrillic and vietnamese alphabets

jandrade:
## Summary:
- Added lato font files to the repository so we can test the same font files we currently use in webapp.
- Added `Lato` to support cyrillic alphabets (e.g. Russian, serbian).
- Added `Lato extended` to add full support for vietnamese.
- Fixed Armenian font.

Issue: https://khanacademy.atlassian.net/browse/WB-1003 https://khanacademy.atlassian.net/browse/WB-857

## Test plan:

Navigate to /#!/Typography and verify that in the following examples we render the correct fonts:

1. Cyrillic: `Lato is used for Latin and Cyrillic languages` should use the lato (non-latin) version.
https://deploy-preview-1082--wonder-blocks.netlify.app/#!/Typography/6
Screen Shot 2020-12-07 at 12 47 36 PM

2. Vietnamese: `Lato (Latin and Latin extended) for Vietnamese` should use LatoLatin and LatoLatinExtended.
https://deploy-preview-1082--wonder-blocks.netlify.app/#!/Typography/8
Screen Shot 2020-12-07 at 12 48 12 PM


static/lato.css
```diff
@@ -0,0 +1,89 @@
...
+@font-face {
+    font-family: 'Lato';
+    font-style: normal;
+    font-weight: 900;
+    src: local('Lato Black'), local('Lato-Black'), url(fonts/lato/Lato-Black.woff2) format('woff2');
+}
+
+/* latin */
+@font-face {
+    font-family: 'Lato';
```
jeresig, line 34: Since this font-family is the same as the one above, and doesn't have any unicode ranges, won't that mean that we end up overriding the regular one with the latin one?

jeresig (approved):
This generally seems ok to me, but I have some questions, I think I might be not understanding something!

static/lato.css
jandrade: Good point, I'll add unicode ranges to the ones above (cyrillic).

jandrade: @jeresig I'm going to implement these changes in parallel on webapp so I'll make sure to pass the correct fonts and unicodes. I'll update this PR in case I found something else missing.

jeresig: Ok, sounds good!

jeresig (approved):
Thanks for adding the ranges!

static/lato.css
jandrade: Ugh.. so far these are the results of webapp using the new fonts:

### LOHP Overall font size

Without new fonts: 167 KB
New fonts included (current PR): 393 KB

I've noticed that the website is loading the LatinExtended font (where it shouldn't be). I'll have to make some teaks and will come back here after getting better results.

static/lato.css
```diff
@@ -0,0 +1,93 @@
...
+    unicode-range: U+0400-04FF, U+0500-052F, U+2DE0-2DFF, U+A640-A69F, U+1D00-1D7F;
+}
+
+@font-face {
+    font-family: 'Lato';
+    font-style: normal;
+    font-weight: 400;
+    src: local('Lato Regular'), local('Lato-Regular'), url(fonts/lato/Lato-Regular.woff2) format('woff2');
+    unicode-range: U+0400-04FF, U+0500-052F, U+2DE0-2DFF, U+A640-A69F, U+1D00-1D7F;
+}
```
kevinbarabash, line 18: There was some discussion in Slack about using `:lang()` pseudo selector when deciding which font files to load.  Why aren't we using it?

static/lato.css
```diff
@@ -0,0 +1,93 @@
...
+
+/* Latin Extended (e.g. vietnamese) */
+@font-face {
+    font-family: 'Lato';
+    font-style: italic;
+    font-weight: 400;
+    src: local('Lato Italic'), local('Lato-Italic'), url(fonts/lato/LatoLatinExtended-Italic.woff2) format('woff2');
+    unicode-range: U+0020-002F, U+0030-0039, U+003A-0040, U+0041-005A, U+005B-0060, U+0061-007A, U+007B-007E, U+00C0-00C3, U+00C8-00CA, U+00CC-00CD, U+00D0, U+00D2-00D5, U+00D9-00DA, U+00DD, U+00E0-00E3, U+00E8-00EA, U+00EC-00ED, U+00F2-00F5, U+00F9-00FA, U+00FD, U+0102-0103, U+0110-0111, U+0128-0129, U+0168-0169, U+01A0-01B0, U+1EA0-1EF9, U+02C6-0323;
+}
```
kevinbarabash, line 72: Some of the ranges overlap what the non-extended LatoLatin fonts provide.  Are the glyphs different or are we trying to avoid both LatoLatinExtended and LatoLatin on Vietnamese pages?

static/noto.css
```diff
@@ -31,29 +31,29 @@
 @font-face {
     font-family: "Noto Serif";
     font-weight: 400;
-    src: url("fonts/armenian/NotoSansArmenian-Regular.woff2");
-    unicode-range: U+0530–U+058F;
```
kevinbarabash, line 35: Was the second `U+` causing an issue?

static/lato.css
```diff
@@ -0,0 +1,93 @@
+/* Adapted from https://fonts.googleapis.com/css?family=Lato:400,400i,700,900 */
+
+/* non-latin (cyrillic, hungarian, serbian) */
+@font-face {
+    font-family: 'Lato';
+    font-style: italic;
+    font-weight: 400;
+    src: local('Lato Italic'), local('Lato-Italic'), url(fonts/lato/Lato-Italic.woff2) format('woff2');
```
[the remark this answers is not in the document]
jandrade: Yes! https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face#Description

Side note: I copy pasted what we have in webapp for compatibility issues.

static/lato.css
```diff
@@ -0,0 +1,93 @@
...
+    unicode-range: U+0400-04FF, U+0500-052F, U+2DE0-2DFF, U+A640-A69F, U+1D00-1D7F;
```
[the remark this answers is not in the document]
jandrade: Absolutely! that's a great idea indeed.

static/lato.css
jandrade: Yeah, but the problem is that we could not do that currently because we are hard-coding the font-family values inside each Typography component. Maybe we can explore that possibility once we try a different CSS approach.

static/lato.css
jandrade: Agree with your observation... I'm running some tests on webapp and will change the ranges based on my findings :)

static/noto.css
jandrade: Yeah!

static/lato.css
kevinbarabash: One of the downsides of CSS-in-JS.

kevinbarabash (approved):
LGTM.  Thanks for updating the ranges and answering all my questions.

PR #1087, opened 2021-02-05
WB-1031: don't defer opening of a URL in a new tab

kevinbarabash:
## Summary:
We were deferring the opening of a new URL in a new tab and this was trigger the action to be block by popup blockers built into some browsers (Safari and Firefox).  Deferring navigation is necessary if the navigation is happening in the same tab, but in this case it isn't.  The originating tab is staying around so it can do whatever it needs to do in 'safeWithNav' **while** the other tab opens.

NOTE: using "target='_blank'" with 'beforeNav' won't work and should be disallowed.

Issue: WB-1031

## Test plan:
- run unit tests
- yarn start, open localhost:6060 in Safari or Firefox
- click on the new "Async action, open URL in new tab" button, see that it opens Google in a new tab

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,6 @@
 module.exports = {
     extends: ["@khanacademy"],
-    plugins: ["import", "promise", "monorepo"],
+    plugins: ["import", "jest", "promise", "monorepo"],
     settings: {
         react: {
             version: "detect",
@@ -37,6 +37,7 @@ module.exports = {
                 ignorePackages: true,
             },
         ],
+        "jest/no-focused-tests": "error",
         "promise/always-return": "error",
         "promise/no-return-wrap": "error",
         "promise/param-names": "error",
--- a/packages/wonder-blocks-clickable/components/__tests__/clickable-behavior.test.js
+++ b/packages/wonder-blocks-clickable/components/__tests__/clickable-behavior.test.js
@@ -1038,6 +1038,7 @@ describe("ClickableBehavior", () => {
 
     describe("target='_blank'", () => {
         it("opens a new tab", () => {
+            // Arrange
             const link = mount(
                 <ClickableBehavior
                     disabled={false}
@@ -1054,8 +1055,11 @@ describe("ClickableBehavior", () => {
                     }}
                 </ClickableBehavior>,
             );
+
+            // Act
             link.simulate("click");
 
+            // Assert
             expect(window.open).toHaveBeenCalledWith(
                 "https://www.khanacademy.org",
                 "_blank",
@@ -1063,13 +1067,15 @@ describe("ClickableBehavior", () => {
         });
 
         it("opens a new tab when using 'safeWithNav'", () => {
+            // Arrange
+            const safeWithNavMock = jest.fn().mockResolvedValue();
             const link = mount(
                 <ClickableBehavior
                     disabled={false}
                     href="https://www.khanacademy.org"
                     role="link"
                     target="_blank"
-                    safeWithNav={() => Promise.resolve()}
+                    safeWithNav={safeWithNavMock}
                 >
                     {(state, handlers) => {
                         return (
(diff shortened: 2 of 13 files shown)


packages/wonder-blocks-clickable/components/__tests__/clickable-behavior.test.js
```diff
@@ -20,7 +20,7 @@ const wait = (delay: number = 0) =>
         return setTimeout(resolve, delay);
     });
 
-describe("ClickableBehavior", () => {
+describe.only("ClickableBehavior", () => {
```
jeresig, line 23: oops!

packages/wonder-blocks-clickable/components/clickable-behavior.js
```diff
@@ -361,6 +360,12 @@ export default class ClickableBehavior extends React.Component<
                 // indicating that we're waiting for navigation to occur.
                 this.setState({waiting: true});
             }
+
+            if (this.props.target === "_blank") {
+                this.navigateOrReset(shouldNavigate);
```
jeresig, line 365: Shouldn't we be calling `safeWithNav` before we call this? And we don't need to set `waiting: true` in this case either. I think we can simplify this a lot by moving the target logic up to the top, like you did in `navigateOrReset`. In fact, it seems like we should just do: `if (history && !skipClientNav || this.props.target === "_blank") {` and that would probably do the right thing.

packages/wonder-blocks-clickable/components/__tests__/clickable-behavior.test.js
kevinbarabash, line 23: I'm going to add a lint rule to check for this.

jeresig (changes requested):
Thank you for working to resolve this! Pushing back a bit as I think we're not calling safeWithNav, and that feels wrong.

I also feel like we need to be updating some docs as a part of this PR, to clarify what's happening now!

packages/wonder-blocks-clickable/components/clickable-behavior.js
kevinbarabash: Good call.  Not sure what I was thinking not calling `safeWithNav`.  I was thinking of checking that it was called in our tests.  I need to remember my testing ABCs (Always Be Checking). 😅

packages/wonder-blocks-clickable/components/clickable-behavior.js
```diff
@@ -96,6 +96,9 @@ type Props = {|
      *
      * If both safeWithNav and beforeNav are provided, beforeNav will be run
      * first and safeWithNav will only be run if beforeNav does not reject.
+     *
+     * WARNING: Using this with `target="_blank"` will trigger built-in popup
+     * blockers in Firefox and Safari.
```
jeresig, line 101: I dunno, this feels like a huge issue - big enough that I think we should prevent people from doing it! We could at least change the Flow types to make it mutually exclusive. And then update the docs to make it clear that `beforeNav` is impossible to use with target=_blank and that folks should instead use safeWithNav.

.eslintrc.js
```diff
@@ -37,6 +37,7 @@ module.exports = {
                 ignorePackages: true,
             },
         ],
+        "jest/no-focused-tests": "error",
```
jeresig, line 40: Ah, interesting!

jeresig (changes requested):
Pushing back a bit more - we really shouldn't allow folks to write code that just won't work!

packages/wonder-blocks-clickable/components/clickable-behavior.js
kevinbarabash: Normally I would immediately jump to enforce this with flow, but the props are already pretty complicated with an existing disjoint union.  I'll give it a try and see how hard it is to make it work.

jeresig (approved):
Very nice, thank you for fixing that up! Would be good to update the beforeNav docs to make this interaction clear

.eslintrc.js
kevinbarabash: We use this rule in `webapp` too.  Will be nice to move some of this stuff into our shared eslint config so that we don't have to remember to add this to every project.

PR #1088, opened 2021-02-05
WB-1028: Set rel="noopener noreferrer" when 'target="_blank"'

kevinbarabash:
## Summary:
This addresses security concerns raised in https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/.  It's something that we were doing in link.jsx and would like to continue doing in all of our components that can render <a> tags in wonder-blocks.

This PR also renames `ClickableHandlers` to `ChildrenProps` since there are more things than just handlers in that type.

Issue: WB-1028

## Test Plan:
- run the unit tests
- open localhost:6060 and check that the button labeled "Async action, open URL in new tab" has the correct 'rel' tag


packages/wonder-blocks-button/__tests__/generated-snapshot.test.js
```diff
@@ -443,7 +443,6 @@ describe("wonder-blocks-button", () => {
                     <Button
                         href="https://google.com"
                         target="_blank"
-                        rel="noopener noreferrer"
```
kevinbarabash, line 446: The test changed but the snapshot didn't. 🙂

packages/wonder-blocks-button/components/button.js
```diff
@@ -83,20 +83,7 @@ export type SharedProps = {|
...
-        | "help"
-        | "license"
-        | "next"
-        | "nofollow"
-        | "noreferrer"
-        | "noopener"
-        | "prev"
-        | "search"
-        | "tag",
+    rel?: string,
```
kevinbarabash, line 86: An enum isn't sufficient because you can set multiple relationships with `rel`.

packages/wonder-blocks-clickable/components/clickable-behavior.js
```diff
@@ -589,7 +596,17 @@ export default class ClickableBehavior extends React.Component<
                   // things that aren't buttons or anchors.
                   tabIndex: 0,
               };
+
+        // When the link is set to open in a new window, we want to append some
+        // `rel` attributes. This is to ensure that the links we're sending folks
+        // to can't hijack the existing page.
+        // More info: https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/
```
kevinbarabash, line 603: I copied this comment from webapp.

jeresig (approved):
Looks good, thank you! This seems like a solid improvement.

packages/wonder-blocks-clickable/components/clickable-behavior.js
```diff
@@ -589,7 +596,17 @@ export default class ClickableBehavior extends React.Component<
                   // things that aren't buttons or anchors.
                   tabIndex: 0,
               };
+
+        // When the link is set to open in a new window, we want to append some
```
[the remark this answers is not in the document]
kevinbarabash: I think overriding is more flexible.  I'll change the language and make a not in the doctoring for the `rel` prop.

PR #1090, opened 2021-02-05
[breaking change] WB-1029: Add default focus ring a way to override it

kevinbarabash:
## Summary:
The default focus ring is very bare bones.  webapp already has a bunch of focus rings like this in it so I think it's fine to ship this now and then iterator on it later.  This PR also adds a style so that the "pointer" cursor is shown when hovering over "Clickable"s just like regular buttons and links.

Issue: WB-1029

## Test plan:
- yarn start
- load http://localhost:6060/#!/Clickable
- hover over the clickables to see the "pointer" cursor
- tab through the page to see Clickable's default focus rin around the "Client-Side routing" examples

Screen Shot 2021-02-05 at 6 42 01 PM


packages/wonder-blocks-clickable/components/clickable.js
```diff
@@ -84,6 +87,12 @@ type CommonProps = {|
      * Respond to raw "keyup" event.
      */
     onKeyUp?: (e: SyntheticKeyboardEvent<>) => mixed,
+
+    /**
+     * Don't show the default focus ring.  This should be used when implementing
+     * a custom focus ring withing your own component that uses Clickable.
```
jeresig, line 93:
```suggestion
     * a custom focus ring within your own component that uses Clickable.
```

packages/wonder-blocks-clickable/components/clickable.js
```diff
@@ -329,4 +346,10 @@ const styles = StyleSheet.create({
         WebkitFontSmoothing: "inherit",
         MozOsxFontSmoothing: "inherit",
     },
+    link: {
+        cursor: "pointer",
+    },
+    focused: {
```
jeresig, line 352: Clickable is used by other WB components, right? How does this change impact them?

jeresig (approved):
Interesting! How closely does the outline match what we have now for links/buttons/etc.? Ideally it should match it very closely. Also, it seems like this will be a breaking change, correct? We'll have to ensure that all existing uses of Clickable are ok with this change, including inside WB.

kevinbarabash:
This is definitely a breaking change although it shouldn't be too difficult to update webapp since there are only 11 uses of it and the upgrade path is pretty simple.  We don't use `Clickable` within webapp itself.  This is something that Kumail and I tried to do but there were issues that we couldn't resolve.

packages/wonder-blocks-clickable/components/clickable.js
kevinbarabash: `ClickableBehavior` is used in other components, but not `Clickable`.

kevinbarabash:
The parent pull-request ([#1088]](https://github.com/Khan/wonder-blocks/pull/1088)) has been merged into `master`, but this branch (`wb-1029`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #1091, opened 2021-02-17
[FEI-3308] Types-first in Wonder Blocks

somewhatabstract:
## Summary:
This upgrades WB to 0.125.1 of Flow and enables types-first.

I will do a subsequent PR or two to get us upto the latest Flow version.

Issue: FEI-3308

## Test plan:
`yarn flow`
`yarn test`

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,5 @@
 [version]
-0.123.0
+0.124.0
 
 [declarations]
 <PROJECT_ROOT>/node_modules
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,5 @@
 [version]
-0.124.0
+0.125.1
 
 [declarations]
 <PROJECT_ROOT>/node_modules
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,5 @@
 [version]
-0.122.0
+0.123.0
 
 [declarations]
 <PROJECT_ROOT>/node_modules
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,5 @@
 [version]
-0.121.0
+0.122.0
 
 [declarations]
 <PROJECT_ROOT>/node_modules
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,5 @@
 [version]
-0.119.1
+0.120.1
 
 [declarations]
 <PROJECT_ROOT>/node_modules
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,5 @@
 [version]
-0.116.1
+0.117.0
 
 [declarations]
 <PROJECT_ROOT>/node_modules
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,5 @@
 [version]
-0.120.1
+0.121.0
 
 [declarations]
 <PROJECT_ROOT>/node_modules
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,5 @@
 [version]
-0.118.0
+0.119.1
 
 [declarations]
 <PROJECT_ROOT>/node_modules
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,5 @@
 [version]
-0.117.0
+0.118.0
 
 [declarations]
 <PROJECT_ROOT>/node_modules
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,5 @@
 [version]
-0.115.0
+0.116.1
 
 [declarations]
 <PROJECT_ROOT>/node_modules
--- a/.flowconfig
+++ b/.flowconfig
@@ -11,6 +11,7 @@ implicit-inexact-object=error
 
 [options]
 well_formed_exports=true
+types_first=true
 
 # $FlowFixMe
 # This suppression hould be used to suppress an issue that requires additional
--- a/.flowconfig
+++ b/.flowconfig
@@ -10,6 +10,8 @@ deprecated-utility=error
 implicit-inexact-object=error
 
 [options]
+well_formed_exports=true
+
 # $FlowFixMe
 # This suppression hould be used to suppress an issue that requires additional
 # effort to be resolved. That effort should be recorded in our issue tracking
--- a/flow-typed/ka-types.js
+++ b/flow-typed/ka-types.js
@@ -1,0 +1,24 @@
+// @flow
+/**
+ * This is the best way to specify an empty exact type in Flow.
+ *
+ * When a component has empty props, use this type like this:
+ *     type Props = Empty;
+ *
+ * Then, when adding props to the component:
+ *     type Props = {| myComponentProp: string |};
+ *
+ * See https://github.com/facebook/flow/issues/2977 for details.
+ */
+declare type Empty = $Shape<{||}>;
+
+// Marks all properties of an object optional.
+//
+// We use this instead of $Shape<T> because spreading $Shape<$Exact<T>> ends up
+// spreading $Exact<T> instead, losing the optionality of keys.
+// https://github.com/facebook/flow/issues/6906#issuecomment-453439922
+declare type Partial<T> = $ReadOnly<$Rest<T, {...}>>;
+
+// TODO(jeff, WEB-616): Remove this in favour of real React type for refs when
+// it is available.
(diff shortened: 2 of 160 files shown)


flow-typed/ka-types.js
```diff
@@ -0,0 +1,24 @@
...
+// Marks all properties of an object optional.
+//
+// We use this instead of $Shape<T> because spreading $Shape<$Exact<T>> ends up
+// spreading $Exact<T> instead, losing the optionality of keys.
+// https://github.com/facebook/flow/issues/6906#issuecomment-453439922
+declare type Partial<T> = $ReadOnly<$Rest<T, {...}>>;
+
+// TODO(jeff, WEB-616): Remove this in favour of real React type for refs when
+// it is available.
+declare type RefObject<T> = {current: null | React.ElementRef<T>, ...};
```
somewhatabstract, line 24: Haven't used this one yet, so we could drop it from WB types if we wished.

.flowconfig
```diff
@@ -10,6 +10,9 @@ deprecated-utility=error
 implicit-inexact-object=error
 
 [options]
+well_formed_exports=true
+types_first=true
```
somewhatabstract, line 14: :tada:

flow-typed/npm/react-dom_v16.x.x.js
```diff
@@ -0,0 +1,110 @@
+// flow-typed signature: a333d1cdbb5a4103d0be63a412070e22
```
somewhatabstract, line 1: Added with `flow-typed install react-dom`

packages/wonder-blocks-button/components/button-core.js
```diff
@@ -29,14 +29,18 @@ type Props = {|
     type?: "submit",
 |};
 
+type ContextTypes = {|
+    router: PropTypes.Requireable<any>,
```
somewhatabstract, line 33: Technically, `PropTypes` is an untyped module, I believe so this is equivalent to just `any`, but it would be something like `PropTypes.Requireable<any>` if it were typed (I believe that's what TypeScript would use).

packages/wonder-blocks-timing/util/types.js
```diff
@@ -241,6 +241,10 @@ export type WithActionSchedulerProps = {|
     schedule: IScheduleActions,
 |};
 
+export type WithoutActionScheduler<TProps: {...}> = $Exact<
+    $Rest<TProps, WithActionSchedulerProps>,
+>;
```
somewhatabstract, line 246: We can now consider WB-1027 as done.

flow-typed/ka-types.js
kevinbarabash: Yet another example where TypeScript types for React are ahead of the Flow types:
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L92-L96
- https://github.com/facebook/flow/blob/master/lib/react.js#L211-L214

packages/wonder-blocks-button/components/button-core.js
kevinbarabash: I couldn't find anything about `Requireable` in https://github.com/facebook/prop-types.  Is this a made up type?

somewhatabstract: It's what TypeScript says it should be. As in my comment, it really equates to `any` here.

packages/wonder-blocks-color/index.js
```diff
@@ -4,7 +4,35 @@ import {mix, fade} from "./util/utils.js";
 const offBlack = "#21242c";
 const white = "#ffffff";
 
-const Color = {
+type ColorType = {|
```
kevinbarabash, line 7: In retrospect we probably should've used named exports for the color package.  We could change in this in the future.

packages/wonder-blocks-core/components/__tests__/view.test.js
```diff
@@ -1,4 +1,4 @@
-import React from "react";
```
kevinbarabash, line 1: Thanks for fixing all of these.

packages/wonder-blocks-data/util/request-fulfillment.js
```diff
@@ -45,7 +47,10 @@ export class RequestFulfillment {
      * This will return an inflight request if one exists, otherwise it will
      * make a new request. Inflight requests are deleted once they resolve.
      */
-    fulfill = <TOptions, TData: ValidData>(
+    fulfill: <TOptions, TData: ValidData>(
+        handler: IRequestHandler<TOptions, TData>,
+        options: TOptions,
+    ) => Promise<CacheEntry<TData>> = <TOptions, TData: ValidData>(
```
kevinbarabash, line 53: At least GitHub is using the same color for all of the type params now. ;)

packages/wonder-blocks-dropdown/components/action-menu.js
```diff
@@ -120,14 +125,17 @@ export default class ActionMenu extends React.Component<Props, State> {
      * Used to sync the `opened` state when this component acts as a controlled
      * component
      */
-    static getDerivedStateFromProps(props: Props, state: State) {
+    static getDerivedStateFromProps(
```
kevinbarabash, line 128: I didn't realize we used `getDerivedStateFromProps` in so many places.

packages/wonder-blocks-dropdown/util/constants.js
```diff
@@ -41,8 +41,10 @@ export const defaultLabels = {
...
     selectNoneLabel: "Select none",
-    selectAllLabel: (numOptions: number) => `Select all (${numOptions})`,
+    selectAllLabel: (numOptions: number): string =>
+        `Select all (${numOptions})`,
     noneSelected: "0 items",
-    someSelected: (numSelectedValues: number) => `${numSelectedValues} items`,
+    someSelected: (numSelectedValues: number): string =>
+        `${numSelectedValues} items`,
     allSelected: "All items",
 };
```
kevinbarabash, line 50: Why don't we need to create a type like `ColorTypes` for `defaultLabels`?

packages/wonder-blocks-grid/components/cell.js
```diff
@@ -39,6 +39,13 @@ type Props = {|
     style?: StyleType,
 |};
 
+type DefaultProps = {|
+    smallCols: $PropertyType<Props, "smallCols">,
+    mediumCols: $PropertyType<Props, "mediumCols">,
+    largeCols: $PropertyType<Props, "largeCols">,
+    cols: $PropertyType<Props, "cols">,
```
kevinbarabash, line 46: These props in `Props` are optional, but we're providing default values for them so they should be required.

packages/wonder-blocks-modal/components/modal-launcher.js
```diff
@@ -107,11 +111,14 @@ export default class ModalLauncher extends React.Component<Props, State> {
...
-    static defaultProps = {
+    static defaultProps: DefaultProps = {
         backdropDismissEnabled: true,
     };
 
-    static getDerivedStateFromProps(props: Props, state: State) {
+    static getDerivedStateFromProps(
+        props: Props,
+        state: State,
+    ): ?Partial<State> {
```
kevinbarabash, line 121: I like this type here because I always forget if I have to return the full object for `state` or not.

kevinbarabash (approved):
LGTM.  A few nits inline, but nothing blocking.  Thanks for updating all `"react"` imports to use namespace imports.

packages/wonder-blocks-dropdown/util/constants.js
somewhatabstract: Because flow didn't ask.

I think it's because they are primitive literals.

packages/wonder-blocks-grid/components/cell.js
somewhatabstract: Possibly, but I don't think it matters in a types first world.

packages/wonder-blocks-modal/components/modal-launcher.js
somewhatabstract: Yeah, I originally was typing "just enough" to get it to pass which was `State` in some cases, `Partial<State>` in others, etc. I updated them all to be `?Partial<State>` to reflect the full range of possibilities.

PR #1092, opened 2021-02-17
[FEI-3308].2 Tidy up some things that didn't work

somewhatabstract:
## Summary:
So even though flow passed for Wonder Blocks, it didn't for webapp
consuming Wonder Blocks. These changes address the problematic types
I was able to identify.

Issue: FEI-3308

## Test plan:
`yarn flow`
Update webapp and see where we're at.


kevinbarabash (approved):
LGTM.  :shipit:

PR #1094, opened 2021-02-23
FEI-3316: Make dispatched 'submit' events cancelable

kevinbarabash:
## Summary:
This was causing login forms to redirect to the LOHP without logging users in when using Firefox.  This issue arose because of a recent change in webapp to use forms properly.

In wonder-blocks we do a bunch of custom handling of events which involves calling 'preventDefault' on the user generated events.  Normally the browser would generate 'submit' events for us, but we have to do it ourselves in wonder-blocks because of this custom handling.

In webapp we need to call 'preventDefault' on the generate 'submit' event as part of the login flow.  Firefox does not support calling 'preventDefault' on non-cancelable events (which is what we were generating).  Safari and Chrome allow this but this may change in the future.

This PR fixes the issue by making the 'submit' event we generate cancelable.

Issue: FEI-3316

## Test plan:
- yarn clean
- yarn start-storybook
- open the "Submit Button in Form" story in Firefox
- submit the form by pressing "enter" when the cursor is in the textfield, see that the story doesn't reload
- submit the form by pressing "spacebar" when the cursor is on the button, see that the story doesn't reload
- submit the form by clicking the "submit" button, see that the story doesn't reload

### DIFF
--- a/packages/wonder-blocks-button/components/button.stories.js
+++ b/packages/wonder-blocks-button/components/button.stories.js
@@ -247,3 +247,29 @@ export const buttonWithSpinner: React.ComponentType<Empty> = () => (
         </Button>
     </View>
 );
+
+export const submitButtonInForm: React.ComponentType<Empty> = () => (
+    <form
+        onSubmit={(e) => {
+            e.preventDefault();
+            window.alert("form submitted"); // eslint-disable-line no-alert
+        }}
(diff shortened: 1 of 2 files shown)


jandrade (approved):
Looks good to me, thanks for the quick fix!

PR #1096, opened 2021-02-23
Fix flow errors that only manifest in webapp

kevinbarabash:
## Summary:
There were a couple of flow errors that appear in webapp in the node_modules after upgrading the wonder-blocks deps there.  I made the changes in this PR directly within node_modules are restarted to flow to make sure things were working as expected before backporting them here.

Issue: none

## Test plan:
- yarn flow

### DIFF
--- a/packages/wonder-blocks-clickable/util/get-clickable-behavior.js
+++ b/packages/wonder-blocks-clickable/util/get-clickable-behavior.js
@@ -34,6 +34,8 @@ export default function getClickableBehavior(
     if (router && skipClientNav !== true && href && !isExternalUrl(href)) {
         // We cast to `any` here since the type of ClickableBehaviorWithRouter
         // is slightly different from the return type of this function.
+        // TODO(WB-1037): Always return the wrapped version once all routes have
+        // been ported to the app-shell in webapp.
         return (ClickableBehaviorWithRouter: any);
     }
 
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-clickable/util/get-clickable-behavior.js
```diff
@@ -32,7 +32,9 @@ export default function getClickableBehavior(
     router?: any,
 ): React.ComponentType<React.ElementConfig<typeof ClickableBehavior>> {
     if (router && skipClientNav !== true && href && !isExternalUrl(href)) {
-        return ClickableBehaviorWithRouter;
+        // We cast to `any` here since the type of ClickableBehaviorWithRouter
+        // is slightly different from the return type of this function.
+        return (ClickableBehaviorWithRouter: any);
```
somewhatabstract, line 37: This seems reasonable. Could you add a TODO or raise a ticket to change this to component that flips between the two properly (or at least to address this better)?

packages/wonder-blocks-dropdown/components/dropdown-core.js
```diff
@@ -813,7 +813,9 @@ const styles = StyleSheet.create({
     },
 });
 
-type ExportProps = WithoutActionScheduler<React.ElementConfig<DropdownCore>>;
```
somewhatabstract, line 816: When I get WB to a newer flow, this will get caught properly.

somewhatabstract (approved):
LGTM

PR #1097, opened 2021-02-24
[FEI-3327].1 Upgrade to Flow 0.127.0

somewhatabstract:
## Summary:

Issue: FEI-3327

## Test plan:
`yarn flow`

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,6 +1,6 @@
-[version]
-0.125.1
-
+# [version]
+# 0.126.1
+#
 [declarations]
 <PROJECT_ROOT>/node_modules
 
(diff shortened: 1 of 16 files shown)


.flowconfig
```diff
@@ -1,6 +1,6 @@
-[version]
-0.125.1
-
+# [version]
+# 0.126.1
+#
```
somewhatabstract, line 3: I'll reinstate this once I've finished the upgrade process (though, I don't really see the point in keeping this and wonder if we should drop it so dependabot can auto update us).

kevinbarabash (approved):
LGTM.  Easy breezy.  🎉

PR #1098, opened 2021-02-24
[FEI-3327].2 Update to Flow 0.145.0

somewhatabstract:
## Summary:
The were a few different points where changes were made, but they
seemed small enough to keep it all bundled into one review.

Issue: FEI-3327

## Test plan:
`yarn flow`
`yarn test`

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,6 +1,3 @@
-# [version]
-# 0.126.1
-#
 [declarations]
 <PROJECT_ROOT>/node_modules
 
@@ -8,7 +5,3 @@
 deprecated-type=error
 deprecated-utility=error
 implicit-inexact-object=error
-
-[options]
-well_formed_exports=true
-types_first=true
--- a/flow-typed/npm/jest_v24.x.x.js
+++ b/flow-typed/npm/jest_v24.x.x.js
@@ -1,7 +1,5 @@
 // flow-typed signature: f5a484315a3dea13d273645306e4076a
 // flow-typed version: 7c5d14b3d4/jest_v23.x.x/flow_>=v0.39.x
-import type {ComponentType} from "react";
-
 type JestMockFn<TArguments: $ReadOnlyArray<*>, TReturn> = {|
     (...args: TArguments): TReturn,
     /**
@@ -267,7 +265,7 @@ type JestJQueryMatchersType = {
 
 // TODO(WEB-902): create flow types for jest-enzyme
 type JestEnzymeMatchersType = {
-    toContainMatchingElement(selector: string | ComponentType<*>): void,
+    toContainMatchingElement(selector: string | React$ComponentType<*>): void,
     ...
 };
 
--- a/flow-typed/storybook.flow.js
+++ b/flow-typed/storybook.flow.js
@@ -3,7 +3,7 @@ declare module "@storybook/react" {
     /**
      * Made up type to decorate a React component with the story field.
      */
-    declare type StoryComponentType<T> = React$ComponentType<*> & {
+    declare type StoryComponentType = React$ComponentType<Empty> & {
         story?: $FlowFixMe,
         ...
     };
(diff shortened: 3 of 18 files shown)


.flowconfig
```diff
@@ -1,14 +1,7 @@
-# [version]
-# 0.126.1
-#
```
somewhatabstract, line 3: I propose that we don't specify a version and let dependabot keep us up-to-date when it can.

flow-typed/npm/jest_v24.x.x.js
```diff
@@ -1,7 +1,5 @@
 // flow-typed signature: f5a484315a3dea13d273645306e4076a
 // flow-typed version: 7c5d14b3d4/jest_v23.x.x/flow_>=v0.39.x
-import type {ComponentType} from "react";
```
somewhatabstract, line 3: No global imports are allowed.

flow-typed/storybook.flow.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+declare module "@storybook/react" {
+    /**
+     * Made up type to decorate a React component with the story field.
+     */
+    declare type StoryComponentType<T> = React$ComponentType<*> & {
+        story?: $FlowFixMe,
+        ...
+    };
+}
```
somewhatabstract, line 10: Couldn't add flow-typed typelib for this module and I checked its types; it doesn't cater to this specific situation. So I hand-crafted something simple to suit our needs/usage.

I could be persuaded to make `<*>` be `<Empty>` since all current uses seem to be that way.

.flowconfig
kevinbarabash, line 3: Is it okay to drop this?

flow-typed/npm/jest_v24.x.x.js
kevinbarabash, line 3: This is to avoid having to move everything into a `module { }` block?

packages/wonder-blocks-data/components/data.js
```diff
@@ -136,14 +136,23 @@ export default class Data<TOptions, TData: ValidData> extends React.Component<
...
                     );
+
+                    /**
+                     * Need to share our types with InternalData so Flow
+                     * doesn't need to infer them and find mismatches.
+                     */
+                    class TypedInternalData extends InternalData<
+                        TOptions,
+                        TData,
+                    > {}
```
somewhatabstract, line 147: Took me a while to work out how to resolve the errors that were coming up here. I realised it was quite possibly because we weren't explicitly parameterizing `InternalData` - so I tried this and it worked.

flow-typed/storybook.flow.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+declare module "@storybook/react" {
+    /**
+     * Made up type to decorate a React component with the story field.
+     */
+    declare type StoryComponentType<T> = React$ComponentType<*> & {
+        story?: $FlowFixMe,
+        ...
+    };
```
kevinbarabash, line 9: nit: `*` should probably be `any` since `*` is deprecated.

kevinbarabash (approved):
LGTM.  :shipit:

flow-typed/npm/jest_v24.x.x.js
somewhatabstract: Yup.

.flowconfig
somewhatabstract: I don't see what value it gives us. I don't use it in `render-gateway`, `checksync` or `ancesdir`.

somewhatabstract:
The parent pull-request ([#1097]](https://github.com/Khan/wonder-blocks/pull/1097)) has been merged into `master`, but this branch (`fei3327.2`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #1099, opened 2021-02-25
Move source files into 'src' dirs

kevinbarabash:
## Summary:
This PR adds 'src' directories that are siblings to 'dist'.  This will simplify possible future migration to TypeScript by ensuring that source code and built code (in dist) and distrint from one another.  Although we won't be using tsc to build out dist files we don't want it including dist in the source files that it looks at when type checking.  Having these folders be siblings is the easiest way to achieve this.

Interestingly a bunch of our CI jobs weren't building the packages and broken when the location of the files moved.  This PR fixes the issues but I'm confused as to how things were still working in other recent PRs.  I assume it has something to do with `node_module` gets cached by our CI services between builds.  I thought the site was being updated, but it wasn't and I just thought it was. 😓

Issue: none

## Test plan:
- yarn test
- yarn flow


jeresig (approved):
Nice! I wonder if you should also updated `lerna.json` to exclude the entire `src` directory, I think that would help fix a lot of issues!

kevinbarabash:
Unfortunately we can't ignore the `src` directory at the current moment since `dist/index.js.flow` imports `../src/index.js`.  This is needed so that Flow can use the types for each package.

PR #1100, opened 2021-02-25
Roll back flow-related change and comment why

somewhatabstract:
## Summary:
Having this wrapper component was causing the children of InternalData to render
more than we wanted. I was unable to easily prevent it from doing so. So, instead
I removed the extra component and addressed the flow error with a suppression.

Issue: FEI-3327

## Test plan:
1. `yarn build:all`
2. copy contents of packages/wonder-blocks-data to webapp's node_modules
3. Exit `javascript/data-access/with-content-data_test.js` in webapp to remove `forceStatic:true`
4. Run `yarn test javascript/data-access/with-content-data_test.js`


jeresig (approved):
Thanks for the updated comment!

PR #1101, opened 2021-02-27
Experiment: refactor conditional props

kevinbarabash:
## Summary:
This is part of a series of experimental PRs to determine the feasibility of migrating to TypeScript.

This PR in particular is some pre-work to refactor our conditional props into a form that TypeScript knows how to deal with.  This is based on the pattern described in https://www.benmvp.com/blog/conditional-react-props-typescript/.

Note: Flow does not like passing the results of function calls as props that are conditional props.  This is a bug in Flow since Flow is just fine passing a literal or variable that is the same type as the return type of the function.  This is fine because we deploy wonder-blocks to webapp in this state.  We'll convert it to TypeScript and then deploy it webapp.  TypeScript doesn't have this bug.

Issue: none

## Test plan:
- yarn flow

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -79,5 +79,8 @@ module.exports = {
                 ],
             },
         ],
+
+        // customization of rules
+        "react/prop-types": "off", // we let Flow/TS keep track of valid props
     },
 };
--- a/packages/wonder-blocks-button/src/components/button.js
+++ b/packages/wonder-blocks-button/src/components/button.js
@@ -80,20 +80,6 @@ export type SharedProps = {|
     testId?: string,
 
     /**
-     * Specifies the type of relationship between the current document and the
-     * linked document. Should only be used when `href` is specified. This
-     * defaults to "noopener noreferrer" when `target="_blank"`, but can be
-     * overridden by setting this prop to something else.
-     */
-    rel?: string,
-
-    /**
-     * A target destination window for a link to open in. Should only be used
-     * when `href` is specified.
-     */
-    target?: "_blank",
-
-    /**
      * Set the tabindex attribute on the rendered element.
      */
     tabIndex?: number,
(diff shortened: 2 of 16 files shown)


.eslintrc.js
```diff
@@ -79,5 +79,8 @@ module.exports = {
                 ],
             },
         ],
+
+        // customization of rules
+        "react/prop-types": "off", // we let Flow/TS keep track of valid props
```
kevinbarabash, line 84: This plugin doesn't understand conditional props.  We were getting around this by setting default values while destructuring.  I think it's better to let the static type checker worry about this.  There are some files in webapp that aren't using Flow yet so we may want to figure out how to enable this rule but only for those files.

packages/wonder-blocks-button/src/components/button.js
```diff
@@ -184,44 +167,40 @@ type Props =
...
-           * browser does a full page load navigation, the callback promise must be
-           * settled before the navigation will occur. Errors are ignored so that
-           * navigation is guaranteed to succeed.
-           */
-          safeWithNav: () => Promise<mixed>,
+          href?: empty, // can't be used with submit
+          beforeNav?: empty, // must be used with href
+          safeWithNav?: empty, // must be used with href
+          target?: empty, // must be used with href
+          rel?: empty, // must be used with href
```
kevinbarabash, line 192: I've added inline comments here so that the comment shows up in the error message if Flow/TS finds any violations of these conditional props.

packages/wonder-blocks-data/src/components/intercept-data.js
```diff
@@ -26,39 +26,43 @@ type BaseProps<TOptions, TData> = {|
...
-    shouldRefreshCache: InterceptShouldRefreshCacheFn<TOptions, TData>,
-|};
+// type ShouldRefreshCacheProps<TOptions, TData> = {|
+//     /**
+//      * Called to determine if the cache should be refreshed.
+//      * If this returns null, the handler being intercepted will be asked if
+//      * the cache should be refreshed.
+//      */
+//     shouldRefreshCache: InterceptShouldRefreshCacheFn<TOptions, TData>,
+// |};
```
kevinbarabash, line 45: I'll go back to spreading these but include `prop? empty` as necessary.

packages/wonder-blocks-data/src/components/intercept-data.js
```diff
@@ -26,39 +26,43 @@ type BaseProps<TOptions, TData> = {|
...
 
-type Props<TOptions, TData> =
```
kevinbarabash, line 47: I think this original formulation would've worked in TS.  I changed this to define a separate `ConditionalProps` type to match the rest of the changes in this PR.

packages/wonder-blocks-dropdown/src/components/action-item.js
```diff
@@ -146,74 +146,81 @@ export default class ActionItem extends React.Component<ActionProps> {
...
+            return (
+                <ClickableBehavior
+                    disabled={disabled}
+                    onClick={onClick}
+                    role={role}
+                >
+                    {children}
+                </ClickableBehavior>
+            );
+        }
```
kevinbarabash, line 223: `ClickableBehavior`'s props contain conditional props which means we have to be careful about how we pass `ActionItem`'s props down to it even if the conditional props for the two components are the same.  We should probably have an internal version of `ClickableBehavior` that doesn't have any conditional props.  That would be internal use easier and we can export a facade for it that has conditional props.

packages/wonder-blocks-link/src/components/link.js
```diff
@@ -127,19 +127,16 @@ type CommonProps = {|
     onKeyUp?: (e: SyntheticKeyboardEvent<>) => mixed,
 |};
 
-export type SharedProps =
```
kevinbarabash, line 130: Exporting a shared type that include conditional props makes it difficult to use in `LinkCore`.  Instead we only export the common props (like button.js) and then add any missing props to `LinkCore`'s props manually.

kevinbarabash:
Good idea.  I did a quick search and couldn't find any existing tickets for the issue I'm seeing.  I'll file a ticket later today.

kevinbarabash:
@jeresig I can't repro this issue in the [flow playground](https://flow.org/try/#0PTAEAEDMBsHsHcBQBLAtgB1gJwC6gFSgCGAzqAEoCmRAxnpFrKqAERbV0sDciiOAnukqgAwk1SwAdgAVG6MgF5QAbwA+iUJtAAjIgC8AXKEkBXVNspYANIlUBfHn0HCxkgCbIcyKUWizY8qBKahpakLCwRiQ4WMiSAOY2WjpEWAD8RpQYAklaAI4mAB4ZoFnoObZ2oKoq6snhsCVlFcm6WEbaEdDUkrmaBcWZ2fw29tW1oZoNTcN9KelD5SOToANGpuaWo3a8AkKg-oHBdVoAdOdiqBIyciRz56euHl4+frfbjjTQpGQAYhEAIVSpUKOEo7jIVFoOEeTEwknBOAAPIcSAA+FQrdjuSwACgAlJjksl2DgTFhJMYTNBoDxkjsHLwaFJoqBINoAIxBUBI-6wIFYFJ6BTKACsVQaChYkGQADdKCxQMA0TxmZJWeyAEzc3mA4G6YViqptEUxEyUKrKxwgEFCOiUNxGVDIEgkOLxUBSYSwSBsiKewVtRBqjXaADMOr5AqFIvFSpVvBtlEKdrBjtAAAtSDpYDgM37YMR3PNQPAM+DQJ5QCQM7BqW5PZJoPxM0R5Y3KMGWfRtAAWSN6wP6WMSiJSmXyxUm5Rmi3xz7d0Ck8nq35y4RKAlRGLuoIY6Xr7hd9U90UD-n64dGgsi5cUkhr+UEy0J4+su-qgAqWHN3K3OdgboiEpBQMVnVVF3ZAA2c9owNEd5lvSgyXvb9zWfecmUXD8SHIYC3CYP98XWMwLEFUDQAAWSIPNTiwfCmAJCCTzZbQAHZYMvQ04wGJCUPVPD3EY-EXx4IA).  I believe this issue only manifests when cross the module boundary.  I'll have to create a minimal reproducible repo to include in the bug report.

kevinbarabash:
I was able to repro the issue in the [flow playground](https://flow.org/try/#0PTAEAEDMBsHsHcBQBLAtgB1gJwC6gFSgCGAzqAEoCmRAxnpFrKqAERbV0sDciiOAnukqgACo3RkAvKADeAH0SgloABbtIALlAkcWZADsA5gBpFygUID8WyhgGm5AX1BzZC5avXXQt9PbNKFpTeLCQArgBGqMg4LMagIKDIkKBBSWTsAI5hyOwAJqkqlPqFwsgk4WUk+gDkeLrIhoaUWIhOPIg00KRkAEJhODiwJZQAHjjFeWRUtDgAdADCTJj6xTgAPGKwEgB8sgGg7Pp5LQAUAJRaM3RzAHKwJ-seHuw4YVgl+mHQ0FwJYGh0NBbGtQCccERkNADo5ELDEOt+oNhp5KJBJCxgJBYLAWAkdjxEQMhiUghjwlEYnjgASEUiSfiOokxkI6JQCi1GFh4gAxHFg2CUap1Yg0GiUPygCKwHAqUDocRkIj1IraIioYQ4NCUOnElFqNEYrE4vFk0KRaKxRmgXg0YY6UDNHAAVSw0FA0guWh0eiMHr2mOxuJ4dv0Dve7ukTtd0AuIfteAjAGUGkZvanDB7HZQXW647wiciSgb0TJo3nzs4aX9EiQVLBvgV9DKfFgubqi6jSxGqwSlLX643QM28JzsB2GSXJDJkxnezwgA) and have filed a ticket.

kevinbarabash:
I'm going to close this ticket because we can't land it as is.  I'm going to leave the branch up so that I can continue to run tests on `flow-to-ts` on it and generate example PRs from those test runs.

PR #1102, opened 2021-02-27
Update enzyme shim config to better align with webapp

kevinbarabash:
## Summary:
This PR moves the call to 'unmountAll()' into an 'afterEach()' block jest-setup.js.  This guarantees that it's being called after each test without developers having to manually add a call to it in each test suite they right.  It was missing in a few them as it is.

This will also make it easier to use package references in the TypeScript experimeents I'm working on.

Issue: none

## Test plan:
- yarn test
- yarn flow

### DIFF
--- a/packages/wonder-blocks-modal/src/components/__tests__/close-button.test.js
+++ b/packages/wonder-blocks-modal/src/components/__tests__/close-button.test.js
@@ -1,16 +1,12 @@
 // @flow
 import * as React from "react";
+import {mount} from "enzyme";
 
-import {mount, unmountAll} from "../../../../../utils/testing/mount.js";
 import expectRenderError from "../../../../../utils/testing/expect-render-error.js";
 import CloseButton from "../close-button.js";
 import ModalContext from "../modal-context.js";
 
 describe("CloseButton", () => {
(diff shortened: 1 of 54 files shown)


packages/wonder-blocks-modal/src/components/__tests__/close-button.test.js
```diff
@@ -23,9 +19,11 @@ describe("CloseButton", () => {
...
         // Arrange
         const wrapper = mount(
-            <ModalContext.Provider value={{closeModal: () => {}}}>
-                <CloseButton testId="modal-example-close" />,
-            </ModalContext.Provider>,
+            <div>
+                <ModalContext.Provider value={{closeModal: () => {}}}>
+                    <CloseButton testId="modal-example-close" />,
+                </ModalContext.Provider>
+            </div>,
```
kevinbarabash, line 26: We were calling `unmountAll` in `beforeEach()` and this test was running last in this suite so we weren't calling `unmount` on it before.

jeresig (approved):
Thank you for cleaning this up! This always bothered me... and this is a good approach!

PR #1103, opened 2021-02-28
Use namesapce imports for 'prop-types'

kevinbarabash:
## Summary:
'prop-types' doesn't actually have a default export even though we're doing a default import.  webpack/babel make this work at run time.  TypeScript complains about this.  We could use TS' "allowSyntheticDefaultImports" setting to have it paper of this as well, but changing it a namespace import is easy and better reflect the true nature of this module.

Issue: none

## Test plan:
- yarn flow
- yarn test

### DIFF
--- a/packages/wonder-blocks-button/src/components/button-core.js
+++ b/packages/wonder-blocks-button/src/components/button-core.js
@@ -2,7 +2,7 @@
 import * as React from "react";
 import {StyleSheet} from "aphrodite";
 import {Link} from "react-router-dom";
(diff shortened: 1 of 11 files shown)


jeresig (approved):
Very interesting... I actually wonder if it matters if we use PropTypes or not, but this is useful to know!

kevinbarabash:
Not sure.  I think more recent versions of react-router use the new context.  I looks like it's available but marked as internal.  See https://github.com/ReactTraining/react-router/issues/7393. :(

PR #1106, opened 2021-04-20
Initial version of prototype for shading package

nefaurk:


jeresig (approved):
How exactly we integrate this into Wonder Blocks remains to be seen - but I really like this approach. I think the end result looks sharp and it seems easy enough to configure. Thanks for making the story with the knobs, that made it really easy to experiment with!

jandrade (approved):

somewhatabstract:
@jandrade I am closing this PR for now. It's just noise. We can reopen, etc, if we need to.

PR #1107, opened 2021-05-03
FEI-3288: Don't render <Link> when URL is external

kevinbarabash:
## Summary:
We want to be able to use external URLs with links and buttons and such without having to specify 'skipClientNav={true}' in webapp.  While we were checking if 'href' is external when deciding whether or not to wrap ClickableBehavior in 'withRouter', but we missing the other decision point in button-core, link-core, etc. where we decide whether to render a <Link>, `<a>`, or `<button>`.  This diff fixes this.  It also, fixes an issue where we were setting 'disabled={true}' on <a> which doesn't support that attribute.

Issue: FEI-3288

## Test plan:
- yarn test


packages/wonder-blocks-clickable/src/__tests__/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -20,10 +20,9 @@ exports[`wonder-blocks-clickable example 1 1`] = `
     }
   }
 >
-  <button
+  <a
```
kevinbarabash, line 23: I'll double check why this one switched to an `a`.

kevinbarabash: I checked the example in clickable.md which generates this snapshot and the `Clickable` doesn't have an `href` so this should be a `button`.  I'll work on fixing this.

consistency-tests/__tests__/clickables.test.js
```diff
@@ -66,16 +63,80 @@ describe.each`
...
+});
+
+// NOTE: Link doesn't work without an href so it isn't included in this suite
+describe.each`
+    Component            | name
+    ${ActionItem}        | ${"ActionItem"}
+    ${Button}            | ${"Button"}
+    ${ClickableWrapper}  | ${"Clickable"}
+    ${IconButtonWrapper} | ${"IconButton"}
+`("$name without an href", ({Component, name}) => {
```
kevinbarabash, line 107: I added these tests since relying on snapshots to verify that the right underlying component is not ideal since it can take some time to validate the snapshots are correct.

kevinbarabash:
In addition to external URLs we'll want to the same when using `#` or `javascript:void(0)` as the `href`.

consistency-tests/__tests__/clickables.test.js
jandrade: Nice set of tests!

packages/wonder-blocks-clickable/src/util/__tests__/is-external-url.js.test.js
```diff
@@ -1,28 +1,33 @@
 // @flow
-import isExternalUrl from "../is-external-url.js";
+import {isClientSideUrl} from "../is-client-side-url.js";
```
jandrade, line 2: What do you think about changing this file name from `is-external-url.js.test.js` to `is-client-side-url.js.test.js` to match the file that is being tested?

packages/wonder-blocks-clickable/src/util/__tests__/is-external-url.js.test.js
```diff
@@ -1,28 +1,33 @@
...
 
 describe("isExternalUrl", () => {
```
jandrade, line 4: same comment here

packages/wonder-blocks-clickable/src/components/clickable.js
```diff
@@ -211,7 +212,10 @@ export default class Clickable extends React.Component<Props> {
         commonProps: {[string]: any, ...},
     ) => React.Node = (clickableState, commonProps) => {
         const activeHref = this.props.href && !this.props.disabled;
-        const useClient = this.context.router && !this.props.skipClientNav;
+        const useClient =
+            this.context.router &&
+            !this.props.skipClientNav &&
```
jandrade, line 217: If I understand correctly... is the idea to remove this `skipClientNav` check after removing all webapp usages?

jandrade (approved):
Looks good to me! Just a couple of minor comments related to a file name, but nothing blocking. Thanks for the tests, really nice setup!

packages/wonder-blocks-clickable/src/components/clickable.js
kevinbarabash: I don't have a task to remove this prop from the API.  There's a couple of uses in webapp where `skipClientNav` is being used in such a way that it can't be removed without making other changes.  See the uses of `skipClientNav` in student-assignment-table.jsx in webapp.

PR #1108, opened 2021-05-03
Fix pre-publish-check script

kevinbarabash:
## Summary:
In #1099 I moved the source code in each package into 'src', but forgot to update the pre-publish script.  Now the script looks for index.js in the right place.

Issue: none

## Test plan:
- After landing this I'll do a publish.

### DIFF
--- a/utils/pre-publish-check.js
+++ b/utils/pre-publish-check.js
@@ -3,8 +3,8 @@
  * Pre-publish checks to verify that our publish will go smoothly.
  */
 const path = require("path");
-const glob = require("glob");
 const {exec} = require("child_process");
+const glob = require("glob");
 
 const inquirer = require("inquirer");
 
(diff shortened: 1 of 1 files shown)


utils/pre-publish-check.js
```diff
@@ -3,8 +3,8 @@
  * Pre-publish checks to verify that our publish will go smoothly.
  */
 const path = require("path");
-const glob = require("glob");
 const {exec} = require("child_process");
+const glob = require("glob");
```
somewhatabstract, line 7: Oooo, we should update this to fast-glob.

No action required.

somewhatabstract (approved):
LGTM

utils/pre-publish-check.js
```diff
@@ -103,6 +103,12 @@ glob(
                         if (!skipWarnings) {
                             process.exit(1);
                         }
+                        return;
+                    })
+                    .catch((e) => {
+                        console.warn("inquirer prompt failed");
+                        console.warn(e);
+                        process.exit(1);
```
kevinbarabash, line 111: The linter was complaining about not having `catch` block so I added one.

utils/pre-publish-check.js
kevinbarabash, line 7: The linter auto fixed this.

kevinbarabash: I'll update it separately.  It feels like we should almost have a linter for this, lol.

PR #1109, opened 2021-05-04
Handle mailto:, tel:, etc. hrefs in isClientSideUrl

kevinbarabash:
## Summary:
In #1107 I updated the logic to decide when to render a <a> vs a <Link>.  I didn't realize we used mailto: and tel: in webapp.  This PR handles those cases in a more general way which should hopefully be more future proof.

Issue: none

## Test plan:
- yarn jest packages/wonder-blocks-clickable/src/util/__tests__/is-client-side-url.js.test.js

### DIFF
--- a/packages/wonder-blocks-clickable/src/util/is-client-side-url.js
+++ b/packages/wonder-blocks-clickable/src/util/is-client-side-url.js
@@ -6,9 +6,5 @@
  * - true for all other values, e.g. /foo/bar
  */
 export const isClientSideUrl = (href: string): boolean => {
-    return (
-        !/^(https?:)?\/\//i.test(href) &&
-        !/^[a-zA-Z]+:/.test(href) &&
-        href !== "#"
-    );
+    return !/^(https?:)?\/\//i.test(href) && !/^(#[\w-]*|[\w\-.]+:)/.test(href);
 };
--- a/packages/wonder-blocks-clickable/src/util/is-client-side-url.js
+++ b/packages/wonder-blocks-clickable/src/util/is-client-side-url.js
@@ -2,12 +2,13 @@
 /**
  * Returns:
  * - false for hrefs staring with http://, https://, //.
- * - false for '#' and 'javascript:void(0);'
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-clickable/src/util/is-client-side-url.js
```diff
@@ -2,12 +2,13 @@
...
- * - false for '#' and 'javascript:void(0);'
+ * - false for '#', 'javascript:...', 'mailto:...', 'tel:...', etc.
  * - true for all other values, e.g. /foo/bar
  */
 export const isClientSideUrl = (href: string): boolean => {
     return (
         !/^(https?:)?\/\//i.test(href) &&
-        !/^(#|javascript:void\(0\);?)$/.test(href)
+        !/^[a-zA-Z]+:/.test(href) &&
+        href !== "#"
```
somewhatabstract, line 12: I think that these protocols may also be able to include hyphens, perhaps we can be more permissive than just a-zA-Z?

somewhatabstract (approved):
LGTM though I think maybe we should be more broad

packages/wonder-blocks-clickable/src/util/is-client-side-url.js
jandrade, line 12: I'm not sure if we have any case where we define a hash (e.g. `href="#math"`), would we need to cover that scenario as well?

kevinbarabash: We don't do that for any `Link`, but we'll probably swap out some of our `<a>`s for `<Link>`s in the future so we may as well be ready for that.

jandrade (approved):
LGTM! just a question related to matching "#".

packages/wonder-blocks-clickable/src/util/is-client-side-url.js
kevinbarabash: Good call.  Looking at https://en.wikipedia.org/wiki/List_of_URI_schemes it looks like `.` can be included too.

PR #1111, opened 2021-05-04
nefaurk/typography

nefaurk:
Foundation v2 prototype


packages/wonder-blocks-typography/src/components/text.stories.js
```diff
@@ -0,0 +1,478 @@
...
+        TypeSize.XSmall,
+    ];
+
+    return (
+        <ul className={css(sharedStyles.ul)}>
+            {Object.values(languages)
+                .sort()
+                .map((lCode, i) =>
+                    sizes.map((size, i) => (
+                        <String
```
jandrade, line 279: Neat!

packages/wonder-blocks-typography/src/components/text.js
```diff
@@ -0,0 +1,344 @@
...
+    HeadingSmall: {
+        fontSize: "20px",
+        lineHeight: "22px",
+    },
+    HeadingXSmall: {
+        fontSize: "12px",
+        lineHeight: "16px",
+    },
+    // Compact Heading
+    CompactHeading: {
```
jandrade, line 95: I remember seeing this in the Figma specs. In which situations do you think we would need to use `CompactHeading`?

static/source-serif-pro.css
```diff
@@ -0,0 +1,51 @@
+/* Adapted from https://fonts.googleapis.com/css?family=Source+Serif+Pro:400,600 */
```
jandrade, line 1: Thanks for moving these fonts to WB.

PR #1119, opened 2021-05-11
WB-1014 - Add `style` prop to `Spring` and `Strut`

RivasCVA:
## Summary:
Integrated the style prop to the Spring and Strut components under wonder-blocks-layout. Included new examples to StyleGuidist and React Storybook demonstrating the new change.

Also changed the components call in the Styleguide config for the Layout and Clickable components to prevent Storybook components from being picked up by StyleGuidist.

Issue: https://khanacademy.atlassian.net/browse/WB-1014

## Test plan:
1. Open StyleGuidist in your browser
2. Go to http://localhost:6060/#spring (strut is right below it)
3. See the code examples to see how you can now pass styles directly to the Spring and Strut components.

You can also view the React Storybook preview link to see the new additions for Spring and Strut.

Screen Shot 2021-05-11 at 5 10 04 PM
###### Simple SS example for Spring on StyleGuidist

### DIFF
--- a/packages/wonder-blocks-layout/src/components/strut.js
+++ b/packages/wonder-blocks-layout/src/components/strut.js
@@ -2,9 +2,11 @@
 import * as React from "react";
 
 import {View} from "@khanacademy/wonder-blocks-core";
+import type {StyleType} from "@khanacademy/wonder-blocks-core";
 
 type Props = {|
     size: number,
+    style?: StyleType,
 |};
 
 /**
@@ -14,7 +16,12 @@ type Props = {|
  */
 export default class Strut extends React.Component<Props> {
     render(): React.Node {
-        return <View aria-hidden="true" style={strutStyle(this.props.size)} />;
+        const {size, style} = this.props;
+        return (
+            <View style={style}>
+                <View aria-hidden="true" style={strutStyle(size)} />
+            </View>
+        );
     }
 }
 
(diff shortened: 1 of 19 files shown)


packages/wonder-blocks-layout/src/components/strut.js
```diff
@@ -14,7 +16,12 @@ type Props = {|
  */
 export default class Strut extends React.Component<Props> {
     render(): React.Node {
-        return <View aria-hidden="true" style={strutStyle(this.props.size)} />;
+        const {size, style} = this.props;
+        return (
+            <View style={style}>
+                <View aria-hidden="true" style={strutStyle(size)} />
+            </View>
```
kevinbarabash, line 23: It be preferable to return only a single `View`.  Is there a reason why `style` and `strutStyle(size)` can't both be set on a single `View`?

jeresig: Yeah, it looks like Spring is a single element, so I think we should change this to match.

jeresig (changes requested):
One small request regarding the construction of the Strut, but other than that it's looking good!

packages/wonder-blocks-layout/src/components/strut.js
RivasCVA: Hey guys. Thanks for bringing this up. I did originally have `Strut` be a single `View`, but it seemed that it was not applying any color from `style`. Now that i'm further looking into to it, it was because single-`View` `Strut` would have a natural height of zero if no height was specified. I got that fixed 👍🏻

jeresig (approved):
Awesome, thank you for the fix!

PR #1120, opened 2021-05-13
WB-1043 - Fix Dropdown list scrollbar padding

RivasCVA:
## Summary:
Fixed a wrong width calculation with the `wonder-blocks-dropdown` list
which was causing a padding issue with the scrollbar (when the list was
large enough for the scrollbar to show).

Also fixed an issue with the single-select dropdown highlighting the incorrect
selected item when the dropdown was re-opened.

Issue: https://khanacademy.atlassian.net/browse/WB-1043

## Test plan:
1. Open Styleguidist in browser (or React Storybook)
2. Go to: http://localhost:6060/#!/SingleSelect/17 (Storybook: http://localhost:59197/?path=/story/dropdown-multiselect--custom-styles-opened)
3. See that the dropdown scrollbar is perfectly aligned to the right edge (and also the selected item is correctly highlighted when dropdown is re-opened)

#### Screenshots of the scrollbar aligning correctly on different widths
dd-ss1
dd-ss2
dd-ss3

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/dropdown-core-virtualized.js
+++ b/packages/wonder-blocks-dropdown/src/components/dropdown-core-virtualized.js
@@ -101,11 +101,12 @@ class DropdownCoreVirtualized extends React.Component<Props, State> {
      */
     setWidth() {
         const rootNode = ((ReactDOM.findDOMNode(this): any): ?HTMLElement);
+        const parentNode = rootNode?.parentElement;
 
         // after the non-virtualized items are rendered, we get the container
         //  width to pass it to react-window's List
-        if (rootNode) {
-            const width = rootNode.getBoundingClientRect().width;
+        if (parentNode) {
+            const width = parentNode.getBoundingClientRect().width;
 
             this.setState({
                 width,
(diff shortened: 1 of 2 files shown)


jeresig (approved):
Thank you!! The scrollbar issue had been really bugging me... thanks for identifying and fixing both of these!

jandrade (approved):
LGTM! thanks for applying both fixes.

kevinbarabash:
I was playing around with the netlify preview and am still seeing some weirdness on Safari:
Screen Shot 2021-05-13 at 4 03 34 PM

After scrolling around a bit, it's a little better, but there horizontal scrollbar is still there:
Screen Shot 2021-05-13 at 4 03 48 PM

This was the "ActionMenu with custom dropdown style" example.

RivasCVA:
After some investigation, it seems that the issue is related to `action-item` when the user passes an `href` prop. In those situations, the `action-item` seems to have some type of padding on the right.

#### In this demo, I changed the background color of the dropdown to see the individual `action-items`
###### All three `action-items` that have the right "padding" have an href prop.
https://user-images.githubusercontent.com/60367213/118197087-de300780-b413-11eb-8f95-7a023009ee11.mp4

I'm going to create a new ticket for this because I snooped around for over an hour and could not seem to find where this extra padding is coming from. My guesses are that maybe `addStyle()` is doing this, but I can't seem to find where exactly.

The only thing I can fix right now is the horizontal bar by adding an `overflowX: hidden` to the Dropdown list (already made commit). I believe the other issue with **Safari** is linked to this padding problem.

PR #1121, opened 2021-05-14
WB-1032 - Add `light` prop to `Clickable`

RivasCVA:
## Summary:
Added the `light` prop to the `Clickable` component in
`wonder-blocks-clickable` which will cause the default focus ring to be
white instead of blue when the `light` prop is set to true.

Issue: https://khanacademy.atlassian.net/browse/WB-1032

## Test plan:
1. Open Styleguidist in browser
2. Go to: http://localhost:6060/#!/Clickable/3
3. Press 'tab' and see that the default focus ring is white when the
   `light` prop is set to true.

SS1

##### When `light` is set to false.
SS2


jeresig (approved):
Very nice! Definitely a gap in our API, thanks for fixing this!

jandrade (approved):
LGTM! Thanks for adding the story too 🎉

PR #1123, opened 2021-05-21
WB-1048 - Add TextField and the basic styles.

RivasCVA:
## Summary:
Adding TextField to `wonder-blocks-form`. Implemented the basic look and
styles of the TextField from the [Figma](https://www.figma.com/file/VbVu3h2BpBhH80niq101MHHE/Wonder-Blocks-(Web)?node-id=495%3A2) concept. No props or any other
abilities have been implemented. Only its styles.

###### Figma Default
figma-default
###### WB Default
default-text
default-placeholder

###### Figma Disabled
figma-disabled
###### WB Disabled
disabled-text
disabled-placeholder

###### Figma Error
figma-error
###### WB Error
error-text
error-placeholder

Issue: https://khanacademy.atlassian.net/browse/WB-1048

## Test plan:
1. You can view the `TextField` section on Styleguidist, but since no error
and disabled props have been implemented yet, it is stuck on its default state.
2. I attached screenshots of the different styles above. Those screenshots are
the result of directly changing the style of the TextField using the 'default', 'disabled', and 'error' style.

### DIFF
--- a/packages/wonder-blocks-form/src/components/text-field.js
+++ b/packages/wonder-blocks-form/src/components/text-field.js
@@ -3,33 +3,38 @@
 import * as React from "react";
 import {StyleSheet, css} from "aphrodite";
 
-import {View} from "@khanacademy/wonder-blocks-core";
 import Color from "@khanacademy/wonder-blocks-color";
+import Spacing from "@khanacademy/wonder-blocks-spacing";
 import {styles as typographyStyles} from "@khanacademy/wonder-blocks-typography";
 
 export default class TextField extends React.Component<{||}> {
     render(): React.Node {
         return (
-            <View style={[styles.container, styles.default]}>
-                <input
-                    className={css([
-                        styles.input,
-                        typographyStyles.LabelMedium,
-                    ])}
-                    placeholder="Placeholder"
-                />
-            </View>
+            <input
+                className={css([
+                    styles.input,
+                    typographyStyles.LabelMedium,
+                    styles.default,
+                ])}
+                placeholder="Placeholder"
+            />
         );
     }
 }
 
 const styles = StyleSheet.create({
-    container: {
-        width: 288,
+    input: {
+        width: "100%",
         height: 40,
         borderRadius: 4,
         boxSizing: "border-box",
-        padding: "10px 10px 10px 16px",
+        paddingLeft: Spacing.medium_16,
+        margin: 0,
+        outline: "none",
+        boxShadow: "none",
+        "::placeholder": {
+            color: Color.offBlack64,
+        },
     },
     default: {
         background: Color.white,
@@ -37,7 +42,7 @@ const styles = StyleSheet.create({
         color: Color.offBlack,
     },
     error: {
-        background: `linear-gradient(0deg, rgba(217, 41, 22, 0.06), rgba(217, 41, 22, 0.06)), ${Color.white}`,
+        background: "rgba(217, 41, 22, 0.06)",
         border: `1px solid ${Color.red}`,
         color: Color.offBlack,
     },
@@ -45,21 +50,5 @@ const styles = StyleSheet.create({
         background: Color.offWhite,
         border: `1px solid ${Color.offBlack16}`,
         color: Color.offBlack64,
-    },
-    input: {
-        flex: 1,
-        display: "flex",
-        alignItems: "center",
-        width: "100%",
-        background: "none",
-        color: "inherit",
-        border: "none",
-        outline: "none",
-        boxShadow: "none",
-        padding: 0,
-        margin: 0,
-        "::placeholder": {
-            color: Color.offBlack64,
-        },
     },
 });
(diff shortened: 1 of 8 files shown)


packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -0,0 +1,65 @@
...
+                    placeholder="Placeholder"
+                />
+            </View>
+        );
+    }
+}
+
+const styles = StyleSheet.create({
+    container: {
+        width: 288,
```
jandrade, line 28: We would prefer leaving this decision to the parent component so we don't add any initial constrains to these fields. The `width` should be set to `100%` and override it if we pass a custom width with the `style` prop.

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -0,0 +1,65 @@
...
+        height: 40,
+        borderRadius: 4,
+        boxSizing: "border-box",
+        padding: "10px 10px 10px 16px",
```
jandrade, line 32: Ugh I think I added this in the implementation spec, but maybe it's better if we only pass `padding-left=16px` and let the browser vertically center the input based on the `height` value (40 in our case, or whatever we pass via `style` prop).

Also, it'd be nice if you could use Spacing when changing this (`paddingLeft: Spacing.medium_16`).

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -0,0 +1,65 @@
...
+import {styles as typographyStyles} from "@khanacademy/wonder-blocks-typography";
+
+export default class TextField extends React.Component<{||}> {
+    render(): React.Node {
+        return (
+            <View style={[styles.container, styles.default]}>
+                <input
+                    className={css([
+                        styles.input,
+                        typographyStyles.LabelMedium,
```
jandrade, line 17: Nice! thanks for reusing `typography` here!.

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -0,0 +1,65 @@
...
+        boxSizing: "border-box",
+        padding: "10px 10px 10px 16px",
+    },
+    default: {
+        background: Color.white,
+        border: `1px solid ${Color.offBlack16}`,
+        color: Color.offBlack,
+    },
+    error: {
+        background: `linear-gradient(0deg, rgba(217, 41, 22, 0.06), rgba(217, 41, 22, 0.06)), ${Color.white}`,
```
jandrade, line 40: 🤔 I see that this is the value in Figma, but I think that gradient is using the same color in both steps, so maybe using a solid color instead of a gradient would get the same result.

jandrade (changes requested):
Thanks for adding the first set of changes and splitting the TextField into smaller PRs/tickets 🥇. 

Holding of approval to get your thoughts on the `View > input` vs `input`, and also to make some changes on the way we set styles (esp. padding and width).

Also, it would be great if you could add some Storybook examples so we can start tracking changes in Chromatic for every PR you add related to TextField 😃.

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -0,0 +1,65 @@
...
+import {StyleSheet, css} from "aphrodite";
+
+import {View} from "@khanacademy/wonder-blocks-core";
+import Color from "@khanacademy/wonder-blocks-color";
+import {styles as typographyStyles} from "@khanacademy/wonder-blocks-typography";
+
+export default class TextField extends React.Component<{||}> {
+    render(): React.Node {
+        return (
+            <View style={[styles.container, styles.default]}>
```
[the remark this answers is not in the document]
RivasCVA: **I'm on board with just having the `input` alone**. I actually did start off by just having the `input` alone without a `View` wrapper, but then I noticed on Figma that the TextField had a few design layers so I assumed that it meant it needed a wrapper `View` to apply those styles.

packages/wonder-blocks-form/src/components/text-field.js
RivasCVA: Got it 👍

packages/wonder-blocks-form/src/components/text-field.js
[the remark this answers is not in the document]
RivasCVA: Also, in terms of the "icon" or another element next to the `input`, I guess if that's something we might want later it would be better to have a separate component that derives from TextField that handles that.

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -0,0 +1,54 @@
...
+        boxShadow: "none",
+        "::placeholder": {
+            color: Color.offBlack64,
+        },
+    },
+    default: {
+        background: Color.white,
+        border: `1px solid ${Color.offBlack16}`,
+        color: Color.offBlack,
+    },
```
jandrade, line 43: It's really nice how you splitted core CSS and theme here. This will make it easier to override with `error` and `disabled` once you include more props :)

jandrade (approved):
Looks great to me! Thanks for making the changes and adding the story.

PR #1124, opened 2021-05-24
WB-1051 - Add States to TextField

RivasCVA:
## Summary:
Integrated the States that TextField will use. This includes `enabled`,
`disabled`, `error`, and `focused`. Included the `focused` style for
when the user is focused on (editing) the TextField. I attached images
below of this new style.

Issue: https://khanacademy.atlassian.net/browse/WB-1051

## Test plan:
1. View the generated Storybook link. Check the "Form / TextField" section.
2. You can also view Styleguidist under the "Form" section.
##### onFocus [no text]
focus-notext

##### onFocus [text]
focus-text

### DIFF
--- a/packages/wonder-blocks-form/src/components/text-field.js
+++ b/packages/wonder-blocks-form/src/components/text-field.js
@@ -7,46 +7,75 @@ import Color from "@khanacademy/wonder-blocks-color";
 import Spacing from "@khanacademy/wonder-blocks-spacing";
 import {styles as typographyStyles} from "@khanacademy/wonder-blocks-typography";
 
-type State = {|
-    enabled: boolean,
+type Props = {|
+    /**
+     * Makes a read-only input field that cannot be focused.
+     */
     disabled: boolean,
+|};
+
+type DefaultProps = {|
+    disabled: $PropertyType<Props, "disabled">,
+|};
+
+type State = {|
+    /**
+     * Displayed when the validation fails.
+     */
     error: ?string,
+
+    /**
+     * The user focuses on this field.
+     */
     focused: boolean,
 |};
 
-export default class TextField extends React.Component<{||}, State> {
-    state: State = {
-        enabled: true,
+/**
+ * A TextField is an element used to accept a single line of text from the user.
+ */
+export default class TextField extends React.Component<Props, State> {
+    static defaultProps: DefaultProps = {
         disabled: false,
+    };
+
+    state: State = {
         error: null,
         focused: false,
     };
 
-    handleOnFocus: () => void = () => {
+    handleOnFocus: (
+        event: SyntheticFocusEvent<HTMLInputElement>,
+    ) => mixed = () => {
         this.setState({
             focused: true,
         });
     };
 
-    handleOnBlur: () => void = () => {
+    handleOnBlur: (
+        event: SyntheticFocusEvent<HTMLInputElement>,
+    ) => mixed = () => {
         this.setState({
             focused: false,
         });
     };
 
     render(): React.Node {
+        const {disabled} = this.props;
         return (
             <input
                 className={css([
                     styles.input,
                     typographyStyles.LabelMedium,
-                    this.state.enabled && styles.default,
-                    this.state.disabled && styles.disabled,
-                    this.state.error && styles.error,
-                    this.state.focused && styles.focused,
+                    styles.default,
+                    // Prioritizes disabled, then focused, then error (if any)
+                    disabled
+                        ? styles.disabled
+                        : this.state.focused
+                        ? styles.focused
+                        : this.state.error && styles.error,
                 ])}
                 placeholder="Placeholder"
-                disabled={this.state.disabled}
+                disabled={disabled}
                 onFocus={this.handleOnFocus}
                 onBlur={this.handleOnBlur}
             />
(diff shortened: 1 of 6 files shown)


packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -7,16 +7,48 @@ import Color from "@khanacademy/wonder-blocks-color";
 import Spacing from "@khanacademy/wonder-blocks-spacing";
 import {styles as typographyStyles} from "@khanacademy/wonder-blocks-typography";
 
-export default class TextField extends React.Component<{||}> {
+type State = {|
+    enabled: boolean,
```
jandrade, line 11: In this case it's better to have only one variable that holds the `enabled/disabled` state. There could be situations for X reasons where both could be `true` and that would lead to an invalid state. In that case it's better to get rid of `enabled` and use only `disabled=true|false` instead.

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -7,16 +7,48 @@ import Color from "@khanacademy/wonder-blocks-color";
...
+    disabled: boolean,
```
jandrade, line 12: Based on the previous comment, `disabled` should be treated as a prop that is only defined in the component instance and shouldn't change internally (only via the `disabled` prop).

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -7,16 +7,48 @@ import Color from "@khanacademy/wonder-blocks-color";
...
+
+export default class TextField extends React.Component<{||}, State> {
+    state: State = {
+        enabled: true,
+        disabled: false,
+        error: null,
+        focused: false,
+    };
+
+    handleOnFocus: () => void = () => {
```
jandrade, line 25: You would want to add flow types for the function signature here. Not needed right now, but it'd be good to have it in mind when you add the `onFocus` and `onBlur` props later.

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -32,23 +64,37 @@ const styles = StyleSheet.create({
...
     },
     disabled: {
         background: Color.offWhite,
         border: `1px solid ${Color.offBlack16}`,
         color: Color.offBlack64,
+        "::placeholder": {
+            color: Color.offBlack32,
+        },
+    },
+    focused: {
```
jandrade, line 92: Nice! thanks.

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -7,16 +7,48 @@ import Color from "@khanacademy/wonder-blocks-color";
...
+    };
+
     render(): React.Node {
         return (
             <input
                 className={css([
                     styles.input,
                     typographyStyles.LabelMedium,
-                    styles.default,
+                    this.state.enabled && styles.default,
```
jandrade, line 43: Based on the first comment, this would be translated to `styles.default, ....` and it's replaced with whatever UI state changes in the component (via `state` with `error, focused` or via `props` with `disabled`).

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -7,16 +7,48 @@ import Color from "@khanacademy/wonder-blocks-color";
...
+            focused: true,
+        });
+    };
+
+    handleOnBlur: () => void = () => {
+        this.setState({
+            focused: false,
+        });
+    };
+
```
jandrade, line 36: As you are adding `state` and handlers here, this is a good point to start adding unit tests so we can verify that some behaviors in the component work as we intend :). Happy to chat more about this part in detail.

jandrade (changes requested):
Thanks for adding more functionality on top of the first PR. Things are looking good, but I'm requesting some changes before approving:

1. Removing `enabled` from `state`.
2. Moving `disabled` to `Props`. (it'd be great if you add a new story with this prop set to `true`).
3. Adding unit tests to start verifying that the component works as expected.

packages/wonder-blocks-form/src/components/text-field.js
RivasCVA: Yes. That makes total sense. I was a bit confused on why the spec had both when I started to code this up. I went too deep into this and just thought that I would have to internally ensure that `enabled` and `disabled` were never true at the same time. This clears it up 😄

jandrade: yeah no worries, sometimes there will be some stuff that doesn't translate exactly from design to UI :). And I think that's one of the reasons why we use implementation specs, so we can create a bridge between Design and Code.

RivasCVA:
On this new change, I moved `disabled` from a state to a prop. I removed the `enabled` state. I added a `disabled` example on Storyboard, as well as on Styleguidist. I added a few unit tests to test the states and prop.

jandrade (approved):
Looks great! Thanks for implementing the changes 💯

RivasCVA:
The parent pull-request ([#1123]](https://github.com/Khan/wonder-blocks/pull/1123)) has been merged into `master`, but this branch (`wb-1051`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #1126, opened 2021-05-26
WB-1049 - Add `id`, `type`, `value`, and `onChange` Props to TextField

RivasCVA:
## Summary:
* Added the Props `id`, `type`, `value`, and `onChange` to the TextField.
* Included unit tests for all of these new props (Coverage displays 100% for the TextField component after the new unit tests).
* Added more examples to Styleguidist and React Storybook.

Issue: WB-1049

## Test plan:
1. Open Styleguidist in the browser
2. Check the new examples under Form -> TextField
3. You can also view the new examples on React Storybook under Form -> Textfield

###### Number
tf-number

###### Password
tf-password

##### NOTE: Will also add `email` and `tel` examples when the `validation` prop is added.

PR #1128, opened 2021-05-28
WB-1052 - Add `validation` and `onError` Props to TextField

RivasCVA:
## Summary:
Adding the `validation` and `onError` Props to TextField.
The `validation` prop will be called after every value change.
The `onError` prop will be called after every failed `validation`.

Added unit tests to test both new props.

Included examples on Styleguidist and React Storybook of `validation`
and `onError` working with the `password`, `email`, and `telephone`
type TextFields.

Issue: https://khanacademy.atlassian.net/browse/WB-1052

## Test plan:
1. Open Styleguidist (or React Storybook)
2. Go to the TextField (under Form -> TextField)
3. See the given examples, specially the `password`, `email`, and
   `telephone` TextField examples where `validation` and `onError` are at work.

---

##### Failed validation on `password` TextField
tf-password

##### Failed validation on `email` TextField
tf-email

#### Failed validation on `telephone` TextField
tf-telephone

### DIFF
--- a/packages/wonder-blocks-form/src/components/__tests__/text-field.test.js
+++ b/packages/wonder-blocks-form/src/components/__tests__/text-field.test.js
@@ -176,7 +176,7 @@ describe("TextField", () => {
         wrapper.simulate("change", {target: {value: newValue}});
 
         // Assert
-        expect(handleValidation.mock.results[0].value).toEqual(undefined);
+        expect(handleValidation).toHaveReturnedWith(undefined);
     });
 
     it("validation is given an invalid input", () => {
@@ -203,7 +203,7 @@ describe("TextField", () => {
         wrapper.simulate("change", {target: {value: newValue}});
 
         // Assert
-        expect(handleValidation.mock.results[0].value).toEqual(errorMessage);
+        expect(handleValidation).toHaveReturnedWith(errorMessage);
     });
 
     it("onValidation is called after input validation", () => {
--- a/packages/wonder-blocks-form/src/components/__tests__/text-field.test.js
+++ b/packages/wonder-blocks-form/src/components/__tests__/text-field.test.js
@@ -130,4 +130,108 @@ describe("TextField", () => {
         // Assert
         expect(handleOnChange).toHaveBeenCalledWith(newValue);
     });
+
+    it("validation is called when value changes", () => {
+        // Arrange
+        const handleValidation = jest.fn((value: string): ?string => {});
+
+        const wrapper = mount(
+            <TextField
+                id={"tf-1"}
+                value="Text"
+                validation={handleValidation}
+                onChange={() => {}}
+                disabled={true}
+            />,
+        );
+
+        // Act
+        const newValue = "Text2";
+        wrapper.simulate("change", {target: {value: newValue}});
+
+        // Assert
+        expect(handleValidation).toHaveBeenCalledWith(newValue);
+    });
+
+    it("validation is given a valid input", () => {
+        // Arrange
+        const handleValidation = jest.fn((value: string): ?string => {
+            if (value.length < 8) {
+                return "Value is too short";
+            }
+        });
+
+        const wrapper = mount(
+            <TextField
+                id={"tf-1"}
+                value="TextIsLong"
+                validation={handleValidation}
+                onChange={() => {}}
+                disabled={true}
+            />,
+        );
+
+        // Act
+        const newValue = "TextIsLongerThan8";
+        wrapper.simulate("change", {target: {value: newValue}});
+
+        // Assert
+        expect(handleValidation.mock.results[0].value).toEqual(undefined);
+    });
+
+    it("validation is given an invalid input", () => {
+        // Arrange
+        const errorMessage = "Value is too short";
+        const handleValidation = jest.fn((value: string): ?string => {
+            if (value.length < 8) {
+                return errorMessage;
+            }
+        });
+
+        const wrapper = mount(
+            <TextField
+                id={"tf-1"}
+                value="TextIsLongerThan8"
+                validation={handleValidation}
+                onChange={() => {}}
+                disabled={true}
+            />,
+        );
+
+        // Act
+        const newValue = "Text";
+        wrapper.simulate("change", {target: {value: newValue}});
+
+        // Assert
+        expect(handleValidation.mock.results[0].value).toEqual(errorMessage);
+    });
+
+    it("onError is called on invalid validation", () => {
+        // Arrange
+        const errorMessage = "Value is too short";
+        const handleOnError = jest.fn((errorMessage: string) => {});
+        const handleValidation = jest.fn((value: string): ?string => {
+            if (value.length < 8) {
+                return errorMessage;
+            }
+        });
+
+        const wrapper = mount(
+            <TextField
+                id={"tf-1"}
+                value="TextIsLongerThan8"
+                validation={handleValidation}
+                onChange={() => {}}
+                onError={handleOnError}
+                disabled={true}
+            />,
+        );
+
+        // Act
+        const newValue = "Text";
+        wrapper.simulate("change", {target: {value: newValue}});
+
+        // Assert
+        expect(handleOnError).toHaveBeenCalledWith(errorMessage);
+    });
 });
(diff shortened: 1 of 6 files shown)


packages/wonder-blocks-form/src/components/text-field.md
```diff
@@ -54,31 +54,215 @@ class TextFieldExample extends React.Component {
...
 class TextFieldExample extends React.Component {
     constructor(props) {
         super(props);
         this.state = {
             value: "Password123",
+            errorMessage: null,
         };
+        this.handleOnChange = this.handleOnChange.bind(this);
+        this.handleValidation = this.handleValidation.bind(this);
+        this.handleOnError = this.handleOnError.bind(this);
```
jandrade, line 76: Ugh this is unfortunate, but hopefully we'll avoid using this once we fully migrate to Storybook 🤞

packages/wonder-blocks-form/src/components/text-field.md
```diff
@@ -54,31 +54,215 @@ class TextFieldExample extends React.Component {
...
+    handleOnChange(newValue) {
+        this.setState({value: newValue});
+    }
+
+    handleValidation(value) {
+        const emailRegex = /^[^@\s]+@[^@\s.]+\.[^@.\s]+$/;
+        if (!emailRegex.test(value)) {
+            return "Please enter a valid email";
+        }
+        this.setState({errorMessage: null})
```
jandrade, line 160: mmm I'm thinking that this function should only take care of validating the value without having to modify the state (or anything related about React lifecycle and state). I understand that you had to set state here to reset the error state in case the field is valid, but I wonder if we need to modify the `onError` prop to be called no matter if it's an error or not.

That said, I can see different approaches:
1. Modify `onError` to be called every time the validation is performed.
2. Rename `onError` to something more meaningful: e.g. `onValidation(errorMessage: string)`.

IMO using `onValidation` feels more natural in the sense that it's the function that is called after performing `validation`, but ofc I'd like to hear your thoughts on this, and probably explore other possibilities.

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -70,8 +81,16 @@ export default class TextField extends React.Component<Props, State> {
...
-        const {onChange} = this.props;
-        onChange(event.target.value);
+        const {validation, onChange, onError} = this.props;
+        const newValue = event.target.value;
+        if (validation) {
+            const isError = validation(newValue);
+            this.setState({error: isError});
+            if (isError && onError) {
+                onError(isError);
+            }
```
jandrade, line 91: Please see my comment related to renaming/calling `onError`.

packages/wonder-blocks-form/src/components/text-field.md
jandrade: BTW Thanks for adding this example! It helped to identify this situation ahead of time :)

jandrade (commented):
Things are looking good, but want to open a conversation about `onError` before continuing with the review.

Thanks for all the examples and tests that you have provided 💯.

packages/wonder-blocks-form/src/components/__tests__/text-field.test.js
```diff
@@ -130,4 +130,108 @@ describe("TextField", () => {
...
+                disabled={true}
+            />,
+        );
+
+        // Act
+        const newValue = "TextIsLongerThan8";
+        wrapper.simulate("change", {target: {value: newValue}});
+
+        // Assert
+        expect(handleValidation.mock.results[0].value).toEqual(undefined);
```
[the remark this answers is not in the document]
RivasCVA: Actually... I remember now why I used that:
The `toHaveBeenCalledWith` matcher only tests the parameter passed to the `handleValidation` function, but I wanted to test against the **return value** of that function (to ensure it returned `undefined`, meaning a valid input).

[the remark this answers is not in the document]
jandrade: Got it! thanks for the context. In that case, maybe you could try with `toHaveReturnedWith` and see if it gets the same result?

packages/wonder-blocks-form/src/components/__tests__/text-field.test.js
```diff
@@ -130,4 +130,108 @@ describe("TextField", () => {
...
+                disabled={true}
+            />,
+        );
+
+        // Act
+        const newValue = "Text";
+        wrapper.simulate("change", {target: {value: newValue}});
+
+        // Assert
+        expect(handleValidation.mock.results[0].value).toEqual(errorMessage);
```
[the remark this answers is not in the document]
RivasCVA: As commented [above](https://github.com/Khan/wonder-blocks/pull/1128#discussion_r641780627), I want to test the **return value** of the `handleValidation` function, and not the parameter.

Here I want to ensure it returns the string error message that the function is expected to return when the validation fails.

packages/wonder-blocks-form/src/components/text-field.md
RivasCVA: 🤞

packages/wonder-blocks-form/src/components/text-field.md
RivasCVA: I see.... I agree with the `onValidation` route.

At first, I viewed it as redundant since `validation` will call `onValidation`, so why not just put everything in the `validation` function? 

But then thinking deeper about it, I can see why it would be better to have both:
* `validation` will **only** handle validating the value
* `onValidation` will handle or modify any states or variables.

RivasCVA:
What changed:
* Replaced all instances of `onError` with `onValidation`. The new prop, `onValidation`, is called every time the input is validated, whether there is an error or not.
* Updated all examples in Styleguidist and React Storybook to use `onValidation` instead of `onError`.
* Updated the unit test handling `onError` to instead test `onValidation`.

packages/wonder-blocks-form/src/components/__tests__/text-field.test.js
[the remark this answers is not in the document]
RivasCVA: It works with `toHaveReturnedWith` 🎉 . Thanks for teaching me of that new matcher 🙏  (I Googled it and no such matcher popped up in my search).

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -70,8 +81,16 @@ export default class TextField extends React.Component<Props, State> {
     handleOnChange: (event: SyntheticInputEvent<HTMLInputElement>) => mixed = (
         event,
     ) => {
-        const {onChange} = this.props;
-        onChange(event.target.value);
+        const {validation, onValidation, onChange} = this.props;
+        const newValue = event.target.value;
+        if (validation) {
+            const maybeError = validation(newValue);
```
jandrade, line 87: Great! it feels more consistent

jandrade (approved):
This looks great! thank you so much for changing to `onValidation`.

One more thing, could you please update the implementation spec to match this new approach?

Great progress!! 🚀

RivasCVA:
> This looks great! thank you so much for changing to `onValidation`.
> 
> One more thing, could you please update the implementation spec to match this new approach?
> 
> Great progress!! 🚀

I'm on it 📝

jandrade:
I noticed that we missed one case... We need to validate too if the initial value is valid (and if the `validation` prop is set).
Not sure if you have plans to add that in the future, but if not, I think it would be good to add it with this PR.

RivasCVA:
> I noticed that we missed one case... We need to validate too if the initial value is valid (and if the `validation` prop is set).
> Not sure if you have plans to add that in the future, but if not, I think it would be good to add it with this PR.

Good pointer. I've already started working on the next PR, so I'll attach this "addition" on that new PR.

RivasCVA:
> We need to validate too if the initial value is valid (and if the `validation` prop is set).

BTW, you actually meant "if the initial value is **invalid**," right?

jandrade:
> > We need to validate too if the initial value is valid (and if the `validation` prop is set).
> 
> BTW, you actually meant "if the initial value is **invalid**," right? So that there is an error by default at the start.

that's right :)

RivasCVA:
The parent pull-request ([#1126]](https://github.com/Khan/wonder-blocks/pull/1126)) has been merged into `master`, but this branch (`wb-1052`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #1130, opened 2021-05-31
[WB-498] Fix obscurity check for when popper can obscure the anchor

somewhatabstract:
## Summary:
Firstly, this adds flow types to popper.js. From that, I fixed an edge case
where the anchor element passed to `isObscured` could be just a reference
object - I don't know under what circumstances that occurs, though I had seen
an error along those lines in the past, so that should be fixed if it happens
anywhere.

Finally, and most importantly, I have fixed the issue that would cause the
popper (whether a tooltip, menu, dropdown, or popover) to flicker or not appear
when being scrolled near the top of the screen. It turns out this was because
the popper portion was coming back as something covering the anchor element.
To work around this, the popper (or child of a popper, as was the case with
the action menu) is set to not receive pointer events while the obscurity check
is completed.

Issue: WB-498

## Test plan:
With the test site, you can compare with "prod".
In this video, you can see the behavior in production that clearly illustrates
the problem being fixed.

https://user-images.githubusercontent.com/1266297/120145758-a9f56e80-c1a9-11eb-9879-131406e755a6.mov

Using the test site, one can see that this no longer occurs.

https://user-images.githubusercontent.com/1266297/120146006-09ec1500-c1aa-11eb-853f-922662992445.mov

### DIFF
--- a/config/jest/test.config.js
+++ b/config/jest/test.config.js
@@ -22,6 +22,7 @@ module.exports = {
     },
     collectCoverageFrom: [
         "packages/**/*.js",
+        "shared-unpackaged/**/*.js",
         "!packages/**/*.stories.js",
         "!packages/**/*.flowtest.js",
         "!packages/**/dist/**/*.js",
--- a/flow-typed/popper.js
+++ b/flow-typed/popper.js
@@ -1,0 +1,169 @@
+// @flow
+
+declare module "popper.js" {
+    declare type Position = "top" | "right" | "bottom" | "left";
+
+    declare type Placement =
+        | "auto-start"
+        | "auto"
+        | "auto-end"
+        | "top-start"
+        | "top"
+        | "top-end"
+        | "right-start"
+        | "right"
+        | "right-end"
+        | "bottom-end"
+        | "bottom"
+        | "bottom-start"
+        | "left-end"
+        | "left"
+        | "left-start";
+
+    declare type Boundary = "scrollParent" | "viewport" | "window";
+
+    declare type Behavior = "flip" | "clockwise" | "counterclockwise";
+
+    declare type ModifierFn = (data: Data, options: Object) => Data;
+
+    declare type Attributes = {|
+        "x-out-of-boundaries": "" | false,
+        "x-placement": Placement,
+    |};
+
+    declare type Padding = {|
+        top?: number,
+        bottom?: number,
+        left?: number,
+        right?: number,
+    |};
+
+    declare type BaseModifier = {|
+        order?: number,
+        enabled?: boolean,
+        fn?: ModifierFn,
+    |};
+
+    declare type Modifiers = {
+        shift?: BaseModifier,
+        offset?: {|
+            ...BaseModifier,
+            offset?: number | string,
+        |},
+        preventOverflow?: {|
+            ...BaseModifier,
+            priority?: Position[],
+            padding?: number | Padding,
+            boundariesElement?: Boundary | Element,
+            escapeWithReference?: boolean,
+        |},
+        keepTogether?: BaseModifier,
+        arrow?: {|
+            ...BaseModifier,
+            element?: string | Element,
+        |},
+        flip?: {|
+            ...BaseModifier,
+            behavior?: Behavior | Position[],
+            padding?: number | Padding,
+            boundariesElement?: Boundary | Element,
+            flipVariations?: boolean,
+            flipVariationsByContent?: boolean,
+        |},
+        inner?: BaseModifier,
+        hide?: BaseModifier,
+        applyStyle?: {|
+            ...BaseModifier,
+            onLoad?: Function,
+            gpuAcceleration?: boolean,
+        |},
+        computeStyle?: {|
+            ...BaseModifier,
+            gpuAcceleration?: boolean,
+            x?: "bottom" | "top",
+            y?: "left" | "right",
+        |},
+
+        [name: string]: ?{
+            ...BaseModifier,
+            [key: string]: any,
+            ...
+        },
+        ...
+    };
+
+    declare type Offset = {|
+        top: number,
+        left: number,
+        width: number,
+        height: number,
+    |};
+
+    declare type Data = {|
+        instance: Popper,
+        placement: Placement,
+        originalPlacement: Placement,
+        flipped: boolean,
+        hide: boolean,
+        arrowElement: Element,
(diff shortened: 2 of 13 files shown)


flow-typed/popper.js
```diff
@@ -0,0 +1,170 @@
+// @flow
+// This hand-crafted file is based on the TypeScript types.
```
somewhatabstract, line 2: Added these first so that I could be sure things were getting the right types.

config/jest/test.config.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
     },
     collectCoverageFrom: [
         "packages/**/*.js",
+        "shared-unpackaged/**/*.js",
```
somewhatabstract, line 25: Turns out we weren't collecting coverage data for the unpackaged code that gets shared in several places (the sharing is done this way because we didn't want to expose these pieces as part of a public API).

shared-unpackaged/get-bounds.js
```diff
@@ -0,0 +1,35 @@
+// @flow
+import type {ReferenceObject} from "popper.js";
+import type {Bounds} from "./types.js";
+import getEdges from "./get-edges.js";
+
+export default function getBounds(
+    element: Element | ReferenceObject,
+    withoutEdges: boolean = false,
```
somewhatabstract, line 8: I am not sure why the bounds were calculated with the margin, padding, and border removed. I imagine it was to base things off the "visible" boundary of the item.

I did play around with not doing this and did not immediately notice a difference, but that doesn't mean there isn't one, so I left this as-is so that we can pick and choose.

shared-unpackaged/is-obscured.js
```diff
@@ -26,55 +34,70 @@ export default function isObscured(element: Element): boolean {
     // going to end up hiding, so, you know, probably don't do that.
     // We're not explicitly checking for that CSS since it's a corner-case and
     // would impact perf of the regular cases if we were always checking it.
-    // TODO(somewhatabstract): Consider how we might mitigate the pointer-events
-    // issue and make this call more robust.
-    const anchorRect = element.getBoundingClientRect();
```
somewhatabstract, line 31: This code is all refactored elsewhere for re-use and such.

shared-unpackaged/is-obscured.js
```diff
@@ -26,55 +34,70 @@ export default function isObscured(element: Element): boolean {
...
+    // viewport is zoomed such that both corners are off screen but the rest
+    // isn't. In this case some browsers don't return the element from
+    // `elementFromPoint` then doesn't return the element.
+    // Also, consider how we might mitigate the pointer-events issue and make
+    // this call more robust.
+    const bounds = getBounds(anchorElement);
+
+    // This method does the main work, taking some coordinates and determining
+    // if our element is visible at that point or not.
+    const isVisible = (x: number, y: number): boolean => {
```
somewhatabstract, line 48: I turned the previous code into a reusable "isVisible" call that takes coordinates. This reduced repetition and also meant we can limit the work we do...see further on...

shared-unpackaged/is-obscured.js
```diff
@@ -26,55 +34,70 @@ export default function isObscured(element: Element): boolean {
...
+        // If element is a reference object, all we have to work with is
+        // intersection for checking obscurity. Since this doesn't cover
+        // parent/child relationships in the DOM, it's not really effective
+        // on its own and is possibly about as good as just returning `true`.
+        const intersection =
+            elAtPoint && getIntersection(bounds, getBounds(elAtPoint, true));
+        return (
+            intersection?.horizontal !== "within" ||
+            intersection?.vertical !== "within"
+        );
```
somewhatabstract, line 88: Since adding flowtypes, I discovered that this function can get an `Element` or `ReferenceObject` and in the latter case, we seem to have much less to work with. I had noted, especially after hot-reloading, that we got errors around this problem, so I added some checks and code to cope with that. In truth, if we get here, our visibility check is much less useful, but at least it will do something half right now.

shared-unpackaged/is-obscured.js
```diff
@@ -26,55 +34,70 @@ export default function isObscured(element: Element): boolean {
...
+    // NOTE: We are using functions here so that we only do as much work
+    // as we need to, short-circuiting as soon as we have a definitive
+    // answer.
+    const isTopLeftVisible = () => isVisible(bounds.left, bounds.top);
+    const isBottomRightVisible = () => isVisible(bounds.right, bounds.bottom);
+    const isCenterVisible = () =>
+        isVisible(
+            bounds.left + (bounds.right - bounds.left) / 2,
+            bounds.top + (bounds.bottom - bounds.top) / 2,
+        );
```
somewhatabstract, line 100: Added a check for the center of the anchor too, to improve this check a bit. Also, made these function calls so that we only execute as much code as we need to, which should speed this check up in many cases.

shared-unpackaged/is-obscured.js
```diff
@@ -26,55 +34,70 @@ export default function isObscured(element: Element): boolean {
...
+
+    return !isTopLeftVisible() && !isBottomRightVisible() && !isCenterVisible();
```
somewhatabstract, line 102: So now, with the function calls, we short circuit the work we do if an earlier check passes.

shared-unpackaged/is-obscured.js
```diff
@@ -26,55 +34,70 @@ export default function isObscured(element: Element): boolean {
...
+            // Remove pointer events so that we can look through it.
+            elAtPoint.style.pointerEvents = "none";
+            try {
+                const visible = isVisible(x, y);
+                return visible;
+            } finally {
+                // Make sure we put things back the way we found them. :)
+                elAtPoint.style.pointerEvents = pointerEventsStyle;
+            }
+        }
```
somewhatabstract, line 66: This is the fix for the bug. It turned out that sometimes (especially when the anchor was scrolling off the top of the screen or was in the top-left region of the screen), the popper itself (or a child thereof) would be the result of `elementFromPoint` for the bottom right check. So, now if that occurs, we briefly deprive the popper (or the child) of pointer events and reperform the check, making sure to put the pointer events back to what they were when we're done.

This then means that we get a proper check of the anchor visibility based on non-anchor/popper elements, and all works as we want.

jeresig (approved):
Thanks for fixing this!

PR #1131, opened 2021-06-01
WB-1053 - Add `onKeyDown`, `onFocus`, and `onBlur` Props to TextField

RivasCVA:
## Summary:
Adding `onKeyDown`, `onFocus`, and `onBlur` props to TextField. Also
adding the ability to validate the initial value of TextField on
component mount and call `onValidation` as necessary.

Issue: https://khanacademy.atlassian.net/browse/WB-1053

## Test plan:
1. Open Styleguidist (or React Storybook).
2. Go to the TextField under Form -> TextField.
3. Notice how pressing "Enter" while the TextField is focused will blur the TextField.
4. Also notice how the error message (on invalid input) will only show when the TextField is not focused (aka blurred).

---

##### Error message only showing when TextField is blurred (1)
tf-password

##### Error message only showing when TextField is blurred (2)
tf-email

##### Error message shows on component mount if initial value is invalid
tf-email-initial

### DIFF
--- a/packages/wonder-blocks-form/src/components/text-field.js
+++ b/packages/wonder-blocks-form/src/components/text-field.js
@@ -54,12 +54,12 @@ type Props = {|
     /**
      * Called when the element has been focused.
      */
-    onFocus?: () => mixed,
+    onFocus?: (event: SyntheticFocusEvent<HTMLInputElement>) => mixed,
 
     /**
      * Called when the element has been blurred.
      */
-    onBlur?: () => mixed,
+    onBlur?: (event: SyntheticFocusEvent<HTMLInputElement>) => mixed,
 |};
 
 type DefaultProps = {|
@@ -125,27 +125,27 @@ export default class TextField extends React.Component<Props, State> {
         onChange(newValue);
     };
 
-    handleOnFocus: (
-        event: SyntheticFocusEvent<HTMLInputElement>,
-    ) => mixed = () => {
+    handleOnFocus: (event: SyntheticFocusEvent<HTMLInputElement>) => mixed = (
+        event,
+    ) => {
         const {onFocus} = this.props;
         this.setState({
             focused: true,
         });
         if (onFocus) {
-            onFocus();
+            onFocus(event);
         }
     };
 
-    handleOnBlur: (
-        event: SyntheticFocusEvent<HTMLInputElement>,
-    ) => mixed = () => {
+    handleOnBlur: (event: SyntheticFocusEvent<HTMLInputElement>) => mixed = (
+        event,
+    ) => {
         const {onBlur} = this.props;
         this.setState({
             focused: false,
         });
         if (onBlur) {
-            onBlur();
+            onBlur(event);
         }
     };
 
--- a/packages/wonder-blocks-form/src/components/text-field.js
+++ b/packages/wonder-blocks-form/src/components/text-field.js
@@ -45,6 +45,21 @@ type Props = {|
      * Called when the value has changed.
      */
     onChange: (newValue: string) => mixed,
+
+    /**
+     * Called when a key is pressed.
+     */
+    onKeyDown?: (event: SyntheticKeyboardEvent<HTMLInputElement>) => mixed,
+
+    /**
+     * Called when the element has been focused.
+     */
+    onFocus?: () => mixed,
+
+    /**
+     * Called when the element has been blurred.
+     */
+    onBlur?: () => mixed,
 |};
 
 type DefaultProps = {|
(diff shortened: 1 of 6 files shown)


packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -78,39 +93,56 @@ export default class TextField extends React.Component<Props, State> {
...
-        const {validation, onValidation, onChange} = this.props;
-        const newValue = event.target.value;
+    componentDidMount() {
+        this.maybeValidate(this.props.value);
+    }
+
+    maybeValidate: (newValue: string) => void = (newValue) => {
+        const {validation, onValidation} = this.props;
         if (validation) {
             const maybeError = validation(newValue);
```
jeresig, line 103: Somewhat unrelated but I wonder if we should make sure the value falls back to being `null` as the default, if no response is provided, that way it'll match the initial state of `error` (right now if `undefined` is returned that will cause a re-render).
```suggestion
            const maybeError = validation(newValue) || null;
```

jeresig (approved):
Looks great, thank you! Just a few thoughts inline, nothing blocking.

packages/wonder-blocks-form/src/components/text-field.js
RivasCVA: Thanks on that! I was treating `undefined` and `null` as the same, but you're right, it would cause a re-render when `error` goes from its initial `null` state to an `undefined` state on the first invalid input.

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -78,39 +93,56 @@ export default class TextField extends React.Component<Props, State> {
...
     };
 
-    handleOnChange: (event: SyntheticInputEvent<HTMLInputElement>) => mixed = (
-        event,
-    ) => {
-        const {validation, onValidation, onChange} = this.props;
-        const newValue = event.target.value;
+    componentDidMount() {
+        this.maybeValidate(this.props.value);
+    }
```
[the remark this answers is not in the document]
RivasCVA: Ideally we **do** want to fire `onValidation` on the initial render mainly because we would want to show any error message if the initial value for the TextField is invalid.

How do you think we could go about ensuring that the `error` state is set on the server-side, but still be able to call `onValidation` on the first render? My attempt is to replace line 99 on your suggested change with just calling my `maybeValidate()` method, but that does not seem to update the `error` state because I believe `setState` does not work inside a constructor.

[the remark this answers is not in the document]
RivasCVA: **UPDATE:**
Doing someting like this would also work...
```js
constructor(props: Props) {
    super(props);
    if (props.validation) {
        this.state.error = props.validation(props.value) || null;
        if (props.onValidation) {
            props.onValidation(this.state.error);
        }
    }
}
```
Thoughts (between all possible options)?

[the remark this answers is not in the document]
jeresig: Yeah, I like your proposal to set the error state in the constructor and then call the onValidation callback inside componentDidMount. I'm not sure about calling callbacks inside a constructor but that feels like it could cause problems, somehow? (Some sort of loop that is constantly re-triggering renders.) so handling it in componentDidMount feels more correct, but at least we'll get the correct error state in the text field on the initial render!

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -73,44 +88,69 @@ export default class TextField extends React.Component<Props, State> {
...
 
     handleOnFocus: (
         event: SyntheticFocusEvent<HTMLInputElement>,
     ) => mixed = () => {
+        const {onFocus} = this.props;
         this.setState({
             focused: true,
         });
+        if (onFocus) {
+            onFocus();
```
jandrade, line 136: Oops I think we missed one important part in the implementation spec.

It would be better if `onFocus` returns the `event` handled here, in case the parent component needs to do something with that event.

`onFocus(event);`

Same thing should happen with `onBlur(event)`.

Sorry that I missed that part when I originally wrote the impl. spec.

RivasCVA: It's all good. I'll made that change and update the implementation spec as well...

jandrade (approved):
LGTM!

RivasCVA:
The parent pull-request ([#1128]](https://github.com/Khan/wonder-blocks/pull/1128)) has been merged into `master`, but this branch (`wb-1053`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #1132, opened 2021-06-02
WB-1056 - Add `placeholder`, `required`, `light`, and `style` Props to TextField

RivasCVA:
## Summary:
Adding the `placeholder`, `required`, `light`, and `style` props to TextField.

When it comes to the `light` prop, I referenced off of the TextFields in Khan Academy's ["Donate"](https://www.khanacademy.org/donate) page (see images below). This was because the Figma did not specify any specific look for the `light` version of the TextField. Please let me know if there is another `light` style that is preferred.

Issue: https://khanacademy.atlassian.net/browse/WB-1056

## Test plan:
1. Open Styleguidist (or React Storybook)
2. View TextField under Form -> TextField
3. See how all example TextFields have a unique placeholder text.
4. See the "Light" TextField example and how the "default" and "error" states fit the dark background.
5. See the "Custom Style" TextField and view the ease of passing custom styles.

---

#### TextField with `light="true"` (Focus)
tf-light-focus

#### TextField with `light="true"` (Error)
tf-light-error

#### ["Donate"](https://www.khanacademy.org/donate) Page TextField (used as reference for the `light` style)
tf-donate

### DIFF
--- a/packages/wonder-blocks-form/src/components/text-field.js
+++ b/packages/wonder-blocks-form/src/components/text-field.js
@@ -6,6 +6,7 @@ import {StyleSheet, css} from "aphrodite";
 import Color from "@khanacademy/wonder-blocks-color";
 import Spacing from "@khanacademy/wonder-blocks-spacing";
 import {styles as typographyStyles} from "@khanacademy/wonder-blocks-typography";
+import type {StyleType} from "@khanacademy/wonder-blocks-core";
 
 type TextFieldType = "text" | "password" | "email" | "number" | "tel";
 
@@ -60,11 +61,32 @@ type Props = {|
      * Called when the element has been blurred.
      */
     onBlur?: (event: SyntheticFocusEvent<HTMLInputElement>) => mixed,
+
+    /**
+     * Provide hints or examples of what to enter.
+     */
+    placeholder?: string,
+
+    /**
+     * Whether this component is required.
+     */
+    required?: boolean,
+
+    /**
+     * Change the default focus ring color to fit a dark background.
+     */
+    light: boolean,
+
+    /**
+     * Custom styles for the input.
+     */
+    style?: StyleType,
 |};
 
 type DefaultProps = {|
     type: $PropertyType<Props, "type">,
     disabled: $PropertyType<Props, "disabled">,
+    light: $PropertyType<Props, "light">,
 |};
 
 type State = {|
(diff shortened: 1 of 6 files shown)


packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -198,7 +236,7 @@ const styles = StyleSheet.create({
         },
     },
     error: {
-        background: "rgba(217, 41, 22, 0.06)",
+        background: `linear-gradient(0deg, rgba(217, 41, 22, 0.06), rgba(217, 41, 22, 0.06)), ${Color.white}`,
```
jeresig, line 239: Is there any way to compute these RGBA values based on one of our existing colors? Possibly using one of the methods from the `wonder-blocks-color` package?

jeresig (approved):
This looks great, thank you!

packages/wonder-blocks-form/src/components/text-field.js
RivasCVA, line 239: I returned the `linear-gradient` because it gave the TextField a white background with the red hue. The old version would only give the red hue. I now understand why Nefaur had the linear gradient on the Figma like this.

RivasCVA: I'll try to see...

RivasCVA: I did not realize that the WB Color package had such methods. I was able to replicate the same effect using the `mix` and `fade` methods 👍.

jeresig (approved):
That's awesome! Thank you 😄

packages/wonder-blocks-form/src/components/text-field.js
RivasCVA: See comment [above](https://github.com/Khan/wonder-blocks/pull/1132#discussion_r644277785)

jandrade (approved):
LTGM! The only thing that I notice is that we might not need to use the `required` prop for now, mostly because we handle everything with `validation` and `onValidation`. But I suspect that this prop will be mostly helpful for when we implement `LabeledTextfield` to mark in the UI when the field is required.

RivasCVA:
I'll keep the `required` prop just in case we do use it in the LabeledTextField. If it's never used, then it would still be an easy change to remove.

RivasCVA:
The parent pull-request ([#1131]](https://github.com/Khan/wonder-blocks/pull/1131)) has been merged into `master`, but this branch (`wb-1056`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #1136, opened 2021-06-08
WB-919 - Enable auto-dismiss of Popover component

RivasCVA:
## Summary:
Adding the ability for the `Popover` component to be auto-dismissed when
the user clicks outside of the element. This "auto-dismisse" feature
is set to only work when `dismissEnabled=true` as requested.

Issue: https://khanacademy.atlassian.net/browse/WB-919

## Test plan:
1. Open Styleguidist in the browser.
2. Go to the "Popover" section.
3. Test clicking outside the popovers that have `dismissEnabled=true`.
   You will see the popover auto-dismiss itself.

---

#### Popover Example 1:
popover-ad-1

#### Popover Example 2:
popover-ad-2

#### Popover Example 3 - Dismissing when opening another popover (a big expectation from devs / users):
popover-ad-3


packages/wonder-blocks-popover/src/components/popover.js
```diff
@@ -118,6 +118,11 @@ type State = {|
      * Current popper placement
      */
     placement: Placement,
+
+    /**
+     * Popover Content ref
+     */
+    contentRef: RefObject<PopoverContent | PopoverContentCore>,
```
jandrade, line 125: It's better if this variable is declared as a class field declaration (or instance property).

e.g.
```js
class Foo {
    contentRef:.... = React.createRef();
  
   render() {}
}
```


```suggestion
    // somewhere at class level
    contentRef: RefObject<PopoverContent | PopoverContentCore> = React.createRef();
```

packages/wonder-blocks-popover/src/components/popover.js
```diff
@@ -156,6 +161,7 @@ export default class Popover extends React.Component<Props, State> {
         opened: !!this.props.opened,
         anchorElement: null,
         placement: this.props.placement,
+        contentRef: React.createRef(),
```
jandrade, line 164: Then you wouldn't need to to this.

BTW the main reason to do it is to avoid using state for accessing refs, as we might not need to update this internally. React should be able handle this for us (but I might be missing something 😄).

packages/wonder-blocks-popover/src/components/popover.js
```diff
@@ -188,11 +194,14 @@ export default class Popover extends React.Component<Props, State> {
...
-                  close: this.handleClose,
-              })
-            : content;
+        return React.cloneElement(
+            typeof content === "function"
+                ? content({
+                      close: this.handleClose,
+                  })
+                : content,
+            {ref: this.state.contentRef},
```
jandrade, line 203: This would need to be replaced to `this.contentRef`.

packages/wonder-blocks-popover/src/components/popover-keypress-listener.js
```diff
@@ -39,6 +65,24 @@ export default class PopoverKeypressListener extends React.Component<Props> {
         }
     };
 
+    _handleClick: (e: KeyboardEvent) => void = (e) => {
```
jandrade, line 68: This should be `MouseEvent`.

packages/wonder-blocks-popover/src/components/popover-keypress-listener.js
```diff
@@ -1,25 +1,51 @@
...
+    Props,
+    State,
+> {
+    state: State = {
+        isFirstClick: true,
+    };
+
     componentDidMount() {
         window.addEventListener("keyup", this._handleKeyup);
+        window.addEventListener("click", this._handleClick);
```
jandrade, line 43: It would be good to add a unit test verifying that `onClose` is called if we press `click` outside the popover ref.

packages/wonder-blocks-popover/src/components/popover-keypress-listener.js
```diff
@@ -1,25 +1,51 @@
...
+     */
+    isFirstClick: boolean,
 |};
 
 /**
  * A component that, when mounted, calls `onClose` when Escape is pressed.
  * @see @khanacademy/wonder-blocks-modal/components/modal-launcher.js
  */
-export default class PopoverKeypressListener extends React.Component<Props> {
+export default class PopoverKeypressListener extends React.Component<
```
jandrade, line 33: I'm thinking that this whole file should be renamed to `PopoverEventListener` or something like that because now we handle `keyboard` and `mouse` events. Probably the description above should be updated as well to reflect these changes.

packages/wonder-blocks-popover/src/components/popover.js
```diff
@@ -188,11 +194,14 @@ export default class Popover extends React.Component<Props, State> {
     renderContent(): PopoverContents {
         const {content} = this.props;
 
-        return typeof content === "function"
-            ? content({
-                  close: this.handleClose,
-              })
-            : content;
+        return React.cloneElement(
```
jandrade, line 197: Nice way to get the `ref` to the Popover content instance :)

jandrade (changes requested):
Things are looking good. The core change makes a lot of sense, but I'm requesting a couple of changes before approving:

- Moving `contentRef` to the class level (instead of `state`).
- Adding a unit test to cover the `window.onClick` case.
- Renaming the file and class to (PopoverEventListener).

Thanks for adding all the examples in the PR description, it's super helpful 💯.

packages/wonder-blocks-popover/src/components/popover.js
RivasCVA: Got it! ✅

packages/wonder-blocks-popover/src/components/popover-keypress-listener.js
RivasCVA: Ohhh. Thanks for catching that!

packages/wonder-blocks-popover/src/components/popover-keypress-listener.js
RivasCVA: I'll rename the file to "popover-event-listener.js" and make the component be called `PopoverEventListener`.

jandrade (approved):
Looks great to me! thanks for updating the PR and adding the test.

PR #1140, opened 2021-06-10
WB-792 - Fix Modal header subtitle out of sync with spec

RivasCVA:
## Summary:
Fix the `ModalHeader`'s subtitle label that had the Typography style of
`LabelMedium` while the [design spec](https://www.figma.com/file/CRryTHqyr1l6aOOltD32KMh5/Modal-v2?node-id=412%3A195) requested `LabelSmall`.

Issue: https://khanacademy.atlassian.net/browse/WB-792

## Test plan:
1. Open Styleguidist on the browser.
2. Go to the Modal section, under "Modals".
3. See the "OnePaneDialog with above container" example.
4. (Optional) You can also inspect-element and view that `LabelSmall` is used

---

##### Using inspect-element demonstrates that `LabelSmall` is now being used:
modal-test


jandrade (approved):
Looks good to me, thanks!

PR #1141, opened 2021-06-10
WB-1070 - Fix ModalHeader not setting testId properly (OnePanelDialog)

RivasCVA:
## Summary:
Fixed `ModalHeader` not being passed a testId when used inside
`OnePaneDialog`. It was previously working in the case that `ModalHeader`
had a subtitle, but not in other cases. That was fixed.

Issue: https://khanacademy.atlassian.net/browse/WB-1070

## Test plan:
1. You can view the code and see that testId is now being passed in all cases.
2. You can also open Styleguidist and check that testId is being passed
   on all cases of `ModalHeader` when used in a `OnePaneDialog`. This
   can be see through "inspect element".

---

##### * You can see that the "testId" is indeed passed on the `ModalHeader` (even when there is no subtitle).

modal-test-id

### DIFF
--- a/packages/wonder-blocks-modal/src/__tests__/__snapshots__/generated-snapshot.test.js.snap
+++ b/packages/wonder-blocks-modal/src/__tests__/__snapshots__/generated-snapshot.test.js.snap
@@ -1179,7 +1179,7 @@ exports[`wonder-blocks-modal example 7 1`] = `
           </button>
           <div
             className=""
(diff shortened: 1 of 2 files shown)


jandrade (approved):
LGTM! One minor request: Would it be possible to rename the suffix from `ModalHeader` to just `header`?

I see that shortening that would be better for when we write tests using this `testId` :)

Sorry I missed that in the original implementation.

PR #1144, opened 2021-06-14
WB-1064 - Add LabeledTextField to WB (file setup)

RivasCVA:
## Summary:
Added `LabeledTextField` to Wonder Blocks under `wonder-blocks-form`.
Created all files that will be used and added basic skeleton
examples in Styleguidist and React Storybook.

**Note: This PR is just a small stop before beginning to add states and props.**

Issue: https://khanacademy.atlassian.net/browse/WB-1064

## Test plan:
1. View all the new files added
2. View the generated Styleguidist and React Storybook links (under Form / LabeledTextField)

---

#### The Basic State-less, Prop-less Look:
labeled-text-field-ex-1


jandrade (approved):
LGTM!

packages/wonder-blocks-form/src/__tests__/__snapshots__/generated-snapshot.test.js.snap
```diff
@@ -6184,3 +6184,193 @@ exports[`wonder-blocks-form example 18 1`] = `
...
+        "fontSize": 14,
+        "fontWeight": 400,
+        "lineHeight": "18px",
+      }
+    }
+  >
+    Error
+  </span>
+</div>
+`;
```
somewhatabstract, line 6376: Snip snappity snapshot snappery.
😏📸

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+import * as React from "react";
+
+import FieldHeading from "./field-heading.js";
+import TextField from "./text-field.js";
+
+export default class LabeledTextField extends React.Component<{||}> {
```
somewhatabstract, line 7: NIT: Since we almost always will end up adding a prop some day, even when props are empty could you define the type:

`type Props = {||};`

It keeps things consistent which helps folks read the code and it simplifies maintenance when we have to start adding props.

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -0,0 +1,25 @@
...
+                        onChange={() => {}}
+                    />
+                }
+                label="Label"
+                description="Description"
+                error="Error"
+            />
+        );
+    }
+}
```
somewhatabstract, line 25: This component could be a functional stateless component. Would be less indentation - just an option. No change required.

```js
export default (props: Props) => (
    <FieldHeading... etc. etc.
);
```

somewhatabstract (approved):
Looks good. Couple of inline thoughts. :)

PR #1145, opened 2021-06-15
WB-1065 - Add States to LabeledTextField

RivasCVA:
## Summary:
Added the `error` and `focused` state to the `LabeledTextField`. The `enabled` state was added in the form of a `disabled` prop where it is set to `false` by default.

Included unit tests to cover these new states and prop (except the `error` state because it will be covered once validation is added).

Issue: https://khanacademy.atlassian.net/browse/WB-1065

## Test plan:
1. View the code to see the new states and prop
2. View Styleguidist and / or React Storybook examples under Form / LabeledTextField (see generated links)

---

#### The `disabled` Prop in Action:
labeled-text-field-disabled

### DIFF
--- a/packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
+++ b/packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
@@ -1,0 +1,50 @@
+//@flow
+import * as React from "react";
+import {mount} from "enzyme";
+
+import LabeledTextField from "../labeled-text-field.js";
+
+const wait = (delay: number = 0) =>
+    new Promise((resolve, reject) => {
+        // eslint-disable-next-line no-restricted-syntax
+        return setTimeout(resolve, delay);
+    });
+
+describe("LabeledTextField", () => {
+    it("labeledtextfield becomes focused", () => {
+        // Arrange
+        const wrapper = mount(<LabeledTextField />);
+        const field = wrapper.find("TextField");
+
+        // Act
+        field.simulate("focus");
+
+        // Assert
+        expect(wrapper).toHaveState("focused", true);
+    });
+
+    it("labeledtextfield becomes blurred", async () => {
+        // Arrange
+        const wrapper = mount(<LabeledTextField />);
+        const field = wrapper.find("TextField");
+
+        // Act
+        field.simulate("focus");
+        await wait(0);
+        field.simulate("blur");
+
+        // Assert
+        expect(wrapper).toHaveState("focused", false);
+    });
+
+    it("disabled prop disables the input", async () => {
+        // Arrange
+
+        // Act
+        const wrapper = mount(<LabeledTextField disabled={true} />);
+
+        // Assert
+        const input = wrapper.find("input");
+        expect(input).toBeDisabled();
+    });
+});
--- a/packages/wonder-blocks-form/src/components/labeled-text-field.js
+++ b/packages/wonder-blocks-form/src/components/labeled-text-field.js
@@ -4,8 +4,57 @@ import * as React from "react";
 import FieldHeading from "./field-heading.js";
 import TextField from "./text-field.js";
 
-export default class LabeledTextField extends React.Component<{||}> {
+type Props = {|
+    /**
+     * Makes a read-only input field that cannot be focused. Defaults to false.
+     */
+    disabled: boolean,
+|};
+
+type DefaultProps = {|
+    disabled: $PropertyType<Props, "disabled">,
+|};
+
+type State = {|
+    /**
+     * Displayed when the validation fails.
+     */
+    error: ?string,
+
+    /**
+     * The user focuses on the textfield.
+     */
+    focused: boolean,
+|};
+
+export default class LabeledTextField extends React.Component<Props, State> {
+    static defaultProps: DefaultProps = {
+        disabled: false,
+    };
+
+    state: State = {
+        error: null,
+        focused: false,
+    };
+
+    handleOnFocus: (
+        event: SyntheticFocusEvent<HTMLInputElement>,
+    ) => mixed = () => {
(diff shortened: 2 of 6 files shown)


packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
```diff
@@ -0,0 +1,50 @@
...
+        // Arrange
+        const wrapper = mount(<LabeledTextField />);
+        const field = wrapper.find("TextField");
+
+        // Act
+        field.simulate("focus");
+
+        // Assert
+        expect(wrapper).toHaveState("focused", true);
+    });
```
kevinbarabash, line 24: It might be worth adding a snapshot test for this state.

packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
```diff
@@ -0,0 +1,50 @@
...
+    it("disabled prop disables the input", async () => {
+        // Arrange
+
+        // Act
+        const wrapper = mount(<LabeledTextField disabled={true} />);
+
+        // Assert
+        const input = wrapper.find("input");
+        expect(input).toBeDisabled();
+    });
```
kevinbarabash, line 49: Nice!

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -4,16 +4,68 @@ import * as React from "react";
...
+            focused: true,
+        });
+    };
+
+    handleOnBlur: (
+        event: SyntheticFocusEvent<HTMLInputElement>,
+    ) => mixed = () => {
+        this.setState({
+            focused: false,
+        });
```
kevinbarabash, line 53: Where is `this.state.focused` used?

kevinbarabash (approved):
LGTM.  I've suggested one additional test you may want to write before merging this.

packages/wonder-blocks-form/src/components/labeled-text-field.js
RivasCVA: It will be used later when I add validation. I will only be showing the error message (if any) when the user is **not** focused on the textfield.

packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
RivasCVA: How do you suggest I add this snapshot? Would it be in the `custom-snapshot.test.js` file? If so, how would I change the `focused` state since it is **not** a prop state?

kevinbarabash: I was going to suggest doing something similar to what's in https://github.com/Khan/wonder-blocks/blob/75334ca695dc97adcc0c557c2530b3dc95af8925/packages/wonder-blocks-core/src/components/__tests__/view.test.js, but I see now that none of those tests are simulating events.  I think you might be to trigger an event on the object returned by `renderer.create(node)` but since this isn't something we already please ignore.  A lot of our other components have a stateless (*-core.js) and stateful halves so testing the markup of the focus state can be done by taking a snapshot of the stateless component with the right props.  Please ignore my suggestion for now.  In the future this kind of testing might get easy if we adopt React Testing Library.

RivasCVA: Alright. Thanks for the info! It seems React Testing Library will be giving us so much more flexibility in many domains, can't wait to [hopefully] see it soon.

packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
```diff
@@ -0,0 +1,50 @@
...
+import * as React from "react";
+import {mount} from "enzyme";
+
+import LabeledTextField from "../labeled-text-field.js";
+
+const wait = (delay: number = 0) =>
+    new Promise((resolve, reject) => {
+        // eslint-disable-next-line no-restricted-syntax
+        return setTimeout(resolve, delay);
+    });
```
somewhatabstract, line 11: No change required.

I think we should copy over my new `wait` implementation from webapp into Wonder Blocks and then reuse it across our tests rather than defining this in every test.

https://github.com/Khan/webapp/blob/master/services/static/javascript/testing/wait.js

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -4,16 +4,68 @@ import * as React from "react";
...
-export default class LabeledTextField extends React.Component<{||}> {
+type Props = {|
+    /**
+     * Makes a read-only input field that cannot be focused. Defaults to false.
+     */
+    disabled: boolean,
+|};
+
+type DefaultProps = {|
+    disabled: $PropertyType<Props, "disabled">,
```
somewhatabstract, line 15: Yes! Thanks for following this pattern.

somewhatabstract (approved):
LGTM

One inline suggestion but it's broader than your changes. Just something for us to consider as a possible update to our WB testing infrastructure.

packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
RivasCVA: I could possibly partake in this (if the team's okay with this new change). Under what directory would this `"wait.js"` file go within Wonder Blocks?

RivasCVA:
The parent pull-request ([#1144]](https://github.com/Khan/wonder-blocks/pull/1144)) has been merged into `master`, but this branch (`wb-1065`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #1146, opened 2021-06-15
WB-1066 - Add the first set of Props to LabeledTextField

RivasCVA:
## Summary:
Adding the first set of props to `LabeledTextField`.
This includes:
* id
* type
* label
* description
* initialValue

I added unit tests to cover these new props (current Coverage is 100%).

Issue: https://khanacademy.atlassian.net/browse/WB-1066

## Test plan:
1. Open Styleguidist or React Storybook in browser (see generated links).
2. View the new props in action.
Note: I will be adding examples that use different TextField `type`s when I add validation

### DIFF
--- a/packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
+++ b/packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
@@ -13,7 +13,7 @@ const wait = (delay: number = 0) =>
 describe("LabeledTextField", () => {
     it("labeledtextfield becomes focused", () => {
         // Arrange
-        const wrapper = mount(<LabeledTextField />);
+        const wrapper = mount(<LabeledTextField label="Label" />);
         const field = wrapper.find("TextField");
 
         // Act
@@ -25,7 +25,7 @@ describe("LabeledTextField", () => {
 
     it("labeledtextfield becomes blurred", async () => {
         // Arrange
-        const wrapper = mount(<LabeledTextField />);
+        const wrapper = mount(<LabeledTextField label="Label" />);
         const field = wrapper.find("TextField");
 
         // Act
@@ -37,11 +37,103 @@ describe("LabeledTextField", () => {
         expect(wrapper).toHaveState("focused", false);
     });
 
-    it("disabled prop disables the input", async () => {
+    it("value state changes when the user types", () => {
+        // Arrange
+        const wrapper = mount(<LabeledTextField label="Label" />);
+        const input = wrapper.find("input");
+
+        // Act
+        const newValue = "New Value";
+        input.simulate("change", {target: {value: newValue}});
+
+        // Assert
+        expect(wrapper).toHaveState("value", newValue);
+    });
+
+    it("id prop is passed to input", () => {
+        // Arrange
+        const id = "exampleid";
+
+        // Act
+        const wrapper = mount(
+            <LabeledTextField id={id} label="Label" disabled={true} />,
+        );
+
+        // Assert
+        const input = wrapper.find("input");
+        expect(input).toContainMatchingElement(`[id="${id}-field"]`);
+    });
+
+    it("auto-generated id is passed to input when id prop is not set", () => {
         // Arrange
 
         // Act
-        const wrapper = mount(<LabeledTextField disabled={true} />);
+        const wrapper = mount(<LabeledTextField label="Label" />);
+
+        // Assert
+        // Since the generated id is unique, we cannot know what it will be
+        // so we only test if the id attribute is set.
+        const input = wrapper.find("input");
+        expect(input).toContainMatchingElement("[id]");
+    });
+
+    it("type prop is passed to input", () => {
+        // Arrange
+        const type = "email";
+
+        // Act
+        const wrapper = mount(<LabeledTextField type={type} label="Label" />);
+
+        // Assert
+        const input = wrapper.find("input");
+        expect(input).toContainMatchingElement(`[type="${type}"]`);
+    });
+
+    it("label prop is rendered", () => {
+        // Arrange
+        const label = "Label";
+
+        // Act
+        const wrapper = mount(<LabeledTextField label={label} />);
+
+        // Assert
+        expect(wrapper).toIncludeText(label);
+    });
+
+    it("description prop is rendered", () => {
+        // Arrange
+        const description = "Description";
+
+        // Act
+        const wrapper = mount(
+            <LabeledTextField label="Label" description={description} />,
+        );
+
+        // Assert
+        expect(wrapper).toIncludeText(description);
+    });
+
+    it("initialValue prop is set on mount", () => {
+        // Arrange
+        const initialValue = "Value";
+
+        // Act
+        const wrapper = mount(
+            <LabeledTextField initialValue={initialValue} label="Label" />,
+        );
+
+        // Assert
+        const input = wrapper.find("input");
+        expect(input).toContainMatchingElement(`[value="${initialValue}"]`);
+    });
+
+    it("disabled prop disables the input", () => {
+        // Arrange
+
+        // Act
+        const wrapper = mount(
+            <LabeledTextField label="Label" disabled={true} />,
+        );
 
         // Assert
         const input = wrapper.find("input");
--- a/packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
+++ b/packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
@@ -71,10 +71,10 @@ describe("LabeledTextField", () => {
         const wrapper = mount(<LabeledTextField label="Label" />);
 
         // Assert
-        // Since the generated id is unique, we cannot know what it will be
-        // so we only test if the id attribute is set.
+        // Since the generated id is unique, we cannot know what it will be.
+        // We only test if the id attribute starts with "uid-" and ends with "-field".
         const input = wrapper.find("input");
-        expect(input).toContainMatchingElement("[id]");
+        expect(input.props()["id"]).toMatch(/uid-.*-field/);
     });
 
     it("type prop is passed to input", () => {
@@ -124,7 +124,7 @@ describe("LabeledTextField", () => {
 
         // Assert
         const input = wrapper.find("input");
-        expect(input).toContainMatchingElement(`[value="${initialValue}"]`);
+        expect(input).toHaveValue(initialValue);
     });
 
     it("disabled prop disables the input", () => {
--- a/packages/wonder-blocks-form/src/components/labeled-text-field.js
+++ b/packages/wonder-blocks-form/src/components/labeled-text-field.js
@@ -70,16 +70,12 @@ export default class LabeledTextField extends React.Component<Props, State> {
 
     constructor(props: Props) {
         super(props);
-        if (props.initialValue) {
-            this.state.value = props.initialValue;
-        }
+        this.state = {
+            value: props.initialValue ? props.initialValue : "",
+            error: null,
+            focused: false,
+        };
     }
-
-    state: State = {
-        value: "",
-        error: null,
-        focused: false,
-    };
 
     handleOnFocus: (
         event: SyntheticFocusEvent<HTMLInputElement>,
@@ -107,7 +103,7 @@ export default class LabeledTextField extends React.Component<Props, State> {
         const {id, type, label, description, disabled} = this.props;
 
         return (
-            <IDProvider id={id} scope="field">
+            <IDProvider id={id} scope="labeled-text-field">
                 {(uniqueId) => (
                     <FieldHeading
                         id={uniqueId}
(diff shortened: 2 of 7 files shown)


packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -29,10 +64,19 @@ type State = {|
...
+        if (props.initialValue) {
+            this.state.value = props.initialValue;
+        }
+    }
+
     state: State = {
+        value: "",
         error: null,
         focused: false,
     };
```
RivasCVA, line 82: Is it good to have it like this? Or would removing the "constructor" and changing the state class property to something like below be better?
```js
state: State = {
    value: this.state.value ? this.state.value : "",
    error: null,
    focused: false,
};
```

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -1,21 +1,56 @@
...
 import FieldHeading from "./field-heading.js";
-import TextField from "./text-field.js";
+import TextField, {type TextFieldType} from "./text-field.js";
 
 type Props = {|
+    /**
+     * An optional unique identifier for the TextField.
+     * If no id is specified, a unique id will be auto-generated.
+     */
+    id?: string,
```
RivasCVA, line 15: This is another change from the implementation spec (will update it). I have now made `id` optional since `IDProvider` is used to generated unique ids.

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -1,21 +1,56 @@
...
 type DefaultProps = {|
+    type: $PropertyType<Props, "type">,
     disabled: $PropertyType<Props, "disabled">,
 |};
 
 type State = {|
+    /**
+     * The value of the input.
+     */
+    value: string,
```
RivasCVA, line 52: This is another area that I forgot to add in the implementation spec. I have added `value` as a state (will update implementation spec).

packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
```diff
@@ -37,11 +37,103 @@ describe("LabeledTextField", () => {
...
+        const id = "exampleid";
+
+        // Act
+        const wrapper = mount(
+            <LabeledTextField id={id} label="Label" disabled={true} />,
+        );
+
+        // Assert
+        const input = wrapper.find("input");
+        expect(input).toContainMatchingElement(`[id="${id}-field"]`);
```
kevinbarabash, line 64: Does `.toHaveProp()` not work here?

packages/wonder-blocks-form/src/components/labeled-text-field.js
kevinbarabash: Are there plans to add a `testId` prop eventually as well?

packages/wonder-blocks-form/src/components/labeled-text-field.js
kevinbarabash: Personally I prefer doing `this.state = { ... }` in the constructor.  I find it easier to reason about especially since not everyone is up-to-date on all the new-fangled JavaScript features.

jandrade: I think it's better if you centralize that in a single place, and also think that it would be better moving everything to the `constructor`.

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -53,25 +97,42 @@ export default class LabeledTextField extends React.Component<Props, State> {
...
-                field={
-                    <TextField
-                        id="tf-1"
-                        value="Value"
-                        placeholder="Placeholder"
-                        disabled={disabled}
-                        onChange={() => {}}
-                        onFocus={this.handleOnFocus}
-                        onBlur={this.handleOnBlur}
+            <IDProvider id={id} scope="field">
```
jandrade, line 110: If I remember correctly, the scope will be prefixed as part of the generated unique id, right?

packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
```diff
@@ -37,11 +37,103 @@ describe("LabeledTextField", () => {
...
+        // Arrange
+
+        // Act
+        const wrapper = mount(<LabeledTextField label="Label" />);
+
+        // Assert
+        // Since the generated id is unique, we cannot know what it will be
+        // so we only test if the id attribute is set.
+        const input = wrapper.find("input");
+        expect(input).toContainMatchingElement("[id]");
```
jandrade, line 77: I think you can also try matching a given pattern. Something like this could probably work.

```suggestion
        expect(input).toMatch(/field-field/);
```

packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
```diff
@@ -37,11 +37,103 @@ describe("LabeledTextField", () => {
...
+        const initialValue = "Value";
+
+        // Act
+        const wrapper = mount(
+            <LabeledTextField initialValue={initialValue} label="Label" />,
+        );
+
+        // Assert
+        const input = wrapper.find("input");
+        expect(input).toContainMatchingElement(`[value="${initialValue}"]`);
```
jandrade, line 127: Nit: I think you can also use `expect(input).toHaveValue(initialValue)`.

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -53,25 +97,42 @@ export default class LabeledTextField extends React.Component<Props, State> {
...
+                                placeholder="Placeholder"
+                                disabled={disabled}
+                                onChange={this.handleOnChange}
+                                onFocus={this.handleOnFocus}
+                                onBlur={this.handleOnBlur}
+                            />
+                        }
+                        label={label}
+                        description={description}
+                        error="Error"
```
jandrade, line 132: Is the plan to add the error `prop` as part of https://khanacademy.atlassian.net/browse/WB-1068?

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -8,7 +8,7 @@ import Spacing from "@khanacademy/wonder-blocks-spacing";
 import {styles as typographyStyles} from "@khanacademy/wonder-blocks-typography";
 import type {StyleType, AriaProps} from "@khanacademy/wonder-blocks-core";
 
-type TextFieldType = "text" | "password" | "email" | "number" | "tel";
+export type TextFieldType = "text" | "password" | "email" | "number" | "tel";
```
jandrade, line 11: Nice!

jandrade (approved):
Looks good to me! Just added a couple of comments, but nothing blocking. Great stuff!

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -53,25 +97,42 @@ export default class LabeledTextField extends React.Component<Props, State> {
...
+                                aria-invalid={
+                                    this.state.error ? "true" : "false"
+                                }
+                                type={type}
+                                value={this.state.value}
+                                placeholder="Placeholder"
+                                disabled={disabled}
+                                onChange={this.handleOnChange}
+                                onFocus={this.handleOnFocus}
+                                onBlur={this.handleOnBlur}
```
kevinbarabash, line 127: style nit: usually these are called `handleChange`, `handleBlur`, etc.

packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
```diff
@@ -37,11 +37,103 @@ describe("LabeledTextField", () => {
...
+        // Arrange
+        const wrapper = mount(<LabeledTextField label="Label" />);
+        const input = wrapper.find("input");
+
+        // Act
+        const newValue = "New Value";
+        input.simulate("change", {target: {value: newValue}});
+
+        // Assert
+        expect(wrapper).toHaveState("value", newValue);
```
kevinbarabash, line 50: It might be good to have a test that checks that the value of the `input` is also `"value"`.

kevinbarabash (approved):
LGTM.  A few nits in the comments, but nothing blocking.

packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
kevinbarabash: Nevermind, I see there already is one.

packages/wonder-blocks-form/src/components/labeled-text-field.js
RivasCVA: Yes. That is in the last [ticket](https://khanacademy.atlassian.net/browse/WB-1069) for the [epic](https://khanacademy.atlassian.net/browse/WB-1055).

packages/wonder-blocks-form/src/components/labeled-text-field.js
RivasCVA: I don't necessarily plan to add `error` as a prop, but keep it as a state. The `error` state will be used to pass the error message `onValidation` returns to `FieldHeading`'s error label. I will indeed be making use of this on that next ticket.

packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
RivasCVA: I can't seem to get this to work. Does `toMatch(...)` match against the attributes in the HTML element, or against something else?

packages/wonder-blocks-form/src/components/labeled-text-field.js
RivasCVA: The scope is added in the middle of the unique id:

idprovider-example

packages/wonder-blocks-form/src/components/labeled-text-field.js
RivasCVA: Just to be sure... you are referring to the handle function, right? Not the prop.

So I would just remove the "On" in the function name.
```js
this.handleOnChange
```
to
```js
this.handleChange
```

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -29,10 +64,19 @@ type State = {|
...
     static defaultProps: DefaultProps = {
+        type: "text",
         disabled: false,
     };
 
+    constructor(props: Props) {
+        super(props);
+        if (props.initialValue) {
+            this.state.value = props.initialValue;
+        }
```
somewhatabstract, line 75: Would it matter if this were just `this.state.value = props.initialValue` without the check? Seems like the effect would be the same.

If it needs to retain a default, then what about.

`this.state.value = props.initialValue || "";`

somewhatabstract (approved):
LGTM

packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
RivasCVA: Never mind, I got it working doing something like:
```js
expect(input.props()["id"]).toMatch(/uid-.*-field/);
```

packages/wonder-blocks-form/src/components/labeled-text-field.js
RivasCVA: I also have now changed the scope to be `"labeled-text-field"` to be more descriptive.

packages/wonder-blocks-form/src/components/labeled-text-field.js
RivasCVA: Good suggestion! I could also just make initialValue a default prop and set it to `""` by default.

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -70,17 +70,13 @@ export default class LabeledTextField extends React.Component<Props, State> {
...
     constructor(props: Props) {
         super(props);
-        if (props.initialValue) {
-            this.state.value = props.initialValue;
-        }
+        this.state = {
+            value: props.initialValue ? props.initialValue : "",
+            error: null,
+            focused: false,
+        };
```
somewhatabstract, line 77: Nice :)

packages/wonder-blocks-form/src/components/labeled-text-field.js
jandrade: Oh good point! I totally forgot about that part 😅. That makes total sense.

RivasCVA:
The parent pull-request ([#1145]](https://github.com/Khan/wonder-blocks/pull/1145)) has been merged into `master`, but this branch (`wb-1066`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #1148, opened 2021-06-16
WB-1068 - Add Validation and Event Props to LabeledTextField

RivasCVA:
## Summary:
Added all props dealing with Validation and Event Handling to `LabeledTextField`.

The following were the props added:
* validation
* onValidation
* onChange
* onKeyDown
* onFocus
* onBlur

Added unit tests to ensure the new props work as expected.

Issue: https://khanacademy.atlassian.net/browse/WB-1068

## Test plan:
1. Open Styleguidist or React Storybook (see generated links).
2. View the `LabeledTextField` examples under "Form / LabeledTextField".
3. View at all the new examples that show validation and events in action.
4. View the examples that provide error message on validation errors.
5. Test all examples by pressing "Enter" and notice how the TextField blurs.

---

#### Normal
labeled-text-field-normal

#### Validated
labeled-text-field-validated

### DIFF
--- a/packages/wonder-blocks-form/src/components/labeled-text-field.js
+++ b/packages/wonder-blocks-form/src/components/labeled-text-field.js
@@ -38,6 +38,37 @@ type Props = {|
      * Makes a read-only input field that cannot be focused. Defaults to false.
      */
     disabled: boolean,
+
+    /**
+     * Provide a validation for the input value.
+     * Return a string error message or null | void for a valid input.
+     */
+    validation?: (value: string) => ?string,
+
+    /**
+     * Called when the TextField input is validated.
+     */
+    onValidation?: (errorMessage: ?string) => mixed,
+
+    /**
+     * Called when the value has changed.
+     */
+    onChange?: (newValue: string) => mixed,
+
+    /**
+     * Called when a key is pressed.
+     */
+    onKeyDown?: (event: SyntheticKeyboardEvent<HTMLInputElement>) => mixed,
+
+    /**
+     * Called when the element has been focused.
+     */
+    onFocus?: (event: SyntheticFocusEvent<HTMLInputElement>) => mixed,
+
+    /**
+     * Called when the element has been blurred.
+     */
+    onBlur?: (event: SyntheticFocusEvent<HTMLInputElement>) => mixed,
 |};
 
 type DefaultProps = {|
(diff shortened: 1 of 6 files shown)


packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -38,6 +38,37 @@ type Props = {|
      * Makes a read-only input field that cannot be focused. Defaults to false.
      */
     disabled: boolean,
+
+    /**
+     * Provide a validation for the input value.
+     * Return a string error message or null | void for a valid input.
+     */
+    validation?: (value: string) => ?string,
```
somewhatabstract, line 46: I know this is inherited from `TextField`, so I guess this is a suggestion for that component too. Since this is an action, perhaps it should be renamed to `validate` or `onValidate` and then change `onValidation` to `onValidated`. I think that would be clearer for folks using this API.

What do folks think?

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -71,36 +102,64 @@ export default class LabeledTextField extends React.Component<Props, State> {
...
 
-    handleOnFocus: (
-        event: SyntheticFocusEvent<HTMLInputElement>,
-    ) => mixed = () => {
+    handleValidation: (errorMessage: ?string) => mixed = (errorMessage) => {
+        const {onValidation} = this.props;
+        this.setState({error: errorMessage});
+        if (onValidation) {
+            onValidation(errorMessage);
+        }
```
somewhatabstract, line 116: Should this be invoked once the state has been applied (as in using the `setState` callback)?

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -71,36 +102,64 @@ export default class LabeledTextField extends React.Component<Props, State> {
...
+
+    handleChange: (newValue: string) => mixed = (newValue) => {
+        const {onChange} = this.props;
         this.setState({
-            focused: true,
+            value: newValue,
         });
+        if (onChange) {
+            onChange(newValue);
+        }
```
somewhatabstract, line 126: Should this be invoked once the state has been applied (as in using the `setState` callback)?

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -71,36 +102,64 @@ export default class LabeledTextField extends React.Component<Props, State> {
...
+        event,
+    ) => {
+        const {onFocus} = this.props;
         this.setState({
-            focused: false,
+            focused: true,
         });
+        if (onFocus) {
+            onFocus(event);
+        }
```
somewhatabstract, line 138: Should this be invoked once the state has been applied (as in using the `setState` callback)?

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -71,36 +102,64 @@ export default class LabeledTextField extends React.Component<Props, State> {
...
+        event,
+    ) => {
+        const {onBlur} = this.props;
         this.setState({
-            value: newValue,
+            focused: false,
         });
+        if (onBlur) {
+            onBlur(event);
+        }
```
somewhatabstract, line 150: Should this be invoked once the state has been applied (as in using the `setState` callback)?

somewhatabstract (approved):
Some questions inline. Nothing blocking, I don't think.

packages/wonder-blocks-form/src/components/labeled-text-field.js
RivasCVA: Hmm 🤔 . I wasn't aware `setState(...)` had a callback function. I guess I would also need to update the same for `TextField` because I called all `onEvent(...)` props in the same manner.

packages/wonder-blocks-form/src/components/labeled-text-field.js
RivasCVA: This makes senes, `"validate"` is more of an action word (aka a verb). Although, a lot of the TextFields built in Webapp use the name `"onValidation"` as well. I'm not sure if changing the name would complicate things when bringing WB `TextField`s to Webapp.

What do other people think?

jandrade: I think changing to `validate` makes sense. On the other hand, I think that changing to `onValidate` would make more sense instead of `onValidation`. In understand the meaning of changing to `onValidated`, but I think that `onValidate` adheres better to the way these event handlers are named in React (e.g. `onChange`, `onClick`, etc).

I think it's ok to rename all the occurrences in `webapp`, because we are going to also rename from `validator` to `validate` (or the prop name that we'll end up using).

Any thoughts?

RivasCVA: That seems fair enough. I will be changing `validation` --> `validate` and `onValidation` --> `onValidate`. I also updated a TextField ticket to add those changes later.

RivasCVA:
The parent pull-request ([#1146]](https://github.com/Khan/wonder-blocks/pull/1146)) has been merged into `master`, but this branch (`wb-1068`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #1153, opened 2021-06-29
WB-1080 - Make TextField and LabeledTextField forward ref to input

RivasCVA:
## Summary:
Added the ability for `TextField` and `LabeledTextField`'s `ref` prop to
forward into the `<input/>` element. This would simplify accessing the
`input` to focus or blur under certain circumstances.

**Note:** Because Styleguidist uses the name of the default export and we were exporting the `React.forwardRef` instance, we had to rename the classes of both `LabeledTextField` and `TextField` in order for Styleguidist to display the correct name. Ideally when Styleguidist is gone and fully replaced with React Storybook, we would like to have both class names reverted back to their original name (See [WB-1081](https://khanacademy.atlassian.net/browse/WB-1081)).

Issue: https://khanacademy.atlassian.net/browse/WB-1080

## Test plan:
1. View the new "Ref" example in Styleguidist and React Storybook under Form / TextField
2. View the new "Ref" example in Styleguidist and React Storybook under Form / LabeledTextField

---

#### `ref` Demo on `TextField`
text-field-ref-demo

#### `ref` Demo on `LabeledTextField`
labeled-text-field-ref-demo


jeresig (approved):
Looks good to me - thank you for adding the tests + stories!

packages/wonder-blocks-form/src/components/text-field.md
```diff
@@ -614,3 +614,71 @@ const styles = StyleSheet.create({
 
 <TextFieldExample />
 ```
+
```
jandrade, line 617: Thanks for adding this example. It's super helpful if folks want to try that out!

jandrade (approved):
LGTM!

PR #1154, opened 2021-06-30
Properly support SSR request fulfillment without hydration support

somewhatabstract:
## Summary:
This fixes our SSR support for non-hydrateable handlers to still support
request fulfillment during SSR.

Issue: FEI-3597

## Test plan:
`yarn test`

I'll also be incorporating this version of WB Data into webapp to verify it
all works with the intended use-case; the new GqlQuery component.

### DIFF
--- a/packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
+++ b/packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
@@ -354,7 +354,7 @@ describe("../response-cache.js", () => {
         });
 
         describe("no hydrate", () => {
-            it("should store the entry in the NoCache cache", () => {
+            it("should store the entry in the ssr-only cache", () => {
                 // Arrange
                 jest.spyOn(Server, "isServerSide").mockReturnValue(true);
                 const ssrOnlyCache = new MemoryCache();
--- a/packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
+++ b/packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
@@ -2,7 +2,6 @@
 import {Server} from "@khanacademy/wonder-blocks-core";
 import {ResponseCache} from "../response-cache.js";
 import MemoryCache from "../memory-cache.js";
-import NoCache from "../no-cache.js";
 
 import type {IRequestHandler} from "../types.js";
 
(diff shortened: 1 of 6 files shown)


packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
```diff
@@ -359,9 +357,10 @@ describe("../response-cache.js", () => {
             it("should store the entry in the NoCache cache", () => {
```
somewhatabstract, line 357: TODO: Rename the test case

packages/wonder-blocks-data/src/util/response-cache.js
```diff
@@ -43,25 +31,49 @@ export class ResponseCache {
...
+    _ssrOnlyCache: ?MemoryCache<any, any>;
 
-    constructor(memoryCache: ?MemoryCache<any, any> = null) {
-        this._cache = memoryCache || new MemoryCache();
+    constructor(
+        memoryCache: ?MemoryCache<any, any> = null,
+        ssrOnlyCache: ?MemoryCache<any, any> = null,
+    ) {
+        this._ssrOnlyCache = Server.isServerSide()
+            ? ssrOnlyCache || new MemoryCache()
```
somewhatabstract, line 42: Only want to support this if we are in SSR mode. That includes during testing since that way, we will properly fail still if the test doesn't also mock out Server.isServerSide as it should.

packages/wonder-blocks-data/src/util/response-cache.js
```diff
@@ -204,21 +233,36 @@ export class ResponseCache {
...
 
     /**
-     * Deep clone the cache.
+     * Deep clone the hydration cache.
      *
      * By design, this does not clone anything held in custom caches.
      */
-    cloneCachedData: () => $ReadOnly<Cache> = (): $ReadOnly<Cache> => {
-        return this._cache.cloneData();
+    cloneHydratableData: () => $ReadOnly<Cache> = (): $ReadOnly<Cache> => {
```
somewhatabstract, line 264: Renamed this to make it clearer what it is for.

packages/wonder-blocks-data/src/util/types.js
```diff
@@ -15,12 +15,12 @@ export type Result<TData: ValidData> =
...
-          error: string,
-          data?: ?void,
+          +error: string,
+          +data?: ?void,
       |}
     | {|
-          data: TData,
-          error?: ?void,
+          +data: TData,
+          +error?: ?void,
```
somewhatabstract, line 23: Flow was outputting some new errors about these being read-only in some cases and not others, so I just marked them as read-only since that's right anyway.

jeresig (approved):
Looks good, thank you!

PR #1155, opened 2021-07-01
Update to Flow 0.154.0

somewhatabstract:
## Summary:

Issue: XXX-XXXX

## Test plan:
`yarn flow`
`yarn test`

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "eslint-plugin-react": "^7.18.0",
     "eslint-plugin-react-native-animation-linter": "^0.1.2",
     "eslint-watch": "^6.0.1",
-    "flow-bin": "0.145.0",
+    "flow-bin": "0.154.0",
     "glob": "^7.1.2",
     "husky": "^4.2.5",
     "jest": "^25.1.0",
--- a/packages/wonder-blocks-dropdown/src/components/dropdown-core.js
+++ b/packages/wonder-blocks-dropdown/src/components/dropdown-core.js
@@ -573,6 +573,7 @@ class DropdownCore extends React.Component<Props, State> {
     };
 
     handleDropdownMouseUp: (event: SyntheticMouseEvent<>) => void = (event) => {
+        // $FlowIgnore[method-unbinding]
         if (event.nativeEvent.stopImmediatePropagation) {
             event.nativeEvent.stopImmediatePropagation();
         } else {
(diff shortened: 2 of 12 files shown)


packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -573,6 +573,7 @@ class DropdownCore extends React.Component<Props, State> {
     };
 
     handleDropdownMouseUp: (event: SyntheticMouseEvent<>) => void = (event) => {
+        // $FlowIgnore[method-unbinding]
         if (event.nativeEvent.stopImmediatePropagation) {
             event.nativeEvent.stopImmediatePropagation();
```
kevinbarabash, line 578: Why does `method-unbinding` come into play here?

packages/wonder-blocks-timing/src/util/__tests__/action-scheduler.test.js
```diff
@@ -66,6 +66,7 @@ describe("ActionScheduler", () => {
                 actionScheduler.clearAll();
 
                 // Assert
+                // $FlowIgnore[method-unbinding]
                 expect(testTimeout.clear).toHaveBeenCalledWith(true);
```
kevinbarabash, line 70: And here?  Does Flow just not like `.toHaveBeenCalledWith()` and `.toHaveBeenCalledTimes()` calls?

package.json
```diff
@@ -73,7 +73,7 @@
     "eslint-plugin-react": "^7.18.0",
     "eslint-plugin-react-native-animation-linter": "^0.1.2",
     "eslint-watch": "^6.0.1",
-    "flow-bin": "0.145.0",
```
kevinbarabash, line 76: wonder-blocks was quite behind.  Thanks for bring us up-to-date. 🎉

kevinbarabash (approved):
Thanks for updating.  LGTM :shipit:

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
somewhatabstract: Because their rule isn't very good yet? 🤷🏻‍♂️

packages/wonder-blocks-timing/src/util/__tests__/action-scheduler.test.js
somewhatabstract: It has nothing to do with the calls. It sees any method that is referenced without being called as a potential unbinding right now.

jeresig (approved):

packages/wonder-blocks-timing/src/util/__tests__/action-scheduler.test.js
kevinbarabash: That makes sense.

PR #1158, opened 2021-07-16
WB-1086 - Make LabeledTextField's `style` prop Apply to FieldHeading

RivasCVA:
## Summary:
Currently, the style prop on `LabeledTextField` applies to the internal `TextField` element.
Due to problems with trying to set width, margin, padding, etc., to the entire `FieldHeading` container,
we would like to instead have the style prop apply to the FieldHeading container.

Issue: https://khanacademy.atlassian.net/browse/WB-1086

## Test plan:
- Added unit tests to ensure `LabeledTextField`'s style is passed to FieldHeading
- Tested examples where setting flexGrow applies to the entire FieldHeading container

### DIFF
--- a/packages/wonder-blocks-form/src/components/labeled-text-field.js
+++ b/packages/wonder-blocks-form/src/components/labeled-text-field.js
@@ -84,6 +84,15 @@ type Props = {|
 
     /**
      * Custom styles for the container.
+     *
+     * Note: This style is passed to the field heading container
+     * due to scenarios where we would like to set a specific
+     * width for the text field. If we apply the style directly
+     * to the text field, the container would not be affected.
+     * For example, setting text field to "width: 50%" would not
+     * affect the container since text field is a child of the container.
+     * In this case, the container would maintain its width ocuppying
+     * unnecessary space when the text field is smaller.
      */
     style?: StyleType,
 
--- a/packages/wonder-blocks-form/src/components/labeled-text-field.js
+++ b/packages/wonder-blocks-form/src/components/labeled-text-field.js
@@ -83,7 +83,7 @@ type Props = {|
     light: boolean,
 
     /**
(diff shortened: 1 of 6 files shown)


RivasCVA:
I have also `yarn link` in Webapp, and my original issue is now fixed :)

webapp-new-labeled-text-field

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -83,7 +83,7 @@ type Props = {|
     light: boolean,
 
     /**
-     * Custom styles for the TextField component.
+     * Custom styles for the container.
```
jandrade, line 86: It would be worth adding a NOTE here why we apply that to the container and not to the `TextField` component.

packages/wonder-blocks-form/src/components/labeled-text-field.md
```diff
@@ -333,57 +333,6 @@ const styles = StyleSheet.create({
 <LabeledTextFieldExample />
 ```
 
-The field can have custom styles
```
jandrade, line 336: I wouldn't remove this test here, and would update it with an example of how custom styles can be applied in the container, so devs can understand possible uses. (same would apply to the story).

One good example could be the "two fields in a row" that seems to be a real case that devs could benefit from!

jandrade (approved):
The implementation works perfect! Just requesting some changes about documentation to show examples using this prop, but nothing blocking.

RivasCVA:
I've added new `style` examples in React Storybook (and Styleguidist):

new-example

packages/wonder-blocks-form/src/components/labeled-text-field.js
jandrade: Cool, thanks!

PR #1159, opened 2021-07-21
Fix SSR bug that would clear hydration cache incorrectly

somewhatabstract:
## Summary:
Before this change, rendering data repeatedly in SSR with WB Data and a handler with a custom client-side cache would cause data to be erased from the hydration cache. This was not what we wanted. This simple bug fix addresses that bug.

Issue: FEI-3701

## Test plan:
`yarn test`

I'll be putting this into webapp and it will be clear real quick from content and test prep pages if this is working or not.

### DIFF
--- a/packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
+++ b/packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
@@ -451,137 +451,205 @@ describe("../response-cache.js", () => {
         });
 
         describe("with custom cache", () => {
-            it("SSR should use framework cache, not custom cache", () => {
-                // Arrange
-                jest.spyOn(Server, "isServerSide").mockReturnValue(true);
-                const internalCache = new MemoryCache();
-                jest.spyOn(internalCache, "retrieve");
-                const cache = new ResponseCache(internalCache);
-                const customCache = {
-                    store: jest.fn(),
-                    retrieve: jest.fn(),
-                    remove: jest.fn(),
-                    removeAll: jest.fn(),
-                };
-                const fakeHandler: IRequestHandler<string, string> = {
-                    getKey: () => "MY_KEY",
-                    type: "MY_HANDLER",
-                    shouldRefreshCache: () => false,
-                    fulfillRequest: jest.fn(),
-                    cache: customCache,
-                    hydrate: true,
-                };
+            describe("SSR", () => {
+                it("should use framework cache, not custom cache", () => {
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
```diff
@@ -451,137 +451,205 @@ describe("../response-cache.js", () => {
...
+                });
             });
 
-            it("CSR should remove framework entry when custom cache misses", () => {
-                // Arrange
-                jest.spyOn(Server, "isServerSide").mockReturnValue(false);
-                const internalCache = new MemoryCache();
-                jest.spyOn(internalCache, "retrieve").mockReturnValue({
-                    data: "data!",
+            describe("CSR", () => {
```
somewhatabstract, line 550: These test cases are unchanged, I just scoped them in a `describe`

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
```diff
@@ -451,137 +451,205 @@ describe("../response-cache.js", () => {
...
-            });
-
-            it("CSR should store framework entry in custom cache when custom cache misses", () => {
-                // Arrange
-                jest.spyOn(Server, "isServerSide").mockReturnValue(false);
-                const internalCache = new MemoryCache();
-                jest.spyOn(internalCache, "retrieve").mockReturnValue({
-                    data: "data!",
+            describe("SSR", () => {
+                it("should use framework cache, not custom cache", () => {
```
somewhatabstract, line 455: This test case is unchanged, it's just scoped in a `describe`

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
```diff
@@ -451,137 +451,205 @@ describe("../response-cache.js", () => {
...
-                    type: "MY_HANDLER",
-                    shouldRefreshCache: () => false,
-                    fulfillRequest: jest.fn(),
-                    cache: customCache,
-                    hydrate: true,
-                };
 
-                // Act
-                const result = cache.getEntry(fakeHandler, "options");
+                it("should not store framework entry in custom cache", () => {
```
somewhatabstract, line 487: This and the following test case are new, to verify that the SSR behavior is correct. Without my change, these would fail.

jeresig (approved):
Thanks for adding regression test cases!

PR #1161, opened 2021-07-22
Add ability to query if there are unfulfilled requests

somewhatabstract:
## Summary:
Now that WonderBlocks is responsible for cached queries, it is important that we support repeated render loops to get all the data requests. React Apollo's getDataFromTree handled this with its own render loop. Since we already have a
render loop in WebApp, we need to know when to stop looping in order to recreate similar behavior. We already do this for PageSettings by checking if anymore changes got made in the previous render and for loadables by checking for pending loadables. This adds the ability to check if we have pending tracked data requests.

Once this is in webapp, we can update the render loop to incorporate WB Data fulfillment rather that it only occurring once after a single tracking cycle.

Issue: FEI-3701

## Test plan:
`yarn test`

Incorporate into webapp, modify rendering to include WB Data fulfillment and tracking in the render loop, and update snapshots. They should then resemble the snapshots from before the move off React Apollo.

### DIFF
--- a/packages/wonder-blocks-data/src/components/internal-data.js
+++ b/packages/wonder-blocks-data/src/components/internal-data.js
@@ -209,7 +209,7 @@ export default class InternalData<
         const result = this._resultFromState();
         // We only track data requests when we are server-side and we don't
         // already have a result. The existence of a result is indicated by the
-        // loading flag being true.
+        // loading flag being false.
         if (result.loading && Server.isServerSide()) {
             return this._renderWithTrackingContext(result);
         }
--- a/packages/wonder-blocks-data/src/components/internal-data.js
+++ b/packages/wonder-blocks-data/src/components/internal-data.js
@@ -207,6 +207,9 @@ export default class InternalData<
 
     render(): React.Node {
         const result = this._resultFromState();
+        // We only track data requests when we are server-side and we don't
+        // already have a result. The existence of a result is indicated by the
+        // loading flag being true.
         if (result.loading && Server.isServerSide()) {
             return this._renderWithTrackingContext(result);
         }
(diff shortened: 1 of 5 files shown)


packages/wonder-blocks-data/src/index.js
```diff
@@ -21,8 +22,21 @@ export type {
...
     ResCache.Default.initialize(source);
 
-export const fulfillAllDataRequests = (): Promise<ResponseCache> =>
-    RequestTracker.Default.fulfillTrackedRequests();
+export const fulfillAllDataRequests = (): Promise<ResponseCache> => {
+    if (!Server.isServerSide()) {
+        return Promise.reject(
+            new Error("Data requests are not tracked when client-side"),
+        );
+    }
```
somewhatabstract, line 30: This would've helped catch that we weren't setting WB to server side mode in our route tests if we had it place before my 2 hour search for that bug the other day.

packages/wonder-blocks-data/src/util/request-tracking.js
```diff
@@ -98,6 +98,13 @@ export class RequestTracker {
         this._trackedRequests = {};
     };
 
+    /**
+     * Indicates if we have requests waiting to be fulfilled.
+     */
+    get hasUnfulfilledRequests(): boolean {
+        return Object.keys(this._trackedRequests).length > 0;
+    }
+
```
somewhatabstract, line 107: This is the bit I really need :)

packages/wonder-blocks-data/src/components/internal-data.js
```diff
@@ -207,6 +207,9 @@ export default class InternalData<
 
     render(): React.Node {
         const result = this._resultFromState();
+        // We only track data requests when we are server-side and we don't
+        // already have a result. The existence of a result is indicated by the
+        // loading flag being true.
```
somewhatabstract, line 212: NIT: should be `false`

kevinbarabash (approved):
LGTM.  The tests make sense.  Thanks for adding those `isServerSide()` checks. :shipit:

PR #1163, opened 2021-07-29
Allow setting a role of alertdialog for ModalDialog and OnePaneDialog

jeanettehead:
## Summary:
Alert dialog is similar to dialog, but it cues the screen reader techologoy to announce the alert message when the modal is opened
For more context: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alertdialog_role

Issue: WB-1089

## Test plan:
`yarn test`


kevinbarabash:
I used admin privileges to merge this PR because we're having issues with checks not running on CircleCI.

PR #1166, opened 2021-08-24
Fix wallaby config

kevinbarabash:
We'd like to use the new "exclusive test run" feature.  The new version of wallaby didn't understand our current config.  I've updated it to auto-detect the config, but to do that jest.config.js needs to live in the root of the repo.  I've added a jest.config.js which re-exports config/jest/test.config.js.

### DIFF
--- a/wallaby.js
+++ b/wallaby.js
@@ -1,91 +1,20 @@
 /* eslint-disable import/no-commonjs */
-const babelConfig = require("./build-settings/babel.config.js");
-
-module.exports = function (wallaby) {
-    const tests = [
-        "packages/**/__tests__/*.test.js",
-        "shared-unpackaged/**/*.test.js",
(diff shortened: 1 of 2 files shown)


wallaby.js
```diff
@@ -1,91 +1,16 @@
 /* eslint-disable import/no-commonjs */
-const babelConfig = require("./build-settings/babel.config.js");
-
-module.exports = function (wallaby) {
```
jandrade, line 4: Does the change mean that all the contents of this function are now handled with `autoDetect=true`?

jandrade (approved):
LGTM! Just a question about how the current Wallaby configuration works.

wallaby.js
kevinbarabash: I think it looks at jest.config.js and figures things out from that.

PR #1167, opened 2021-08-24
Fei 3856: Fix SingleSelect so that we can test it using @testing-library

kevinbarabash:
For some reason popper includes `pointerEvents: "none"` in the `style` it passes to us, but only during testing.  It must be autodetecting jsdom.  I'm not sure why it does this, but removing this property from `style` fixes the issue and all existing tests pass.  This PR also updates all `SingleSelect` tests to use `@testing-library`.


packages/wonder-blocks-dropdown/src/components/search-text-input.js
```diff
@@ -130,7 +130,7 @@ export default class SearchTextInput extends React.Component<Props, State> {
 
     maybeRenderDismissIconButton(): React.Node {
         const {searchText} = this.props;
-        const {clearSearch} = this.state.labels;
+        const {clearSearch} = this.state.labels; // ?
```
jandrade, line 133: is this change intended?

packages/wonder-blocks-dropdown/src/components/__tests__/single-select.test.js
```diff
@@ -1,111 +1,207 @@
...
-import DropdownOpener from "../dropdown-opener.js";
-import SelectOpener from "../select-opener.js";
 import OptionItem from "../option-item.js";
 import SingleSelect from "../single-select.js";
-import {keyCodes} from "../../util/constants.js";
-import SearchTextInput from "../search-text-input.js";
 
 jest.mock("../dropdown-core-virtualized.js");
 
+const FRAME_DURATION = 17;
```
jandrade, line 13: Is there a specific reason why we use this value?

packages/wonder-blocks-dropdown/src/components/__tests__/single-select.test.js
```diff
@@ -1,111 +1,207 @@
...
+                // Act
+                userEvent.click(screen.getByText("item 1")); // closed
+
+                // Assert
+                expect(onChange).toHaveBeenCalledWith("1"); // value
+            });
+        });
+
+        describe("keyboard", () => {
+            describe.each([{key: "{enter}"}, {key: "{space}"}])(
```
jandrade, line 123: Nice improvement!

packages/wonder-blocks-dropdown/src/components/__tests__/single-select.test.js
```diff
@@ -428,21 +514,21 @@ describe("SingleSelect", () => {
...
-            });
+            const searchInput = screen.getByRole("textbox");
+            userEvent.paste(searchInput, "Item 2");
 
             // Assert
-            expect(wrapper.find(OptionItem).at(0).text()).toEqual("item 2");
-            expect(wrapper.find(OptionItem).at(1).exists()).toBe(false);
+            const options = screen.getAllByRole("option");
+            expect(options).toHaveLength(1);
+            expect(options[0]).toHaveTextContent("item 2");
```
jandrade, line 526: So much simpler!

jandrade (approved):
Looks great! It's awesome to see every time how RTL makes tests simpler.

packages/wonder-blocks-dropdown/src/components/search-text-input.js
kevinbarabash: whoopsie, I'll remove that

packages/wonder-blocks-dropdown/src/components/__tests__/single-select.test.js
kevinbarabash: Yes, it's approximately 1000/60 or 1/60th of a second.  It's not really necessary to use that value though since we're using fake timers.

packages/wonder-blocks-dropdown/src/components/__tests__/single-select.test.js
kevinbarabash: The tests within this describe are testing more than we were testing before.

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -748,10 +748,13 @@ class DropdownCore extends React.Component<Props, State> {
                     }}
                 >
                     {({placement, ref, style, outOfBoundaries}) => {
+                        // For some reason popper.js includes `pointerEvents: "none"` in the `style`
+                        // it passes to us, but only when running the tests.
+                        const {pointerEvents: _, ...restStyle} = style;
```
somewhatabstract, line 753: To add some reasoning - it's possibly our mock PopperJS that's setting that style. When it was just used for Tooltips, this made sense as generally, we wouldn't want pointer events going to the tooltip, but instead to what's under it (we've changed hit detection since then, so this isn't "necessary" anymore, I don't think). My guess is that our mock of PopperJS is still adding this style.

kevinbarabash: I'll have a look at the popper mock to see if that's what's happening.

kevinbarabash: I looked at all of the __mocks__/popper.js.js files and I don't see anything in them that is mocking the style that's being returned.  I did some digging and found that our version of react-popper sets
```
  var initialStyle = {
    position: 'absolute',
    top: 0,
    left: 0,
    opacity: 0,
    pointerEvents: 'none'
  };
```
and defines a `getPopperStyle` helper like so
```
      _defineProperty(_assertThisInitialized(_this), "getPopperStyle", function () {
        return !_this.popperNode || !_this.state.data ? initialStyle : _extends({
          position: _this.state.data.offsets.popper.position
        }, _this.state.data.styles);
      });
```
I'll update the comment to say "react-popper" instead of "popper.js".

PR #1170, opened 2021-09-01
Update dropdown-core to focus the menu container (element with role "menu" or "lostbox" that contains the menu items) when the menu is opened

jeanettehead:
## Summary:
This is so that mobile screen readers can access the menu. Without this change, focus is not shifted into the menu when it is opened with a screen reader, and screen reader users cannot interact with the menu.

Issue: LP-10008

## Test plan:
yarn jest

Open a dropdown menu with a mobile screen reader (VoiceOver/Safari/iOS or TalkBack/Chrome/Android)
Note that your focus jumps to the menu container

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/dropdown-core.js
+++ b/packages/wonder-blocks-dropdown/src/components/dropdown-core.js
@@ -187,6 +187,11 @@ class DropdownCore extends React.Component<Props, State> {
     listRef: {|
         current: null | React.ElementRef<typeof List>,
     |};
+    // Keeps a reference of the menu instance so we can focus it when the menu
+    // is opened for screen reader support
+    menuRef: {|
+        current: null | React.ElementRef<typeof View>,
+    |};
 
     // Figure out if the same items are focusable. If an item has been added or
     // removed, this method will return false.
@@ -261,6 +266,7 @@ class DropdownCore extends React.Component<Props, State> {
         };
 
         this.listRef = React.createRef();
+        this.menuRef = React.createRef();
     }
 
     componentDidMount() {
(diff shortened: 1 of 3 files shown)


jeanettehead:
Obviously this code is incomplete - opening this PR as a conversation starter

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -332,30 +338,21 @@ class DropdownCore extends React.Component<Props, State> {
...
-            searchText,
-        } = this.props;
+        const {open, onSearchTextChanged, searchText} = this.props;
 
         if (open) {
-            // Reset focused index
-            this.focusedIndex = initialFocusedIndex;
-            // We explicitly set focus to the first item only if we sense
-            // that the user opened the menu via the keyboard
-            if (keyboard) {
```
jeanettehead, line 348: In general this should be avoided because screen readers are keyboard interfaces that generate regular click events

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -332,30 +338,21 @@ class DropdownCore extends React.Component<Props, State> {
...
-            initialFocusedIndex,
-            open,
-            onSearchTextChanged,
-            searchText,
-        } = this.props;
+        const {open, onSearchTextChanged, searchText} = this.props;
 
         if (open) {
-            // Reset focused index
-            this.focusedIndex = initialFocusedIndex;
```
jeanettehead, line 345: So, I just broke this behavior and I don't know how to fix it - the ability to specify which item in the list to focus initially is in conflict with the behavior of focusing the menu container when it is opened.

It might be possible to focus the focusedIndex if one is specified in props? And default to focusing the menu? And warn about breaking a11y best practices by setting it in the documentation?

jeanettehead: (related - I commented out a bunch of tests above that were at least partly about this behavior that obviously need addressing once we know what the plan is)

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -332,30 +338,21 @@ class DropdownCore extends React.Component<Props, State> {
...
-            // We explicitly set focus to the first item only if we sense
-            // that the user opened the menu via the keyboard
-            if (keyboard) {
-                this.keyboardNavOn = true;
```
jeanettehead, line 349: There's other code in here that uses this value that would need to be addressed in this change as well once we have more clearly defined behavior

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -402,6 +399,17 @@ class DropdownCore extends React.Component<Props, State> {
         this.props.schedule.animationFrame(() => this.focusCurrentItem());
     }
 
+    focusMenu() {
+        requestAnimationFrame(() => {
```
jeanettehead, line 403: For the next version of this code, this should be this.prop.schedule

jeanettehead:
Whoever takes a look at this, let's schedule some time to chat about it! This is a first draft - but it does work on android, which the previous version did not at all.

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
jandrade: Oh, just a heads up... @PeppeMiller is working on fixing the focus when the dropdown is opened, so there might be some things you both would like to align on: https://khanacademy.atlassian.net/browse/WB-1091

PR #20, opened 2018-04-02
Add new colors to match the color spec

jangmi-jo:
Summary:
Related task: https://khanacademy.atlassian.net/browse/CLASS-1340
Mockup: zpl://screen?sid=5a6226d4ba218e7607ff5661&pid=5a622684dd6d876f24a8cb20

Test Plan:
Run `npm run storybook`
Go to http://localhost:9001/?selectedKind=Color&selectedStory=All%20colors&full=0&addons=1&stories=1&panelRight=0&addonPanel=kadira%2Fjsx%2Fpanel
See all new colors

screen shot 2018-04-02 at 1 11 07 pm

PR #1171, opened 2021-09-02
Remove keyboard tracking from dropdown-core so that menus can be accessed with a screen reader

jeanettehead:
## Summary:
The focus events that we have for keyboard users are also important for screen readers - but screen readers generate click events that cannot be differentiated from regular click events. For that reason I removed all of the keyboard specific code and merged the click interface with the keyboard interface to the bet of my ability.

This is so that mobile screen readers can access the dropdowns. Without this change, focus is not shifted into the menu when it is opened with a screen reader, and screen reader users cannot interact with the menu.
It also greatly improves screen reader access on web. Currently dropdowns can be only accessed if the user knows to switch to the keyboard controls (enter/arrow) instead of using the screen reader controls (ie VO + enter / VO + arrow keys) which isn't a reasonable assumption to make of our users.

Issue: LP-10008

## Test plan:
yarn jest

Open a dropdown menu with a mobile screen reader (VoiceOver/Safari/iOS or TalkBack/Chrome/Android). Note that your focus jumps into the menu

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/__tests__/dropdown-core.test.js
+++ b/packages/wonder-blocks-dropdown/src/components/__tests__/dropdown-core.test.js
@@ -76,7 +76,6 @@ describe("DropdownCore", () => {
                     },
                 ]}
                 role="listbox"
-                keyboard={true}
                 light={false}
                 open={false}
                 // mock the opener elements
@@ -95,7 +94,6 @@ describe("DropdownCore", () => {
         const handleOpen = jest.fn();
         dropdown.setProps({
             initialFocusedIndex: 0,
-            keyboard: true,
             onOpenChanged: (open) => handleOpen(open),
             open: true,
         });
@@ -232,7 +230,6 @@ describe("DropdownCore", () => {
                         },
                     ]}
                     role="listbox"
-                    keyboard={true}
                     light={false}
                     open={true}
                     // mock the opener elements
@@ -292,7 +289,6 @@ describe("DropdownCore", () => {
         const handleOpen = jest.fn();
         dropdown.setProps({
             initialFocusedIndex: 2,
-            keyboard: true,
             onOpenChanged: (open) => handleOpen(open),
             open: true,
         });
@@ -307,43 +303,10 @@ describe("DropdownCore", () => {
         expect(document.activeElement).toBe(elementAtIndex(dropdown, 0));
     });
 
-    it("focuses correct item when opened with click, then switching to keyboard", () => {
-        dropdown.setProps({
-            initialFocusedIndex: 1,
-            keyboard: false,
-            open: true,
-        });
-
-        dropdown.simulate("keydown", {keyCode: keyCodes.down});
-        dropdown.simulate("keyup", {keyCode: keyCodes.down});
-        jest.runAllTimers();
-        expect(document.activeElement).toBe(elementAtIndex(dropdown, 1));
-    });
-
-    it("focuses correct item after clicking on option, then switching to keyboard", () => {
-        dropdown.setProps({
-            initialFocusedIndex: 0,
-            keyboard: false,
-            open: true,
-        });
-
-        const option1 = dropdown.find("OptionItem").at(1);
-
-        // Click on item at index 1
-        option1.simulate("click");
-
-        // When starting to use keyboard behavior, should move to next item
-        dropdown.simulate("keydown", {keyCode: keyCodes.down});
-        dropdown.simulate("keyup", {keyCode: keyCodes.down});
-        jest.runAllTimers();
-        expect(document.activeElement).toBe(elementAtIndex(dropdown, 2));
-    });
-
     it("focuses correct item with clicking/pressing with initial focused of not 0", () => {
         // Same as the previous test, expect initialFocusedIndex is 2 now
         dropdown.setProps({
             initialFocusedIndex: 2,
-            keyboard: false,
             open: true,
         });
 
@@ -400,7 +363,6 @@ describe("DropdownCore", () => {
                     populatedProps: {},
                 },
             ],
-            keyboard: true,
             open: true,
         });
 
@@ -423,7 +385,6 @@ describe("DropdownCore", () => {
     it("focuses correct item after different items become focusable", () => {
         dropdown.setProps({
             initialFocusedIndex: 0,
-            keyboard: true,
             open: true,
         });
 
@@ -639,7 +600,7 @@ describe("DropdownCore", () => {
                 },
             ],
         });
-        // SearchTextInput should be focused (since keyboard is true)
+        // SearchTextInput should be focused
         const searchInput = dropdown.find(SearchTextInput);
         jest.runAllTimers();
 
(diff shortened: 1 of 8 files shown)


jeanettehead:
I had a more complicated version of this one https://github.com/Khan/wonder-blocks/pull/1170 but this approach is simpler and behaves better in testing

kevinbarabash:
I'll have a look at this on Tuesday.

packages/wonder-blocks-dropdown/src/components/__tests__/dropdown-core.test.js
```diff
@@ -317,7 +315,7 @@ describe("DropdownCore", () => {
         dropdown.simulate("keydown", {keyCode: keyCodes.down});
         dropdown.simulate("keyup", {keyCode: keyCodes.down});
         jest.runAllTimers();
-        expect(document.activeElement).toBe(elementAtIndex(dropdown, 1));
+        expect(document.activeElement).toBe(elementAtIndex(dropdown, 2));
```
jandrade, line 318: Do you know why the focus changed to a different element in this case?

packages/wonder-blocks-dropdown/src/components/__tests__/single-select.test.js
```diff
@@ -55,15 +55,17 @@ describe("SingleSelect", () => {
                 expect(screen.queryByRole("listbox")).toBeInTheDocument();
             });
 
-            it("the opener should keep the focus after opening", () => {
+            it("should focus the first item in the dropdown", () => {
```
jandrade, line 58: Nice! it should behave this way :)

jandrade (commented):
Taking a first pass on the changes, it looks that everything makes sense, but I'd like to hear thoughts from other folks to try to understand why we introduced this API in the first time (this was implemented a couple of years ago, so I'm not sure if we'll get enough context about this).

@jeanettehead One thing that I noticed is that the focus works fine if the user selects an element that is currently visible (e.g. one of the first 8 elements in the list)... but if the user selects an element down the list (e.g. item #20), then the focus is lost. This means that we might need to make more changes in order to fix this issue when we display long option lists.

packages/wonder-blocks-dropdown/src/components/__tests__/dropdown-core.test.js
jeanettehead: the initialFocusedIndex is 1 so keying down should focus 2 so this assertion is logical to me.

I would assume it's either because the click behavior ignored the "initialFocusedIndex" prop.

We could probably remove this assertion altogether since it's logic that doesn't really do much.

jeanettehead:
> Taking a first pass on the changes, it looks that everything makes sense, but I'd like to hear thoughts from other folks to try to understand why we introduced this API in the first time (this was implemented a couple of years ago, so I'm not sure if we'll get enough context about this).
> 
> @jeanettehead One thing that I noticed is that the focus works fine if the user selects an element that is currently visible (e.g. one of the first 8 elements in the list)... but if the user selects an element down the list (e.g. item #20), then the focus is lost. This means that we might need to make more changes in order to fix this issue when we display long option lists.

👍 As long as it works within the first 8 that will at least fix the BofA use case (dropdowns on the discussion page) that triggered the need for this fix, so it's at least a good start in its current form. If we feel we should definitely tackle longer lists, I could probably make some time to poke at that later this week?

When looking at how other design libraries handle this (like material UI) - the popular structure seems to be to make the menu container (thing with `role=menu` or `role=listbox`` that holds the list items) focusable and focus that element and let them navigate from there. That would mean removing (or frequently bypassing) the initialFocusedIndex code, but maybe it would help in the case of longer lists?

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.js
```diff
@@ -379,9 +379,6 @@ describe("MultiSelect", () => {
             .find(OptionItem)
             .at(2)
             .find(ClickableBehavior);
-        // The focus is on opener. Press up (or down) should focus the input
-        select.simulate("keydown", {keyCode: keyCodes.up});
-        select.simulate("keyup", {keyCode: keyCodes.up});
```
jandrade, line 384: Oh this makes total sense! I'm glad that the browser now handles focus correctly without having to specify that.

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -343,12 +323,7 @@ class DropdownCore extends React.Component<Props, State> {
         if (open) {
             // Reset focused index
             this.focusedIndex = initialFocusedIndex;
-            // We explicitly set focus to the first item only if we sense
-            // that the user opened the menu via the keyboard
-            if (keyboard) {
-                this.keyboardNavOn = true;
-                this.scheduleToFocusCurrentItem();
-            }
+            this.scheduleToFocusCurrentItem();
```
jandrade, line 326: It seems that this is called again in line # 333. Could you please get rid of one of these fn calls?

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -301,16 +291,7 @@ class DropdownCore extends React.Component<Props, State> {
...
-                        // If keyboard navigation was already on, use that
-                        this.scheduleToFocusCurrentItem();
-                    } else {
-                        // Otherwise shift focus to the original focus item
-                        // (the opener) to listen for further keyboard events
-                        if (this.props.openerElement) {
-                            this.props.openerElement.focus();
-                        }
-                    }
+                    this.scheduleToFocusCurrentItem();
```
kevinbarabash, line 294: I thought that using the scroll wheel would cause a new item to be focus, but scrolling causes the focused item to become unfocused.  This is the same behaviour as before this PR so not an issue for this review.  I'm not sure if we even care about the interaction between using the keyboard and the mouse at the same time.  It seems likely that most people only use one at a time.

jandrade (approved):
Looks good to me! Thanks for tackling this issue and removing the unnecessary code 🚀 

> When looking at how other design libraries handle this (like material UI) - the popular structure seems to be to make the menu container (thing with role=menu or `role=listbox`` that holds the list items) focusable and focus that element and let them navigate from there. That would mean removing (or frequently bypassing) the initialFocusedIndex code, but maybe it would help in the case of longer lists?

Let's keep it this way for now. Regarding removing `initialFocusedIndex` code... I'd say that it's better to keep it around until we fix an issue with the focus being lost on longer lists that Pepper is working on. Then we can revisit the whole logic after that.

kevinbarabash (commented):
I'm happy to see all of the keyboard specific code in dropdown-core.js go away.  I think some of the logic may have been to facilitate switching between keyboard and mouse while using the dropdowns.  The fact that the extra logic was breaking screen readers is enough motivation for me to remove it.  I added some more keyboard nav tests for `SingleSelect` recently.  Since those are still passing I think we can have confidence in this change.  @jeanettehead you may want to have a look at the tests in https://github.com/Khan/wonder-blocks/blob/master/packages/wonder-blocks-dropdown/src/components/__tests__/single-select.test.js to make sure I didn't miss any key keyboard test cases.  @jandrade I'll leave the final approval to you since I haven't looked at the code changes in this PR as closely as you have.

jeanettehead:
opened https://github.com/Khan/wonder-blocks/pull/1173 which is a duplicate but maybe circle ci will run

PR #1173, opened 2021-09-09
Remove keyboard tracking from dropdown-core so that menus can be accessed with a screen reader

jeanettehead:
## Summary:
The focus events that we have for keyboard users are also important for screen readers - but screen readers generate click events that cannot be differentiated from regular click events. For that reason I removed all of the keyboard specific code and merged the click interface with the keyboard interface to the bet of my ability.

This is so that mobile screen readers can access the dropdowns. Without this change, focus is not shifted into the menu when it is opened with a screen reader, and screen reader users cannot interact with the menu.
It also greatly improves screen reader access on web. Currently dropdowns can be only accessed if the user knows to switch to the keyboard controls (enter/arrow) instead of using the screen reader controls (ie VO + enter / VO + arrow keys) which isn't a reasonable assumption to make of our users.

Issue: LP-10008

## Test plan:
yarn jest

Open a dropdown menu with a mobile screen reader (VoiceOver/Safari/iOS or TalkBack/Chrome/Android). Note that your focus jumps into the menu

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/__tests__/dropdown-core.test.js
+++ b/packages/wonder-blocks-dropdown/src/components/__tests__/dropdown-core.test.js
@@ -76,7 +76,6 @@ describe("DropdownCore", () => {
                     },
                 ]}
                 role="listbox"
-                keyboard={true}
                 light={false}
                 open={false}
                 // mock the opener elements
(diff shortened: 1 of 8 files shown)


jeanettehead:
this is a duplicate PR to https://github.com/Khan/wonder-blocks/pull/1171 but hopefully circle ci checks will run this time

jeanettehead:
this didn't advance the circle ci check so I'm closing this one

PR #1175, opened 2021-09-17
Changed initialFocusedIndex on drop downs to scroll into view

PeppeMiller:
## Summary:
If you have a select list with more than 9 items, they won't
all fit in view.  If you have the 10th or later item selected,
upon open that item should not only be selected, but visible
too.  To accomidate this, I updated the code so that
initialFocusedIndex not only focuses the item, but scrolls
it into view.  This brought to light a few bugs that were
fixed along the way.

Issue: https://khanacademy.atlassian.net/browse/WB-1091
https://khanacademy.atlassian.net/browse/CLASS-7007

## Test plan:
Verify that all unit tests work correctly.
In the Wonder Blocks documentation,
"Single select with search filter and custom styles"
example:
  * Open the drop down
  * Select an item that is initially out of view
  * Re-open drop down
  * Verify that the item visible
Change the example so no filter is available and
rerun the changes.

Open the dropdown
Select the filter
Type values
Enter a space, verifying that it works
Press tab, verifying that it moves you to the X

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/dropdown-core-virtualized.js
+++ b/packages/wonder-blocks-dropdown/src/components/dropdown-core-virtualized.js
@@ -57,7 +57,10 @@ type State = {|
 |};
 
 /**
- * Maximum visible items inside the dropdown list
+ * Maximum visible items inside the dropdown list.
+ * Based on the defined height that we're using, this is the maximium
+ * number of items that can fit into the visible porition of the
+ * dropdowns list box.
  */
 const MAX_VISIBLE_ITEMS = 9;
 
--- a/packages/wonder-blocks-dropdown/src/components/dropdown-core-virtualized.js
+++ b/packages/wonder-blocks-dropdown/src/components/dropdown-core-virtualized.js
@@ -59,7 +59,7 @@ type State = {|
 /**
  * Maximum visible items inside the dropdown list
  */
-const MAX_VISIBLE_ITEMS = 10;
+const MAX_VISIBLE_ITEMS = 9;
 
 /**
  * A react-window's List wrapper that instantiates the virtualized list and
--- a/packages/wonder-blocks-dropdown/src/components/dropdown-core.js
+++ b/packages/wonder-blocks-dropdown/src/components/dropdown-core.js
@@ -42,7 +42,7 @@ type DefaultProps = {|
      * An index that represents the index of the focused element when the menu
      * is opened.
      */
-    initialFocusedIndex: number,
+    initialFocusedIndex?: number,
 
     /**
      * Whether this menu should be left-aligned or right-aligned with the
@@ -198,7 +198,7 @@ class DropdownCore extends React.Component<Props, State> {
 
     static defaultProps: DefaultProps = {
         alignment: "left",
-        initialFocusedIndex: 0,
+        initialFocusedIndex: undefined,
         labels: {
             noResults: defaultLabels.noResults,
         },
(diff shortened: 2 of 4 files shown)


packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -198,7 +198,7 @@ class DropdownCore extends React.Component<Props, State> {
 
     static defaultProps: DefaultProps = {
         alignment: "left",
-        initialFocusedIndex: 0,
+        initialFocusedIndex: undefined,
```
kevinbarabash, line 201: This should probably be removed completely.

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -310,14 +318,39 @@ class DropdownCore extends React.Component<Props, State> {
...
         if (open) {
             // Reset focused index
-            this.focusedIndex = initialFocusedIndex;
+            // If we are given an initial focus index, select it.  Otherwise
+            // default to the first item
+            if (initialFocusedIndex) {
+                this.focusedIndex = initialFocusedIndex + this.getIndexOffset();
+            } else {
+                this.focusedIndex = 0;
+            }
```
kevinbarabash, line 353: We do pretty much the same thing in the constructor.  Can you create a helper method that gets the initial focus index?  `getIndexOffset` should probably be inlined into the new helper.

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -475,20 +504,13 @@ class DropdownCore extends React.Component<Props, State> {
...
+        const {onOpenChanged, open} = this.props;
         const keyCode = event.which || event.keyCode;
-        const showSearchTextInput =
-            !!onSearchTextChanged && typeof searchText === "string";
         switch (keyCode) {
             case keyCodes.space:
                 // When we display SearchTextInput and the focus is on it,
                 // we should let the user type space.
-                if (showSearchTextInput && this.focusedIndex === 0) {
+                if (this.hasSearchBox() && this.focusedIndex === 0) {
```
kevinbarabash, line 513: Nice cleanup from this helper.

kevinbarabash (commented):
Thanks for these improvements to dropdown-core.js.  Given that the behaviour has changed, it would be good to have some more automated tests for this.  I'm not sure how much of this is testable using `@testing-library`, but maybe test some of the logic around:
- the difference in behaviour when `initialFocusedIndex` is undefined and the dropdown has a search box vs. not having a searchbox
- restoring the focus index after closing and reopening the dropdown

packages/wonder-blocks-dropdown/src/components/dropdown-core-virtualized.js
```diff
@@ -59,7 +59,7 @@ type State = {|
 /**
  * Maximum visible items inside the dropdown list
  */
-const MAX_VISIBLE_ITEMS = 10;
+const MAX_VISIBLE_ITEMS = 9;
```
jandrade, line 62: One more thing that'd be nice to add is explaining why we set 9 items here. It was a bit tricky to identify why the focused element didn't appear on the screen at some point.

jandrade (commented):
Things are looking good! Agreed with @kevinbarabash's comments around testing. Also requesting a minor improvement on the docs/comments.

PeppeMiller:
I updated with all of the suggestions, thanks!  The one thing I didn't do was write a unit test for verifying that a value is kept after re-opening.  I wasn't sure how to make that work in a unit test (though it works in the UI).

jandrade (approved):
Looks great! Thanks for all the refactors and for adding the tests. I checked the extra test you mentioned in [Storybook](https://5e1bf4b385e3fb0020b7073c-hwrxnznjrk.chromatic.com/?path=/story/dropdown-singleselect--with-filter-opened) and everything looks great!

jandrade:
test

PR #1181, opened 2021-09-29
Disable scheduling actions after unmount

somewhatabstract:
## Summary:
This makes it so that calling to schedule an action after a component has been unmounted will result in a noop action.

Issue: WB-1093

## Test plan:
`yarn test`


packages/wonder-blocks-timing/src/util/action-scheduler.js
```diff
@@ -58,4 +75,13 @@ export default class ActionScheduler implements IScheduleActions {
...
     }
+
+    /**
+     * Prevents this scheduler from creating any additional actions.
+     * This also clears any pending actions.
+     */
+    disable(): void {
+        this._disabled = true;
+        this.clearAll();
+    }
```
jeresig, line 86: Perfect!

jeresig (approved):
This is great, thank you!

PR #1183, opened 2021-09-30
Improved API for `withActionScheduler`

somewhatabstract:
## Summary:
- [TYPES] Improve HOC types so ref forwarded instances are not just `mixed`
- [BREAKING CHANGE] Use options and named policies instead of boolean values

Issue: WB-1094

## Test plan:
`yarn test`
`yarn flow`

## Update plan:
When updating Webapp, the flow type changes should catch places that need changing with respect to the new policy approach, but I'll also do a search for `withActionScheduler` uses so that the better ref-forwarding can be introduced.


packages/wonder-blocks-timing/src/components/with-action-scheduler.js
```diff
@@ -21,12 +21,16 @@ type WithoutActionScheduler<T> = $Exact<$Diff<T, WithActionSchedulerProps>>;
...
+    Component: React.AbstractComponent<WithActionScheduler<Config>, Instance>,
 >(
     WrappedComponent: Component,
-): React.ComponentType<WithoutActionScheduler<React.ElementConfig<Component>>> {
-    return React.forwardRef((props, ref) => (
+): React.AbstractComponent<
+    WithoutActionScheduler<React.ElementConfig<Component>>,
+    Instance,
+> {
+    return React.forwardRef<_, Instance>((props, ref) => (
```
somewhatabstract, line 33: These changes mean that when ref forwarding is used, instead of the ref being `mixed` it will be typed to the wrapped instance.

packages/wonder-blocks-timing/src/util/animation-frame.js
```diff
@@ -19,22 +24,23 @@ export default class AnimationFrame implements IAnimationFrame {
...
+        schedulePolicy: SchedulePolicy = SchedulePolicies.Immediately,
     ) {
         if (typeof action !== "function") {
             throw new Error("Action must be a function");
         }
 
         this._action = action;
 
-        if (autoSchedule || autoSchedule == null) {
+        if (schedulePolicy === SchedulePolicies.Immediately) {
```
somewhatabstract, line 43: I think this use of named policies makes the implementation easier to read.

packages/wonder-blocks-timing/src/util/animation-frame.js
```diff
@@ -61,10 +67,10 @@ export default class AnimationFrame implements IAnimationFrame {
      */
     set(): void {
         if (this.isSet) {
-            this.clear(false);
+            this.clear(ClearPolicies.Cancel);
         }
         this._animationFrameId = requestAnimationFrame((time) =>
-            this.clear(true, time),
+            this.clear(ClearPolicies.Resolve, time),
```
somewhatabstract, line 73: I think this use of named policies makes the implementation easier to read.

packages/wonder-blocks-timing/src/util/types.js
```diff
@@ -161,12 +172,7 @@ export interface IScheduleActions {
      * @returns {ITimeout} A interface for manipulating the created timeout.
      * @memberof IScheduleActions
      */
-    timeout(
-        action: () => mixed,
-        period: number,
-        autoSchedule?: boolean,
-        resolveOnClear?: boolean,
-    ): ITimeout;
+    timeout(action: () => mixed, period: number, options?: Options): ITimeout;
```
somewhatabstract, line 175: Having strongly typed options with named values should make it a lot easier to interpret calls instead of looking into what the last two boolean values mean.

jeresig (approved):
Looks ok to me - will this be a major version bump for the package then?

PR #1184, opened 2021-09-30
WB-962: Upgrade to Popper v2

jandrade:
### General

- Upgrades Wonder Blocks to use the latest Popper version (v2).
- By doing this upgrade, we can solve a long-standing issue that cuts off the dropdown options list container.
- Modified Dropdown, Popover and Tooltip packages to match Popper's v2 API.
- Removed `visibility-modifier.js` as it is no longer needed due to some Popper internal improvements that handle the logic added in wbVisibility.
- Added some new stories to verify that changes don't introduce any regressions in the existing packages.

### Tooltip

- Changed the logic to keep the tooltip visible when the user hovers on the tooltip bubble.
- Removed `onTimeoutChanged` from `TooltipAnchor`.
- Updated some integration tests to match the changes.
- Migrated tooltip.integration.test.js to use RTL.

Issue: https://khanacademy.atlassian.net/browse/WB-962

## Test plan:

Navigate to all the examples in Styleguidist and Storybook that use Popper (Tooltips, Popovers and Dropdowns) and verify that we don't introduce any regresions.

https://user-images.githubusercontent.com/843075/135547709-f5e74c7b-f9ab-4174-9144-53c5858a4b20.mov

https://user-images.githubusercontent.com/843075/135547726-abaaacab-badd-4ca2-87c4-9f7f15fef061.mov


packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -699,15 +695,26 @@ class DropdownCore extends React.Component<Props, State> {
...
-                            boundariesElement: "viewport",
-                            escapeWithReference: true,
+                    modifiers={[
+                        {
+                            name: "preventOverflow",
+                            options: {
+                                rootBoundary: "viewport",
+                                // Allows to overlap the popper in case there's
+                                // no more vertical room in the viewport.
+                                altAxis: true,
```
jandrade, line 705: This is the line that fixes WB-962. This is possible only with v2, and that's why we had to do the rest of the work involved on this PR 😅.

jeresig: Wow, so much work!

shared-unpackaged/visibility-modifier.js
```diff
@@ -1,76 +0,0 @@
-// @flow
```
jeresig, line 1: Great that we were able to delete this!

jeresig (approved):
Wow! Thank you so much for working through all the edge cases here... got to clean up a lot of old code!

package.json
```diff
@@ -103,13 +103,13 @@
     "webpack-cli": "^3.3.9"
   },
   "dependencies": {
+    "@popperjs/core": "^2.10.1",
```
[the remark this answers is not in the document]
jandrade: That's a great question! According to bundlephobia:

Popper v1: 20.7 kB (minified) / 7.3 kB (min + gzipped)
react-popper v1: 46 kB (min) / 15.3 kB (min + gz)

Popper v2: 19.9kB (minified) / 6.9 kB (min + gzipped)
react-popper v2: 5.4 kB (min) / 2 kB (min + gz)

It looks that the major gains will be with the `react-popper` upgrade :).

Side question: what do you think about adding a GH action to verify the bundle size of the packages per PR?
This seems to be pretty popular: https://github.com/ai/size-limit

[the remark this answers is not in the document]
jeresig: Wow! That's a huge improvement! Yeah, something like this for WB would be super useful!

PR #1187, opened 2021-10-01
[Experiment] Add filesize actions

jandrade:
## Summary:
Adding some Github actions to experiment with file sizes in our WB bundles.

Issue: XXX-XXXX

## Test plan:

Verify the PR results in the PR!

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -41,8 +41,6 @@ jobs:
             ${{ runner.OS }}-node_modules-
       - name: yarn install
         run: yarn install
-      - name: yarn build:all
-        run: yarn build:all
       # Check bundle sizes
       - name: Compressed size
         uses: preactjs/compressed-size-action@v2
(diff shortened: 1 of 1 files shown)


kevinbarabash:
How does this action stores the previous bundle sizes?

jandrade:
> How does this action stores the previous bundle sizes?

It does it by building the PR, then building the target (`master`) and comparing between the two

.github/workflows/node-ci.yml
```diff
@@ -48,6 +43,16 @@ jobs:
         run: yarn install
       - name: yarn build:all
         run: yarn build:all
```
jeresig, line 45: I think we can remove this now - since the Compressed size action is doing another build anyway!

.github/workflows/node-ci.yml
```diff
@@ -48,6 +43,16 @@ jobs:
...
+      # Check bundle sizes
+      - name: Compressed size
```
jeresig, line 47: I think a better name might be "Check Builds"?

jeresig (approved):
This is great, thank you for adding this!

PR #1190, opened 2021-10-04
Fix tests and include babel.config.js in PRs

jandrade:
Summary:

- Removed all the `index.test.js` files because we are only testing public APIs.
- Popover: InitialFocus => Changed spread operator to `Array.from` to convert NodeList to Array.
- Actions: Included `babel.config.js` in the `Jest` step to trigger JS tests when that file is changed.

Issue: XXX-XXXX

Test plan:

Verify that the Jest tests pass.
Also verify that Popover receives focus as expected (that means one of the elements inside the Popover is focused once opened).


.github/workflows/node-ci.yml
```diff
@@ -68,7 +68,7 @@ jobs:
       - name: Run jest tests
         uses: Khan/jest-action@v1.2.0
         with:
-          run-all-if-changed: package.json,jest.config.js
+          run-all-if-changed: package.json,jest.config.js,babel.config.js
```
jeresig, line 71: Thank you!!

packages/wonder-blocks-popover/src/components/__tests__/initial-focus.test.js
```diff
@@ -1,6 +1,8 @@
 // @flow
 import * as React from "react";
-import {mount} from "enzyme";
+import {render, screen} from "@testing-library/react";
```
jandrade, line 3: I took the opportunity to migrate these tests to RTL while figuring out the NodeList spread issue.

jeresig (approved):
Very good, thank you for fixing this up!

PR #1215, opened 2021-11-02
FEI-2243: Upgrade react-router-dom to 5.3.0

kevinbarabash:
## Summary:
We'd like to start using the hooks from react-router-dom in webapp but before we can do that we need to upgrade wonder-blocks to use a version of react-router-dom that exports hooks.

This PR also changes how we check to see if we're inside a react-router context.  This is accomplished by using __RouterContext.Provider to get the 'router' context object since this.context is no longer available.

I also added some additional mocking in some places to get rid of some warnings that I saw when I temporarily disabled the log-on-fail-reporter when doing some debugging.

Issue: FEI-2243

## Test plan:
- yarn jest


jeresig (approved):
Too bad that we have to depend on react-router now, as well - but thanks for fixing this up!

packages/wonder-blocks-button/src/components/button-core.js
```diff
@@ -2,7 +2,7 @@
 import * as React from "react";
 import {StyleSheet} from "aphrodite";
 import {Link} from "react-router-dom";
-import * as PropTypes from "prop-types";
```
jandrade, line 5: oh I totally missed that! Kinda related question: why do we have `prop-types` as a peer dep in some of the packages?

packages/wonder-blocks-clickable/package.json
```diff
@@ -23,7 +23,8 @@
     "prop-types": "^15.6.2",
```
jandrade, line 23: For example here: `prop-types`.

jandrade (approved):
LGTM! Just added a question about `prop-types` that doesn't need to be addressed here.

kevinbarabash:
I'll remove `"prop-types"` as a dep before landing and publishing.

packages/wonder-blocks-button/src/components/button-core.js
kevinbarabash: It's no longer needed.  Thanks for flagging this.  I've removed it as a peerDep now.

PR #1217, opened 2021-11-08
Add module.name_mapper entry in .flowconfig

kevinbarabash:
## Summary:
Without this people would have build packages first.  Also, the build packages could be out of sync with other local changes which is confusing.  This avoids the issue, but have 'yarn flow' work out of the box (well after running 'yarn install').

Issue: None

## Test plan:
- yarn clean followed by yarn flow


jeresig (approved):
oh interesting, that's handy! thank you

PR #1218, opened 2021-11-08
LabeledTextField: Allow passing `aria-describedby` as a prop.

nishasy:
## Summary:
Here, we allow the user to pass in aria-describedby as a prop,
but default to the UniqueIdProvider error otherwise.

Issue: https://khanacademy.atlassian.net/browse/WB-1115

## Test plan:
`yarn flow`
`yarn jest packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js`

### DIFF
--- a/packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
+++ b/packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
@@ -1,6 +1,7 @@
 //@flow
 import * as React from "react";
 import {mount} from "enzyme";
+import {render, screen} from "@testing-library/react";
 
 import {StyleSheet} from "aphrodite";
 import LabeledTextField from "../labeled-text-field.js";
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -10,6 +14,8 @@ import TextField, {type TextFieldType} from "./text-field.js";
 type WithForwardRef = {|forwardedRef: React.Ref<"input">|};
 
 type Props = {|
+    ...AriaProps,
```
kevinbarabash, line 17: This means that all aria props would be accepted, but we're only passing down `aria-describedby`.

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -219,7 +225,11 @@ class LabeledTextFieldInternal extends React.Component<
                         field={
                             <TextField
                                 id={`${uniqueId}-field`}
-                                aria-describedby={`${uniqueId}-error`}
+                                aria-describedby={
+                                    ariaDescribedby
+                                        ? ariaDescribedby
```
jandrade, line 230: Could you please add a unit test to verify this case? This will help us in the long term to identify any possible errors in case we make any changes.

packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
```diff
@@ -186,6 +187,49 @@ describe("LabeledTextField", () => {
...
+                label="Label"
+                value=""
+                onChange={() => {}}
+                ariaDescribedby={ariaDescription}
+            />,
+        );
+
+        // Assert
+        const input = screen.getByRole("textbox");
+        expect(input.getAttribute("aria-describedby")).toEqual(ariaDescription);
```
jandrade, line 206: Nit: You can also use a custom matcher `expect(input).toHaveAttribute("aria-describedby", ariaDescription)`. No need to change it, but just wanted to share in case you find it helpful.

jandrade (approved):
LGTM! thank you for adding tests 🎉

PR #1219, opened 2021-11-10
FEI-4048: Add useUniqueId* and useRenderState hooks

kevinbarabash:
## Summary:
I ended up creating multiple hooks for useUniqueId since I wasn't able to get function overloading working with flow.  We can replace these with a single useUniqueId once we transition to TypeScript.  Since WithUniqueId uses WithSSRPlaceholder internally, I need ended up creating a useRenderState hook as well.

Note: while this PR adds some bytes to our bundles, this is only because this PR doesn't update existing components to use these new hooks.  This will happen once the majority of tests in webapp have been updated to use `@testing-library`.  The hope of migrating the tests first is that it'll simplify using components that use hooks.

Issue: FEI-4048

## Test plan:
- yarn jest use-unique-id.test.js
- look at the [docs](https://5e1bf4b385e3fb0020b7073c-hecvguyszk.chromatic.com/?path=/docs/core-useuniqueid--without-mock-example) published to chromatic


kevinbarabash:
This is ready for review now.  Please ignore the netlify build errors.  It's a config error.

.nvmrc
```diff
@@ -0,0 +1 @@
+12
```
kevinbarabash, line 1: This was necessary to fix the netlify build.

packages/wonder-blocks-core/src/hooks/use-unique-id.stories.mdx
```diff
@@ -0,0 +1,130 @@
...
+<Canvas>
+    <Story name="WithMockExample">
+        <WithMockExample />
+    </Story>
+</Canvas>
+
+# Scoped
+
+Both of these hooks ensure that every identifier factory is unique using a
+unique number for each on. However, this isn't very readable when wanting to
```
jaredly, line 93:
```suggestion
unique number for each one. However, this isn't very readable when wanting to
```

packages/wonder-blocks-core/src/hooks/use-render-state.js
```diff
@@ -0,0 +1,24 @@
...
+ * This is a thin wrapper around React.useContext(RenderStateContext).
+ *
+ * RenderStateContext is used by WithSSRPlaceholder to hide a component from
+ * rendered during SSR.  It's also used by useUniqueId(With|Without)Mock to
+ * to return a UniqueIDFactory instance in a SSR-safe way.
+ *
+ * This hook isn't part of wonder-blocks-core's public interface since we still
+ * need to use WithSSRPlaceholder to update the value in the context.
+ *
+ * TODO(kevinb): Figure out whether we should always render a WithSSRPlaceholder
```
jeresig, line 12: Jira issue to track this?

packages/wonder-blocks-core/src/hooks/use-unique-id.js
```diff
@@ -0,0 +1,65 @@
...
+ * Returns a unique identifier factory.  If the parent component hasn't
+ * been mounted yet, the global SsrIDFactory will be returned until the
+ * component becomes mounted.
+ *
+ * @param {string} [scope] optional string to prefix generated ids with.
+ * @returns {IIdentifierFactory}
+ */
+export const useUniqueIdWithMock = (scope?: string): IIdentifierFactory => {
+    const renderState = useRenderState();
+    const [isMounted, setIsMounted] = React.useState(false);
```
jeresig, line 20: Huh, do we have to use `React.` for all hooks? Is there a way to avoid that?

packages/wonder-blocks-core/src/hooks/use-unique-id.md
```diff
@@ -0,0 +1,41 @@
+### useUniqueIdWithoutMock (hook)
```
jeresig, line 1: Hmm, are you missing docs for `useUniqueIdWithMock`?

jeresig (approved):
Looks good to me!

packages/wonder-blocks-core/src/hooks/use-unique-id.md
kevinbarabash: I need to remove this file.  All the docs are in use-unique-id.stories.mdx.  I couldn't get the hooks examples to work in styleguidist.  I guess I could just have the text parts in the .md file and leave out the examples.

packages/wonder-blocks-core/src/hooks/use-unique-id.js
kevinbarabash: I can do named imports.  I went with `React.` because we do namespace imports of "react" everywhere in wonder-blocks.

packages/wonder-blocks-core/src/hooks/use-unique-id.md
kevinbarabash: I ended up duplicating the docs without the examples from use-unique-id.stories.mdx.

PR #1214, opened 2021-10-21
WB-1109: Migrate Button docs to Storybook

jandrade:
## Summary:
First part of the Storybook migration.

- This PR covers converting all the existing Button docs to Storybook.
- It also creates a new `mdx` page to include the best practices section
  in a single sub-page inside `Button`.

Issue: https://khanacademy.atlassian.net/browse/WB-1109

Test plan:

Verify the changes in the Storybook instance generated by chromatic.
 https://5e1bf4b385e3fb0020b7073c-xyuiwjfhdi.chromatic.com/?path=/story/navigation-button--default-button

Also verify that the Docs pages are generated correctly:
- [Examples](https://5e1bf4b385e3fb0020b7073c-xyuiwjfhdi.chromatic.com/?path=/docs/navigation-button--default-button) (includes props table and descriptions for a lot of the examples).
- [Accessibility](https://5e1bf4b385e3fb0020b7073c-xyuiwjfhdi.chromatic.com/?path=/docs/navigation-button-accessibility--labeling)
- [Best practices](https://5e1bf4b385e3fb0020b7073c-xyuiwjfhdi.chromatic.com/?path=/docs/navigation-button-best-practices--full-bleed-button)
Screen Shot 2021-11-01 at 11 26 43 AM

PR #1220, opened 2021-11-12
Improve navigation callback documentation

aag:
## Summary:
This commit adds additional documentation describing the `onClick`,
`beforeNav`, and `safeWithNav` props. These props have been available in
Wonder Blocks for a while, but there has been some confusion about when to
use each one. This commit attempts to add enough detail about these
props to make it clear what the use case is for each prop.

Additionally, this commit changes the section headings in button.md from
h4 to h3 elements. This avoids a gap in heading levels and also brings the
section in line with other sections on the page.

Issue: none

## Test plan:
- Load the documentation and read the Button, Clickable, and Link
  sections.

Button
image

Clickable
image

Link
image

Storybook Documentation Page
image

Storybook Examples
image

### DIFF
--- a/packages/wonder-blocks-button/src/components/__docs__/navigation-callbacks.stories.mdx
+++ b/packages/wonder-blocks-button/src/components/__docs__/navigation-callbacks.stories.mdx
@@ -1,0 +1,59 @@
+import {Meta, Story, Canvas} from "@storybook/addon-docs";
+import {StyleSheet} from "aphrodite";
+
+import Button from "@khanacademy/wonder-blocks-button";
+import {View} from "@khanacademy/wonder-blocks-core";
+
+<Meta
+    title="Navigation/Button/Navigation Callbacks"
+    component={Button}
+    parameters={{
+        previewTabs: {
+            canvas: {hidden: true},
+        },
+        viewMode: "docs",
+        chromatic: {
+            // Disables chromatic testing for these stories.
+            disableSnapshot: true,
+        },
+    }}
+/>
+
+## Running Callbacks on Navigation
+
+Sometimes you may need to run some code and also navigate when the user
+clicks the button. For example, you might want to send a request to the
+server and also send the user to a different page. You can do this by
+passing in a URL to the `href` prop and also passing in a callback
+function to either the `onClick`, `beforeNav`, or `safeWithNav` prop.
+Which prop you choose depends on your use case.
+
+- `onClick` is guaranteed to run to completion before navigation starts,
+  but it is not async aware, so it should only be used if all of the code
+  in your callback function executes synchronously.
+
+- `beforeNav` is guaranteed to run async operations before navigation
+  starts. You must return a promise from the callback function passed in
+  to this prop, and the navigation will happen after the promise
+  resolves. This prop should be used if it's important that the async code
+  completely finishes before the next URL starts loading.
+
+- `safeWithNav` runs async code concurrently with navigation when safe,
+  but delays navigation until the async code is finished when
+  concurrent execution is not safe. You must return a promise from the
+  callback function passed in to this prop, and Wonder Blocks will run
+  the async code in parallel with client-side navigation or while opening
+  a new tab, but will wait until the async code finishes to start a
+  server-side navigation. This prop should be used when it's okay to load
+  the next URL while the async callback code is running.
+
+This table gives an overview of the options:
+
+| Prop         | Async safe? | Completes before navigation? |
+|--------------|-------------|------------------------------|
+| onClick      | no          | yes                          |
+| beforeNav    | yes         | yes                          |
+| safeWithNav  | yes         | no                           |
+
+
+
--- a/packages/wonder-blocks-button/src/components/__docs__/navigation-callbacks.stories.mdx
+++ b/packages/wonder-blocks-button/src/components/__docs__/navigation-callbacks.stories.mdx
@@ -35,7 +35,8 @@ Which prop you choose depends on your use case.
 - `beforeNav` is guaranteed to run async operations before navigation
   starts. You must return a promise from the callback function passed in
   to this prop, and the navigation will happen after the promise
-  resolves. This prop should be used if it's important that the async code
+  resolves. If the promise rejects, the navigation will not occur.
+  This prop should be used if it's important that the async code
   completely finishes before the next URL starts loading.
 
 - `safeWithNav` runs async code concurrently with navigation when safe,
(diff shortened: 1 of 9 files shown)


packages/wonder-blocks-button/src/components/button.md
```diff
@@ -420,12 +420,47 @@ const styles = StyleSheet.create({
     </View>
 </MemoryRouter>
 ```
+### Running callbacks on navigation
```
jandrade, line 423: This is really great! and totally fine to leave it here until we finish the docs migration.

Could you please replicate this section in Storybook?

Probably this could be an option: https://github.com/Khan/wonder-blocks/blob/master/packages/wonder-blocks-button/src/components/__docs__/best-practices.stories.mdx

But if you think this deservers to be on its own section, then we can create a new `.mdx` file in that folder to show this.

jeresig, line 423: This is great!

packages/wonder-blocks-button/src/components/button.md
```diff
@@ -452,7 +487,7 @@ const styles = StyleSheet.create({
                 setTimeout(resolve, 1000);
             })}
         >
-            Async action, client-side nav
+            beforeNav, client-side nav
```
jandrade, line 490: Probably it would be a good idea too to move the current async stories next to this description. Those stories are defined here: 

https://github.com/Khan/wonder-blocks/blob/master/packages/wonder-blocks-button/src/components/button.stories.js#L356-L411

jeresig (approved):
Wow, thank you for improving this, it looks great!

jandrade (changes requested):
Thanks for improving the documentation!

Changes look great, but I'm requesting changes because we recently started migrating documentation over to Storybook, and `Button` was recently migrated in #1214.

The idea is to keep the Styleguidist documentation as it is until we finish the Storybook migration, then we officially switch the website to SB. That means that as soon as we port one section, we stop maintenance on the other one (e.g. adding new documentation, adding more examples). 

Changes on `Clickable` and `Link` are totally fine, because we'll migrate these docs in the next couple of weeks.

`Button` docs changes in Styleguidist are fine if we keep them, but I'm requesting the following changes just to replicate the docs on the Storybook side:

- Move the new async documentation to a Storybook MDX file (e.g. `__docs__/best-practices.stories.mdx` or even into a new MDX file inside `__docs__`).
- Update the examples in `button.stories.js`.

Sorry for the confusion around this. Happy to chat more about that in case you have any questions or suggestions.

jandrade:
Couple of extra things that I forgot to mention:

You can see the results of this PR here: https://5e1bf4b385e3fb0020b7073c-motrxagwqb.chromatic.com/

If you go to the `Checks` section and scroll down, you'll see a `Storybook Publish` check with a link to this version.

You can run `storybook` locally with `yarn start:storybook`.

packages/wonder-blocks-button/src/components/button.md
```diff
@@ -420,12 +420,47 @@ const styles = StyleSheet.create({
...
 </MemoryRouter>
 ```
+### Running callbacks on navigation
+
+Sometimes you may need to run some code and also navigate when the user
+clicks the button. For example, you might want to send a request to the
+server and also send the user to a different page. You can do this by
+passing in a URL to the `href` prop and also passing in a callback
+function to either the `onClick`, `beforeNav`, or `safeWithNav` prop.
+Which prop you choose depends on your use case.
```
kevinbarabash, line 430: Only passing one of these callbacks is super important so it might warrant calling this out with a special note or warning.

packages/wonder-blocks-button/src/components/button.md
```diff
@@ -420,12 +420,47 @@ const styles = StyleSheet.create({
...
+  server-side navigation. This prop should be used when it's okay to load
+  the next URL while the async callback code is running.
+
+This table gives an overview of the options:
+
+| Prop         | Async safe? | Completes before navigation? |
+|--------------|-------------|------------------------------|
+| onClick      | no          | yes                          |
+| beforeNav    | yes         | yes                          |
+| safeWithNav  | yes         | no                           |
```
kevinbarabash, line 457: I love this table.

kevinbarabash (commented):
Thanks for improving the docs around these props.  Having docs that actually appear in styleguidist/storybook is way more helpful than source-only docs. 😅

packages/wonder-blocks-button/src/components/button.md
aag: Can you say more about this? The Flow types explicitly allow any combination of the 3 props, and code comments document what happens if you pass both `beforeNav` and `safeWithNav`. Why is it important to only pass one of them?

packages/wonder-blocks-button/src/components/__docs__/navigation-callbacks.stories.mdx
```diff
@@ -0,0 +1,59 @@
...
+This table gives an overview of the options:
+
+| Prop         | Async safe? | Completes before navigation? |
+|--------------|-------------|------------------------------|
+| onClick      | no          | yes                          |
+| beforeNav    | yes         | yes                          |
+| safeWithNav  | yes         | no                           |
+
+
+
```
jandrade, line 59: Thank you so much for doing this!!

Just in case it's helpful for doc purposes.... if you want to display the story here, you can pass the `id` of the existing story so it can be embedded on this view. Of course, no need to change this, but just wanted to share it :).

e.g.
```
<Story id="navigation-button--before-nav-callbacks" />
```

jandrade (approved):
LGTM! Thank you for adding the Storybook docs 🚀.

packages/wonder-blocks-button/src/components/button.md
aag: I dug into the code some and determined that it is indeed safe for multiple callbacks to be used together, then updated the documentation with the details of how they interact.

packages/wonder-blocks-button/src/components/__docs__/navigation-callbacks.stories.mdx
```diff
@@ -44,16 +45,24 @@ Which prop you choose depends on your use case.
...
+| onClick     | no          | yes                          |
+| beforeNav   | yes         | yes                          |
+| safeWithNav | yes         | no                           |
 
+It is possible to use more than one of these props on the same element.
+If multiple props are used, they will run in this order: first `onClick`,
+then `beforeNav`, then `safeWithNav`. If both `beforeNav` and `safeWithNav`
+are used, the `safeWithNav` callback will not be called until the
+`beforeNav` promise resolves successfully. If the `beforeNav` promise
+rejects, `safeWithNav` will not be run.
```
kevinbarabash, line 65: My vague recollection of the props was incorrect.  While we are using disjoint props, it's `_target` and `beforeNav` that can't be used together.

Thanks for making it explicit that these props can be used together and the ordering in which they're run.

kevinbarabash (approved):
LGTM.  :shipit:

PR #1221, opened 2021-11-16
WB-586: Add changesets to automate the release process

jandrade:
## Summary:

- Introduces `changesets` to the repository.
- Adds a new workflow to handle releases. This will be triggered after
  the "release coordinator" (one of WB's admins) has approved the
  release (by hitting the `Merge` button).
- Creates a new `release` script to handle releases using `changesets`
  instead of `lerna`. (keeping `lerna/publish` for backwards compatibility).

More info about the new release process here:
https://www.figma.com/file/xaYyzIuKLxqXAJVAl3y1S4/Wonder-Blocks-Web-Eng-process?node-id=50%3A400
Wonder Blocks Web Eng process

NOTE: To include a regular PR in the release PR, the PR's author will
need to run `yarn changesets` to include the changes for each package
they have changed.

More info about using it on this part of the process here: https://www.figma.com/file/xaYyzIuKLxqXAJVAl3y1S4/Wonder-Blocks-Web-Eng-process?node-id=81%3A549

Issue: WB-586

## Test plan:

Merge this PR and verify that the "Release PR" gets created after a
contributor lands a PR (merges it into `master`).

Here's an example of how the `Release PR` will look: https://github.com/emotion-js/emotion/pull/2394


.github/workflows/release.yml
```diff
@@ -0,0 +1,49 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - master
```
kevinbarabash, line 6: Does this mean if a person wants to batch work together from multiple PRs before release, they'll need to land those PRs on to a non-master branch?  If so, what impact does this have on our required checks?

.changeset/config.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@1.6.2/schema.json",
+  "changelog": "@changesets/cli/changelog",
+  "commit": false,
+  "linked": [],
+  "access": "public",
+  "baseBranch": "master",
+  "updateInternalDependencies": "patch",
```
kevinbarabash, line 8: Does this take into account whether that internal dep was bump by a patch, minor, or major version number?

.changeset/config.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@1.6.2/schema.json",
+  "changelog": "@changesets/cli/changelog",
```
kevinbarabash, line 3: Are there different tools that can be used to generate a change log file?

.github/workflows/release.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Send a Slack notification if a publish happens
+      - uses: rtCamp/action-slack-notify@v2
+        if: steps.changesets.outputs.published == 'true'
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+          SLACK_CHANNEL: wonder-blocks-eng
+          SLACK_MSG_AUTHOR: ${{ github.event.pull_request.user.login }}
+          SLACK_USERNAME: GithubGoose
+          SLACK_ICON_EMOJI: ":goose:"
```
kevinbarabash, line 45: lol

.github/workflows/release.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Setup Node.js 12.x
+        uses: actions/setup-node@master
+        with:
+          node-version: 12.x
+
+      - name: Install Dependencies
+        run: yarn
+
+      - name: Create Release Pull Request or Publish to npm
```
kevinbarabash, line 28: Given that this is running `yarn release`, should the name of this maybe be `Publish to npm`.

package.json
```diff
@@ -21,7 +21,8 @@
     "jest": "jest --config config/jest/test.config.js",
     "lint:watch": "esw --watch --config ./eslint/eslintrc packages",
     "lint": "eslint --config .eslintrc.js packages",
-    "publish": "node utils/pre-publish-check.js && git diff --stat --exit-code HEAD && yarn build:cjs && yarn build:es6 && lerna publish",
+    "publish": "node utils/pre-publish-check.js && git diff --stat --exit-code HEAD && yarn build:all && lerna publish",
+    "release": "node utils/pre-publish-check.js && git diff --stat --exit-code HEAD && yarn build:all && changeset publish",
```
kevinbarabash, line 25: Can we rename these to `publish:lerna` and `publish:changesets`?  Someone looking at this might think that these two commands do different things, but really they both publish, but use different tools to do so.

kevinbarabash (commented):
I think it would be good to have an ADR to support this change.  `changesets` looks like a great tool, but we should have documentation in confluence explaining why we picked this tool over other tools.

jandrade:
Here's the ADR: https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/1878523905/ADR+523+Automating+the+release+process+on+Wonder+Blocks

.github/workflows/release.yml
jandrade: hmm good question, for more clarification... do you mean about required checks at PR level? or when pushing to `master`?

package.json
jandrade: I'm going to use `publish:ci` so we follow the `*:ci` pattern that @somewhatabstract recently introduced in the repo :).

.github/workflows/release.yml
kevinbarabash: I originally thought the release process was automatically like it is with semantic-release, but that isn't the case with changesets because someone needs to create a release PR.

kevinbarabash: I chatted with Juan in Slack and learned that the action runs once to generate the PR if there's any changeset markdown files.  This PR converts those markdown files to changelogs.  Once the PR lands then the action runs a second time and will do the publish since there are no pending changeset markdown files.

kevinbarabash (approved):
LGTM.  Thanks for taking the time to walk me through the process.  I think I have a better understanding of it now.

PR #1223, opened 2021-11-17
WB-1121: Add chromatic GH action back on dependabot PRs

jandrade:
## Summary:

There was an issue with the Chromatic action (CLI) that didn't allow
skipping tests using the --skip flag (or `skip` env var in this case).

This now can be added back because the fix was applied on the Chromatic CLI repo.

Issue: WB-1121

## Test plan:

- Merge this PR.
- Comment `@dependabot recreate` on one of the dependabot-generated PRs.
- Verify that the Chromatic tests are skipped and the PR can be landed automatically.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -48,4 +48,9 @@ jobs:
     # (else) Run this step on dependabot PRs.
     - name: Skip Chromatic builds
       if: ${{ github.actor == 'dependabot[bot]' }}
-      run: yarn chromatic --skip
+      uses: chromaui/action@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        projectToken: ${{ secrets.CHROMATIC_APP_CODE }}
+        exitZeroOnChanges: true
+        skip: true
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -53,4 +53,7 @@ jobs:
         token: ${{ secrets.GITHUB_TOKEN }}
         projectToken: ${{ secrets.CHROMATIC_APP_CODE }}
         exitZeroOnChanges: true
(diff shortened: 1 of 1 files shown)


kevinbarabash:
Do we actually want Chromatic running on all of the Dependabot PRs?  I think my original reason for stopping it from running was that I didn't want Dependabot using up all of our free snapshots.  Is not running it causing issues?

kevinbarabash:
> Verify that the Chromatic tests are skipped and the PR can be landed automatically.

Thanks for including this in the test plan.

kevinbarabash (approved):
LGTM.  I'm glad we're able to use the Chromatic Action again.

jeresig (approved):

.github/workflows/chromatic.yml
```diff
@@ -48,4 +48,9 @@ jobs:
     # (else) Run this step on dependabot PRs.
     - name: Skip Chromatic builds
       if: ${{ github.actor == 'dependabot[bot]' }}
-      run: yarn chromatic --skip
+      uses: chromaui/action@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        projectToken: ${{ secrets.CHROMATIC_APP_CODE }}
+        exitZeroOnChanges: true
+        skip: true
```
somewhatabstract, line 56: Should this still be there? What is it skipping? Can you perhaps add a comment so that it's clear why we need this?

somewhatabstract (approved):
Thanks for getting this working again. Very exciting.

.github/workflows/chromatic.yml
jandrade: Fair enough! It's added so the Chromatic build reports back to GH, but that we don't need to process any stories inside this PR. That saves running some unnecessary commands (that means in saving time when running this workflow on these PRs).

PR #1225, opened 2021-11-22
FEI-4022: implement useTimeout hook

kevinbarabash:
## Summary:
This hook follows similar patterns to ActionScheduler.  I had to tweak things a little be since we're dropping the need for the scheduler itself with this hook.

I've also added lint rules for hooks in this PR.  I'll create a separate PR with that change and rebase this one on top of it.

Issue: FEI-4022

## Test plan:
- yarn jest use-timeout.test.js
- yarn start:storybook
- load http://localhost:6061/?path=/docs/timing-usetimeout
- see that the timeout starts immediately with the first example
- see that the timeout doesn't start until you click 'set' in the seconde example
- play around with setting and clearing the timeouts


packages/wonder-blocks-timing/src/hooks/use-timeout.js
```diff
@@ -0,0 +1,73 @@
...
+                window.clearTimeout(timeout);
+                if (!mountedRef.current) {
+                    clear();
+                }
+            };
+        }
+    }, [clear, isActive, timeoutMs]);
+
+    return {
+        get isSet() {
```
jeresig, line 67: is `isSet` clearer than `isActive`? Is there a harm in just doing: `isSet: isActive`?

jeresig (approved):
Looks good to me - thanks for the clear docs and tests!

packages/wonder-blocks-timing/src/hooks/use-timeout.js
kevinbarabash: `isSet` conforms to the `ITimeout` interface.  I going to rename `isActive` to `isSet`.  If we get feedback from devs that this name is confusing, we should rename it `ITimeout`.

I'm going to drop the getter method and use `isSet: isActive` instead.  The tests pass just fine after that change.  I'm also adding an additional test case to make sure that `isSet` is false after the timer expires.

package.json
```diff
@@ -88,7 +89,7 @@
     "react-docgen": "^5.0.0",
     "react-router-enzyme-context": "^1.2.0",
     "react-styleguidist": "^11.0.5",
-    "react-test-renderer": "^16.4.1",
+    "react-test-renderer": "^16.9.0",
```
kevinbarabash, line 92: This was needed by `@testing-library/react-hooks`.

jandrade (approved):
LGTM! Just a minor suggestion for code formatting but the implementation and tests look great!.

packages/wonder-blocks-timing/src/hooks/use-timeout.stories.mdx
```diff
@@ -0,0 +1,152 @@
...
+        },
+    }}
+/>
+
+# `useTimeout`
+
+`useTimeout` is a hook that provides a convenient API for setting and clearing
+a timeout. It is defined as follows:
+
+```
```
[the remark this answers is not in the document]
kevinbarabash: I ended up going with `jsx` since it does a better job of highlighting things.

PR #1226, opened 2021-11-23
Add lint rules for hooks

kevinbarabash:
## Summary:
There were a bunch of lint because the rule expects components to use PascalCase instead of camelCase.  This PR updates the names of all the stories where that was an issue.  Hopefully that doesn't result in the names of snapshots in Chromatic changing. :crossed_fingers:

Issue: None

## Test plan:
- yarn lint
- yarn start:storybook
- see that the stories are still there

### DIFF
--- a/packages/wonder-blocks-form/src/components/text-field.stories.js
+++ b/packages/wonder-blocks-form/src/components/text-field.stories.js
@@ -2,7 +2,7 @@
 import * as React from "react";
 import {StyleSheet} from "aphrodite";
 
-import {View, Text} from "@khanacademy/wonder-blocks-core";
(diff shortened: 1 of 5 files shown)


packages/wonder-blocks-form/src/components/text-field.stories.js
```diff
@@ -15,7 +15,7 @@ export default {
     title: "Form / TextField",
 };
 
-export const text: StoryComponentType = () => {
+export const Text: StoryComponentType = () => {
```
jeresig, line 18: Why not name this one `TextStory` and leave the other as Text?

jeresig (approved):

jandrade:
Just for the posteriority, this is a great idea! thanks!

PR #1229, opened 2021-11-24
[🔥AUDIT🔥] Address webpack not liking optional chaining

somewhatabstract:
🖍 _This is an audit!_ 🖍

## Summary:
Webpack v4 does not support optional chaining. When building for node 14
or above, babel does not transpile optional chaining away, which means
the webpack build fails, so for webpack's sake, we fix the target node
to 12, even if running on a different version. Long term, either
webpack needs updating or replacing (for example, why not produce a cjs
build using rollup instead?).

Issue: XXX-XXXX

## Test plan:
Merge on master and watch the magic happen


build-settings/babel.config.js
```diff
@@ -9,7 +9,7 @@ module.exports = (api) => {
             {
                 targets: api.env("test")
                     ? {
-                          node: "current",
+                          node: 12,
```
kevinbarabash, line 12: Tricky.

kevinbarabash (commented):
LGTM.  Thanks for fixing this.

PR #1230, opened 2021-11-24
Enable flow enums

somewhatabstract:
## Summary:
- updates flow, eslint, babel, jest, storybook, and more.
- gets linters and tests passing
- adds flow enum support to babel and the root package.json
- removes some unused code (webapp doesn't use it either)
- changed render state to use an enum in Wonder Blocks Core

Issue: FEI-4138

## Test plan:
`yarn build:all`
`yarn test`
`yarn lint`
`yarn flow`


somewhatabstract:
The coverage being 0 is disconcerting. Taking a look at that

somewhatabstract:
OK, so the jest and jest coverage actions are not properly detecting that all tests need running.

packages/wonder-blocks-timing/src/hooks/__tests__/use-timeout.test.js
```diff
@@ -42,7 +42,9 @@ describe("useTimeout", () => {
             renderHook(() => useTimeout(action, 1000));
 
             // Act
-            jest.advanceTimersByTime(1000);
+            act(() => {
+                jest.advanceTimersByTime(1000);
+            });
```
kevinbarabash, line 47: I didn't realize that `renderHook` was actually rendering a real component.  I had a quick look at its code and it is.  I wonder why we didn't need the `act()` wrappers before. 🤔

packages/wonder-blocks-timing/src/util/__tests__/interval.test.js
```diff
@@ -114,12 +124,15 @@ describe("Interval", () => {
...
         it("should invoke setInterval to call the given action", () => {
             // Arrange
+            const intervalSpy = jest.spyOn(global, "setInterval");
             const action = jest.fn();
-            const interval = new Interval(() => action(), 500);
+            const interval = new Interval(
+                () => action(),
+                500,
+                SchedulePolicy.OnDemand,
+            );
```
kevinbarabash, line 133: Why the switch to `OnDemand` for this test?

packages/wonder-blocks-timing/src/util/__tests__/interval.test.js
```diff
@@ -99,13 +104,18 @@ describe("Interval", () => {
     describe("#set", () => {
         it("should call setInterval", () => {
             // Arrange
-            const interval = new Interval(() => {}, 500);
+            const intervalSpy = jest.spyOn(global, "setInterval");
+            const interval = new Interval(
+                () => {},
+                500,
+                SchedulePolicy.OnDemand,
+            );
```
kevinbarabash, line 112: Why the switch to `OnDemand` for this test?

packages/wonder-blocks-timing/src/util/__tests__/interval.test.js
```diff
@@ -131,12 +144,16 @@ describe("Interval", () => {
         it("should clear the active interval", () => {
             // Arrange
             const action = jest.fn();
-            const interval = new Interval(() => action(), 500);
+            const interval = new Interval(
+                () => action(),
+                500,
+                SchedulePolicy.OnDemand,
+            );
```
kevinbarabash, line 151: Why the switch to `OnDemand` for this test?

packages/wonder-blocks-timing/src/util/__tests__/timeout.test.js
```diff
@@ -110,12 +116,15 @@ describe("Timeout", () => {
...
         it("should invoke setTimeout to call the given action", () => {
             // Arrange
+            const timeoutSpy = jest.spyOn(global, "setTimeout");
             const action = jest.fn();
-            const timeout = new Timeout(() => action(), 500);
+            const timeout = new Timeout(
+                () => action(),
+                500,
+                SchedulePolicy.OnDemand,
+            );
```
kevinbarabash, line 125: Why the switch to `OnDemand` for this test?

kevinbarabash (approved):
LGTM.  Thanks for upgrading so many things.  There was quite a bit to review though so upgrading things separately when possible is preferred.  No action req'd for this PR.  I left a few questions and suggestions but nothing blocking.  :shipit:

packages/wonder-blocks-timing/src/util/__tests__/interval.test.js
somewhatabstract: Because that was what the test was written to work with, but it didn't use it and so it fails.

packages/wonder-blocks-timing/src/util/__tests__/interval.test.js
somewhatabstract: Because that was what the test was written to work with, but it didn't use it and so it fails.

packages/wonder-blocks-timing/src/util/__tests__/interval.test.js
somewhatabstract: Because that was what the test was written to work with, but it didn't use it and so it fails.

packages/wonder-blocks-timing/src/util/__tests__/timeout.test.js
somewhatabstract: Because that was what the test was written to work with, but it didn't use it and so it fails.

somewhatabstract:
> There was quite a bit to review though so upgrading things separately when possible is preferred. No action req'd for this PR.

Agreed. Sadly, there was no easy path to do that since everything was so intrinsically linked - at least, not an easy path to see from the starting point.

.storybook/.babelrc.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Storybook doesn't seem to like the "exports a function" style of babel
+ * config. And it pics up our root .babelrc.js file that is added to workaround
+ * jest. So, we need to provide its own.
+ *
+ * We should remove this when jest is fixed.
+ * See FEI-4139 in Jira.
```
jandrade, line 7: Thanks for the comment!

packages/wonder-blocks-dropdown/src/components/__tests__/dropdown-core-virtualized.test.js
```diff
@@ -24,6 +26,12 @@ describe("DropdownCoreVirtualized", () => {
     });
 
     afterEach(() => {
+        // We have to explicitly unmount before clearing mocks, otherwise jest
+        // timers will throw because we'll try to clear an animation frame that
+        // we set with a setTimeout but are clearing with clearAnimationFrame
+        // because we restored the clearAnimationFrame mock (and we won't
+        // have cleared the timeout we actually set!)
+        unmountAll();
```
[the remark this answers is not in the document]
somewhatabstract: Because of what the comment says.

```
        // We have to explicitly unmount before clearing mocks, otherwise jest
        // timers will throw because we'll try to clear an animation frame that
        // we set with a setTimeout but are clearing with clearAnimationFrame
        // because we restored the clearAnimationFrame mock (and we won't
        // have cleared the timeout we actually set!)
```

package.json
```diff
@@ -35,76 +35,77 @@
...
-    "jest": "^25.1.0",
-    "jest-enzyme": "^7.0.2",
+    "jest": "^27.3.1",
+    "jest-enzyme": "^7.1.2",
     "lerna": "^3.16.5",
     "lint-staged": "^10.2.2",
     "marked": "1.1.0",
     "mkdirp": "^1.0.4",
-    "prettier": "^2.0.5",
+    "prettier": "^2.4.1",
```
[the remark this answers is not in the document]
jandrade: +++

build-settings/babel.config.js
```diff
@@ -26,30 +26,20 @@ module.exports = (api) => {
...
          * https://babeljs.io/docs/en/babel-plugin-transform-flow-strip-types
          */
         "@babel/plugin-transform-flow-strip-types",
 
-        /**
-         * Adds support for the class properties proposal.
-         * https://babeljs.io/docs/en/babel-plugin-proposal-class-properties
-         */
         [
-            "@babel/plugin-proposal-class-properties",
```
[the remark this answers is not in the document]
somewhatabstract: Yup :)

packages/wonder-blocks-data/package.json
```diff
@@ -24,4 +24,4 @@
   },
   "author": "",
   "license": "MIT"
-}
+}
```
jandrade, line 27: Out of curiosity: Did prettier add this change? or was it manually included?

build-settings/babel.config.js
```diff
@@ -60,12 +50,12 @@ module.exports = (api) => {
...
         // needed in production because webpack converts dynamic
         // imports for us.
-        plugins.unshift("dynamic-import-node");
+        plugins.push("dynamic-import-node");
     } else {
         // If we're building for non-test then we want to transform the babel
         // runtime stuff so that it's all included from a single shared module.
         // This helps to cut down on the code repetition and save bytes.
-        plugins.unshift([
+        plugins.push([
```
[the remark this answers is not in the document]
somewhatabstract: Not necessarily, but the flow enum stuff has to come first and it was confusing to see these being pushed onto the front. I was reworking things to do this conditional stuff first, as that felt easier to follow, but by the time I fixed things up, it did not seem necessary any more.

packages/wonder-blocks-core/src/components/with-ssr-placeholder.js
```diff
@@ -115,35 +117,10 @@ export default class WithSSRPlaceholder extends React.Component<Props, State> {
...
-            //
-            // We ignore this from coverage. It's a maintenance case to help
-            // us catch code changes that affect the control flow unexpectedly,
-            // but it's not something we need to write a test case for.
-            //
-            /* istanbul ignore next */
-            default:
-                // Let's log this case so we can debug it easily.
-                // Then fall through to the root case.
-                /* eslint-disable-next-line no-console */
```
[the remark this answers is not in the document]
somewhatabstract: Yup

packages/wonder-blocks-modal/src/components/__tests__/modal-backdrop.test.js
```diff
@@ -98,10 +124,11 @@ describe("ModalBackdrop", () => {
                     footer={<div data-modal-footer />}
                 />
             </ModalBackdrop>,
+            {attachTo: attachElement},
```
jandrade, line 127: Was this introduced due to the Jest upgrade? or to fix some false positives?

packages/wonder-blocks-modal/src/util/find-focusable-nodes.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+
+/**
+ * List of elements that can be focused
+ * @see https://www.w3.org/TR/html5/editing.html#can-be-focused
+ */
+const FOCUSABLE_ELEMENTS =
+    'a[href], details, input, textarea, select, button:not([aria-label^="Close"])';
+
+export function findFocusableNodes(
```
jandrade, line 10: I see this is pretty similar to what we have in `wb-popover`. Probably we can abstract this more but it can be done separately. (No change required). In fact, I can see now that we have some shared utils with Modal, Popover and Tooltip that could be abstracted out into a separate (private) package.

jandrade (approved):
LGTM! Thanks for working on this! There's a lot going on here, but the changes are understandable. Just added a couple of comments, but mostly for the posteriority (and probably one comment about adding context about understanding better the `attachTo` change).

package.json
```diff
@@ -35,76 +35,77 @@
...
-    "eslint-watch": "^6.0.1",
-    "flow-bin": "0.154.0",
+    "eslint-plugin-testing-library": "^5.0.0",
+    "eslint-watch": "^8.0.0",
+    "flow-bin": "^0.165.1",
     "glob": "^7.1.2",
     "husky": "^4.2.5",
-    "jest": "^25.1.0",
-    "jest-enzyme": "^7.0.2",
+    "jest": "^27.3.1",
```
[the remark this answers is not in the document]
somewhatabstract: Yes, probably. The updating of babel led me to update all the things that relied on it and that led to me updating this. In hindsight, I agree this likely could've been left as-was.

packages/wonder-blocks-data/package.json
somewhatabstract: I don't know.

packages/wonder-blocks-modal/src/components/__tests__/modal-backdrop.test.js
somewhatabstract: This is enzyme specific. For some reason, certain things involving the DOM (like focus() changing document.activeElement), the component must be actually in the DOM, not just mounted on an element outside of the document.

somewhatabstract:
This JSDOM environment nonsense is killing me. I am removing `jest-enzyme` to eliminate that as part of the problem (just copying over the bits we need into our setup script).

somewhatabstract:
OK, it's not jest-enzyme that's at fault. I think it must be caching related. Going to temporarily disable caching and see what happens.

somewhatabstract:
I think the accepting an older version of cache is causing continuous cache pollution with the wrong versions somehow

somewhatabstract:
OK, I think I have an idea of what is happening. 

1. `jest-enzyme` was causing an old jsdom env to get installed
2. The caches were getting incorrectly setup with that version and saved due to mismatched workflow caching configuration that didn't run `yarn install` if the cache was retrieved and since we fell back to old caches, this happened every time.
3. With the cache polluted that way, any use of the cache seemingly caused a bad jsdom env to get used

I have disabled all caching to prove this out so `yarn install` is "pure". Assuming that works, I will reintroduce caching with updated keys and removing the fallback patterns to see if it still works - may also look at moving the cache steps to a workflow file that can be triggered by the other workflows.

If that all still works, I may look at allowing fallback cache retrieval for the yarn cache but not the node-modules cache and see what happens.

somewhatabstract:
OK, now things are weird. It's like it's just not installing things properly 👎🏻 

I'll come back to this later. I am thoroughly confused.

.github/workflows/node-ci-pr.yml
```diff
@@ -61,22 +82,10 @@ jobs:
...
-          find-related-tests: 1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      - name: Run Jest with coverage
-        uses: Khan/jest-coverage-action@v1.1.3
-        with:
-          jest-bin: ./node_modules/.bin/jest
-          coverage-data-path: ./coverage/coverage-final.json
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
somewhatabstract: I'm of the opinion that they aren't valuable anymore and that just running these things from `yarn` is the right way. I don't know that continuing work on them makes sense.

PR #1240, opened 2021-12-07
Add pre-publish-check for changesets

jandrade:
## Summary:

Now that we have `changesets` as part of the workflow, we need to ensure that
our publish checks run smoothly, no matter if it's a manual or automated release
process.

This PR makes some modifications to the existing `pre-publish-check.js` file to
be able to reuse it with our CI env.

You can see the original failure here: https://github.com/Khan/wonder-blocks/actions/runs/1546721356

- Moved script to `utils/publish/` folder.
- Extracted some shared functions into `pre-publish-utils.js`.
- Created a lite version of `pre-publish-check.js` to make it work with `changesets` by:
    - removing the npm user check, because this process runs using an automation token instead.
    - Removing the warnings and/or inquirer so the deploy process is not interactive.

Issue: XXX-XXXX

## Test plan:

Run the following scripts and verify that it runs without any problems:

```
node utils/publish/pre-publish-check.js
node utils/publish/pre-publish-check-ci.js
```


utils/publish/pre-publish-check.js
```diff
@@ -1,62 +1,22 @@
...
 /**
  * Pre-publish checks to verify that our publish will go smoothly.
+ * This file is used by the npm publish script (manual publish).
  */
 const path = require("path");
 const {exec} = require("child_process");
 const glob = require("glob");
 
 const inquirer = require("inquirer");
-
```
jandrade, line 10: I just moved all this code to `pre-publish-check-utils.js`.

utils/publish/pre-publish-check-ci.js
```diff
@@ -0,0 +1,30 @@
...
+    path.join(__dirname, "..", "..", "packages", "**", "package.json"),
+    (err, pkgPaths) => {
+        for (const pkgPath of pkgPaths) {
+            const pkgJson = require(path.relative(__dirname, pkgPath));
+
+            checkPublishConfig(pkgJson);
+            checkPackageMain(pkgJson);
+            checkPackageModule(pkgJson);
+            checkPackageSource(pkgJson);
+        }
```
jandrade, line 28: The main difference of this script is that we don't need to run `checkNpmUser` here because that's handled via the changesets action that uses a [npm automation token](https://github.blog/changelog/2020-10-02-npm-automation-tokens/).

jeresig (approved):
Looks reasonable to me! Thank you

PR #1242, opened 2021-12-08
Add skeleton wonder-blocks-testing package

somewhatabstract:
## Summary:
This introduces a new package for our test tooling and components.

It does absolutely nothing in this PR.

Issue: FEI-4126

## Test plan:
`yarn test`

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,13 +1,14 @@
 *   @kevinbarabash @jeresig @somewhatabstract @jandrade @github-actions[bot]
 packages/wonder-blocks-breadcrumbs/*   @jandrade @kevinbarabash @jeresig
 packages/wonder-blocks-button/*   @kevinbarabash @jeresig
-packages/wonder-blocks-color/*   @jangmi-jo @kevinbarabash @jeresig
+packages/wonder-blocks-color/*   @kevinbarabash @jeresig
 packages/wonder-blocks-core/*   @somewhatabstract @kevinbarabash @jeresig
-packages/wonder-blocks-data/*   @somewhatabstract @jeresig @jangmi-jo
+packages/wonder-blocks-data/*   @somewhatabstract @jeresig
 packages/wonder-blocks-icon-button/*   @kevinbarabash @jeresig
 packages/wonder-blocks-link/*   @kevinbarabash @jeresig
 packages/wonder-blocks-modal/*   @jandrade @kevinbarabash @jeresig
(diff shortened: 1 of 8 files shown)


CODEOWNERS
```diff
@@ -1,13 +1,14 @@
...
-packages/wonder-blocks-color/*   @jangmi-jo @kevinbarabash @jeresig
+packages/wonder-blocks-color/*   @kevinbarabash @jeresig
 packages/wonder-blocks-core/*   @somewhatabstract @kevinbarabash @jeresig
-packages/wonder-blocks-data/*   @somewhatabstract @jeresig @jangmi-jo
+packages/wonder-blocks-data/*   @somewhatabstract @jeresig
 packages/wonder-blocks-icon-button/*   @kevinbarabash @jeresig
 packages/wonder-blocks-link/*   @kevinbarabash @jeresig
 packages/wonder-blocks-modal/*   @jandrade @kevinbarabash @jeresig
 packages/wonder-blocks-popover/*   @jandrade @jeresig
+packages/wonder-blocks-testing/*   @somewhatabstract @jandrade @jeresig
```
jandrade, line 11: oh now that I see this I think we can get rid of CODEOWNERS now that we support Gerald in the repo. No need to change this, but more like we have to work on removing this from the repo.

styleguide.config.js
```diff
@@ -203,6 +203,10 @@ module.exports = {
             name: "Spacing",
             content: "packages/wonder-blocks-spacing/docs.md",
         },
+        {
+            name: "Testing",
+            content: "packages/wonder-blocks-testing/docs.md",
+        },
```
jandrade, line 209: Probably worth mentioning that we won't support Styleguide in the short term. The plan is to migrate to Storybook docs in January/February, so probably it would be better if you can add the story for that instead 😃. That would reduce the effort to migrating later to Storybook, and that'd be highly appreciated 🙏.

jandrade (approved):
Looks good. The only thing that would be appreciated is to use Storybook instead of Styleguidist for the documentation (now that we won't be supporting any new work with Styleguidist in the short term).

styleguide.config.js
somewhatabstract: I don't intend on updating the styleguidist doc beyond this one file. I'll add stories when there's something to add stories for :D

jandrade: sounds good, thanks for the clarification!

PR #1243, opened 2021-12-08
Prevent nesting of RenderStateRoots

kevinbarabash:
## Summary:
Nesting of RenderStateRoots will result in unnecessary renders at best and incorrect behavior at worst.  This PR updates the component to throw if it detects that its a descendant of another RenderStateRoot.

Issue: none

## Test plan:
- yarn test

### DIFF
--- a/packages/wonder-blocks-core/src/hooks/use-unique-id.stories.mdx
+++ b/packages/wonder-blocks-core/src/hooks/use-unique-id.stories.mdx
@@ -24,7 +24,7 @@ factory for each subsequent render. The identifier factory is unique to
 each component.
 
 NOTE: All uses of `uniqueUniqueIdWithoutMock` should appear as descendants
-of `<RenderStateRoot>`.  It's customary to place `<RenderStateRoot>` near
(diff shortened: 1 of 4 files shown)


packages/wonder-blocks-core/src/hooks/use-unique-id.stories.mdx
```diff
@@ -23,6 +23,11 @@ It will return `null` on the initial render and then the same identifier
 factory for each subsequent render. The identifier factory is unique to
 each component.
 
+NOTE: All uses of `uniqueUniqueIdWithoutMock` should appear as descendants
```
somewhatabstract, line 26: `uniqueUnique...`? Should that be `useUnique...`?

somewhatabstract (approved):
A possibly typo in the docs, but otherwise looks great. Thanks!

packages/wonder-blocks-core/src/hooks/use-unique-id.stories.mdx
kevinbarabash: copy 🍝

PR #1244, opened 2021-12-09
FEI-4023: Create useInterval hook

kevinbarabash:
## Summary:
This is very similar to the useTimeout hook from #1225.  I ported the tests from timeout.test.js and added a few additional tests.  This should provide a high level of confidence that the hook and withActionScheduler's schedule.interval() method have consistent behavior.

Issue: FEI-4023

## Test plan:
- yarn jest use-interval.test.js
- http://localhost:6061/?path=/docs/timing-useinterval--immediately
- test that the example stories behave as expected

### DIFF
--- a/packages/wonder-blocks-timing/src/hooks/use-interval.js
+++ b/packages/wonder-blocks-timing/src/hooks/use-interval.js
@@ -1,0 +1,79 @@
+// @flow
+import {useEffect, useState, useCallback, useRef} from "react";
+
+import {
+    SchedulePolicy as SchedulePolicies,
+    ClearPolicy as ClearPolicies,
+} from "../util/policies.js";
+import type {IInterval, ClearPolicy, Options} from "../util/types.js";
+
+export function useInterval(
+    action: () => mixed,
+    intervalMs: number,
+    options?: Options,
+): IInterval {
+    if (typeof action !== "function") {
+        throw new Error("Action must be a function");
+    }
+
+    if (intervalMs < 1) {
+        throw new Error("Interval period must be >= 1");
+    }
+
+    const schedulePolicy =
+        options?.schedulePolicy ?? SchedulePolicies.Immediately;
+    const [isSet, setIsSet] = useState(
+        schedulePolicy === SchedulePolicies.Immediately,
+    );
+    const actionRef = useRef(action);
+    const mountedRef = useRef(false);
+
+    useEffect(() => {
+        mountedRef.current = true;
+        return () => {
+            mountedRef.current = false;
+        };
+    }, []);
+
+    useEffect(() => {
+        actionRef.current = action;
+    }, [action]);
+
+    const clear = useCallback(
+        (policy?: ClearPolicy) => {
+            if (
+                isSet &&
+                (policy ?? options?.clearPolicy) === ClearPolicies.Resolve
+            ) {
+                actionRef.current();
+            }
+            // This will cause the useEffect below to re-run
+            setIsSet(false);
+        },
+        [isSet, options?.clearPolicy],
+    );
+
+    const set = useCallback(() => {
+        if (!isSet) {
+            // This will cause the useEffect below to re-run
+            setIsSet(true);
+        }
+    }, [isSet]);
+
+    useEffect(() => {
+        if (isSet && mountedRef.current) {
+            const timeout = setInterval(() => {
+                actionRef.current();
+            }, intervalMs);
+
+            return () => {
+                clearInterval(timeout);
+                if (!mountedRef.current) {
+                    clear();
+                }
+            };
+        }
+    }, [clear, isSet, intervalMs]);
+
+    return {isSet, set, clear};
+}
--- a/packages/wonder-blocks-timing/src/hooks/use-interval.stories.mdx
+++ b/packages/wonder-blocks-timing/src/hooks/use-interval.stories.mdx
@@ -1,0 +1,147 @@
+import {Meta, Story, Source, Canvas} from "@storybook/addon-docs";
+
+import {Body, HeadingSmall} from "@khanacademy/wonder-blocks-typography";
+import {View} from "@khanacademy/wonder-blocks-core";
+import Button from "@khanacademy/wonder-blocks-button";
+
+import {ClearPolicy, SchedulePolicy} from "../util/policies.js";
+import {useInterval} from "./use-interval.js";
+
+<Meta
+    title="Timing/useInterval"
+    parameters={{
+        chromatic: {
+            disableSnapshot: true,
+        },
+    }}
+/>
+
+# `useInterval`
+
+`useInterval` is a hook that provides a convenient API for setting and clearing
+an interval. It is defined as follows:
+
+```ts
+function useInterval(
+    action: () => mixed,
+    timeoutMs: number,
+    options?: {|
+        schedulePolicy?: "schedule-immediately" | "schedule-on-demand",
+        clearPolicy?: "resolve-on-clear" | "cancel-on-clear",
+    |},
+): ITimeout;
+
+interface ITimeout {
+    get isSet(): boolean;
+    set(): void;
+    clear(policy?: ClearPolicy): void;
+}
+```
+
+By default the interval will be set immediately up creation. The `options` parameter can
+be used to control when when the interval is schedule and whether or not `action` should be
+called when the interval is cleared.
+
+Notes:
+
+-   Because `clear` takes a param, it's import that you don't pass it directly to an event handler,
+    e.g. `<Button onClick={clear} />` will not work as expected.
+-   Calling `set` after the interval has been cleared will restart the interval.
+-   Updating the second paramter, `timeoutMs`, will also restart the interval.
+-   When the component using this hooks is unmounted, the interval will automatically be cleared.
+-   Calling `set` after the interval is already set does nothing.
+
+export const Immediately = () => {
+    const [callCount, setCallCount] = React.useState(0);
+    const callback = React.useCallback(() => {
+        setCallCount((callCount) => callCount + 1);
+    }, []);
+    const {isSet, set, clear} = useInterval(callback, 1000);
+    return (
+        <View>
+            <View>isSet = {isSet.toString()}</View>
+            <View>callCount = {callCount}</View>
+            <View style={{flexDirection: "row"}}>
+                <Button onClick={set}>Set interval</Button>
+                <Button onClick={clear}>Clear interval</Button>
+            </View>
+        </View>
+    );
+};
+
+<Canvas>
+    <Story name="Immediately">
+        <Immediately />
+    </Story>
+</Canvas>
+
+```jsx
+const Immediately = () => {
+    const [callCount, setCallCount] = React.useState(0);
+    const callback = React.useCallback(() => {
+        setCallCount((callCount) => callCount + 1);
+    }, []);
+    const {isSet, set, clear} = useInterval(callback, 1000);
+    return (
+        <View>
+            <View>isSet = {isSet.toString()}</View>
+            <View>callCount = {callCount}</View>
+            <View style={{flexDirection: "row"}}>
+                <Button onClick={() => set()}>Set interval</Button>
+                <Button onClick={() => clear()}>Clear interval</Button>
+            </View>
+        </View>
+    );
+};
+```
+
+export const OnDemandAndResolveOnClear = () => {
+    const [callCount, setCallCount] = React.useState(0);
+    const callback = React.useCallback(() => {
+        console.log("action called");
+        setCallCount((callCount) => callCount + 1);
+    }, []);
+    const {isSet, set, clear} = useInterval(callback, 1000, {
+        clearPolicy: ClearPolicy.Resolve,
+        schedulePolicy: SchedulePolicy.OnDemand,
+    });
+    return (
+        <View>
+            <View>isSet = {isSet.toString()}</View>
+            <View>callCount = {callCount}</View>
+            <View style={{flexDirection: "row"}}>
+                <Button onClick={() => set()}>Set interval</Button>
+                <Button onClick={() => clear()}>Clear interval</Button>
+            </View>
+        </View>
+    );
+};
+
+<Canvas>
+    <Story name="OnDemandAndResolveOnClear">
+        <OnDemandAndResolveOnClear />
+    </Story>
+</Canvas>
+
+```jsx
+const OnDemandAndResolveOnClear = () => {
+    const [callCount, setCallCount] = React.useState(0);
+    const callback = React.useCallback(() => {
+        setCallCount((callCount) => callCount + 1);
+    }, []);
+    const {isSet, set, clear} = useInterval(callback, 1000, {
+        clearPolicy: ClearPolicy.Resolve,
+        schedulePolicy: SchedulePolicy.OnDemand,
+    });
+    return (
+        <View>
+            <View>isSet = {isSet.toString()}</View>
+            <View>callCount = {callCount}</View>
+            <View style={{flexDirection: "row"}}>
+                <Button onClick={() => set()}>Set interval</Button>
+                <Button onClick={() => clear()}>Clear interval</Button>
+            </View>
+        </View>
+    );
+};
+```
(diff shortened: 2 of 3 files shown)


packages/wonder-blocks-timing/src/hooks/use-interval.js
```diff
@@ -0,0 +1,79 @@
...
+    );
+    const actionRef = useRef(action);
+    const mountedRef = useRef(false);
+
+    useEffect(() => {
+        mountedRef.current = true;
+        return () => {
+            mountedRef.current = false;
+        };
+    }, []);
```
jaredly, line 36: this would be nice as a hook by itself:
```js
const useMounted = () => {
    const ref  = useRef(false);
    useEffect(() => {
        ref.current = true;
        return () => ref.current = false;
    }, [])
    return ref
}
```

packages/wonder-blocks-timing/src/hooks/use-interval.js
```diff
@@ -0,0 +1,79 @@
...
+
+    useEffect(() => {
+        actionRef.current = action;
+    }, [action]);
```
jaredly, line 40: I've also done a `useCurrent()` hook in other projects that's quite handy
```js
const useCurrent = <T>(value: T) => {
    const ref = useRef(value);
    useEffect(() => ref.current = value, [value]);
    return ref
}
```

packages/wonder-blocks-timing/src/hooks/use-interval.js
```diff
@@ -0,0 +1,79 @@
...
+    const set = useCallback(() => {
+        if (!isSet) {
+            // This will cause the useEffect below to re-run
+            setIsSet(true);
+        }
+    }, [isSet]);
+
+    useEffect(() => {
+        if (isSet && mountedRef.current) {
+            const timeout = setInterval(() => {
```
jaredly, line 65: `timeout` here should probably be `intervalID`?

packages/wonder-blocks-timing/src/hooks/use-interval.stories.mdx
```diff
@@ -0,0 +1,147 @@
...
+        setCallCount((callCount) => callCount + 1);
+    }, []);
+    const {isSet, set, clear} = useInterval(callback, 1000);
+    return (
+        <View>
+            <View>isSet = {isSet.toString()}</View>
+            <View>callCount = {callCount}</View>
+            <View style={{flexDirection: "row"}}>
+                <Button onClick={set}>Set interval</Button>
+                <Button onClick={clear}>Clear interval</Button>
```
jaredly, line 66: is this supposed to be a counter-example? because above it says "don't pass clear to onClick"

jaredly (changes requested):
So, thinking about this API, I think it could be simpler if instead of the hook owning the "isSet" and "setIsSet", it just accepted an `enabled` boolean.
so
```
useInterval(action, interval, enabled)
```
that would remove the need for SchedulePolicy, and maybe also for ClearPolicy? In fact, I don't think it would need to return any of these functions; the return type could be `void`.

Basically, I think maybe putting the state of whether the interval is "on" or not at the higher level would better follow react/functional style, whereas the current format requires the hook to pass back several "internal" handles so that the parent can control it. Does that make sense?
It's also possible that I'm overlooking something important -- I'm definitely willing to be convinced!

packages/wonder-blocks-timing/src/hooks/use-interval.js
kevinbarabash: Great idea.  We use this same pattern in `useTimeout`.  I'm going to keep it internal for now, but we can make it public later if we want.

packages/wonder-blocks-timing/src/hooks/use-interval.js
kevinbarabash: That seems handy too.

packages/wonder-blocks-timing/src/hooks/use-interval.js
kevinbarabash: copy 🍝

packages/wonder-blocks-timing/src/hooks/use-interval.stories.mdx
kevinbarabash: This is me forgetting my own guidance.  Thankfully flow picks this up (I just checked).  The reason why Flow didn't pick this up here is that it doesn't run in .mdx files.

kevinbarabash:
> So, thinking about this API, I think it could be simpler if instead of the hook owning the "isSet" and "setIsSet", it just accepted an `enabled` boolean. so
> 
> ```
> useInterval(action, interval, enabled)
> ```
>
> that would remove the need for SchedulePolicy, and maybe also for ClearPolicy? In fact, I don't think it would need to return any of these functions; the return type could be `void`.
>
> Basically, I think maybe putting the state of whether the interval is "on" or not at the higher level would better follow react/functional style, whereas the current format requires the hook to pass back several "internal" handles so that the parent can control it. Does that make sense? It's also possible that I'm overlooking something important -- I'm definitely willing to be convinced!

The API you've proposed is what most of the third-party APIs that I evaluated looked like (or something similar).  The issue I found with this API is that it's more difficult to use if you want to start the interval/timeout in response to some event (lots of code in webapp does this).  In that situation you'll need to also use a `useState` hook in addition to `useInterval`/`useTimeout` vs just calling `interval.set()`.

The API in this PR is based on the API of the `Interval` class that `ActionScheduler` uses.  We use `withActionScheduler` in a bunch of places in webapp already so having `useInterval` conform to `IInterval` will make it easier for people to use since they already have experience with `withActionScheduler`.

jaredly:
@kevinbarabash hmmmm ok. Could we have `useInterval` be the simple/externally-controlled one, and then `useStatefulInterval` or something be just wrapping that with a useState?

packages/wonder-blocks-timing/src/hooks/use-interval.js
```diff
@@ -0,0 +1,97 @@
...
+
+    const set = useCallback(() => {
+        if (!isSet) {
+            // This will cause the useEffect below to re-run
+            setIsSet(true);
+        }
+    }, [isSet]);
+
+    useEffect(() => {
+        if (isSet && isMounted) {
```
jaredly, line 82: ooh this needs to be `isMounted.current`

packages/wonder-blocks-timing/src/hooks/use-interval.js
```diff
@@ -0,0 +1,97 @@
...
+        throw new Error("Interval period must be >= 1");
+    }
+
+    const schedulePolicy =
+        options?.schedulePolicy ?? SchedulePolicies.Immediately;
+    const [isSet, setIsSet] = useState(
+        schedulePolicy === SchedulePolicies.Immediately,
+    );
+    const actionRef = useUpdatingRef(action);
+    const isMounted = useMounted();
```
jaredly, line 58: maybe this should be `const isMountedRef = useMountedRef()` to make it clear that it's a ref

packages/wonder-blocks-timing/src/hooks/use-interval.js
jaredly: also: 🤔 I think it's maybe not possible for this useEffect function to be called once the component is unmounted.... so this might be superfluous.

kevinbarabash: Good catch.  I'm surprised that none of the test case failed like this.  I think I need add a test case that explicitly tests unmounting.

packages/wonder-blocks-timing/src/hooks/use-interval.js
```diff
@@ -0,0 +1,97 @@
...
+        if (isSet && isMounted) {
+            const intervalId = setInterval(() => {
+                actionRef.current();
+            }, intervalMs);
+
+            return () => {
+                clearInterval(intervalId);
+                if (!isMounted) {
+                    clear();
+                }
```
jaredly, line 91: ooh I bet calling clear here, which will call setIsSet, will cause a react error "you tried to set state after it unmounted"

kevinbarabash: I reworked things a bit and was able to get rid of the `setIsSet()` call.

packages/wonder-blocks-timing/src/hooks/internal/use-simple-interval.js
```diff
@@ -0,0 +1,34 @@
...
+ * @param action called every `intervalMs` when `isSet` is true
+ * @param intervalMs the duration between calls to `action`
+ * @param isSet whether or not the interval is active
+ */
+export function useSimpleInterval(
+    action: () => mixed,
+    intervalMs: number,
+    isSet: boolean,
+) {
+    const actionRef = useUpdatingRef(action);
```
tony-dinh, line 18: I'm curious as to what benefits we get from `useUpdatingRef`? It feels a bit counter-intuitive to me. 

My first thought when reading this was wondering whether or not `useCallback` could be used here for the same purpose

packages/wonder-blocks-timing/src/hooks/internal/use-simple-interval.js
```diff
@@ -0,0 +1,34 @@
...
+            }, intervalMs);
+
+            return () => {
+                clearInterval(intervalId);
+            };
+        }
+        // react-hooks/exhaustive-deps doesn't require refs to be
+        // listed in the deps array.  Unfortunately, in this situation
+        // it doesn't recognized actionRef as a ref.
+    }, [action, actionRef, intervalMs, isSet]);
```
tony-dinh, line 33: Did you mean to remove action from this dep list?

packages/wonder-blocks-timing/src/hooks/use-interval.js
```diff
@@ -0,0 +1,97 @@
...
+    ClearPolicy as ClearPolicies,
+} from "../util/policies.js";
+import type {IInterval, ClearPolicy, Options} from "../util/types.js";
+
+/**
+ * Returns `true` if the component is currently mount and `false`
+ * if it is not.
+ * @returns {boolean}
+ */
+const useMounted = () => {
```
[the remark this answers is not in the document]
kevinbarabash: It needs to return a ref.  I've updated `useMounted` to be called `useMountedRef`.

packages/wonder-blocks-timing/src/hooks/internal/use-simple-interval.js
kevinbarabash: Using a ref avoids recreating the interval.

packages/wonder-blocks-timing/src/hooks/internal/use-simple-interval.js
kevinbarabash: It's gone now.

kevinbarabash:
I chatted with Jared on a call and we decided to rewrite `useInterval` in terms of a simpler version of the hook.  Separately, I will audit our uses of action scheduler to see how many places in webapp where it actually makes sense to use the more complex action-scheduler type API over the simpler API and then use that to make a decision on whether which API ends up being the default, i.e. the one named `useInterval`, `useTimeout`, etc. and whether or not we ship the other one.

packages/wonder-blocks-timing/src/hooks/internal/use-mounted-ref.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+import {useEffect, useRef} from "react";
+
+/**
+ * Returns `true` if the component is currently mount and `false`
+ * if it is not.
+ * @returns {boolean}
```
tony-dinh, line 7: Should we update the docstring here to reflect the React Ref return value?

packages/wonder-blocks-timing/src/hooks/internal/use-simple-interval.js
tony-dinh: Might've forgotten to push up that change, it's still here! 😱

packages/wonder-blocks-timing/src/hooks/use-interval.js
```diff
@@ -0,0 +1,74 @@
...
+
+    const clear = useCallback(
+        (policy?: ClearPolicy) => {
+            policy = policy ?? options?.clearPolicy;
+            if (isSet && policy === ClearPolicies.Resolve) {
+                actionRef.current();
+            }
+            setIsSet(false);
+        },
+        // react-hooks/exhaustive-deps doesn't require refs to be
```
tony-dinh, line 46: Bummer... seems like there might be a chance this will be addressed if this lands
https://github.com/facebook/react/pull/20477

tony-dinh (approved):
I'll defer the final approval to Jared but things look sane to me, thanks for clarifying the usage of useUpdatingRef!

packages/wonder-blocks-timing/src/hooks/internal/use-simple-interval.js
```diff
@@ -0,0 +1,34 @@
...
+import {useEffect} from "react";
+
+import {useUpdatingRef} from "./use-updating-ref.js";
+
+/**
+ * A simple hook for using `setInterval`.
+ *
+ * @param action called every `intervalMs` when `isSet` is true
+ * @param intervalMs the duration between calls to `action`
+ * @param isSet whether or not the interval is active
```
jaredly, line 11: nit: can we rename `isSet` to `enabled` or `active`? `isSet` doesn't really make sense to me... (but not a blocker)

packages/wonder-blocks-timing/src/hooks/internal/use-updating-ref.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+import {useEffect, useRef} from "react";
+
+/**
+ * Returns a ref whose .current value is updated whenever
+ * the `value` passed to this hook changes.
```
jaredly, line 6: maybe add something about "this is great for values that you want to reference from within a useCallback or useEffect event listener, without re-triggering the effect when the value changes"?

packages/wonder-blocks-timing/src/hooks/use-interval.js
```diff
@@ -0,0 +1,74 @@
...
+                if (isSet && options?.clearPolicy === ClearPolicies.Resolve) {
+                    // eslint-disable-next-line react-hooks/exhaustive-deps
+                    actionRef.current();
+                }
+            }
+        };
+        // react-hooks/exhaustive-deps doesn't require refs to be
+        // listed in the deps array.  Unfortunately, in this situation
+        // it doesn't recognized actionRef and mountedRef as refs.
+    }, [actionRef, isSet, mountedRef, options?.clearPolicy]);
```
jaredly, line 69: hmmm I don't think this is unfortunate... by definition, actionRef and mountedRef will always be `===` to the previous value, so they're superfluous in this list.

packages/wonder-blocks-timing/src/hooks/use-interval.js
```diff
@@ -0,0 +1,74 @@
...
+            // eslint-disable-next-line react-hooks/exhaustive-deps
+            if (!mountedRef.current) {
+                // This code should only run when the component using the hook
+                // is unmounted.
+                if (isSet && options?.clearPolicy === ClearPolicies.Resolve) {
+                    // eslint-disable-next-line react-hooks/exhaustive-deps
+                    actionRef.current();
+                }
+            }
+        };
```
jaredly, line 65: hmm given that this `useEffect` /only/ wants to run on unmount, it would make more sense to me to write it like this:
```js
const runOnUnmount = useUpdatingRef(isSet && options?.clearPolicy === ClearPolicies.Resolve);
useEffect(() => {
    if (runOnUnmount.current) {
        actionRef.current();
    }
}, []);
```
up to you though

jaredly (approved):
nice, thanks for splitting this out!

packages/wonder-blocks-timing/src/hooks/use-interval.js
kevinbarabash: This lint rule will not require refs in the list if defined using `useRef`.  Having to add deps that wouldn't normally be required isn't great.

packages/wonder-blocks-timing/src/hooks/internal/use-updating-ref.js
kevinbarabash: Great suggestion!  I gotta up my docs game and go beyond just what a thing does to why you'd want to do such a thing.

packages/wonder-blocks-timing/src/hooks/use-interval.js
jaredly: ahhhh gotcha I totally misunderstood. it would be cool to be able to configure `exhaustive-deps` to indicate which custom hooks return refs.

kevinbarabash: I'm planning on forking that lint rule and making it better.  In addition to fixing this issue I also want to make it check that aren't any extra deps in the dependency array.

PR #1245, opened 2021-12-09
Chromatic: Add delay to Tooltip and Popover snapshots

jandrade:
## Summary:

Initial work for investigating on Chromatic about some flaky tests with PopperJS.

This first step is ~just to disable~ add a delay to the Popover and Tooltip
snapshots until we get a better sense of how (and if we can) remove the
flakyness.

Issue: WB-1170

## Test plan:

Verify that the chromatic build doesn't report any changes on the Tooltip and
Popover stories.


kevinbarabash (approved):
:shipit:

PR #1246, opened 2021-12-10
[fei4126.2] Implement framework

somewhatabstract:
## Summary:
This brings over the implementation from webapp with some improvements.

Sadly the ability to export with `module.exports` is not working in Wonder Blocks.
However, we know it works in webapp because we tried it already. Not
sure what the issue is in Wonder Blocks yet.

If not, then I may need to write a custom babel plugin that makes that
magic happen specifically for storybook (it will work fine for our
sandbox stuff).

Anyway, this gets is much closer.

Issue: FEI-4126

## Test plan:
`yarn start:storybook`


packages/wonder-blocks-testing/src/fixtures/fixtures.defaultwrapper.stories.js
```diff
@@ -0,0 +1,59 @@
...
+                },
+                Wrapper,
+            );
+        },
+    ) ?? {},
+);
+
+export default stories[0];
+export const f1 = stories[1];
+export const f2 = stories[2];
```
somewhatabstract, line 59: Disappointed that `module.exports = fixtures(...)` was insufficient in WB, but this works to demonstrate things for now. Really wish there was a way to export multiple stories without naming them all upfront.

packages/wonder-blocks-testing/src/fixtures/combine.js
```diff
@@ -0,0 +1,40 @@
...
+ * If the values are objects, it will merge them at the top level, with val2
+ * overwriting val1 where there are conflicts.
+ *
+ * If the values are arrays, it will concatenate and dedupe them.
+ * NOTE: duplicates in either val1 or val2 will also be deduped.
+ *
+ * If the values are any other type, or val2 has a different type to val1, val2
+ * will be returned.
+ */
+export const combine = (val1: $ReadOnly<any>, val2: $ReadOnly<any>): any => {
```
kevinbarabash, line 19: There's a number of different npm packages that do deep object merging. Could we have used one of those?

packages/wonder-blocks-testing/src/fixtures/fixtures.js
```diff
@@ -0,0 +1,77 @@
...
+        description: groupDescription,
+        defaultWrapper,
+        additionalAdapterOptions,
+    } = options;
+
+    // 1. Create a new adapter group.
+    const group = adapter.declareGroup<TProps>({
+        title: title || component.displayName || component.name || "Component",
+        description: groupDescription,
+    });
```
kevinbarabash, line 53: I'm not sure what Storybook does with a test title that's the same between different files. I think we might be take more precautions to make this unique.

There's a secondary issue with the organization of stories into groups. Titles with / in them will turn into folders in the Storybook UI. Maybe we could check if there's no / in the title and if there isn't we could put all of those stories into an "Uncategorized" folder by prepending the title with "Uncategorized/". This will keep the number of top-level folders to a minimum.

packages/wonder-blocks-testing/src/fixtures/__tests__/combine-options.test.js
```diff
@@ -0,0 +1,65 @@
...
+            {a: "test1", b: "test1"},
+            {b: "test2"},
+            {a: "test3", c: "test3"},
+        ];
+        jest.spyOn(CombineModule, "combine").mockImplementation(
+            // Just for testing, we know the values are strings, so let's
+            // combine them with concatenation so we see the order of
+            // combination in the result.
+            (v1, v2) => `${v1 === undefined ? "" : v1}${v2}`,
+        );
```
kevinbarabash, line 53: Can't the order be inferred by which values end up in the result?

kevinbarabash (commented):
I copied over the two highest priority comments from https://github.com/Khan/webapp/pull/2035.  I'll finish reviewing Monday.

packages/wonder-blocks-testing/src/fixtures/fixtures.js
somewhatabstract: That would be something to do at the top group level rather than here.

packages/wonder-blocks-testing/src/fixtures/combine.js
somewhatabstract: This doesn't do a deep merge and I don't think we want it to.

packages/wonder-blocks-testing/src/fixtures/__tests__/combine-options.test.js
somewhatabstract: Yes, that's what the comment says. Can you clarify what you mean? Is there a change you're suggesting?

packages/wonder-blocks-testing/src/fixtures/combine.js
kevinbarabash: Not sure why I thought it was a deep merge.😳

packages/wonder-blocks-testing/src/fixtures/fixtures.js
kevinbarabash: I didn't realize the groups could be nested.  That's a neat feature.

It could be an optional behaviour.  Maybe I'm being paranoid.  Maybe telling people to organize their stories into folders will be enough and if enough people forget, the result will be so annoying that people will self correct.

somewhatabstract: To be clear, the nesting is a storybook feature based on the slashes in the name. There's no procedural nesting of the groups in the framework itself (at least not yet).

packages/wonder-blocks-testing/src/fixtures/__tests__/combine-options.test.js
kevinbarabash: If we didn't have this mock implementation here, the result would be:
```
{
  a: "test3",
  b: "test2",
  c: "test3",
}
```
We could assert that result.  This mock doesn't seem like it's adding much.  Apologies if I misunderstanding what this test is testing.

somewhatabstract: I can see how you may reach that conclusion, but it won't work like that. The jest.mock call at the top of the file ensures we isolate the code so the original implementation won't be available here. I want to assert the code runs as-is without relying that other code works correctly.

packages/wonder-blocks-testing/src/fixtures/adapters/storybook.js
```diff
@@ -0,0 +1,115 @@
...
+} from "../types.js";
+
+type StoryContext = {|
+    args: $ReadOnly<any>,
+    argTypes: $ReadOnly<any>,
+    globals: $ReadOnlyArray<any>,
+    hooks: $ReadOnlyArray<any>,
+    parameters: $ReadOnly<any>,
+    viewMode: mixed,
+|};
```
kevinbarabash, line 19: TIL that StoryBook has its own context object that we can use.

packages/wonder-blocks-testing/src/fixtures/adapters/storybook.js
```diff
@@ -0,0 +1,115 @@
...
+            const exports = declaredFixtures.reduce(
+                (acc, {description, getProps, component: Component}, i) => {
+                    const storyName = `${i + 1} ${description}`;
+                    const exportName = storyName
+                        // Make word boundaries start with an upper case letter.
+                        .replace(/\b\w/g, (c) => c.toUpperCase())
+                        // Remove all non-alphanumeric characters.
+                        .replace(/[^\w]+/g, "")
+                        // Remove all underscores.
+                        .replace(/[_]+/g, "");
```
kevinbarabash, line 67: Thank you for documenting these.

packages/wonder-blocks-testing/src/fixtures/combine.js
```diff
@@ -0,0 +1,40 @@
...
+        typeof val2 === "object"
+    ) {
+        const val1IsArray = Array.isArray(val1);
+        const val2IsArray = Array.isArray(val2);
+
+        if (val1IsArray && val2IsArray) {
+            return Array.from(new Set([...val1, ...obj2Clone]));
+        } else if (!val1IsArray && !val2IsArray) {
+            return {...val1, ...obj2Clone};
+        }
```
kevinbarabash, line 37: nit: Cloning `obj2` isn't necessary in these cases.

packages/wonder-blocks-testing/src/fixtures/fixtures.defaultwrapper.stories.js
kevinbarabash: We could create a babel plugin to help with this.  No action req'd in this PR.

packages/wonder-blocks-testing/src/fixtures/fixtures.js
```diff
@@ -0,0 +1,77 @@
...
+ * fixture group.
+ * @param {FixtureFn<TProps> => void} fn A function that provides a `fixture`
+ * function for defining fixtures.
+ * @returns {Exports} The object to be exported as `module.exports`.
+ *
+ * TODO(somewhatabstract): Determine a way around this requirement so we
+ * can support named exports and default exports via the adapters in a
+ * deterministic way. Currently this is imposed on us because of how
+ * storybook, the popular framework, uses both default and named exports for
+ * its interface.
```
kevinbarabash, line 28: 👍

packages/wonder-blocks-testing/src/fixtures/types.js
```diff
@@ -0,0 +1,180 @@
...
+export type FixturesOptions<TProps: {...}> = {|
+    /**
+     * The component being tested by the fixtures.
+     */
+    component: React.ComponentType<TProps>,
+
+    /**
+     * Optional title of the fixture collection.
+     */
+    title?: string,
```
kevinbarabash, line 35: Any plans to make this non-optional in the future?

packages/wonder-blocks-testing/src/jest/isolate-modules.js
```diff
@@ -0,0 +1,31 @@
+// @flow
+// Opt this file out of coverage because it's super hard to test.
+/* istanbul ignore file */
```
kevinbarabash, line 3: :+1:

kevinbarabash (approved):
LGTM.  This is a very nice way of defining fixtures/stories.  Figuring out a better story for name exports will be keep for future adoption, but it's fine to punt on that for now.  Thanks for all the detailed jsdocs.

packages/wonder-blocks-testing/src/fixtures/__tests__/combine-options.test.js
somewhatabstract: No worries. It's easy to do. Would be great if there were some sort of IDE highlighting for mocked calls to make it clearer.

packages/wonder-blocks-testing/src/fixtures/__tests__/fixtures.test.js
```diff
@@ -0,0 +1,483 @@
...
+        );
+
+        // Assert
+        expect(adapter.declareGroup).toHaveBeenCalledWith({
+            title: "DISPLAYNAME",
+        });
+    });
+
+    it("should default the title to the component.name in the absence of component.displayName", () => {
+        // Arrange
```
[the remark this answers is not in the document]
somewhatabstract: Yeah, basically.

packages/wonder-blocks-testing/src/fixtures/__tests__/setup.test.js
```diff
@@ -0,0 +1,67 @@
...
+        // Act
+        const act = () => getConfiguration();
+
+        // Assert
+        expect(act).toThrowErrorMatchingInlineSnapshot(`"Not configured"`);
+    });
+});
+
+describe("#setup", () => {
+    it("should set the configuration returned by getConfiguration", () => {
```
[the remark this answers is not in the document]
somewhatabstract: The test case title, for me, represents the requirement, not the implementation. It is a requirement that it set the config that getConfiguration returns, but not necessarily a requirement that it can be called more than once.

packages/wonder-blocks-testing/src/fixtures/adapters/adapter.js
```diff
@@ -0,0 +1,63 @@
...
+    +_name: string;
+    +_closeGroupFn: CloseGroupFn<any, Options, Exports>;
+
+    /**
+     * @param {string} name The name of the adapter.
+     * @param {CloseGroupFn<any, Options, Exports>} closeGroupFn The function
+     * an adapter group should call when the group is closed. This is invoked
+     * by an adapter group when it is closed. This function is where an
+     * adapter implements the logic to generate the actual fixtures for the
+     * adapter's target framework.
```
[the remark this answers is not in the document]
somewhatabstract: Understood. Naming things is hard so I went with something really generic. It could also be called just `generateExports` or something? Anyway, I'll leave as-is for now since it fits the "group" nomenclature adopted in the framework for this general concept.

packages/wonder-blocks-testing/src/fixtures/adapters/adapter.js
```diff
@@ -0,0 +1,63 @@
...
+        this._name = name;
+        this._closeGroupFn = closeGroupFn;
+    }
+
+    /**
+     * The name of the adapter.
+     */
+    get name(): string {
+        return this._name;
+    }
```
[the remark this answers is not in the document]
somewhatabstract: The name keys the options overrides that a group can give. It seemed handly to expose it in case folks want to make sure they are keying those overrides with the right adapter name (like the adapter names in withTestHarness, for example).

packages/wonder-blocks-testing/src/fixtures/adapters/storybook.js
somewhatabstract: Yeah, I learned while I was reading up on StoryBook's API to try and flesh out this framework adapter a bit.

packages/wonder-blocks-testing/src/fixtures/combine.js
somewhatabstract: I'll rename the function to clear this up a bit.

packages/wonder-blocks-testing/src/fixtures/fixtures.defaultwrapper.stories.js
somewhatabstract: Yeah - it would need to at least be able to know how many fixtures there are though.

packages/wonder-blocks-testing/src/fixtures/types.js
somewhatabstract: No - the reason it's optional is that many folks will just want this to be the component name, which is standard for storybook - so providing no title defaults to that.

somewhatabstract:
The parent pull-request ([#1242](https://github.com/Khan/wonder-blocks/pull/1242)) has been merged into `feature/testingframework`, but this branch (`fei4126.2`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

packages/wonder-blocks-testing/src/fixtures/combine.js
somewhatabstract: I believe it is in order to ensure nested items are independent of the originally passed object - otherwise changes to the source object after the combination would result in changes to the combined object.

PR #1247, opened 2021-12-10
Add some tests that test the interaction of RenderStateRoot with other components

kevinbarabash:
## Summary:
I added a few tests that verify that `WithSSRPlaceholder` and `UniqueIDProvider` can operate as descendants of `RenderStateRoot` without blowing up.

Issue: none

## Test plan:
- yarn jest unique-id-provider.test.js with-ssr-placeholder.test.js

### DIFF
--- a/packages/wonder-blocks-core/src/components/__tests__/with-ssr-placeholder.test.js
+++ b/packages/wonder-blocks-core/src/components/__tests__/with-ssr-placeholder.test.js
@@ -4,6 +4,7 @@ import * as ReactDOMServer from "react-dom/server.js";
 import {mount} from "enzyme";
 
 import WithSSRPlaceholder from "../with-ssr-placeholder.js";
+import {RenderStateRoot} from "../render-state-root.js";
 
 describe("WithSSRPlaceholder", () => {
     describe("client-side rendering", () => {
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-core/src/components/__tests__/with-ssr-placeholder.test.js
```diff
@@ -186,4 +187,73 @@ describe("WithSSRPlaceholder", () => {
...
+
+                // Act
+                mount(nodes);
+            });
+
+            // Assert
+            // Our promise doesn't resolve until the children render, therefore
+            // we don't get here until that and so if the placeholder has been
+            // called, it must have been called first.
+            expect(mockPlaceholder).toHaveBeenCalledTimes(1);
```
jaredly, line 215: hmm I wonder if it would be clearer if we did something like
`const trace = []`;
`placeholder={() => trace.push("placeholder")}`
`{() => trace.push("content")}`

then we could assert the order more explicitly, with `expect(trace).toEqual(["placeholder", "content"])`

jaredly (approved):
cool

packages/wonder-blocks-core/src/components/__tests__/with-ssr-placeholder.test.js
kevinbarabash: I copied some of the existing tests so I'd like to stay with this style.

kevinbarabash:
This was rebased on top of the change to add delays to Chromatic snapshots.  I'm approving the changes to the screenshots.  Unfortunately, it looks like the delay isn't making things more stable.

PR #1252, opened 2021-12-14
Explicitly import jest-enzyme in tests that use jest matchers

kevinbarabash:
## Summary:
jest-enzyme and @testing-library/jest-dom have some matchers with the same name, e.g. .toBeDisabled() and jest doesn't support overriding matchers defined in setupFilesAfterEnv.  Hopefully, moving the matcher imports into the test files is enough to allow .toBeDisabled() to be used by enzyme and testing-library tests.

Issue: None

## Test plan:
- yarn test


config/jest/test-setup.js
```diff
@@ -20,40 +19,6 @@ StyleSheetTestUtils.suppressStyleInjection();
 Enzyme.configure({adapter: new EnzymeAdapter()});
 expect.addSnapshotSerializer(enzymeSerializer);
 
-// Copied from jest-enzyme.
```
jandrade, line 23: Would this still be a problem? or is it ok if we leave that wrong JSDOM env open in the meantime until we fully migrate to RTL?

jandrade (approved):
LGTM! I just added a question about removing the matchers from the setup, but I think that doesn't block these changes from being landed. Thanks for fixing this so quickly!!

config/jest/test-setup.js
kevinbarabash: I'll see which version of JSDOM the tests are actually running with.  If it's an old version I can add a resolution to our `package.json`.

kevinbarabash: I added a test to verify that we're using v16 of JSDOM in our tests.  No changes appear to be necessary in terms of how we're consuming `jest-enzyme`.

PR #1255, opened 2021-12-14
FEI-4212.1: rename timing hooks based on API design spec

kevinbarabash:
## Summary:
I thought initial that the more complicated hooks might be the default but after auditing our current use of 'withActionScheduler', it turned out that the majority of existing use cases can be met with the simpler API so that will become the default.  In the next PR, I'll export add a 'useTimeout' that mimics the current 'useInterval'.

Issue: FEI-4212

## Test plan:
- yarn jest

### DIFF
--- a/packages/wonder-blocks-timing/src/index.js
+++ b/packages/wonder-blocks-timing/src/index.js
@@ -21,4 +21,5 @@ export type {
 
 export {SchedulePolicy, ClearPolicy} from "./util/policies.js";
 export {default as withActionScheduler} from "./components/with-action-scheduler.js";
-export {useTimeout} from "./hooks/use-timeout.js";
+export {useScheduledInterval} from "./hooks/use-scheduled-interval.js";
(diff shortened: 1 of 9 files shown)


packages/wonder-blocks-timing/src/index.js
```diff
@@ -21,4 +21,5 @@ export type {
 
 export {SchedulePolicy, ClearPolicy} from "./util/policies.js";
 export {default as withActionScheduler} from "./components/with-action-scheduler.js";
-export {useTimeout} from "./hooks/use-timeout.js";
+export {useScheduledInterval} from "./hooks/use-scheduled-interval.js";
```
jeresig, line 24: It seems like we should be exporting `useInterval` too?

jeresig (approved):

packages/wonder-blocks-timing/src/index.js
kevinbarabash: I'll export it in the next PR.  I want to add docs for it as well.

PR #1256, opened 2021-12-14
WB-1128: Add lang attribute to ActionItem

jandrade:
## Summary:

- Adds the `lang` prop/attribute to the `ActionItem` element so screen readers
can use this to work properly with JAWS in Windows.
- Adds a new Story to verify this.
- Converts one test from Enzyme to RTL.

Issue: WB-1128

## Test plan:

Verify the Storybook instance to see if the `lang` is passed down to the `span` element in the ActionItem component defined in the `ActionMenu` story.

https://5e1bf4b385e3fb0020b7073c-bkcroffsww.chromatic.com/?path=/story/dropdown-actionmenu--action-menu-with-lang
Screen Shot 2021-12-14 at 5 15 57 PM


packages/wonder-blocks-dropdown/src/components/__tests__/action-item.test.js
```diff
@@ -1,33 +1,53 @@
...
+        // Assert
+        expect(screen.getByRole("menuitem")).toBeDisabled();
     });
 
     it("should render an anchor if there's no router", () => {
-        const wrapper = mount(<ActionItem href="/foo" />);
+        // Arrange
 
+        // Act
+        const wrapper = mount(<ActionItem href="/foo" label="Example" />);
```
jandrade, line 24: Keeping this with Enzyme because I still don't find a clean way to find elements by tagName with RTL.

packages/wonder-blocks-dropdown/src/components/__tests__/action-item.test.js
```diff
@@ -1,33 +1,53 @@
...
+
+        // Act
         const wrapper = mount(
             <MemoryRouter>
-                <ActionItem href="/foo" />
+                <ActionItem href="/foo" label="Example" />
             </MemoryRouter>,
         );
 
+        // Assert
```
jandrade, line 40: Same thing here because RTL cannot query components. I think that's what's ADR 526 is about.

packages/wonder-blocks-dropdown/src/components/__tests__/action-item.test.js
kevinbarabash: You should be able to use `getByRole("link")` in this situation.  If you really needed to query by tag name you could do `document.querySelectorAll("a")`.

packages/wonder-blocks-dropdown/src/components/action-item.js
```diff
@@ -38,6 +38,12 @@ type ActionProps = {|
      */
     href?: string,
 
+    /**
+     * Optional attribute to indicate to the Screen Reader which language the
+     * item text is in.
+     */
+    lang?: string,
```
kevinbarabash, line 45: Are there any other places that we should be adding this prop?  If so, maybe file a JIRA to take care of the rest in the new year.

kevinbarabash (approved):
LGTM.  Thanks for updating the tests to use @testing-library.  I think one of the tests that's still using `enzyme`, may be able to be converted right now, see inline comments for details.

packages/wonder-blocks-dropdown/src/components/__tests__/action-item.test.js
jandrade: The thing is that we are passing down a `role` prop so the component uses `menuitem` by default. That's why I needed to keep using `a` for that case, but the RTL linter complains about that, so we would need to use `// eslint-disable-next-line testing-library/no-node-access` to fix that. I'm wondering if this test becomes redundant with RTL.

packages/wonder-blocks-dropdown/src/components/action-item.js
jandrade: Not required yet, but better to tackle this in advance for sure!

packages/wonder-blocks-dropdown/src/components/__tests__/action-item.test.js
kevinbarabash: Having to add `// eslint-disable-next-line testing-library/no-node-access` from time to time doesn't seem that bad.  You could even disable that rule for this entire test file if need be.

PR #1257, opened 2021-12-14
Allow disabled buttons to be focusable by screenreaders

nishasy:
## Summary:
"There might be controls like a button that's disabled and we want to be able to communicate why it's disabled with a tooltip.  We want that tooltip to be accessible even if the control is disabled.  Disabled controls may not tabbable but users can still navigate to them via Voice Over and other screen readers."

- Change `disabled` prop to `aria-disabled`
- Change tabIndex to 0
- We don't have to add an eventListener to stop clicks from happening because that is already handled here:
  https://github.com/Khan/wonder-blocks/blob/master/packages/wonder-blocks-clickable/src/components/clickable-behavior.js#L213

Issue: https://khanacademy.atlassian.net/browse/WB-393

Screen reader:

https://user-images.githubusercontent.com/13231763/146093822-5fc4a327-2228-43a3-b2dc-f291c6b90cf9.mov

Just tabbing:

https://user-images.githubusercontent.com/13231763/148317084-3bf55d9d-0f87-4885-8e45-e50430bf61b6.mov

With tooltips:

https://user-images.githubusercontent.com/13231763/148317089-58cd708e-1548-4200-a771-e7e182fc94b9.mov

## Test plan:
- Open storybook and go to the Basic Buttons tab
- Tab through the buttons. They should all receive focus
  (but you won't be able to see it for the disabled buttons)
- Turn on a screen reader
- Tab through the buttons. Confirm that the screenreader
  reads the disabled buttons as "dimmed"


packages/wonder-blocks-clickable/src/components/clickable-behavior.js
```diff
@@ -223,7 +223,9 @@ const disabledHandlers = {
     onKeyUp: () => void 0,
     onFocus: () => void 0,
     onBlur: () => void 0,
-    tabIndex: -1,
+    // Clickable components should still be tabbable so they can
+    // be used as anchors.
+    tabIndex: 0,
 };
```
kevinbarabash, line 229: I think it would be good to have some unit tests that verify that this is the case.

packages/wonder-blocks-button/src/components/button-core.js
```diff
@@ -162,7 +162,7 @@ export default class ButtonCore extends React.Component<Props> {
                 <StyledButton
                     type={type || "button"}
                     {...commonProps}
-                    disabled={disabled}
+                    aria-disabled={disabled}
```
kevinbarabash, line 165: How does `aria-disabled` differ from `disabled`?  Should we be using `aria-disabled` in other clickable components like `IconButton` and the various drop-downs and the items the contain?

kevinbarabash (commented):
Thanks for improving button accessibility.  I think there are maybe a few other components that could probably use the same treatment.  You may want to create a separate JIRA task and separate PR for them (not sure what the scope of the original task was).

Is being focusable by screenreader different from just tabbing?  Would the disabled button still not be tabbable to?

packages/wonder-blocks-button/src/components/button-core.js
jandrade: Yeah that's a good question. I remember listing out some of the elements that would be nice to update with this pattern, but probably a good follow up ticket after this is landed.

kevinbarabash:
I read the Slack thread that the ticket linked to and it sounds like the suggestion is to make disabled buttons actually focusable/tabbable.

Also, the JIRA ticket is about allowing tooltips to appear on disabled items so I think that the test plan should include something to show that tabbing to disabled items will show any associated tooltip.  Ideally there would be `@testing-library` tests for this so that we don't have to worry about accidentally breaking this behaviour in the future.

packages/wonder-blocks-button/src/components/__tests__/button.test.js
```diff
@@ -785,6 +786,56 @@ describe("Button", () => {
...
+
+            // Act
+            const button = screen.getByTestId("button-focus-test");
+            button.focus();
+
+            // Assert
+            expect(button).toHaveFocus();
+        });
+
+        test("can be focused when disabled", async () => {
```
jeresig, line 818: I think we need to update the names of these tests to make it clear we're testing primary/tertiary/etc.

packages/wonder-blocks-clickable/src/components/clickable-behavior.js
jeresig: Nisha and I worked on implementing some focusable tests - but testing focus in RTL (or even Enzyme) is a giant mess. And in the end we're not even confident that it's working correctly.

jeresig (approved):
This is looking good to me, sorry it proved to be so challenging to test!

packages/wonder-blocks-button/src/components/button-core.js
```diff
@@ -327,9 +332,18 @@ const _generateStyles = (color, kind, light, iconWidth, size) => {
                 color: light ? white50 : offBlack32,
                 borderColor: light ? fadedColor : offBlack32,
                 cursor: "default",
+                ":focus": {
+                    borderColor: light ? white50 : offBlack32,
+                    borderWidth: 2,
+                    // The left padding for the button with icon should have 4px
+                    // less padding
+                    paddingLeft: iconWidth ? padding - 5 : padding - 1,
```
jandrade, line 340: Nit: It would be probably good to include a comment explaining why this differs from the `default` state. Also, it would probably be helpful to move these values into a constant so it's easier to keep track of those subtle changes (-5, -1, etc).

packages/wonder-blocks-tooltip/src/components/tooltip.stories.js
```diff
@@ -48,6 +51,25 @@ const BaseTooltipExample = ({placement}: {|placement: Placement|}) => {
...
+                    content="This is a tooltip on a disabled button."
+                    placement="bottom"
+                >
+                    <Button disabled={true}>Example 2</Button>
+                </Tooltip>
+            </View>
+        </View>
+    );
+};
+
```
jandrade, line 72: We can probably skip publishing this story snapshot to Chromatic because the tooltip will only be displayed if the user focuses on the button.

```suggestion

TooltipOnButtons.parameters = {
    chromatic: {
        disableSnapshot: true,
    },
};

```

For more context, this is what you see in Chromatic and I don't think this visual test will need to check the tooltip itself: https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=61dcdd6e6e34b6003afb27ba

jandrade (approved):
Looks good to me!  I just added a couple of minor comments for you to consider before landing. Thanks for working on these changes and adding the tests.

PR #1270, opened 2021-12-16
Update RenderStateRoot to support nesting

kevinbarabash:
## Summary:
While we want to only have a single RenderStateRoot, the way that client-side and server-side rendering is currently set up in webapp requires us to allow nesting.  When nesting though, we only rendering a single context provider.  To support the nesting of this component, I've add a 'throwIfNested' prop.

Issue: none

## Test plan:
- yarn jest

### DIFF
--- a/packages/wonder-blocks-core/src/components/__tests__/render-state-root.test.js
+++ b/packages/wonder-blocks-core/src/components/__tests__/render-state-root.test.js
@@ -25,7 +25,7 @@ describe("RenderStateRoot", () => {
 
         // Act
         render(
-            <RenderStateRoot>
+            <RenderStateRoot throwIfNested={true}>
                 <TestComponent />
             </RenderStateRoot>,
         );
@@ -47,7 +47,7 @@ describe("RenderStateRoot", () => {
 
         // Act
         render(
-            <RenderStateRoot>
+            <RenderStateRoot throwIfNested={true}>
                 <TestComponent />
             </RenderStateRoot>,
         );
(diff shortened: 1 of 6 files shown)


packages/wonder-blocks-core/src/components/render-state-root.js
```diff
@@ -11,23 +11,35 @@ const {useContext, useEffect, useState} = React;
...
-    if (contextValue !== RenderState.Root) {
-        throw new Error(
-            "There's already a <RenderStateRoot> above this instance in " +
-                "the render tree.  This instance should be removed.",
-        );
-    }
-
     useEffect(() => {
         setFirstRender(false);
     }, []); // This effect will only run once.
```
kevinbarabash, line 29: I had to move this up because of hooks rules.

packages/wonder-blocks-core/src/components/__tests__/render-state-root.test.js
```diff
@@ -69,4 +69,23 @@ describe("RenderStateRoot", () => {
...
+                </RenderStateRoot>,
+            );
+
+        expect(underTest).not.toThrowError();
+    });
+
+    // This test can be written once ADR #526 has been implemented.  I've left
+    // a comment in the ADR about this kind of test case being something we
+    // should support.
+    it.todo("should only render a single context provider when nesting");
```
jandrade, line 90: Cool thanks!

jandrade (approved):
Makes sense to me 🚀

PR #1273, opened 2021-12-21
WB-1173.1: Add wonder-blocks-cell package

jandrade:
## Summary:

This PR only introduces a new package for Cell components: `wonder-blocks-cell`.

Issue: WB-1173

## Test plan:

Run `yarn test`.


packages/wonder-blocks-cell/src/components/__docs__/basic-cell.argtypes.js
```diff
@@ -0,0 +1,88 @@
+// @flow
```
jandrade, line 1: This file is used by Storybook so we can display the Props documentation properly.
Screen Shot 2021-12-21 at 6 45 29 PM

packages/wonder-blocks-cell/src/components/__docs__/basic-cell.stories.js
```diff
@@ -0,0 +1,204 @@
...
+import BasicCell from "../basic-cell.js";
+
+export default {
+    title: "Layout / Cell / Basic Cell",
+    component: BasicCell,
+    argTypes: BasicCellArgTypes,
+    parameters: {
+        componentSubtitle: ((
+            <ComponentInfo name={name} version={version} />
+        ): any),
```
jandrade, line 27: This is used to include a link to GH and show the current package version.
Screen Shot 2021-12-21 at 6 47 28 PM

packages/wonder-blocks-cell/src/util/types.js
```diff
@@ -0,0 +1,155 @@
...
+ * A set of values that can be used to configure the horizontal rule appearance.
+ */
+type HorizontalRuleVariant = "full-width" | "inset" | "none";
+
+/**
+ * A subset of CSS Properties to allow overriding some of the default styles set
+ * on the accessory wrapper (loosely based on StyleType).
+ */
+export type AccessoryStyle = {|
+    minWidth?: 16 | 24 | 32 | 48,
```
lillialexis, line 17: Do we want to use raw numbers or `Spacing....`?

packages/wonder-blocks-cell/src/util/types.js
```diff
@@ -0,0 +1,155 @@
...
+ * `rightAccessory` are set.
+ */
+export type CellProps =
+    /**
+     * leftAccessory
+     */
+    | {|
+          ...SharedProps,
+          leftAccessory?: React.Node,
+      |}
```
lillialexis, line 66: Do we need to do this kind of `CellProps` where it's a union of exact types? Feels like we will have _way_ too many combos once we add in DetailCell/subtitle, etc.

lillialexis (commented):
I did a quick once-over before I take off for the holidays. It wasn't a super thorough review. Is there anyone else around who can also review this? I don't want to block you. Just slack-DM me if you need me to deeper-review and approve this.

jandrade:
> I did a quick once-over before I take off for the holidays. It wasn't a super thorough review. Is there anyone else around who can also review this? I don't want to block you. Just slack-DM me if you need me to deeper-review and approve this.

Don't worry about that. The idea is that I'm going to be creating all the stacked PRs in the next couple of days, so y'all can review the week after the holidays. Then I'll make changes based on the reviews once I'm back.

packages/wonder-blocks-cell/src/util/types.js
jandrade: Yep, sadly it's one of the downsides of using this approach, we need to cover all these cases. I'm going to try it out this week with the DetailCell to see how it goes.

We could also use a single type (making leftAccessoryStyle? optional directly in `SharedProps`), but then we would got into the "make the impossible states unrepresentable" opposite case, because there would be the case where devs don't set `leftAccessory`, but set a value for `leftAccessoryStyle`.

jandrade:
Update: Moved the BasicCell implementation to a separate PR.

lillialexis (approved):

PR #1277, opened 2021-12-23
WB-1173.4: Add states to BasicCell (hovered, focused, pressed, active, disabled)

jandrade:
## Summary:
- Adds the cell states (hovered, focused, pressed, active, disabled).
- Adds `onClick` to include a pressable cell.
- Adds the `aria-label` prop.

### Plan:
1. Add wonder-blocks-cell package #1273
2. Add BasicCell component layout (`title, leftAccessory, leftAccessoryStyle, rightAccessory, rightAccessoryStyle`). #1275
3. Add `horizontalRule, style, testId`.  1276
4. **Add states to BasicCell (hovered, focused, pressed, active, disabled) [CURRENT]**

Issue: WB-1173

## Test plan:

- Verify the visual snapshots in Chromatic.
- Also verify that the Storybook instance renders all the stories and
  includes the documentation properly.

[BasicCell active](https://5e1bf4b385e3fb0020b7073c-fwzsprnuan.chromatic.com/?path=/story/cell-basiccell--basic-cell-active)
Screen Shot 2021-12-23 at 5 08 16 PM

[BasicCell disabled](https://5e1bf4b385e3fb0020b7073c-fwzsprnuan.chromatic.com/?path=/story/cell-basiccell--basic-cell-disabled)
Screen Shot 2021-12-23 at 5 09 19 PM

PR #1278, opened 2021-12-23
WB-1174: Implement DetailCell component

jandrade:
## Summary:

- Adds the `DetailCell` component to `wonder-blocks-cell`.
- Adds documentation and unit tests

NOTE: This is the last PR and it depends on all the other PRs created for `BasicCell`.

Issue: WB-1174

## Test plan:

- Verify the visual snapshots in Chromatic.
- Also verify that the Storybook instance renders all the stories and includes the documentation properly.

https://5e1bf4b385e3fb0020b7073c-vmauauvfua.chromatic.com/?path=/story/cell-detailcell--default-detail-cell
Screen Shot 2021-12-23 at 5 36 31 PM


packages/wonder-blocks-cell/src/components/detail-cell.js
```diff
@@ -0,0 +1,86 @@
...
+ *  title="Detail cell"
+ *  subtitle1="Subtitle 2"
+ *  rightAccessory={<Icon icon={icons.caretRight} size="medium" />}
+ * />
+ * ```
+ */
+function DetailCell(props: DetailCellProps): React.Node {
+    const {title, subtitle1, subtitle2, ...coreProps} = props;
+
+    const maybeRenderSubtitle = (subtitle?: TypographyText): React.Node => {
```
jeresig, line 53: Same note here as in the BasicCell, you could have this be `Subtitle` and do `<Subtitle .../>`

jeresig (approved):
Great, thank you!

packages/wonder-blocks-cell/src/components/detail-cell.js
```diff
@@ -0,0 +1,86 @@
...
+     * You can either provide a string or a custom node Typography element (or
+     * nothing at all). Both a string or a custom node Typography element will
+     * occupy the “Subtitle2” area of the Cell.
+     */
+    subtitle2?: TypographyText,
+|};
+
+/**
+ * This is a variant of BasicCell that allows adding subtitles, before and after
+ * the cell title.
```
lillialexis, line 34: "It also have more vertical padding, and the Title is bold."

packages/wonder-blocks-cell/src/components/detail-cell.js
```diff
@@ -0,0 +1,86 @@
...
+    return (
+        <CellCore {...coreProps}>
+            {maybeRenderSubtitle(subtitle1)}
+            {typeof title === "string" ? (
+                <LabelMedium>{title}</LabelMedium>
+            ) : (
+                title
+            )}
+            {/* Add a vertical spacing between the title and the subtitle */}
+            {subtitle2 && <Strut size={Spacing.xxxxSmall_2} />}
```
lillialexis, line 74: I think we need to land on final values for these spacing values.  https://khanacademy.atlassian.net/browse/MOB-4368

lillialexis: Also, it should probably be the same as the vertical spacing between the above one and the title. (cc @megaroeny )

packages/wonder-blocks-cell/src/components/detail-cell.js
```diff
@@ -0,0 +1,86 @@
...
+
+        if (typeof subtitle === "string") {
+            return <LabelSmall style={styles.subtitle}>{subtitle}</LabelSmall>;
+        }
+
+        return subtitle;
+    };
+
+    return (
+        <CellCore {...coreProps}>
```
lillialexis, line 66: I think because of the way that this is constructed, the `disabled` state isn't changing the subtitles, and that looks weird to me. It's also not consistent w mobile in this regard.

lillialexis (approved):
I'm still not sure if all the Stories are working as expected. 

Also, with really long (unbroken) text, it overlaps and overruns the edge of the cell:
Screen Shot 2022-01-05 at 5 33 39 PM


See mobile cell wrapping:
File

packages/wonder-blocks-cell/src/components/detail-cell.js
megaroeny: I will double-check that what is in the Figma `.base` component is what we want. But need to confirm with you and Juan. Let's quickly review together tomorrow if you have a brief moment. I think I wanted to have it be 2px above/below the title, especially with letters like a _g_, it's descender/tail was too close to the title.

jandrade: Yeah, that was the main reasoning about using 2px. We tested that directly in the browser and ended up picking `2px` on the subtitle because it rendered better with the cases that @megaroeny mentioned.

packages/wonder-blocks-cell/src/components/detail-cell.js
jandrade: I'm going to be revisiting the `disabled` states after our last conversation.

jandrade:
The parent pull-request ([#1277](https://github.com/Khan/wonder-blocks/pull/1277)) has been merged into `feature/cell`, but this branch (`cell-d1`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #1280, opened 2022-01-05
Implement useInternalData

somewhatabstract:
## Summary:
This hook does all the work that the `InternalData` component did.

This also modifies the `Result` type to use a `status` value so that
the data is validated as loading, success, or error up front.

Finally, this removes the client-side caching.

Those last two things make this a breaking change, but that should be easy to cope with
given that WB Data is as yet not widely adopted.

I wanted to try and split this into smaller PRs but it was not so easy to do given the order in which I had been doing the work. However, since much of the change is deleting code and the only new code is either tests or the `useDataInternal` hook, I felt that it wouldn't complicate things terribly.

### Next up
The PR after this one will then implement the `useData` hook, supporting
the cache and request interception context as per the current `Data`
component and we can then replace the `Data` component with one that
uses the hooks. The `InternalData` component will go away in the next PR.

Issue: FEI-4219

## Test plan:
`yarn test`


packages/wonder-blocks-data/src/hooks/use-data-internal.js
```diff
@@ -0,0 +1,91 @@
...
+                    `Unexpected error occurred during data fulfillment: ${e}`,
+                );
+                setLocalError(typeof e === "string" ? e : e.message);
+                return;
+            });
+
+        return () => {
+            cancel = true;
+        };
+    }, [cachedData, handler, options]);
```
somewhatabstract, line 80: Instead of `options` here, we really want `handler.getKey(options)`. Our linter hates that, but it's what we really want - so I'll probably make that change and suppress the linter.

somewhatabstract: This is not relevant anymore

config/jest/test.config.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
     setupFilesAfterEnv: [
         "@testing-library/jest-dom/extend-expect",
         "<rootDir>/config/jest/test-setup.js",
+        "jest-extended/all"
```
somewhatabstract, line 20: Surprised we weren't already using this in WB.

ejanzer:
The hook looks good to me! Will let some other folks take a look as well

packages/wonder-blocks-data/src/hooks/use-data-internal.js
```diff
@@ -0,0 +1,113 @@
...
+        // The request handler is in control of whether that request actually
+        // happens or not.
+        let cancel = false;
+        RequestFulfillment.Default.fulfill(handler, options)
+            .then((updateEntry) => {
+                if (cancel) {
+                    return;
+                }
+                setResult(updateEntry);
+                return;
```
kevinbarabash, line 86: style nit: `return` isn't necessary here.

packages/wonder-blocks-data/src/hooks/use-data-internal.js
```diff
@@ -0,0 +1,113 @@
...
+                 */
+                // eslint-disable-next-line no-console
+                console.error(
+                    `Unexpected error occurred during data fulfillment: ${e}`,
+                );
+                setResult({
+                    data: null,
+                    error: typeof e === "string" ? e : e.message,
+                });
+                return;
```
kevinbarabash, line 104: style nit: `return` isn't necessary here.

packages/wonder-blocks-data/src/util/__tests__/memory-cache.test.js
```diff
@@ -343,5 +319,9 @@ describe("MemoryCache", () => {
         });
     });
 
-    describe("#cloneData", () => {});
+    describe("#cloneData", () => {
+        it("should return a copy of the cache data", () => {});
+
+        it("should throw if there is an error during cloning", () => {});
+    });
```
kevinbarabash, line 326: Are these to be written in a future PR?  If so, consider using `it.todo("...")` in the meantime.

packages/wonder-blocks-data/src/util/__tests__/request-fulfillment.test.js
```diff
@@ -17,32 +17,30 @@ describe("RequestFulfillment", () => {
...
-                    MY_KEY: {
-                        error: "OH NO!",
-                    },
-                },
-            });
+            expect(cacheErrorSpy).toHaveBeenCalledWith(
+                fakeBadHandler,
+                "OPTIONS",
+                error,
+            );
```
kevinbarabash, line 43: Nice!

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
```diff
@@ -112,779 +108,448 @@ describe("../response-cache.js", () => {
...
-                expect(storeSpy).toHaveBeenCalledWith(fakeHandler, "options", {
-                    data: "data",
-                });
-            });
-
-            it("should not store the entry in the framework cache", () => {
-                // Arrange
+        describe("when server-side", () => {
+            beforeEach(() => {
                 jest.spyOn(Server, "isServerSide").mockReturnValue(true);
```
kevinbarabash, line 154: This feels like a good use of shared `Arrange` since it's literally in the title of the describe block.

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
```diff
@@ -112,779 +108,448 @@ describe("../response-cache.js", () => {
...
-                    shouldRefreshCache: () => false,
-                    fulfillRequest: jest.fn(),
-                    cache: null,
-                    hydrate: false,
-                };
+            describe("when the handler does not hydration", () => {
+                it("should store the entry in the ssr-only cache", () => {
+                    // Arrange
+                    const hydrationCache = new MemoryCache();
+                    const ssrOnlyCache = new MemoryCache();
```
kevinbarabash, line 212: I totally forgot that these were two separate things. 😅

packages/wonder-blocks-data/src/util/request-fulfillment.js
```diff
@@ -76,13 +76,17 @@ export class RequestFulfillment {
                     delete handlerRequests[key];
                     /**
                      * Let's cache the data!
+                     *
+                     * NOTE: This only caches when we're server side.
                      */
                     return cacheData<TOptions, TData>(handler, options, data);
```
kevinbarabash, line 82: This is because we only track the request fulfillments on the server, right?

packages/wonder-blocks-data/src/util/response-cache.js
```diff
@@ -101,6 +83,8 @@ export class ResponseCache {
 
     /**
      * Cache data for a specific response.
+     *
+     * This is a noop when client-side.
```
kevinbarabash, line 87: Thanks for adding in comments like this to indicate differences in behaviour between client and server.

packages/wonder-blocks-data/src/util/response-cache.js
```diff
@@ -140,42 +124,27 @@ export class ResponseCache {
...
             // This does mean that if another handler of the same type but
-            // without a custom cache won't get the value, but that's not an
-            // expected valid usage of this framework - two handlers with
-            // different caching options shouldn't be using the same type name.
-            this._hydrationAndDefaultCache.remove(handler, options);
+            // without some sort of linked cache won't get the value, but
+            // that's not an expected use-case. If two different places use the
+            // same handler and options (i.e. the same request), then the
+            // handler should cater to that to ensure they share the result.
+            this._hydrationCache.remove(handler, options);
```
kevinbarabash, line 147: I was wondering about this.  Thanks for adding this comment.

packages/wonder-blocks-data/src/util/types.js
```diff
@@ -1,16 +1,19 @@
...
     | {|
-          loading: false,
+          status: "success",
           data?: TData,
-          error?: string,
+      |}
+    | {|
+          status: "error",
+          error: string,
       |};
```
kevinbarabash, line 17: Nice!

kevinbarabash (approved):
LGTM.  Thank you for all of the comments in the code, it help with my understanding of things.  I have some minor perf concerns about parts of `useInternalData` so please look at those comments before landing.  It's nothing blocking though.

packages/wonder-blocks-data/src/hooks/__tests__/use-data-internal.test.js
```diff
@@ -0,0 +1,578 @@
...
+            jest.spyOn(Server, "isServerSide").mockReturnValue(false);
+        });
+
+        it("should initially return loading state when no cached result", () => {
+            // Arrange
+            const fakeHandler: IRequestHandler<string, string> = {
+                fulfillRequest: jest.fn().mockReturnValue(
+                    new Promise(() => {
+                        /*prevent act() warning*/
+                    }),
```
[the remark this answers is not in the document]
somewhatabstract: Leaving the promise pending is what prevents the act warnings because we won't change state if the promise doesn't resolve or reject.

packages/wonder-blocks-data/src/util/request-fulfillment.js
somewhatabstract: It's because we only cache for the server to make it so that we can render things on initial render.

packages/wonder-blocks-data/src/hooks/use-data-internal.js
```diff
@@ -0,0 +1,113 @@
...
+
+export const useDataInternal = <TOptions, TData: ValidData>(
+    handler: IRequestHandler<TOptions, TData>,
+    options: TOptions,
+): Result<TData> => {
+    // If we're server-side or hydrating, we'll have a cached entry to use.
+    // So we get that and use it to initialize our state.
+    // This works in both hydration and SSR because the very first call to
+    // this will have cached data in those cases as it will be present on the
+    // initial render - and subsequent renders on the client it will be null.
```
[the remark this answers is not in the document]
somewhatabstract: All the ones that were used, which should be all of them. If it isn't, then hydration is broken because we requested stuff server-side for things that don't get rendered client-side.

packages/wonder-blocks-data/src/hooks/use-data-internal.js
```diff
@@ -0,0 +1,113 @@
...
+    );
+    const [result, setResult] = useState<?CacheEntry<TData>>(cachedResult);
+
+    // We only track data requests when we are server-side and we don't
+    // already have a result, as given by the cachedData (which is also the
+    // initial value for the result state).
+    const maybeTrack = useContext(TrackerContext);
+    if (result == null && Server.isServerSide()) {
+        maybeTrack?.(handler, options);
+    }
```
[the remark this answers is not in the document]
somewhatabstract: Yes, if the code rendering on the server didn't use `TrackData` to enable tracking, then tracking won't be there.

packages/wonder-blocks-data/src/hooks/use-data-internal.js
```diff
@@ -0,0 +1,113 @@
...
+
+    // We need to update our request when the handler changes or the key
+    // to the options change, so we keep track of those.
+    // However, even if we are hydrating from cache, we still need to make the
+    // request at least once, so we do not initialize these references.
```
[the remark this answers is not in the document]
somewhatabstract: No, request always means request from the handler.

packages/wonder-blocks-data/src/hooks/use-data-internal.js
```diff
@@ -0,0 +1,113 @@
...
+        // NOTE: This shouldn't happen since effects would not run on the server
+        // but let's be defensive - I think it makes the code clearer.
+        if (Server.isServerSide()) {
+            return;
+        }
+
+        // If the handler hasn't changed and the key that the options represent
+        // hasn't changed, then we don't need to make our request.
+        if (
+            handler === handlerRef.current &&
```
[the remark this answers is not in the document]
somewhatabstract: It depends on what triggers the effect to run again. If it is something in this check, then yes, likely not needed, but if it were something else (like the cachedResult), it would be. Since it would be easy to introduce a problem there, I felt being explicit made the code more robust and easier to grok.

packages/wonder-blocks-data/src/hooks/use-data-internal.js
somewhatabstract: It is because of our linter.

packages/wonder-blocks-data/src/hooks/use-data-internal.js
somewhatabstract: It is, because of our linter.

packages/wonder-blocks-data/src/hooks/use-data-internal.js
```diff
@@ -0,0 +1,113 @@
...
+                    data: null,
+                    error: typeof e === "string" ? e : e.message,
+                });
+                return;
+            });
+
+        return () => {
+            cancel = true;
+        };
+    }, [handler, options, handlerRef, keyRef, cachedResult]);
```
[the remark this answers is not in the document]
somewhatabstract: The linter is really annoying then as it demands they be included.

I think the linter for deps is really leading to bad code and would be better to be disabled permanently.

packages/wonder-blocks-data/src/util/__tests__/memory-cache.test.js
somewhatabstract: They are getting added shortly. Just felt it was more work than necessary for this PR that was already giant.

packages/wonder-blocks-data/src/hooks/use-data-internal.js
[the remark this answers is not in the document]
kevinbarabash: I missed `RequestFulfillment.Default.fulfill(handler, options)`.  I think disabling the lint rule in this situation makes sense because we have a custom condition of when things should be re-run.  We could probably work around the linter by storing `options` in a `ref`, but I'd that seems like the tail wagging the dog.

somewhatabstract:
@kevinbarabash I think your main comments are addressed in the next PR #1282.

PR #1282, opened 2022-01-11
Implement useData hook

somewhatabstract:
## Summary:
This builds on the `useDataInternal` hook, combining the data interception logic of `Data` with `useDataInternal` to create `useData`.

In doing so, we delete `InternalData` and `useDataInternal`, and simplify the `Data` component to just invoking the new hook.
This also converts tests to React Testing Library.

Issue: FEI-4219

## Test plan:
`yarn jest packages/wonder-blocks-data`


packages/wonder-blocks-data/src/components/__tests__/data.test.js
```diff
@@ -1,8 +1,7 @@
 /* eslint-disable max-lines */
 // @flow
 import * as React from "react";
-import {mount, shallow} from "enzyme";
-import "jest-enzyme";
+import {render, act} from "@testing-library/react";
```
kevinbarabash, line 4: Thanks for updating these tests to use `@testing-library`.

packages/wonder-blocks-data/src/components/__tests__/data.test.js
```diff
@@ -261,22 +262,24 @@ describe("Data", () => {
...
-                const wrapper = mount(
+                const wrapper = render(
                     <Data handler={fakeHandler} options={"options"}>
                         {fakeChildrenFn}
                     </Data>,
                 );
                 // We want to make sure we render the error state so we can
                 // see our switch back to loading.
-                await fulfillSpy.mock.results[0].value;
+                await act(() => fulfillSpy.mock.results[0].value);
```
kevinbarabash, line 272: Seeing this got me thinking maybe we should have something similar to trigger GraphQL request fulfilment in tests so that we can wrap that inside an `await act(() => {})` as well instead of having the `await wait(0)` or `await waitFor(() => {...})`.

packages/wonder-blocks-data/src/components/__tests__/track-data.test.js
```diff
@@ -30,10 +29,11 @@ describe("TrackData", () => {
         jest.spyOn(Server, "isServerSide").mockReturnValue(true);
 
         // Act
-        const result = shallow(<TrackData>SOME CHILDREN</TrackData>);
+        render(<TrackData>SOME CHILDREN</TrackData>);
+        const result = screen.getByText("SOME CHILDREN");
 
         // Assert
-        expect(result).toHaveHTML("SOME CHILDREN");
+        expect(result).toBeDefined();
```
kevinbarabash, line 36: nit: `.toBeInTheDocument()` might be better here.

packages/wonder-blocks-data/src/components/data.js
```diff
@@ -52,63 +45,10 @@ type Props<
...
-            </InterceptContext.Consumer>
-        );
-    }
-}
+const Data = <TOptions, TData: ValidData>(
+    props: Props<TOptions, TData>,
+): React.Node => {
+    const data = useData(props.handler, props.options);
+    return props.children(data);
+};
```
kevinbarabash, line 53: Nice!

packages/wonder-blocks-data/src/hooks/__tests__/use-data.test.js
```diff
@@ -446,7 +448,99 @@ describe("#useDataInternal", () => {
...
             // Act
             const render = clientRenderHook(
-                ({options}) => useDataInternal(oldHandler, options),
+                ({options}) => useData(oldHandler, options),
+                {
+                    initialProps: {options: "OLD OPTIONS"},
+                },
+            );
+            render.rerender({options: "NEW OPTIONS"});
+            await act(() => oldRequest.catch(() => {}));
```
kevinbarabash, line 457: The `.catch()` is here just to ignore errors, right?

packages/wonder-blocks-data/src/hooks/__tests__/use-data.test.js
```diff
@@ -446,7 +448,99 @@ describe("#useDataInternal", () => {
...
+            const result = render.result.all;
+
+            // Assert
+            expect(result).not.toIncludeAnyMembers([
+                {
+                    status: "error",
+                    error: "OLD ERROR",
+                },
+            ]);
```
kevinbarabash, line 466: Nice!

packages/wonder-blocks-data/src/hooks/__tests__/use-data.test.js
```diff
@@ -446,7 +448,99 @@ describe("#useDataInternal", () => {
...
+            // Act
+            const render = clientRenderHook(
+                ({handler}) => useData(handler, "options"),
+                {
+                    initialProps: {handler: oldHandler},
+                },
+            );
+            render.rerender({handler: newHandler});
+            await act(() => oldRequest.catch(() => {}));
+            const result = render.result.all; //?
```
kevinbarabash, line 509: nit: remove `//?`

packages/wonder-blocks-data/src/hooks/use-data.js
```diff
@@ -107,7 +127,12 @@ export const useDataInternal = <TOptions, TData: ValidData>(
         return () => {
             cancel = true;
         };
-    }, [handler, options, handlerRef, keyRef, cachedResult]);
+        // - handler.getKey is a proxy for options
+        // - We don't want to trigger on cachedResult changing, we're
+        //   just using that as a flag for render state if the other things
+        //   trigger this effect.
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [handler, handler.getKey(options), interceptor]);
```
kevinbarabash, line 135: Thanks! ❤️

packages/wonder-blocks-data/src/hooks/use-data.js
```diff
@@ -73,11 +74,30 @@ export const useDataInternal = <TOptions, TData: ValidData>(
...
+            if (interceptor == null) {
+                return handler;
+            }
+
+            const fulfillRequestFn = (options) =>
+                interceptor.fulfillRequest(options) ??
+                handler.fulfillRequest(options);
+            return {
+                fulfillRequest: fulfillRequestFn,
+                getKey: (options) => handler.getKey(options),
```
kevinbarabash, line 87: couldn't this be `getKey: handler.getKey`?

packages/wonder-blocks-data/src/util/__tests__/memory-cache.test.js
```diff
@@ -267,26 +284,80 @@ describe("MemoryCache", () => {
...
+
             // Act
             const result = cache.removeAll(
                 fakeHandler,
                 (k, d) => d.data === "2",
             );
-            const after = cache.cloneData();
 
             // Assert
             expect(result).toBe(2);
```
kevinbarabash, line 327: style nit: using `"2"` for the data and the result also being `2` might make people think they're related even though they're not.

packages/wonder-blocks-data/src/util/__tests__/memory-cache.test.js
```diff
@@ -306,22 +377,70 @@ describe("MemoryCache", () => {
...
             // Assert
             expect(result).toBe(3);
-            expect(after).toStrictEqual({
-                MY_HANDLER: {},
-                OTHER_HANDLER: {
+        });
+    });
+
+    describe("#cloneData", () => {
+        it("should return a copy of the cache data", () => {});
```
kevinbarabash, line 387: Are you planning on writing this test in a future PR?

packages/wonder-blocks-data/src/util/__tests__/memory-cache.test.js
```diff
@@ -306,22 +377,70 @@ describe("MemoryCache", () => {
...
+            // Assert
+            expect(act).toThrowErrorMatchingInlineSnapshot(
+                `"An error occurred while trying to clone the cache: Error: BANG!"`,
+            );
         });
     });
 
-    describe("#cloneData", () => {
-        it("should return a copy of the cache data", () => {});
+    describe("@inUse", () => {
```
kevinbarabash, line 410: Is the `#` for methods and `@` for getters a personal pattern or is this a "standard" of some sort?

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
```diff
@@ -436,105 +459,371 @@ describe("../response-cache.js", () => {
...
-            const fakeHandler: IRequestHandler<string, string> = {
-                getKey: () => "MY_KEY",
-                type: "MY_HANDLER",
-                fulfillRequest: jest.fn(),
-                hydrate: true,
-            };
 
-            // Act
-            const result = cache.getEntry(fakeHandler, "options");
+            describe("handler wants hydration", () => {
```
kevinbarabash, line 540: style nit: This describe block appears to be the same for client-side and server-side.  You could use a `describeEach` to de-dupe these.  This would also make it more explicit that this behaviour is the same between the two environments.

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
```diff
@@ -436,105 +459,371 @@ describe("../response-cache.js", () => {
...
+                    expect(result).toStrictEqual({data: "data!"});
+                });
+            });
+
+            describe("handler does not want hyrdation", () => {
+                it("should return null if not in the ssr-only cache", () => {
+                    // Arrange
+                    const hydrationCache = new MemoryCache();
+                    const ssrOnlyCache = new MemoryCache();
+                    jest.spyOn(ssrOnlyCache, "retrieve").mockReturnValue(null);
```
kevinbarabash, line 589: Do we not bother looking in the `hydrationCache` if the `ssrOnlyCache` returns `null`?

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
```diff
@@ -436,105 +459,371 @@ describe("../response-cache.js", () => {
...
+                    expect(result).toBeNull();
+                });
+
+                it("should return the cached entry if in the hydration cache", () => {
+                    // Arrange
+                    const hydrationCache = new MemoryCache();
+                    const ssrOnlyCache = new MemoryCache();
+                    jest.spyOn(ssrOnlyCache, "retrieve").mockReturnValue({
+                        data: "data!",
+                    });
```
kevinbarabash, line 614: Is the title correct for this test?  It looks like the `ssrOnlyCache` is the one with the data.

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
```diff
@@ -436,105 +459,371 @@ describe("../response-cache.js", () => {
...
-            const hydrationCache = new MemoryCache();
-            jest.spyOn(hydrationCache, "removeAll").mockReturnValue(1);
-            const cache = new ResponseCache(hydrationCache);
-            const fakeHandler: IRequestHandler<string, string> = {
-                getKey: () => "MY_KEY",
-                type: "MY_HANDLER",
-                fulfillRequest: jest.fn(),
-                hydrate: true,
-            };
+        describe("when handler wants hydration", () => {
```
kevinbarabash, line 755: Does this only happen on the client?

kevinbarabash (approved):
LGTM.  Thanks implementing this hook.  I'm curious to see how it gets used for the GraphQL hooks. 🚀

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
somewhatabstract: We don't even look in the hydrationCache if the handler says it doesn't want hydration, which is what these tests are all about.

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
somewhatabstract: I'll check

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
somewhatabstract: I'm not sure I understand the question - can you elaborate?

kevinbarabash: Does the scenario of the handling "wanting hydration" only occur on the client or would this test fail if we were on the server?

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
```diff
@@ -436,105 +459,371 @@ describe("../response-cache.js", () => {
...
-                hydrate: true,
-            };
+        describe("when client-side", () => {
+            beforeEach(() => {
+                jest.spyOn(Server, "isServerSide").mockReturnValue(false);
+            });
 
-            // Act
-            const result = cache.getEntry(fakeHandler, "options");
+            describe("handler wants hyrdation", () => {
```
somewhatabstract, line 467: NIT: spelling `hydration`

packages/wonder-blocks-data/src/util/__tests__/memory-cache.test.js
somewhatabstract: Oh, no, it should be written here. Hmm. Good catch, thank you.

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
```diff
@@ -436,105 +459,371 @@ describe("../response-cache.js", () => {
...
+                    jest.spyOn(ssrOnlyCache, "retrieve").mockReturnValue(null);
+                    const cache = new ResponseCache(
+                        hydrationCache,
+                        ssrOnlyCache,
+                    );
+                    const fakeHandler: IRequestHandler<string, string> = {
+                        getKey: () => "MY_KEY",
+                        type: "MY_HANDLER",
+                        fulfillRequest: jest.fn(),
+                        hydrate: true,
```
somewhatabstract, line 598: NIT: this should say `hydrate: false` 😱

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
somewhatabstract: We likely wouldn't call this function on the server, but if we did, it would work just fine in both places.

packages/wonder-blocks-data/src/hooks/use-data.js
somewhatabstract: Flow will complain about detaching the method from the object if we do that.

packages/wonder-blocks-data/src/util/__tests__/memory-cache.test.js
somewhatabstract: Oh yeah, nice. I'll fix that.

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
somewhatabstract: I kept them separate because they are semantically different and could easily diverge. I don't feel strongly about it, but I have a mild preference for the separation and the clarity it has in telling a story (plus matching the pattern of other tests in the file).

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
somewhatabstract: I've made a change here to mock out hydrationCache (even though it shouldn't be referenced) so that we catch if the `hydrate` value below is wrong, since it was wrong, but the test had passed. Now if it's wrong, the test will fail.

packages/wonder-blocks-data/src/util/__tests__/response-cache.test.js
somewhatabstract: Good catch. Thanks

PR #1284, opened 2022-01-12
WB-1185: Fix changesets release process

jandrade:
## Summary:

Includes the following fixes:

1. Use ss-agent so that the release PR can self-update and initiate test runs

2. Fix that message so it doesn’t say ${GITHUB_REPOSITORY} and markdown formatting issues with the link

3. Remove lerna script from package.json

4. Add experimental flag to include bump updates to dependents. This is
   added so we avoid having mismatches between dependencies in webapp
   (we used a similar approach with lerna).

Issue: WB-1185

## Test plan:

Merge then verify if the next release process works smoothly.


jeresig (approved):
Very cool - looks good to me!

.github/workflows/node-ci-push.yml
```diff
@@ -137,3 +137,53 @@ jobs:
         token: ${{ secrets.GITHUB_TOKEN }}
         projectToken: ${{ secrets.CHROMATIC_APP_CODE }}
         autoAcceptChanges: true
+
+  publish_to_branch:
```
somewhatabstract, line 141: This is pretty much a copy from the render-gateway, right, except for the repo being cloned?

I wonder if it is worth annotating that. If we end up doing some shared workflows it might be useful to know.

.github/workflows/release.yml
```diff
@@ -64,7 +64,7 @@ jobs:
           SLACK_MSG_AUTHOR: ${{ github.event.pull_request.user.login }}
           SLACK_USERNAME: GithubGoose
           SLACK_ICON_EMOJI: ":goose:"
-          SLACK_MESSAGE: "A new version of ${GITHUB_REPOSITORY} was published! [Release notes →](<https://github.com/Khan/wonder-blocks/releases/>)"
+          SLACK_MESSAGE: "A new version of ${github.event.repository.name} was published! Release notes → https://github.com/Khan/${github.event.repository.name}/releases/)"
```
somewhatabstract, line 67: Nice, but what about more emoji?

package.json
```diff
@@ -23,7 +23,6 @@
     "lint:watch": "esw --watch --config ./eslint/eslintrc packages",
     "lint": "eslint --config .eslintrc.js packages",
     "lint:ci": "eslint --config .eslintrc.js",
-    "publish": "node utils/publish/pre-publish-check.js && git diff --stat --exit-code HEAD && yarn build:all && lerna publish",
```
somewhatabstract, line 26: 👋🏻

package.json
```diff
@@ -85,7 +84,6 @@
     "husky": "^4.2.5",
     "jest": "^27.4.3",
     "jest-enzyme": "^7.1.2",
-    "lerna": "^3.16.5",
```
somewhatabstract, line 88: 👋🏻  🖖🏻

somewhatabstract (approved):
This looks good to me. Some inline things but nothing blocking or important. :)

PR #1295, opened 2022-01-14
Added `required` prop to LabeledTextField

nishasy:
## Summary:
- Add a red asterisk after the label on a LabeledTextField
- Added validation for required fields, provided the user has not already
  added their own validation through the `validate` prop

Issue: https://khanacademy.atlassian.net/browse/FEI-3485

Screen Shot 2022-01-20 at 12 45 51 PM

https://user-images.githubusercontent.com/13231763/150419796-2ea2e8ab-3dcc-4666-b76b-94dce9af6a66.mov

## Test plan:
Storybook
http://localhost:6061/?path=/story/form-textfield--required
http://localhost:6061/?path=/story/form-labeledtextfield--required-with-default-text
http://localhost:6061/?path=/story/form-labeledtextfield--required-with-specified-text
http://localhost:6061/?path=/story/form-labeledtextfield--email-required

Tests
`yarn jest packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js`


packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
```diff
@@ -485,3 +486,133 @@ describe("LabeledTextField", () => {
...
+            />,
+        );
+
+        const textField = screen.getByTestId("test-labeled-text-field-field");
+        textField.focus();
+        userEvent.type(textField, "a");
+
+        await wait(0);
+
+        console.log("value", textField.value);
```
nishasy, line 580: (WIP comment):
I can't seem to get the field to change its value in the test.

It looks like the stories in storybook have to use some sort of helper to get the value to change on keyboard input?

https://github.com/Khan/wonder-blocks/blob/master/packages/wonder-blocks-form/src/components/labeled-text-field.stories.js#L20

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -134,11 +149,13 @@ class TextFieldInternal extends React.Component<PropsWithForwardRef, State> {
         type: "text",
         disabled: false,
         light: false,
+        requiredErrorMessage: "This field is required.",
     };
 
     constructor(props: PropsWithForwardRef) {
         super(props);
-        if (props.validate) {
+        // Value should never be undefined since it is a required prop
```
nishasy, line 157: oops, I'll remove this comment, it's not supposed to be there

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -150,18 +167,30 @@ class TextFieldInternal extends React.Component<PropsWithForwardRef, State> {
     };
 
     componentDidMount() {
-        this.maybeValidate(this.props.value);
+        // Value should never be undefined since it is a required prop
```
nishasy, line 170: same with this comment

packages/wonder-blocks-form/src/components/text-field.stories.js
```diff
@@ -40,6 +40,30 @@ export const Text: StoryComponentType = () => {
...
+    };
+
+    return (
+        <TextField
+            id="tf-1"
+            type="text"
+            onChange={handleChange}
+            onKeyDown={handleKeyDown}
+            required={true}
+        />
```
nishasy, line 63: oops, I forgot to add `value={value}` here.

Why do we need to do it that way anyway? When I use `value=""`, it doesn't let me type anything into the box

packages/wonder-blocks-form/src/components/field-heading.js
```diff
@@ -28,6 +28,11 @@ type Props = {|
      */
     description?: string | React.Element<Typography>,
 
+    /**
+     * Whether this field is required to to continue.
```
jandrade, line 32: Typo: 
```suggestion
     * Whether this field is required to continue.
```

What do you think about framing this in a way where you describe the validation intention?

Something like `Whether this field is required and validated`

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -41,6 +41,27 @@ type Props = {|
...
+    /**
+     * Whether this field is required to to continue.
+     *
+     * If the `validate` prop is not being passed in in conjuction with
+     * the `required` prop, please use the `requiredErrorMessage` prop
+     * to specify the error message.
+     */
+    required?: boolean,
+
+    /**
```
jandrade, line 53: This is super well explained!

packages/wonder-blocks-form/src/components/labeled-text-field.stories.js
```diff
@@ -37,6 +37,49 @@ export const Text: StoryComponentType = () => {
...
+            description="Please enter your name"
+            value={value}
+            onChange={(newValue) => setValue(newValue)}
+            onKeyDown={handleKeyDown}
+            required={true}
+        />
+    );
+};
+
+export const RequiredWithSpecifiedText: StoryComponentType = () => {
```
jandrade, line 61: I definitively think that we don't need/want to send this story to Chromatic for visual regression, specially because this visual behavior is already tested in the previous story.

```
export const RequiredWithSpecifiedText: =  () => {
}
...

RequiredWithSpecifiedText.parameters = {
    chromatic: {
        // We have screenshots of other stories that cover this case.
        disableSnapshot: true,
    },
};
```

packages/wonder-blocks-form/src/components/labeled-text-field.stories.js
```diff
@@ -121,6 +164,36 @@ export const Email: StoryComponentType = () => {
     );
 };
 
+export const EmailRequired: StoryComponentType = () => {
```
jandrade, line 167: Same comment related to chromatic applies with this story: `disableSnapshot: true`

jandrade (commented):
This looks super great! Nothing blocking, but just "unofficially" requesting some changes to remove some Chromatic snapshots before giving final approval.

This is so we get a new Chromatic build with the changes and we are able to approve the changes on Chromatic's side.

You can see an example of the current Chromatic build to get a better context: https://www.chromatic.com/pullrequest?appId=5e1bf4b385e3fb0020b7073c&number=1295&view=changes

Happy to chat more about the Chromatic part if needed.

packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
```diff
@@ -485,3 +486,138 @@ describe("LabeledTextField", () => {
         expect(textField).toHaveProp("autoComplete", autoComplete);
     });
 });
+
+describe("Required LabeledTextField", () => {
+    test("Has * when required prop is true", async () => {
+        // Arrange
```
somewhatabstract, line 492: NIT: The style we generally use is to place a newline after `// Arrange` if it is empty, to separate it from `// Act`. I find it tends to help with readability.

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -41,6 +41,27 @@ type Props = {|
...
+     * The error message that should render if this field is
+     * required and left blank. Note that this message will not be used
+     * if a `validate` prop is passed in.
+     *
+     * Implementation note: Added this field so that users can pass in
+     * translated strings.
+     *
+     * Example message: i18n._("A password is required to log in.")
+     */
+    requiredErrorMessage?: string,
```
somewhatabstract, line 63: Should the `required` prop be merged with this one since this error message is meaningless without the field being required?

Perhaps:

```
type Required = boolean | string;

type Props = {|
    ....
    required?: Required,
    ....
```

That way it can either be true/falsy or a string that is the error.

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -76,6 +76,17 @@ type Props = {|
...
+    /**
+     * The error message that should render if this field is
+     * required and left blank. Note that this message will not be used
+     * if a `validate` prop is passed in. Also note that error messages
+     * only redner for the LabeledTextField component, not for TextField.
+     *
+     * Implementation note: Added this field so that users can pass in
+     * translated strings.
+     */
+    requiredErrorMessage: string,
```
somewhatabstract, line 88: See my earlier comment regarding the `required` field and this one being combined.

somewhatabstract (changes requested):
I think that the `required` prop should be combined with the `requiredErrorMessage` prop for a cleaner API since `requiredErrorMessage` is meaningless when `required` is false. Otherwise, this looks OK to me.

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -42,25 +42,28 @@ type Props = {|
...
+     * This can be a boolean or a string.
      *
-     * Implementation note: Added this field so that users can pass in
-     * translated strings.
+     * String:
+     * Please pass in a translated string to use as the error message that will
+     * render if the user leaves this field blank. If this field is required,
+     * and a string is not passed in, a default untranslated string will render
+     * upon error.
+     * Note: The string will not be used if a `validate` prop is passed in.
```
somewhatabstract, line 55: Given this info, I wonder if we should consolidate our validation API here.

1. `required` is a type of validation. It's a shortcut for "not empt"
2. The error message is what to show if that `required` validation does not pass
3. `validate` is a custom validator

Rather than the `required` prop and `validate` prop, should we have a single `validation` prop that supports validation rules or something?

Not requesting a change right now, but I think it's worth considering.

Imagine a `validation` prop of type `Validation<T> | $ReadOnlyArray<Validation<T>>` where `Validation` is simply a `validate` callback that takes in a value and returns a `?string`. If the string is null/undefined/empty, it's considered valid. `T` is the type of the thing to be validated, so in this case it would be `string`.

We then have helpers:

1. `<T>runValidations(value: ?T, validations: Validation<T> | $ReadOnlyArray<Validation<T>>): ?string`
    This would run all the validations in order and return the first error string.
2. `Validate.Required<T>(translatedErrorMessage?: string)`
    This would be the `required: true` equivalent. You could just set `validation` prop to `Validate.Required<string>()` and that would return a proper `Validation<string>` function that asserted it wasn't empty, for example. Could also take an error string to return if it was missing, rather than the default.
3. `Validate.Custom<T>(validator: (value: T) => ?string): Validation<T>`
     To define a custom rule.
4. Could be other `Validate.` rules like `Validate.Min(minimum: number, translatedErrorMessage: string)` as we needed

I feel this is scalable and reusable and avoids us having these issues where "prop A is only valid if prop B is not set", etc.

Again, not requesting change here, just something for us to think about.

/cc @jandrade

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -72,20 +74,28 @@ type Props = {|
...
-     * Whether this component is required.
-     */
-    required?: boolean,
-
-    /**
-     * The error message that should render if this field is
-     * required and left blank. Note that this message will not be used
-     * if a `validate` prop is passed in. Also note that error messages
-     * only redner for the LabeledTextField component, not for TextField.
+     * Whether this field is required to to continue, or the error message to
```
somewhatabstract, line 77: See my earlier comment on how I think we can simplify this API in the long term.

somewhatabstract (approved):
LGTM 

Thanks for making the change to `required`. This has, of course, spurred more thoughts on this API. Nothing that need be done right now, but worth considering for some future rework, I think.

packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
```diff
@@ -485,3 +486,141 @@ describe("LabeledTextField", () => {
         expect(textField).toHaveProp("autoComplete", autoComplete);
     });
 });
+
+describe("Required LabeledTextField", () => {
+    test("Has * when `required` prop is true", () => {
```
somewhatabstract, line 491: SUPER NIT: We tend to prefer lower case in test names, I think, so `has`?

packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.js
```diff
@@ -485,3 +486,141 @@ describe("LabeledTextField", () => {
...
+                onChange={() => {}}
+                required={true}
+            />,
+        );
+
+        // Assert
+        expect(screen.getByText("*")).toBeInTheDocument();
+    });
+
+    test("Does not have * when `required` prop is false", () => {
```
somewhatabstract, line 508: SUPER NIT: We tend to prefer lower case in test names, I think, so `does`?

packages/wonder-blocks-form/src/components/labeled-text-field.js
nishasy: I'm wondering if we should just go back to what I was originally going to do, and just have `required` for the asterisk visual, and if someone want to validate the required field, they can do so using the `validate` prop. I was just trying to save work for people writing required fields in the future, but that's how it was being implemented up until now anyway.

Example: 
https://github.com/Khan/webapp/blob/master/services/static/javascript/auth/components/login-form.jsx#L269

jandrade: @somewhatabstract that's super interesting! Agree with you about proposing that change in the near future.

How do you see the `required` (aka add the `*` label) would fit in that model?

jandrade (approved):
LGTM! Thanks for making the changes proposed by Jeff and modifying the stories to prevent unnecessary chromatic snapshots 🥇 🚀

packages/wonder-blocks-form/src/components/labeled-text-field.js
somewhatabstract: Yes, good question. There are a couple of things I could imagine.

1. We just have a `markAsRequired` type prop that adds the indicator
2. We have a `decorator` prop that takes a component to render as the decorator, and we can pass in one that's to indicate required.

I think option 1 is probably the best.

PR #1296, opened 2022-01-14
[fei4225.1] Implement useGql hook

somewhatabstract:
## Summary:
This adds the `useGql` hook. It also refactors our API a little to cater to the ways in which GQL can be requested, deferring the main work of building the request URL and payload to the fetch function given to our `GqlRouter`.
We also add `KindError` from Wonder Stuff Core here to give better error information.

Issue: FEI-4225

## Test plan:
`yarn test`


packages/wonder-blocks-data/src/util/gql-error.js
```diff
@@ -0,0 +1,36 @@
+// @flow
```
jandrade, line 1: Neat!

packages/wonder-blocks-data/src/util/gql-types.js
```diff
@@ -14,43 +19,41 @@ export type GqlOperation<
...
-> = {|
+> = {
     type: TType,
     id: string,
-|};
+    // We allow other things here to be passed along to the fetch function.
+    // For example, we might want to pass the full query/mutation definition
+    // as a string here to allow that to be sent to an Apollo server that
+    // expects it.
+    ...
```
jandrade, line 29: What do you think about adding a test to cover this case?

packages/wonder-blocks-data/src/hooks/use-gql.js
```diff
@@ -0,0 +1,71 @@
...
+            ) => {
+                const {variables, context} = options;
+
+                // Invoke the fetch and extract the data.
+                return fetch(operation, variables, {
+                    ...defaultContext,
+                    ...context,
+                }).then(getGqlDataFromResponse, (error) => {
+                    // Return null if the request was aborted.
+                    if (error.name === "AbortError") {
```
jandrade, line 63: How do we set "AbortError" here. Is this using `AbortController`? If I understand correctly, `fetch` is injected using Context, but I'm not sure which `fetch` fn is being passed in WB.

jeresig (approved):

packages/wonder-blocks-data/src/hooks/use-gql.js
somewhatabstract: This bit of code is based entirely on how Apollo handles aborted errors. I would not expect us to be setting this - it would entirely be down to how the fetch works and is out of context of this library - we're just making sure we handle aborts. Most likely, any fetch will end up calling the standard `fetch` and we'd get abort errors as expected.

packages/wonder-blocks-data/src/util/gql-types.js
somewhatabstract: Well, we won't do anything with these values other than passing it all through so I don't know that it matters. There's not really anything to test.

jandrade: Fair enough :)

jandrade (approved):
Looks great! thanks for your answers.

PR #1297, opened 2022-01-15
[fei4244.1] Implemented mocking for gqlFetch stuff

somewhatabstract:
## Summary:

Issue: FEI-4244

## Test plan:
`yarn test`


packages/wonder-blocks-testing/src/gql/gql-request-matches-mock.js
```diff
@@ -0,0 +1,70 @@
+// @flow
+import type {GqlOperation, GqlContext} from "@khanacademy/wonder-blocks-data";
+import type {GqlMockOperation} from "./types.js";
+
+const safeHasOwnProperty = (obj: any, prop: string): boolean =>
+    // Flow really shouldn't be raising this error here.
+    // $FlowFixMe[method-unbinding]
+    Object.prototype.hasOwnProperty.call(obj, prop);
+
+const areObjectsEqual = (a: any, b: any): boolean => {
```
jeresig, line 10: Hmm - is this something we should be using from elsewhere? Seems like the sort of thing we don't want to be writing ourselves.

jeresig (approved):
Very nice, thank you!

packages/wonder-blocks-testing/src/gql/gql-request-matches-mock.js
somewhatabstract: Agreed. I intend to find an import for this, but I don't want to bloat the library more than necessary. That said, this is just the testing package so it's less of a concern. I didn't do that bit yet just out of interests of getting things "done" and usable. I'll raise a ticket for this bit to make sure it's not forgotten.

What I would really like is to leverage jest's partial matchers like `expect.objectContaining` as it would make this even nicer to work with but I haven't researched that bit yet.

jeresig: Ah yeah - the fact that this is "just" a part of testing makes this less critical - thanks for filing a follow-up!

somewhatabstract: OK, raised FEI-4268 and tagged code with comment.

PR #1310, opened 2022-01-21
Stop modals from closing when user releases click in the backdrop

nishasy:
## Summary:
Right now, if you click in the modal backround and release in the panel, the modal closes even though it should not.
Also in Chrome, if you click in a modal panel and release in the backdrop, the modal closes even though it should not.

- Instead of handling `click`, handle `mouseup` and `mouedown` and keep track of whether both happened in the backdrop.
  Only close if the user pressed AND released in the backdrop.

Issue: https://khanacademy.atlassian.net/browse/WB-602

Not sure how to show where my presses and releases are, but this video contains a variety of presses
and releases on Google Chrome.

https://user-images.githubusercontent.com/13231763/150456137-778f6880-bbb1-473a-9310-d34c8b475eac.mov

## Test plan:
`yarn test packages/wonder-blocks-modal/src/components/__tests__/modal-backdrop.test.js`

Manual testing - USE MULTIPLE BROWSERS.
- Open the Netlify wonder blocks app (localhost:6060)
- Click on a modal launcher (ex. the "Open flexible modal" button)
- Press inside the panel, drag out, and release in the backdrop. The modal should not close.
- Press outside the panel, drag out, and release in the panel.
- Try different combinations of pressing and releasing inside and outside the panel.
- Click in the panel. The modal should not close.
- Click in the backdrop. The modal should close.

### DIFF
--- a/packages/wonder-blocks-modal/src/components/__tests__/modal-backdrop.test.js
+++ b/packages/wonder-blocks-modal/src/components/__tests__/modal-backdrop.test.js
@@ -3,6 +3,7 @@ import * as React from "react";
 import {mount} from "enzyme";
 import "jest-enzyme";
 import {render, screen, fireEvent} from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
 import ModalBackdrop from "../modal-backdrop.js";
 import OnePaneDialog from "../one-pane-dialog.js";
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-modal/src/components/__tests__/modal-backdrop.test.js
```diff
@@ -47,19 +49,33 @@ describe("ModalBackdrop", () => {
...
-        wrapper.simulate("click");
+        // Simulate click
+        //
+        // The component looks for mouseup and mousedown instead of just click
+        // on the backdrop to make sure the user didn't drag into or out from
+        // the modal panel.
+        // eslint-disable-next-line testing-library/prefer-user-event
+        fireEvent.mouseDown(backdrop);
+        // eslint-disable-next-line testing-library/prefer-user-event
+        fireEvent.mouseUp(backdrop);
```
kevinbarabash, line 76: `userEvent.click()` should be simulating the mousedown and mouseup events as well.  Using `fireEvent` should only be necessary when the mousedown is on one element and the mouseup is on different one.

kevinbarabash (approved):
LGTM.  Thanks for using testing-library for the new tests.  I think we may also want to do the same thing for touch events as well.  That could be a separate PR though.  Talk it over with @jandrade and see what he thinks.

jandrade (commented):
Adding to what @kevinbarabash said, I think this issue might not apply to small devices (phones) because the Dialog takes the whole screen, but probably worth checking how this would work on a tablet.

You can take a look on this story that points to this PR: https://5e1bf4b385e3fb0020b7073c-ykjuhjpdky.chromatic.com/?path=/story/floating-modal-onepanedialog--with-opener

PR #1313, opened 2022-01-26
Update changeset for wonder-blocks-clickable

nishasy:
## Summary:
Update changeset for wonder-blocks-clickable so that buttons can have right behavior pick ca3c4bb6fde Migrate simple analytics functions

Issue: https://khanacademy.atlassian.net/browse/WB-393

## Test plan:
Verify that the version gets bumped in the release PR


jandrade (approved):
Thanks for fixing this.

NOTE: We need to find a way to let changesets know when a package needs to be bumped.

PR #1317, opened 2022-01-27
WB-1200.2: Migrate BirthdayPicker component

jandrade:
## Summary:

- Migrated `BirthdayPicker` from webapp.
- Refactored unit tests to avoid depending on `waitForAnimationFrame`.
- Added Storybook stories and documentation.
- Added `moment` and `jest-date-mock` to the root `package.json`.
- Fixed a couple of Jest warnings on the `wb-dropdown` package. (changed `backgroundColor` to `background`).

Issue: WB-1200

## Test plan:

```
yarn jest packages/wonder-blocks-birthday-picker/
```

Manual testing:

Check the changes in the Storybook instance associated to this PR, or test it locally:

```
yarn start:storybook
```

Navigate to http://localhost:6061/?path=/docs/birthdaypicker--birthday-picker-default

Verify that the stories are rendered correctly and the documentation is correct.
Screen Shot 2022-01-27 at 12 08 24 PM

https://user-images.githubusercontent.com/843075/151408556-280425c0-d7e2-49c9-a9ce-4fc81d4c362c.mov


packages/wonder-blocks-birthday-picker/src/components/birthday-picker.js
```diff
@@ -0,0 +1,236 @@
...
+    day: string | null,
+    year: string | null,
+    error: string | null,
+|};
+
+// Flow doesn't know about the getYear property on Date for some reason!
+// $FlowFixMe[prop-missing]
+const CUR_YEAR = new Date().getYear() + 1900;
+// TODO(WB-1200.2): Add labels prop to handle translations.
+const BIRTHDAY_ERROR = "Please select a valid birthdate.";
```
jandrade, line 38: NOTE: These hardcoded strings will be changed to use a `labels` prop in the next PR (same applies to month, day, year).

packages/wonder-blocks-dropdown/src/components/option-item.js
```diff
@@ -190,7 +190,7 @@ const {blue, white, offBlack, offBlack32} = Color;
 const styles = StyleSheet.create({
     itemContainer: {
         flexDirection: "row",
-        backgroundColor: white,
+        background: white,
```
jandrade, line 193: This is unrelated to these changes, but I changed this so Jest stops complaining about this:

```
 console.error
    Warning: Removing a style property during rerender (background) when a conflicting property is set (backgroundColor) can lead to styling bugs. To avoid this, don't mix shorthand and non-shorthand properties for the same value; instead, replace the shorthand with separate values.
        in button (created by StyleComponent)
        in StyleComponent (created by ClickableBehavior)
        in ClickableBehavior (created by Context.Consumer)
        in SelectOpener (created by SingleSelect)
        in div (created by StyleComponent)
        in StyleComponent (created by View)
        in View (created by DropdownCore)
        in DropdownCore (created by ActionSchedulerProvider)
        in ActionSchedulerProvider (created by ForwardRef)
        in ForwardRef (created by SingleSelect)
        in SingleSelect (created by BirthdayPicker)
        in div (created by StyleComponent)
        in StyleComponent (created by View)
        in View (created by BirthdayPicker)
        in BirthdayPicker
```

packages/wonder-blocks-birthday-picker/src/components/__tests__/birthday-picker.test.js
```diff
@@ -0,0 +1,343 @@
...
+            // Assert
+            expect(
+                screen.getByTestId("birthday-picker-month"),
+            ).toHaveTextContent("Month");
+            expect(screen.getByTestId("birthday-picker-day")).toHaveTextContent(
+                "Day",
+            );
+            expect(
+                screen.getByTestId("birthday-picker-year"),
+            ).toHaveTextContent("Year");
```
somewhatabstract, line 128: Can you break the `screen.getByTestId` calls out into their own thing, possibly in the Act section so that the expectations are easier to parse?

Aside: This seems like a good example of where multiple expectations kinda make sense. However, I wonder if instead of getting each part of the picker, we took a holistic view and did `toHaveTextContent("MonthDayYear")` on the picker itself - would that work even?

Hmm, I dunno. I'm just riffing. No change requested.

packages/wonder-blocks-birthday-picker/src/components/__tests__/birthday-picker.test.js
```diff
@@ -0,0 +1,343 @@
...
+                    ref={(node) => (maybeInstance = node)}
+                />,
+            );
+
+            if (!maybeInstance) {
+                throw new Error("BirthdayPicker instance is undefined");
+            }
+            const instance = maybeInstance;
+
+            // This test was written by calling methods on the instance because
```
somewhatabstract, line 308: Useful comment, thanks!

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.js
```diff
@@ -0,0 +1,236 @@
...
+     * format or `null`.
+     */
+    onChange: (date: ?string) => mixed,
+|};
+
+type State = {|
+    month: string | null,
+    day: string | null,
+    year: string | null,
+    error: string | null,
```
somewhatabstract, line 31: What do you think about commenting on these like we do props? I don't feel strongly about it, but perhaps it's useful to understand how the state changes reflect in the component behavior?

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.js
```diff
@@ -0,0 +1,236 @@
...
+ * ```
+ */
+export default class BirthdayPicker extends React.Component<Props, State> {
+    constructor(props: Props) {
+        super(props);
+
+        let month = null;
+        let day = null;
+        let year = null;
+        let error = null;
```
somewhatabstract, line 73: I find `let` to be a smell that we need a separate function. What do you think about a `getStateFromDefault` call that returned a state object?

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.js
```diff
@@ -0,0 +1,236 @@
...
+     * valid, and then reporting the result back to the calling component via
+     * reportChange.
+     */
+    handleChange: () => void = (): void => {
+        const {month, day, year} = this.state;
+
+        // If any of the values haven't been set then our overall value is
+        // equal to null
+        if (month === null || day === null || year === null) {
+            this.reportChange(null);
```
somewhatabstract, line 124: Add early return and drop the else to reduce indent and improve readability?

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.js
```diff
@@ -0,0 +1,236 @@
...
+                                icon={icons.info}
+                                color={Color.red}
+                                style={{marginTop: 3}}
+                                aria-hidden="true"
+                            />
+                            <Strut size={Spacing.xxxSmall_4} />
+                            <Body style={{color: Color.red}}>{error}</Body>
+                        </View>
+                    </>
+                )}
```
somewhatabstract, line 232: I prefer a `maybeRenderXxx` type function or even a new component to inline things like this, I think it makes things more readable.

somewhatabstract (approved):
LGTM - one comment in Chromatic for your attention.

I have some suggestions inline but I don't think any of them are blocking.

packages/wonder-blocks-birthday-picker/src/components/__tests__/birthday-picker.test.js
jandrade: oh oops, that's was John's comment, but yeah, it's supper helpful :)

I forgot to mention the minor detail that I moved these tests from webapp to wb.

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.js
jandrade: Good point! I'll do this in the next PR to avoid some conflicts.

packages/wonder-blocks-birthday-picker/src/components/__tests__/birthday-picker.test.js
```diff
@@ -0,0 +1,335 @@
...
+            render(<BirthdayPicker onChange={() => {}} />);
+
+            const monthPicker = screen.getByTestId("birthday-picker-month");
+            const dayPicker = screen.getByTestId("birthday-picker-day");
+            const yearPicker = screen.getByTestId("birthday-picker-year");
+
+            // Assert
+            expect(monthPicker).toHaveTextContent("Month");
+            expect(dayPicker).toHaveTextContent("Day");
+            expect(yearPicker).toHaveTextContent("Year");
```
somewhatabstract, line 32: Thanks! :)

packages/wonder-blocks-birthday-picker/src/components/__tests__/birthday-picker.test.js
```diff
@@ -0,0 +1,335 @@
...
+            // Pick Another Date
+            userEvent.click(screen.getByTestId("birthday-picker-month"));
+            const monthOptionNew = await screen.findByRole("option", {
+                name: "Aug",
+            });
+            userEvent.click(monthOptionNew, undefined, {
+                skipPointerEventsCheck: true,
+            });
+
+            userEvent.click(screen.getByTestId("birthday-picker-day"));
```
somewhatabstract, line 224: Oh, rather than get this thing multiple times, I think you can just get it once and use a variable. It may make these more complex tests easier to read.

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.js
jandrade: Done, but GH didn't notice the change here (see next line).

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.js
```diff
@@ -0,0 +1,259 @@
...
+     * reportChange.
+     */
+    handleChange: () => void = (): void => {
+        const {month, day, year} = this.state;
+
+        // If any of the values haven't been set then our overall value is
+        // equal to null
+        if (month === null || day === null || year === null) {
+            this.reportChange(null);
+            return;
```
somewhatabstract, line 147: I see the change here. I think GH pinned the comment to the commit somehow, or something.

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.js
```diff
@@ -0,0 +1,259 @@
...
+
+            // If the date is in the future or is invalid then we want to show
+            // an error to the user.
+            if (date.isAfter() || !date.isValid()) {
+                initialState.error = BIRTHDAY_ERROR;
+            }
+        }
+
+        return initialState;
+    }
```
somewhatabstract, line 117: Nice. Could completely decouple this from the component by taking the default in as an arg - no change requested.

jandrade:
The parent pull-request ([#1316](https://github.com/Khan/wonder-blocks/pull/1316)) has been merged into `feature/birthday-picker`, but this branch (`bdp-2`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #1319, opened 2022-01-27
[onlybuildactualpackages] Only build actual packages

somewhatabstract:
## Summary:
When things are modified between checkouts, package folders aren't deleted, so we can't use them as an indicator of a package to be built. Instead, check that there is a package.json.

This should fix [things like this](https://github.com/Khan/wonder-blocks/runs/4969439614?check_suite_focus=true) that occur when adding a new package.

The linked failure happens because the build first builds the current code with the new package, then it switches to the base branch and builds that, but the switch doesn't delete the new package folder, and so it tries to build that and fails.

With my change, it should ignore that new package folder in the second case because it will no longer have a package.json file.

Issue: XXX-XXXX

## Test plan:
`yarn build:all`

### DIFF
--- a/.changeset/spotty-birds-collect.md
+++ b/.changeset/spotty-birds-collect.md
@@ -1,0 +1,5 @@
+---
+"wb-dev-build-settings": minor
+---
+
+Builds only include actual packages, not remnant directories from a branch switch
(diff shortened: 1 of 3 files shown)


kevinbarabash (approved):
LGTM.  Thanks for fixing this. ❤️

jandrade (approved):
Awesome!! thanks for fixing this, super helpful 🥇

PR #1324, opened 2022-02-01
[🔥AUDIT🔥] Disable Tooltip and Popover snapshots

jandrade:
🖍 _This is an audit!_ 🖍

## Summary:

Temporary fix that disables these snapshots so Chromatic stops reporting flaky
visual regressions.

Issue: WB-1170

## Test plan:

Verify in Chromatic that these stories are no longer reported.


jeresig (approved):
Too bad! :(

PR #1325, opened 2022-02-01
WB-1226: [dropdown] Update disabled state to match design specs

jandrade:
## Summary:

Updates the `disabled` styles on the `SelectOpener` component (used by `SingleSelect` and `MultiSelect`) to match the design specs.

NOTE: It also uses the same color used in `TextField` to be consistent across form fields.

Issue: WB-1226

## Test plan:

```
yarn start:storybook
```

Navigate to the following pages:
http://localhost:6061/?path=/story/dropdown-singleselect--disabled-single-select
Screen Shot 2022-02-01 at 11 34 13 AM

http://localhost:6061/?path=/story/dropdown-multiselect--disabled-multi-select
Screen Shot 2022-02-01 at 11 34 19 AM

Verify that the dropdown opener is using `offWhite` as the bg color.


kevinbarabash (approved):
LGTM.  Thanks for fixing that color. :shipit:

PR #1329, opened 2022-02-01
WB-1227: Storybook - Fix WB aliases on the webpack config file

jandrade:
## Summary:

Overriding the Storybook's webpack config so we can work with custom WB aliases
to be able to run Storybook inside our monorepo.

Issue: https://khanacademy.atlassian.net/browse/WB-1227

## Test plan:

```
yarn start:storybook
```

1. Verify that storybook starts locally.
2. Verify that the Storybook chromatic build works fine.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,3 +1,6 @@
+const path = require("path");
+const webpack = require("webpack");
+
 module.exports = {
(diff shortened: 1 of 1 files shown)


jeresig (approved):
Ohh, very tricky!

PR #1330, opened 2022-02-01
WB-1181: Allow birthday component to be disabled

jandrade:
## Summary:

- Added the `disabled` prop.
- Added docs and tests covering this case.
- Added documentation to the existing stories.
- Removed some Storybook `argTypes` not needed anymore. SB is taking correctly the info from the `Props` type.

Issue: WB-1181

## Test plan:

```
yarn jest packages/wonder-blocks-birthday-picker/
yarn start:storybook
```

Navigate to http://localhost:6061/?path=/story/birthdaypicker--disabled-birthday-picker
Screen Shot 2022-02-01 at 5 23 56 PM

Verify that the birthday picker is disabled and cannot be interacted with.

https://user-images.githubusercontent.com/843075/152061785-6f30f361-1b06-40dd-b9f4-87af39d27fe0.mov

### DIFF
--- a/packages/wonder-blocks-birthday-picker/src/components/__docs__/birthday-picker.argtypes.js
+++ b/packages/wonder-blocks-birthday-picker/src/components/__docs__/birthday-picker.argtypes.js
@@ -2,37 +2,13 @@
 import {defaultLabels} from "../birthday-picker.js";
 
 export default {
-    defaultValue: {
-        control: {type: "text"},
-        description:
-            "The default value to populate the birthdate with. Should be in the format: `YYYY-MM-DD` (e.g. 2021-05-26). It's only used to populate the initial value as this is an uncontrolled component.",
-        table: {
-            type: {
-                summary: "string",
-            },
-        },
-    },
     labels: {
-        control: {type: "object"},
-        description:
-            "The object containing the custom labels used inside this component.",
         defaultValue: defaultLabels,
-        table: {
-            type: {
-                summary: "Labels",
-                detail: "{month?: string, year?: string, day?: string, errorMessage?: string}",
-            },
-        },
     },
     onChange: {
         action: "onChanged",
-        description:
-            "Listen for changes to the birthdate. Could be a string in the `YYYY-MM-DD` format or `null`.",
         table: {
             category: "Events",
-            type: {
-                summary: "(date: ?string) => mixed",
-            },
         },
     },
 };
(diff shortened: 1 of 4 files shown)


packages/wonder-blocks-birthday-picker/src/components/__docs__/birthday-picker.argtypes.js
```diff
@@ -2,37 +2,13 @@
 import {defaultLabels} from "../birthday-picker.js";
 
 export default {
-    defaultValue: {
```
jandrade, line 5: I noticed that we don't need this anymore, at least for this component.

packages/wonder-blocks-birthday-picker/src/components/__docs__/birthday-picker.stories.js
```diff
@@ -64,6 +64,10 @@ BirthdayPickerWithDefaultValue.parameters = {
     chromatic: {
         disableSnapshot: true,
     },
+    docs: {
```
jandrade, line 67: Also added some missing documentation from the previous stories.

packages/wonder-blocks-birthday-picker/src/components/__docs__/birthday-picker.stories.js
```diff
@@ -87,3 +98,24 @@ BirthdayPickerWithCustomLabels.args = {
...
+    onChange: () => {},
+    disabled: true,
+};
+
+DisabledBirthdayPicker.parameters = {
+    docs: {
+        storyDescription:
+            "A BirthdayPicker can be disabled by passing the `disabled` prop. This will disable all the dropdown controls and prevent them from being interacted with.",
+    },
+};
```
jandrade, line 121: This is the new story relevant to the `disabled` change.

packages/wonder-blocks-birthday-picker/src/components/__docs__/birthday-picker.argtypes.js
jeresig: Huh - why is that? Are the descriptions still being generated?

jeresig (approved):

packages/wonder-blocks-birthday-picker/src/components/__docs__/birthday-picker.argtypes.js
jandrade: Yes :) I'm not sure what happened, but I tried that the first time and didn't work. maybe it's working now that I've fixed the Webpack config (just a random guessing)? I'll check with the Cell docs to see what happens.

https://5e1bf4b385e3fb0020b7073c-bmwroxzzmz.chromatic.com/?path=/docs/birthdaypicker--birthday-picker-default

PR #1335, opened 2022-02-04
[fei4261.contextmerge] Only merge properties with non-undefined values

somewhatabstract:
## Summary:
I was using the API and found it frustrating that `undefined` values explicitly included would override the defaults. I don't think this is what folks would expect and it certainly makes things harder in our case since we have to do some wrappers around the Wonder Blocks API in webapp.

So, this merges with an algorithm that ignores values explicitly set to undefined.

Also tidied up some typing.

Issue: FEI-4261

## Test plan:
`yarn test`
`yarn flow`

### DIFF
--- a/.changeset/thick-cobras-deny.md
+++ b/.changeset/thick-cobras-deny.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-data": minor
+---
+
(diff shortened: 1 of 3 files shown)


jaredly (approved):
makes sense to me!

kevinbarabash (approved):
LGTM :shipit:

PR #1337, opened 2022-02-08
Clickable: Allow disabled state to be focusable

jandrade:
## Summary:

- Clickable: Replaced `disabled` attribute with `aria-disabled` to make Clickable
components keyboard focusable.  
- Cell: Changed styles to include the `not-allowed` cursor when the user hovers
on a disabled `cell`.  
- Fixed unit tests to account for this change.

Issue: WB-1213

## Test plan:

Navigate to http://localhost:6061/?path=/story/cell-compactcell--compact-cell-disabled

Verify that the cell is disabled and it can be navigated using the keyboard.

Also verify that the `not-allowed` cursor is displayed when the user hovers on the cell.

Screenshot:
Screen Shot 2022-02-08 at 11 09 29 AM

Video
https://user-images.githubusercontent.com/843075/153028015-fe0eb112-83cb-4a7c-9d97-04004174dca7.mov

### DIFF
--- a/packages/wonder-blocks-clickable/src/components/__tests__/clickable.test.js
+++ b/packages/wonder-blocks-clickable/src/components/__tests__/clickable.test.js
@@ -3,6 +3,8 @@ import * as React from "react";
 import {MemoryRouter, Route, Switch} from "react-router-dom";
 import {mount} from "enzyme";
 import "jest-enzyme";
+import {render, screen} from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
 import {View} from "@khanacademy/wonder-blocks-core";
 import Clickable from "../clickable.js";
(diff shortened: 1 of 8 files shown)


packages/wonder-blocks-clickable/src/components/__tests__/clickable.test.js
```diff
@@ -452,6 +454,56 @@ describe("Clickable", () => {
...
+    test("allow keyboard navigation when disabled is set", () => {
+        // Arrange
+        render(
+            <Clickable testId="button" disabled={true}>
+                {(eventState) => <h1>Click Me!</h1>}
+            </Clickable>,
+        );
+
+        // Act
+        userEvent.tab();
```
nishasy, line 482: Is the button receiving focus because of this `userEvent.tab()`? It is possible that the button could be receiving focus because it's the only thing being rendered here? Sorry, I'm not entirely sure how this works but I'm just wondering because I struggled to test with `tab()` in the past.

packages/wonder-blocks-clickable/src/components/__tests__/clickable.test.js
```diff
@@ -452,6 +454,56 @@ describe("Clickable", () => {
...
+
+        const button = screen.getByTestId("button");
+
+        // Assert
+        expect(button).toHaveFocus();
+    });
+
+    test("allow keyboard navigation when disabled is set", () => {
```
nishasy, line 490: Is this test a duplicate of the one before it?

packages/wonder-blocks-clickable/src/components/__tests__/clickable.test.js
jandrade: That's an excellent question 🤔. By default, RTL gives focus to the `body` element, so this test is kinda tricky. I'll change it to be more explicit about this, and will probably add a couple of extra focusable elements (other buttons that are not `Clickable` elements) to make the test more robust.

packages/wonder-blocks-clickable/src/components/__tests__/clickable.test.js
jandrade: ouch good finding, I think that for some reason these tests got duplicated. I'll remove the duplicated ones.

packages/wonder-blocks-clickable/src/components/__tests__/clickable.test.js
jandrade: I updated the test to make it clearer. Let me know what you think.

megaroeny (commented):
Thanks for adding the not-allowed cursor!

nishasy (approved):
Thanks for updating the tests! LGTM

PR #1338, opened 2022-02-08
[fei4269.1.useServerEffect] Architecture tidy-up

somewhatabstract:
Summary:
After finally implementing the `useFastlyQuery` hook in webapp, it was clear that things would be a lot simpler with some of the changes I had been plannig. So I did them.

1. `useData` hook is gone. It was doing too much. The `Data` component does everything it was doing.
2. New `useServerEffect` hook that just handles invoking an async operation on the server and caching its result for hydration or only SSR. On the client, it only handles giving back whatever is in the hydration cache and clearing that value.
3. No more `IRequestHandler` - which should simplify general usage significantly. All the caching has been updated to be much simpler with this change.
4. Requests are just identified by a single identifier. The associated handler would be shared by things that use the same ID - this is OK as usually, that will be encapsulated in a component or hook, rather than just direct duplicate usage, anyway.

This also updates the gen-snapshot-tests tool to cope with exporting a variable instead of directly exporting a function or class.

Comprehensive docs will be created for the FEI-4191 task (hopefully we can replace the styleguidist docs with stories at that point too).

Issue: FEI-4269

## Test plan:
`yarn test`
`yarn flow`


packages/wonder-blocks-data/src/util/response-cache.js
```diff
@@ -29,32 +28,35 @@ export class ResponseCache {
...
-        handler: IRequestHandler<TOptions, TData>,
-        options: TOptions,
+    _setCacheEntry<TData: ValidData>(
+        id: string,
         entry: CacheEntry<TData>,
+        hydrate: boolean,
     ): CacheEntry<TData> {
         const frozenEntry = Object.freeze(entry);
-        if (this._ssrOnlyCache != null) {
+        if (Server.isServerSide()) {
```
somewhatabstract, line 50: I did this just so that the TrackData docs example would still work. I didn't like it, but it seemed OK for now. better docs will be coming that make it less ...flaky (the docs are flaky, not the code)?

packages/wonder-blocks-data/src/util/result-from-cache-entry.js
```diff
@@ -15,24 +15,21 @@ export const resultFromCacheEntry = <TData: ValidData>(
...
+            status: "success",
+            data,
         };
     }
 
     return {
-        status: "error",
-        error,
+        status: "aborted",
     };
```
somewhatabstract, line 34: To support the GQL framework, we need to allow `null` data, and so we have the `aborted` status as a first-class citizen. Hopefully I can add fetch aborting to WB Data in a future update to make full use of this.

packages/wonder-blocks-data/src/hooks/use-server-effect.js
```diff
@@ -0,0 +1,30 @@
+// @flow
+import {Server} from "@khanacademy/wonder-blocks-core";
+import {useContext} from "react";
+import {TrackerContext} from "../util/request-tracking.js";
+import {ResponseCache} from "../util/response-cache.js";
+
+import type {CacheEntry, ValidData} from "../util/types.js";
+
+export const useServerEffect = <TData: ValidData>(
```
jeresig, line 9: Comment here explaining usage?

packages/wonder-blocks-data/src/util/result-from-cache-entry.js
jeresig: cc: @kevinbarabash so that you're aware!

jeresig (approved):
Seems like some substantial improvements, thank you!

PR #1339, opened 2022-02-09
WB-1182: Remove `day` field from the `BirthdayPicker`

jandrade:
## Summary:

- Added `monthYearOnly` prop to hide the day field from the UI.
- Modified the logic to set the FIRST day of the SELECTED month when setting this new prop.
- Modified the `minWidth` of the components to account for this scenario.
- Added unit tests to cover this prop.

Issue: WB-1182

## Test plan:

Navigate to: http://localhost:6061/?path=/story/birthdaypicker--birthday-picker-with-year-and-month-only

Verify that the UI hides the day field.

Screen Shot 2022-02-09 at 12 33 03 PM

Chromatic: https://5e1bf4b385e3fb0020b7073c-yrgchxqdjx.chromatic.com/?path=/story/birthdaypicker--birthday-picker-with-year-and-month-only

Extra: Inspect the component using React DevTools and verify that the state is setting `day=1`.

### DIFF
--- a/packages/wonder-blocks-birthday-picker/src/components/__tests__/birthday-picker.test.js
+++ b/packages/wonder-blocks-birthday-picker/src/components/__tests__/birthday-picker.test.js
@@ -387,7 +387,7 @@ describe("BirthdayPicker", () => {
             expect(onChange).toHaveBeenCalledWith("2018-08-01");
         });
 
-        it("onChange triggers the first day of the month when defaultValue and monthYearOnly are set", async () => {
+        it("onChange triggers the passed-in day intact when defaultValue and monthYearOnly are set", async () => {
             // Arrange
             const onChange = jest.fn();
 
@@ -417,8 +417,8 @@ describe("BirthdayPicker", () => {
             });
 
             // Assert
-            // Verify that we passed the first day of the month
-            expect(onChange).toHaveBeenCalledWith("2018-08-01");
+            // Verify that we passed the same day originally passed in.
+            expect(onChange).toHaveBeenCalledWith("2018-08-17");
         });
     });
 
--- a/packages/wonder-blocks-birthday-picker/src/components/__tests__/birthday-picker.test.js
+++ b/packages/wonder-blocks-birthday-picker/src/components/__tests__/birthday-picker.test.js
@@ -181,9 +181,9 @@ describe("BirthdayPicker", () => {
             // Arrange
             const onChange = jest.fn();
 
-            // Act
             render(<BirthdayPicker onChange={onChange} />);
 
+            // Act
             userEvent.click(screen.getByTestId("birthday-picker-month"));
             const monthOption = await screen.findByRole("option", {
                 name: "Jul",
@@ -270,7 +270,6 @@ describe("BirthdayPicker", () => {
             // Arrange
             const onChange = jest.fn();
 
-            // Act
             render(
                 <BirthdayPicker
                     defaultValue="2017-07-17"
@@ -278,6 +277,7 @@ describe("BirthdayPicker", () => {
                 />,
             );
 
+            // Act
             userEvent.click(screen.getByTestId("birthday-picker-month"));
             const monthOption = await screen.findByRole("option", {
                 name: "Aug",
@@ -339,7 +339,6 @@ describe("BirthdayPicker", () => {
             // Arrange
             const onChange = jest.fn();
 
-            // Act
             render(
                 <BirthdayPicker
                     defaultValue="2021-02-31"
@@ -347,6 +346,7 @@ describe("BirthdayPicker", () => {
                 />,
             );
 
+            // Act
             userEvent.click(screen.getByTestId("birthday-picker-year"));
             const yearOption = await screen.findByRole("option", {
                 name: "2020",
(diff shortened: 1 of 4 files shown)


jeresig (approved):
Great! A question about how we should be treating the day when there is a defaultValue, but this otherwise looks good to me!

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.js
```diff
@@ -148,7 +158,12 @@ export default class BirthdayPicker extends React.Component<Props, State> {
         // If a default value was provided then we use moment to convert it
         // into a date that we can use to populate the
         if (defaultValue) {
-            const date = moment(defaultValue);
+            let date = moment(defaultValue);
+
+            // Override the default day if monthYearOnly mode is set.
+            if (monthYearOnly) {
+                date = date.startOf("month");
+            }
```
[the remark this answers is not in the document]
jandrade: Yeah that's a super valid point. I think we can keep the existing day as far as the callsite(s) replace that day with the logic that will be proposed for the Account Records project. I'm going to remove this part and update the unit test related to it.

packages/wonder-blocks-birthday-picker/src/components/__tests__/birthday-picker.test.js
```diff
@@ -346,6 +358,68 @@ describe("BirthdayPicker", () => {
             // Assert
             expect(onChange).toHaveBeenCalledTimes(1);
         });
+
+        it("onChange triggers the first day of the month when monthYearOnly is set", async () => {
+            // Arrange
+            const onChange = jest.fn();
+
+            // Act
+            render(<BirthdayPicker monthYearOnly={true} onChange={onChange} />);
```
somewhatabstract, line 367: This seems like part of the `Arrange` and the stuff after it is the actual action being tested. What do you think?

packages/wonder-blocks-birthday-picker/src/components/__tests__/birthday-picker.test.js
```diff
@@ -346,6 +358,68 @@ describe("BirthdayPicker", () => {
...
+            const onChange = jest.fn();
+
+            // Act
+            render(
+                <BirthdayPicker
+                    defaultValue="2017-07-17"
+                    monthYearOnly={true}
+                    onChange={onChange}
+                />,
+            );
```
somewhatabstract, line 401: See earlier comment

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.js
```diff
@@ -49,6 +49,16 @@ type Props = {|
...
     labels?: Labels,
 
+    /**
+     * Whether we want to hide the day field.
+     *
+     * **NOTE:** We will set the day to the _FIRST_ day of the _SELECTED_ month
+     * if the day field is hidden. Please make sure to modify the passed date
+     * value to fit different needs (e.g. if you want to set the _FIRST_ day of
+     * the _FOLLOWING_ month instead).
+     */
```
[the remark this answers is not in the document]
somewhatabstract: I think you can drop the caps and keep the emphasis.

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.js
```diff
@@ -87,6 +97,11 @@ export const defaultLabels: Labels = Object.freeze({
     day: "Day",
 });
 
+// Default minWidth value when we include the full DOB.
+const FIELD_MIN_WIDTH_FULL = 110;
+// Alternative minWidth value when we hide the day field.
+const FIELD_MIN_WIDTH_MONTH_YEAR = 167;
```
somewhatabstract, line 103: How are these values determined? Should we document a design reference? They seem arbitrary when all our other sizes are usually multiples of our spacing constants.

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.js
```diff
@@ -252,63 +267,99 @@ export default class BirthdayPicker extends React.Component<Props, State> {
...
+            ? FIELD_MIN_WIDTH_MONTH_YEAR
+            : FIELD_MIN_WIDTH_FULL;
+
+        return (
+            <SingleSelect
+                disabled={disabled}
+                placeholder={this.labels.month}
+                onChange={this.handleMonthChange}
+                selectedValue={month}
+                style={{minWidth}}
```
somewhatabstract, line 283: Thanks for moving this to a constant.

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.js
```diff
@@ -252,63 +267,99 @@ export default class BirthdayPicker extends React.Component<Props, State> {
...
-                            <OptionItem
-                                key={String(CUR_YEAR - yearOffset)}
-                                label={String(CUR_YEAR - yearOffset)}
-                                value={String(CUR_YEAR - yearOffset)}
-                            />
-                        ))}
-                    </SingleSelect>
+
+                    {this.renderYear()}
                 </View>
```
somewhatabstract, line 363: This is not i18n friendly. I think long term we need to support date formats. In most countries outside the US, the day comes before the month.

somewhatabstract (approved):
Some inline NITS but nothing blocking.

:ship:

PR #1340, opened 2022-02-09
[fei4261.1.useInMemoryCache] Add useSharedCache hook

somewhatabstract:
## Summary:
This adds a `useSharedCache` hook that uses a shared instance of in-memory cache (originally https://github.com/Khan/webapp/pull/3522?w=1).

It also reworks the in-memory cache implementation so that we only have one implementation that is shared across our different use cases (that doesn't mean instances are shared across all use-cases - that would not make sense).

Issue: FEI-4261

## Test plan:
`yarn test`
`yarn flow`


packages/wonder-blocks-data/src/hooks/use-shared-cache.js
```diff
@@ -0,0 +1,106 @@
...
+ * be toggled to cause a rerender whenever the referenced cached data changes
+ * so that callers can re-render on cache changes. However, we should make
+ * sure this toggling is optional - or we could use a callback argument, to
+ * achieve this on an as-needed basis.
+ */
+export const useSharedCache = <TValue: ValidData>(
+    id: string,
+    scope: string,
+    initialValue?: ?TValue | (() => ?TValue),
+): [?TValue, CacheValueFn<TValue>] => {
```
somewhatabstract, line 57: Signature matches the return of `useState`, with a result and a setting function. I like this pattern since it allows folks to easily name the function and value something that makes sense for their context. Returning an object allows for that, but it's more cumbersome.

packages/wonder-blocks-data/src/util/memory-cache.js
```diff
@@ -1,134 +0,0 @@
-// @flow
-import type {ValidData, CacheEntry, Cache} from "./types.js";
```
somewhatabstract, line 2: This is replaced by our new `in-memory-cache.js` implementation.

packages/wonder-blocks-data/src/util/memory-cache.js
```diff
@@ -1,134 +0,0 @@
...
-     * We want to deep clone the source cache to dodge mutations by external
-     * references. So we serialize the source cache to JSON and parse it
-     * back into a new object.
-     *
-     * NOTE: This doesn't work for get/set property accessors.
-     */
-    const serializedInitCache = JSON.stringify(source);
-    const cloneInitCache = JSON.parse(serializedInitCache);
-    return Object.freeze(cloneInitCache);
-}
```
somewhatabstract, line 15: This is replaced by wonder-stuff-core's `clone`. This is a change in the cloning since it does far more than just stringify/parse.

packages/wonder-blocks-data/src/util/response-cache.js
```diff
@@ -9,6 +9,8 @@ import type {
     ResponseCache as ResCache,
 } from "./types.js";
 
+const DefaultScope = "default";
```
somewhatabstract, line 12: Since the new in-memory cache is scoped and we don't need that, we just use a default value for scope.

packages/wonder-blocks-data/src/util/response-cache.js
```diff
@@ -28,17 +30,17 @@ export class ResponseCache {
         return _default;
     }
 
-    _hydrationCache: MemoryCache;
-    _ssrOnlyCache: ?MemoryCache;
+    _hydrationCache: InMemoryCache;
+    _ssrOnlyCache: ?InMemoryCache;
```
somewhatabstract, line 34: Now that we have a scoped in-memory cache, we could feasibly use one instance and two different scopes, but I think keeping them separate makes things a little easier to understand when each cache comes into play. It also makes it less likely that we introduce a bug like returning the ssr-only cache when we're asked to clone data for the hydration cache.

packages/wonder-blocks-data/src/util/types.js
```diff
@@ -1,8 +1,15 @@
 // @flow
-export type ValidData = string | boolean | number | {...} | Array<ValidData>;
-
-export type Status = "loading" | "success" | "error";
```
somewhatabstract, line 4: Nothing was using this.

packages/wonder-blocks-data/src/util/types.js
```diff
@@ -1,8 +1,15 @@
...
+/**
+ * Define what can be cached.
+ *
+ * We disallow functions and undefined as undefined represents a cache miss
+ * and functions are not allowed.
+ */
```
somewhatabstract, line 7: Wanted to add some commentary. In hindsight, I'd rename a bunch of these types but I didn't want to introduce more breaking changes.

packages/wonder-blocks-data/src/util/types.js
```diff
@@ -34,9 +44,29 @@ export type InterceptContextData = {
     ...
 };
 
+/**
+ * A cache of fulfilled request responses.
+ */
 export type Cache = {
     [key: string]: CacheEntry<any>,
     ...
 };
```
somewhatabstract, line 53: Ideally, this would get renamed to ResponseCache or something, but then we'd have to rename other things that are already called that. Naming things is hard. For now, I decided to leave the existing types as they are for this change.

packages/wonder-blocks-data/src/util/types.js
```diff
@@ -19,6 +26,9 @@ export type Result<TData: ValidData> =
           status: "aborted",
       |};
 
+/**
+ * A cache entry for a fulfilled request response.
+ */
 export type CacheEntry<TData: ValidData> =
```
somewhatabstract, line 32: Ideally, `ResponseCacheEntry` or something, but naming things is hard and I decided to leave existing types as they are for now.

packages/wonder-blocks-data/src/util/types.js
somewhatabstract: See https://github.com/Khan/wonder-blocks/pull/1341?w=1 where I fix up the naming. :)

packages/wonder-blocks-data/src/util/types.js
somewhatabstract: Naming is fixed up in https://github.com/Khan/wonder-blocks/pull/1341?w=1

packages/wonder-blocks-data/src/hooks/use-shared-cache.js
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Hook to retrieve data from and store data in an in-memory cache.
+ *
+ * @returns {[?ReadOnlyCacheValue, CacheValueFn]}
+ * Returns an array containing the current cache entry (or undefined), a
+ * function to set the cache entry (passing null or undefined to this function
+ * will delete the entry).
+ *
+ * To clear a single scope within the cache or the entire cache,
+ * the `clearScopedCache` export is available.
```
kevinbarabash, line 41: Nice.  It seems like there aren't many cases where you'd want to do that so not including it in the hook's API make sense.

packages/wonder-blocks-data/src/hooks/use-shared-cache.js
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * NOTE: Unlike useState or useReducer, we don't automatically update folks
+ * if the value they reference changes. We might add it later (if we need to),
+ * but the likelihood here is that things won't be changing in this cache in a
+ * way where we would need that. If we do (and likely only in specific
+ * circumstances), we should consider adding a simple boolean useState that can
+ * be toggled to cause a rerender whenever the referenced cached data changes
+ * so that callers can re-render on cache changes. However, we should make
+ * sure this toggling is optional - or we could use a callback argument, to
+ * achieve this on an as-needed basis.
```
kevinbarabash, line 51: Thank you for including this comment.

packages/wonder-blocks-data/src/hooks/use-shared-cache.js
kevinbarabash: Love it! 🙂

packages/wonder-blocks-data/src/hooks/use-shared-cache.js
```diff
@@ -0,0 +1,106 @@
...
+ */
+export const useSharedCache = <TValue: ValidData>(
+    id: string,
+    scope: string,
+    initialValue?: ?TValue | (() => ?TValue),
```
kevinbarabash, line 56: 🤘

packages/wonder-blocks-data/src/hooks/__tests__/use-shared-cache.test.js
```diff
@@ -0,0 +1,307 @@
...
+            // Rerender so the hook retrieves this new value.
+            wrapper.rerender();
+            const result = wrapper.result.current[0];
+
+            // Assert
+            expect(result).toBeNull();
+        });
+    });
+
+    it("should share cache across all uses", () => {
```
kevinbarabash, line 210: ❤️

packages/wonder-blocks-data/src/hooks/use-shared-cache.js
```diff
@@ -0,0 +1,106 @@
...
+ * A function for inserting a value into the cache or clearing it.
+ */
+type CacheValueFn<TValue: ValidData> = (value: ?TValue) => void;
+
+/**
+ * This is the cache.
+ * It's incredibly complex.
+ * Very in-memory. So cache. Such complex. Wow.
+ */
+const cache = new InMemoryCache();
```
kevinbarabash, line 17: Something we may want to explore in the future is providing a way to swap out cache implementations.  During development it may be useful to debug/test with a cache that supports time-travel debugging by storing values using immutable-js (or something like that) so that devs could easily move between different cache states.  No action req'd.  Just an idle thought.

packages/wonder-blocks-data/src/util/in-memory-cache.js
```diff
@@ -0,0 +1,142 @@
...
+        scope,
+        id,
+    ): ?ValidData => {
+        return this._cache[scope]?.[id] ?? null;
+    };
+
+    /**
+     * Purge an item from the cache.
+     */
+    purge: (scope: string, id: string) => void = (scope, id) => {
```
kevinbarabash, line 74: In the other PR I seem to remember the equivalent of this function would return `false` if there was nothing to purge and `true` if purging was successful.  I'm curious why you dropped that behavior?

packages/wonder-blocks-data/src/util/__tests__/in-memory-cache.test.js
```diff
@@ -0,0 +1,396 @@
...
+            // Act
+            cache.purgeAll();
+            const result = cache.clone();
+
+            // Assert
+            expect(result).toStrictEqual({});
+        });
+    });
+
+    describe("#clone", () => {
```
kevinbarabash, line 326: Is there a test elsewhere that checks the cloned object is structurally equivalent?

packages/wonder-blocks-data/src/util/response-cache.js
kevinbarabash: If we find ourselves using single-scope caches in lots of places, we may want to provide a facade where specifying the scope isn't necessary.  No action req'd.

kevinbarabash (approved):
LGTM.  Reviewing this went much faster than reviewing the previous PR. 😄 :shipit:

packages/wonder-blocks-data/src/hooks/use-shared-cache.js
somewhatabstract: Yes. WB Data used to have a much more complex, defined caching API. I removed it because ultimately, we didn't in fact use it like we thought we would.

I was trying to envisage a scenario where we might want the cache to be in a specific state before starting a test and I wasn't able to convince myself that folks shouldn't do the right behavior to get it to that state rather than just jump there.

However, I think real world pain will illustrate it clearly and we can then act accordingly.

packages/wonder-blocks-data/src/util/in-memory-cache.js
somewhatabstract: Because nothing was using it and it was otherwise just bloating the code.

packages/wonder-blocks-data/src/util/__tests__/in-memory-cache.test.js
somewhatabstract: Technically, no, but since we're using Wonder Blocks Core's `clone` method, which does test that, we're probably safe.

packages/wonder-blocks-data/src/util/response-cache.js
somewhatabstract: Yes. I did have one, but I felt it wasn't being very helpful and just bloated the WB Data package unnecessarily.

PR #1341, opened 2022-02-09
[fei4261.2.typenames] Rename types to be more helpful

somewhatabstract:
## Summary:
This is just some housekeeping since we're about to do a breaking release of Wonder Blocks Data anyway.

Issue: FEI-4261

## Test plan:
`yarn flow`


packages/wonder-blocks-data/docs.md
```diff
@@ -81,30 +81,30 @@ If the predicate is not given, all items are removed.
...
     | {|
         error: string,
       |};
 
-type ResponseCache = {
-    [id: string]: CacheEntry,
+type CachedResponses = {
+    [id: string]: CachedResponse,
     ...,
 };
```
kevinbarabash, line 109: I tend to avoid pluralized type names since it's hard to know if the type is an array or an object just by looking at it.  I can hover over it thought to get that info so it's not a big deal.

packages/wonder-blocks-data/src/components/data.js
```diff
@@ -5,15 +5,15 @@ import {Server} from "@khanacademy/wonder-blocks-core";
 import {RequestFulfillment} from "../util/request-fulfillment.js";
 import InterceptContext from "./intercept-context.js";
 import {useServerEffect} from "../hooks/use-server-effect.js";
-import {resultFromCacheEntry} from "../util/result-from-cache-entry.js";
+import {resultFromCachedResponse} from "../util/result-from-cache-entry.js";
```
kevinbarabash, line 8: I like this.

kevinbarabash (approved):
LGTM.  The renaming I think will really aid in understanding, thanks for doing this.  This PR really benefited from the "hide whitespace" option.  🚢

packages/wonder-blocks-data/docs.md
somewhatabstract: Naming things is hard and with something already called `ResponseCache`, I didn't want to reuse it. 

However, we could rename that `ResponseCache` to `SsrCache` and then call this `ResponseCache`.

packages/wonder-blocks-data/src/components/data.js
somewhatabstract: The good fortune of using find/replace 😁

somewhatabstract: I'll rename the file to match

somewhatabstract:
> LGTM. The renaming I think will really aid in understanding, thanks for doing this. This PR really benefited from the "hide whitespace" option. 🚢

I use this Chrome extension so that it opens with `?w=1` automatically on my PRs.

https://chrome.google.com/webstore/detail/github-ignore-code-whites/hjlhnnpagncjflbcjfpoligpcppmkioi

PR #1343, opened 2022-02-09
Allow user to specify what element should receieve focus after modal is closed

nishasy:
## Summary:
We need to find a way to pass the return the focus to a specific element outside
the modal component once this is closed.
The JIRA ticket mentions three possible routes - we're going with #2.

- Add a new prop to pass a reference to the element we want to return focus after the modal is closed (`closedFocusId`).

Issue: https://khanacademy.atlassian.net/browse/WB-1090

https://user-images.githubusercontent.com/13231763/153308190-0655802b-87ed-46ce-a4a6-7c183cc5404b.mov

## Test plan:
- `yarn jest packages/wonder-blocks-modal/src/components/__tests__/modal-launcher.test.js`

Storybook
- Go to http://localhost:6061/?path=/story/floating-modal-onepanedialog--with-closed-focus-id
- Open the dropdown
- Click on or press enter on "open modal"
- Click on or press enter on "close modal"
- Confirm that the "Focus here after close" button has focus
- Press enter
- Confirm that the `actions` menu now has focus


packages/wonder-blocks-modal/src/components/__tests__/modal-launcher.test.js
```diff
@@ -156,7 +161,7 @@ describe("ModalLauncher", () => {
 
         // Rather than test this rigorously, we'll just check that a
         // ScrollDisabler is present, and trust ScrollDisabler to do its job.
-        const wrapper = shallow(
```
nishasy, line 159: `shallow` stopped working because I added an HOC to `modal-launcher.js`

packages/wonder-blocks-modal/src/components/__tests__/modal-launcher.test.js
```diff
@@ -289,45 +294,113 @@ describe("ModalLauncher", () => {
...
+        });
+    });
+
+    test("if `closedFocusId` is passed, shift focus to specified element after the modal closes", async () => {
+        // Arrange
+        const ModalLauncherWrapper = () => {
+            const [opened, setOpened] = React.useState(false);
+
+            const handleOpen = () => {
+                console.log("opening modal");
```
jandrade, line 342: I think it'd be better getting rid of these console.logs in the tests.

packages/wonder-blocks-modal/src/components/__tests__/modal-launcher.test.js
```diff
@@ -289,45 +294,113 @@ describe("ModalLauncher", () => {
...
+                setOpened(true);
+            };
+
+            const handleClose = () => {
+                console.log("closing modal");
+                setOpened(false);
+            };
+
+            return (
+                <View style={{gap: 20}}>
```
jandrade, line 352: Maybe this is copypasta from the stories :) Yo won't need to use `style` on tests, unless you are testing something related to styling (that is usually tested with Chromatic) :).

packages/wonder-blocks-modal/src/components/__tests__/modal-launcher.test.js
```diff
@@ -289,45 +294,113 @@ describe("ModalLauncher", () => {
...
+                </View>
+            );
+        };
+
+        render(<ModalLauncherWrapper />);
+
+        // Act
+        // Launch modal
+        const launcherButton = await screen.findByTestId("launcher-button");
+        userEvent.click(launcherButton);
```
jandrade, line 391: Nit: This section should probably go before act, b/c in this case the action related to the test is closing the modal, so opening the modal is part of `arrange`.

packages/wonder-blocks-modal/src/components/modal-launcher.js
```diff
@@ -43,10 +49,19 @@ type CommonProps = {|
      */
     initialFocusId?: string,
 
+    /**
+     * The selector for the element that will be focused after the dialog
+     * closes. When not set, the last element focused outside modal will be
```
jandrade, line 54:
```suggestion
     * closes. When not set, the last element focused outside the modal will be
```

packages/wonder-blocks-modal/src/components/__tests__/modal-launcher.test.js
```diff
@@ -289,45 +294,113 @@ describe("ModalLauncher", () => {
...
 
         // Act
         savedCloseModal(); // close the modal
-        wrapper.update();
 
         // Assert
-        expect(document.activeElement).toBe(lastButton);
+        await waitFor(() => {
+            expect(lastButton).toHaveFocus();
+        });
```
jandrade, line 333: Thanks for refactoring this test!

packages/wonder-blocks-modal/src/components/modal-launcher.js
```diff
@@ -158,11 +173,33 @@ export default class ModalLauncher extends React.Component<Props, State> {
...
+                    return;
+                }
+            }
+
+            if (lastElement != null) {
+                // Wait for the modal to leave the DOM before trying to
+                // return focus to the element that triggered the modal.
+                schedule.animationFrame(() => {
+                    lastElement.focus();
+                });
```
jandrade, line 202: And also for fixing this issue 🥇

jandrade, line 202: Nit: Taking another look at this code, I can see that you could extract this returning focus part into a separate function (the `onClose` part could remain here). Some benefits about that:

- It reduces indentation and improves readability.
- Abstracts the "return focus" logic into a separate logic that can help for future refactors.

wdyt?

packages/wonder-blocks-modal/src/components/one-pane-dialog.stories.js
```diff
@@ -200,3 +201,41 @@ WithOpener.parameters = {
         disableSnapshot: true,
     },
 };
+
+export const WithClosedFocusId: StoryComponentType = () => {
```
jandrade, line 205: Could you please add the `disableSnapshot` config here? This story doesn't need to be to be sent to Chromatic/visual testing because the thing we want to test doesn't appear on first render, but it happens until we complete a series of steps.

jandrade (commented):
Thanks for fixing this! The implementation is super close, but wanted to give some suggestions before approving:

- Don't forget to fix the Eslint warnings.
- Possible refactor on the "return focus" logic.
- Removing the story from chromatic snapshots.

jandrade (approved):
Looks great!! thanks for the recent changes :)

packages/wonder-blocks-modal/src/components/__tests__/modal-launcher.test.js
```diff
@@ -284,50 +294,140 @@ describe("ModalLauncher", () => {
         await wait();
 
         // Assert
+        // eslint-disable-next-line testing-library/no-node-access
         expect(document.activeElement).not.toBe(lastButton);
     });
 
     test("if modal is closed, return focus to the last element focused outside the modal", async () => {
```
nishasy, line 301: @jandrade This is the test I updated for the coverage test to stop failing!

packages/wonder-blocks-modal/src/components/__tests__/modal-launcher.test.js
```diff
@@ -284,50 +294,140 @@ describe("ModalLauncher", () => {
...
+        userEvent.click(lastButton);
 
-        // wait for styles to be applied
-        await wait();
+        // Act
+        // Close modal
+        const modalCloseButton = await screen.findByTestId(
+            "modal-close-button",
+        );
+        userEvent.click(modalCloseButton);
```
jandrade, line 357: Ah ofc this makes way more sense than calling the function directly!

packages/wonder-blocks-modal/src/components/__tests__/modal-launcher.test.js
jandrade: Thanks for fixing this test! the solution is great!

PR #1344, opened 2022-02-10
[fixdependencies] Make wonder-blocks package deps be peers

somewhatabstract:
## Summary:
Nisha has been seeing [flow issues](https://khanacademy.slack.com/archives/C4PE1QM5Y/p1644281707112309) when trying to update wonder blocks in webapp. I was also able to replicate this issue.

The flow issue is caused because it is encountering typography as a dependency of form in the form package's own nested node_modules before it encounters it elsewhere and this is causing some conflict. It turns out this is happening because we're bundling wonder-blocks deps with each package and it's loading them in an order that means they're not compatible.

I think having packages reference one another as peer dependencies resolves this for our main packages, but it doesn't necessarily help with when we want to alias a package (like we do with Wonder Blocks modal) - however, I think solving the main issue is relevant here and modal v1 should be fine because it uses the old broken "ship with my fellow packages as dependencies" model.

This is currently just a suggested change. We don't have to accept it, but if we don't then I think Flow is going to continue to get confused about mismatching copies of the exact same library. :(

Issue: XXX-XXXX

## Test plan:
Release this change and try it in webapp.

### DIFF
--- a/.changeset/famous-crabs-sit.md
+++ b/.changeset/famous-crabs-sit.md
@@ -1,0 +1,25 @@
+---
+"@khanacademy/wonder-blocks-birthday-picker": patch
+"@khanacademy/wonder-blocks-breadcrumbs": patch
+"@khanacademy/wonder-blocks-button": patch
+"@khanacademy/wonder-blocks-cell": patch
+"@khanacademy/wonder-blocks-clickable": patch
+"@khanacademy/wonder-blocks-core": patch
+"@khanacademy/wonder-blocks-data": patch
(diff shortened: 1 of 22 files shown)


somewhatabstract:
I'm not convinced this will solve the issue but it was a quick thing to do.

I notice that some WB packages include their peers in node_modules and some don't. I don't know what's causing them to install that way, but perhaps that's the real bug here. Something is perhaps calling install on sub-packages in node_modules? Maybe `make deps` is too eager?

PR #1345, opened 2022-02-10
[DRAFT] WB-646: Add aria-live to announce dropdown count on Screen Readers

jandrade:
## Summary:

Adding `aria-live="polite"` to announce the dropdown count on Screen Readers.

Issue: 

## Test plan:
TBD


jandrade:
The official implementation was done as part of #1468

PR #1346, opened 2022-02-10
[fei4261.interceptdata] Add useRequestInterception hook

somewhatabstract:
## Summary:
Well, here comes WB Data v5.

As I was integrating v4 into webapp, I realised that the `InterceptData` component was not very useful. So, this reworks it into `InterceptRequests` and introduces a new `useRequestInterception` hook to leverage it. The `Data` component uses this hook, but it also means that non-Data components (and hooks) can use it too, which introduces another way of mocking GraphQL requests in tests and stories, should folks want to intercept the raw WB request that doesn't have any of the nice polished types and things that our `mockGqlFetch` stuff provides.

Issue: FEI-4261

## Test plan:
`yarn test`
`yarn flow`


jeresig (approved):

somewhatabstract:
Forgot to update the docs and snapshots. On it.

PR #1348, opened 2022-02-11
[🔥AUDIT🔥] [fei4261.resultsnotcachedresponses] Export resultFromCachedResponse

somewhatabstract:
🖍 _This is an audit!_ 🖍

## Summary:
This just adds an export for code that already exists.

Issue: FEI-4261

## Test plan:
`yarn test`
`yarn flow`

### DIFF
--- a/packages/wonder-blocks-data/src/index.js
+++ b/packages/wonder-blocks-data/src/index.js
@@ -52,6 +52,7 @@ export {useServerEffect} from "./hooks/use-server-effect.js";
 export {useRequestInterception} from "./hooks/use-request-interception.js";
 export {useSharedCache, clearSharedCache} from "./hooks/use-shared-cache.js";
 export {ScopedInMemoryCache} from "./util/scoped-in-memory-cache.js";
+export {resultFromCachedResponse} from "./util/result-from-cache-response.js";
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-data/src/index.js
```diff
@@ -52,6 +52,7 @@ export {useServerEffect} from "./hooks/use-server-effect.js";
 export {useRequestInterception} from "./hooks/use-request-interception.js";
 export {useSharedCache, clearSharedCache} from "./hooks/use-shared-cache.js";
 export {ScopedInMemoryCache} from "./util/scoped-in-memory-cache.js";
+export {resultFromCachedResponse} from "./util/result-from-cache-response.js";
```
somewhatabstract, line 55: The `useServerEffect` hook returns the `CachedResponse` type. I think we want to keep it that way rather than have it return the `Result` type, though I could be convinced otherwise. So, rather than change that hook, I export the utility that can convert from the `CachedResponse` type to the `Result` type.

We could, in future remove this and just keep `CachedResponse` as an internal type instead, but for now, this will enable us to do the things we need to do without code duplication :)

PR #1351, opened 2022-02-15
[fei4327.1.useForceUpdate] Add useForceUpdate hook

somewhatabstract:
## Summary:
This adds a simple utility hook to allow components and hooks to trigger a re-render without requiring their own state to be updated.
This is useful when, for example, the state of the component comes from some shared cache or other location that doesn't warrant also sharing it in component state.

Issue: FEI-4327

## Test plan:
`yarn test`
`yarn flow`

### DIFF
--- a/packages/wonder-blocks-core/src/hooks/__tests__/use-force-update.test.js
+++ b/packages/wonder-blocks-core/src/hooks/__tests__/use-force-update.test.js
@@ -1,0 +1,54 @@
+// @flow
+import * as React from "react";
+import {render, act} from "@testing-library/react";
+import {renderHook} from "@testing-library/react-hooks";
+
+import {useForceUpdate} from "../use-force-update.js";
+
+describe("#useForceUpdate", () => {
+    it("should return a function", () => {
+        // Arrange
+
+        // Act
+        const {
+            result: {current: result},
+        } = renderHook(() => useForceUpdate());
+
+        // Assert
+        expect(result).toBeInstanceOf(Function);
+    });
+
+    describe("returned function", () => {
+        beforeEach(() => {
+            jest.useFakeTimers();
+        });
+
+        it("should cause component to render", () => {
+            // Arrange
+            const Component = (props): React.Node => {
+                const countRef = React.useRef(0);
+                const forceUpdate = useForceUpdate();
+                React.useEffect(() => {
+                    countRef.current++;
+
+                    setTimeout(forceUpdate, 50);
+                });
+                return countRef.current;
+            };
+
+            // Act
+            const wrapper = render(<Component />);
+            act(() => {
+                // Advance enough for the timeout to run 4 times.
+                // Which means the component should have rendered 4 times,
+                // with one more pending for the timeout that was setup in
+                // the last render.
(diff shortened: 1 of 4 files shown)


jeresig (approved):

packages/wonder-blocks-core/src/hooks/use-force-update.js
```diff
@@ -0,0 +1,16 @@
...
+ * Hook for forcing a component to update on demand.
+ *
+ * @returns {() => void} A function that forces the component to update.
+ */
+export const useForceUpdate = (): (() => void) => {
+    const [, setState] = React.useState(false);
+    const forceUpdate = React.useCallback(
+        () => setState((state) => !state),
+        [],
+    );
```
kevinbarabash, line 14: Passing a callback to `setState` ftw. 🤘

packages/wonder-blocks-core/src/hooks/use-force-update.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+import * as React from "react";
+
+/**
+ * Hook for forcing a component to update on demand.
+ *
```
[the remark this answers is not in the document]
kevinbarabash: At the very least this should probably be restricted for use in other hooks.

packages/wonder-blocks-core/src/index.js
```diff
@@ -12,6 +12,7 @@ export {
     useUniqueIdWithMock,
     useUniqueIdWithoutMock,
 } from "./hooks/use-unique-id.js";
+export {useForceUpdate} from "./hooks/use-force-update.js";
```
kevinbarabash, line 15: There should probably be a storybook documentation page for this hook since it's part of the public API.

kevinbarabash (approved):
Thanks for creating this hook.  Hopefully we won't have to use it that much.

packages/wonder-blocks-core/src/hooks/__tests__/use-force-update.test.js
```diff
@@ -0,0 +1,54 @@
...
+        // Assert
+        expect(result).toBeInstanceOf(Function);
+    });
+
+    describe("returned function", () => {
+        beforeEach(() => {
+            jest.useFakeTimers();
+        });
+
+        it("should cause component to render", () => {
```
kevinbarabash, line 26: It might be good to have a test that checks that calling `forceUpdate()` multiple times will result in multiple re-renders.

packages/wonder-blocks-core/src/index.js
somewhatabstract: I have a separate task for the docs work.

packages/wonder-blocks-core/src/hooks/__tests__/use-force-update.test.js
somewhatabstract: This test does just that.

packages/wonder-blocks-core/src/index.js
somewhatabstract: See FEI-4191

PR #1355, opened 2022-02-16
[fei4327.4.firstclassresult] Result<TData> as first class citizen in external API

somewhatabstract:
## Summary:
In this PR:

1. `useGql` delegates some responsibility to `useGqlRouterContext` and the new `mergeGqlContext` function
2. `useGql` can now take context adjustments directly so that the returned fetch can have them applied to all calls
3. `Result<TData` is now the default return of our data architecture, which partly paves the way for a universal "opinion" on how aborts will be done (that is currently unprioritized work), but also simplifies integration to webapp

Issue: FEI-4327

## Test plan:
`yarn test`
`yarn flow`

### DIFF
--- a/packages/wonder-blocks-data/src/components/__tests__/data.test.js
+++ b/packages/wonder-blocks-data/src/components/__tests__/data.test.js
@@ -168,14 +168,16 @@ describe("Data", () => {
                 });
             });
 
-            it("should render with aborted if the request resolves to null data", async () => {
+            it("should render with aborted if the request rejects with an abort error", async () => {
                 // Arrange
                 const fulfillSpy = jest.spyOn(
                     RequestFulfillment.Default,
                     "fulfill",
                 );
 
-                const fakeHandler = () => Promise.resolve(null);
+                const abortError = new Error("bang bang, abort!");
+                abortError.name = "AbortError";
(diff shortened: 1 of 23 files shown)


packages/wonder-blocks-data/src/components/__tests__/data.test.js
```diff
@@ -527,39 +535,6 @@ describe("Data", () => {
             });
         });
 
-        describe("with cached abort", () => {
```
somewhatabstract, line 530: We don't cache aborts.

packages/wonder-blocks-data/src/util/request-tracking.js
```diff
@@ -118,37 +118,62 @@ export class RequestTracker {
...
 
                 try {
                     promises.push(
                         this._requestFulfillment
-                            .fulfill(requestKey, options)
-                            .then((data) => {
-                                if (data == null) {
-                                    // Request aborted. We won't cache this.
-                                    return null;
+                            .fulfill(requestKey, {...options})
```
somewhatabstract, line 125: We spread options because the `fulfill` method currently has `hydrate` as optional whereas we assert it can never be null. That messes with flow.

jeresig (approved):
Looks good to me, thank you!

PR #1356, opened 2022-02-18
[fei4327.5.useHydratableEffect] useHydratableEffect hook

somewhatabstract:
## Summary:
This adds `useHydratableEffect` to WB Data. This is basically a generic version of `useFastlyQuery` from webapp (it was created by comparing the `Data` element implementation and `useFastlyQuery` to incorporate all the requirements).

The next PR will simplify the implementation a little by moving some common functionality to a new `useCachedEffect` hook, and then I'll be writing stories to document (and integrating all this into webapp).

Note that to achieve this, we now have a `Status` API for generating result statuses, and so that client-side Results can contain real errors, I have split the `ScopedInMemoryCache` and the serializable version of it into two things (one based on the other).

You can also now see that the `Data` element is now back to being entirely implemented by a hook, but this time, unlike `useData`, the hook is better architected/typed/abstracted, etc. based off real usage rather than old usage and guesswork.

Issue: FEI-4327

## Test plan:
`yarn test`
`yarn flow`


packages/wonder-blocks-data/src/components/__tests__/data.test.js
```diff
@@ -132,8 +143,8 @@ describe("Data", () => {
                 );
 
                 // Assert
-                expect(fakeChildrenFn).toHaveBeenCalledTimes(2);
-                expect(fakeChildrenFn).toHaveBeenLastCalledWith({
+                // expect(fakeChildrenFn).toHaveBeenCalledTimes(2);
```
jeresig, line 146: Old test?

jeresig (approved):
Looks reasonable to me!

packages/wonder-blocks-data/src/components/__tests__/data.test.js
somewhatabstract: Ah yes, I was having to tinker with the test cases to work out what was going on. Good catch. Thanks.

PR #1362, opened 2022-03-01
Update CODEOWNERS and REVIEWERS to use the Wonder Blocks team.

jeresig:
## Summary:
Previously we were tagging individuals, which doesn't feel right. Instead let's use the Wonder Blocks team for this!

Issue: XXX-XXXX

## Test plan:

### DIFF
--- a/.github/REVIEWERS
+++ b/.github/REVIEWERS
@@ -50,4 +50,4 @@ Regex Examples:
 [ON PULL REQUEST] (DO NOT DELETE THIS LINE)
 
 # Make these people reviewers for any changes made to any file
-**/*                @Khan/frontend-infra!
+**/*                @Khan/wonder-blocks!
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,14 +1,1 @@
-*   @kevinbarabash @jeresig @somewhatabstract @jandrade @github-actions[bot]
-packages/wonder-blocks-breadcrumbs/*   @jandrade @kevinbarabash @jeresig
-packages/wonder-blocks-button/*   @kevinbarabash @jeresig
-packages/wonder-blocks-color/*   @kevinbarabash @jeresig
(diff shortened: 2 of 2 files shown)


.github/REVIEWERS
```diff
@@ -50,4 +50,4 @@ Regex Examples:
 [ON PULL REQUEST] (DO NOT DELETE THIS LINE)
 
 # Make these people reviewers for any changes made to any file
-**/*                @Khan/frontend-infra!
+**/*                @Khan/wonder-blocks!
```
jandrade, line 53: Good call!

CODEOWNERS
```diff
@@ -1,14 +1 @@
...
-packages/wonder-blocks-data/*   @somewhatabstract @jeresig
-packages/wonder-blocks-icon-button/*   @kevinbarabash @jeresig
-packages/wonder-blocks-link/*   @kevinbarabash @jeresig
-packages/wonder-blocks-modal/*   @jandrade @kevinbarabash @jeresig
-packages/wonder-blocks-popover/*   @jandrade @jeresig
-packages/wonder-blocks-testing/*   @somewhatabstract @jandrade @jeresig
-packages/wonder-blocks-timing/*   @somewhatabstract @jeresig
-packages/wonder-blocks-tooltip/*   @somewhatabstract @kevinbarabash @jeresig
-packages/wonder-blocks-typography/*   @briangenisio @kevinbarabash @jeresig
+*   @Khan/wonder-blocks
```
jandrade, line 1: Out of curiosity... do you think we still need this file now that we use Gerald to notify the team?

jandrade (approved):
Thanks for simplifying this! Just added a question about the `CODEOWNERS` file.

CODEOWNERS
jeresig: I don't know - I guess we may not need it! We could just rely on Gerald... maybe I'll just remove this file instead!

somewhatabstract (approved):

PR #1366, opened 2022-03-03
[fei4322.fix] Fix bug where changing requestId after hydrating result would actually update

somewhatabstract:
## Summary:
In investigating an SSR issue I saw, I found this client-side bug. It doesn't explain why the SSR was not working in the case I saw, but it is definitely a bug. I've added a test case to cover this issue which failed before I implemented the fix.

Issue: FEI-4322

## Test plan:
`yarn test`

Tried this out in dev server for webapp to verify it fixed the client-side bug I saw. Interestingly, SSR and hydration is working in dev which makes me think the SSR issue I saw in staging was down to something else entirely.

### DIFF
--- a/packages/wonder-blocks-data/src/hooks/use-hydratable-effect.js
+++ b/packages/wonder-blocks-data/src/hooks/use-hydratable-effect.js
@@ -178,9 +178,16 @@ export const useHydratableEffect = <TData: ValidCacheData>(
                 }
                 return null;
         }
-        // There is no reason for this to change after the first render.
+        // There is no reason for this to change after the first render,
+        // you might thing, but the function closes around serverResult and if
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-data/src/hooks/use-hydratable-effect.js
```diff
@@ -178,9 +178,16 @@ export const useHydratableEffect = <TData: ValidCacheData>(
                 }
                 return null;
         }
-        // There is no reason for this to change after the first render.
+        // There is no reason for this to change after the first render,
+        // you might thing, but the function closes around serverResult and if
```
jeresig, line 182:
```suggestion
        // you might think, but the function closes around serverResult and if
```

jeresig (approved):

PR #1369, opened 2022-03-04
[fei4350.1.node16andchangesetfix] Workflow amendments

somewhatabstract:
## Summary:
- Node 16 as test target

- minimum node version as 12.22 (needed for enums)

- Fix release workflow to ensure pushes and new PRs trigger other workflows

- Remove unneeded publish job from ci workflow

Issue: FEI-4350

## Test plan:
See what happens in CI when we run with these updated workflows.


jeresig (approved):
Looks good to me, thank you!

somewhatabstract:
OK, the build size check will fail because it will run the old package.json against yarn in Node 16 and that will fail due to the node-sass issue that this PR fixes.

So, I'm going to split that bit into another PR and get that up first.

PR #1370, opened 2022-03-04
[fei4350.0.node16precursor] Remove need to even care about node-sass

somewhatabstract:
## Summary:
This removes the dependency on node-sass, which we don't need, but that would cause yarn to fail when operating under Node 16. This means that when we actually update things to Node 16 in the subsequent PR, actions will pass properly.

Issue: FEI-4350

## Test plan:
Run styleguidist and see all is well
Run `yarn` under node 16


jeresig (approved):
Good catch, thank you!

PR #1371, opened 2022-03-07
Always show error alert for LabeledTextField.

jeresig:
## Summary:
In order for Aria alerts to be readable by a screenreader they must exist in a page before the text is inserted into them. So we can fix this by making sure that the alert wrapper exists in the page for all text field components.

See: https://pauljadam.com/demos/aria-alert-assertive-validation.html

Issue: FEI-3472

## Test plan:
Using a screen reader (Voiceover in Safari) I visited:
https://wonder-blocks.netlify.app/#labeledtextfield

And navigated to a field with an error. I confirmed that entering and exiting the field the error text was read as an alert.

### DIFF
--- a/packages/wonder-blocks-form/src/components/field-heading.js
+++ b/packages/wonder-blocks-form/src/components/field-heading.js
@@ -121,14 +121,15 @@ export default class FieldHeading extends React.Component<Props> {
     maybeRenderError(): ?React.Node {
         const {error, id, testId} = this.props;
 
-        if (!error) {
-            return null;
-        }
-
+        // NOTE(jeresig, FEI-3472): We need to always display the error
+        // element, even if there is no error. This is because the error
+        // uses an ARIA alert role, which requires that the element be
+        // in the DOM before the text is inserted. Doing otherwise will
+        // mean that the alert will not be read.
         return (
             <React.Fragment>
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-form/src/components/field-heading.js
```diff
@@ -121,14 +121,15 @@ export default class FieldHeading extends React.Component<Props> {
...
+        // element, even if there is no error. This is because the error
+        // uses an ARIA alert role, which requires that the element be
+        // in the DOM before the text is inserted. Doing otherwise will
+        // mean that the alert will not be read.
         return (
             <React.Fragment>
-                <Strut size={Spacing.small_12} />
-                {typeof error === "string" ? (
+                {error && <Strut size={Spacing.small_12} />}
+                {typeof error === "string" || !error ? (
```
jandrade, line 132: Super nit: I was taking a look at the source code, and it looks that we can even simplify this more.

We currently define error as: `error: string | React.Element<Typography>`, but...

we are only passing `error: string` from the callers, so maybe we can simplify the `error` type:

```
error?: string,
```

That way we can move the condition inside line # 139.

Of course no need to do it here, but wanted to point out that finding :)

jandrade (approved):
Looks great! thanks for adding the test.

PR #1372, opened 2022-03-07
Add wonder-blocks-search-field package

nishasy:
## Summary:
This PR introduces a new package for Search Field: `wonder-blocks-saerch-field`.

Issue: https://khanacademy.atlassian.net/browse/WB-1117

## Test plan:
`yarn test`


jeresig (approved):
Woohoo! Exciting 😄

jandrade (approved):
Looks great! thanks for creating this as a separate PR.

-------

One minor suggestion about landing these changes (apologies for not mention it before).... given that this is your first new package, I'd recommend you to land this PR only after you finish the whole `SearchField` component implementation.

**Why:** By doing this, you can avoid landing premature search-field changes in `main`, and at the same time, other folks can release other features/bug fixes/etc independently without having to publish the new `search-field` package.

### PR strategies

There are two options I can think of:

1. Use a different base branch to land these `SearchField` related PRs into a separate branch (something like a "deploy branch"). Once you are done, then you can merge that deploy branch into `main`. Note: This would probably require an extra PR from your side to merge into `main`. Also if you take this approach, that'd mean that you'd need to change the base branch on this commit.

2. Another alternative, that would be creating stacked PRs, then land all the PRs once the last one gets approved. 
e.g. `create-search-field-package` -> `search-field-component`. so you land first this PR, then land right after the next PR.

#### Recommendation
I'd recommend to take this approach if you think that you'll only create another PR, but if there are more PRs in the way, probably option 1 requires less branch maintenance (e.g. having to merge the parent/base branch).

Happy to chat more in detail about that.

PR #1373, opened 2022-03-10
[fei4385.1.skipservereffect] Add skip to useServerEffect

somewhatabstract:
## Summary:
This fixes an issue with `useServerEffect` and `useHydratableEffect` where skipping the effect would cause an infinite render loop because the server effect would get tracked but would always result in an abort that never updated the server/hydrate cache, and therefore it would just get tracked again.

Issue: FEI-4385

## Test plan:
`yarn test`

When we update webapp with this version of WB Data, skipping `useFastlyQuery` usage will no longer cause an infinite loop in SSR tests nor in the render-gateway.


packages/wonder-blocks-data/src/hooks/__tests__/use-hydratable-effect.test.js
```diff
@@ -107,34 +107,11 @@ describe("#useHydratableEffect", () => {
...
-            serverRenderHook(() =>
-                useHydratableEffect("ID", fakeHandler, {skip: true}),
-            );
-            const underTest = useServerEffectSpy.mock.calls[0][1]();
-
-            // Assert
-            await expect(underTest).rejects.toMatchInlineSnapshot(
-                `[AbortError: skipped]`,
-            );
-        });
```
somewhatabstract, line 136: We don't do this anymore. :)

packages/wonder-blocks-data/src/hooks/__tests__/use-server-effect.test.js
```diff
@@ -109,6 +109,57 @@ describe("#useServerEffect", () => {
...
+            );
+
+            // Act
+            serverRenderHook(() => useServerEffect("ID", fakeHandler), {
+                wrapper: TrackData,
+            });
+
+            // Assert
+            expect(trackDataRequestSpy).not.toHaveBeenCalled();
+        });
```
somewhatabstract, line 161: Tests to make sure we are not tracking requests that we shouldn't track.

PR #1375, opened 2022-03-14
[🔥AUDIT🔥] Add turbosnap to only run affected stories on Chromatic

jandrade:
🖍 _This is an audit!_ 🖍

## Summary:

Adding `turbosnap` to optimize visual testing with Chromatic. Adding this change
means that Chromatic will only generate visual snapshots for the stories that
are really affected (instead of generating snapshots for ALL stories).

See https://www.chromatic.com/docs/turbosnap#configure

Issue: XXX-XXXX

## Test plan:

Verify the chromatic action and see if the Chromatic build changes.

For reference: https://www.chromatic.com/docs/turbosnap#how-can-i-tell-if-turbosnap-is-working


kevinbarabash (approved):
So cool!  Thanks for adding this.  I wonder how many snapshots this will save? 🤔

PR #1376, opened 2022-03-14
WB-1257: Add href to Cell

jandrade:
## Summary:

Added `href` to both `CompactCell` and `DetailCell`.

Issue: WB-1257

## Test plan:

- Navigate to /?path=/story/cell-detailcell--detail-cell-navigation  
- Verify that the "See navigation changes here" text changes after clicking on any of the
cells.


jeresig (approved):
Very nice, thank you!

PR #1377, opened 2022-03-14
[🔥AUDIT🔥] Add GitHub pages action to publish storybook in gh-pages branch

jandrade:
🖍 _This is an audit!_ 🖍

## Summary:

Adds a new step as part of the Release job to build the Storybook static version
and push those generated files to the `gh-pages` branch.

Issue: XXX-XXXX

## Test plan:

Verify that https://khan.github.io/wonder-blocks works a couple of minutes after
landing this PR.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,6 +47,18 @@ jobs:
       - name: Install Dependencies
         run: yarn
 
+      - name: Build Storybook
+        # Generate a static version of storybook inside "build/"
+        run: yarn build-storybook
+
+      - name: Deploy to GitHub pages
+        uses: JamesIves/github-pages-deploy-action@v4.2.5
+        with:
+          # The branch the action should deploy to.
+          branch: gh-pages
+          # The folder the action should deploy.
+          folder: build
+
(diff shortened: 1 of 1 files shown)


kevinbarabash (approved):
That was easy.  I think I'm going to start using this in my own projects.  Thanks for setting this up.  Are there plans to also use this for PRs?

jandrade:
> That was easy. I think I'm going to start using this in my own projects. Thanks for setting this up. Are there plans to also use this for PRs?

🎉 I think it wouldn't be strictly necessary on this repo as we still can use Chromatic for that. But still it's a really good idea to add it in the future!

If we add this for PRs, one thing that we could do is to generate a subfolder per PR, that maps directly to the PR ID.

e.g.
`PR #1 => gh-docs/1/ => khan.github.io/wonder-blocks/1/`

Just food for thought, but I think this would simplify that a lot, and would also help devs to map the PR id to a "friendly" URL. Wdyt?

jeresig:
@jandrade Yeah, I like that! It may be worth using something like `wonder-blocks/pulls/1377` so that the URL matches more exactly what's on Github.

PR #1380, opened 2022-03-15
Add cell components to our clickable consistency test suite

kevinbarabash:
## Summary:
The consistency tests found that the 'target' prop was missing.  We may also want to expand this test suite to also include 'tabIndex' in the future since that's another common prop that people may want to use on clickables.  I'll leave that for another day.

Issue: None

## Test plan:
- yarn jest clickables.test.js

### DIFF
--- a/packages/wonder-blocks-cell/src/util/types.js
+++ b/packages/wonder-blocks-cell/src/util/types.js
@@ -136,6 +136,8 @@ export type CellProps = {|
     /**
      * A target destination window for a link to open in. Should only be used
      * when `href` is specified.
+     *
+     * TODO(WB-1262): only allow this prop when `href` is also set.t
      */
     target?: "_blank",
 |};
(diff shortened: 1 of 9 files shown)


packages/wonder-blocks-cell/src/util/types.js
```diff
@@ -132,4 +132,10 @@ export type CellProps = {|
      * by default if react-router is present.
      */
     href?: string,
+
+    /**
+     * A target destination window for a link to open in. Should only be used
+     * when `href` is specified.
+     */
+    target?: "_blank",
```
somewhatabstract, line 140: Should this be a variation of the `href` prop instead, since it's only relevant when that's set? For example:

```js
    href?: string | {|href: string, target: "_blank"|},
```

somewhatabstract (approved):
Seems good to me. One unimportant suggestion inline.

PR #1385, opened 2022-03-23
WB-1258: Use react-window only for long lists

jandrade:
## Summary:

- Modified `DropdownCore` to only use `react-window` if there are more than 80
options in the menu.
- Improved the performance a bit by avoiding to recalculate rendering the options list on every re-render.
- Created `DropdownPopper` to abstract out the `PopperJS` functionality into a
separate component. This will help us to make improvements in the future (e.g.
porting to hooks, using PopperV3/floating-ui).
- Modified unit tests inside `wb-dropdown` to avoid relying on timers to test the dropdowns.

Current benchmarks for showing the dropdown menu without `react-window`:

BEFORE IMPROVING PERFORMANCE: 173ms for 70 items.

Screen Shot 2022-03-23 at 11 27 27 AM

AFTER IMPROVING PERFORMANCE: ~90ms for 70 items (average).

dropdown-core-70

Issue: WB-1258

## Test plan:

Navigate to the Stories related to Dropdown components and verify that everything continues working as expected.

Verify that this example (that doesn't use `react-window`) works as expected: keyboard navigation works, focus, selecting items, filtering items, etc.

https://5e1bf4b385e3fb0020b7073c-tzbtketfim.chromatic.com/?path=/story/dropdown-singleselect--default-single-select-opened

NO-WINDOWING VERSION
https://user-images.githubusercontent.com/843075/159743719-7ca3078f-3886-41c2-9515-e981077633a1.mov

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/__tests__/dropdown-core.test.js
+++ b/packages/wonder-blocks-dropdown/src/components/__tests__/dropdown-core.test.js
@@ -32,13 +32,32 @@ const items = [
 ];
 
 describe("DropdownCore", () => {
-    beforeEach(() => {
-        jest.useFakeTimers();
-    });
+    it("should throw for invalid role", () => {
+        // Arrange
+        // Passing an invalid role will throw an error.
+        jest.spyOn(console, "error").mockImplementation(() => {});
 
-    afterEach(() => {
-        jest.runOnlyPendingTimers();
-        jest.useRealTimers();
+        // Act
+        const underTest = () =>
+            render(
+                <div>
+                    <button data-test-id="external-button" />
+                    <DropdownCore
+                        initialFocusedIndex={0}
+                        // mock the items
+                        items={[]}
+                        role={("invalid": any)}
+                        open={true}
+                        // mock the opener elements
+                        opener={<button />}
+                        openerElement={null}
+                        onOpenChanged={jest.fn()}
+                    />
+                </div>,
+            );
+
+        // Assert
+        expect(underTest).toThrow();
     });
 
     it("focus on the correct option", () => {
@@ -711,5 +730,103 @@ describe("DropdownCore", () => {
 
         // Assert
         expect(preventDefaultMock).toHaveBeenCalledTimes(0);
+    });
+
+    describe("VirtualizedList", () => {
+        const optionItems = new Array(100).fill(null).map((_, i) => ({
+            component: (
+                <OptionItem
+                    key={i}
+                    value={(i + 1).toString()}
+                    label={`Fruit # ${i + 1}`}
+                    testId={`item-${i}`}
+                />
+            ),
+            focusable: true,
+            populatedProps: {},
+        }));
+
+        it("should render a virtualized list of options and focus on the search field", async () => {
+            // Arrange
+            render(
+                <DropdownCore
+                    initialFocusedIndex={undefined}
+                    onSearchTextChanged={jest.fn()}
+                    searchText=""
+                    // mock the items
+                    items={[
+                        {
+                            component: (
+                                <SearchTextInput
+                                    testId="search-text-input"
+                                    key="search-text-input"
+                                    onChange={jest.fn()}
+                                    searchText={""}
+                                />
+                            ),
+                            focusable: true,
+                            populatedProps: {},
+                        },
+                        ...optionItems,
+                    ]}
+                    role="listbox"
+                    open={true}
+                    // mock the opener elements
+                    opener={<button />}
+                    openerElement={null}
+                    onOpenChanged={jest.fn()}
+                />,
+            );
+
+            await screen.findByRole("listbox");
+
+            // Act
+            const searchField = await screen.findByPlaceholderText("Filter");
+
+            // Assert
+            expect(searchField).toHaveFocus();
+        });
+
+        it("should focus on the item after clicking on it", async () => {
+            // Arrange
+            render(
+                <DropdownCore
+                    initialFocusedIndex={undefined}
+                    onSearchTextChanged={jest.fn()}
+                    searchText=""
+                    // mock the items
+                    items={[
+                        {
+                            component: (
+                                <SearchTextInput
+                                    testId="search-text-input"
+                                    key="search-text-input"
+                                    onChange={jest.fn()}
+                                    searchText={""}
+                                />
+                            ),
+                            focusable: true,
+                            populatedProps: {},
+                        },
+                        ...optionItems,
+                    ]}
+                    role="listbox"
+                    open={true}
+                    // mock the opener elements
+                    opener={<button />}
+                    openerElement={null}
+                    onOpenChanged={jest.fn()}
+                />,
+            );
+
+            await screen.findByRole("listbox");
+
+            // Act
+            const item = await screen.findByTestId("item-1");
+            userEvent.click(item);
+
+            // Assert
+            expect(item).toHaveFocus();
+        });
     });
 });
--- a/packages/wonder-blocks-dropdown/src/components/__tests__/dropdown-core.test.js
+++ b/packages/wonder-blocks-dropdown/src/components/__tests__/dropdown-core.test.js
@@ -1,13 +1,11 @@
 // @flow
 import * as React from "react";
-import {fireEvent, render, screen, waitFor} from "@testing-library/react";
-import userEvent from "../../../../../utils/testing/user-event.js";
+import {fireEvent, render, screen} from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
 import OptionItem from "../option-item.js";
 import SearchTextInput from "../search-text-input.js";
 import DropdownCore from "../dropdown-core.js";
-
-jest.mock("../dropdown-core-virtualized.js");
 
 const items = [
     {
@@ -43,7 +41,7 @@ describe("DropdownCore", () => {
         jest.useRealTimers();
     });
 
-    it("focus on the correct option", async () => {
+    it("focus on the correct option", () => {
         // Arrange
         render(
             <DropdownCore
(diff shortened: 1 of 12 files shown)


packages/wonder-blocks-dropdown/src/components/single-select.md
```diff
@@ -477,7 +477,7 @@ class ExampleWithFilter extends React.Component {
     constructor() {
         super();
         this.state = {
-            opened: true,
+            opened: false,
```
jandrade, line 480: Made this change to avoid an issue with `Enzyme` (using gen-snapshots). This case is already covered by Chromatic.

packages/wonder-blocks-dropdown/src/components/__tests__/dropdown-core.test.js
```diff
@@ -1,14 +1,12 @@
 // @flow
 import * as React from "react";
-import {fireEvent, render, screen, waitFor} from "@testing-library/react";
-import userEvent from "../../../../../utils/testing/user-event.js";
+import {fireEvent, render, screen} from "@testing-library/react";
```
jandrade, line 3: Now that we are not using `react-window` by default, I was able to remove the custom `user-event.js` implementation and remove some extra steps in all the tests inside this file.

kevinbarabash: Yay! Do this mean that we no longer need to wait in the webapp tests for things that use wonder-blocks-dropdown?

packages/wonder-blocks-dropdown/src/components/__tests__/dropdown-core.test.js
```diff
@@ -88,13 +102,6 @@ describe("DropdownCore", () => {
             />,
         );
 
-        // Wait until the dropdown is open
-        await screen.findByRole("listbox");
-
-        // RTL's focuses on `document.body` by default, so we need to focus on
-        // the dropdown menu
-        userEvent.tab();
```
kevinbarabash, line 96: Why can we get rid of this?

packages/wonder-blocks-dropdown/src/components/__tests__/dropdown-core.test.js
```diff
@@ -391,7 +363,7 @@ describe("DropdownCore", () => {
                                 key="1"
                             />
                         ),
-                        focusable: false,
+                        focusable: true,
```
kevinbarabash, line 366: Why did this change?

packages/wonder-blocks-dropdown/src/components/__tests__/dropdown-core.test.js
```diff
@@ -536,19 +492,12 @@ describe("DropdownCore", () => {
                 role="listbox"
                 open={true}
                 // mock the opener elements
-                opener={<button />}
+                opener={<button data-test-id="opener" />}
```
kevinbarabash, line 495: This test doesn't seem to make use of this `data-test-id`.

packages/wonder-blocks-dropdown/src/components/__tests__/dropdown-core.test.js
```diff
@@ -797,4 +731,102 @@ describe("DropdownCore", () => {
...
+                    role="listbox"
+                    open={true}
+                    // mock the opener elements
+                    opener={<button />}
+                    openerElement={null}
+                    onOpenChanged={jest.fn()}
+                />,
+            );
+
+            await screen.findByRole("listbox");
```
kevinbarabash, line 822: Do we need to wait for this role, it seems like it should be available right away.

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -382,21 +395,28 @@ class DropdownCore extends React.Component<Props, State> {
...
     scheduleToFocusCurrentItem() {
-        // wait for windowed items to be recalculated
-        this.props.schedule.animationFrame(() => this.focusCurrentItem());
+        if (this.shouldVirtualizeList()) {
+            // wait for windowed items to be recalculated
+            this.props.schedule.animationFrame(() => this.focusCurrentItem());
+        } else {
+            // immediately focus the current item if we're not virtualizing
+            this.focusCurrentItem();
+        }
```
kevinbarabash, line 404: Nice!

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -579,11 +599,98 @@ class DropdownCore extends React.Component<Props, State> {
...
+
+    /**
+     * Renders the non-virtualized list of items.
+     */
+    renderList(): React.Node {
+        let focusCounter = 0;
+        const itemRole = this.getItemRole();
+
+        // if we don't need to virtualize, we can render the list directly
+        return this.props.items.map((item, index) => {
```
kevinbarabash, line 636: How much of this logic is shared with dropdown-core-virtualized.js?

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -661,83 +776,37 @@ class DropdownCore extends React.Component<Props, State> {
...
 
-        return null;
+        // Preprocess the items that are used inside the Popper instance. By
+        // doing this, we optimize the list to be processed only one time
+        // instead of every time popper changes.
+        // NOTE: This improves the performance impact of the dropdown by
+        // reducing the execution time up to 2.5X.
+        const listRenderer = this.shouldVirtualizeList()
+            ? this.renderVirtualizedList()
+            : this.renderList();
```
kevinbarabash, line 795: I wonder if memoizing some of this computation would result in further perf gains.  No action req'd for this PR.

packages/wonder-blocks-dropdown/src/components/dropdown-popper.js
```diff
@@ -0,0 +1,105 @@
...
+    alignment = "left",
+    onPopperElement,
+    referenceElement,
+}: Props): React.Node {
+    // If we are in a modal, we find where we should be portalling the menu by
+    // using the helper function from the modal package on the opener element.
+    // If we are not in a modal, we use body as the location to portal to.
+    const modalHost =
+        maybeGetPortalMountedModalHostElement(referenceElement) ||
+        document.querySelector("body");
```
kevinbarabash, line 68: I forgot why we did this.  Is this to simplify focus handling?

packages/wonder-blocks-dropdown/src/components/dropdown-popper.js
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * Styles that will be applied to the children.
+     */
+    style?: StyleType,
+|};
+
+/**
+ * A wrapper for PopperJS that renders the children inside a portal.
+ */
+export default function DropdownPopper({
```
kevinbarabash, line 57: Good call splitting this out into its own component.  dropdown-core.js is getting pretty complicated.

kevinbarabash (approved):
LGTM.  This really simplifies the tests.  We should be able to simplify some of the tests in webapp too once this ships, right?  I left a few comments, but nothing blocking. 🚢

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
jandrade: It's like 60% - 70%. The virtualized version process this in two steps:

1. Modifies the items in this file (`parseVirtualizedItems()`), so we can pass any logic related to keyboard/focus navigation. (I added a note in that function explaining that distinction).
2. Then, in `dropdown-core-virtualized-item`, it maps every item to clone it using the data processed in point # 1.

The "regular" version does both steps directly here, so there are no performance implications.

I can probably add another note here explaining all of this?, wdyt?

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
jandrade: Good point! I made some initial explorations here, but decided to leave it for the next phase (see a [comment above](https://github.com/Khan/wonder-blocks/pull/1385/files#diff-2f8aa6cd2de3d71f2a379edd27259800ac222fa77bbee3584ff143e64402c2f5R37) pointing to that JIRA ticket :)). I'll add this note to the ticket!

One point to clarify, do you see this as memoizing the computation of the logic (`React.useMemo()`) or memoizing rendering the elements (`React.memo()`)? What do you think what model would fit better for this scenario?

packages/wonder-blocks-dropdown/src/components/dropdown-popper.js
jandrade: Yep, and I even have more plans to reorganize this file (to be working on that on a separate ticket).

packages/wonder-blocks-dropdown/src/components/dropdown-popper.js
jandrade: It's to avoid having layout context issues, in case we are attaching the Dropdown below a modal dialog (if there's one).

jandrade:
> We should be able to simplify some of the tests in webapp too once this ships, right?

Right, if these dropdowns have less than 80 items. So in the case of the BirthdayPicker we will still have to rely on the `react-window` version because we are displaying 120 years.

That makes me wonder... what if we move up the threshold to those 120 items? Just checked perf. and the execution time goes to `~110ms` (instead of `90ms`). cc @kevinbarabash @jeresig

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
kevinbarabash: No comment is necessary.  I was curious what the potentially for deduplication was.

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
kevinbarabash: That's a good question.  I think `React.memo()` avoids re-renders if the props are the same, but doesn't do anything for state changes (I could be wrong).  `React.useMemo()` seems like a better fit, but would require migration to hooks.  Maybe something to revisit in the future as wonder-blocks gets more hook-ified. 🙂

packages/wonder-blocks-dropdown/src/components/dropdown-popper.js
kevinbarabash: Was the z-index issue between wonder-blocks-dropdown and wonder-blocks-modal (or was it an issue with one webapp's modal components)?

kevinbarabash:
> That makes me wonder... what if we move up the threshold to those 120 items? Just checked perf. and the execution time goes to ~110ms (instead of 90ms). cc @kevinbarabash @jeresig

SGTM.  Go for it!

kevinbarabash (approved):

PR #1386, opened 2022-03-24
Add new SearchField component

nishasy:
## Summary:
Added new SearchField component
- search-field.js
- search-field.test.js
- search-field.stories.js

Issue: https://khanacademy.atlassian.net/browse/WB-1117

Demo:
https://user-images.githubusercontent.com/13231763/160504207-b0ce9594-21c1-4de1-a7ce-460320e35de2.mov

Docs:
Screen Shot 2022-03-28 at 4 30 35 PM

Light and Disabled stories:
image

Light mode focused:
Screen Shot 2022-03-28 at 4 32 24 PM

Light mode focused with X:
Screen Shot 2022-03-28 at 4 33 02 PM

## Test plan:
`yarn test packages/wonder-blocks-search-field/src/components/__tests__/search-field.test.js`

Manual testing in storybook
- Check that you can type in the textbox
- Clear button should show up after typing
- Clear button should go away if the field is clear
- Check spacing and borders
- Test focusing in and out with clicking and with keyboard nav


packages/wonder-blocks-search-field/src/components/search-field.stories.js
```diff
@@ -0,0 +1,34 @@
+// @flow
```
nishasy, line 1: For some reason, this is showing up under BirthdayPicker in storybook in the first (unnamed?) section. How do I change the placement? Where should it go?

packages/wonder-blocks-search-field/src/components/__tests__/search-field.test.js
```diff
@@ -0,0 +1,181 @@
+// @flow
```
nishasy, line 1: I tried to cover all the test cases I could think of for now, but it's very possible that I missed some. Are there any other tests I should add?

packages/wonder-blocks-search-field/src/components/search-field.stories.js
```diff
@@ -0,0 +1,34 @@
...
+
+import type {StoryComponentType} from "@storybook/react";
+
+import SearchField from "../index.js";
+
+export default {
+    title: "SearchField",
+};
+
+export const Default: StoryComponentType = () => {
```
kevinbarabash, line 12: There are a bunch of other variants that should probably be showcased as well, e.g. `disabled`, `light`, something with custom styling?  Once the flow issue is fixed and the chromatic action starts working again it'll snapshot each of the stories.  For some of the other components we minimize the number of snapshots by having one story that shows a whole bunch of variants together and then we disable chromatic on all of the stories except for the one with multiple variants.

kevinbarabash (changes requested):
This is a great start on this component.  It'll be nice to finally have an official `SearchField` component.  Requesting a few changes (see inline comments for more details):
- more stories to covert different variants
- tests for event handlers
- using the "action" storybook add-on to log events in storybook (this is more of a nice-to-have)

packages/wonder-blocks-search-field/src/components/__tests__/search-field.test.js
```diff
@@ -0,0 +1,181 @@
...
+                value=""
+                onChange={() => {}}
+                testId="search-field-test"
+                aria-label="some-aria-label"
+            />,
+        );
+
+        // Assert
+        const input = screen.getByRole("textbox");
+        expect(input).toHaveAttribute("aria-label", "some-aria-label");
```
jandrade, line 50: Nice test!

packages/wonder-blocks-search-field/src/components/__tests__/search-field.test.js
```diff
@@ -0,0 +1,181 @@
...
+        userEvent.tab();
+        // blur
+        userEvent.tab();
+
+        // Assert
+        const searchField = screen.getByTestId("search-field-test");
+        expect(searchField).not.toHaveFocus();
+    });
+
+    test("does not have clear icon by default", () => {
```
jandrade, line 113: Another test that would be good including is verifying that `clearAriaLabel` is applied to the dismiss icon.

packages/wonder-blocks-search-field/src/components/search-field.js
```diff
@@ -0,0 +1,257 @@
+// @flow
+// A TextField with a search icon on its left side and X icon on its right side
```
jandrade, line 2: Could you please move this just before you declare the component (line # 104)? That way, Storybook Docs can get this description and add it to the Docs section:

https://5e1bf4b385e3fb0020b7073c-hxnnfisaue.chromatic.com/?path=/docs/searchfield--default

Here's a Cell example for reference: https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-cell/src/components/compact-cell.js#L10-L27
Output: https://khan.github.io/wonder-blocks/?path=/docs/cell-compactcell--default-compact-cell

packages/wonder-blocks-search-field/src/components/search-field.js
```diff
@@ -0,0 +1,257 @@
...
+                <IconButton
+                    icon={icons.dismiss}
+                    kind="tertiary"
+                    onClick={this.handleDismiss}
+                    style={styles.dismissIcon}
+                    aria-label={clearAriaLabel}
+                />
+            );
+        }
+        return null;
```
jandrade, line 138: Nit: This works perfectly, but you could also do the opposite... check for null inside the if, otherwise return the `IconButton` instance. Any option works, this just improves indentation a bit (so no worries if you keep it as it is).

packages/wonder-blocks-search-field/src/components/search-field.js
```diff
@@ -0,0 +1,257 @@
...
+        flex: 1,
+        background: "inherit",
+        border: "none",
+        outline: "none",
+        "::placeholder": {
+            color: Color.offBlack64,
+        },
+        width: "100%",
+        color: "inherit",
+        paddingLeft: 4,
```
jandrade, line 255: Could you please use `Spacing` here?

packages/wonder-blocks-search-field/src/components/search-field.stories.js
```diff
@@ -0,0 +1,34 @@
+// @flow
+import * as React from "react";
+
+import type {StoryComponentType} from "@storybook/react";
+
+import SearchField from "../index.js";
+
+export default {
+    title: "SearchField",
```
jandrade, line 9: If you include `component: SearchField` here, SB Docs can use that to include the Props table in the `Docs` tab.

For reference: https://5e1bf4b385e3fb0020b7073c-hxnnfisaue.chromatic.com/?path=/docs/cell-compactcell--default-compact-cell

NOTE: You might need to add another key...`argTypes` that is used to customize how each prop is rendered in the Props table (or might not).

jandrade (changes requested):
The implementation looks really solid!

Requesting some changes around:
- Testing (Kevin pointed out some of the possible solutions).
- Storybook: The stories file could be included in a separate location, and more importantly, there are some code comments and Storybook features that should be included to take advantage of the Docs feature.

packages/wonder-blocks-search-field/src/components/search-field.js
```diff
@@ -0,0 +1,257 @@
...
+    state: State = {
+        focused: false,
+    };
+
+    handleDismiss: () => void = () => {
+        const {onClick, onChange} = this.props;
+        // Empty the search text and focus the SearchField
+        onChange("");
+        if (onClick) {
+            onClick();
```
nishasy, line 120: I took the `handleDismiss` code directly from `search-text-input.js`, but I'm a bit confused about why it calls `onClick` in this function. I feel like `onClick` should be reserved for clicking into the text field? Should this be a different prop named `onClear` instead?

nishasy: I think `onClick` _is_ passed into the render function, so it's kind of weird that it's also here? What do yall think?

nishasy: The description for the prop says the following

```
/**
     * Handler that is triggered when this component is clicked. For example,
     * use this to adjust focus in parent component. This gets called when we
     * click the dismiss icon button within the SearchField.
     */
    onClick?: () => mixed,
    ```
    
Should `onClick` not be passed into the render function `View`? Should it only be used on the dismiss button? I feel like that's a bit unintuitive.

**Comment by kevinbarabash** [comment]
@nishasy we'd like start migrating wonder-blocks to using hooks.  Since the `SearchField` component doesn't make use of lifecycle methods it should be relatively simple to write as a functional component using hooks.  In a follow-up PR could you convert this component to be a functional component? 🙏

File: packages/wonder-blocks-search-field/src/components/search-field.js (hunk 10 above)
Comment (kevinbarabash) [reply packages/wonder-blocks-search-field/src/components/search-field.js:120]: Juan and I have responded in the corresponding Slack thread. 🙂

File: packages/wonder-blocks-search-field/src/components/search-field.stories.js (hunk 1 above)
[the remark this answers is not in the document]
Comment (nishasy) [reply packages/wonder-blocks-search-field/src/components/search-field.stories.js:1]: Do you think I need an `argTypes` file for search field?

File: packages/wonder-blocks-tooltip/src/util/__tests__/active-tracker.test.js [hunk 11]
```diff
@@ -1,5 +1,6 @@
 // @flow
-import ActiveTracker, {IActiveTrackerSubscriber} from "../active-tracker.js";
+import ActiveTracker from "../active-tracker.js";
+import type {IActiveTrackerSubscriber} from "../active-tracker.js";
```
nishasy, line 3: Added this because I was getting a flow error for it 🤷🏽‍♀️

packages/wonder-blocks-search-field/src/components/search-field.stories.js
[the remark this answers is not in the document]
jandrade: hmmm not completely sure if you need it, but you can try seeing the result in the page just with including `component: SearchField`. If the types are not good enough, then you'll need to add `argTypes`.

packages/wonder-blocks-search-field/src/components/search-field.stories.js
```diff
@@ -0,0 +1,34 @@
...
+
+    const handleChange = (newValue: string) => {
+        setValue(newValue);
+    };
+
+    const handleKeyDown = (event: SyntheticKeyboardEvent<HTMLInputElement>) => {
+        if (event.key === "Enter") {
+            event.currentTarget.blur();
+        }
+    };
```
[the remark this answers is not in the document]
nishasy: I'm not really sure what all the use cases for `SearchField` will be. I could see there being cases where maybe pressing enter causes a dropdown and maybe in that case it might make more sense not to blur?

This is one of those things where I _think_ maybe it makes more sense to leave it for now and then add it later if it becomes relevant.

packages/wonder-blocks-search-field/src/components/search-field.stories.js
```diff
@@ -0,0 +1,34 @@
...
+
+    return (
+        <SearchField
+            id="sf-1"
+            value={value}
+            placeholder="Placeholder"
+            onChange={handleChange}
+            onKeyDown={handleKeyDown}
```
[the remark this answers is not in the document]
nishasy: Do you know how to get multiple actions to work? I'm following the guid here:
https://github.com/storybookjs/storybook/blob/f09053cd8f28737e7b278ac80da6bb76e970ac63/addons/actions/ADVANCED.md

But I don't think it's working properly. I'm getting a ton of console warnings saying the following
```
Warning: This synthetic event is reused for performance reasons. If you're seeing this, you're accessing the property `defaultPrevented` on a released/nullified synthetic event. This is set to null. If you must keep the original synthetic event around, use event.persist(). See https://fb.me/react-event-pooling for more information.
```

And all the fields in the not-first event are null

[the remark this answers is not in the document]
kevinbarabash: I'm not sure what's going on.  I think the event handlers would look something like:
```
onChange={(e) => action("clicked")(e)}
onKeyDown={(e) => action("keydown")(e)}
```
I haven't tried this though, I'm going off the this code from button.stories.js:
```
            <Button
                href="/foo"
                style={styles.button}
                onClick={(e) => {
                    action("clicked")(e);
                    e.preventDefault();
                }}
            >
                This button prevents navigation.
            </Button>
```

packages/wonder-blocks-search-field/src/components/__docs__/search-field.stories.js
```diff
@@ -0,0 +1,117 @@
...
+
+    const handleChange = (newValue: string) => {
+        setValue(newValue);
+    };
+
+    const handleKeyDown = (event: SyntheticKeyboardEvent<HTMLInputElement>) => {
+        if (event.key === "Enter") {
+            event.currentTarget.blur();
+        }
+    };
```
kevinbarabash, line 91: Will this ever get called if the component is disabled?

kevinbarabash (approved):
LGTM.  Thanks for adding more stories and tests. 🎉

packages/wonder-blocks-search-field/src/components/__docs__/search-field.stories.js
```diff
@@ -0,0 +1,117 @@
...
+            disabled={true}
+        />
+    );
+};
+
+Disabled.parameters = {
+    docs: {
+        storyDescription:
+            "SearchField takes a `disabled` prop, which makes it unusable. Try to avoid using this if possible as it is bad for accessibility.",
+    },
```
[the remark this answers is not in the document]
nishasy: I think the guideline says that if a clickable item has to be disabled, it's better for it not to be on the screen at all. Its existence is confusing, especially if it never explains why it's disabled.

packages/wonder-blocks-search-field/src/components/search-field.stories.js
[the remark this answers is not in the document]
nishasy: I'm going to skip this change for now. Juan confirmed that it seems there’s an issue with the actions addon.

packages/wonder-blocks-search-field/src/components/__docs__/search-field.stories.js
nishasy: Great question! It should not get called if the component is disabled.

This also reminded me that we don't currently treat `TextField`/`SearchField` the same as `Clickable` when disabled, but we should. I'm gonna create a ticket for it.

packages/wonder-blocks-search-field/src/components/search-field.js
```diff
@@ -0,0 +1,215 @@
...
+ * const handleChange = (newValue: string) => {
+ *     setValue(newValue);
+ * };
+ *
+ * <SearchField
+ *     id="some-id"
+ *     value={value}
+ *     onChange={handleChange}
+ * />
+ * ```
```
jandrade, line 108: Awesome! thanks for this explanation and code snippet!

jandrade (approved):
Looks great! thanks for applying the changes 💯

nishasy:
The parent pull-request ([#1372](https://github.com/Khan/wonder-blocks/pull/1372)) has been merged into `main`, but this branch (`search-field-component`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #1387, opened 2022-03-25
FEI-3762.1: Copy i18n code from webapp to create wonder-blocks-i18n package

kevinbarabash:
## Summary:
This package is part of the Perseus decoupling work.  Once this package is published webapp will be updated to import this instead of its local shared-package/i18n.js.  When we go to extract Perseus, we'll be able to 'npm install' this new package in that repo.

This PR makes a number of changes from the original webapp code:
- Always use translators for "boxes" and "accents" (before, we only did this in dev)
- Replaces `RequestInfo.kaLocale` with `getLocale()`
- FakeTranslate is removed and we no longer parse the query string (our Fastly code is doing this for us to set the `X-KA-Locale` header)
- Removes handlebars functions
- Stops setting i18n globals for live-editor (we can do that in webapp or update live-editor to import `wonder-blocks-i18n`)

TODO:
- [x] write tests for 'getLocale()'

Issue: FEI-3762

## Test plan:
- use wallaby to run the test files
- [x] yarn test (to build and run all tests)


packages/wonder-blocks-i18n/docs.md
```diff
@@ -0,0 +1,7 @@
+Progress spinners to be used when indicating that some piece of content is loading. They come in three sizes (default, small, and xsmall) and also in light and dark varities. By default `CircularSpinner` should be used for most loading indicators.
+
+```js
+import {CircularSpinner} from "@khanacademy/wonder-blocks-progress-spinner";
+
+<CircularSpinner />
+```
```
kevinbarabash, line 7: This is copy/pasta.  I'm going to change this to some placeholder text and will tackle docs in a separate PR (and probably a separate JIRA task).

packages/wonder-blocks-i18n/src/components/i18n-inline-markup.jsx
```diff
@@ -0,0 +1,179 @@
+// @flow
+
+/**
+ * Render some HTML, using functions passed in as props to render tags.
```
kevinbarabash, line 4: Should be able to migrate all of this into a storybook MDX file when I work on docs.

packages/wonder-blocks-i18n/src/index.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+export {
+    _,
+    $_,
+    ngettext,
+    doNotTranslate,
+    doNotTranslateYet,
+    localeToFixed,
+    getDecimalSeparator,
```
kevinbarabash, line 9: I check to see that these were being used in webapp before exporting them.

kevinbarabash:
Since the coverage went down a bit, I'm going to see if I can add a few more test cases.

packages/wonder-blocks-i18n/src/functions/get-locale.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+const HEADERS_GLOBAL = "__HEADERS__";
```
jeresig, line 2: I'll admit that this scares me a bit. I know that this will work with our current system, but it definitely feels fragile. I'd love for us to have a way to work around this - maybe something we can set on initialization?

jeresig (approved):
Looking good to me! Will be nice to be able to re-use this in more places.

I think it's possible that we don't have to set i18n for CS stuff anymore, I think we do that in the import files themselves now.

packages/wonder-blocks-i18n/src/functions/i18n.js
```diff
@@ -0,0 +1,429 @@
...
+ * represent strings that may undergo revisions and should not be
+ * translated yet.
+ */
+export const doNotTranslateYet: _Overloads = doNotTranslate;
+
+/**
+ * Rounds num to X places, and uses the proper decimal seperator.
+ * But does *not* insert thousands separators.
+ */
+export const localeToFixed = function (num: number, places: number): string {
```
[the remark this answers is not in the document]
kevinbarabash: `localToFixed` is not, but `getDecimalSeparator` is.

packages/wonder-blocks-i18n/src/functions/i18n.js
```diff
@@ -0,0 +1,429 @@
...
+//
+// If it prints anything else, use the `plural` expression, which should
+// hopefully be usable in JavaScript.
+//
+// TODO(csilvers): auto-generate this list instead.
+//
+// NOTE(mdr): Disable Prettier for this object, because intl_js_test.py
+//     expects it to be in a particular format.
+//     prettier-ignore
+const allPluralForms: PluralFormsMap = {
```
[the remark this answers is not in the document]
kevinbarabash: I couldn't think of how to use snapshots for this so I wrote `test.each` tests for all the odd-ball languages.

packages/wonder-blocks-i18n/src/functions/i18n.js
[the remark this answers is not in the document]
kevinbarabash: Sounds good.  I'll move them into l10n.js as suggested.

packages/wonder-blocks-i18n/src/functions/get-locale.js
kevinbarabash: It's super fragile.  Calling some initialization function feels fragile too but maybe there's some way we can guarantee that it's being called before anything else by calling it one of the page building functions.

kevinbarabash:
I still need to add a changesets .md file.  I'll do that after lunch.

packages/wonder-blocks-i18n/src/functions/i18n-faketranslate.js
```diff
@@ -0,0 +1,171 @@
+// @flow
+import Accents from "./i18n-accents.js";
+import Boxes from "./i18n-boxes.js";
```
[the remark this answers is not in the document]
kevinbarabash: Agreed. 😞

PR #1390, opened 2022-03-29
WB-1267: Fix top alignment between accessory and content

jandrade:
## Summary:

There’s an extra spacing on the content container that doesn’t work well when we try to align the accessory to the top. We need to fix the layout to align both the accessories and the content using the same padding (applies for both Compact and Detail cells).

Issue: https://khanacademy.atlassian.net/browse/WB-1267

## Test plan:

Navigate to the Storybook instance and verify that the
?path=/story/cell-detailcell--detail-cell-with-custom-styles example is now
vertically aligned to the top.

BEFORE

Screen Shot 2022-03-29 at 10 21 08 AM

AFTER
fix

### DIFF
--- a/.changeset/curly-doors-draw.md
+++ b/.changeset/curly-doors-draw.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-cell": patch
+---
+
+Fix top vertical alignment between accessories and content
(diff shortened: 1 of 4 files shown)


jeresig (approved):

matthewmorgan (approved):
Thanks @jandrade !

PR #1392, opened 2022-03-29
Add getLocale()/setLocale(locale) methods to wonder-blocks-i18n

kevinbarabash:
## Summary:
Change how we get the current locale, but exporting a get/set methods for the locale.  This will allow mobile and webapp set the current locale for the rest of the functions exported by this package.

Issue: None

## Test plan:
- yarn jest wonder-blocks-i18n
- yarn flow


packages/wonder-blocks-i18n/src/index.js
```diff
@@ -9,3 +9,5 @@ export {
 } from "./functions/i18n.js";
 
 export {localeToFixed, getDecimalSeparator} from "./functions/l10n.js";
+
+export {getLocale, setLocale} from "./functions/locale.js";
```
jeresig, line 13: Whoops, think you forgot to include locale.js!

kevinbarabash: It's there now.

jeresig (approved):
Seems reasonable, thank you!

PR #1395, opened 2022-03-29
Export I18nInlineMarkup from wonder-blocks-i18n package

kevinbarabash:
## Summary:
Export I18nInlineMarkup React component

Issue: None
## Test plan:
- yarn start:storybook
- open localhost:6061, see that the stories render

Screen Shot 2022-03-29 at 4 20 56 PM


jeresig (approved):

packages/wonder-blocks-i18n/src/components/i18n-inline-markup.stories.js
```diff
@@ -0,0 +1,63 @@
+// @flow
+import * as React from "react";
+
+import {I18nInlineMarkup} from "./i18n-inline-markup.js";
+
+export default {
+    title: "Translations/I18nInlineMarkup",
+    parameters: {
```
jandrade, line 8: You might want to include `component: I18nInlineMarkup` here in case you want to expose the Props table.

jeresig (commented):
I just realized that the code coverage went way down - do we not have tests for this component?

kevinbarabash:
> I just realized that the code coverage went way down - do we not have tests for this component?

There are tests for the parser but not for the component itself. I'll add some tests for it.

kevinbarabash:
@jeresig the reason why the coverage went down is that when I copied the file over in a previous PR it still had a .jsx extension which jest ignores when computing coverage.  In this PR I changed the extension to .js so now it's picking up the file in the report.

jeresig:
@kevinbarabash Gotcha - so just confirming, the reason why it went down is that there are no tests for this code?

packages/wonder-blocks-i18n/src/components/i18n-inline-markup.stories.js
jandrade: Oh one more thing 🙏 ... could you please move this file inside the `__docs__` folder?

kevinbarabash:
> @kevinbarabash Gotcha - so just confirming, the reason why it went down is that there are no tests for this code?

That's right.  The drop should've happened when the file was first added, but was deferred until this PR because of the .jsx vs .js issue.

PR #1396, opened 2022-03-29
[fei4127.allowstorybookautogentitles] Allow auto-gen titles in storybook

somewhatabstract:
## Summary:
This makes the `title` field optional/nullable and adds a `getDefaultTitle` function so that adapters that cope with nullish titles can do their own thing, but those that can't (like if we had a React Sandbox adapter) can just use the default title as needed.

Issue: FEI-4127

## Test plan:
`yarn test`
`yarn flow`

### DIFF
--- a/packages/wonder-blocks-testing/src/fixtures/adapters/storybook.js
+++ b/packages/wonder-blocks-testing/src/fixtures/adapters/storybook.js
@@ -26,7 +26,7 @@ export type StorybookOptions = {|
 |};
 
 type DefaultExport = {|
-    title: string,
+    title?: ?string,
     ...StorybookOptions,
 |};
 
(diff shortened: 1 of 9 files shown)


packages/wonder-blocks-testing/src/fixtures/adapters/storybook.js
```diff
@@ -47,6 +47,9 @@ export const getAdapter: AdapterFactory<StorybookOptions, Exports<any>> = (
             {
                 title,
                 description: groupDescription,
+                // We don't use the default title in Storybook as storybook
+                // will generate titles for us if we pass a nullish title.
+                getDefaultTitle: _,
```
kevinbarabash, line 52: neat

jeresig (approved):
Looks good, thank you!

kevinbarabash (approved):
LGTM. 🎉

PR #1398, opened 2022-03-30
Don't bundle 'react' in dist bundle for wonder-blocks-i18n

kevinbarabash:
## Summary:
This will save some bytes and hopefully also fix some errors I was seeing in webapp tests around "key" prop missing when using some of the i18n functions.

Issue: none

## Test plan:
- yarn build:all
- check the dist bundles for wonder-blocks-i18n, grep for "Facebook", see no hits

### DIFF
--- a/packages/wonder-blocks-i18n/package.json
+++ b/packages/wonder-blocks-i18n/package.json
@@ -17,6 +17,9 @@
   "dependencies": {
     "@babel/runtime": "^7.16.3"
   },
+  "peerDependencies": {
+    "react": "16.14.0"
+  },
   "devDependencies": {
     "wb-dev-build-settings": "^0.3.0"
   }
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-i18n/package.json
```diff
@@ -17,6 +17,9 @@
   "dependencies": {
     "@babel/runtime": "^7.16.3"
   },
+  "peerDependencies": {
+    "react": "16.14.0"
```
jeresig, line 21: Do we need to do this for other WB packages too?

jeresig (approved):

packages/wonder-blocks-i18n/package.json
kevinbarabash: It's set on the rest of the packages.

PR #1404, opened 2022-04-04
Add forwardRef to SearchField

nishasy:
Add ref forwarding to `SearchField` so it can be used by wb dropdowns.

I used the same ref forwarding method as that used in [TextField](https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-form/src/components/text-field.js#L351) and [LabeledTextField](https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-form/src/components/labeled-text-field.js#L296).

Issue: WB-1117 https://khanacademy.atlassian.net/browse/WB-1117

## Test plan:
`yarn test packages/wonder-blocks-search-field/src/components/__tests__/search-field.test.js`


jeresig (approved):
Looks good to me, thank you!

packages/wonder-blocks-search-field/src/components/__tests__/search-field.test.js
```diff
@@ -285,4 +285,42 @@ describe("SearchField", () => {
...
+                value="some-value"
+                onChange={() => {}}
+                testId="search-field-test"
+                ref={ref}
+            />,
+        );
+
+        // Assert
+        expect(ref.current.value).toBe("some-value");
+    });
```
kevinbarabash, line 325: Neat test!  I wouldn't've thought to check for this.

packages/wonder-blocks-search-field/src/components/search-field.js
```diff
@@ -213,3 +222,17 @@ const styles = StyleSheet.create({
...
+    React.ElementConfig<typeof SearchField>,
+    WithForwardRef,
+>;
+
+const ForwardedSearchField: React.AbstractComponent<
+    ExportProps,
+    HTMLInputElement,
+> = React.forwardRef<ExportProps, HTMLInputElement>((props, ref) => (
+    <SearchField {...props} forwardedRef={ref} />
+));
```
kevinbarabash, line 236: We could just implement the component inside of the callback.  This would avoid the need to use `forwardedRef` and differentiate between `Props` and `ExportProps`.  I don't think we do this anywhere yet, check with @jandrade first and see what he thinks.

kevinbarabash (approved):
LGTM.  Thanks for adding support for ref forwarding.

packages/wonder-blocks-search-field/src/components/__tests__/search-field.test.js
nishasy: It was Juan's idea! 🙂

packages/wonder-blocks-search-field/src/components/__tests__/search-field.test.js
```diff
@@ -285,4 +285,42 @@ describe("SearchField", () => {
...
+                id="sf-id"
+                value="some-value"
+                onChange={() => {}}
+                testId="search-field-test"
+                ref={ref}
+            />,
+        );
+
+        // Assert
+        expect(ref.current.value).toBe("some-value");
```
nishasy, line 324: I'm actually getting a flow error here saying `Cannot get `ref.current.value` because property `value` is missing in null [1].`.

Would the test still test correctly if I change it to `expect(ref.current?.value).toBe("some-value");`? Or would it pass when it shouldn't in that case?

kevinbarabash: `expect(ref.current?.value).toBe("some-value");` should work just fine.

jandrade: In that case you can use `ref.current?.value` safely.

packages/wonder-blocks-search-field/src/components/search-field.js
jandrade: ohhh yeah good point! that should work fine with this component. I'm happy that it was refactored to a functional one b/c it makes easier this change.

packages/wonder-blocks-search-field/src/components/__tests__/search-field.test.js
nishasy: I added `ref.current?.value` but Wallaby shows that line as yellow now. In a component file, I know yellow means that the line is only partially covered. What does yellow mean in a test file? Do I need to add an if/else and auto-fail in the else case?

nishasy: The following is green on wallaby and passes flow:

```expect(ref.current && ref.current.value).toBe("some-value");```

Should I just keep it like this instead of using `?`?

Comment (jandrade) [reply packages/wonder-blocks-search-field/src/components/__tests__/search-field.test.js]: mmm I'm not sure what yellow means in the test context.... @kevinbarabash maybe do you have any thoughts on that?

File: packages/wonder-blocks-search-field/src/components/search-field.js [hunk 4]
```diff
@@ -107,78 +107,83 @@ type Props = {|
...
-        ): any);
-        currentField.focus();
-    };
-
-    const maybeRenderClearIconButton: () => React.Node = () => {
-        if (!value.length) {
-            return null;
-        }
+const SearchField: React.AbstractComponent<Props, HTMLInputElement> =
+    React.forwardRef<Props, HTMLInputElement>((props: Props, ref) => {
```
jandrade, line 111: 🎉 Way more cleaner! Thanks @kevinbarabash for the suggestion!

jandrade (approved):
Looks great! thanks for the improvements! ⏩

PR #1407, opened 2022-04-05
Fix perf issue in FakeTranslate

kevinbarabash:
## Summary:
When I moved i18n-faketranslate.js over from webapp, I tweaked the logic a bit so that we'd call the translator all the time, even in prod.  I made the mistake of also haveing ._translator fallback to the identity function if no translator could be found.  This had the unintended result of _parseAndTranslate() no longer short circuiting for actual locales.  This PR fixes the issue by dropping the fallback.

Issue: None

## Test plan:
- yarn test i18n-faketranslate

### DIFF
--- a/.changeset/chilly-panthers-act.md
+++ b/.changeset/chilly-panthers-act.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-i18n": patch
+---
+
+Pass-through strings if the current locale is not a pseudo-lang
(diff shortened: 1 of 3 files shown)


jeresig (approved):

kevinbarabash:
@jeresig I was thinking that we could make accents/boxes faster if we skip the DOM manipulation if there's no HTML tags in the input string.  That's an optimization for a different day.

jeresig:
@kevinbarabash Yeah, true! Although considering how rarely they're using, I think it's ok to not really prioritize that.

PR #1410, opened 2022-04-06
Cell: Fix rightAccessory style

jandrade:
## Summary:

Fixes `CellCore` to set the styles properly on the `rightAccessory` element. The previous styles where applied incorrectly.

With this fix, now we are applying the remaining space available to the content, so the accessories are correctly aligned to each side.

Issue: XXX-XXXX

## Test plan:

Navigate to http://localhost:6061/?path=/story/cell-detailcell--default-detail-cell&globals=outline:true

https://5e1bf4b385e3fb0020b7073c-nkzomwxcrj.chromatic.com/?path=/story/cell-detailcell--default-detail-cell

Verify that the cells render as expected.

BEFORE:
Screen Shot 2022-04-06 at 10 08 58 AM
Screen Shot 2022-04-06 at 10 09 21 AM

AFTER:
Screen Shot 2022-04-06 at 10 09 46 AM

### DIFF
--- a/.changeset/neat-cobras-sing.md
+++ b/.changeset/neat-cobras-sing.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-cell": patch
+---
+
+Fixes `CellCore` to set the styles properly (removed `marginLeft: auto` from accessory and expanded the content wrapper).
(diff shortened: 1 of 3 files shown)


kevinbarabash:
@jandrade I noticed that the Chromatic snapshots didn't change.  Should we maybe add a snapshot to prevent regression of this in the future?

jandrade:
> @jandrade I noticed that the Chromatic snapshots didn't change. Should we maybe add a snapshot to prevent regression of this in the future?

Good question! I think that no changes is a good signal in this case, because the change is not introducing any regression.

About the specific bug... it happened in 321 (@matthewmorgan shared it with me), and the issue happens when we have an external wrapper where `text-align: left` is set and the contents are short.

e.g.
```
<View style={{ textAlign: "left" }}>
  <DetailCell ... />
</View>
```

Do you think it would be helpful to add this story to cover this particular case?

jandrade:
@kevinbarabash I ended up modifying one of the existing stories to try passing custom styles to see that the contents are horizontally centered.

PR #1418, opened 2022-04-13
Update parseSimpleHTML to warn instead of throw when the use of I18nInlineMarkup isn't necessary

kevinbarabash:
## Summary:
The translation isn't correct, but it's better to show the translation than to throw an exception which results in us showing an error screen.

Issue: None

## Test plan:
- yarn jest parse-simple-html


jeresig (approved):
Thank you!

PR #1420, opened 2022-04-15
Add UniqueIDProvider to SearchField

nishasy:
## Summary:
- Wrap SearchField elements in `UniqueIDProvider`so that
  users don't have to pass in an ID every time they
  use the component.
- Instead of using IDs to find the `<input>` element in
  the DOM on clear, pass in a ref to the `<input>`.

Issue: none

## Test plan:
`yarn jest packages/wonder-blocks-search-field/src/components/__tests__/search-field.test.js`

Storybook:
- http://localhost:6061/?path=/docs/search-field-searchfield--default
- http://localhost:6061/?path=/docs/dropdown-singleselect--default-single-select-opened

### DIFF
--- a/packages/wonder-blocks-search-field/src/components/__tests__/search-field.test.js
+++ b/packages/wonder-blocks-search-field/src/components/__tests__/search-field.test.js
@@ -334,6 +334,7 @@ describe("SearchField", () => {
     });
 
     test("uses a unique ID if one is not provided", () => {
+        // Arrange
         render(
             <SearchField
                 testId="search-field-test"
@@ -342,8 +343,10 @@ describe("SearchField", () => {
             />,
         );
 
+        // Act
         const searchField = screen.getByTestId("search-field-test");
 
+        // Assert
         expect(searchField.getAttribute("id")).toMatch(/^uid-.*-field$/);
     });
 });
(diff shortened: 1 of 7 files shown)


packages/wonder-blocks-search-field/src/components/__tests__/search-field.test.js
```diff
@@ -321,6 +310,40 @@ describe("SearchField", () => {
...
+        waitFor(() => {
+            expect(ref.current?.value).toBe("some-value");
+        });
+    });
+
+    test("uses the passed in ID if one is provided", () => {
+        // Arrange
+        render(
+            <SearchField
+                id={"some-random-id"}
```
jeresig, line 322:
```suggestion
                id="some-random-id"
```

jeresig (approved):
Nice! This seems like a solid improvement, thank you! Don't forget to add in the Changeset docs.

packages/wonder-blocks-search-field/src/components/__tests__/search-field.test.js
```diff
@@ -321,6 +310,40 @@ describe("SearchField", () => {
...
+
+        // Act
+        const searchField = screen.getByTestId("search-field-test");
+
+        // Assert
+        expect(searchField).toHaveAttribute("id", "some-random-id-field");
+    });
+
+    test("uses a unique ID if one is not provided", () => {
+        render(
```
jandrade, line 337: Nice test! One minor thing... could you please add `// Arrange, Act, Assert` comments on this test?

PR #1421, opened 2022-04-15
Add Breadcrumbs to Storybook

nishasy:
## Summary:
Add Breadcrumbs to storybook.
- Create a default Breadcrumbs story
- Add BreadcrumbsItem as a subcomponent, found at the bottom
  of the Docs tab (not as a separate component because 
  it's pretty bare bones)
- Add an Accessibility section

Issue: https://khanacademy.atlassian.net/browse/WB-1130

Screen Shot 2022-04-14 at 10 23 25 PM
Screen Shot 2022-04-14 at 10 23 15 PM
Screen Shot 2022-04-14 at 10 22 31 PM
Screen Shot 2022-04-14 at 10 22 40 PM

## Test plan:
- Find the default breadcrumbs here: http://localhost:6061/?path=/docs/breadcrumbs-breadcrumbs--default-breadcrumbs
- Make sure the explanation makes sense
- Make sure the code segments display correctly
- Make sure the control interactions are correct
- Go to the Docs tab and confirm that there is
  a BreadcrumbsItem tab at the bottom that lists its props
- Confirm that there is an Accessibility section


packages/wonder-blocks-breadcrumbs/src/components/breadcrumbs.js
```diff
@@ -36,6 +36,30 @@ const StyledList = addStyle("ol");
...
+ *         <Link href="">Course</Link>
+ *     </BreadcrumbsItem>
+ *     <BreadcrumbsItem>
+ *         <Link href="">Unit</Link>
+ *     </BreadcrumbsItem>
+ *     <BreadcrumbsItem>
+ *         <Link href="">Lesson</Link>
+ *     </BreadcrumbsItem>
+ * </Breadcrumbs>
+ * ```
```
jandrade, line 62: Ohhh I see that the bundle file size was increased ~145b due to these new comments. I still think this shouldn't count as a version bump as we are not introducing any changes to the code itself.

jandrade (approved):
Looks great! only a couple of nits. thanks for also including the accessibility section.

packages/wonder-blocks-breadcrumbs/src/components/breadcrumbs.js
jandrade: @jeresig ☝️ This is the reason for the file size increase.

jeresig: But why do the comments increase the file size - don't we strip the comments out when generating the bundles? I thought the bundles were minified ahead of time. At the very least, I think we should do that if we're comparing the output to get a more accurate view of what changes are being made and how that might impact users. Out of scope of this PR, but would be good to look in to this in the future!

jandrade: It turns out we are not stripping the comments out :|. I've created this ticket to track that change: https://khanacademy.atlassian.net/browse/WB-1299

jeresig: Thank you!

PR #1423, opened 2022-04-20
WB-1299: Strip comments out from ES bundles

jandrade:
## Summary:

Strip comments out from ES bundles (rollup).

Issue: WB-1299

## Test plan:

Verify that the bundle sizes check shows some improvements on each bundled package.

### DIFF
--- a/.changeset/tame-ghosts-poke.md
+++ b/.changeset/tame-ghosts-poke.md
@@ -1,0 +1,5 @@
+---
+"wb-dev-build-settings": minor
+---
+
+Strip comments out from ES generated bundles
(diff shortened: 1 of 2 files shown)


jeresig (approved):
Wow! Yeah, that's a solid improvement. This probably won't matter much for Webapp as we already compress things there, but it's probably good to do this anyway.

kevinbarabash (approved):
LGTM.  The size improvements are awesome!  Were the comments not being stripped out by webpack in webapp?

jeresig:
@kevinbarabash Webapp is stripping out the comments, so this is purely a Wonder Blocks issue.

PR #1425, opened 2022-04-20
WB-1131.2: Migrate `wonder-blocks-clickable` to Storybook - ClickableBehavior

jandrade:
## Summary:

- Add stories for the `ClickableBehavior` component.
- Reuse from of the argtypes defined for `Clickable`.
- Modify JSDocs to improve the way the comments are displayed in Storybook docs.

NOTE: This component doesn't include any accessibility notes as the recommended
approach is to use `Clickable` for an improved and more robust experience.

Issue: WB-1131

## Test plan:

1. Navigate to the newly created `ClickableBehavior` section in storybook.
2. Verify that there's a new story explaining how to use that component.
3. Verify that the props table is displayed correctly.

https://5e1bf4b385e3fb0020b7073c-wjhmfkabvs.chromatic.com/?path=/story/clickable-clickablebehavior--default
Screen Shot 2022-04-20 at 12 57 09 PM
Screen Shot 2022-04-20 at 12 57 22 PM

### DIFF
--- a/packages/wonder-blocks-clickable/src/components/__docs__/clickable-behavior.argtypes.js
+++ b/packages/wonder-blocks-clickable/src/components/__docs__/clickable-behavior.argtypes.js
@@ -3,10 +3,7 @@ import clickableArgtypes from "./clickable.argtypes.js";
 export default {
     children: {
         description:
-            "A function that returns the a React `Element`.  The function is passed an object with three boolean properties: hovered, focused, and pressed, and a `childrenProps` argument that contains all the event handlers that should be passed to the React `Element` itself.",
-        control: {
-            type: "text",
-        },
+            "A function that returns the a React `Element`. The function is passed an object with three boolean properties: hovered, focused, and pressed, and a `childrenProps` argument that contains all the event handlers that should be passed to the React `Element` itself.",
         type: {
             required: true,
         },
--- a/packages/wonder-blocks-clickable/src/components/__docs__/clickable-behavior.argtypes.js
+++ b/packages/wonder-blocks-clickable/src/components/__docs__/clickable-behavior.argtypes.js
@@ -1,0 +1,58 @@
+import clickableArgtypes from "./clickable.argtypes.js";
+
+export default {
+    children: {
+        description:
+            "A function that returns the a React `Element`.  The function is passed an object with three boolean properties: hovered, focused, and pressed, and a `childrenProps` argument that contains all the event handlers that should be passed to the React `Element` itself.",
+        control: {
+            type: "text",
+        },
+        type: {
+            required: true,
+        },
+        table: {
+            type: {
+                summary:
+                    "(state: ClickableState, childrenProps: ChildrenProps) => React.Node",
+            },
+        },
+    },
+    /**
+     * States
+     */
+    disabled: {
+        ...clickableArgtypes.disabled,
+        description:
+            "Whether the component is disabled.\n\n" +
+            "If the component is disabled, this component will return handlers that do nothing.",
+    },
+    /**
+     * Events
+     */
+    onClick: {
+        ...clickableArgtypes.onClick,
+        description:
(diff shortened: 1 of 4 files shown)


jeresig (approved):
Looks really good to me, thank you!

packages/wonder-blocks-clickable/src/components/__docs__/clickable-behavior.argtypes.js
```diff
@@ -0,0 +1,58 @@
+import clickableArgtypes from "./clickable.argtypes.js";
+
+export default {
+    children: {
+        description:
+            "A function that returns the a React `Element`.  The function is passed an object with three boolean properties: hovered, focused, and pressed, and a `childrenProps` argument that contains all the event handlers that should be passed to the React `Element` itself.",
```
nishasy, line 6:
```suggestion
            "A function that returns the a React `Element`. The function is passed an object with three boolean properties: hovered, focused, and pressed, and a `childrenProps` argument that contains all the event handlers that should be passed to the React `Element` itself.",
```

packages/wonder-blocks-clickable/src/components/__docs__/clickable-behavior.argtypes.js
```diff
@@ -0,0 +1,58 @@
...
+        control: {
+            type: "text",
```
nishasy, line 8: Is `"text"` usually the type we want for `children`'s control? I think storybook usually default this to a JSON object, and I've been keeping it like that.

packages/wonder-blocks-clickable/src/components/__docs__/clickable-behavior.stories.js
```diff
@@ -0,0 +1,92 @@
...
+    chromatic: {
+        // we don't need screenshots because this story only displays the
+        // resting/default state.
+        disableSnapshot: true,
+    },
+};
+
+const styles = StyleSheet.create({
+    clickable: {
+        cursor: "pointer",
```
nishasy, line 77: It looks like the cursor remains `pointer` when using the controls to set this story in the disabled state.
Wha do you think about adding a `disabled` style and adding `disabled && styles.disabled` in the `View`'s style prop? 

Or do you think it would be better to leave that out to demonstrate the inherent behavior of `ClickableBehavior`? In that case, I'd argue that we shouldn't add the `cursor: "pointer"` style either.

nishasy (approved):
Nice! A couple of small questions, but overall looks really good!

packages/wonder-blocks-clickable/src/components/__docs__/clickable-behavior.argtypes.js
jandrade: Good question.. Storybook has this existing issue that is that is not able to parse JSX in controls, so we have to use a different control for that. The issue with `json` is that it's not possible to modify it, and `string` is a bit more configurable in that regard.

Here's an extract from their docs:
> Complex values such as JSX elements cannot be serialized to the manager (e.g., the Controls addon) or synced with the URL. Arg values can be "mapped" from a simple string to a complex type using the mapping property in argTypes to work around this limitation. It works in any arg but makes the most sense when used with the select control type.

packages/wonder-blocks-clickable/src/components/__docs__/clickable-behavior.stories.js
jandrade: Good point, I agree that the latter is better, so I'm going to remove that cursor from the example.

packages/wonder-blocks-clickable/src/components/__docs__/clickable-behavior.argtypes.js
jandrade: Giving a second thought, I think that it doesn't matter which option we use, so I'll change it to whatever SB sets as default, that is `json` as you suggested :).

PR #1426, opened 2022-04-20
Use SearchField inside wonder blocks dropdown

nishasy:
## Summary:
We want to replace the current field in wb dropdown with
the newly created wb SearchField. We do this by replacing
the render content of search-text-input.js with a
`<SearchField />`. We do this instead of replacing all
of `<SearchTextInput />` entirely due to the static `isClassOf`
function and the static `_IS_SEARCH_TEXT_INPUT` variable.

- Use `SearchField` in `SearchTextInput`
- Update the tests for `SearchTextInput to RTL
- Add a test for the focus shifting to the input on clear.

Issue: WB-1281

Screen Shot 2022-04-20 at 12 47 30 PM

## Test plan:
`yarn jest packages/wonder-blocks-dropdown/src/components/__tests__/search-text-input.test.js`

Storybook
- Go to Dropdown/SingleSelect/Default Single Select Opened
- Click on the Filter input box
- Check there is no clear button when it's empty
- Make sure typing in it works
- Make sure clear button appears when there is text
- Use the keyboard to press the clear button
- Confirm that the focus shifts back to the input


packages/wonder-blocks-dropdown/src/components/__tests__/search-text-input.test.js
```diff
@@ -1,145 +1,212 @@
...
-import {shallow} from "enzyme";
-import "jest-enzyme";
+import {render, screen} from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
 import SearchTextInput from "../search-text-input.js";
 
 describe("SearchTextInput", () => {
-    it("text input container should be focused when focusing on the input", () => {
+    test("text input container should be focused when focusing on the input", () => {
```
jandrade, line 9: It's great that you migrated most of these tests to RTL!

Nit: One thing to note is that some of these tests might be redundant b/c some of these cases are already tested in `search-field`, but I guess it doesn't hurt to increase testing overall!

packages/wonder-blocks-dropdown/src/components/__tests__/search-text-input.test.js
```diff
@@ -1,145 +1,212 @@
...
+        // Act
+        userEvent.paste(input, "value");
+
+        // Assert
+        const clearIconButton = screen.queryByRole("button");
+        expect(clearIconButton).toBeInTheDocument();
+    });
 
-        const input = wrapper.find(`[data-test-id="search-text-input"]`);
+    test("search should be cleared if the clear icon is clicked", () => {
```
jandrade, line 96: You could get rid of this test b/c it's already being tested in `search-field`.

packages/wonder-blocks-dropdown/src/components/__tests__/search-text-input.test.js
```diff
@@ -1,145 +1,212 @@
...
-        });
+        userEvent.click(clearIconButton);
 
         // Assert
-        expect(wrapper.find("IconButton")).toExist();
+        expect(input).toHaveValue("");
     });
 
-    it("search should be dismissed if the close icon is clicked", () => {
+    test("focus should return to the input element after clear button is clicked", () => {
```
jandrade, line 121: Same here.

packages/wonder-blocks-dropdown/src/components/__tests__/search-text-input.test.js
```diff
@@ -1,145 +1,212 @@
...
-        const wrapper = shallow(
+        // Act
+        const clearIconButton = screen.queryByRole("button");
+        userEvent.click(clearIconButton);
+
+        // Assert
+        expect(input).toHaveFocus();
+    });
+
+    test("placeholder should be updated by the parent component", () => {
```
jandrade, line 147: These `rerender` tests are neat!

jandrade (approved):
Looks great! just a couple of comments that I leave for your consideration.

One thing important that is missing is to include the changeset describing these changes. This can probably be treated  as a `patch` version.

packages/wonder-blocks-dropdown/src/components/__tests__/search-text-input.test.js
nishasy: I added this test again in this file on purpose, because it initially wasn't working in dropdown but it wasn't showing up as a test failure! If it's okay, I think I'd prefer to keep it.

jandrade: Sounds good :)

PR #1428, opened 2022-04-20
[fei4362.1.requestid] Add getGqlRequestId

somewhatabstract:
## Summary:
This takes the `getGqlRequestId` function that was created for webapp and makes it suitable for any Wonder Blocks Data-compatible GraphQL request.

Once this is released, we can replace webapp's version with this one. This may also pave the way for a `useGqlQuery` hook to exist within Wonder Blocks Data that implements the bulk of what the webapp equivalent currently does.

Issue: FEI-4362

## Test plan:
`yarn test`
`yarn storybook`
`yarn flow`

### DIFF
--- a/packages/wonder-blocks-data/src/util/get-gql-request-id.js
+++ b/packages/wonder-blocks-data/src/util/get-gql-request-id.js
@@ -21,24 +21,30 @@ export const getGqlRequestId = <TData, TVariables: {...}>(
     const parts = [];
 
     // First, we push the context values.
-    const sortedContext = Object.keys(context)
-        .sort()
-        .map((key) => `${key}=${context[key]}`)
-        .join(";");
-    parts.push(sortedContext);
+    const sortableContext = new URLSearchParams(context);
+    // $FlowIgnore[prop-missing] Flow has incomplete support for URLSearchParams
+    sortableContext.sort();
+    parts.push(sortableContext.toString());
 
     // Now we add the operation identifier.
     parts.push(operation.id);
 
     // Finally, if we have variables, we add those too.
     if (variables != null) {
-        // Turn the variables into a string of the form,
-        //     "key1=;key2=value"
-        const sortedVariables = Object.keys(variables)
-            .sort()
-            .map((key) => `${key}=${toString(variables[key])}`)
-            .join(";");
-        parts.push(sortedVariables);
+        // We need to turn each variable into a string.
+        const stringifiedVariables = Object.keys(variables).reduce(
+            (acc, key) => {
+                acc[key] = toString(variables[key]);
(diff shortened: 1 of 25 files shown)


packages/wonder-blocks-data/src/util/get-gql-request-id.js
```diff
@@ -0,0 +1,44 @@
...
+    // We add all the bits for this into an array and then join them with
+    // a chosen separator.
+    const parts = [];
+
+    // First, we push the context values.
+    const sortedContext = Object.keys(context)
+        .sort()
+        .map((key) => `${key}=${context[key]}`)
+        .join(";");
+    parts.push(sortedContext);
```
somewhatabstract, line 28: Differs from webapp since:

1. We don't know what the context may have (in webapp, we know what keys we have)
2. Includes the key name in the output; not just the context value

packages/wonder-blocks-data/src/util/get-gql-request-id.js
```diff
@@ -0,0 +1,44 @@
...
+    // Now we add the operation identifier.
+    parts.push(operation.id);
+
+    // Finally, if we have variables, we add those too.
+    if (variables != null) {
+        // Turn the variables into a string of the form,
+        //     "key1=;key2=value"
+        const sortedVariables = Object.keys(variables)
+            .sort()
+            .map((key) => `${key}=${toString(variables[key])}`)
```
somewhatabstract, line 39: Differs from webapp in that we don't quote strings. The `toString` helper just returns the value if it is already a string.

packages/wonder-blocks-data/src/util/get-gql-request-id.js
```diff
@@ -0,0 +1,44 @@
...
+    operation: GqlOperation<TData, TVariables>,
+    variables: ?TVariables,
+    context: GqlContext,
+): string => {
+    // We add all the bits for this into an array and then join them with
+    // a chosen separator.
+    const parts = [];
+
+    // First, we push the context values.
+    const sortedContext = Object.keys(context)
```
jeresig, line 24: I think you could use https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams to simplify this logic a bit. I think calling `new URLSearchParams(context)` would do it? (and same below)

jeresig (approved):
Great, thank you!

packages/wonder-blocks-data/src/util/get-gql-request-id.js
somewhatabstract: Oh it does. It's just `sort()` doesn't return anything. Cool. :)

PR #1429, opened 2022-04-20
WB-1301: Restrict the dropdown menu height

jandrade:
## Summary:

- Add `maxHeight` to the dropdown menu container (listbox) to avoid overflow
issues when the options list is longer than the screen height.
- Created a new helper to abstract out calculating the dropdown menu height, so it can be used by `DropdownCore` and the virtualized version.
- Added unit tests to verify that the styles are applied correctly.

Issue: WB-1301

## Test plan:

Navigate to any of the Dropdown stories and verify that after opening the
dropdown, you can see a maximum of 9 items in the screen.
Screen Shot 2022-04-20 at 6 00 35 PM

Navigate to the `BirthdayPicker` stories and verify that all the dropdowns are
now correctly displayed.

BEFORE
Screen Shot 2022-04-20 at 1 04 30 PM

AFTER
Screen Shot 2022-04-20 at 1 08 16 PM

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/dropdown-core.js
+++ b/packages/wonder-blocks-dropdown/src/components/dropdown-core.js
@@ -25,6 +25,10 @@ import SearchTextInput from "./search-text-input.js";
 import {defaultLabels, keyCodes, searchInputStyle} from "../util/constants.js";
 import type {DropdownItem} from "../util/types.js";
 import DropdownPopper from "./dropdown-popper.js";
+import {
+    generateDropdownMenuStyles,
+    getDropdownMenuHeight,
+} from "../util/dropdown-menu-styles.js";
 
 /**
  * The number of options to apply the virtualized list to.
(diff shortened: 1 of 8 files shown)


packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -772,7 +784,11 @@ class DropdownCore extends React.Component<Props, State> {
                     styles.dropdown,
                     light && styles.light,
                     isReferenceHidden && styles.hidden,
-                    {minWidth: minDropdownWidth},
```
jandrade, line 775: I'm not sure if I found an issue with how we merge styles in `wonder-blocks-core`.

Turns out that in this scenario:

```
<SingleSelect dropdownStyle={{ minWidth: 100 }} ...>
```

`minWidth=100` was overridden by `DropdownCore` internally because using the object in line #775 takes more precedence because WB core takes this as an inline style so it appends it at the last position of the array styles.

I'm not sure if this is expected or if it's some issue that we haven't noticed yet.

jeresig (approved):
Looks reasonable to me!

PR #1431, opened 2022-04-21
Add Checkbox to Storybook

nishasy:
## Summary:
Add Checkbox Storybook documentation.
All of the documentation is taken directly from styleguidist.

Issue: https://khanacademy.atlassian.net/browse/WB-1289

Screen Shot 2022-04-29 at 7 00 18 PM
Screen Shot 2022-04-29 at 7 00 22 PM
Screen Shot 2022-04-29 at 6 58 42 PM
Screen Shot 2022-04-29 at 6 58 46 PM

## Test plan:
- Go to http://localhost:6061/?path=/docs/form-checkbox--checkbox-default
- Compare with the docs in https://wonder-blocks.netlify.app/#checkbox
- Make sure that the controls work for the default story.


nishasy:
Oops, just realized I forgot to add the Accessibility section!

jeresig (approved):
I reviewed this as khan-actions-bot, sorry about that!

packages/wonder-blocks-form/src/components/__docs__/checkbox.stories.js
```diff
@@ -0,0 +1,156 @@
+// @flow
```
jandrade, line 1: One general suggestion that came from @megaroeny in this ticket: https://khanacademy.atlassian.net/browse/WB-1278

> Remove redundant button(s) words
> Basic Buttons → Basic or combine with Default
> Buttons with Colors → with Colors
> Buttons with Size → Sizes

What do you think about this?

(Sorry that I didn't mention this before, I just started using this pattern with `Clickable`).

packages/wonder-blocks-form/src/components/__docs__/checkbox.stories.js
```diff
@@ -0,0 +1,156 @@
...
+export const CheckboxDefault: StoryComponentType = (args) => (
+    <Checkbox {...args} />
+);
+
+CheckboxDefault.args = {
+    checked: false,
+    onChange: () => {},
+};
+
+export const CheckboxBasic: StoryComponentType = () => {
```
jandrade, line 33: I think it would make a bit more sense to call this story `Controlled`, b/c with this case we are using the Checkbox as a controlled component. wdyt?

packages/wonder-blocks-form/src/components/__docs__/checkbox.stories.js
```diff
@@ -0,0 +1,156 @@
...
+    const [checked, setChecked] = React.useState(false);
+    return <Checkbox checked={checked} onChange={setChecked} />;
+};
+
+CheckboxBasic.parameters = {
+    docs: {
```
jandrade, line 39: Probably worth removing this story from Chromatic snapshots as we are trying to illustrate here that the `state` works fine.

packages/wonder-blocks-form/src/components/__docs__/checkbox.stories.js
```diff
@@ -0,0 +1,156 @@
...
+        storyDescription:
+            "Use state to keep track of whether the checkbox is checked or not",
+    },
+};
+
+export const CheckboxVariations: StoryComponentType = () => (
```
jandrade, line 45: Nit: oh maybe we could use the term `Variants` here. This is a popular term in Design Systems and Figma as well, so we can have some consistency between Design and Eng :) (cc @megaroeny in case you have any thoughts).

https://www.figma.com/best-practices/creating-and-organizing-variants/

packages/wonder-blocks-form/src/components/__docs__/checkbox.stories.js
```diff
@@ -0,0 +1,156 @@
...
+    title: "Form / Checkbox",
+    component: Checkbox,
+    parameters: {
+        componentSubtitle: ((
+            <ComponentInfo name={name} version={version} />
+        ): any),
+    },
+};
+
+export const CheckboxDefault: StoryComponentType = (args) => (
```
jandrade, line 24: I think we can also skip this Story from Chromatic snapshots as this is being also tested in the `Variations / Variants` example.

packages/wonder-blocks-form/src/components/__docs__/checkbox-accessibility.stories.mdx
```diff
@@ -0,0 +1,137 @@
...
+        },
+        viewMode: "docs",
+        chromatic: {
+            // Disables chromatic testing for these stories.
+            disableSnapshot: true,
+        },
+    }}
+/>
+
+export const ErrorTemplate = (args) => {
```
jandrade, line 23: Ah this makes a lot of sense! thanks for adding the Stories in a separate place so the MDX docs are easier to read.

packages/wonder-blocks-form/src/components/__docs__/checkbox-accessibility.stories.mdx
```diff
@@ -0,0 +1,137 @@
...
+See the Error section for information about `aria-invalid` and
+`aria-required`.
+
+### Error state
+
+The Wonder Blocks `Checkbox` component takes an `error` boolean prop. Setting
+this prop to true will set `aria-invalid` to true, and the color of the
+checkbox to red.
+
+When a form input is invalid, the dev should provide a reason for why
```
jandrade, line 83: What do you think about changing the term `dev` for something more generic (user or something like that)? I know this is mainly intended for devs 😅, but maybe we can leave this more open to other audiences :)

packages/wonder-blocks-form/src/components/__docs__/checkbox-accessibility.stories.mdx
```diff
@@ -0,0 +1,137 @@
...
+        }}
+    >
+        {ErrorTemplate.bind({})}
+    </Story>
+</Canvas>
+
+### Disabled state
+
+The Wonder Blocks `Checkbox` compoenent takes a `disabled` boolean prop.
+This sets the underlying `input` element's `disabled` prop to true.
```
jandrade, line 111:
```suggestion
This sets the underlying `input` element's `disabled` prop to `true`.
```

jandrade (commented):
This is looking really great! and the accessibility section looks amazing :) thanks for putting so much care on this!

I'm going to hold off reviewing the Chromatic snapshots for now as I'd like to get your thoughts on some comments I added related to the stories. Let me know what you think, Great work!

jandrade:
Ah one more thing that I noticed that would be nice to include.

Could you please include the `Usage` section in the JS docs, like the one defined here: 

https://5e1bf4b385e3fb0020b7073c-vzutziqmpy.chromatic.com/?path=/docs/birthdaypicker--birthday-picker-default

https://github.com/Khan/wonder-blocks/blob/5536d12ecdee792fc9eba82bfadd984dd4d1f0a9/packages/wonder-blocks-birthday-picker/src/components/birthday-picker.js#L122

megaroeny:
Wow, just saw the SB page and agreed with @jandrade - thanks for putting so much detail into this! A couple of things of feedback on my end:

1. The very first checkbox at the top of the Docs page, isn't allowing me to check it. This intentional?
2. If you need ideas on a good description for a component, check out https://component.gallery/. If you search for _checkbox_, you'll land [on this page](https://component.gallery/components/checkbox/) which has a pretty nice description! I like to also look at Material Design and Apple's HIG descriptions too.
3. Agreed with @jeresig's comment above - Let's simplify our labels so they're easier to glance/scan at, and remove redundant works. In this case, we could remove the first _Checkbox_ word on all the nav items in this case. Maybe for _Checkbox Label_, it could be something like _Label for checkbox_. I think _Label_ would be too generic maybe, or maybe that's fine. Curious what others think.

Awesome work, thanks again!

packages/wonder-blocks-form/src/components/__docs__/checkbox.stories.js
nishasy: Interesting! I'll go ahead and update my titles to match this.

I'm not personally bothered by the redundancy. I actually think it could be helpful if you have a lot of the Storybook directories open so it's clearer which directory you're under.

nishasy:
@megaroeny Yes, it's expected that the first checkbox can't be clicked. The the default default, like without a `checked` or `onChange` prop passed in, so it has no functionality. If you want to toggle the `checked` value, you can do it from the control panel!

packages/wonder-blocks-form/src/components/__docs__/checkbox.stories.js
jandrade: That's a good point. I took a couple of screenshots with multiple subdirs open and also using the filter option, and I think the short versions might have sense. Another point is that the stories in the `Docs` page look better with the shorten version. Well it's my personal opinion :), so I'm not sure what y'all think.

Screen Shot 2022-05-05 at 5 19 59 PM

Screen Shot 2022-05-05 at 5 21 43 PM

I'm fine with either option. We'll see how folks react to the docs.

jandrade (approved):
Looks really awesome! thanks @nishasy for the detailed docs and examples 🚢

PR #1433, opened 2022-04-21
[fei4413.2.usingfetchpolicies] Implement fetch policy in useCachedEffect

somewhatabstract:
## Summary:
This reworks `useCachedEffect` to add fetch policy and refetch support.

This was a heavier lift than I originally hoped for, however, it works. I think things will get a little nicer when the RequestFulfillment properly supports aborting requests, but that's a while away (see FEI-4276) as we need to determine how to exactly do that. For now, I've added basic "aborting" to the RequestFulfillment that we'll have to expand later.

Issue: FEI-4413

## Test plan:
`yarn test`

### DIFF
--- a/packages/wonder-blocks-data/src/__docs__/exports.use-cached-effect.stories.mdx
+++ b/packages/wonder-blocks-data/src/__docs__/exports.use-cached-effect.stories.mdx
@@ -19,9 +19,9 @@ function useCachedEffect<TData: ValidCacheData>(
 ): [Result<TData>, () => void];
 ```
 
-This hook invokes the given handler and caches the result using the [`useSharedCache`](/docs/data-exports-usesharedcache--page) hook. The `requestId` is used to both share identify inflight requests that can be shared, and to identify the cached value to use.
+This hook invokes the given handler and caches the result using the [`useSharedCache`](/docs/data-exports-usesharedcache--page) hook. The `requestId` is used to both identify inflight requests that can be shared, and to identify the cached value to use.
 
-The hook returns an array containing the current state of the requested data, and a function that can be used to `refetch` the data on demand. Note that calling `refetch` while an inflight request is in progress for the given `requestId` will be a no-op.
+The hook returns an array containing the current state of the request, and a function that can be used to `refetch` that request on demand. Calling `refetch` while an inflight request is in progress for the given `requestId` will be a no-op.
 
 The behavior of the hook can be modified with the options.
 
--- a/packages/wonder-blocks-data/src/__docs__/exports.use-cached-effect.stories.mdx
+++ b/packages/wonder-blocks-data/src/__docs__/exports.use-cached-effect.stories.mdx
@@ -16,13 +16,14 @@ function useCachedEffect<TData: ValidCacheData>(
     requestId: string,
     handler: () => Promise<TData>,
     options?: CachedEffectOptions<TData>,
-): Result<TData>;
+): [Result<TData>, () => void];
 ```
 
-This hook is a wrapper around `useEffect`. Internally it uses the [`useSharedCache`](/docs/data-exports-usesharedcache--page) hook to store the result of the effect and, if that result is available, returns it without rerunning the effect. The precise behavior of the
-hook can be modified using the options.
+This hook invokes the given handler and caches the result using the [`useSharedCache`](/docs/data-exports-usesharedcache--page) hook. The `requestId` is used to both share identify inflight requests that can be shared, and to identify the cached value to use.
 
-There is currently no way to reinvoke the effect without changing the `requestId`.
+The hook returns an array containing the current state of the requested data, and a function that can be used to `refetch` the data on demand. Note that calling `refetch` while an inflight request is in progress for the given `requestId` will be a no-op.
+
+The behavior of the hook can be modified with the options.
 
 ```ts
 type CachedEffectOptions<TData: ValidCacheData> = {|
@@ -30,6 +31,7 @@ type CachedEffectOptions<TData: ValidCacheData> = {|
     retainResultOnChange?: boolean,
     onResultChanged?: (result: Result<TData>) => void,
     scope?: string,
+    fetchPolicy?: FetchPolicy,
 |};
 ```
 
@@ -39,3 +41,4 @@ type CachedEffectOptions<TData: ValidCacheData> = {|
 | `retainResultOnChange` | `false` | When `true`, the effect will not reset the result to the loading status while executing if the requestId changes, instead, returning the existing result from before the change; otherwise, the result will be set to loading status.  If the status is loading when the changes are made, it will remain as loading; old pending effects are discarded on changes and as such this value has no effect in that case.|
 | `onResultChanged` | `undefined` | Callback that is invoked if the result for the given hook has changed. When defined, the hook will invoke this callback whenever it has reason to change the result and will not otherwise affect component rendering directly. When not defined, the hook will ensure the component re-renders to pick up the latest result. |
 | `scope` | `"useCachedEffect"` | Scope to use with the shared cache. When specified, the given scope will be used to isolate this hook's cached results. Otherwise, the default scope will be used. Changing this value after the first call is not supported. |
+| `fetchPolicy` | [`FetchPolicy`](/docs/data-types-fetchpolicy--page) | Fetch policy to use when fetching the data. Defaults to `FetchPolicy.CacheBeforeNetwork`. |
--- a/packages/wonder-blocks-data/src/hooks/__tests__/use-cached-effect.test.js
+++ b/packages/wonder-blocks-data/src/hooks/__tests__/use-cached-effect.test.js
@@ -109,7 +109,9 @@ describe("#useCachedEffect", () => {
 
                 // Act
                 const {
-                    result: {current: result},
+                    result: {
+                        current: [result],
+                    },
                 } = serverRenderHook(() => useCachedEffect("ID", fakeHandler));
 
                 // Assert
@@ -129,7 +131,9 @@ describe("#useCachedEffect", () => {
 
                 // Act
                 const {
-                    result: {current: result},
+                    result: {
+                        current: [result],
+                    },
                 } = serverRenderHook(() => useCachedEffect("ID", fakeHandler));
 
                 // Assert
@@ -311,7 +315,7 @@ describe("#useCachedEffect", () => {
             );
 
             // Assert
-            expect(result.current).toStrictEqual(Status.success("DATA2"));
+            expect(result.current[0]).toStrictEqual(Status.success("DATA2"));
         });
 
         it("should not fulfill request when skip is true", () => {
@@ -366,7 +370,7 @@ describe("#useCachedEffect", () => {
             );
 
             // Assert
-            expect(result.current).toStrictEqual(Status.success("DATA1"));
+            expect(result.current[0]).toStrictEqual(Status.success("DATA1"));
         });
 
         it("should not ignore inflight request if options (other than skip) change", async () => {
@@ -389,7 +393,7 @@ describe("#useCachedEffect", () => {
             await act((): Promise<mixed> => response1);
 
             // Assert
-            expect(result.current).toStrictEqual(Status.success("DATA1"));
+            expect(result.current[0]).toStrictEqual(Status.success("DATA1"));
         });
 
         it("should return previous result when requestId changes and retainResultOnChange is true", async () => {
@@ -417,7 +421,7 @@ describe("#useCachedEffect", () => {
             );
             await act((): Promise<mixed> => response1);
             rerender({requestId: "ID2"});
-            const result = hookResult.current;
+            const [result] = hookResult.current;
             await waitForNextUpdate();
 
             // Assert
@@ -449,7 +453,7 @@ describe("#useCachedEffect", () => {
             rerender({requestId: "ID2"});
 
             // Assert
-            expect(result.current).toStrictEqual(Status.loading());
+            expect(result.current[0]).toStrictEqual(Status.loading());
         });
 
         it("should trigger render when request is fulfilled and onResultChanged is undefined", async () => {
@@ -464,7 +468,7 @@ describe("#useCachedEffect", () => {
             await act((): Promise<mixed> => response);
 
             // Assert
-            expect(result.current).toStrictEqual(Status.success("DATA"));
+            expect(result.current[0]).toStrictEqual(Status.success("DATA"));
         });
 
         it("should not trigger render when request is fulfilled and onResultChanged is defined", async () => {
@@ -481,7 +485,7 @@ describe("#useCachedEffect", () => {
             await act((): Promise<mixed> => response);
 
             // Assert
-            expect(result.current).toStrictEqual(Status.loading());
+            expect(result.current[0]).toStrictEqual(Status.loading());
         });
 
         it("should call onResultChanged when request is fulfilled and onResultChanged is defined", async () => {
--- a/packages/wonder-blocks-data/src/hooks/__tests__/use-cached-effect.test.js
+++ b/packages/wonder-blocks-data/src/hooks/__tests__/use-cached-effect.test.js
@@ -650,6 +650,40 @@ describe("#useCachedEffect", () => {
                 (e) => e !== FetchPolicy.CacheOnly,
             ),
         )(
+            "should trigger render once per inflight request being fulfilled and onResultChanged is undefined for FetchPolicy.%s",
+            async (fetchPolicy) => {
+                // Arrange
+                const response = Promise.resolve("DATA");
+                const fakeHandler = jest.fn().mockReturnValue(response);
+                let renderCount = 0;
+                const Component = React.memo(() => {
+                    const [, refetch] = useCachedEffect("ID", fakeHandler, {
+                        fetchPolicy,
+                    });
+                    React.useEffect(() => {
+                        refetch();
+                        refetch();
+                        refetch();
+                        refetch();
+                    }, [refetch]);
+                    renderCount++;
+                    return <div>Hello :)</div>;
+                });
+
+                // Act
+                render(<Component />);
+                await reactAct((): Promise<mixed> => response);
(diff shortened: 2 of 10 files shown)


packages/wonder-blocks-data/src/hooks/use-cached-effect.js
```diff
@@ -81,8 +96,9 @@ export const useCachedEffect = <TData: ValidCacheData>(
     options: CachedEffectOptions<TData> = ({}: $Shape<
         CachedEffectOptions<TData>,
     >),
-): Result<TData> => {
+): [Result<TData>, () => void] => {
```
somewhatabstract, line 99: Now returning the result and a `refetch` method in a tuple.

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
```diff
@@ -104,107 +120,152 @@ export const useCachedEffect = <TData: ValidCacheData>(
...
-
-            // If our caller provided a cacheUpdated callback, we use that.
-            // Otherwise, we toggle our little state update.
-            if (onResultChanged != null) {
-                onResultChanged(value);
-            } else {
-                forceUpdate();
+    // For the NetworkOnly fetch policy, we ignore the cached value.
+    // So we need somewhere else to store the network value.
+    const networkResultRef = React.useRef();
```
somewhatabstract, line 126: For the NetworkOnly fetch policy, we need a way to track that network result outside of the shared cache. I did try using state and replacing the force update but I was getting some infinite render loop issues and I couldn't quite track it down. So for now, I'll do it this way where the value is stored as a ref and we just look at it on render to determine what to do.

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
```diff
@@ -104,107 +120,152 @@ export const useCachedEffect = <TData: ValidCacheData>(
...
+
+    // Set up the function that will do the fetching.
+    const currentRequestRef = React.useRef();
+    const fetchRequest = React.useMemo(() => {
```
somewhatabstract, line 130: We move this out of the `useEffect` so we can reference it for our `refetch` ability.

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
```diff
@@ -104,107 +120,152 @@ export const useCachedEffect = <TData: ValidCacheData>(
...
+        // We aren't using useCallback here because we need to make sure that
+        // if we are rememo-izing, we cancel any inflight request for the old
+        // callback.
+        currentRequestRef.current?.cancel();
+        currentRequestRef.current = null;
+        networkResultRef.current = null;
```
somewhatabstract, line 136: We use `useMemo` so that when we're asked to recreate the callback, we have a chance to reset some stuff from any previous invocation.

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
```diff
@@ -104,107 +120,152 @@ export const useCachedEffect = <TData: ValidCacheData>(
...
+            const request = RequestFulfillment.Default.fulfill(requestId, {
+                handler: interceptedHandler,
+            });
+
+            if (request === currentRequestRef.current?.request) {
+                // The request inflight is the same, so do nothing.
+                // NOTE: Perhaps if invoked via a refetch, we will want to
+                // override this behavior and force a new request?
+                return;
+            }
```
somewhatabstract, line 157: If we get invoked more than once (like someone calling `refetch` twice in a row or something), this just protects us from adding our `.then` handler more than once.

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
```diff
@@ -104,107 +120,152 @@ export const useCachedEffect = <TData: ValidCacheData>(
...
+
+            // Clear the last network result.
+            networkResultRef.current = null;
+
+            // Cancel the previous request.
+            currentRequestRef.current?.cancel();
```
somewhatabstract, line 163: So, a `refetch` occurs but the inflight request is different. I don't "think" this is technically possible since the callback itself would've been recreated, but being defensive to make sure.

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
```diff
@@ -104,107 +120,152 @@ export const useCachedEffect = <TData: ValidCacheData>(
...
+
+            // TODO(somewhatabstract, FEI-4276):
+            // Until our RequestFulfillment API supports cancelling/aborting, we
+            // will have to do it.
+            let cancel = false;
```
somewhatabstract, line 168: When aborts are real aborts that give AbortError responses, I don't believe we'll need to guard the `then` with a cancel token anymore. That'll be nice. Once day.

jeresig (approved):
Looks good to me! Glad we are building up parity on this.

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
```diff
@@ -104,107 +120,152 @@ export const useCachedEffect = <TData: ValidCacheData>(
...
+                    cancel = true;
+                    RequestFulfillment.Default.abort(requestId);
+                },
+            };
+        };
+        // We deliberately ignore the handler here because we want folks to use
+        // interceptor functions inline in props for simplicity. This is OK
+        // since changing the handler without changing the requestId doesn't
+        // really make sense - the same requestId should be handled the same as
+        // each other.
```
kevinbarabash, line 211: This feels like it could admit subtle issues if the interceptor function captures variables from its closure.  Is this the existing behavior?

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
kevinbarabash: Thanks for adding this comment about the use of `useMemo` over `useCallback`.

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
```diff
@@ -104,107 +120,152 @@ export const useCachedEffect = <TData: ValidCacheData>(
...
+                );
             }
-        },
-        [setMostRecentResult, onResultChanged, forceUpdate],
-    );
+            // We use our request fulfillment here so that in-flight
+            // requests are shared.
+            const request = RequestFulfillment.Default.fulfill(requestId, {
+                handler: interceptedHandler,
+            });
```
kevinbarabash, line 150: Is `fulfill` memoized?

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
```diff
@@ -104,107 +120,152 @@ export const useCachedEffect = <TData: ValidCacheData>(
...
+    const fetchRequest = React.useMemo(() => {
+        // We aren't using useCallback here because we need to make sure that
+        // if we are rememo-izing, we cancel any inflight request for the old
+        // callback.
+        currentRequestRef.current?.cancel();
+        currentRequestRef.current = null;
+        networkResultRef.current = null;
+
+        // Now we can return the new fetch function.
+        return () => {
```
kevinbarabash, line 139: style nit: It took me a while to find this `return` statement.  Assigning this to a variable and returning a variable from the bottom of the `useMemo()` might improve the readability a bit.

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
```diff
@@ -104,107 +120,152 @@ export const useCachedEffect = <TData: ValidCacheData>(
...
+        // interceptor functions inline in props for simplicity. This is OK
+        // since changing the handler without changing the requestId doesn't
+        // really make sense - the same requestId should be handled the same as
+        // each other.
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [
+        requestId,
+        onResultChanged,
+        forceUpdate,
+        setMostRecentResult,
```
kevinbarabash, line 217: I know that setters from `useState()` don't need to be listed.  I wonder if `setMostRecentResult` falls into that category.  The lint rule isn't smart enough to know unfortunately.

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
```diff
@@ -104,107 +120,152 @@ export const useCachedEffect = <TData: ValidCacheData>(
...
+        fetchPolicy,
+    ]);
```
kevinbarabash, line 219: This feels like it could be extracted into a separate `useFetch()` hook.  Not necessary for this PR, but maybe something for a followup PR.

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
```diff
@@ -104,107 +120,152 @@ export const useCachedEffect = <TData: ValidCacheData>(
...
-            // least once and so we already made the request at least once. So
-            // we can bail out right here.
-            return true;
+    // indicates its a different request.
+    const requestIdRef = React.useRef(requestId);
+
+    // Calculate if we want to fetch the result or not.
+    // If this is true, we will do a new fetch, cancelling the previous fetch
+    // if there is one inflight.
+    const shouldFetch = React.useMemo(() => {
```
kevinbarabash, line 228: I like how this logic has been consolidated.  It reads quite nicely.

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
```diff
@@ -104,107 +120,152 @@ export const useCachedEffect = <TData: ValidCacheData>(
...
+    }, [requestId, mostRecentResult, fetchPolicy, hardSkip]);
 
-        return false;
-    }, [requestId, previousRequestId, mostRecentResult]);
+    // Let's make sure our ref is set to the most recent requestId.
+    requestIdRef.current = requestId;
 
-    // So now we make sure the client-side request happens per our various
-    // options.
     React.useEffect(() => {
```
kevinbarabash, line 259: The effect is comically simple now, love it.

packages/wonder-blocks-data/src/hooks/__tests__/use-cached-effect.test.js
```diff
@@ -91,49 +100,94 @@ describe("#useCachedEffect", () => {
             },
         );
 
-        it("should not request data", () => {
-            // Arrange
-            const fakeHandler = jest.fn().mockResolvedValue("data");
+        it.each(Array.from(FetchPolicy.members()))(
+            "should not request data for FetchPolicy.%s",
```
kevinbarabash, line 104: Could you add something to the title about the reason for it not requesting data?  It's because `fakeHandler` is resolved with data, right?

packages/wonder-blocks-data/src/hooks/__tests__/use-cached-effect.test.js
```diff
@@ -190,13 +229,137 @@ describe("#useCachedEffect", () => {
...
+                } = clientRenderHook(() =>
+                    useCachedEffect("ID", fakeHandler, {fetchPolicy}),
+                );
+                fakeHandler.mockClear();
+                await act((): Promise<mixed> => response);
+                act(refetch);
+                await act((): Promise<mixed> => response);
+
+                // Assert
+                expect(fakeHandler).toHaveBeenCalledTimes(1);
```
kevinbarabash, line 257: What's the connection between fakeHandler being called once and the test title?

packages/wonder-blocks-data/src/hooks/__tests__/use-cached-effect.test.js
```diff
@@ -190,13 +229,137 @@ describe("#useCachedEffect", () => {
...
+            // Assert
+            expect(underTest).toThrowErrorMatchingInlineSnapshot(
+                `"Cannot fetch with CacheOnly policy"`,
+            );
+        });
+
+        it.each(
+            Array.from(FetchPolicy.members()).filter(
+                (e) => e !== FetchPolicy.CacheOnly,
+            ),
```
kevinbarabash, line 286: style nit: assigning this value to `AllFetchPoliciesExceptCacheOnly` would help with readability or list them out in an array like you do in the next `it.each()`.

packages/wonder-blocks-data/src/hooks/__tests__/use-cached-effect.test.js
```diff
@@ -449,59 +616,158 @@ describe("#useCachedEffect", () => {
...
+                let renderCount = 0;
+                const Component = React.memo(() => {
+                    const [, refetch] = useCachedEffect("ID", fakeHandler, {
+                        fetchPolicy,
+                    });
+                    React.useEffect(() => {
+                        refetch();
+                        refetch();
+                        refetch();
+                        refetch();
```
kevinbarabash, line 667: These get collapsed to a single request right?

packages/wonder-blocks-data/src/hooks/__tests__/use-hydratable-effect.test.js
```diff
@@ -407,7 +413,7 @@ describe("#useHydratableEffect", () => {
             rerender({requestId: "ID2"});
 
             // Assert
-            expect(result.current).toStrictEqual(Status.loading());
+            expect(result.current).toStrictEqual(Status.success("GOODDATA"));
```
kevinbarabash, line 416: Why did this behavior change?

kevinbarabash (approved):
The implementation changes look reasonable to me.  The tests are very comprehensive, but I do have a couple of questions about some of the changes in behavior in a couple of the tests.

I'm also a little concerned about the move to return an array instead of an object like before.  If you haven't done so already, you may want to do a pulse check with the 321 team about that detail publishing.

packages/wonder-blocks-data/src/__docs__/exports.use-cached-effect.stories.mdx
```diff
@@ -16,20 +16,22 @@ function useCachedEffect<TData: ValidCacheData>(
     requestId: string,
     handler: () => Promise<TData>,
     options?: CachedEffectOptions<TData>,
-): Result<TData>;
+): [Result<TData>, () => void];
```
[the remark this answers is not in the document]
somewhatabstract: I don't think `Result` should have an API like that. It's a cacheable piece of state information. The hooks pattern is well-defined to have APIs that return methods in a tuple like this.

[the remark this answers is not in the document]
somewhatabstract: Also, `Result` is used in many more places than just this hook - it wouldn't make sense to have a `refetch` in those other circumstances.

packages/wonder-blocks-data/src/__docs__/exports.use-cached-effect.stories.mdx
```diff
@@ -16,20 +16,22 @@ function useCachedEffect<TData: ValidCacheData>(
...
 ```
 
-This hook is a wrapper around `useEffect`. Internally it uses the [`useSharedCache`](/docs/data-exports-usesharedcache--page) hook to store the result of the effect and, if that result is available, returns it without rerunning the effect. The precise behavior of the
-hook can be modified using the options.
+This hook invokes the given handler and caches the result using the [`useSharedCache`](/docs/data-exports-usesharedcache--page) hook. The `requestId` is used to both share identify inflight requests that can be shared, and to identify the cached value to use.
 
-There is currently no way to reinvoke the effect without changing the `requestId`.
+The hook returns an array containing the current state of the requested data, and a function that can be used to `refetch` the data on demand. Note that calling `refetch` while an inflight request is in progress for the given `requestId` will be a no-op.
```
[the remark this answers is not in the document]
somewhatabstract: I disagree. THat would be like the `useState` method returning the current state with a `set` function on it. It doesn't make sense. The current result is different from the API to update the current result.

packages/wonder-blocks-data/src/hooks/__tests__/use-cached-effect.test.js
somewhatabstract: No, it's because it's server-side. That's the entire context here. This hook should never invoke the effect when server-side. I feel like adding that to the test name duplicates information already held in the parent describe.

packages/wonder-blocks-data/src/hooks/__tests__/use-cached-effect.test.js
somewhatabstract: Just being explicit. When refetch is called, the handler should be called. Seems fair to also expect it to only invoke the handler once in that scenario.

packages/wonder-blocks-data/src/hooks/__tests__/use-cached-effect.test.js
somewhatabstract: Exactly. That's what the test is asserting.

packages/wonder-blocks-data/src/hooks/__tests__/use-hydratable-effect.test.js
somewhatabstract: Because the previous test case was actually copypasta and not doing what the test title stated. I updated it to have cached data for the new request ID to assert that it does indeed do what the test title says.

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
somewhatabstract: No, but why would it need to be? It's not being recreated inline or anything. It's just defined function.

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
somewhatabstract: Yes, it is.

Since interception is only used in fixtures and tests, I don't think that subtlety is an issue and we can call it out in docs if it becomes one.

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
somewhatabstract: Yeah, I don't know, but those setters are memoized, so it doesn't seem to be an issue. :)

PR #1437, opened 2022-04-25
[fei4371.1.consolidatecacheclearingapis] Tidy up APIs

somewhatabstract:
## Summary:
This changes our cache purging APIs to use a common nomenclature, clearer naming with respect to the hydration cache, as well as adding a new `purgeCaches` method that clears all of the caches in one go.

This also renames the tracked requests API to be more specific, since this package is now much more than just SSR data request tracking and hydration.

- Removed `removeFromCache` export as nothing uses it
- Renames `removeAllFromCache` to `purgeHydrationCache`
- Renames `clearSharedCache` to `purgeSharedCache`
- Renames `SsrCache.removeAll`  to `SsrCache.purgeData`
- Adds new export, `purgeCaches` that purges all caches
- Renames `hasUnfulfilledRequests` to `hasTrackedRequestsToBeFetched`
- Renames `fulfillAllDataRequests` to `fetchTrackedRequests`

Issue: FEI-4371

## Test plan:
`yarn test`
`yarn flow`

I also grepped webapp for calls to `removeFromCache` before I removed it.


.changeset/good-pants-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-data": minor
+---
+
+Add `purgeCaches` export for purging all caches with one call
```
somewhatabstract, line 5: I did each chunk of API or renaming as a commit and did a changeset for each one to keep track. Put up as a single PR because it didn't seem worth splitting - I can if necessary though.

packages/wonder-blocks-data/src/index.js
```diff
@@ -23,79 +13,16 @@ export type {
...
- * @param {(id: string) => boolean} [predicate] The predicate to match against
- * the cached hydration responses. If no predicate is provided, all cached
- * hydration responses will be removed.
- */
-export const removeAllFromCache = (
-    predicate?: (
-        key: string,
-        cacheEntry: ?$ReadOnly<CachedResponse<ValidCacheData>>,
-    ) => boolean,
-): void => SsrCache.Default.removeAll(predicate);
```
somewhatabstract, line 90: No more untested functionality hidden in index.js. These now live in `xxx- api.js` files.

jeresig (approved):
Makes sense to me - thanks for the clear commits and PR message.

PR #1438, opened 2022-04-25
[fei4454.1] client-side inflight requests consider scope

somewhatabstract:
## Summary:
This stops exporting our internal `RequestFulfillment` API (webapp doesn't use it and we don't really want to support this as an externally facing API).
This then updates the `useCachedEffect` hook to add the scope to request IDs when fulfilling them so that different scopes don't share an ID.

The next PR will add an `abortInflightRequests` export as well as add knowledge of `scope` to the request interception API.

Issue: FEI-4454

## Test plan:
`yarn test`

Grepped webapp for `RequestFulfillment` usage. If we find we want this for say, `gqlFetch` calls in the future, we should look at how we might want to expose that in an effective way (for example, internally handling that in the `gqlFetch` call returned by Wonder Blocks Data's `useGql` hook).


jeresig (approved):

packages/wonder-blocks-data/src/hooks/use-cached-effect.js
```diff
@@ -143,10 +143,15 @@ export const useCachedEffect = <TData: ValidCacheData>(
...
             // We use our request fulfillment here so that in-flight
-            // requests are shared.
-            const request = RequestFulfillment.Default.fulfill(requestId, {
-                handler: interceptedHandler,
-            });
+            // requests are shared. In order to ensure that we don't share
+            // in-flight requests for different scopes, we add the scope to the
+            // requestId. This is just internal, so nothing else will care.
+            const request = RequestFulfillment.Default.fulfill(
+                `${requestId}|${scope}`,
```
somewhatabstract, line 150: This solves the immediate issue that Project321 was experiencing. Subsequent PRs need to address this for SSR tracked requests - however, that's less urgent since we don't have different scoped uses of that yet (`useFastlyQuery` is currently the only SSR usage we have and its scope value is fixed); i.e. the code is broken, but in practice, we never hit the broken paths.

I'll fix that up in separate PRs.

PR #1448, opened 2022-04-28
[fei4370.2.fixcaching] Fix shared cache usage

somewhatabstract:
## Summary:
The two parallel jobs were missing node-version from their matrices and as such, they were looking at the wrong cache keys. This should now properly share the same keys and speed things up considerably.

Issue: FEI-4370

## Test plan:
See the jobs run on GitHub. Run jobs, run. 🐇

### DIFF
--- a/.github/actions/shared-node-cache/action.yml
+++ b/.github/actions/shared-node-cache/action.yml
@@ -1,0 +1,28 @@
+name: 'Shared Node Cache'
+description: 'Install & cache our JavaScript dependencies'
+inputs:
+  node-version:
+    description: 'Node version to use'
+    required: false
+    default: '16.x'
+runs:
+  using: "composite"
+  steps:
+
+    - name: Use Node.js ${{ inputs.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ inputs.node-version }}
+
+    - uses: actions/cache@v2
+      id: node-modules-cache # use this to check for `cache-hit` (`steps.node-modules-cache.outputs.cache-hit != 'true'`)
+      with:
+        path: node_modules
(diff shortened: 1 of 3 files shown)


.github/actions/shared-node-cache/action.yml
```diff
@@ -0,0 +1,28 @@
+name: 'Shared Node Cache'
```
somewhatabstract, line 1: This is based on the action that @jaredly wrote for https://github.com/Khan/webapp.

.github/workflows/node-ci-pr.yml
```diff
@@ -21,21 +21,10 @@ jobs:
         os: [ubuntu-latest]
         node-version: [16.x]
```
somewhatabstract, line 22: This has the `node-version` so it would create a cache key starting with `Linux16.x-`

.github/workflows/node-ci-pr.yml
```diff
@@ -44,22 +33,18 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
```
somewhatabstract, line 35: This did not have the `node-version` and so was creating a cache key starting with just `Linux-`

.github/workflows/node-ci-pr.yml
```diff
@@ -104,22 +89,18 @@ jobs:
...
```
somewhatabstract, line 91: This did not have the node-version and so was creating a cache key starting with just Linux-

somewhatabstract:
NOTE: The check build sizes step will fail because it changes the checked out code to the earlier branch and as such, the `action.yml` used to set up the caching check is no longer available to run tidy up.

jeresig (approved):
Ahh - thank you for fixing this!

PR #1451, opened 2022-04-29
Fix focus issue with filterable dropdowns

nishasy:
## Summary:
Currently, if a dropdown switches between a virtualized list (125+ results) and
a non-virtualized list (< 125 results), then the search field rerenders and
loses focus. This means that a user is typing in the search field and the filter
results change mid-typing, the focus leaves the text field and the user will have
to click back in to keep typing.

As a temporary fix, here we force focus on the field on mount if the focus was already
previously on the field.

The issue should go away with WB-1310, but we are adding this quick fix as a bandaid for now.

Issue: https://khanacademy.atlassian.net/browse/WB-1281

## Before:
Screen Shot 2022-04-29 at 5 46 49 PM

https://user-images.githubusercontent.com/13231763/166083939-c26ea043-c6d0-43b0-98b4-10c765965d21.mov

## After:
https://user-images.githubusercontent.com/13231763/166083866-5b3a9284-1444-49e1-8440-8bccfa4f16a7.mov

## Test plan:

Storybook
- Go to storybook
- Go to Dropdown/SingleSelect/With Filter Opened
- Type in "fruit # 5 banana" and confirm that the focus doesn't leave the search field.
- Backspace back to "fruit" and confirm that the focus doesn't leave the search field.

NOTE: I tried adding a unit test in `dropdown-core.test.js` and `single-select.test.js` but it doesn't look like there's a good way to test the behavior with the current implementation of SingleSelect/Dropdown and the current tools in React Testing Library. I wrote some tests that looked like they were intuitively testing the behavior, but when I commented out the bug fix, the tests didn't fail like they were supposed to, due to how the rendering works with dropdowns. So they wouldn't even help prevent this from happening again. While it would be ideal to have unit tests for this, Dropdowns are supposed to be refactored soon as part of WB-1310, so hopefully it's okay to forego adding tests for now.

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/single-select.stories.js
+++ b/packages/wonder-blocks-dropdown/src/components/single-select.stories.js
@@ -79,6 +79,7 @@ const optionItems = new Array(1000)
     ));
 
 type Props = {|
+    selectedValue: string,
     opened: boolean,
 |};
 
@@ -88,16 +89,18 @@ type State = {|
 |};
 
 type DefaultProps = {|
+    selectedValue: $PropertyType<Props, "selectedValue">,
     opened: $PropertyType<Props, "opened">,
 |};
 
 class SingleSelectWithFilter extends React.Component<Props, State> {
     static defaultProps: DefaultProps = {
+        selectedValue: "2",
         opened: false,
     };
 
     state: State = {
-        selectedValue: "2",
+        selectedValue: this.props.selectedValue,
         opened: this.props.opened,
     };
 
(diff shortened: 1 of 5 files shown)


packages/wonder-blocks-dropdown/src/components/single-select.stories.js
```diff
@@ -167,6 +167,23 @@ export const WithFilterOpened: StoryComponentType = () => (
     <SingleSelectWithFilter opened={true} />
 );
```
nishasy, line 169: I added this story because I wanted to see if the search field automatically receives focus if the filter is open and no value is selected. It does.

nishasy: Not sure why the open dropdown isn't showing up in the snapshot. I guess it's not there on initial render? Should I disable the snapshot?

jandrade: oh that's an issue with Chromatic not being able to capture some "async" snapshots (we have seen this already happening with some Tooltip and Popover stories that we had to disable for now).

One thing that you can try out is to use the [delay](https://www.chromatic.com/docs/delay) feature and see if that fixes the issue.

packages/wonder-blocks-dropdown/src/util/constants.js
```diff
@@ -24,6 +24,7 @@ export const filterableDropdownStyle = {
 export const searchInputStyle = {
     margin: Spacing.xSmall_8,
     marginTop: Spacing.xxxSmall_4,
+    minHeight: "auto",
```
jandrade, line 27: Could you also add a comment here to explain why we have to override this please?

jandrade (approved):
This looks great!! Thanks for fixing this issue 🥇 . Hopefully we'll land on a more optimal solution once we refactor the `dropdown-core` component in WB-1310.

packages/wonder-blocks-dropdown/src/components/single-select.stories.js
nishasy: Hmm... the delay doesn't seem to be helping. I'm wondering if the current on chromatic the same as it is for the `WithFilterOpened` story

nishasy: I updated the story so that it uses the same class as the other ones, and now the chromatic issue is gone!

PR #1466, opened 2022-05-09
Move `flow-enums-runtime` dep to individual packages

jeremywiebe:
## Summary:
Ensure individual packages that use enums depend on flow-enums-runtime package. In this monorepo, we don't "hoist" the root package.json dependencies entries up to individual packages. As a result, even though everything works in this repo when building and testing, when a project depends on @khanacademy/wonder-blocks-core that doesn't also depend on `flow-enums-runtime`, we encounter resolution errors.

This PR moves the dependeny up into the individual packages that use Flow enums. Unfortunately, the dependency isn't obvious in our code because it's used by the babel flow plugin. I've removed the dependency from the root package.json so that it is more obvious when a package takes on this dependency (ie. you'll need to add it to each new package that starts to use Flow enums).

Issue: "none"

## Test plan:

I discovered this when I `yarn link`'d in Perseus (which depends on `@khanacademy/wonder-blocks-core`)


kevinbarabash (approved):
LGTM.  Thanks for fixing this.  This needs a .changeset file though otherwise you won't be able to publish a new version of the packages you updated.

PR #1469, opened 2022-05-16
Make flow-enums-runtime a peer dependency

kevinbarabash:
## Summary:
The reason for making this change is to avoid having to install this dependency multiple times.

Issue: None

## Test plan:
- let GitHub Actions run all the checks


jandrade (approved):
LGTM!

PR #1473, opened 2022-05-17
WB-1132.1: Add SB docs for addStyle and View

jandrade:
## Summary:

- Added Storybook documentation for `View` (including a couple of new examples).
- Migrated `addStyle` docs and refactored the documentation a bit.
- Removed `enumerate-scroll-ancestors.js` b/c it's not being used anymore.

NOTE: First PR for migrating `wb-core` to Storybook (1/3).

Issue: WB-1132

## Test plan:

Navigate to SB under the `core/` section and verify the documentation for `View` and `addStyle`.

View: https://5e1bf4b385e3fb0020b7073c-qnuztwnegk.chromatic.com/?path=/docs/core-view--default
addStyle: https://5e1bf4b385e3fb0020b7073c-qnuztwnegk.chromatic.com/?path=/docs/core-addstyle--with-default-style


packages/wonder-blocks-core/src/components/view.js
```diff
@@ -45,16 +49,23 @@ const StyledSection = addStyle<"section">("section", styles.default);
...
- * `tag` prop.
+ * maps to `div`. You can override which tag is used to render the component
+ * (for semantic purposes) by specifying the `tag` prop.
  *
  * These components can take styles (via the `style` prop) in a variety of
  * manners:
  *
  * - An inline style object
  * - An `aphrodite` StyleSheet style
  * - An array combining the above
```
kevinbarabash, line 60: One thing that's missing that should probably be add is a description of the default styling for `View`s, in particular:
- use flexbox by default
- flex-direction is set to `column`
- they each get their own stacking context

Additionally, you may want to include what the inspiration and rationale were for these choices.

packages/wonder-blocks-core/src/util/enumerate-scroll-ancestors.js
```diff
@@ -1,98 +0,0 @@
-// @flow
-/**
- * Utility for finding the scroll ancestors of a child
- *
- * Inspired by https://github.com/olahol/scrollparent.js
- * with modifications to make an iterator that returns a sequence of all scroll
- * ancestors.
- *
- * Also modified for our standards (and commented a bit).
- */
```
kevinbarabash, line 10: Are we not using this any more?

kevinbarabash (approved):
LGTM.  Thanks for adding stories and docs for these things. 💚

packages/wonder-blocks-core/src/util/enumerate-scroll-ancestors.js
jandrade: That's right. We are not importing this module anywhere in the codebase. This was probably some leftovers that I forgot to remove when we migrated from Popper v1 to v2.

PR #1478, opened 2022-05-19
Set an aria-label on ModalDialog if one is passed in

jeanettehead:
## Summary:
Doing some accessibility testing on webapp with jest-axe I found a component that uses this dialog and has no way to pass the test if a label cannot be set (a good labelledby attribute isn't available)

Issue: wb-1324

## Test plan:
Set an aria-label attribute
Verify the dialog has the aria-label


jeresig (approved):
Very nice, thank you!

jandrade:
@jeanettehead could you please add changesets info? otherwise the change won't be published in `npm`.

You can see how this is done here: https://khanacademy.atlassian.net/wiki/spaces/WB/pages/100827261/Developing+Web+Wonder+Blocks#Versioning

Let me know if you have any questions about that.

PR #1498, opened 2022-05-24
Fix the autocomplete example for TextField in storybook

nishasy:
## Summary:
The autocomplete example wasn't providing autocompletion options
when you type something in. That's because it was missing the
submit action that a form needs to save this information.

- Put the TextField in a `<form>` and add a submit button.

Issue: https://khanacademy.atlassian.net/browse/WB-1294

Screen Shot 2022-05-24 at 4 41 47 PM

## Test plan:
- Go to storybook
- Go to the text field autocomplete example
- type something in
- press the submit button
- refresh the page
- click on the field or start typing something with the same starting letters as the input from before
- Confirm that options show up


jeresig (approved):
Ahh - tricky!! Good catch.

PR #1500, opened 2022-05-25
Upgrade Storybook to v6.5.5

jandrade:
## Summary:

- Upgrade Storybook to v6.5.5
- Upgrade SB addons to their latest versions.
- Include `react-refresh` and `@pmmmwh/react-refresh-webpack-plugin` as direct dependencies to avoid some issues with SB v6.5.5 due to mismatching versions.

Issue: XXX-XXXX

## Test plan:

Verify that the Storybook Chromatic instance continues working as expected.


jandrade:
> LGTM. It looks like the Chromatic job didn't actually build Storybook. You may need to push a commit that makes a change to a component/story to test that part of the workflow.

@kevinbarabash The job was built :) https://github.com/Khan/wonder-blocks/runs/6596996612?check_suite_focus=true

I'm thinking that maybe the existing chromatic workflow(s) are a bit confusing, so I'll probably make some changes around that.

PR #1501, opened 2022-05-25
WB-1311: Rename Button's `xlarge` size to `large`

jandrade:
## Summary:

- Changed `size` prop to use `large` (instead of `xlarge`).
- Updated stories to match this new value.
- Removed StyleGuidist Button docs.

Issue: WB-1311

## Test plan:

Navigate to the Button examples and verify that the `large` size works as designed in Figma.

SB: https://5e1bf4b385e3fb0020b7073c-ypnywjshye.chromatic.com/?path=/story/button--size

Screen Shot 2022-05-25 at 4 02 00 PM

Figma: https://www.figma.com/file/VbVu3h2BpBhH80niq101MHHE/Main-Components?node-id=389%3A0


megaroeny:
My _only_ piece of feedback would be if you change the button labels to say the size that they are (like I do in Figma), could be nice to make it super obvious. 😃  Thanks for the work on this!

packages/wonder-blocks-button/src/components/__docs__/button.argtypes.js
```diff
@@ -66,15 +66,15 @@ export default {
...
     size: {
         description: "The size of the button.",
-        options: ["small", "medium", "xlarge"],
+        options: ["small", "medium", "large"],
         control: {type: "select"},
         table: {
             category: "Layout",
             defaultValue: {
-                detail: `"medium" = height: 40; "small" = height: 32; "xlarge" = height: 60;`,
+                detail: `"medium" = height: 40; "small" = height: 32; "large" = height: 56;`,
```
kevinbarabash, line 74: We're sure we'll never need a button with height `48`?

packages/wonder-blocks-button/src/components/button-core.js
```diff
@@ -100,11 +100,13 @@ export default class ButtonCore extends React.Component<Props> {
 
         const Label = size === "small" ? LabelSmall : LabelLarge;
 
+        const iconSize = size === "small" ? "small" : "medium";
```
kevinbarabash, line 103: Could you add a comment to the code explaining why the icon sizes don't line up exactly with the button sizes?

packages/wonder-blocks-button/src/components/button-core.js
```diff
@@ -215,8 +217,9 @@ const sharedStyles = StyleSheet.create({
     small: {
         height: 32,
     },
-    xlarge: {
-        height: 60,
+    large: {
+        borderRadius: Spacing.xxSmall_6,
```
kevinbarabash, line 221: Nice!

kevinbarabash (approved):
LGTM. 🚀

packages/wonder-blocks-button/src/components/__docs__/button.argtypes.js
jandrade: Good question... that's what we ended up upon (@megaroeny in case you have any thoughts).

PR #1511, opened 2022-06-01
WB-1134.1: Add stories for ActionMenu

jandrade:
## Summary:

First part of the wb-dropdown migration to storybook:

- Adding stories for ActionMenu.
- Adding props table for both `ActionMenu` and `ActionItem`.

Issue: WB-1134

## Test plan:

Open storybook and navigate to `Dropdown > ActionMenu`.

Verify that all the examples are similar from the ones in the StyleGuidist site:

https://wonder-blocks.netlify.app/#actionmenu

Screen Shot 2022-06-01 at 3 35 22 PM

Screen Shot 2022-06-01 at 3 35 36 PM

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -8,4 +8,9 @@
     body {
         font-family: "Lato", sans-serif;
     }
+
+    /* Overrides the ArgTypes table tabs to avoid some issues with Popper */
+    .os-padding {
+        z-index: initial !important;
+    }
 </style>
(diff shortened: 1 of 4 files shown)


.storybook/preview-head.html
```diff
@@ -8,4 +8,9 @@
     body {
         font-family: "Lato", sans-serif;
     }
+
+    /* Overrides the ArgTypes table tabs to avoid some issues with Popper */
```
jeresig, line 12: Would be good to clarify what issues so that we can know if we can remove this safely in the future.

jeresig (approved):
Looking really nice!

PR #1514, opened 2022-06-01
Add controls for LabeledTextField to storybook

nishasy:
## Summary:
Add controls for LabeledTextField in storybook and add
descriptions for each story.

I also made a few small changes to TextField stories
as I was using them as a reference and figured I'd
udpate both at the same time when I spotted
areas of improvement.

I added an "Error Light" story to both `LabeledTextField` and
`TextField` in order to view the light error style.

View the following Storybook publish to see changes:

https://5e1bf4b385e3fb0020b7073c-gndecaotnq.chromatic.com/?path=/docs/form-labeledtextfield--default

Issue: https://khanacademy.atlassian.net/browse/WB-1295

## Test plan:
- Go to http://localhost:6061/?path=/docs/form-labeledtextfield--default
- Check that the description and usage info are there.
- Check that the default story is at the top.
- Read through the props and confirm that they all look right.
- Read through the story descriptions and interact with them
  all to confirm that they all look and behave as expected.


packages/wonder-blocks-form/src/components/__docs__/text-field.stories.js
```diff
@@ -419,23 +419,6 @@ Error.parameters = {
...
-    />
-);
-
-Disabled.parameters = {
-    docs: {
-        storyDescription: `If the \`disabled\` prop is set to true,
-        \`TextField\` will have disabled styling and will not be interactable.`,
-    },
-};
-
```
nishasy, line 438: Moved underneath the `ErrorLight` story.

packages/wonder-blocks-form/src/components/__docs__/labeled-text-field.argtypes.js
```diff
@@ -0,0 +1,248 @@
+// @flow
+
+export default {
+    id: {
+        description: `An optional unique identifier for the TextField.
```
jandrade, line 5: One thing that I'm not completely sure: you might not need to add descriptions here. I took a look at the labeled-text-field prop types, and it looks like Storybook might be able to get the description directly from the JSDocs, mostly because these definitions are located within the same file (again, not completely sure about this).

packages/wonder-blocks-form/src/components/__docs__/labeled-text-field.stories.js
```diff
@@ -51,13 +93,29 @@ export const RequiredWithDefaultText: StoryComponentType = () => {
             label="Name"
             description="Please enter your name"
             value={value}
-            onChange={(newValue) => setValue(newValue)}
+            onChange={setValue}
```
jandrade, line 96: Nice catch!

packages/wonder-blocks-form/src/components/text-field.js
```diff
@@ -348,6 +348,27 @@ type ExportProps = $Diff<
...
+ * const [value, setValue] = React.useState("");
+ *
+ * const handleChange = (newValue: string) => {
+ *     setValue(newValue);
+ * };
+ *
+ * <TextField
+ *     id="some-unique-text-field-id"
+ *     value={value}
+ *     onChange={handleChange}
```
jandrade, line 368: Same here

jandrade (approved):
Looks great! thanks for adding the Error Light examples.

packages/wonder-blocks-form/src/components/__docs__/labeled-text-field.argtypes.js
nishasy: Hmm... it seems kind of inconsistent? It pulls them fine sometimes but it wasn't able to here. I think it wasn't working in this case because `TextField` is a different element than `TextFieldInternal`, and the props are under `TextFieldInternal`. I had to add the Usage comment above the `TextField` at the bottom of the file too.

packages/wonder-blocks-form/src/components/labeled-text-field.js
```diff
@@ -293,6 +293,30 @@ type ExportProps = $Diff<
...
+ * const handleChange = (newValue: string) => {
+ *     setValue(newValue);
+ * };
+ *
+ * <LabeledTextField
+ *     label="Label"
+ *     description="Hello, this is the description for this field"
+ *     placeholder="Placeholder"
+ *     value={value}
+ *     onChange={handleChange}
```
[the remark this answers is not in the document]
nishasy: Ah yes. I considered doing that but I wasn't sure if it was done that way in the `TextField` examples on purpose to demonstrate a specific kind of usage.

PR #1519, opened 2022-06-07
Migrate Choice to Storybook

nishasy:
## Summary:
Migrate Choice to Storybook. In the styleguidist app,
the docs for Choice just say that it's meant to be used
with CheckboxGroup or RadioGroup and to go to those
sections to see the usage.

Here, I wanted the prop controls to be available, so I
one example with both checkbox and radios. The controls
update the `Choice` prop values on the same choice in
each respective list in the same story. I thought this
might be more clear than adding separate stories, as the
choice has the same props in either case.

Issue: https://khanacademy.atlassian.net/browse/WB-1291

Screen Shot 2022-06-06 at 5 41 27 PM
Screen Shot 2022-06-06 at 5 41 34 PM
Screen Shot 2022-06-06 at 5 41 39 PM

## Test plan:
- Go to http://localhost:6061/?path=/docs/form-choice--default
- Play with the controls and check that the label and
  description change the same way on both lists.


jandrade (approved):
Looks great! thanks for adding both examples in the usage section :shipit:

PR #1523, opened 2022-06-09
Don't rerun gerald for title/body edits to the PR

jaredly:
## Summary:
<Add your summary here>

Issue: XXX-XXXX

## Test plan:

### DIFF
--- a/.github/workflows/gerald-pr.yml
+++ b/.github/workflows/gerald-pr.yml
@@ -7,8 +7,8 @@ name: Gerald - Notify and Request Reviewers On Pull Request
     types: [opened, synchronize, reopened, ready_for_review, edited]
 jobs:
   gerald:
-    # Skip this workflow on dependabot PRs
-    if: github.actor != 'dependabot[bot]' && github.actor != 'dependabot-preview[bot]'
+    # Skip this workflow on dependabot PRs. Don't re-run if only the title or body changes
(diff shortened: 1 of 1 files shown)


.github/workflows/gerald-pr.yml
```diff
@@ -7,8 +7,8 @@ name: Gerald - Notify and Request Reviewers On Pull Request
     types: [opened, synchronize, reopened, ready_for_review, edited]
 jobs:
   gerald:
-    # Skip this workflow on dependabot PRs
-    if: github.actor != 'dependabot[bot]' && github.actor != 'dependabot-preview[bot]'
+    # Skip this workflow on dependabot PRs. Don't re-run if only the title or body changes
+    if: github.actor != 'dependabot[bot]' && github.actor != 'dependabot-preview[bot]' && (github.event.action != 'edited' || github.event.changes.base != null)
```
jeresig, line 11: Ahh - can we not put this into the shared action?

jeresig (approved):
Thanks for the fixes!

somewhatabstract (approved):
This is great. Thanks, @jaredly.

I also wonder the same thing as @jeresig; is this something we can include in the shared action(s)?

jaredly:
So, the shared action wouldn't allow us to "not run the workflow at all". It's possible we could do something like that with reusable workflows, but I haven't played with those yet https://docs.github.com/en/actions/using-workflows/reusing-workflows

PR #1530, opened 2022-06-14
[🔥AUDIT🔥] Fix flaky dropdown tests

jandrade:
🖍 _This is an audit!_ 🖍

## Summary:

Fixing some flaky tests that were probably introduced when worked on the
Dropdown search field changes.

Issue: "none"

## Test plan:

Verify that the GH action test steps pass constantly.


packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.js
```diff
@@ -19,6 +19,19 @@ const labels: $Shape<Labels> = {
...
+        window.scrollTo = jest.fn();
+
+        // We mock console.error() because React logs a bunch of errors pertaining
+        // to the use href="javascript:void(0);".
+        jest.spyOn(console, "error").mockImplementation(() => {});
+    });
+
+    afterEach(() => {
+        window.scrollTo.mockClear();
+        jest.spyOn(console, "error").mockReset();
```
jeresig, line 32: Do you have to do this for the console error? I think this happens automatically.

jeresig (approved):

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.js
jandrade: It happens automatically in the `webapp` repo, but we don't do it here. Probably worth copying that implementation in a separate PR.

PR #1539, opened 2022-06-29
[fei4635.1] Make fixtures framework CSF/storybook only and simplify API

somewhatabstract:
## Summary:
Issue: FEI-4635

## Test plan:
`yarn start:storybook`
`yarn test`


packages/wonder-blocks-testing/src/fixtures/fixtures.js
```diff
@@ -1,127 +1,78 @@
 // @flow
 import * as React from "react";
-import {getConfiguration} from "./setup.js";
-import {combineOptions} from "./combine-options.js";
+import {action} from "@storybook/addon-actions";
```
somewhatabstract, line 3: This file is basically most of what was in `storybook.js` but modified to work with our new syntax.

PR #1544, opened 2022-06-30
Update Spring and Strut in Storybook

nishasy:
## Summary:
Wonder Blocks Layout -- Spring and Strut are already in Storybook,
but they're extremely minimal. I added the controls here by
adding info in the `export default` block. I also added a
`Default` story in which the (very minimal) props can be modified.
I kept the `Simple` and `WithStyle` stories as is.

Issue: https://khanacademy.atlassian.net/browse/WB-1139

## Spring
Screen Shot 2022-06-30 at 3 33 29 PM
Screen Shot 2022-06-30 at 3 33 36 PM

## Strut
Screen Shot 2022-06-30 at 3 33 51 PM
Screen Shot 2022-06-30 at 3 33 57 PM

## Test plan:
- Confirm that adding a style to the style object in controls works.
- For Strut, confirm that changing the size works.


jeresig (approved):
Great! Thanks for improving these!

PR #1546, opened 2022-07-05
[fixfixturesbug] Make sure wrapper component is used if provided

somewhatabstract:
## Summary:
When I reworked the framework, I somehow dropped the ability to render the wrapper. This brings that back and adds a test case to cover it.

Issue: FEI-4603

## Test plan:
`yarn test`


jeresig (approved):
Ah ha! Thanks for the fix!

PR #1553, opened 2022-07-19
WB-1264: Add keyboard support for searching options

jandrade:
## Summary:

Added keyboard support to `DropdownCore` so we can try to match a given option
from the listbox.

This is done replicating the behavior from the `<select>` element:

- If the user focuses on the dropdown opener (but dropdown menu is CLOSED),
typing in will automatically SELECT the first occurrence.
- If the user focuses on the dropdown opener (and the dropdown menu is OPEN),
typing in will FOCUS on the first occurrence, but won’t select it unless the
user press space.

Issue: WB-1264

## Test plan:

Navigate to any SingleSelect example and verify that keyboard search is enabled
when you focus on the dropdown.

Navigate to the BirthdayPicker examples and try out the different fields (day,
month, year).

https://user-images.githubusercontent.com/843075/179804174-dbb16c45-257c-4398-8b77-de6d0c990628.mov

https://5e1bf4b385e3fb0020b7073c-dlukqygytd.chromatic.com/?path=/story/birthdaypicker--birthday-picker-default

### DIFF
--- a/.changeset/khaki-swans-add.md
+++ b/.changeset/khaki-swans-add.md
@@ -1,5 +1,0 @@
----
-"@khanacademy/wonder-blocks-birthday-picker": patch
----
-
-Included keyboard support to search options when the dropdown opener is focused
--- a/.changeset/khaki-swans-add.md
+++ b/.changeset/khaki-swans-add.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-birthday-picker": patch
+---
+
+Included keyboard support to search options when the dropdown opener is focused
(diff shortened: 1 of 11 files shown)


.changeset/khaki-swans-add.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-birthday-picker": patch
```
jeresig, line 2: Did you meant to have two changeset files?

jeresig (approved):
Ah - this seems like a great usability improvement, thank you!

.changeset/khaki-swans-add.md
jandrade: ahh oops! I thought the `patch` needed to be handled separately. I'll remove this file.

packages/wonder-blocks-dropdown/src/components/dropdown-core.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
```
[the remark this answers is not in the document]
jandrade: True, I'll probably change this in a separate PR/audit.

PR #1555, opened 2022-07-22
[fei4677.1] Make sure that if there is a cached response, we don't fetch

somewhatabstract:
## Summary:
I didn't add any new test cases here. If folks feel strongly that there should be test cases to verify this, then I can work on adding some.

Issue: FEI-4677

## Test plan:
`yarn test`

### DIFF
--- a/.changeset/weak-bugs-attend.md
+++ b/.changeset/weak-bugs-attend.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-data": patch
+---
+
+Do not fetch if FetchPolicy is CacheBeforeNetwork and there is already a cached value, even if the requestId changes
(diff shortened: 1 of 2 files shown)


jeresig (approved):
Hmm! I'm not sure I fully understand the ramifications of this change, but if you feel confident in it then it seems ok!

ejanzer (approved):
Thank you!

somewhatabstract:
> Hmm! I'm not sure I fully understand the ramifications of this change, but if you feel confident in it then it seems ok!

The impact is that if the `requestId` changes, and there is already a cached value, and the fetch policy is `CacheBeforeNetwork`, we won't bother making a network request.

All other cases remain the same.

PR #1561, opened 2022-08-04
Change darkBlue value in Wonder Blocks Color

nishasy:
## Summary:
Update the value of `darkBlue` from #0a2a66 to #0b2149.
Also, update the value of the `darkblue` background color for
Storybook accordingly.

Note that the manually written darkBlue colors in webapp will also
need to be updated in a different PR.

Issue: https://khanacademy.atlassian.net/browse/WB-1381

## Before:
Screen Shot 2022-08-03 at 5 07 20 PM

## After:
Screen Shot 2022-08-03 at 5 07 25 PM

## Test plan:
- Type `0a2a66` into the VS Code search back and confirm
  that there are no files (except for md files for now)
  that contain this value.
- Go to the Dark story under Button and check that the background
  is using the new blue.


jeresig (approved):
Looks great, thank you!

somewhatabstract (approved):

PR #1562, opened 2022-08-04
[fei4702.1] Export the SharedCache API

somewhatabstract:
## Summary:
This provides direct access to the in-memory cache used by `useSharedCache` (and therefore any hooks and components that also use that). This will enable us to have non-React code use the shared cache as needed.

See changesets for more info.

Issue: FEI-4702

## Test plan:
`yarn flow`
`yarn test`


packages/wonder-blocks-data/src/__docs__/exports.scoped-in-memory-cache.stories.mdx
```diff
@@ -11,12 +11,12 @@ import {Meta} from "@storybook/addon-docs";
 
 # ScopedInMemoryCache
 
-This class implements an in-memory cache that can contain different scopes of cached data. This allows for quick removal of entire classes of data as identified by their scopes without having to iterate each cached item to find them.
+This class implements an in-memory cache that can contain different scopes of cached data. This allows for quick removal of entire classes of data as identified by their scopes without having to iterate each cached item to find them. It implemnents the [`ScopedCache`](/docs/data-types-scopedcache--page) interface.
```
somewhatabstract, line 14: **nit:** `implemnents` -> `implements`

somewhatabstract:
> **Size Change:** -28 B (0%)

I love when adding functionality can still reduce overall code size (because I deleted a whole function here, and then just exported something that was already implemented in the code).

packages/wonder-blocks-data/src/__docs__/types.scoped-cache.stories.mdx
```diff
@@ -12,16 +12,103 @@ import {Meta} from "@storybook/addon-docs";
...
 
+    /**
+     * Retrieve a value from the cache.
+     */
+    get(scope: string, id: string): ?ValidCacheData;
+
+    /**
+     * Purge an item from the cache.
+     */
+    purge(scope: string, id: string): void;
```
kevinbarabash, line 26: **praise:** I like the change to a method-based interface, nice!

packages/wonder-blocks-data/src/__docs__/types.scoped-cache.stories.mdx
```diff
@@ -12,16 +12,103 @@ import {Meta} from "@storybook/addon-docs";
...
+```
+
+Sets a value in the cache within a given scope.
+
+### Throws
+
+| Error Type | Error Name | Reason |
+| ------ | ------ | ------ |
+| [`DataError`](/docs/data-exports-dataerror--page) | `InvalidInputDataError` | `id` and `scope` must be non-empty strings |
+| [`DataError`](/docs/data-exports-dataerror--page) | `InvalidInputDataError` | `value` must be a non-function value |
```
kevinbarabash, line 72: **off-topic**: I wish Flow/TS had a way to indicate what errors a function can throw.  Thanks for including it in the docs.  No action req'd.

packages/wonder-blocks-data/src/util/types.js
```diff
@@ -112,3 +112,40 @@ export type ErrorOptions = {|
      */
     cause?: ?Error,
 |};
+
+export interface ScopedCache {
```
kevinbarabash, line 116: I was confused when I first saw this because I we were redefining the interface, but what I had seen before was just the docs for the interface.  It's too bad our doc tools don't provide ways to import type definitions from code into the docs.  No action req'd.

kevinbarabash (approved):
LGTM.  The new interface and docs look great!  Thanks for refactoring this so that the cache can be used outside of React code. 🚀

PR #1568, opened 2022-08-13
[fei4681.1] Introduce response settling

somewhatabstract:
## Summary:
This introduces a pattern similar to `AbortController` and `AbortSignal` (but with a little less implementation - adding a `onsettled` event listener property seemed overkill, for example), but for settling a promise.

I chose this pattern so that it was familiar with existing APIs/solutions.

Now folks can create a `SettleController`, then pass it's `signal` property to the various `RespondWith` methods to create a pending outcome that can be settled (rejected or resolved) by calling `settle()` on the `SettleController`.

This means we can support more complex test scenarios where the order of operations is important. Previously, folks would need to create their own promises and responses in order to create this kind of behavior.

In the next PR, I'll be exposing a first class way to turn a `RespondWith` API result directly into a promise so that folks can retrieve a promise directly, as well as a special `RespondWith` call that can take such a promise and use it as the response. When combined with the changes in this PR, it should cater to all the various test scenarios we have encountered.

Issue: FEI-4681

## Test plan:
`yarn test`

### DIFF
--- a/.changeset/strong-penguins-kneel.md
+++ b/.changeset/strong-penguins-kneel.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-testing": patch
+---
+
+`RespondWith` API now supports a signal for controlling the settlement of the promise. Introduces `SettleController` to support this new feature.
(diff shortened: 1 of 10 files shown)


jeresig (approved):
Thanks for adding this, seems like this will make it easier to test some complex states...

PR #1569, opened 2022-08-15
[fei4682.1] Add toPromise to MockResponse

somewhatabstract:
## Summary:
This modifies the `RespondWith` API to abstract away the internals of response promise generation and to expose that via a `toPromise` function on the returned `MockResponse` instances. This will allow folks to use `RespondWith` outside of our mocking APIs, as needed, to generate their own `Promise<Response>` instances for use with fetch-based APIs.

Since the behavior is ultimately the same and this just adds a new function, I decided this was a minor update.

Issue: FEI-4682

## Test plan:
`yarn test`


packages/wonder-blocks-testing/src/respond-with.js
```diff
@@ -196,7 +219,18 @@ export const makeMockResponse = (
...
             });
 
+        /* istanbul ignore next */
         default:
-            throw new Error(`Unknown response type: ${response.type}`);
+            if (process.env.NODE_ENV !== "production") {
+                // If we're not in production, give an immediate signal that the
+                // dev forgot to support this new type.
+                throw new Error(`Unknown response type: ${response.type}`);
+            }
```
somewhatabstract, line 228: **note:** I did some checking when writing this to see what this all compiles down to. And although our testing library is likely only used in test, I thought it useful to still keep things useful in the various usage paths.

For folks that didn't know, like me, turns out that:

1. Such checks compile to `if (true) { /* the code */ }` or `if (false) { /*no code*/ }`; I expected the whole condition to be removed if the value was `false`.
2. If code is not in an `if` like this, it gets included regardless.

Not a big deal for this library, but worth noting when doing this elsewhere. In addition, this affects the Wonder Blocks build that we package; not the final usage point (i.e. webapp gets the "production" version, even in dev and testing).

jeresig (approved):
Looks reasonable to me!

PR #1570, opened 2022-08-15
Migrate OnePaneDialog documentation to storybook

nishasy:
## Summary:
Update the doucmentation for OnePaneDialog in storybook.
- Add argtypes file so that the control panel can be used.
- Create viewports like those in Styleguidist so that
  multiple modals can be viewed.
- Add stories for more cases.
- Add story descriptions.

Storybook publish: https://5e1bf4b385e3fb0020b7073c-kvwgzciqev.chromatic.com/

## Test plan:
- Read through the props and make sure they're right.
- Play around the controls with the control panel and
  confirm that they all behave as expected.
- Read the descriptions for all the stories and confirm that
  there are no typos and that they make sense.
- Try to open and close the modals with the launcher.


jeresig (approved):
This looks awesome - a huge improvement, thank you!

packages/wonder-blocks-modal/src/components/__docs__/one-pane-dialog.stories.js
```diff
@@ -3,15 +3,26 @@
...
 import {ActionMenu, ActionItem} from "@khanacademy/wonder-blocks-dropdown";
+import {Strut} from "@khanacademy/wonder-blocks-layout";
+import Link from "@khanacademy/wonder-blocks-link";
 import {ModalLauncher, OnePaneDialog} from "@khanacademy/wonder-blocks-modal";
-import {Body} from "@khanacademy/wonder-blocks-typography";
+import {Body, LabelLarge} from "@khanacademy/wonder-blocks-typography";
 
 import type {StoryComponentType} from "@storybook/react";
 
 import type {ModalElement} from "../../util/types.js";
```
nishasy, line 21: Hmm... so the `ModalElement` import was already here before I started working on this task. Something about this doesn't feel right. It's not a public type, so I don't think we should use it in the examples. Is there another way to type a modal? In the `WithLauncher` story, this is used where a modal is defined as follows:

```
const MyModal = ({closeModal}: MyModalProps): ModalElement => (...)
```

nishasy: @kevinbarabash Do you have any thoughts on this? Not sure if you're the right person to tag, but your name is on the last git blame.

jandrade (approved):
Looks great to me!

One minor suggestion would be adding the `Usage` section as part of the `one-pane-dialog.js` doc block, so we can show an inline snippet on how to use this component (and preserve consistency with the rest of WB packages). wdyt?

For reference: https://github.com/Khan/wonder-blocks/blob/85c31780fabbbb2c11af39827609b8e30ba46b1a/packages/wonder-blocks-button/src/components/button.js#L243

PR #1572, opened 2022-08-24
Migrate ModalLauncher documentation to storybook

nishasy:
## Summary:
Migrate ModalLauncher documentation to Storybook using
Styeguidist docs as a foundation.

Update OnePaneDialog with the "Flexible Modal" example from
Styleguidist.

NOTE: The JS Docs don't currently show up at the top of the
page because there is something wrong with the integration.
@jandrade is working on a fix.

Storybook publish: 
https://5e1bf4b385e3fb0020b7073c-eyfgklpjyu.chromatic.com/?path=/docs/modal-modallauncher--default

## Demo of the flexible modal in OnePaneDialog:

https://user-images.githubusercontent.com/13231763/186538331-213edb69-105e-47a8-b417-43f9cb80d0b1.mov

## Test plan:
- Open storybook and go to ModalLauncher
- Check that all the prop types look correct.
- Read through and make sure all the story descriptions make
  sense and that there are no typos.
- Interact with all the modals and make sure they behave as expected.


jeresig (approved):
Awesome - thanks for adding this!

packages/wonder-blocks-modal/src/components/__docs__/modal-launcher.stories.js
```diff
@@ -0,0 +1,380 @@
...
+        },
+        chromatic: {
+            viewports: [320, 640, 1024],
+        },
+    },
+    argTypes: ModalLauncherArgTypes,
+};
+
+export const Default: StoryComponentType = (args) => (
+    <ModalLauncher modal={DefaultModal}>
```
jandrade, line 86: **suggestion:** Spreading `...args` here would help us to try out the `Default` story with different combinations.

```suggestion
    <ModalLauncher modal={DefaultModal} {...args}>
```

jandrade (approved):
LGTM! just added a minor (non-blocking) comment around spreading args to the Default story. Everything looks awesome! 🏅

packages/wonder-blocks-modal/src/components/__docs__/modal-launcher.stories.js
nishasy: Oops, I actually meant to remove args altogether because I thought it doesn't need any of them specified. But I just tested it out and that would mean that none of the props would even work! Thanks for pointing this out.

PR #1574, opened 2022-08-26
Migrate ModalDialog to Storybook

nishasy:
## Summary:
Using Styleguidist as a foundation, migrate the documentation for
ModalDialog to Storybook.

Styleguidist does not have specific examples for each building
block under modal, but I decided to start adding them all with their own
examples (as opposed to using subcomponents), particularly so that
users can experiment with the control panel. This one was a little
tricky since ModalDialog does not have a visual element.

Storybook publish: 
https://5e1bf4b385e3fb0020b7073c-sbvqecqsiq.chromatic.com/?path=/docs/modal-building-blocks-modaldialog--default

## Test plan:
- Check that all the documentation makes sense.
- Check there are no typos.
- Interact with the modals to confirm they all behave as expected.


packages/wonder-blocks-modal/src/components/__docs__/modal-dialog.stories.js
```diff
@@ -0,0 +1,255 @@
...
+        below: {
+            control: {type: null},
+        },
+    },
+};
+
+export const Default: StoryComponentType = (args) => (
+    <View style={styles.previewSizer}>
+        <View style={styles.modalPositioner}>
+            <ModalDialog {...args}>
```
jandrade, line 78: **suggestion:** What do you think about adding `aria-labelledby` referencing to a title within the `ModalPanel` contents? 

That could be a good way to show the a11y case presented in the component description and we could show engineers how to use custom Dialogs while preserving accessibility.

Screen Shot 2022-08-26 at 8 52 28 AM

packages/wonder-blocks-modal/src/components/__docs__/modal-dialog.stories.js
```diff
@@ -0,0 +1,255 @@
...
+        height: 521,
+        position: "absolute",
+        top: -100,
+        left: -300,
+    };
+
+    return (
+        <View style={styles.previewSizer}>
+            <View style={styles.modalPositioner}>
+                <ModalDialog
```
jandrade, line 145: Same comment related to `aria-describedby` here.

jandrade (approved):
This looks great! Thanks for adding all these examples that will be super helpful for creating custom Dialogs. Just added a note for you to consider.

packages/wonder-blocks-modal/src/components/__docs__/modal-dialog.stories.js
nishasy: Great idea!

jandrade (approved):
Awesome! thanks for adding `aria-describedby` to the examples :shipit:

Issue #31, opened 2018-04-11
Add SPEC.md docs for existing v1 components


kevinbarabash:

kevinbarabash:
Tracking in JIRA.

PR #1575, opened 2022-08-29
WB-1142.2: Move `PopoverContent` to Storybook

jandrade:
## Summary:
Migrating the `PopoverContent` component to Storybook:

- Added stories to cover all variants.
- Included action mappings to provide flexible examples for the actions
  prop.
- Modified the JSDoc block to include the "Usage" guide.

Issue: WB-1142

## Test plan:

Navigate to the `Popover / PopoverContent` section and verify that all
the stories match with the React Styleguidist version:

Styleguidist: https://wonder-blocks.netlify.app/#popovercontent


somewhatabstract (approved):
LGTM

packages/wonder-blocks-popover/src/components/__docs__/popover-content.stories.js
```diff
@@ -0,0 +1,134 @@
...
+
+export default {
+    title: "Popover/PopoverContent",
+    component: PopoverContent,
+    argTypes: PopoverContentArgtypes,
+    parameters: {
+        componentSubtitle: ((
+            <ComponentInfo name={name} version={version} />
+        ): any),
+        docs: {
```
nishasy, line 24: I haven't putting this `docs` section in the components I've been migrating. Is this something I should have been doing?

packages/wonder-blocks-popover/src/components/__docs__/popover-content.argtypes.js
```diff
@@ -0,0 +1,116 @@
...
+    content: {
+        description: "The content to render inside the popover.",
+        type: {required: true, summary: "string"},
+    },
+    title: {
+        description: "The popover title",
+        type: {required: true, summary: "string"},
+    },
+    actions: {
+        description: `User-defined actions.\n\nIt can be either a Node or a function using the children-as-function pattern to pass a close function for use anywhere within the actions. This provides a lot of flexibility in terms of what actions may trigger the Popover to close the popover dialog`,
```
nishasy, line 62: Do you think you could specify that the `Emphasized` options are intended to be used when the `emphasized` prop is true? I found it confusing initially because if you just do this without setting the `emphasized` prop, it just shows white buttons on a white background and it doesn't look emphasized. Maybe `detail` could be appropriate for this.

packages/wonder-blocks-popover/src/components/__docs__/popover-content.argtypes.js
```diff
@@ -0,0 +1,116 @@
...
+    testId: {
+        description: "Test ID used for e2e testing.",
+        control: {type: "text"},
+        table: {
+            type: {
+                summary: "string",
+            },
+        },
+    },
+    icon: {
```
nishasy, line 100: It looks like pressing the "Set Object" button on the `Icon` prop in the controls causes the site to go black and crash for me. Is this happening for you too?

nishasy: Okay, that was happening in the Docs tab. In the canvas tab, it gives me a "Minified React error #31" error.

packages/wonder-blocks-popover/src/components/__docs__/popover-content.argtypes.js
```diff
@@ -0,0 +1,116 @@
...
+        description: `Decorate the popover with an illustrated icon. It cannot be used at the same time with image.`,
+        type: {
+            summary: `string | React.Element<"img"> | React.Element<"svg">`,
+        },
+    },
+    image: {
```
nishasy, line 106: I'm also getting an error if I try to set the object for the `image` prop.

packages/wonder-blocks-popover/src/components/__docs__/popover-content.argtypes.js
nishasy: If I understand the intent correctly, then it would be more clear if they were named with "Light" at the end instead of "Emphasized"?

nishasy (changes requested):
Requesting to at least fix the "Set Object" button for the `icon` and `image` props in the control panel.

packages/wonder-blocks-popover/src/components/__docs__/popover-content.argtypes.js
jandrade: oh nice catch! I'll create some `mappings` to provide some valid examples.

packages/wonder-blocks-popover/src/components/__docs__/popover-content.argtypes.js
jandrade: Will do the same here.

packages/wonder-blocks-popover/src/components/__docs__/popover-content.argtypes.js
jandrade: I originally named them `*Light` but noticed that there was no connection with the `Emphasized` state. I'll try to provide a better label for the dropdown.

packages/wonder-blocks-popover/src/components/__docs__/popover-content.stories.js
jandrade: ahh yeah, sorry that I totally missed that.

The thing is that `storyDescription` still works as of v6, but it will be deprecated in v7 in favor of `docs.description`. You can do that on your next stories, and we can modify the previous ones once we migrate to Storybook v7 (I suspect we'll be able to update at some point in Q4).

packages/wonder-blocks-popover/src/components/__docs__/popover-content.argtypes.js
jandrade: You can see the changes here: https://5e1bf4b385e3fb0020b7073c-wzaxfiexxw.chromatic.com/?path=/story/popover-popovercontent--default

jandrade: Great idea! FTR I added this in a separate PR... see: https://github.com/Khan/wonder-blocks/pull/1577/commits/b8bcee515c936c86d2e497d559c14c89d061fbd1

PR #1576, opened 2022-08-29
WB-1142.3: Move `PopoverContentCore` to Storybook

jandrade:
## Summary:
Migrating the `PopoverContentCore` component to Storybook:

- Ported the existing "Custom popover" example.
- Included two new stories to show an WB Icon example and a DetailCell
  inside a popover example.
- Modified the JSDoc block to include the "Usage" guide.

Issue: WB-1142

## Test plan:
Navigate to the Popover / PopoverContentCore section and verify that all
the stories match with the React Styleguidist version:

Styleguidist: https://wonder-blocks.netlify.app/#popovercontentcore


packages/wonder-blocks-popover/src/components/__docs__/popover-content-core.stories.js
```diff
@@ -0,0 +1,186 @@
...
+                // See https://github.com/storybookjs/storybook/issues/12596
+                excludeDecorators: true,
+            },
+        },
+    },
+    decorators: [
+        (Story: any): React.Element<typeof View> => (
+            <View style={styles.example}>{Story()}</View>
+        ),
+    ],
```
nishasy, line 47: How do you feel about adding an a small inline `argTypes` here and then setting the control type for children to `null`? I usually do this because I find the giant json object under the control panel for React component children to be unreadable and unhelpful.

Screen Shot 2022-08-29 at 2 52 33 PM

nishasy (approved):
LGTM! I did leave one small nonblocking suggestion.

PR #1578, opened 2022-08-30
Migrate ModalPanel to storybook

nishasy:
## Summary:
Using Netlify as a guideline, migrate the documentation
for ModalPanel to Storybook.

View changes here:
https://5e1bf4b385e3fb0020b7073c-fetfhuleis.chromatic.com/?path=/docs/modal-building-blocks-modalpanel--default

## Test plan:
- Test the prop values in the control panel and confirm
  they all work as expected. Confirm nothing crashes.
- Read through the documentation and confirm that there
  are no typos.
- Resize the window and confirm that the modals'
  responsive behavior is as expected.


packages/wonder-blocks-modal/src/components/__docs__/modal-panel.stories.js
```diff
@@ -0,0 +1,434 @@
...
+                                        Lorem ipsum dolor sit amet, consectetur
+                                        adipiscing elit, sed do eiusmod tempor
+                                        incididunt ut labore et dolore magna
+                                        aliqua.
+                                    </Body>
+                                    <Strut size={Spacing.large_24} />
+                                    <Button>Primary action</Button>
+                                </View>
+                            }
+                            closeButtonVisible={false}
```
nishasy, line 346: The original example in Netlify kept the default close buttons in the panels. It used the `mediaSize` from `MediaLayout` to determine when this prop should be true or false. I didn't do that here since we are moving away from `MediaLayout` and I decided it might be easier to remove the button entirely rather than figure out how to get this to work with CSS grid or adding a window resize listener something.

jeresig (approved):
Love all the examples, looks really nice, thank you!

packages/wonder-blocks-modal/src/components/__docs__/modal-panel.stories.js
```diff
@@ -0,0 +1,459 @@
...
+            control: {type: null},
+        },
+        footer: {
+            control: {type: null},
+        },
+    },
+};
+
+export const Default: StoryComponentType = (args) => (
+    <View style={styles.previewSizer}>
```
jandrade, line 143: **nit (non-blocking):** Not sure if you know this already..... as I see that these two views are being reused in all the stories (`previewSizer` and `modalPositioner`), we could also use the `decorator` setting in the default export so we could create that common wrapper for all stories in this file. No need to do it here, but might come handy in future PRs.

e.g.

```
export default {
  ...
  decorators: [
        (Story: StoryComponentType): React.Element<typeof View> => (
            <View style={styles.previewSizer}>
                <View style={styles.modalPositioner}>
                    {Story()}
               </View> 
            </View>
        ),
    ],
};

This also helps to avoid including these two decorative wrappers into the "Show code" tab.
```

jandrade (approved):
Looks great! Thanks for adding all these clear examples 👏

PR #1580, opened 2022-08-30
Add ModalHeader to storybook

nishasy:
## Summary:
Using Styleguidist as a guide, migrate the documentation
for ModalHeader to storybook.

Warning -- this element has a few weird issues as a result
of its `light` prop:

If you play around with the control
panel, you'll see that the close button is not visible on
the header when it's dark; this is because the close
button style is determined by the encompassing ModalPanel.
This is understandable because not all panels have headers,
but that does mean that we can't have a case where
only the header is dark.

Similarly, breadcrumbs are not really visible on the
dark background. It's also not easy to style the text
in breadcrumbs to be white since you can't pass a View
into the `breadcrumbs` prop, you can't style the
Breadcrumbs element, you can pass a View into the
Breadcrumbs element, you can't style the BreadcrumbsItem,
and you can't pass anything that's not a link or a
string into BreadcrumbsItem.

Storybook publish:
https://5e1bf4b385e3fb0020b7073c-rohxzexeos.chromatic.com/?path=/story/modal-building-blocks-modalheader--default

## Test plan:
- Test different configurations in the control panel.
- Read all the descriptions and confirm it makes sense
  and there are no typos.


packages/wonder-blocks-modal/src/components/__docs__/modal-header.argtypes.js
```diff
@@ -0,0 +1,76 @@
...
+import * as React from "react";
+import {
+    Breadcrumbs,
+    BreadcrumbsItem,
+} from "@khanacademy/wonder-blocks-breadcrumbs";
+import Link from "@khanacademy/wonder-blocks-link";
+
+const BreadcrumbsMappings: {[key: string]: React.Node} = {
+    none: null,
+    "lesson path": (
```
jandrade, line 11: **praise:** Ohh that's a nice touch! I don't know why I didn't use strings with the Popover mappings :).

jandrade (approved):

PR #1584, opened 2022-09-01
[fei4139.wb.1] Update to v29 of jest

somewhatabstract:
## Summary:
This updates to jest@29 and then updates the snapshots. It also deletes all the `.babelrc.js` symlinks and changes the root `.babelrc.js` to `babel.config.js`.

This last part is because Jest hasn't quite fixed their bug with snapshots and the babel config transform. Instead of using the config we pass to the transformer, it appears to look in the `cwd` (which is equivalent to the `rootDir`) and it seems it is either trying to match our filename or only looking for `babel.config.js`. Before I made this filename change, it was exhibiting issues like we used to see with flow types not getting stripped - I tried clearing cache and such and that didn't fix it, but the rename did.

So, it's better than it was, but still not quite right.

Issue: FEI-4139

## Test plan:
`yarn test`
Also try making an inline snapshot invalid and then running `yarn jest -u`.


jeresig (approved):
Very nice, thank you!

PR #1594, opened 2022-09-08
add aria-hidden to tooltip-tail

handeyeco:
Slack convo: https://khanacademy.slack.com/archives/C8Z9DSKC7/p1662657335181759

The tooltip-tail is an SVG which the OSX VoiceOver was seeing as an image. The image is decorative and isn't super useful for users using screen readers, so I'm hiding it from screen readers with [aria-hidden](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden).

I can write a unit test if that's helpful, but I figured it'd be more noise than anything. I manually tested with Storybook and the results of using aria-hidden were what I was hoping for.

Screen Shot 2022-09-08 at 12 10 08 PM

### DIFF
--- a/.changeset/sour-otters-poke.md
+++ b/.changeset/sour-otters-poke.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tooltip": patch
+---
+
+hide tooltip-tail from screen readers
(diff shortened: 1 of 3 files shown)


jandrade (approved):
LGTM! Thanks for fixing this 🚀

PR #1596, opened 2022-09-08
Add CircularSpinner to storybook

nishasy:
Add Circular Spinner to storybook using styleguidist as a foundation.

Storybook publish:
https://5e1bf4b385e3fb0020b7073c-kxhrvmtwtf.chromatic.com/?path=/docs/progressspinner-circularspinner--default

Issue: https://khanacademy.atlassian.net/browse/WB-1143

## Testing Plan
- Check for typos
- Play around with the control panel and confirm it behaves as expected


packages/wonder-blocks-progress-spinner/src/components/__docs__/circular-spinner.stories.js
```diff
@@ -0,0 +1,181 @@
...
+            <td>
+                <LabelLarge>small</LabelLarge>
+            </td>
+            <td>
+                <LabelLarge>medium</LabelLarge>
+            </td>
+            <td>
+                <LabelLarge>large</LabelLarge>
+            </td>
+        </tr>
```
kevinbarabash, line 60: **suggestion (non-blocking):** since these are headings, `th` would be a better choice than `td` here.  I think the output would look a little nicer with the headings centered.

packages/wonder-blocks-progress-spinner/src/components/__docs__/circular-spinner.stories.js
```diff
@@ -0,0 +1,181 @@
...
+    darkBackground: {
+        background: Color.darkBlue,
+        padding: Spacing.xLarge_32,
+        width: "100%",
+        alignItems: "center",
+        justifyContent: "center",
+    },
+    distanced: {
+        margin: Spacing.large_24,
+    },
```
kevinbarabash, line 172: **style nit (non-blocking):** Instead of applying this style to each spinner, you could set `borderSpacing: 24` on the table itself.

kevinbarabash (approved):
LGTM.  I had a look the chromatic link and things and was surprised to see the props documentation.  Is Storybook able to extract those docs from the comments in the source code without extra help?

packages/wonder-blocks-progress-spinner/src/components/__docs__/circular-spinner.stories.js
```diff
@@ -0,0 +1,181 @@
...
+};
+
+export const Light: StoryComponentType = () => (
+    <View style={styles.darkBackground}>
+        <CircularSpinner light={true} />
+    </View>
+);
+
+Light.parameters = {
+    docs: {
```
jandrade, line 123: **thought:** You could also set the Storybook's `darkBlue` background by including it as a key here:

```suggestion
    backgrounds: {
        default: "darkBlue",
    },
    docs: {
```

By doing this, the whole canvas takes that `darkBlue` bg.

(Sorry if I missed this in previous reviews, but just remembered that).

packages/wonder-blocks-progress-spinner/src/components/__docs__/circular-spinner.stories.js
```diff
@@ -0,0 +1,181 @@
...
+    </Body>
+);
+
+Inline.parameters = {
+    docs: {
+        storyDescription: `Circular spinners also work inline.`,
+    },
+};
+
+export const WithStyle: StoryComponentType = () => {
```
jandrade, line 142: **praise:** Great example!

jandrade (approved):
LGTM! 🚢

packages/wonder-blocks-progress-spinner/src/components/__docs__/circular-spinner.stories.js
nishasy: I'm not sure why, but this ends up looking very different 🤔 

Screen Shot 2022-09-19 at 1 48 39 PM

packages/wonder-blocks-progress-spinner/src/components/__docs__/circular-spinner.stories.js
nishasy: OOOOOOOO I like that!

PR #1597, opened 2022-09-12
[hack2022.1.useonline] Add useOnline

somewhatabstract:
## Summary:
I was tinkering with some code that needed the online state and made this hook. I can already see places in webapp where it (or an HOC that wraps it) would be used to simplify existing code, so I thought I'd put it up as a PR.

Also had a stab at expanding our docs a little, based on the format of the Wonder Blocks Data and Wonder Blocks Testing docs.

Issue: XXX-XXXX

## Test plan:
`yarn test`
`yarn flow`


packages/wonder-blocks-core/src/hooks/use-online.js
```diff
@@ -0,0 +1,29 @@
...
+        window.addEventListener("online", handleChange);
+        window.addEventListener("offline", handleChange);
+
+        return () => {
+            window.removeEventListener("online", handleChange);
+            window.removeEventListener("offline", handleChange);
+        };
+    }, [forceUpdate]);
+
+    return navigator.onLine;
```
somewhatabstract, line 28: **explanation:** Seems silly to store the value in state when it's already got a source of truth. So instead, we use force updates on the events but always use `navigator.onLine` as the actual value.

jeresig (approved):
Looks good to me! I'm curious where you see this being used?

somewhatabstract:
> Looks good to me! I'm curious where you see this being used?

I originally wrote it to support my anti-cheating work, but then I found that the latest code already does something similar. However, the approach taken in webapp stores a value of online based on the window events only, rather than having the browser be the source of truth, which I think is likely prone to issues somewhere down the line. So, in the very least, I expect to see this hook used in that specific case.

I added it to WB since it's a generic hook and I suspect that we may find it useful in some other circumstances.

PR #1599, opened 2022-09-17
Experiment with @storybook/interactions addon

kevinbarabash:
## Summary:
`@storybook/interactions` allows you to write tests using `@testing-library` within Storybook stories.  If a story has a test, it will show up in the "Interactions" panel and will be run when you view the story.  This PR also enable the interactive debugger which allows you to step through, as well as rewind, the test.

It is possible to import the tests from storybook files into our jest unit tests and run them there ~but this will require update `@testing-library/user-event` to a more recent version.  The version that `@storybook/interactions provides`, uses an async API whereas the one we're using for our tests is not.~ This currently doesn't work with the test I added in this PR.  This is b/c tests written using `@storybook/interactions` use async/await within the test and simulating events on the birthday picker doesn't work when there are pauses after each interaction.  This will likely require a fix in the dropdown component itself.  This is may be related to an issue we say when simulating events on dropdowns in cypress tests failing in certain scenarios.

The reason why `@storybook/interactions` tests use async/await is to enable stepping through the test one action at a time from the Storybook UI.

Issue: None

## Test plan:
- yarn storybook
- http://localhost:6061/?path=/story/birthdaypicker--birthday-picker-default
- step through the test

Screen Shot 2022-09-16 at 7 36 30 PM


jandrade:
Closing it in favor of #1658. We can probably revisit the BirthdayPicker later (cc @kevinbarabash).

PR #1601, opened 2022-09-19
WB-1402: FocusTrap breaks keyboard navigation when the close button is hidden

jandrade:
## Summary:
For some reason, the FocusTrap functionality is conflicting when we hide
the `Close (X)` button. The issue occurs when the user presses
`Shift+Tab` because the focus starts working inconsistently right after
that.

This PR fixes the issue by simplyfing the focus trap logic, so instead
of listening on the global focus and recursively trying to focus on the
next available element, we now focus on the sentinel elements and choose
the next available element based on the list of focusable elements
within the modal dialog.

Issue: WB-1402

## Test plan:

Navigate to the new Focus trap example: `?path=/docs/modal-modallauncher--focus-trap`

https://5e1bf4b385e3fb0020b7073c-jcywaioyxt.chromatic.com/?path=/story/modal-modallauncher--focus-trap

1. Open the modal and verify that the focus is set on the first radio option.
2. Press `Shift+Tab` and verify that the focus now moves to the `Next button`.
3. Try using the keyboard in both directions to verify that the focus
   trap is working as expected (focus never leaves the current modal).
4. Press to the `Open another modal` button and see that a new modal appears on top.
5. Use the keyboard to navigate on this modal.
6. Focus on the `Close (X)` button and press it.
7. Verify that the focus returns to the original modal and that the
   keyboard navigation continue working as expected.

https://user-images.githubusercontent.com/843075/191094490-4f0bbe24-3827-4046-9bf3-312a11e2c0f7.mov

NOTE: If you want to see the issue, you can check here https://5e1bf4b385e3fb0020b7073c-iynbeohsku.chromatic.com/?path=/story/modal-modallauncher--with-radio-group


jeresig (approved):
Interesting, well the logic definitely seems simpler, which is good! Thanks for resolving this!

PR #1602, opened 2022-09-20
Improve flow types for icons dict

kevinbarabash:
## Summary:
I typo'd a key when using the 'icons' dict in webapp, but flow didn't pick it up because of how we were typing 'icons'.  This PR changes icont-assets.js to use named exports.  I had to update some of the tests to ignore '__esModule'.  I believe this is a test-only issue and this key shouldn't show up in actual use.

Issue: None

## Test plan:
- yarn jest wonder-blocks-icon


jeresig (approved):
Nice, looks good to me!

packages/wonder-blocks-icon/src/util/icon-assets.test.js
```diff
@@ -2,13 +2,16 @@
 describe("icons", () => {
     test("default contains icons we expect", async () => {
         // Arrange
-        const importedModule = import("./icon-assets.js");
+        const importedModule = await import("./icon-assets.js");
```
somewhatabstract, line 5: **nit:** This should be in the `// Act` - an empty `// Arrange` is ok, but an empty `// Act` or `// Assert` is a smell, I think.

packages/wonder-blocks-icon/src/index.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 import Icon from "./components/icon.js";
-import icons from "./util/icon-assets.js";
+import * as icons from "./util/icon-assets.js";
```
somewhatabstract, line 3: **suggestion:** This could be changed to:

```
export * as icons from "./util/icon-assets.js";
```

And then drop the additional export on line 7.

However, since this also exports the `__esModule` export, that's going to be confusing. Instead, how about...

```
import * as iconAssets from "./util/icon-assets.js";

const {__esModule, ...icons} = iconAssets;

export {icons};
```

somewhatabstract (changes requested):
Looks good, though I think we can lose the need to worry about the `__esModule` piece by changing our `index.js`.

packages/wonder-blocks-icon/src/index.js
kevinbarabash: I love it!  I was trying to think of a way to do this this morning.  Thank you. 💚

packages/wonder-blocks-icon/src/util/icon-assets.test.js
kevinbarabash: I forgot to do this.  I'll do it just in a abit.

somewhatabstract (approved):
LGTM :)

PR #1603, opened 2022-09-20
Allow I18nInlineMarkup to be used with LabeledTextField

kevinbarabash:
## Summary:
This PR updates LabeledTextField to accept <I18nInlineMarkup> for the 'label' and 'description' props and updates field-heading.js to wrap them in the appropriate typography component.  If developers need to customize the styling they can do so by wrapping <I18nInlineMarkup> in another component.

Issue: None

## Test plan:
- yarn jest wonder-blocks-form

### DIFF
--- a/packages/wonder-blocks-form/src/components/field-heading.js
+++ b/packages/wonder-blocks-form/src/components/field-heading.js
@@ -4,6 +4,7 @@ import {StyleSheet} from "aphrodite";
 
 import {View, addStyle, type StyleType} from "@khanacademy/wonder-blocks-core";
 import Color from "@khanacademy/wonder-blocks-color";
+import {typeof I18nInlineMarkup} from "@khanacademy/wonder-blocks-i18n";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
 import Spacing from "@khanacademy/wonder-blocks-spacing";
 import {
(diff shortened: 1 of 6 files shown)


packages/wonder-blocks-form/src/components/field-heading.js
```diff
@@ -21,12 +22,12 @@ type Props = {|
     /**
      * The title for the label element.
      */
-    label: string | React.Element<Typography>,
+    label: string | React.Element<Typography | I18nInlineMarkup>,
```
jeresig, line 25: I bet there are a *ton* of places in Wonder Blocks that don't handle this right now, not just field headings - should we be updating the other ones too?

jeresig (approved):

packages/wonder-blocks-form/src/components/field-heading.js
kevinbarabash: Probably.  I can expand the scope of this PR.

kevinbarabash: I've converted this to a draft and will resubmit once I've updated all the other places.

jeresig:
Thanks Kevin, appreciate it! If landing this is blocking other work, then please just land it - but I feel like if this is a "nice to have" then let's update more things!

kevinbarabash:
I totally forgot about this PR.  I'll update it on Monday.

kevinbarabash:
#1620 is even more flexible so I'm going to close this PR in preference of that one.  I see that there's some things I need to port from this one to that one (namely the removal of the `typeof label` checks that we have in a few places).

PR #1605, opened 2022-09-21
[fei4762.1.betterrequestids] Handle non-standard objects better

somewhatabstract:
## Summary:
Although I think we really should only use primitive types within variables, there are currently some cases where we pass `Date` objects (and maybe other non-primitive object types).

Previously, the request ID generation ignored these because they went down a path that only converted them if `Object.entries` returned any entries.

Now, we do a little more effort to ensure that even if there are no results from `Object.entries`, we still get that object value represented in the ID somehow.

Issue: FEI-4762

## Test plan:
`yarn test`

### DIFF
--- a/.changeset/breezy-wasps-allow.md
+++ b/.changeset/breezy-wasps-allow.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-data": major
+---
+
+Make sure request ID generation handles objects like Date and Error instances better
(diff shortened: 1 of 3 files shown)


.changeset/breezy-wasps-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-data": major
```
somewhatabstract, line 2: **explanation:** This changes the output of `getGqlRequestId` and so I put this as a major update since that ultimately can effect the behaviour of things that rely on that for cache keys and inflight requests.

jeresig (approved):

PR #1608, opened 2022-09-23
WB-1318: Add `labels` prop to `SingleSelect` to pass translated strings

jandrade:
## Summary:
We previously created a new functionality to be able to pass translated
strings to `MultiSelect` using a `labels prop` that is used by the consumers
passing `i18n.()...` strings.

This PR is responsible for adding a similar functionality to
`SingleSelect`, but with some minor differences from `MultiSelect` (we
don't need to support all the texts multiselect does).

Also removed some WB interdependencies that are not really used by the
Dropdown package.

JIRA: https://khanacademy.atlassian.net/browse/WB-1318

## Test plan:

Verify that the `Custom Labels` story allows passing different labels.

### DIFF
--- a/packages/wonder-blocks-dropdown/package.json
+++ b/packages/wonder-blocks-dropdown/package.json
@@ -16,12 +16,10 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.18.6",
-    "@khanacademy/wonder-blocks-button": "^3.0.3",
     "@khanacademy/wonder-blocks-clickable": "^2.3.1",
     "@khanacademy/wonder-blocks-color": "^1.2.0",
     "@khanacademy/wonder-blocks-core": "^4.4.0",
     "@khanacademy/wonder-blocks-icon": "^1.2.30",
-    "@khanacademy/wonder-blocks-icon-button": "^3.4.11",
     "@khanacademy/wonder-blocks-layout": "^1.4.11",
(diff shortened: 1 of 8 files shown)


packages/wonder-blocks-dropdown/package.json
```diff
@@ -40,6 +38,7 @@
     "react-window": "^1.8.5"
   },
   "devDependencies": {
+    "@khanacademy/wonder-blocks-button": "^3.0.3",
```
jandrade, line 41: Discovered that this package is only used in SB stories.

I have a separate ticket to check for inter-dependencies that I'll be working on next sprint.

packages/wonder-blocks-dropdown/src/index.js
```diff
@@ -17,4 +18,4 @@ export {
     MultiSelect,
 };
 
-export type {Labels};
+export type {Labels, SingleSelectLabels};
```
jandrade, line 21: This still gets me thinking... should we rename `Labels` to `MultiSelectLabels` to be more explicit and differentiate with the SingleSelect ones?

jeresig: Hmm... we probably should! Might as well wait until the next major release for that, at least.

jeresig (approved):
This is great, thank you!

PR #1609, opened 2022-09-23
Remove Styleguidist deprecated docs

jandrade:
## Summary:
Now that we have migrated most of the Web WB package docs to Storybook,
we can start deprecating all the Styleguidist docs that are now fully
migrated to Storybook.

This is based on the feedback provided here:

https://khanacademy.slack.com/archives/C8Z9DSKC7/p1663694718583589

Issue: XXX-XXXX

## Test Plan

Verify that the Styleguidist (Netlify) site is updated and doesn't throw
any error.

https://deploy-preview-1609--wonder-blocks.netlify.app/


packages/wonder-blocks-popover/src/components/__tests__/popover.test.js
```diff
@@ -97,14 +100,86 @@ describe("Popover", () => {
...
         // we try to close it from inside the content
-        wrapper.find("[data-close-button]").simulate("click");
+        userEvent.click(screen.getByRole("button", {name: "close popover"}));
+
+        // Assert
+        expect(screen.queryByText("Title")).not.toBeInTheDocument();
+        expect(onCloseMock).toBeCalled();
+    });
+
+    it("should close the Popover using the default close button", () => {
```
jandrade, line 114: After removing the gen snapshot tests, I noticed that we were not covering some important unit tests, so I decided to add some here.

styleguide.config.js
```diff
@@ -21,24 +21,14 @@ module.exports = {
         {
             name: "Breadcrumbs",
             content: "packages/wonder-blocks-breadcrumbs/docs.md",
-            components: [
```
jeresig, line 24: So much moved over!

jeresig (approved):
Great, thanks for cleaning this all up!

styleguide.config.js
jandrade: Totally! this shows that we are pretty close to finish the migration 🤞

PR #1615, opened 2022-10-13
Add new Banner component to Wonder Blocks

nishasy:
## Summary:
This is the base implementation of the Banner component. It
is mostly taken from the existing implementation of Banner
in mobile Wonder Blocks.

- 8px of spacing around floating banners
- Updating the icon to be an Icon with IconAsset instead of an img

**Updated Storybook publish:**
https://5e1bf4b385e3fb0020b7073c-omnvzidaqs.chromatic.com/?path=/story/banner--default

Old Storybook publishes:
https://5e1bf4b385e3fb0020b7073c-fdyhsmnhub.chromatic.com/Ï
https://5e1bf4b385e3fb0020b7073c-meqidgklni.chromatic.com/
https://5e1bf4b385e3fb0020b7073c-aclcwsrtdq.chromatic.com/
https://5e1bf4b385e3fb0020b7073c-eytllurxkb.chromatic.com/
https://5e1bf4b385e3fb0020b7073c-gnzcmyyvts.chromatic.com/

Issue: https://khanacademy.atlassian.net/browse/WB-1384

## Test plan:
Storybook
- Open Banner in storybook.
- Check that all the controls in the control panel work.
- Check that all the descriptions make sense.
- Check that the banner looks correct using the Storybook
  ruler and outline tools.

Unit tests
- Run `yarn jest packages/wonder-blocks-banner/src/components/__tests__/banner.test.js`


packages/wonder-blocks-banner/src/components/banner.js
```diff
@@ -0,0 +1,281 @@
...
+import Button from "@khanacademy/wonder-blocks-button";
+import Color from "@khanacademy/wonder-blocks-color";
+import {View} from "@khanacademy/wonder-blocks-core";
+import {icons} from "@khanacademy/wonder-blocks-icon";
+import Link from "@khanacademy/wonder-blocks-link";
+import IconButton from "@khanacademy/wonder-blocks-icon-button";
+import Spacing from "@khanacademy/wonder-blocks-spacing";
+import {LabelSmall} from "@khanacademy/wonder-blocks-typography";
+
+type BannerKind =
```
nishasy, line 14: Note: The banner types are in a different file in the mobile repo, but I kept them in the same file here to be more consistent with other web wonder blocks components. I can move them to a different file if that is preferable.

packages/wonder-blocks-banner/src/components/banner.js
```diff
@@ -0,0 +1,281 @@
...
+ *     actions={[
+ *         {title: "Button 1", onClick: () => {}},
+ *         {title: "Button 2", onClick: () => {}},
+ *     ]}
+ *     onDismiss={() => {console.log("Has been dismissed.")}}
+ * />
+ * ```
+ */
+const Banner = (props: Props): React.Node => {
+    const {actions, onDismiss, kind, layout = "full-width", text} = props;
```
nishasy, line 134: I used `layout = "full-width"` to set "full-width" as the default value for `layout`. Should I use `type DefaultProps = {|layout: $PropertyType<Props, "full-width">|}` alongside a `static defaultProps: DefaultProps` like in other files? Would that work for functional components?

packages/wonder-blocks-banner/src/components/banner.js
```diff
@@ -0,0 +1,281 @@
...
+        >
+            <View
+                style={[
+                    styles.backgroundColor,
+                    layoutstyle,
+                    {backgroundColor: colorForKind(kind ?? "info")},
+                ]}
+            />
+            <View style={styles.containerInner}>
+                <img
```
nishasy, line 192: This `img` will be changed to a Wonder Blocks `Icon` in a following PR. I should be able to remove the svg files after that.

packages/wonder-blocks-banner/src/components/banner.js
```diff
@@ -0,0 +1,281 @@
...
+                        <LabelSmall>{text}</LabelSmall>
+                    </View>
+                    {actions && (
+                        <View style={styles.actionsContainer}>
+                            {renderActions()}
+                        </View>
+                    )}
+                </View>
+                {onDismiss ? (
+                    <IconButton
```
nishasy, line 208: In a following PR, I will add an `aria-label` prop for the dismiss button and a default aria-label otherwise.

packages/wonder-blocks-banner/src/components/__docs__/banner.stories.js
```diff
@@ -0,0 +1,355 @@
...
+            kind="critical"
+            actions={[
+                {title: "Link", href: "/"},
+                {title: "Button", onClick: () => {}},
+            ]}
+        />
+        <Strut size={Spacing.medium_16} />
+        <Banner
+            text={
+                <span>
```
nishasy, line 210: Oops I forgot to update this to a `LabelSmall`. I'll do that when I update this PR with other changes.

packages/wonder-blocks-banner/src/components/banner.js
```diff
@@ -0,0 +1,281 @@
...
+     * This is the default.
+     */
+    | "full-width";
+
+type ActionTrigger = {|
+    title: string,
+    onClick?: () => void,
+    href?: string,
+    ariaLabel?: string,
+|};
```
jandrade, line 49: **suggestion:** Now that I see, this type could be improved so we can enforce that an action item includes only a `onClick` or a `href` prop (not both at the same time). You can see what I refer to here:

https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgCCAxhgJZwB2AKgE5kDmDetYAvGAN4A+qY-Ychhh4AXGADOGepQYAaPgKoBhGGWIBrAPziAFAEp2APjAA3OGQAmCgWAAWtPFB2TpZWTYEBDel4AyXgBGeDAuUjLyqNwAvgDc6MDAYNR2ZBJglnB4EpQA5Bj2XqYELLRwtBJyYIEArgVkBRJ2cDUwlmBQXmQwYGRQYAgE7sQwNZYEGKnpxF4SeKjEVFK9ALY4cBISZIEiAMoYXhhiRKQUNPRMLOxcikIi4gBEdiHwD54qapp6hmwmnNGeBxOR4SOArPAAWjUlA0b1QcQSSQAsl4NO4GKt1pttiJXIdspiNlsdvNsPgTuQqHRGMxaAAhWYEDg8W4Ne6uCKeHxkfxBEJhNweKIIzC4AgkSnnGksADqDTsdLqGCo1xZ-AAdJqJWdqZd6Yz3pRVOoNN9jGYLNZhfFReTtVSLrS5ZM-O4NKreBqtacHdL9XNAY4oOJwuiFDEbWTxT6pXqAEzXe2xp3yxUYZWUMDcCk6x2y+WumE2hZLeqUUxeNSWJPiJO62kJ5ms4THJ4vOBw-gfE1m35cAGKIHBsAPUHgqFuuEixaUZYVqtJwgSNMZ2sx+ssRs3fh3VvPGCvQ3Gr5gAzm-4KaelsyVqyLiSF00533x1XN9ltg8dwPAkdjyHQrCl6xEAA

packages/wonder-blocks-banner/src/components/banner.js
```diff
@@ -0,0 +1,281 @@
...
+                        >
+                            {action.title}
+                        </Button>
+                    </View>
+                );
+            } else {
+                throw new Error(
+                    "An action must have an href or an onClick field.",
+                );
+            }
```
jandrade, line 172: **suggestion:** Expanding on my previous comment about the `ActionTrigger` type. If we modify the types, we can get rid of this runtime check.

jandrade (commented):
This is great progress! Including some initial comments focused on the component implementation. I'll make another pass in tests and stories in a bit and will update my review if I find something else.

packages/wonder-blocks-banner/src/components/banner.js
nishasy: Oh interesting. I feel like we should provide the option to have both an `href` and an `onClick` as there are some links that run an `onClick` before navigation. For example, links that get analytics data. I guess we could just ignore that case for now and then add support for it later if someone requests it?

packages/wonder-blocks-banner/src/components/__docs__/banner.stories.js
```diff
@@ -0,0 +1,355 @@
...
+            <Banner text="kind: warning" kind="warning" layout="floating" />
+        </View>
+        <View style={styles.spacing}>
+            <Banner text="kind: critical" kind="critical" layout="floating" />
+        </View>
+    </View>
+);
+
+DarkBackground.parameters = {
+    docs: {
```
jandrade, line 155: **suggestion:** instead of including the `darkBackground` View wrapper, we could use the SB `backgrounds` option:

```suggestion
    backgrounds: {
        default: "darkBlue",
    },
    docs: {
```

packages/wonder-blocks-banner/src/components/__docs__/banner.stories.js
```diff
@@ -0,0 +1,355 @@
...
+        ]}
+    />
+);
+
+WithLinks.parameters = {
+    docs: {
+        storyDescription: `This is a banner with links. An action, passed
+            into the \`actions\` prop, becomes a link when it has an
+            \`href\` value. It can also have an \`onClick\` value, but it
+            will be a link regardless if it navigates to a URL via \`href\`.`,
```
jandrade, line 193: This makes me think about my previous comment of `href` vs `onClick` 🤔.

packages/wonder-blocks-banner/src/components/__tests__/banner.test.js
```diff
@@ -0,0 +1,162 @@
...
+
+        // Act
+        const actionTest = () =>
+            render(<Banner text="" actions={[{title: "some button"}]} />);
+
+        // Assert
+        expect(actionTest).toThrow(
+            "An action must have an href or an onClick field.",
+        );
+    });
```
jandrade, line 161: **thought:** I'm seeing that we are not including tests for `kind`, but maybe you are planning to add them in the next PR?

jandrade (commented):
A couple of extra comments related to SB stories and tests.

packages/wonder-blocks-banner/src/components/banner.js
nishasy: I definitely like your solution way better. I just wasn't sure if there's a case that needs both href and onClick

jandrade: Here you can see an updated version that includes this valid scenario: https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgCCAxhgJZwB2AKgE5kDmDetYAvGAN4A+qY-Ychhh4AXGADOGepQYAaPgKoBhGGWIBrAPziAFAEp2APjAA3OGQAmCgWAAWtPFB2TpZWTYEBDel4AyXgBGeDAuUjLyqNwAvgDc6MDAYNR2ZBJglnB4EpQA5Bj2XqYELLRwtBJyYIEArgVkBRJ2cDUwlmBQXmQwYGRQYAgE7sQwNZYEGKnpxF4SeKjEVFK9ALY4cBISZIEiAMoYXhhiRKQUNPRMLOxcikIi4gBEdiHwD54qapp6hmwmnNGeBxOR4SOArPAAWjUlA0b1QcQSSQAsl4NO4GKt1pttiJXIdspiNlsdvNsPgTuQqHRGMxaAAhWYEDg8W4Ne6uCKeHxkfxBEJhNweKIIzC4AgkSnnGksADqDTsdLqGCo1xZ-AAdJqJWdqZd6Yz3pRVOoNN9jGYLNZhfFReTtVSLrS5ZM-O4NKreBqtacHdL9XNAY4oOJwujDcbNC4DObzFYFDEbWTxT6pXqAEzXe2pp3yxUYZWUMDcCk6x2y+WumE2xL2INgcr1o2fd1NFptaoEcZQdx4SwLJb1SimLxqSxZ8RZ3W0jPM1nCY5PF5wOEixaUZbD0dZwgSPMFicpqcsGc3fh3BfPGCvcPNs2-LgA+E2tcbkdWbcSSumku+9OqufsouV7LoGwJgA8oLglCbors+A5mG+Y4pjuX7OnYAASXiUJYIi0AekpHrQJ6cABF5LnC-BAsG4GQZC0KwjeJp3n80RPkAA

One nice thing of this approach is that we can get rid of the runtime validation that was defined in the component.

packages/wonder-blocks-banner/src/components/banner.js
```diff
@@ -0,0 +1,281 @@
...
+        position: "absolute",
+        top: 0,
+        bottom: 0,
+        left: 0,
+        right: 0,
+        opacity: 0.08,
+    },
+    containerOuter: {
+        borderInlineStartWidth: Spacing.xxSmall_6,
+        minHeight: 56,
```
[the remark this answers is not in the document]
megaroeny: Yeah in Figma, I use the 8px of vertical padding and it hugs the contents to get the height (I believe it's 56px in total for single-line). We can look at this more closely later.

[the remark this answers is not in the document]
nishasy: Oh! You're absolutely right. I believe that's left over from when I was building up the layout; it doesn't need to be there.

packages/wonder-blocks-banner/src/components/__docs__/banner.stories.js
```diff
@@ -0,0 +1,355 @@
...
+            source: {
+                // See https://github.com/storybookjs/storybook/issues/12596
+                excludeDecorators: true,
+            },
+        },
+    },
+    argTypes: BannerArgTypes,
+};
+
+const longText = `Lorem ipsum dolor sit amet, consectetur adipiscing elit.
```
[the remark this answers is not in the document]
megaroeny: Yeah. @nishasy here are some examples that I think would be generic still, but create more distinction between each kind and also a clearer example:

- Info = "This is a message about something informative like an announcement."
- Success = "This is a message about something positive or successful!"
- Warning = "This is a message warning the user about a potential issue."
- Critical = "This is a message about something critical or an error."
- etc.

What do you think?

packages/wonder-blocks-banner/src/components/banner.js
```diff
@@ -0,0 +1,281 @@
...
+ *     ]}
+ *     onDismiss={() => {console.log("Has been dismissed.")}}
+ * />
+ * ```
+ */
+const Banner = (props: Props): React.Node => {
+    const {actions, onDismiss, kind, layout = "full-width", text} = props;
+    const layoutstyle = {
+        borderRadius: layout && layout === "full-width" ? 0 : 4,
+    };
```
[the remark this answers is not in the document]
megaroeny: @jandrade yes confirmed. `floating` should have 8px margin around it, preventing the rounded corners from touching the edge of something (which looks bad).

megaroeny (changes requested):
I left some comments on a few threads in this PR. Thanks Nisha! :)

packages/wonder-blocks-banner/src/components/banner.js
```diff
@@ -0,0 +1,281 @@
...
+        alignContent: "center",
+        flexWrap: "wrap",
+        justifyContent: "space-between",
+        marginLeft: Spacing.xSmall_8,
+        marginRight: Spacing.xSmall_8,
+        paddingInlineStart: Spacing.xSmall_8,
+    },
+    labelContainer: {
+        flexShrink: 1,
+        marginTop: Spacing.small_12,
```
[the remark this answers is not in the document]
nishasy: I'm updating all the padding/margin values, so you should probably see this change go into effect.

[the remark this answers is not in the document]
megaroeny: @nishasy if this simplifies it for you, you'll see in Figma that I am applying 16px to the colored background container (in Figma it's the `background + margins` layer).

If you have that selected, you'll see these properties:
image

lillialexis:
Is this PR ready to be looked at again?

lillialexis:
Is this PR ready to be looked at again?

packages/wonder-blocks-banner/src/components/banner.js
jandrade: That's right, and that's the purpose of static types (e.g Flow, TS)..... that we give signals that something is wrong, but that at the same time we cannot stop devs from bypassing that (if they want). If they still need to bypass that, they would need to use a `$FlowIgnore` comment, and that's something we don't recommend to do (unless there's a very specific reason to do so).

packages/wonder-blocks-banner/src/components/banner.js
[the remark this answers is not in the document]
jandrade: Just to understand better the change... in order to preserve the `16px` gap, would this remain as `12` or would change to `8`?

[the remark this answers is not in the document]
megaroeny: I also [left a comment (below)](https://github.com/Khan/wonder-blocks/pull/1615#discussion_r998440447) that is related to this I believe.

nishasy:
@lillialexis yes please!

packages/wonder-blocks-banner/src/components/banner.js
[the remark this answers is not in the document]
nishasy: @jandrade I changed it to 8!

packages/wonder-blocks-banner/src/components/banner.js
kevinbarabash: For default props on functional component you should do something that looks like this:
```
const Foo = (props: Props): React.Node => {
    return null;
};

type DefaultProps = {|
   bar: string,
|};
Foo.defaultProps: DefaultProps = {bar: "hello"};

export {Foo};
```
This way when the component is used, flow make those props in `Foo.defaultProps` behave as optional props.

packages/wonder-blocks-banner/src/components/__docs__/banner.stories.js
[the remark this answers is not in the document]
lillialexis: Oh, I didn't know we weren't supposed to use lorem ipsum, but maybe that directive is just for shorter text, because, yeah, what would you use for longer text? I also like "Lorem Office-Speak-um" personally, but just because I think it's funny.

packages/wonder-blocks-banner/src/components/__docs__/banner.stories.js
```diff
@@ -0,0 +1,380 @@
...
+            into the \`actions\` prop, becomes a link when it has an
+            \`href\` value. It can also have an \`onClick\` value, but it
+            will be a link regardless if it navigates to a URL via \`href\`.`,
+    },
+};
+
+export const WithInlineLinks: StoryComponentType = () => (
+    <>
+        <Banner
+            text="Oh no! The button and link look different."
```
lillialexis, line 202: I wonder if it would be better to be even more explicit here? Like `"Oh no! The button and link on the right look different! Don't mix button and link actions."` and the next one being `"Use inline links in the body of the text instead."`

nishasy:
@lillialexis sorry, I forgot to update the storybook publish in the description after pushing the fix for that yesterday! Does it still look wrong to you?

lillialexis (approved):
No, the icon v-alignment in Storybook doesn't look wrong any more. The buttons still look big, but you're fixing that elsewhere?

packages/wonder-blocks-banner/src/components/__docs__/banner.stories.js
jandrade: I think this comment is still valid, b/c the semantics of the action will remain the same (`<a>` for this case).

packages/wonder-blocks-banner/src/components/banner.js
nishasy: I got it to work!

packages/wonder-blocks-banner/src/components/__docs__/banner.stories.js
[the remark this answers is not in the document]
megaroeny: @nishasy yeah, I think forcing it to be more narrow is also more realistic to what we would probably see. Especially on Mobile. 🙂 

The text examples I gave could be used for the "[Variants](https://5e1bf4b385e3fb0020b7073c-fdyhsmnhub.chromatic.com/?path=/story/banner--variants)" story/page you have, to further clarify their purpose/meaning.

jandrade (approved):
Looks good! thanks for making the changes based on our suggestions 💯

PR #1617, opened 2022-10-20
Make `layout` a required prop on Banner

nishasy:
## Summary:
- Updated stories to show that devs should manually add spacing around floating banners, it is _not_ part of the banner itself.
- Made `layout` a required prop with no default.

I actually got rid of the "Variants" story because it
starting looking really awkward and I didn't think it was
worth updating. Instead, I added a specific "Kinds" story
and a specific "Layouts" story.

- Updating the icon to be an Icon with IconAsset instead of an img

**Storybook publish:**
https://5e1bf4b385e3fb0020b7073c-jannpabsfz.chromatic.com/?path=/docs/banner--default

**Chromatic:**
https://www.chromatic.com/pullrequest?appId=5e1bf4b385e3fb0020b7073c&number=1617

## Test plan:
- Check the stories at http://localhost:6061/?path=/docs/banner--layouts
- Make sure the descriptions make sense and have no typos.
- Confirm that the stories look correct.


packages/wonder-blocks-banner/src/components/banner.js
```diff
@@ -256,6 +262,10 @@ const styles = StyleSheet.create({
         flexDirection: "row",
         padding: Spacing.xSmall_8,
     },
+    containerFloating: {
+        padding: Spacing.xSmall_8,
```
nishasy, line 266: I opted to use padding instead of margins because it was much more consistent with the spacing. Margins kept collapsing.

packages/wonder-blocks-banner/src/components/banner.js
```diff
@@ -224,6 +224,12 @@ const Banner = (props: Props): React.Node => {
             </View>
         </View>
     );
+
+    return layout === "full-width" ? (
+        bannerBody
+    ) : (
+        <View style={styles.containerFloating}>{bannerBody}</View>
+    );
```
nishasy, line 232: I was thinking of just adding a container around the whole thing instead of only in the floating case, but that required adding two different styles and switching between them depending on layout. I think that was also a completely find way of doing this, and it's not a big change if people would prefer using that method instead.

jandrade (approved):
LGTM! Thanks for splitting the stories to make it more concise 💯

megaroeny (approved):
Great approach leveraging the dark blue background to show the difference in margins/layouts. ✨ Thank you!

lillialexis (changes requested):
I thought we weren't going to do this, and leave it up to the consumer of the Banner to add the space. This is because, what if the containing layout already has padding, and by us forcing more padding, it gets pushed in too much?

jandrade:
> I thought we weren't going to do this, and leave it up to the consumer of the Banner to add the space. This is because, what if the containing layout already has padding, and by us forcing more padding, it gets pushed in too much?

@lillialexis IIRC this was part of the plan, specially that works great for site banners (I think that was the main driver for creating this kind of variant). @megaroeny could confirm this.

For the consumer having already padding, then it would probably be better if they use the `full-width` variant.

lillialexis:
>For the consumer having already padding, then it would probably be better if they use the full-width variant.

But then the corners aren't rounded.

It was the plan that they would always have the padding, but that was supposed to be in the designs only. Designers weren't supposed to give developers a design with rounded-corner banners that didn't have at least 8 px padding. But even in the mobile app right now, I have a screen that has a container that has built-in padding so that all the components are away from the edge of the screen. If I added the 8px padding to Banner on top of that, it would be pushed in a further 8 px from the rest of the things, and then the consumer would have to add -8px to compensate.

I don't think this should be built-in to the code. Just a rule for designers.

nishasy:
Sharing the slack thread here in case anyone ends up needing to follow this conversation:

https://khanacademy.slack.com/archives/C4PE1QM5Y/p1666378904188279

megaroeny (changes requested):
Think you removed the `border-radius` on some of the stories with the `floating` variant. Besides that, looks good and glad we can keep some examples with the margin still (for a guideline)

megaroeny (approved):

packages/wonder-blocks-banner/src/components/__docs__/banner.stories.js
```diff
@@ -50,107 +51,120 @@ Default.args = {
...
-                />
-            </View>
-        </View>
-        <View style={styles.variantColumn}>
-            <LabelMedium>Floating layout</LabelMedium>
-            <View style={styles.spacing}>
-                <Banner
-                    text="kind: info - This is a message about something informative like an announcement."
-                    kind="info"
+                    text="This banner has floating layout. Space has been added around it."
```
lillialexis, line 134: I would add to this sentence to say that space has been added around it by an outside component that isn't part of the Banner itself. And that we always want the space added around it, as indicated by a notation in the designs, but that it's up to the developer to ensure that this rule is adhered to on the screen... or something like that... maybe not in the Banner itself, but in a note below it? I do see it somewhat explained in the docs, but are those on a tab that someone might not see unless clicked? It's also not _super_ obvious that it's a separate view adding the padding. Maybe make that view have a dashed-line border to really call it out?

lillialexis: OMG!! I read the code and added the comment before clicking into the Storybook!!!! IGNORE ME!!! YOU DID THIS! 🤦‍♀️

lillialexis: Yeah, the "docs text" is right there above the story, but I would _maaaaybe_ add a tiny bit to that text blob that does explicitly call out that the "spacing added" is a separate view managed by the developer? Because, to me, that's the critical difference. Not just that there should be a space, but that it has to manually added. 😄

lillialexis (approved):
This is great! But before you land this, can you please change the Title/Description of the PR to reflect what the PR is really doing? Because, now, the critical changes here are:
- Banner layout is now required, with no default, and
- The spacing added around `floating` is _not_ part of the Banner, but instead added to the Stories as a separate thing to call out that it's needed

PR #1618, opened 2022-10-20
Use wonder blocks Icon instead of img in Banner

nishasy:
## Summary:
- Use Wonder Blocks Icon for the icons in Banner.
- Remove the use of `img` for icons as well as the svg files.
- Add aria-label to icons
- Add tests for kind

This PR should get rid of the code coverage complaints.

**Storybook publish:**
https://5e1bf4b385e3fb0020b7073c-opbliurcxg.chromatic.com/

### A11y Demo
https://user-images.githubusercontent.com/13231763/197080461-752b974f-09bc-4848-92b8-68bd76d2fc03.mov

## Test plan:
- Go to http://localhost:6061/?path=/story/banner--kinds
- Use a screenreader and confirm the icons are labeled correctly.

`yarn jest packages/wonder-blocks-banner/src/components/__tests__/banner.test.js`


packages/wonder-blocks-banner/src/components/banner-icons.js
```diff
@@ -0,0 +1,18 @@
+// @flow
```
nishasy, line 1: lol this is not a generated file

packages/wonder-blocks-banner/src/components/__tests__/banner.test.js
```diff
@@ -146,4 +146,61 @@ describe("Banner", () => {
...
+
+        // Act
+        render(<Banner text="" />);
+
+        // Assert
+        const icon = screen.getByTestId("banner-kind-icon");
+        expect(icon).toHaveAttribute("aria-label", "info");
+    });
+
+    test("info kind displays info icon", () => {
```
kevinbarabash, line 163: **style suggestion (non-blocking):** You could use `it.each` to dedupe these tests.

packages/wonder-blocks-banner/src/components/banner.js
```diff
@@ -196,10 +203,12 @@ const Banner = (props: Props): React.Node => {
                 ]}
             />
             <View style={styles.containerInner}>
-                <img
-                    src={iconForKind(kind ?? "info")}
-                    alt={kind}
-                    className={css(styles.icon)}
+                <Icon
+                    icon={iconForKind(kind ?? "info")}
```
kevinbarabash, line 207: If you align the names of the exports from banner-icons.js with `BannerKind`, you could drop the `iconForKind` banner.  It would look something like:
```
import * as bannerIcons from "./banner-icons.js";

...

<Icon icon={bannerIcons[kind]} ... />
```

packages/wonder-blocks-banner/src/components/banner-icons.js
```diff
@@ -0,0 +1,18 @@
...
+import {type IconAsset} from "@khanacademy/wonder-blocks-icon";
+
+export const infoIcon: IconAsset = {
+    medium: "M12 4C7.58172 4 4 7.58172 4 12C4 16.4183 7.58172 20 12 20C16.4183 20 20 16.4183 20 12C20 7.58172 16.4183 4 12 4ZM2 12C2 6.47715 6.47715 2 12 2C17.5228 2 22 6.47715 22 12C22 17.5228 17.5228 22 12 22C6.47715 22 2 17.5228 2 12ZM10.25 11.25C10.25 10.6977 10.6977 10.25 11.25 10.25H12C12.5523 10.25 13 10.6977 13 11.25V15.5315C13.4313 15.6425 13.75 16.034 13.75 16.5C13.75 17.0523 13.3023 17.5 12.75 17.5H12C11.4477 17.5 11 17.0523 11 16.5V12.2185C10.5687 12.1075 10.25 11.716 10.25 11.25ZM12.9375 7.875C12.9375 8.49632 12.4338 9 11.8125 9C11.1912 9 10.6875 8.49632 10.6875 7.875C10.6875 7.25368 11.1912 6.75 11.8125 6.75C12.4338 6.75 12.9375 7.25368 12.9375 7.875Z",
+};
```
kevinbarabash, line 6: **question:** this icon looks similar to the "info" icon from `wonder-blocks-icon`.  Is it the same?  Could we reuse that icon here?

packages/wonder-blocks-banner/src/components/banner-icons.js
```diff
@@ -0,0 +1,18 @@
...
+
+export const criticalIcon: IconAsset = {
+    medium: "M12 4C7.58172 4 4 7.58172 4 12C4 16.4183 7.58172 20 12 20C16.4183 20 20 16.4183 20 12C20 7.58172 16.4183 4 12 4ZM2 12C2 6.47715 6.47715 2 12 2C17.5228 2 22 6.47715 22 12C22 17.5228 17.5228 22 12 22C6.47715 22 2 17.5228 2 12ZM12 6.5C12.5523 6.5 13 6.94772 13 7.5V12.75C13 13.3023 12.5523 13.75 12 13.75C11.4477 13.75 11 13.3023 11 12.75V7.5C11 6.94772 11.4477 6.5 12 6.5ZM13.125 16.125C13.125 16.7463 12.6213 17.25 12 17.25C11.3787 17.25 10.875 16.7463 10.875 16.125C10.875 15.5037 11.3787 15 12 15C12.6213 15 13.125 15.5037 13.125 16.125Z",
```
kevinbarabash, line 17: **suggestion (non-blocking):** It seems like these other icons could be useful outside of wonder-blocks-banner.  Maybe we could define them in that module and then consume them in this one.

kevinbarabash (approved):
LGTM. 🎉  Thanks for including a video of manual testing with Voice Over. 💚

packages/wonder-blocks-banner/src/components/banner-icons.js
nishasy: Ah, I believe that's true. I think I might keep it this way anyway just for the sake of consistency, and because we're switching to Phosphor icons soon and I'm not sure how that will affect our existing icons.

jandrade: It's totally good to keep it this way. One small thing that could be helpful is adding a TODO comment pointing to the JIRA ticket to mention that this will be migrated over to Phosphor.

Here's the ticket to track that: https://khanacademy.atlassian.net/browse/WB-1409

nishasy: Ah yes, I do have a comment about that in `banner.js`, but I can add it to `banner-icons.js` too.

packages/wonder-blocks-banner/src/components/banner-icons.js
nishasy: I don't know if it's worth considering doing this until after we start using Phosphor icons

packages/wonder-blocks-banner/src/components/banner-icons.js
jandrade: yes please! I think it would be helpful to see that this will change in the (near) future.

packages/wonder-blocks-banner/src/components/banner-icons.js
jandrade: +1 as we will be able just to get rid of this file after we officially support Phosphor.

PR #1620, opened 2022-10-21
WB-1430: Update form component types to accept React.Node

kevinbarabash:
## Summary:

label, description, and error props on components in wonder-blocks-form are too restrictive (sometimes accepting just strings).  Sometimes designs call for markup in these props such as links or bolded words.  This PR makes these props accept any React.Node so that developers can implement designs without having to use workarounds.

## Test Plan:
- yarn jest
- yarn flow

Screen Shot 2022-10-21 at 4 00 38 PM

### DIFF
--- a/packages/wonder-blocks-form/src/components/__docs__/labeled-text-field.stories.js
+++ b/packages/wonder-blocks-form/src/components/__docs__/labeled-text-field.stories.js
@@ -528,6 +528,27 @@ CustomStyle.parameters = {
     },
 };
 
+export const WithMarkup: StoryComponentType = (args) => {
+    return (
+        <LabeledTextField
+            {...args}
+            label="Name"
+            description={
+                <span>
+                    Please enter your <strong>name</strong>
+                </span>
+            }
+        />
+    );
+};
+
+WithMarkup.parameters = {
+    docs: {
+        storyDescription: `\`LabeledTextField\`'s \`label\` and \`description\` props
+        can accept \`React.Node\`s.`,
+    },
+};
+
 export const Ref: StoryComponentType = () => {
     const [value, setValue] = React.useState("Khan");
     const inputRef = React.createRef();
--- a/packages/wonder-blocks-form/src/components/__docs__/labeled-text-field.stories.js
+++ b/packages/wonder-blocks-form/src/components/__docs__/labeled-text-field.stories.js
@@ -9,6 +9,7 @@ import Color from "@khanacademy/wonder-blocks-color";
 import Spacing from "@khanacademy/wonder-blocks-spacing";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
 import Button from "@khanacademy/wonder-blocks-button";
+import Link from "@khanacademy/wonder-blocks-link";
 
 import type {StoryComponentType} from "@storybook/react";
 
@@ -535,7 +536,8 @@ export const WithMarkup: StoryComponentType = (args) => {
             label="Name"
             description={
                 <span>
-                    Please enter your <strong>name</strong>
+                    Description with <strong>strong</strong> text and a{" "}
+                    <Link href="/path/to/resource">link</Link>
                 </span>
             }
         />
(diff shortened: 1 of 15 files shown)


erikhelal (commented):
thank you!!

packages/wonder-blocks-form/src/components/__docs__/labeled-text-field.stories.js
```diff
@@ -528,6 +528,27 @@ CustomStyle.parameters = {
...
 
+export const WithMarkup: StoryComponentType = (args) => {
+    return (
+        <LabeledTextField
+            {...args}
+            label="Name"
+            description={
+                <span>
+                    Please enter your <strong>name</strong>
+                </span>
```
jandrade, line 539: **suggestion:** It would be better including an example using WB components.

e.g.
```
<LabelSmall>Please enter your <strong>name</strong><LabelSmall>
```

or

```
<View>something custom...</View>
```

packages/wonder-blocks-form/src/components/__docs__/labeled-text-field.stories.js
```diff
@@ -528,6 +528,27 @@ CustomStyle.parameters = {
...
+            }
+        />
+    );
+};
+
+WithMarkup.parameters = {
+    docs: {
+        storyDescription: `\`LabeledTextField\`'s \`label\` and \`description\` props
+        can accept \`React.Node\`s.`,
```
jandrade, line 548: **suggestion:** It'd be helpful adding a bit more context.... something like: .... `This is helpful when you need to decorate or use specific elements in your form field (e.g. including Popovers, Tooltips or emphasized text)`.

jandrade (approved):
Looks good to me! A couple of small suggestions related to the stories, but the implementation looks solid. Thanks for making this change to make folks easier to use these components 🚀

packages/wonder-blocks-form/src/components/__tests__/checkbox-group.test.js
```diff
@@ -1,85 +1,138 @@
...
+                        selectedValues={[]}
+                    >
+                        <Choice label="a" value="a" aria-labelledby="test-a" />
+                        <Choice label="b" value="b" aria-labelledby="test-b" />
+                        <Choice label="c" value="c" aria-labelledby="test-c" />
+                    </CheckboxGroup>,
+                );
+
+            // Assert
+            expect(action).not.toThrow();
```
jandrade, line 110: **suggestion:** Based on the test description, I think than rather than verifying that the component renders, we should probably check that the `label` is rendered (e.g. `screen.getByText(/label with.../i).toBeInTheDocument`). wdyt?

packages/wonder-blocks-form/src/components/checkbox-group.js
```diff
@@ -136,19 +132,15 @@ export default class CheckboxGroup extends React.Component<CheckboxGroupProps> {
             <StyledFieldset data-test-id={testId} style={styles.fieldset}>
                 {/* We have a View here because fieldset cannot be used with flexbox*/}
                 <View style={style}>
-                    {typeof label === "string" ? (
+                    {label && (
                         <StyledLegend style={styles.legend}>
                             <LabelMedium>{label}</LabelMedium>
```
jandrade, line 137: **thought:** Probably out of scope for this PR, but I noticed that we are rendering a `span` here, and there would probably be an opportunity here to use `<label />` instead. Again, this could be an improvement that can be done in the future.

jandrade (approved):
Looks great! Thanks for porting the tests to RTL. I left a couple of comments in the tests for you to consider.

packages/wonder-blocks-form/src/components/checkbox-group.js
jandrade: oh wait, this is the general `label` for the group 🤦. Please disregard this comment.

kevinbarabash: Good idea.  I can do this in a separate PR.

PR #1622, opened 2022-10-21
Update wonder-blocks-clickable tests to use @testing-library

kevinbarabash:
## Summary:
I'd like to add a `useClickable` hook and update `ClickableBehavior` to use it.  This will allow use to get rid of `getClickable` since we'll be able to determine if there's a router or not by simply using `useHistory`.  Before we can do any of that though, we need to convert all of our enzyme tests to testing library because enzyme (our at least the version we're on) doesn't support testing components using hooks.

The reason why I want to create a `useClickable` hook is that I was going through the "Advanced React Patterns" workshop that's part of Epic React and learned about hooks that return a props object.  That seemed pretty similar to what `ClickableBehavior` does expect it's a full blown component that provides props for its children to spread into some other component.

Updating all of the tests to use `@testing-library` is going to take some time though.  I'm going to spread this work out and do a little bit here and there when I have time.  I'll open an epic to track this.

Issue: None

## Test plan:
- yarn jest clickable

### DIFF
--- a/.changeset/flat-snails-tease.md
+++ b/.changeset/flat-snails-tease.md
@@ -1,5 +1,2 @@
 ---
-"@khanacademy/wonder-blocks-clickable": patch
 ---
-
-Remove uses of enzyme from wonder-blocks-clickable tests
--- a/.changeset/flat-snails-tease.md
+++ b/.changeset/flat-snails-tease.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-clickable": patch
+---
+
+Remove uses of enzyme from wonder-blocks-clickable tests
--- a/packages/wonder-blocks-clickable/src/components/__tests__/clickable-behavior.test.js
+++ b/packages/wonder-blocks-clickable/src/components/__tests__/clickable-behavior.test.js
@@ -1,13 +1,14 @@
+/* eslint-disable testing-library/prefer-user-event */
 /* eslint-disable max-lines */
 // @flow
 import * as React from "react";
-import {render, screen} from "@testing-library/react";
+import {render, screen, fireEvent} from "@testing-library/react";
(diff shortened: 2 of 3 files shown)


.changeset/flat-snails-tease.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-clickable": patch
+---
+
+Remove uses of enzyme from wonder-blocks-clickable tests
```
jandrade, line 5: **suggestion:** These changes don't require to release a new version, mostly because we are not making any changes on the code released to users (only testing changes).

I think that for that case, you can just use `yarn changeset --empty` so the GH action won't complain.

packages/wonder-blocks-clickable/src/components/__tests__/clickable-behavior.test.js
```diff
@@ -21,6 +22,20 @@ const wait = (delay: number = 0) =>
         return setTimeout(resolve, delay);
     });
 
+const labelForState = (state: ClickableState): string => {
```
jandrade, line 25: **praise:** This is great! love how states got moved to the DOM so RTL can do what it does best!.

jandrade (approved):
Thank you for migrating these tests! 💯

.changeset/flat-snails-tease.md
kevinbarabash: Good call.  I'll do that for all of the test-only changes.

packages/wonder-blocks-clickable/src/components/__tests__/clickable-behavior.test.js
kevinbarabash: Thankfully this was pretty easy to do since the component renders children.

PR #1624, opened 2022-10-25
Add a11y features to banners

nishasy:
## Summary:
- Add role values to banner
- Add aria-live values to banner
- Add an aria label for the dismiss button
- Add an aria label to the icons
- Add tests for accessibility

**Storybook publish:**
https://5e1bf4b385e3fb0020b7073c-jnsslrwmuu.chromatic.com/?path=/docs/banner--default

**Chromatic:**
https://www.chromatic.com/build?appId=5e1bf4b385e3fb0020b7073c&number=2396

### Demo: Icon and button aria labels
https://user-images.githubusercontent.com/13231763/197654765-8a0b8ed4-d94c-41f4-b5b5-06db9dba13ed.mov

### Demo: Different icon labels for different icons
https://user-images.githubusercontent.com/13231763/197654799-5723ddb2-75a6-45c6-be75-937b8908dfcc.mov

### Demo: assertive role makes the screen reader read out banner on render
https://user-images.githubusercontent.com/13231763/197654885-cc3b5c20-9d22-4cbc-83b3-97fca6054673.mov

## Test plan:
- yarn jest `packages/wonder-blocks-banner/src/components/__tests__/banner.test.js`

Storybook
- Go to Storybook at http://localhost:6061/?path=/story/banner--default
- Go through all the stories and use a screenreader with all of them
- Confirm that the icons and buttons have accessible labels
- Use the rotor to search by role. Confirm that banners show up under their expected roles.


packages/wonder-blocks-banner/src/components/banner.js
```diff
@@ -88,20 +94,38 @@ type Props = {|
...
+                color: Color.gold,
+                role: "alert",
+                ariaLive: "polite",
+            };
         case "critical":
-            return Color.red;
+            return {
+                color: Color.red,
+                role: "alert",
+                ariaLive: "assertive",
```
jandrade, line 122: **nit:** I think `assertive` is defined by default with the `alert` role, so there might not be need to add this key here.

For reference: https://www.w3.org/TR/wai-aria-1.1/#alert

> Elements with the role alert have an implicit [aria-live](https://www.w3.org/TR/wai-aria-1.1/#aria-live) value of assertive, and an implicit [aria-atomic](https://www.w3.org/TR/wai-aria-1.1/#aria-atomic) value of true.

packages/wonder-blocks-banner/src/components/__tests__/banner.test.js
```diff
@@ -179,4 +179,162 @@ describe("Banner", () => {
             expect(icon).toHaveAttribute("aria-label", kind);
         },
     );
+
+    // Test accessibility
```
jandrade, line 183: **praise:** All these tests are amazing! 👏

packages/wonder-blocks-banner/src/components/banner.js
```diff
@@ -88,20 +94,38 @@ type Props = {|
      * If present, dismiss button is on right side. If not, no button appears
      */
     onDismiss?: ?() => void,
+
+    /**
+     * The accessible label for the dismiss button.
+     * Please pass in a translated string.
+     */
+    dismissAriaLabel: string,
```
jandrade, line 102: **suggestion:** probably you have to specify this in the `argTypes.js` file.

jandrade (approved):
Looks great!! I left a couple of suggestions for you to consider. Great job on Banners in general 👏🚢🚀

packages/wonder-blocks-banner/src/components/banner.js
nishasy: Oh, good catch! Thanks

PR #1629, opened 2022-10-26
Update channel name when posting Slack message in Release workflow

kevinbarabash:
## Summary:
We changed the channel name from wonder-blocks-eng to wonder-blocks-dev, but hadn't updated it in the Release workflow so it failed when did a release.

Issue: None

## Test plan:
- cross fingers and wait for the next release


.changeset/giant-beers-grow.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
kevinbarabash, line 2: I'm going to try this again.  I think there was an issue with me creating an "empty" file after first creating a non-empty file. 🤷‍♂️

.github/workflows/release.yml
```diff
@@ -90,7 +90,7 @@ jobs:
         uses: rtCamp/action-slack-notify@v2
         env:
           SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
-          SLACK_CHANNEL: wonder-blocks-eng
```
somewhatabstract, line 93: **question:** I wonder if we should move these to configurable environment vars so that we don't have to ship code changes just to udpate them?

kevinbarabash: That's a great idea!  I'll do that.

kevinbarabash: I ended up having to add it a secret.

jandrade (approved):

PR #1631, opened 2022-10-27
WB-1430.1: Allow falsy children in CheckboxGroup and RadioGroup

kevinbarabash:
## Summary:
As I was upgrading components in webapp to use v3 of wonder-blocks-form (which removes Radio and requires RadioGroup), I noticed that in some places we provide conditional children.  Neither RadioGroup or CheckboxGroup handle this case, but our dropdown components do.  This PR updates RadioGroup and CheckboxGroup to allow falsy children (and filter them out) when rendering.  This makes them consistent with the dropdown components in terms of how we deal with children.

Issue: WB-1430

## Test plan:
- yarn jest wonder-blocks-form
- yarn start:storybook
- check that the new stories I added render without their last child

### DIFF
--- a/.changeset/stupid-starfishes-heal.md
+++ b/.changeset/stupid-starfishes-heal.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-form": minor
+---
+
+Allow CheckboxGroup and RadioGroup to accept falsy children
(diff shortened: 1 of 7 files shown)


jandrade (approved):
Looks great! Thanks for adding this improvement 🚀

PR #1633, opened 2022-10-27
Remove default aria-label from WB Clickable

nishasy:
## Summary:
Currently, if no aria label is provided, Clickable components
have `aria-label=""`. Instead of having an empty aria label,
it should just not have the aria label attribute.

- Remove the default value of aria-label in Wonder Blocks Clickable
- Add tests for aria label values

Issue: https://khanacademy.atlassian.net/browse/WB-1429

## Test plan:
`yarn jest packages/wonder-blocks-clickable/src/components/__tests__/clickable.test.js`

Storybook
- Go to https://khan.github.io/wonder-blocks/?path=/story/clickable-clickable--basic
- Inspect the <a> element.
- Observe that it has an empty aria-label attribute.
- Now go to http://localhost:6061/?path=/story/clickable-clickable--basic
- Confirm there is no aria-label by default when inspecting the element

Before (prod storybook):
Screen Shot 2022-10-27 at 2 35 13 PM

After (local storybook):
Screen Shot 2022-10-27 at 2 35 26 PM

### DIFF
--- a/.changeset/real-rocks-attack.md
+++ b/.changeset/real-rocks-attack.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-clickable": patch
+---
+
+Remove the default value of aria-label ""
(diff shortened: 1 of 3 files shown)


jeresig (approved):
Ah ha - thanks for fixing this!

jandrade (approved):
LGTM! thanks for adding the tests :shipit:

PR #1635, opened 2022-10-28
WB-1430.2: Don't wrap cloned Choices in fragments

kevinbarabash:
## Summary:
Converting children to an array in #1631 cause issues in webapp with React complaining about children not having a 'key' prop.  We do set the 'key' prop on the Choice, but React doesn't see this because the Choice is wrapped in a Fragment.  The wrapper isn't necessary and removing it should fix the issue in webapp.

Issue: WB-1430

## Test plan:
- let CI run the tests

### DIFF
--- a/.changeset/clean-cameras-admire.md
+++ b/.changeset/clean-cameras-admire.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-form": patch
+---
+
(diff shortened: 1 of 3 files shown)


jandrade (approved):
Makes sense to me. I don't understand why we had that Fragment in the first place, but glad you fixed it! 🚀

PR #1638, opened 2022-11-03
Add 'useRenderState' hook to wonder-blocks-core

kevinbarabash:
## Summary
This hook provides access to the value of RenderStateContext which differentiates between initial renders (server and rehydration) to standard renders (on the client after rehydration).  This information can be useful in helping the client avoid rehydration errors.

This work was motivated by https://github.com/Khan/webapp/pull/9508.

## Test Plan:
- let CI run its checks
- yarn start:storybook, check that the docs are appearing as intended

Screen Shot 2022-11-02 at 10 40 31 PM

### DIFF
--- a/packages/wonder-blocks-core/src/__docs__/exports.use-render-state.stories.mdx
+++ b/packages/wonder-blocks-core/src/__docs__/exports.use-render-state.stories.mdx
@@ -1,0 +1,25 @@
+import {Meta} from "@storybook/addon-docs";
+
+<Meta
+    title="Core / Exports / useRenderState()"
+    parameters={{
+        chromatic: {
+            disableSnapshot: true,
+        },
+    }}
+/>
+
+# useRenderState()
+
+```ts
+function useRenderState(): RenderState;
+```
+
+The `useRenderState` hook will return either:
+- `RenderState.Initial` if the component is being rendered on the server **or** it's
+  the initial rehydration render on the client.
+- `RenderState.Standard` if the component renders on the client after the initial
+  rehydration.
+
+NOTE: Although the `RenderState` enum has a third state `Root`, this value is never
+returned by `useRenderState`.
--- a/packages/wonder-blocks-core/src/components/render-state-root.js
+++ b/packages/wonder-blocks-core/src/components/render-state-root.js
@@ -6,8 +6,9 @@ import * as React from "react";
 // https://github.com/import-js/eslint-plugin-import/issues/2073
 // eslint-disable-next-line import/named
 import {RenderState, RenderStateContext} from "./render-state-context.js";
+import {useRenderState} from "../hooks/use-render-state.js";
 
-const {useContext, useEffect, useState} = React;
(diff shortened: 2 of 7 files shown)


packages/wonder-blocks-core/src/hooks/__tests__/use-render-state.test.js
```diff
@@ -0,0 +1,61 @@
+// @flow
+import * as React from "react";
+import {renderHook as renderHookOnServer} from "@testing-library/react-hooks/server";
+import {renderHook} from "@testing-library/react-hooks";
```
kevinbarabash, line 4: I like this a little better than what we did in some of the other hook tests.  It makes it clearer which tests are server-side tests vs. client-side tests.  It also makes the tests more consistent b/c they're rendering the hook using the same API.

packages/wonder-blocks-core/src/components/render-state-root.js
```diff
@@ -6,8 +6,9 @@ import * as React from "react";
 // https://github.com/import-js/eslint-plugin-import/issues/2073
 // eslint-disable-next-line import/named
 import {RenderState, RenderStateContext} from "./render-state-context.js";
+import {useRenderState} from "../hooks/use-render-state.js";
```
kevinbarabash, line 9: I updated existing uses of `useContext(RenderStateContext)` which should hopefully save some bytes.

packages/wonder-blocks-core/src/__docs__/exports.use-render-state.stories.mdx
```diff
@@ -0,0 +1,25 @@
...
+```
+
+The `useRenderState` hook will return either:
+- `RenderState.Initial` if the component is being rendered on the server **or** it's
+  the initial rehydration render on the client.
+- `RenderState.Standard` if the component renders on the client after the initial
+  rehydration.
+
+NOTE: Although the `RenderState` enum has a third state `Root`, this value is never
+returned by `useRenderState`.
```
kevinbarabash, line 25: I included this because I thought people might be confused by seeing `Root` in the list of auto-completions for `RenderState` without it being mentioned in the docs.

packages/wonder-blocks-core/src/components/render-state-root.js
```diff
@@ -49,9 +47,10 @@ export const RenderStateRoot = ({
...
-// Flow doesn't know about defaultProps on functional components like this,
-// and it doesn't seem worth the effort to teach it.
-// $FlowIgnore[prop-missing]
+// We can set `defaultProps` on a functional component if we move the `export` to appear
+// afterwards.
 RenderStateRoot.defaultProps = {
     throwIfNested: true,
 };
+
+export {RenderStateRoot};
```
kevinbarabash, line 56: This is unrelated to the other changes, but I thought I may as well fix this since I was in the neighborhood.

jeresig (approved):
Thanks for adding this!

PR #1640, opened 2022-11-03
Add a dependency to WB Link in WB Banner

nishasy:
## Summary:
I missed a dependency to Wonder Blocks Link in Wonder Blocks Banner.
I'm adding it here.

Issue: none

## Test plan:
N/A


packages/wonder-blocks-banner/package.json
```diff
@@ -21,6 +21,7 @@
     "@khanacademy/wonder-blocks-core": "^4.4.0",
     "@khanacademy/wonder-blocks-icon": "^1.2.31",
     "@khanacademy/wonder-blocks-icon-button": "^3.4.12",
+    "@khanacademy/wonder-blocks-link": "^3.8.10",
```
jandrade, line 24: **suggestion:** It looks like the latest version is `3.8.12`:

For reference: https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-link/package.json#L3

```suggestion
    "@khanacademy/wonder-blocks-link": "^3.8.12",
```

jandrade (approved):
LGTM! Just one minor change before landing.

PR #1642, opened 2022-11-04
Fix Banner border

nishasy:
## Summary:
Currently, the inner part of the banner has a border radius on
the left when it shouldn't. The fix here is to use the
`border-start-end-radius` and `border-end-end-radius` style properties.
By using these instead of right/left, we can support RTL languages too.

Issue: none

**Storybook:**
https://5e1bf4b385e3fb0020b7073c-fbmdtatpxs.chromatic.com/?path=/story/banner--default&args=layout:floating

### Before:
image

### After:
Screen Shot 2022-11-04 at 3 37 19 PM

## Test plan:
Look at storybook and zoom in a lot
http://localhost:6061/?path=/story/banner--simple


kevinbarabash (approved):
LGTM 🎉  Thanks for spotting this issue and fixing it. 💚

PR #1644, opened 2022-11-07
Add ariaLabel prop to Banner

nishasy:
## Summary:
There are [cases](https://github.com/Khan/webapp/blob/master/services/static/javascript/site-banners/shared/base-banner.jsx#L58
) of legacy banners in webapp that have
screen reader only headers. We can add this functionality
in WB Banners by adding an aria label prop.

https://user-images.githubusercontent.com/13231763/200427306-d89ea648-d45d-4349-8304-334c02982597.mov

Issue: none

## Test plan:
Storybook
- Go to the "With Dismissal" story.
- Dismiss the banner.
- Turn on the screenreader.
- Press the button to make the banner come back.
- Confirm that the screenreader reads that aria label,
  and then the rest of the contents of the banner.


packages/wonder-blocks-banner/src/components/banner.js
```diff
@@ -68,6 +68,12 @@ type BannerValues = {|
 |};
 
 type Props = {|
+    /**
+     * Accessible label for the banner.
+     * This is read out before the other contents of the banner.
+     */
+    ariaLabel?: string,
```
kevinbarabash, line 75: **suggestion:** Some of our other components use `aria-label` so we should probably be consistent with those.

packages/wonder-blocks-banner/src/components/__tests__/banner.test.js
```diff
@@ -340,4 +340,25 @@ describe("Banner", () => {
...
         expect(banner).toHaveAttribute("aria-live", "polite");
     });
+
+    test("ariaLabel prop becomes aria-label on the banner", () => {
+        // Arrange
+
+        // Act
+        render(
+            <Banner
+                text=""
```
kevinbarabash, line 350: **question:** why is `text` empty?

kevinbarabash (changes requested):
Thanks for adding this prop.  It'll make it easier to use in `webapp`.  Requesting a change to the name of the prop so that it's more consistent with other components that already have this prop.  See inline comments for details.

packages/wonder-blocks-banner/src/components/__tests__/banner.test.js
nishasy: oh it doesn't have to be. I can add something. I figured the text wasn't important since this is inside a unit test.

kevinbarabash (approved):
LGTM :shipit:

PR #1645, opened 2022-11-07
WB-1145.1: Migrate wonder-blocks-toolbar to Storybook

jandrade:
## Summary:
Moved all the current `wonder-blocks-toolbar` examples and documentation
defined in Styleguidist to Storybook Docs.

https://wonder-blocks.netlify.app/#section-toolbar

Issue: WB-1145

## Test plan:

Navigate to the Toolbar section in Storybook and verify that the
examples are working as expected.


packages/wonder-blocks-toolbar/src/components/__docs__/toolbar.argtypes.js
```diff
@@ -0,0 +1,61 @@
...
+import {icons} from "@khanacademy/wonder-blocks-icon";
+import IconButton from "@khanacademy/wonder-blocks-icon-button";
+import {Strut} from "@khanacademy/wonder-blocks-layout";
+import Link from "@khanacademy/wonder-blocks-link";
+import Spacing from "@khanacademy/wonder-blocks-spacing";
+import {LabelLarge} from "@khanacademy/wonder-blocks-typography";
+
+type Mappings = {[key: string]: React.Node};
+
+export const leftContentMappings: Mappings = {
```
nishasy, line 14: I love these!

packages/wonder-blocks-toolbar/src/components/__docs__/toolbar.argtypes.js
```diff
@@ -0,0 +1,61 @@
...
+    ),
+};
+
+const buttonStyle = {width: 160};
+
+export const rightContentMappings: Mappings = {
+    none: null,
+    primaryButton: <Button>Submit</Button>,
+    tertiaryButton: <Button kind="tertiary">Import...</Button>,
+    nextVideoButton: <Button>Next Video</Button>,
```
nishasy, line 33: Do you think you could add an example with a light button for the dark background case?

packages/wonder-blocks-toolbar/src/components/__docs__/toolbar.stories.js
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * Small toolbar with multiple left side buttons.
+ */
+export const Small: StoryComponentType = Template.bind({});
+
+Small.args = {
+    size: "small",
+    leftContent: leftContentMappings.multipleContent,
+    rightContent: rightContentMappings.tertiaryButton,
```
nishasy, line 81: Ooo nice! Good idea to reuse these.

packages/wonder-blocks-toolbar/src/components/__docs__/toolbar.stories.js
```diff
@@ -0,0 +1,199 @@
...
+};
+
+const styles = StyleSheet.create({
+    example: {
+        padding: Spacing.large_24,
+        alignItems: "center",
+        justifyContent: "center",
+    },
+    darkContainer: {
+        backgroundColor: "#0a2a66",
```
nishasy, line 50: Is this supposed to be the old dark blue? I have a feeling the background color was written in the .md file to match our old dark blue but then we updated `Color.darkBlue`. Maybe we should update this to use `Color.darkBlue` too?

Also, would it make sense to use `background` under parameters instead?

```
Dark.parameters = {
    backgrounds: {
        default: "darkBlue",
    },
}
```

nishasy (approved):
Nice!! I only left a couple small suggestions.

One consideration though: When I used the `darkBlue` background in storybook, I realized that the outline doesn't show up for any of the toolbars because the gray is too dark. It looks like even in the Dark story where you have color set to "dark" there's no outline around the toolbar. Is this something we should consider looking into?

Screen Shot 2022-11-07 at 3 34 19 PM

packages/wonder-blocks-toolbar/src/components/__docs__/toolbar.argtypes.js
jandrade: Good idea! I'll add it to the "dark" example.

packages/wonder-blocks-toolbar/src/components/__docs__/toolbar.stories.js
jandrade: hmmm yeah good point! I'll use `darkBlue` instead + include the light button to illustrate better this example.

jandrade: @nishasy I updated the dark example following your suggestions. wdyt?

https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=636a90f833a0e76923978f37

nishasy (approved):
Awesome! Thanks for updating the dark examples

PR #1646, opened 2022-11-08
Remove default tabIndex 0 from ClickableBehavior

nishasy:
## Summary:
Currently, ClickableBehavior includes a default tabIndex of 0.
This causes tabIndex to show up even if it's wrapping an element
that is already keyboard navigable, like buttons or links.
As this is redundant, we should not have a tabIndex in this case.

As a first pass to fix this issue, this PR is giving ClickableBehavior
an optional `tabIndex` prop and removing `tabIndex` from defaults.
This also includes updating how the `tabIndex` prop is passed through
elements that use ClickableBehavior (buttons, icon buttons, and links), as
well as updating OptionItem to have a tabIndex of 0.

For more information about possible fixes, see:
https://khanacademy.atlassian.net/wiki/spaces/WB/pages/2093678742/Possible+Fix+for+Redundant+tabIndex

Issue: https://khanacademy.atlassian.net/browse/WB-1404

## Test plan:
Storybook
- Go through all the clickable elements and navigate using
  the keyboard to confirm they recieve keyboard focus.
  - Buttons
  - Links
  - Icon Buttons
  - All dropdown components
  - ClickableBehavior

`yarn jest`


packages/wonder-blocks-clickable/src/components/__tests__/clickable-behavior.test.js
```diff
@@ -297,16 +297,40 @@ describe("ClickableBehavior", () => {
...
+                {(state, childrenProps) => {
+                    return (
+                        <button data-test-id="test-button-1" {...childrenProps}>
+                            Label
+                        </button>
+                    );
+                }}
+            </ClickableBehavior>,
+        );
+
```
jandrade, line 323: **suggestion:** I think it would be useful to include another test where we include a non-interactive element and pass in `tabIndex=0` then verify if it can be focused to ensure that the tabIndex is set correctly (something similar to the `WithTabIndex` story might work).

packages/wonder-blocks-clickable/src/components/clickable-behavior.js
```diff
@@ -88,6 +88,8 @@ type CommonProps = {|
 
     skipClientNav?: boolean,
 
+    tabIndex?: number,
```
jandrade, line 91: **suggestion:** It would be helpful to add a comment explaining this prop here. I see that you added it in `argtypes.js`, but it's also helpful to add it in the implementation context.

Side note: I hope we can get rid of these comments duplication in the future 🤞

jandrade (approved):
Looks great! Thanks for all the changes in the different packages 🏅.

I left a couple of comments but nothing blocking before landing.

packages/wonder-blocks-clickable/src/components/__docs__/clickable-behavior.stories.js
```diff
@@ -72,12 +72,98 @@ Default.parameters = {
     },
 };
 
+export const WrappingButton: StoryComponentType = (args) => {
+    const ClickableBehavior = getClickableBehavior();
+    const StyledButton = addStyle("button");
+
+    return (
+        <ClickableBehavior role="button" {...args}>
```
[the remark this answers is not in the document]
nishasy: Oh oops! Sorry, that's left over from when I copied the code from another story.

kevinbarabash (approved):
Thanks for updating ClickableBehavior so that it isn't setting `tabIndex=0` by default.

I see that some of the snapshots were updated, but I think it would be even better if we had a [consistency test](https://github.com/Khan/wonder-blocks/blob/main/consistency-tests/__tests__/clickables.test.js) to ensure that we aren't setting `tabIndex` on buttons, links, etc.  Could you add one before landing this PR?

I was a little confused by this part of the PR summary:
> This also includes updating the use of tabIndex in updating elements
that use ClickableBehavior (buttons, icon buttons, and links), as
well as updating OptionItem to pass in a tabIndex of 0.

It sounds kind of like buttons, icons buttons, and links will be passed a tabIndex of 0.

nishasy:
@kevinbarabash re-requesting a review because I'm not sure I wrote the consistency tests how you intended.

consistency-tests/__tests__/clickables.test.js
```diff
@@ -156,3 +163,133 @@ describe.each`
...
+    });
+});
+
+// These components do not wrap an inherently clickable component, so
+// it is expected that they have a tabIndex of 0 so they can be keybaord
+// navigable.
+describe.each`
+    Component     | name
+    ${OptionItem} | ${"OptionItem"}
+`("$name", ({Component, name}) => {
```
kevinbarabash, line 204: **suggestion (non-blocking):** This could be combined with the previous test by introducing an additional column `hasTabIndex` with boolean values.  This variable could then be used to decide what assertion to use in the test case.

consistency-tests/__tests__/clickables.test.js
```diff
@@ -156,3 +163,133 @@ describe.each`
...
+        );
+
+        // Assert
+        const checkbox = screen.getByTestId(
+            "checkbox-choice-clickable-test-id",
+        );
+        expect(checkbox).not.toHaveAttribute("tabIndex");
+    });
+
+    test("doesn't have a redunant tabIndex of 0 (radio)", () => {
```
kevinbarabash, line 252: **suggestion (non-blocking):** You could use `test.each` to dedupe this test case with the previous one.

consistency-tests/__tests__/clickables.test.js
```diff
@@ -156,3 +163,133 @@ describe.each`
...
+                onChange={jest.fn()}
+                testId="checkbox-clickable-test-id"
+            />,
+        );
+
+        // Assert
+        const checkbox = screen.getByTestId("checkbox-clickable-test-id");
+        expect(checkbox).not.toHaveAttribute("tabIndex");
+    });
+});
```
kevinbarabash, line 295: **suggestion (non-blocking):** this is already being covered by `Choice doesn't have a redunant tabIndex of 0 (checkbox)` and could be removed.

kevinbarabash (approved):
Thanks for adding consistency tests for the changes in this PR. 🎉  I've left a few non-blocking suggestions.

consistency-tests/__tests__/clickables.test.js
nishasy: talked about this offline and decided not to do it this way ✅

PR #1648, opened 2022-11-11
Add custom icon Story (with styles)

jandrade:
## Summary:

The 321 team came with a specific use case where they needed to use a Phosphor
icon that contains a couple of extra SVG attributes.

They came up with a clever solution, and while looking at it, I noticed that we
don't have an example for custom icons.

This story is responsible for showcasing how to include a custom path in the
`Icon` component and also support custom SVG attributes via CSS styles.

Thanks to @adinardi for the example.

Issue: XXX-XXXX

## Test plan:

Screen Shot 2022-11-11 at 11 03 28 AM

Verify that the /?path=/docs/icon-icon--custom-icon looks as expected.


kevinbarabash (approved):
LGTM 🎉  No action req'd for this PR, but I wonder if we should also provide away to set the width/height of coordinates used for the icon.  I heard that the phosphor icons paths were 256px by 256px (or something like that) and that doesn't align with any of our sizes.

PR #1650, opened 2022-11-17
Add comment explaining overflow hidden style on Banner

nishasy:
## Summary:
I learned from a CSS course that you should pretty much
always add a clarifying comment for why an
`overflow: hidden` is necessary. So here we go!

Issue: none

## Test plan:
N/A


jandrade (approved):
Looks great! Thanks for this improvement 🎉

PR #1651, opened 2022-11-21
Bug Fix: Toolbar height can become unconstrained in Support Bar in webapp

kevinbarabash:
## Summary:
In certain circumstances (Chrome + ZND?) the dev support bar height can become quite large.  This component renders the wonder-blocks Toolbar and that appears to be the source of the issue.  I fixed it in the browser by editing the CSS to set the 'max-height' to match the 'min-height'.  This PR implements this same fix.

Issue: None

Screenshot 2022-11-15 at 2 11 46 PM

## Test plan:
- Let Chromatic run
- Deploy changes to webapp and see if people are still experiencing this after that

### DIFF
--- a/.changeset/shy-grapes-melt.md
+++ b/.changeset/shy-grapes-melt.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-toolbar": patch
+---
+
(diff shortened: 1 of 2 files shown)


jeresig:
It sounds like you're abandoning this approach?

PR #1652, opened 2022-11-21
WB-1146.1: Port `Tooltip` to Storybook

jandrade:
## Summary:
- Port the `Tooltip` and `TooltipContent` components from `wonder-blocks-tooltip` to Storybook.
- Removed some stories that will be reintroduced in a separate PR using a different approach..
- Added more documentation to the stories.

Issue: WB-1146

Test plan

Verify that the stories look correct in Storybook (Docs).


jeresig (approved):
Looks great - thanks for porting this!

PR #1653, opened 2022-11-21
WB-1146.2: Remove Tooltip styleguidist docs

jandrade:
## Summary:

Removing the styleguidist docs now that these are migrated to Storybook.

Issue: WB-1146

## Test plan:

Verify that in the Styleguidist site, the Tooltip docs now point to Storybook.


jeresig (approved):
Great!

PR #1654, opened 2022-11-28
Handle `aria-describedby` in `OnePaneDialog` and `ModalDialog`, and remove unused aria props from Flow types

timmcca-be:
## Summary:
This allows us to add attributes like `aria-describedby` to the modal. Note that these props were already in the types, but did nothing.

Issue: TP-7580

## Test plan:
Open `ModalLauncher`'s "navigation with focus trap" story in Safari with VoiceOver enabled and verify that it reads the body text

### DIFF
--- a/packages/wonder-blocks-modal/src/components/__tests__/one-pane-dialog.test.js
+++ b/packages/wonder-blocks-modal/src/components/__tests__/one-pane-dialog.test.js
@@ -68,4 +68,21 @@ describe("OnePaneDialog", () => {
         // Assert
         expect(screen.getByLabelText("Breadcrumbs")).toBeInTheDocument();
     });
+
+    it("applies aria-describedby to the modal", () => {
+        // Arrange
+        render(
+            <OnePaneDialog
+                title="unused"
+                content={<p id="description">cool dialog</p>}
+                aria-describedby="description"
+            />,
+        );
+
+        // Act
+        const modal = screen.getByRole("dialog");
+
+        // Assert
+        expect(modal).toHaveDescription(/cool dialog/i);
+    });
 });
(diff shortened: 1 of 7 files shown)


packages/wonder-blocks-modal/src/components/modal-dialog.js
```diff
@@ -94,9 +86,8 @@ export default class ModalDialog extends React.Component<Props> {
                             {below && <View style={styles.below}>{below}</View>}
                             <View
                                 role={role}
+                                {...ariaProps}
```
jandrade, line 89: **suggestion:** It would be better to only pass what is needed. There are a lot of ARIA attributes that we surely won't need here.

jandrade (changes requested):
Requesting changes to revert these changes and only pass a subset of `ARIA` attributes.

Also, could you please add `aria-describedby` here?
https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-modal/src/components/__docs__/one-pane-dialog.argtypes.js

packages/wonder-blocks-modal/src/components/__tests__/one-pane-dialog.test.js
```diff
@@ -68,4 +68,21 @@ describe("OnePaneDialog", () => {
...
+                content={<p id="description">cool dialog</p>}
+                aria-describedby="description"
+            />,
+        );
+
+        // Act
+        const modal = screen.getByRole("dialog");
+
+        // Assert
+        expect(modal).toHaveDescription(/cool dialog/i);
```
jandrade, line 86: **praise:** Neat! TIL about the `toHaveDescription` matcher.

packages/wonder-blocks-modal/src/components/modal-dialog.js
```diff
@@ -48,6 +47,21 @@ type Props = {|
      * Test ID used for e2e testing.
      */
     testId?: string,
+
+    /**
+     * The label for this dialog, if applicable.
+     */
+    "aria-label"?: string,
```
jandrade, line 54: **suggestion:** just a reminder about removing this prop (as you already suggested in Slack)

jandrade (approved):
Looks awesome! Thank you so much for implementing the changes and looking at the `aria-label` case 👏

PR #1659, opened 2022-12-06
Add support for custom actions to Banner

nishasy:
## Summary:
Sometimes the actions need more stuff than the current
implementation allows them to do (for example: button
with spinner, button with beforeNav).

Here, Banner actions can now have custom actions using
the "custom" type with a node passed in.

Issue: none

## Test plan:
http://localhost:6061/?path=/docs/banner--with-mixed-actions

yarn jest packages/wonder-blocks-banner/src/components/__tests__/banner.test.js

image

### DIFF
--- a/packages/wonder-blocks-banner/src/components/__docs__/banner.stories.js
+++ b/packages/wonder-blocks-banner/src/components/__docs__/banner.stories.js
@@ -417,15 +417,17 @@ export const WithCustomAction: StoryComponentType = () => (
 
 WithCustomAction.parameters = {
     docs: {
-        storyDescription: `There are a number of other props that
-            Buttons and Links may have that are not currently supported by
-            the \`actions\` prop in Banner. These would require the use of
-            custom actions. If it absolutely necessary to have a custom
-            action, it can be done by passing in an object into the \`actions\`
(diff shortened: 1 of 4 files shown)


packages/wonder-blocks-banner/src/components/__docs__/banner.stories.js
```diff
@@ -393,6 +393,83 @@ WithDismissal.parameters = {
...
+        ]}
+    />
+);
+
+WithCustomAction.parameters = {
+    docs: {
+        storyDescription: `There are a number of other props that
+            Buttons and Links may have that are not currently supported by
+            the \`actions\` prop in Banner. These would require the use of
+            custom actions. If it absolutely necessary to have a custom
```
jandrade, line 423: **suggestion:** Even that you mention here that this can be done (if necessary), I feel that it would be important to state as the first phrase that this is discouraged (unless it's the last resource - not sure about the exact wording about that). wdyt?

nishasy: How about this?

image

lillialexis (approved):
Yeah, love it!

PR #1661, opened 2022-12-07
Fix the text of a popover story button

aag:
## Summary:
Previously, the button that opens the popover in the "Popover With
initialFocusId" story referenced a "Next" button in the popover, which
doesn't exist. This commit changes the button text to reference the "It
is focused!" button, which does exist.

Issue: none

## Test plan:
- Run `npm run start:storybook`.
- View the Popover > Popover > With initialFocusId story.
- Verify that the button says: Open with initial focus on the "It is
  focused!" button.
- Click the button.
- Verify that the "It is focused!" button in the popover has focus.

Before change:
popover_before_change

After change:
popover_after_change

### DIFF
--- a/.changeset/chilled-pens-fetch.md
+++ b/.changeset/chilled-pens-fetch.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-popover": patch
+---
+
+Fix button text in the Popover With initialFocusId Storybook story
--- a/.changeset/chilled-pens-fetch.md
+++ b/.changeset/chilled-pens-fetch.md
@@ -1,5 +1,2 @@
 ---
-"@khanacademy/wonder-blocks-popover": patch
 ---
-
-Fix button text in the Popover With initialFocusId Storybook story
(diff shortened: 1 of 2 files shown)


.changeset/chilled-pens-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-popover": patch
+---
+
+Fix button text in the Popover With initialFocusId Storybook story
```
jandrade, line 5: Could you please change this file to have an empty message?

The short answer is that Storybook/Test only changes shouldn't be generating a new version of the WB package as this is not affecting the bundle that is later downloaded in webapp. This allow us to avoid publishing changes that are not required in npm.

```suggestion
---
---
```

jandrade: Apologies for not mentioning this before. We are planning to revamp the WB contribution docs in Confluence sometime in January or February to make all of these clearer.

PR #1662, opened 2022-12-13
Fix banner stories options

nishasy:
## Summary:
There was an error and the dropdown was glitching whenever
selecting something from the `actions` or `onDismiss` dropdown
in the props panel for Banner.

This is fixed by adding `Object.keys` for the `options` field
and adding a `mapping` field.

Issue: https://khanacademy.atlassian.net/browse/WB-1455

## Test plan:
- Go to http://localhost:6061/iframe.html?args=kind:success;layout:floating&id=banner--default&viewMode=story
- Open the console
- Confirm there are no errors
- Go to http://localhost:6061/?path=/story/banner--default
- Open the console
- Change the dropdown values for `actions` and `onDismiss`
- Confirm that does no cause any errors


packages/wonder-blocks-banner/src/components/__docs__/banner.argtypes.js
```diff
@@ -60,7 +60,8 @@ export default {
         description: `Links or tertiary Buttons that appear to the right of the
             text.\n\nThe ActionTrigger must have either an onClick or an href
             field, or both.`,
-        options: actionsMappings,
+        options: (Object.keys(actionsMappings): Array<React.Node>),
```
jandrade, line 63: It looks like there's a flow error here.

jandrade (approved):
Looks great! thanks for fixing this 💯

PR #1664, opened 2022-12-19
Adds an "opened" prop to make Tooltips controllable by their parent.

handeyeco:
## Summary:
I copied the "controlled by parent" logic from Popover to make Tooltip optionally controllable.

Issue: WB-1411

## Test plan:
- added unit tests
- added a story to try out the functionality

I wanted to add better tests, but I had a little trouble with RTL (I think because of Portals) and I don't know Enzyme very well. I was trying to something like the following example. I never got it to work so I thought maybe someone could give me feedback if they thought it was important.

``` JS
        test("can be closed programmatically", async () => {
            // Arrange
            function StatefulTooltipWrapper({children}) {
                const [opened, setOpened] = React.useState(false);
                console.log(opened);

                return children(opened, setOpened);
            }

            const buttonRef = await new Promise((resolve) => {
                const nodes = (
                    <StatefulTooltipWrapper>
                        {(opened, setOpened) => (
                            <View>
                                <Tooltip
                                    id="tooltip"
                                    content="Content"
                                    opened={opened}
                                >
                                    Anchor
                                </Tooltip>
                                <button
                                    ref={resolve}
                                    onClick={() => setOpened(!opened)}
                                >
                                    Toggle
                                </button>
                            </View>
                        )}
                    </StatefulTooltipWrapper>
                );
                mount(nodes);
            });

            const node = (ReactDOM.findDOMNode(buttonRef): any);
            node && node.click();
            jest.runOnlyPendingTimers();

            // Act
            // Flow doesn't like jest mocks
            // $FlowFixMe[prop-missing]
            const result = TooltipBubble.mock.instances[0].props["children"];

            // Assert
            expect(result).toMatchSnapshot(
                `Similar to <TooltipContent>Content</TooltipContent>`,
            );
        });
```


packages/wonder-blocks-tooltip/src/components/__tests__/tooltip.test.js
```diff
@@ -399,4 +399,54 @@ describe("Tooltip", () => {
             });
         });
     });
+
+    describe("Controlled", () => {
```
kevinbarabash, line 403: I can convert these tests cases to use `@testing-library` whenever I get around to landing my PR.

kevinbarabash (approved):
LGTM.  Thanks for adding this prop to `Tooltip`.  There's a failing test that will need to be fixed before this can be landed.

packages/wonder-blocks-tooltip/src/components/__docs__/tooltip.stories.js
```diff
@@ -262,6 +262,40 @@ TooltipOnButtons.parameters = {
...
+            <Tooltip
+                content="You opened the tooltip with a button"
+                opened={opened}
+            >
+                tooltip
+            </Tooltip>
+            <Button onClick={() => setOpened(!opened)}>{buttonText}</Button>
+        </View>
+    );
+};
```
kevinbarabash, line 283: I had a look at this story in Chromatic and I think it would be better to start the story with `opened` set to `true`.

handeyeco:
The test was unrelated to my changes; they resulted in a year change.

1. The test was expecting that typing `"20"` would autocomplete to `"2022"` in the birthday picker. However it's 2023 now.
2. Also the birthday picker doesn't allow for future dates, so changing `"2022"` to the current year wasn't enough since the month was July (so the test was inputting `"2023-07-05"` which is the future).

I figured the safest thing to do was to just use last year instead.

kevinbarabash (approved):
LGTM.  This is good to go now.

PR #1665, opened 2022-12-20
Update tooltip.test.js to use @testing-library

kevinbarabash:
## Summary:
We'd like to use @testing-library for all of our tests eventually so that we can start using hooks in wonder-blocks (enzyme doesn't support them).  This PR updates only one of the test suites in wonder-blocks-tooltip.  I've removed some of the tests that were checking the rendering of 'title' and 'content' b/c they didn't seem particularly high-value especially now that we're using Chromatic to do snapshot tests of this component.

Issue: None

## Test plan:
- let the tests run on CI


jandrade (approved):
LGTM! thanks for porting this to RTL. 💯 

The only remaining part is to include the empty changeset before landing (`yarn changeset --empty`).

packages/wonder-blocks-tooltip/src/components/__tests__/tooltip.test.js
```diff
@@ -32,234 +32,144 @@ describe("Tooltip", () => {
...
-                `Similar to <HeadingSmall>Title</HeadingSmall>`,
-            );
-            expect(result.props["children"]).toMatchSnapshot(
-                `Similar to <HeadingSmall>Some custom content</HeadingSmall>`,
-            );
+            expect(tooltip).toBeInTheDocument();
         });
 
-        test("with title, overrides title of TooltipContent", async () => {
+        it("should show the tooltip on focusin", async () => {
```
handeyeco, line 74: This has the same name as the previous test. Maybe "closes tooltip on unhover"?

packages/wonder-blocks-tooltip/src/components/__tests__/tooltip.test.js
```diff
@@ -32,234 +32,144 @@ describe("Tooltip", () => {
...
+                    <Tooltip title="Title" content="Content">
+                        <View>Anchor</View>
+                    </Tooltip>
+                </View>,
+            );
+
+            const node = screen.getByText("Anchor");
+            userEvent.hover(node);
+            jest.runOnlyPendingTimers();
+            userEvent.unhover(node);
```
handeyeco, line 87: We've been talking about this in Perseus. In this test, do you think we should assert that the tooltip shows before asserting that it disappears? There's been conversation about whether you should have that assertion in the "disappears" test or whether it's enough that the preceding test is asserting that the thing shows.

I worry about false positives (ie "it's not in the document, not because it left the DOM, but because it never appeared"), but obviously the previous test would catch this.

handeyeco (approved):
Thank you again for looking at this!

kevinbarabash:
@jandrade I'm not sure why the chromatic screenshots changed.

kevinbarabash:
I self approved the Chromatic changes.  It was a couple of tooltips that were shifted horizontally a bit a partially missing hair-line.

PR #1666, opened 2022-12-20
Remove manual pre-publish script

jandrade:
## Summary:
We moved from manual releases to automated releases some time ago (using
changesets).

Looking at the code, I noticed that we never cleaned up the pre-publish
manual script completely. In this PR, I remove the script and the
related dependencies.

Issue: XXX-XXXX

Test plan:

Verify that all the checks still pass.


utils/publish/pre-publish-check.js
```diff
@@ -1,76 +0,0 @@
...
-        let warnings = false;
-
-        for (const pkgPath of pkgPaths) {
-            const pkgJson = require(path.relative(__dirname, pkgPath));
-
-            checkPublishConfig(pkgJson);
-            checkPackageMain(pkgJson);
-            checkPackageModule(pkgJson);
-            checkPackageSource(pkgJson);
-            warnings = checkPackagePrivate(pkgJson);
```
kevinbarabash, line 44: Does our `changeset` setup do these checks for us?

kevinbarabash (approved):
LGTM.  Thanks for cleaning this up. 🎉

utils/publish/pre-publish-check.js
jandrade: Kinda... we have a separate `pre-publish-check-ci.js` file that was created for that case, but I probably forgot to remove this non-ci file (that was the previous way we used to do releases).

PR #1667, opened 2022-12-20
WB-1344: Fix tertiary button underlines

jandrade:
## Summary:
This fixes the underline on tertiary buttons:
- The underline was not being applied to the button when it was
disabled+focused.
- The `pressed` state was not being applied to the button properly. This
  fix now applies the `1px` border to the button when it is pressed.

Issue: WB-1344

## Test Plan:

Navigate to the Button stories and verify that all the tertiary buttons
work as expected. This means:

1. The tertiary+disabled+focused button should include an underline of
   1px (grayed out).
2. The tertiary+pressed button should display an underline of 1px as
   well.

https://user-images.githubusercontent.com/843075/208765748-4819c8c0-083e-4a44-a30d-6ab1cb499a12.mov


packages/wonder-blocks-button/src/components/button-core.js
```diff
@@ -381,14 +380,26 @@ const _generateStyles = (color, kind, light, iconWidth, size) => {
...
                 color: light ? fadedColor : offBlack32,
                 cursor: "default",
+                ":focus": {
+                    ":after": {
+                        content: "''",
+                        position: "absolute",
+                        height: 2,
+                        width: `calc(100% - ${iconWidth}px)`,
+                        right: 0,
+                        bottom: `calc(50% - 11px)`,
```
kevinbarabash, line 398: **question:** could we do `bottom: -1` here as well so that it's consistent with the `active` style?

kevinbarabash (approved):
LGTM.  Thanks for fixing this issue. 🎉

packages/wonder-blocks-button/src/components/button-core.js
jandrade: ahh good question. I tried to do that originally, but the issue is that in that case the underline is applied to the `button` instead of the internal `span`, so the line is shown too far from the `span` element. Note that this is a specific case for `tertiary` buttons. I'll add a comment there to explain all of this.

PR #1668, opened 2022-12-20
Updated aria-roles

drewkiimon:
## Summary:
- Created funky script to scrape roles without manually inputting them
- Updated `roles` in associated file

 Scrape roles by using terminal
```
    const getAriaRoles = () => {
        const nodes = document.querySelectorAll("#index_role .role-reference");

        let roles = [];
        for (let i = 0; i <= nodes.length - 1; i ++) {
            roles.push(nodes[i].innerHTML);
        }

        roles = roles.map((value) => value.substring(6, value.indexOf("</code>")));
        roles = [...new Set(roles)].sort();
        return '\'' + roles.join('\'|\'') + '\'';
    }

    copy(getAriaRoles())
```

Issue: none

## Test plan:
- N/A


drewkiimon:
@jandrade A lot of automated tests here compared to Webapp. Let me know what you think! I put in the script that I used to grab the values.

jandrade:
@drewkiimon This PR check is failing: https://github.com/Khan/wonder-blocks/actions/runs/3751292386/jobs/6387567301

One reason might be the comment script you added, but I'm not fully sure about it.

jandrade:
Update: It looks like merging `main` would fix that issue.

packages/wonder-blocks-core/src/util/types.js
```diff
@@ -11,31 +11,54 @@ export type StyleType =
...
+            roles.push(nodes[i].innerHTML);
+        }
+
+        roles = roles.map((value) => value.substring(6, value.indexOf("</code>")));
+        roles = [...new Set(roles)].sort();
+        return '\'' + roles.join('\'|\'') + '\'';
+    }
+
+    copy(getAriaRoles())
+*/
```
jandrade, line 31: This is great! but I feel this would be better having it in the PR description instead of the code itself.

drewkiimon:
Thanks for the PR @jandrade ! I removed the comment and put it into this PR instead, and merged `main` into my branch.

packages/wonder-blocks-core/src/util/types.js
```diff
@@ -16,26 +16,31 @@ type roles =
     | "alertdialog"
     | "application"
     | "article"
+    | "articles"
```
jandrade, line 19: This role `articles` does not exist.

```suggestion
```

packages/wonder-blocks-core/src/util/types.js
```diff
@@ -16,26 +16,31 @@ type roles =
...
     | "banner"
     | "button"
+    | "cell"
     | "checkbox"
     | "columnheader"
     | "combobox"
-    | "command" // abstract role
+    | "command"
```
jandrade, line 26: Could you please leave this line as it was before? I think it's important to state that this is an abstract role and it shouldn't be included in any custom implementation.

`| "command" // abstract role`

> Abstract roles are used for the ontology. Authors MUST NOT use abstract roles in content.

More about it here: https://www.w3.org/TR/wai-aria/#abstract_roles
https://www.w3.org/TR/wai-aria/#isAbstract

packages/wonder-blocks-core/src/util/types.js
```diff
@@ -16,26 +16,31 @@ type roles =
...
     | "complementary"
-    | "composite" // abstract role
+    | "composite"
```
jandrade, line 28: Same here:

```suggestion
    | "composite" // abstract role
```

packages/wonder-blocks-core/src/util/types.js
```diff
@@ -16,26 +16,31 @@ type roles =
...
     | "form"
     | "grid"
     | "gridcell"
     | "group"
     | "heading"
     | "img"
-    | "input" // abstract role
-    | "landmark" // abstract role
+    | "input"
+    | "landmark"
```
jandrade, line 43: Here too... `... // abstract role`.

packages/wonder-blocks-core/src/util/types.js
```diff
@@ -50,40 +55,44 @@ type roles =
...
     | "navigation"
+    | "none"
     | "note"
     | "option"
     | "presentation"
     | "progressbar"
     | "radio"
     | "radiogroup"
-    | "range" // abstract role
+    | "range"
```
jandrade, line 65: same here: 
```suggestion
    | "range" // abstract role
```

packages/wonder-blocks-core/src/util/types.js
```diff
@@ -50,40 +55,44 @@ type roles =
...
     | "region"
-    | "reletype" // abstract role
+    | "roletype"
```
jandrade, line 67:
```suggestion
    | "roletype" // abstract role
```

jandrade (changes requested):
Thanks for working on this. Requesting changes to revert some of the lines that were updated and mainly to remove an invalid role (`articles`).

kevinbarabash:
This is superseded by #1780.  It uses React's `AriaRole` type.  Please note that that type does not include abstract roles like `landmark`, but it does allow any string to be use.  See https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v16/index.d.ts#L1741-L1811 for the full definition of the `AriaRole`.

PR #1679, opened 2022-12-22
Stop dropdown box from shifting position on change from empty to filled state

Myranae:
## Summary:
A bug was discovered where the dropdown box alignment shifts when the selected option changes from a blank option to an option with text. This PR adds a story to test the behavior of the dropdown box when nested in text as an inline-block, plus adds CSS for inline-block dislay and middle vertical alignment to styles.menuWrapper. As a result, the alignment no longer shifts when an option with text replaces an empty option.

With inline-block setting, before vertical alignment
image

With inline-block setting, after vertical alignment
image

Issue: LP-13114

## Test plan:
- confirm all checks pass
- confirm story works as expected
- confirm other stories work as expected

## Questions:
-> Should there be a changeset for this? If so, is the summary I added to the changeset acceptable?
-> Will changing the display to inline-block for all dropdowns interfere with design decisions?

[LP-13114]: https://khanacademy.atlassian.net/browse/LP-13114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/__docs__/single-select.stories.js
+++ b/packages/wonder-blocks-dropdown/src/components/__docs__/single-select.stories.js
@@ -150,6 +150,7 @@ const items = [
     <OptionItem label="Grape" value="grape" key={6} />,
     <OptionItem label="Lemon" value="lemon" key={7} />,
     <OptionItem label="Mango" value="mango" key={8} />,
+    <OptionItem label="" value="" key={9} />,
 ];
 
 const Template = (args) => {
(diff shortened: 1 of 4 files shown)


packages/wonder-blocks-dropdown/src/components/__docs__/single-select.stories.js
```diff
@@ -150,6 +150,7 @@ const items = [
     <OptionItem label="Grape" value="grape" key={6} />,
     <OptionItem label="Lemon" value="lemon" key={7} />,
     <OptionItem label="Mango" value="mango" key={8} />,
+    <OptionItem label="" value="" key={9} />,
```
kevinbarabash, line 153: **suggestion:** This change should be reverted since it affects existing stories.  For the story you've added, you could do `{[...items, <OptionItem label="" value="" key={9} />]}`.

packages/wonder-blocks-dropdown/src/components/select-opener.js
```diff
@@ -107,6 +107,9 @@ export default class SelectOpener extends React.Component<SelectOpenerProps> {
 
         const ClickableBehavior = getClickableBehavior(router);
 
+        // Note(TB): Prevents unwanted vertical shift for empty selection
+        const space = "\u00A0";
```
kevinbarabash, line 111: **suggestions:** maybe use `tamarab` or `myranae`.  `TB` on its own isn't much to go on.  Also, could you rename the variable to `nbsp` since I would expect `space` to contain `\u0020`.  Lastly, inlining this code (or at least placing the comment down below where `space` is used), I think would make more sense.  As it stands, if I read the comment here, I need to hunt for where `space` is used to understand how its achieving its purpose.

kevinbarabash (approved):
LGTM.  Please revert the change to `items` in single-select.stories.js before landing this.  I've provided a suggested workaround to keep the new story you've added behaving as expected.

packages/wonder-blocks-dropdown/src/components/select-opener.js
jandrade: **question:** Do you know how a Screen Reader will announce this text?

Myranae: When you read the blank option in the dropdown, it just reads as "text". When the placeholder is the blank spot, it does not read as anything though. I think most of the situations for us are having a blank placeholder.

Updated note reference to tamarab. Moved the code inline and got rid of the variable for space. Moved the comment down to where the nbsp unicode is used.

packages/wonder-blocks-dropdown/src/components/__docs__/single-select.stories.js
Myranae: Updated!

packages/wonder-blocks-dropdown/src/components/select-opener.js
```diff
@@ -146,7 +146,9 @@ export default class SelectOpener extends React.Component<SelectOpenerProps> {
                             {...childrenProps}
                         >
                             <LabelMedium style={styles.text}>
-                                {children}
+                                {/* // Note(tamarab): Prevents unwanted vertical
+                                shift for empty selection */}
+                                {children || "\u00A0"}
```
kevinbarabash, line 151: **style nit:** you don't need the additional `//`.

kevinbarabash (approved):
LGTM. :shipit: Thanks for fixing this issue.

packages/wonder-blocks-dropdown/src/components/select-opener.js
jandrade: 🤔 Maybe for all situations (if possible ofc) it would be nice adding an `aria-label` attribute that could help describing the intention of the dropdown control. No change required, but maybe it would be nice adding some sort of documentation around it.

PR #1692, opened 2023-01-06
Clickable components no longer click when being dragged into

nishasy:
## Summary:
Currently, Clickable components are activated on mouse release, when
they should only be activated on click.

Changes
- Removed the redundant `this.handleClick` from `handleMouseEnter`
  in clickable-behavior.js.
- After that step, the dragging logic doesn't actually do anything
  anymore, so I removed that too.
- Made it so that mouseenter only affects the hover state, not pressed.

Issue: https://khanacademy.atlassian.net/browse/WB-1425

## Test plan:
- Open http://localhost:6061/?path=/docs/clickable-clickable--default
- Click the clickable element. Confirm that the click alert pops up.
- Press inside the clickable element, drag out, and unpress. Confirm it does NOT result in a click alert.
- Press outside the clickable element, drag in, and unpress. Confirm it does NOT result in a click alert.
- Press inside the clickable element, drag within it, and unpress. Confirm that the click alert pops up.
- Press inside the clickable element, drag outside it, DON'T unpress, drag back into it, and then unpress.
  Confirm that the click alert pops up.
---
- Go to mobile storybook locally.
- Confirm that tapping clicks the button.
- Confirm that dragging into the button does not click it.
- Confirm that dragging out of the button does not click it.
- Dragging within and the button and out/back in does not apply in mobile.
---
`yarn jest packages/wonder-blocks-clickable/src/components/__tests__/clickable-behavior.test.js`

## Before:
(Hovering and then dragging)

https://user-images.githubusercontent.com/13231763/210906862-2c54fb4b-6503-48e0-ade3-e01a7550e28e.mov

## After:
(Hovering, dragging, and clicking)

https://user-images.githubusercontent.com/13231763/210906877-158ffcc1-bb4a-443f-ad77-0bb5b25d9eae.mov


packages/wonder-blocks-clickable/src/components/__tests__/clickable-behavior.test.js
```diff
@@ -763,7 +778,7 @@ describe("ClickableBehavior", () => {
         expect(onClick).toHaveBeenCalledTimes(expectedNumberTimesCalled);
     });
 
-    it("calls onClick on mouseup when the mouse was dragging", () => {
+    it("does not call onClick on mouseup when the mouse was dragging", () => {
```
kevinbarabash, line 781: **suggestion (non-blocking):** I think this test case should be split into three and the titles should be more specific since there's really three different test cases being tested here.

packages/wonder-blocks-clickable/src/components/__tests__/clickable-behavior.test.js
```diff
@@ -93,18 +93,33 @@ describe("ClickableBehavior", () => {
...
         );
         const button = screen.getByRole("button");
+        expect(button).not.toHaveTextContent("hovered");
+        expect(button).not.toHaveTextContent("pressed");
+
+        fireEvent.mouseEnter(button, {buttons: 1});
+        expect(button).toHaveTextContent("hovered");
+    });
+
+    it("changes only pressed state on mouse leave while dragging", () => {
```
kevinbarabash, line 103: It looks like the hover state goes away as well when dragging the mouse outside the bounds of the button.

packages/wonder-blocks-clickable/src/components/__tests__/clickable-behavior.test.js
```diff
@@ -93,18 +93,33 @@ describe("ClickableBehavior", () => {
             </ClickableBehavior>,
         );
         const button = screen.getByRole("button");
+        expect(button).not.toHaveTextContent("hovered");
+        expect(button).not.toHaveTextContent("pressed");
+
+        fireEvent.mouseEnter(button, {buttons: 1});
+        expect(button).toHaveTextContent("hovered");
```
kevinbarabash, line 100: **suggestion (non-blocking):** This test should also assert that the button is not in the pressed state after the call to `mouseEnter`.

kevinbarabash (approved):
Looks good to me.  Thanks for taking the time to simplify this code.  I did a little bit of testing on mobile, but it looks like you already had that covered by the test plan.  I left a few suggestions to improve the tests, but nothing blocking. 🎉

packages/wonder-blocks-clickable/src/components/__tests__/clickable-behavior.test.js
nishasy: Now that I removed `onDrag` altogether, I don't think this can be tested using `fireEvent.dragStart`. Also the middle test case (dragging within the button) doesn't show the same result as manual testing does.

nishasy: I guess I can keep 2/3 of these tests.

nishasy: I added back the two tests that did seem to work. But dragging within the button and letting go should result in a click, which I can't seem to replicate with RTL/jest. I also initially wanted to add a test for pressing in a button, leaving, reentering, and then letting go - this should click, but it doesn't seem to be able to in a jest/RTL test. It works manually though.

packages/wonder-blocks-clickable/src/components/__tests__/clickable-behavior.test.js
```diff
@@ -763,7 +779,17 @@ describe("ClickableBehavior", () => {
         expect(onClick).toHaveBeenCalledTimes(expectedNumberTimesCalled);
     });
 
-    it("calls onClick on mouseup when the mouse was dragging", () => {
+    // The following two tests involve click behavior when dragging.
```
jandrade, line 782: **praise:** This comment is great!

jandrade (approved):
Looks great!! 🚢

PR #1696, opened 2023-01-12
FEI-4917: Add useOnMountEffect hook to wonder-blocks-core

kevinbarabash:
## Summary:
Using useEffect() to run an effect once on mount is awkward b/c you have to pass an empty deps array and then disable the 'react-hooks/exhaustive-deps' lint rule.  The 'useOnMountEffect' takes care of this for us allow devs to focus on more important things.

Issue: FEI-4917

## Test plan:
- yarn jest use-on-mount-effect
- check that the docs are showing up in storybook

Screen Shot 2023-01-12 at 12 00 37 PM


packages/wonder-blocks-core/src/__docs__/exports.use-on-mount-effect.stories.mdx
```diff
@@ -0,0 +1,23 @@
...
+```ts
+function useOnMountEffect(callback: (void | () => void)) void;
+```
+
+The `useOnMountEffect` can be used to run an effect once on mount. This avoids
+having to pass `useEffect` an empty deps array and disable the
+`react-hooks/exhaustive-deps` lint.
+
+If `callback` returns a cleanup function, it will be called when the component
+is unmounted.
```
jandrade, line 23: **thought:** Based on the example you shared with me. What do you think about expanding the docs here to include an example?

I usually use the `Usage` section for other WB components, so maybe that could work here to explain that. wdyt?

https://github.com/Khan/wonder-blocks/blob/a8deae7200aefaa8c5f69b3ef3db3cae9be2b4fd/packages/wonder-blocks-spacing/src/__docs__/spacing.stories.mdx?plain=1#L33-L43

jandrade (approved):
Looks great! just a couple of comments for you to consider 🚀

packages/wonder-blocks-core/src/__docs__/exports.use-on-mount-effect.stories.mdx
somewhatabstract: I think it may be useful to also include in the docs what the equivalent `useEffect` code looks like. This would not only help explain to folks what this is doing for them, but it would also help them identify places that could be tidied up by using this hook.

.changeset/rich-dragons-sneeze.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-core": minor
+---
+
+useOnMountEffect hook
```
somewhatabstract, line 5: **nit:** I find it useful for our release notes to include a verb describing what happened, like "Added  " or "Fixed ".

packages/wonder-blocks-core/src/hooks/__tests__/use-on-mount-effect.test.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+import {renderHook} from "@testing-library/react-hooks";
+
+import {useOnMountEffect} from "../use-on-mount-effect.js";
+
+describe("#useOnMountEffect", () => {
+    it("shoul call the callback once", () => {
```
somewhatabstract, line 7: **nit:** 
```suggestion
    it("should call the callback once", () => {
```

packages/wonder-blocks-core/src/hooks/__tests__/use-on-mount-effect.test.js
```diff
@@ -0,0 +1,25 @@
...
+        const callback = jest.fn();
+        const {rerender} = renderHook(() => useOnMountEffect(callback));
+
+        rerender();
+
+        expect(callback).toHaveBeenCalledTimes(1);
+    });
+
+    it("should call the cleanup function is one is returned by the callback", () => {
```
somewhatabstract, line 16: **question:** What is this test case trying to say? The grammar is off somehow.

packages/wonder-blocks-core/src/hooks/__tests__/use-on-mount-effect.test.js
somewhatabstract: **change:** Arrange/Act/Assert comments are missing from these test cases.

See best practices: https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/98402353/JavaScript+Testing+Best+Practices

packages/wonder-blocks-core/src/hooks/use-on-mount-effect.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+import * as React from "react";
+
+/**
+ * Runs a callback once on mount.
+ *
+ * If the callback returns a cleanup function, it will be called with the component is unmounted.
```
somewhatabstract, line 7: **suggestion:** I think it's worth including in this commentary a snippet of code showing the equivalent `useEffect` code.

somewhatabstract (approved):
Generally looks good to me although there are some typos and other updates that are needed before landing.

packages/wonder-blocks-core/src/hooks/__tests__/use-on-mount-effect.test.js
kevinbarabash: Sorry about that.  I'll add them in.

kevinbarabash: **aside:** this is totally the kind of thing we could have an action for that checks for the comments and if there missing as the link in a comment.

packages/wonder-blocks-core/src/hooks/use-on-mount-effect.js
kevinbarabash: It felt redudant b/c the code is right below, but I can see people just reading the comment and not looking the code.  I'll add some comment about it.

somewhatabstract: Folks using this in their code are going to see the function comment in the intellisense tooltip - they aren't going to see the code.

PR #1707, opened 2023-01-24
Port Link tests from enzyme to RTL

jandrade:
## Summary:

Porting the `wonder-blocks-link` tests from `enzyme` to `@testing-library/react`.

Issue: XXX-XXXX

## Test plan:

Verify that all the unit tests pass.


packages/wonder-blocks-link/src/components/__tests__/link.test.js
```diff
@@ -41,16 +35,16 @@ describe("Link", () => {
             );
 
             // Act
-            const linkWrapper = wrapper.find("Link").first();
-            linkWrapper.simulate("click", {button: 0});
+            const linkWrapper = screen.getByText("Click me!");
```
nishasy, line 38: Nit: I think this can be just `link` instead of `linkWrapper` now.

nishasy (approved):
Nice! Thanks for doing this! No blocking requests

packages/wonder-blocks-link/src/components/__tests__/link.test.js
jandrade: ah yeah, good call! will update after fixing the failed test.

packages/wonder-blocks-link/src/components/__tests__/link.test.js
```diff
@@ -179,21 +164,19 @@ describe("Link", () => {
...
 
             // Act
-            const linkWrapper = wrapper.find("Link").first();
-            linkWrapper.simulate("click", {
-                button: 0,
-            });
-            await wait(0);
-            wrapper.update();
+            const link = screen.getByText("Click me!");
+            userEvent.click(link);
```
kevinbarabash, line 168: So much easier. 🙂

packages/wonder-blocks-link/src/components/__tests__/link.test.js
```diff
@@ -325,38 +298,38 @@ describe("Link", () => {
...
+                <Link href="/" onKeyUp={(e) => (keyCode = e.keyCode)}>
                     Click me!
                 </Link>,
             );
 
             // Act
-            wrapper.find(Link).simulate("keydown", {keyCode: 32});
+            const link = screen.getByText("Click me!");
+            // eslint-disable-next-line testing-library/prefer-user-event
+            fireEvent.keyUp(link, {keyCode: 32});
```
kevinbarabash, line 329: **question:** why did this change to a keyup from a keydown?

kevinbarabash (commented):
Thanks for porting these tests to RTL. 💚

packages/wonder-blocks-link/src/components/__tests__/link.test.js
jandrade: I think the previous test was giving false positives as we are intending to test `keyUp` and the previous version was testing `keyDown` instead.

PR #1709, opened 2023-01-26
WB-1081: TextField - Replace all "internal" Class Names to intended Names

jandrade:
## Summary:
We had to use "internal" class names for the components used in
`TextField` and `LabeledTextField` because Styleguidist was not able to
resolve the names properly (this was an issue with `Styleguidist` +
`forwardRef`).

Now that Styleguidist is gone, we can replace the "internal" class names
with the intended ones, as this no longer causes any issues with
Storybook.

NOTE: This is a super small change that will help us with better
readability when using React DevTools.

Issue: WB-1081

Test Plan:

Verify that all the checks pass (specially unit tests) and that the
`TextField` and `LabeledTextField` stories work fine.

### DIFF
--- a/.changeset/green-chicken-rush.md
+++ b/.changeset/green-chicken-rush.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-form": patch
+---
+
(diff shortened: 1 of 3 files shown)


kevinbarabash (approved):
LGTM. 🚢

PR #1711, opened 2023-01-30
hack-enzyme.2: Port `wonder-blocks-core` to RTL

jandrade:
## Summary:
This ports remaining the `wonder-blocks-core` tests from Enzyme to RTL.

Issue: XXX-XXXX

## Test plan:

Verify that unit tests pass.


packages/wonder-blocks-core/src/components/__tests__/unique-id-provider.test.js
```diff
@@ -51,26 +50,38 @@ describe("UniqueIDProvider", () => {
...
+                    UniqueIDFactory,
+                );
+            });
+
+            waitFor(() => {
+                // Check the second render gets the same UniqueIDFactory instance.
+                expect(children.mock.calls[1][0]).toBe(
+                    children.mock.calls[1][0],
+                );
+            });
```
somewhatabstract, line 84: **question**: Shouldn't this be a single `waitFor` containing all the expectations? Also, doesn't `waitFor` need to be `await`'d?

packages/wonder-blocks-core/src/components/__tests__/unique-id-provider.test.js
```diff
@@ -112,26 +123,38 @@ describe("UniqueIDProvider", () => {
...
+                    UniqueIDFactory,
+                );
+            });
+
+            waitFor(() => {
+                // Check the third render gets the same UniqueIDFactory instance.
+                expect(children.mock.calls[2][0]).toBe(
+                    children.mock.calls[1][0],
+                );
+            });
```
somewhatabstract, line 157: **question**: Shouldn't this be a single `waitFor` containing all the expectations? Also, doesn't `waitFor` need to be `await`'d?

packages/wonder-blocks-core/src/components/__tests__/unique-id-provider.test.js
```diff
@@ -171,11 +194,15 @@ describe("UniqueIDProvider", () => {
...
 
             // Assert
-            expect(foo).toHaveBeenCalled();
-            expect(foo.mock.calls[0][0]).toBeTruthy();
+            waitFor(() => {
+                expect(foo).toHaveBeenCalled();
+            });
+            waitFor(() => {
+                expect(foo.mock.calls[0][0]).toBeTruthy();
+            });
```
somewhatabstract, line 205: **question**: Shouldn't this be a single `waitFor` containing all the expectations? Also, doesn't `waitFor` need to be `await`'d?

somewhatabstract (approved):
I think this can be tidied up a bit and it is missing some `await`ing.

packages/wonder-blocks-core/src/components/__tests__/unique-id-provider.test.js
jandrade: Oof thanks for pointing this out! I found out that there was an issue with `rerender` that was leading to some false positives.

I made a refactor to these tests so it now works consistently + we don't have to use `waitFor` at all.

The problem was that I was using the reference to the nodes, instead of using a reference to a component instance. This was causing `rerender` not to re-render at all.

BEFORE:

```
const nodes = (<Some><Tree/></Some>);
const {rerender} = render(nodes);
rerender(nodes); // this doesn't re-render internally.
```

AFTER:
```
const UnderTest = () => (<Some><Tree/></Some>);
const {rerender} = render(<UnderTest />);
rerender(<UnderTest />); // this now forces a re-render as expected.
```

PR #1713, opened 2023-01-31
hack-enzyme.3: Port Breadcrumbs, Data, Dropdown, and Grid to Enzyme

jandrade:
## Summary:

Ports some more packages to Enzyme. I noticed that there were some
files/specs that needed to be ported in those packages, so the main goal
here is to get those files/specs ported to RTL.

Issue: XXX-XXXX

## Test plan:

Verify that tests pass.


packages/wonder-blocks-dropdown/src/components/__tests__/action-item.test.js
```diff
@@ -28,20 +26,6 @@ describe("ActionItem", () => {
...
-        // Act
-        const wrapper = mount(
-            <MemoryRouter>
-                <ActionItem href="/foo" label="Example" />
-            </MemoryRouter>,
-        );
-
-        // Assert
-        expect(wrapper.find(Link)).toHaveLength(1);
-    });
```
kevinbarabash, line 43: **suggestion (non-blocking):** This test could be ported by mocking `react-router-dom`.

kevinbarabash (approved):
LGTM. 🎉  Thanks for porting these tests.

packages/wonder-blocks-dropdown/src/components/__tests__/action-item.test.js
jandrade: Do you have an example of how that test would look like? you mean something like doing some client-side navigation to verify that the `Link` works as expected? or something completely different?

kevinbarabash: See https://github.com/Khan/webapp/pull/11326/files#diff-a38bc63689752bc546ffae6d8a74d17f7c502b0aeacdd9dec7816baf8ffb1140.  You might not need to mock the whole module in order to mock Link.

PR #1716, opened 2023-01-31
hack-enzyme.6: Port `wonder-blocks-popover` to RTL

jandrade:
## Summary:

This is the next stop in a series of changes to port all of our tests to
RTL.

This change ports the `wonder-blocks-popover` package to RTL. Also
deletes some unused test utilities.

Issue: XXX-XXXX

## Test plan:

```
yarn test
```


packages/wonder-blocks-popover/src/components/__tests__/popover-dialog.test.js
```diff
@@ -1,45 +1,27 @@
...
-            <PopoverDialog placement="top" onUpdate={onUpdateMock}>
-                <PopoverContentCore color="darkBlue">
-                    popover content
-                </PopoverContentCore>
-            </PopoverDialog>,
-        );
-
-        // Assert
-        expect(wrapper.find(TooltipTail).prop("color")).toBe("darkBlue");
-    });
```
somewhatabstract, line 29: **question:** Is there another way we can validate this? It seems like we want to make sure it's still the case somehow. One option is to just mock the tail import and capture the props, for example. There may be some other way such as looking for specific HTML in the `container.innerHTML` returned by `render`.

somewhatabstract (approved):
LGTM - nice to see some deleted files.

I think the deleted test case likely is needed to avoid a regression, so perhaps there's a way to bring that back?

packages/wonder-blocks-popover/src/components/__tests__/popover-dialog.test.js
```diff
@@ -13,33 +12,37 @@ jest.mock("@khanacademy/wonder-blocks-tooltip");
 describe("PopoverDialog", () => {
     it("should update the tail color to match the content's color", () => {
         // Arrange
-        const onUpdateMock = jest.fn();
+        const tooltipTailSpy = jest.spyOn(Tooltip, "TooltipTail");
```
jandrade, line 15: @somewhatabstract that's a fair point! I added back the test and mocked TooltipTail as you suggested. I felt that this assertion would be better as with the HTML one is hard to understand/read what do we need to expect from the test + it would easily break by a minor SVG change.

PR #1717, opened 2023-01-31
hack-enzyme.7: Port testing and timing to RTL

jandrade:
Summary:

Ported some remaining tests in the following packages to RTL:

- wonder-blocks-testing
- wonder-blocks-timing

Issue: XXX-XXXX

Test Plan:

```
yarn test
```


packages/wonder-blocks-testing/src/harness/adapters/__tests__/data.test.js
```diff
@@ -10,30 +9,10 @@ describe("WonderBlocksData.adapter", () => {
...
-        const wrapper = mount(Data.adapter(children, []), {
-            includeDefaultTestHarness: false,
-        });
+        render(Data.adapter(children, []));
 
         // Assert
-        expect(wrapper).toHaveHTML("<div>CONTENT</div>");
-    });
-
-    it("should render children within InterceptRequests when dataIntercepts configured", () => {
```
jandrade, line 21: **note:** I don't think this test has value within the RTL context. Please let me know if you think otherwise.

somewhatabstract: **suggestion:** I think it is important to test this still as we want to make sure that intercepts work within this component.

Perhaps the children need to make a request and render the result, showing the request was intercepted.

somewhatabstract (changes requested):
Looks good, but we still need a test in the `Data` component to verify that request interception will work.

packages/wonder-blocks-testing/src/harness/adapters/__tests__/data.test.js
jandrade: I added back this test and used `useCachedEffect` for that. TBH I'm not fully sure if that's the right way to use the data adapter. Let me know if there's a better approach for this test. Happy to make the change if you think so :).

packages/wonder-blocks-testing/src/harness/adapters/__tests__/data.test.js
```diff
@@ -10,57 +10,57 @@ describe("WonderBlocksData.adapter", () => {
...
-        });
+        render(Data.adapter(children, []));
 
         // Assert
-        expect(wrapper).toHaveHTML("<div>CONTENT</div>");
+        expect(screen.getByText("CONTENT")).toBeInTheDocument();
     });
 
-    it("should render children within InterceptRequests when dataIntercepts configured", () => {
+    it("should render children within InterceptRequests when dataIntercepts configured", async () => {
```
somewhatabstract, line 19: **suggestion:** Probably need to change the test name to something like - `should supports request interception via configured dataIntercepts`

packages/wonder-blocks-testing/src/harness/adapters/__tests__/data.test.js
```diff
@@ -10,57 +10,57 @@ describe("WonderBlocksData.adapter", () => {
...
         // Arrange
-        const children = <div>CONTENT</div>;
+
+        const TestFixture = () => {
+            const [result] = useCachedEffect("ID", jest.fn());
```
somewhatabstract, line 23: **suggestion:** This works well enough. It would involve less "infrastructure" to use `useRequestInterception` although that would be more work here in the test, I think. This seems good enough to me (as long as the shared cache is getting cleared between tests).

somewhatabstract (approved):

PR #1718, opened 2023-02-01
hack-enzyme.8: Port `wonder-blocks-tooltip` to RTL

jandrade:
Summary:

Final part of the porting of `wonder-blocks` from Enzyme to RTL.

This ports the `wonder-blocks-tooltip` package.

Test Plan:

```
yarn test
```

### DIFF
--- a/packages/wonder-blocks-tooltip/src/components/__tests__/tooltip-anchor.test.js
+++ b/packages/wonder-blocks-tooltip/src/components/__tests__/tooltip-anchor.test.js
@@ -41,11 +41,6 @@ describe("TooltipAnchor", () => {
 
     test("on mount, subscribes to focus and hover events", () => {
         // Arrange
-        const nodes = (
-            <TooltipAnchor anchorRef={() => {}} onActiveChanged={() => {}}>
-                Anchor text
-            </TooltipAnchor>
-        );
         const addEventListenerSpy = jest.spyOn(
             HTMLElement.prototype,
             "addEventListener",
@@ -53,7 +48,11 @@ describe("TooltipAnchor", () => {
         addEventListenerSpy.mockClear();
 
         // Act
-        render(nodes);
+        render(
+            <TooltipAnchor anchorRef={() => {}} onActiveChanged={() => {}}>
+                Anchor text
+            </TooltipAnchor>,
+        );
 
         // Assert
         expect(addEventListenerSpy).toHaveBeenCalledWith(
@@ -76,17 +75,16 @@ describe("TooltipAnchor", () => {
 
     test("on unmount, unsubscribes from focus and hover events", () => {
         // Arrange
-        const nodes = (
-            <TooltipAnchor anchorRef={() => {}} onActiveChanged={() => {}}>
-                Anchor text
-            </TooltipAnchor>
-        );
         const removeEventListenerSpy = jest.spyOn(
             HTMLElement.prototype,
             "removeEventListener",
         );
         removeEventListenerSpy.mockClear();
-        const wrapper = render(nodes);
+        const wrapper = render(
+            <TooltipAnchor anchorRef={() => {}} onActiveChanged={() => {}}>
+                Anchor text
+            </TooltipAnchor>,
+        );
 
         // Act
         wrapper.unmount();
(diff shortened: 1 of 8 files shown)


packages/wonder-blocks-tooltip/src/components/__tests__/tooltip-tail.test.js
```diff
@@ -26,92 +28,12 @@ describe("TooltipTail", () => {
...
+            const testees = testPoints.map((tp) => () => render(makeNode(tp)));
 
             // Assert
             for (const testee of testees) {
                 expect(testee).not.toThrowError();
             }
         });
     });
-
-    describe("INTERNALS", () => {
```
jandrade, line 38: **note:** I removed all the following tests b/c I noticed that these introduced false positives in the context of our current testing strategy. 

`_minDistanceFromCorners` was still being used in the code b/c it was being tested here, but it was not being used directly within the code implementation.

Instead of testing these methods, we can rely on FlowJS (or TS in the future) to ensure that we won't reach the impossible states that we are testing here.

packages/wonder-blocks-tooltip/src/components/__tests__/tooltip-anchor.test.js
```diff
@@ -113,122 +111,116 @@ describe("TooltipAnchor", () => {
...
+            );
+            render(nodes);
 
             // Act
-            const result = ref?.getAttribute("tabindex");
+            const result = screen.getByText("This is the anchor");
 
             // Assert
-            expect(result).toBe("0");
+            expect(result).toHaveAttribute("tabindex", "0");
```
kevinbarabash, line 131: **praise:** thanks for using a precise matcher.

packages/wonder-blocks-tooltip/src/components/__tests__/tooltip-anchor.test.js
```diff
@@ -113,122 +111,116 @@ describe("TooltipAnchor", () => {
...
-                    >
-                        <View tabIndex={-1}>This is the anchor</View>
-                    </TooltipAnchor>
-                );
-                mount(nodes);
-            });
+            const nodes = (
+                <TooltipAnchor
+                    forceAnchorFocusivity={true}
+                    anchorRef={jest.fn()}
```
kevinbarabash, line 139: **question:** are there any tests that check that we set the anchor ref correctly?

packages/wonder-blocks-tooltip/src/components/__tests__/tooltip-anchor.test.js
```diff
@@ -248,26 +240,20 @@ describe("TooltipAnchor", () => {
...
+            );
 
             // Act
-            // Let's fake a focusin (this is the event that the anchor gets
-            // whether focused directly or a child is focused). We have to
-            // fake directly because there's no real browser here handling
-            // focus and real events.
-            ref?.dispatchEvent(new FocusEvent("focusin"));
+            // Let's focus the anchor
+            userEvent.tab();
```
kevinbarabash, line 256: Yes! 🎉

packages/wonder-blocks-tooltip/src/components/__tests__/tooltip-anchor.test.js
```diff
@@ -293,52 +279,46 @@ describe("TooltipAnchor", () => {
...
-                            activeState = active;
-                        }}
-                    >
-                        Anchor Text
-                    </TooltipAnchor>
-                );
-                mount(nodes);
-            });
-
-            ref?.dispatchEvent(new FocusEvent("focusin"));
```
kevinbarabash, line 341: I love that we're no longer dispatching events like this.

packages/wonder-blocks-tooltip/src/components/__tests__/tooltip-anchor.test.js
```diff
@@ -474,41 +448,42 @@ describe("TooltipAnchor", () => {
...
+                <TooltipAnchor
+                    anchorRef={jest.fn()}
+                    onActiveChanged={(active) => {
+                        activeState = active;
+                    }}
+                >
+                    Anchor Text
+                </TooltipAnchor>
+            );
+            render(nodes);
```
kevinbarabash, line 472: **style nit (non-blocking):** we could pass the JSX directly to `render()` instead of using `nodes`.  This is what we do in most of the other test cases.

packages/wonder-blocks-tooltip/src/components/__tests__/tooltip-tail.test.js
kevinbarabash: Using Chromatic for these tests cases makes sense. 👍

packages/wonder-blocks-tooltip/src/components/tooltip-tail.js
```diff
@@ -247,24 +247,6 @@ export default class TooltipTail extends React.Component<Props> {
...
-                return minDistanceFromCornersForTopBottom;
-
-            case "left":
-            case "right":
-                return minDistanceFromCornersForLeftRight;
-
-            default:
-                throw new Error(`Unknown placement: ${placement}`);
-        }
-    }
```
kevinbarabash, line 266: **praise:** I didn't realize that this code wasn't being used in production.  Thanks for removing it.

kevinbarabash (approved):
LGTM. 🚢 Thanks for migrating these tests to RTL. 💚

packages/wonder-blocks-tooltip/src/components/__tests__/tooltip-anchor.test.js
jandrade: oh good point, I'll add one!

PR #1720, opened 2023-02-01
[experimental] Remove webpack from the build process

jandrade:
## Summary:
This is an experimental change to remove webpack from the build process.
This is done to see if we can get a faster build process and to see if
we can start migrating to Storybook v7.

Issue: XXX-XXXX

## Test plan:

TBD


package.json
```diff
@@ -5,13 +5,12 @@
...
-    "node": ">=12.22"
+    "node": ">=16"
   },
   "scripts": {
     "alex": "alex packages/",
-    "build:all": "yarn run build:cjs && yarn run build:es6",
-    "build:cjs": "webpack --config ./build-settings/webpack.config.js --display-modules && node ./utils/package-flow-types.js",
-    "build:es6": "rollup -c ./build-settings/rollup.config.js",
+    "build": "rollup -c ./build-settings/rollup.config.js",
+    "build:es6": "yarn run build",
```
jandrade, line 13: **note:** I kept this command just to make the `Check build sizes` workflow step happy. I'll remove it later in an Audit.

.changeset/tender-windows-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+"wb-dev-build-settings": major
```
jandrade, line 2: **note:** I'm not fully sure if this should be marked as a `major`. It's a breaking change for this package, but it's not a "public" package, so I don't know how to bump this, and how that will affect the WB packages that depend on this (as a dev dep).

jeresig (approved):
Awesome! Really glad we can get rid of this!

PR #1726, opened 2023-02-03
Experiment: v8 coverage provider

kevinbarabash:
## Summary:
The v8 coverage provider provides more details coverage of JSX code.  It uses v8 itself to instrument the code being run which means that the JS is unmodified.  The other coverage provider - istanbul - injects a bunch of additional JS statements to track coverage.  While v8 coverage is faster on line-by-line basis, the fact that it's tracking coverage for more lines than istanbul does may actually result in longer test run times.  The benefit of switching to v8 is that the line-by-line coverage reports should be more accurate.

Issue: None

## Test plan:
- yarn coverage
- look at the coverage reports
- see that more lines are being coverage than in the old report


config/jest/test.config.js
```diff
@@ -34,6 +34,7 @@ module.exports = {
         "!<rootDir>/node_modules/",
         "!packages/**/node_modules/",
     ],
+    coverageProvider: "v8",
```
jeresig, line 37: Could you add a comment explaining why we need to use this?

jeresig (approved):
Great - thanks for adding this!

PR #1729, opened 2023-02-07
Add inline variant to Link

nishasy:
## Summary:
Add the `inline` variant to Link. Make it match the design spec
https://www.figma.com/file/VbVu3h2BpBhH80niq101MHHE/%F0%9F%92%A0-Main-Components?node-id=11512%3A7203&t=eFPXHDHuxnSBGKzI-0

Issue: https://khanacademy.atlassian.net/browse/WB-1450

## Test plan:
`yarn jest`

Storybook interaction tests
https://5e1bf4b385e3fb0020b7073c-wqfijoyjxy.chromatic.com/?path=/story/link--inline
https://5e1bf4b385e3fb0020b7073c-wqfijoyjxy.chromatic.com/?path=/story/link--inline-light

Storybook manual testing
- Check colors of resting links
- Check colors of hovered links
- Check colors of mousdown links
- Check outline of focused links
- Add focus to link and hover at the same time. The focus ring should stay.


packages/wonder-blocks-link/src/__tests__/custom-snapshot.test.js
```diff
@@ -43,35 +43,38 @@ describe("LinkCore", () => {
...
-                                        {...stateProps}
-                                        {...defaultHandlers}
-                                    >
-                                        Click me
-                                    </LinkCore>,
-                                )
-                                .toJSON();
-                            expect(tree).toMatchSnapshot();
-                        });
+                    for (const inline of [true, false]) {
```
nishasy, line 49: I added this line
```
for (const inline of [true, false]) {
```
with a closing `}` later.

The rest of it is just indented.

packages/wonder-blocks-link/src/components/__docs__/link.stories.js
```diff
@@ -42,30 +47,32 @@ Default.args = {
     children: "Hello, world!",
 };
 
-export const Basic: StoryComponentType = () => (
+export const Primary: StoryComponentType = () => (
```
nishasy, line 50: I thought changing `Basic` to `Primary` would be more clear.

packages/wonder-blocks-link/src/components/__docs__/link.stories.js
```diff
@@ -42,30 +47,32 @@ Default.args = {
...
     },
 };
 
-Basic.play = async ({canvasElement}) => {
+Primary.play = async ({canvasElement}) => {
     const canvas = within(canvasElement);
 
     const link = canvas.getByRole("link");
 
+    await expect(link).toHaveStyle(`color: ${Color.blue}`);
```
nishasy, line 66: I added the initial color check to all the interaction tests because I ran into an issue at one point where the initial color was wrong and it wasn't being caught anywhere.

packages/wonder-blocks-link/src/components/__docs__/link.stories.js
```diff
@@ -164,48 +173,188 @@ LightVisitable.parameters = {
...
+        `text-decoration: underline ${Color.white} dashed 2px`,
+    );
+
+    await fireEvent.mouseDown(primaryLink);
+    await expect(primaryLink).toHaveStyle(
+        `text-decoration: underline solid ${fadedBlue} 1px`,
+    );
+};
+
+export const Variants: StoryComponentType = () => (
```
nishasy, line 280: I consolidated all the variants into one story.

packages/wonder-blocks-link/src/components/link-core.js
```diff
@@ -92,8 +96,14 @@ const sharedStyles = StyleSheet.create({
     },
 });
 
-const _generateStyles = (kind, light, visitable) => {
-    const buttonType = kind + light.toString() + visitable.toString();
+const _generateStyles = (inline, kind, light, visitable) => {
+    // Even though `inline` is defined and set the same as all the other
+    // boolean props, it is coming up as `undefined` in tests, which means
+    // the toString() function can't be used. Setting it explicitly to avoid
+    // a TypeError.
```
nishasy, line 103: I have NO idea why this is happening. 😕 Maybe someone can help me look at this before I land this PR?

packages/wonder-blocks-link/src/components/link-core.js
```diff
@@ -136,14 +157,20 @@ const _generateStyles = (kind, light, visitable) => {
         : Object.freeze({});
 
     const newStyles: StyleDeclaration = {
-        default: {
+        resting: {
```
nishasy, line 160: I thought "resting" would be more clear than "default"

packages/wonder-blocks-link/src/components/link-core.js
```diff
@@ -43,11 +44,14 @@ export default class LinkCore extends React.Component<Props> {
...
-        const linkStyles = _generateStyles(kind, light, visitable);
+        const linkStyles = _generateStyles(inline, kind, light, visitable);
+        const restingStyles = inline
+            ? linkStyles.restingInline
+            : linkStyles.resting;
 
         const defaultStyles = [
             sharedStyles.shared,
-            !(hovered || focused || pressed) && linkStyles.default,
+            !(hovered || focused || pressed) && restingStyles,
```
jandrade, line 54: **question:** I'm not sure if this is an issue that recently surfaced, but I noticed that if we focus on the link, then hover on it and move the mouse out, the focus ring is no longer present (but the link is still focused).

Do you know if a recent change introduced that behavior? or this is something we do on purpose?

For reference:

https://user-images.githubusercontent.com/843075/217328219-12f4c11f-bd08-419d-a262-d852719b088b.mov

jandrade (commented):
Giving a first pass on the changes! looking great so far, but want to discuss about an interesting behavior with focus/hover states.

packages/wonder-blocks-link/src/components/link-core.js
nishasy: @jandrade re: focus vs hover, I thought that was the intended behavior based on the previous logic in the code. But I just tested links on Wikipedia to see how this is handled there, and you're absolutely right - the focus ring is still there when hovering and clicking! This should be an easy fix.

I just want to confirm that this sounds okay with @megaroeny.

megaroeny: @nishasy @jandrade , so does that mean that we should have variants for focus while hovering/pressing? I don't think the designers would ever use that, but I could try to add specs for that in Figma, at least!

I think it would look poor if it kept the underline, though, right?

nishasy: @megaroeny If you don't want the underline and focus ring at the same time, I can just change the logic so that focus takes priority over hover. 

The examples I looked at so far are:
- Wikipedia: Focus includes both the ring and the underline. Hovering over a focused link doesn't make it look different. Pressing changes the color and keeps the focus ring.
- Google: Focus includes just the focus ring. Hovering over a focused link adds an underline. Pressing removes the focus ring.
- w3: All links are underlined. Focus adds the ring and changes the color. Hovering over a focused link doesn't make it look different. Pressing doesn't make it look different.

packages/wonder-blocks-link/src/components/link-core.js
jandrade: what value `buttonType` is using? I'm not sure what was the original intention, but to me it looks like it's setting something like: `primarytruetrue`? I'm not sure if that's what we want (to clarify this is not something that you introduced).

nishasy: Yeah, I believe that is right. Each variant of button has its own style saved in this `styles` array with `buttonType` being the index. I'm not sure why it's done this way.

jandrade: I wonder what would happen if you move `inline` after `kind`? maybe it would also help to separate each condition:

e.g.
```
const buttonType = `${kind}-${inline}-${light}-${visitable}`;
```

jandrade (approved):
Looks great to me! :shipit: Hopefully we'll be able to resolve the focus/mouse issue with ClickableBehavior in a separate PR.

packages/wonder-blocks-link/src/components/__docs__/link.stories.js
```diff
@@ -307,4 +477,8 @@ const styles = StyleSheet.create({
         flexDirection: "row",
         alignItems: "center",
     },
+    standaloneLinkWrapper: {
+        display: "inline-block",
```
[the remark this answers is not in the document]
nishasy: I used `inline-block` so that the outline wraps just the text instead of taking up the full width. I tried using `flex` and it doesn't fix the full width outline issue. I'll add a comment to clarify.

Screen Shot 2023-02-09 at 2 58 32 PM
Screen Shot 2023-02-09 at 2 57 45 PM

nishasy:
The parent pull-request ([#1700](https://github.com/Khan/wonder-blocks/pull/1700)) has been merged into `main`, but this branch (`link-inline`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #1735, opened 2023-02-10
WB-1190: Cleanup WB dependencies between packages

jandrade:
## Summary:
We noticed a while that we have an incorrect dependency in wb-dropdown.
I checked and we were not using wb-button internally in the code (only
for stories/documentation), so that might be bumping the dropdown
version when button changed unnecessarily.

This PR changes the other packages that are out-of-date to only include
the WB deps that are used in the code (excluding tests, stories, etc) so
the releases are a bit more contained.

In VS Code, I checked each `package.json` individually and used the
search panel to look for the other packages. I then checked the
`package.json` of the package I was looking for and compared any missing
WB packages that were not included in package.json, or removed the ones
that were not being imported in the code.

NOTE: I filtered out the contents of the `__tests__` and `__docs__`
folders for this search.

Issue: WB-1190

## Test plan:

Verify that the "Check builds" workflow job passes.


somewhatabstract (approved):
I like these changes, but what stops us getting into this situation again. Is there a tool we can write that can make sure each package.json is only referencing the things it uses? I feel like that would be really helpful.

Perhaps the parsing tool I wrote for webapp that we use for lambdatest-config could be used here. Maybe I can migrate that to Wonder Stuff and then we can reuse it in our other repos to make sure deps are correct.

jandrade:
> Perhaps the parsing tool I wrote for webapp that we use for lambdatest-config could be used here. Maybe I can migrate that to Wonder Stuff and then we can reuse it in our other repos to make sure deps are correct.

That would be super helpful! and we could even extend that to other monorepos to improve these best practices (like perseus).

PR #1736, opened 2023-02-10
Dropdown: Prototype with combobox role

jandrade:
## Summary:

Just playing around with dropdowns setting role="combobox" to the opener, to see
if we can find better testing practices for these components.

Issue: XXX-XXXX

## Test plan:

Navigate to /?path=/story/dropdown-singleselect--selected-single-select

Verify that the interaction test passes.


jandrade:
Just experimenting

PR #1737, opened 2023-02-15
Add Typography documentation to Storybook

nishasy:
## Summary:
Migrate the Typography documentation from styleguidist to Storybook.

(I used the same examples as the ones from styleguidist.)

Issue: https://khanacademy.atlassian.net/browse/WB-1147

## Test plan:
`yarn flow`
`yarn run lint`

Storybook
- Alter values in the control panel and confirm that the example changes accordingly.
  - Add {"color": "blue"} to style prop and confirm the text becomes blue.
- Go to "Noto for non-Latin languages" story and change the values in the
  dropdown. Confirm it changes the languages correctly.


packages/wonder-blocks-typography/src/__docs__/typography.stories.js
```diff
@@ -0,0 +1,294 @@
...
+            source: {
+                // See https://github.com/storybookjs/storybook/issues/12596
+                excludeDecorators: true,
+            },
+        },
+    },
+    argTypes: TypographyArgTypes,
+};
+
+export const ControlProps: StoryComponentType = (args) => <Title {...args} />;
```
nishasy, line 62: I didn't name this `Default` because I thought that would be confusing considering there's no default in the Typography package. Please let me know if you can think of something better than `ControlProps` for this.

megaroeny (commented):
Curious, for the [Typography Elements](https://5e1bf4b385e3fb0020b7073c-bjsyszjzwy.chromatic.com/?path=/story/typography--typography-elements) story showing every type style/scale, are there line-heights applied there? Seems like at a glance there isn't, because I'd expect to see more whitespace between things like the headings at least. But maybe I'm wrong.

packages/wonder-blocks-typography/src/__docs__/typography.argtypes.js
```diff
@@ -0,0 +1,31 @@
...
+        control: {type: "text"},
+        description: "Test ID used for e2e testing.",
+        table: {type: {summary: "string"}},
+        type: {required: false},
+    },
+    tabIndex: {
+        control: {type: "number"},
+        table: {type: {summary: "number"}},
+        type: {required: false},
+    },
```
jandrade, line 24: **sugestion:** Even that the types use `tabIndex`, I think it would be better to not include this in the documentation as it might mislead devs about using this prop for headings? wdyt? do you think there would still be cases where we should/need to use `tabIndex` for headings and/or paragraphs?

packages/wonder-blocks-typography/src/__docs__/typography.stories.js
```diff
@@ -0,0 +1,294 @@
...
+
+ControlProps.args = {
+    children: "This is a Title typography element",
+    id: "example-Title",
```
jandrade, line 66: **nit:** nothing blocking, probably copy pasta from the styleguidist version.

```suggestion
    id: "example-title",
```

jandrade (approved):
Looks good to me!  I feel that we still have a lot of opportunities for improving these docs, but this is good enough for the migration part of it 🚢.

nishasy:
@megaroeny The line height is present. Maybe it looks weird because it's different on each line?

Your comment did lead to me find out that they're all using the mobile styles for some reason though, so I'll try to fix that now.

packages/wonder-blocks-typography/src/__docs__/typography.argtypes.js
nishasy: I agree with you. I was about to ask why that's even there, but `TextViewSharedProps` is supposed to be pretty versatile so I guess it makes sense.

nishasy:
@megaroeny nvm it was just the size of the window, which is expected

PR #1743, opened 2023-02-22
FEI-4966.2: Add 'format' script

kevinbarabash:
## Summary:
This will come in handy later on to fix formatting issues after running the codemod.  The reason for not using 'eslint --fix' is that it's super slow.  Running 'prettier' directly only takes 6s to format all of our .js files.

Issue: FEI-4966

## Test plan:
- add some formatting lint to a file
- run 'yarn format'
- see that it's fixed

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,4 +3,9 @@
 docs
 *.css
 *.yml
+flow-typed
 flow-typed/npm
+coverage
+storybook-static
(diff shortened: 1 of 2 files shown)


kevinbarabash:
I was trying to run `eslint --fix` on all files including files including storybook-static, that's why it was so slow. 😞

PR #1744, opened 2023-02-22
FEI-4966.2: Convert Flow enums to POJOs

kevinbarabash:
## Summary:
The codemod we're using can't deal with Flow enums so we're temporarily converting them to POJOs.

Issue: FEI-4966

## Test plan:
- let CI do its thing


jandrade (approved):
:shipit:

PR #1746, opened 2023-02-22
FEI-4966.3: Configure typescript and type definitions for dependencies

kevinbarabash:
## Summary:
Before we can convert all of the source code from Flow to TypeScript we need to have TypeScript configured properly and lib defs for third party libraries.  The codemod will be run in the next PR.

Issue: FEI-4966

## Test plan:
- run 'yarn tsc --noEmit' in the next PR to check that things are configured properly


packages/wonder-blocks-banner/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+    "exclude": ["dist"],
+    "extends": "../tsconfig-shared.json",
+    "compilerOptions": {
+        "outDir": "./dist",
+        "rootDir": "src",
+    },
+    "references": [
+        {"path": "../wonder-blocks-button"},
```
jeresig, line 9: Are these generated automatically?

jeresig (approved):
Nice, exciting!!

packages/wonder-blocks-banner/tsconfig.json
kevinbarabash: I wrote them all by hand.  I copy/pasted them and then tweaked the references.  That works in the short term, but in the long term we need tooling to make sure that these are correct.  It should be pretty obvious if there's a missing reference path though since the types for that dependency won't be available.

jeresig (approved):

PR #1747, opened 2023-02-22
FEI-4966.2a: Replace .defaultProps on functional components

kevinbarabash:
## Summary:
TypeScript doesn't like this pattern.  Instead, we provide defaults when destructuring props inside the component.

Issue: FEI-4966

## Test plan:
- yarn flow

### DIFF
--- a/.changeset/seven-mangos-leave.md
+++ b/.changeset/seven-mangos-leave.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-banner": patch
+---
+
+Replace .defaultProps on functional component with corresponding defaults when destructuring props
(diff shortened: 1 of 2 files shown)


jeresig (approved):
Nice - to clarify there is only a single component following this pattern?

kevinbarabash:
> Nice - to clarify there is only a single component following this pattern?

That's correct.  I was a little surprised by that.  I'm pretty sure we there are more components that do this in webapp.

jandrade:
> Nice - to clarify there is only a single component following this pattern?

@jeresig I think so... we haven't recently created new components (we previously created class-based components). The one that I could think of is `Cell` and I used the pattern used in this PR when we built that code.

e.g. https://github.com/Khan/wonder-blocks/blob/f74b7379e93875b9013a05b3f5a11655a5c67c72/packages/wonder-blocks-cell/src/components/internal/cell-core.js#L123

PR #1750, opened 2023-02-23
FEI-4966.5: Fix errors that weren't suppressed by the codemod

kevinbarabash:
## Summary:
I was able to fix most of the issues, but in a few places I had to add a TODO with a JIRA ticket to tackle later.

Issue: FEI-4966

## Test plan:
- yarn tsc --package tsconfig-check.json, see that there are no errors

### DIFF
--- a/packages/wonder-blocks-testing/src/fixtures/fixtures.basic.stories.tsx
+++ b/packages/wonder-blocks-testing/src/fixtures/fixtures.basic.stories.tsx
@@ -8,7 +8,7 @@ type Props = {
 };
 
 const MyComponent: React.FC<Props> = (props): React.ReactElement =>
-    <>`I am a component. Here are my props: ${JSON.stringify(props, null, 2)}`</>;
+    <>{`I am a component. Here are my props: ${JSON.stringify(props, null, 2)}`}</>;
 
 const Wrapper = (props: any) => (
     <>
--- a/packages/wonder-blocks-testing/src/fixtures/fixtures.basic.stories.tsx
+++ b/packages/wonder-blocks-testing/src/fixtures/fixtures.basic.stories.tsx
@@ -7,9 +7,8 @@ type Props = {
     propB?: string;
 };
 
-// @ts-expect-error [FEI-5019] - TS2322 - Type 'string' is not assignable to type 'ReactElement<any, string | JSXElementConstructor<any>>'.
 const MyComponent: React.FC<Props> = (props): React.ReactElement =>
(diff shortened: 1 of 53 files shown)


jeresig (approved):
Lots of fixes - thank you!

jeresig (approved):

packages/wonder-blocks-testing/src/fixtures/fixtures.basic.stories.tsx
```diff
@@ -7,9 +7,8 @@ type Props = {
     propB?: string;
 };
 
-// @ts-expect-error [FEI-5019] - TS2322 - Type 'string' is not assignable to type 'ReactElement<any, string | JSXElementConstructor<any>>'.
 const MyComponent: React.FC<Props> = (props): React.ReactElement =>
-    `I am a component. Here are my props: ${JSON.stringify(props, null, 2)}`;
+    <>`I am a component. Here are my props: ${JSON.stringify(props, null, 2)}`</>;
```
kevinbarabash, line 11: This kind of annoying.  Hopefully they improve the React types to allow returning strings from functional components.

kevinbarabash: I'm not sure if I did this right, since I think I may need to wrap the backticks in `{}`.  I double check this.

packages/wonder-blocks-timing/src/util/types.flowtest.ts
```diff
@@ -19,8 +19,8 @@ type Props1 = {
     test: string;
 } & WithActionSchedulerProps;
 
-// @ts-expect-error [FEI-5019] - TS2322 - Type 'string' is not assignable to type 'ReactElement<any, string | JSXElementConstructor<any>>'.
 const InnerComponent1: React.FC<Props1> = (props): React.ReactElement =>
+    // @ts-expect-error [FEI-5019] - TS2322 - Type 'string' is not assignable to type 'ReactElement<any, string | JSXElementConstructor<any>>'.
     props.test;
```
kevinbarabash, line 24: This can be updated to use the `<>{string}</>` so we can drop this `ts-expect-error`.

packages/wonder-blocks-testing/src/fixtures/fixtures.basic.stories.tsx
kevinbarabash: I got rid of this and added a suppression comment in a later PR in the stack.

PR #1751, opened 2023-02-23
FEI-4966.6: Add flow.d.ts, Remove flow suppressions and some flow types

kevinbarabash:
## Summary:
There are still a number of TypeScript errors that need to be fix.  Most of these appear to be HOC issues.  I'll migrate our HOCs in the next PR.  I've also changed the "Run Flow" check to "Typecheck" and it now runs "yarn typecheck" which is what I've been used to check for TS errors locally.  This will give reviewers a better idea of the progress being made.

Issue: FEI-4966

## Test plan:
- yarn typecheck, see that there are still about two dozen errors

### DIFF
--- a/packages/wonder-blocks-button/src/components/button-core.tsx
+++ b/packages/wonder-blocks-button/src/components/button-core.tsx
@@ -23,10 +23,7 @@ import type {SharedProps} from "./button";
 
 type Props = SharedProps &
     ChildrenProps &
-    ClickableState & {
-        href?: string;
-        type?: "submit";
-    };
+    ClickableState;
 
 const StyledAnchor = addStyle<"a">("a");
 const StyledButton = addStyle<"button">("button");
@@ -130,20 +127,19 @@ export default class ButtonCore extends React.Component<Props> {
             </Label>
         );
 
+        const sizeMapping = {
+            medium: "small",
+            small: "xsmall",
+            large: "medium",
+        } as const;
+
         const contents = (
             <React.Fragment>
                 {label}
                 {spinner && (
                     <CircularSpinner
                         style={sharedStyles.spinner}
-                        size={
-                            // @ts-expect-error [FEI-5019] - TS7053 - Element implicitly has an 'any' type because expression of type 'any' can't be used to index type '{ medium: string; small: string; large: string; }'.
-                            {
-                                medium: "small",
-                                small: "xsmall",
-                                large: "medium",
-                            }[size]
-                        }
+                        size={sizeMapping[size]}
                         light={kind === "primary"}
                         testId={`${testId || "button"}-spinner`}
                     />
--- a/packages/wonder-blocks-button/src/components/button.tsx
+++ b/packages/wonder-blocks-button/src/components/button.tsx
@@ -1,10 +1,8 @@
-// @ts-expect-error [FEI-5019] - TS2307 - Cannot find module 'flow-to-typescript-codemod' or its corresponding type declarations.
-import {Flow} from "flow-to-typescript-codemod";
 import * as React from "react";
 import {__RouterContext} from "react-router";
 
 import {getClickableBehavior} from "@khanacademy/wonder-blocks-clickable";
-import type {ClickableState} from "@khanacademy/wonder-blocks-clickable";
+import type {ClickableState, ChildrenProps} from "@khanacademy/wonder-blocks-clickable";
 import type {AriaProps, StyleType} from "@khanacademy/wonder-blocks-core";
 import type {IconAsset} from "@khanacademy/wonder-blocks-icon";
 import ButtonCore from "./button-core";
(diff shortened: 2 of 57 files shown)


packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -23,10 +23,7 @@ import type {SharedProps} from "./button";
 
 type Props = SharedProps &
     ChildrenProps &
-    ClickableState & {
-        href?: string;
-        type?: "submit";
-    };
```
kevinbarabash, line 29: There are included in `SharedProps` now.

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -130,20 +127,19 @@ export default class ButtonCore extends React.Component<Props> {
             </Label>
         );
 
+        const sizeMapping = {
+            medium: "small",
+            small: "xsmall",
+            large: "medium",
+        } as const;
```
kevinbarabash, line 134: The `as const` makes the object lookup behave the way we want it to.

packages/wonder-blocks-button/src/components/button.tsx
```diff
@@ -143,64 +143,30 @@ export type SharedProps =
...
-           */
-          href?: string;
-      })
-    | (SharedProps & {
-          /**
-           * Used for buttons within <form>s.
-           */
-          type: "submit";
-      })
-    | (SharedProps & {
```
kevinbarabash, line 165: I missed these in the pre-work b/c I was grepping for `CommonProps`.

packages/wonder-blocks-button/src/components/button.tsx
```diff
@@ -264,7 +230,7 @@ export default class Button extends React.Component<Props> {
             router,
         );
 
-        const renderProp = (state: ClickableState, {...restChildProps}) => {
```
kevinbarabash, line 267: The spread copies the object.  I don't think we need to do that here, but if people have concerns I can revert this change.

packages/wonder-blocks-clickable/src/components/clickable-behavior.ts
```diff
@@ -175,7 +175,7 @@ type DefaultProps = {
 };
 
 export type ChildrenProps = {
-    onClick: (e: React.MouseEvent) => unknown;
+    onClick: (e: React.SyntheticEvent) => unknown;
```
kevinbarabash, line 178: This is to align with `onClick` in a different object type that interacts with `ChildrenProps`.

packages/wonder-blocks-form/src/components/labeled-text-field.tsx
```diff
@@ -63,7 +61,7 @@ type Props = {
     /**
      * Identifies the element or elements that describes this text field.
      */
-    ariaDescribedby?: string | Array<string>;
+    ariaDescribedby?: string | undefined;
```
kevinbarabash, line 64: This aligns with TS' ARIA types.

packages/wonder-blocks-form/src/components/labeled-text-field.tsx
```diff
@@ -296,4 +294,4 @@ type ExportProps = Omit<
 
 export default React.forwardRef<HTMLInputElement, ExportProps>((props, ref) => (
     <LabeledTextField {...props} forwardedRef={ref} />
-)) as Flow.AbstractComponent<ExportProps, HTMLInputElement>;
+)) as React.ForwardRefExoticComponent<ExportProps & React.RefAttributes<HTMLInputElement>>;
```
kevinbarabash, line 297: This is based on the return type of `React.forwardRef`.  We might be able to drop it since these type params are already being passed to `React.forwardRef` itself.

packages/wonder-blocks-form/tsconfig.json
```diff
@@ -11,6 +11,7 @@
         {"path": "../wonder-blocks-core"},
         {"path": "../wonder-blocks-icon"},
         {"path": "../wonder-blocks-layout"},
+        {"path": "../wonder-blocks-link"},
```
kevinbarabash, line 14: I think the deps in some of our package.json files aren't quite correct.  `wonder-blocks-link` isn't listed even though it's used in stories.  It should probably be listed as a dev dependency.

packages/wonder-blocks-dropdown/src/components/dropdown-core-virtualized.tsx
```diff
@@ -177,20 +177,17 @@ class DropdownCoreVirtualized extends React.Component<Props, State> {
 
         return (
             // react-window has some issues for typing lists when passing refs
```
jeresig, line 179: We can remove these comments now.

jeresig (approved):
Thanks for all the cleanup!

PR #1753, opened 2023-02-24
Update SingleSelect documentation in Storybook

nishasy:
## Summary:
Currently the documentation for SingleSelect says the label is
passed as children, but the label is passed as a `label` prop.

I updated the documentation so it shows the label prop correctly.
I also added an example for how to add OptionItems with a map
because that's what it seems like most of the usage in webapp is.

image

Issue: none

## Test plan:
Copy/paste the example into a temporary test story and check that it works.


megaroeny (approved):

dnerdy (approved):
Lgtm. Thanks!

PR #1754, opened 2023-02-24
FEI-4966.8: Update eslint configuration and fix lint

kevinbarabash:
## Summary:
Some of the lint fixes were done using `yarn lint --fix` while other where done using find'n'replace to update rule names in `eslint-disable` comments.  The rest were manual changes.  The next PRs will update the babel and jest configs so that we can ensure that things build and that tests are passing.

TODO:
- [x] tweak some of the disabled rules to be disable for test files only

Issue: FEI-4966

## Test plan:
- yarn lint
- yarn typecheck


.eslintrc.js
```diff
@@ -88,7 +97,7 @@ module.exports = {
         "react/react-in-jsx-scope": "error",
         "react/require-render-return": "error",
         "monorepo/no-internal-import": "error",
-        "monorepo/no-relative-import": "error",
+        // "monorepo/no-relative-import": "error",
```
jeresig, line 100: Might want to copy the comment for this from wonder-stuff.

packages/wonder-blocks-banner/src/components/__docs__/banner.stories.tsx
```diff
@@ -12,6 +11,7 @@ import {LabelSmall} from "@khanacademy/wonder-blocks-typography";
 
 // @ts-expect-error [FEI-5019] - TS2305 - Module '"@storybook/react"' has no exported member 'StoryComponentType'.
 import type {StoryComponentType} from "@storybook/react";
+import Banner from "@khanacademy/wonder-blocks-banner";
```
jeresig, line 14: Why the move?

jeresig (approved):
Lots of fixes! Thanks for resolving all of this.

packages/wonder-blocks-banner/src/components/__docs__/banner.stories.tsx
kevinbarabash: Bad autofixing from `import/order`.  I can fix this.

.eslintrc.js
```diff
@@ -30,17 +39,32 @@ module.exports = {
...
+        // from node_modules/@types/react/index.d.ts
+        JSX: false,
+        // from node_modules/typescript/lib/lib.dom.d.ts
+        DOMHighResTimeStamp: false,
+        RequestInfo: false,
+        RequestInit: false,
+        // from types/utility.d.ts
+        Empty: false,
+        ObjMap: false,
+        SpreadType: false,
```
kevinbarabash, line 62: Some people suggest disable `no-undef` all the time, e.g. https://stackoverflow.com/questions/65054079/eslint-with-typescript-and-firebase-no-undef-error.  That doesn't seem like a good solution.  Eslint should be able to tell the difference between value identifier and type identifiers and ignore just the undefined type identifiers.

PR #1761, opened 2023-02-27
Remove Clickable from Choice

nishasy:
## Summary:
Currently, there is an accessibility issue where screenreaders
are reading the wrong checked state for checkboxes and radio
buttons on iOS.

It seems like the cause of this issue has to do with
ClickableBehavior. We can fix this by removing ClickableBehavior
from choice-js.

- Remove ClickableBehavior from choice.js
- Make the styles use pseudo selectors instead of relying on the
  clickable states

Issue: https://khanacademy.atlassian.net/browse/WB-1427

## Test plan:
`yarn jest`

Manual testing
- Go to http://localhost:58796/?path=/story/form-checkbox--variants-controlled
- Hover on the unchecked checkboxes and confirm it looks the same as before.
- Unhover on the checked checkboxes and confirm it looks the same as before.
- Press down on the unchecked checkboxes and confirm it looks the same as before.
- Unpress and confirm that the checked state changes, and it the newly checked checkbox look the same as before.
- Repleat steps for checked checkboxes.
- Turn on a screen reader and confirm that the checked states are all correct.

- Repeat the above steps for radio buttons at http://localhost:58796/?path=/story/form-radiogroup--basic

- Repeat the above steps for checkboxes and radio buttons on storybook on a mobile device.

### Checkbox Before

https://user-images.githubusercontent.com/13231763/221702662-0bc6a99a-84ce-4a27-a281-97cad2134888.mov

### Checkbox After

https://user-images.githubusercontent.com/13231763/221702718-bed8bbf6-9a07-4e7e-a8a3-7b9ecfb7d16a.mov

### Radio Before

https://user-images.githubusercontent.com/13231763/221702683-14ae6137-482f-4ad8-9556-703133ad84b7.MP4

### Radio After

https://user-images.githubusercontent.com/13231763/221702820-61b105ab-e736-447f-bd64-f1abaac92266.mov


nishasy:
Marked as a draft because I want to wait for input about whether we need to keep the labels clickable or not.

packages/wonder-blocks-form/src/__tests__/custom-snapshot.test.tsx
```diff
@@ -5,60 +5,45 @@ import CheckboxCore from "../components/checkbox-core";
...
-                                waiting={false}
-                            />,
-                        )
-                        .toJSON();
-                    expect(tree).toMatchSnapshot();
-                });
+        checkedStates.forEach((checked: any) => {
+            test(`type:${state} checked:${String(checked)}`, () => {
+                const disabled = state === "disabled";
+                const tree = renderer
```
kevinbarabash, line 15: **off-topic:** We should probably update this to use React Testing Library at some point.  Not req'd for this PR.

packages/wonder-blocks-form/src/components/choice-internal.tsx
```diff
@@ -115,51 +106,53 @@ type DefaultProps = {
...
         return (
             <UniqueIDProvider mockOnFirstRender={true} scope="choice">
                 {(ids) => {
+                    // A choice element should always have a unique ID set
+                    // so that the label can always refer to this element.
+                    // This guarantees that clicking on the label will
+                    // always click on the choice as well. If an ID is
+                    // passed in as a prop, use that one. Otherwise,
+                    // create a unique ID using the provider.
+                    const uniqueId = id || ids.get("main");
```
kevinbarabash, line 129: Nice!

kevinbarabash (approved):
LGTM.  I like that the consistency tests still pass without any modifications.  Can you wait on landing this until I've done a publish with the TypeScript changes?

PR #1756, opened 2023-02-27
FEI-4966.10: Experimental branch to get .d.ts files building [DO NOT LAND]

kevinbarabash:
## Summary:
This is only an experiment to see if moving docs and stories somewhere else will fix the issues with circular dependencies.  This PR confirms that is a valid solution by deleting all of the docs and stories.  In a seperate PR I'll do the work of moving these files and creating a new home for them.

Issue: FEI-4966

## Test plan:
- yarn build
- yarn build:types
- see that .d.ts files are written to the dist folder for each package

PR #1762, opened 2023-02-28
FEI-4966.10: Move docs out of packages and into top-level docs folder

kevinbarabash:
## Summary:
This was necessary b/c TypeScript doesn't support circular dependencies when using project references.  The only reason why we had circular deps was because of the documentation.  This was borne out in #1756.

One downside of this approach is that inline docstrings aren't included in the generated docs.  That issue can be resolved later.

Issue: FEI-4966

## Test plan:
- yarn build:storybook
- yarn start

**Overview: After/Before**
Screen Shot 2023-02-28 at 11 32 38 AM

**Choice: After/Before**
Screen Shot 2023-02-28 at 11 35 55 AM


jeresig (approved):
Blurgh... oh well :( Thanks for fixing this!

kevinbarabash:
@jeresig re-requesting review due to all of the fixes for type errors in the stories.  I also bumped `@babel/core` in order to support the use of the `satisfies` operator.

jeresig (approved):
Wow - it looks like this caught a bunch of issues? Thanks for fixing this up!

PR #1764, opened 2023-03-01
FEI-4966.11: Remove almost all traces of Flow

kevinbarabash:
## Summary:
There are still a few links to flow related docs.  I'm going to be updating the aphrodite types in the next PR to use TypeScript DOM lib defs so that'll go away.  As for HOCs we don't have that fully sorted yet so I'm reluctant to update docs to point to something that's half-baked.

Issue: FEI-4966

## Test plan:
- let CI do its thing


kevinbarabash:
I'm not sure why `rollup` is failing in the "Check build sizes" job.  I tried running the same command locally and it works without error.  I need to create another PR anyways so maybe things will resolve themselves in the next one. 🤞

jeresig (approved):
Nice - thanks for cleaning this up!

PR #1767, opened 2023-03-01
FEI-4966.13: Add build and build:types to PR workflow

kevinbarabash:
## Summary:
This PR adds npmPkgJsonLint to make sure our package.json files have all the fields we want them to have.  This tool unfortunately does not check that the fields point to valid files so we still need our existing pre-flight check.

Issue: FEI-4966

TODO:
- [x] add a 'types' field to all of our package.json files
- [x] fix issue with i18n-inline-markup.d.ts

## Test plan:
- yarn build:types
- yarn lint:pkg-json
- node utils/publish/pre-publish-check-ci.js


jeresig (approved):
Very nice!

PR #1768, opened 2023-03-01
Fix babel config to work with Storybook

kevinbarabash:
## Summary:
Storybook was working in #1762, but now it isn't.  The issue is that we're missing a number of plugins.  This PR adds those plugins to our babel config.  

~~We might be able to get away with only adding them when running storybook, but I'd rather keep the config as close as possible to what we use when building the published components.~~

It turns out that the extra babel plugins break out tests.  I trust our tests more than I trust Storybook so I've decided to move the new plugins into storybook's babel config which builds atop of ours.

Issue: None

## Test plan:
- yarn build
- yarn build:storybook
- yarn start

```
...
info => Preview built (31 s)
WARN asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
WARN This can impact web performance.
WARN Assets: 
WARN   main.6604b071.iframe.bundle.js (1.77 MiB)
WARN   vendors~main.6835c72b.iframe.bundle.js (3.56 MiB)
WARN   6.3a3124c2.iframe.bundle.js (746 KiB)
WARN   10.bc961892.iframe.bundle.js (682 KiB)
WARN entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
WARN Entrypoints:
WARN   main (5.34 MiB)
WARN       runtime~main.8d470283.iframe.bundle.js
WARN       vendors~main.6835c72b.iframe.bundle.js
WARN       main.6604b071.iframe.bundle.js
WARN 
info => Output directory: /Users/kevinbarabash/khan/wonder-blocks/storybook-static
✨  Done in 33.96s.
```

Screen Shot 2023-03-01 at 6 28 10 PM


.storybook/main.js
```diff
@@ -27,7 +27,6 @@ module.exports = {
     // monorepo.
     // See https://storybook.js.org/docs/react/configure/webpack
     webpackFinal: async (config, {configType}) => {
-        const wbPkgJsonRegex = /^@khanacademy\/wonder-blocks(-.*)\/package.json$/;
```
kevinbarabash, line 30: I removed this style of imports a while ago but forgot to clean this up so I'm doing so now.

nishasy (approved):

PR #1775, opened 2023-03-02
Make text highlightable for checkboxes and radio buttons

nishasy:
## Summary:
Currently, choice-internal has the style `userSelect: "none"`,
which stops checkboxes' and radio buttons' labels from being
text highlightable.

[Here](https://github.com/Khan/wonder-blocks/pull/287#issue-351326095) is a comment explaining why it was added in the first place.

There has been a discussion since then about text highlighting a11y,
and we want to bring back highlighting and remove `userSelect: "none"`,
at least just for checkboxes and radios. (Discussion can be found in the 
Jira ticket below.)

Issue: https://khanacademy.atlassian.net/browse/WB-1512

## Test plan:
- Go to http://localhost:53078/?path=/story/form-checkboxgroup--basic
- Drag across the text and confirm that it highlights.
- Confirm that dragging along the text does not toggle the checkbox state.
- Confirm the checkbox works as expected otherwise.

- Do the same with http://localhost:53078/?path=/story/form-radiogroup--basic


kevinbarabash (approved):
🚢

PR #1776, opened 2023-03-03
Fix misc TS types issues

kevinbarabash:
## Summary:
- make sure that the generated flow type use 'extends' instead of 'mixins'
- fix export types of i18n functions

Issue: None

## Test plan:
- yarn typecheck
- let CI run


.changeset/flat-clouds-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-i18n": patch
```
jeresig, line 2: I notice that only i18n is flagged as having changes, even though the build-script will almost certainly change all the packages (and require republishing them all).

build-scripts/gen-flow-types.ts
```diff
@@ -14,6 +15,18 @@ for (const inFile of files) {
...
+        // so we do some post-processing on the files to clean that up.
+        const contents = fs.readFileSync(path.join(rootDir, outFile), "utf-8");
+        if (contents.includes(" mixins ")) {
+            console.log("replacing 'mixins' with 'extends'");
+            fs.writeFileSync(
+                path.join(rootDir, outFile),
+                contents.replace(/ mixins /g, " extends "),
+                "utf-8",
+            );
+        }
```
kevinbarabash, line 29: I must've been looking at one of the earlier PRs instead what's on `main` in the wonder-stuff repo.

jeresig (approved):

packages/wonder-blocks-i18n/src/functions/i18n.ts
```diff
@@ -233,8 +232,12 @@ export const $_: (
  * In Russian, any number ending in 1 (e.g. 21, 61) uses singular form.
  * You should instead still parameterize the number.
  */
-// @ts-expect-error [FEI-5019] - TS2322 - Type '(singular: string, plural: string, num: number | null | undefined, options: NGetOptions | undefined) => string' is not assignable to type 'ngettextOverloads'.
-export const ngettext: ngettextOverloads = (singular, plural, num, options) => {
```
kevinbarabash, line 237: This was causing issues in the perseus repo when I tried to update it.

.changeset/flat-clouds-sniff.md
kevinbarabash: Good catch.  I'll move the build-scripts into build-settings.

PR #1779, opened 2023-03-04
Use solid underline for links (until audit can be completed)

nishasy:
## Summary:
While the new dashed link is available, we are updating it to the old underline
style to match the non-wonder blocks links across webapp. An audit is currently
being run so we can update all these to wonder blocks link.

Once all the legacy links have been updated to use Wonder Blocks Link, we can
finally use the dashed link style.

- Update styles so that the links use solid 1px underlines.
- Comment out the old styles and add a TODO comment above them so we can easily go back.
  - The TODOs should reference WB-1521.

Issue: https://khanacademy.atlassian.net/browse/WB-1520

## Test plan:
- Go to http://localhost:53078/?path=/story/link--variants
- Confirm all the chromatic interaction tests are passing
- Confirm all the links have 1px solid underlines with no offset.

### DIFF
--- a/packages/wonder-blocks-link/src/components/link-core.tsx
+++ b/packages/wonder-blocks-link/src/components/link-core.tsx
@@ -94,8 +94,6 @@ const sharedStyles = StyleSheet.create({
         textDecoration: "none",
         outline: "none",
         display: "inline-flex",
-        fontSize: 16,
-        lineHeight: "22px",
     },
 });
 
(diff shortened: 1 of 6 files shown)


nishasy:
Note that all the UI Test updates on chromatic are from the line height change.

packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -165,14 +163,24 @@ const _generateStyles = (
         },
         restingInline: {
             color: defaultTextColor,
+            // TODO(WB-1521): Update text decoration to the dashed underline
+            // after the Link audit.
+            // textDecoration: "underline currentcolor solid 1px",
```
kevinbarabash, line 168: **question:** should this be `2px`?

packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -165,14 +163,24 @@ const _generateStyles = (
...
             textDecoration: "underline currentcolor solid 1px",
-            textUnderlineOffset: 4,
+            // TODO(WB-1521): Update the underline offset to be 4px after
+            // the Link audit.
+            // textUnderlineOffset: 4,
```
kevinbarabash, line 172: **praise:** I like how you left the old code with a JIRA ticket TODO.

kevinbarabash (approved):
LGTM. Thanks for temporarily reverting this until the audit can be completed.💚

packages/wonder-blocks-link/src/components/link-core.tsx
nishasy: Hmm... I actually think it should be `auto`.

PR #1780, opened 2023-03-06
FEI-5038: Update AriaProps to use React's lib defs

kevinbarabash:
## Summary:
This will keep our aria props in sync with the official ones included in React's lib defs.  One downside is that 'AriaRole' can be any string, although it is constructed in such a way that intellisense will suggest valid roles.

Issue: FEI-5038

## Test plan:
- yarn typecheck


jeresig (approved):
That's too bad that the role is a string now!

kevinbarabash:
> That's too bad that the role is a string now!

Yeah.  I asked the author of the types the reason behind that choice.  Hopefully they get back to me.  We could also open up a PR and try to get rid of that part of the type.

PR #1784, opened 2023-03-07
Replace new TS React types

kevinbarabash:
## Summary:
flowgen hasn't been updated in a while and the TS lib def for React keeps getting updated.  As a result there are some new types that the tool doesn't know about.  This PR adds some more post-processing logic to gen-flow-types.ts to rewrite "JSX.LibraryManagedAttributes<>" as "React.ElementConfig<>".

Issue: None

## Test plan:
- test the regex replacement code manually on a test string
- yarn build:types
- yarn build:flowtypes
- see that it's doing the expected replacement in the logging

```
✅ wrote: packages/wonder-blocks-clickable/dist/util/get-clickable-behavior.js.flow
replacing 'JSX.LibraryManagedAttributes<
    typeof ClickableBehavior,
    React.ComponentProps<typeof ClickableBehavior>
  >' with 'React.ElementConfig<typeof ClickableBehavior>'
```

### DIFF
--- a/.changeset/fluffy-humans-kiss.md
+++ b/.changeset/fluffy-humans-kiss.md
@@ -1,0 +1,32 @@
+---
+"wb-dev-build-settings": patch
+"@khanacademy/wonder-blocks-banner": patch
+"@khanacademy/wonder-blocks-birthday-picker": patch
+"@khanacademy/wonder-blocks-breadcrumbs": patch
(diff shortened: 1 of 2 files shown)


kevinbarabash:
FYI, I bumped all versions b/c I'm too lazy to see which packages are affected by this issue.

PR #1786, opened 2023-03-07
add title to link

n1colemejia:
## Summary:
Add title prop to Link

Issue: WB-1453

## Test plan:
- Ran yarn test & yarn typecheck
- Set title prop in control panel and inspected title attribute change in the DOM
- Added WithTitle story that renders link with a title attribute that appears when hovering over the link
  Screenshot 2023-03-08 at 10 28 42 AM
  Screenshot 2023-03-08 at 10 28 50 AM

### DIFF
--- a/.changeset/metal-worms-heal.md
+++ b/.changeset/metal-worms-heal.md
@@ -1,5 +1,0 @@
----
-"@khanacademy/wonder-blocks-link": minor
----
-
-adds title to link.argtypes
--- a/.changeset/metal-worms-heal.md
+++ b/.changeset/metal-worms-heal.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-link": minor
+---
+
+adds title to link.argtypes
--- a/__docs__/wonder-blocks-link/link.stories.tsx
+++ b/__docs__/wonder-blocks-link/link.stories.tsx
@@ -479,6 +479,30 @@ Navigation.parameters = {
     },
 };
 
+export const WithTitle: StoryComponentType = () => (
+    <Body>
+        <Link href="#" title={"I am a title 😎"}>
+            This link has a title.
+        </Link>
+    </Body>
+);
+
+WithTitle.parameters = {
+    docs: {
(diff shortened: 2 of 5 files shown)


.changeset/metal-worms-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-link": minor
+---
+
+adds title to link.argtypes
```
nishasy, line 5: We only need on changeset per package, and we'd only include changes in the component implementation. Since argytpes are part of the documentation update, I don't think you need to include this in the changset. So you can remove `metal-worms-heal.md` altogether 🙂

nishasy:
In the test plan, I'd also add that you tested setting a title in the control panel and then hovering over the link and confirming that the title shows up after a few seconds. It would be nice to include a screenshot of this in the description 🙂

nishasy:
I also think it could be cool to add a story (maybe called `WithTitle` where you just render a link with a title attribute. You could set the story description to explain what the `title` prop does and what behavior people should expect to see when interacting with this story.

.changeset/metal-worms-heal.md
n1colemejia: Got it! 👌

__docs__/wonder-blocks-link/link.stories.tsx
```diff
@@ -479,6 +479,30 @@ Navigation.parameters = {
     },
 };
 
+export const WithTitle: StoryComponentType = () => (
+    <Body>
+        <Link href="#" title={"I am a title 😎"}>
```
somewhatabstract, line 484: **nit:** Since you're just passing a string literal here, you don't need the braces. It doesn't matter if you leave them either. Up to you if you change this:
```suggestion
        <Link href="#" title="I am a title 😎">
```

somewhatabstract (approved):
Thanks for addressing @nishasy's comments. This looks good to me.

nishasy (approved):
Thanks for the updates! LGTM 🎉

PR #1791, opened 2023-03-09
Add useLatestRef hook to wonder-blocks-core

benchristel:
## Summary:
Jeff suggested this here: https://khanacademy.slack.com/archives/C0918TZ5G/p1677879618274909?thread_ts=1677878402.389499&cid=C0918TZ5G

Issue: https://khanacademy.atlassian.net/browse/FEI-5036

## Test plan:

View the Storybook docs at
/?path=/story/core-exports-uselatestref--page

Screen Shot 2023-03-15 at 2 24 01 PM


__docs__/wonder-blocks-core/exports.use-latest-ref.stories.mdx
```diff
@@ -0,0 +1,95 @@
...
+```ts
+function useLatestRef<T>(value: T): React.RefObject<T>;
+```
+
+The `useLatestRef` hook returns a ref that always contains the `value` passed
+to the hook during the most recent render.
+
+It can be used to wrap a possibly-changing prop in a stable value that can
+be passed to `useEffect`'s dependency array without causing unnecessary
+re-renders. For example:
```
somewhatabstract, line 30: **suggestion:** I wonder if it's worth adding more illustrative info about why this is a valuable alternative to `useMemo`; for example, it allows for a value to be used by an effect without having to retrigger the effect function when it changes (or the effect's "unmount" call).

benchristel: I added a bit of detail. I'm not sure I understand the comparison to `useMemo`. Can `useMemo` be used to solve this problem?

somewhatabstract: No, but the normal way to track a changed object but only change it when appropriate is to `useMemo` on that object, then use the memoized value in the dependency array. This removes the need to do that - you could save a whole object in the ref and then close around that in the effect, and avoid needing to update the effect (or any other hook that uses deps, like `useCallback`) just because the value changed.

To put it more plainly, the first instinct for folks might be to use `useMemo` when this may be a better, simpler option if the value is an object to be used in a closure.

somewhatabstract: Writing hooks is really hard to get right as it involves thinking through a lot of interactions to determine how things will behave, including how React will do things.

__docs__/wonder-blocks-core/exports.use-latest-ref.stories.mdx
```diff
@@ -0,0 +1,95 @@
...
+        <View style={{flexDirection: "row", gap: Spacing.xSmall_8}}>
+            <Button onClick={remount}>Remount</Button>
+            <Button kind="secondary" onClick={unmount}>Unmount</Button>
+        </View>
+        <Strut size={Spacing.medium_16}/>
+        <Context>
+            {remounts > 0 && <ComponentWithDelayedAlert key={remounts} message={message} />}
+        </Context>
+    </Context>
+}
```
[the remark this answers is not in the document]
somewhatabstract: I wonder if the example should have something that emphasizes how the effect is not rerunning on the ref value changing. I think an example that demonstrates the value of this hook, rather than just demonstrates it working at all might be more useful.

Also, since the component is remounting over and over, I don't think the `useLatestRef` is adding any value as the whole component is tossed and remounted - i.e. there is no last ref to use. I think an example where the prop changes repeatedly and is passed through `useLatestRef` makes more sense otherwise, in the example as it stands, it isn't demonstrating the value of `useLatestRef` at all.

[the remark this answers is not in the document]
benchristel: I think `useLatestRef` is affecting the behavior, but I agree that the demo is pretty unmotivated, to the point where it's hard to see what the expected behavior is.

[The original motivation](https://khanacademy.slack.com/archives/C0918TZ5G/p1677878402389499) for `useLatestRef` was the need to send an analytics event when the component _unmounts_, using the props from the most recent render. I've updated the example to be closer to that. What do you think?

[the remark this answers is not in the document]
somewhatabstract: Oh yeah, that's cleaner. Now it can update without being remounted but you can show how remounting would work. Long term, it'd be really cool to show the comparison of render cycles and performance of this approach versus the "traditional" deps array approach to show how just using `useIsMounted` and `useEffect` with the prop in the deps array would lead to more render cycles when doing it that way.

somewhatabstract (approved):
I think this looks good. Thanks for taking the time to get a good example in the docs, @benchristel. Hopefully we can build on this in future to demonstrate the perf benefits of using this too.

benchristel:
@somewhatabstract how should I go about landing this? I notice that there are UI tests that require manual review. Glancing at those tests, it seems like there are regressions, but I don't understand how my change would have caused them.

somewhatabstract:
We are holding off on landing changes for a little bit while we get the initial TypeScript migration steps done such that we can integrate into webapp. Then we'll let folks land again, so don't worry about it for now - I'll ping you as soon as we're good to go.

As for the regressions, if they are in the tooltip component, it's a known issue and is down to a race between the component's position stabilizing and when the CI server takes the snapshot. If it is the tooltip component, we generally just accept the changes ourselves.

somewhatabstract:
@benchristel WB is open for developers to land changes now :)

Make sure to verify that the generated flow types work within WebApp. I found the easiest way to do this was to build the packages locally, then delete the WB package from static service's node modules and copy in the locally built one, then restart flow, then try using the thing and checking that it can be used as expected with appropriate flow coverage.

It's a bit arduous while we're in this temporary flow->TS transition period, but it is effective.

benchristel:
I don't think the generated types are working. Using `yarn flow type-at-pos`, I determined that the return type for `useLatestRef` is `any (implicit)`. The generated type for `useLatestRef` is:

```ts
export declare function useLatestRef<T>(value: T): React.RefObject<T>;
```

But the `React.RefObject<T>` type doesn't seem to exist in Flow. The Flow types for React declare `useRef` as

```
declare export function useRef<T>(initialValue: ?T): {current: T | null};
```

So I think the generated return type for `useLatestRef` should be something like `{|current: T|}`.

Is the path to fixing this to add a string replacement to `gen-flow-types.ts`?

somewhatabstract:
@benchristel Yeah, the quickest fix is to add a custom flow file for this so that it doesn't try and auto-generate. There are examples in Wonder Blocks. Basically, just create a file like `useLatestRef.js.flow` and have that export the right thing :)

benchristel:
I've added custom Flow types and verified that they work in webapp. @somewhatabstract want to take a final look at this before I merge?

packages/wonder-blocks-core/src/hooks/use-latest-ref.js.flow
```diff
@@ -0,0 +1,14 @@
...
+ * @flow
+ */
+
+/**
+ * The useLatestRef hook returns a ref that always contains the `value` passed
+ * to the hook during the most recent render.
+ *
+ * It can be used to wrap a possibly-changing prop in a stable value that can
+ * be passed to useEffect without causing unnecessary re-renders. See the
+ * Storybook docs for a usage example.
```
somewhatabstract, line 12: **nit:** We use the imperative and a single line for the first part of a function description.
```suggestion
 * Return a ref that always contains the `value` passed.
 *
 * The useLatestRef hook returns a ref that always contains the `value` passed
 * to the hook during the most recent render.
 *
 * It can be used to wrap a possibly-changing prop in a stable value that can
 * be passed to useEffect without causing unnecessary re-renders. See the
 * Storybook docs for a usage example.
```

packages/wonder-blocks-core/src/hooks/use-latest-ref.ts
```diff
@@ -0,0 +1,15 @@
+import * as React from "react";
+
+/**
+ * The useLatestRef hook returns a ref that always contains the `value` passed
+ * to the hook during the most recent render.
+ *
+ * It can be used to wrap a possibly-changing prop in a stable value that can
+ * be passed to useEffect without causing unnecessary re-renders. See the
+ * Storybook docs for a usage example.
```
somewhatabstract, line 9: **nit:** We use the imperative and a single line for the first part of a function description.
```suggestion
 * Return a ref that always contains the `value` passed.
 *
 * The useLatestRef hook returns a ref that always contains the `value` passed
 * to the hook during the most recent render.
 *
 * It can be used to wrap a possibly-changing prop in a stable value that can
 * be passed to useEffect without causing unnecessary re-renders. See the
 * Storybook docs for a usage example.
```

somewhatabstract (approved):
LGTM @benchristel 

A couple of inline nits, but nothing blocking.

somewhatabstract:
Thanks for taking the time to verify the flow types with webapp :)

packages/wonder-blocks-core/src/hooks/use-latest-ref.js.flow
```diff
@@ -0,0 +1,14 @@
...
+ */
+declare export function useLatestRef<T>(value: T): {|current: T|};
```
somewhatabstract, line 14: **nit:** I believe we do have a type for this in Flow - but this should be fine as-is.

PR #1792, opened 2023-03-13
Tweak TS Types #1

kevinbarabash:
## Summary:
When updating wonder-blocks deps in webapp, I ran into a number of issues.  This PR fixes some of them, namely:
- isSet() is supposed to be a getter, but flow-to-typescript-codemod drop the 'getter'-ness of getters on interfaces
- the 'children' function on function-as-children components should be allowed to return a ReactNode so that children are more flexible in what they can return

I also made the return type of all render methods 'ReactNode' since that's what it is in the React.Component class definition.

Issue: None

## Test plan:
- yarn typecheck
- let CI run

### DIFF
--- a/packages/wonder-blocks-modal/src/components/one-pane-dialog.tsx
+++ b/packages/wonder-blocks-modal/src/components/one-pane-dialog.tsx
@@ -178,7 +178,7 @@ export default class OnePaneDialog extends React.Component<Props> {
         }
     }
 
-    render(): React.ReactElement {
+    render(): React.ReactNode {
         const {
             onClose,
             footer,
--- a/packages/wonder-blocks-modal/src/components/one-pane-dialog.tsx
+++ b/packages/wonder-blocks-modal/src/components/one-pane-dialog.tsx
@@ -74,7 +74,7 @@ type Common = {
     /**
      * The ID of the content describing this dialog, if applicable.
      */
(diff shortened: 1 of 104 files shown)


packages/wonder-blocks-timing/src/util/types.ts
```diff
@@ -18,7 +18,7 @@ export interface ITimeout {
      * false.
      * @memberof ITimeout
      */
-    isSet(): boolean;
+    get isSet(): boolean;
```
jeresig, line 21: This probably isn't common enough to update the codemod, but maybe worth mentioning in the upgrade guide?

kevinbarabash: I was planning on fixing it before converting Perseus.  I don't think it'll take long to do.

packages/wonder-blocks-modal/src/components/one-pane-dialog.tsx
```diff
@@ -74,7 +74,7 @@ type Common = {
     /**
      * The ID of the content describing this dialog, if applicable.
      */
-    ["aria-describedby"]?: string;
```
kevinbarabash, line 77: @jeresig this was causing the properties to change from optional to required by `flowgen`.  This fixes the issue.

PR #1794, opened 2023-03-13
Add external icon when using `target="_blank"`

n1colemejia:
## Summary:
- Adds external icon to link when `target="_blank"`
- Adds story for links that open in a new tab (see screenshot)
- Adds external links with icon to Variants and Inline stories

Issue: WB-1452

## Test plan:
- Ran yarn test & yarn typecheck
- Set target in control panel and inspected attribute change and icon appearance in the DOM
- Added tests in link.test:
    - test the external icon is rendered for a link using `target="_blank"`
    - test no icon is rendered when not using `target="_blank"`
    - test link has a target attribute and target's value equals "_blank"

Screenshot 2023-03-14 at 4 08 44 PM


__docs__/wonder-blocks-link/link.stories.tsx
```diff
@@ -479,6 +563,27 @@ Navigation.parameters = {
...
+);
+
+OpensInANewTab.parameters = {
+    docs: {
+        storyDescription: `When \`target="_blank"\`, the external icon is
+        automatically added to the end of the link. This indicates that the link
+        will open in a new tab.`,
+    },
+};
+
```
nishasy, line 586: *Super nit*: Can you add this story right before the `Variants` story? I personally prefer for all the variants in variants to be listed before the variants story if we're showing them separately, but it really doesn't matter either way.

__docs__/wonder-blocks-link/link.argtypes.ts
```diff
@@ -81,6 +81,15 @@ export default {
...
 
+    target: {
+        control: {type: "text"},
+        description: `A target destination window for a link to open in.
+            We only support "_blank" which opens the URL in a new tab.`,
+        table: {
+            type: {summary: "string"},
+        },
+    },
+
```
nishasy, line 92: Nice!! I'm glad you added this. I'm not sure why it wasn't there before.

nishasy (approved):
This is awesome! Great job navigating this task and communicating with Reony to ensure it's pixel perfect.

__docs__/wonder-blocks-link/link.stories.tsx
somewhatabstract, line 1: **praise:** Love these story updates! Thank you!

packages/wonder-blocks-link/src/components/__tests__/link.test.tsx
```diff
@@ -436,4 +436,49 @@ describe("Link", () => {
...
+            // Act
+            const link = screen.getByText("Click me!");
+            const icon = screen.getByTestId("external-icon");
+
+            // Assert
+            expect(link.innerHTML).toEqual(expect.stringContaining("<svg"));
+            expect(icon).toBeInTheDocument();
+        });
+
+        test("does not render icon", () => {
```
somewhatabstract, line 473: **suggestion:** Test case names should indicate an outcome and the condition that leads to that outcome, like a requirement. For example, in this case, "does not render icon when there is no target".

This helps folks who encounter a test failure understand the intention of the test; that under X circumstances we should expect Y behaviour.

somewhatabstract (changes requested):
It looks like you need to make sure that the `wonder-blocks-link` package is referencing the `wonder-blocks-icon` package in the `package.json` in `packages/wonder-blocks-link`.

In the `dependencies` section:

```
 "dependencies": {
    "@babel/runtime": "^7.18.6",
    "@khanacademy/wonder-blocks-clickable": "^2.4.6",
    "@khanacademy/wonder-blocks-color": "^1.2.2",
    "@khanacademy/wonder-blocks-core": "^4.8.0"
  },
```

packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -6,12 +6,15 @@ import {__RouterContext} from "react-router";
 import {addStyle} from "@khanacademy/wonder-blocks-core";
 import Color, {mix, fade} from "@khanacademy/wonder-blocks-color";
 import {isClientSideUrl} from "@khanacademy/wonder-blocks-clickable";
+import Icon from "@khanacademy/wonder-blocks-icon";
+import Spacing from "@khanacademy/wonder-blocks-spacing";
```
kevinbarabash, line 10: These error messages mean that references for these packages need to be added to the tsconfig.json file for wonder-blocks-link.

packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -63,16 +67,37 @@ export default class LinkCore extends React.Component<Props> {
...
+            small: "M14 6.5C14 6.63261 13.9473 6.75979 13.8536 6.85355C13.7598 6.94732 13.6326 7 13.5 7C13.3674 7 13.2402 6.94732 13.1464 6.85355C13.0527 6.75979 13 6.63261 13 6.5V3.7075L8.85437 7.85375C8.76055 7.94757 8.63331 8.00028 8.50062 8.00028C8.36794 8.00028 8.2407 7.94757 8.14688 7.85375C8.05306 7.75993 8.00035 7.63268 8.00035 7.5C8.00035 7.36732 8.05306 7.24007 8.14688 7.14625L12.2925 3H9.5C9.36739 3 9.24021 2.94732 9.14645 2.85355C9.05268 2.75979 9 2.63261 9 2.5C9 2.36739 9.05268 2.24021 9.14645 2.14645C9.24021 2.05268 9.36739 2 9.5 2H13.5C13.6326 2 13.7598 2.05268 13.8536 2.14645C13.9473 2.24021 14 2.36739 14 2.5V6.5ZM11.5 8C11.3674 8 11.2402 8.05268 11.1464 8.14645C11.0527 8.24021 11 8.36739 11 8.5V13H3V5H7.5C7.63261 5 7.75979 4.94732 7.85355 4.85355C7.94732 4.75979 8 4.63261 8 4.5C8 4.36739 7.94732 4.24021 7.85355 4.14645C7.75979 4.05268 7.63261 4 7.5 4H3C2.73478 4 2.48043 4.10536 2.29289 4.29289C2.10536 4.48043 2 4.73478 2 5V13C2 13.2652 2.10536 13.5196 2.29289 13.7071C2.48043 13.8946 2.73478 14 3 14H11C11.2652 14 11.5196 13.8946 11.7071 13.7071C11.8946 13.5196 12 13.2652 12 13V8.5C12 8.36739 11.9473 8.24021 11.8536 8.14645C11.7598 8.05268 11.6326 8 11.5 8Z",
```
kevinbarabash, line 75: **suggestion (non-blocking):** The numeric precision on these paths is higher than it needs to be.  3 digits of precision is more than enough.

kevinbarabash (changes requested):
Thanks for adding the external icon.  I've been asking about this for a long time so I'm excited to see it get added.

Requesting changes to fix an issue with the tsconfig.json file in wonder-blocks-link.  See inline comments for details.

packages/wonder-blocks-link/src/components/link-core.tsx
n1colemejia: Thanks for pointing that out! 🫢 Is there a tool or extension I should use to adjust that?

somewhatabstract (approved):
LGTM

kevinbarabash (approved):
LGTM.  Thanks for updating the tsconfig.json.

PR #1799, opened 2023-03-15
FEI-5042.2: Fix wonder-blocks-core flow types (take 2)

kevinbarabash:
## Summary:
Flowgen's handling of inexact types wasn't good enough to make Flow happy when the types were deploy in webapp.  This PR fixes the issue by introducing override files.  If an .js.flow override exists for a given output path, we copy that file over to the output path and skip running flowgen completely.  I expect there to only be a handful of files where this is necessary.

Issue: FEI-5042

## Test plan:
- temporarily update the "files" glob to only capture wonder-blocks-core files
- yarn build:types
- yarn build:flowtypes
- check that view.js.flow and text.js.flow were successfully copied into the dist folder


packages/wonder-blocks-core/src/components/text.tsx
```diff
@@ -1,3 +1,5 @@
+// WARNING: If you modify this file you must update the following file:
+// build-settings/overrides/wonder-blocks-core/components/view.js.flow
```
jeresig, line 2:
```suggestion
// build-settings/overrides/wonder-blocks-core/components/text.js.flow
```

jeresig (approved):

jeresig:
Hmm - you have these files live in build-settings, but it seems like it would make more sense for them to live in the package themselves, correct? That way we can know how to bump the version of the package.

jeresig (changes requested):
Requesting changes to move the js.flow files into the packages themselves.

kevinbarabash:
> Hmm - you have these files live in build-settings, but it seems like it would make more sense for them to live in the package themselves, correct? That way we can know how to bump the version of the package.

I was worried about TypeScript getting confused about the files, but there's a setting to tell it to ignore .js files.  I'll try moving them into the packages see what happens.

kevinbarabash:
It seems my worries were unfounded. 😅

packages/wonder-blocks-core/src/components/text.tsx
```diff
@@ -1,3 +1,5 @@
+// WARNING: If you modify this file you must update the following file:
+// build-settings/overrides/wonder-blocks-core/components/text.js.flow
```
jeresig, line 2:
```suggestion
// src/components/text.js.flow
```
Need to update this!

packages/wonder-blocks-core/src/components/view.tsx
```diff
@@ -1,3 +1,5 @@
+// WARNING: If you modify this file you must update the following file:
+// build-settings/overrides/wonder-blocks-core/components/view.js.flow
```
jeresig, line 2:
```suggestion
// src/components/view.js.flow
```
Need to update this!

jeresig (approved):

PR #1801, opened 2023-03-15
Update gen-flow-types.ts to better handle JSX.LibraryManagedAttributes

kevinbarabash:
## Summary:
Some uses of JSX.LibraryManagedAttributes<> use React.ComponentProps and other use React.ElementProps.  This PR updates gen-flow-types.ts to handle both cases.

Issue: None

## Test plan:
- change the glob in gen-flow-types.ts to wonder-blocks-typography
- yarn build:types
- yarn build:flowtypes
- see the following in the console log

  replacing 'JSX.LibraryManagedAttributes<
    typeof Text,
    React.ElementProps<typeof Text>
  >' with 'React.ElementConfig<typeof Text>'

### DIFF
--- a/.changeset/metal-games-compete.md
+++ b/.changeset/metal-games-compete.md
@@ -1,0 +1,5 @@
+---
+"wb-dev-build-settings": patch
+---
+
+Convert JSX.LibraryManagedAttributes<> containing a React.ElementProps<> to React.ElementConfig<>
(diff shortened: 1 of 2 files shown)


somewhatabstract (approved):
LGTM

Thanks for tweaking this. Hopefully we can get a change into flowgen that handles this thing so it will work for webapp too.

PR #1802, opened 2023-03-15
[fei5045.1.tooltweaks] Some little tweaks to our tooling

somewhatabstract:
## Summary:
- Tweaked our flow type generation command to complain if there are no typescript type definitions built first
- Replaced `glob` with `fast-glob` in the root package.json dev deps and migrated any uses of `glob` to `fast-glob` (we were using `fast-glob` already but it was working because of a third-party inclusion rather than explicit inclusion by our `package.json`)

Issue: FEI-5045

## Test plan:
`yarn clean && yarn build:flowtypes` and see the new error
`yarn build:types && yarn build:flowtypes` and see that it works

`node ./utils/publish/pre-publish-check-ci.js` and see that it works


kevinbarabash (approved):
LGTM.  Thanks for adding error handling to gen-flow-types and updating the pre-publish script. 💚

PR #1803, opened 2023-03-15
Update <Data>'s 'children' prop to allow a ReactNode to be returned

kevinbarabash:
## Summary:
Callbacks should always allow React.ReactNodes to be returned.  I missed updating this one in wonder-blocks-data.  I'm updating wonder-blocks-data at the same time as wonder-blocks-core.  I tested this change in webapp by editing node_modules directly.

Issue: None

## Test plan:
- yarn build:types
- yarn build:flowtypes
- see that the 'children' prop in data.js.flow now allows a 'React.Node' to be returned.

### DIFF
--- a/packages/wonder-blocks-data/src/components/__tests__/data.test.tsx
+++ b/packages/wonder-blocks-data/src/components/__tests__/data.test.tsx
@@ -54,7 +54,6 @@ describe("Data", () => {
                 // Act
                 render(
                     <Data handler={fakeHandler} requestId="ID">
-                        {/* @ts-expect-error [FEI-5019] - TS2322 - Type 'Mock<null, [], any>' is not assignable to type '((result: Result<TData>) => ReactElement<any, string | JSXElementConstructor<any>>) & ReactNode'. */}
                         {fakeChildrenFn}
                     </Data>,
                 );
@@ -74,7 +73,6 @@ describe("Data", () => {
                 // Act
                 render(
                     <Data handler={fakeHandler} requestId="ID">
-                        {/* @ts-expect-error [FEI-5019] - TS2322 - Type 'Mock<null, [], any>' is not assignable to type '((result: Result<TData>) => ReactElement<any, string | JSXElementConstructor<any>>) & ReactNode'. */}
                         {fakeChildrenFn}
                     </Data>,
                 );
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-data/src/components/__tests__/data.test.tsx
```diff
@@ -54,7 +54,6 @@ describe("Data", () => {
                 // Act
                 render(
                     <Data handler={fakeHandler} requestId="ID">
-                        {/* @ts-expect-error [FEI-5019] - TS2322 - Type 'Mock<null, [], any>' is not assignable to type '((result: Result<TData>) => ReactElement<any, string | JSXElementConstructor<any>>) & ReactNode'. */}
```
somewhatabstract, line 57: **question[no-change-required]:** Is there a way to suppress this for a whole file?

somewhatabstract: **suggestion:** Could just set the `fakeChildrenFn` value to `: any` to get around this. That seems like the easy fix.

packages/wonder-blocks-data/src/components/data.ts
```diff
@@ -57,18 +57,18 @@ type Props<
  * support server-side rendering and efficient caching.
  */
 // TODO(FEI-5000): Update this support generic props correctly
-const Data: React.FC<Props<any>> = <TData extends ValidCacheData>({
+const Data: React.FC<Props<any>> = (<TData extends ValidCacheData>({
```
somewhatabstract, line 60: **question:** It's a shame we lose the generic support here. Is that going to lead to incorrect usage?

somewhatabstract (approved):
LGTM

Some things inline but nothing blocking.

packages/wonder-blocks-data/src/components/data.ts
kevinbarabash: I think we can fix this, I just don't know how at the moment. :(

packages/wonder-blocks-data/src/components/__tests__/data.test.tsx
kevinbarabash: This change removes a bunch of suppressions.  There are still a bunch of suppressions like:
```
// @ts-expect-error [FEI-5019] - TS2769 - No overload matches this call.
                await act(() => response1);
```
We could look at tweaking the types for `act` in the future.

somewhatabstract: It's likely we just need to tweak the variables, like `response1` to be a compatible type like `any`.

PR #1805, opened 2023-03-15
Speed up flowgen

kevinbarabash:
## Summary:
gen-flow-types.ts was using the CLI version of flowgen but this is expensive b/c it incurs overhead from node starting up and reparsing the flowgen's code each time it runs.  This PR uses the flowgen's API directly.  It does require us to do a little extra work to include a Flow header in each file.

This change reduced the run-time on CI from 10m14s to 3m49s.

Issue: None

## Test plan:
- compare run-times on CI between the new and old versions of the script


build-settings/gen-flow-types.ts
```diff
@@ -27,22 +37,15 @@ for (const inFile of files) {
...
-        outFile,
-        "--add-flow-header",
-        "--no-inexact",
-    ];
-
     try {
-        execFileSync("yarn", args, {cwd: rootDir});
-        console.log(`✅ wrote: ${outFile}`);
+        const source = fs.readFileSync(path.join(rootDir, inFile), "utf-8");
+        let contents: string = compiler.compileDefinitionString(source, {
```
jeremywiebe, line 42: Not sure if it's worth it, but there's also a `compiler.comileDefinitionFile(path, options)` ([source](https://github.com/joarwilk/flowgen/blob/master/src/cli/compiler.ts#L124)).  It would remove one more line here but probably not affect performance.

build-settings/gen-flow-types.ts
```diff
@@ -27,22 +37,15 @@ for (const inFile of files) {
...
-        "flowgen",
-        inFile,
-        "-o",
-        outFile,
-        "--add-flow-header",
-        "--no-inexact",
-    ];
-
     try {
-        execFileSync("yarn", args, {cwd: rootDir});
```
jeremywiebe, line 40: Yeah, not having to fork out to a new process through `yarn`. That makes sense that it's much faster!! Nice find.

jeremywiebe (approved):
Makes sense to me!

build-settings/gen-flow-types.ts
kevinbarabash: I tried that and it looked like it wasn't working, so I switch to `compileDefinitionString`, but it was actual a different error in the try-catch block.  After I fixed the real error, I decided not to switch back.

PR #1808, opened 2023-03-16
Fix build config

kevinbarabash:
## Summary:
When updating webapp to use wonder-blocks-data I ran into a very strange error. It was something along the lines of:

```
TypeError: Can't call n without new
```

I traced the issue back the to how we were building our packages.  'n' is actually 'KindError' from wonder-stuff-core.  We should probably be include dev bundles in our wonder-stuff-core to make debugging easier.

The issue with the wonder-blocks packages is that we were transpiling ES6 classes.  This was necessary due to a bunch of errors I Was seeing when we first switched over from @babel/preset-flow to @babel/preset-typescript.  In turns out that if we jettison @babel/plugin-transform-spread we can drop the class transforms as well.

~~In order to deal with the issue that @babel/plugin-transform-spread was addressing, I've remove the 'targets' option from @babel/env (this causes the preset to use our .browserslistrc file) and added core-js settings so the appropriate polyfills are added.  I've modified the rollup config to external all core-js modules.  We'll need to make sure that webapp is including those.  I've also updated the .browserslistrc file to reflect our supported browsers as documented in Confluence.~~

I've updated our babel config to match what's in webapp (minus the use of `@babel/plugin-transform-spread`).  That plugin will be remove from webapp's config when we migrate the static service to TypeScript.

Issue: None

## Test plan:
- `yarn build`
- look at packages/wonder-blocks-data/dist/index.js and the esm build
- see that classes aren't being compiled
- see that there are some core-js require (import) statements at the top of the file


jeresig (approved):
Nice! Just to confirm, core-js and the env settings now all match what's in webapp, correct?

kevinbarabash:
> Nice! Just to confirm, core-js and the env settings now all match what's in webapp, correct?

No.  But I will be updating webapp in https://github.com/Khan/webapp/pull/12542 to work with these change.  In webapp we also ignore .browserslistrc and aren't using core-js afaict.  `@babel/plugin-transform-runtime` is using `corejs: false,` which doesn't seem right.  We can either set it to `true` or we can use the same `@babel/preset-env` settings from this PR.  We already have a cache group set up in webpack so they should be equivalent.  My preference would be use the same `@babel/preset-env` settings.

.browserslistrc
```diff
@@ -1,8 +1,5 @@
-node 10
-last 1 Chrome version
-last 1 ChromeAndroid version
-last 1 Edge version
-last 1 Firefox version
-last 1 iOS version
-last 1 Safari version
-IE 10
+chrome >= 87
```
jeresig, line 1: We can delete this file now that we're using esmodules: true.

package.json
```diff
@@ -128,4 +128,4 @@
     "@types/react": "16",
     "@types/react-dom": "16"
   }
-}
```
jeresig, line 131: I think you can revert this!

jeresig (approved):
Great - thanks for working on this. Exciting to see such a file size decrease!

kevinbarabash:
> Great - thanks for working on this. Exciting to see such a file size decrease!

I don't think this is actually a decrease from pre-TS transition.  This likely just gets us back to where we were before in terms of bundle size.

PR #1811, opened 2023-03-16
Add startIcon and endIcon props to Link

n1colemejia:
## Summary:
- Adds optional `startIcon` and `endIcon` props to `Link`
- `endIcon` overrides the default external icon when `target="_blank"`, so if `endIcon` is passed in, it will show `endIcon` after the link label instead of `externalIcon`.
- Adds story for links with start and end icons 
- [Figma specs for `startIcon` and `endIcon`](https://www.figma.com/file/VbVu3h2BpBhH80niq101MHHE/%F0%9F%92%A0-Main-Components?node-id=11512-7046&t=uQJAilK1acQeFt4B-0)

    Screenshot 2023-04-06 at 12 36 05 PM

Issue: WB-1085

## Test plan:
- Yarn test and yarn typecheck
- Set start and end icons in control panel and inspected change in the DOM
- Set `endIcon` and `target` to `"_blank"`, confirmed that `endIcon` renders instead of `externalIcon`, and inspected in the DOM 
  
   - Screenshot 2023-03-16 at 1 59 59 PM
   - Screenshot 2023-03-16 at 2 00 10 PM
- Added unit tests to check if:
  - icon props are passed down correctly
  - icons are rendered with link when icon props are passed in
  - icons are **not** rendered when icon props are **not** passed in 
  - `endIcon` is rendered instead of `externalIcon` when `target="_blank"`


packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -68,9 +70,12 @@ export default class LinkCore extends React.Component<Props> {
             "data-test-id": testId,
             style: [defaultStyles, style],
             target,
+            startIcon,
+            endIcon,
```
nishasy, line 74: I don't think you need to add `startIcon` and `endIcon` to `commonProps` since they've already been used to render the props here and no longer need to continue being passed into the `StyledLink` and `StyledAnchor` components.

packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -114,7 +142,11 @@ export default class LinkCore extends React.Component<Props> {
 const styles: Record<string, any> = {};
 
 const iconStyles = StyleSheet.create({
-    icon: {
+    startIcon: {
+        marginRight: Spacing.xxxSmall_4,
+    },
```
nishasy, line 147: I think the icon is sitting a bit low within the line height, so it doesn't align with the text.

image

n1colemejia: This worked! I wrapped the link contents in `View` and styled with `display: inline-flex`, `flexDirection: row`, and `alignItems: center`. The flex direction seemed to do the trick. It was set to `column` in my dev tools.

n1colemejia: Screenshot 2023-03-21 at 1 06 16 PM

n1colemejia:
The parent pull-request ([#1794](https://github.com/Khan/wonder-blocks/pull/1794)) has been merged into `main`, but this branch (`link-start-end-icons`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

packages/wonder-blocks-link/src/components/link-core.tsx
n1colemejia: SO I ended up changing this because I noticed when the viewport shrank, the icons got all wonky and didn't stay on the same row as the label:

Screenshot 2023-03-27 at 3 51 49 PM

I tried a bunch of different things still wrapping `linkContent` in a `View` but ultimately got rid of `View` and got the closest by just adding `Spacing.xxxxSmall_2` to the top & bottom margin.

Here's what it looks like now:

Screenshot 2023-03-27 at 4 13 50 PM

somewhatabstract:
I left some comments in the Chromatic review :) This is looking great now! Nice work on fixing the wrapping issue.

packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -79,15 +82,38 @@ export default class LinkCore extends React.Component<Props> {
...
+            />
+        ) : (
+            target === "_blank" && externalIcon
+        );
+
         const linkContent = (
             <>
-                {children}
-                {target === "_blank" && externalIcon}
+                {startIconChild}
```
nishasy, line 114: I think you can remove the ternary operator for `startIconChild` definition and only set it to just the `<Icon>`, and then down here, use `{startIcon && startIconChild}`.

__docs__/wonder-blocks-link/link.stories.tsx
```diff
@@ -197,9 +198,100 @@ export const OpensInANewTab: StoryComponentType = () => (
...
+                style={styles.standaloneLinkWrapper}
+            >
+                This link has a start icon and an end icon
+            </Link>
+            <Link
+                href="#"
+                endIcon={icons.info}
+                target="_blank"
+                style={styles.standaloneLinkWrapper}
+            >
```
nishasy, line 239: Do you you think you could add an example that's explicitly multiline with the start and end icons? One approach is to add a max width to the link so that it wraps naturally for the length of the link. You could also have a whole paragraph as a link if you'd prefer.

It's likely that a lot of links won't be designed to wrap with icons anyway, but links can get pretty long when translated to other languages.

nishasy (commented):
@n1colemejia I think we might have to do another style pass to match the [spec](https://www.figma.com/file/VbVu3h2BpBhH80niq101MHHE?node-id=11512:7145#405203487).

Right now, the start and end icons are inline with the text, but I believe they should be outside the text boundary.

Current implemenation:
Screen Shot 2023-03-30 at 3 20 30 PM

Design spec:
Screen Shot 2023-03-30 at 3 21 44 PM

__docs__/wonder-blocks-link/link.argtypes.ts
```diff
@@ -9,6 +10,17 @@ export default {
         type: {name: "string", required: true},
     },
 
+    endIcon: {
+        control: {type: "select"},
+        description: `Icon to appear after the link label. If
+        \`target="_blank"\` and an \`endIcon\` is passed in, \`endIcon\` will
+        override the default \`externalIcon\`.`,
+        options: icons,
```
nishasy, line 18: Can you add a `none: null` option to the `startIcon` and `endIcon` options? Currently, there's no way to get rid of the icon once you select one. I believe you can do this by using:

```
options: {none: null, ...icons},
```

__docs__/wonder-blocks-link/link.argtypes.ts
```diff
@@ -98,6 +110,15 @@ export default {
...
 
+    startIcon: {
+        control: {type: "select"},
+        description: "Icon to appear before the link label.",
+        options: icons,
+        table: {
+            type: {summary: "IconAsset"},
+        },
+    },
+
```
nishasy, line 121: If you want `startIcon` and `endIcon` to be next to each other in the table, you could add a `category` field to `table`. Maybe you could call the category "Icons" or something?

packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -79,15 +82,38 @@ export default class LinkCore extends React.Component<Props> {
...
```
nishasy, line 114: I think the convention would be to remove the `startIconChild &&` from the const definition (for both start icon and end icon), and add it down here.

```
const startIconChild = <Icon />

const linkContent = (
    <>
         {startIcon && startIconChild}
         ...
     </>
 )
```

nishasy (approved):
Nice!! Great stories, tests, and comments. 

I left a couple of small comments, but I don't want to block you from landing once you're done with those.

packages/wonder-blocks-link/src/components/link-core.tsx
n1colemejia: When I add `as IconAsset` the error goes away, but I'm not sure if this goes against the point of using Typescript.

Screenshot 2023-04-11 at 1 40 41 PM

n1colemejia: ^ I'm going to go with the last implementation because I see it done for other components and it passes the typecheck! 😊

PR #1813, opened 2023-03-16
Fix issue with some React.Element<> Flow types and more

kevinbarabash:
## Summary:
Many of our components use 'children: React.Element<typeof Component>' (Flow) to restrict which components can be used as children.  In TypeScript the pattern is 'React.ReactElement<React.ComponentProps<typeof Component>>' which flowgen was converting to 'React.ReactElement<React.ComponentProps<typeof Component>>' which is wrong for Flow.

This PR updates gen-flow-types.ts to fix this using a regex to post process the incorrect type.  We'll want to port this to flowgen for use with converting perseus and webapp.

Issue: FEI-5042

## Test plan:
- change the glob in gen-flow-types.ts to wonder-blocks-breadcrumbs
- yarn build:types
- yarn build:flowtypes
- check that the generate .js.flow files for wonder-blocks-breadcrumbs are correct

### DIFF
--- a/packages/wonder-blocks-i18n/src/components/i18n-inline-markup.tsx
+++ b/packages/wonder-blocks-i18n/src/components/i18n-inline-markup.tsx
@@ -109,8 +109,8 @@ type Props = {
         elem: React.ReactNode,
         type: string,
         i: number,
-    ) => React.ReactElement;
-    onError?: (e: Error) => React.ReactElement;
+    ) => React.ReactNode;
+    onError?: (e: Error) => React.ReactNode;
 };
 
 export class I18nInlineMarkup extends React.PureComponent<Props> {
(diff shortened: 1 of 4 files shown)


packages/wonder-blocks-i18n/src/components/i18n-inline-markup.tsx
```diff
@@ -109,16 +109,16 @@ type Props = {
...
+    onError?: (e: Error) => React.ReactNode;
 };
 
 export class I18nInlineMarkup extends React.PureComponent<Props> {
     /**
      * If an error occurs, we either call the onError prop, or throw the
      * error.
      */
-    handleError(error: Error): React.ReactElement {
+    handleError(error: Error): React.ReactNode {
```
kevinbarabash, line 121: I missed this in other PRs where I update the return types on callback props.

jeresig (approved):

PR #1817, opened 2023-03-20
TextField: Update input to StyledInput

n1colemejia:
## Summary:
css() doesn't support nested arrays. Updated input to be a StyledInput and used style prop instead.

Issue: WB-1530

## Test plan:
Open TextField in Storybook and confirm disabled and error styles look right.

### DIFF
--- a/packages/wonder-blocks-form/src/components/text-field.tsx
+++ b/packages/wonder-blocks-form/src/components/text-field.tsx
@@ -1,9 +1,11 @@
 import * as React from "react";
-import {StyleSheet, css} from "aphrodite";
+import {StyleSheet} from "aphrodite";
 
 import Color, {mix, fade} from "@khanacademy/wonder-blocks-color";
+import {addStyle} from "@khanacademy/wonder-blocks-core";
 import Spacing from "@khanacademy/wonder-blocks-spacing";
 import {styles as typographyStyles} from "@khanacademy/wonder-blocks-typography";
+
 import type {StyleType, AriaProps} from "@khanacademy/wonder-blocks-core";
 
 export type TextFieldType = "text" | "password" | "email" | "number" | "tel";
@@ -13,6 +15,8 @@ type WithForwardRef = {
 };
 
 const defaultErrorMessage = "This field is required.";
+
+const StyledInput = addStyle("input");
 
 type Props = AriaProps & {
     /**
@@ -233,12 +237,10 @@ class TextField extends React.Component<PropsWithForwardRef, State> {
             // Should only include Aria related props
             ...otherProps
         } = this.props;
+
         return (
-            <input
-                // @ts-expect-error: we shouldn't be passing `style` to `css()`
-                // here b/c `style` allows nested arrays of styles, but `css()`
-                // only allows a flat array.
-                className={css([
+            <StyledInput
+                style={[
                     styles.input,
                     typographyStyles.LabelMedium,
                     styles.default,
@@ -247,12 +249,15 @@ class TextField extends React.Component<PropsWithForwardRef, State> {
                         ? styles.disabled
                         : this.state.focused
                         ? [styles.focused, light && styles.defaultLight]
-                        : this.state.error && [
+                        : !!this.state.error && [
                               styles.error,
                               light && styles.errorLight,
                           ],
                     style && style,
-                ])}
+                    // Cast `this.state.error` into boolean since it's being
+                    // used as a conditional
+                    !!this.state.error && styles.error,
+                ]}
                 id={id}
                 type={type}
                 placeholder={placeholder}


kevinbarabash (commented):
Thanks for fixing this issue.  Were you able to reproduce the issue in Storybook before applying the fix?  I wonder why our jest tests didn't catch this, maybe `StyleSheetTestUtils.suppressStyleInjection()` in test-setup.js prevents the code in aphrodite that was triggering the error from running.

packages/wonder-blocks-form/src/components/text-field.tsx
```diff
@@ -247,12 +249,15 @@ class TextField extends React.Component<PropsWithForwardRef, State> {
...
-                        : this.state.error && [
+                        : !!this.state.error && [
                               styles.error,
                               light && styles.errorLight,
                           ],
                     style && style,
-                ])}
+                    // Cast `this.state.error` into boolean since it's being
+                    // used as a conditional
+                    !!this.state.error && styles.error,
```
kevinbarabash, line 259: I don't see this in the originally logic.  Was there an issue with the render of the error state that necessitated this change?

packages/wonder-blocks-form/src/components/text-field.tsx
```diff
@@ -247,12 +249,15 @@ class TextField extends React.Component<PropsWithForwardRef, State> {
                         ? styles.disabled
                         : this.state.focused
                         ? [styles.focused, light && styles.defaultLight]
-                        : this.state.error && [
+                        : !!this.state.error && [
                               styles.error,
                               light && styles.errorLight,
                           ],
                     style && style,
```
kevinbarabash, line 256: **suggestion:** If a component has a `style` prop, it's usually placed last so that it can override the default styles.  Can you move this to be last?

kevinbarabash (commented):
It looks like some of the snapshot tests are failing.  These can be updated by running `yarn jest -u`.

n1colemejia:
We didn't reproduce it in Storybook before fixing it 😕 Maybe you're right about `StyleSheetTestUtils.suppressStyleInjection()` though since the type error was inside the styles? I'm not sure 🤷🏻‍♀️

kevinbarabash:
> @kevinbarabash We didn't reproduce it in Storybook before fixing it 😕 Maybe you're right about `StyleSheetTestUtils.suppressStyleInjection()` though since the type error was inside the styles? I'm not sure 🤷🏻‍♀️

Interesting.  Storybook doesn't use `suppressStyleInjection`.  I wonder why it wasn't reproducible in Storybook. 🤔

packages/wonder-blocks-form/src/components/text-field.tsx
kevinbarabash: Sorry, I didn't explain myself well.  The item `style && style,` within the array we pass to `style` should come last.  The order of the props passed to the `StyledInput` component don't matter because `<StyledInput id={id} type={type} style={[...]} />` gets turned into:
```
React.createElement(StyledInput, {
   id: id,
   type: type,
   style: [ ... ],
});
```
The only time order of props matters is when prop spreading is involved, e.g.
```
const props = {message: "hello", name: "world"};
<MyComponent {...props} message="bonjour" />; // "bonjour" will override props.message
<MyComponent message="bonjour" {...props} />; // props.message will override "bonjour"
```

packages/wonder-blocks-core/src/util/add-style.js.flow
```diff
@@ -0,0 +1,13 @@
...
+declare export default function addStyle<
+    T: React.AbstractComponent<any> | string,
+>(
+    Component: T,
+    defaultStyle?: StyleType,
+): React.AbstractComponent<{
+    ...React.ElementConfig<T>,
+    style?: StyleType,
+    ...
+}>;
```
kevinbarabash, line 13: I added this file because `flowgen` wasn't able to generate Flow type definitions from the new code that work in webapp.  These types were copied from the original add-style.js file before it was converted to TypeScript.

packages/wonder-blocks-core/src/util/add-style.tsx
```diff
@@ -22,18 +40,19 @@ export default function addStyle<
             processStyleList([reset, defaultStyle, style]);
 
         return (
-            // @ts-expect-error [FEI-5019] - TS2604 - JSX element type 'Component' does not have any construct or call signatures.
+            // @ts-expect-error: TS says this is not assignable to the return forwardRef()'s return type.
+            // Type 'Omit<PropsWithChildren<Props>, "style" | "className"> & { ref: ForwardedRef<ElementMap[T]>; className: string; style: CSSProperties; }' is not assignable to type 'IntrinsicAttributes & LibraryManagedAttributes<T, SVGProps<SVGSymbolElement> & ClassAttributes<HTMLObjectElement> & ... 179 more ... & SVGProps<...>>'.
+            //   Type 'Omit<PropsWithChildren<Props>, "style" | "className"> & { ref: ForwardedRef<ElementMap[T]>; className: string; style: CSSProperties; }' is not assignable to type 'LibraryManagedAttributes<T, SVGProps<SVGSymbolElement> & ClassAttributes<HTMLObjectElement> & ObjectHTMLAttributes<HTMLObjectElement> & ... 178 more ... & SVGProps<...>>'
```
kevinbarabash, line 45: I'm not sure how helpful having this error message here is, but I'm going to leave it just case people are curious why the types aren't compatible.

packages/wonder-blocks-core/src/util/add-style.tsx
```diff
@@ -6,14 +6,32 @@ import {processStyleList} from "./util";
...
+    } & Omit<React.ComponentProps<T>, "style">, // Removes the 'style' prop from the original component
+>(
+    Component: T,
+    defaultStyle?: StyleType,
+): React.ForwardRefExoticComponent<
+    React.PropsWithoutRef<Props> &
+        React.RefAttributes<
+            // We need to lookup the HTML/SVG element type based on the tag name, but only
+            // for JSX intrinsics (aka HTML/SVG tags).
+            T extends keyof JSX.IntrinsicElements ? IntrinsicElementsMap[T] : T
```
kevinbarabash, line 25: I think there might be a way to lookup what we want from `JSX.IntrinsicElements` instead of having to duplicate a list of instrinsics below in `IntrinsicElementsMap`, but I haven't found it yet and this works despite it being a little hacky.

packages/wonder-blocks-form/src/components/text-field.tsx
```diff
@@ -233,12 +237,10 @@ class TextField extends React.Component<PropsWithForwardRef, State> {
...
         } = this.props;
+
         return (
-            <input
-                // @ts-expect-error: we shouldn't be passing `style` to `css()`
-                // here b/c `style` allows nested arrays of styles, but `css()`
-                // only allows a flat array.
-                className={css([
+            <StyledInput
+                style={[
```
kevinbarabash, line 243: @n1colemejia I moved the `style` prop up to where the `className` prop was.  This makes it easier to see what was changed in the array.

kevinbarabash (approved):
LGTM. Thanks for working through all of the issues with this PR.

jandrade (approved):
Looks great! 🚀

PR #1749, opened 2023-02-22
FEI-4966.4: Run codemod to convert .js(x) files to .ts(x)

kevinbarabash:
## Summary:
This PR was created by running the following commands:
- cd ../flow-to-typescript-codemod
- yarn typescriptify convert -p ../wonder-blocks --write --delete
- yarn typescriptify fix --autoSuppressErrors --jiraSlug FEI-5019 -p ../wonder-blocks --config ../wonder-blocks/tsconfig-check.json
- cd ../wonder-blocks
- yarn prettier --write packages/**/*.ts packages/**/*.tsx

Issue: FEI-4966

## Test plan:
- yarn tsc --package tsconfig-check.json, see that there are a handful of unsuppressed errors, these will be address in a subsequent PR

PR #1820, opened 2023-03-21
[🔥AUDIT🔥] [fixmockgqlfetch] Fix the indentation for outputting GQL operations

somewhatabstract:
🖍 _This is an audit!_ 🖍

## Summary:
The codemod used for the changes in https://github.com/Khan/wonder-blocks/pull/1749 stripped the indentation from this template literal. This fixes that.

Issue: FEI-5042

## Test plan:
The snapshot tests in webapp caught this. If they pass without udpate, we know it's fixed.

### DIFF
--- a/packages/wonder-blocks-testing/src/gql/mock-gql-fetch.ts
+++ b/packages/wonder-blocks-testing/src/gql/mock-gql-fetch.ts
@@ -8,8 +8,11 @@ import type {GqlFetchMockFn, GqlMockOperation} from "./types";
 export const mockGqlFetch = (): GqlFetchMockFn =>
     mockRequester<GqlMockOperation<any, any, any>, any>(
         gqlRequestMatchesMock,
+        // Note that the identation at the start of each line is important.
         (operation, variables, context) =>
             `Operation: ${operation.type} ${operation.id}
-Variables: ${variables == null ? "None" : JSON.stringify(variables, null, 2)}
-Context: ${JSON.stringify(context, null, 2)}`,
+    Variables: ${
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-testing/src/gql/mock-gql-fetch.ts
```diff
@@ -8,8 +8,11 @@ import type {GqlFetchMockFn, GqlMockOperation} from "./types";
...
         gqlRequestMatchesMock,
+        // Note that the identation at the start of each line is important.
         (operation, variables, context) =>
             `Operation: ${operation.type} ${operation.id}
-Variables: ${variables == null ? "None" : JSON.stringify(variables, null, 2)}
-Context: ${JSON.stringify(context, null, 2)}`,
+    Variables: ${
+        variables == null ? "None" : JSON.stringify(variables, null, 2)
+    }
+    Context: ${JSON.stringify(context, null, 2)}`,
```
somewhatabstract, line 17: **nit:** I didn't add a test for this right now with a view to getting this fixed quickly.  It really should have one so we can catch this more easily in future.

somewhatabstract: NVM, there IS a test for this...but the snapshot got updated instead of the code getting fixed....now I have to update the snapshot in another audit. lol

My bad.

somewhatabstract: Fixed in https://github.com/Khan/wonder-blocks/pull/1822

PR #1822, opened 2023-03-21
[🔥AUDIT🔥] [fixthetest] Fix the test case and add some commentary

somewhatabstract:
🖍 _This is an audit!_ 🖍

## Summary:
This modifies the test to make it harder to break the formatting and adds some additional commentary.

Issue: FEI-5042

## Test plan:
`yarn jest`

(`yarn test` is failing due to `alex` and possible problematic language)


packages/wonder-blocks-testing/src/gql/__tests__/mock-gql-fetch.test.tsx
```diff
@@ -172,16 +172,18 @@ describe("#mockGqlFetch", () => {
...
+        await expect(result).rejects.toThrowError(
+            `No matching mock response found for request:
+    Operation: query getMyStuff
+    Variables: {
+  "a": "variable"
+}
+    Context: {
+  "my": "context"
+}`,
+        );
```
somewhatabstract, line 186: **note:** Changed to not use a snapshot so that if the test fails, it isn't just overwritten easily. Also added a comment indicating the indentation is deliberate.

packages/wonder-blocks-testing/src/gql/mock-gql-fetch.ts
```diff
@@ -9,6 +9,8 @@ export const mockGqlFetch = (): GqlFetchMockFn =>
     mockRequester<GqlMockOperation<any, any, any>, any>(
         gqlRequestMatchesMock,
         // Note that the identation at the start of each line is important.
+        // TODO(somewhatabstract): Make a stringify that indents each line of
+        // the output properly too.
```
somewhatabstract, line 13: **note:** Would love to fix the stringified bits long term too for a nicer output. Not in scope for this change though

PR #1825, opened 2023-03-22
Fix moment import

kevinbarabash:
## Summary:
Using a named import confuses vite since moment doesn't have true named exports.  It isn't an ES6 module.

Issue: None

## Test plan:
- apply this change manually in webapp's node_modules folder
- clear the vite cache by deleting node_modules/.vite
- restart vite
- see that the logged out homepage loads at localhost:8224/


packages/wonder-blocks-birthday-picker/src/components/birthday-picker.tsx
```diff
@@ -1,4 +1,4 @@
-import moment, {monthsShort} from "moment";
+import moment from "moment"; // NOTE: 'moment' does not support named imports
```
somewhatabstract, line 1: **question:** I wonder if this comment should be more strongly worded:
```suggestion
import moment from "moment"; // NOTE: DO NOT use named imports; 'moment' does not support named imports
```

somewhatabstract (approved):
LGTM

PR #1827, opened 2023-03-22
[updatereleaseflow] Update release workflow to use shared-node-cache and update actions to non-Node 12 variants

somewhatabstract:
## Summary:
This makes sure we're not using the deprecated node 12 in our actions. It also brings in the best practice already used in Wonder Stuff and other places that uses our shared-node-cache action to speed up the build and reduce the number of dependencies we have to maintain.

Issue: XXX-XXXX

## Test plan:
We will have to wait until the next release cycle to verify fully. However, this works in other places so I have no reason to believe it won't work here.


jeresig (approved):
Ahh - nice!

.github/workflows/release.yml
```diff
@@ -31,28 +31,23 @@ jobs:
     name: Release
     runs-on: ubuntu-latest
     steps:
-      - uses: webfactory/ssh-agent@v0.4.1
+      - uses: webfactory/ssh-agent@v0.7.0
```
somewhatabstract, line 34: **note:** This is the same version as Wonder Stuff uses for its release.yml

.github/workflows/release.yml
```diff
@@ -31,28 +31,23 @@ jobs:
...
         with:
           ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
       - name: Clone repository
         run: git clone git@github.com:Khan/wonder-blocks.git .
 
-      - name: Setup Node.js 16.x
-        uses: actions/setup-node@master
+      - name: Install & cache node_modules
```
somewhatabstract, line 40: **note:** This section was copied Wonder Stuff release.yml

.github/workflows/release.yml
```diff
@@ -31,28 +31,23 @@ jobs:
...
+        uses: Khan/actions@shared-node-cache-v0
         with:
-          # This is our target Node version.
```
somewhatabstract, line 43: **note:** This comment wasn't in Wonder Stuff so it got removed when I pasted in the new step. I can add it back if folks feel strongly about it.

PR #1828, opened 2023-03-22
Adds indeterminate state to checkbox

n1colemejia:
## Summary:
Adds an indeterminate state to checkbox component by setting `checked` to `null`/`undefined`. When `checked` is `null`/`undefined`, checkbox will show an indeterminate icon :heavy_minus_sign:. 

Issue: https://khanacademy.atlassian.net/browse/WB-1460

## Test plan:
- `yarn jest` `yarn typecheck`
- In the [Indeterminate story](https://5e1bf4b385e3fb0020b7073c-obfmyxbzyo.chromatic.com/?path=/story/form-checkbox--indeterminate) for Checkbox, inspect the checkboxes in default, disabled and error state to see the correct indeterminate icon appears and that the `aria-checked` attribute is `mixed`.

  Screenshot 2023-03-28 at 12 39 06 PM


jandrade:
@n1colemejia I think that indeterminate states are not directly connected to the `checked` state. You can see more about that here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes

> In addition to the checked and unchecked states, there is a third state a checkbox can be in: indeterminate. This is a state in which it's impossible to say whether the item is toggled on or off.

I'm not sure if this was discussed before, so in case of that, I'd like to open the conversation and get your thoughts about this.

packages/wonder-blocks-form/src/components/checkbox.tsx
```diff
@@ -6,9 +6,9 @@ import ChoiceInternal from "./choice-internal";
 // Keep synced with ChoiceComponentProps in ../util/types.js
 type ChoiceComponentProps = AriaProps & {
     /**
-     * Whether this component is checked
+     * Whether this component is checked or indeterminate
      */
-    checked: boolean;
+    checked: boolean | "mixed";
```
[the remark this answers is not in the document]
n1colemejia: When `checked: boolean | null | undefined` in `checkbox.tsx`, I get type errors that say `checked` is incompatible: 

Screenshot 2023-03-31 at 10 27 18 AM
Screenshot 2023-03-31 at 10 28 30 AM

I changed the `checked` type to match in `choice.tsx`, `choice-internal.tsx`, and `types.ts` and the error didn't go away.

Also recieved type error in `checkbox-core.tsx` for `StyledInput` `checked` prop that led to `index.d.ts` for `InputHTMLAttributes` where `checked?: boolean | "mixed" | undefined`.

Screenshot 2023-03-31 at 10 29 59 AM

In Storybook:

The indeterminate icon appears when `checked={null}` or `checked="mixed"`, but doesn't appear when `checked={undefined}`.

The `aria-checked` doesn't appear in the DOM when the `checked` value is `null` or `undefined`, only for `checked="mixed"`

I could totally by missing something or going about this wrong! This is just the scavenger hunt that led me to settle on `boolean | "mixed"`, but I agree that `null`/`undefined` feels like a better approach.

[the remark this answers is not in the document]
jandrade: ahh ok, thanks for all the context, it's super helpful :).

So the issue that you experienced was because, I suspect there's one missing step to be able to use `null/undefined` in checkboxes.

The `props.checked` value can still have `true | false | null | undefined`, but we'd need to map that value to `aria-checked` so that ARIA attribute can have the right value and avoid that TS error.

I could imagine something like this would work (not fully sure)...

Just adding a snippet for illustration purposes:

```
type Checked = boolean | null | undefined;
type AriaChecked = "true" | "false" | "mixed";

function mapCheckedToAriaChecked(value: Checked): AriaChecked {
  switch(value) {
    case true:
      return "true";
    case false:
      return "false";
    default:
      return "mixed";
  }
}

// then in the component
const ariaChecked = mapCheckedToAriaChecked(props.checked);
...
return (
  <input type="checkbox" aria-checked={ariaChecked} ... />
);
```

What do you think? happy to chat more about it.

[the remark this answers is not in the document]
n1colemejia: Oh awesome, thank you Juan! I will try this and check back to see about pairing 😊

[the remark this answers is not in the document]
n1colemejia: So that worked! The only issue is that `undefined` sets `aria-checked` to `false` instead of `"mixed"` even though it "looks" right.

megaroeny (approved):
The new story looks great Nicole!! 🎉

packages/wonder-blocks-form/src/components/checkbox-core.tsx
```diff
@@ -6,17 +6,35 @@ import {addStyle} from "@khanacademy/wonder-blocks-core";
 import Icon from "@khanacademy/wonder-blocks-icon";
 
 import type {IconAsset} from "@khanacademy/wonder-blocks-icon";
-
 import type {ChoiceCoreProps} from "../util/types";
 
+type Checked = boolean | null | undefined;
```
jandrade, line 11: **suggestion:** this type can be exported so it can be reused in `checkbox.tsx`. It help us to centralize these types in case there's a future change:

```suggestion
export type Checked = boolean | null | undefined;
```

jandrade: Also, it would be really helpful adding a comment about what `null | undefined` means in this context (or more specifically that this intends to be used when the checkbox is in a indeterminate state.

packages/wonder-blocks-form/src/components/checkbox-core.tsx
```diff
@@ -6,17 +6,35 @@ import {addStyle} from "@khanacademy/wonder-blocks-core";
...
+
+type AriaChecked = "true" | "false" | "mixed";
+
+function mapCheckedToAriaChecked(value: Checked): AriaChecked {
```
jandrade, line 15: **suggestion:** It would be also helpful adding comments here about the why we define this intermediate step... or describing the purpose of this function (e.g. expressing that we convert the prop value to a value that a Screen Reader can understand via the `aria-checked` attribute).

packages/wonder-blocks-form/src/components/checkbox-core.tsx
```diff
@@ -50,13 +68,31 @@ export default class CheckboxCore extends React.Component<ChoiceCoreProps> {
...
+            />
+        );
+
+        const ariaChecked = mapCheckedToAriaChecked(checked);
+
+        // TODO (nicolemejia) - We're not using `.indeterminate` on
+        // `StyledInput` because `addStyle()` doesn't support ref forwarding. We
+        // should consider adding it in the future to round out the component to
+        // align with input checkbox indeterminate property. Read more here:
+        // https://www.w3schools.com/jsref/prop_checkbox_indeterminate.asp
```
jandrade, line 86: **question:** Do you think we could fix this by merging your changes from #1817 (after that PR is landed)? so we could solve this right away :).

packages/wonder-blocks-form/src/components/checkbox-core.tsx
```diff
@@ -135,7 +164,7 @@ const colors = {
 
 const styles: Record<string, any> = {};
 
-const _generateStyles = (checked: boolean, error: boolean) => {
+const _generateStyles = (checked: Checked, error: boolean) => {
```
jandrade, line 167: **praise:** This is a great use of reusability here!

packages/wonder-blocks-form/src/components/checkbox.tsx
```diff
@@ -6,9 +6,9 @@ import ChoiceInternal from "./choice-internal";
 // Keep synced with ChoiceComponentProps in ../util/types.js
 type ChoiceComponentProps = AriaProps & {
     /**
-     * Whether this component is checked
+     * Whether this component is checked or indeterminate
      */
-    checked: boolean;
+    checked: boolean | null | undefined;
```
jandrade, line 11: **suggestion:** we could use the `Checked` type exported from `checkbox-core.tsx` here.

packages/wonder-blocks-form/src/components/radio-core.tsx
```diff
@@ -28,6 +28,8 @@ const StyledInput = addStyle("input");
             testId,
             ...sharedProps
         } = this.props;
+
+        // @ts-expect-error [WB-1460] - TS2345 - Argument of type 'string | boolean' is not assignable to parameter of type 'boolean'. Type 'string' is not assignable to type 'boolean'.
```
jandrade, line 32: **question:** does this happen b/c the `checked` type differs from the one in `ChoiceInternal/Checkbox`?

packages/wonder-blocks-form/src/components/checkbox-core.tsx
```diff
@@ -50,13 +68,31 @@ export default class CheckboxCore extends React.Component<ChoiceCoreProps> {
...
+        // `StyledInput` because `addStyle()` doesn't support ref forwarding. We
+        // should consider adding it in the future to round out the component to
+        // align with input checkbox indeterminate property. Read more here:
+        // https://www.w3schools.com/jsref/prop_checkbox_indeterminate.asp
         return (
             <React.Fragment>
                 <StyledInput
                     {...sharedProps}
                     type="checkbox"
+                    role="checkbox"
```
jandrade, line 92: **suggestion:** you can get rid of this `role` as it is implicitly defined by the `input[checkbox]` element.

```suggestion         
```

packages/wonder-blocks-form/src/__tests__/__snapshots__/custom-snapshot.test.tsx.snap
```diff
@@ -2,12 +2,14 @@
 
 exports[`CheckboxCore type:default checked:false 1`] = `
 <input
+  aria-checked="false"
```
jandrade, line 5: **suggestion:** that makes me realize that we should add a new variant to cover the `indeterminate` case.

This test can be included here: https://github.com/Khan/wonder-blocks/blob/c2ec490259d77ec44dcd6b868d050314c93d72e9/packages/wonder-blocks-form/src/__tests__/custom-snapshot.test.tsx#L9

packages/wonder-blocks-form/src/util/types.ts
```diff
@@ -9,7 +9,7 @@ import Choice from "../components/choice";
 // Shared props for radio-core and checkbox-core
 export type ChoiceCoreProps = AriaProps & {
     /** Whether this component is checked */
-    checked: boolean;
+    checked: boolean | null | undefined;
```
jandrade, line 12: **thought:** Giving a second thought about from where we should export `Checked`... maybe this is the best place from where we should define the `Checked` type as it is the shared/central place for both checkboxes and radiobuttons. wdyt?

jandrade (commented):
Thanks for working on this, great progress!! Things look really well, but I'd like to get your thoughts in a couple of points before giving final approval.

packages/wonder-blocks-form/src/components/checkbox-core.tsx
n1colemejia: Yes I like that idea! I landed #1817 but I can't merge the changes until Version Packages are merged, right?

packages/wonder-blocks-form/src/components/radio-core.tsx
n1colemejia: The error goes away when I update the type for `checked` to `null | undefined` where `_generateStyles` is defined later in the file. 

Not sure if this makes sense though since `Radio` doesn't have an indeterminate state. Should the `checked` type for `Checkbox`/`Radio` be the same?

packages/wonder-blocks-form/src/components/checkbox-core.tsx
jandrade: good question! Yes, you can merge changes right now :)

You can see it here: https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-core/src/util/add-style.tsx#L48

-----------

For more context, you could say that releasing features are a two-step process: 

1. The "code implementation" changes for a particular PR are merged into `main` right after a PR is landed.
2. The version updates for a particular package are merged into `main` right after `Version Packages` are merged.

This means that, in (1) all the code associated to your PR is directly merged into the main branch as soon as these are landed, and this code is available inside the `wonder-blocks` repo, but it's not available for consumers (like the `webapp` repo) yet. That's where (2) comes into the process, because that's where we make these changes available to consumers by releasing the package changes to npm.

packages/wonder-blocks-form/src/components/radio-core.tsx
jandrade: Ahh got it. I think it's should be fine to change it to `Checked` as `RadioCore` is already using the props defined in `ChoiceInternal`.

packages/wonder-blocks-form/src/components/checkbox-core.tsx
n1colemejia: Oh I see! Thank you so much for the context. That makes more sense to me now.

packages/wonder-blocks-form/src/components/checkbox-core.tsx
```diff
@@ -6,21 +6,57 @@ import {addStyle} from "@khanacademy/wonder-blocks-core";
 import Icon from "@khanacademy/wonder-blocks-icon";
 
 import type {IconAsset} from "@khanacademy/wonder-blocks-icon";
-
 import type {ChoiceCoreProps} from "../util/types";
 
+// Checkbox is in indeterminate state when `checked` is `null` | `undefined`
+export type Checked = boolean | null | undefined;
```
jandrade, line 12: **suggestion:** Move this to `/util/types.ts` so you can reference this type everywhere.

packages/wonder-blocks-form/src/components/choice-internal.tsx
```diff
@@ -12,7 +12,7 @@ import RadioCore from "./radio-core";
 
 type Props = AriaProps & {
     /** Whether this choice is checked. */
-    checked: boolean;
+    checked: boolean | null | undefined;
```
jandrade, line 15: **suggestion:** By moving `Checked` to `types.ts`, you can do `import type {Checked} from "./util/type"`, then:

```suggestion
    checked: Checked;
```

packages/wonder-blocks-form/src/components/radio-core.tsx
```diff
@@ -112,7 +113,10 @@ const colors = {
     },
 } as const;
 const styles: Record<string, any> = {};
-const _generateStyles = (checked: boolean, error: boolean) => {
+const _generateStyles = (
+    checked: boolean | null | undefined,
```
jandrade, line 117: Same here:

```suggestion
    checked: Checked,
```

packages/wonder-blocks-form/src/util/types.ts
jandrade, line 12: And here:

```suggestion
    checked: Checked;
```

packages/wonder-blocks-form/src/__tests__/custom-snapshot.test.tsx
```diff
@@ -5,7 +5,7 @@ import CheckboxCore from "../components/checkbox-core";
 import RadioCore from "../components/radio-core";
 
 const states = ["default", "error", "disabled"];
-const checkedStates = [false, true];
+const checkedStates = [false, true, null || undefined];
```
jandrade, line 8: **nit:** This will evaluate to `undefined`, so you can probably use `null` here to be more explicit the indeterminate case:

```suggestion
const checkedStates = [false, true, null];
```

P.S. Don't forget to update the snapshots :).

packages/wonder-blocks-form/src/util/types.ts
n1colemejia: I agree! That makes more sense to me as well. Especially since we're using the same type for radio

jandrade (approved):
Looks really great! thanks for applying the suggestions 🚀.

PR #1830, opened 2023-03-24
Add `:focus-visible` pseudoselector to link

nishasy:
## Summary:
Currently, if you click on a clickable element in Chrome or Firefox,
it sets that element as the the active element (`document.activeElement`).

If that active element happens to have a focus or hover styles,
both of those styles will appear when revisiting this tab, as Chrome and Firefox
automatically force focus on the last active element.

To remove the focus out line for mouse users, I added the `:focus-visible`
pseudoselector to the styles in `link-core.js`. This means that the focus
outline should still appear for keyboard users.

Note that the hover still shows up when revisiting the tab. This behavior
is browser-specfic and not really up to us. This is likely an a11y feature
so that it is immediately obvious where the focus on the page is, if it's
not at the top.

Issue: https://khanacademy.atlassian.net/browse/WB-1536

## Test plan:
todo


packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -49,7 +49,7 @@ export default class LinkCore extends React.Component<Props> {
 
         const defaultStyles = [
             sharedStyles.shared,
-            !(hovered || focused || pressed) && restingStyles,
```
nishasy, line 52: tl;dr: we want the resting style to be part of the default style and add new state styles on top.

Context: The browser (chrome, firefox) automatically adds focus to the link if it's the active element when tabbing back to the page. This triggers ClickableBehavior's handleFocus(), which sets focused to false. So then the resting styles go away. This removes the underline for inline links since the default sharedStyles has textDecoration: "none". We can avoid this by just keeping the resting styles there by default and then adding new state styles on top of that.

packages/wonder-blocks-link/src/components/__tests__/link.test.tsx
nishasy, line 1: I moved the outline tests from the .test file to the .stories file.

packages/wonder-blocks-link/src/__tests__/__snapshots__/custom-snapshot.test.tsx.snap
```diff
@@ -114,13 +114,17 @@ exports[`LinkCore kind:primary href:# light:false visitable:false focused 1`] =
...
+      },
       "alignItems": "center",
-      "borderRadius": 3,
       "color": "#1865f2",
       "cursor": "pointer",
       "display": "inline-flex",
-      "outline": "1px solid #1865f2",
-      "textDecoration": "none",
+      "outline": "none",
+      "textDecoration": "underline currentcolor solid",
```
jandrade, line 127: **question:** why does `textDecoration` now changes to this value?

nishasy: @jandrade I believe that comes from the `restingInline` style, now that I removed the conditional before resting styles.

__docs__/wonder-blocks-link/link.stories.tsx
```diff
@@ -434,7 +502,7 @@ Inline.play = async ({canvasElement}) => {
 export const InlineLight: StoryComponentType = () => (
     <Body style={{color: Color.white}}>
         This is an inline{" "}
-        <Link href="#" inline={true} light={true}>
+        <Link href="#" inline={true} light={true} testId="link-inline-light">
```
jandrade, line 505: **nit:** I think this is no longer needed as `getByRole` is now being used (don't need to changed if you feel so).

jandrade (approved):
Looks great, thanks for answering my question 🚢

__docs__/wonder-blocks-link/link.stories.tsx
nishasy: Ah, good catch! Thanks

PR #1831, opened 2023-03-28
Remove comment that says full-width is the default Banner layout

nishasy:
## Summary:
At some point, we decided that banners should not have any
default value for the `layout` prop. Here, I'm removing
a leftover comment from before then.

Issue: None

## Test plan:
N/A


nishasy:
@lillialexis I wasn't able to find that other comment you were talking about. I did a search for the word "default" in the wonder-blocks-banner directory. If you can find it, please let me know!

jeresig (approved):

lillialexis (approved):

PR #1833, opened 2023-03-29
[versionpackages.1.actionsonpush] Use the checkout action for releases

somewhatabstract:
## Summary:
This uses the checkout action with ssh-key and persist-credentials to see if it fixes the release workflow so that updates to Version Packages run our expected actions.

This approach was garnered through reading various responses to https://github.com/changesets/action/issues/70. Specifically:

- https://github.com/changesets/action/issues/70#issuecomment-813015134
- https://github.com/changesets/action/issues/70#issuecomment-1028232705

Issue: XXX-XXXX

## Test plan:
Land this PR and see what happens with the Version Packages PR when it is created, updated, and landed.


somewhatabstract:
🤞🏻

PR #1836, opened 2023-04-03
Don't use React.FC<> for functional components

kevinbarabash:
## Summary:
Stop using React.FC<> for components.  It has the following downsides:
- adds an implicit 'children' prop
- prevents proper typing of functional components with parametric/generic props

Issue: None

## Test plan:
- yarn build:types
- yarn build:flowtypes
- check that the following new flow types work in webapp (the rest of the packages contain trivial changes):
  - wonder-blocks-data
  - wonder-blocks-dropdown
  - wonder-blocks-modal


packages/wonder-blocks-modal/src/components/modal-launcher.tsx
```diff
@@ -26,7 +26,7 @@ type Props = {
      * Note: Don't call `closeModal` while rendering! It should be used to
      * respond to user intearction, like `onClick`.
      */
-    modal: ModalElement | React.FC<{closeModal: () => void}>;
+    modal: ModalElement | ((props: {closeModal: () => void}) => ModalElement);
```
kevinbarabash, line 29: `ModalElement` is defined as `React.ReactElement | null` which is the same as the return type for things typed as `React.FC<>`.

jeresig (approved):

PR #1838, opened 2023-04-10
Remove display value from Link

nishasy:
## Summary:
Currently, Links have `display: inline-flex` in their styles.
This was added as part of the recent Link updates, but it is
unclear as to why.

This caused some course names to be cut off without ellipses.
(See https://khanacademy.slack.com/archives/C04E2SED13J/p1681138678986699)

After doing some experimentation, it seems like the `display`
style doesn't even need to be set for Links by default. So here,
I am removing it.

Issue: none

## Test plan:
- Go to http://localhost:51027/?path=/story/link--variants.
- Resize the window.
- Confirm that all the links text wrap when they overflow.

## Before (inline-flex): 

https://user-images.githubusercontent.com/13231763/231018582-60225e45-2b82-41c9-8a95-a69faadf21f6.mov

## After (no display, default to inline in storybook):

https://user-images.githubusercontent.com/13231763/231018701-240b6228-cc5f-4950-996f-31b1f5893ac2.mov

### Note the outline on text wrap

Screen Shot 2023-04-10 at 4 37 19 PM

### DIFF
--- a/.changeset/fair-olives-nail.md
+++ b/.changeset/fair-olives-nail.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-link": patch
+---
+
+Remove default display style
--- a/.changeset/tiny-clouds-turn.md
+++ b/.changeset/tiny-clouds-turn.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-link": minor
+---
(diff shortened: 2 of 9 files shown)


jandrade (approved):
Thanks for fixing this! 🚢

PR #1840, opened 2023-04-14
FEI-5092: Remove ref-forwarding from withActionScheduler

kevinbarabash:
## Summary:
It isn't possible to define HOCs in TS in a such a way that supports ref-forwarding across multiple HOCs. In order to make the TS migration as smoothe as possible, we're removing ref-forwarding from our HOCs so that the types and semantics match both before and after migration.

Since this is a change to the semantics of withActionScheduler, I've bumped wonder-blocks-timing by a major version.

Issue: FEI-5092

## Test plan:
- let CI run the checks

### DIFF
--- a/.changeset/sweet-pets-hear.md
+++ b/.changeset/sweet-pets-hear.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-timing": major
+---
+
+Remove ref-forwarding from withActionScheduler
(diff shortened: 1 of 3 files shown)


__docs__/wonder-blocks-link/link.stories.tsx
```diff
@@ -393,8 +393,8 @@ Inline.parameters = {
 Inline.play = async ({canvasElement}) => {
     const canvas = within(canvasElement);
 
-    const primaryLink = canvas.getByText("Primary link");
-    const secondaryLink = canvas.getByText("Secondary link");
```
kevinbarabash, line 397: These were capturing the `span` that was inside the `a` tag instead of the `a` tag itself.

kevinbarabash (commented):
The interaction test failures for `Link` are a pre-existing issue on `main`.  I decided to fix them as part of this PR so that we'd have a fix for them on `main` sooner.

PR #1844, opened 2023-04-19
Fix Link inline style

n1colemejia:
## Summary:
Fixed link to use aphrodite style instead of inline style.
Updated span to StyledSpan

Issue:

## Test plan:
Inspect in Storybook to confirm styling is correct.
http://localhost:6061/?path=/story/link--start-and-end-icons&globals=outline:true


jandrade (approved):
Looks great to me! thanks for the fix 🚢

PR #1846, opened 2023-04-20
Remove span from links without icons

n1colemejia:
## Summary:
Initially, the link text/`children` was nested inside a `span` element in order to use `verticalAlign` to center the text with icons.

Nesting `children` in a `span` broke other links that don't have icons (screenshots). 

This change checks first if icon props are present. If true, the text/`children` is nested inside a `span` element.

If the link has no icons, no `span` is needed.

Broken:

Screenshot 2023-04-20 at 1 11 13 PM

From production:

Screenshot 2023-04-20 at 1 11 06 PM

Also removed `textUnderlineOffset` from `sharedStyles` because without the `span` element, the underline sat too low when inline:

image

Added `textUnderlineOffset` to `sharedStyles.withIcon` and `restingInline` to match the underline when inline:

Screenshot 2023-04-20 at 4 35 58 PM

Issue: n/a

## Test plan:
For start and end icon props:
- In Storybook, inspect the Default `Link` story. 
- Confirm the link text is not nested inside a `span` element. 
- Add an icon via the control panel and notice in the DOM that a `span` element wraps the link text.

For external icon (since external icon relies on `target` prop:
- In Storybook, inspect the Default `Link` story. 
- Confirm the link text is not nested inside a `span` element. 
- Add "_blank" to `target` prop via the control panel and notice in the DOM that a `span` element wraps the link text.

### DIFF
--- a/packages/wonder-blocks-link/src/components/link-core.tsx
+++ b/packages/wonder-blocks-link/src/components/link-core.tsx
@@ -49,6 +49,11 @@ export default class LinkCore extends React.Component<Props> {
             ...restProps
         } = this.props;
 
+        // If icon props are present, link `children` will render inside a
+        // `StyledSpan` element to vertically align text with icons and add
+        // `textUnderlineOffset`
+        const withIcon = startIcon || endIcon || target === "_blank";
+
         const linkStyles = _generateStyles(inline, kind, light, visitable);
         const restingStyles = inline
             ? linkStyles.restingInline
@@ -65,6 +70,7 @@ export default class LinkCore extends React.Component<Props> {
             // focused link even after hovering and un-hovering on it.
             !pressed && hovered && linkStyles.hover,
             !pressed && focused && linkStyles.focus,
+            withIcon && sharedStyles.withIcon,
         ];
 
         const commonProps = {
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -49,6 +49,11 @@ export default class LinkCore extends React.Component<Props> {
             ...restProps
         } = this.props;
 
+        // If icon props are present, link `children` will render inside a
+        // `StyledSpan` element to vertically align text with icons and add
+        // `textUnderlineOffset`
+        const withIcon = startIcon || endIcon || target === "_blank";
```
jandrade, line 55: **praise:** Nice! I like this const name a lot, it's very descriptive.

packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -240,6 +252,7 @@ const _generateStyles = (
             // TODO(WB-1521): Update the underline offset to be 4px after
             // the Link audit.
             // textUnderlineOffset: 4,
+            textUnderlineOffset: 1.5,
```
jandrade: **question:** any particular reason why you choose 1.5 over 1? same question goes for 3.5 vs 3 in the icon style.

n1colemejia: I was messing with decimals to get a closer match, but noticed they cut off part of the parenthesis when used inline, so I just changed them to whole numbers again.

PR #1850, opened 2023-04-27
Add eslint rules for safe use of TS enums

kevinbarabash:
## Summary:
The new rules require that all enum values have initializers and that those initializers are literal values. TS enums work different from Flow enums, in particular they default to numeric value, e.g. 0, 1, 2, ... instead of a string representation of the identifier.

It's safe to use TS enums now since flowgen converts them to POJOs for us.

Issue: XXX-XXXX

## Test plan:
- n/a since we aren't using any enums yet


jeresig (approved):

kevinbarabash:
I'm not sure why coverage went down given that I only changed the eslint config.

PR #1856, opened 2023-05-03
Link broken layout fixes

n1colemejia:
## Summary:
- Move `verticalAlign` to only apply to links `withIcon`
- Add check for external vs relative link; only display external icon for external links with `target="_blank"`
- Update external links in SB
- Update unit tests to reflect external link changes

Issue: n/a

## Test plan:
`yarn jest`
Go to Default Link in SB.

For `verticalAlign` update:
- Inspect Default Link and confirm `verticalAlign` is not set to `bottom`.
- In control panel, add a `startIcon` or an `endIcon`.
- Inspect Link again and see that `verticalAlign: bottom` appear.
- You can also inspect the links in the Inline story and compare `verticalAlign` for links with and without the external icon.

For external icon update:
- In control panel, change `href` to an external path and set `target` to `_blank`.
- An external icon should only appear if both are true.
- Icon should not appear if `href` is relative OR if `target` is not set to `_blank`.


packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -80,7 +80,10 @@ export default class LinkCore extends React.Component<Props> {
             ...restProps,
         } as const;
 
-        // Default external icon
+        const linkUrl = new URL(href, window.location.origin);
+
+        const externalLink = linkUrl.origin !== window.location.origin;
```
jandrade, line 85: **nit (non-blocking):** It's probably a good practice to indicate that this is the evaluation of an expression to read it more as a boolean. In this case something like `isExternalLink` could help to improve readability a bit. Of course, No strong opinion on this, just a minor suggestion.

packages/wonder-blocks-link/src/components/__tests__/link.test.tsx
```diff
@@ -370,7 +370,18 @@ describe("Link", () => {
             expect(icon).toBeInTheDocument();
         });
 
-        test("does not render external icon when there is no target", () => {
+        test("does not render external icon when there is no target and link is external", () => {
```
jandrade, line 373: **question:** Is there a test for when `target="_blank" is set and link is relative?

jandrade (approved):
Looks great! thanks for updating all the stories and adding unit tests. Just a couple of (non-blocking) minor things for your consideration before landing. 🚀

PR #1858, opened 2023-05-03
Update flowgen to convert more React types properly

kevinbarabash:
## Summary:
Flowgen wasn't converting React.ForwardExoticRefComponent<> or React.ForwardedRef<> and Flow was just ignoring those types in webapp.  This was resulting it issues with incorrect props not being detected in some components.

Issue: None

## Test plan:
- yarn build:types
- yarn build:flowtypes
- check text-field.js.flow and add-style.js.flow in the dist folders of wonder-blocks-form and wonder-blocks-core to see that they contain React.AbstractComponent<> and React.Ref<> instead of the types listed in the summary.

### DIFF
--- a/.changeset/dirty-waves-remain.md
+++ b/.changeset/dirty-waves-remain.md
@@ -1,0 +1,7 @@
+---
+"@khanacademy/wonder-blocks-core": patch
+"@khanacademy/wonder-blocks-form": patch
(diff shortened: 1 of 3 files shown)


jandrade (approved):
LGTM! Thanks for fixing this so quickly 💯

PR #1909, opened 2023-05-12
Experiment with TS language service plugins

kevinbarabash:
## Summary:
The TypeScript language service supports plugins that can provide additional diagnostics.  This PR adds plugins to run eslint and type coverage.  Although VSCode plugins for eslint and type coverage exist, running them via plugins offers the following benefits:
- easier set up, devs don't have to install extra plugins
- a single point a failure, easier to notice when the service dies because all diagnostics go out together
- provides type coverage to editors/IDEs that don't have a type coverage plugin (JetBrains editors, vim, etc.)

I had to disable "typescript.tsserver.experimental.enableProjectDiagnostics" in .vscode/settings.json.  This option causes all files to be checked by the TS server even if the files aren't open.  As a result, eslint was running on files in node_modules even though those files are ignored.

People can still get project wide TS errors by running 'yarn tsc -w' in a terminal.

Issue: None

## Test plan:
- disable the eslint and type coverage VSCode plugins
- add some lint to a file
- add some uncovered code to file
- see that it's detected by the plugins

Screen Shot 2023-05-11 at 11 11 32 PM

Screen Shot 2023-05-11 at 11 11 22 PM

Screen Shot 2023-05-11 at 11 13 15 PM

## Performance
I used the following code for the type checking (toggling `y`'s type between `string` and `number`) and type coverage (toggling `x`'s type between `any` and `number`) tests:
```
export function foo(x: any, y: string) {
    return x + y;
}

foo(5, 10);
```

For the eslint tests, I added a `console.log()` by pasting it and seeing how long it took for the new lint to be reported.

NOTE: The tools reported diagnostics in response to either document update or document save events.  I've noted this in parentheses for each test.

### VSCode plugins:
- type checking (on update): 3.85s, 4.50s, 3.36s, 4.06s
- type coverage (on save): 3.87s, 3.93s, 3.38s, 4.30s
- eslint (on save): almost instantaneous, but requires a save

### TypeScript plugins:
- type checking (on update): 2.93, 2.90s, 2.78s, 3.17s
- type coverage (on save): 2.30s, 2.91s, 3.38s, 2.76s
- eslint (on update): 2.80s, 3.00s, 2.50s, 3.25s

NOTE: `typescript-eslint-language-service` can be disabled by setting the `TS_ESLINT_SERVICE_DISABLED` and restarting your editor/IDE.


tsconfig-common.json
```diff
@@ -24,6 +24,16 @@
...
+        "skipLibCheck": false,
+
+        "plugins": [
+            {
+                "name": "typescript-eslint-language-service"
+            },
+            {
+                "name": "ts-plugin-type-coverage",
+                "ignoreCatch": true,
+            }
```
kevinbarabash, line 36: There are a bunch of options that we can experiment, see https://github.com/plantain-00/type-coverage#config-in-packagejson.  TS doesn't support type annotations on the `e` in `catch (e)` which is why I've decided to start with `ignoreCatch: true`.  The tool also provides a special comments to ignore type coverage on certain parts of code, see https://github.com/plantain-00/type-coverage#ignore-line.

kevinbarabash:
>     * What is the performance like compared with using the other plugins directly?

I will try to get some numbers for this by timing how long it takes for diagnostics to update after making changes.

>     * If I have this and the ESLint plugin enabled at the same time, what happens?

Things will be double reported.  This means that VSCode users will need to disable the eslint plugin for any workspace using this configuration if they want to avoid this issue.

kevinbarabash:
@jeresig I added a "Performance" section to the initial comment with some perf numbers.

jeremywiebe (approved):
I'm not sure how I feel about this change. :)  I feel like it's still early days with TypeScript and there'll be some learning curve with what works "in anger".  I'm a bit sad to lose the whole project analysis setting. 

I say we land it and keep an eye open for rough edges that slow us down or make dev annoying. We can always revert this change.

packages/wonder-blocks-dropdown/src/components/dropdown-opener.tsx
```diff
@@ -56,18 +64,19 @@ class DropdownOpener extends React.Component<Props> {
     ): React.ReactElement {
         const {disabled, testId, text} = this.props;
         const renderedChildren = this.props.children({...eventState, text});
-        const childrenProps = renderedChildren.props;
-        const childrenTestId = this.getTestIdFromProps(childrenProps);
+        const rendererdChildrenProps: RenderedChildrenProps =
+            renderedChildren.props; // type-coverage:ignore-line
```
kevinbarabash, line 68: I wasn't able to cover this line with proper types.  Thankfully the type coverage tooling supports coverage exclusion comments.  The reported generated by `yarn coverage:types` uses the same underlying tooling so these comments should be respected when that report is generated.

kevinbarabash:
I'm going to abandon this for now since it breaks some people's workflows.

PR #1911, opened 2023-05-12
Turn off dependabot until we have time to configure it to run checks

kevinbarabash:
## Summary:
Dependabot landed some PRs that broken Storybook. We're going to turn of dependabot for a bit until we can fix the issue and put safeguards in place to prevent it from happening again.

Issue: None

## Test plan:
- n/a


jeresig (approved):

kevinbarabash:
I had the wrong base branch. 🤦‍♂️

PR #1914, opened 2023-05-16
FEI-5134: Remove all TypeScript error suppressiosn on JSX attributes

kevinbarabash:
## Summary:
We don't want there to be any @ts-expect-error comments on JSX attributes since this causes TS to opt out of all type checking on the entire JSX element.  This PR updates wonder-blocks to use our new custom ts-no-error-suppressions eslint rule but updating @khanacademy/eslint-config and adding @khanacademy/eslint-plugin.

Issue: FEI-5134

## Test plan:
- yarn lint
- yarn jest


packages/wonder-blocks-banner/src/components/banner.tsx
```diff
@@ -18,10 +18,12 @@ type ActionTriggerBase = {
...
 type ActionTriggerWithButton = ActionTriggerBase & {
+    type: "button";
     onClick: () => void;
 };
 
 type ActionTriggerWithLink = ActionTriggerBase & {
+    type: "link";
     href: string;
     onClick?: () => void;
 };
```
kevinbarabash, line 29: I bumped `wonder-blocks-banner` by a "major" version because this is a breaking change.

jeresig (approved):
Nice - seems like some good improvements, thank you!

PR #1917, opened 2023-05-22
FEI-5161: Update component using getDerivedStateFromProps to have readonly props and state

kevinbarabash:
## Summary:
The problem is that 'getDerivedStateFromProps' ends up in the .d.ts files we generate and the React lib defs have its 'props' param wrapped in Readonly<>. This is causing type errors in webapp.  We need to make sure all components that make use of 'getDerivedStateFromProps' are also using readonly props for this static method.  I've updated all props and state types to be readonly. This is generally a good practice since neither of those should be mutated directly.

I've created https://khanacademy.atlassian.net/browse/FEI-5164 to track writing a lint rule for this.  Ideally all components should have readonly props and state.

Issue: FEI-5161

## Test plan:
- yarn tsc
- let the tests run on CI

### DIFF
--- a/.changeset/wise-experts-sell.md
+++ b/.changeset/wise-experts-sell.md
@@ -1,0 +1,11 @@
+---
+"@khanacademy/wonder-blocks-clickable": patch
+"@khanacademy/wonder-blocks-dropdown": patch
+"@khanacademy/wonder-blocks-popover": patch
+"@khanacademy/wonder-blocks-tooltip": patch
+"@khanacademy/wonder-blocks-timing": patch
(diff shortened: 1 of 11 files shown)


jeresig (approved):
Great! This doesn't change our exposed API, does it?

kevinbarabash:
> Great! This doesn't change our exposed API, does it?

It does, but not in a meaningful way.  `getDerivedStateFromStore` shouldn't be called directly.  Also, it's fine to pass a mutable object as props to a component that only accepts readonly props.  Making `props` and `state` readonly should only affect the implementation of components and only if they were modifying `props` or `state` directly.

PR #1919, opened 2023-06-01
[namedcontexts] Make sure contexts have a name

somewhatabstract:
## Summary:
I was debugging some context-related changes in webapp and using React Dev Tools as well as the React Context extension for dev tools. I noticed that many of the contexts were given unhelpful generated names, which made it hard to find the context that I was looking for. This diff adds a display name to all of our contexts so that they can be more easily identified in dev tools.

This was found while working on the guide context stabilization work, so I'm tagging with that issue.

Issue: GL-568

## Test plan:
`yarn test`

### DIFF
--- a/.alexrc
+++ b/.alexrc
@@ -1,3 +1,3 @@
 {
-    "allow": ["invalid", "color", "colors", "white", "latin", "hook", "hooks"]
+    "allow": ["invalid", "color", "colors", "white", "latin", "hook", "hooks", "trap"]
 }
--- a/packages/wonder-blocks-core/src/components/render-state-context.ts
+++ b/packages/wonder-blocks-core/src/components/render-state-context.ts
@@ -22,7 +22,9 @@ export const RenderState = {
  * standard:
  *   means that we're all now doing non-SSR rendering
  */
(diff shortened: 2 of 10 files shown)


.alexrc
```diff
@@ -1,3 +1,3 @@
 {
-    "allow": ["invalid", "color", "colors", "white", "latin", "hook", "hooks"]
+    "allow": ["invalid", "color", "colors", "white", "latin", "hook", "hooks", "trap"]
```
somewhatabstract, line 2: **note:** We were getting a warning about this, but I think the use of this word is OK in our case.

packages/wonder-blocks-core/src/components/render-state-context.ts
```diff
@@ -22,7 +22,9 @@ export const RenderState = {
  * standard:
  *   means that we're all now doing non-SSR rendering
  */
-// @ts-expect-error [FEI-5019] - TS2322 - Type 'Context<"root">' is not assignable to type 'Context<"initial" | "root" | "standard">'.
```
somewhatabstract, line 25: **note:** I fixed this typing issue which also removed suppressions in a test file.

jandrade (approved):
LGTM! 🚀

PR #1920, opened 2023-06-07
[wb1556.1.notail] Make popover tail optional

somewhatabstract:
## Summary:
This makes it so that the popover tail can be hidden for designs that warrant it. This new prop defaults to true so that existing uses are unchanged.

I made this a minor update for popover since it's a new prop on the primary usage. I made it only patch for tooltip as the primary usage is unchanged (it's only the TooltipTail component that changes).

Here's the default story for the popover that shows the tail:
Screen Shot 2023-06-07 at 14 36 24

Here's the new No Tail story that shows the popover without the tail:
Screen Shot 2023-06-07 at 14 36 18

Issue: WB-1556

## Test plan:
`yarn start` and view the new story (and the old ones) to see that it works.

### DIFF
--- a/packages/wonder-blocks-tooltip/src/components/tooltip-bubble.tsx
+++ b/packages/wonder-blocks-tooltip/src/components/tooltip-bubble.tsx
@@ -84,6 +84,7 @@ export default class TooltipBubble extends React.Component<Props, State> {
             >
                 <View style={styles.content}>{children}</View>
                 <TooltipTail
+                    show={true}
                     updateRef={updateTailRef}
                     placement={placement}
                     offset={tailOffset}
--- a/packages/wonder-blocks-tooltip/src/components/tooltip-bubble.tsx
+++ b/packages/wonder-blocks-tooltip/src/components/tooltip-bubble.tsx
@@ -84,7 +84,6 @@ export default class TooltipBubble extends React.Component<Props, State> {
             >
                 <View style={styles.content}>{children}</View>
                 <TooltipTail
-                    show={true}
                     updateRef={updateTailRef}
                     placement={placement}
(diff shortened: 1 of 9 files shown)


packages/wonder-blocks-tooltip/src/components/tooltip-bubble.tsx
```diff
@@ -84,6 +84,7 @@ export default class TooltipBubble extends React.Component<Props, State> {
             >
                 <View style={styles.content}>{children}</View>
                 <TooltipTail
+                    show={true}
```
jeresig, line 87: Seems like we could remove this, since it default to true?

jeresig (approved):
Great - thanks for tweaking this!

packages/wonder-blocks-tooltip/src/components/tooltip-bubble.tsx
somewhatabstract: Yes, totally. I made this change before I set the default. Good catch

packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx
```diff
@@ -39,9 +39,7 @@ export default class TooltipPopper extends React.Component<Props> {
         // We'll hide some complexity from the children here and ensure
         // that our placement always has a value.
         const placement: Placement =
-            // We know that popperProps.placement will only be one of our
-            // supported values, so just cast it.
-            (popperProps.placement as any) || this.props.placement;
```
somewhatabstract, line 44: **note:**  Caught this while working through how to make this change. Tidy up.

PR #1921, opened 2023-06-20
Update "Mastery" purple

drewkiimon:
Update mastery color from `#9059FF` to `#594094`.

Makes me think the 4 different shades of "Mastery" should be under `Color`. 

# ON PAUSE
This color used used pretty frequently, and needs more discussion.

### DIFF
--- a/.changeset/swift-ties-happen.md
+++ b/.changeset/swift-ties-happen.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-color": minor
+---
+
+Updated the Color purple to updated spec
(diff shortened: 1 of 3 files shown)


jandrade:
@drewkiimon Has this change been tested against different backgrounds? We use this purple for the WB `Link` component, and from my first look it seems that the color difference makes it harder to see it as a visited link.

Screenshot 2023-06-20 at 2 40 19 PM

drewkiimon:
@jandrade I am not aware if it has or not. All I am aware of is:

1. We have changed the mastery colors based off design and a11y preferences
2. People use `Color.purple` as a way to show "Mastery" in the UI

Besides that I am unsure. It does not look good as a link color. The new purple for Mastered is darker than the previous color.  Should `Link` be using a different color?

PR #1927, opened 2023-07-01
Updating tertiary button focus state to be a border instead of underline

marianmwang:
## Summary:
Updated styling for tertiary button focus state to be a border instead of underline so that the states can be differentiated. Also updated some styling for other buttons.

Issue: WB-1420

## Test plan:
- Wrote new interaction test/story for tertiary button
- Updated snapshots in unit tests

| Focus state before | Focus state after |
|--------|--------|
| Screen Shot 2023-06-30 at 5 14 46 PM | Screen Shot 2023-07-05 at 10 58 47 AM |

### DIFF
--- a/packages/wonder-blocks-button/src/components/button-core.tsx
+++ b/packages/wonder-blocks-button/src/components/button-core.tsx
@@ -317,16 +317,14 @@ const _generateStyles = (
                 borderColor: light ? white50 : offBlack50,
                 borderStyle: "solid",
                 borderWidth: 1,
-                paddingLeft: iconWidth ? padding - 4 : padding,
+                paddingLeft: padding,
                 paddingRight: padding,
             },
             focus: {
                 background: light ? "transparent" : white,
                 borderColor: light ? white : color,
                 borderWidth: 2,
-                // The left padding for the button with icon should have 4px
-                // less padding
-                paddingLeft: iconWidth ? padding - 5 : padding - 1,
+                paddingLeft: padding - 1,
                 paddingRight: padding - 1,
             },
             active: {
@@ -336,9 +334,7 @@ const _generateStyles = (
                 borderWidth: 2,
                 // We need to reduce padding to offset the difference
                 // caused by the border becoming thicker on focus.
-                // The left padding for the button with icon should have 4px
-                // less padding
-                paddingLeft: iconWidth ? padding - 5 : padding - 1,
+                paddingLeft: padding - 1,
                 paddingRight: padding - 1,
             },
             disabled: {
@@ -350,9 +346,7 @@ const _generateStyles = (
                     borderWidth: 2,
                     // We need to reduce padding to offset the difference
                     // caused by the border becoming thicker on focus.
-                    // The left padding for the button with icon should have 4px
-                    // less padding
-                    paddingLeft: iconWidth ? padding - 5 : padding - 1,
+                    paddingLeft: padding - 1,
                     paddingRight: padding - 1,
                 },
             },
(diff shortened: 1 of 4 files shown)


marianmwang:
> @marianmwang I was reviewing the snapshots in Chromatic and it seems like the buttons have changed somehow, but it's unclear what may have changed, from looking at your code - can you take a look and confirm what may be happening? Thanks!
@jeresig Yep, the snapshots changed because of a 4px padding update (See this [commit](https://github.com/Khan/wonder-blocks/pull/1927/commits/ec3c65fd9b6ef0b0770927f6940b43a56ab33105)). The main change, though, was not captured in snapshots because it involves the "focus" state of the button being updated.

jeresig:
@marianmwang Gotcha - I think I missed a conversation around this, it seems like this is affecting the width of all buttons, not just the tertiary focus state. Is this something that you discussed with Reony? Can you link me to the Slack thread? Thanks!

jeresig (approved):
Discussed the spacing changed with Marian on Slack - I think this sounds reasonable to me. cc: @jandrade I asked Marian to provide some more context here.

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -377,17 +372,26 @@ const _generateStyles = (
...
+                    // Since we are using a pseudo element, we need to manually
+                    // calculate the width/height and use absolute position to
+                    // prevent other elements from being shifted around.
                     position: "absolute",
-                    height: 2,
-                    width: `calc(100% - ${iconWidth}px)`,
-                    right: 0,
-                    bottom: -1,
+                    width: `calc(100% + 4px)`,
+                    height: `calc(100% - 4px)`,
```
jandrade, line 383: **suggestion:** Using `Spacing` as much as we can would be helpful. Specially with the theming work starting very soon :).

```suggestion
                    width: `calc(100% + ${Spacing.xxxSmall_4}px)`,
                    height: `calc(100% - ${Spacing.xxxSmall_4}px)`,
```

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -377,17 +372,26 @@ const _generateStyles = (
...
+                    borderStyle: "solid",
+                    borderColor: light ? white : color,
+                    borderWidth: 2,
+                    borderRadius: 4,
```
jandrade, line 387: Same suggestion around using Spacing.
```suggestion
                    borderWidth: Spacing.xxxxSmall_2,
                    borderRadius: Spacing.xxxSmall_4,
```

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -105,19 +111,14 @@ export default class ButtonCore extends React.Component<Props> {
...
-                    kind === "tertiary" &&
-                        disabled &&
-                        focused && [
-                            buttonStyles.focus,
-                            buttonStyles.disabledFocus,
-                        ],
+                            ? [buttonStyles.hover, buttonStyles.active]
+                            : hovered && buttonStyles.hover),
                 ]}
+                testId="button-inner-label"
```
jandrade, line 121: **suggestion:** I'd recommend using a composed `testId` value. In this case works fine, but if for any reason we have multiple `Button` instances in `webapp`, then identifying this test-id for a specific button instance will be more complex.

As we already have a `testId` prop in `Button`, we can probably do something like:

```suggestion
                testId={testId ? `${testId}-inner-label` : undefined}
```

Then, you'd need to assign a `testId` value in the `Button` instance included in the interaction test for this to work.

jandrade (approved):
Looks good! thanks for refactoring and cleaning up the existing code. I left some minor suggestions to consider before landing this PR. :shipit:

PR #1928, opened 2023-07-06
Add style prop to tooltip

marianmwang:
## Summary:
Exposed new `contentStyle` (color and padding) and `backgroundColor` props on `Tooltip` to allow some custom styling.
image

Issue: WB-1410

## Test plan:
- `yarn test packages/wonder-blocks-tooltip`
- Experiment with styles in storybook

### DIFF
--- a/.changeset/pink-guests-accept.md
+++ b/.changeset/pink-guests-accept.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tooltip": minor
+---
+
+Added an optional style prop to Tooltip
--- a/.changeset/pink-guests-accept.md
+++ b/.changeset/pink-guests-accept.md
@@ -2,4 +2,4 @@
 "@khanacademy/wonder-blocks-tooltip": minor
 ---
 
-Added an optional style prop to Tooltip
+Added custom styling for background color, text color, and padding to Tooltip
--- a/__docs__/wonder-blocks-tooltip/tooltip.argtypes.ts
+++ b/__docs__/wonder-blocks-tooltip/tooltip.argtypes.ts
@@ -1,11 +1,13 @@
+import Color from "@khanacademy/wonder-blocks-color";
+
 export default {
     placement: {
         description:
             "Where the tooltip should appear in relation to the anchor element. Defaults to `top`.",
         control: {
             type: "select",
-            options: ["top", "bottom", "right", "left"],
         },
+        options: ["top", "bottom", "right", "left"],
     },
     title: {
         description: "Optional title for the tooltip content.",
@@ -21,8 +23,21 @@ export default {
     contentStyle: {
         description:
             "Optional subset of CSS styles for the tooltip content. Currently supports `color` and `padding`.",
+        control: {
+            type: "object",
+        },
+        table: {
+            type: {
+                summary: "ContentStyle",
+                detail: "Only supports color and padding styles.",
+            },
+        },
     },
     backgroundColor: {
         description: "Optional background color for the tooltip content.",
+        control: {
+            type: "select",
+        },
+        options: Object.keys(Color) as Array<string>,
     },
 };
--- a/__docs__/wonder-blocks-tooltip/tooltip.argtypes.ts
+++ b/__docs__/wonder-blocks-tooltip/tooltip.argtypes.ts
@@ -1,14 +1,28 @@
 export default {
     placement: {
+        description:
+            "Where the tooltip should appear in relation to the anchor element. Defaults to `top`.",
         control: {
             type: "select",
             options: ["top", "bottom", "right", "left"],
         },
     },
-
     title: {
+        description: "Optional title for the tooltip content.",
         control: {
             type: "text",
         },
+        table: {
+            type: {
+                summary: "string",
+            },
+        },
+    },
+    contentStyle: {
+        description:
+            "Optional subset of CSS styles for the tooltip content. Currently supports `color` and `padding`.",
+    },
+    backgroundColor: {
+        description: "Optional background color for the tooltip content.",
     },
 };
--- a/__docs__/wonder-blocks-tooltip/tooltip.stories.tsx
+++ b/__docs__/wonder-blocks-tooltip/tooltip.stories.tsx
@@ -315,6 +315,7 @@ export const WithStyle: StoryComponentType = () => {
                 content={`This is a styled tooltip.`}
                 backgroundColor="darkBlue"
                 opened={true}
+                testId="test-tooltip"
             >
                 My tooltip is styled!
             </Tooltip>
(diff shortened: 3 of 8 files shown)


nishasy (commented):
Hey, this is awesome! Thanks for doing this. One request though: do you think you could include a new story showing how styles can be used with tooltip? I'd probably call it something like `WithStyle`. Here's an example:

Storybook: https://5e1bf4b385e3fb0020b7073c-hzkkdjtzco.chromatic.com/?path=/story/modal-building-blocks-modalpanel--with-style

Github: https://github.com/Khan/wonder-blocks/blob/main/__docs__/wonder-blocks-modal/modal-panel.stories.tsx#L355

packages/wonder-blocks-tooltip/src/components/tooltip.tsx
```diff
@@ -209,7 +212,8 @@ export default class Tooltip extends React.Component<Props, State> {
                 {(props) => (
                     <TooltipBubble
                         id={bubbleId}
-                        style={props.style}
+                        // Allow additional styling to be applied
+                        style={[props.style, style]}
```
somewhatabstract, line 216: **question:** Do we have a business need for this? If not, I don't think we should make this change. If we do, then I think we need to discuss the right way to provide it.

I am concerned about styling the bubble directly this way as it creates a competition between the calculated styles from the popper that govern positioning and such, and the ones passed in. If we do want to provide these custom styles, I think they need to be provided in a way that doesn't compete this way, otherwise I think we're potentially inviting a really easy way to get buggy tooltips.

I think this is worth discussing more when @jandrade is back next week.

somewhatabstract (changes requested):
I am not comfortable with this change since the custom styles can override the important styles provided via the `TooltipPopper`, which are provided to position the bubble.

I think we need some additional discussion to identify the business case for this, and to work out a way to support that with something that is less fragile. Perhaps worth discussing when @jandrade is back next week.

packages/wonder-blocks-tooltip/src/components/tooltip.tsx
jandrade: Good question @somewhatabstract. This was needed for supporting some custom backgrounds in Perseus tooltips.

You can see some context here: https://github.com/Khan/perseus/pull/285

I agree with your concerns about allowing to override certain styles passed by Popper, so maybe we can find a way to pass custom styles using a different prop, or applying it to a different container inside the bubble.

somewhatabstract: Could the `TooltipContent` component take the prop instead?

marianmwang: i modified it so that we can only pass a subset of styles (`color` of the text, `padding`, and `backgroundColor`). this keeps the scope of the styling small but still covers the original intent of this ticket.

.changeset/pink-guests-accept.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tooltip": minor
+---
+
+Added an optional style prop to Tooltip
```
nishasy, line 5: I think you can update this description now that you're not adding a whole `style` prop anymore.

__docs__/wonder-blocks-tooltip/tooltip.argtypes.ts
```diff
@@ -1,14 +1,28 @@
...
+                summary: "string",
+            },
+        },
+    },
+    contentStyle: {
+        description:
+            "Optional subset of CSS styles for the tooltip content. Currently supports `color` and `padding`.",
+    },
+    backgroundColor: {
+        description: "Optional background color for the tooltip content.",
```
nishasy, line 26: Can you add the `control` field to these? Currently, it doesn't let you put in your own values in the control panel in the `Default` example, but I think that's something we should be able to do. And having `table { type: }}` would also be good for showing the types, especially for `contentStyles` so people can see exactly which styles can be changed.

__docs__/wonder-blocks-tooltip/tooltip.stories.tsx
```diff
@@ -306,6 +295,30 @@ Controlled.parameters = {
...
+export const WithStyle: StoryComponentType = () => {
+    return (
+        <View style={[styles.centered, styles.row]}>
+            <Tooltip
+                contentStyle={{
+                    color: Color.white,
+                    padding: Spacing.xLarge_32,
+                }}
+                content={`This is a styled tooltip.`}
+                backgroundColor="darkBlue"
```
nishasy, line 307: Optional suggestion: I think it might be helpful to have this tooltip always be open so that it's easier for devs to inspect.

packages/wonder-blocks-tooltip/src/components/tooltip.tsx
```diff
@@ -210,6 +228,8 @@ export default class Tooltip extends React.Component<Props, State> {
                     <TooltipBubble
                         id={bubbleId}
                         style={props.style}
+                        contentStyle={contentStyle}
+                        backgroundColor={backgroundColor}
```
nishasy, line 232: Hmm... It looks like the border radius and maybe drop shadow might be getting overridden with the custom styles? I'm not entirely sure off the top of my head why this could be happening. Let me know if you'd like help investigating!

image

packages/wonder-blocks-tooltip/src/components/tooltip.tsx
```diff
@@ -102,6 +102,14 @@ type Props = AriaProps &
...
          */
         testId?: string;
+        /**
+         * Optional custom styles for the tooltip content which are a subset of valid CSS styles.
+         */
+        contentStyle?: ContentStyle;
+        /**
+         * Optional background color.
+         */
+        backgroundColor?: keyof typeof Color;
```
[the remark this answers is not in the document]
marianmwang: content style contains subsets of CSS styles but here background color is more restrictive than the CSS background color prop. i thought it would be more appropriate to factor it out! 

in the code, it gets passed down in two places (i think to `TooltipContent` and `TooltipTail`) and i felt it was more clear if it was a separate prop. open to changing it if you think it would be fine inside `contentStyle` tho!

packages/wonder-blocks-tooltip/src/components/tooltip.tsx
marianmwang: good catch! i was changing the background color on the wrong component. it should have been on an inner view. this actually simplifies this a bit. thanks nisha :)

packages/wonder-blocks-tooltip/src/components/__tests__/tooltip.test.tsx
```diff
@@ -113,6 +113,35 @@ describe("Tooltip", () => {
...
+
+            // Act
+            const tooltipContent = screen.getByTestId("test-tooltip-content");
+            expect(tooltipContent).toBeInTheDocument();
+
+            // Assert
+            expect(tooltipContent.style.backgroundColor).toBe(
+                "rgb(24, 101, 242)",
+            );
+        });
```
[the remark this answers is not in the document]
marianmwang: that makes sense! i wasn't super clear on the purpose of interaction vs unit tests but this clears things up. updated it to an interaction test.

nishasy (approved):
Nice! Thanks for making the requested changes :tada:

.changeset/pink-guests-accept.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tooltip": minor
+---
+
+Added custom styling for background color, text color, and padding to Tooltip
```
somewhatabstract, line 5: **praise:** Love the clear description. 👍🏻

somewhatabstract (approved):
This looks great. Thanks for working on this to get a solid solution in place. Love the new story and the clarity of the API. Good work.

Thanks for being patient while I got round to reviewing again. 🙇🏻

PR #1929, opened 2023-07-07
try to repro issue with eslint results being different on CI

kevinbarabash:
## Summary:
For some reason running eslint on .ts(x) files behaves different on CI than locally for the static service in webpap.  This PR is attempting to repro the error in wonder-blocks.

Issue: None

## Test plan:
- let CI run


kevinbarabash:
I wasn't able to repro the issue. 😞

PR #1930, opened 2023-07-07
Added tabIndex and ref props to Clickable component

marianmwang:
## Summary:
Add tabIndex and ref props to Clickable

Issue: WB-1549

## Test plan:
- Write new unit tests for both props
- Add new story for ref prop
- Test on storybook

### DIFF
--- a/__docs__/wonder-blocks-clickable/clickable.argtypes.ts
+++ b/__docs__/wonder-blocks-clickable/clickable.argtypes.ts
@@ -30,7 +30,7 @@ export default {
     },
     tabIndex: {
         description: "Set the tabindex attribute on the rendered element.",
-        control: {type: "number", min: -1},
+        control: {type: "number", min: -1, max: 1},
         table: {
             type: {
                 summary: "number",
--- a/__docs__/wonder-blocks-clickable/clickable.argtypes.ts
+++ b/__docs__/wonder-blocks-clickable/clickable.argtypes.ts
@@ -28,6 +28,15 @@ export default {
             type: "text",
         },
     },
+    tabIndex: {
+        description: "Set the tabindex attribute on the rendered element.",
+        control: {type: "number", min: -1},
+        table: {
+            type: {
+                summary: "number",
+            },
+        },
+    },
     /**
      * States
      */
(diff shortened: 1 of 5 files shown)


packages/wonder-blocks-clickable/src/components/clickable.tsx
```diff
@@ -156,65 +155,67 @@ const StyledLink = addStyle(Link);
...
-export default class Clickable extends React.Component<Props> {
-    static defaultProps: DefaultProps = {
-        light: false,
-        disabled: false,
-    };
 
-    getCorrectTag: (
+const Clickable = React.forwardRef(function Clickable(
+    props: Props,
+    ref: React.ForwardedRef<any>,
```
marianmwang, line 161: Note: I used `any` type because the inner element has 3 possible types. I'm not sure if this is the best way to accomplish this or if I should try to limit it and typecast based on the situation.

__docs__/wonder-blocks-clickable/clickable.argtypes.ts
```diff
@@ -28,6 +28,15 @@ export default {
             type: "text",
         },
     },
+    tabIndex: {
+        description: "Set the tabindex attribute on the rendered element.",
+        control: {type: "number", min: -1},
```
jandrade, line 33: **praise:** Thanks for restricting this! one minor suggestion would be to also include `max: 1`, so users are aware about not putting any arbitrary number for a11y reasons (like letting the browser follow its natural flow order).

__docs__/wonder-blocks-clickable/clickable.stories.tsx
```diff
@@ -247,21 +248,57 @@ ClientSideNavigation.storyName = "Client-side Navigation";
...
-                "**NOTE:** If you want to navigate to an external URL and/or reload the window, make sure to use `href` and `skipClientNav={true}`",
+        storyDescription: `If you need to save a reference to the \`Clickable\` element , you can do
+        so using the \`ref\` prop. In this example, we want the element to receive focus when the
+        button is pressed. We can do this by creating a React ref of type \`HTMLButtonElement\` and
+        passing it into \`Clickable\`'s \`ref\` prop. Now we can use the ref variable in the
+        \`handleSubmit\` function to shift focus to the field.`,
+        chromatic: {
+            // we don't need screenshots because this story only tests behavior.
+            disableSnapshot: true,
         },
```
jandrade, line 302: **issue:** It looks like the snapshot is still generated in Chromatic. This key should be moved one level up (to `parameters).

jandrade (approved):
Looks great! There's one minor change to avoid generating the Chromatic snapshot to modify before landing :shipit:

PR #1931, opened 2023-07-10
[fei5213.1.useforceupdatefix] Fix useForceUpdate

somewhatabstract:
## Summary:
This update makes sure that no matter how many times `forceUpdate` is called, it will trigger a re-render. Before this change, the new test case would fail because the boolean value, after multiple calls, would not have actually changed, so React never forced a new render.

With this update, things work as intended.

Issue: FEI-5213

## Test plan:
`yarn test`

### DIFF
--- a/.changeset/tough-beans-matter.md
+++ b/.changeset/tough-beans-matter.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-core": patch
+---
+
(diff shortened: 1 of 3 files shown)


somewhatabstract:
> LGTM. 🚢 Thanks for fixing this. A simpler approach might be to set the state to be a random number. That would avoid the need for a ref. The solution put forward in this PR is fine as well. Thanks for adding a test for the edge case that you encountered. 💚

There would need to be a comparison of

PR #1932, opened 2023-07-10
Center inline link text with icon by removing styled span and updating vertical align

marianmwang:
## Summary:
Removed `StyledSpan` from links with icons because the `verticalAlign: center` and underline offset styles were not being applied correctly on Safari.
Also removed `withIcon` styling to keep link underline offset consistent.

| Before (in Safari) | After (in Safari) |
| --- | --- |
|Screen Shot 2023-07-10 at 2 18 28 PM | Screen Shot 2023-07-10 at 2 18 56 PM |

Issue: WB-1550

## Test plan:
- Manual testing by checking storybook Link stories on both Chrome and Safari to ensure no inconsistencies between them
- Run `yarn test`

### DIFF
--- a/packages/wonder-blocks-link/src/components/link-core.tsx
+++ b/packages/wonder-blocks-link/src/components/link-core.tsx
@@ -25,7 +25,6 @@ type Props = SharedProps &
 
 const StyledAnchor = addStyle("a");
 const StyledLink = addStyle(Link);
-const StyledSpan = addStyle("span");
 
 export default class LinkCore extends React.Component<Props> {
     renderInner(router: any): React.ReactNode {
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -165,7 +152,8 @@ const linkContentStyles = StyleSheet.create({
         marginInlineStart: Spacing.xxxSmall_4,
     },
     centered: {
-        verticalAlign: "middle",
+        // Manually align the bottom of start/end icons with the text baseline.
+        verticalAlign: "-10%",
```
jandrade, line 156: **question:** have you tested this change in different browsers? I'm not fully sure if this will be consistent or not. Also, what would happen if the button text has a different font size?

marianmwang: if the button text has a different font size, the icon will sit closer to the baseline of the text instead of upper-middle area. the main change here has to do with removing this style from the actual link text, which was what caused it to sit lower than expected.
image

jandrade: Got it, thanks for the example. Let's land this in `webapp` and see how it works with all the different instances.

jandrade (approved):
The changes make sense to me. Thanks for simplifying this logic. 🚢

PR #1933, opened 2023-07-12
[forceupdatefix2] Tweak implementation so this works in all scenarios

somewhatabstract:
## Summary:
While the previous "fix" corrected the issue of multiple calls before a render, it broke the situation where a hook that is based off `useForceUpdate` may need to trigger an async update outside of a render.

I discovered the issue when integrating the update into webapp where a hook failed because of the issue resolved by this PR.

This change ensures that every call to force an update will do a change, and thus make sure that calling hooks can trigger updates asynchronously between renders.

I have added an additional test case to capture the scenario where things broke. The original implementation handled it, but the updated one did not. This implementation should handle both situations.

Issue: FEI-5213

## Test plan:
`yarn test`

### DIFF
--- a/packages/wonder-blocks-core/src/hooks/use-force-update.ts
+++ b/packages/wonder-blocks-core/src/hooks/use-force-update.ts
@@ -13,25 +13,17 @@ import * as React from "react";
  * @returns {() => void} A function that forces the component to update.
  */
 export const useForceUpdate = (): (() => void) => {
-    const updatePendingRef = React.useRef(false);
-    const [, setUpdateToggle] = React.useState(false);
+    const [, setUpdateState] = React.useState(0);
 
     const forceUpdate = React.useCallback(() => {
-        if (updatePendingRef.current) {
-            // If an update is already pending, then we do nothing.
-            return;
-        }
-
-        // Otherwise, if we haven't been asked to force an update since our
-        // last render then we toggle the state to invoke a render.
-        setUpdateToggle((toggle) => !toggle);
-        updatePendingRef.current = true;
+        setUpdateState((state) => {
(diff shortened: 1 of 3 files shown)


jeresig (approved):

packages/wonder-blocks-core/src/hooks/use-force-update.ts
```diff
@@ -13,25 +13,16 @@ import * as React from "react";
...
-        // Otherwise, if we haven't been asked to force an update since our
-        // last render then we toggle the state to invoke a render.
-        setUpdateToggle((toggle) => !toggle);
-        updatePendingRef.current = true;
+        // We leverage here that every new object instance will be seen
+        // as a state change. This is a little hacky but it works better than
+        // a boolean that would just flip-flop and could not trigger a render,
+        // or a random number that could repeat values and also then not
+        // trigger a render. This will always work.
+        setUpdateState({});
```
kevinbarabash, line 24: Neat!  This feels like a variation on my "random number" suggestion.

kevinbarabash (approved):
LGTM. 🚢

packages/wonder-blocks-core/src/hooks/use-force-update.ts
somewhatabstract: Yes, I tried that but abandoned it for two reasons:

1. It didn't guarantee a fix to the original issue since a random number could repeat a number already seen.
2. It was more complicated.

kevinbarabash: I like that this guarantees the state value is always new.  Good call. 🙂

PR #1935, opened 2023-07-14
Fix the issue where the anchor is no longer shift-tabbable after closing popover

nishasy:
## Summary:
Currently, you can't tab back to the anchor after closing a popover.
It seems to be caused by the line changed in this PR.

For context, the `|| this.nextElementAfterPopover` was added
in order to reduce the number of re-calculations, but it
seems like removing this condition is necessary for correct behavior.

Issue: https://khanacademy.atlassian.net/browse/WB-1559

## Test plan:
`yarn jest packages/wonder-blocks-popover/src/components/__tests__/popover.test.tsx`

^ I confirmed that the test fails with the old code (`if (!anchorElement || this.nextElementAfterPopover)`) and passes with the new code (`if (!anchorElement)` ) 🎉 

## Demo (includes both before and after in video)

https://github.com/Khan/wonder-blocks/assets/13231763/cf5e3d1e-25ba-410f-896a-7150af22fb15

### DIFF
--- a/packages/wonder-blocks-popover/src/components/__tests__/popover.test.tsx
+++ b/packages/wonder-blocks-popover/src/components/__tests__/popover.test.tsx
@@ -2,6 +2,9 @@ import * as React from "react";
 import {render, screen, waitFor} from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 
+import {View} from "@khanacademy/wonder-blocks-core";
+import Button from "@khanacademy/wonder-blocks-button";
+
 import Popover from "../popover";
 import PopoverContent from "../popover-content";
 import {PopoverContentCore} from "../../index";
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-popover/src/components/__tests__/popover.test.tsx
```diff
@@ -181,4 +184,58 @@ describe("Popover", () => {
             expect(screen.queryByText("Title")).not.toBeInTheDocument();
         });
     });
+
+    it("should shift-tab back to the anchor after popover is closed", async () => {
```
jandrade, line 188: **praise:** Thanks for adding this test!

jandrade (approved):
Looks great! :shipit:

PR #1938, opened 2023-07-18
Truncate text when icon is added to button

marianmwang:
## Summary:
Long text on buttons was not getting truncated when an icon was present. I updated the styling so that the text box width is correctly constrained and the button text is truncated instead of cut off.

Issue: WB-1561

| Before | After | 
| --- | --- |
| before | after |

## Test plan:
- Create new story that displays this behavior
- Test on storybook
- Run `yarn test`


packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -140,6 +131,15 @@ export default class ButtonCore extends React.Component<Props> {
 
         const contents = (
             <React.Fragment>
+                {icon && (
+                    <Icon
+                        size={iconSize}
+                        color="currentColor"
+                        icon={icon}
+                        style={sharedStyles.icon}
+                        aria-disabled="true"
```
jandrade, line 140: **suggestion:** probably using `aria-hidden` here would be better. That way the icon is not announced by screen readers. Did you add `aria-disabled` for a different purpose?

```suggestion
                        aria-hidden="true"
```

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -140,6 +131,15 @@ export default class ButtonCore extends React.Component<Props> {
 
         const contents = (
             <React.Fragment>
+                {icon && (
```
jandrade, line 134: **praise:** Neat!

jandrade (approved):
LGTM! Just a minor change to improve a11y a little bit. 🚀

packages/wonder-blocks-button/src/components/button-core.tsx
marianmwang: my bad! i think i mixed up the aria labels. i'll update that

PR #1939, opened 2023-07-19
Fix banner overflow

nishasy:
## Summary:
There are a couple of banners where the copy is being cut off because
of the overflow on the Wonder Blocks banner component. Here, I'm updating
it so that the text in WB Banner has `overflowWrap: break-word` so that
the long text line wraps

Issue: https://khanacademy.atlassian.net/browse/FEI-4898
also https://khanacademy.atlassian.net/browse/FEI-4823

## Test plan:
Storybook

http://localhost:50875/?path=/docs/banner--long-text

- Resize the page and confirm that the text never goes under the
  dismiss button, and that it never gets cut off of to the side.
  The text should wrap.

### Before:
Screenshot 2023-07-19 at 2 23 03 PM

### After:
Screenshot 2023-07-19 at 2 24 48 PM

### DIFF
--- a/.changeset/fuzzy-buses-breathe.md
+++ b/.changeset/fuzzy-buses-breathe.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-banner": patch
+---
+
+Fix the banner text overflow so that long text line wraps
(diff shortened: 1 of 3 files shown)


lillialexis (approved):

jandrade (approved):
LGTM! Thanks for fixing this 🚢

PR #1949, opened 2023-07-26
Add ref forwarding to Link

nishasy:
## Summary:
Forward refs in Link and LinkCore.
Link and LinkCore have been converted to functional components.

## Test plan:
`yarn jest packages/wonder-blocks-link/src/components/__tests__/link.test.tsx`


packages/wonder-blocks-link/src/components/__tests__/link.test.tsx
nishasy, line 1: I added the tests to `link.test.tsx` instead of `ref-forwarded.test.tsx` so that I can test the specific case where it uses client side navigation.

nishasy: Actually, I think I can move this to `ref-forwarded.test.tsx` and add it to the section that Marian added in the [Button PR](https://github.com/Khan/wonder-blocks/pull/1951/files) after that is landed.

jandrade (approved):
LGTM! :shipit:

PR #1950, opened 2023-07-26
WB-1566: Add `wonder-blocks-theming` package

jandrade:
## Summary:
Includes the skeleton of the new `wonder-blocks-theming` package.

Issue: WB-1566

## Test plan:

Verify that the checks pass.

NOTE: The build size check will fail as there's no way to compare against a
previous version of this newly-created package.


packages/wonder-blocks-theming/package.json
```diff
@@ -0,0 +1,28 @@
+{
+    "name": "@khanacademy/wonder-blocks-theming",
+    "version": "0.0.1",
+    "publishConfig": {
+        "access": "public"
+    },
+    "description": "",
```
somewhatabstract, line 7: **nit:** Add a simple description, like:

`Provide theming support for Wonder Blocks components"

packages/wonder-blocks-theming/package.json
```diff
@@ -0,0 +1,28 @@
+{
+    "name": "@khanacademy/wonder-blocks-theming",
+    "version": "0.0.1",
+    "publishConfig": {
+        "access": "public"
```
somewhatabstract, line 5: **question:** Should this be a private package until it's ready for release?

somewhatabstract (approved):
LGTM Some inline things.

packages/wonder-blocks-theming/package.json
jeresig: I think it's ok to have this be public - even if we release as an early 0.0.1 that's fine!

jandrade: FYI This won't be released yet as it will be initially landed in a separate branch.

My plan is to land all the theming-related PRs into this branch (`feature/theming`), but I can land all of this directly in `main` instead. Do y'all have a workflow preference?

jandrade: If I land everything in `feature/theming`, then I'll have to create another PR at the end (which includes all the changes) to land everything in `main`. We've used this approach in the past with `Cell` and `Banner`, but happy to use `main` directly if there's fit.

PR #1951, opened 2023-07-26
Add ref forwarding for Button and IconButton

marianmwang:
## Summary:
- Added ref forwarding to Button
- Added ref forwarding to IconButton
- Removed duplicate Clickable test
- Wrote new tests for ref forwarding

Issue: WB-1579

## Test plan:
`yarn test consistency_tests`

### DIFF
--- a/.changeset/great-cherries-compete.md
+++ b/.changeset/great-cherries-compete.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/wonder-blocks-icon-button": minor
+"@khanacademy/wonder-blocks-button": minor
+---
+
+Add ref forwarding to Button and IconButton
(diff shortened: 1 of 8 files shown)


.changeset/great-cherries-compete.md
```diff
@@ -0,0 +1,7 @@
+---
+"@khanacademy/wonder-blocks-icon-button": minor
+"@khanacademy/wonder-blocks-button": minor
+"@khanacademy/wonder-blocks-clickable": patch
```
jandrade, line 4: **nit:** In this case, Clickable doesn't require a version bump as we are only modifying tests (not code that will be included in the bundle).

jandrade (approved):
LGTM! Thanks for updating clickable as well 🚢

.changeset/great-cherries-compete.md
marianmwang: would removing that line be enough to not generate the new patch? i saw that you can do `yarn changeset --empty` but i think that only works if it's just one package

PR #1952, opened 2023-07-26
Add ref forwarding to View

marianmwang:
## Summary:
Added ref forwarding prop to View

Issue: WB-1578

## Test plan:
`yarn test consistency-tests`

### DIFF
--- a/consistency-tests/__tests__/ref-forwarded.test.tsx
+++ b/consistency-tests/__tests__/ref-forwarded.test.tsx
@@ -24,6 +24,7 @@ import LabelSmall from "../../packages/wonder-blocks-typography/src/components/l
 import LabelXSmall from "../../packages/wonder-blocks-typography/src/components/label-xsmall";
 import Tagline from "../../packages/wonder-blocks-typography/src/components/tagline";
 import Title from "../../packages/wonder-blocks-typography/src/components/title";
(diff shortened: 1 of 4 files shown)


nishasy (commented):
LGTM but I'll wait for someone else to approve because I'm not entirely sure how the `ForwardRefExoticComponent` works.

consistency-tests/__tests__/ref-forwarded.test.tsx
```diff
@@ -52,3 +53,16 @@ describe("Typography elements", () => {
         expect(ref.current).toBeInstanceOf(type);
     });
 });
+
+describe("View element", () => {
```
jandrade, line 57: **suggestion:** I think it would be good to at least include one of the other tags to see how it behaves in that case (and see how it is typed).

jandrade: Also good to include here some tests like the ones @kevinbarabash mentioned here.

https://khanacademy.slack.com/archives/C4PE1QM5Y/p1690409647322639?thread_ts=1690406084.915429&cid=C4PE1QM5Y

jandrade (approved):
Looks great! thanks for adding the extra tests 🏅

PR #1953, opened 2023-07-26
Add ref forwarding to Breadcrumbs

nishasy:
## Summary:
Add ref forwarding to Breadcrumbs. To make this easier, I also
changed it to a functional component.

## Test plan:
`yarn jest packages/wonder-blocks-breadcrumbs/src/components/__tests__/breadcrumbs.test.tsx`

Confirm that the Breadcrumbs work as expected in storybook

### DIFF
--- a/packages/wonder-blocks-breadcrumbs/src/components/__tests__/breadcrumbs.test.tsx
+++ b/packages/wonder-blocks-breadcrumbs/src/components/__tests__/breadcrumbs.test.tsx
@@ -35,20 +35,4 @@ describe("Breadcrumbs", () => {
             "test",
         );
     });
-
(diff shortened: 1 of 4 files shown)


packages/wonder-blocks-breadcrumbs/src/components/__tests__/breadcrumbs.test.tsx
nishasy, line 1: Adding the test here instead of in `ref-forwarded.test.tsx` because it needs the BreadcrurmbsItem children, and I figured it would be easier to put it here than to complicate the other file.

jandrade (approved):
🚀

PR #1956, opened 2023-07-28
Forward refs in Checkbox

nishasy:
## Summary:
Forward refs in Checkbox, ChoiceInternal, and CheckboxCore.
To make this easier, I converted all of these into functional components.

## Test plan:

Go to the storybook page for Checkbox, and confirm that everything
works as expected.


__docs__/wonder-blocks-form/checkbox.stories.tsx
```diff
@@ -88,6 +100,77 @@ Indeterminate.parameters = {
...
+
+IndeterminateWithGroup.parameters = {
+    docs: {
+        storyDescription: `Here is an example of how you can use the
+            indeterminate checkbox to select all or deselect all options
+            in a checkbox group. If only some of the options are selected,
+            the indeterminate checkbox will be in the indeterminate state.`,
+    },
+};
+
```
nishasy, line 173: I figured it would be useful to add an example of how someone might want to use indeterminate checkboxes in practice.

packages/wonder-blocks-form/src/components/checkbox-core.tsx
```diff
@@ -39,27 +39,13 @@ const indeterminatePath: IconAsset = {
 /**
  * The internal stateless ☑️ Checkbox
  */
-export default class CheckboxCore extends React.Component<ChoiceCoreProps> {
-    componentDidMount(): void {
-        if (this.props.checked == null && this.inputRef.current != null) {
-            this.inputRef.current.indeterminate = true;
```
nishasy, line 45: I'm a little confused here... removing this `componentDidMount()` and `componentDidUpdate()` didn't seem to change anything. All the behavior works as expected. And I don't think we're using the internal value of `indeterminate` anywhere?

Can I keep it this way without setting this value? Or would it be preferable for me to add this back in a `useEffect()`? (cc @kevinbarabash @jandrade)

nishasy: Is this so that the `:indeterminate` pseudoselector can be used? I don't think we use pseudoselectors in our code though..

__docs__/wonder-blocks-form/checkbox.stories.tsx
```diff
@@ -37,8 +38,19 @@ Default.parameters = {
...
+    const [checked, setChecked] = React.useState<boolean | null>(null);
+
+    const handleChange = () => {
+        if (checked === false) {
+            setChecked(true);
+        } else {
+            // If `checked` is true (checked) OR null/undefined (indeterminate),
+            // we want to change it to false (unchecked).
+            setChecked(false);
+        }
```
marianmwang, line 50: this can be condensed to `setChecked(checked === false)`.

packages/wonder-blocks-form/src/components/choice-internal.tsx
```diff
@@ -142,22 +138,27 @@ type DefaultProps = {
                                 tabIndex={-1}
                             >
                                 <ChoiceCore
-                                    {...coreProps}
+                                    // {...coreProps}
```
marianmwang, line 141: should this be removed entirely?

marianmwang (commented):
nice! you are powering through ref forwarding 🚀

consistency-tests/__tests__/ref-forwarded.test.tsx
```diff
@@ -6,6 +6,7 @@ import * as ReactDOM from "react-dom";
 import Breadcrumbs from "../../packages/wonder-blocks-breadcrumbs/src/components/breadcrumbs";
 import BreadcrumbsItem from "../../packages/wonder-blocks-breadcrumbs/src/components/breadcrumbs-item";
 import Button from "../../packages/wonder-blocks-button/src/components/button";
+import Checkbox from "../../packages/wonder-blocks-form/src/components/checkbox";
```
marianmwang, line 9: forgot to mention this in my review but this path should be added to `tsconfig.json` in `consistency-tests/`. all the dependencies are supposed to be listed there. more context in this slack thread: https://khanacademy.slack.com/archives/C4PE1QM5Y/p1690396563362009

__docs__/wonder-blocks-form/checkbox.stories.tsx
```diff
@@ -53,9 +61,10 @@ Controlled.parameters = {
         disableSnapshot: true,
     },
     docs: {
-        description: {
-            story: "Use state to keep track of whether the checkbox is checked or not",
-        },
+        description: `Use state to keep track of whether the checkbox
```
jandrade, line 64: **suggestion:** `description.story` is needed so Storybook displays the description next to the example. You can see how it's missing here: 

https://5e1bf4b385e3fb0020b7073c-nwtpfilbpc.chromatic.com/?path=/docs/form-checkbox--docs#controlled

__docs__/wonder-blocks-form/checkbox.stories.tsx
```diff
@@ -96,6 +105,77 @@ Indeterminate.parameters = {
...
+                    ))}
+                </CheckboxGroup>
+            </View>
+        </View>
+    );
+};
+
+IndeterminateWithGroup.parameters = {
+    docs: {
+        description: `Here is an example of how you can use the
```
jandrade, line 172: Same comment here about adding the `story` key.

https://5e1bf4b385e3fb0020b7073c-nwtpfilbpc.chromatic.com/?path=/docs/form-checkbox--docs#indeterminate-with-group
Screenshot 2023-08-01 at 10 39 16 AM

packages/wonder-blocks-form/src/components/checkbox-core.tsx
jandrade: This is to programmatically add the `indeterminate` value. You can see an explanation here:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes

> This is set using the [HTMLInputElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement) object's indeterminate property via JavaScript (it cannot be set using an HTML attribute)

jandrade (commented):
Looking good so far. I think we still need to programmatically include the `indeterminate` state so the `input` can have that information set via JS.

jandrade (approved):
Looks great! thanks for doing the last changes 💯

PR #1958, opened 2023-07-29
Forward refs in Choice and Radio

nishasy:
## Summary:
Forward refs in Choice, Radio, and RadioCore.
To make this easier, converted them into functional components.

## Test plan:

Confirm that the Radio and Choice storybook pages behave right.


__docs__/wonder-blocks-form/choice.stories.tsx
nishasy, line 1: Added different IDs to the checkboxes vs radios so that hovering/clicking on their labels works as expected.

jandrade (approved):
LGTM! :shipit:

PR #1960, opened 2023-08-01
Forward refs in RadioGroup

nishasy:
## Summary:
Forward refs in RadioGroup. To make this easier,
I converted it into a functional component.

While I'm here, I also noticed that clicking on most of the
labels in the CheckboxGroup and RadioGroup stories caused
a jump to the top of the page. This is because the values on the
Choice components were not unique, they were duplicates of the
values on the first story. This caused clicking the label to jump to
the first input with the associated value. To fix this, I updated
all the inputs to have unique `value` fields.

## Test plan:

Go to the storybook page for RadioGroup and confirm that
everything behaves as expected.


packages/wonder-blocks-form/src/components/radio-group.tsx
```diff
@@ -149,13 +154,15 @@ export default class RadioGroup extends React.Component<RadioGroupProps> {
                             groupName: groupName,
                             id: `${groupName}-${value}`,
                             key: value,
-                            onChange: () => this.handleChange(value),
+                            onChange: () => handleChange(value),
```
jandrade, line 157: **nit:** Out of scope for this PR, but I noticed that `handleChange` could go away and instead we could assign directly the prop:

```suggestion
                            onChange: onChange,
```

jandrade (approved):
Looks great! thanks for refactoring forms 🏅

PR #1965, opened 2023-08-02
Use named functions in all components with forwardRef()

nishasy:
## Summary:
After updating a bunch of components to use forwardRef(),
we realized that most of them were anonymous functions, which
caused the components to be named `<ForwardRef />` instead of
their action name. (This showed up in the snapshots when
attempting to release to webapp.)

Here, I changed all of these to named functions so that they
show up correctly in snapshot tests.

Issue: none

## Test plan:
N/A?

### DIFF
--- a/packages/wonder-blocks-breadcrumbs/src/components/breadcrumbs-item.tsx
+++ b/packages/wonder-blocks-breadcrumbs/src/components/breadcrumbs-item.tsx
@@ -29,41 +29,42 @@ const StyledSvg = addStyle("svg");
 /**
  * The BreadcrumbsItem represents an individual item in the breadcrumbs list.
  */
-const BreadcrumbsItem = React.forwardRef(
-    (props: Props, ref: React.ForwardedRef<HTMLLIElement>) => {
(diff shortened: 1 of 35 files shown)


packages/wonder-blocks-breadcrumbs/src/components/breadcrumbs-item.tsx
```diff
@@ -29,41 +29,42 @@ const StyledSvg = addStyle("svg");
...
-                    width={16}
-                    height={16}
-                    viewBox="0 0 16 16"
-                    aria-hidden={true}
-                >
-                    <circle cx="8" cy="9" r="1.5" />
-                </StyledSvg>
-            );
-        };
+const BreadcrumbsItem = React.forwardRef(function BreadCrumbsItem(
```
jandrade, line 32: **nit:** No change required tbh... 

```suggestion
const BreadcrumbsItem = React.forwardRef(function BreadcrumbsItem(
```

jandrade (approved):
Makes sense to me! thanks for fixing all the names 🚀

PR #1966, opened 2023-08-03
Add deploy workflow using the official GitHub Action

jandrade:
## Summary:

After upgrading to Storybook v7, the current GH pages site is no longer working.
This PR adds a new workflow that uses the official GitHub Action to deploy the
site.

NOTE: We switched to the official action as the previous one was using Jekill
(which I suspect had some conflicts with the way SB7 was building the site). The
official version only upload the static files instead of using Jekill as an
intermediate step.

Issue: XXX-XXXX

## Test plan:

Run the deploy workflow using the `Deploy` workflow_dispatch event.

Verify that https://khan.github.io/wonder-blocks/ is working.


jeresig (approved):
Seems reasonable to me!

PR #1968, opened 2023-08-04
FEI-5000: Remove Flow infra and update code to use TS enums and disjoint props

kevinbarabash:
## Summary:
Removing 'build:flowtypes' will speed up CI.  The TS enums should behave the same as the POJOs but some type annotations may need updating in downstream repos.

Issue: FEI-5000

TODO:
- [x] write .typestset.tsx files for these changes

## Test plan:
- yarn tsc
- yarn test


jeresig (approved):
yay! Thanks for cleaning this up!

PR #1976, opened 2023-08-08
Create new primitive Switch component based on implementation spec

marianmwang:
## Summary:
- Add 'switch' role to ClickableBehavior
- Create new Switch component
- Add unit tests for Switch
- Add stories for Switch
- Add ref forwarding test for Switch

image 
image

Issue: WB-1581

## Test plan:
- Write new unit tests and run `yarn test packages/wonder-blocks-switch`
- Create new stories for use cases

### DIFF
--- a/__docs__/wonder-blocks-switch/switch-best-practices.mdx
+++ b/__docs__/wonder-blocks-switch/switch-best-practices.mdx
@@ -1,0 +1,63 @@
+import {Canvas, Meta} from "@storybook/blocks";
+
+import * as SwitchBestPracticesStories from "./switch-best-practices.stories";
+
+<Meta of={SwitchBestPracticesStories} />
+
+# Switch
+
+## Best Practices
+
+### With Labelling
+
+The switch can be paired with a visible label which should be an html `label` element. The label
+should include the `htmlFor` attribute, and the switch should include the `aria-labelledby`
+attribute.
+
+#### Label
+
+The label text should __not__ change as the state of the switch changes.
+
+<Canvas of={SwitchBestPracticesStories.WithLabel} />
+
+#### Label, Description
+
+If a description is also provided, the switch should include the `aria-describedby` attribute.
+
+<Canvas of={SwitchBestPracticesStories.WithLabelAndDescription} />
+
+#### Label, ON/OFF Labels
+
+If on/off labels are desired, they should include the `aria-hidden` attribute to prevent
+redundant descriptions of the state for screen readers.
+
+<Canvas of={SwitchBestPracticesStories.WithLabelAndOnOff} />
+
+### Inside Cells
+
+The switch can be placed inside a cell as a left/right accessory.
+
+In the following examples, the title is a `label` element with the `htmlFor` attribute set to the
+switch, so that clicking the title also changes the state of the switch. The `onClick` attribute
+is omitted from the cell, and the `onChange` exists on the switch as normal.
+
+#### Compact Cell
+
+<Canvas of={SwitchBestPracticesStories.InsideCell} />
+
+#### Detailed Cell
+
+In the detailed cell, the title should change the state of the switch, and the description should
+not.
+
+<Canvas of={SwitchBestPracticesStories.InsideDetailCell} />
+
+### With a Tooltip
+
+The switch can be wrapped with a tooltip that describes the purpose and state of the switch.
+
+<Canvas of={SwitchBestPracticesStories.WithTooltip} />
+
+### References
+
+For more details, see the [Accessibility section](https://www.w3.org/WAI/ARIA/apg/patterns/switch/#wai-ariaroles,states,andproperties) in w3.org.
--- a/__docs__/wonder-blocks-switch/switch.argtypes.tsx
+++ b/__docs__/wonder-blocks-switch/switch.argtypes.tsx
@@ -1,0 +1,105 @@
+import * as React from "react";
+
+import Icon, {icons} from "@khanacademy/wonder-blocks-icon";
+
+const iconsMap: Record<string, React.ReactElement<typeof Icon>> = {};
+
+Object.entries(icons).forEach(([iconLabel, iconValue]) => {
+    iconsMap[iconLabel] = <Icon icon={iconValue} />;
+});
+
+export default {
+    "aria-describedby": {
+        description: "The id of an element that describes the switch.",
+        table: {
+            category: "Accessibility",
+            type: {
+                summary: "string",
+            },
+        },
+    },
+    "aria-label": {
+        description:
+            "An accessible label for the switch. If a visible label is not provided, aria-label should be used to describe the purpose of the switch.",
+        table: {
+            category: "Accessibility",
+            type: {
+                summary: "string",
+            },
+        },
+    },
+    "aria-labelledby": {
+        description:
+            "The id of an element that is a label for the switch. Aria-labelledby takes precedence over all other methods of providing an accessible name.",
+        table: {
+            category: "Accessibility",
+            type: {
+                summary: "string",
+            },
+        },
+    },
+    checked: {
+        description: "Whether this component is checked.",
+        control: {
+            type: "boolean",
+        },
+    },
+    disabled: {
+        description: "Whether the switch is disabled. Defaults to `false`.",
+        control: {
+            type: "boolean",
+        },
+        table: {
+            type: {
+                summary: "boolean",
+            },
+        },
+    },
+    icon: {
+        description: "Optional icon to display on the slider.",
+        control: {
+            type: "select",
+            labels: {
+                null: "none",
+            },
+        },
+        options: [null, ...Object.keys(iconsMap)],
+        mapping: iconsMap,
+        table: {
+            type: {
+                summary: "Icon",
+            },
+        },
+    },
+    id: {
+        description:
+            "The unique identifier for the switch. If not provided, a unique id will be generated.",
+        control: {
+            type: "text",
+        },
+        table: {
+            type: {
+                summary: "string",
+            },
+        },
+    },
+    onChange: {
+        description: "Function to call when the switch is clicked.",
+        table: {
+            type: {
+                summary: "(newCheckedState: boolean) => unknown",
+            },
+        },
+    },
+    testId: {
+        description: "Optional test ID used for e2e testing.",
+        control: {
+            type: "text",
+        },
+        table: {
+            type: {
+                summary: "string",
+            },
+        },
+    },
+};
--- a/__docs__/wonder-blocks-switch/switch.stories.tsx
+++ b/__docs__/wonder-blocks-switch/switch.stories.tsx
@@ -1,0 +1,144 @@
+import * as React from "react";
+import type {Meta, StoryObj} from "@storybook/react";
+import {StyleSheet} from "aphrodite";
+import {expect} from "@storybook/jest";
+import {userEvent, within} from "@storybook/testing-library";
+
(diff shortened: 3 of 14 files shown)


packages/wonder-blocks-switch/src/components/switch.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+const StyledSpan = addStyle("span");
+
+const Switch = React.forwardRef(function Switch(
+    props: Props,
+    ref: React.ForwardedRef<HTMLButtonElement>,
+) {
+    const {ariaLabel, checked, disabled = false, id, onChange, testId} = props;
+
+    const renderSwitch = (
```
nishasy, line 54: Optimization: I think it's technically better to break this out into its own little component (still inside this file) because arrow functions are re-run on every render. (I only learned this recently, so please correct me if I'm wrong.)

packages/wonder-blocks-switch/src/components/switch.tsx
```diff
@@ -0,0 +1,224 @@
...
+                    !disabled &&
+                        (eventState?.hovered || eventState?.focused) &&
+                        sharedStyles.outline,
+                    disabled &&
+                        eventState?.focused &&
+                        sharedStyles.disabledOutline,
+                    !disabled && eventState?.pressed && stateStyles.pressed,
+                ]}
+                {...sharedProps}
+            >
```
nishasy, line 74: I think you might need to add an `aria-checked` here. Currently if you try to use a screen reader in your storybook examples, it doesn't read out the `checked` value of the switch.

marianmwang: hm... it should already be in the shared props. i was able to get it to read it out?
image

nishasy: Oh hm... I was trying it on the `Inside Cell` story. I was hoping it would work kind of like the [WithLabel](https://khan.github.io/wonder-blocks/?path=/story/form-checkbox--with-label) story on checkbox. This might be something that would need to be updated in `Cell` though, which would be out of scope for this PR.

packages/wonder-blocks-switch/src/components/switch.tsx
nishasy, line 54: Instead of having this be a function inside of the `Switch` definition, could you break this out into its own little component (still within this file) outside of `Switch`? This would help with optimization and be more in line with best practices, which you can read about [here](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unstable-nested-components.md).

packages/wonder-blocks-switch/src/components/switch.tsx
marianmwang: ooh. yea that would be something that needs to be specified on `Cell`. i'm not sure if we can enforce it from the `Switch` side of things but i'll add a note in the story about it.

__docs__/wonder-blocks-switch/switch.argtypes.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+const iconsMap: Record<string, React.ReactElement<typeof Icon>> = {};
+
+Object.entries(icons).forEach(([iconLabel, iconValue]) => {
+    iconsMap[iconLabel] = <Icon icon={iconValue} />;
+});
+
+export default {
+    ariaLabel: {
+        description: "A label for the switch.",
```
nishasy, line 13:
```suggestion
        description: "An accessible label for the switch.",
```

__docs__/wonder-blocks-switch/switch.argtypes.tsx
```diff
@@ -0,0 +1,85 @@
...
+        table: {
+            category: "Accessibility",
+            type: {
+                summary: "string",
+                detail: "If a visible label is not provided, aria-label should be used to describe the purpose of the switch.",
+            },
+        },
+    },
+    checked: {
+        description: "Whether this compoonent is checked.",
```
nishasy, line 23:
```suggestion
        description: "Whether this component is checked.",
```

__docs__/wonder-blocks-switch/switch.argtypes.tsx
```diff
@@ -0,0 +1,85 @@
...
+});
+
+export default {
+    ariaLabel: {
+        description: "A label for the switch.",
+        table: {
+            category: "Accessibility",
+            type: {
+                summary: "string",
+                detail: "If a visible label is not provided, aria-label should be used to describe the purpose of the switch.",
```
nishasy, line 18: **Optional suggestion:** I think I'd want this detail to be added to the description itself instead so it has more visibility.

__docs__/wonder-blocks-switch/switch.stories.tsx
```diff
@@ -0,0 +1,167 @@
...
+};
+
+WithIcon.parameters = {
+    docs: {
+        description: {
+            story: `The \`medium\` size switch can take an \`icon\` prop, which will be rendered inside the slider.`,
+        },
+    },
+};
+
```
nishasy, line 137: Do you think you could add a `WithLabel` story to demonstrate how this would be used with a `<label>`? Note that this would have to be updated after the `LabeledField` component is created.

consistency-tests/__tests__/ref-forwarded.test.tsx
```diff
@@ -355,6 +356,19 @@ describe("Form elements", () => {
         // Assert
         expect(ref.current).toBeInstanceOf(HTMLFieldSetElement);
     });
+
+    describe("Switch", () => {
+        test("forwards ref to an HTMLInputElement", () => {
```
nishasy, line 361: 🎉

__docs__/wonder-blocks-switch/switch.stories.tsx
```diff
@@ -0,0 +1,167 @@
...
+    return (
+        <CompactCell
+            title="Switch as a right accessory"
+            onClick={() => {
+                setChecked(!checked);
+            }}
+            rightAccessory={<Switch checked={checked} />}
+        />
+    );
+};
```
nishasy, line 150: Hmmmmmmmmm... This story still isn't behaving the way I'd expect with a screen reader. When you press on the cell, it doesn't read out the new checked value. I would expect it to work similarly to option items in a single select or multiselect dropdown.

Let me know if you'd like help investigating this!

packages/wonder-blocks-switch/src/components/__tests__/switch.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+                />,
+            );
+
+            // Act
+            const icon = screen.getByTestId("test-icon");
+
+            // Assert
+            expect(icon).toBeInTheDocument();
+        });
+    });
```
nishasy, line 88: Can you add a test to make sure that clicking on a `<label>` also toggles the switch?

packages/wonder-blocks-switch/src/components/switch.tsx
```diff
@@ -0,0 +1,215 @@
...
+                    >
+                        <StyledInput
+                            aria-checked={checked}
+                            aria-disabled={disabled}
+                            aria-label={ariaLabel}
+                            disabled={disabled}
+                            id={uniqueId}
+                            ref={ref}
+                            role="switch"
+                            style={[sharedStyles.hidden]}
```
nishasy, line 104:
```suggestion
                            style={sharedStyles.hidden}
```

nishasy: Could you also leave a comment pointing out that this input is visually hidden and explain why?

__docs__/wonder-blocks-switch/switch.argtypes.tsx
marianmwang: my keyboard has been acting up lately with these O's. i guess it made it through...

packages/wonder-blocks-switch/src/components/switch.tsx
```diff
@@ -0,0 +1,240 @@
...
+const styles: Record<string, any> = {};
+const _generateStyles = (
+    checked: boolean,
+    disabled: boolean,
+    clickable: boolean,
+) => {
+    const checkedStyle = `${checked}-${disabled}-${clickable}`;
+    if (sharedStyles[checkedStyle]) {
+        return sharedStyles[checkedStyle];
+    }
```
jandrade, line 179: **suggestion:** What do you think about adding a comment here mentioning that the return value in this function is cached to avoid regenerating a `StyleSheet.create` Aphrodite for every re-render?

jandrade (approved):
Looks great! I've left some suggestions before landing. This looks pretty solid! 🚀 

Side note: Probably some things will need to change in the style generation part when you include theming to this component, but this is a really good foundation for this component!

packages/wonder-blocks-switch/src/components/switch.tsx
```diff
@@ -0,0 +1,240 @@
...
+                        onClick={handleClick}
+                        style={[
+                            sharedStyles.switch,
+                            stateStyles.switch,
+                            disabled && sharedStyles.disabled,
+                        ]}
+                        testId={testId}
+                    >
+                        <StyledInput
+                            aria-checked={checked}
```
jandrade, line 102: **question:** Giving a second thought on this, maybe it's better to use `checked` directly as this is an input[checkbox] element? wdyt? (cc @nishasy)

__docs__/wonder-blocks-switch/switch.stories.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+import packageConfig from "../../packages/wonder-blocks-switch/package.json";
+import ComponentInfo from "../../.storybook/components/component-info";
+
+import SwitchArgtypes from "./switch.argtypes";
+
+type StoryComponentType = StoryObj<typeof Switch>;
+
+export default {
+    title: "Switch / Switch",
```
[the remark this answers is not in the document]
marianmwang: hmm... not sure if the behaviour is what was expected. it's still a folder :/
image but i think this is still better.

__docs__/wonder-blocks-switch/switch-best-practices.mdx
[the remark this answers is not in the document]
marianmwang: saw a new pattern on storybook docs! i def think it's a lot easier too write them this way. it says they've deprecated .mdx stories and will no longer support them later on but there's a migration script!

__docs__/wonder-blocks-switch/switch.stories.tsx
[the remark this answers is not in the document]
jandrade: ah yeah, that's because you have the other `Best Practices` section, but now the switch docs are rendered first, so it works better :)

__docs__/wonder-blocks-switch/switch.stories.tsx
marianmwang: update from dicussions offline: we decided the whole cell should _not_ be clickable, and it is best to pass in html `label` tag elements so that clicking them also toggles the switch. this way, the screen reader only focuses on the switch and the click area is also expanded to include the `label` element.

packages/wonder-blocks-switch/src/components/switch.tsx
marianmwang: good point! it seems since it's a native checkbox using `checked` is fine.
source: https://stackoverflow.com/questions/49179545/do-i-need-aria-checked-on-an-actual-checkbox#:~:text=The%20aria%20attributes%20are%20for,and%20other%20attributes%20as%20well).

PR #1962, opened 2023-08-02
Create skeleton for new switch package

marianmwang:
## Summary:
- Created a skeleton for the new wonder blocks Switch component

Issue: WB-1581

## Test plan:
Verify that checks pass
NOTE: Build size check will fail because there is no previous version of this package to comopare against

PR #1983, opened 2023-08-15
Add wonder-blocks-pill package

nishasy:
## Summary:
This PR introduces the new package for pill components: `wonder-blocks-pill`.

Issue: https://khanacademy.atlassian.net/browse/WB-1583

## Test plan:
`yarn test`


jeresig:
@nishasy is this supposed to still be a draft?

packages/wonder-blocks-pill/package.json
```diff
@@ -0,0 +1,30 @@
+{
+    "name": "@khanacademy/wonder-blocks-pill",
+    "version": "0.0.1",
+    "design": "v1",
+    "description": "Pill components for Wonder Blocks.",
```
jandrade, line 5: **suggestion:** This config should include the following settings to avoid releasing the newly created package along other packages. Of course, we'll switch to this version when `Pill` is ready for PROD (or `v1.0.0`)

```suggestion
    "description": "Pill components for Wonder Blocks.",
    "private": true,
    "publishConfig": {
        "access": "restricted"
    },
```

jandrade (changes requested):
This is looking good, but after migrating to TS, there are some extra files that we need to include as part of creating a new package:

- wonder-blocks-pill/tsconfig-build.json (to be able to build the required WB deps inside this package.
- Creating a symlink to the `types` folder: this can be done by doing: `cd packages/wonder-blocks-pill` && ln -s ../../types types`.
- Adding a reference to the pill `tsconfig-build.json` file in the global `tsconfig.build.json` file.

You can see a couple of examples here:
https://github.com/Khan/wonder-blocks/pull/1962/files
https://github.com/Khan/wonder-blocks/pull/1963/files

PR #1985, opened 2023-08-16
Fix destructured props in Choice Internal

nishasy:
After forwarding refs, the snapshots showed that the checkbox and
radio `input` elements had a random `variant` attribute that isn't
supposed to be there. This was an oversight on my part when I updated
these to be functional components and shifted around their props in #1956.

I'm fixing that here by making sure I don't pass in all the props when it should
just be a subset.

Issue: none

Test plan:
- Go to the Checkbox and Radio (Internal) stories in storybook.
- Use the dev tools to confirm that they don't have `variant` in their attributes.

### Before

image

### After

image

### DIFF
--- a/.changeset/khaki-pianos-brush.md
+++ b/.changeset/khaki-pianos-brush.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-form": patch
+---
+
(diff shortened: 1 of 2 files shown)


jandrade (approved):
LGTM! thanks for fixing this so quickly 👏

PR #1988, opened 2023-08-16
Make switch package public

marianmwang:
## Summary:
- Change `publishCofig` to `access: public` for new Switch component

Issue: N/A

## Test plan:
Make sure Github actions pass and verify package is published on next version release


jeresig (approved):
Yay!

.changeset/fuzzy-buttons-beg.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
jandrade, line 2: Just one minor thing to mention... as this becomes public, it would be a good idea to mark it as a `major` version. Probably can be done after you integrate theming into it :)

marianmwang: i actually marked it as a major verson in the previous PR so i think that should apply there!

PR #1989, opened 2023-08-17
[fei5242.2.wbsettings] Move build settings package to Khan namespace

somewhatabstract:
## Summary:
This renames the build settings package so that it sits in the Khan Academy namespace. This should prevent folks squatting the package on NPM, since it would require being able to publish to the Khan Academy NPM org.

Issue: FEI-5242

## Test plan:
Not much to test here but I did run `yarn build` to make sure it still works.


jeresig (approved):
Phew, excellent!

PR #1990, opened 2023-08-17
[fei5147.1.nodata] Add no-data state to useCachedEffect

somewhatabstract:
## Summary:
This introduces a new status to the `Result<TData>` type; `no-data`. This new result is returned when the effect is not waiting on a pending fetch, and there is no data in the cache.

I also addressed the suppressed ts errors in WB Data while I was doing this - most of them were very simple.

Issue: FEI-5147

## Test plan:
`yarn test`
`yarn typecheck`

### DIFF
--- a/packages/wonder-blocks-core/src/hooks/__tests__/use-is-mounted.test.ts
+++ b/packages/wonder-blocks-core/src/hooks/__tests__/use-is-mounted.test.ts
@@ -1,3 +1,5 @@
+// eslint-disable-next-line import/no-unassigned-import
+import "jest-extended";
 import {renderHook} from "@testing-library/react-hooks";
 
 import {useIsMounted} from "../use-is-mounted";
@@ -10,7 +12,6 @@ describe("useIsMounted", () => {
         const {result} = renderHook(useIsMounted);
 
         // Assert
-        // @ts-expect-error [FEI-5019] - TS2339 - Property 'toBeTrue' does not exist on type 'JestMatchers<boolean>'.
         expect(result.current()).toBeTrue();
     });
 
@@ -22,7 +23,6 @@ describe("useIsMounted", () => {
         rerender();
 
         // assert
-        // @ts-expect-error [FEI-5019] - TS2339 - Property 'toBeTrue' does not exist on type 'JestMatchers<boolean>'.
         expect(result.current()).toBeTrue();
     });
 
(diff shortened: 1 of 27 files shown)


packages/wonder-blocks-core/src/hooks/__tests__/use-is-mounted.test.ts
```diff
@@ -10,7 +12,6 @@ describe("useIsMounted", () => {
         const {result} = renderHook(useIsMounted);
 
         // Assert
-        // @ts-expect-error [FEI-5019] - TS2339 - Property 'toBeTrue' does not exist on type 'JestMatchers<boolean>'.
```
somewhatabstract, line 13: **note:** Importing jest-extended in any file during the typecheck caused it to be available for ALL files, so these suppressions were no longer needed. However, I added the `jest-extended` import anyway since that will ensure it works for folks editing files too where the global typechecking isn't going to be running.

packages/wonder-blocks-data/src/hooks/use-cached-effect.ts
```diff
@@ -265,29 +262,36 @@ export const useCachedEffect = <TData extends ValidCacheData>(
...
         };
     }, [shouldFetch, fetchRequest]);
 
     // We track the last result we returned in order to support the
-    // "retainResultOnChange" option.
-    const lastResultAgnosticOfIdRef = React.useRef(Status.loading());
+    // "retainResultOnChange" option. To begin, the last result is no-data.
+    const lastResultAgnosticOfIdRef = React.useRef<Result<TData>>(
+        Status.noData<TData>(),
+    );
```
somewhatabstract, line 274: **note:** Here we default our last result to the new `no-data` state.

packages/wonder-blocks-data/src/hooks/use-cached-effect.ts
```diff
@@ -265,29 +262,36 @@ export const useCachedEffect = <TData extends ValidCacheData>(
...
+    // - The no-data state if shouldFetch is false, and therefore there is no
+    //   in-flight request.
+    // - Otherwise, the loading state (we can assume there's an inflight
+    //   request if skip is not true).
     const loadingResult = retainResultOnChange
         ? lastResultAgnosticOfIdRef.current
-        : Status.loading();
+        : shouldFetch
+        ? Status.loading<TData>()
+        : Status.noData<TData>();
```
somewhatabstract, line 285: **note:** Here we default our "loadingResult" to no-data if we aren't fetching (which encompasses both the hard skip of the skip prop but also states where we won't fetch at all, such as the CacheOnly fetch policy).

jeresig (approved):
Phew... thanks for working through all of this! The end result seems ok to me.

PR #1991, opened 2023-08-17
[fei5235.1.ssradapter] Add SSR adapter for test harness

somewhatabstract:
## Summary:
This adds a new `ssr` adapter to the Wonder Blocks Testing package. It ensures a `RenderStateRoot` component is rendered around the component under test.

This adapter is default off since it may add the need for additional `act` or `waitFor` calls when using Testing Library due to the state changes that `RenderStateRoot` will perform.

In addition, default off ensures that updating to this new version won't break tests that were rendering their own `RenderStateRoot` component; nested `RenderStateRoot` components throw an error by default.

This change also addresses the TypeScript error suppressions in Wonder Blocks Testing as well as one in Wonder Blocks Core.

I also added a `yarn typewatch` command to package.json. This is a useful command to run in a separate terminal while developing to ensure that your changes don't break the type definitions, saves a few keystrokes, and is easier to remember/discover.

Issue: FEI-5235

## Test plan:
`yarn jest`
`yarn typecheck`

### DIFF
--- a/.changeset/blue-ways-cough.md
+++ b/.changeset/blue-ways-cough.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-testing": major
+---
+
+Added new SSR adapter for test harnesses to support `RenderStateRoot` in tests and stories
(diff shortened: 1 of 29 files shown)


jeresig (approved):
Great - thank you for adding this in, Jeff - the changes look good to me!

PR #1995, opened 2023-08-21
Fix typo in switch style check

marianmwang:
## Summary:
I realized that the styles check in switch was checking the wrong object. Corrected the typo

Issue: XXX-XXXX

## Test plan:
- Add console logs and make sure the styles are not being regenerated every render


jandrade (approved):
LGTM! goog catch :shipit:

nishasy (approved):
Great catch!

PR #2000, opened 2023-08-22
Allow 'beforeNav' and 'safeWithNav' to be used without an 'href'

kevinbarabash:
## Summary:
Looking at the implementation of ClickableBehavior, 'beforeNav' and 'safeWithNav' can be used without an 'href'.  This change to the types allows us to use 'Clicakble' in accordion-item.tsx in webapp without splitting its existing use in two.

Issue: None

## Test plan:
- yarn tsc
- let CI run tests
- copy the types over to webapp and check that we can collapse the uses of `Clickable` in https://github.com/Khan/webapp/pull/16140

### DIFF
--- a/packages/wonder-blocks-clickable/src/components/__tests__/clickable.typestest.tsx
+++ b/packages/wonder-blocks-clickable/src/components/__tests__/clickable.typestest.tsx
@@ -47,3 +47,18 @@ import Clickable from "../clickable";
 >
     {(_) => "Hello, world!"}
 </Clickable>;
+
+<Clickable beforeNav={() => Promise.resolve()}>
+    {(_) => "Hello, world!"}
+</Clickable>;
+
+<Clickable safeWithNav={() => Promise.resolve()}>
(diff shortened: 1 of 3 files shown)


jeresig (approved):

packages/wonder-blocks-clickable/src/components/__tests__/clickable.typestest.tsx
```diff
@@ -47,3 +47,18 @@ import Clickable from "../clickable";
...
+<Clickable safeWithNav={() => Promise.resolve()}>
+    {(_) => "Hello, world!"}
+</Clickable>;
+
+<Clickable
+    beforeNav={() => Promise.resolve()}
+    safeWithNav={() => Promise.resolve()}
+>
+    {(_) => "Hello, world!"}
+</Clickable>;
```
kevinbarabash, line 64: We should probably make either `href` or `onClick` required, but I've fiddled enough with the types for today and would like to move on to other things.

PR #2002, opened 2023-08-22
Add `themeName` to `useScopedTheme` hook

jandrade:
## Summary:

This PR adds the `themeName` to the `useScopedTheme` hook. This is useful for
when you want to use the theme name in a component that needs a the theme name
to be used as a key/cache key.

This is going to be useful for cases like `Button` where we need to cache the
`StyleSheet.create` call for generating dynamic styles.

e.g.
https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-button/src/components/button-core.tsx#L279-L283

Issue: XXX-XXXX

## Test plan:

Verify that the tests pass and the documentation is updated and clear.

/?path=/docs/theming-usescopedtheme--docs

Screenshot 2023-08-22 at 1 04 37 PM

PR #2005, opened 2023-08-23
WB-1577: Add theming support to Button component

jandrade:
## Summary:

This PR adds theming support to the Button component via the
`wonder-blocks-theming` package. It also creates a new `khanmigo` theme for the
Button component so it can be used within the Khanmigo widget.

Another change is that we include a new "Theme" toolbar option in the Storybook
UI to be able to switch between the default and khanmigo themes for any
story/component that supports theming.

Issue: WB-1577

## Test plan:

1. Verify that the new theme story works as expected in Storybook: `/?path=/story/button--khanmigo-theme`.

Screenshot 2023-08-23 at 5 23 31 PM

2. Also verify that the "Theme" toolbar option switches between themes.

https://github.com/Khan/wonder-blocks/assets/843075/6187c035-300b-4d31-86a4-378f2316c7a9


__docs__/wonder-blocks-button/button.stories.tsx
```diff
@@ -120,51 +122,49 @@ export const styles: StyleDeclaration = StyleSheet.create({
...
-                <Strut size={16} />
-                <Button onClick={() => {}} kind="secondary">
-                    Hello, world!
-                </Button>
-                <Strut size={16} />
-                <Button onClick={() => {}} kind="tertiary">
-                    Hello, world!
-                </Button>
-            </View>
+export const Variants: StoryComponentType = () => (
```
jandrade, line 125: **note:** Removed the `render` fn to be able to reuse the story in the `KhanmigoTheme` story at the end of this file. This change caused a lot of indentation changes.

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -228,17 +232,18 @@ const sharedStyles = StyleSheet.create({
...
         // The left padding for the button with icon should have 4px less padding
-        paddingLeft: 12,
+        paddingLeft: theme.padding.medium,
     },
     disabled: {
         cursor: "auto",
     },
     small: {
-        height: 32,
+        borderRadius: theme.border.radius.small,
```
jandrade, line 241: **note:** Added a borderRadius value here b/c the `khanmigo` theme overrides this value for `small` buttons.

packages/wonder-blocks-theming/src/tokens.ts
```diff
@@ -15,8 +15,12 @@ const tokens = {
         // New colors
         activeBlue: mix(Color.offBlack32, Color.blue),
         fadedBlue: mix(fade(Color.blue, 0.32), Color.white),
+        fadedBlue16: fade(Color.blue, 0.16),
```
jandrade, line 18: **note:** New tokens added while working on the button theming impl.

jandrade:
The parent pull-request ([#2002](https://github.com/Khan/wonder-blocks/pull/2002)) has been merged into `main`, but this branch (`theming-button2`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -392,12 +437,14 @@ const _generateStyles = (
                     // calculate the width/height and use absolute position to
                     // prevent other elements from being shifted around.
                     position: "absolute",
-                    width: `calc(100% + ${Spacing.xxxSmall_4}px)`,
-                    height: `calc(100% - ${Spacing.xxxSmall_4}px)`,
+                    width: `calc(100% + ${theme.border.width.focused * 2}px)`,
```
jeresig, line 440: Huh... why the `* 2`?

jeresig (approved):
Wow, this is very cool! Really like seeing all of this come together. It definitely seems like a bunch of work to make a component themeable the first time, but future themes will be much simpler...

packages/wonder-blocks-button/src/components/button-core.tsx
jeresig: Ahh, gotcha - ok! Yeah, a comment would be helpful.

somewhatabstract:
For theming, do we want Khanmigo branding stuff to be publicly released? I know when Wonder Blocks was made, there was some concern about our branding being publicly available, and theming affords us the ability to keep that private. Do we want to keep the Khanigo stuff private and use some other non-branding theme for demo purposes?

jandrade:
> For theming, do we want Khanmigo branding stuff to be publicly released? I know when Wonder Blocks was made, there was some concern about our branding being publicly available, and theming affords us the ability to keep that private. Do we want to keep the Khanigo stuff private and use some other non-branding theme for demo purposes?

That's a good point @somewhatabstract. We discussed about the possibility of allowing to create themes in the consumer repos, but decided to take this direction so we can still have control of the tokens people will use for new themes. This change is intentional so we can still own the consistency (while providing flexibility) from within the repo (cc @jeresig).

That said, this approach might change in the future, but we wanted to start from a more contained place before making bigger/wider changes.

jeresig:
@somewhatabstract I think the concerns with have with the khanmigo theme are the same for the main Khan Academy theme. Theoretically we'd have a generic non-KA theme (whatever that is) as the default that we ship, but that's not really something that's been pushed for from design, so I'm fine just keeping things as-is right now.

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -263,36 +268,55 @@ const sharedStyles = StyleSheet.create({
...
+            : theme.color.bg.action;
+
     const buttonType =
-        color + kind + light.toString() + iconWidth.toString() + size;
+        color +
+        kind +
+        light.toString() +
+        iconWidth.toString() +
+        size +
+        themeName;
```
marianmwang, line 297: nit: use a template literal for readability aka ``${color}-${kind}-...``

marianmwang (approved):
so excited for this! :)

PR #2007, opened 2023-08-23
[fei5235.2.fixadapterrendering] Render adapters in a more React-friendly manner

somewhatabstract:
## Summary:
I was updating webapp to use the new `ssr` adapter for rendering the `RenderStateRoot` component and I discovered it wasn't working right. The children were definitely getting rendered inside a `RenderStateRoot` but the behavior was all wrong, and they were getting the context value BEFORE the root component had provided one.

This appears to be down to how the adapters were rendered, since we render the innermost first with the children and then wrap each rendered child accordingly. This means the tree of components isn't built so much as a tree of rendered results.

This update changes that by rendering the adapters as components in a component tree. The `renderAdapters` function is replaced by a recursive `Adapt` component and a companion `Adapter` component. The behavior is such that it is compatible with previous usage, but it should fix the issue I was seeing in webapp.

Issue: FEI-5235

## Test plan:
`yarn test`
`yarn typecheck`


packages/wonder-blocks-testing/src/harness/adapt.tsx
```diff
@@ -0,0 +1,55 @@
...
+ * test harness.
+ */
+export const Adapt = <TAdapters extends TestHarnessAdapters>({
+    children,
+    adapters,
+    configs,
+}: Props<TAdapters>): React.ReactElement => {
+    // We start at the end of the adapter list and work backwards to be
+    // compatible with previous releases.
+    const thisAdapterName = Object.keys(adapters).at(-1);
```
kevinbarabash, line 28: `.at(-1)` so fancy 🙂

packages/wonder-blocks-testing/src/harness/adapt.tsx
```diff
@@ -0,0 +1,55 @@
...
+        // we must filter by adapter name here.
+        Object.entries(configs).filter(([name]) => name !== thisAdapterName),
+    );
+    return (
+        <Adapter adapter={thisAdapter} config={thisConfig}>
+            <Adapt adapters={restAdapters} configs={restConfigs}>
+                {children}
+            </Adapt>
+        </Adapter>
+    );
```
kevinbarabash, line 54: It's cool that this recursive approach works.

packages/wonder-blocks-testing/src/harness/adapter.tsx
```diff
@@ -0,0 +1,27 @@
...
+ */
+export const Adapter = <TConfig, TAdapter extends TestHarnessAdapter<TConfig>>({
+    children,
+    adapter,
+    config,
+}: Props<TConfig, TAdapter>): React.ReactElement => {
+    if (config == null) {
+        return <>{children}</>;
+    }
+    return adapter(children, config);
```
kevinbarabash, line 26: This probably isn't necessary, but it also isn't hurting having a component wrapper around `adapter()`.  No action req'd.

packages/wonder-blocks-testing/src/harness/make-test-harness.tsx
```diff
@@ -43,13 +43,11 @@ export const makeTestHarness = <TAdapters extends TestHarnessAdapters>(
...
-            renderAdapters<TAdapters>(
-                adapters,
-                fullConfig,
-                <Component {...(props as TProps)} ref={ref} />,
-            ),
-        );
+        const harnessedComponent = React.forwardRef((props: TProps, ref) => (
+            <Adapt adapters={adapters} configs={fullConfig}>
+                <Component {...(props as TProps)} ref={ref} />
+            </Adapt>
```
kevinbarabash, line 49: **praise:** this is much cleaner.

kevinbarabash (approved):
LGTM. 🎉  Nice tests. Thanks for taking the time to refactor the API while fixing this issue. 💚

packages/wonder-blocks-testing/src/harness/adapter.tsx
somewhatabstract: I think it might be, since rendering it as a component changes the order of invocation of the render function itself.

PR #2012, opened 2023-08-24
[🔥AUDIT🔥] [fixtimingstories] The examples use React but that isn't a globally available thing, we have to import it

somewhatabstract:
🖍 _This is an audit!_ 🖍

## Summary:
We were seeing errors in the docs for these examples. It turns out they were trying to use React without importing it.

Issue: XXX-XXXX

## Test plan:
`yarn start` and view the various timing hooks docs to see they work rather than showing a stacktrace.

### DIFF
--- a/__docs__/wonder-blocks-timing/use-interval.stories.mdx
+++ b/__docs__/wonder-blocks-timing/use-interval.stories.mdx
@@ -1,3 +1,4 @@
+import * as React from "react";
 import {Meta, Story, Source, Canvas} from "@storybook/blocks";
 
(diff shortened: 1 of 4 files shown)


__docs__/wonder-blocks-timing/use-interval.stories.mdx
```diff
@@ -59,7 +60,7 @@ export const BasicUsage = () => {
     </Story>
 </Canvas>
 
-```jsx
+```tsx
```
somewhatabstract, line 63: **note:** Leftover from me tinkering to determine what the error could be.

PR #2016, opened 2023-08-24
Provide better debug component name for withActionScheduler HoC

jeremywiebe:
## Summary:

When a component is wrapped with `withActionScheduler`, the React dev tools aren't able to derive a meaningful name for the wrapping component. This PR uses a [trick](https://github.com/remix-run/react-router/blob/7a891467987b164b64f39eaf72415b609ed2d4a7/packages/react-router/modules/withRouter.js) I learned from `react-router` to provide a friendly name for the wrapping component. 

Issue: "none"

## Test plan:

Run an app that uses `withActionScheduler`
Inspect a page using the React dev tools and note that the component created by `withActionScheduler` is named `withActionSchedule(SomeComponent)`.

### DIFF
--- a/packages/wonder-blocks-timing/src/components/with-action-scheduler.tsx
+++ b/packages/wonder-blocks-timing/src/components/with-action-scheduler.tsx
@@ -18,11 +18,20 @@ type WithoutActionScheduler<T> = Omit<T, "schedule">;
 export default function withActionScheduler<
     Props extends WithActionSchedulerProps,
 >(WrappedComponent: React.ComponentType<Props>) {
-    return (props: WithoutActionScheduler<Props>) => (
+    const displayName = `withActionScheduler(${
+        WrappedComponent.displayName || WrappedComponent.name
+    })`;
+
+    const C = (props: WithoutActionScheduler<Props>) => (
         <ActionSchedulerProvider>
             {(schedule: IScheduleActions) => (
                 <WrappedComponent {...(props as Props)} schedule={schedule} />
             )}
         </ActionSchedulerProvider>
     );
+
+    C.displayName = displayName;
+    C.WrappedComponent = WrappedComponent;
+
+    return C;
 }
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-timing/src/components/with-action-scheduler.tsx
```diff
@@ -18,11 +18,20 @@ type WithoutActionScheduler<T> = Omit<T, "schedule">;
...
+    const C = (props: WithoutActionScheduler<Props>) => (
         <ActionSchedulerProvider>
             {(schedule: IScheduleActions) => (
                 <WrappedComponent {...(props as Props)} schedule={schedule} />
             )}
         </ActionSchedulerProvider>
     );
+
+    C.displayName = displayName;
+    C.WrappedComponent = WrappedComponent;
```
somewhatabstract, line 34: **question:** Is exposing this bit necessary?

jeremywiebe: Possibly not. It seems to be part of the pattern for HoCs in fairly core React libs (Redux, for example, [does this also](https://github.com/reduxjs/react-redux/blob/4a7e129ee537d35c53da258ccf7924a0376391ee/src/components/connect.tsx#L785)).

However, given I'm unsure what it does, I'm happy to remove this and just set the `.displayName`

somewhatabstract: We use the `.displayName` part in our code a lot. Not seen the `WrappedComponent` piece; I wonder if that's there for their debug tools or something.

jeremywiebe: Removed for now. If we find out it helps, we can always do another PR.

kevinbarabash (approved):
LGTM.  Thanks for specifying a `displayName` on this HOC. 🚢

PR #2018, opened 2023-08-25
[fei5235.4.cacheadaptercomponents] Cache adapter components so we reuse the same one rather than causing remounting by auto-generating them

somewhatabstract:
## Summary:
My change to add a name to each adapter component broke our testing because it meant that every render of the harness would cause things to remount unnecessarily.

This works around that problem by caching the named components so they can be reused across renders. This works great since only the name changes; everything else is done at render time, allowing even different adapters with the same name to share a cached component.

I also added a test case to capture the non-cached situation to show that things are not remounting in this new implementation (comment out the cache retrieval step in `getNamedAdapterComponent` to see the test fail).

Issue: FEI-5235

## Test plan:
`yarn test`
`yarn typecheck`

Putting this into my PR for adding the new `ssr` test harness adapter to webapp will show that it works. Currently, with the old "remounting" bug, there were a number of test failures. Now, with this change, they should just pass.

### DIFF
--- a/packages/wonder-blocks-testing/src/harness/get-named-adapter-component.tsx
+++ b/packages/wonder-blocks-testing/src/harness/get-named-adapter-component.tsx
@@ -1,0 +1,36 @@
+import * as React from "react";
+
+import type {TestHarnessAdapter} from "./types";
+
+type Props<TConfig = any> = {
+    children: React.ReactNode;
+    config: TConfig;
+    adapter: TestHarnessAdapter<TConfig>;
+};
+
+const componentCache = new Map<string, React.FunctionComponent<Props>>();
+
+/**
+ * Get a component tagged with the given name for rendering an adapter.
+ *
+ * We can share these across invocations because only the name is used.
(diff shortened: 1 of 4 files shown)


jeresig (commented):
Wow, this is a tricky edge case! Thanks for fixing this.

packages/wonder-blocks-testing/src/harness/get-named-adapter-component.tsx
```diff
@@ -0,0 +1,36 @@
...
+        return existing;
+    }
+
+    // Otherwise, create a new component, name it, cache it, and return it.
+    const newComponent: React.FunctionComponent<Props> = ({
+        children,
+        config,
+        adapter,
+    }: any) => adapter(children, config);
+    newComponent.displayName = `Adapter(${name})`;
```
kevinbarabash, line 33: Nice touch!

kevinbarabash (approved):
LGTM.  Nice test case for verifying the solution! 🎉

PR #2020, opened 2023-08-25
Fix the return type of useLatestRef

benchristel:
## Summary:
React defines the RefObject type as:

```
interface RefObject<T> {
  readonly current: T | null;
}
```

Previously, the `useLatestRef` hook was typed to return `RefObject<T>`,
but that wasn't really correct. The `current` value returned by
`useLatestRef` cannot be null if `T` is not nullable.

This commit changes the return type to `{readonly current: T}`.

Issue: none

## Test plan:

Run `yarn start` to launch Storybook.

View the documentation for `useLatestRef`. The return type should be `React.MutableRefObject<T>`.

`yarn lint`, `yarn test`, and `yarn typecheck` should pass.

### DIFF
--- a/packages/wonder-blocks-core/src/hooks/__tests__/use-latest-ref.test.ts
+++ b/packages/wonder-blocks-core/src/hooks/__tests__/use-latest-ref.test.ts
@@ -37,4 +37,15 @@ describe("useLatestRef", () => {
         // Assert
         expect(result.current).toBe(refFromFirstRender);
     });
+
+    // Type tests
+    const _ = () => {
+        // The correct, non-nullable type should be inferred for `current`.
+        useLatestRef(123) satisfies {current: number};
+        // The return value should be assignable to React's ref types.
+        useLatestRef(123) satisfies React.RefObject<number>;
+        useLatestRef(123) satisfies React.MutableRefObject<number>;
+        // @ts-expect-error the result should not be assignable to the wrong type
+        useLatestRef(123) satisfies {current: string};
+    };
 });
--- a/packages/wonder-blocks-core/src/hooks/__tests__/use-latest-ref.test.ts
+++ b/packages/wonder-blocks-core/src/hooks/__tests__/use-latest-ref.test.ts
@@ -37,15 +37,4 @@ describe("useLatestRef", () => {
         // Assert
         expect(result.current).toBe(refFromFirstRender);
     });
-
-    // Type tests
-    const _ = () => {
-        // The correct, non-nullable type should be inferred for `current`.
-        useLatestRef(123) satisfies {current: number};
-        // The return value should be assignable to React's ref types.
-        useLatestRef(123) satisfies React.RefObject<number>;
-        useLatestRef(123) satisfies React.MutableRefObject<number>;
-        // @ts-expect-error the result should not be assignable to the wrong type
-        useLatestRef(123) satisfies {current: string};
-    };
 });
--- a/packages/wonder-blocks-core/src/hooks/__tests__/use-latest-ref.typestest.ts
+++ b/packages/wonder-blocks-core/src/hooks/__tests__/use-latest-ref.typestest.ts
@@ -1,21 +1,21 @@
+// Suppress the linter so we can call hooks at the top level.
+/* eslint-disable react-hooks/rules-of-hooks */
 import * as React from "react";
 import {useLatestRef} from "../use-latest-ref";
 
-const _ = () => {
-    // The correct, non-nullable type should be inferred for `current`.
-    useLatestRef(123) satisfies {current: number};
+// The correct, non-nullable type should be inferred for `current`.
+useLatestRef(123) satisfies {current: number};
 
-    // The return value should be assignable to React's ref types.
-    useLatestRef(123) satisfies React.RefObject<number>;
-    useLatestRef(123) satisfies React.MutableRefObject<number>;
+// The return value should be assignable to React's ref types.
+useLatestRef(123) satisfies React.RefObject<number>;
+useLatestRef(123) satisfies React.MutableRefObject<number>;
 
-    // @ts-expect-error the result should not be assignable to the wrong type
-    useLatestRef(123) satisfies {current: string};
+// @ts-expect-error the result should not be assignable to the wrong type
+useLatestRef(123) satisfies {current: string};
 
-    {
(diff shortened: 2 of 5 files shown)


packages/wonder-blocks-core/src/hooks/use-latest-ref.ts
```diff
@@ -10,7 +10,7 @@ import * as React from "react";
  * be passed to useEffect without causing unnecessary re-renders. See the
  * Storybook docs for a usage example.
  */
-export function useLatestRef<T>(value: T): React.RefObject<T> {
+export function useLatestRef<T>(value: T): React.MutableRefObject<T> {
```
somewhatabstract, line 13: **question:** Do you want folks to be able to set current outside the ref? Should this be our own type it returns that meets your requirements without making the return value mutable?

benchristel: Yeah, I think using a custom type `{readonly current: T}` would be good. Note that `readonly` doesn't actually give you much protection because readonly types are assignable to mutable types! See: https://www.typescriptlang.org/play?#code/C4TwDgpgBAShCGATA9gOwDYjgMwPICMArCAY2AB4AVAPigF4oBvAJwRQxChIFdnXVgALiiUAvgChQkKAFluwePnQQcBYmSq0GjHnwgDhY8eOzdUZAJZooAW3gBrFWzSYcmgBS7+QkQEphcEguWBB4RKQUNEziULFQrMC8qExe+sASEiZmltZ2jnIKSk7YHqkGfsIFisqq4RpRjDFxCUkpvN4ZxiRoAM7AUPCV8tXFahHkqNw2+BDMWrYOTkEcOO4AjABMAMy+4t2ofVD4Ac4roWMak9Oz83kQVUWrmzviQA

packages/wonder-blocks-core/src/hooks/__tests__/use-latest-ref.test.ts
```diff
@@ -37,4 +37,12 @@ describe("useLatestRef", () => {
...
         expect(result.current).toBe(refFromFirstRender);
     });
+
+    // Type tests
+    const _ = () => {
+        // The correct, non-nullable type should be inferred for `current`.
+        useLatestRef(123) satisfies {current: number};
+        // @ts-expect-error the result should not be assignable to the wrong type
+        useLatestRef(123) satisfies {current: string};
+    };
```
benchristel, line 47: Is the convention to put type tests in a `*.typestest.ts` file? I see a couple such files. @somewhatabstract

packages/wonder-blocks-core/src/hooks/use-latest-ref.ts
somewhatabstract: That isn't a fair example. After all, it's creating a new object in each case and `readonly` only governs the property it is declared on, not the value. You would need `as const` to cover the value too.

By creating a new object, you remove the readonly-ness.

packages/wonder-blocks-core/src/hooks/__tests__/use-latest-ref.test.ts
somewhatabstract: Yes, that's the convention we've used for types-only testing.

packages/wonder-blocks-core/src/hooks/use-latest-ref.ts
benchristel: Is this a better example? https://www.typescriptlang.org/play?#code/C4TwDgpgBAShCGATA9gOwDYjgMwPICMArCAY2AB4AVAPigF4oBvAJwRQxChIFdnXVgALiiUAvgChQkKAFluwePnQQcBYmSq0GjHnwgDhY8eOzdUZAJZooAW3gBrFWzSYcmgBS7+QkQEphcEguWBB4RKQUNEziULFQrMC8qExe+sASEuLKwPFsuBwBzhyq4Rqo3Db4EMxatg5OQcWh7gCMAEwAzL7GAPQ9UAACwADOALQQAB6QZON8yMzifbLy8MAWqADmUPCwRa6hahFQFsNQqMg58OjoyADuEIgAdOKsSPmYj6kC9FAALACsADZev0ABJ3CAAN2qj1hi36IGQ3C48GS8GGwwsGzRu0a+zC6hywGQ2ygkPgzAsimUUCkD22pzkCiUTgJEVhz2ythWLOETOprMOZQqVRqP1eiHeIHh21QiFpAAt9FBEdwAOSsbbXCHymwrYDQCzAACE4j1zOUn143h+LSAA

somewhatabstract: Yes, that's much better and really shows the issue. It's a shame that TypeScript doesn't handle better the assignment of readonly objects to mutable equivalents. I think even wrapping with `ReadOnly<>` doesn't help that. Probably need to exercise the muscle of using `Object.freeze` on things to enforce readonliness at runtime to help mitigate.

packages/wonder-blocks-core/src/hooks/__tests__/use-latest-ref.typestest.ts
```diff
@@ -1,18 +1,21 @@
+import * as React from "react";
 import {useLatestRef} from "../use-latest-ref";
 
-// The correct, non-nullable type should be inferred for `current`.
-useLatestRef(123) satisfies {current: number};
+const _ = () => {
```
somewhatabstract, line 4: **praise:** Nice way to avoid the linter; though I think I would've just disabled it for the whole file.

packages/wonder-blocks-core/src/hooks/use-latest-ref.ts
```diff
@@ -10,7 +10,7 @@ import * as React from "react";
  * be passed to useEffect without causing unnecessary re-renders. See the
  * Storybook docs for a usage example.
  */
-export function useLatestRef<T>(value: T): React.RefObject<T> {
+export function useLatestRef<T>(value: T): {readonly current: T} {
```
somewhatabstract, line 13: **suggestion:** Could do `Readonly<React.MutableRefObject<T>>`? :shrug: If you change it, don't forget to update the docs.

somewhatabstract (approved):
LGTM

Some inline comments, but nothing that I feel strongly about. Your call :)

PR #2023, opened 2023-08-28
Allow 'title' as a prop on Clickable

kevinbarabash:
## Summary:
We're passing 'title' to Clickable in a number of places in webapp.  When I havemore time I'll fix this in a comprehensive way so that we aren't playing whack-a-mole.

Issue: None

## Test plan:
- n/a, yolo

### DIFF
--- a/packages/wonder-blocks-clickable/src/components/clickable.tsx
+++ b/packages/wonder-blocks-clickable/src/components/clickable.tsx
@@ -91,6 +91,7 @@ type CommonProps =
          * Set the tabindex attribute on the rendered element.
          */
         tabIndex?: number;
+        title?: string;
         /**
          * Run async code before navigating. If the promise returned rejects then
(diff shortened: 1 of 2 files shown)


kevinbarabash:
This change shouldn't impact snapshots in Chromatic.

packages/wonder-blocks-clickable/src/components/clickable.tsx
```diff
@@ -91,6 +91,7 @@ type CommonProps =
          * Set the tabindex attribute on the rendered element.
          */
         tabIndex?: number;
+        title?: string;
```
somewhatabstract, line 94: **suggestion:** Can you please include doc comments on the prop?

somewhatabstract (approved):
LGTM, but needs a quick edit before landing. See inline.

PR #2025, opened 2023-08-28
[fei5275.1.useismountedlifecyclechange] Update useIsMounted to use useLayoutEffect

somewhatabstract:
## Summary:
This updates `useIsMounted` effect to use `useLayoutEffect` so that the mounted state is setup before any child components run their effects. This ensures that if those child effects cause parent event handlers or other state changing code to be invoked, the mounted state won't improperly block guarded state updates.

This also introduces `useIsomorphicLayoutEffect` to support this feature so that we can use `useLayoutEffect` in our codebase without worrying about SSR errors.

I have also updated the eslint configuration to ensure that uses of `useIsomorphicLayoutEffect` have exhaustive dependency checks enabled.

Issue: FEI-5275

## Test plan:
`yarn test`
`yarn typecheck`


packages/wonder-blocks-core/src/hooks/use-isomorphic-layout-effect.ts
```diff
@@ -0,0 +1,34 @@
...
+import Server from "../util/server";
+
+/**
+ * Hook that works like useLayoutEffect but does not error on the server.
+ *
+ * Effects from `useLayoutEffect` cause errors on the server. Rather than
+ * erroring, we want the server side to just silently no-op like `useEffect`
+ * calls do. This hook allows that.
+ */
+export const useIsomorphicLayoutEffect: typeof React.useLayoutEffect = (
```
somewhatabstract, line 12: **note:** This is based on the hook of the same name in webapp except:

1. It does not use `React.useEffect` when server-side; it just invokes a no-op.
2. It does not decide which call to use on import, but instead on first invocation - this is to simplify testing, since things that vary on import require module isolation and what have you when trying to test CSR vs SSR scenarios. It uses a ref so that the decision is only made once, on initial invocation.

packages/wonder-blocks-core/src/hooks/__tests__/use-is-mounted.test.ts
```diff
@@ -1,39 +0,0 @@
...
-describe("useIsMounted", () => {
-    it("should return false on first call", () => {
-        // Arrange
-
-        // Act
-        const {result} = renderHook(useIsMounted);
-
-        // Assert
-        expect(result.current()).toBeTrue();
-    });
```
somewhatabstract, line 16: **note:** This test case was not doing what was intended. Note that the assertion is different from the test case title; it was mounting the component rather than testing initial render.

.changeset/seven-zebras-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-core": major
```
somewhatabstract, line 2: **note:** Since this changes the behavior of the `useIsMounted` hook, I thought a `major` release was appropriate. I could be convinced that this is `minor` though if folks disagree 😄

__docs__/wonder-blocks-core/exports.use-isomorphic-layout-effect.stories.mdx
```diff
@@ -0,0 +1,37 @@
...
+{
+    "rules": {
+        "react-hooks/exhaustive-deps": [
+            "error",
+            {
+                "additionalHooks": "(useIsomorphicLayoutEffect)"
+            }
+        ]
+    }
+}
```
kevinbarabash, line 36: **praise:** Cool!

packages/wonder-blocks-core/src/hooks/use-is-mounted.ts
```diff
@@ -1,22 +1,28 @@
 import * as React from "react";
 
-import {useOnMountEffect} from "./use-on-mount-effect";
```
kevinbarabash, line 3: **question:** should `useOnMounted` be updated to use `useIsomorphicLayoutEffect` internally?

packages/wonder-blocks-core/src/hooks/__tests__/use-isomorphic-layout-effect.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+                useIsomorphicLayoutEffect(() => {}, []);
+                return null;
+            };
+
+            // Act
+            const underTest = () => renderToString(<Component />);
+
+            // Assert
+            expect(underTest).not.toThrowError();
+        });
```
kevinbarabash, line 55: **question:** would `useLayoutEffect()` throw in the tests if `Server.isServerSide()` returns `true`?  Maybe we should check that `useLayoutEffect()` isn't being called in this scenario.

packages/wonder-blocks-core/src/hooks/__tests__/use-is-mounted.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+                const isMounted = useIsMounted();
+                return isMounted() ? <>MOUNTED</> : <>UNMOUNTED</>;
+            };
+
+            // Act
+            const result = renderToString(<Component />);
+
+            // Assert
+            expect(result).toBe("UNMOUNTED");
+        });
```
kevinbarabash, line 45: **suggestion (non-blocking):** you could also use `result.all` with `renderHook()` to check all the values that were returned from the hook.

kevinbarabash (approved):
LGTM 🎉  I really like the introduction of `useIsomorphicLayoutEffect`.  Thanks for fixing the issues with `useIsMounted()`. 💚

packages/wonder-blocks-core/src/hooks/use-is-mounted.ts
somewhatabstract: No, I don't think so. I think that it should still use `useEffect`. That's the path of least surprise and what most folks are expecting when they use it, in terms of order of execution.

packages/wonder-blocks-core/src/hooks/__tests__/use-isomorphic-layout-effect.test.tsx
somewhatabstract: It does give an error in those scenarios and I feel like adding a test for that is tautological in that we are adding this hook because it causes problems. So to then test that it causes problems seems odd to me.

packages/wonder-blocks-core/src/hooks/__tests__/use-is-mounted.test.tsx
somewhatabstract: Hmm, I could, but this test case feels more succinct in that we know for sure this is the initial render we want.

packages/wonder-blocks-core/src/hooks/__tests__/use-isomorphic-layout-effect.test.tsx
somewhatabstract: OK, so the production version of React throws an error. The development version logs an error instead. I can test for the latter, not the former. Updating accordingly.

packages/wonder-blocks-core/src/hooks/__tests__/use-is-mounted.test.tsx
somewhatabstract: I tried `result.all` but that only had one value and that returned `true`. I don't think that Testing Library can represent that very first, pre-mount server/hydration render.

PR #2028, opened 2023-08-30
[gei5275.2.ismountedwasfine] Put useIsMounted back the way it was but the updated tests

somewhatabstract:
## Summary:
After making my earlier change and having a general discussion with @kevinbarabash about the complexity of hooks in general and `useLayoutEffect` in specific, I realised some more things about `useLayoutEffect` that changed my mind.

The error thrown during SSR is a proxy for the fact that useLayoutEffect runs before hydration on the client - that’s where the real issue lies; the client could do something to change the output of the pre-hydration render (the initial render) that would lead to hydration errors.

Having `isMounted` return true in that initial render cycle is definitely not what we want.
So, I have reverted that change (while keeping the updated test cases), and also renamed `useIsomorphicLayoutEffect` to `usePreHydrationEffect` to make it clear that it is only for use in the pre-hydration render cycle. I have also updated its documentation (in code and in docs) to make it clear the dangers of using it, when to use it, and why folks should use alternatives (like `useLayoutEffect`) unless they know for sure that their use case is a safe use of `usePreHydrationEffect`.

Issue: FEI-5275

## Test plan:
`yarn test`
`yarn typecheck`


somewhatabstract:
Intrigued to see how the changeset actions handle this since I have deleted a changeset from an earlier PR and reworded another that already existed, but have not added a new one. I'm hoping it just does the right thing and updates the release notes properly, removing the one that's no longer there and updating the other. 🤞🏻

packages/wonder-blocks-core/src/hooks/__tests__/use-is-mounted.test.tsx
```diff
@@ -24,24 +25,22 @@ describe("useIsMounted", () => {
...
-                // use `renderToString` from `react-dom/server`, but that is
-                // going to log an error because we use `useLayoutEffect`.
-                // So, we suppress that error for our testing purposes.
-                /* no-op */
-            });
+            // We cannot use renderHook on useIsMounted directly here and then
+            // look at the result.all[0] to see the first render as the result
+            // is a function that accesses the underlying ref, and that ref will
+            // have changed to `true` even if it was `false` in the initial
+            // render. So, we use a component.
```
somewhatabstract, line 32: **note:** @kevinbarabash, I realised today why `result.all` would not work for checking the initial render. This new test follows the spirit of your original suggestion though :)

kevinbarabash (approved):
LGTM. 🎉 Thanks for continuing to ruminate on this.

packages/wonder-blocks-core/src/hooks/__tests__/use-is-mounted.test.tsx
somewhatabstract: I think another approach here would likely be to have a hook in the test that wraps the `useIsMounted` hook and returns the result of `isMounted` as its return value, then render that hook. That way the `result.all` approach would work, I think.

PR #2029, opened 2023-08-30
Add Khanmigo theme to Switch

marianmwang:
## Summary:
- Fix typo and refactor parts of Switch
- Added new color tokens for khanmigo
- Add theme files and theming to switch

image

Issue: WB-1600

## Test plan:
- Create new story for switch style and update snapshots

### DIFF
--- a/__docs__/wonder-blocks-switch/switch.stories.tsx
+++ b/__docs__/wonder-blocks-switch/switch.stories.tsx
@@ -28,6 +28,13 @@ export default {
             />
         ),
     },
+    decorators: [
+        (Story) => (
+            <RenderStateRoot>
+                <Story />
+            </RenderStateRoot>
+        ),
+    ],
     argTypes: SwitchArgtypes,
 } as Meta<typeof Switch>;
 
--- a/__docs__/wonder-blocks-switch/switch.stories.tsx
+++ b/__docs__/wonder-blocks-switch/switch.stories.tsx
@@ -5,10 +5,10 @@ import {expect} from "@storybook/jest";
 import {userEvent, within} from "@storybook/testing-library";
 
 import Switch from "@khanacademy/wonder-blocks-switch";
-import {View} from "@khanacademy/wonder-blocks-core";
-import {Strut} from "@khanacademy/wonder-blocks-layout";
-import Spacing from "@khanacademy/wonder-blocks-spacing";
+import {RenderStateRoot, View} from "@khanacademy/wonder-blocks-core";
 import Icon, {icons} from "@khanacademy/wonder-blocks-icon";
+import {ThemeSwitcherContext, tokens} from "@khanacademy/wonder-blocks-theming";
+import {LabelMedium} from "@khanacademy/wonder-blocks-typography";
 
 import packageConfig from "../../packages/wonder-blocks-switch/package.json";
 import ComponentInfo from "../../.storybook/components/component-info";
@@ -52,7 +52,7 @@ export const Controlled: StoryComponentType = () => {
     return (
         <View style={styles.column}>
             <Switch checked={checkedOne} onChange={setCheckedOne} />
-            <Strut size={Spacing.xSmall_8} />
+
             <Switch
                 testId="test-switch"
                 aria-label="test switch"
@@ -100,15 +100,12 @@ Controlled.parameters = {
 export const Disabled: StoryComponentType = () => (
     <View style={styles.column}>
         <Switch checked={false} disabled={true} />
-        <Strut size={Spacing.xSmall_8} />
         <Switch checked={true} disabled={true} />
-        <Strut size={Spacing.xSmall_8} />
         <Switch
             checked={false}
             disabled={true}
             icon={<Icon icon={icons.search} />}
         />
-        <Strut size={Spacing.xSmall_8} />
         <Switch
             checked={true}
             disabled={true}
@@ -129,7 +126,7 @@ export const WithIcon: StoryComponentType = () => {
     return (
         <View style={styles.column}>
             <Switch checked={false} icon={<Icon icon={icons.search} />} />
-            <Strut size={Spacing.xSmall_8} />
+
             <Switch checked={true} icon={<Icon icon={icons.search} />} />
         </View>
     );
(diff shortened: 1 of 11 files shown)


packages/wonder-blocks-color/src/index.ts
```diff
@@ -48,12 +51,16 @@ const Color: ColorType = {
     white,
     white64: fade(white, 0.64),
     white50: fade(white, 0.5),
+    white32: fade(white, 0.32),
```
jandrade, line 54: **suggestion:** could you please move this to `tokens.ts`? The mid-term plan is to get rid of this `color` package and instead rely directly on the `theming` package for colors and spacing (and any other tokens).

https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-theming/src/tokens.ts#L15

packages/wonder-blocks-color/src/index.ts
```diff
@@ -48,12 +51,16 @@ const Color: ColorType = {
...
 
     // Brand
     darkBlue: "#0b2149",
     teal: "#14bf96",
     lightBlue: "#37c5fd",
     pink: "#fa50ae",
+
+    // Khanmigo
+    eggplant: "#5f1e5c",
```
jandrade, line 63: Same suggestion this about moving this to `tokens.ts`.

packages/wonder-blocks-switch/src/themes/default.ts
```diff
@@ -0,0 +1,66 @@
+import {tokens} from "@khanacademy/wonder-blocks-theming";
+
+const theme = {
+    color: {
+        bg: {
+            primary: {
```
jandrade, line 6: **suggestion:** I think the `primary/secondary/etc` category doesn't apply here.

For context, in `Button`, I used these terms b/c it was descriptive for the `kind` prop which is used by the same component (`<button/>`).

As I understand (please correct me if I'm wrong), in switch, these variants are used for different elements/surfaces within the component (`switch`, `slider`, `icon`).

My recommendation is that if we go with a more semantic approach (from the component-level perspective) is to provide names that align more to the structure of the component:

e.g.
- primary => `color.bg.switch`.
- secondary => `color.bg.slider`
- icon => `color.bg.icon`

You can read more about that concept here:
https://spectrum.adobe.com/page/design-tokens/#Use-component-specific-tokens-for-their-respective-component
https://uxdesign.cc/design-tokens-cheatsheet-927fc1404099#:~:text=Anatomy%20of%20a%20Design%20Token%20and%20Organization (see the surfaces part in the image).

packages/wonder-blocks-switch/src/themes/default.ts
```diff
@@ -0,0 +1,66 @@
...
+                disabledOff: tokens.color.offBlack32,
+                activeOff: tokens.color.offBlack64,
+                on: tokens.color.blue,
+                disabledOn: tokens.color.fadedBlue,
+                activeOn: tokens.color.activeBlue,
+            },
+            secondary: {
+                on: tokens.color.white,
+                off: tokens.color.white,
+            },
```
jandrade, line 17: **praise:** This is a great use of adding an extra level, as it helps to separate elements within the component! Also love the distinction between different states (so descriptive).

packages/wonder-blocks-switch/src/themes/default.ts
```diff
@@ -0,0 +1,66 @@
...
+        height: {
+            none: 0,
+            // switch
+            medium: 20,
+            // slider
+            large: tokens.spacing.large_24,
+        },
+        width: {
+            none: 0,
+            xxxxSmall: tokens.spacing.xxxxSmall_2,
```
jandrade, line 47: **suggestion:** It would be better to try to dissociate the value from the token naming.

Instead of `xxxxSmall: tokens.spacing.xxxxSmall_2`, this could be something like `small: tokens.spacing.xxxxSmall_2`.

**Why:**
Using this approach would help to have a more flexible naming in case we want to provide a different value in a different theme.

For example, in the future, we want to provide a different `width.small` value to the `khanmigo` theme (e.g. `8px`), so instead of enforcing this to a `xxxxSmall` size, we can group this into a more flexible `small` size.

packages/wonder-blocks-switch/src/themes/default.ts
```diff
@@ -0,0 +1,66 @@
...
+            // NOTE: This token is specific to the Switch component
+            // switch
+            medium: 20,
+            // NOTE: This token is specific to the Switch component
+            // slider
+            xLarge: 40,
```
jandrade, line 53: **suggestion/nit:** Following up on the `width` names...

In this case, these component tokens could be defined in a scalar way: `small`, `medium`, `large`, so there are no gaps in the t-shirt size and feels more linear. Wdyt?

```suggestion
            small: tokens.spacing.xxxxSmall_2,
            // NOTE: This token is specific to the Switch component
            // switch
            medium: 20,
            // NOTE: This token is specific to the Switch component
            // slider
            large: 40,
```

packages/wonder-blocks-switch/src/themes/default.ts
```diff
@@ -0,0 +1,66 @@
...
+        },
+        offset: {
+            default: 1,
+        },
+    },
+    spacing: {
+        xxxxSmall: tokens.spacing.xxxxSmall_2,
+        xxxSmall: tokens.spacing.xxxSmall_4,
+        medium: tokens.spacing.medium_16,
```
jandrade, line 62: **suggestion:** Similar suggestion here around naming. In fact, by looking at the implementation, I could see another pattern emerging:

https://github.com/Khan/wonder-blocks/pull/2029/files#diff-f0d673a56005e5635e2d3ead871ffdb309a4166f246fedfd77e63a396865c896R159-R160

How do you feel about using a different naming convention in here (similar to `surface` as mentioned in a previous comment):

- `xxxxSmall` => `sliderPosition` or `slider.position` (as it is used for `top` and `left`).
- `xxxSmall` => `iconPosition` or `icon.position`.
- `medium` => `switchTransform` or `switch.transform`. (+ the whole transform value could be included in here).

Of course, this are some thoughts I'm having while I'm looking at the code, so happy to chat more offline about this! I'm even checking about if/how this can be integrated with the `button` theme.

packages/wonder-blocks-switch/src/components/switch.tsx
```diff
@@ -3,20 +3,27 @@ import {StyleSheet} from "aphrodite";
 
 import {
     AriaProps,
-    UniqueIDProvider,
     View,
     addStyle,
+    useUniqueIdWithMock,
```
jandrade, line 8: **praise:** Thanks for switching to the hooks version 👏

packages/wonder-blocks-switch/package.json
```diff
@@ -19,7 +19,8 @@
     "@khanacademy/wonder-blocks-color": "^2.0.1",
     "@khanacademy/wonder-blocks-core": "^6.1.1",
     "@khanacademy/wonder-blocks-icon": "^2.1.5",
-    "@khanacademy/wonder-blocks-spacing": "^4.0.1"
+    "@khanacademy/wonder-blocks-spacing": "^4.0.1",
```
jandrade, line 22: **suggestion:** By including `theming` here, I think you can get rid of `color` and `spacing` as direct dependencies to this package.

If you still need to use them for Storybook or tests, you can move them to `devDependencies`.... although I think Storybook/tests can use `{tokens} from `wb-theming` instead so we can start introducing tokens gradually in the examples.

jandrade (changes requested):
Taking a first pass on this. Great progress so far! I'm requesting changes mainly to cover two important aspects:

1. Removing/changing `color` and `spacing` in `package.json`.
2. Suggesting some changes / opening the convo around component token naming for Switch.

Great work on integrating the `theming` package so smoothly!!

packages/wonder-blocks-switch/src/themes/default.ts
marianmwang: that makes sense to me.

packages/wonder-blocks-switch/src/themes/default.ts
marianmwang: that makes sense! i didn't think about how using a new theme would affect the naming.

__docs__/wonder-blocks-switch/switch.stories.tsx
```diff
@@ -17,6 +17,21 @@ import SwitchArgtypes from "./switch.argtypes";
 
 type StoryComponentType = StoryObj<typeof Switch>;
 
+/**
```
jandrade, line 20: **praise:** YAY thanks for adding the description here!

packages/wonder-blocks-switch/src/components/switch.tsx
```diff
@@ -154,85 +160,92 @@ const sharedStyles = StyleSheet.create({
...
+                transform: theme.spacing.transform,
             },
         };
     } else {
         newStyles = {
             switch: {
-                backgroundColor: disabled ? Color.offBlack32 : Color.offBlack50,
+                backgroundColor: disabled
+                    ? theme.color.bg.switch.disabledOff
+                    : theme.color.bg.switch.off,
```
jandrade, line 233: **praise:** Awesome! I love who these tokens are super readable and map to their elements.

jandrade (approved):
Looks awesome! thanks for adding this 💯

packages/wonder-blocks-switch/src/components/switch.tsx
```diff
@@ -154,85 +160,92 @@ const sharedStyles = StyleSheet.create({
...
-        width: `calc(${Spacing.medium_16}px + ${Spacing.xxxSmall_4}px)`,
-        borderRadius: "50%",
-        backgroundColor: Color.white,
-        transition: "transform 0.15s ease-in-out",
+        top: theme.spacing.slider.position,
+        left: theme.spacing.slider.position,
+        height: theme.size.height.medium,
+        width: theme.size.width.medium,
+        borderRadius: theme.border.radius.full,
+        backgroundColor: theme.color.bg.slider.on,
```
jandrade, line 168: **question:** What happens if you set this as `backgroundColor: theme.color.bg.slider`? does TS throw an error?

marianmwang: discussed offline: it won't throw an error here, but inside the `_generateStyles` function, the `newStyles` object should have a type of `Record<string, CSSProperties>` instead of `Record<string, any>`.

PR #2030, opened 2023-08-30
[🔥AUDIT🔥] [fixsomeerrorsuppressions] Fix a few error suppressions

somewhatabstract:
🖍 _This is an audit!_ 🖍

## Summary:
Fixed a few error suppressions that I happened to know how to fix. Functionally no changes so I didn't add a changeset here.

Issue: FEI-5019

## Test plan:
`yarn typecheck`

### DIFF
--- a/packages/wonder-blocks-timing/src/util/interval.ts
+++ b/packages/wonder-blocks-timing/src/util/interval.ts
@@ -15,7 +15,7 @@ import type {IInterval, SchedulePolicy, ClearPolicy} from "./types";
  * @implements {IInterval}
  */
 export default class Interval implements IInterval {
-    _intervalId: number | null | undefined;
+    _intervalId: ReturnType<typeof setInterval> | null | undefined;
     _action: () => unknown;
     _intervalMs: number;
 
@@ -75,7 +75,6 @@ export default class Interval implements IInterval {
         if (this.isSet) {
             this.clear(ClearPolicies.Cancel);
         }
-        // @ts-expect-error [FEI-5019] - TS2322 - Type 'Timer' is not assignable to type 'number'.
         this._intervalId = setInterval(() => this._action(), this._intervalMs);
     }
 
--- a/packages/wonder-blocks-timing/src/util/timeout.ts
+++ b/packages/wonder-blocks-timing/src/util/timeout.ts
@@ -15,7 +15,7 @@ import type {ITimeout, SchedulePolicy, ClearPolicy} from "./types";
  * @implements {ITimeout}
  */
 export default class Timeout implements ITimeout {
-    _timeoutId: number | null | undefined;
+    _timeoutId: ReturnType<typeof setTimeout> | null | undefined;
     _action: () => unknown;
     _timeoutMs: number;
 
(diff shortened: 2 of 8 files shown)


packages/wonder-blocks-timing/src/util/timeout.ts
```diff
@@ -15,7 +15,7 @@ import type {ITimeout, SchedulePolicy, ClearPolicy} from "./types";
  * @implements {ITimeout}
  */
 export default class Timeout implements ITimeout {
-    _timeoutId: number | null | undefined;
+    _timeoutId: ReturnType<typeof setTimeout> | null | undefined;
```
somewhatabstract, line 18: **note**: This was a tricky one. The global `setTimeout` has a return type of `NodeJS.Timeout`. This is what is likely correct for in tests where we're not in a JSDOM, but the `window.setTimeout` call that will run in browsers is typed to return a number. 

The easy fix to satisfy TS without a functional change is what we have here.

packages/wonder-blocks-timing/src/util/interval.ts
```diff
@@ -15,7 +15,7 @@ import type {IInterval, SchedulePolicy, ClearPolicy} from "./types";
  * @implements {IInterval}
  */
 export default class Interval implements IInterval {
-    _intervalId: number | null | undefined;
+    _intervalId: ReturnType<typeof setInterval> | null | undefined;
```
somewhatabstract, line 18: **note**: This was a tricky one. The global `setInterval` has a return type of `NodeJS.Timer`. This is what is likely correct for in tests where we're not in a JSDOM, but the `window.setInterval` call that will run in browsers is typed to return a number. 

The easy fix to satisfy TS without a functional change is what we have here.

packages/wonder-blocks-tooltip/src/util/ref-tracker.ts
```diff
@@ -41,8 +39,7 @@ export default class RefTracker {
                 throw new Error("targetFn must be a function");
             }
 
-            // @ts-expect-error [FEI-5019] - TS2322 - Type '((instance: any) => void) | null' is not assignable to type '(arg1?: HTMLElement | null | undefined) => void | null | undefined'.
-            this._targetFn = targetFn || null;
+            this._targetFn = targetFn || undefined;
```
kevinbarabash, line 42: I feel like TS' type system tends to favour `undefined` over `null` for some reason.

kevinbarabash (approved):
LGTM 🚢  Thanks for fixing these `@ts-expect-error`s. 💚

packages/wonder-blocks-timing/src/util/interval.ts
kevinbarabash: neat

packages/wonder-blocks-tooltip/src/util/ref-tracker.ts
somewhatabstract: It's not that. It's that because I don't initialize `_targetFn`, it is defaulted to `undefined`. And I didn't bother keeping the ability to type it to `null`, so `null` wasn't valid here.

PR #2037, opened 2023-08-31
WB-1596: Fix cell styles to allow displaying outline with custom background color

jandrade:
## Summary:

While testing the Activities Sidebar updates in the website, I noticed that the
focus outline is missing for all of the DetailCell elements in the sidebar.

Upon closer look in the WB Storybook site, I found out that the outline is
hidden behind the custom background color applied to a cell via the styles prop.

This PR focuses in the following changes:

- Move the `outline/border` state to the `:focus-visible` pseudo-class, so that it
  is only applied when the element is focused and the user is using a keyboard.
- Move the overall cell states to pseudo-classes directly to the `Clickable`
instance, so that the inner button/link element gets the correct styles. By
doing this change we are reducing the complexity on handling the cell states as
we rely that to CSS directly.

Issue: WB-1596

## Test plan:

Verify that the focus outline is displayed when the user is using a keyboard to
navigate on the Cell examples in Storybook.

https://5e1bf4b385e3fb0020b7073c-fqqtpbnmal.chromatic.com/?path=/story/cell-compactcell--compact-cells-as-list-items

https://5e1bf4b385e3fb0020b7073c-fqqtpbnmal.chromatic.com/?path=/docs/cell-detailcell--docs#detail-cells-as-list-items

https://github.com/Khan/wonder-blocks/assets/843075/194a97c6-e17c-478f-87aa-92d58b1a526b


packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -201,28 +178,52 @@ const CellCore = (props: CellCoreProps): React.ReactElement => {
...
     wrapper: {
         background: Color.white,
         color: Color.offBlack,
-        minHeight: CellMeasurements.cellMinHeight,
+        // minHeight: CellMeasurements.cellMinHeight,
         textAlign: "left",
     },
 
     innerWrapper: {
+        // minHeight: CellMeasurements.cellMinHeight,
```
matthewmorgan, line 214: Did you mean to remove this?

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -201,28 +178,52 @@ const CellCore = (props: CellCoreProps): React.ReactElement => {
...
+        </View>
+    );
 };
 
 const styles = StyleSheet.create({
     wrapper: {
         background: Color.white,
         color: Color.offBlack,
-        minHeight: CellMeasurements.cellMinHeight,
+        // minHeight: CellMeasurements.cellMinHeight,
```
matthewmorgan, line 209: Did you mean to remove this?

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -250,43 +251,76 @@ const styles = StyleSheet.create({
...
-        outline: `solid ${Spacing.xxxxSmall_2}px ${Color.blue}`,
-        // The focus ring is not visible when there are stacked cells.
-        // Using outlineOffset to display the focus ring inside the cell.
-        outlineOffset: -Spacing.xxxxSmall_2,
-        // To hide the internal corners of the cell.
-        overflow: "hidden",
-    },
+        // focus (only visible when using keyboard navigation)
+        ":focus-visible": {
+            // borderRadius: Spacing.xxxSmall_4,
```
matthewmorgan, line 268: Did you mean to remove this?

matthewmorgan (commented):
Thanks for the sneak peek!

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
jandrade: ahh good call! I was testing some stuff, but forgot to revert those changes.

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
jandrade: same here thanks!

__docs__/wonder-blocks-cell/compact-cell.stories.tsx
```diff
@@ -317,6 +317,73 @@ CompactCellDisabled.parameters = {
...
+ * We recommend using a faded background color (third cell) to make the cell
+ * look as expected with different states (e.g. hover, focus, active).
+ *
+ * If you use a solid background color (last cell), the cell states will not
+ * change the background color.
+ */
+export const CompactCellsAsListItems: StoryComponentType = {
+    render: () => (
+        <section role="list">
+            <View role="listitem">
```
nishasy, line 333: **question**: Out of curiosity, why did you use `section` and `View` with `role="list"` and `role="listitem"` respectively, instead of just using `ul` and `li`?

jandrade: hmmm great question... I was replicating the code where this was originally introduced in `webapp` so I could make sure that the issue is fixed, but I didn't think very carefully about how the code was created. The `View` part is to get View's default styles, which make sense for the `listitem` case, but the `section` is incorrectly used, and could probably use something different.

Side note: I've noticed that sometimes we tend to overuse `View` instead of relying on semantic elements via the `addStyle` function. This could have used `ul/li` in that case.

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -201,14 +177,28 @@ const CellCore = (props: CellCoreProps): React.ReactElement => {
...
+                ]}
+                aria-current={active ? "true" : undefined}
             >
-                {(eventState) => renderCell(eventState)}
+                {() => renderCell()}
             </Clickable>
         );
     }
 
     // No click event attached, so just render the cell as-is.
```
nishasy, line 193: > No click event attached, so just render the cell as-is.

Is this comment still needed?

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -220,9 +210,19 @@ const styles = StyleSheet.create({
     },
 
     innerWrapper: {
+        minHeight: CellMeasurements.cellMinHeight,
```
nishasy, line 213: Do you need to have both `minHeight` and `height` here?

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -250,43 +250,76 @@ const styles = StyleSheet.create({
...
+            // calculate the width/height and use absolute position to
+            // prevent other elements from being shifted around.
+            position: "absolute",
+            top: 0,
+            left: 0,
+            zIndex: 1,
+            // We remove the border width from the width/height to ensure
+            // that the focus ring is drawn inside the cell.
+            width: `calc(100% - ${Spacing.xxxSmall_4}px)`,
+            height: `calc(100% - ${Spacing.xxxSmall_4}px)`,
```
nishasy, line 288: **Praise:** oooo I see. It's unfortunate that `calc` is necessary, but I get why it has to be done this way. I'll probably end up having to use this myself soon too. Nice solve!

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -250,43 +250,76 @@ const styles = StyleSheet.create({
...
-    },
+    clickable: {
+        outline: "none",
+        /**
+         * States
+         */
+        // disabled
+        // NOTE: We use `aria-disabled` instead of `disabled` because we want
+        // to allow the cell to be focusable even when it's disabled.
+        [":hover[aria-disabled=true]" as any]: {
```
nishasy, line 261: Nice!

nishasy (approved):
Nice! I did want to point out one thing, but I don't think it's related to your change - I noticed that the focus ring has a border radius even if the cell itself is still squared. It creates this awkward squared + rounded corner on the cell when focused. Is this intentional?

Screenshot 2023-09-11 at 7 44 33 PM

jandrade:
> Nice! I did want to point out one thing, but I don't think it's related to your change - I noticed that the focus ring has a border radius even if the cell itself is still squared. It creates this awkward squared + rounded corner on the cell when focused. Is this intentional?
> 
> Screenshot 2023-09-11 at 7 44 33 PM

I noticed that too, but this is intentional by design: https://www.figma.com/file/VbVu3h2BpBhH80niq101MHHE/%F0%9F%92%A0-Main-Components?type=design&node-id=4337-2500&mode=design&t=hZQUu47EvHf1mFCp-4

I think the main goal was to allow using a consistent focus ring, but preserving the "list-view" style for all the other Cell states.

__docs__/wonder-blocks-cell/compact-cell.stories.tsx
jandrade: I've updated the docs to reflect this and recommend this `addStyle` approach.

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
jandrade: Nice catch! I'll delete it.

jandrade: Yeah, I think it's just to mention that we render the cell without any `Clickable` wrapper around. I'll update it to match this case.

jandrade:
Using the WB Snapshot Release, I confirmed that the fix works as expected (cc @matthewmorgan)


https://github.com/Khan/wonder-blocks/assets/843075/4f2afa85-1bee-4191-a202-44faa5631b03

PR #2051, opened 2023-09-08
Add note about addStyle usage

jeremywiebe:
## Summary:

I recently had a question about when to use `addStyle`. Juan [explained](https://khanacademy.slack.com/archives/C4PE1QM5Y/p1694188336590029) and so I figured it'd be useful to add into our docs for other folks.

Issue: "none"

## Test plan:

Review the docs changes. Do they make sense? Are they correct?


jandrade (approved):
Looks great! thanks for adding that clarification :shipit:

PR #2053, opened 2023-09-13
Disblae snapshot release action

marianmwang:
## Summary:
There were some issues found where the latest tag was pointing to the wrong version in perseus, where the snapshot action was borrowed from. See https://khanacademy.atlassian.net/browse/LC-1257.
- Disabled snapshot release action
- Changed branch name condition 

Issue: XXX-XXXX
- WB-1606 has been opened for re-enabling the action once it is resolved

## Test plan:
- Ensure action does not run on this pr


jandrade (approved):
Makes sense. Hopefully the Perseus team will understand better the issue with npm releases so we can turn this on again.

PR #2054, opened 2023-09-14
Add invalid state to single and multi select

marianmwang:
## Summary:
- Added `isInvalid` prop to single and multi select with styling to match
- Added new stories

Issue: WB-1501
image

## Test plan:
- Check new storybook story to ensure styles are being rendered correctly

### DIFF
--- a/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
@@ -275,13 +275,13 @@ export const Invalid: StoryComponentType = {
         expect(dropdown).toHaveStyle("border-width: 2px");
 
         // Focus style
-        fireEvent.focus(dropdown);
+        await fireEvent.focus(dropdown);
         expect(dropdown).toHaveStyle("border-color: #d92916");
         expect(dropdown).toHaveStyle("border-width: 2px");
 
         // Active (mouse down) style
         // eslint-disable-next-line testing-library/prefer-user-event
-        fireEvent.mouseDown(dropdown);
+        await fireEvent.mouseDown(dropdown);
         expect(dropdown).toHaveStyle("background-color: rgb(243, 187, 180)");
     },
 };
--- a/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
@@ -9,7 +9,7 @@ import Color from "@khanacademy/wonder-blocks-color";
 import {Checkbox} from "@khanacademy/wonder-blocks-form";
 import {OnePaneDialog, ModalLauncher} from "@khanacademy/wonder-blocks-modal";
 import Spacing from "@khanacademy/wonder-blocks-spacing";
-import {HeadingLarge} from "@khanacademy/wonder-blocks-typography";
+import {HeadingLarge, LabelMedium} from "@khanacademy/wonder-blocks-typography";
 import {MultiSelect, OptionItem} from "@khanacademy/wonder-blocks-dropdown";
 import type {Labels} from "@khanacademy/wonder-blocks-dropdown";
 
@@ -260,12 +260,15 @@ const ErrorWrapper = (args: any) => {
 
     return (
         <View style={styles.wrapper}>
+            <LabelMedium style={{marginBottom: Spacing.xSmall_8}}>
+                Select at least 2 options to clear the error!
+            </LabelMedium>
             <MultiSelect
                 {...args}
                 error={error}
                 onChange={(values) => {
                     setSelectedValues(values);
-                    setError(false);
+                    setError(values.length < 2);
                 }}
                 onToggle={setOpened}
                 opened={opened}
@@ -279,7 +282,7 @@ const ErrorWrapper = (args: any) => {
 
 /**
  * Here is an example of a dropdown that is in an error state.
- * Selecting one or more options will clear the error by setting the `error` prop to `false`.
+ * Selecting two or more options will clear the error by setting the `error` prop to `false`.
  */
 export const Error: StoryComponentType = {
     render: ErrorWrapper,
--- a/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
@@ -28,6 +28,7 @@ export default {
     argTypes: multiSelectArgtypes,
     args: {
         isFilterable: false,
+        isInvalid: false,
         opened: false,
         disabled: false,
         light: false,
@@ -250,6 +251,16 @@ CustomStylesOpened.parameters = {
             story: "Here you can see an example of the previous dropdown opened",
         },
     },
+};
+
+/**
+ * Here is an example of a dropdown that is in an invalid state.
+ */
+export const Invalid: StoryComponentType = {
+    render: Template,
+    args: {
+        isInvalid: true,
+    } as MultiSelectArgs,
 };
 
 /**
--- a/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
@@ -30,7 +30,7 @@ export default {
     argTypes: multiSelectArgtypes,
     args: {
         isFilterable: false,
-        isInvalid: false,
+        error: false,
         opened: false,
         disabled: false,
         light: false,
@@ -261,7 +261,7 @@ CustomStylesOpened.parameters = {
 export const Invalid: StoryComponentType = {
     render: Template,
     args: {
-        isInvalid: true,
+        error: true,
     } as MultiSelectArgs,
     play: async ({canvasElement}) => {
         const canvas = within(canvasElement);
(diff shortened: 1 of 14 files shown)


__docs__/wonder-blocks-dropdown/single-select.stories.tsx
```diff
@@ -340,6 +343,45 @@ Disabled.parameters = {
...
+        // Focus style
+        await fireEvent.focus(dropdown);
+        expect(dropdown).toHaveStyle("border-color: #d92916");
+        expect(dropdown).toHaveStyle("border-width: 2px");
+
+        // Active (mouse down) style
+        // eslint-disable-next-line testing-library/prefer-user-event
+        await fireEvent.mouseDown(dropdown);
+        expect(dropdown).toHaveStyle("background-color: rgb(243, 187, 180)");
+    },
```
jandrade, line 382: **suggestion:** I would suggest to move this into a Jest unit test. I don't think these kind of examples require Storybook interaction tests mostly because the visuals for this can be snapshotted and confirmed in Chromatic (at least the `resting` state). For the other states, I think this can be moved to Jest.

packages/wonder-blocks-dropdown/src/components/dropdown-core.tsx
```diff
@@ -964,6 +976,10 @@ class DropdownCore extends React.Component<Props, State> {
                             maxDropdownHeight,
                         ),
                     ]}
+                    aria-invalid={role === "listbox" ? ariaInvalid : undefined}
+                    aria-required={
+                        role === "listbox" ? ariaRequired : undefined
+                    }
```
jandrade, line 982: **suggestion:** It would be good adding a comment explaining why we have to add the condition here.

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -32,9 +33,10 @@ type SelectOpenerProps = AriaProps & {
      * id should match the field id.
      */
     id?: string;
-    //TODO: error state
-    // error: boolean,
-
+    /**
+     * Whether or not the input has an invalid value. Defaults to false.
+     */
+    isInvalid: boolean;
```
jandrade, line 39: **suggestion:** As I mentioned in the first comment, it would be better using the `error` name here to describe that state for consistency purposes across the Design System (e.g. `Checkbox`, `Radio`, `TextField` all use `error` to define this state).

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -57,11 +55,16 @@ type SelectOpenerProps = AriaProps & {
      * Whether the dropdown is open.
      */
     open: boolean;
+    /**
+     * Test ID used for e2e testing.
+     */
+    testId?: string;
```
jandrade, line 61: **praise:** Ah nice catch!

packages/wonder-blocks-dropdown/src/components/single-select.tsx
```diff
@@ -141,6 +141,10 @@ type Props = AriaProps &
          * top. The items will be filtered by the input.
          */
         isFilterable?: boolean;
+        /**
+         * Whether or not the input has an invalid value. Defaults to false.
+         */
+        isInvalid?: boolean;
```
jandrade, line 147: Same suggestion here around using the `error` name instead.

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -11,6 +11,7 @@ import Icon, {icons} from "@khanacademy/wonder-blocks-icon";
 import Spacing from "@khanacademy/wonder-blocks-spacing";
 import {LabelMedium} from "@khanacademy/wonder-blocks-typography";
 import {DROPDOWN_ITEM_HEIGHT} from "../util/constants";
+import tokens from "../../../wonder-blocks-theming/src/tokens";
```
jandrade, line 14: **issue(blocking):** This would be better if it imports that from `@ka/wb-theming` instead as the file would be directly imported and bundled as part of the `wb-dropdown` package.

This would mean that you'd need to make changes to `package.json` and other files like the TS config to include `theming` as a dependency.

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -222,28 +235,31 @@ const _generateStyles = (light: boolean, placeholder: boolean) => {
...
-                background: "transparent",
+                background: invalid ? tokens.color.fadedRed8 : "transparent",
                 color: placeholder ? white50 : white,
-                borderColor: white50,
+                borderColor: invalid ? Color.red : white50,
                 borderWidth: 1,
             },
             focus: {
-                borderColor: white,
+                borderColor: invalid ? tokens.color.fadedRed8 : white,
```
jandrade, line 244: **praise:** Nice! thanks for starting to use `theming` in other components.

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -222,28 +235,31 @@ const _generateStyles = (light: boolean, placeholder: boolean) => {
...
                 borderWidth: 2,
                 paddingLeft: adjustedPaddingLeft,
                 paddingRight: adjustedPaddingRight,
             },
             active: {
                 paddingLeft: adjustedPaddingLeft,
                 paddingRight: adjustedPaddingRight,
-                borderColor: mix(fade(blue, 0.32), white),
+                borderColor: invalid ? Color.red : mix(fade(blue, 0.32), white),
```
jandrade, line 252: **suggestion:** Now that you are going to import theming, you could use `fadedBlue` here:

```suggestion
                borderColor: invalid ? Color.red : tokens.color.fadedBlue,
```

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -222,28 +235,31 @@ const _generateStyles = (light: boolean, placeholder: boolean) => {
...
                 borderWidth: 2,
                 color: placeholder
                     ? mix(fade(white, 0.32), blue)
                     : mix(fade(blue, 0.32), white),
-                backgroundColor: mix(offBlack32, blue),
+                backgroundColor: Color
+                    ? tokens.color.fadedRed
+                    : mix(offBlack32, blue),
```
jandrade, line 259: **suggestion:**
```suggestion
                backgroundColor: Color
                    ? tokens.color.fadedRed
                    : tokens.color.activeBlue
```

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -252,20 +268,22 @@ const _generateStyles = (light: boolean, placeholder: boolean) => {
...
                 paddingLeft: adjustedPaddingLeft,
                 paddingRight: adjustedPaddingRight,
             },
             active: {
-                background: mix(fade(blue, 0.32), white),
-                borderColor: mix(offBlack32, blue),
+                background: invalid
+                    ? tokens.color.fadedRed
+                    : mix(fade(blue, 0.32), white),
+                borderColor: invalid ? Color.red : mix(offBlack32, blue),
```
jandrade, line 286: same suggestion for using `tokens.color` values here instead of `mix`, `fade`, etc. This would be really helpful to start cleaning up the repo from those one-off colors and centralize everything with tokens :)

__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
```diff
@@ -252,6 +255,37 @@ CustomStylesOpened.parameters = {
...
 };
 
+/**
+ * Here is an example of a dropdown that is in an invalid state.
+ */
+export const Invalid: StoryComponentType = {
+    render: Template,
+    args: {
+        isInvalid: true,
+    } as MultiSelectArgs,
```
jandrade, line 265: **thought:** How do you feel about using an example that uses a wrapper to manage state, where the `error` state is set `true` by default, and it's set `false` if the user chooses an option, so they can see how the Dropdown goes from `error` to valid in the example?

jandrade (changes requested):
Thanks for working on this! This is looking really good. Requesting changes mainly for:

1. Fixing an issue with direct import of the `theming/tokens` module.
2. Renaming the prop from `isInvalid` to `error` (apologies for not mentioning this earlier as I think I used `invalid`/`isInvalid` in the prototype).
3. Moving interaction tests to unit tests.
4. [optional] Changing the `error` stories to be a bit more interactive in Storybook (managing state between error / no error).

__docs__/wonder-blocks-dropdown/single-select.stories.tsx
marianmwang: ahh i see. i thought interaction tests were more for visual styles but i guess chromatic snapshots makes sense. i'll update it.

__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
marianmwang: that's a good idea! i've added the stories

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -4,19 +4,16 @@ import {__RouterContext} from "react-router";
 
 import type {AriaProps} from "@khanacademy/wonder-blocks-core";
 
-import Color, {mix, fade} from "@khanacademy/wonder-blocks-color";
```
jandrade, line 7: **praise:** Neat! thanks for moving everything from `Color, Spacing` to `tokens` 👏

jandrade (approved):
Looks awesome!! Thanks for adding this new prop while cleaning up the code (TS, Color, Spacing, typos, etc). Great job 👏 🚢 :shipit: 🚀

PR #2057, opened 2023-09-18
[🔥AUDIT🔥] WB-1593.2: Optimize Breadcrumbs docs to reduce number of stories

jandrade:
🖍 _This is an audit!_ 🖍

## Summary:

This PR refactors the Breadcrumbs docs to reduce the number of stories. The
accessibility stories have been removed and replaced with a call to the existing
story in the Breadcrumbs docs.

Issue: WB-1593

## Test plan:

Check if the number of Breadcrumbs stories has been reduced. Also check for the
total number of stories (currently at 362).


jandrade:
> Oh interesting! I also note that it only generated 1 snapshot, is that typical? So we would only be getting charged for 1 in this case?

@jeresig from what I can see.. in theory, as we are using [TurboSnap](https://www.chromatic.com/docs/turbosnap/), that would mean that Chromatic only generated snapshots for the affected changes, which is the `Breadcrumbs` docs in this case.

My change here was to remove an extra story previously defined in the `accessibility.stories` file and instead reuse the one from the `breadcrumbs.stories` file so Chromatic would only generate 1 story instead of 2. I'm happy to see this is the case!

I've reached out the Chromatic team to understand better how Snapshots work and to see if there's a way to programmatically know how many snapshots we have taken and they answered that they include that info in the `Usage Reports` section in the admin panel.

PR #2062, opened 2023-09-21
WB-1609: Add support to disabled + focus state to Switch

jandrade:
## Summary:

We are not currently allowing to focus a disabled switch component. This PR adds
support to that via the `aria-disabled` attribute.

Also made the default story fully interactive so users can play/interact with
it.

Issue: WB-1609

## Test plan:

1. Verify that the switch elements in the `Disabled` state can be focused via
keyboard.

/?path=/story/switch--disabled

https://github.com/Khan/wonder-blocks/assets/843075/7f6c4f2f-424b-4fed-8695-33e4399b2c2c

2. Verify that the `Default` story is fully interactive (switch can be
checked/unchecked).

/?path=/story/switch--default

### DIFF
--- a/.changeset/fast-jokes-hammer.md
+++ b/.changeset/fast-jokes-hammer.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-switch": patch
+---
+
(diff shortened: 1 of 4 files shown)


jeresig (approved):
Great! Thanks for fixing this!

PR #2065, opened 2023-09-26
Fix how `mix` from `wonder-blocks-color` formats single-digit color components

timmcca-be:
## Summary:
Previously, if the red, green, or blue component was less than 16, its hex digit would be repeated (e.g. `f` would become `ff`). With this change, we preserve the real value of the number by prepending a `0` instead (e.g. `f` becomes `0f`).

## Test plan:
`yarn run jest packages/wonder-blocks-color/src/util/__tests__/utils.test.ts`

### DIFF
--- a/.changeset/afraid-rings-knock.md
+++ b/.changeset/afraid-rings-knock.md
@@ -1,5 +1,5 @@
 ---
-"@khanacademy/wonder-blocks-color": patch
+"@khanacademy/wonder-blocks-color": major
 ---
 
 Fix how mix formats single-digit color components
--- a/.changeset/afraid-rings-knock.md
+++ b/.changeset/afraid-rings-knock.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-color": patch
+---
+
+Fix how mix formats single-digit color components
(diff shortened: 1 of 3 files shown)


.changeset/afraid-rings-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-color": patch
```
somewhatabstract, line 2: **nit:** This is a breaking change for code that relies on the old behavior. As such I think this likely needs to be a `major` change, though I could be persuaded down to `minor` since it is unlikely that there is code relying on this.

packages/wonder-blocks-color/src/util/utils.ts
```diff
@@ -61,10 +61,9 @@ const format = (color: Color): string => {
     const b = Math.round(color.b);
 
     if (color.a === 1) {
-        // @ts-expect-error [FEI-5019] - TS7006 - Parameter 'c' implicitly has an 'any' type.
```
somewhatabstract, line 64: **praise:** Thanks for taking the time to get rid of an error suppression!

somewhatabstract (approved):
This looks good to me, in the most part, though I think this should be a `major` bump given that it changes functionality.

In addition, the test plan `make tesc` doesn't make sense for Wonder Blocks as there is no `make tesc`. Should be `yarn test` or `yarn jest packages/wonder-blocks-color`, I believe.

---
My review uses [Conventional Comments][1] to add context and set expectations for the comments I am leaving.

[1]:https://conventionalcomments.org/

PR #2071, opened 2023-09-27
Verify changeset for every changed package

jandrade:
## Summary:

Another implementation borrowed from the Perseus repo, that help us to verify
that there's a changeset for every changed package.

Previously, we were using `check-for-changeset` to verify the existence of a
changeset file, but now we switch to `changeset-per-package` to make sure that
there's a changeset for every changed package. This is useful in monorepos like
this one.

See original PR: https://github.com/Khan/perseus/pull/735 (thanks @nedredmond!)

Issue: XXX-XXXX

## Test plan:

Verify that the CI passes and that in other PRs with actual changes, the
changeset workflow works as expected.

### DIFF
--- a/.github/workflows/node-ci-pr.yml
+++ b/.github/workflows/node-ci-pr.yml
@@ -27,6 +27,37 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
+  changeset:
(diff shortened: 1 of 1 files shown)


nedredmond (approved):
I'm honored.

PR #2073, opened 2023-09-29
WB-1623: Cell states don't work with anchor tags correctly [fix]

jandrade:
## Summary:

There’s an issue with wb-cell that was introduced after making some layout/css
changes to the components.

Turns out that Cell states won’t be displayed as expected when the cell top node
is `display: inline`. As cells are meant to use the full-width of it’s parent
container, this PR fixes that by making the top node to always use `display: flex`.

Issue: https://khanacademy.atlassian.net/browse/WB-1623

## Test plan:

Verify that cell states are displayed as expected (hover, active change the bg
color).

### DIFF
--- a/.changeset/hot-starfishes-sleep.md
+++ b/.changeset/hot-starfishes-sleep.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-cell": patch
+---
+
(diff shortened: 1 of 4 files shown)


jeresig (approved):

adinardi:
yay thanks!

PR #2077, opened 2023-10-02
Handle custom media layout specs.

jeresig:
## Summary:
We weren't taking into account custom ones, which was a problem if you wante to use a custom one!

Issue: XXX-XXXX

## Test plan:


jeresig:
Landing to unblock @matthewmorgan - he confirmed that this is working for him, though!

PR #2078, opened 2023-10-02
Fix filters for changeset workflow step

nedredmond:
By default, filters are inclusive disjunctions (OR). This makes the `packages/` filter useless, since the glob filter will match anything that does not include `__tests__` and `dist` in the path.

This field that I added to the `filter-files` action will ensure that file pathnames match EVERY provided filter. This prevents accidental inclusions of packages outside of `packages/`.


jandrade (approved):
Thanks for fixing this! :shipit:

PR #2081, opened 2023-10-05
Fix glob match issue in `filter-files` action

nedredmond:
Corrects an issue where we unexpectedly match all files except those with both `tests` and `dist` in the path.


jeresig (approved):

jandrade (approved):
🎉

PR #2086, opened 2023-10-13
Add `endIcon` prop and change existing `icon` prop to `startIcon`

nishasy:
## Summary:
Currently, the `icon` prop on WB Button added an icon on the left.
Now, `icon` has been renamed to `startIcon`, and we are adding
an `endIcon` prop so people can add an icon on the other side.

Issue: https://khanacademy.atlassian.net/browse/WB-1630

## Test plan:
In storybook, confirm that the icon story has the right
description and that the icons look correct.

`yarn jest packages/wonder-blocks-button/src/components/__tests__/button.test.tsx`

image

### DIFF
--- a/.changeset/rich-jars-jog.md
+++ b/.changeset/rich-jars-jog.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-button": minor
+---
+
+Add `endIcon` prop and change existing `icon` prop to `startIcon`
--- a/__docs__/wonder-blocks-button/button.argtypes.ts
+++ b/__docs__/wonder-blocks-button/button.argtypes.ts
@@ -7,8 +7,22 @@ export default {
         description: "Text to appear on the button.",
         type: {name: "string", required: true},
     },
-    icon: {
-        description: "A Phosphor icon asset (imported as a static SVG file).",
+    startIcon: {
+        description: `A Phosphor icon asset (imported as a static SVG file)
+            that will appear at the start of the button (left for LTR, right
+            for RTL).`,
+        type: {name: "other", value: "PhosphorIconAsset", required: false},
+        control: {type: "select"},
+        options: IconMappings,
+        table: {
+            category: "Layout",
+            type: {summary: "PhosphorIconAsset"},
+        },
+    },
+    endIcon: {
+        description: `A Phosphor icon asset (imported as a static SVG file)
+            that will appear at the end of the button (right for LTR, left
+            for RTL).`,
         type: {name: "other", value: "PhosphorIconAsset", required: false},
         control: {type: "select"},
         options: IconMappings,
--- a/__docs__/wonder-blocks-button/button.argtypes.ts
+++ b/__docs__/wonder-blocks-button/button.argtypes.ts
@@ -13,7 +13,8 @@ export default {
             for RTL).`,
         type: {name: "other", value: "PhosphorIconAsset", required: false},
         control: {type: "select"},
-        options: IconMappings,
+        options: Object.keys(IconMappings),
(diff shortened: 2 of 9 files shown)


packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -54,22 +54,17 @@ const ButtonCore: React.ForwardRefExoticComponent<
...
             id,
             waiting: _,
             ...restProps
         } = props;
 
-        const iconWidth = icon
-            ? size === "small"
-                ? theme.size.width.medium
-                : theme.size.width.large
-            : 0;
```
nishasy, line 67: I realized this isn't being used anywhere. It used to be used in `_generatestyles()`, but that got replaced by the `padding` value.

.changeset/rich-jars-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-button": minor
```
jandrade, line 2: **suggestion:** It's better to treat this as a breaking change as it is changing the existing API (`icon` => `startIcon`).

For reference: https://khanacademy.atlassian.net/wiki/spaces/WB/pages/2217640195/Versioning+and+releasing+Wonder+Blocks+packages#Major-changes-1.0.0

__docs__/wonder-blocks-button/button.argtypes.ts
```diff
@@ -7,8 +7,22 @@ export default {
         description: "Text to appear on the button.",
         type: {name: "string", required: true},
     },
-    icon: {
-        description: "A Phosphor icon asset (imported as a static SVG file).",
+    startIcon: {
+        description: `A Phosphor icon asset (imported as a static SVG file)
+            that will appear at the start of the button (left for LTR, right
+            for RTL).`,
```
jandrade, line 13: **praise:** Amazing as always, thanks for adding a11y docs even in the smallest details.

__docs__/wonder-blocks-button/button.argtypes.ts
```diff
@@ -7,8 +7,22 @@ export default {
...
+        type: {name: "other", value: "PhosphorIconAsset", required: false},
+        control: {type: "select"},
+        options: IconMappings,
```
jandrade, line 16: **suggestion:** This is something that I missed when I changed this. Turns out that we need to use both keys to preserve the icon selection in a separate Storybook tab.

```suggestion
        options: Object.keys(IconMappings),
        mapping: IconMappings,
```


e.g. https://khan.github.io/wonder-blocks/?path=/story/button--default&args=color:destructive

If you change the props in the `Controls` tab, the icon is not currently preserved, but if we make this change, then it's added to the query string so it can be included in a "permalink" version.

Permalink version: https://khan.github.io/wonder-blocks/iframe.html?args=color:destructive&id=button--default&viewMode=story

packages/wonder-blocks-button/src/components/button-core.tsx
jandrade: oh nice catch!

Now that you got rid of this, I think you can also remove these tokens from the default theme: https://github.com/Khan/wonder-blocks/blob/1644f6510760c26c8ab407d49654faae36fddcc9/packages/wonder-blocks-button/src/themes/default.ts#L107-L110

jandrade (approved):
Looks really great! A couple of small suggestions before landing, thanks for working on this so quickly! 💯

.changeset/rich-jars-jog.md
nishasy: Ah yeah, that's a good point

PR #2087, opened 2023-10-16
WB-1587.1: Add theming support to IconButton component

jandrade:
## Summary:

This PR adds theming support to the IconButton component. It also adds a new
theme file for the default theme.

Also updated the `disabled + light` state to match how we handle disabled states across WB button components.

The next PR will add a new Khanmigo theme to the IconButton component.

Issue: WB-1587

## Test plan:

Verify in Storybook that the IconButton component looks as expected.

https://github.com/Khan/wonder-blocks/pull/2087#issuecomment-1764927773


__docs__/wonder-blocks-icon-button/icon-button.stories.tsx
```diff
@@ -167,14 +168,57 @@ export const Variants: StoryComponentType = {
 };
 
 /**
- * An IconButton on a dark background.
- * Only the primary kind is allowed to have the `light` prop set to true.
+ * IconButton has a `color` that is either `default` (the default, as shown
+ * above) or `destructive` (as can seen below):
+ */
+export const WithColor: StoryComponentType = {
```
jandrade, line 174: **note:** I noticed that we don't have docs for the `color` prop, so I added a new story to cover that.

__docs__/wonder-blocks-icon-button/icon-button.stories.tsx
```diff
@@ -167,14 +168,57 @@ export const Variants: StoryComponentType = {
 };
 
 /**
- * An IconButton on a dark background.
- * Only the primary kind is allowed to have the `light` prop set to true.
+ * IconButton has a `color` that is either `default` (the default, as shown
+ * above) or `destructive` (as can seen below):
```
nishasy, line 172: How do you feel about adding a sentence like "Note that the focus ring has the color even when the `kind` is `secondary` or `tertiary`"?

__docs__/wonder-blocks-icon-button/icon-button.stories.tsx
```diff
@@ -167,14 +168,57 @@ export const Variants: StoryComponentType = {
...
-            <View style={styles.dark}>
+            <View style={[styles.dark, styles.row]}>
+                <IconButton
+                    icon={magnifyingGlass}
+                    aria-label="search"
+                    light={true}
+                    onClick={(e) => console.log("Click!")}
+                />
                 <IconButton
+                    color="destructive"
```
nishasy, line 221: Is this supposed to look different than the default normally and on hover? It looks like the color only shows in the "active" state when clicking on the button.

nishasy (commented):
Love the tokens! I did leave one suggestion and one question, but I didn't want to block your PR.

__docs__/wonder-blocks-icon-button/icon-button.stories.tsx
jandrade: ah yeah good point, I'll add that!

__docs__/wonder-blocks-icon-button/icon-button.stories.tsx
jandrade: That's expected (the color only changes in the pressed state). You can see how that matches to what we do with `Button`: https://www.figma.com/file/VbVu3h2BpBhH80niq101MHHE/%F0%9F%92%A0-Main-Components?type=design&node-id=6396-5801&mode=design&t=Z5d5gJ1FZQ2VDI8L-4

__docs__/wonder-blocks-icon-button/icon-button.stories.tsx
jandrade: hmmm giving a second thought, the color changes based on different factors (whether is light, whether is using the khanmigo theme), so I prefer to leave it as it is.

nishasy (approved):

PR #2088, opened 2023-10-16
Add Khanmigo theming for buttons with icons

nishasy:
## Summary:
We want to allow buttons to have a khanmigo-themed version of the icons.
The idea is that icons will be wrapped by a rounded container and have
different focus/hover states (probably worth defining a pressed state as well).

Still waiting on design confirmation that this looks right when there
are icons on both sides.

More info in this slack thread:
https://khanacademy.slack.com/archives/C8Z9DSKC7/p1697490308355979

Issue: https://khanacademy.atlassian.net/browse/WB-1631

## Test plan:
Manual testing on Storybook (/button--khanmigo-theme)
For both the regular theme and the Khanmigo theme:
- Make sure everything looks normal when idle and all the icons are visible
- Make sure that the focus ring looks correct when tabbing through the buttons

Khanmigo theme:
- Confirm that the background circle is visible for the secondary and
  tertiary buttons.
- Confirm that the hoveer/active styles are as expected for all buttons.


packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -271,10 +287,41 @@ const themedSharedStyles: ThemedStylesFn<ButtonThemeContract> = (theme) => ({
     },
     startIcon: {
         marginInlineEnd: theme.padding.small,
+        marginInlineStart: -theme.padding.small,
```
jandrade, line 290: **thought:** One thing that got me thinking is that `marginInlineStart` (and `end`) is not supported by Safari v14 (we currently have to support this version). I think it would be fine to keep it this way as we haven't gotten any reports about this, but it's worth keeping an eye on it.

Hopefully we'll update our support list to support Safari v15+ in the near future.

jandrade (approved):
Looks great! just to confirm... is the final design approved? or is there any feedback we need from the Design team?

__docs__/wonder-blocks-button/button.stories.tsx
```diff
@@ -546,12 +550,33 @@ Spinner.parameters = {
...
+            <Button
+                onClick={() => {}}
+                style={styles.truncatedButton}
+                startIcon={plus}
+            >
+                label too long for the parent container
+            </Button>
+            <Strut size={Spacing.medium_16} />
+            <Button
+                size="small"
```
nishasy, line 567: Adding small buttons to story

jandrade (approved):
Changes look great, thanks! 🚢

PR #2089, opened 2023-10-19
Make `useOnMountEffect` accept functions with void return types

timmcca-be:
## Summary:
Previously, `useOnMountEffect` accepted functions that either return a cleanup function or explicitly return undefined. This means that functions with no return value (e.g. `() => {}`) result in errors, even though they're accepted by `useEffect`. This PR updates the argument type so that it accepts functions with void returns. Minor version bump because the API changed, but it is non-breaking.

## Test plan:
N/A

### DIFF
--- a/packages/wonder-blocks-core/src/hooks/use-on-mount-effect.ts
+++ b/packages/wonder-blocks-core/src/hooks/use-on-mount-effect.ts
@@ -28,9 +28,7 @@ import * as React from "react";
  *
  * If you only need to do something on mount, don't return a cleanup function from `callback`.
  */
-export const useOnMountEffect = (
-    callback: () => void | undefined | (() => void),
-): void => {
+export const useOnMountEffect = (callback: () => void | (() => void)): void => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-core/src/hooks/use-on-mount-effect.ts
```diff
@@ -29,7 +29,7 @@ import * as React from "react";
  * If you only need to do something on mount, don't return a cleanup function from `callback`.
  */
 export const useOnMountEffect = (
-    callback: () => undefined | (() => void),
+    callback: () => void | undefined | (() => void),
```
kevinbarabash, line 32: This can be `callback: () => void | (() => void)` since a `void` return type essentially ignores the type of the return, see https://www.typescriptlang.org/play?ssl=1&ssc=31&pln=1&pc=1#code/DYUwLgBAZg9jBcEAUBKCBeAfBAbjAlgCYbJpYQCsA3AFBA.

kevinbarabash (approved):
Thanks for improving the types for this hook.  One minor suggestion inline.

PR #2090, opened 2023-10-19
WB-1587.2: IconButton - Switch to pseudo-classes instead of JS style objects

jandrade:
## Summary:

Before we implement the Khanmigo theme, we need to make sure that we can support
pseudo-classes in our components. Now that browsers have caught up with better
support for buttons/links, we can switch from `ClickableBehavior` to
pseudo-classes. This is done to bring the following benefits:

- Allow browsers to handle these states via CSS (instead of relying on custom JS
logic).
- Some minimal performance improvements (but still improvements): no need to
re-render the component if these states change b/c CSS now handles that for us.
- Allow us to use Chrome DevTools to inspect the different states.
- Improved Storybook DX as it allow us to use an add-on to display different
states.

This PR is responsible for making these changes:

- Removed `ClickableBehavior` from `IconButton` and `IconButtonCore`.
- Reworked the `IconButton` component to use pseudo-classes instead of JS style
objects.
- Added the `storybook-addon-pseudo-states` package to be able to display
pseudo-classes in Storybook. This will be helpful for taking snapshots in
Chromatic.
- Removed the existing IconButton snapshots in favor of the new ones that
display pseudo-classes (via the pseudo-states addon).
- Modified the `secondary` and `tertiary` variants to remove the focus ring on
hover. This was a change requested by design.

NOTE: The next PR will introduce the khanmigo theme.

Issue: WB-1587

## Test plan:

Verify that the `IconButton` component still works as expected. Verify that the
new "All variants" stories look correct in all the different states (default,
hover, focus, active).


consistency-tests/__tests__/clickables.test.tsx
```diff
@@ -38,14 +38,13 @@ const IconButtonWrapper = (props: any) => (
 );
 
 describe.each`
-    Component            | name             | role
-    ${ActionItem}        | ${"ActionItem"}  | ${"menuitem"}
-    ${Button}            | ${"Button"}      | ${"button"}
-    ${ClickableWrapper}  | ${"Clickable"}   | ${"link"}
-    ${CompactCell}       | ${"CompactCell"} | ${"link"}
-    ${DetailCell}        | ${"DetailCell"}  | ${"link"}
-    ${IconButtonWrapper} | ${"IconButton"}  | ${"link"}
```
jandrade, line 47: **note:** The relevant tests have now been moved to `icon-button.test.tsx`. Note that I removed the `window.open` test as we no longer require to test this with `IconButton` because we let the browser handle this for us, so there's no need to test implementation details in that case from now on.

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
```diff
@@ -149,7 +134,7 @@ const IconButtonCore: React.ForwardRefExoticComponent<
                 <StyledButton
                     type="button"
                     {...commonProps}
-                    disabled={disabled}
+                    aria-disabled={disabled}
```
jandrade, line 137: **note:** Changed from `disabled` to `aria-disabled` to match what we do with other clickable components (allowing to focus on disabled interactive elements to allow Screen Readers to detect these elements).

See: https://css-tricks.com/making-disabled-buttons-more-inclusive/

packages/wonder-blocks-icon-button/src/components/icon-button.tsx
```diff
@@ -173,56 +171,28 @@ export const IconButton: React.ForwardRefExoticComponent<
...
     } = props;
-    const renderClickableBehavior = (router: any): React.ReactNode => {
-        const ClickableBehavior = getClickableBehavior(
-            href,
-            skipClientNav,
-            router,
-        );
 
-        return (
-            <ClickableBehavior
```
nishasy, line 191: YAYYYYY

nishasy (approved):
Very cool!! I'm so excited for this!

consistency-tests/__tests__/clickables.test.tsx
jandrade: ah I meant about the `IconButtonWrapper` part :). All the other tests for other clickable elements will remain here.

You can see the tests copied to icon-button here: https://github.com/Khan/wonder-blocks/pull/2090/files/753c9ebc30f9ea81fb1494644abd5d0e434e8f74#diff-281fe82fd57699c210462f743c18eb5434a47b22ef793de121d7151dce6f5bdeR165

nishasy: I see, thanks!

jandrade:
The parent pull-request ([#2087](https://github.com/Khan/wonder-blocks/pull/2087)) has been merged into `main`, but this branch (`theming-icon-button-2`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #2096, opened 2023-10-21
Accordion animations!

nishasy:
## Summary:
Add animations to accordion if `includeAnimation` prop
is set to true.

Animations are done using grid layout because it's
smooth and more performant than using transitions
with height or max-height.

TBD: performance metrics in webapp with this implementation

## Test plan:
`yarn jest packages/wonder-blocks-accordion/src/components/__tests__/accordion.test.tsx`
`yarn jest packages/wonder-blocks-accordion/src/components/__tests__/accordion-section.test.tsx`
`yarn jest packages/wonder-blocks-accordion/src/components/__tests__/accordion-section-header.test.tsx`

Storybook
http://localhost:6061/?path=/story/accordion-accordion--with-animation


packages/wonder-blocks-accordion/src/components/accordion.tsx
```diff
@@ -60,6 +60,15 @@ type Props = AriaProps & {
...
     cornerKind?: AccordionCornerKindType;
+    /**
+     * Whether to include animation on the header. This should be false
+     * if the user has `prefers-reduced-motion` opted in. Defaults to false.
+     *
+     * If this prop is specified both here in the Accordion and within
+     * a child AccordionSection component, the Accordion’s includeAnimation
+     * value is prioritized.
+     */
+    includeAnimation?: boolean;
```
kevinbarabash, line 71: **suggestion (non-blocking):** I think `animated?: boolean` might fit better with existing boolean props in other components such as `opened`.

packages/wonder-blocks-accordion/src/components/accordion-section.tsx
```diff
@@ -227,32 +251,51 @@ const AccordionSection = React.forwardRef(function AccordionSection(
...
+        gridTemplateRows: "min-content 1fr",
+    },
     contentWrapper: {
+        overflow: "hidden",
+    },
+    conentWrapperCollapsed: {
+        // Make sure screen readers don't read the content when it's
+        // collapsed.
+        visibility: "hidden",
+    },
```
kevinbarabash, line 296: **suggestion:** this style should be applied after the animation completes.  Currently, it gets applied at the start which causes the contents to pop out of existence.

packages/wonder-blocks-accordion/src/components/__tests__/accordion.test.tsx
```diff
@@ -359,6 +359,32 @@ describe("Accordion", () => {
...
+            {wrapper: RenderStateRoot},
+        );
+
+        // Act
+        const sectionHeader = screen.getByTestId("section-header-test-id");
+
+        // Assert
+        expect(sectionHeader).toHaveStyle({
+            transition: "border-radius 300ms",
+        });
```
kevinbarabash, line 385: **suggestion (non-blocking):** Could you leave a comment about why this assertion verifies what the test's title says it's testing?

kevinbarabash (commented):
This is looking pretty good.  There's a slight issue with the content disappearing before a section is finished the collapse animation.  I think it's worth creating a test with more content in the sections so it's easier to see this when manually testing.  Also, you may also want to test what happens if the user clicks on the header again before the animation completes.  This admittedly hard to do because the duration is so short so you may need to temporarily make it longer.  This is just to verify that the accordion doesn't get it a bad state if you click the header before the animation completes.

packages/wonder-blocks-accordion/src/components/accordion.tsx
jandrade: That's a good point @kevinbarabash, I was thinking about using something like this as well.

packages/wonder-blocks-accordion/src/components/accordion-section.tsx
nishasy: Oh sorry! This isn't the correct approach. Let me push that really quickly.

packages/wonder-blocks-accordion/src/components/accordion.tsx
```diff
@@ -60,6 +60,15 @@ type Props = AriaProps & {
      * value is prioritized.
      */
     cornerKind?: AccordionCornerKindType;
+    /**
+     * Whether to include animation on the header. This should be false
+     * if the user has `prefers-reduced-motion` opted in. Defaults to false.
+     *
+     * If this prop is specified both here in the Accordion and within
+     * a child AccordionSection component, the Accordion’s animated
```
nishasy, line 68: I actually think we're gonna change all these props so that the AccordionSection's props are prioritized. I'll add this in the following PR.

__docs__/wonder-blocks-accordion/accordion.stories.tsx
```diff
@@ -221,6 +276,134 @@ export const WithInitialExpandedIndex: StoryComponentType = {
     },
 };
 
+/**
+ * An Accordion can be animated using the `animated` prop. This
+ * animation includes the caret, the expansion/collapse, and the last
+ * section's border radius. In this example, animated accordions with
+ * different corner kinds are shown to demonstrate the border radius transition,
+ * as well as accordions with \`allowMultipleExpanded\` set to \`false\`, and
```
jandrade, line 284: **nit:**
```suggestion
 * as well as accordions with `allowMultipleExpanded` set to `false`, and
```

packages/wonder-blocks-accordion/src/components/accordion-section.tsx
```diff
@@ -227,32 +251,51 @@ const AccordionSection = React.forwardRef(function AccordionSection(
...
+                )}
+            </View>
         </View>
     );
 });
 
 const styles = StyleSheet.create({
     wrapper: {
-        flexDirection: "column",
+        // Use grid layout for clean animations.
```
jandrade, line 276: **praise:** Thanks for adding this context here.

__docs__/wonder-blocks-accordion/accordion-section.stories.tsx
```diff
@@ -371,6 +414,29 @@ export const CornerKinds: StoryComponentType = {
...
+ *
+ * If the user has `prefers-reduced-motion` opted in, this animation should
+ * be disabled. This can be done by passing `animated={false}` to
+ * the AccordionSection.
+ *
+ * If `animated` is specified both here in the AccordionSection
+ * and within a parent Accordion component, the Accordion's
+ * `animated` value is prioritized.
+ */
+export const WithAnimation: StoryComponentType = {
```
jandrade, line 430: **suggestion:** Disable this chromatic snapshot as we cannot visually test this in chromatic.

__docs__/wonder-blocks-accordion/accordion.stories.tsx
```diff
@@ -221,6 +276,134 @@ export const WithInitialExpandedIndex: StoryComponentType = {
...
+ *
+ * If the user has `prefers-reduced-motion` opted in, this animation should
+ * be disabled. This can be done by passing `animated={false}` to
+ * the Accordion.
+ *
+ * If `animated` is specified both here in the Accordion
+ * and within a child AccordionSection component, the Accordion's
+ * `animated` value is prioritized.
+ */
+export const WithAnimation: StoryComponentType = {
```
jandrade, line 295: **suggestion:** Disable this chromatic snapshot as we cannot visually test this in chromatic.

__docs__/wonder-blocks-accordion/accordion.stories.tsx
```diff
@@ -282,9 +465,60 @@ SingleSection.parameters = {
     },
 };
 
+/**
+ * This is an example of an Accordion with many sections, as well as
+ * a lot of content within each section.
+ */
+export const LongSections: StoryComponentType = {
```
jandrade, line 472: **suggestion:** Disable this chromatic snapshot as we cannot visually test this in chromatic.

jandrade (approved):
LGTM! The animations look really smooth! 👏. Left some comments to disable the new stories in Chromatic for you to change before landing.

PR #2098, opened 2023-10-25
WB-1634: Add supported themes list to wb-theming

jandrade:
## Summary:

Exports supported themes list to enforce available themes.

This change will allow us to enforce that all themes are supported by the theme
switcher. This will prevent us from accidentally shipping a theme that is not
allowed.

Issue: https://khanacademy.atlassian.net/browse/WB-1634

## Test plan:

Verify that the new types are exported and that the theme switcher context
enforces the supported themes list.

### DIFF
--- a/packages/wonder-blocks-theming/src/types.ts
+++ b/packages/wonder-blocks-theming/src/types.ts
@@ -1,6 +1,6 @@
 import {StyleDeclaration} from "aphrodite";
 
-export type ThemedStylesFn<T> = (theme: T) => StyleDeclaration;
+export type ThemedStylesFn<T extends object> = (theme: T) => StyleDeclaration;
 
 export type SupportedThemes = "default" | "khanmigo";
-export type Themes<T> = Partial<Record<SupportedThemes, T>>;
+export type Themes<T extends object> = Partial<Record<SupportedThemes, T>>;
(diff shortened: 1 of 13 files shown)


jeresig (approved):
Nice!

nishasy (approved):
🎉

packages/wonder-blocks-theming/src/types.ts
```diff
@@ -1,3 +1,6 @@
 import {StyleDeclaration} from "aphrodite";
 
 export type ThemedStylesFn<T> = (theme: T) => StyleDeclaration;
+
+export type SupportedThemes = "default" | "khanmigo";
+export type Themes<T> = Partial<Record<SupportedThemes, T>>;
```
somewhatabstract, line 6: **question:** Is `T` here an object? I wonder if we need to be specific about that with `T extends Object`?

packages/wonder-blocks-theming/src/types.ts
```diff
@@ -1,3 +1,6 @@
 import {StyleDeclaration} from "aphrodite";
 
 export type ThemedStylesFn<T> = (theme: T) => StyleDeclaration;
```
somewhatabstract, line 3: **question:** Does `T` need to be constrained here to object or something?

packages/wonder-blocks-theming/src/types.ts
jandrade: nice catch! I'll add that, and yes, it expects an object.

PR #2102, opened 2023-10-31
Start creating component gallery (sticker sheet) on Wonder Blocks overview page

nishasy:
## Summary:
We want to add a view to the overview page that shows all available
Wonder Blocks components in one location.

I took some design inspiration from the [Material UI gallery](https://m3.material.io/components).

SOME NOTES:

1. The rest of the descriptions will be added in the following PR. This one just includes one
description (Accordion) to demonstrate what that looks like.
2. I made the tiles open the pages in a new tab so that they wouldn't
open inside the iframe.
3. I think there is something wrong with the keyboard nav for the tile headers in Firefox and Safari, but I'm not sure why.
4. There are a few things I wasn't sure if I should add, so I left them out for now:

- Clickable
- Core
- Data
- Layout
- Translations
- Modal building blocks
- Testing
- Timing
- Theming
- Typography

Issue: https://khanacademy.atlassian.net/browse/WB-1595

## Test plan:
- Go to http://localhost:6061/?path=/docs/overview--docs
- Confirm that the tiles look okay (doesn't matter how the actual components look for now).
- Confirm that the hover/focus styles works.
- Confirm that the tiles link to the correct pages.


zhiyi-zhang-duke:
Very nice, excited to see the finished product!

.storybook/components/component-gallery.tsx
jandrade, line 1: **suggestion:** For some reason, Chromatic and Storybook fail to use TurboSnap when we store non-related config files in the `.storybook` folder. The issue with this is that this will cause more snapshots to be generated in Chromatic than we need. We'll probably need to move the other component I created a while ago in that folder too.

https://github.com/Khan/wonder-blocks/actions/runs/6701414235/job/18208805704?pr=2102#step:7:41

Do you think you could move it to a different location? probably `__docs__/components/` would work there 891

__docs__/components/gallery/component-gallery.tsx
```diff
@@ -0,0 +1,71 @@
...
+    },
+    cellExample: {
+        backgroundColor: tokens.color.offWhite,
+        padding: tokens.spacing.large_24,
+    },
+    centerContent: {
+        width: "fit-content",
+        alignSelf: "center",
+    },
+});
```
jandrade, line 71: **suggestion(non-blocking):** It would be probably good moving this to a shared styles file or something like that. It got me confused for a moment that the styles were not used here.

__docs__/wonder-blocks-search-field/search-field.stories.tsx
```diff
@@ -16,7 +16,7 @@ import packageConfig from "../../packages/wonder-blocks-search-field/package.jso
 
 export default {
     component: SearchField,
-    title: "Search Field/SearchField",
+    title: "SearchField",
```
jandrade, line 19: **question:** oh so that means that due to this key missing, Chromatic wasn't snapshotting these files?

jandrade (approved):
Looks really solid!! thanks for improving the docs 👏 🚀

__docs__/wonder-blocks-search-field/search-field.stories.tsx
nishasy: No, I don't think that was the case. I just updated it because I noticed that there was only one component in the search field package, so I updated it to match the other components with the same structure. I noticed it when I was copying the link for it.

nishasy: I see the new snapshots for SearchField and Toolbar though 🤔

PR #2103, opened 2023-11-01
Export `ModalContent` building block from modal package

tony-dinh:
## Summary:
There is a desire to export the `ModalContent` component to give consumers more control when constructing Modals using the building blocks. 

Currently, we always wrap the `ModalPanel`'s `content` prop in a `ModelContent` component if it isn't already wrapped in one (which is typically the case); the `ModelContent` component has pre-defined styles and there is currently no way to customize it, so we export the `ModelContent` component to allow for this as part of the Modal building blocks.

Issue: "none"

Test plan:

None (there is no functionality change)

### DIFF
--- a/.changeset/lucky-jokes-stare.md
+++ b/.changeset/lucky-jokes-stare.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-modal": patch
+---
+
+Exports `ModalContent` component
--- a/packages/wonder-blocks-modal/src/index.ts
+++ b/packages/wonder-blocks-modal/src/index.ts
@@ -1,3 +1,4 @@
+import ModalContent from "./components/modal-content";
 import ModalDialog from "./components/modal-dialog";
 import ModalFooter from "./components/modal-footer";
 import ModalHeader from "./components/modal-header";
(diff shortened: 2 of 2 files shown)


jandrade:
@tony-dinh do you have an example of how `ModalContent` would be used in webapp? trying to understand better the use case/need for this.

tony-dinh:
> @tony-dinh do you have an example of how ModalContent would be used in webapp? trying to understand better the use case/need for this.

Hi @jandrade yes absolutely, this came up in our hackathon project where we are trying to create an experience that looks like:
image

In this modal, we use a `ModalPanel` where we specify the `header`, `footer`, and `content` props.
Without this change.. our use of the content prop would look something like:
```
<ModalDialog>
    <ModalPanel
        header={<ModalHeader />}
        content={
            <View style={styles.body}>
                {...}
            </View>
        }
        footer={
            <ModalFooter />
        }
    />
</ModalDialog>
```

But as is, the `content` will get wrapped in a `ModalContent` which has a pre-defined padding on it and the result would look something closer to:
Screen Shot 2023-11-01 at 3 10 27 PM

If we were able to do something like this instead (where the body styles remove the padding), we'd get our desired results:
```
<ModalDialog>
    <ModalPanel
        header={<ModalHeader />}
        content={
            <ModalContent style={styles.body}>
                {...}
            </ModalContent>
        }
        footer={
            <ModalFooter />
        }
    />
</ModalDialog>
```

kevinbarabash:
@tony-dinh I gave bad advice.  It turns out that you can set `margin: -32` on whatever you pass to the content prop of the `OnePaneDialog`.  Using `ModalDialog` and the building block components isn't actually necessary.  Props to @jandrade for suggesting this alternative solution.

tony-dinh:
> @tony-dinh I gave bad advice. It turns out that you can set margin: -32 on whatever you pass to the content prop of the OnePaneDialog. Using ModalDialog and the building block components isn't actually necessary. Props to @jandrade for suggesting this alternative solution.

Ah yea, I forgot about using negative margins, that should work for this use-case! Thanks @jandrade & @kevinbarabash, I'll go with that approach instead :) At least I got to learn a bit about the WB release process – hope to make a contribution to the package another time!

PR #2105, opened 2023-11-03
Experiment: try using CSS Modules

kevinbarabash:
## Summary:
The purpose of this PR is figure out what the issues would be adopting CSS Modules in wonder-blocks.  This PR updates cell-core.tsx and its dependencies (View, Clickable, and addStyle) to use CSS Modules for styles.

TODO:
- [ ] fix the rollup config to that we can get pass the build step on CI
- [ ] get all of the Cell stories behaving as expected
- [ ] figure out how to get these components to work with styles from either CSS Modules or an Aphrodite stylesheet
- [ ] make sure that the changes to View and Clickable don't break any other components that depend on them

Issue: None

## Test plan:
- TBD


kevinbarabash:
This was just an experiment.

PR #2107, opened 2023-11-08
Add `collapsible` prop to AccordionSection

nishasy:
## Summary
In the event that an accordion section needs to stay open at all times, the user can now use `collapsible={false}` to keep it open. This also sets the header button's `aria-disabled` to `true`, as outlined in the [accessibility guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/).

## Test plan
`yarn jest packages/wonder-blocks-accordion/src/components/__tests__/accordion-section.test.tsx`
`yarn jest packages/wonder-blocks-accordion/src/components/__tests__/accordion-section-header.test.tsx`

Storybook
- Go to http://localhost:6061/?path=/docs/accordion-accordionsection--docs
- Check that there is no caret icon 
- Confirm that the header button cannot be clicked

### DIFF
--- a/packages/wonder-blocks-accordion/src/components/accordion-section-header.tsx
+++ b/packages/wonder-blocks-accordion/src/components/accordion-section-header.tsx
@@ -21,7 +21,7 @@ type Props = {
     // Corner roundedness type.
     cornerKind: AccordionCornerKindType;
     // Whether the section is collapsible or not. If false, the header will
-    // not be clickable.
+    // not be clickable, and the section will stay expanded at all times.
     collapsible?: boolean;
     // Whether the section is expanded or not.
     expanded: boolean;
--- a/packages/wonder-blocks-accordion/src/components/accordion-section-header.tsx
+++ b/packages/wonder-blocks-accordion/src/components/accordion-section-header.tsx
@@ -20,6 +20,9 @@ type Props = {
     caretPosition: "start" | "end";
     // Corner roundedness type.
     cornerKind: AccordionCornerKindType;
+    // Whether the section is collapsible or not. If false, the header will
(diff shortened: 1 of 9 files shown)


packages/wonder-blocks-accordion/src/components/accordion-section-header.tsx
```diff
@@ -20,6 +20,9 @@ type Props = {
     caretPosition: "start" | "end";
     // Corner roundedness type.
     cornerKind: AccordionCornerKindType;
+    // Whether the section is collapsible or not. If false, the header will
+    // not be clickable.
```
jandrade, line 24: **suggestion:** maybe it would be good to include that the contents are visible/expanded all the time?

packages/wonder-blocks-accordion/src/components/accordion-section.tsx
```diff
@@ -53,6 +53,11 @@ type Props = AriaProps & {
      * value is prioritized.
      */
     cornerKind?: AccordionCornerKindType;
+    /**
+     * Whether this section is collapsible. If false, the header will not be
+     * clickable, and the section will be stay expanded at all times.
```
jandrade, line 58: ah you mention it here! so probably worth copy pasting this in the other file.

jandrade (approved):
Looks great! :shipit:

packages/wonder-blocks-accordion/src/components/accordion-section-header.tsx
nishasy: Ah, I didn't include that here since the content panel is not part of the header. I'll still add it now though.

PR #2109, opened 2023-11-10
Add keyboard nav support to Accordion (arrow keys, home, end)

nishasy:
## Summary:
As per the [w3 guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/) for accordion, 
it is recommended (but optional... but we're doing it anyway!) to have keyboard nav support for
arrow up, arrow down, home, and end.

I've added that keyboard support here using the `onKeyDown` listener on `ul` and
`onFocus` on AccordionSection. To get this to work, I changed the ref for AccordionSection
from its container View/div to its header button. I also had to update Clickable because its
`onFocus` prop wasn't working on buttons for some reason.

## Test plan:
Manual testing on Accordion stories
- Go to a story with multiple accordions, such as http://localhost:6061/?path=/story/accordion-accordion--caret-positions
- try keyboard nav in all sort of order + clicking across multiple accordions
- make sure to include tabbing in from elsewhere and then switching to arrow keys, both to the start and the end of the accordion

`yarn jest packages/wonder-blocks-accordion/src/components/__tests__/accordion.test.tsx`

### DIFF
--- a/packages/wonder-blocks-accordion/src/components/__tests__/accordion.test.tsx
+++ b/packages/wonder-blocks-accordion/src/components/__tests__/accordion.test.tsx
@@ -1,5 +1,6 @@
 import * as React from "react";
 import {render, screen} from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
 import {RenderStateRoot} from "@khanacademy/wonder-blocks-core";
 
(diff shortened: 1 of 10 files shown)


packages/wonder-blocks-accordion/src/components/accordion-section.tsx
```diff
@@ -158,7 +164,9 @@ type Props = AriaProps & {
  */
 const AccordionSection = React.forwardRef(function AccordionSection(
     props: Props,
-    ref: React.ForwardedRef<HTMLDivElement>,
+    // Using a button ref here beacuse the ref is pointing to the
+    // section header, which is a button.
+    ref: React.ForwardedRef<HTMLButtonElement>,
```
nishasy, line 169: I guess I could have done this a different way so that the main ref is to the container and there's a separate ref for the header button? I personally liked this approach better because I think it makes sense for people to want a ref to the button itself?

I tried seeing if `useImperativeHandle` would be appropriate here, but the ref types weren't matching up.

packages/wonder-blocks-accordion/src/components/accordion.tsx
```diff
@@ -141,14 +147,59 @@ const Accordion = React.forwardRef(function Accordion(
...
+    const handleKeyDown = (event: React.KeyboardEvent) => {
+        switch (event.key) {
+            // ArrowUp focuses on the previous section.
+            case "ArrowUp":
+                const previousSectionIndex = currentlyFocusedSection - 1;
+                if (previousSectionIndex >= 0) {
+                    const previousChildRef = childRefs[previousSectionIndex];
+                    previousChildRef.current.focus();
+                }
+                break;
```
jandrade, line 177: **suggestion:** I think it would be good to add support to cyclic navigation as per w3's guidelines:
> If focus is on the first accordion header, either does nothing or moves focus to the last accordion header.

packages/wonder-blocks-accordion/src/components/accordion.tsx
```diff
@@ -141,14 +147,59 @@ const Accordion = React.forwardRef(function Accordion(
...
+            // ArrowDown focuses on the next section.
+            case "ArrowDown":
+                const nextSectionIndex = currentlyFocusedSection + 1;
+                if (nextSectionIndex < children.length) {
+                    const nextChildRef = childRefs[nextSectionIndex];
+                    nextChildRef.current.focus();
+                }
+                break;
```
jandrade, line 185: **suggestion:** Similar to the previous comment, cyclic navigation would be helpful too:
> If focus is on the last accordion header, either does nothing or moves focus to the first accordion header.

packages/wonder-blocks-accordion/src/components/accordion.tsx
```diff
@@ -120,13 +120,19 @@ const Accordion = React.forwardRef(function Accordion(
...
     if (initialExpandedIndex !== undefined) {
         startingArray[initialExpandedIndex] = true;
     }
-
     const [sectionsOpened, setSectionsOpened] = React.useState(startingArray);
 
+    // Setting up focus state and refs for keyboard navigation.
+    const [currentlyFocusedSection, setCurrentlyFocusedSection] =
+        React.useState(0);
+    const childRefs = Array(children.length).fill(null);
```
jandrade, line 134: **thought:** One important thing to consider here is to filter out any possible `disabled` accordion sections. Are we including them as part of the focus order?

jandrade (commented):
This is looking great! Leaving some comments with suggestions before doing a final review.

packages/wonder-blocks-accordion/src/components/accordion.tsx
nishasy: Documenting our discussion from a separate meeting: I think we should keep the disabled sections in the focus order since they'd receive focus anyway as part of aria-disabled. Screenreaders should read them out as disabled anyway, so it should still be clear. I'll add a gray outline to make it more obvious for visual users.

packages/wonder-blocks-accordion/src/components/__tests__/accordion.test.tsx
```diff
@@ -397,4 +398,340 @@ describe("Accordion", () => {
...
+                    </AccordionSection>
+                </Accordion>,
+                {wrapper: RenderStateRoot},
+            );
+
+            const button1 = screen.getByRole("button", {name: "Section 1"});
+            const button3 = screen.getByRole("button", {name: "Section 3"});
+
+            // Act
+            button3.focus();
```
jandrade, line 613: **praise:** 🎉 thanks for adding these tests!

packages/wonder-blocks-accordion/src/components/accordion-section-header.tsx
```diff
@@ -217,6 +225,10 @@ const styles = StyleSheet.create({
     disabled: {
         pointerEvents: "none",
         color: "inherit",
+
+        ":focus-visible": {
+            outline: `2px solid ${tokens.color.offBlack32}`,
+        },
     },
```
jandrade, line 232: **suggestion:** It would be great adding support to Safari v14 as well.

For reference: https://github.com/Khan/wonder-blocks/blob/0d875501f8387b6445fc9a77f753f8f0ba23c09b/packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx#L333-L355

packages/wonder-blocks-accordion/src/components/accordion-section.tsx
jandrade: I think this change makes sense to me, specially from the consumer perspective.

packages/wonder-blocks-accordion/src/components/accordion.tsx
jandrade: Awesome thanks! It would be nice adding a summary of what you said here as a code comment :).

packages/wonder-blocks-accordion/src/components/accordion.tsx
```diff
@@ -141,14 +147,64 @@ const Accordion = React.forwardRef(function Accordion(
...
+     */
+    const handleKeyDown = (event: React.KeyboardEvent) => {
+        switch (event.key) {
+            // ArrowUp focuses on the previous section.
+            case "ArrowUp":
+                // Get the previous section, or cycle to last section if
+                // the first section is currently focused.
+                const previousSectionIndex =
+                    (currentlyFocusedSection + children.length - 1) %
+                    children.length;
```
jandrade, line 176: **praise:** modulo operator FTW!

jandrade (approved):
Looks great!! thanks for implementing all the different keyboard behaviors. 🚢

PR #2110, opened 2023-11-10
WB-1636: Add `describedBy` prop to announce the popover dialog contents to screen readers.

jandrade:
## Summary:

The Popover dialog element is not assigning correctly the `aria-describedby`
value, thus causing Screen Readers not being able to read the popover contents
correctly.

The current implementation is trying to reference `aria-describedby` to the
anchor/trigger element (or button that triggers the popover to open/close). This
is incorrect, as the `aria-describedby` value should reference to a particular
content inside the dialog itself.

This PR adds a new prop to the Popover component, `describedBy`, which will
allow the user to pass a string referencing the element that contains the
dialog's contents. This will allow Screen Readers to read the popover contents
correctly.

Issue: https://khanacademy.atlassian.net/browse/WB-1636

## Test plan:

1. Navigate to /?path=/story/popover-popover--described-by.
2. Open the Popover.
3. Verify that the Screen Reader reads the title of the Popover dialog.
4. Change the `describedBy` prop to reference the `content` element by changing the
`describedBy` prop to `content` control in the Storybook.
5. Now verify that the SR reads the contents of the Popover dialog.


packages/wonder-blocks-popover/src/components/popover.tsx
```diff
@@ -233,12 +253,12 @@ export default class Popover extends React.Component<Props, State> {
                     {(props: PopperElementProps) => (
                         <PopoverDialog
                             {...props}
-                            aria-describedby={`${uniqueId}-anchor`}
+                            aria-describedby={`${uniqueId}-${describedBy}`}
```
timmcca-be, line 256: So it looks like this existing `aria-describedby` prop makes it so the popover is described by the content that it is pointing at. I think it would be best not to totally remove that, as we want some way to tie the popover to the content it points at. [Adobe's `react-aria-components` library](https://react-spectrum.adobe.com/react-aria/Popover.html) uses `aria-labelledby` for this, so maybe a solution would be:
```suggestion
                            aria-labelledby={`${uniqueId}-anchor`}
                            aria-describedby={`${uniqueId}-${describedBy}`}
```

packages/wonder-blocks-popover/src/components/popover.tsx
```diff
@@ -56,6 +56,19 @@ type Props = AriaProps &
...
+         * - title: References to the title of the popover.
+         * - content: References to the content of the popover.
+         * - all-content: References to both the title and content of the
+         *   popover.
+         *
+         * This is used to allow screen readers to announce the popover content
+         * when it is opened.
+         * Defaults to "title".
+         */
+        describedBy: "title" | "content" | "all-content";
```
timmcca-be, line 71:
```suggestion
        describedBy: "title" | "content" | "title-and-content";
```

I think this is slightly clearer

timmcca-be (commented):
Thanks so much for looking at this!! One extra consideration: I wonder how we can make this work with `PopoverContentCore` too

packages/wonder-blocks-popover/src/components/popover.tsx
jandrade: In this case pointing `labelledby` to the anchor element would not work as we expect. If we make that, then the dialog would be announced with the contents of the button/anchor element and the intention would be to use that to announce the dialog title instead.

From: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role#associated_aria_roles_states_and_properties

> [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby)
Use this attribute to label the dialog. Often, the value of the aria-labelledby attribute will be the id of the element used to title the dialog.

That makes me think that we can even use that instead of the `describedBy` solution. Let me try that alternative.

jandrade:
Closed as we are going with a different approach added in #2134

PR #2112, opened 2023-11-14
WB-1635: Button - Allow styling the inner label element

jandrade:
## Summary:

The current button doesn’t allow to display the text contents in multiple lines.

This is required so we can remove the truncate functionality for certain cases
like the `ActionBubble` component in webapp. This is needed so we can wrap the
text instead and allow the button’s content to be displayed in two (or more)
lines when needed.

This PR adds a new prop `labelStyle` to the `Button` component that allows to
style the inner label element.

Also changed the `border` property to `outline` to be consistent with the
overall strategy for visual indicatorss in all the WB components. This will also
help us to avoid some layout issues in case folks override the button styles.

Issue: https://khanacademy.atlassian.net/browse/WB-1635

## Test plan:

Verify that the following story works as expected:

/?path=/story/button--custom-styles

Screenshot 2023-11-14 at 5 00 45 PM

By working as expected, it means that the custom button should be displayed
in multiple lines and the text should be wrapped instead of being truncated.

Also verify that the other stories work as expected.


__docs__/wonder-blocks-button/button.stories.tsx
```diff
@@ -591,6 +596,51 @@ TruncatingLabels.parameters = {
...
+            maxWidth: 200,
+            minHeight: 32,
+            height: "auto",
+        },
+        labelStyle: {
+            textOverflow: "initial",
+            whiteSpace: "normal",
+        },
+    },
+    render: (args: any) => (
```
nishasy, line 624: **Praise**: Great time to use args!

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -488,25 +482,12 @@ export const _generateStyles = (
                 },
             },
             focus: {
-                ":after": {
```
nishasy, line 491: **Praise**: I love that we can remove this!

__docs__/wonder-blocks-button/button.stories.tsx
nishasy, line 1: I think the test on line 124 is failing here because of the border/outline update. (It won't let me comment that line directly)

```
await expect(computedStyleButton.borderWidth).toBe("2px");
```

nishasy (approved):
Looks AWESOME! Very excited about the border -> outline update.

Don't forget to update the chromatic interaction test before landing 🙂

__docs__/wonder-blocks-button/button.stories.tsx
jandrade: Thanks! I'll fix that.

PR #2119, opened 2023-11-16
Accordion cleanup and release

nishasy:
## Summary:

All that's left is some polishing, and the Accordion component should be ready for launch!

In this PR:
- Prioritize children's values for overloaded props
- Fix the issue where the bottom border radius was applying to non-last section content when it wasn't supposed to
- Update the Uncontrolled AccordionSection story so it actually says it's uncontrolled
- Add categories to prop tables
- Fixed typo
- Update comments and documentation accordingly
- Make Accordion package.json say that the package is public! 🎉 

## Test plan:

`yarn jest`

Storybook
- Confirm that Accordion's prop table has the props split up into sections now
- Confirm that AccordionSection's prop table has the props split the sections now
- Confirm that all the documentation says that the children override the parents' props
- Go to Accordion's "React Element in Children" story and confirm that the corners don't have overflow.

### DIFF
--- a/.changeset/wet-apricots-wait.md
+++ b/.changeset/wet-apricots-wait.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-accordion": minor
+---
+
+Cleanup: prioritize all children overloaded props, fix typos, update documentation, make the accordion component public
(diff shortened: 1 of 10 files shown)


jandrade (approved):
Everything looks great!! Congrats on finishing this complex component and taking care of all the considerations 👏 🚢 :shipit:  🚀

PR #2123, opened 2023-11-16
Fix font.family.sans token

SeanMcP:
## Summary:
There was an issue with the `font.family.sans` token.

Issue: XXX-XXXX

## Test plan:
- Good question

### DIFF
--- a/.changeset/mean-toes-think.md
+++ b/.changeset/mean-toes-think.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-theming": patch
+---
+
+fix font.family.sans value
--- a/packages/wonder-blocks-theming/src/tokens.ts
+++ b/packages/wonder-blocks-theming/src/tokens.ts
@@ -45,7 +45,7 @@ const tokens = {
     },
     font: {
         family: {
-            sans: 'Lato,"Noto Sans" sans-serif',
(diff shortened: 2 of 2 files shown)


jeresig (approved):
Ah, good catch! cc: @jandrade

jandrade:
>Test plan:
> Good question

😆 This could be tested in this story: ?path=/docs/theming-tokens-typography--docs

By making sure that the CSS prop is applied correctly.

https://5e1bf4b385e3fb0020b7073c-mcxpyiegwx.chromatic.com/?path=/docs/theming-tokens-typography--docs#family

@SeanMcP Thanks for catching this! the only missing part before landing this is to add a changeset. To do this, you'll need to run: `yarn changeset`, then mark `tokens` as a patch version and adding a comment specifying the change. That way, our release tooling is capable to publish this change to npm.

More info here: https://khanacademy.atlassian.net/wiki/spaces/WB/pages/2217640195/Versioning+and+releasing+Wonder+Blocks+packages

jandrade (approved):
🎉 thanks for fixing this!!

SeanMcP:
@jandrade What are the next steps? Just land?

jandrade:
@SeanMcP yeah, land the change and I'll start a release right after that :). And thanks for contributing to Wonder Blocks! 🌯

PR #2128, opened 2023-11-27
WB-1617: Remove Icon component

jandrade:
## Summary:

This removes the Icon component from Wonder Blocks. This is possible now that we
have fully migrated to the new icon system! (Phosphor).

Issue: WB-1617

## Test plan:

Verify that the tests pass and the docs look good.


.changeset/itchy-books-laugh.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-icon": major
+---
+
+Delete Icon component in favor of PhosphorIcon
```
somewhatabstract, line 5: **praise:** ❗ 😮 ❗

packages/wonder-blocks-icon/src/types.ts
```diff
@@ -2,3 +2,8 @@
  * All the possible icon weights.
  */
 export type PhosphorIconAsset = PhosphorRegular | PhosphorBold | PhosphorFill;
+
+/**
+ * All the possible icon sizes.
+ */
+export type IconSize = "small" | "medium" | "large" | "xlarge";
```
somewhatabstract, line 9: **question:** Since in TypeScript 5, all enums are union enums, I wonder if we should do an enum here instead of a string union? 🤷🏻 I don't feel strongly about it. Just asking the question.

somewhatabstract (approved):
Awesome! Congrats!

---
My review uses [Conventional Comments][1] to add context and set expectations for the comments I am leaving.

[1]:https://conventionalcomments.org/

packages/wonder-blocks-icon/src/types.ts
jandrade: That's fair, I'll make the change. Thanks for the suggestion!

jandrade: After quickly making the change, I realized that this would mean that we'd have to replace a bunch of callers to make it compatible with this change. (same would happen with `Button`, `IconButton`, etc).

PR #2136, opened 2023-12-12
WB-1607: Allow displaying the listbox based on the viewport size

jandrade:
## Summary:

This PR changes the logic to display the listbox based on the viewport size.
Before this, we were calculating a fixed height for the listbox based on the
number of items in the listbox. This was causing the listbox to be displayed
partially and the scrollbar was not visible by default.

I'm changing that logic to display the listbox based on the viewport size. This
is done using a PopperJS modifier so we can detect the viewport size and
calculate the height of the listbox accordingly. By implementing this change, we
no longer display a fixed height and instead we display as many items as we can.

Issue: WB-1607

## Test plan:

Navigate to any of the `SingleSelect/MultiSelect` stories and verify that the
listbox is displayed based on the viewport size (resize the screen to see how
the listbox changes its height).

https://github.com/Khan/wonder-blocks/assets/843075/fee493a6-619b-4a16-b5bf-f59a40a6c467

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/dropdown-core.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/dropdown-core.tsx
@@ -955,9 +955,6 @@ class DropdownCore extends React.Component<Props, State> {
                     styles.dropdown,
                     light && styles.light,
                     isReferenceHidden && styles.hidden,
-                    {
-                        maxHeight: "var(--popper-max-height)",
-                    },
                     dropdownStyle,
                 ]}
                 testId="dropdown-core-container"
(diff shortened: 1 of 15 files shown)


packages/wonder-blocks-dropdown/src/components/dropdown-core.tsx
```diff
@@ -961,6 +955,9 @@ class DropdownCore extends React.Component<Props, State> {
                     styles.dropdown,
                     light && styles.light,
                     isReferenceHidden && styles.hidden,
+                    {
+                        maxHeight: "var(--popper-max-height)",
+                    },
```
nishasy, line 960: **Suggestion**: It might help consistency to move this down into the Aphrodite styles.

nishasy (approved):
This is awesome!

packages/wonder-blocks-dropdown/src/components/dropdown-core.tsx
jandrade: Good points! I'll move that to the Aphrodite styles + include a comment mentioning that comes from that file.

PR #2138, opened 2023-12-13
WB-1645.pre: Append className value in Text component (fix)

jandrade:
## Summary:

While working on adding custom option items to Dropdowns, I noticed that the
`className` prop was not being appended to the `Text` component. This PR fixes
that by changing the logic to append the `className` prop to the `Text` if it
includes both a `className` and a `style` prop.

Also updated the typography docs to include an example of using the `style` and
`className` props.

Issue: WB-1645

## Test plan:

Verify that the unit tests pass.

Also verify that the following story works as expected:

http://localhost:6061/?path=/docs/typography--docs#with-style

1. Verify that the text is blue (assigned via `style`).
2. Verify that the has a bg color (assigned via `className` + `css()`).
3. Lastly, verify that it includes a black border (assigned via `className` + and external css import).

Screenshot 2023-12-13 at 11 29 24 AM

### DIFF
--- a/.changeset/stale-steaks-call.md
+++ b/.changeset/stale-steaks-call.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-core": patch
+---
+
+Append className if set via props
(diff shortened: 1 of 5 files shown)


jeresig (approved):
Ahh, tricky! Thanks for the fix.

PR #2142, opened 2023-12-14
Banner: Fix a11y issue by adding role=img to icons

jandrade:
## Summary:

- Banner: Fix a11y issue that caused that icons were not read by screen readers
properly.

- PhosphorIcon: Improved docs to explain how to use the `role` prop +
added unit tests.

Issue: XXX-XXXX

## Test plan:

1. Navigate to the Banner component in the docs and verify that the icons are read
by screen readers properly (e.g. includes `info` before reading the message).

http://localhost:6061/?path=/story/banner--default

2. Navigate to the PhosphorIcon component in the docs and verify that the `role`
prop is documented properly.

http://localhost:6061/?path=/docs/icon-phosphoricon--docs#announcing-the-icon-to-assistive-technology

### DIFF
--- a/.changeset/perfect-chefs-lay.md
+++ b/.changeset/perfect-chefs-lay.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/wonder-blocks-banner": patch
+"@khanacademy/wonder-blocks-icon": patch
+---
+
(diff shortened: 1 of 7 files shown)


jeresig (approved):
Ahh - good fix!

PR #2143, opened 2023-12-15
Accordion: Use z-index instead of adding a 2px margin to bring the header outline in front of the content panel

nishasy:
## Summary:
Currently, there is a 2px added to expanded accordions in order to account for
the outline and stop it from being cut off by the content panel. This is causing
an issue in webapp where these 2px is causing a weird visible gap, so we
want to remove it.

While using z-index here is unfortunate, this is still an appropriate use
for z-index, and it will remove spacing issues.

Example of the gap this 2px margin was causing:

image

Outline cutoff issue that is making the use of z-index necessary:

image

Issue: none

## Test plan:
- Make sure the outline shows up on hover and via keyboard nav in Accordion storybook
- Specifically this story is a good test:
  http://localhost:6061/?path=/story/accordion-accordionsection--react-element-in-header

### DIFF
--- a/.changeset/itchy-penguins-dress.md
+++ b/.changeset/itchy-penguins-dress.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-accordion": patch
+---
+
+Use z-index to put the header outline in front of the content panel instead of adding a 2px margin that can cause weird spacing issues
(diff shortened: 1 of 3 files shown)


jandrade (approved):
🎉

PR #2144, opened 2023-12-15
[fei5403.1.workflowstonode20] Move workflows to node 20

somewhatabstract:
## Summary:
This updates the GitHub workflows to use Node 20 instead of 16.

Issue: FEI-5403

## Test plan:
I verified that with Node 20, I could install node modules and also run a variety of commands:

- `yarn test`
  - Some snapshots needed to be updated due to a change in the error thrown by `JSON.parse`. Instead of `SyntaxError: Unexpected token B in JSON at position 0`, the error looks like `SyntaxError: Unexpected token 'B', "BAD JSON" is not valid JSON`
- `yarn typecheck`
- `yarn lint`


jeresig (approved):
How'd this drop our code coverage by 1%??

somewhatabstract:
> How'd this drop our code coverage by 1%??

I have no idea. Very curious. If we're using Node's built-in coverage tracking, maybe that changed. I can investigate after the main migration of we need to

somewhatabstract:
@jeresig https://github.com/Khan/wonder-blocks/blob/60f8bbd47d749dd53db4702c10d521137d7a41c2/config/jest/test.config.js#L47-L50

We're using the v8 code coverage and I expect there have been some significant changes to that between v16 and v20 of Node. That probably accounts for the change in coverage stats.

jeresig:
@somewhatabstract Ahh, ok - that's a bummer!

PR #2139, opened 2023-12-13
WB-1645: Allow custom option item elements

jandrade:
## Summary:

1. Allow custom `OptionItem` components by using `DetailCell` internally.

- Removed `ClickableBehavior` from `OptionItem` and replaced it with
  `DetailCell` (which internally uses `Clickable`).
- Added stories and included docs for `OptionItem`.

2. Modified Cell to support a new prop required for the `Optionitem` changes:
- `aria-selected` is used to allow the `OptionItem` to be selectable (via aria attributes).

Issue: WB-1645

### SingleSelect:

1. Navigate to http://localhost:6061/?path=/docs/dropdown-singleselect--docs#custom-option-items
2. Verify that the custom option items are rendered as expected.
3. Verify that the custom option items are clickable and that clicking on them
   selects them.

https://github.com/Khan/wonder-blocks/assets/843075/6d171969-cc1e-4922-bbcf-16e23cb8e4d4

### MultiSelect:

1. Navigate to http://localhost:6061/?path=/docs/dropdown-multiselect--docs#custom-option-items
2. Verify that the custom option items are rendered as expected.
3. Verify that the custom option items are clickable and that clicking on them
   selects them.

https://github.com/Khan/wonder-blocks/assets/843075/e60716a7-533c-45c4-b239-c282de7dbaf9

### OptionItem docs:

1. Navigate to http://localhost:6061/?path=/docs/dropdown-optionitem--docs
2. Verify that the documentation for `OptionItem` is correct and up to date.
Screenshot 2023-12-13 at 12 07 11 PM

PR #2146, opened 2023-12-15
[🔥AUDIT🔥] WB-1645.fix Dropdown: Fix regression in SingleSelect

jandrade:
🖍 _This is an audit!_ 🖍

## Summary:

Fixes a regression in `SingleSelect` where the opener is not able to use an
empty string as a label.

In #2139, I changed the logic to use `1 item` for those cases, but turns out
that this "empty" value feature is required by Perseus.

Issue: WB-1645

## Test plan:

Verify that the "Two with Text" example in the `SingleSelect` storybook page
shows empty string as the label for both openers:

http://localhost:6061/?path=/story/dropdown-singleselect--two-with-text

Screenshot 2023-12-15 at 1 52 16 PM

### DIFF
--- a/.changeset/warm-hats-walk.md
+++ b/.changeset/warm-hats-walk.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-dropdown": patch
+---
+
+Fix issue with empty option items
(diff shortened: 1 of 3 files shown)


jeresig (approved):
Ahh - good call!

jeremywiebe (approved):
Thank-you for the quick fix!

PR #2149, opened 2023-12-18
Dropdown: Fix focus state for custom openers

jandrade:
## Summary:

There's a bug in the dropdown component where the focus state is not applied to
custom openers. This PR solves that by adding `tabIndex` to the custom opener so
`ClickableBehavior` can apply the focus state as the element can trigger its
`onFocus`/`onBlur` handlers.

Issue: XXX-XXXX

## Test plan:

Verify that the SingleSelect > Custom opener example displays the focus state
when tabbing to the opener.

https://github.com/Khan/wonder-blocks/assets/843075/f30cfeea-b6b0-427f-a0bd-22a04ff149ef

### DIFF
--- a/__docs__/wonder-blocks-dropdown/single-select.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/single-select.stories.tsx
@@ -6,7 +6,7 @@ import {action} from "@storybook/addon-actions";
 import type {Meta, StoryObj} from "@storybook/react";
 
 import Button from "@khanacademy/wonder-blocks-button";
-import Color, {mix} from "@khanacademy/wonder-blocks-color";
+import Color, {fade} from "@khanacademy/wonder-blocks-color";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {TextField} from "@khanacademy/wonder-blocks-form";
 import {PhosphorIcon} from "@khanacademy/wonder-blocks-icon";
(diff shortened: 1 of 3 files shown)


jeresig (approved):
Ahh, good catch!

__docs__/wonder-blocks-dropdown/single-select.stories.tsx
```diff
@@ -125,6 +126,7 @@ const styles = StyleSheet.create({
         padding: Spacing.medium_16,
     },
     focused: {
+        backgroundColor: mix(Color.lightBlue, Color.offWhite),
```
nishasy, line 129: **Question**: Is this focused style being applied? I do see the outline, but I'm not sure I see a different background color when I focus on the custom opener in the [story](https://5e1bf4b385e3fb0020b7073c-yxpybycqjl.chromatic.com/?path=/story/dropdown-singleselect--custom-opener).

nishasy (approved):
Nice! 🎉

__docs__/wonder-blocks-dropdown/single-select.stories.tsx
jandrade: hmm yeah, I was experimenting with it but forgot to get rid of this. I'll do it before landing.

PR #2152, opened 2023-12-19
Accordion header overflow fix

nishasy:
## Summary:

Currently, there is an issue with a particular accordion section in webapp where long titles in the
custom header cause the accordion contents to overflow. This is fixed by adding `minWidth: 0`
to the styles in Wonder Blocks Accordion Header.

I also included a story in this PR that replicates the issue, which is fixed with the new style.

| Before | After |
| --- | --- |
| Screenshot 2023-12-19 at 3 41 54 PM | Screenshot 2023-12-19 at 3 41 37 PM |

Issue: none

## Test plan:
- Go to the following story: /?path=/story/accordion-accordionsection--react-element-in-header
- Confirm that the second accordion header has an ellipsis when the window is small

### DIFF
--- a/packages/wonder-blocks-accordion/src/components/accordion-section-header.tsx
+++ b/packages/wonder-blocks-accordion/src/components/accordion-section-header.tsx
@@ -159,6 +159,9 @@ const ANIMATION_LENGTH = "300ms";
 
 const styles = StyleSheet.create({
     heading: {
+        // This also needs to have display: grid (matching its container)
+        // so that the header doesn't overflow when the title text is long.
+        display: "grid",
         marginTop: 0,
     },
     headerWrapper: {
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-accordion/src/components/accordion-section-header.tsx
```diff
@@ -159,6 +159,9 @@ const ANIMATION_LENGTH = "300ms";
 
 const styles = StyleSheet.create({
     heading: {
+        // minWidth 0 is necessary to stop custom headers from overflowing
+        // when their content is too long.
+        minWidth: 0,
```
kevinbarabash, line 164: You might also want to include the following in this comment:
> Grid items have an initial size of min-width: auto and min-height: auto.

kevinbarabash (approved):
LGTM.  Thanks for fixing this.

PR #2154, opened 2023-12-22
Pill updates - updated colors, small pills are bigger for a11y

nishasy:
## Summary:
Some updates need to be made to pill so they're more usable
within webapp.

- small size needs 14px font and 24px height
- Banner colors included
- Needs fixed height
- Needs a story for vertically stacked pills

More context for some of these changes can be found in
the following slack thread:
https://khanacademy.slack.com/archives/C8Z9DSKC7/p1693923761646739

Issue: https://khanacademy.atlassian.net/browse/WB-1602

## Test plan:
`yarn jest packages/wonder-blocks-pill/src/components/__tests__/pill.test.tsx`

Storybook
- All variants
  /?path=/story/pill--variants
- Vertically stacked pills
  /?path=/story/pill--vertically-stacked

### DIFF
--- a/packages/wonder-blocks-pill/src/components/__tests__/pill.test.tsx
+++ b/packages/wonder-blocks-pill/src/components/__tests__/pill.test.tsx
@@ -146,10 +146,10 @@ describe("Pill", () => {
         kind          | color
         ${"neutral"}  | ${tokens.color.offBlack8}
         ${"accent"}   | ${tokens.color.blue}
-        ${"info"}     | ${tokens.color.pillBlue}
-        ${"success"}  | ${tokens.color.pillGreen}
-        ${"warning"}  | ${tokens.color.pillGold}
-        ${"critical"} | ${tokens.color.pillRed}
+        ${"info"}     | ${tokens.color.fadedBlue16}
+        ${"success"}  | ${tokens.color.fadedGreen16}
+        ${"warning"}  | ${tokens.color.fadedGold16}
+        ${"critical"} | ${tokens.color.fadedRed16}
     `(
         "renders the correct background color for $kind kind",
         ({kind, color}) => {
--- a/packages/wonder-blocks-pill/src/components/__tests__/pill.test.tsx
+++ b/packages/wonder-blocks-pill/src/components/__tests__/pill.test.tsx
@@ -2,6 +2,8 @@ import * as React from "react";
 import {render, screen} from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 
+import {tokens} from "@khanacademy/wonder-blocks-theming";
+
 import Pill from "../pill";
 
 describe("Pill", () => {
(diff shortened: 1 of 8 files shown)


packages/wonder-blocks-theming/src/tokens.ts
```diff
@@ -25,6 +25,11 @@ const tokens = {
         // Khanmigo
         white32: fade(Color.white, 0.32),
         eggplant: "#5f1e5c",
+        // Pill
+        pillBlue: fade(Color.blue, 0.24),
+        pillGreen: fade(Color.green, 0.24),
+        pillGold: fade(Color.gold, 0.24),
+        pillRed: fade(Color.red, 0.24),
```
nishasy, line 32: FYI: I did check and confirm that all of these pass AA (and AAA) contrast guidelines when they're used as the background color with black text.

nishasy: Also FYI: These colors are based on the colors in `coverage-badge.tsx` in webapp. You can see this being used on Khan Academy here:

https://www.khanacademy.org/standards/CCSS.Math/3.OA

I originally tried using the banner background colors, but I felt like they were just way too light and you couldn't even see the pill.

jandrade: Could you please change the token names here? These new colors represent semantic names and in here we are only using core/primitive tokens.

As for the color values, let me check with Design if these are the colors that we want to officially introduce to WB.

jandrade: We discussed this with the design team, and the plan is to use the 16% option.... we could change to 24% in the future, but the plan is to still with 16% for now, for consistency across the system.

packages/wonder-blocks-pill/src/components/__tests__/pill.test.tsx
```diff
@@ -96,4 +98,74 @@ describe("Pill", () => {
...
+            </Pill>,
+        );
+        const pill = screen.getByTestId("pill-test-id");
+
+        // Assert
+        expect(screen.getByText("Hello, world!")).toBeInTheDocument();
+        // Font size should be at least 14px for accessibility.
+        expect(pill).toHaveStyle({
+            fontSize: 14,
+        });
```
jandrade, line 116: **praise:** Thanks for adding this assertion :accessibility:

packages/wonder-blocks-pill/src/components/__tests__/pill.test.tsx
```diff
@@ -96,4 +98,74 @@ describe("Pill", () => {
...
+
+    test.each`
+        kind          | color
+        ${"neutral"}  | ${tokens.color.offBlack8}
+        ${"accent"}   | ${tokens.color.blue}
+        ${"info"}     | ${tokens.color.fadedBlue16}
+        ${"success"}  | ${tokens.color.fadedGreen16}
+        ${"warning"}  | ${tokens.color.fadedGold16}
+        ${"critical"} | ${tokens.color.fadedRed16}
+    `(
```
jandrade, line 153: **praise:** nice!

packages/wonder-blocks-pill/src/components/pill.tsx
```diff
@@ -147,57 +155,89 @@ const pillStyles = StyleSheet.create({
...
     },
     wrapperSmall: {
-        paddingLeft: Spacing.xSmall_8,
-        paddingRight: Spacing.xSmall_8,
-        borderRadius: Spacing.xxSmall_6,
+        paddingLeft: tokens.spacing.xSmall_8,
+        paddingRight: tokens.spacing.xSmall_8,
+        borderRadius: tokens.spacing.xxSmall_6,
+        // Minimum tap area recommendation for a11y
+        height: tokens.spacing.large_24,
```
jandrade, line 162: **question:** I remember that one of the changes was to make this `20px` tall. So if I understand correctly, we are using 24 instead for the reasons you stated in the comment right? If that's the case, it would be worth checking how that looks in the places where we are hard-coding 20px.

jandrade (approved):
Looks great! thanks for working on improving this! 🚢 🚀 

I only have an open question about 20px vs 24px that I'd like to understand better before landing, but everything looks really solid to me.

packages/wonder-blocks-pill/src/components/pill.tsx
nishasy: I believe that was originally the case, but then we were asked to just bump it up to 24 in this thread:
https://khanacademy.slack.com/archives/C8Z9DSKC7/p1693925664214919?thread_ts=1693923761.646739&cid=C8Z9DSKC7

I can go ahead and use a snapshot release to check that it still looks okay in webapp.

jandrade: Got it, that makes sense, thanks!

PR #2155, opened 2024-01-02
Fix search/filter logic on SingleSelect

jandrade:
## Summary:

After introducing the custom option items, the search/filter logic was broken.
This PR fixes that by using the correct lookup value for the search/filter.

Also added some unit tests and modified some stories to ensure that this
scenario is fully covered for both `SingleSelect` and `MultiSelect`.

Issue: XXX-XXXX

## Test plan:

Verify that the search/filter logic works as expected in the following stories:

- /?path=/story/dropdown-singleselect--virtualized-filterable
- /?path=/story/dropdown-singleselect--custom-option-items-virtualized

https://github.com/Khan/wonder-blocks/assets/843075/0f8fbed5-aaff-4ce2-b5be-cea449d41248

### DIFF
--- a/.changeset/tough-rabbits-cough.md
+++ b/.changeset/tough-rabbits-cough.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-dropdown": patch
+---
+
+Fix filter functionality (SingleSelect)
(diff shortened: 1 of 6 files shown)


jeresig (approved):
Ahhh - good catch, thanks for adding tests!

PR #2157, opened 2024-01-02
Accordion only handles keydown when header focused

nedredmond:
Summary:
Adjusts Accordion element to only handle keydown events when an Accordion header is focused. Previously, an "active header" index would be set, but not unset when the header was blurred. Rather than adding `onBlur` handlers throughout `Clickable`, instead remove active index state and check the `activeElement` from the document at the time a keydown event is fired.

Issue: WB-1656

Test:
Added a story, "With Dropdown" to Accordion docs.

### DIFF
--- a/__docs__/wonder-blocks-accordion/accordion.stories.tsx
+++ b/__docs__/wonder-blocks-accordion/accordion.stories.tsx
@@ -563,7 +563,7 @@ export const WithDropdown: StoryComponentType = {
                     </View>
                 </AccordionSection>
                 <AccordionSection header={`Multi Select`}>
-                    <View style={singleOpened && {height: 200}}>
+                    <View style={multiOpened && {height: 200}}>
                         <MultiSelect
                             selectedValues={values}
                             onChange={setValues}
--- a/__docs__/wonder-blocks-accordion/accordion.stories.tsx
+++ b/__docs__/wonder-blocks-accordion/accordion.stories.tsx
@@ -527,11 +527,6 @@ export const LongSections: StoryComponentType = {
  * This demonstrates how the accordion keyboard interactions do not interfere
  * with the dropdown's keyboard interactions.
  */
-const items = [
-    <OptionItem label="Banana" value="banana" key={0} />,
-    <OptionItem label="Strawberry" value="strawberry" disabled key={1} />,
-    <OptionItem label="Pear" value="pear" key={2} />,
-];
 export const WithDropdown: StoryComponentType = {
     render: function Render() {
         const [value, setValue] = React.useState<any>(null);
@@ -539,6 +534,17 @@ export const WithDropdown: StoryComponentType = {
 
         const [values, setValues] = React.useState<any>([]);
         const [multiOpened, setMultiOpened] = React.useState(false);
+
+        const items = [
+            <OptionItem label="Banana" value="banana" key={0} />,
+            <OptionItem
+                label="Strawberry"
+                value="strawberry"
+                disabled
+                key={1}
+            />,
+            <OptionItem label="Pear" value="pear" key={2} />,
+        ];
 
         return (
             <Accordion animated={true}>
@@ -576,7 +582,7 @@ export const WithDropdown: StoryComponentType = {
             // Disabling because this doesn't test anything visual.
             disableSnapshot: true,
         },
-    }
+    },
 };
 
 LongSections.parameters = {
--- a/__docs__/wonder-blocks-accordion/accordion.stories.tsx
+++ b/__docs__/wonder-blocks-accordion/accordion.stories.tsx
@@ -571,6 +571,12 @@ export const WithDropdown: StoryComponentType = {
             </Accordion>
         );
     },
+    parameters: {
+        chromatic: {
+            // Disabling because this doesn't test anything visual.
+            disableSnapshot: true,
+        },
+    }
 };
 
 LongSections.parameters = {
--- a/__docs__/wonder-blocks-accordion/accordion.stories.tsx
+++ b/__docs__/wonder-blocks-accordion/accordion.stories.tsx
@@ -543,6 +543,7 @@ export const WithDropdown: StoryComponentType = {
         return (
             <Accordion animated={true}>
                 <AccordionSection header={`Single Select`}>
+                    {/* Adding height because overflow hidden in sections. */}
                     <View style={singleOpened && {height: 200}}>
                         <SingleSelect
                             placeholder="Select an option"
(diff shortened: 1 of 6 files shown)


__docs__/wonder-blocks-accordion/accordion.stories.tsx
```diff
@@ -517,6 +522,56 @@ export const LongSections: StoryComponentType = {
...
+                            onToggle={setMultiOpened}
+                        >
+                            {items}
+                        </MultiSelect>
+                    </View>
+                </AccordionSection>
+            </Accordion>
+        );
+    },
+};
```
nishasy, line 573: **Praise**: Thank you for adding this story!

packages/wonder-blocks-accordion/src/components/accordion.tsx
```diff
@@ -162,56 +161,63 @@ const Accordion = React.forwardRef(function Accordion(
...
+
+        const currentlyFocusedHeaderIndex = childRefs.findIndex(
+            (ref) => ref.current === document.activeElement,
+        );
+
+        // If the currently focused element is not a header, do nothing.
+        if (currentlyFocusedHeaderIndex === -1) {
+            return;
+        }
+
```
nishasy, line 187: This is awesome! Do you think you could add unit test(s) for the new logic in `accordion.test.tsx` under the "Keyboard navigation" section? 

Specifically, I think we'd want to add a test for making sure that nothing happens if the focus isn't on a header. We could also check that keyboard nav works for stuff inside the content (like the dropdowns in the story you added) within an accordion.

packages/wonder-blocks-accordion/src/components/accordion-section.tsx
```diff
@@ -120,12 +120,6 @@ type Props = AriaProps & {
      * @ignore
      */
     isRegion?: boolean;
-    /**
-     * Called when the header is focused.
-     * For internal use only.
-     * @ignore
-     */
-    onFocus?: () => void;
```
nishasy, line 128: **Praise**: Thanks for cleaning this up!

nishasy (commented):
Nice!! Your story works great! Just on suggestion to add unit tests, but everything other than that looks good to me.

packages/wonder-blocks-accordion/src/components/accordion.tsx
nedredmond: Tests added!

__docs__/wonder-blocks-accordion/accordion.stories.tsx
```diff
@@ -517,6 +522,57 @@ export const LongSections: StoryComponentType = {
     },
 };
 
+/**
+ * This is an example of an Accordion with a dropdown within each section.
+ * This demonstrates how the accordion keyboard interactions do not interfere
+ * with the dropdown's keyboard interactions.
+ */
```
jandrade, line 529: **suggestion:** Could you please move this comment block before the `export const WithDropdown` line? That way SB will be able to include that in the docs page.

__docs__/wonder-blocks-accordion/accordion.stories.tsx
```diff
@@ -517,6 +522,57 @@ export const LongSections: StoryComponentType = {
...
+                <AccordionSection header={`Multi Select`}>
+                    <View style={singleOpened && {height: 200}}>
+                        <MultiSelect
+                            selectedValues={values}
+                            onChange={setValues}
+                            opened={multiOpened}
+                            onToggle={setMultiOpened}
+                        >
+                            {items}
+                        </MultiSelect>
```
jandrade, line 568: **question:** Is it expected that this `MultiSelect` instance should open on click/Enter/Space? I was trying this out in SB and the SingleSelect works as expected, but MultiSelect never opens.

For reference, I'm using this instance: https://5e1bf4b385e3fb0020b7073c-doqpquuorc.chromatic.com/?path=/story/accordion-accordion--with-dropdown

packages/wonder-blocks-accordion/src/components/__tests__/accordion.test.tsx
```diff
@@ -836,5 +836,38 @@ describe("Accordion", () => {
             expect(button2).not.toHaveFocus();
             expect(button3).toHaveFocus();
         });
+
+        test.each(["{end}", "{home}", "{arrowup}", "{arrowdown}"])(
```
jandrade, line 840: **praise:** Nice set of tests!

packages/wonder-blocks-accordion/src/components/accordion.tsx
```diff
@@ -162,56 +161,63 @@ const Accordion = React.forwardRef(function Accordion(
...
-    /** Keyboard navigation for keys: ArrowUp, ArrowDown, Home, and End.
-     *
-     * Note that we don't have to use `setCurrentlyFocusedSection` in this
-     * function because the focus is handled by the browser + the
-     * section's onFocus handler (the `handleSectionFocus` function above).
+    /**
+     * Keyboard navigation for keys: ArrowUp, ArrowDown, Home, and End.
      */
     const handleKeyDown = (event: React.KeyboardEvent) => {
+        // From https://www.w3.org/WAI/ARIA/apg/patterns/accordion/
```
jandrade, line 173: **praise:** This is great! really helps to future selves to understand why we are doing it this way.

jandrade (approved):
Looks great! just a couple of things for you to consider before landing.

__docs__/wonder-blocks-accordion/accordion.stories.tsx
nedredmond: @jandrade, in that version, it's cut off by the boundaries of the AccordionSection, which has `overflow: hidden`. I pushed a change to extend its height in [fd8dd7e](https://github.com/Khan/wonder-blocks/pull/2157/commits/fd8dd7edff68192b9fcca688b51a6bc1dae652cf), but this seems to be an old version.

__docs__/wonder-blocks-accordion/accordion.stories.tsx
```diff
@@ -517,6 +522,57 @@ export const LongSections: StoryComponentType = {
...
+                    </View>
+                </AccordionSection>
+            </Accordion>
+        );
+    },
```
jandrade, line 573: **suggestion:** Ah one extra thing.... as this story doesn't need to be snapshotted by Chromatic b/c it's more for interactive purposes (keyboard navigation), I think we can skip it from Chromatic.

```suggestion
    },
    parameters: {
        chromatic: {
            // Disabling because this doesn't test anything visual.
            disableSnapshot: true,
        },
    }
```

__docs__/wonder-blocks-accordion/accordion.stories.tsx
jandrade: Gotcha, I now see it working! thanks for the context.

PR #2159, opened 2024-01-04
WB-1638: Fix tabbing in popover

jandrade:
## Summary:
There is a bug in the popover where tabbing out of the popover (while it
is still open) will cause the focus order to be incorrect. More
specifically, the focus will be returned to the trigger element, while
it should be returned to the next element in the tab order.

This is because the popover is not properly handling the case where the
trigger element is the last element in the tab order. This PR fixes that
by making all the focusable elements in the popover non-tabbable, and
then make them tabbable again when the popover gains focus.

For more reference on the actual error, see: https://khanacademy.atlassian.net/browse/CL-1259

https://github.com/Khan/wonder-blocks/assets/843075/ca72e266-9091-409e-9065-3754e4c681c9

Issue: WB-1638

## Test plan:

- Navigate to /?path=/story/popover-popover--keyboard-navigation
- Start tabbing through the elements in the story
- Verify that the focus order is correct
- Open the popover
- Tab through the elements in the popover and verify that the focus
  order is still correct
- Tab backwards through the elements in the popover and verify that the
  focus order is still correct (using shift+tab).
- Close the popover and verify that the focus order is still correct

https://github.com/Khan/wonder-blocks/assets/843075/a8831575-84a8-4bb3-a0f3-11a26d872076

### DIFF
--- a/packages/wonder-blocks-popover/src/components/__tests__/focus-manager.test.tsx
+++ b/packages/wonder-blocks-popover/src/components/__tests__/focus-manager.test.tsx
@@ -1,31 +1,19 @@
 import * as React from "react";
-import * as ReactDOM from "react-dom";
 import {render, screen} from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 
 import FocusManager from "../focus-manager";
-import {findFocusableNodes} from "../../util/util";
 
 describe("FocusManager", () => {
     it("should focus on the first focusable element inside the popover", async () => {
         // Arrange
-        const ref = await new Promise((resolve: any) => {
-            const nodes = (
-                <div ref={resolve}>
-                    <button>Open popover</button>
-                    <button>Next focusable element outside</button>
-                </div>
-            );
-            render(nodes);
-        });
-        // @ts-expect-error [FEI-5019] - TS2345 - Argument of type 'unknown' is not assignable to parameter of type 'ReactInstance | null | undefined'.
-        const domNode = ReactDOM.findDOMNode(ref) as HTMLElement;
-
-        // mock focusable elements in document
-        // eslint-disable-next-line testing-library/no-node-access
-        global.document.querySelectorAll = jest
-            .fn()
-            .mockImplementation(() => findFocusableNodes(domNode));
+        const externalNodes = (
+            <div>
+                <button>Open popover</button>
+                <button>Next focusable element outside</button>
+            </div>
+        );
+        render(externalNodes);
 
         // get the anchor reference to be able pass it to the FocusManager
         const anchorElementNode = screen.getByRole("button", {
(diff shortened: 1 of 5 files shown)


lillialexis (approved):

packages/wonder-blocks-popover/src/components/__tests__/focus-manager.test.tsx
```diff
@@ -109,4 +87,105 @@ describe("FocusManager", () => {
...
+            name: "Open popover",
+        });
+
+        render(
+            <FocusManager anchorElement={anchorElementNode}>
+                <div>
+                    <button>first focusable element inside</button>
+                </div>
+            </FocusManager>,
+        );
```
kevinbarabash, line 113: This is cool.  I didn't realize we had tests that used multiple calls to `render()` in this way.

packages/wonder-blocks-popover/src/components/__tests__/focus-manager.test.tsx
```diff
@@ -109,4 +87,105 @@ describe("FocusManager", () => {
...
+        userEvent.tab();
+
+        // The elements inside the focus manager should not be focusable anymore.
+        const focusableElementInside = screen.getByRole("button", {
+            name: "first focusable element inside",
+        });
+
+        // Assert
+        expect(focusableElementInside).toHaveAttribute("tabIndex", "-1");
+    });
```
kevinbarabash, line 190: **praise:** thanks for labelling what each `userEvent` method call is for.

kevinbarabash: Does tabbing back into the popover making elements inside the popover focusable again?

packages/wonder-blocks-popover/src/components/__tests__/popover.test.tsx
```diff
@@ -356,4 +356,191 @@ describe("Popover", () => {
...
+                <>
+                    <Button>Prev focusable element outside</Button>
+                    <Popover
+                        onClose={jest.fn()}
+                        content={
+                            <PopoverContent
+                                title="Popover title"
+                                content="content"
+                                actions={<Button>Button inside popover</Button>}
+                            />
```
kevinbarabash, line 373: **question:** are there multiple focusable items in this popover content?  If not, could you add another button?

packages/wonder-blocks-popover/src/components/__tests__/popover.test.tsx
```diff
@@ -356,4 +356,191 @@ describe("Popover", () => {
...
+            userEvent.keyboard("{enter}");
+
+            // Wait for the popover to be open.
+            await screen.findByRole("dialog");
+
+            // Focus on the next element after the popover
+            userEvent.tab();
+
+            // Focus on the document body
+            userEvent.tab();
```
kevinbarabash, line 477: **suggestion (non-blocking):** it might good to assert what some of these elements are in addition to the comments.

packages/wonder-blocks-popover/src/components/focus-manager.tsx
```diff
@@ -86,6 +87,11 @@ export default class FocusManager extends React.Component<Props> {
      */
     focusableElementsInPopover: Array<HTMLElement> = [];
```
kevinbarabash, line 88: The name of this field is a little confusing because we're setting these elements to be non-focusable when the focus exits the popover or whatever.  

**suggestion:** consider renaming this to something like `elementsThatCanBeFocusable` or something like that to avoid confusion as to whether the elements in the array are currently focusable or not.

kevinbarabash (approved):
LGTM.  A few minor feedbacks in the comments.  Thanks for fixing this behavior. 🎉

packages/wonder-blocks-popover/src/components/__tests__/focus-manager.test.tsx
jandrade: That's correct.... I make them focusable as soon as the focus goes back to the popover wrapper. One important thing to mention is that I just noticed an issue when the trigger element is the last focusable element in the DOM. I might need to completely change this approach if needed, so wanted to give a heads up.

PR #2162, opened 2024-01-05
WB-1654: Popover trigger element is stealing focus from other focusable elements [FIX]

jandrade:
## Summary:

As soon as a popover is closed, we return focus to its trigger element (or
element that caused the popover to open). This works fine in most cases, but
there are situations where we don’t want to do that.

For example, when there are multiple Popovers within the same context and the
user opens a separate popover, the previous opened popover closes, but in this
case the focus is being moved from the new popover to the trigger element of the
previous popover.

This PR fixes this by detecting if the element that caused the popover to be
dismissed is an element that can be focused, and if so, we let the focus flow
naturally to that element. If not, we return focus to the trigger element.

Issue: https://khanacademy.atlassian.net/browse/WB-1654

## Test plan:

- Open the popover docs page.
- Scroll down to the first story.
- Click on the popover trigger button and verify that the popover opens.
- Now click on the popover trigger button of the second story and wait until
  the first popover closes and the second popover opens.
- Verify that the focus is not being moved to the trigger element of the
  first popover.

| BEFORE | AFTER |
|--------|--------|
| https://github.com/Khan/wonder-blocks/assets/843075/3024ed9c-491e-4963-9b1c-53ae3034870e | https://github.com/Khan/wonder-blocks/assets/843075/a0a2fc3a-9656-4505-a20e-6ee2e795ec83 |

### DIFF
--- a/.changeset/clever-trees-confess.md
+++ b/.changeset/clever-trees-confess.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-popover": patch
+---
+
+Don't return focus to trigger element if the focus has to go to a different interactive element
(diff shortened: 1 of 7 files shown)


jeresig (approved):
oof, this is tricky! thanks for the fix.

PR #2163, opened 2024-01-05
Add `medium` size to Pill

nishasy:
## Summary:
The current `small` pill is too small for a11y, but we
can't make it bigger because there are some areas in
Khan Academy that use unfortunately small font and don't
work well with larger pills.

To remedy this, I'm adding a `medium` pill size that's the
recommended "smaller" size, but `small` is still available
for the smaller text inline pills.

Issue: https://khanacademy.atlassian.net/browse/WB-1602

## Test plan:
`yarn jest`

Go to storybook (/?path=/docs/pill--docs) and confirm that
- the docs make sense
- the interaction test in the Variants story passes
- the dev tools show the correct styles for everything
  in the Variants story


__docs__/wonder-blocks-pill/pill.stories.tsx
```diff
@@ -334,52 +322,28 @@ Variants.play = async ({canvasElement}) => {
...
-    await userEvent.tab();
-    computedStyle = getComputedStyle(warningSmallClickable, ":hover");
-    await expect(computedStyle.outline).toBe("rgb(24, 101, 242) solid 2px");
-
-    await userEvent.tab();
-    computedStyle = getComputedStyle(criticalSmallClickable, ":hover");
-    await expect(computedStyle.outline).toBe("rgb(217, 41, 22) solid 2px");
-
-    await userEvent.tab();
-    computedStyle = getComputedStyle(neutralLargeClickable, ":hover");
```
nishasy, line 362: Removed these because I realized the tests were redundant. It was testing the same style block for all the different sizes.

packages/wonder-blocks-pill/src/components/pill.tsx
```diff
@@ -216,12 +196,46 @@ const _generateColorStyles = (
...
+        default:
+            // Medium size
+            wrapperStyles = {
+                paddingLeft: tokens.spacing.xSmall_8,
+                paddingRight: tokens.spacing.xSmall_8,
+                borderRadius: tokens.spacing.xxSmall_6,
+                // Minimum tap area recommendation for a11y
+                height: tokens.spacing.large_24,
+            };
+    }
```
jandrade, line 228: **suggestion:** Could you please move this back to the previous static `StyleSheet.create()` implementation?

This "dynamic" styles generation is more meant for things like colors as sizes can be more contained into smaller variants.

packages/wonder-blocks-pill/src/components/pill.tsx
```diff
@@ -149,32 +151,10 @@ const Pill = React.forwardRef(function Pill(
...
-        height: tokens.spacing.xLarge_32,
-    },
-});
-
-const _generateColorStyles = (
+const _generateStyles = (
     clickable: boolean,
     kind: PillKind,
-    size: "small" | "large",
+    size: PillSize,
```
jandrade, line 157: **suggestion:** Based on my previous comment, this makes me realize that we don't need the `size` argument here.

jandrade (commented):
The end result looks as expected, but it would be better to revert to how styles were constructed previously before going forward with the PR approval.

jandrade (approved):
Looks great! thanks so much for adding this new variant and applying the recent changes 🙏

PR #2164, opened 2024-01-08
WB-1657: Add support to `closedFocusId` prop in Popover

jandrade:
## Summary:

Allows manually redirecting the focus after the popover is closed/hidden. This
feature already exists in the WB Modal package, and this commit adds support to
the popover package.

The `closedFocusId` prop is a string that represents the id of the element that
should receive focus after the popover is closed. If the element is not found,
the focus will be set to the element that triggered the popover.

This feature has been requested by the Perseus team.

Issue: https://khanacademy.atlassian.net/browse/WB-1657

## Test plan:

- Navigate to /?path=/docs/popover-popover--docs#with-closedfocusid
- Verify that the popover returns focus to the first button after it is closed.

https://github.com/Khan/wonder-blocks/assets/843075/a6ab6c20-2abe-4eec-b6b0-513c2392cc52

### DIFF
--- a/.changeset/strong-lemons-crash.md
+++ b/.changeset/strong-lemons-crash.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-popover": minor
+---
+
+Add `closedFocusId` prop to manually return focus to a specific element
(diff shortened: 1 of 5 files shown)


somewhatabstract (approved):
LGTM

Thanks for adding tests.

PR #2167, opened 2024-01-09
Add `name` prop to SearchField and TextField + add SearchField argtypes

nishasy:
## Summary:
WB SearchField doesn't have a `name` prop, but this is something that is
common on search fields. Adding it here. Also adding argtypes while I'm here.

Issue: https://khanacademy.atlassian.net/browse/WB-1661

## Test plan:
`yarn jest packages/wonder-blocks-form/src/components/__tests__/text-field.test.tsx`
`yarn jest packages/wonder-blocks-search-field/src/components/__tests__/search-field.test.tsx`

Check argtypes in search field and confirm that all the types and default
values are correct. Also confirm that the demo works as expected when the
controls values are changed.


packages/wonder-blocks-form/src/components/text-field.tsx
```diff
@@ -95,6 +106,10 @@ type Props = AriaProps & {
      * Custom styles for the input.
      */
     style?: StyleType;
+    /**
+     * Optional CSS classes for the entire dropdown component.
+     */
+    className?: string;
```
jandrade, line 112: question: is there a particular reason why `className` needs to be included? We tend to use `style` in our WB components for that purpose, but I'm wondering if we are missing something. Same question applies to `SearchField`.

nishasy: Ah, you're right. I thought it was being used for something else, but it's just styles.

jandrade (approved):
Looks great!! thanks for adding this 🏅

PR #2168, opened 2024-01-11
Accordion keyboard nav: stop nav keys from causing page scroll

nishasy:
Currently, pressing the home, end, and arrow keys causes
scrolling on the page (the default behavior of those keys).

To fix this, I added `preventDefault()` for all the keys.

While I'm updating accordion, I also added a comment to the
`WithAnimation` stories warning people about the performance
impact of using height animations.

Issue: none

## Test plan:
- Go to local accordion stories docs tab.
- Click just before the "Default" story accordion.
- Press tab so that the focus shifts to the first section header.
- Press the end key and confirm that the page does not jump
  to the bottom. Confirm that it focuses on the last section header.
- Press the home key and confirm that the page does not jump
  to the top. Confirm that it focuses on the first section header.
- Press the up arrow key repeatedly. Confirm that it cycles through
  the accordion sections and does not slowly scroll up the page.
- Press the down key repeatedly. Confirm that it cycles through
  the accordion sections and does not slowly scroll down the page.

### Demo: Before and after

https://github.com/Khan/wonder-blocks/assets/13231763/829993a2-a484-4987-8d71-54bdd0cc338c

### DIFF
--- a/.changeset/fifty-countries-fry.md
+++ b/.changeset/fifty-countries-fry.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-accordion": patch
+---
+
+Prevent default in home, end, and arrow keys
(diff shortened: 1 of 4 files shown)


jandrade (approved):
Looks great! Nice catch 🚀

PR #2171, opened 2024-01-31
WB-1660: Create `wb-codemod` package

jandrade:
## Summary:

This PR creates a new package `wb-codemod` that will be used to run code
migrations using codemods. The package is a wrapper around `jscodeshift` that
provides a CLI to run WB codemods from different packages/locations.

This will be useful now that Web Wonder Blocks is being actively used in the
main repo and in Perseus as well.

The goal/motivation is for anyone to run a particular codemod without having to
install jscodeshift or the codemod package itself locally.

```
npx @khanacademy/wb-codemod -t <transform-file-name> <path>
```

e.g.
```
npx @khanacademy/wb-codemod -t "template" src/
```

Issue: WB-1660

## Test plan:

Verify that unit tests pass and that the package can be used to run codemods.

If you want to run this locally while the package is not published yet, you can
use the following command:

```
cd wb-codemod
npx . -t "template" ../packages/wonder-blocks-button/
```

### DIFF
--- a/wb-codemod/src/wb-codemod.ts
+++ b/wb-codemod/src/wb-codemod.ts
@@ -41,12 +41,14 @@ export async function run(
             babel: true,
             extensions: "js,jsx,ts,tsx",
             parser: "tsx",
-            runInBand: true,
+            runInBand: !!options.dryRun,
             silent: false,
             ignorePattern: ["**/node_modules/**", "**/dist/**"],
             // Allows to format the transformed code.
             printOptions: {
                 objectCurlySpacing: false,
+                quote: "double",
+                trailingComma: true,
             },
         });
 
--- a/wb-codemod/src/wb-codemod.ts
+++ b/wb-codemod/src/wb-codemod.ts
@@ -1,0 +1,58 @@
+/* eslint-disable no-console */
+import path from "path";
+import chalk from "chalk";
+import jscodeshift from "jscodeshift/src/Runner";
+
+type RunOptions = {
+    /**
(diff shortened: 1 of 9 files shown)


wb-codemod/src/wb-codemod.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    try {
+        const result = await jscodeshift.run(transformFile, filePaths, {
+            dry: options.dryRun || false,
+            print: options.print || false,
+            verbose: 2,
+            babel: true,
+            extensions: "js,jsx,ts,tsx",
+            parser: "tsx",
+            runInBand: true,
```
jeresig, line 44: Why are you doing this? IMO, this makes it a lot slower. You could make this `true` if `dryRun` is true, instead?

wb-codemod/src/wb-codemod.ts
```diff
@@ -0,0 +1,58 @@
...
+            silent: false,
+            ignorePattern: ["**/node_modules/**", "**/dist/**"],
+            // Allows to format the transformed code.
+            printOptions: {
+                objectCurlySpacing: false,
+            },
```
jeresig, line 50: Could add in:

```
        quote: "double",
        trailingComma: true,
```

jeresig (approved):
Nice - this should be handy, thank you!

wb-codemod/src/wb-codemod.ts
jandrade: Good point! I'll change that.

PR #2176, opened 2024-02-01
Popover: Add reason param to onClose callback

rjcorwin:
This diff adds a `reason` param to Popover component's `onClose` callback. Popovers are closed for the following reasons:

1. The dismiss button is clicked. In this case the reason is "dismissed".
2. The user pressed escape.  In this case the reason is "dismissed".
3. The user clicked outside of the scope of the Popover. In this case the reason is "blur".
4. The Popover is opened again, in which case the the reason is currently "unknown". There is possibly room for improvement here, please see TODO.


rjcorwin:
Hi @jandrade! I'm going to post a thread and tag you for some context on this PR.

PR #2177, opened 2024-02-01
Add onMouseDown and onMouseUp props to Clickable

nishasy:
## Summary:

Currently, Clickable does not have `onMouseDown` or
`onMouseUp` as props. Up until now, it has not been
an issue since people can just use `onClick`. However,
there is a case in Perseus that would benefit from
having an `onMouseDown` prop here - preventing 
default in `onMouseDown` would allow the focus to stay
on the math input when keypad buttons (clickable) are
clicked.

In this PR, I added `onMouseDown` and `onMouseUp`
to Clickable. (I don't need `onMouseUp`, but I figured
it makes sense to have if `onMouseDown` is there for
symmetry.)

Issue: none

## Test plan:
`yarn jest packages/wonder-blocks-clickable/src/components/__tests__/clickable.test.tsx`


jandrade (approved):
Looks good to me! Thanks for adding the tests.

One thing before landing... could you please include argTypes to both `clickable` and `clickable-behavior`?

You can see how `onkeyDown` is set as reference.

PR #2186, opened 2024-02-21
WB-1665: MultiSelect: Don't select disabled items when select all is selected

jandrade:
## Summary:

We are currently selecting disabled items when the "Select all" option is
selected. Even that this is not a common use case, it is still a bug in the
sense that it doesn't match how `<select>` works.

In this PR we are trying to match the behavior of `<select>` by not selecting
disabled option items when "Select all" is selected.

Another change is that we are counting the number of available options to
select in the `Select All (n)` label.

Issue: WB-1665

## Test plan:

1. Navigate to /?path=/story/dropdown-multiselect--shortcuts.
2. Open the dropdown and select "Select all".
3. Verify that the disabled items are not selected.

https://github.com/Khan/wonder-blocks/assets/843075/199ae0d9-f59e-4ffe-bc15-fc1da44b9555

### DIFF
--- a/.changeset/pink-islands-grin.md
+++ b/.changeset/pink-islands-grin.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-dropdown": patch
+---
+
+Don't select disabled items using the "Select all" shortcut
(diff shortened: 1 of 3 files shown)


jeresig (approved):
Whoops - good catch, thank you!

kevinbarabash (commented):
LGTM 🚀

PR #2187, opened 2024-02-21
WB-1666: OptionItem not vertically centered when horizontalRule is set (fix)

jandrade:
## Summary:

There is a visual issue when the `horizontalRule` prop is set to `true` in the
`OptionItem` component. The `OptionItem` component is not vertically centered
as causes to include a padding after the horizontal rule.

This PR fixes that by applying the correct padding to the `OptionItem` component
to make it vertically centered when the `horizontalRule` prop is set.

Issue: WB-1666

## Test plan:

1. Navigate to /?path=/story/dropdown-singleselect--custom-option-items.
2. Focus on different option items by pressing `Tab`.
2. Verify that the `OptionItem` component is vertically centered and the
horizontal rule is correctly aligned to the bottom of the component.

Alternatively, navigate to /?path=/story/dropdown-optionitem--horizontal-rule and verify the alignment.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2024-02-15 at 3 40 27 PM (1) | Screenshot 2024-02-21 at 12 00 58 PM |

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/option-item.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/option-item.tsx
@@ -235,8 +235,6 @@ const styles = StyleSheet.create({
         // Reset the default styles for the cell element so it can grow
         // vertically.
         minHeight: "unset",
-        // Make sure that the item is always at least as tall as 40px.
-        paddingBlock: spacing.xxxxSmall_2,
 
         /**
          * States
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-dropdown/src/components/option-item.tsx
```diff
@@ -314,7 +312,10 @@ const styles = StyleSheet.create({
     },
     itemContainer: {
         minHeight: "unset",
-        padding: spacing.xSmall_8,
+        // Make sure that the item is always at least as tall as 40px.
```
jandrade, line 315: **note:** By moving the `2px` to the container, we can ensure that the horizontal rule is fully positioned at the bottom of the cell.

Before that, we where adding a `bottom` padding of 2px after the hr.

jeresig (approved):
Ahh good fix, thank you!

PR #2189, opened 2024-02-27
WB-1668: Tokens: Add 24% color shades

jandrade:
## Summary:

- Add some new color tokens to existing primary colors. These new tokens are 24%
lighter than the original ones, and will be used for some cards in the WHAM
project.

- Also fixed a very small issue with the existing faded colors, which were not
including a mix with white. The change is minimal, but still worth as it allows
to convert the shade from an alpha value to a solid color.

Issue: https://khanacademy.atlassian.net/browse/WB-1668

## Test plan:

Verify that the new colors are included and that the snapshots look correct.

Screenshot 2024-02-23 at 3 54 31 PM

### DIFF
--- a/.changeset/rotten-cheetahs-guess.md
+++ b/.changeset/rotten-cheetahs-guess.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": minor
+---
+
+Add new shades of existing colors (24%).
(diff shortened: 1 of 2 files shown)


jandrade:
@kevinbarabash The idea is to use these new `fadedX24` versions in the WHAM cards (maybe there are a few cases where we could use `fadedX16` instead.

jeresig (approved):
Nice and pastel-y!

PR #2192, opened 2024-02-28
Add 'contentStyle' prop to DetailCell

kevinbarabash:
## Summary:
We need a way to align content to the top the content container so that we can can implement cards with all items aligned to the top.  This PR adds a 'contentStyle' prop to provide this functional.  It's similar to the existing leftAccessoryStyle and rightAccessoryStyle props.

I also added a 'start:storybook' script to package.json that matches the command we use in webapp.  'yarn start' will still work to start Storybook.

Issue: None

## Test plan:
- yarn start:storybook
- view the DetailCell story for custom styles, see that the title is top-aligned like the two accessories are

Screen Shot 2024-02-28 at 10 11 16 AM


__docs__/wonder-blocks-cell/detail-cell.stories.tsx
```diff
@@ -123,8 +123,9 @@ DetailCellDisabled.parameters = {
 export const DetailCellWithCustomStyles: StoryComponentType = () => (
     <DetailCell
         title="Title for article item"
-        subtitle1="Subtitle for article item"
-        subtitle2="Subtitle for article item"
+        contentStyle={{
+            alignSelf: "flex-start",
+        }}
```
jandrade, line 128: **suggestion:** It would be nice adding a short description that this prop can be used in `docs.description.story`.

__docs__/wonder-blocks-cell/detail-cell.stories.tsx
jandrade, line 1: **suggestion:** Could you please add the arg description in compact-cell.argtyles.ts?

You can use this as reference: https://github.com/Khan/wonder-blocks/blob/56a896c6b6b1169360ad1b6e63638a8cc1d8a593/__docs__/wonder-blocks-cell/compact-cell.argtypes.tsx#L114-L127

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -127,7 +128,7 @@ function CellInner(props: CellCoreProps): React.ReactElement {
             />
 
             {/* Cell contents */}
-            <View style={styles.content} testId={testId}>
+            <View style={[styles.content, contentStyle]} testId={testId}>
```
jandrade, line 131: **suggestion:** It would be helpful adding a unit test for this case. For reference: https://github.com/Khan/wonder-blocks/blob/860d9ef9358a4bf5b8c27126c2e0d047d62a7642/packages/wonder-blocks-cell/src/components/internal/__tests__/cell-core.test.tsx#L128-L140

jandrade (approved):
The implementation looks great! I left some comments for adding some docs and unit test before landing. Thanks for adding this feature, it will come very handy in some situations 🚢

PR #2193, opened 2024-02-28
WB-1667: Submit event bubbling on button[type=submit] is incorrect in React 17.0.2

jandrade:
## Summary:

This PR updates `ClickableBehavior` (that is internally used by the `Button`
component) so we can fix an issue with the bubbling of the `submit` event.

This was causing forms not to submit when a `Button` component was used as a
submit button, due to some event changes in React 17.0.2.

Issue: WB-1667

## Test plan:

Verify that the tests fail without the changes and pass with the changes.

1. 🔴 Test Failure: https://github.com/Khan/wonder-blocks/actions/runs/8087704244/job/22100292494?pr=2193#step:5:17
2. 🟢 Test succeeded: https://github.com/Khan/wonder-blocks/actions/runs/8087764486/job/22100453501?pr=2193#step:5:17


jandrade:
@jeresig In order to test that this would work with React v17 (without having to actually upgrade to that version), I created 3 different commits:
- The first one included only the upgrade to show the test failure.
- The second included the actual fix + the test passing.
- The third reverted the first commit.

You can see the results of the tests in the PR description.

jeresig (approved):
Awesome!! Thank you, Juan.

PR #2195, opened 2024-02-29
Upgrade user-event to v14, update all tests.

jeresig:
## Summary:
This gets all the tests working with v14. Unfortunately, I had to skip a lot of tests, they are much more complicated than the ones in Webapp.

1) There are the keyboard-using ones that are failing, same as webapp. I have an issue to track those.
2) There are tests that are no longer working due to the focus state behaving differently.
3) There are tests failing due to issues with element visibility and findByRole not finding them.
4) There are tests failing that were checking for an un-resolved promise, which no longer works as promises are now resolved during renders.

I assume that Juan will probably want to do another pass on these at some point to make some improvements, but at least this un-blocks us from using user-event v14.

Issue: WB-1266

## Test plan:
Tests are passing!

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,11 +63,10 @@
     "@storybook/testing-library": "^0.2.2",
     "@swc-node/register": "^1.6.5",
     "@swc/core": "^1.3.36",
-    "@testing-library/dom": "^8.19.0",
     "@testing-library/jest-dom": "^5.16.5",
-    "@testing-library/react": "^12.1.5",
-    "@testing-library/react-hooks": "^8.0.1",
-    "@testing-library/user-event": "^13.5.0",
+    "@testing-library/react": "^12.1.2",
+    "@testing-library/react-hooks": "^7.0.2",
+    "@testing-library/user-event": "^14.5.1",
     "@types/jest": "29",
     "@types/jscodeshift": "^0.11.11",
     "@types/node": "^18.14.1",
@@ -102,7 +101,7 @@
     "fast-glob": "^3.2.12",
     "jest": "^29.5.0",
     "jest-date-mock": "^1.0.8",
-    "jest-environment-jsdom": "^28.1.2",
+    "jest-environment-jsdom": "^29.0.2",
     "jest-extended": "^3.2.4",
     "jscodeshift": "^0.15.1",
     "npm-package-json-lint": "^6.4.0",
@@ -142,6 +141,7 @@
   "resolutions": {
     "@types/react": "16",
     "@types/react-dom": "16",
+    "aria-query": "5.3.0",
     "strip-ansi": "6.0.1",
     "strip-ansi-explanation": "There's an issue with strip-ansi v7 which causes conflicts with the Khan/changeset-per-package action"
   }
--- a/package.json
+++ b/package.json
@@ -139,9 +139,11 @@
     "wb-codemod"
   ],
   "resolutions": {
+    "@testing-library/dom": "^8.0.0",
+    "@testing-library/jest-dom": "^5.16.5",
+    "@testing-library/user-event": "^14.5.1",
     "@types/react": "16",
     "@types/react-dom": "16",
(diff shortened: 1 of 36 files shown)


package.json
```diff
@@ -142,7 +141,8 @@
   "resolutions": {
     "@types/react": "16",
     "@types/react-dom": "16",
+    "aria-query": "5.3.0",
```
jandrade, line 144: **suggestion:** What do you think about adding a comment explaining why we had to pin this version? (similar to `strip-ansi` below).

packages/wonder-blocks-accordion/src/components/__tests__/accordion.test.tsx
```diff
@@ -175,20 +175,24 @@ describe("Accordion", () => {
...
 
-    test("Styles the corners based on the square cornerKind", () => {
+    test("Styles the corners based on the square cornerKind", async () => {
         // Arrange
         render(
             <Accordion cornerKind="square">
                 {[
-                    <AccordionSection header="Title" testId="section-test-id">
+                    <AccordionSection
+                        key="section-id"
```
jandrade, line 192: **question:** Does `render` throw a warning about keys now? not sure if this was happening before, but you just uncovered (the warning makes sense btw).

packages/wonder-blocks-birthday-picker/src/components/__tests__/birthday-picker.test.tsx
```diff
@@ -534,27 +568,31 @@ describe("BirthdayPicker", () => {
             jest.useFakeTimers();
         });
 
-        it("should find and select an item using the keyboard", () => {
+        it("should find and select an item using the keyboard", async () => {
             // Arrange
+            const ue = userEvent.setup({
+                advanceTimers: jest.advanceTimersByTime,
+                pointerEventsCheck: PointerEventsCheckLevel.Never,
+            });
```
jandrade, line 576: This is cool! By reading the docs, I thought that we always needed to use `setup`, but it looks like it is only needed when we need to create an instance with certain settings.

packages/wonder-blocks-dropdown/src/components/__tests__/action-menu.test.tsx
```diff
@@ -51,14 +55,18 @@ describe("ActionMenu", () => {
...
         // Assert
-        expect(screen.getByRole("menu")).toBeInTheDocument();
+        expect(
+            await screen.findByRole("menu", {hidden: true}),
+        ).toBeInTheDocument();
     });
 
-    it("closes itself on escape", () => {
+    // TODO(FEI-5533): Key press events aren't working correctly with
+    // user-event v14. We need to investigate and fix this.
```
jandrade, line 68: **question:** What's happening in this case? is the focus order being set incorrectly?

jandrade (approved):
Looks great! Thanks a lot for working on this migration :shipit:.

Hopefully the `userEvent.keyboard` issues will be solved in the near future 🤞.

packages/wonder-blocks-accordion/src/components/__tests__/accordion.test.tsx
jeresig: I'm not sure if it's new, or not! But I noticed it so decided to clean it up while I was in there...

package.json
jeresig: Oh yeah, honestly, we probably don't even need this change, it turned out to not matter. I'll remove it.

packages/wonder-blocks-birthday-picker/src/components/__tests__/birthday-picker.test.tsx
jeresig: Yeah, exactly! The default instance works well for us 99% of the time - it turns out that WB is particularly complex, with the mocked timers and pointer checks.

PR #2200, opened 2024-03-28
Automatically format the number for the current locale in ngettext.

jeresig:
## Summary:
We can't support a custom-formatted num with ngettext when we move to LinguiJS, so we're removing that logic and making it so that the numbers are now formatted using NumberFormat instead (which matches what LinguiJS will do).

I'll have to make a few changes in webapp to remove the few cases where this is happening, so I will release this change at the same time as I release those changes in webapp.

Issue: FEI-5547

## Test plan:
The tests should pass.

### DIFF
--- a/packages/wonder-blocks-i18n/src/functions/i18n.ts
+++ b/packages/wonder-blocks-i18n/src/functions/i18n.ts
@@ -5,6 +5,7 @@ import * as React from "react";
 
 import FakeTranslate from "./i18n-faketranslate";
 import {allPluralForms} from "./plural-forms";
+import {getLocale} from "./locale";
 
 type InterpolationOptions<T> = {
     [key: string]: T;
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-i18n/src/functions/i18n.ts
```diff
@@ -260,12 +261,23 @@ export const ngettext: ngettextOverloads = (
 
     // Get the options to substitute into the string.
     // We automatically add in the 'magic' option-variable 'num'.
-    actualOptions.num = actualOptions.num || actualNum;
+    actualOptions.num = formatNumber(actualNum);
```
somewhatabstract, line 264: **question:** I am confused by how this change relates to custom formatting - perhaps it's down to the test examples only using numbers and not strings. Basically, the `num` option could override what number was actually getting formatted. Is this being used in webapp code to pass a string instead of a number, and therefore do custom formatting?

I'm all for removing the extra confusing option - I just want to understand how that impacts formatting.

somewhatabstract (approved):
This change looks fine to me although I wanted some clarification on how this relates to custom formatting. Perhaps a clearer example - the test cases from before just don't seem to correlate with custom formatting, only with number substitution (which seems odd in the first place - here's a 2, but I put a 5 in the options, so it's a 5 - not sure what's going on there).

packages/wonder-blocks-i18n/src/functions/i18n.ts
jeresig: Yes, it was being used to pass in a string - there were test cases for this here in wonder-blocks-i18n that were testing this case, too, but I removed them as part of this change. So now there will be no way to have a custom format of the string, it'll just be formatted with `Intl.NumberFormat` instead.

somewhatabstract: Yeah, I saw the deleted test cases; it's just they always used a primitive number so I didn't get that you could just pass a string. Thanks for clarifying.

PR #2201, opened 2024-04-01
[fei5384.1.medialayout] Make sure MediaLayout respects Server.isServerSide()

somewhatabstract:
## Summary:
I was working on addressing the source of some hydration issues in webapp and discovered the `MediaLayout` component was using a heuristic to determine SSR that meant it was not doing the SSR behavior in our route tests.

I addressed this from the webapp side by deleting the `matchMedia` function when doing our route testing, but I also wanted to fix the core issue here by ensuring the `MediaLayout` component utilizes our central `Server.isServerSide()` function from Wonder Blocks Core.

This PR does that and also modifies the tests acccordingly. I also did some tweaks to the testing to make use of `describe.each` and reduce some nesting.

NOTE: I am not convinced the `MediaLayout` component hydrates properly since I don't see how it guarantees the initial client-side render will match the SSR'd result. However, addressing that is not a goal of this change. If that is an issue, a separate ticket should be logged to address it.

Issue: FEI-5384

## Test plan:
`yarn test`

### DIFF
--- a/packages/wonder-blocks-layout/src/components/__tests__/media-layout.test.tsx
+++ b/packages/wonder-blocks-layout/src/components/__tests__/media-layout.test.tsx
@@ -5,11 +5,11 @@ import {render, screen} from "@testing-library/react";
 
 import MediaLayout from "../media-layout";
 import {resizeWindow, matchMedia} from "../../util/test-util";
+import {MediaSize} from "../../util/types";
 
 describe("MediaLayout", () => {
     beforeEach(() => {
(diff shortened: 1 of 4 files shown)


packages/wonder-blocks-layout/src/components/__tests__/media-layout.test.tsx
```diff
@@ -78,185 +76,177 @@ describe("MediaLayout", () => {
...
             expect(args.mediaSize).toEqual("large");
         });
     });
 
-    describe("styleSheets", () => {
-        const testSizes = {
-            small: [640, 480],
-            medium: [800, 600],
-            large: [1200, 800],
-        } as const;
```
somewhatabstract, line 91: The size arrays were not being used. I changed this to a `describe.each` of the media sizes instead of this data type and the `for of` loops.

jeresig (approved):
Great, thank you Jeff!

PR #2208, opened 2024-04-15
WB-1681: Tooltip: Add `autoUpdate` prop to force updating the tooltip position

jandrade:
## Summary:

Adds a new `autoUpdate` prop to update the tooltip position when the trigger element changes.

This is useful for cases where the trigger element changes its position after
the tooltip is shown, like when the trigger element is a draggable element.

PopperJS does not update the tooltip position when the trigger element changes
by default, so we need to force the update.

After different approaches, I decided to use `MutationObserver` to listen to
changes in the trigger element and update the tooltip position when needed.

**NOTE:** The newer version of PopperJS (floating-ui) have an `autoUpdate`
middleware that does the same thing, but it's not available in the version we
are using. This prop uses the same naming to be consistent with the newer
version (if we decide to upgrade in the future).

Issue: XXX-XXXX

## Test plan:

Navigate to the tooltip storybook and test the new `autoUpdate` prop.

/?path=/story/tooltip-tooltip--auto-update

https://github.com/Khan/wonder-blocks/assets/843075/ca2e592e-b144-447a-9a64-ec5ca6a9751d

### DIFF
--- a/__docs__/wonder-blocks-tooltip/tooltip.stories.tsx
+++ b/__docs__/wonder-blocks-tooltip/tooltip.stories.tsx
@@ -351,6 +351,79 @@ WithStyle.play = async ({canvasElement}) => {
     });
 };
 
+/**
+ * Tooltip by default (and for performance reasons) only updates its position
+ * under the following conditions:
+ *
+ * 1. When the window is resized.
+ * 2. When the scroll position changes.
+ *
+ * However, there are cases where you might want the tooltip to update its
+ * position when the trigger element changes. This can be done by setting the
+ * `autoUpdate` prop to `true`.
+ */
+export const AutoUpdate: StoryComponentType = {
+    render: function Render() {
+        const [position, setPosition] = React.useState<{
+            x: number;
+            y: number;
+        } | null>(null);
+        return (
+            <View style={[styles.centered, styles.row, {position: "relative"}]}>
+                <Tooltip
+                    content="This is a tooltip that auto-updates its position when the trigger element changes."
+                    opened={true}
+                    autoUpdate={true}
+                >
+                    <View
+                        style={[
+                            position && {
+                                position: "absolute",
+                                top: position.y,
+                                left: position.x,
+                            },
+                        ]}
+                    >
+                        Trigger element
+                    </View>
+                </Tooltip>
+                <Button
+                    onClick={() => {
+                        setPosition({
+                            x: Math.floor(Math.random() * 200),
+                            y: Math.floor(Math.random() * 200),
+                        });
+                    }}
+                >
+                    Click to update trigger position
+                </Button>
+            </View>
+        );
+    },
+    play: async ({canvasElement}) => {
+        // Arrange
+        const canvas = within(canvasElement.ownerDocument.body);
+
+        // Get HTML elements
+        const tooltip = await canvas.findByRole("tooltip");
+        const initialLeft = tooltip.getBoundingClientRect().left;
+        const initialTop = tooltip.getBoundingClientRect().top;
+
+        // Act
+        await userEvent.click(canvas.getByRole("button"));
+
+        // Wait for the tooltip to update its position
+        const newTooltip = await canvas.findByRole("tooltip");
+        const newLeft = newTooltip.getBoundingClientRect().left;
+        const newTop = newTooltip.getBoundingClientRect().top;
+
+        // Assert
+        // The tooltip should have updated its position
+        await expect(initialLeft).not.toEqual(newLeft);
+        await expect(initialTop).not.toEqual(newTop);
+    },
+};
+
 const styles = StyleSheet.create({
     storyCanvas: {
         // NOTE: This is needed for Chromatic to include the tooltip bubble.
--- a/packages/wonder-blocks-tooltip/src/components/tooltip-bubble.tsx
+++ b/packages/wonder-blocks-tooltip/src/components/tooltip-bubble.tsx
@@ -3,26 +3,9 @@ import * as React from "react";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
 
-import type {StyleType} from "@khanacademy/wonder-blocks-core";
 import TooltipContent from "./tooltip-content";
 import TooltipTail from "./tooltip-tail";
-
-import type {getRefFn, Offset, Placement} from "../util/types";
-
-export type PopperElementProps = {
-    /** The placement of the bubble with respect to the anchor. */
-    placement: Placement;
-    /** Whether the bubble is out of bounds or not. */
-    isReferenceHidden?: boolean | null | undefined;
-    /** A callback for updating the ref of the bubble itself. */
-    updateBubbleRef?: getRefFn;
-    /** A callback for updating the ref of the bubble's tail. */
-    updateTailRef?: getRefFn;
-    /** Where the tail is to be rendered. */
-    tailOffset?: Offset;
-    /** Additional styles to be applied by the bubble. */
-    style?: StyleType;
-};
+import {PopperElementProps} from "../util/types";
 
 export type Props = {
     /** The unique identifier for this component. */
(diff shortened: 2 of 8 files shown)


packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx
```diff
@@ -25,26 +29,91 @@ type Props = {
...
+                attributes: true,
+                childList: true,
+                subtree: true,
+            });
+        }
+
+        return function cleanup() {
+            _observer?.disconnect();
+        };
+    }, [isMounted, anchorElement, autoUpdate]);
```
jandrade, line 98: **note:** This is the logic that allows auto-updating the tooltip position.

__docs__/wonder-blocks-tooltip/tooltip.stories.tsx
```diff
@@ -351,6 +351,79 @@ WithStyle.play = async ({canvasElement}) => {
...
+        // Wait for the tooltip to update its position
+        const newTooltip = await canvas.findByRole("tooltip");
+        const newLeft = newTooltip.getBoundingClientRect().left;
+        const newTop = newTooltip.getBoundingClientRect().top;
+
+        // Assert
+        // The tooltip should have updated its position
+        await expect(initialLeft).not.toEqual(newLeft);
+        await expect(initialTop).not.toEqual(newTop);
+    },
```
jandrade, line 424: **note:** I started writing a unit test to cover this `autoUpdate` feature, but PopperJS cannot work fine with JSDOM (at least for the styling part), so I decided to switch to Storybook Interaction tests to test that behavior in a real browser.

packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx
jandrade, line 1: **note:** Took the opportunity to refactor this component to be function-based instead of class-based.

packages/wonder-blocks-tooltip/src/components/tooltip-bubble.tsx
```diff
@@ -3,26 +3,9 @@ import * as React from "react";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
 
-import type {StyleType} from "@khanacademy/wonder-blocks-core";
 import TooltipContent from "./tooltip-content";
 import TooltipTail from "./tooltip-tail";
-
-import type {getRefFn, Offset, Placement} from "../util/types";
-
-export type PopperElementProps = {
```
jandrade, line 12: **note:** Moved this to `util/types.ts` (below) as it is used in different files, so it made sense to move it to a more shared place.

packages/wonder-blocks-tooltip/src/components/tooltip.tsx
```diff
@@ -258,6 +265,9 @@ export default class Tooltip extends React.Component<Props, State> {
 
         const popperHost = this._getHost();
 
+        const shouldBeVisible =
+            popperHost && (active || activeBubble) && this.state.anchorElement;
```
somewhatabstract, line 269: **question:** Why do we need to now check the existence of the anchor element when we didn't before?

packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx
```diff
@@ -25,26 +29,91 @@ type Props = {
...
+
+    const _renderPositionedContent = (
         popperProps: PopperChildrenProps,
-    ): React.ReactNode {
-        const {children} = this.props;
+    ): React.ReactNode => {
```
somewhatabstract, line 102: **suggestion:** I think this needs to be in a `useCallback` otherwise it will cause the Popper to re-render every time since it keeps changing.

packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx
```diff
@@ -25,26 +29,91 @@ type Props = {
...
+const _tailRefTracker: RefTracker = new RefTracker();
+/**
+ * A MutationObserver that watches for changes to the anchor element.
+ */
+let _observer: MutationObserver | null = null;
+/**
+ * A function that can be called to update the popper.
+ * @see https://popper.js.org/docs/v2/lifecycle/#manual-update
+ */
+let _popperUpdate: PopperUpdateFn | null = null;
```
somewhatabstract, line 55: **suggestion:** This feels wrong. Previously when this was a class, these were unique per component instance. Now they are shared by every instance of the component. Is that what we want here?

I think keeping this as a class-based component is the least complex thing - otherwise, i think these have to be `useRef` or `useState` things in the component instance.

somewhatabstract (changes requested):
I don't think the refactor from class-based to functional is correct - what were once per instance values are now shared across all instances and I think that's going to be a problem.

packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx
jandrade: hmmm yeah that's a good point, I'll revert it back to a class-based component.

packages/wonder-blocks-tooltip/src/components/tooltip.tsx
jandrade: This is so the `MutationObserver` can observe the correct/existent anchor element in the DOM. I'll add a comment about that.

jandrade:
> I don't think the refactor from class-based to functional is correct - what were once per instance values are now shared across all instances and I think that's going to be a problem.

@somewhatabstract that's a very valid concern. Thanks for pointing that out, I reverted those changes back to the original implementation.

somewhatabstract (approved):
LGTM

Are there any unit tests that should be added for this new behavior to verify that we're handling the observer correctly, etc.?

jandrade:
> LGTM
> 
> Are there any unit tests that should be added for this new behavior to verify that we're handling the observer correctly, etc.?

I tried adding unit tests, but didn't work because PopperJS doesn't play well with JSDOM, so the tests ended up being useless (positions didn't update).

I added Storybook integration/interaction tests instead to verify that behavior.

You can see it here: https://5e1bf4b385e3fb0020b7073c-mxscivqvby.chromatic.com/?path=/story/tooltip-tooltip--auto-update

Screenshot 2024-04-17 at 10 41 21 AM

somewhatabstract:
> > LGTM
> > Are there any unit tests that should be added for this new behavior to verify that we're handling the observer correctly, etc.?
> 
> @somewhatabstract I tried adding unit tests, but didn't work because PopperJS doesn't play well with JSDOM, so the tests ended up being useless or didn't offer a useful way to cover/test that feature (positions didn't update).
> 
> I added Storybook integration/interaction tests instead to verify that behavior.
> 
> You can see it here: [5e1bf4b385e3fb0020b7073c-mxscivqvby.chromatic.com/?path=/story/tooltip-tooltip--auto-update](https://5e1bf4b385e3fb0020b7073c-mxscivqvby.chromatic.com/?path=/story/tooltip-tooltip--auto-update)
> 
> Screenshot 2024-04-17 at 10 41 21 AM

@jandrade I don't think it matters if the position updates. Like with other Tooltip tests, we mock out the popper library so that we can test the logistics such as the observer being properly tidied up. I think tests for that should be added.

jandrade:
> > > LGTM
> > > Are there any unit tests that should be added for this new behavior to verify that we're handling the observer correctly, etc.?
> > 
> > 
> > @somewhatabstract I tried adding unit tests, but didn't work because PopperJS doesn't play well with JSDOM, so the tests ended up being useless or didn't offer a useful way to cover/test that feature (positions didn't update).
> > I added Storybook integration/interaction tests instead to verify that behavior.
> > You can see it here: [5e1bf4b385e3fb0020b7073c-mxscivqvby.chromatic.com/?path=/story/tooltip-tooltip--auto-update](https://5e1bf4b385e3fb0020b7073c-mxscivqvby.chromatic.com/?path=/story/tooltip-tooltip--auto-update)
> > Screenshot 2024-04-17 at 10 41 21 AM
> 
> @jandrade I don't think it matters if the position updates. Like with other Tooltip tests, we mock out the popper library so that we can test the logistics such as the observer being properly tidied up. I think tests for that should be added.

@somewhatabstract I get your point and it's fair enough. I already landed this PR, so I'll add that test in a new PR.

PR #2211, opened 2024-04-16
WB-1679: Set aria-invalid inside TextField when validation fails

jandrade:
## Summary:

Currently, `TextField` accepts a `validate` prop, which updates an internal error
state. We don’t update `aria-invalid` if the validation fails, and that might
mislead folks as they would expect this value to be changed internally.

This PR fixes that by setting `aria-invalid` directly in `TextField` to inform
the user when the validation fails and there's an error in the input field.

Also removed the `aria-invalid` from `LabeledTextField` as it's not necessary
since the error state is already being handled in `TextField`.

Issue: WB-1679

## Test plan:

1. Navigate to the `TextField > Error` story in Storybook.
2. Verify that the `aria-invalid` attribute is set to `true` when the validation
fails.
3. Verify that the `aria-invalid` attribute is set to `false` when the validation
passes.

Do the same for the `LabeledTextField > Error` story.

### DIFF
--- a/.changeset/ten-pumas-tap.md
+++ b/.changeset/ten-pumas-tap.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-form": patch
+---
+
+Set aria-invalid directly in `TextField` to inform the user when the validation fails and there's an error in the input field.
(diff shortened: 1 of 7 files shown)


jeresig (approved):
Ah ha! This does seem like an improvement, thank you!

SeanMcP (commented):
🙌

PR #2213, opened 2024-04-17
Require `aria-labelledby` in ModalDialog (WCAG 2.1 4.1.2)

nedredmond:
Requires `aria-labelledby` to comply with A-level WCAG success criterion 4.1.2

### DIFF
--- a/packages/wonder-blocks-modal/src/components/modal-dialog.tsx
+++ b/packages/wonder-blocks-modal/src/components/modal-dialog.tsx
@@ -44,9 +44,10 @@ type Props = {
      */
     testId?: string;
     /**
-     * The ID of the title labelling this dialog, if applicable.
+     * The ID of the title labelling this dialog. Required.
+     * See WCAG 2.1: 4.1.2 Name, Role, Value
      */
-    "aria-labelledby"?: string;
+    "aria-labelledby": string;
     /**
      * The ID of the content describing this dialog, if applicable.
      */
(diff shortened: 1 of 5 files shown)


packages/wonder-blocks-modal/src/components/modal-dialog.tsx
```diff
@@ -44,9 +44,10 @@ type Props = {
      */
     testId?: string;
     /**
-     * The ID of the title labelling this dialog, if applicable.
+     * The ID of the title labelling this dialog. Required.
+     * See WCAG 2.1: 4.1.2 Name, Role, Value
      */
```
jandrade, line 49: **suggestion:** Could you also please add this comment to the argtypes file? tbh it's a bit annoying having to repeat this, but Storybook Docs has some limitations around that.

https://github.com/Khan/wonder-blocks/blob/main/__docs__/wonder-blocks-modal/modal-dialog.argtypes.tsx#L55-L62

packages/wonder-blocks-modal/src/components/modal-dialog.tsx
jandrade, line 1: **suggestion**: It would be great if you could update the comment block in the component to state that this prop is now required, so we can reflect this change in the docs.

https://github.com/Khan/wonder-blocks/blob/1b21747ae1045357309fc04205b3067c3b601af1/packages/wonder-blocks-modal/src/components/modal-dialog.tsx#L63-L64
Screenshot 2024-04-17 at 12 27 00 PM

PR #2215, opened 2024-04-25
WB-1682: Add faded versions of offBlack

jandrade:
## Summary:

Based on previous discussions with Design, we are going to add faded versions of
the existing `offBlack` and white `color` tokens.

Currently, our "offBlack" versions are using the `rgba` format, and sometimes we just need the solid version of it, so we are adding those versions to the color tokens.

https://khanacademy.slack.com/archives/C8Z9DSKC7/p1713907216050369?thread_ts=1713907205.849639&cid=C8Z9DSKC7

Issue: https://khanacademy.atlassian.net/browse/WB-1682

## Test plan:

Verify that the new colors are available in the `Tokens > Color` page.

Screenshot 2024-04-25 at 4 44 01 PM

### DIFF
--- a/.changeset/smooth-lies-build.md
+++ b/.changeset/smooth-lies-build.md
@@ -2,4 +2,4 @@
 "@khanacademy/wonder-blocks-tokens": minor
 ---
 
-Add faded versions of offBlack and white
+Add faded versions of offBlack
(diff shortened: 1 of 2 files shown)


nishasy:
This is awesome! Thank you for this!

jeresig (approved):

PR #2216, opened 2024-04-26
WB-1589: Add `Listbox` component with Single and Multiple selection support

jandrade:
## Summary:

This PR adds a new `Listbox` component to the `wonder-blocks-dropdown` package.

This is the first step in the implementation of the `Combobox` component. The
`Listbox` component is a dropdown list that can be used to display a list of
options. It supports both single and multiple selection modes, and will be used
by `Combobox` in future PRs, but at this time, we are only adding the `Listbox`
component to make it publicly available.

https://khanacademy.atlassian.net/wiki/spaces/WB/pages/2753298546/Combobox+Implementation+Spec#Listbox-public-new

### Combobox Implementation Plan:

1. **Add `Listbox` component with Single and Multiple selection support. [CURRENT]**
2. Add `Combobox` component with Single selection support.
3. Add multiple selection support to `Combobox` component.
4. Add autocomplete support to `Combobox` component.
5. Add async/dynamic support to `Combobox` component.

Issue: WB-1589

## Test plan:

Verify that the new `Listbox` component works as expected and that the new
documentation is accurate.

/?path=/docs/dropdown-listbox--docs

Screenshot 2024-04-26 at 2 18 42 PM


packages/wonder-blocks-dropdown/src/components/option-item.tsx
```diff
@@ -220,31 +258,85 @@ export default class OptionItem extends React.Component<OptionProps> {
...
+                }
                 {...sharedProps}
             />
         );
     }
+
+    render(): React.ReactNode {
+        const {disabled, focused, parentComponent, role, selected} = this.props;
+
+        if (parentComponent === "listbox") {
```
jandrade, line 272: **note:** I'm making these changes to add support to the new `Listbox` component, while supporting the existing use in `SingleSelect` and `MultiSelect` (used by `DropdownCore`). The long-term goal is to replace some parts of `DropdownCore` with `Listbox`. When that happens, I'll get rid of these customizations.

packages/wonder-blocks-dropdown/src/hooks/use-listbox.tsx
jandrade, line 1: **note:** Created this logic as a custom hook as I'm planning to reuse it with the `Combobox` component (to be included in the next PR).

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
jandrade, line 1: **note:** The changes in this file are mostly TS changes.

packages/wonder-blocks-dropdown/src/components/listbox.tsx
```diff
@@ -0,0 +1,176 @@
...
+        focusedIndex,
+        isListboxFocused,
+        renderList,
+        selected,
+        // event handlers
+        handleKeyDown,
+        handleKeyUp,
+        handleFocus,
+        handleBlur,
+    } = useListbox({children, disabled, id: uniqueId, selectionType, value});
```
jeresig, line 133: I really like this! I was like "wait, how is this component so small??" ha!

jeresig (approved):
Very nice, looking good to me!

packages/wonder-blocks-dropdown/src/components/listbox.tsx
jandrade: 😃 and the best part is that I'm going to reuse this hook in the Combobox component. All these handlers will be attached to the input element (coming in the next PR).

PR #2218, opened 2024-04-26
Remove strings from Popover placement prop that throw errors

nedredmond:
The `Placement` type looked to be copy-pasted from the `react-popper` code which had many variations, but WB limits valid placements to `top | bottom | left | right`. Anything else will throw an error.

This change limits the allowed values to the valid values.


jandrade (approved):
Looks great! thanks for fixing the the types :shipit:

PR #2219, opened 2024-04-26
Storybook: Change sidebar structure

jandrade:
## Summary:

The "Using colors" guide was not too discoverable, so based on feedback from
folks, we decided to move it to the top-level category.

By doing this change, we also had to add prefixes to all the existing pages to
be able to show them into a single category (packages) and preserve the current
structure.

The previous structure was:

- Overview
- Package 1
- Package 2
- ...
- Package N

From now on:

- Overview
- Foundations
  - Using colors
- Packages
  - Package 1
  - Package 2
  - ...
  - Package N

Issue: XXX-XXXX

## Test plan:

Verify that the Storybook sidebar has the new structure and makes sense.

Also verify that the `Using color` page is now easy to find.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2024-04-25 at 5 38 59 PM | Screenshot 2024-04-25 at 5 37 02 PM |


nishasy (approved):
Yay! Thanks for doing this!

jeresig (approved):
Nice, thank you!

PR #2222, opened 2024-05-04
[fei5585.1.hooksrefresh] Updates to timing hooks

somewhatabstract:
## Summary:
This PR updates the timing hooks APIs. This work was initially, to address an issue where calling `set` on the returned APIs would not reset the corresponding timers. However, in doing this work, I consolidated us back down to just two hooks and modified their implementation to reuse the `Interval` and `Timeout` implementations we have that are used by the scheduling API.

First:
- `useInterval` and `useTimeout` have been deleted.
- `useScheduledInterval` and `useScheduledTimeout` have been renamed to `useInterval` and `useTimeout` respectively.

The `active` argument was only ever set to `true` in webapp usage. This is the same as the default behavior of the hooks as they are. Options can be passed to change this to `SchedulePolicy.OnDemand` if the old `false` behavior is desired.

Second:
- The `set` call on the returned API for both hooks now properly resets the timer since it is deferring to the core, tested implementation used elsewhere.

Third:
- A new `ActionPolicy` has been added so that if folks want, the timers will reset when the invoked action changes. The default behavior is to not do this as that could introduce bugs where folks aren't using `useCallback` on their actions. However, adding this ability supports a wider range of use cases.

Issue: FEI-5585

## Test plan:
`yarn test`
`yarn typecheck`

## Upgrade notes:
When migrating a code base that uses an earlier release, these steps should help.

### For migrating `useScheduledInterval` and `useScheduledTimeout`
1. Rename any imports and uses of `useScheduledInterval` to `useInterval` and `useScheduledTimeout` to `useTimeout`.

### For migrating `useInterval` and `useTimeout`
1. For cases where `useInterval` or `useTimeout` are passing a truthy value as the third argument, either remove the third argument (since this is default behavior) or pass `{schedulePolicy: SchedulePolicy.Immediately}` in the third argument.
2. For cases where `useInterval` or `useTimeout` are passing a falsy value as the third argument, pass `{schedulePolicy: SchedulePolicy.OnDemand}` in the third argument.


__docs__/wonder-blocks-icon/accessibility.mdx
```diff
@@ -32,15 +32,11 @@ More information about all these points can be found in the
 
 ### Demo: Contrast
 
-<Canvas>
-  <Story of={AccessibilityStories.IconContrast} />
-</Canvas>
+<Canvas of={AccessibilityStories.IconContrast} />
```
somewhatabstract, line 35: **note:** Regarding the docs changes. Turns out the Storybook syntax changed but we didn't get any error indicating that.

Now, `Canvas` doesn't render children - it uses the `of` prop to know what story to show and without that, will always render the default story from the imported stories.

I have fixed all uses including those that used additional properties on Canvas to show source code.

packages/wonder-blocks-timing/src/hooks/use-timeout.ts
```diff
@@ -1,36 +1,73 @@
...
+                timeoutRef.current?.clear(policy);
+            },
+            get isSet() {
+                return timeoutRef.current?.isSet ?? false;
+            },
+        }),
+        [],
+    );
+
+    return externalApi;
```
jeremywiebe, line 72: As I was looking through this PR it struck me how similar the `useTimeout` and `useInterval` hooks were. We handle things virtually identically between the two implementations. I'm wondering if it'd be worth unifying them into a single underlying implementation and just passing in which `setXyz()` JS function to use to drive the hook (eg. `setTimeout` or `setInterval`). 🤔  Or maybe that just is too much abstraction.

jeremywiebe (commented):
Nice round of bug fixes to these hooks (and improvements)!  Thanks. Given this is WB code, I'll leave it for someone closer to WB to approve.

packages/wonder-blocks-timing/src/hooks/__tests__/use-interval.test.ts
```diff
@@ -1,123 +1,520 @@
...
+        // Act
+        jest.advanceTimersByTime(250);
         rerender({action: action2});
-        jest.advanceTimersByTime(501);
+        jest.advanceTimersByTime(751);
 
         // Assert
-        expect(timeoutSpy).toHaveBeenCalledTimes(callCount);
+        expect(action1).not.toHaveBeenCalled();
+        expect(action2).toHaveBeenCalledTimes(2);
```
kevinbarabash, line 117: Does this mean that `action2` is called immediately when the hook is re-rendered?  It seems a little weird to me, but also I can't make a strong argument as to why any other default behaviour would be preferable and people can change the reset behaviour anyways.  Also, I'm guessing that changing the `action` is not something that's a very common use case.

packages/wonder-blocks-timing/src/hooks/__tests__/use-interval.test.ts
```diff
@@ -1,123 +1,520 @@
...
+            // Act
+            act(() => {
+                result.current.set();
+                jest.advanceTimersByTime(501);
+            });
+
+            // Assert
+            expect(action).toHaveBeenCalledTimes(1);
+        });
+
```
kevinbarabash, line 290: Given the title of the tests I would expect it to spy on `clearInterval` and assert that it was called instead of the action.

packages/wonder-blocks-timing/src/hooks/__tests__/use-timeout.test.ts
```diff
@@ -7,129 +8,484 @@ describe("useTimeout", () => {
...
-        rerender({active: false});
-        jest.advanceTimersByTime(501);
+        act(() => {
+            unmount();
+        });
 
-        // Assert
-        expect(action).not.toHaveBeenCalled();
+        expect(action).toHaveBeenCalled();
     });
```
kevinbarabash, line 74: **suggestion (non-blocking):** Arrange/Act/Assert.

packages/wonder-blocks-timing/src/hooks/__tests__/use-timeout.test.ts
```diff
@@ -7,129 +8,484 @@ describe("useTimeout", () => {
...
-            {initialProps: {action: action1}},
+            ({action}: any) => useTimeout(action, 500),
+            {
+                initialProps: {action: action1},
+            },
         );
         // NOTE: For some reason setTimeout is called twice by the time we get
         // here.  I've verified that it only gets called once inside the hook
         // so something else must be calling it.
         const callCount = timeoutSpy.mock.calls.length;
```
kevinbarabash, line 109: **question:** Is this still the case after your changes?

packages/wonder-blocks-timing/src/hooks/__tests__/use-timeout.test.ts
```diff
@@ -7,129 +8,484 @@ describe("useTimeout", () => {
...
+
+            // Act
+            unmount();
+
+            // Assert
+            expect(action).not.toHaveBeenCalled();
+        });
+    });
+
+    describe("SchedulePolicies.OnDemand", () => {
```
kevinbarabash, line 393:
```suggestion
    describe("SchedulePolicy.OnDemand", () => {
```

packages/wonder-blocks-timing/src/util/interval.ts
```diff
@@ -1,9 +1,6 @@
-import {
-    SchedulePolicy as SchedulePolicies,
-    ClearPolicy as ClearPolicies,
-} from "./policies";
```
kevinbarabash, line 4: **praise:** I like that we've dropped this renaming.

packages/wonder-blocks-timing/src/util/policies.ts
```diff
@@ -1,9 +1,14 @@
...
-} as const;
+export enum ClearPolicy {
+    Resolve = "resolve-on-clear",
+    Cancel = "cancel-on-clear",
+}
+
+export enum ActionPolicy {
+    Reset = "reset",
+    Passive = "passive",
+}
```
kevinbarabash, line 14: **praise:** +1 for enums

kevinbarabash (approved):
This is a very nice cleanup of the timing hooks APIs and refactoring of the code and tests. 🚀

__docs__/wonder-blocks-timing/use-timeout.mdx
```diff
@@ -6,44 +6,85 @@ import {Meta, Story, Canvas} from "@storybook/blocks";
...
-  will stop it
-* Changing the value of `timeoutMs` will reset the timeout, changing `action`
-  will not.
+* Because `clear` takes a param, it's important that you don't pass it directly to an event handler,
+  e.g. `<Button onClick={clear} />` will not work as expected.
+* Calling `set` after the timeout has expired will restart the timeout.
+* Updating the second paramter, `timeoutMs`, will also restart the timeout.
+* When the component using this hooks is unmounted, the timeout will automatically be cleared.
+* Calling `set` after the timeout is set but before it expires means that the timeout will be
+  reset and will call `action`, `timeoutMs` after the most recent call to `set` was made.
```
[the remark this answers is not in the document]
somewhatabstract: Hmm, it does the same in both cases. The docs may just be wrong.

packages/wonder-blocks-timing/src/hooks/use-interval.ts
```diff
@@ -1,36 +1,101 @@
...
+ * consumers to wrap their action in a `useCallback`. To change this behavior,
+ * see the `actionPolicy` option.
+ * @param intervalMs The interval period. If this changes, the interval will
+ * be reset.
+ * @param options Options for the hook.
+ * @param options.actionPolicy Determines how the action is handled when it
+ * changes. By default, the action is replaced but the interval is not reset,
+ * and the updated action will be invoked when the interval next fires.
+ * If you want to reset the interval when the action changes, use
+ * "replace-and-reset".
```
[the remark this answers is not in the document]
somewhatabstract: I'll check. I made some changes before I moved to enums.

packages/wonder-blocks-timing/src/hooks/use-interval.ts
```diff
@@ -1,36 +1,101 @@
...
+ * @param options.clearPolicy Determines how the interval is cleared when the
+ * component is unmounted or the interval is recreated. By default, the
+ * interval is cleared immediately. If you want to let the interval run to
+ * completion, use "resolve-on-clear". This is NOT applied if the interval
```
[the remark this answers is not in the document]
somewhatabstract: Yeah, this is old docs before the enum change was made.

packages/wonder-blocks-timing/src/hooks/__tests__/use-timeout.test.ts
somewhatabstract: This was an existing test from your old changes. I didn't edit them all, but the diff has made some unusual choices in showing what changed. I can update them to AAA though.

packages/wonder-blocks-timing/src/hooks/__tests__/use-timeout.test.ts
somewhatabstract: Not sure. I can check.

__docs__/wonder-blocks-icon/accessibility.mdx
somewhatabstract: > Thanks for fixing this. Was the only way to see that it was broken to look at these pages in Storybook?

Yes, unfortunately. I thought at first I had broken the hooks, and had to dig into the stories more before I discovered this issue.

__docs__/wonder-blocks-timing/use-interval.stories.tsx
```diff
@@ -15,20 +19,43 @@ export default {
...
+            </View>
+        </View>
+    );
+};
+
+export const OnDemandAndResolveOnClear = () => {
+    const [callCount, setCallCount] = React.useState(0);
+    const callback = React.useCallback(() => {
+        // eslint-disable-next-line no-console
+        console.log("action called");
```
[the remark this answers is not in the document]
somewhatabstract: Aha! That was debug code that I forgot to remove :D

packages/wonder-blocks-timing/src/hooks/use-timeout.ts
somewhatabstract: Yes, you may be right. We could possibly replace this with a single underlying hook like `useActionScheduler` that takes the type of scheduler somehow. Not sure if it's worth the effort in golfing it down to that though.

packages/wonder-blocks-timing/src/hooks/__tests__/use-interval.test.ts
somewhatabstract: No, the action is not called when anything renders. It is called when the interval fires. This test shows that changing the action means it will get called when the next interval occurs - it won't reset the interval.

packages/wonder-blocks-timing/src/hooks/__tests__/use-timeout.test.ts
somewhatabstract: The confusion here is that this is a rename of the old `use-scheduled-timeout.test.ts` file but it's getting diffed against the now deleted `use-timeout.text.ts` file.

PR #2224, opened 2024-05-14
Allow TextField number input to use min, max, and step props

nishasy:
## Summary:
HTML `<input>` elements have `min`, `max`, and `step` props.

Adding them here to TextField to better align with HTML props.

Issue: none

## Test plan:
Storybook
http://localhost:6061/?path=/story/packages-form-textfield--number

### DIFF
--- a/__docs__/wonder-blocks-form/text-field.stories.tsx
+++ b/__docs__/wonder-blocks-form/text-field.stories.tsx
@@ -138,7 +138,7 @@ export const Number: StoryComponentType = () => {
     };
 
     return (
-        <View style={styles.column}>
+        <View>
             <TextField
                 id="tf-3"
                 type="number"
@@ -149,7 +149,7 @@ export const Number: StoryComponentType = () => {
             />
             <Strut size={spacing.small_12} />
             <Body>
-                The following text field has a min of 0, a max of 15, and a snap
+                The following text field has a min of 0, a max of 15, and a step
                 of 3
             </Body>
             <TextField
@@ -177,7 +177,7 @@ Number.parameters = {
                 restrictions, while the second number input has a minimum
                 value of 0, a maximum value of 15, and a step of 3. Observe
                 that using the arrow keys will automatically snap to the
-                snap step, and stop at the min and max values.`,
+                step, and stop at the min and max values.`,
         },
     },
 };
@@ -896,8 +896,5 @@ const styles = StyleSheet.create({
     },
     fieldWithButton: {
         marginBottom: spacing.medium_16,
-    },
-    column: {
-        flexDirection: "column",
     },
 });
--- a/__docs__/wonder-blocks-form/text-field.stories.tsx
+++ b/__docs__/wonder-blocks-form/text-field.stories.tsx
@@ -6,7 +6,7 @@ import {View, Text as _Text} from "@khanacademy/wonder-blocks-core";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
 import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
 import Button from "@khanacademy/wonder-blocks-button";
-import {LabelLarge} from "@khanacademy/wonder-blocks-typography";
+import {LabelLarge, Body} from "@khanacademy/wonder-blocks-typography";
 
 import {TextField} from "@khanacademy/wonder-blocks-form";
 import packageConfig from "../../packages/wonder-blocks-form/package.json";
@@ -124,7 +124,8 @@ Required.parameters = {
 };
 
 export const Number: StoryComponentType = () => {
-    const [value, setValue] = React.useState("12345");
+    const [value, setValue] = React.useState("1234");
+    const [value2, setValue2] = React.useState("12");
 
     const handleChange = (newValue: string) => {
         setValue(newValue);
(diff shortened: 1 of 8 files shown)


packages/wonder-blocks-form/src/components/text-field.tsx
```diff
@@ -248,6 +269,12 @@ class TextField extends React.Component<PropsWithForwardRef, State> {
             ...otherProps
         } = this.props;
 
+        if ((min || max || step) && type !== "number") {
+            throw new Error(
+                'The props `min`, `max`, and `step` can only be used with `type="number"`.',
+            );
+        }
```
jandrade, line 276: **thought:** this should probably be checked at build time instead of runtime. Probably using a TS discriminated union type would be better for this case.

You can see an example of this here: https://www.typescriptlang.org/play/?&q=387#code/C4TwDgpgBAwg9gWwXAdgBQE5zAZygXigG8AoKcqAegCpqyKpqoAVAC2gFcUBLARw+jcAJhBTBuAM24QMUCXFnB2UbijAdgAOnoVqlHSqEAuKDmAZVAcwDcBmnQaMWy1euBQAbgEMANgO2OegbefhAmZhYoNna0Bk5s0CheCNDyii5qGlAAxqjmcD6aztx4JaYcAEYI3MDAEEJQvqiWUADuNaxxTEqpCghQQl7AXgEMQQxJKQD84eZWtgC+tiSgkFAA8j0YAJKZwJjYeITwSKgHuFAAZMQGqxAz5ABEdQAewI9QAD5Qj2BeODhWgohB9vo8IAgvNwfKCfnUYYtlncoAA5DgpCzZXZuc5HWCIZDoLAXa6kBh3B6PFDoioyR62Bj2LrOaDVHgIdFQakY7jZTy+ARyBRQHoqPajXT6Bhsh7c2kYBkUJmBFlQSEvbgc-rcmS8-mhIXpQTirpSijq2U0mSK8jKsaqnWY-WC4BwMXZDAQ0TuYUiD1esRQCogCXkJgAJQg-G4nrwotcWVdQdZHB84jAPggeDgEhFrDKIX8poMZggYEtCHliJIK3A0FxBA2W2xGgb3zRPKxe1xy1yKDMIogbxbwBMDcIZIowhMjxqOAAtF55693gAaYICsI-HCICDLodr251mcrx4kJY1vsD1QhYTMA8jsfEvGT8jTn5zxf7t6PdcMQtbo8O4pN+h7ksecIHr+BhsiYACM64XiQV7uNyj5QOONwMO+s7AAuS6nn+FAATOwF7oRR6QDOcp0kR5CwVACHnkAA

__docs__/wonder-blocks-form/text-field.stories.tsx
```diff
@@ -137,21 +138,46 @@ export const Number: StoryComponentType = () => {
...
+                id="tf-3"
+                type="number"
+                value={value}
+                placeholder="Number"
+                onChange={handleChange}
+                onKeyDown={handleKeyDown}
+            />
+            <Strut size={spacing.small_12} />
+            <Body>
+                The following text field has a min of 0, a max of 15, and a snap
```
jeremywiebe, line 152: "snap" or "step"?  The prop is named `step` but is "snap" another term for it?

packages/wonder-blocks-form/src/components/labeled-text-field.tsx
```diff
@@ -122,6 +120,15 @@ type Props = {
     autoComplete?: string;
 };
 
+type OtherInputProps = CommonProps & {
+    /**
+     * Determines the type of input. Defaults to text.
+     */
+    type?: "text" | "password" | "email" | "tel";
```
jeremywiebe, line 127: Given we have this in the `DefaultProps`, I think we can mark this as non-optional.  See `disabled` prop above which also has a default. 

```suggestion
    type: "text" | "password" | "email" | "tel";
```

jeremywiebe (commented):
Just giving my 1.5¢ worth. 😄

packages/wonder-blocks-form/src/components/text-field.tsx
```diff
@@ -117,6 +114,30 @@ type Props = AriaProps & {
     autoComplete?: string;
 };
 
+type OtherInputProps = CommonProps & {
+    type?: "text" | "password" | "email" | "tel";
```
kevinbarabash, line 118: This should be required.

packages/wonder-blocks-form/src/components/text-field.tsx
```diff
@@ -117,6 +114,30 @@ type Props = AriaProps & {
...
+};
+
+// Props that are only available for inputs of type "number".
+export type NumericInputProps = {
+    type?: "number";
```
kevinbarabash, line 123: This should be required as well.

jandrade (approved):
Looks great! agree with Jeremy's and Kevin's suggestions about marking `type` as required. This is a great improvement overall, thanks! 🚢

__docs__/wonder-blocks-form/text-field.stories.tsx
nishasy: You're right! It's just step.

packages/wonder-blocks-form/src/components/labeled-text-field.tsx
nishasy: Is it better to do it this way?

I've always found it odd to make a prop non-optional if it has a default prop, because I think it inherently becomes optional for a user to pass in that prop if it has a default value... right?

PR #2227, opened 2024-05-16
Pass aria attributes to pill

nedredmond:
Previously, aria attributes were accepted as props, but nothing was done with them. I have passed the aria attributes to the `clickable` and the `view` so that they can be used.


jeresig (approved):
oof, thank you!

PR #2231, opened 2024-05-20
Allow all special characters in IDs, except whitespace.

jeresig:
## Summary:
This came up with the Perseus team, where they had some Unicode characters in an ID and that was throwing errors. I did some more research and learned that only whitespace is considered to be invalid, so I updated our logic around this.

Issue: WB-1704

## Test plan:
I ran the tests and they passed.

### DIFF
--- a/packages/wonder-blocks-core/src/util/unique-id-factory.ts
+++ b/packages/wonder-blocks-core/src/util/unique-id-factory.ts
@@ -33,14 +33,13 @@ export default class UniqueIDFactory implements IIdentifierFactory {
      * identifier. It does not assert that a string IS a valid identifier (for
      * example, that it doesn't start with numbers). We don't need to do that
      * here because all identifiers are prefixed to avoid needing that check.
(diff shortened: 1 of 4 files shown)


packages/wonder-blocks-core/src/util/unique-id-factory.ts
```diff
@@ -33,14 +33,13 @@ export default class UniqueIDFactory implements IIdentifierFactory {
      * identifier. It does not assert that a string IS a valid identifier (for
      * example, that it doesn't start with numbers). We don't need to do that
      * here because all identifiers are prefixed to avoid needing that check.
+     *
+     * According to this post:
+     * https://stackoverflow.com/questions/70579/html-valid-id-attribute-values/31773673#31773673
```
jandrade, line 38: Thanks for adding this link!

jandrade (approved):
Looks great and nice catch! :shipit:

PR #2233, opened 2024-05-28
Remove react-test-renderer

jandrade:
## Summary:

We still use `react-test-renderer` in some of our tests. This is not completely
necessary as we are using `@testing-library/react` which we can use to generate
more accurate snapshots (that print out the actual HTML that is rendered).

Issue: XXX-XXXX

## Test plan:

Verify that the tests still pass and that the new snapshots present relevant
information.


jandrade:
This was pointed out by @jeremywiebe, thanks!

jeresig (approved):
Very nice!

PR #2234, opened 2024-05-29
Prevent dropdown item focus when enableTypeAhead is false

osha90:
## Summary:
Requirement: When enableTypeAhead is set to false, Single Select component should not auto focus on a dropdown item when the search text changes.
This PR prvents the dropdown item from from focus when the search text changes the first time. 

Issue: DIST-4569

## Test plan:

### DIFF
--- a/.changeset/afraid-coats-call.md
+++ b/.changeset/afraid-coats-call.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-dropdown": minor
+---
+
+Prevent dropdown item focus when enableTypeAhead is false
--- a/.changeset/afraid-coats-call.md
+++ b/.changeset/afraid-coats-call.md
@@ -1,5 +1,5 @@
 ---
-"@khanacademy/wonder-blocks-dropdown": minor
+"@khanacademy/wonder-blocks-dropdown": patch
 ---
 
 Prevent dropdown item focus when enableTypeAhead is false
--- a/packages/wonder-blocks-dropdown/src/components/dropdown-core.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/dropdown-core.tsx
@@ -571,9 +571,10 @@ class DropdownCore extends React.Component<Props, State> {
     }
 
     focusPreviousItem(): void {
-        if (this.isSearchFieldFocused() && !this.props.enableTypeAhead) {
-            this.focusedIndex = this.state.itemRefs.length - 1;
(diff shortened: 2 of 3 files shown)


.changeset/afraid-coats-call.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-dropdown": minor
```
jandrade, line 2: **suggestion:** This is more of a `patch` version as you are fixing a bug in the existing implementation.

More of that here: https://khanacademy.atlassian.net/wiki/spaces/WB/pages/2217640195/Versioning+and+releasing+Wonder+Blocks+packages#Patch-changes-0.0.1

packages/wonder-blocks-dropdown/src/components/dropdown-core.tsx
```diff
@@ -571,7 +571,9 @@ class DropdownCore extends React.Component<Props, State> {
     }
 
     focusPreviousItem(): void {
-        if (this.focusedIndex === 0) {
+        if (this.isSearchFieldFocused() && !this.props.enableTypeAhead) {
+            this.focusedIndex = this.state.itemRefs.length - 1;
+        } else if (this.focusedIndex === 0) {
```
jandrade, line 576: **suggestion:** I think this could be moved to the original `if`, so we can simplify assigning `focusedIndex` to this condition:

```suggestion
        if (this.focusedIndex === 0 || (this.isSearchFieldFocused() && !this.props.enableTypeAhead)) {
```

packages/wonder-blocks-dropdown/src/components/dropdown-core.tsx
```diff
@@ -585,7 +587,9 @@ class DropdownCore extends React.Component<Props, State> {
     }
 
     focusNextItem(): void {
-        if (this.focusedIndex === this.state.itemRefs.length - 1) {
+        if (this.isSearchFieldFocused() && !this.props.enableTypeAhead) {
+            this.focusedIndex = 0;
+        } else if (this.focusedIndex === this.state.itemRefs.length - 1) {
```
jandrade, line 592: Same suggestion here about combining both conditions into a single one:

```suggestion
    if (t(his.focusedIndex === this.state.itemRefs.length - 1) || (this.isSearchFieldFocused() && !this.props.enableTypeAhead)) {
```

jandrade (approved):
Looks good to me! I added some suggestions for you to consider before landing.

PR #2235, opened 2024-05-29
LabeledTextField: Fix `onValidate` prop to be called properly

jandrade:
## Summary:

Fixes an issue where the `onValidate` prop was not being called when the
callback was provided in the call site. This caused a weird behavior where
the `handleValidate` function inside `LabeledTextField` never got called.

See https://github.com/Khan/webapp/actions/runs/9273558563/job/25513924364?pr=22121
for more context in the failure.

Issue: XXX-XXXX

## Test plan:

Verify that unit tests pass.

### DIFF
--- a/.changeset/ten-forks-repair.md
+++ b/.changeset/ten-forks-repair.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-form": patch
+---
+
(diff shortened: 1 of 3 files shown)


nishasy (approved):
Yayyy!!!

PR #2238, opened 2024-05-30
[🔥AUDIT🔥] LabeledTextField: Fix onValidate prop to be called properlyLabeledTextField: Fix onBlur and onFocus props to be called properly

jandrade:
🖍 _This is an audit!_ 🖍

## Summary:

While working on #2235, I missed a few extra props from being passed to the
underlying `TextField` component. This PR fixes that.

Issue: XXX-XXXX

## Test plan:

Verify that unit tests work here and in webapp.

### DIFF
--- a/.changeset/long-zoos-think.md
+++ b/.changeset/long-zoos-think.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-form": patch
+---
+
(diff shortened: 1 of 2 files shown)


jandrade:
I've confirmed that the tests pass in webapp by installing the snapshot release and running the previously failed tests!

Screenshot 2024-05-30 at 11 56 56 AM

PR #2240, opened 2024-05-30
Single/Multi-select Storybook fixes

beaesguerra:
## Summary:
- adds `offWhite` background setting in Storybook
- remove single/multi-select story decorators for background styling and utilize Storybook's background parameter feature instead so that it can be controlled like the `light` prop control
- ensure control args are applied before component props in single select stories

Issue: WB-1543

## Test plan:
- Confirm that stories can now have an `offWhite` background
- Confirm that the SingleSelect - Disabled story (?path=/story/packages-dropdown-singleselect--disabled) no longer has a hardcoded offWhite container so that the background can be changed to `darkBlue` with the `light` prop control enabled
- Confirm the `disabled` control works in the SingleSelect Light story (?path=/story/packages-dropdown-singleselect--light)

Screenshots:
New `offWhite` background setting
Screenshot 2024-05-30 at 11 22 57 AM

SingleSelect - Disabled story (before & after)
Screenshot 2024-05-30 at 11 19 03 AM

SingleSelect - Light story with disabled control enabled (before & after)
Screenshot 2024-05-30 at 11 20 55 AM


jandrade (approved):
LGTM! This looks way better! Thanks for taking the `backgrounds` approach, the examples look cleaner than before and also for the detailed Test Plan.

Congrats on landing your first official WB PR 🚢 :shipit: 🚀

PR #2241, opened 2024-05-30
Updates to icon a11y docs and examples

beaesguerra:
## Summary:
- Adds instructions for adding `aria-hidden` to decorative icons to hide it from screen readers
- Use `aria-label` terminology in the icon a11y docs instead of `alt text` since the PhosphorIcon component takes an `aria-label` prop rather than `alt` since a `span` element is used for the icon instead of an `img` or `svg`
- Group docs related to aria-labels together
- Storybook: Address a11y warning where the `aria-label` was set in stories and `role` was not
- Storybook: Add new story showing an example of a decorative icon

Issue: WB-1599

## Test plan:
- Confirm the icon a11y docs make sense (`?path=/docs/packages-icon-accessibility--docs`)
- Confirm there are no more a11y warnings in Storybook for icons (`?path=/story/packages-icon-phosphoricon--default`)
- Review the new Decorative Icon story (`?path=/story/packages-icon-phosphoricon--decorative-icon`)

## Screenshots
Storybook a11y warning for PhosphorIcon story (before & after)
Screenshot 2024-05-30 at 4 29 27 PM


__docs__/wonder-blocks-icon/accessibility.mdx
```diff
@@ -13,19 +13,19 @@ import * as AccessibilityStories from './accessibility.stories';
 * The color contrast should pass WCAG.
   * At least 3:1 contrast ratio for icons.
 * Ideally, an icon should be accompanied by text. However, if an icon
-  is not accompanied by text, provide alt text.
```
beaesguerra, line 16: I replaced references to "alt text" with "aria-label" since the [PhosphorIcon component](https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-icon/src/components/phosphor-icon.tsx#L13) has an `aria-label` prop instead of an `alt` prop. Please let me know though if I'm missing some context!

__docs__/wonder-blocks-icon/accessibility.mdx
```diff
@@ -13,19 +13,19 @@ import * as AccessibilityStories from './accessibility.stories';
...
+    text.
+  * If the icon is decorative, set `aria-hidden` to `true` and do not provide an
+   `aria-label`. This will hide the icon from screen readers.
   * The `title` attribute is often used to display a description on mouse
     hover, but please note that use of the `title` attribute is discouraged
     as titles are not available to keyboard-only or touch-only users.
     More information can be found in the [References](#references) below.
 * Icons should go on the opposite side as left-to-right languages
   for right-to-left languages.
   * The icon may also need to be mirrored.
```
beaesguerra, line 28: For my own curiosity, when should icons be mirrored? And how should developers do this?

__docs__/wonder-blocks-icon/accessibility.mdx
```diff
@@ -13,19 +13,19 @@ import * as AccessibilityStories from './accessibility.stories';
 * The color contrast should pass WCAG.
   * At least 3:1 contrast ratio for icons.
 * Ideally, an icon should be accompanied by text. However, if an icon
-  is not accompanied by text, provide alt text.
-  * Do not provide alt text for decorative icons. Set the alt text
-    to an empty string (`""`).
+  is not accompanied by text, provide an `aria-label`.
```
jandrade, line 16: **nit:** probably would be nice adding `... attribute (or prop) here.
```suggestion
  is not accompanied by text, provide an `aria-label` attribute.
```

__docs__/wonder-blocks-icon/accessibility.mdx
jandrade: 🤔 I never thought about this.... my take would be to use `rotate(180deg)`, but not sure if it needs to be mirrored in a different way (cc @nishasy in case you have any thoughts about this).

__docs__/wonder-blocks-icon/phosphor-icon.stories.tsx
```diff
@@ -81,6 +81,7 @@ export const Default: StoryComponentType = {
         icon: IconMappings.magnifyingGlassBold,
         size: "small",
         "aria-label": "Search",
+        role: "img",
```
jandrade, line 84: **praise:** Thanks for fixing a11y in this story :accessibility:

jandrade (approved):
Looks great! thanks for adding the new story as well :accessibility: 🚀

PR #2243, opened 2024-06-03
Allow SingleSelect and MultiSelect to be focusable when disabled

beaesguerra:
## Summary:
- Use `aria-disabled` instead of `disabled` for SelectOpener so that the disabled state can be focusable. This makes it consistent with disabled buttons
- Add focus styling for disabled SelectOpener for both light and non-light types
- Add single/multiselect variant stories with states. When creating the new stories to show all the variants, I noticed two things (I will address these in a separate pull request!):
  - The error state styling needs to be updated when the `light` prop is `true` 
  - The state styling (focus, hover, active) can be moved from JS to CSS pseudoclasses (ie. `:focus-visible`, `:hover`, `:active`). By doing this, we can simplify logic and also utilize the Storybook pseudostates addon for visual regression tests (I'll add stories for the different states once this is updated!) 

Issue: WB-1238

## Test plan:
- Verify that the components using SelectOpener (SingleSelect, MultiSelect, BirthdayPicker) can be focused on when they are in a disabled state
- Verify focus styling on the SingleSelect and MultiSelect components (specifically when they are disabled)
- Verify that screen readers still communicate that the components are in a disabled state (from the `aria-disabled` attribute)
- Verify new all variants stories:
  - `?path=/docs/packages-dropdown-singleselect-all-variants--docs`
  - `?path=/docs/packages-dropdown-multiselect-all-variants--docs`

## Screen recording:
Single Select - VoiceOver (Safari) and Keyboard Navigation 

https://github.com/Khan/wonder-blocks/assets/14334617/017ee324-2fd5-4da5-9cf2-3d2ee69dfe77

### DIFF
--- a/.changeset/rotten-actors-flow.md
+++ b/.changeset/rotten-actors-flow.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-dropdown": patch
+---
+
+Allow `SingleSelect` and `MultiSelect` to be focusable when disabled. These components now have `aria-disabled` when the `disabled` prop is `true`. This makes it so screenreaders will continue to communicate that the component is disabled, while allowing focus on the disabled component. Focus styling is also added to the disabled state.
--- a/__docs__/wonder-blocks-dropdown/multi-select-variants.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/multi-select-variants.stories.tsx
@@ -1,0 +1,145 @@
+import * as React from "react";
+import {StyleSheet} from "aphrodite";
+import type {Meta, StoryObj} from "@storybook/react";
+
+import {View} from "@khanacademy/wonder-blocks-core";
+import {ThemeSwitcherContext} from "@khanacademy/wonder-blocks-theming";
+import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {HeadingLarge, LabelMedium} from "@khanacademy/wonder-blocks-typography";
+import {MultiSelect, OptionItem} from "@khanacademy/wonder-blocks-dropdown";
+
+/**
+ * The following stories are used to generate the pseudo states for the
+ * multi select component. This is only used for visual testing in Chromatic.
+ */
+export default {
+    title: "Packages / Dropdown / MultiSelect / All Variants",
+    parameters: {
+        docs: {
+            autodocs: false,
+        },
+    },
+} as Meta;
+
+type StoryComponentType = StoryObj<typeof MultiSelect>;
+
+const defaultProps = {
+    onChange: () => {},
+};
+
+const selectItems = [
+    <OptionItem label="item 1" value="1" />,
+    <OptionItem label="item 2" value="2" />,
+    <OptionItem label="item 3" value="3" />,
+];
+
+const KindVariants = ({light}: {light: boolean}) => {
+    return (
+        <ThemeSwitcherContext.Consumer>
+            {(theme) => (
+                <>
+                    <View
+                        style={[
+                            styles.gridRow,
+                            light &&
+                                (theme === "khanmigo"
+                                    ? styles.darkKhanmigo
+                                    : styles.darkDefault),
+                        ]}
+                    >
+                        <LabelMedium style={light && {color: color.white}}>
+                            Default
+                        </LabelMedium>
+                        <MultiSelect {...defaultProps} light={light}>
+                            {selectItems}
+                        </MultiSelect>
+                    </View>
+                    <View
+                        style={[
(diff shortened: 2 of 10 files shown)


.changeset/rotten-actors-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-dropdown": patch
```
beaesguerra, line 2: I made this a `patch` update since it's a small update, but let me know if this should be a minor update instead. 

I'm thinking tests for things using Single/MultiSelect may break if the unit tests are checking `.toBeDisabled` (like [BirthdayPicker](packages/wonder-blocks-birthday-picker/src/components/__tests__/birthday-picker.test.tsx) was)

Let me know what you think!

__docs__/wonder-blocks-dropdown/multi-select-variants.stories.tsx
```diff
@@ -0,0 +1,145 @@
+import * as React from "react";
```
beaesguerra, line 1: Inspired by the [Icon Button Variants stories](https://github.com/Khan/wonder-blocks/blob/main/__docs__/wonder-blocks-icon-button/icon-button-variants.stories.tsx) :D

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -261,6 +261,9 @@ const _generateStyles = (
                 borderColor: mix(tokens.color.white32, tokens.color.blue),
                 color: mix(tokens.color.white32, tokens.color.blue),
                 cursor: "auto",
+                ":focus": {
```
beaesguerra, line 264: Disabled focus styling was based on [these changes](https://github.com/Khan/wonder-blocks/pull/1257/files#diff-134d42ab9fc77b18bd4bade6c66d5ac4fdb937902b895eacde259cb39fd8529eR294-R298)!

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
jandrade, line 1: **praise:** Thanks for adding all the possible test scenarios.

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -293,6 +296,9 @@ const _generateStyles = (
                 borderColor: tokens.color.offBlack16,
                 color: tokens.color.offBlack64,
                 cursor: "auto",
+                ":focus": {
```
jandrade, line 299: **suggestion/issue:** This probably needs to be changed to `:focus-visible`. Right now, if we use `focus`, then the disabled state will also apply when the user presses on the button (via mouse for example), and the would misguide the user as it would think that the button is interactive.

jandrade: That said, I understand if you want to leave this fix for the next PR (as you mentioned in the description :)).

jandrade (approved):
Looks great! thanks for adding the "All variants" stories. This will really increase the confidence in all the pseudo-classes :).

One thing that also want to mention is that I think it would be nice if we could add an extra description in the Docs mentioning that we use `aria-disabled` instead of `disabled` for a11y purposes. For example, it would be nice adding that here:

https://5e1bf4b385e3fb0020b7073c-hxlrveljzq.chromatic.com/?path=/docs/packages-dropdown-multiselect--docs#disabled
And here: https://5e1bf4b385e3fb0020b7073c-hxlrveljzq.chromatic.com/?path=/docs/packages-dropdown-singleselect--docs#disabled

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
beaesguerra: Good catch! I will update this in my next PR that makes sure all the states are correct!

beaesguerra:
> One thing that also want to mention is that I think it would be nice if we could add an extra description in the Docs mentioning that we use `aria-disabled` instead of `disabled` for a11y purposes. For example, it would be nice adding that here:
> 
> https://5e1bf4b385e3fb0020b7073c-hxlrveljzq.chromatic.com/?path=/docs/packages-dropdown-multiselect--docs#disabled And here: https://5e1bf4b385e3fb0020b7073c-hxlrveljzq.chromatic.com/?path=/docs/packages-dropdown-singleselect--docs#disabled

@jandrade Good idea! I updated the documentation for the Disabled stories for the SingleSelect, MultiSelect, and BirthdayPicker components 👍

PR #2244, opened 2024-06-04
Update styling for Single/MultiSelect states

beaesguerra:
## Summary:
Addresses the state styling issues noticed when fixing WB-1238 ([PR](https://github.com/Khan/wonder-blocks/pull/2243)). 

- Move hover, focus, active state styling to CSS pseudo classes instead of relying on JS logic. This makes it so we can simplify styling logic and also use the Storybook pseudo states add on for visual regression tests
  - Keep js logic for showing pressed styling when Enter/Space is pressed
- Add stories for different pseudo states
- Replace `:focus` with `:focus-visible` for disabled focused styling
- Replace ClickableBehavior in SelectOpener: Similar to IconButton, we replace ClickableBehavior with keyUp/keyDown handlers to achieve normalized behaviour in the different browsers. All browsers should open the select when pressing Space or Enter. Since the SelectOpener doesn't support routes, the router logic was also removed while removing ClickableBehaviour
- UI updates:
  - Fix dropdown error state in light mode. Updates the text and icon color
  - Update focus/hover border color to red instead of fadedRed for light variants of select components
  - Update light variant text color when select opener is active and in an error state
  - Update disabled state cursor to use `not-allowed`

Note: It seems the styling for [select components in Figma for light mode](https://www.figma.com/design/VbVu3h2BpBhH80niq101MHHE/%F0%9F%92%A0-Main-Components?node-id=13693-11142&t=GuAXJN1oDmW6YMVn-4) are different than the ones in the [code components](https://khan.github.io/wonder-blocks/?path=/story/packages-dropdown-singleselect--default&args=light:!true&globals=backgrounds.value:!hex(0b2149)) (even in `main`). I've only updated styling for scenarios where the text was not readable - Let me know though if we would like to restyle these to match Figma! We may also not need to do this now since the upcoming Combobox is intended to replace the internals of the select components

Issue: WB-1238

## Test plan:
- Verify the All Variants stories for SingleSelect and MultiSelect
  - `?path=/docs/packages-dropdown-singleselect-all-variants--docs`
  - `?path=/docs/packages-dropdown-multiselect-all-variants--docs`
- Verify SingleSelect and MultiSelect behave as expected in different browsers (clicking, pressing Enter, and pressing Space all open the select and show the pressed/active state styling)

## Recording:
Opening the select by clicking, pressing Enter, and pressing Space in Chrome, Firefox, Safari and Edge

https://github.com/Khan/wonder-blocks/assets/14334617/95c72089-4427-451b-acfa-e156d5ea0c43


packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -229,44 +225,58 @@ const _generateStyles = (
...
-                borderColor: error ? tokens.color.red : tokens.color.fadedBlue,
-                borderWidth: tokens.border.width.thin,
-                color: placeholder
-                    ? mix(tokens.color.white32, tokens.color.blue)
-                    : tokens.color.fadedBlue,
-                backgroundColor: error
-                    ? tokens.color.fadedRed
-                    : tokens.color.activeBlue,
+                ":hover:not([aria-disabled=true])": focusHoverStyling,
+                ":focus-visible:not([aria-disabled=true])": focusHoverStyling,
```
beaesguerra, line 245: I believe we can use `:focus-visible` now ([caniuse](https://caniuse.com/?search=focus-visible)) without fallback `:focus` styling (like this [example of fallback focus styling](https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx#L355-L360)) since the [minimum Safari version](https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/103612417/Browser+Support) we support now is 15.6.

Let me know otherwise though!

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
```diff
@@ -1539,58 +1539,6 @@ describe("MultiSelect", () => {
         });
     });
 
-    describe("error state styles", () => {
```
beaesguerra, line 1542: Since we use css pseudo-classes now to handle hover/focus/active styling, these tests failed to verify style attributes because the state styling was applied using js previously. We have the Chromatic regression tests now to verify styling visually so I believe these are no longer necessary. Let me know though if you have any concerns with this!

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
jandrade, line 1: **issue (blocking):** There are some extra changes that will probably need to be addressed in this component:

1. `ClickableBehavior` introduces some custom behaviors that prevent pseudo-classes from being trigged correctly when navigating with the keyboard. For example, if you focus on the opener, then press `Enter` or `Space`, there won't be visual feedback for the user to see when the button was pressed.

I'm mentioning this as Browsers don't trigger the `:active` state consistently. For example, Chrome and Safari only do it on `Space`. Firefox doesn't apply this at all (https://bugzilla.mozilla.org/show_bug.cgi?id=68851).

I suggest looking at `IconButton` to see how to remove `ClickableBehavior` and apply some custom event handlers to apply the Enter/Space logic consistently.

https://github.com/Khan/wonder-blocks/blob/9092363d7d50ebe9816b115e3ae024e2139e38d4/packages/wonder-blocks-icon-button/src/components/icon-button.tsx#L214-L215

------------

2. Disabling `hover` on touch devices: There's another part that would need to be addressed in the styles, which is overriding the hover styles so it don't become sticky when pressing the opener in a mobile device (for example). You can see an example of this here: https://github.com/Khan/wonder-blocks/blob/58075352c336b1c3c85ddeccba906058ca579c49/packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx#L343-L354

jandrade (changes requested):
Changes are looking good so far, but I'm requesting some changes to remove `ClickableBehavior` and moving some of its functionality to `SelectOpener` directly (taking `IconButton` as reference).

Thanks for moving us to CSS to use pseudo-classes btw :).

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
beaesguerra: Thanks for sharing this context Juan, this is helpful!

1. Replaced `ClickableBehaviour` logic with the `keyDown`/`keyUp` handlers. ✅ 
`Enter` and `Space` will open the select in all browsers now. However, the active state is only shown when clicked with a mouse in all browsers. When we paired earlier today, I showed that I had the active state showing in Chrome when using `Space` (I did this by removing the event.preventDefault call for the `Space` key). After more testing, I realized the active styling was persisting in Safari when using `Space`, so I removed that workaround. Let me know though if you have any other suggestions around this! 

2. Updated the styles so hover styles are reset on touch devices ✅

__docs__/wonder-blocks-dropdown/multi-select-variants.stories.tsx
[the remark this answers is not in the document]
beaesguerra: Removed the Khanmigo versions in the stories! We can add back later easily if we support it later on!

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -97,74 +114,51 @@ export default class SelectOpener extends React.Component<SelectOpenerProps> {
...
-    }
+        const style = [
+            styles.shared,
+            stateStyles.default,
+            disabled && stateStyles.disabled,
+        ];
 
-    render(): React.ReactNode {
         return (
-            <__RouterContext.Consumer>
```
beaesguerra, line 165: There weren't any router related props for SelectOpener so I removed this while removing ClickableBehavior. I'm assuming we had this previously since ClickableBehavior needed a `router`, but let me know otherwise!

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -229,44 +223,68 @@ const _generateStyles = (
...
-                borderColor: error ? tokens.color.red : tokens.color.fadedBlue,
-                borderWidth: tokens.border.width.thin,
-                color: placeholder
-                    ? mix(tokens.color.white32, tokens.color.blue)
-                    : tokens.color.fadedBlue,
-                backgroundColor: error
-                    ? tokens.color.fadedRed
-                    : tokens.color.activeBlue,
+                ":hover:not([aria-disabled=true])": focusHoverStyling,
+                ["@media not (hover: hover)"]: {
```
jandrade, line 243: **suggestion:** I think it would be good adding a comment here explaining why we add this custom media query.

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -82,7 +80,26 @@ export default class SelectOpener extends React.Component<SelectOpenerProps> {
...
     };
 
-    renderClickableBehavior(router: any): React.ReactNode {
+    handleKeyDown: (e: React.KeyboardEvent) => void = (e) => {
+        const keyCode = e.key;
+        // Prevent default behavior for Enter key. Without this, the select
+        // is only open while the Enter key is pressed.
+        // Prevent default behavior for Space key. Without this, Safari stays in
+        // active state visually
+        if (keyCode === "Enter" || keyCode === " ") {
```
jandrade, line 89: **praise:** Thanks for adding comments per browser here! it's super helpful for understanding why we have to do this.

jandrade (approved):
Everything looks great! thank you so much for this improvement + all the code cleanup 👏 :shipit:

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
beaesguerra: I updated this PR to also always show the pressed styling when Enter/Space is pressed!

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -66,23 +64,59 @@ type DefaultProps = {
     isPlaceholder: SelectOpenerProps["isPlaceholder"];
 };
 
+type SelectOpenerState = {
+    pressed: boolean;
```
jandrade, line 68: **suggestion:** It would be valuable adding a comment explaining the context behind this decision (or why we only store state for pressed but not for the other states).

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -97,74 +131,52 @@ export default class SelectOpener extends React.Component<SelectOpenerProps> {
...
-                    );
-                }}
-            </ClickableBehavior>
-        );
-    }
+        const style = [
+            styles.shared,
+            stateStyles.default,
+            disabled && stateStyles.disabled,
+            this.state.pressed && stateStyles.pressed,
```
jandrade, line 150: **suggestion(change-required):** This also needs to check that the opener is not `disabled`, otherwise this will display the pressed/active state when the control is disabled and the user presses `Enter/Space`.

jandrade (approved):
Looks great! Just requiring a minor change to fix one behavior with disabled + pressed, but your last changes look great overall. Thank you so much for all the work you've put in this PR :) 🎉

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
beaesguerra: Good idea - updated!

jandrade (approved):

beaesguerra:
The parent pull-request ([#2243](https://github.com/Khan/wonder-blocks/pull/2243)) has been merged into `main`, but this branch (`dropdown-states`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #2245, opened 2024-06-05
WB-1092: Make id prop optional in TextField

jandrade:
## Summary:

- Changed `id` prop to be optional.
- Use `IDProvider` to automatically generate the `id` if it’s not provided by
the caller.

Issue: WB-1092

## Test plan:

1. In Storybook, verify that the Default story auto-generates the `id` for the
input field: `?path=/story/packages-form-textfield--default`.
Screenshot 2024-06-05 at 4 30 53 PM

2. Modify the Default story to provide an `id` prop and verify that the `id` is
now used in the input field: `?path=/story/packages-form-textfield--default&args=id:some-custom-id`.
Screenshot 2024-06-05 at 4 31 49 PM

### DIFF
--- a/.changeset/chatty-crews-visit.md
+++ b/.changeset/chatty-crews-visit.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/wonder-blocks-form": major
+"@khanacademy/wonder-blocks-search-field": patch
+---
+
+Make `id` prop optional in `TextField`.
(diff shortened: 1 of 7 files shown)


jeresig (approved):
Great, thank you!

.changeset/chatty-crews-visit.md
```diff
@@ -0,0 +1,6 @@
+---
+"@khanacademy/wonder-blocks-form": major
```
beaesguerra, line 2: (As we discussed on Slack) This could be a `minor` update since changing a prop from being required to optional won't be a breaking change to existing callers. It could also remain a `major` update if we want to emphasize there's a change to the external API!

beaesguerra (approved):
Changes look good to me - just added a comment regarding versioning!

PR #2247, opened 2024-06-06
Fix hydration error in MediaLayout component

somewhatabstract:
## Summary:
In working with React 18, I identified some hydration issues in our main website. One of these is in how the `MediaLayout` component renders. Previous to this change, it would render one way on the server and then another locally. Now we use the `WithSSRPlaceholder` to ensure we retain the same structure on initial render, whether on the server or on the client.

It's a reasonable simple change - I've done my best to leave the original code intact to minimize the change.

I have also updated the tests:
- to remove promise use (when we update the Testing Library and React in Wonder Blocks, the promise approach won't work right)
- to fix the tests that were failing after updating things to properly hydrate

Issue: FEI-5498

## Test plan:
`yarn test`

Integrating this into webapp, and verifying that the hydration issue is resolved.

### DIFF
--- a/packages/wonder-blocks-layout/src/components/media-layout.tsx
+++ b/packages/wonder-blocks-layout/src/components/media-layout.tsx
@@ -2,7 +2,7 @@ import * as React from "react";
 import type {StyleDeclaration} from "aphrodite";
 import type {StyleType} from "@khanacademy/wonder-blocks-core";
 
-import {Server} from "@khanacademy/wonder-blocks-core";
+import {WithSSRPlaceholder} from "@khanacademy/wonder-blocks-core";
 import MediaLayoutContext from "./media-layout-context";
 import type {MediaSize, MediaSpec} from "../util/types";
 import type {Context} from "./media-layout-context";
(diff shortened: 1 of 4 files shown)


packages/wonder-blocks-layout/src/components/media-layout.tsx
```diff
@@ -138,15 +138,10 @@ class MediaLayoutInternal extends React.Component<CombinedProps, State> {
...
-    // available.
-    isServerSide() {
-        return (
-            Server.isServerSide() ||
-            typeof window === "undefined" ||
-            !window.matchMedia
-        );
+    // We assume that we're running on an unsupported environment) if there is
+    // no window object or matchMedia function available.
+    isUnsupportedEnvironment() {
```
somewhatabstract, line 143: **note:** Renamed to reflect the purpose. I'm not certain we need this anymore, but wanted to keep the changes minimal.

packages/wonder-blocks-layout/src/components/media-layout.tsx
```diff
@@ -209,7 +204,7 @@ class MediaLayoutInternal extends React.Component<CombinedProps, State> {
         return mockStyleSheet;
     }
 
-    render(): React.ReactNode {
+    renderContent(initialRender: boolean): React.ReactNode {
```
somewhatabstract, line 207: **note:** This is the old render function but with a parameter to indicate whether it's the initial render or not. This way we can use the same code and just invoke it from the `WithSSRPlaceholder`.

packages/wonder-blocks-layout/src/components/media-layout.tsx
```diff
@@ -239,14 +234,22 @@ class MediaLayoutInternal extends React.Component<CombinedProps, State> {
...
         return children({mediaSize, mediaSpec, styles});
     }
+
+    render() {
+        return (
+            <WithSSRPlaceholder placeholder={() => this.renderContent(true)}>
+                {() => this.renderContent(this.isUnsupportedEnvironment())}
+            </WithSSRPlaceholder>
+        );
+    }
```
somewhatabstract, line 252: **note:** This now ensures we'll hydrate correctly. The initial "placeholder" render will do the old server-side behavior, and then the subsequent "children" render will do the old behavior based on window and matchMedia availability.

jandrade (approved):
Makes sense to me! 🚢

PR #2249, opened 2024-06-06
Add support for an icon in the Banner component

beaesguerra:
## Summary:
Adds support for an icon (phosphor or custom) to the Banner component. This overrides the default icon used to represent the kind of Banner

Issue: WB-1447

## Test plan:
- Review Banner documentation (`?path=/docs/packages-banner--docs`)
  - new `icon` prop documentation
  - the example story descriptions for `With Phosphor Icon` and `With Custom Icon` 
- Verify that the Banner works with a PhosphorIcon (`?path=/story/packages-banner--with-phosphor-icon`)
- Verify that the Banner works with a custom icon (`?path=/story/packages-banner--with-custom-icon`)
- Verify that the aria-label of the icon matches the `kind` prop ("info", "success", "warning", "critical")
  - Note: I created WB-1713 since the aria label for the icon doesn't support translations at the moment

## Screenshots
Banner with Phosphor Icon
Banner with Phosphor Icon

Banner with Custom Icon
Banner with Custom Icon

### DIFF
--- a/__docs__/wonder-blocks-banner/banner.stories.tsx
+++ b/__docs__/wonder-blocks-banner/banner.stories.tsx
@@ -1,6 +1,7 @@
 import * as React from "react";
 import {StyleSheet} from "aphrodite";
 import type {Meta, StoryObj} from "@storybook/react";
+import magnifyingGlass from "@phosphor-icons/core/regular/magnifying-glass.svg";
 
 import Button from "@khanacademy/wonder-blocks-button";
 import {View} from "@khanacademy/wonder-blocks-core";
@@ -13,6 +14,7 @@ import Banner from "@khanacademy/wonder-blocks-banner";
 import BannerArgTypes from "./banner.argtypes";
 import ComponentInfo from "../../.storybook/components/component-info";
 import packageConfig from "../../packages/wonder-blocks-banner/package.json";
+import crownIcon from "../wonder-blocks-icon/icons/crown.svg";
 
 type StoryComponentType = StoryObj<typeof Banner>;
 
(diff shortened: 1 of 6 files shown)


__docs__/wonder-blocks-banner/banner.stories.tsx
```diff
@@ -468,6 +470,68 @@ export const WithMixedActions: StoryComponentType = {
...
+ * Icons](https://phosphoricons.com/) library.
+ *
+ * To use a Phosphor icon, you can use the following syntax:
+ *
+ * ```jsx
+ * import magnifyingGlass from "@phosphor-icons/core/regular/magnifying-glass.svg";
+ * <Banner icon={magnifyingGlass} layout="floating" text="text" />
+ * ```
+ *
+ * __Accessibility__: The icon chosen for the banner is decorative and
```
beaesguerra, line 488: Let me know what your thoughts are on letting the `kind` prop be the `aria-label` for the icon! (Semi-related: I created WB-1713 for supporting translations for this aria-label)

An alternative option I was considering was adding a new prop for the aria-label of the icon so that devs can provide their own aria-label. After seeing the examples of icons that could be used in the banner (like the clock icon or Khanmigo icon), these seem to be more decorative and the `kind` label is more helpful to communicate the status

jandrade: hmmm good point. I think mapping by default to the `kind` prop makes sense, but I also see your point about using a separate `aria-label` prop.

So yeah, I like of your idea about adding a new `aria-label` for the icon (I understand it would still default to `kind`, right?).

jandrade (approved):
Looks great overall! Thanks for adding the tests 💯 🚀

__docs__/wonder-blocks-banner/banner.stories.tsx
beaesguerra: I'm wondering if the `kind` should always be there and communicated to screen readers since we only use colour to communicate that the message is critical/warning/info/success. Or would the guideline be that if teams provide their own `aria-label` for the icon, it should include details about the banner `kind`?

jandrade: I think they should rely on the default `aria-label` but they could optionally override it if they want (or need to). That means that we could create an `iconAriaLabel` (or such) optional prop in a separate ticket as a lower priority. wdyt?

beaesguerra: That sounds good to me! I created WB-1715 for this, thanks for the feedback!

PR #2250, opened 2024-06-11
Disable keyboard interactions for DropdownCore if component is disabled

beaesguerra:
## Summary:
- Remove storybook controls for SingleSelect disabled story. Because the storybook args included opened:false, the select was always closed, which was hiding a bug where the disabled select could still be opened using keyboard arrows
- Add disabled prop to DropdownCore so that it's interactions can be disabled if it is true. This is needed now that the SelectOpener does not have the disabled attribute set (since we use aria-disabled now so that it can be focusable)
- Add tests to make sure event handlers aren't called when a component is disabled
- Prevent Single and Multi Select from being in an open state when the `opened` prop and `disabled` prop are both true

Issue: WB-1238

## Test plan:
- Confirm that the SingleSelect and MultiSelect cannot be opened when disabled, especially by using the down arrow key to open the listbox
  - `?path=/story/packages-dropdown-singleselect--disabled`
  - `?path=/story/packages-dropdown-multiselect--disabled`

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/__tests__/select-opener.test.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/__tests__/select-opener.test.tsx
@@ -5,7 +5,7 @@ import {userEvent} from "@testing-library/user-event";
 import SelectOpener from "../select-opener";
 
 describe("SelectOpener", () => {
-    describe("onClick", () => {
+    describe("onOpenChanged", () => {
         const children = "text";
         it("should trigger using the mouse", async () => {
             // Arrange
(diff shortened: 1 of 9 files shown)


packages/wonder-blocks-dropdown/src/components/multi-select.tsx
jandrade, line 1: suggestion: This makes me realize that we should force the listbox/dropdown to be closed when it is `disabled` even if `open/opened` is set to `true`.

Right now, if I use the `Default` story and turn on both the `opened` and `disabled` controls, then the listbox will be shown, but to me it looks incorrect as disabled should take more precedence (IMO). What are your thoughts on that?

https://5e1bf4b385e3fb0020b7073c-nhmqyuvdyi.chromatic.com/iframe.html?args=opened:!true;disabled:!true&id=packages-dropdown-singleselect--default&viewMode=story

beaesguerra: That's a good point! I've updated it so the derived open state will always be false if the `disabled` prop is true!

packages/wonder-blocks-dropdown/src/components/__tests__/select-opener.test.tsx
jandrade, line 1: **praise:** Thanks for adding all the tests (including the `open + disabled` case.

jandrade (approved):
LGTM! Thanks for fixing this :shipit:

PR #2252, opened 2024-06-13
Fixing edge case for opening the Single/MultiSelect when there are no items

beaesguerra:
## Summary:
- Fixing bug where the SingleSelect and MultiSelect components can be opened via keyboard when it is disabled because it has 0 items. The select components should not be able to be opened when the select opener is disabled
- Update SingleSelect to have the same behaviour as MultiSelect: if all option items are disabled, the select should be disabled

Issue: WB-1238

## Test plan:
- Confirm when there are no items, the select components should not be able to be opened using the keyboard
  - `?path=/story/packages-dropdown-singleselect--disabled`
  - `?path=/story/packages-dropdown-multiselect--disabled`

### DIFF
--- a/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
@@ -20,6 +20,7 @@ import multiSelectArgtypes from "./multi-select.argtypes";
 import {defaultLabels} from "../../packages/wonder-blocks-dropdown/src/util/constants";
 import {allCountries, allProfilesWithPictures} from "./option-item-examples";
 import {OpenerProps} from "../../packages/wonder-blocks-dropdown/src/util/types";
+import Strut from "../../packages/wonder-blocks-layout/src/components/strut";
 
 type StoryComponentType = StoryObj<typeof MultiSelect>;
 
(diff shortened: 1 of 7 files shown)


__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
```diff
@@ -385,18 +386,40 @@ export const DropdownInModal: StoryComponentType = {
 
 /**
  * `MultiSelect` can be disabled by passing `disabled={true}`. This can be
- * useful when you want to disable a control temporarily.
+ * useful when you want to disable a control temporarily. It is also disabled
+ * when:
+ * - there are no items
+ * - there are items and they are all disabled
+ *
```
jandrade, line 393: **praise:** This is great! thanks for including docs about these edge cases.

packages/wonder-blocks-dropdown/src/components/single-select.tsx
```diff
@@ -456,10 +455,17 @@ export default class SingleSelect extends React.Component<Props, State> {
             disabled,
         } = this.props;
         const {searchText} = this.state;
-        const allChildren = React.Children.toArray(children).filter(Boolean);
-        // @ts-expect-error [FEI-5019] - TS2345 - Argument of type '(ReactChild | ReactFragment | ReactPortal)[]' is not assignable to parameter of type 'ReactElement<{}, string | JSXElementConstructor<any>>[]'.
+        const allChildren = (
```
jandrade, line 458: **praise:** oh great!

jandrade (approved):
Looks really great! 🚢

packages/wonder-blocks-dropdown/src/components/single-select.tsx
beaesguerra: Got it from the [similar line for MultiSelect](https://github.com/Khan/wonder-blocks/pull/2252/files#diff-f7d0f4de4f6447c1f878dd6c51e6646adae276658cf9d7f29fcea7f630b17430R560)!

PR #2255, opened 2024-06-19
Communicate checked status to screenreaders when OptionItems are used in ActionMenu

beaesguerra:
## Summary:
When OptionItems are used within an ActionMenu, they should have role="menuitemcheckbox" and use aria-checked instead of aria-selected

Issue: WB-1659

## Test plan:
- Confirm that an ActionMenu with OptionItems communicates when an option item is selected (`?path=/story/packages-dropdown-actionmenu--with-option-items`)
  - rendered option items should have role="menuitemcheckbox" and aria-checked="true"
  - screenreaders communicate the checked status

### Before & After
In the before videos, the selected option item in the menu is not communicated to the user using a screen reader. In the after videos, it is communicated (safari reads out "Checkmark", Chrome reads out "checked")

#### Safari

https://github.com/Khan/wonder-blocks/assets/14334617/114ac914-0ece-4e65-a886-987e8ba2f163

https://github.com/Khan/wonder-blocks/assets/14334617/b54486bb-ec8a-4fc0-a38b-1093267782ff

#### Chrome

https://github.com/Khan/wonder-blocks/assets/14334617/24dc0246-34a4-4c1f-a5d8-26386bb55619

https://github.com/Khan/wonder-blocks/assets/14334617/bfb849c7-6dab-44a5-8a0e-5b442b167796

### DIFF
--- a/.changeset/small-avocados-complain.md
+++ b/.changeset/small-avocados-complain.md
@@ -1,10 +1,14 @@
 ---
 "@khanacademy/wonder-blocks-clickable": patch
 "@khanacademy/wonder-blocks-dropdown": patch
-"@khanacademy/wonder-blocks-cell": patch
+"@khanacademy/wonder-blocks-cell": minor
 ---
 
 Improves accessibility of the checked status on `OptionItem` components used
 within the `ActionMenu` component. The checked status is communicated to
 screenreaders by using a `menuitemcheckbox` role with the `aria-checked`
 attribute (instead of `aria-selected`).
+    - `CellCore` (used by `CompactCell` and `DetailCell`) has a new optional
+    prop for `aria-checked`
+    - `ClickableRole` type now supports the `menuitemcheckbox` role
+    - `OptionItem`'s `role` prop now also supports the `menuitemcheckbox` role
--- a/.changeset/small-avocados-complain.md
+++ b/.changeset/small-avocados-complain.md
@@ -1,0 +1,10 @@
+---
+"@khanacademy/wonder-blocks-clickable": patch
+"@khanacademy/wonder-blocks-dropdown": patch
+"@khanacademy/wonder-blocks-cell": patch
+---
+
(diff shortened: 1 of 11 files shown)


packages/wonder-blocks-dropdown/src/components/action-menu.tsx
```diff
@@ -203,14 +203,18 @@ export default class ActionMenu extends React.Component<Props, State> {
...
                 return {
                     ...itemObject,
                     populatedProps: {
                         onToggle: this.handleOptionSelected,
-                        selected: selectedValues
-                            ? selectedValues.includes(value)
-                            : false,
+                        selected,
                         variant: "check",
+                        role: "menuitemcheckbox",
```
beaesguerra, line 215: This is where the main change is! When an ActionMenu is used, we set these attributes for OptionItem components. We set these attributes here since this should only apply to OptionItems in an ActionMenu (not when it is used in a Single/MultiSelect). 

The remaining changes make it so these attributes are applied/received as props

jandrade (approved):
Looks good to me! One minor thing that I noticed is that `aria-checked` is not correctly set in the Storybook Docs section. Could you please fix that before landing?

Thanks for fixing this 🚢

.changeset/small-avocados-complain.md
```diff
@@ -0,0 +1,10 @@
+---
+"@khanacademy/wonder-blocks-clickable": patch
+"@khanacademy/wonder-blocks-dropdown": patch
+"@khanacademy/wonder-blocks-cell": patch
```
jandrade, line 4: **suggestion:** One more thing that I missed. `wb-cell` should be `minor` here as you are adding support to a new prop.

beaesguerra: Good catch - Updated! I also added more details to the changeset!

jandrade (approved):
Thanks! Love the level of detail in the changesets file.

PR #2259, opened 2024-06-24
Fix active state styling for tertiary button

beaesguerra:
## Summary:
Use the same theme token for the height of the underline on the active state of the tertiary button so that the hover and active styling are consistent

Issue: WB-1025

## Test plan:
- Confirm that the hover and active state of the tertiary button have a consistent underline thickness

## Screen Recording:
Before & After

https://github.com/Khan/wonder-blocks/assets/14334617/2ce3725e-48f1-49f8-a111-761d1ce48944

### DIFF
--- a/packages/wonder-blocks-button/src/components/button-core.tsx
+++ b/packages/wonder-blocks-button/src/components/button-core.tsx
@@ -495,7 +495,7 @@ export const _generateStyles = (
             active: {
                 color: light ? fadedColor : activeColor,
                 ":after": {
-                    height: 1,
+                    height: theme.size.height.tertiaryHover,
                     background: light ? fadedColor : activeColor,
                 },
             },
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -495,7 +495,7 @@ export const _generateStyles = (
             active: {
                 color: light ? fadedColor : activeColor,
                 ":after": {
-                    height: 1,
+                    height: theme.size.height.tertiaryHover,
```
beaesguerra, line 498: Using the [same token as the height of the underline in the hover state](https://github.com/Khan/wonder-blocks/pull/2259/files#diff-ede6855e27a1ecfe88d6f2e68ba1979e36ea896776f7e023078b353df3b84b94R479)!

jandrade (approved):
Looks great! :shipit:

PR #2260, opened 2024-06-25
Fix Banner styling for custom action buttons

beaesguerra:
## Summary:
Fixing styling issue where non-tertiary buttons get cropped when used as a custom action in the banner component.

Issue: WB-1716

## Test plan:
- Confirm buttons look as expected in the banner (`?path=/story/packages-banner--with-custom-action-primary`)
- Confirm a mix of actions and custom actions look as expected in the banner (`?path=/story/packages-banner--with-mixed-actions`)

## Screenshots:
Mixed actions (before & after changes)
before-mixed
after-mixed

Primary button for custom action (before & after changes) 
before-primary
after-primary

### DIFF
--- a/packages/wonder-blocks-banner/src/components/banner.tsx
+++ b/packages/wonder-blocks-banner/src/components/banner.tsx
@@ -204,7 +204,7 @@ const Banner = (props: Props): React.ReactElement => {
             if (action.type === "custom") {
                 return (
                     <View style={styles.action} key={`custom-action-${i}`}>
-                        <View style={styles.customAction}>{action.node}</View>
+                        {action.node}
                     </View>
                 );
             }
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-banner/src/components/banner.tsx
```diff
@@ -204,7 +204,7 @@ const Banner = (props: Props): React.ReactElement => {
             if (action.type === "custom") {
                 return (
                     <View style={styles.action} key={`custom-action-${i}`}>
-                        {action.node}
+                        <View style={styles.customAction}>{action.node}</View>
```
jandrade, line 207: **question:** Would it be possible to merge this style in the parent `View`? I'm not sure if that will affect applying `flexShrink` in there? if it can be done, it would be better as we avoid introducing more divs in that component.

```
<View style={[styles.action, styles.customAction]} key...>{action.node}</div>
```

beaesguerra: I found another solution that doesn't require adding an extra div! Thanks for the suggestion of moving the height to the actions container :D

jeresig (approved):
Thank you for the fixes, Bea!

PR #2262, opened 2024-06-27
Set up jest-axe for a11y testing

beaesguerra:
## Summary:
- Configure `jest-axe`
- Adds `toHaveNoA11yViolations` matcher for jest tests
- Configure jest await async lint rule for toHaveNoA11yViolations matcher. This is so we don't forget to await the a11y checks. Also fixing some linting issues that were caught when this rule was enabled
- Adds an example a11y test for Title component

Note: this set up is similar to how `jest-axe` is set up in webapp (I referenced this [jest-axe webapp PR](https://github.com/Khan/webapp/pull/5850)!)

Issue: WB-466

## Test plan:
- Run unit tests and linting to make sure there are no errors
- Manually tested jest-axe is working by making sure that it catches expected a11y errors locally. Used this example with headings that are out of order:
```tsx
test("expect a11y test to fail", async () => {
  // Arrange
  // Act
  const {container} = render(
      <div>
        <h1>title1</h1>
        <h3>title2</h3>
      </div>,
  );

  // Assert
  await expect(container).toHaveNoA11yViolations();
});
```


jeresig (approved):
These changes all look reasonable to me, thank you Bea!

PR #2263, opened 2024-06-27
add aria label to view component

anakaren-rojas:
## Summary:
<Add your summary here>

Issue: LEMS-2089

## Test plan:

### DIFF
--- a/packages/wonder-blocks-popover/src/components/popover-content-core.tsx
+++ b/packages/wonder-blocks-popover/src/components/popover-content-core.tsx
@@ -70,6 +70,7 @@ export default class PopoverContentCore extends React.Component<Props> {
 
     render(): React.ReactNode {
         const {
+            "aria-label": ariaLabel,
             children,
             closeButtonLight,
             closeButtonLabel,
@@ -87,6 +88,7 @@ export default class PopoverContentCore extends React.Component<Props> {
                     color !== "white" && styles[color],
                     style,
                 ]}
+                aria-label={ariaLabel}
             >
                 {closeButtonVisible && (
                     <CloseButton


packages/wonder-blocks-popover/src/components/popover-content-core.tsx
```diff
@@ -70,6 +70,7 @@ export default class PopoverContentCore extends React.Component<Props> {
 
     render(): React.ReactNode {
         const {
+            "aria-label": ariaLabel,
```
catandthemachines, line 73: Oh interesting, so the ariaLabel prop is exposed, but the value is not piped through to View component??

catandthemachines:
Could you add a unit test that ensures the ariaLabel is set when passed in?

catandthemachines (approved):

jeresig (approved):
Ahh - sounds good, thanks!

packages/wonder-blocks-popover/src/components/popover-content-core.tsx
anakaren-rojas: Yup it accepts all the `aria-*` labels by default, but we have to expose them individually @catandthemachines

packages/wonder-blocks-popover/src/components/__tests__/popover.test.tsx
```diff
@@ -582,6 +582,41 @@ describe("Popover", () => {
                 }),
             ).toBeInTheDocument();
         });
+
+        it("should correctly describe the popover content core's arial label", async () => {
```
anakaren-rojas, line 586: Added a unit test for the popover content core's aria label check in the `popover.test` file - hope that works @catandthemachines @jeresig

jeresig, line 586:
```suggestion
        it("should correctly describe the popover content core's aria label", async () => {
```

jeresig (approved):
Oh yes, good idea, thank you

PR #2265, opened 2024-07-09
fix(LEMS-2183): Add optional aria label to popover

anakaren-rojas:
## Summary:
Add conditional Aria Label for popover 

Issue: LEMS-2183

#### Running local: 
1. `yarn start` to run story book 
OR 
#### Chromatic link: 
1. [Chromatic link](https://5e1bf4b385e3fb0020b7073c-ojvdqqxwvd.chromatic.com/?path=/story/packages-popover-popover--with-custom-aria-label)
2. Open the "With Custom Aria Label" story
3. Enter a custom aria label

#### Expected behavior:
- Screen reader reads out the aria label 

### Before 
https://github.com/user-attachments/assets/eeca69d5-d23c-4e8b-89a8-74f9d54b27a2

###  After 
https://github.com/user-attachments/assets/077fac81-d6c2-4cbd-b3ea-4b681866983d


catandthemachines:
I know you shared a clip of this working with Juans suggestion, do you have another clip for this implementation or some testing steps so I can test it out? 😊

jandrade:
@anakaren-rojas +1 @catandthemachines's point, it would be really useful adding a test plan.

anakaren-rojas:
Yup I'll add steps + video shortly @catandthemachines @jandrade

__docs__/wonder-blocks-popover/popover.stories.tsx
```diff
@@ -558,3 +558,21 @@ export const PopoverAlignment: StoryComponentType = {
...
+    args: {
+        children: <Button>Open popover</Button>,
+        content: ContentMappings.withTextOnly,
+        placement: "top",
+        dismissEnabled: true,
+        id: "",
+        initialFocusId: "",
+        testId: "",
+        onClose: () => {},
+        "aria-label": "",
```
catandthemachines, line 576: For the storybook case can you put in an aria-label just to highlight the experience? Thanks!

catandthemachines (approved):
Looks so awesome, I always enjoy simple crisp fixes. Ship it!

.changeset/afraid-buckets-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-popover": minor
```
jandrade, line 2: **suggestion(non-blocking):** Could you please change this to `patch` as you are fixing an issue with the existing implementation?

To clarify a bit more, the public API of Popover is not changing as `aria-label` existed already as a prop, but instead the internal implementation is being fixed by applying the correct prop/attribute to the `PopoverDialog` element.

__docs__/wonder-blocks-popover/popover.stories.tsx
jandrade, line 1: **suggestion:** Non-related to this file, but it would be helpful if you could add some documentation around this change in the `Accessibility` page. You can see the current docs here:

Docs: https://5e1bf4b385e3fb0020b7073c-ojvdqqxwvd.chromatic.com/?path=/docs/packages-popover-popover-accessibility--docs
Code: https://github.com/Khan/wonder-blocks/blob/main/__docs__/wonder-blocks-popover/popover.accessibility.mdx#L32

> ## Popover Dialog:
> The popover component will populate the [aria-labelledby](https://www.w3.org/TR/wai-aria/#aria-labelledby) and [aria-describedby](https://www.w3.org/TR/wai-aria-1.1/#aria-describedby) attributes automatically. Internally, it will be set in the dialog element and the PopoverContent component to reference the title and content elements/props. By doing this, the popover will be announced by screen readers as a dialog with a title and content.

It would be great if you could add documentation about using `aria-label` as an alternative aproach to `aria-labelledBy/aria-describedBy`.

e.g.

```
....and content elements/props.
By doing this, the popover will be announced by screen readers as a dialog with a title and content.

Alternatively, you can use `aria-label` to label the popover dialog in case there is no visible title and/or description inside the dialog.
```

@catandthemachines Do you have any thoughts about this description? something to add/change?

__docs__/wonder-blocks-popover/popover.stories.tsx
jandrade: +1 to this suggestion :)

jandrade (approved):
Looks great! I've added a few suggestions to address before landing that will help improving our documentation about this new feature.

Thanks a lot for contributing to Wonder Blocks and fixing this accessibility issue 💯 🚢

__docs__/wonder-blocks-popover/popover.accessibility.mdx
```diff
@@ -39,6 +39,9 @@ the `PopoverContent` component to reference the `title` and `content`
 elements/props. By doing this, the popover will be announced by screen readers
 as a dialog with a title and content.
 
+Alternatively, you can use `aria-label` to add screen reader accessible text to the popover dialog in case there is no
```
anakaren-rojas, line 42: Tweaked it a tiny bit @jandrade, hope thats ok
Pulling this from the other thread: 
> @catandthemachines Do you have any thoughts about this description? something to add/change?

anakaren-rojas:
Reverting previous addition of aria - described by as written [text will not be read by VoiceOver ](https://a11y-101.com/development/aria-describedby#:~:text=Text%20typed%20into%20this%20field%20will%20not%20be%20displayed%20and%20will%20not%20be%20spoken%20by%20VoiceOver)(Mac screen reader)

PR #2266, opened 2024-07-10
[wbt.react18.updates] Add renderHookStatic and error boundary adapter

somewhatabstract:
## Summary:
This turned out to be a bit of a yak shave, because I chose to update the Storybook version for Wonder Blocks Testing and add some missing deps from the main repo, which then led to a whole host of exciting adventures.

Anyway, this updates Storybook a bit, including some types. It also addresses a failure with `alex` and reworks the `yarn test` command so that it gives important errors a little sooner (like running jest first, rather than everything else, and running alex last).

It also creates a new package, Wonder Blocks Testing Core, which is everything that was in Wonder Blocks Testing, bar anything that took a dependency on any other part of Wonder Blocks. This means that other Wonder Blocks packages can take a dependency on Wonder Blocks Testing Core, which will be important because...

...this also adds a new `renderHookStatic` function to Wonder Blocks Testing Core. This is a function that allows you to render a hook in a test harness, and will replace the deprecated and soon-to-be-removed `renderHook` from `@testing-library/react-hooks/server`. Also, Wonder Blocks Testing Core adds the new `boundary` adapter, which is a harness adapter that acts as an error boundary. Both of these are useful in the migration to React 18 and above as the old ways of doing things won't work. The next PR will update Wonder Blocks Core and Wonder Blocks Data to test with these new features.

Wonder Blocks Testing has been updated to use Wonder Blocks Testing Core, and export all the same things it used to, but with the ssr and data adapters that require dependencies on other parts of Wonder Blocks.

Issue: WB-1730

## Test plan:
`yarn test`
`yarn jest`
`yarn start:storybook` and check things are working (I did a lot of this)


.alexrc
```diff
@@ -1,3 +1,3 @@
 {
-    "allow": ["invalid", "color", "colors", "white", "latin", "hook", "hooks", "trap"]
+    "allow": ["invalid", "color", "colors", "white", "latin", "hook", "hooks", "trap", "whitespace", "just", "special"]
```
somewhatabstract, line 2: **note:** This covers the warnings that were failing alex. The alternative is to rephrase the CHANGELOGs that were causing the issues.

tsconfig-common.json
```diff
@@ -24,6 +28,6 @@
         // "noImplicitAny": true,
         /* Completeness */
         "skipDefaultLibCheck": true, // it's safe to assume that built-in .d.ts files are correct
-        "skipLibCheck": false
+        "skipLibCheck": true
```
somewhatabstract, line 31: **note:** This stops us typechecking node modules.

tsconfig-common.json
```diff
@@ -2,6 +2,10 @@
 /* Visit https://aka.ms/tsconfig to read more about this file */
 {
     "compilerOptions": {
+        "typeRoots": [
+            "./node_modules/@types",
+            "./types"
+        ],
```
somewhatabstract, line 8: **note:** Copied from webapp. Not sure it's necessary, but hasn't caused any harm.

package.json
```diff
@@ -51,22 +51,24 @@
...
+    "@storybook/test": "^8.2.1",
     "@swc-node/register": "^1.6.5",
     "@swc/core": "^1.3.36",
-    "@testing-library/jest-dom": "^5.16.5",
+    "@testing-library/jest-dom": "^6.4.6",
     "@testing-library/react": "^12.1.2",
     "@testing-library/react-hooks": "^7.0.2",
     "@testing-library/user-event": "^14.5.1",
-    "@types/jest": "29",
+    "@types/jest": "^29.5.12",
```
somewhatabstract, line 71: **note:** I updated jest while trying to figure out why it kept failing.

tsconfig-common.json
jandrade: oh I see, that makes sense.

jandrade (approved):
Looks great! thanks for all these refactors + bringing the changes from webapp. This will definitively unblock us from upgrading to v18 👏 🚀

PR #2268, opened 2024-07-10
[wbt.react18.updates.2] Update data and core packages to use renderHookStatic

somewhatabstract:
## Summary:
This is a pre-cursor change to the React 18 work. Here we migrate any uses of `@testing-library/react-hooks/server` imports to use our own `renderHookStatic` function. It's not quite a drop-in replacement, but good enough.

Issue: WB-1730

## Test plan:
`yarn jest`


jandrade (approved):
🎉 🚀 Awesome! this PR is pretty straightforward (after making the big changes in the first PR).

PR #2269, opened 2024-07-11
Allow custom keyboard navigation in Popover

catandthemachines:
## Summary:
Updating Popover to not override custom focus behavior for content inside Popover. Also updating component to not portal Popover into DOM and instead insert it into the DOM along side the trigger. Ideally to ensure all focus functionality works without the need of FocusManager component.

Issue: LEMS-2129

## Test plan:
Go to url: ?path=/story/packages-popover-popover--custom-keyboard-navigation
See that the popover functionality works while allowing the user to only use arrow key navigation for content inside.


__docs__/wonder-blocks-popover/popover.stories.tsx
```diff
@@ -517,6 +519,182 @@ export const KeyboardNavigation: StoryComponentType = {
...
+ *  inside the popover.
+ * - When the popover is closed, the focus is returned to the element that
+ * triggered the popover.
+ * - If the popover is opened and the focus reaches the last focusable element
+ * inside the popover, the next tab will set focus on the next focusable
+ * element that exists after the PopoverAnchor (or trigger element).
+ * - If the focus is set to the first focusable element inside the popover, the
+ * next shift + tab will set focus on the PopoverAnchor element.
+ * - If you have custom keyboard navigation (like with left and right arrow keys)
+ * popover won't override them
```
anakaren-rojas, line 538: nit: Would this explanation be better suited for the `popover.accessibility.mdx` file? 
It has good context and would allow us to remove some duplicate comments (between this and the regular `KeyboardNavigation` story)

packages/wonder-blocks-popover/src/components/popover.tsx
```diff
@@ -101,6 +102,8 @@ type Props = AriaProps &
          * Whether to show the popover tail or not. Defaults to true.
          */
         showTail: boolean;
+
+        disablePortal?: boolean;
```
jandrade, line 110: **suggestion:** This could also be created in a different way... defining a  required `portal` (`true` by default) prop, then setting it as `false` when needed. This way, it would help us describing the need for this portal + explaining in which cases `false` makes sense. Another benefit is that the code could become a bit more explicit,  see the `maybeRenderPortal` example. What do you think?

related: Could you also please add comments to this prop so we can display it in Storybook? you might need to add that to `ArgTypes` as well (not fully sure about this part).

packages/wonder-blocks-popover/src/components/popover.tsx
```diff
@@ -255,32 +258,44 @@ export default class Popover extends React.Component<Props, State> {
...
-                            {this.renderContent(uniqueId)}
-                        </PopoverDialog>
-                    )}
-                </TooltipPopper>
-            </InitialFocus>
-        );
+                    {popperContent}
+                </FocusManager>
+            );
+        }
```
jandrade: **suggestion:** The `else` here could be removed here as the if stops at the return in line 290.

e.g.
```
if () {

}

return (<FocusManager....
```

packages/wonder-blocks-popover/src/components/popover.tsx
```diff
@@ -317,7 +333,14 @@ export default class Popover extends React.Component<Props, State> {
...
                             </PopoverAnchor>
-                            {opened && this.renderPopper(uniqueId)}
+                            {disablePortal
+                                ? opened && this.renderPopper(uniqueId)
+                                : popperHost &&
+                                  opened &&
+                                  ReactDOM.createPortal(
+                                      this.renderPopper(uniqueId),
+                                      popperHost,
+                                  )}
```
jandrade, line 347: **suggestion:** We tend to use separate functions to avoid these kind of complex expressions. It would be ideal moving this to improve readability:

e.g.
```
maybeRenderPortal(uniqueId: string) {
  const {portal} = this.props;

  if (!opened) {
    return null;
  }

  // Attach the dialog to a Portal
  const popperHost = this.getHost();
  if (portal && popperHost) {
    return ReactDOM.createPortal(
      this.renderPopper(uniqueId),
      popperHost,
    );
  }

  // Otherwise, append the dialog next to the trigger element
  return this.renderPopper(uniqueId);
}

**Review by jandrade — COMMENTED** [review]
The implementation makes sense. I'm leaving some comments for you to consider about taking a different approach in the naming for that prop. Let me know what you think.

File: packages/wonder-blocks-popover/src/components/popover.tsx (hunk 3 above)
Comment (catandthemachines) [reply packages/wonder-blocks-popover/src/components/popover.tsx]: Very true, I just enjoy being explicit with my return statements in an if/else, though you are correct it's not strictly neccesary. Let me know if you would prefer I remove the else.

File: packages/wonder-blocks-popover/src/components/popover.tsx (hunk 2 above)
Comment (catandthemachines) [reply packages/wonder-blocks-popover/src/components/popover.tsx]: Makes sense, updated the code and added a comment to the prop. Let me know what you think!

File: packages/wonder-blocks-popover/src/components/__tests__/popover.test.tsx
Comment (catandthemachines) [inline packages/wonder-blocks-popover/src/components/__tests__/popover.test.tsx:1]: GitHub is making it look live everything changed with these tests, but the main thing is I updated each keyboard navigation test to test the portal and non-portal implementation and confirm that both have the expected keyboard navigation. 😉

**Review by anakaren-rojas — APPROVED** [review]

File: packages/wonder-blocks-popover/src/components/popover.tsx [hunk 5]
```diff
@@ -101,6 +102,20 @@ type Props = AriaProps &
...
+         * This is very handy in cases where the Popover can't be easily
+         * injected into the DOM structure and requires portaling to
+         * the trigger location.
+         *
+         * Set to "true" by default.
+         *
+         * CAUTION: Turning off portal could cause some clipping issues
+         * especially around legacy code with usage of z-indexing,
+         * Use caution when turning this functionality off and ensure
+         * your content does not get clipped or hidden.
```
jandrade, line 116: Thanks for adding the "Caution" section! 🙏. Could you please add this to `popover.argtypes.ts` as well?

It's unfortunate that we have to duplicate these comments, but we have to do it due to some issues with `react-docgen` not being able to extract the code comments when we use union types.

You can see how the `portal` docs don't show up here: https://5e1bf4b385e3fb0020b7073c-wphvlpzdmq.chromatic.com/?path=/docs/packages-popover-popover--docs

Screenshot 2024-07-17 at 6 32 33 PM

jandrade (approved):
Looks good! I left you some minor comments to address before landing.

Thanks a lot for all the iterations and adding all the docs 🚀.

PR #2271, opened 2024-07-12
Update tooltip-popper to handle smaller screensizes.

catandthemachines:
## Summary:
Adjusting implementation of Popper in tooltip-popper to handle smaller screen sizes by:
- Adjusting location of popper based on document size, not viewport size. This will ensure the popper doesn't default to inaccessible locations when the popper doesn't fit in the viewport size.
- Disabled isReferenceHidden flag to ensure the popper stays present even when the reference anchor is out of view. This will allow users in higher zoom levels or smaller screen sizes to scroll to interact with the popper if necessary instead of it disappearing when the reference isn't in view.

Issue: LEMS-2085

## Test plan:
- Go to url: https://khan.github.io/wonder-blocks/iframe.html?args=&id=packages-popover-popover--with-actions&viewMode=story
- Set the page body to have padding: "0 0 500px 0" to ensure the page body is larger than the height of the button example.
- Set the viewport size to 320 x 256
- You should see the popover attach at the bottom of the trigger element since there is more room within the page document.


jandrade:
@catandthemachines This approach seems to have a counter effect. When the Tooltip/Popover is overflowing the viewport, the tooltip is not repositioning itself and instead it appears cut off.

You can see an example of that here:

https://5e1bf4b385e3fb0020b7073c-ocepdnhrkc.chromatic.com/?path=/docs/packages-popover-popover--docs#with%20actions

Screenshot 2024-07-12 at 5 11 38 PM

While this fixes this behavior for screen sizes, it will introduce some unexpected behaviors like the one in the screenshot.

catandthemachines:
> @catandthemachines This approach seems to have a counter effect. When the Tooltip/Popover is overflowing the viewport, the tooltip is not repositioning itself and instead it appears cut off.
> 
> You can see an example of that here:
> 
> https://5e1bf4b385e3fb0020b7073c-ocepdnhrkc.chromatic.com/?path=/docs/packages-popover-popover--docs#with%20actions
> 
> Screenshot 2024-07-12 at 5 11 38 PM
> While this fixes this behavior for screen sizes, it will introduce some unexpected behaviors like the one in the screenshot.

Yes this was intentional as it appears, just my own opinion, to be better as the user can scroll to see the full content. Is this a requirement for Popover? I will be meeting with our accessibility designer to ask their opinion and any concerns she might have.

jandrade:
@catandthemachines Updating here to set expectations. As per our last conversation, this change mostly makes sense, but we still want to keep the `flip` functionality around so the Popover dialog is still visible in some situations.

packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx
```diff
@@ -172,21 +217,43 @@ export default class TooltipPopper extends React.Component<Props> {
...
+
+        const smallViewportModifier: SmallViewportModifier = {
+            name: "smallViewport",
+            enabled: true,
+            phase: "main",
+            fn: _modifyPosition,
+        };
+
+        const modifiers: (
+            | Partial<PreventOverflowModifier>
```
handeyeco, line 230: I'm confused by the use of `Partial`. Are we saying that only part of the `PreventOverflowModifier` will be used or are we trying to say that these modifiers are optional?

Like should it be:

``` TS
type Modifier = PreventOverflowModifier | FlipModifier | SmallViewportModifier;
const modifiers: Modifier[] = [smallViewportModifier];
```

Likely I'm just confused.

packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx
```diff
@@ -64,11 +78,42 @@ const filterPopperPlacement = (
...
+function _modifyPosition({state}: ModifierArguments<object>): void {
+    // Calculates the available space for the popper based on the placement
+    // relative to the viewport.
+    const popperHeight =
+        state.rects.popper.height + state.rects.reference.height;
+    const html = document.documentElement;
+    const minHeight = html.clientHeight;
+
+    if (minHeight < popperHeight && state.modifiersData.hide) {
+        state.modifiersData.hide = {
```
handeyeco, line 101: This doesn't need to use `setState` or anything? I'm guessing not, I'm guessing it's Popper state and not React state. Just thought I'd double check.

handeyeco (commented):
Looking good. I'll defer to FEI/WB folks to approve though since I know there's been some back and forth.

packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx
```diff
@@ -64,11 +78,42 @@ const filterPopperPlacement = (
     }
 };
 
+type SmallViewportModifier = Modifier<"smallViewport", object>;
```
handeyeco, line 81: I've never seen `object` used like this before. Is this basically the `any` of objects?

jandrade:
@catandthemachines would you mind adding a Storybook story explaining the use case for this new prop + including the padding necessary for recreating this experience? This one doesn't need to be included in the Chromatic snapshots.

catandthemachines:
> @catandthemachines would you mind adding a Storybook story explaining the use case for this new prop + including the padding necessary for recreating this experience? This one doesn't need to be included in the Chromatic snapshots.

Added a new storybook page, you can find it in Chromatic at `/?path=/story/packages-popover-popover--with-document-root-boundary`

__docs__/wonder-blocks-popover/popover.stories.tsx
```diff
@@ -735,6 +735,40 @@ export const PopoverAlignment: StoryComponentType = {
...
+    );
+};
+
+WithDocumentRootBoundary.parameters = {
+    docs: {
+        description: {
+            story: `Sometimes you need to add actions to be able to
+                control the popover state. For this reason, you can make use of the
+                \`actions\` prop:`,
+        },
```
jandrade, line 768: **suggestion:** Could you please update this description to match the story? Probably the popover content would be useful for this.

packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx
```diff
@@ -172,21 +223,39 @@ export default class TooltipPopper extends React.Component<Props> {
...
+        const {anchorElement, placement, rootBoundary} = this.props;
+
+        const smallViewportModifier: SmallViewportModifier = {
+            name: "smallViewport",
+            enabled: true,
+            phase: "main",
+            fn: _modifyPosition,
+        };
+
+        const modifiers: Modifiers[] = [smallViewportModifier];
```
jandrade, line 235: **suggestion:** Could you please add a TODO comment here pointing to the floating-ui Jira ticket? this would be helpful for us to track this change in the future.

```suggestion
        // TODO(WB-1680): Use floating-ui's 
        const modifiers: Modifiers[] = [smallViewportModifier];
```

packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx
```diff
@@ -172,21 +223,39 @@ export default class TooltipPopper extends React.Component<Props> {
...
     render(): React.ReactNode {
-        const {anchorElement, placement} = this.props;
+        const {anchorElement, placement, rootBoundary} = this.props;
+
+        const smallViewportModifier: SmallViewportModifier = {
+            name: "smallViewport",
+            enabled: true,
+            phase: "main",
+            fn: _modifyPosition,
+        };
```
jandrade, line 233: **suggestion:** I'd recommend to move this modifier definition at the the file level, that way the `_modifyPosition` function won't be recreated on every render.

jandrade (approved):
Looks good! I've left some suggestions to tackle before landing. Thanks for iterating over this solution! 🚢

PR #2272, opened 2024-07-12
[fixexternals] Fix how react-dom/server is imported

somewhatabstract:
## Summary:
Turns out that the rollup plugin, autoexternals, doesn't handle sub-package improts, like `react-dom/server`. However, reading the issue report on that package, [I saw a link to another](https://github.com/stevenbenisek/rollup-plugin-auto-external/issues/16#issuecomment-557172792), more up-to-date package, that does handle this. So I switched to that package, and now everything works as expected.

I did have to explicitly use the `mjs` extension for the rollup config, but that's a small price to pay for the benefits of the new package.

Issue: XXX-XXXX

## Test plan:
`yarn build` to see that the build works.

I'll also verify the packages in webapp to check everything else is fine, though the storybook should also check out.

### DIFF
--- a/.changeset/gentle-houses-listen.md
+++ b/.changeset/gentle-houses-listen.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/wb-dev-build-settings": patch
+"@khanacademy/wonder-blocks-testing-core": patch
+---
+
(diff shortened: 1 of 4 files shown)


jandrade (approved):
Looks great! and the packages got slightly smaller, so that's a win win, thanks! 🚢

PR #2274, opened 2024-07-15
Add test that exhibits Popover warning

jeremywiebe:
## Summary:

This PR doesn't intend to land anything. It does however provide a unit test that causes React to `console.error()` a warning: ``React does not recognize the `uniqueId` prop on a DOM element.`` 

Issue: "none"

## Test plan:

`yarn jest packages/wonder-blocks-popover/src/components/__tests__/popover.regression.test.tsx`


jeremywiebe:
Closing. We discovered that `JSX.Element` is typed as `any`. Thus, it is compatible with the `Popover`'s `content` prop. As a result, it seems like we can't properly restrict the `content` prop to reject arbitrary JSX. 

image

PR #2275, opened 2024-07-16
Adds TextArea component

beaesguerra:
## Summary:
- TextArea component is implemented and documentation following the [spec](https://khanacademy.atlassian.net/wiki/spaces/WB/pages/3041591727/TextArea+Implementation+Spec)
- Also exports `processStyleList` utility from wonder-blocks-core so we can use it in other wonderblocks packages (this was used in TextArea so we can apply aphrodite styles to the textarea element)

Issue: WB-654

## Test plan:
- TextArea stories are reviewed and match the spec/expected behaviour
- TextArea variants stories matches Figma design


__docs__/wonder-blocks-form/text-area.argtypes.ts
```diff
@@ -0,0 +1,126 @@
+export default {
```
beaesguerra, line 1: For the most part, Storybook was automatically extracting types and descriptions from the jsdocs for props in the component. Some notes around this:
- I didn't redefine the descriptions here so that the single source of truth can be the prop jsdocs!
- I've only configured types explicitly for the props that had unclear autogenerated types. For example, any union type like `"on" | "off"` would get extracted as `"union"`, which isn't too informative
- I categorized the event related props and the style related props since this component has a long list of props!
- (Might be helpful if we come across this later:) After wrapping the component with `forwardRef`, Storybook was no longer autogenerating docs for all the props. I found that defining the component and types a certain way fixed it: https://github.com/storybookjs/storybook/issues/8881#issuecomment-1098119636

__docs__/wonder-blocks-form/text-area.stories.tsx
```diff
@@ -0,0 +1,444 @@
...
+    title: "Packages / Form / TextArea",
+    component: TextArea,
+    parameters: {
+        componentSubtitle: (
+            <ComponentInfo
+                name={packageConfig.name}
+                version={packageConfig.version}
+            />
+        ),
+        a11y: {
```
beaesguerra, line 30: There is one remaining Storybook a11y warning, any thoughts on how to resolve it? There doesn't seem to be another element obscuring the textarea 🤔 

image

When I open the Story in a new tab on it's own and use axe DevTools, this contrast warning isn't there:
image


I've left the Storybook a11y warning there for now since it is marked as "Incomplete" rather than "Violation". Let me know if you have thoughts on this!

packages/wonder-blocks-form/src/components/text-area.tsx
```diff
@@ -0,0 +1,416 @@
...
+            // don't render an error when they are initially empty
+            if (value !== "") {
+                handleValidation(value);
+            }
+        });
+
+        // Processing style prop so we can apply aphrodite styles to the textarea
+        // element. The textarea element is used directly so that we can apply inline
+        // styles for width
+        const processedStyle = processStyleList(style);
```
beaesguerra, line 246: Similar to the TextField component, the TextArea component by default has `width: 100%`. When we provide this style to the textarea using Aphrodite, the `!important` property is applied to it. When the built-in browser resize control is used, the width stays at 100% because the Aphrodite style overrides the dynamic width set by the browser.

To get around this, we can use `<textarea>` directly and include an inline style with `width: 100%`. This enables the resize functionality. When we do this, the `style` prop for custom styling can't be applied directly to the `<textarea>` element. To continue to support the `style` prop, we process it and use the resulting `style` with the `css` function from Aphrodite, and the resulting `className` is applied to the `className` prop. The [addStyle utility](https://github.com/Khan/wonder-blocks/blob/4bebd48fb38ae1236394197ea716646b54797aab/packages/wonder-blocks-core/src/util/add-style.tsx#L39-L40) is also using `processStyleList` (that's how I found it 😄 )

For reference, the diff for this change can be found in this commit: https://github.com/Khan/wonder-blocks/pull/2275/commits/d604a49bb8718c9d2123bb6e2f1a416b478ba1e0 .

Let me know if there are any concerns with this or if there is another way to achieve applying non-important styles!

packages/wonder-blocks-form/src/components/text-area.tsx
```diff
@@ -0,0 +1,416 @@
...
+            const lightStyles = [
+                styles.light,
+                !disabled && styles.lightFocus,
+                disabled && styles.lightDisabled,
+                !!error && styles.lightError,
+            ];
+            return [...baseStyles, ...(light ? lightStyles : defaultStyles)];
+        };
+        return (
+            <div>
```
beaesguerra, line 272: Since we may be adding a character count for the textarea component, I wrapped it in a div so that we can still control the layout within the TextArea component if we add a label for it!

beaesguerra (commented):
Sorry for the large-ish PR! I can create smaller ones next time - maybe one PR for attributes, one PR for event handlers, another for documentation. Let me know if you have a preference!

packages/wonder-blocks-form/src/components/text-area.tsx
jandrade: **thought:** I'm still not sure about this implementation. I think there's a different way to tackle this:

Moving the `width: 100%` to the `div` parent container (but this would need to use `View` to work properly), but then I'm unsure if we want to add more DOM elements that the ones we need.

One benefit is that it simplifies the logic for calculating the size as it totally delegates that to the Browser internally.

How do you feel about this approach?

jandrade: To expand on the idea, it would be something like:

```
<View style={{ width: "100%"}}><StyledTextArea ... /></View>
```

beaesguerra: Ah yes you're right! That works with resizing and it is a simpler solution, thank you for the suggestion! I've made this update!

The parent container is now a View component with `width: 100%`. The View component provides `display: flex` and `align-items: stretch`, which makes the textarea expand to the full width by default. Once the textarea resize control is used, the browser is able to update the width of it since we don't have any `!important` width styling directly on it :) 

> I'm unsure if we want to add more DOM elements that the ones we need.

I think this is okay! I was wrapping the `<textarea>` in a `<div>` anyways so we can still have control over the layout within the TextArea component if we add a character count label for it later on! Then we're less likely to break layouts if we introduce this later on :) https://github.com/Khan/wonder-blocks/pull/2275#discussion_r1680185405

__docs__/wonder-blocks-form/text-area.stories.tsx
```diff
@@ -0,0 +1,444 @@
...
+
+/**
+ * If the `autoComplete` prop is set, the browser can predict values for the
+ * textarea. For more details, see the
+ * [MDN docs for the textarea attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#attributes).
+ */
+export const AutoComplete: StoryComponentType = {
+    args: {
+        autoComplete: "on",
+    },
```
jandrade, line 198: **suggestion:** I think this snapshot can be skipped from Chromatic.

__docs__/wonder-blocks-form/text-area.stories.tsx
```diff
@@ -0,0 +1,444 @@
...
+};
+
+/**
+ * When the `autoFocus` prop is set, the TextArea will be focused on page load.
+ * Try to avoid using this if possible as it is bad for accessibility.
+ */
+export const AutoFocus = () => {
```
jandrade, line 205: **suggestion:** I think this snapshot can be skipped from Chromatic.

__docs__/wonder-blocks-form/text-area.stories.tsx
```diff
@@ -0,0 +1,444 @@
...
+};
+
+/**
+ * Spell check can be enabled for the TextArea. It will be checked for spelling
+ * when you try to edit it (ie. once the textarea is focused).
+ *
+ * **Note**: Consider disabling `spellCheck` for
+ *  sensitive information (see [Security and Privacy concerns](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/spellcheck#security_and_privacy_concerns) for more details)
+ */
+export const SpellCheckEnabled: StoryComponentType = {
```
jandrade, line 260: **suggestion:** I think this snapshot can be skipped from Chromatic.

__docs__/wonder-blocks-form/text-area.stories.tsx
```diff
@@ -0,0 +1,444 @@
...
+    args: {
+        spellCheck: true,
+        value: "This exampull will be checkd fur spellung when you try to edit it.",
+    },
+};
+
+export const SpellCheckDisabled: StoryComponentType = {
```
jandrade, line 267: **suggestion:** I think this snapshot can be skipped from Chromatic.

__docs__/wonder-blocks-form/text-area.stories.tsx
```diff
@@ -0,0 +1,444 @@
...
+/**
+ * The `minlength` and `maxlength` textarea attributes can be set using the
+ * `minLength` and `maxLength` props.
+ *
+ * Note: At this time, character length requirements are not displayed as part of
+ * the Text Area component. These props are only setting the underlying HTML
+ * attributes ([minlength](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/minlength)
+ * and [maxlength](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength)).
+ */
+export const MinMaxLength: StoryComponentType = {
```
jandrade, line 353: **suggestion:** I think this snapshot can be skipped from Chromatic.

__docs__/wonder-blocks-form/text-area.stories.tsx
```diff
@@ -0,0 +1,444 @@
...
+
+/**
+ * The behaviour of the built-in resize control can be configured using the
+ * `resizeType` prop. Here are some tips:
+ * - The initial size of the TextArea can be configured using the `rows` prop.
+ * This size should be large enough for the expected user input.
+ * - Avoid having too small of a TextArea and having `resizeType=none`. This
+ * makes it difficult for users to scroll through their input.
+ */
+export const ResizeType: StoryComponentType = {
```
jandrade, line 370: **suggestion:** I think this snapshot can be skipped from Chromatic.

__docs__/wonder-blocks-form/text-area.stories.tsx
```diff
@@ -0,0 +1,444 @@
...
+                <label htmlFor="resize-default">
+                    Resize Type: default (both)
+                </label>
+                <TextArea {...args} id="resize-default" />
+            </div>
+        );
+    },
+};
+
+export const Light: StoryComponentType = {
```
jandrade, line 406: **suggestion:** Could you please add some context to this example?

__docs__/wonder-blocks-form/text-area.stories.tsx
```diff
@@ -0,0 +1,444 @@
...
+    },
+    render: ControlledTextArea,
+    parameters: {
+        backgrounds: {
+            default: "darkBlue",
+        },
+    },
+};
+
+export const CustomStyle: StoryComponentType = {
```
jandrade, line 419: **suggestion:** Could you please add some context to this example?

__docs__/wonder-blocks-form/text-area.stories.tsx
jandrade, line 1: **praise:** Overall the documentation looks great, thanks for adding links for more context 💯

packages/wonder-blocks-form/src/components/__tests__/text-area.test.tsx
```diff
@@ -0,0 +1,1077 @@
...
+            await userEvent.tab();
+
+            // Assert
+            expect(handleOnBlur).toHaveBeenCalledOnce();
+        });
+    });
+
+    describe("Accessibility", () => {
+        describe("Axe", () => {
+            test("has no accessibility violations", async () => {
```
jandrade, line 427: **praise:** Awesome! thanks for start using axe-core 👏

packages/wonder-blocks-form/src/components/__tests__/text-area.test.tsx
```diff
@@ -0,0 +1,1077 @@
...
+                );
+            });
+
+            it("should set aria-invalid to true if the validate prop returns an error message", async () => {
+                // Arrange
+                render(
+                    <TextArea
+                        value="text"
+                        onChange={() => {}}
+                        validate={() => "Error"}
```
jandrade, line 571: **suggestion:** It took me a moment to understand why ariaInvalid is true in this test. Could you please add a comment here explaining how this causes that?

packages/wonder-blocks-form/src/components/__tests__/text-area.test.tsx
```diff
@@ -0,0 +1,1077 @@
...
+                // Act
+
+                // Assert
+                const textArea = await screen.findByRole("textbox");
+                expect(textArea).toHaveAttribute("aria-invalid", "false");
+            });
+        });
+    });
+
+    describe("Validation", () => {
```
jandrade, line 602: **praise:** All these validation tests are great (along with the other tests) 🥇

packages/wonder-blocks-form/src/components/text-area.tsx
```diff
@@ -0,0 +1,401 @@
...
+            // Base styles are the styles that apply regardless of light mode
+            const baseStyles = [
+                styles.textarea,
+                typographyStyles.LabelMedium,
+                resizeType && resizeStyles[resizeType],
+            ];
+            const defaultStyles = [
+                styles.default,
+                !disabled && styles.defaultFocus,
+                disabled && styles.disabled,
```
jandrade, line 256: **thought/suggestion:** I wonder if we should rely on the `:disabled` pseudo-class instead. That way we don't need to define that here and instead let `default` define that directly. What do you think?

https://developer.mozilla.org/en-US/docs/Web/CSS/:disabled

packages/wonder-blocks-form/src/components/text-area.tsx
```diff
@@ -0,0 +1,401 @@
...
+                <StyledTextArea
+                    id={uniqueId}
+                    data-testid={testId}
+                    ref={ref}
+                    className={className}
+                    style={[...getStyles(), style]}
+                    value={value}
+                    onChange={handleChange}
+                    placeholder={placeholder}
+                    disabled={disabled}
```
jandrade, line 278: **question:** I wonder if you explored using `aria-disabled` here in case we want to allow SRs focus on the field? I'm not sure still if this would be the right approach, but interested in hearing from you as we use it in multiple WB components already.

jandrade (approved):
Looks really great! I've left some comments specially around Chromatic, and a few thoughts about some attributes, but everything looks solid. Thanks for working on this 🚀 🚢

__docs__/wonder-blocks-form/text-area.stories.tsx
```diff
@@ -0,0 +1,444 @@
...
+};
+
+export const Default: StoryComponentType = {
+    args: {
+        value: "",
+        onChange: () => {},
+    },
+};
+
+export const Controlled: StoryComponentType = {
```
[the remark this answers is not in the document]
jandrade: One more thing, this probably doesn't need a snapshot.

__docs__/wonder-blocks-form/text-area.stories.tsx
beaesguerra: Tried changing the background color and even hardcoding the border color to black (since it for sure has a high contrast with black) and was still getting this warning! I also rendered a basic unstyled `<textarea>` in a story and the warning is still there (with and without the resize control).  I also noticed this warning shows in all browsers except for Firefox. 🤷 Since the TextArea styling is following the same styling as TextField, I will leave this as is!

__docs__/wonder-blocks-form/text-area.stories.tsx
beaesguerra: I left this one to be tested by Chromatic to verify the resizing control types in case styling impacts it later on!

packages/wonder-blocks-form/src/components/text-area.tsx
beaesguerra: Great thought! I was thinking this too because it'd be great if we could utilize css specificity! I had tried to do this initially, but got aphrodite typescript errors so I assumed it wasn't supported!

Looking into it again, I see in Wonder Blocks, we explicitly define the types for aphrodite and we don't define the `:disabled` pseudo-class ([WB aphrodite types](https://github.com/Khan/wonder-blocks/blob/68dd6059c24c595c62e07b72b9ec17f0ead2bfbb/types/aphrodite.d.ts#L25-L38)). However, the [types in aphrodite](https://github.com/Khan/aphrodite/blob/225f43c5802259a9e042b384a1f4f2e5b48094ea/typings/css-properties.d.ts#L1822) includes it.

I'm thinking we can create a separate task to update the aphrodite type definitions, or maybe even use the types from aphrodite directly (let me know if there's a reason we define the types in WB!). Then we can also update the styling to use the `:disabled` pseudo-class. Other components could be updated to use it too! What do you think?

packages/wonder-blocks-form/src/components/text-area.tsx
beaesguerra: Great question! I was wondering this as well. Since the TextArea component is similar to the TextField component, I followed the behaviour of the TextField component, which was to use the `disabled` attribute directly. As a general guideline, do we want all interactive elements in the design system to utilize `aria-disabled` instead of `disabled` so that they remain focusable?

jandrade: That's right, I think we should try to use that approach as much as we can.

packages/wonder-blocks-form/src/components/text-area.tsx
jandrade: You could directly add `:disabled` in the WB aphrodite types here, but it's totally fine if you prefer to do it in a separate task. Agreed that a lot of components could be benefited from this :), so probably a separate ticket is the best approach.

packages/wonder-blocks-form/src/components/text-area.tsx
beaesguerra: Sounds good! I created a follow up task to make the TextArea focusable when it is `disabled` (WB-1734). I'll work on this after our deprecated component migration work this focus week!

packages/wonder-blocks-form/src/components/text-area.tsx
beaesguerra: Will work on this in the follow up ticket to make the TextArea focusable when disabled (WB-1734). We can use a CSS attribute selector then for aria-disabled since we would no longer use the `disabled` attribute!

jandrade (approved):

PR #2276, opened 2024-07-17
[wb1733.1.adddocsfixdocslinks] Add docs for new WB Testing and fix links

somewhatabstract:
## Summary:
This adds docs for the new `boundary` adapter and the new `renderHookStatic` call to the Wonder Blocks Testing package. It also updates the overview to explain some things about Wonder Blocks Testing Core.

In addition, I discovered that most of the intra-story links in MDX files were not working. After some investigation and discussion with @jandrade, it was determined that the markdown link format just doesn't currently work properly and that converting to `<a>` links with slightly changed href works as needed.

I used some regexp to convert from the markdown version to the markup version.

### 1. Markdown to markup
Find: `\[((?:[^\[\]])+)\]\((/docs/[^\)]+)\)`
Replace: `<a href="./?path=$2">$1</a>`

### 2. Add the missing leading `packages-` to the links
Seems that we reorganized the docs so that there's now a new top-level grouping that the URL needs to include when using the anchor approach.
Find: `<a href="./\?path=/docs/(?!packages)`
Replace: `<a href="./?path=/docs/packages-`

Issue: WB-1733

## Test plan:
`yarn start:storybook` and check the new docs and changed links are working


__docs__/wonder-blocks-testing/_overview_utlities.mdx
somewhatabstract, line 1: **note:** New section for WBT exports that aren't part of the existing fixtures, mocking, or test harness frameworks.

__docs__/wonder-blocks-testing/_overview_utlities.mdx
```diff
@@ -0,0 +1,14 @@
+import {Meta} from "@storybook/blocks";
+
+<Meta
+    title="Packages / Testing / Utilities / Overview"
+    parameters={{
+        chromatic: {
+            disableSnapshot: true,
+        },
+    }}
```
jandrade, line 9: **nit:** This is not needed anymore. SB changed the way MDX pages work as part of v7 or v8 (don't remember which exact version). Now MDX pages are treated as docs-only pages, so Chromatic never take them as reference for snapshots.

__docs__/wonder-blocks-testing/exports.render-hook-static.mdx
```diff
@@ -0,0 +1,25 @@
+import {Meta} from "@storybook/blocks";
+
+<Meta
+    title="Packages / Testing / Utilities / Exports / renderHookStatic()"
+    parameters={{
+        chromatic: {
+            disableSnapshot: true,
+        },
+    }}
```
jandrade, line 9: Same comment here (no change required as it works fine right now, but I leave it to you in case you can get rid of this param).

__docs__/wonder-blocks-timing/with-action-scheduler.mdx
```diff
@@ -37,7 +37,7 @@ The added `schedule` prop is of type [`IScheduleActions`](#ischeduleactions). Th
 The returned value from `withActionScheduler` is a React component with props of
 type `TProps`.
 
-Access to the [timing API](/docs/timing-types-ischeduleactions--page) is provided via the `withActionScheduler` higher order
+Access to the <a href="./?path=/docs/packages-timing-types-ischeduleactions--docs">timing API</a> is provided via the `withActionScheduler` higher order
```
jandrade, line 40: **suggestion:** It looks like this link is broken. It's missing `withactionscheduler`

```suggestion
Access to the <a href="./?path=/docs/packages-timing-withactionscheduler-types-ischeduleactions--docs">timing API</a> is provided via the `withActionScheduler` higher order
```

jandrade (approved):
Looks and works great! thanks for fixing all the broken links 👏 🚢

__docs__/wonder-blocks-timing/with-action-scheduler.mdx
somewhatabstract: Ah, yeah, that must've broken with a rename at some point.

__docs__/wonder-blocks-testing/_overview_utlities.mdx
somewhatabstract: I've removed this from all the MDX files.

__docs__/wonder-blocks-testing/exports.render-hook-static.mdx
somewhatabstract: I've removed this from all the MDX files.

__docs__/wonder-blocks-timing/with-action-scheduler.mdx
somewhatabstract: Ah, the issue is that the exported type shouldn't be nested under `withActionScheduler` like that. The `Types` folder should be at the root level of the package docs like with other packages. I'll fix that.

PR #2277, opened 2024-07-17
Add in loadTranslations/clearTranslations to i18n, use stored messages if they exist.

jeresig:
## Summary:
This is adding in functionality that's similar to LinguiJS, to the wonder-blocks-i18n package, so that we can have a store of translated strings and use them for our translations instead of assuming that the strings will already be translated for us.

Issue: FEI-5682

## Test plan:
Tests should pass! I'll be wiring this up in Webapp in a separate PR, once this lands.

### DIFF
--- a/.changeset/dull-stingrays-mate.md
+++ b/.changeset/dull-stingrays-mate.md
@@ -1,6 +1,5 @@
 ---
 "@khanacademy/wonder-blocks-i18n": minor
-"@khanacademy/wonder-blocks-form": patch
 ---
 
 Add loadTranslations/clearTranslations methods to wonder-blocks-i18n.
--- a/.changeset/dull-stingrays-mate.md
+++ b/.changeset/dull-stingrays-mate.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/wonder-blocks-i18n": minor
+"@khanacademy/wonder-blocks-form": patch
+---
+
+Add loadTranslations/clearTranslations methods to wonder-blocks-i18n.
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -13,5 +13,6 @@
     "files.associations": {
         "*.flow": "plaintext",
     },
-    "flow.enabled": false
+    "flow.enabled": false,
+    "jest.jestCommandLine": "yarn jest"
 }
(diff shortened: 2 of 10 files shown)


packages/wonder-blocks-form/src/components/__tests__/field-heading.test.tsx
```diff
@@ -187,7 +187,11 @@ describe("FieldHeading", () => {
         render(
             <FieldHeading
                 field={<TextField id="tf-1" value="" onChange={() => {}} />}
-                label={<I18nInlineMarkup>Hello, world!</I18nInlineMarkup>}
+                label={
+                    <I18nInlineMarkup b={(s: string) => <b>{s}</b>}>
```
jeresig, line 191: Caught this and the next file during testing - they weren't using I18nInlineMarkup correctly, but that fixes them so that they are now rendered with markup.

.vscode/settings.json
```diff
@@ -13,5 +13,6 @@
     "files.associations": {
         "*.flow": "plaintext",
     },
-    "flow.enabled": false
-}
\ No newline at end of file
+    "flow.enabled": false,
+    "jest.jestCommandLine": "yarn jest"
```
jeresig, line 17: I caught this and the next file during my local test - it was running `yarn test` which caused a bunch of unrelated stuff to run (and the results failed to appear in VS Code). So this allows the tests to now run correctly!

.changeset/dull-stingrays-mate.md
```diff
@@ -0,0 +1,6 @@
+---
+"@khanacademy/wonder-blocks-i18n": minor
+"@khanacademy/wonder-blocks-form": patch
```
jandrade, line 3: **nit(no change required):** You probably don't need to bump `form` (dev-only changes), but I also get if you do it in this case.

jandrade (approved):
Looks great! thanks for cleaning up and refactoring it to make it closer to Lingui :shipit: .

.changeset/dull-stingrays-mate.md
jeresig: Good call - removing!

PR #2281, opened 2024-07-22
Fixing tooltip bug in firefox where it sometimes flickers and stays hidden

beaesguerra:
## Summary:
- In Firefox, we were observing that the tooltip would sometimes flicker and stay hidden even though the tooltip anchor is still hovered on. Other browsers didn't seem to have this issue. It could be reproduced in Firefox by placing the tooltip anchor at the top left with long content and hovering over it. 
- When the `mouseenter` and `mouseleave` events were logged, I found that when the issue occurs, there would be subsequent `mouseenter`, `mouseleave`, and `mouseenter` events. 

https://github.com/user-attachments/assets/911f57d3-90b4-4dbc-9d66-8c62928b46c2

What's happening here:
  - The first `mouseenter` opens the tooltip
  - When the `mouseleave` event occurs, it [triggers the tooltip to close after a delay](https://github.com/Khan/wonder-blocks/blob/68dd6059c24c595c62e07b72b9ec17f0ead2bfbb/packages/wonder-blocks-tooltip/src/components/tooltip-anchor.tsx#L258-L261)
  - When the second `mouseenter` event occurs, it updates the active state to true immediately (because [`instant` resolves to true](https://github.com/Khan/wonder-blocks/blob/68dd6059c24c595c62e07b72b9ec17f0ead2bfbb/packages/wonder-blocks-tooltip/src/components/tooltip-anchor.tsx#L237) due to `active && TRACKER.steal(this)` resolving to true)
  - Then, the delayed action of closing the tooltip occurs (triggered by `mouseleave`). Since the `mouseenter` event already finished opening the tooltip, the tooltip stays hidden

The fix
- When the active state is going to be updated, we check first to see if the current active state is already the same as what it needs to be updated to. If it's the same, we cancel any pending actions since the state is already up to date. 
- This fixes it because when the subsequent events occur, the second `mouseenter` event will recognize that the tooltip is already opened and cancel the action to close the tooltip (which was triggered from the prior `mouseleave` event) 

You may also be wondering why multiple events are being triggered in the first place. Here's what I found:
- When we hover over the tooltip anchor, the tooltip popper is activated. It places the tooltip based on the anchor, and then uses a transform to shift it over. This [styling comes from react-popper](https://github.com/Khan/wonder-blocks/blob/68dd6059c24c595c62e07b72b9ec17f0ead2bfbb/packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx#L148-L159)
image

- I think what is happening is when the tooltip opens, it sometimes opens where the cursor is, which causes the `mouseleave` event on the tooltip anchor. Then when it is moved using a CSS transform, the `mouseenter` event is triggered again on the tooltip. 
- I was exploring some of the react-popper options to see if we could prevent this from happening in the first place, but couldn't find anything that worked. Just for my own curiosity, I disabled the css transform and hardcoded the positioning and wasn't able reproduce the multiple events on hover. This isn't ideal though since react-popper handles all the different positioning cases. Also, I think it makes sense for the TooltipAnchor component to be able to handle multiple events in a row in case it ever happens!

As always, please let me know if you have any other ideas or suggestions, this was a tricky one!!

Issue: WB-1731

## Test plan:
- Firefox: When tooltip is in the top left corner, the tooltip should consistently open when the anchor is hovered (`/iframe.html?args=&id=packages-tooltip-tooltip--in-top-corner&viewMode=story`)
- Tooltip continues to work as expected

## Screenshots/Recordings
After:
Tooltip behaves as expected even though multiple mouse events occurred on hover

https://github.com/user-attachments/assets/17d5a0c5-8e4f-4211-acd1-13b084da6887

Logs of events:
**Before:**
before

**After:**
after


jandrade (approved):
This is great! thanks for that detailed description and adding all the comments in the code 👏 🚢 

I think this approach makes total sense given some known limitations we have with Popper v2, but hopefully all of this will be simplified as soon as we migrate to floating ui.

PR #2283, opened 2024-07-26
fix(LEMS-2225): add conditional aria-describedby logic

anakaren-rojas:
## Summary:
Adds conditional logic to allow aria-describedby to be:
- defined by the user (overrides default)
- available in addition to `aria-label` or `aria-labelledby`

Issue: LEMS-2225

## Test plan:
1. `yarn start` in branch or [chromatic link](https://5e1bf4b385e3fb0020b7073c-lzlzfekdzl.chromatic.com/?path=/docs/overview--docs)
2. Navigate to Popover stories
3. Choose `With custom aria described by` story 
4. Story includes hidden element that should be read out by the screen reader 

## Screenshots

https://github.com/user-attachments/assets/a46ca21a-71ac-442a-891b-98dc675701fb

### DIFF
--- a/.changeset/plenty-rockets-breathe.md
+++ b/.changeset/plenty-rockets-breathe.md
@@ -2,4 +2,4 @@
 "@khanacademy/wonder-blocks-popover": patch
 ---
 
-Testing adding aria-describedby
+Adding logic for aria-describedby
--- a/__docs__/wonder-blocks-popover/popover.stories.tsx
+++ b/__docs__/wonder-blocks-popover/popover.stories.tsx
@@ -756,7 +756,11 @@ export const WithCustomAriaLabel: StoryComponentType = {
 /**
  * With custom aria-describedby - overrides the default aria-describedby
  */
-export const WithCustomAriaDescribedBy = ({placement}: {placement: Placement}) => {
+export const WithCustomAriaDescribedBy = ({
+    placement,
+}: {
+    placement: Placement;
+}) => {
     const [opened, setOpened] = React.useState(false);
 
     return (
--- a/__docs__/wonder-blocks-popover/popover.stories.tsx
+++ b/__docs__/wonder-blocks-popover/popover.stories.tsx
@@ -6,10 +6,10 @@ import Button from "@khanacademy/wonder-blocks-button";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
 import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
-import {LabelLarge} from "@khanacademy/wonder-blocks-typography";
+import {HeadingMedium, LabelLarge} from "@khanacademy/wonder-blocks-typography";
 import type {Placement} from "@khanacademy/wonder-blocks-tooltip";
 
-import {Popover, PopoverContent} from "@khanacademy/wonder-blocks-popover";
+import {Popover, PopoverContent, PopoverContentCore} from "@khanacademy/wonder-blocks-popover";
 import packageConfig from "../../packages/wonder-blocks-popover/package.json";
 
 import ComponentInfo from "../../.storybook/components/component-info";
@@ -86,6 +86,16 @@ const styles = StyleSheet.create({
         padding: spacing.large_24,
         flexDirection: "row",
         gap: spacing.medium_16,
+    },
+    srOnly: {
+        border: 0,
+        clip: "rect(0,0,0,0)",
+        height: 1,
+        margin: -1,
+        overflow: "hidden",
+        padding: 0,
+        position: "absolute",
+        width: 1,
     },
 });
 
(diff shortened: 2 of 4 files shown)


packages/wonder-blocks-popover/src/components/popover.tsx
```diff
@@ -278,10 +278,20 @@ export default class Popover extends React.Component<Props, State> {
...
 
-        const ariaDescribedBy = ariaLabel ? undefined : `${uniqueId}-content`;
+        /**
+         * If aria-describedby is provided, use that id as a reference to the content
+         * If only aria-label is provided, don't use aria-describedby
+         * If both are provided, use aria-label as the label (automatically done by SR)
+         * If neither are provided, use the uniqueId to reference the content
+         *  > Don't want to have aria-describedby set by default in case there isn't something
+         *  > for the SR to reference.
+         */
```
anakaren-rojas, line 292: tentative comment on approach

.changeset/plenty-rockets-breathe.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-popover": patch
+---
+
+Testing adding aria-describedby
```
catandthemachines, line 5: Update this to state you're adding the aria-describedby, and not testing anymore 😉

__docs__/wonder-blocks-popover/popover.stories.tsx
```diff
@@ -87,6 +87,16 @@ const styles = StyleSheet.create({
         flexDirection: "row",
         gap: spacing.medium_16,
     },
+    srOnly: {
```
catandthemachines, line 90: nit: Does WonderBlocks have a utility constant for this? It would be great if it was in a shared file for other WonderBlocks components to use.

packages/wonder-blocks-popover/src/components/__tests__/popover.test.tsx
```diff
@@ -602,19 +606,59 @@ describe("Popover", () => {
...
+            );
 
             // Assert
+            expect(
+                screen.getByLabelText("Popover Aria Label"),
+            ).toBeInTheDocument();
+
+            expect(
+                await screen.findByRole("dialog", {
+                    description: "This is a popover description",
```
catandthemachines, line 622: Shouldn't we be checking the title of the dialog is "This is a popover title"?

catandthemachines (approved):
Just need to clean up one test file and a few comments and I think we're good to go! 🥳

__docs__/wonder-blocks-popover/popover.stories.tsx
anakaren-rojas: It doesn't seem like it; I'd imagine since wonderblocks hosts components, they would all need to be visible

packages/wonder-blocks-popover/src/components/__tests__/popover.test.tsx
anakaren-rojas: I updated this test to be more consistent with the other tests, hopefully the intent is more clear now; since I'm using an aria label in this scenario, I would expect the dialog title to be `Popover Aria Label`

packages/wonder-blocks-popover/src/components/__tests__/popover.test.tsx
```diff
@@ -602,19 +606,67 @@ describe("Popover", () => {
...
+                    name: "Popover Aria Label",
+                }),
+            ).toBeInTheDocument();
 
-            expect(popover).toHaveAttribute("aria-label", "Popover Aria Label");
-            expect(popover).not.toHaveAttribute("aria-labelledby");
-            expect(popover).not.toHaveAttribute("aria-describedby");
+            expect(
+                await screen.findByRole("dialog", {
+                    description: "This is a popover description",
```
jandrade, line 624: **suggestion:** With this refactor in place, I think it makes sense to create a separate test to assert the `description` part of this test.

For some context, we try to follow the AAA approach for testing: https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/98402353/JavaScript+Testing+Best+Practices#Assert-One-Requirement-Per-Test
https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/98402353/JavaScript+Testing+Best+Practices#Assert-One-Requirement-Per-Test

One of the main benefits we have found by using this approach is that it's easier to isolate a failure when it happens and also easier to apply the fix when needed.

jandrade (approved):
Looks great to me! Just left a minor suggestion on refactoring one of the tests, but nothing blocking. Thanks for working on this improvement :shipit:

packages/wonder-blocks-popover/src/components/__tests__/popover.test.tsx
anakaren-rojas: Thanks Juan, I split it out into three tests that each have 1 assertion

PR #2284, opened 2024-07-31
Allow TextArea to be focusable when disabled

beaesguerra:
## Summary:
For better accessibility, interactive components in WB can be focusable, even when disabled. 
These changes to TextArea include:
- Setting aria-disabled instead of the disabled attribute
- Disabling event handlers if the disabled prop is true. onFocus and onBlur continues to work if disabled prop is true
- Adding disabled focus styling
- Setting the readonly attribute if the disabled prop is true to prevent interactions with textarea

Issue: WB-1734

## Test plan:
- Review Disabled Focus TextArea Styling (`?path=/story/packages-form-textarea-all-variants--focus`)
- Review Disabled TextArea docs (`?path=/docs/packages-form-textarea--docs`)
- Review TextArea a11y docs (`?path=/docs/packages-form-textarea-accessibility--docs`)
- TextArea Disabled state (`?path=/story/packages-form-textarea--disabled`)
  - Confirm the user can't type in the TextArea if it is disabled
  - Confirm the user can still focus on the TextArea if it is disabled
  - Confirm screen readers continue to communicate that the field is disabled

## Screen recording
Voiceover + keyboard navigation

https://github.com/user-attachments/assets/b89116a2-8d79-4d5d-9697-f97c8a0623c1


packages/wonder-blocks-form/src/components/__tests__/text-area.test.tsx
jandrade, line 1: **praise:** Bravo 👏 Adding tests to all these cases makes ton of sense.

packages/wonder-blocks-form/src/components/text-area.tsx
```diff
@@ -273,10 +273,10 @@ const TextArea = React.forwardRef<HTMLTextAreaElement, TextAreaProps>(
                     className={className}
                     style={[...getStyles(), style]}
                     value={value}
-                    onChange={handleChange}
+                    onChange={disabled ? undefined : handleChange}
```
jandrade, line 276: **question:** What would happen if the `TextArea` instance is a controlled one? should we still trigger `onChange` in case the parent component updates its `value` prop?

jandrade (commented):
Looks great! love the tests, but just have one minor question about the `onChange` behavior on controlled components that I'd like to get your thoughts on before giving final approval.

packages/wonder-blocks-form/src/components/text-area.tsx
beaesguerra: Good question! If I'm understanding the question correctly, the parent component will still be able to update the `value` prop and have it update what is shown in the textarea:

https://github.com/user-attachments/assets/7dd79f1d-db51-4ed6-b05d-e97f7cfe2e03

So if something other than the TextArea was controlling the value, it will still update (just the TextArea component won't call the onChange callback)

Kinda related - when I set `onChange` to undefined when it is disabled though, I also set the `readOnly` prop to `true` because of this warning (see below). `readOnly` also prevents the textarea cursor from showing when it is disabled :)

Screenshot 2024-08-02 at 9 41 22 AM

Let me know if you have thoughts around this or if I'm missing something!

jandrade: Thanks for the context! My reasoning about enabling `onChange` in those cases is that when the component is in controlled mode, then that means that both `value` and `onChange` are configured ([docs](https://khan.github.io/wonder-blocks/?path=/docs/packages-form-textarea--docs#controlled)). I totally understand your point here, but I feel that allowing value/onChange to work in expected ways would be better as folks will know that the API will remain consistent.

Maybe you could only modify that behavior for when it is disabled and it is controlled?

beaesguerra: Gotcha! Turns out if I change this back, the `onChange` handler doesn't get called anyways since the TextArea is in readonly mode also if it is disabled (so we didn't need this extra logic 😄 ). This simplifies things and is more consistent with expectations as you mentioned! Thank you for flagging!

beaesguerra: Related to controlled components, the prop types for `value` and `onChange` are not optional, does this mean they should always be used in a controlled way?

jandrade: hmmm yeah, and I think that makes sense! parent components should be the ones managing state for these fields.

jandrade (approved):

PR #2295, opened 2024-08-21
WB-1741: Button is missing focus styles when secondary and disabled

jandrade:
## Summary:

Fixes a UI issue where secondary buttons weren't displaying the focus ring when
disabled + focused.

Issue: WB-1741

## Test plan:

Navigate to the Button variants story in Storybook and verify that the focus
ring is displayed when a secondary button is disabled and focused.

/?path=/story/packages-button--variants

Screenshot 2024-08-21 at 3 03 07 PM


beaesguerra (approved):
Nice! Looks good 🚀

PR #2297, opened 2024-08-22
Fix icon color on secondary and tertiary buttons (hover)

jandrade:
## Summary:

While working on some other changes, I noticed that the icon color on secondary
and tertiary buttons was not being set correctly. This PR fixes that issue by
setting the icon color to the same color as the text.

Issue: XXX-XXXX

## Test plan:

Navigate to /?path=/story/packages-button--icon
Hover over the secondary and tertiary buttons
Verify that the icon color on secondary and tertiary buttons is the same as the
text color

### BEFORE

https://github.com/user-attachments/assets/1902a543-463c-443d-824b-6b6d1263fc0d

### AFTER

https://github.com/user-attachments/assets/5b9929a9-3082-4d2b-b439-1a4b2e0222c7

### DIFF
--- a/.changeset/small-walls-destroy.md
+++ b/.changeset/small-walls-destroy.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-button": patch
+---
+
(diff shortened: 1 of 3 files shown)


beaesguerra (approved):
Nice fix! 😄

PR #2301, opened 2024-08-22
Update TextField focus styling to be consistent with other WB components

beaesguerra:
## Summary:
- Updates the focus styling of the TextField so it is consistent with components like SingleSelect, MultiSelect, TextArea.
- This also aligns with the [Figma TextField component](https://www.figma.com/design/VbVu3h2BpBhH80niq101MHHE/%F0%9F%92%A0-Main-Components?node-id=13497-10024&t=LjLznJ6mpKxWbPsI-4)

Issue: WB-1746

## Test plan:
1. Confirm the focus styling of the TextField component when:
- it is focused
- it is focused on a dark background with the `light` prop set to `true`
- it is in an error state and is focused
- it is in an error state on a dark background with the `light` prop set to `true`

### DIFF
--- a/.changeset/itchy-toys-hear.md
+++ b/.changeset/itchy-toys-hear.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-form": patch
+---
+
+Update focus styling for TextField so that it is thicker. This makes it consistent with the focus styling for other components like SingleSelect, MultiSelect, TextArea.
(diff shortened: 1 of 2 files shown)


beaesguerra:
I'm going to attempt to refactor this instead so we can use pseudo classes instead (and so we can use storybook's pseudo states for visual regression tests!)

PR #2302, opened 2024-08-23
Update TextField styling for consistency with other components (focus styling)

beaesguerra:
## Summary:
- Update `TextField` state styling so it is consistent with other components like `TextArea`, `SingleSelect`, `MultiSelect` (especially the focus styling)
- The styling changes use CSS pseudo-classes now so that we can have visual regression tests for the UI states using Storybook's pseudo-states add on.
- These styles align with the [Figma TextField component](https://www.figma.com/design/VbVu3h2BpBhH80niq101MHHE/%F0%9F%92%A0-Main-Components?node-id=13497-10015&t=QZlu8ecGmGU3QceA-4)
- Removed unneeded focused state since we use pseudo classes now
- Added variant stories for Chromatic tests

Note: These styles are based on how TextArea was styled!

Issue: WB-1746

## Test plan:
1. Confirm the focus styling for TextField when (`?path=/docs/packages-form-textfield--docs`):
- it is focused
- it is focused on a dark background with the `light` prop set to `true`
- it is in an error state and is focused
- it is in an error state on a dark background with the `light` prop set to `true`
2. Confirm other states (error, disabled, default)
3. Review the new variant stories (`?path=/docs/packages-form-textfield-all-variants--docs`)

### DIFF
--- a/packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.tsx
+++ b/packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.tsx
@@ -3,7 +3,6 @@ import {render, screen, fireEvent} from "@testing-library/react";
 import {userEvent} from "@testing-library/user-event";
 
 import {StyleSheet} from "aphrodite";
-import {color} from "@khanacademy/wonder-blocks-tokens";
 import LabeledTextField from "../labeled-text-field";
 
 describe("LabeledTextField", () => {
--- a/packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.tsx
+++ b/packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.tsx
@@ -382,28 +382,6 @@ describe("LabeledTextField", () => {
         expect(input).toBeInTheDocument();
     });
 
-    it("light prop is passed to textfield", async () => {
-        // Arrange
-
-        // Act
-        render(
-            <LabeledTextField
-                label="Label"
-                value=""
-                onChange={() => {}}
-                light={true}
-            />,
-        );
-
-        const textField = await screen.findByRole("textbox");
-        textField.focus();
-
-        // Assert
-        expect(textField).toHaveStyle({
-            boxShadow: `0px 0px 0px 1px ${color.blue}, 0px 0px 0px 2px ${color.white}`,
-        });
-    });
-
     it("style prop is passed to fieldheading", async () => {
         // Arrange
         const styles = StyleSheet.create({
--- a/packages/wonder-blocks-form/src/components/text-field.tsx
+++ b/packages/wonder-blocks-form/src/components/text-field.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import {Falsy, StyleSheet} from "aphrodite";
+import {StyleSheet} from "aphrodite";
 
 import {IDProvider, addStyle} from "@khanacademy/wonder-blocks-core";
 import {border, color, mix, spacing} from "@khanacademy/wonder-blocks-tokens";
@@ -231,7 +231,7 @@ class TextField extends React.Component<PropsWithForwardRef, State> {
         }
     };
 
-    getStyles = (): (React.CSSProperties | Falsy)[] => {
+    getStyles = (): StyleType => {
         const {disabled, light} = this.props;
         const {error} = this.state;
         // Base styles are the styles that apply regardless of light mode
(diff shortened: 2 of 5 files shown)


packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.tsx
```diff
@@ -382,28 +381,6 @@ describe("LabeledTextField", () => {
         expect(input).toBeInTheDocument();
     });
 
-    it("light prop is passed to textfield", async () => {
```
beaesguerra, line 385: We can remove the styling related unit tests since we'll have Chromatic tests to verify the styles!

packages/wonder-blocks-form/src/components/text-field.tsx
```diff
@@ -335,35 +327,81 @@ const styles = StyleSheet.create({
...
     disabled: {
         background: color.offWhite,
         border: `1px solid ${color.offBlack16}`,
         color: color.offBlack64,
         "::placeholder": {
-            color: color.offBlack32,
+            color: color.offBlack64,
+        },
+        cursor: "not-allowed",
+        ":focus-visible": {
```
beaesguerra, line 357: Note: The `TextField` isn't focusable yet when disabled! I created a separate ticket for that work: WB-1751

packages/wonder-blocks-form/src/components/text-field.tsx
```diff
@@ -222,22 +217,38 @@ class TextField extends React.Component<PropsWithForwardRef, State> {
...
-            if (onBlur) {
-                onBlur(event);
-            }
-        });
+        if (onBlur) {
+            onBlur(event);
+        }
+    };
+
+    getStyles = (): (React.CSSProperties | Falsy)[] => {
```
jandrade, line 234: **nit:** You could also use `StyleType` from `wonder-blocks-core`.

```suggestion
    getStyles = (): StyleType => {
```

packages/wonder-blocks-form/src/components/text-field.tsx
```diff
@@ -222,22 +217,38 @@ class TextField extends React.Component<PropsWithForwardRef, State> {
...
+            !disabled && styles.defaultFocus,
+            disabled && styles.disabled,
+            !!error && styles.error,
+        ];
+        const lightStyles = [
+            styles.light,
+            !disabled && styles.lightFocus,
+            disabled && styles.lightDisabled,
+            !!error && styles.lightError,
+        ];
```
jandrade, line 250: **praise:** Nice cleanup!

packages/wonder-blocks-form/src/components/text-field.tsx
```diff
@@ -335,35 +327,81 @@ const styles = StyleSheet.create({
...
         },
     },
-    defaultLight: {
-        boxShadow: `0px 0px 0px 1px ${color.blue}, 0px 0px 0px 2px ${color.white}`,
+    lightFocus: {
+        ":focus-visible": {
+            outline: `1px solid ${color.blue}`,
+            outlineOffset: 0, // Explicitly set outline offset to 0 because Safari sets a default offset
+            borderColor: color.blue,
+            boxShadow: `0px 0px 0px 2px ${color.blue}, 0px 0px 0px 3px ${color.white}`,
```
jandrade, line 375: **question:** Probably out of scope for this PR... do you think this could use `outline` instead of `boxShadow`? I don't remember the exact reason why we took this approach.

jandrade (approved):
This looks awesome! thanks for the improvements 🎉

packages/wonder-blocks-form/src/components/text-field.tsx
beaesguerra: This is what I needed! Thank you for pointing this out 😄

packages/wonder-blocks-form/src/components/text-field.tsx
beaesguerra: I think we could! I based the styling on what I implemented in TextArea, and that was based on styling I found in other components (I remember it being tricky - it's using outline, border, and box shadow!). I think we could rework the styles to simplify things. I created WB-1753 for this! 

It would also be nice if we could share common state styles between components so it's consistent (kind of like what we do with Typography styles). What do you think?

packages/wonder-blocks-form/src/components/text-field.tsx
beaesguerra: I also updated this in TextArea since that's where I copied it from!

PR #2307, opened 2024-08-27
[encoderpatch] Patch TextEncoder/TextDecoder for tests

somewhatabstract:
## Summary:
React 18 requires these things but JSDOM doesn't include them, so we shim them

Issue: XXX-XXXX

## Test plan:
`yarn test` before React 18 update and after.


jeresig (approved):
Great, thanks Jeff!

PR #2308, opened 2024-08-28
WB-1677.2: Combobox misc fixes

jandrade:
## Summary:

- Changed  the border color from `offBlack16` to `offBlack50` to be consistent with the rest of form fields.
- Also applied a few Storybook fixes:
  - Refactored arg types to use categories
  - Fixed interactive story to support switching `selectionType` values without breaking the story.

Issue: WB-1677

## Test plan:

Navigate to the Combobox docs in Storybook and verify the following:

1. The border color is now `offBlack50`.

2. The arg types are now categorized.

3. The interactive story should now support switching `selectionType` values
without breaking the story. This means that the story should try to preserve the
`value` when switching between `single` and `multi` selection types.

https://github.com/user-attachments/assets/0d8481d8-5fcf-4c0e-b665-0798c22f3e0c

### DIFF
--- a/__docs__/wonder-blocks-dropdown/combobox.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/combobox.stories.tsx
@@ -1,4 +1,5 @@
 import {action} from "@storybook/addon-actions";
+import {useArgs} from "@storybook/preview-api";
 import {Meta, StoryObj} from "@storybook/react";
 import {StyleSheet} from "aphrodite";
 import * as React from "react";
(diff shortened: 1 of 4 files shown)


__docs__/wonder-blocks-dropdown/combobox.stories.tsx
```diff
@@ -56,12 +59,17 @@ const defaultArgs = {
     disabled: false,
     placeholder: "Select an item",
     testId: "",
+    autoComplete: "none",
+    light: false,
+    loading: false,
+    // labels: defaultComboboxLabels,
```
beaesguerra, line 65: nit: Can this be removed?

beaesguerra (approved):
Changes look good!

PR #2309, opened 2024-08-28
Revert TextField state styling

beaesguerra:
## Summary:
The new focus styling in the TextField can sometimes be cropped off in webapp if an ancestor has overflow: hidden. This 
reverts the changes (commit 982f680865ad1ffc5fb6a0743fdf68c3853a65dc) to unblock updating other WB dependencies in webapp (https://github.com/Khan/webapp/pull/25010). A proper fix will be made in a follow up PR!

Ran: `git revert 982f680865ad1ffc5fb6a0743fdf68c3853a65dc`

Issue: WB-1746

## Test plan:
1. Confirm TextField styling is the same as before (ie. no additional focus outline) (`?path=/story/packages-form-textfield--default`)
image


beaesguerra:
> 🎉 Good news!! We've packaged up the latest commit from this PR ([beac244](https://github.com/Khan/wonder-blocks/commit/beac2440b05f15885726f6e7f05a59213ed6f561)) and published all packages with changesets to npm.
> 
> You can install the packages in webapp by running:
> 
> ```shell
> ./services/static/dev/tools/deploy_wonder_blocks.js --tag="PR2309"
> ```

I tested the WB PR snapshot with `webapp` and it reverts the TextField styling correctly. I will also confirm in @somewhatabstract's [webapp wbupdate PR](https://github.com/Khan/webapp/pull/25010) znd  once these changes are merged and released!

| Before with `@khanacademy/wonder-blocks-form@4.9.1` | After with `@khanacademy/wonder-blocks-form@0.0.0-PR2309-20240828170905` |
| --- | --- | 
| Screenshot 2024-08-28 at 12 12 55 PM | Screenshot 2024-08-28 at 12 11 09 PM|
| Screenshot 2024-08-28 at 12 13 24 PM | Screenshot 2024-08-28 at 12 10 38 PM  |

packages/wonder-blocks-form/src/components/text-field.tsx
```diff
@@ -327,81 +335,35 @@ const styles = StyleSheet.create({
...
         border: `1px solid ${color.offBlack16}`,
         color: color.offBlack64,
         "::placeholder": {
-            color: color.offBlack64,
-        },
-        cursor: "not-allowed",
-        ":focus-visible": {
-            outline: "none",
-            boxShadow: `0 0 0 1px ${color.white}, 0 0 0 3px ${color.offBlack32}`,
+            color: color.offBlack32,
```
jandrade, line 351: **question:** Is this going to be modified in the future PR? See related comment here: https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=66cf79973b48a29341523038

beaesguerra: Yes that's right! I'll put all these changes back in the future PR and then fix the outline styles so it can work in all cases. These changes will just revert the TextField styles to what they were before!

jandrade: got it, sounds good, thanks for the clarification :)

jandrade (approved):
The revert looks good. It's unfortunate that we have to do this, as using `:focus-visible` + outline is the approach that makes more sense, but I totally understand that it's better to be cautious on this change (better safe than sorry).

beaesguerra:
> It's unfortunate that we have to do this, as using :focus-visible + outline is the approach that makes more sense, but I totally understand that it's better to be cautious on this change (better safe than sorry)

@jandrade I'll look into a solution so that maybe we can still do this! I was tinkering around with a negative outline offset and testing the different states. Stay tuned! 😄

PR #2310, opened 2024-08-29
WB-1754: Refactor Combobox to consume handleChange from useListbox instead of relying on useEffect

jandrade:
## Summary:

Currently, the Combobox component uses useEffect to determine when the selected
value changes.

This is causing some sync issues between the hook (`useListbox`) and its caller
(`Combobox`), as we rely on side effects to verify when the value changes.

For example, there was a case where the combobox was not closing when the same
option was selected twice in a row.

This PR takes a different approach to let the consumer component know when an
option is selected via a handler exposed from the hook.

Also took the opportunity to reorganize/cleanup the code in the Combobox
component.

Issue: https://khanacademy.atlassian.net/browse/WB-1754

## Test plan:

Navigate to the Combobox story in Storybook and verify that the component
behaves as expected.

Also open the combobox, select an option, verify that the selected value is
displayed in the input, and that the listbox is closed.

Open it again, select the same option, and verify that the listbox is closed
without changing the input value.

| BEFORE | AFTER |
|--------|--------|
| https://github.com/user-attachments/assets/ca6d6f2b-f7ef-4379-828f-5d51accb178c | https://github.com/user-attachments/assets/0eb10944-2d18-4a29-8d7e-304df8c7b49a |

### DIFF
--- a/__docs__/wonder-blocks-dropdown/combobox.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/combobox.stories.tsx
@@ -1,13 +1,14 @@
 import {action} from "@storybook/addon-actions";
 import {useArgs} from "@storybook/preview-api";
 import {Meta, StoryObj} from "@storybook/react";
+import {expect, userEvent, within} from "@storybook/test";
 import {StyleSheet} from "aphrodite";
 import * as React from "react";
-import {expect, userEvent, within} from "@storybook/test";
-import {PropsFor, View} from "@khanacademy/wonder-blocks-core";
-import {Combobox, OptionItem} from "@khanacademy/wonder-blocks-dropdown";
 import {LabelLarge} from "@khanacademy/wonder-blocks-typography";
 import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {Checkbox} from "@khanacademy/wonder-blocks-form";
+import {Combobox, OptionItem} from "@khanacademy/wonder-blocks-dropdown";
+import {PropsFor, View} from "@khanacademy/wonder-blocks-core";
 import {allProfilesWithPictures} from "./option-item-examples";
 
 import argTypes from "./combobox.argtypes";
@@ -128,6 +129,8 @@ export const Default: Story = {
         children: items,
         selectionType: "single",
     },
+    // Hide the story in the Docs page (useful for snapshots).
+    tags: ["!autodocs"],
 };
 
 /**
@@ -192,31 +195,49 @@ export const ControlledCombobox: Story = {
         const [opened, setOpened] = React.useState(args.opened);
         const [value, setValue] = React.useState(args.value);
 
+        React.useEffect(() => {
+            setOpened(args.opened);
+        }, [args.opened]);
+
         return (
-            <Combobox
-                {...args}
-                opened={opened}
-                onToggle={() => {
-                    setOpened(!opened);
-                    action("onToggle")();
-                }}
-                onChange={(newValue) => {
-                    setValue(newValue);
-                    action("onChange")(newValue);
-                }}
-                value={value}
-            />
+            <View style={{gap: spacing.medium_16}}>
+                <Checkbox label="Open" onChange={setOpened} checked={opened} />
+                <Combobox
+                    {...args}
+                    opened={opened}
+                    onToggle={() => {
+                        setOpened(!opened);
+                        action("onToggle")();
+                    }}
+                    onChange={(newValue) => {
+                        setValue(newValue);
+                        action("onChange")(newValue);
+                    }}
+                    value={value}
+                />
+            </View>
         );
     },
     args: {
         children: items,
-        opened: true,
+        opened: false,
     },
     decorators: [
         (Story): React.ReactElement<React.ComponentProps<typeof View>> => (
             <View style={styles.wrapper}>{Story()}</View>
         ),
     ],
+    play: async ({canvasElement}) => {
+        // Arrange
+        const canvas = within(canvasElement.ownerDocument.body);
+
+        // Act
+        // Open the combobox by clicking the checkbox
+        await userEvent.click(canvas.getByRole("checkbox"));
+
+        // Assert
+        await expect(canvas.getByRole("listbox")).toBeVisible();
+    },
 };
 
 /**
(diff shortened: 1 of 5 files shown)


__docs__/wonder-blocks-dropdown/combobox.stories.tsx
```diff
@@ -192,31 +195,49 @@ export const ControlledCombobox: Story = {
         const [opened, setOpened] = React.useState(args.opened);
         const [value, setValue] = React.useState(args.value);
 
+        React.useEffect(() => {
+            setOpened(args.opened);
+        }, [args.opened]);
+
```
jandrade, line 201: **note:** refactored this story to open the controlled programmatically via the SB `play` feature. There was an odd issue where the combobox was opened on load, then to focus would move to this combobox instead of keeping the scroll at the top of the page. You can see the repro here: https://khan.github.io/wonder-blocks/?path=/docs/packages-dropdown-combobox--docs

Now that I switched to play, Chromatic still snapshots the story when it is opened, but we don't see it opened in the Docs page (which I think it makes sense).

__docs__/wonder-blocks-dropdown/combobox.stories.tsx
```diff
@@ -323,6 +327,24 @@ export const ControlledMultilpleCombobox: Story = {
             <View style={styles.wrapper}>{Story()}</View>
         ),
     ],
+
```
jandrade, line 330: **note:** Moved `play` from the previous story to the new one as it made sense to snapshot the controlled version of the multi-select mode. That way, the default multi-select version appears closed on load: 
https://5e1bf4b385e3fb0020b7073c-uwiczkggys.chromatic.com/?path=/story/packages-dropdown-combobox--multiple-selection

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2024-08-29 at 5 43 05 PM | Screenshot 2024-08-29 at 5 43 12 PM |

Before this change, both stories, the default one and the controlled one were throwing a very similar snapshot. Now both snapshots capture different states.

packages/wonder-blocks-dropdown/src/components/__tests__/combobox.test.tsx
```diff
@@ -153,6 +153,29 @@ describe("Combobox", () => {
         expect(screen.getByRole("combobox")).toHaveValue("");
     });
 
+    it("should reset the input to its previous selected value when the user presses Escape", async () => {
```
jandrade, line 156: **note:** This test verifies one of the cases that you found in the previous PR, where the combobox wasn't being changed back to its previous selected label.

packages/wonder-blocks-dropdown/src/components/combobox.tsx
```diff
@@ -186,6 +186,7 @@ export default function Combobox({
         children: currentOptions,
         disabled,
         id: uniqueId,
+        onChange: (value) => handleChange(value),
```
jandrade, line 189: **note:** This is the main change in this PR. Instead of relying on `useEffect`, now we rely on the hook to call `onChange` so we can control when the consumer needs to act based on that change (instead of `useEffect` controlling that).

packages/wonder-blocks-dropdown/src/components/combobox.tsx
```diff
@@ -330,6 +329,26 @@ export default function Combobox({
         [setFocusedIndex],
     );
 
+    /**
+     * Filters the items based on the input value. This is used when the
+     * `autoComplete` prop is set to `list` or `both`.
+     */
+    const filterItems = React.useCallback(
```
jandrade, line 336: **note:** Just moved this function next to `focusOnFilteredItem` as both did similar things (not the same things, but pretty similar).

packages/wonder-blocks-dropdown/src/components/combobox.tsx
```diff
@@ -400,6 +410,33 @@ export default function Combobox({
         [selected, setSelected],
     );
 
+    React.useEffect(() => {
+        // Focus on the combobox input when the dropdown is opened.
+        if (openState) {
+            comboboxRef.current?.focus();
+        }
+    }, [openState]);
```
jandrade, line 418: **note:** `useEffect` was simplified just to controlling when we need to focus on the input.

packages/wonder-blocks-dropdown/src/components/combobox.tsx
```diff
@@ -400,6 +410,33 @@ export default function Combobox({
...
+                );
+                return item ? getLabel(item?.props) : "";
+            });
+        }
+
+        // Single selection mode still wraps the selected value in an array
+        // to allow for a consistent API in ComboboxLiveRegion.
+        return [labelFromSelected];
+    }, [children, labelFromSelected, selected]);
+
```
jandrade, line 439: **note:** This fixes the ordering issue with the multi-select labels. The problem was that I was iterating over the `children` array and mapping that to the labels.

```
return children
            .filter((item) => selected?.includes(item.props.value))
            .map((item) => getLabel(item.props) as string);
```
Instead of taking this approach, now I iterate over the array of `selected` values and map that to the relevant child that matches the value.

See working example here: https://5e1bf4b385e3fb0020b7073c-uwiczkggys.chromatic.com/?path=/story/packages-dropdown-combobox--multiple-selection

packages/wonder-blocks-dropdown/src/hooks/use-listbox.tsx
```diff
@@ -70,7 +76,7 @@ export function useListbox({
     };
 
     const focusPreviousItem = React.useCallback(() => {
-        if (focusedIndex === 0) {
+        if (focusedIndex <= 0) {
```
jandrade, line 79: **note:** Found an issue when the focus was on the input and I pressed ArrowUp. The problem was that the listbox visual focus was lost in that case, but this is fixed with that change.


Uploading Screen Recording 2024-08-29 at 5.54.08 PM.mov…

beaesguerra (approved):
Lots of fixes and improvements, nice work!! 🎉

packages/wonder-blocks-dropdown/src/hooks/use-listbox.tsx
jandrade: oh oops, here it is for posteriority 😄 

https://github.com/user-attachments/assets/301abd3d-f012-42f2-a8ea-19eda1d99c80

PR #2311, opened 2024-08-29
Updated TextField & TextArea styling (for consistency and addresses cropped outline)

beaesguerra:
## Summary:

Work was initially done in https://github.com/Khan/wonder-blocks/pull/2302 to update the TextField state styling so it was consistent with other components like SingleSelect, MultiSelect, and TextArea. The new styling had some issues in webapp where the focus outline would get cropped off if an ancestor element had `overflow: hidden`. The new styles were reverted in https://github.com/Khan/wonder-blocks/pull/2309 to unblock other WB changes at the time. This PR re-applies those styles and addresses the focus outline issue in both the TextField and TextArea components.

The cropped focus outline issue is resolved by making sure the outlines are within the component. This makes it so the outline is never cropped off if a parent or ancestor element has `overflow: hidden`. The changes that address the outline issue are (see https://github.com/Khan/wonder-blocks/pull/2311/commits/51d9e4f5a87a749542a62359802aeac1cf0733eb):
- using a negative offset for outlines so that they appear within the component (instead of outside)
- simplifying outline styling by removing the box-shadow styling. Only border + outline are used for styling now

Other alternatives considered:
- Keeping the outline outside and adding additional padding + negative margin so there's enough space for the outline - I think adding spacing around our components isn't ideal because spacing is often added to the components and would override the built in spacing
- Adding the outline styles to a pseudo element (`:after`)  - doesn't work with input elements
- Continue using box-shadow - box-shadow styling also gets cropped out. Could use `inset` option though. I think using outline simplifies things though!

## Test Plan
- Confirm that the states for TextField and TextArea are as expected. Note: I added `overflow: hidden` to the component variant stories to simulate the issue (see before & after screenshots)
  - TextField `?path=/docs/packages-form-textfield-all-variants--docs`
  - TextArea `?path=/docs/packages-form-textarea-all-variants--docs`

| Before | After |
| --- | --- |
| Screenshot 2024-08-29 at 8 58 51 AM | Screenshot 2024-08-29 at 8 59 06 AM |
| Screenshot 2024-08-29 at 8 59 37 AM | Screenshot 2024-08-29 at 8 59 14 AM | 

- Confirmed in webapp that focus outline is not cropped off in TextArea and TextField components 

Issue: WB-1746

### DIFF
--- a/.changeset/two-toys-raise.md
+++ b/.changeset/two-toys-raise.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/wonder-blocks-form": patch
+---
+
+- Update `TextField` state styling so that it is consistent with other components like `TextArea`, `SingleSelect`, `MultiSelect` (especially the focus styling). The styling also now uses CSS pseudo-classes for easier testing in Chromatic and debugging in browsers.
+- `TextField` and `TextArea` state styling has also been updated so that any outline styles outside of the component are now applied within the component to prevent cropped focus outlines in places where an ancestor element has `overflow: hidden`.
--- a/__docs__/wonder-blocks-form/text-area-variants.stories.tsx
+++ b/__docs__/wonder-blocks-form/text-area-variants.stories.tsx
@@ -164,5 +164,6 @@ const styles = StyleSheet.create({
     },
     scenario: {
         gap: spacing.small_12,
+        overflow: "hidden",
     },
 });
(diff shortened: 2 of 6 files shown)


beaesguerra:
Chromatic visual tests are available again now! Note that the visual tests includes other visual changes that has been added since it was disabled

jandrade (approved):
Looks great. Thanks for finding a suitable alternative 🎉

PR #2312, opened 2024-08-29
[fei5815.1.removetsinfo] Add .npmignore so we stop packaging ts config and build info

somewhatabstract:
## Summary:
Our packages are shipping the tsconfig and tsbuildinfo files ([example][1]) - this is not necessary. This change adds an `.npmignore` file to prevent these files from being included in the packages.

However, our current publishing approach doesn't seem to honour that ignore file even though `npm pack --dry-run` does. So, this also updates the tooling to remove files according to whether they are ignored by both the `.npmignore` file and the `.gitignore` file (to avoid deleting source-controlled files). In addition, each package.json now explicitly lists what files it should be including.

This combination means we can easily exclude files and folders in the `dist` location, without having really explicit `files` definitions that might be hard to maintain. 

Issue: FEI-5815

## Test plan:
Check the pre-release packages to see that they no longer include these files.

[1]:https://www.npmjs.com/package/@khanacademy/wonder-blocks-core/v/7.0.0?activeTab=code


jeresig (approved):

somewhatabstract:
So, the `.npmignore` change alone wasn't sufficient. It seems that while it worked for `npm pack --dry-run`, it isn't working with changesets. I've raised a bug for that [here][1].

In the meantime, I've adjusted our existing tooling to do the tidy up for us. It airs on the side of caution and only deletes files and directories that match both the `.npmignore` and the `.gitignore`. This uses the `ignore` package which is used by `checksync` for interpreting ignore files.

[1]:https://github.com/changesets/changesets/issues/1447

somewhatabstract:
OK, this is how things probably should be. Each package is explicitly listing what files to package - I've set this to `dist` folder and the `CHANGELOG.md` file. I've excluded `src` since I don't think we need to ship that anymore (we had to with Flow for types to work).

I've left the post-build script to tidy things a bit since that is useful to still clean up some things that live inside of `dist` that we don't want to ship - specifically test file stuff.

somewhatabstract:
Final update. I discovered that some type definitions were incorrectly importing types from source code. This meant that when we stopped shipping source, those types were broken.

I found the fix was to make sure the referenced types are exported from the relevant packages. I have fixed them and added a new check that verifies type definitions. If it finds an incorrect import of a type, it will error like this:

```
packages/wonder-blocks-dropdown/dist/hooks/use-listbox.d.ts:62:24
    Incorrectly imported type HorizontalRuleVariant from wonder-blocks-cell source
    Update the package wonder-blocks-cell to export the type HorizontalRuleVariant

packages/wonder-blocks-dropdown/dist/hooks/use-listbox.d.ts:65:20
    Incorrectly imported type TypographyText from wonder-blocks-cell source
    Update the package wonder-blocks-cell to export the type TypographyText

packages/wonder-blocks-dropdown/dist/hooks/use-listbox.d.ts:66:20
    Incorrectly imported type TypographyText from wonder-blocks-cell source
    Update the package wonder-blocks-cell to export the type TypographyText

packages/wonder-blocks-dropdown/dist/hooks/use-listbox.d.ts:71:25
    Incorrectly imported type HorizontalRuleVariant from wonder-blocks-cell source
    Update the package wonder-blocks-cell to export the type HorizontalRuleVariant
```

jeresig (approved):

PR #2313, opened 2024-08-30
WB-1755: Toolbar - Allow title to also support React nodes

jandrade:
## Summary:

There’s a valid use case where a toolbar needs to use a progress bar in the
center area. Right now, that part of the layout only accepts a title as a
string.

See design screenshot for reference:

Screenshot 2024-08-30 at 1 56 40 PM

This change allows the title to also accept React nodes, which will allow
engineers to use more complex components in that part of the layout.

Issue: https://khanacademy.atlassian.net/browse/WB-1755

## Test plan:

Navigate to the new story in the toolbar documentation and verify that the
progress bar is displayed in the center area.

/?path=/docs/packages-toolbar--docs#custom%20toolbar

Screenshot 2024-08-30 at 1 56 15 PM

### DIFF
--- a/__docs__/wonder-blocks-toolbar/toolbar.stories.tsx
+++ b/__docs__/wonder-blocks-toolbar/toolbar.stories.tsx
@@ -3,7 +3,7 @@ import {StyleSheet} from "aphrodite";
 import type {Meta, StoryObj} from "@storybook/react";
 
 import {View} from "@khanacademy/wonder-blocks-core";
-import {spacing} from "@khanacademy/wonder-blocks-tokens";
+import {semanticColor, spacing} from "@khanacademy/wonder-blocks-tokens";
 
 import Toolbar from "@khanacademy/wonder-blocks-toolbar";
 import packageConfig from "../../packages/wonder-blocks-toolbar/package.json";
@@ -209,5 +209,29 @@ Dark.parameters = {
                 "A toolbar can be also used in a dark color scheme.\n\n" +
                 "**Note:** Notice that there's a white hairline in this case, and that the background is transparent (to allow the Dark Blue or illustrated background to come through).",
         },
+    },
+};
+
+/**
+ * Sometimes we need to have a custom toolbar to include a custom component or a
+ * more complex layout in the center of the toolbar. This can be achieved by
+ * passing a React node to the `title` prop.
+ *
+ * **NOTE:** This approach should be used with caution, as it may break the
+ * layout of the toolbar.
+ */
+export const CustomToolbar: StoryComponentType = {
+    args: {
+        leftContent: leftContentMappings.exitWithTitle,
+        rightContent: rightContentMappings.primaryButton,
+        title: (
+            <View
+                style={{
+                    width: 300,
+                    height: spacing.xSmall_8,
+                    background: semanticColor.mastery.primary,
+                }}
+            />
+        ),
     },
 };
--- a/packages/wonder-blocks-toolbar/src/components/toolbar.tsx
+++ b/packages/wonder-blocks-toolbar/src/components/toolbar.tsx
@@ -40,20 +40,14 @@ type Props = AriaProps & {
      */
     subtitle?: string;
     /**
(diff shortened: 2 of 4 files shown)


packages/wonder-blocks-toolbar/src/components/toolbar.tsx
```diff
@@ -73,74 +67,70 @@ type DefaultProps = {
...
-        rightContent: null,
-        size: "medium",
-    };
-
-    render(): React.ReactNode {
-        const {color, leftContent, rightContent, size, subtitle, title} =
-            this.props;
-
-        const TitleComponent = subtitle ? LabelLarge : HeadingSmall;
+export default function Toolbar({
```
jandrade, line 70: **note:** Took the opportunity to convert this to a functional component. Most of the code remains the same (except for the custom title part).

packages/wonder-blocks-toolbar/src/components/toolbar.tsx
```diff
@@ -73,74 +67,70 @@ type DefaultProps = {
...
-                >
-                    {rightContent}
                 </View>
+            ) : (
+                // We don't use wideColumn here to allow more flexibility with
+                // the custom node.
+                <View style={[sharedStyles.column]}>
+                    <View style={[sharedStyles.titles]}>{title}</View>
+                </View>
+            )}
```
jandrade, line 121: **note:** This is the actual change that allows for more flexibility to the `title` prop.

packages/wonder-blocks-toolbar/src/components/toolbar.tsx
```diff
@@ -40,22 +40,16 @@ type Props = AriaProps & {
      */
     subtitle?: string;
     /**
-     * The main title rendered in larger bold text.
+     * The main title rendered in larger bold text. It also supports rendering
+     * React nodes (use with caution).
      */
-    title?: string;
+    title?: string | React.ReactNode;
```
kevinb-khan, line 46: This is the main functional change right?

Also, I believe that `ReactNode` contains `string` so this can probably be simplified to `title?: React.ReactNode`.

packages/wonder-blocks-toolbar/src/components/toolbar.tsx
```diff
@@ -73,74 +67,70 @@ type DefaultProps = {
...
-                    ]}
-                >
-                    {rightContent}
                 </View>
+            ) : (
+                // We don't use wideColumn here to allow more flexibility with
+                // the custom node.
+                <View style={[sharedStyles.column]}>
+                    <View style={[sharedStyles.titles]}>{title}</View>
+                </View>
```
kevinb-khan, line 120: Nice!

kevinb-khan (approved):
LGTM 🚀   I'm not sure why the new story didn't show up here: https://5e1bf4b385e3fb0020b7073c-jgvoqajhqi.chromatic.com/?path=/docs/packages-toolbar--docs.

__docs__/wonder-blocks-toolbar/toolbar.stories.tsx
```diff
@@ -211,3 +211,27 @@ Dark.parameters = {
         },
     },
 };
+
+/**
+ * Sometimes we need to have a custom toolbar to include a custom component or a
+ * more complex layout in the center of the toolbar. This can be achieved by
+ * passing a React node to the `title` prop.
+ *
+ * **NOTE:** This approach should be used with caution, as it may break the
```
beaesguerra, line 220: Great idea to document this warning!

beaesguerra (approved):
Looks good to me! 🎉

packages/wonder-blocks-toolbar/src/components/toolbar.tsx
jandrade: That's a good point, although, I'd prefer to keep it that way as Storybook autogenerates docs based on this types. By keeping this, we signal devs that "string" is the default way to go, but that they can still rely on `ReactNode` as an alternative approach.

PR #2316, opened 2024-09-06
Add `viewportPadding` prop to `Tooltip` and `Popover` components

beaesguerra:
## Summary:
This new `viewportPadding` prop provides spacing between the popper and the viewport edges in the `Tooltip` and `Popover` components. If this prop is not provided, default spacing is applied so that the popper contents are not too close to the edge by default. This can be overridden to support different use cases.

Note: `viewportPadding` is only applied if the `rootBoundary` prop is set to `viewport`. `Tooltip` will always use `viewport` for the `rootBoundary`, however, this prop is exposed in the `Popover` component

Issue: WB-1756

## Test plan:
- Tooltip
  - Confirm that default spacing is applied to the Tooltip contents when it is in the corners (`?path=/story/packages-tooltip-tooltip--in-corners`)
  - Confirm that changing the `viewportPadding` control in Storybook updates the spacing between the Tooltip popper and the viewport edge
  - Confirm prop documentation makes sense (`?path=/docs/packages-tooltip-tooltip--docs`)
- Popover
  - Confirm that default spacing is applied to the Popover contents when it is in the corners (`?path=/story/packages-popover-popover--in-corners`)
  - Confirm that changing the `viewportPadding` control in Storybook updates the spacing between the Popover popper and the viewport edge 
  - Confirm prop documentation makes sense (`?path=/docs/packages-popover-popover--docs`)
- Confirm that Tooltip and Popover components in webapp continue to work as expected

### DIFF
--- a/__docs__/wonder-blocks-popover/popover.stories.tsx
+++ b/__docs__/wonder-blocks-popover/popover.stories.tsx
@@ -3,7 +3,7 @@ import {StyleSheet} from "aphrodite";
 import type {Meta, StoryObj} from "@storybook/react";
 
 import Button from "@khanacademy/wonder-blocks-button";
-import {View} from "@khanacademy/wonder-blocks-core";
+import {PropsFor, View} from "@khanacademy/wonder-blocks-core";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
 import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
 import {HeadingMedium, LabelLarge} from "@khanacademy/wonder-blocks-typography";
@@ -832,4 +832,60 @@ export const WithCustomAriaDescribedBy = ({
             </Popover>
         </View>
     );
+};
+
+export const InCorners = (args: PropsFor<typeof Popover>) => {
+    const [openedIndex, setOpenedIndex] = React.useState<number>(0);
+    const renderPopover = (index: number) => {
+        return (
+            <Popover
+                {...args}
+                content={
+                    <PopoverContent
+                        closeButtonVisible
+                        content="The default version only includes text."
+                        title="A simple popover"
+                    />
+                }
+                dismissEnabled
+                onClose={() => setOpenedIndex(-1)}
+                opened={openedIndex === index}
+            >
+                <Button onClick={() => setOpenedIndex(index)}>
+                    Open default popover
+                </Button>
+            </Popover>
+        );
+    };
+    return (
+        <View
+            style={{
+                height: "100vh",
+                width: "100vw",
+                justifyContent: "space-between",
+            }}
+        >
+            <View
+                style={{
+                    flexDirection: "row",
+                    justifyContent: "space-between",
+                }}
+            >
+                {renderPopover(0)}
+                {renderPopover(1)}
+            </View>
+            <View
+                style={{
+                    flexDirection: "row",
+                    justifyContent: "space-between",
+                }}
+            >
+                {renderPopover(2)}
+                {renderPopover(3)}
+            </View>
+        </View>
+    );
+};
+InCorners.parameters = {
+    layout: "fullscreen",
 };
--- a/__docs__/wonder-blocks-popover/popover.stories.tsx
+++ b/__docs__/wonder-blocks-popover/popover.stories.tsx
@@ -834,6 +834,14 @@ export const WithCustomAriaDescribedBy = ({
     );
 };
 
+/**
+ * If the Popover is placed near the edge of the viewport, default spacing of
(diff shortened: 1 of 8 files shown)


__docs__/wonder-blocks-popover/popover.stories.tsx
```diff
@@ -833,3 +833,59 @@ export const WithCustomAriaDescribedBy = ({
         </View>
     );
 };
+
+export const InCorners = (args: PropsFor<typeof Popover>) => {
```
jandrade, line 837: **suggestion:** Could you please add a comment here to explain to devs how this example works? (including that they need to use `viewportPadding`)

__docs__/wonder-blocks-tooltip/tooltip.stories.tsx
```diff
@@ -43,9 +43,12 @@ export default {
...
-        (Story): React.ReactElement => (
-            <View style={styles.storyCanvas}>{Story()}</View>
-        ),
+        (Story, {parameters}): React.ReactElement =>
+            parameters.layout === "fullscreen" ? (
+                Story()
+            ) : (
+                <View style={styles.storyCanvas}>{Story()}</View>
+            ),
     ],
```
jandrade, line 52: **praise:** Neat approach!

packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx
```diff
@@ -44,10 +45,17 @@ type Props = {
      * on where there is available room within the document body.
      */
     rootBoundary?: RootBoundary;
+    /**
+     * If `rootBoundary` is `viewport`, this padding value is used to provide
+     * spacing between the popper and the viewport. If not provided, default
+     * spacing is applied.
```
jandrade, line 51: **suggestion:** How do you feel about specifying the default spacing value here? (e.g. 12px)

packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx
```diff
@@ -240,6 +250,7 @@ export default class TooltipPopper extends React.Component<Props> {
                 name: "preventOverflow",
                 options: {
                     rootBoundary: "viewport",
+                    padding: viewportPadding,
```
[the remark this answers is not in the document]
jandrade: I think that makes sense. In fact, this would be a nice addition to the docs: specifying that this will only works if `rootBoundary` is viewport.

packages/wonder-blocks-tooltip/src/components/tooltip.tsx
```diff
@@ -116,6 +116,12 @@ type Props = AriaProps &
          * Optional background color.
          */
         backgroundColor?: keyof typeof color;
+        /**
+         * If `rootBoundary` is `viewport`, this padding value is used to provide
+         * spacing between the popper and the viewport. If not provided, default
+         * spacing is applied.
```
jandrade, line 122: **suggestion:** same comment about including the spacing value in px.

__docs__/wonder-blocks-tooltip/tooltip.stories.tsx
```diff
@@ -509,3 +512,53 @@ export const InTopCorner = {
         </View>
     ),
 };
+
+export const InCorners = {
```
jandrade, line 516: **suggestion:** Same for including a description to this history here. To expand on that, probably worth mentioning that we provide an spacing by default, but that it can be overwitten using .... if needed.

jandrade (approved):
Looks great overall! I just left some comments for improving the docs for you to consider before landing 🚀

__docs__/wonder-blocks-popover/popover.stories.tsx
beaesguerra: Yes good catch! Added 😄

__docs__/wonder-blocks-tooltip/tooltip.stories.tsx
beaesguerra: This was because even if the layout parameter was set to `fullscreen` for a story, the overall decorator for the stories in the file would still get applied and add additional spacing around the story. I couldn't figure out how to opt out of the decorator for one story so I added this condition!

packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx
beaesguerra: I added it in! I was avoiding adding specific implementation details so there was one source of truth, but I could see how explicitly documenting the details would be useful! We will need to make sure to update the docs if the value ever changes 😄

PR #2320, opened 2024-09-11
Allow TextField to be focusable when disabled

beaesguerra:
## Summary:
For better accessibility, interactive components in WB can be focusable, even when disabled.
These changes to TextField include:

- Setting aria-disabled instead of the disabled attribute
- Disabling event handlers if the disabled prop is true. onFocus and onBlur continues to work if disabled prop is true
- Setting the readonly attribute if the disabled prop is true to prevent interactions with textarea

These changes are very similar to the changes applied to the TextArea component in https://github.com/Khan/wonder-blocks/pull/2284

Issue: WB-1751

## Test plan:
- Review docs around disabled state
  - TextField (`?path=/docs/packages-form-textfield--docs`)
  - LabeledTextField (`?path=/docs/packages-form-labeledtextfield--docs`)
- Review TextField disabled state (`?path=/story/packages-form-textfield--disabled`)
  - Confirm the user can't type in the TextField if it is disabled
  - Confirm the user can still focus on the TextField if it is disabled 
  - Confirm screen readers continue to communicate that the field is disabled
- Review LaeledTextField disabled state (`?path=/story/packages-form-labeledtextfield--disabled`)

## Screen recording

Keyboard navigation + voiceover

https://github.com/user-attachments/assets/f106cfb3-600f-4c47-a2fa-8ac744e22282

https://github.com/user-attachments/assets/ff78b8e8-ab55-4eff-bdcc-444ac2f82e89


packages/wonder-blocks-form/src/components/__tests__/text-field.test.tsx
```diff
@@ -158,24 +158,6 @@ describe("TextField", () => {
         expect(input).toBeInTheDocument();
     });
 
-    it("disabled prop disables the input element", async () => {
```
beaesguerra, line 161: Moved this test with the other `Disabled state` tests! It was also updated to check for the `aria-disabled` attribute instead

jandrade (approved):
Looks really great! thanks for adding all the test cases and the comments on the event handlers 👏 🚀

packages/wonder-blocks-form/src/components/text-field.tsx
```diff
@@ -291,13 +299,13 @@ class TextField extends React.Component<PropsWithForwardRef, State> {
...
                         onChange={this.handleChange}
-                        onKeyDown={onKeyDown}
-                        onFocus={this.handleFocus}
-                        onBlur={this.handleBlur}
+                        onKeyDown={disabled ? undefined : onKeyDown}
+                        onFocus={this.handleFocus} // TextField can be focused if disabled
+                        onBlur={this.handleBlur} // TextField can be blurred if disabled
                         data-testid={testId}
-                        readOnly={readOnly}
+                        readOnly={readOnly || disabled} // Set readOnly also if it is disabled, otherwise users can type in the field
```
jandrade, line 308: **side note:** It's great that you are including this here so the `Combobox` gets this by free :)

PR #2321, opened 2024-09-13
WB-1757.1: Combobox - Add error prop

jandrade:
## Summary:

- Added `error` prop to the `Combobox` component to handle the error state.
- Included an X icon to clear the combobox selection in single-selection mode. This is useful for the user to quickly reset the value when needed. 

- Textbox: Modifies the order in which the `aria-invalid` attribute is applied
to the input element.

Issue: WB-1757

## Test plan:

1. Navigate to the `Error` story: /?path=/story/packages-dropdown-combobox--error
2. Verify that the error state is displayed and `aria-invalid` is set `true`.
3. Open the listbox and select one option.
4. Verify that the error is gone and `aria-invalid=false`.
5. Verify that the "Clear selection" button (X) is now visible.
6. Click on the X button.
7. Verify that the error is visible again and `aria-invalid=true`.

https://github.com/user-attachments/assets/22f69acd-9345-45aa-a626-ea0822a85c29

### DIFF
--- a/__docs__/wonder-blocks-dropdown/combobox.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/combobox.stories.tsx
@@ -396,3 +396,27 @@ export const AutoCompleteMultiSelect: Story = {
         },
     },
 };
+
+export const Error: Story = {
+    render: function Render(args: PropsFor<typeof Combobox>) {
+        const [error, setError] = React.useState(args.error);
+        const [value, setValue] = React.useState(args.value);
+
+        return (
+            <Combobox
+                {...args}
+                error={error}
+                value={value}
+                onChange={(newValue) => {
+                    setValue(newValue);
+                    setError(newValue !== "" ? false : true);
+                    action("onChange")(newValue);
+                }}
+            />
+        );
+    },
+    args: {
+        children: items,
+        error: true,
+    },
+};
--- a/__docs__/wonder-blocks-dropdown/combobox.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/combobox.stories.tsx
@@ -397,6 +397,14 @@ export const AutoCompleteMultiSelect: Story = {
     },
 };
 
+/**
+ * This `Combobox` is in an error state. Selecting any option will clear the
+ * error state by updating the `error` prop to `false`.
+ *
+ * **NOTE:** We internally apply the correct `aria-invalid` attribute based on
+ * the `error` prop.
+ */
+
 export const Error: Story = {
     render: function Render(args: PropsFor<typeof Combobox>) {
         const [error, setError] = React.useState(args.error);
--- a/packages/wonder-blocks-dropdown/src/components/__tests__/combobox.test.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/__tests__/combobox.test.tsx
@@ -6,6 +6,7 @@ import {PointerEventsCheckLevel, userEvent} from "@testing-library/user-event";
 import Combobox from "../combobox";
 import OptionItem from "../option-item";
 import {defaultComboboxLabels} from "../../util/constants";
+import {MaybeValueOrValues} from "../../util/types";
 
 const doRender = (element: React.ReactElement) => {
     render(element, {wrapper: RenderStateRoot});
@@ -58,7 +59,9 @@ describe("Combobox", () => {
         );
 
         // Act
-        await userEvent.click(screen.getByRole("button"));
+        await userEvent.click(
+            screen.getByRole("button", {name: /toggle listbox/i}),
+        );
 
         // Assert
         await screen.findByRole("listbox", {hidden: true});
@@ -96,11 +99,15 @@ describe("Combobox", () => {
             </Combobox>,
         );
 
-        await userEvent.click(screen.getByRole("button"));
+        await userEvent.click(
+            screen.getByRole("button", {name: /toggle listbox/i}),
+        );
         await screen.findByRole("listbox", {hidden: true});
 
         // Act
-        await userEvent.click(screen.getByRole("button"));
+        await userEvent.click(
+            screen.getByRole("button", {name: /toggle listbox/i}),
+        );
 
         // Assert
         expect(
@@ -214,7 +221,9 @@ describe("Combobox", () => {
             </Combobox>,
         );
 
-        await userEvent.click(screen.getByRole("button"));
+        await userEvent.click(
+            screen.getByRole("button", {name: /toggle listbox/i}),
+        );
         await screen.findByRole("listbox", {hidden: true});
 
         // Act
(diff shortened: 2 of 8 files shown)


packages/wonder-blocks-dropdown/src/components/__tests__/combobox.test.tsx
```diff
@@ -296,6 +305,109 @@ describe("Combobox", () => {
...
+            </Combobox>,
+        );
+
+        // Act
+        await userEvent.click(
+            screen.getByRole("button", {name: /clear selection/i}),
+        );
+
+        // Assert
+        expect(screen.getByRole("combobox")).toHaveValue("");
```
somewhatabstract, line 324: **suggestion:** I know this is an established pattern. Just wanted to note that I would move the `screen.getByRole` into the `// Act` as a `const result =  screen.getByRole` so that the expectation is purely inert and just doing the expectation check. Not a big deal, just wanted to note it.

somewhatabstract (approved):
Looks good. Thanks for adding these features! I have a couple of inline comments.

__docs__/wonder-blocks-dropdown/combobox.stories.tsx
```diff
@@ -396,3 +396,35 @@ export const AutoCompleteMultiSelect: Story = {
...
+
+        return (
+            <Combobox
+                {...args}
+                error={error}
+                value={value}
+                onChange={(newValue) => {
+                    setValue(newValue);
+                    setError(newValue !== "" ? false : true);
+                    action("onChange")(newValue);
```
beaesguerra, line 421: TIL we could do this!

packages/wonder-blocks-dropdown/src/components/__tests__/combobox.test.tsx
```diff
@@ -296,6 +305,109 @@ describe("Combobox", () => {
...
+            );
+
+            // Assert
+            expect(screen.getByRole("combobox")).toHaveAttribute(
+                "aria-invalid",
+                "false",
+            );
+        });
+
+        it("should display the errror message when the value is invalid", async () => {
```
beaesguerra, line 347: question: Should the test be something like "should use aria-invalid=true if error is true" or something since it is checking for aria-invalid rather than an error message?

packages/wonder-blocks-dropdown/src/components/__tests__/combobox.test.tsx
```diff
@@ -296,6 +305,109 @@ describe("Combobox", () => {
...
+            await userEvent.tab();
+
+            // Assert
+            expect(screen.getByRole("combobox")).toHaveAttribute(
+                "aria-invalid",
+                "true",
+            );
+        });
+
+        it("should clear the error message when the value is valid", async () => {
```
beaesguerra, line 367: Same question about referencing aria-invalid instead of error message in the test description!

packages/wonder-blocks-dropdown/src/util/constants.ts
```diff
@@ -48,6 +48,7 @@ export const defaultLabels = {
 } as const;
 
 export const defaultComboboxLabels: ComboboxLabels = {
+    clearSelection: "Clear selection",
```
beaesguerra, line 51: praise: I like that this copy could be used for both single and multiple selection types!

packages/wonder-blocks-dropdown/src/util/types.ts
```diff
@@ -63,6 +63,11 @@ export type MaybeValueOrValues = MaybeString | Array<MaybeString>;
  * The labels for the combobox component.
  */
 export type ComboboxLabels = {
+    /**
+     * Label for the "Clear" button that removes a selected item. Only used in
+     * single-select mode.
```
beaesguerra, line 68: question: Is there a reason we can't use it with the multiple selection type?

packages/wonder-blocks-dropdown/src/components/combobox.tsx
```diff
@@ -521,18 +530,44 @@ export default function Combobox({
...
+                    <IconButton
+                        disabled={disabled}
+                        icon={xIcon}
+                        onClick={(e) => {
+                            e.stopPropagation();
+                            // Reset the combobox value.
+                            setInputValue("");
+                            setSelected("");
+                            onChange?.("");
+                            comboboxRef.current?.focus();
```
beaesguerra, line 556: question: Would it be helpful to have a screen reader announcement for when the selected value is cleared? What do you think?

beaesguerra (commented):
Looks good! Just had a few questions 😄

packages/wonder-blocks-dropdown/src/util/types.ts
beaesguerra: Right! The clear button for multi-selection could be helpful for a "Clear all" functionality, especially if there are many pills selected! This could be a future improvement though if it's needed 😄

beaesguerra:
I forgot to review the Chromatic tests last time I reviewed, sorry! I left some comments in Chromatic :)

packages/wonder-blocks-dropdown/src/components/combobox.tsx
jandrade: Makes total sense! I tried making this work, but the announcement logic is becoming to complex, so I'm planning to refactor it in a follow up PR (cc @marcysutton).

marcysutton (commented):
This tested well with NVDA on Windows and VoiceOver! The ARIA invalid state is announced and you can navigate the combobox with both platforms. 👍 

Left you a suggestion for a testing improvement!

packages/wonder-blocks-dropdown/src/components/__tests__/combobox.test.tsx
```diff
@@ -296,6 +305,109 @@ describe("Combobox", () => {
...
+        const userEvent = doRender(
+            <Combobox selectionType="single" value="option2">
+                <OptionItem label="option 1" value="option1" />
+                <OptionItem label="option 2" value="option2" />
+                <OptionItem label="option 3" value="option3" />
+            </Combobox>,
+        );
+
+        // Act
+        await userEvent.click(
```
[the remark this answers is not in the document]
jandrade: Thanks! makes total sense. I added some separate tests down below that use keyboard to verify the `aria-invalid` state, but that can be definitively improved to be more explicit about keyboard navigation. I'm going to go with your suggestion of adding a "keyboard" test (or describe block) to be more explicit about the need for testing in both mouse and keyboard devices.

For some extra context, we have done this in some other components, but it's not something that we have standardized as a testing pattern, and it makes sense to make this more of a standard now that you mention it :).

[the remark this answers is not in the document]
jandrade: @marcysutton I added the new test to verify keyboard. Let me know what you think: https://github.com/Khan/wonder-blocks/pull/2321/commits/aca75a023ffd80a2939749fb963ca41988261ae7

beaesguerra (approved):
Looks good, thanks for making those changes! I'm excited for our screen reader announcer :D

PR #2325, opened 2024-09-19
IconButton: Add type=submit prop to allow submitting forms using the button

jandrade:
## Summary:

This PR adds a new prop `type` to the `IconButton` component that allows the
button to submit a form when clicked or pressed. This is useful when we want to
use an IconButton to submit a form, for example, a search form.

This is done as I noticed that the IconButton component is missing the `type`
and was needed in the search page. The prop is currently set, but the type
definition is missing.

Issue: FEI-5728

## Test plan:

Docs:

- Navigate to /?path=/docs/packages-iconbutton--docs#submitting%20forms
- Verify that the docs make sense.

Interactive example:

- Navigate to /?path=/story/packages-iconbutton--submitting-forms
- Verify that the submit action is recorded in the SB `Actions` tab.

Screenshot 2024-09-19 at 10 14 56 AM

### DIFF
--- a/__docs__/wonder-blocks-icon-button/icon-button.stories.tsx
+++ b/__docs__/wonder-blocks-icon-button/icon-button.stories.tsx
@@ -21,6 +21,7 @@ import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
 import ComponentInfo from "../../.storybook/components/component-info";
 import packageConfig from "../../packages/wonder-blocks-icon-button/package.json";
 import IconButtonArgtypes from "./icon-button.argtypes";
+import TextField from "../../packages/wonder-blocks-form/src/components/text-field";
 
 /**
  * An `IconButton` is a button whose contents are an SVG image.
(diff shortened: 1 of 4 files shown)


packages/wonder-blocks-icon-button/src/components/__tests__/icon-button.test.tsx
```diff
@@ -320,4 +320,51 @@ describe("IconButton", () => {
...
+
+            // Act
+            const button = await screen.findByRole("button");
+            await userEvent.click(button);
+
+            // Assert
+            expect(submitFnMock).toHaveBeenCalled();
+        });
+
+        it("should submit button within form via keyboard", async () => {
```
beaesguerra, line 342: priase: nice work adding tests using different interactions! :accessibility:

__docs__/wonder-blocks-icon-button/icon-button.stories.tsx
```diff
@@ -355,6 +356,41 @@ export const WithRouter: StoryComponentType = {
...
+            <View style={styles.row}>
+                <LabelMedium tag="label" style={styles.row}>
+                    Search:{" "}
+                    <TextField
+                        id="foo"
+                        value="press the button"
+                        onChange={() => {}}
+                    />
+                </LabelMedium>
+                <IconButton icon={magnifyingGlass} type="submit" />
```
beaesguerra, line 382: suggestion: The IconButton could have an aria-label in the example!

beaesguerra: thought: I'm wondering, are there any reasons to not have an aria-label on an IconButton? For the most part, IconButtons should always do something since it's a button, so it'd make sense for it to always have an aria-label. What do you think? (no changes required for this, but what just curious!)

beaesguerra (approved):
Changes look good, nice work! 🎉 

Note: Looks like something went wrong with the deploy to Chromatic step, trying it again might fix it!

image

__docs__/wonder-blocks-icon-button/icon-button.stories.tsx
jandrade: hmmm yeah that makes a ton of sense to me! especially seeing how a11y could be improved just by forcing this label in all cases.

PR #2326, opened 2024-09-19
SearchField: Add onKeyUp prop

jandrade:
## Summary:

The SearchField component now accepts an `onKeyUp` prop that will be called when
the user releases a key while the input is focused.

This is to bring consistency with the other input components in the library and
also to allow for more flexibility when using the SearchField component.

Also took the opportunity to clear some of the SB argtype descriptions, as we no
longer need to duplicate them as of SB 8.0.

NOTE: This will be used as part of the Search page refactor.

Issue: FEI-5728

## Test plan:

Verify that the unit tests pass.

### DIFF
--- a/__docs__/wonder-blocks-search-field/search-field.argtypes.ts
+++ b/__docs__/wonder-blocks-search-field/search-field.argtypes.ts
@@ -1,6 +1,5 @@
 export default {
     clearAriaLabel: {
-        description: `ARIA label for the clear button. Defaults to "Clear search".`,
         type: {name: "string", required: false},
         table: {
             type: {
@@ -13,8 +12,6 @@ export default {
         },
     },
     id: {
-        description: `The unique identifier for the input. If one is not
-            provided, a unique id will be generated.`,
         type: {name: "string", required: false},
         table: {
             type: {
@@ -26,7 +23,6 @@ export default {
         },
     },
     value: {
-        description: "The text input value.",
         type: {name: "string", required: true},
         table: {
             type: {
@@ -36,8 +32,6 @@ export default {
         control: {type: "text"},
     },
     name: {
-        description: `The name for the input control. This is submitted along
-            with the form data.`,
         type: {name: "string", required: false},
         table: {
             type: {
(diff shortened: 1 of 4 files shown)


__docs__/wonder-blocks-search-field/search-field.argtypes.ts
```diff
@@ -1,6 +1,5 @@
 export default {
     clearAriaLabel: {
-        description: `ARIA label for the clear button. Defaults to "Clear search".`,
```
beaesguerra, line 3: praise: Thanks for cleaning this up!

I will also start removing descriptions from story argtypes as I go along so we can rely on the prop docs to avoid duplication 😄

packages/wonder-blocks-search-field/src/components/search-field.tsx
```diff
@@ -72,6 +72,10 @@ type Props = AriaProps & {
      * Called when a key is pressed.
      */
     onKeyDown?: (event: React.KeyboardEvent<HTMLInputElement>) => unknown;
+    /**
+     * Called when a key is released.
+     */
+    onKeyUp?: (event: React.KeyboardEvent<HTMLInputElement>) => unknown;
```
beaesguerra, line 78: question: I noticed we often deconstruct the specific props before applying it to the underlying element. Do we do that so that the remaining `otherProps` will usually be the aria props? Are there any drawbacks to letting other props be part of the `otherProps`?

Here's an example where we do that with the other event handlers:
https://github.com/Khan/wonder-blocks/blob/dd2759e5f5c8bc5016cee0773770987a71b28cac/packages/wonder-blocks-search-field/src/components/search-field.tsx#L126-L130

beaesguerra (approved):
Looks good, thanks for making these improvements 🎉

packages/wonder-blocks-search-field/src/components/search-field.tsx
jandrade: That's correct, we use that for aria props, and also for letting other valid attributes be injected in the inner node. I don't think there are drawbacks with that logic.... the only thing is that TS would complain, but this can be fixed by adding the missing props, like the one included here :).

PR #2328, opened 2024-09-19
Form: Fix a11y issue on RadioGroup and ChoiceGroup

jandrade:
## Summary:

`RadioGroup` and `ChoiceGroup` components are not associating correctly their
labels with their `fieldset` elements. The underlying issue is that these
components are appending a `View` element as the only child of the `fieldset`
element, which is causing the `fieldset` element to not associate correctly the
`legend` value as its label.

This PR fixes that issue by removing the `View` element and using the `legend`
element directly as the first child of the `fieldset` element.

Issue: XXX-XXXX

## Test plan:

Navigate to:

- CheckboxGroup: /?path=/docs/packages-form-checkboxgroup--docs
- RadioGroup: /?path=/docs/packages-form-radiogroup--docs

Verify that the first example associates the `label` value with the `fieldset`
element correctly.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2024-09-19 at 5 10 44 PM | Screenshot 2024-09-19 at 5 09 48 PM |
| Screenshot 2024-09-19 at 5 34 49 PM | Screenshot 2024-09-19 at 5 33 58 PM |

### DIFF
--- a/packages/wonder-blocks-form/src/components/checkbox-group.tsx
+++ b/packages/wonder-blocks-form/src/components/checkbox-group.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
 
-import {View, addStyle} from "@khanacademy/wonder-blocks-core";
+import {addStyle} from "@khanacademy/wonder-blocks-core";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
 import {spacing} from "@khanacademy/wonder-blocks-tokens";
 import {LabelMedium, LabelSmall} from "@khanacademy/wonder-blocks-typography";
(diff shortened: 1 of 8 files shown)


packages/wonder-blocks-form/src/components/checkbox-group.tsx
```diff
@@ -130,43 +130,44 @@ const CheckboxGroup = React.forwardRef(function CheckboxGroup(
     const allChildren = React.Children.toArray(children).filter(Boolean);
 
     return (
-        <StyledFieldset data-testid={testId} style={styles.fieldset} ref={ref}>
-            {/* We have a View here because fieldset cannot be used with flexbox*/}
-            <View style={style}>
```
jandrade, line 135: This is an old implementation created ~6 years ago. This `flexbox` issue is no longer happening in current browser versions.

beaesguerra (approved):
Looks good, thanks for the fix! 🚀

PR #2329, opened 2024-09-25
Fix plural forms configiration for Khmer locale

mpolyak:
## Summary:
It only has "other", but was configured "like english", resulting in a crash
on km.khanacademy.org for pages with plural form platform strings.

Further details can be seen on Slack: https://khanacademy.slack.com/archives/C3SLM30AW/p1727281143856849

Issue: https://khanacademy.atlassian.net/browse/CP-8661

Test Plan:
 - After making a release of `wonder-blocks-i18n` package and deploying webapp
 - Verify https://km.khanacademy.org/math/early-math/cc-early-math-counting-topic doesn't crash

### DIFF
--- a/.changeset/thick-maps-provide.md
+++ b/.changeset/thick-maps-provide.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-i18n": patch
+---
+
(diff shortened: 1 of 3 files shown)


jeresig (approved):
Great, thanks!

PR #2331, opened 2024-09-30
[fei5850.1.friendlyobjectmatching] Support undefined variable comparison

somewhatabstract:
## Summary:
Before this change, these two variables objects are considered different:

```ts
{
    a: undefined,
    b: 1
}
```

```ts
{
    b: 1
}
```

This can be annoying when writing tests, because you have to explicitly set the variable to `undefined` in the expectation if the actual request ends up fetching with an explicit `undefined` value.

To make GraphQL mocks easier, we now consider a missing key to be equivalent to an explicit key with an `undefined` value. This brings this approach to GraphQL mocking inline with Apollo mocks that already do this.

Note that with this implementation, there's no way to assert that an explicit `undefined` variable is included in the request. We could add that if it turns out to be something we really need though I suspect that it is not.

Issue: FEI-5850

## Test plan:
`yarn test`

### DIFF
--- a/.changeset/old-islands-relax.md
+++ b/.changeset/old-islands-relax.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-testing": major
+---
+
+When mocking GraphQL, consider explicit undefined values in a request to be equivalent to missing keys in a mock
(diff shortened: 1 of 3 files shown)


jeresig (approved):
Ah interesting, ok!

PR #2332, opened 2024-10-01
IconButton: Add id prop, fix type to set 'button' as default

jandrade:
## Summary:

Following up on #2325, this PR fixes an issue that I introduced in the
IconButton component, which was that the type prop was not being set to 'button'
by default.

This PR fixes that issue and also adds an optional `id` prop to the
IconButton component.

NOTE: We set `button` as the default type because it is the most common use case
of how browsers interpret the type attribute of a button element. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#type

Issue: XXX-XXXX

## Test plan:

Verify that the IconButton component works as expected with the new changes.

Also verify that that the tests pass.

/iframe.html?args=id:some-identifier&id=packages-iconbutton--default&viewMode=story

Screenshot 2024-10-01 at 11 50 01 AM

### DIFF
--- a/.changeset/friendly-moose-type.md
+++ b/.changeset/friendly-moose-type.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-icon-button": minor
+---
+
+Adds `id` prop and fixes `type` prop to set 'button' as default value.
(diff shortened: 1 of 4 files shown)


beaesguerra (approved):
Looks good, thanks for the fix!

PR #2334, opened 2024-10-01
[fei5850.2.hardfail] Add hard fail support to request mocking

somewhatabstract:
## Summary:
An oft annoying but sometimes useful feature of Apollo's `MockLink` is that it will hard fail if a request is made that has not been mocked. This is annoying because it forces folks to provide a mock for every request, even if they don't care about that request. However, it is useful when debugging tests to see why a mock isn't matching as expected or what requests are happening (the KA_LOG_FETCHES is useful for seeing this info too, but less accessible).

This change adds the ability to turn on hard fails for our request mocking system. I have added this using a fluent API so that it is more easily worked into existing uses of this mocking system.

I've also added tests to verify the new functionality, and added some additional commentary to the types for the mocking framework.

Issue: FEI-5850

## Test plan:
`yarn test`
`yarn typecheck`

### DIFF
--- a/.changeset/great-paws-ring.md
+++ b/.changeset/great-paws-ring.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/wonder-blocks-testing-core": minor
+"@khanacademy/wonder-blocks-testing": minor
+---
+
+Add support for hard fails to the request mocking features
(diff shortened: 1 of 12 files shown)


jeresig (approved):
Great!

PR #2335, opened 2024-10-01
[fei5850.3.creator] Add matchGql function

somewhatabstract:
## Summary:
I added a signature change to webapp to make it easier to do simple mocks of GraphQL requests for a query without specifying variables, etc. However, I noticed that it made type-checking much less helpful. It obfuscated the real issue (such as a bad variable type) by raising a signature mismatch problem. This is not going to help devs find it easier to write GraphQL mocks.

So, I want to back that change out over in webapp and instead, have added this new helper to build a mock. I originally had this be a simple 3 argument function, like this:

```ts
const matchGql = <TData, TVariables, TContext>(query: GqlOperation<TData, TVariables>, variables?: TVariables, context: TContext) => {
  // ...
}
```

However, I noticed a weird issue where if there was a type issue in both the context and variables arguments, only the variables one caused an error and suppressing that error did not cause the context one to be raised at all. This too felt like it was not helpful at all.

So, instead, I have changed to this fluent API style whereby the returned value is a valid `GqlMockOperation` to pass to the `mockOperation` and `mockOperationOnce` calls, but it also has methods to set the variable and context matching.

I have added type tests to verify that we now get helpful errors when the values passed are incorrect.

This means these two things should be equivalent, for example:

```ts
mockGqlFetch.mockOperation({operation: query, variables, context}, RespondWith.graphQLData(data));

mockGqlFetch.mockOperation(matchGql(query, variables, context), RespondWith.graphQLData(data));
```

The type checking feedback for the `matchGql` variant will be much easier to understand and resolve. However, this still isn't enough. Because the query is now inside the `matchGql` call, Typescript doesn't appear to know how to infer the type of `TData` for the query and ensure parity between the `matchGql` call and the `RespondWith.graphQLData` call - so this again becomes hard to manage for devs. I'm working on this - for now...this is a proof-of-concept but not the whole story.

Issue: FEI-5850

## Test plan:
`yarn test`
`yarn typecheck`

### DIFF
--- a/.changeset/great-paws-ring.md
+++ b/.changeset/great-paws-ring.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/wonder-blocks-testing-core": minor
+"@khanacademy/wonder-blocks-testing": minor
+---
+
+Add support for hard fails to the request mocking features
--- a/.changeset/old-islands-relax.md
+++ b/.changeset/old-islands-relax.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-testing": major
+---
+
+When mocking GraphQL, consider explicit undefined values in a request to be equivalent to missing keys in a mock
--- a/.changeset/ten-horses-mate.md
+++ b/.changeset/ten-horses-mate.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-testing": minor
+---
+
+Add matchGql function to facilitate type-safe construction of mocked GraphQL requests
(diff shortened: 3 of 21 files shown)


jeresig (commented):
Interesting - you mention this is a proof of concept, are you looking for a review?

somewhatabstract:
> Interesting - you mention this is a proof of concept, are you looking for a review?

Sorry, it was late and I needed to eat. Hold off on review. I learned why things weren't working right last night and I'll be reworking this today based on what I learned (I may just close this PR).

somewhatabstract:
OK, while this might be useful for anyone using Wonder Blocks Data directly - webapp wouldn't use this due to it's need to use a `GqlOperation` type that is compatible with Apollo. If we drop Apollo and use WBD only, then things get simpler, but for now, I'm going to just not add code here that we won't use.

PR #2339, opened 2024-10-04
LabeledField(part3): Wire up attributes

beaesguerra:
## Summary:
- Use an auto-generated unique id if the `id` prop is not provided
- Wire up elements within the component
  - the `label` element's `for` attribute is the field `id`
  -  `aria-describedby` on the field has the ids of the description and/or error message elements
- Apply attributes to the field element:
  - `id`
  - `aria-describedby`
  - `required` if the `required` prop is set to `true`
  - `testId`

Issue: WB-1503

## Test plan:
- Tests are passing
- Updated documentation makes sense:
  - `?path=/docs/packages-labeledfield-labeledfield--docs`
  - `?path=/docs/packages-labeledfield-labeledfield-accessibility--docs`
- There are no a11y warnings in the Storybook Accessibility add on for the LabeledField stories
  - Note: There are warnings for the LabeledField stories with the different field examples. They are related to an issue with the SearchField component and will be addressed in WB-1771!

### DIFF
--- a/.changeset/pink-turtles-sleep.md
+++ b/.changeset/pink-turtles-sleep.md
@@ -1,5 +1,5 @@
 ---
-"@khanacademy/wonder-blocks-labeled-field": patch
+"@khanacademy/wonder-blocks-labeled-field": minor
 ---
 
 LabeledField: Wire up attributes for elements and apply attributes to the field element
--- a/.changeset/pink-turtles-sleep.md
+++ b/.changeset/pink-turtles-sleep.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-labeled-field": patch
+---
+
+LabeledField: Wire up attributes for elements and apply attributes to the field element
--- a/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
+++ b/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
@@ -22,7 +22,7 @@ import SearchField from "@khanacademy/wonder-blocks-search-field";
  * `LabeledField` component so that our form fields are consistent and accessible.
  */
 export default {
-    title: "Packages / LabeledField / LabeledField",
+    title: "Packages / LabeledField",
     component: LabeledField,
     parameters: {
         componentSubtitle: (
--- a/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
+++ b/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
@@ -174,7 +174,7 @@ export const Light: StoryComponentType = {
  * The following story shows what the LabeledField looks like when different
  * props are set.
  */
-export const Combinations = (args: PropsFor<typeof LabeledField>) => {
+export const Scenarios = (args: PropsFor<typeof LabeledField>) => {
     const [textFieldValue, setTextFieldValue] = React.useState("");
 
     return (
--- a/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
+++ b/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
@@ -169,3 +169,49 @@ export const Light: StoryComponentType = {
     },
     render: AllFields,
 };
+
+/**
+ * The following story shows what the LabeledField looks like when different
+ * props are set.
+ */
+export const Combinations = (args: PropsFor<typeof LabeledField>) => {
+    const [textFieldValue, setTextFieldValue] = React.useState("");
+
+    return (
+        <View style={{gap: spacing.large_24}}>
+            <LabeledField
+                {...args}
+                label="Label only"
+                field={
+                    <TextField
+                        value={textFieldValue}
+                        onChange={setTextFieldValue}
+                    />
+                }
+            />
+            <LabeledField
+                {...args}
+                label="With description"
+                description="Description"
+                field={
+                    <TextField
+                        value={textFieldValue}
+                        onChange={setTextFieldValue}
+                    />
+                }
+            />
+            <LabeledField
+                {...args}
+                label="With Error"
+                error="Error message"
+                field={
+                    <TextField
+                        value="invalid value"
+                        onChange={() => {}}
+                        validate={() => "Error message"}
+                    />
+                }
+            />
+        </View>
+    );
+};
--- a/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
+++ b/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
@@ -202,8 +202,21 @@ export const Combinations = (args: PropsFor<typeof LabeledField>) => {
             />
             <LabeledField
                 {...args}
-                label="With Error"
+                label="With error"
                 error="Error message"
+                field={
+                    <TextField
+                        value="invalid value"
+                        onChange={() => {}}
+                        validate={() => "Error message"}
+                    />
+                }
+            />
+            <LabeledField
+                {...args}
+                label="With description and error"
+                error="Error message"
+                description="Description"
                 field={
                     <TextField
                         value="invalid value"
--- a/packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
+++ b/packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
@@ -368,6 +368,8 @@ describe("LabeledField", () => {
                     <LabeledField
                         field={<TextField value="" onChange={() => {}} />}
                         label="Label"
+                        description="Description for the field"
+                        error="Error message"
                     />,
                     defaultOptions,
                 );
--- a/packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
+++ b/packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
@@ -240,8 +240,124 @@ describe("LabeledField", () => {
     });
 
     describe("Attributes", () => {
-        it("should autogenerate ids for the elements if the id prop is not provided", () => {});
-        it("should use the id prop for element ids", () => {});
+        it("should use the id prop for the base of element ids", () => {
+            // Arrange
+            const id = "example-id";
+            const label = "Label";
+            const description = "Description of the field";
+            const error = "Error message";
+            render(
+                <LabeledField
+                    field={<TextField value="" onChange={() => {}} />}
+                    id={id}
+                    label={label}
+                    description={description}
+                    error={error}
+                />,
+                defaultOptions,
+            );
+
+            // Act
+            const labelEl = screen.getByText(label);
+            const descriptionEl = screen.getByText(description);
+            const fieldEl = screen.getByRole("textbox");
+            const errorEl = screen.getByText(error);
+
+            // Assert
+            expect(labelEl.id).toBe(`${id}-label`);
+            expect(descriptionEl.id).toBe(`${id}-description`);
+            expect(fieldEl.id).toBe(`${id}-field`);
+            expect(errorEl.id).toBe(`${id}-error`);
+        });
+
+        it("should autogenerate ids for the elements if the id prop is not provided", () => {
+            // Arrange
+            const label = "Label";
+            const description = "Description of the field";
+            const error = "Error message";
+            render(
+                <LabeledField
+                    field={<TextField value="" onChange={() => {}} />}
+                    label={label}
+                    description={description}
+                    error={error}
+                />,
+                defaultOptions,
+            );
+
+            // Act
+            const labelEl = screen.getByText(label);
+            const descriptionEl = screen.getByText(description);
+            const fieldEl = screen.getByRole("textbox");
+            const errorEl = screen.getByText(error);
+
+            // Assert
+            expect(labelEl.id).toInclude("-label");
+            expect(descriptionEl.id).toInclude("-description");
+            expect(fieldEl.id).toInclude("-field");
+            expect(errorEl.id).toInclude("-error");
+        });
+
+        it("should use the testId prop for the base of test ids", () => {
+            // Arrange
+            const testId = "test-id";
+            const label = "Label";
+            const description = "Description of the field";
+            const error = "Error message";
+            render(
+                <LabeledField
+                    field={<TextField value="" onChange={() => {}} />}
+                    testId={testId}
+                    label={label}
+                    description={description}
+                    error={error}
+                />,
+                defaultOptions,
+            );
+
+            // Act
+            const labelEl = screen.getByText(label);
+            const descriptionEl = screen.getByText(description);
+            const fieldEl = screen.getByRole("textbox");
+            const errorEl = screen.getByText(error);
+
+            // Assert
+            expect(labelEl).toHaveAttribute("data-testid", `${testId}-label`);
+            expect(descriptionEl).toHaveAttribute(
+                "data-testid",
+                `${testId}-description`,
+            );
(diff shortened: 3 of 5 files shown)


packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -10,7 +15,7 @@ type Props = {
     /**
      * The form field component.
      */
-    field: React.ReactNode;
+    field: React.ReactElement;
```
beaesguerra, line 18: Needed to update the `field` prop to be the `ReactElement` type in order to use `React.cloneElement` to apply attributes from this component. I think this makes sense for the `field` prop so that it can't be `boolean | null | undefined` which `ReactNode` allows.

Let me know if there are any concerns with this!

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
```diff
@@ -6,8 +6,13 @@ import {I18nInlineMarkup} from "@khanacademy/wonder-blocks-i18n";
 import {Body} from "@khanacademy/wonder-blocks-typography";
 
 import {TextField} from "@khanacademy/wonder-blocks-form";
+import {RenderStateRoot} from "@khanacademy/wonder-blocks-core";
 import LabeledField from "../labeled-field";
 
+const defaultOptions = {
+    wrapper: RenderStateRoot,
```
beaesguerra, line 13: This is needed for tests because we use `useUniqueIdWithMock` to generate an id if one isn't provided

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
```diff
@@ -215,11 +230,240 @@ describe("LabeledField", () => {
...
+
+                // Act
+                const labelEl = screen.getByText(label);
+                const inputEl = screen.getByRole("textbox");
+
+                // Assert
+                expect(labelEl).toHaveAttribute("for", inputEl.id);
+            });
+
+            it("should set aria-describedby on the field to the id of the description", () => {
```
marcysutton, line 425: Is there an example of this usage in Storybook? I saw the error example, but wasn't sure if this description variant was shown. It looks solid implementation-wise, though!

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -132,12 +160,23 @@ export default function LabeledField(props: Props) {
         );
     }
 
+    function renderField() {
+        return React.cloneElement(field, {
+            id: fieldId,
+            "aria-describedby": [description && descriptionId, error && errorId]
+                .filter(Boolean)
```
marcysutton, line 167: TIL you can do this!

marcysutton (commented):
This is looking great! I left one question about Storybook, but that is all.

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
beaesguerra: Great question! All of the LabeledField stories had a description configured so we are able to inspect the `aria-describedby` relationship with the description! 

Your question made me realize that it would be helpful to have a story (and visual regression tests) for the different combinations of props since right now, all of the LabeledField stories show when both the label and description are present. I've added a new [Combinations story](https://5e1bf4b385e3fb0020b7073c-rnlzivmykl.chromatic.com/?path=/story/packages-labeledfield--combinations)!

__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
```diff
@@ -22,7 +22,7 @@ import SearchField from "@khanacademy/wonder-blocks-search-field";
  * `LabeledField` component so that our form fields are consistent and accessible.
  */
 export default {
-    title: "Packages / LabeledField / LabeledField",
+    title: "Packages / LabeledField",
```
beaesguerra, line 25: I removed the additional nesting of folders in Storybook!

.changeset/pink-turtles-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-labeled-field": patch
```
jandrade, line 2: **nit:** This would be a `minor` version, but I don't think it matters if you are planning to add more LabeledField PRs as part of this initial release.

__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
```diff
@@ -169,3 +169,49 @@ export const Light: StoryComponentType = {
...
+                label="With Error"
+                error="Error message"
+                field={
+                    <TextField
+                        value="invalid value"
+                        onChange={() => {}}
+                        validate={() => "Error message"}
+                    />
+                }
+            />
```
jandrade, line 214: **thought:** How do you feel about adding a `With Description + Error` example here? I understand that this is covered by previous stories, but maybe it would be good to include it here so folks are aware of all possible combinations when they navigate to this example? (no change required).

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
```diff
@@ -215,11 +230,240 @@ describe("LabeledField", () => {
...
+                    error={error}
+                />,
+                defaultOptions,
+            );
+
+            // Act
+            const labelEl = screen.getByText(label);
+            const descriptionEl = screen.getByText(description);
+            const fieldEl = screen.getByRole("textbox");
+            const errorEl = screen.getByText(error);
```
jandrade, line 264: **nit:** this works well, but I wonder if we could use the `byRole` fn here, so we can ensure that we are using the correct element for referencing the error? (maybe this is covered in a different test, but wanted to see your thoughts on this).

```suggestion
            const errorEl = screen.getByRole("alert");
```

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
```diff
@@ -215,11 +230,240 @@ describe("LabeledField", () => {
...
+
+            // Assert
+            expect(labelEl).toHaveAttribute("data-testid", `${testId}-label`);
+            expect(descriptionEl).toHaveAttribute(
+                "data-testid",
+                `${testId}-description`,
+            );
+            expect(fieldEl).toHaveAttribute("data-testid", `${testId}-field`);
+            expect(errorEl).toHaveAttribute("data-testid", `${testId}-error`);
```
jandrade, line 331: **suggestion:** This looks to me that we are testing multiple assertions within the same test. This makes a bit harder to understand when a specific assertion is failing. I think we should try using `it.each` [^1] to test every element within the component, instead of checking for multiple things in these tests.

Note: This comment would applly to all the tests that are asserting multiple things in this file.

[^1]: https://jestjs.io/docs/api#testeachtablename-fn-timeout

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
```diff
@@ -215,11 +230,240 @@ describe("LabeledField", () => {
...
+
+            // Act
+            const fieldEl = screen.getByRole("textbox");
+
+            // Assert
+            expect(fieldEl).toHaveAttribute("name", "name-example");
+        });
+    });
+
+    describe("Accessibility", () => {
```
jandrade, line 386: **question:** Follow up question to one of my prev comments. Is there a test that verifies that the error is logged correctly by checking the element with `role="alert"`? I think this would be helpful in the future to prevent any possible regressions (disregard this comment if there's already a test verifying that).

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -32,14 +37,28 @@ type Props = {
...
      *
-     * The label will assume that the field will have its id formatted as `${id}-field`.
-     * The field can assume that the error will have its id formatted as `${id}-error`.
+     * If the `id` prop is not provided, a base unique id will be auto-generated.
+     * This is important so that the different elements can be wired up together
+     * for accessibility!
+     *
+     * Note: When using the `LabeledField` component, an `id` provided to the
+     * field component (ex: a TextField component) will be overridden.
      */
```
jandrade, line 53: **praise:** Love these comments!

jandrade (commented):
This looks really solid! I added some comments around tests + possibly modifying a story. Let me know what you think :)

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
beaesguerra: There are some tests checking for the alert! I will also make sure we have enough test coverage around errors when I re-visit error validation!

That said, I'm not sure yet if we should be using `role="alert"` for the error messages, depending on how we want to handle error validation. Since there are cases for submitting a form and getting multiple errors, would it be too much to have all the error messages as alerts? I will be bringing up error validation in our cubby meeting later to get some thoughts on this and we can discuss more!

marcysutton: I think whether to use `role="alert"` depends if you're sending focus to the first invalid field. An alert announcement and a focused element announcement would compete with each other, so it's kind of one or the other.

.changeset/pink-turtles-sleep.md
beaesguerra: Updated it anyways to be concise! Good to know we want to use a minor update for setting attributes!

Side note - I've been adding changeset files to the intermediate PRs (these partX PRs are being merged into a `feature/labeled-field` branch). We only need one changeset when we merge to main, however, our pull requests check for a changeset file. I think it should be fine though that there will be multiple changeset files since it should result in just one version number change and consolidate the messages. Let me know what you think!

__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
beaesguerra: Good point, I've added it in! 

I will also add more examples to this story so I renamed it to "Scenarios". I'm going to add examples for handling overflowing text when I work on polishing the styling in another PR!

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
beaesguerra: Updated, though it may change based on the [error message implementation thread](https://github.com/Khan/wonder-blocks/pull/2339#discussion_r1792093660)

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
beaesguerra: Updated in https://github.com/Khan/wonder-blocks/pull/2339/commits/eaf4aa1edf3c68bf82a4ba6977f32b51ea12f48e! Let me know what you think :)

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
beaesguerra: That makes sense! Thinking aloud, we may need to provide a prop that let's the devs decide the behaviour:
- If it's client side validation and the error message is shown onBlur, it would make sense for the error message to have `role="alert"` so it's announced right away
- If the error message is shown after form submission (usually from backend validation), we would focus on the first field with error message (and the error message doesn't have `role="alert"`

But what about when there's both client side validation and backend validation? 🤔 

I tested to see what happens in VoiceOver when multiple error messages with `role="alert"` are shown in the [validation example](https://5e1bf4b385e3fb0020b7073c-cmzmctcwqt.chromatic.com/?path=/story/packages-labeledfield--validation). It only announced the last error message! But maybe that's okay since a user would then fill out that field and get feedback again on a previous error?

I will look into this more when I work on the validation part, but adding some more thoughts here for now. Let me know if anyone has more ideas or alternatives!

jandrade: All of this makes sense, thanks for the context!

As an option, I was thinking that maybe it would be worth experimenting with setting `role="status"` by default (which is `polite` + `aria-atomic="true"`), and allowing to override to `alert` when needed.

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
jandrade: Looks awesome! Neat approach for the `getElement` function.

.changeset/pink-turtles-sleep.md
jandrade: That's true that we only need one changeset for main, but I feel that adding a changeset per PR helps to understand better the scope of the change for that PR. Agreed about having one version number + consolidated msgs. I guess using `minor` here is me being nitpicky 😅, but I guess I feel it's good to keep versions semantic just for keeping consistency with our PRs.

Maybe useful if we are looking back in history to look for specific changes, although I'm not sure if this PR will get squashed an erased as soon as we land the intermediate branch (`feature/labeled-field`) into `main`.

jandrade (approved):
This is looking very strong!, great progress on the `LabeledField` component 🚀

.changeset/pink-turtles-sleep.md
beaesguerra: Using `minor` here is a good call! I'm realizing now that none of the [other PRs I merged already](https://github.com/Khan/wonder-blocks/compare/main...feature/labeled-field) specified a minor change so at least there will be this one!

My plan is to land `feature/labeled-field` into `main` using `git land --feature-branch` so the PR merge commits don't get squashed (based on the [Intermediate Integration Branch example in Confluence](https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/2918416504/Understand+our+Branching+Strategy#Example-2---Intermediate-Integration-Branch))

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
marcysutton: Good thinking and research here @beaesguerra! It's challenging that screen reader behavior is so inconsistent cross-platform. 

A few more thoughts/questions that come to mind after chatting about it:

- Exposing an `alert` client-side onBlur makes a ton of sense. For some reason [this example](https://www.w3.org/WAI/tutorials/forms/notifications/#on-focus-change) doesn't work in Safari and VO but does work in iOS Safari and VO.
    - Maybe this technique could be improved with the forthcoming Announcer component?
- A summary of error messages would be helpful in the event there are multiple errors. Ideally it would be visible, but at the very least something visually hidden could be announced to inform screen reader users. Would that cover the case of client-side validation and backend validation (were you meaning at the same time)? 
- Recipes and examples that we can test thoroughly will be helpful in deciding what to do! I'll be happy to help test when needed.

beaesguerra: Thanks for summarizing Marcy! We can definitely iterate on this to make it more consistent cross-platform :) 

> Would that cover the case of client-side validation and backend validation (were you meaning at the same time)

Yes! That would address both types of validation 👍 

> Recipes and examples that we can test thoroughly will be helpful in deciding what to do! I'll be happy to help test when needed.

I'll be working on some examples next and will appreciate the feedback!

PR #2340, opened 2024-10-04
LabeledField - WIP validation

beaesguerra:
## Summary:
validation example

Issue: XXX-XXXX

## Test plan:

### DIFF
--- a/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
+++ b/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
@@ -22,7 +22,7 @@ import SearchField from "@khanacademy/wonder-blocks-search-field";
  * `LabeledField` component so that our form fields are consistent and accessible.
  */
 export default {
-    title: "Packages / LabeledField / LabeledField",
+    title: "Packages / LabeledField",
     component: LabeledField,
     parameters: {
         componentSubtitle: (
--- a/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
+++ b/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
@@ -169,3 +169,49 @@ export const Light: StoryComponentType = {
     },
     render: AllFields,
 };
+
+/**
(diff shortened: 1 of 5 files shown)


__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
```diff
@@ -169,3 +170,140 @@ export const Light: StoryComponentType = {
...
+                                    value="strawberry"
+                                />
+                                <OptionItem label="Banana" value="banana" />
+                            </MultiSelect>
+                        }
+                    />
+                    <Button
+                        type="submit"
+                        onClick={() => {
+                            // Example where everything has an error
```
beaesguerra, line 290: error message for textarea and textfield will be shown because of LabeledField, however, error styling is not applied to the field element because they don't currently support an error prop (just built-in validation)

__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
```diff
@@ -169,3 +170,140 @@ export const Light: StoryComponentType = {
...
+                                    setSingleSelectValue(value);
+                                    if (value === "strawberry") {
+                                        setSingleSelectError(
+                                            "Don't select strawberry.",
+                                        );
+                                    } else {
+                                        setSingleSelectError("");
+                                    }
+                                }}
+                                error={!!singleSelectError} // should not need to do this! LabeledField should set this
```
beaesguerra, line 247: LabeledField can inject the `error` prop based on if `LabeledField` receives an error message! However, this will only work if the field supports an `error` prop (so it wouldn't currently work with the TextArea/TextField components that only support the `validate`/`onValidate` props)

beaesguerra:
The parent pull-request ([#2339](https://github.com/Khan/wonder-blocks/pull/2339)) has been merged into `feature/labeled-field`, but this branch (`labeled-field-validation`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #2344, opened 2024-10-10
LabeledField(part4): styling and error icon

beaesguerra:
## Summary:
use semantic colors where possible. light styles don't have corresponding semantic tokens
Handle long text with no word break
visually align icon with text
Add long text scenario story
Add  prop for error icon aria-label
Add error icon to error message

Issue: WB-1503

## Test plan:

### DIFF
--- a/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
+++ b/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
@@ -168,7 +168,10 @@ export default function LabeledField(props: Props) {
                 <View style={styles.errorSection}>
                     <PhosphorIcon
                         icon={WarningCircle}
-                        style={light ? styles.lightError : styles.error}
+                        style={[
+                            styles.errorIcon,
+                            light ? styles.lightError : styles.error,
+                        ]}
                         role="img"
                         aria-label={labels.errorIconAriaLabel}
                     />
@@ -229,6 +232,9 @@ const styles = StyleSheet.create({
     },
     lightError: {
         color: color.fadedRed,
+    },
+    errorIcon: {
+        marginTop: "1px", // This vertically aligns the icon with the text
     },
     required: {
         color: color.red,
--- a/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
+++ b/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
@@ -1,5 +1,6 @@
 import * as React from "react";
 import {StyleSheet} from "aphrodite";
+import WarningCircle from "@phosphor-icons/core/bold/warning-circle-bold.svg";
 
 import {
     View,
@@ -10,6 +11,7 @@ import {
 import {Strut} from "@khanacademy/wonder-blocks-layout";
 import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
 import {LabelMedium, LabelSmall} from "@khanacademy/wonder-blocks-typography";
+import {PhosphorIcon} from "@khanacademy/wonder-blocks-icon";
 
 type Props = {
     /**
@@ -148,14 +150,22 @@ export default function LabeledField(props: Props) {
         return (
             <React.Fragment>
                 <Strut size={spacing.small_12} />
-                <LabelSmall
-                    style={light ? styles.lightError : styles.error}
-                    role="alert"
-                    id={errorId}
-                    testId={testId && `${testId}-error`}
-                >
-                    {error}
-                </LabelSmall>
+                <View style={styles.errorSection}>
+                    <PhosphorIcon
+                        icon={WarningCircle}
+                        style={light ? styles.lightError : styles.error}
+                        role="img"
+                        aria-label="Error"
+                    />
+                    <LabelSmall
+                        style={light ? styles.lightError : styles.error}
+                        role="alert"
+                        id={errorId}
+                        testId={testId && `${testId}-error`}
+                    >
+                        {error}
+                    </LabelSmall>
+                </View>
             </React.Fragment>
         );
     }
@@ -194,6 +204,10 @@ const styles = StyleSheet.create({
     },
     lightDescription: {
         color: color.white64,
+    },
+    errorSection: {
+        flexDirection: "row",
+        gap: spacing.xSmall_8,
     },
     error: {
         color: color.red,
--- a/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
+++ b/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
@@ -67,6 +67,20 @@ type Props = {
      * Change the field’s sub-components to fit a dark background.
      */
     light?: boolean;
+    /**
+     * The object containing the custom labels used inside this component.
+     *
+     * This is useful for internationalization. Defaults to English.
+     */
+    labels?: LabeledFieldLabels;
+};
+
+type LabeledFieldLabels = {
+    errorIconAriaLabel: string;
+};
+
+const defaultLabeledFieldLabels: LabeledFieldLabels = {
+    errorIconAriaLabel: "Error",
 };
 
 const StyledSpan = addStyle("span");
@@ -86,6 +100,7 @@ export default function LabeledField(props: Props) {
         light,
         description,
         error,
+        labels = defaultLabeledFieldLabels,
     } = props;
 
     const ids = useUniqueIdWithMock("labeled-field");
@@ -155,7 +170,7 @@ export default function LabeledField(props: Props) {
                         icon={WarningCircle}
                         style={light ? styles.lightError : styles.error}
                         role="img"
-                        aria-label="Error"
+                        aria-label={labels.errorIconAriaLabel}
                     />
                     <LabelSmall
                         style={light ? styles.lightError : styles.error}
--- a/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
+++ b/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
@@ -113,7 +113,10 @@ export default function LabeledField(props: Props) {
     function renderLabel(): React.ReactNode {
         const requiredIcon = (
             <StyledSpan
-                style={light ? styles.lightRequired : styles.required}
+                style={[
+                    styles.textWordBreak,
+                    light ? styles.lightRequired : styles.required,
+                ]}
                 aria-hidden={true}
             >
                 {" "}
@@ -124,7 +127,10 @@ export default function LabeledField(props: Props) {
         return (
             <React.Fragment>
                 <LabelMedium
-                    style={light ? styles.lightLabel : styles.label}
+                    style={[
+                        styles.textWordBreak,
+                        light ? styles.lightLabel : styles.label,
+                    ]}
                     tag="label"
                     htmlFor={fieldId}
                     testId={testId && `${testId}-label`}
@@ -146,7 +152,10 @@ export default function LabeledField(props: Props) {
         return (
             <React.Fragment>
                 <LabelSmall
-                    style={light ? styles.lightDescription : styles.description}
+                    style={[
+                        styles.textWordBreak,
+                        light ? styles.lightDescription : styles.description,
+                    ]}
                     testId={testId && `${testId}-description`}
                     id={descriptionId}
                 >
@@ -176,7 +185,11 @@ export default function LabeledField(props: Props) {
                         aria-label={labels.errorIconAriaLabel}
                     />
                     <LabelSmall
-                        style={light ? styles.lightError : styles.error}
+                        style={[
+                            styles.textWordBreak,
+                            styles.errorMessage,
+                            light ? styles.lightError : styles.error,
+                        ]}
                         role="alert"
                         id={errorId}
                         testId={testId && `${testId}-error`}
@@ -236,10 +249,16 @@ const styles = StyleSheet.create({
     errorIcon: {
         marginTop: "1px", // This vertically aligns the icon with the text
     },
+    errorMessage: {
+        minWidth: "0",
+    },
     required: {
         color: semanticColor.status.critical.foreground,
     },
     lightRequired: {
         color: color.fadedRed,
+    },
+    textWordBreak: {
+        overflowWrap: "break-word",
     },
 });
--- a/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
+++ b/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
@@ -9,7 +9,7 @@ import {
     useUniqueIdWithMock,
 } from "@khanacademy/wonder-blocks-core";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
-import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {color, semanticColor, spacing} from "@khanacademy/wonder-blocks-tokens";
 import {LabelMedium, LabelSmall} from "@khanacademy/wonder-blocks-typography";
 import {PhosphorIcon} from "@khanacademy/wonder-blocks-icon";
 
@@ -212,13 +212,13 @@ export default function LabeledField(props: Props) {
 
 const styles = StyleSheet.create({
     label: {
-        color: color.offBlack,
+        color: semanticColor.text.primary,
     },
     lightLabel: {
-        color: color.white,
+        color: semanticColor.text.inverse,
     },
     description: {
-        color: color.offBlack64,
+        color: semanticColor.text.secondary,
     },
     lightDescription: {
         color: color.white64,
@@ -228,7 +228,7 @@ const styles = StyleSheet.create({
         gap: spacing.xSmall_8,
     },
     error: {
-        color: color.red,
+        color: semanticColor.status.critical.foreground,
     },
     lightError: {
         color: color.fadedRed,
@@ -237,7 +237,7 @@ const styles = StyleSheet.create({
         marginTop: "1px", // This vertically aligns the icon with the text
     },
     required: {
-        color: color.red,
+        color: semanticColor.status.critical.foreground,
     },
     lightRequired: {
         color: color.fadedRed,
--- a/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
+++ b/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
@@ -13,6 +13,11 @@ import {
     SingleSelect,
 } from "@khanacademy/wonder-blocks-dropdown";
 import SearchField from "@khanacademy/wonder-blocks-search-field";
+import {
+    HeadingLarge,
+    HeadingMedium,
+    HeadingSmall,
+} from "@khanacademy/wonder-blocks-typography";
 
 /**
  * The `LabeledField` component provides common elements for a form field such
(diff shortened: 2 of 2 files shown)


packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -184,27 +225,40 @@ export default function LabeledField(props: Props) {
...
     lightLabel: {
-        color: color.white,
+        color: semanticColor.text.inverse,
     },
     description: {
-        color: color.offBlack64,
+        color: semanticColor.text.secondary,
     },
     lightDescription: {
         color: color.white64,
```
beaesguerra, line 237: We don't have semantic color tokens yet for colors used on dark backgrounds so the original color tokens are still used for the light styling (except for the label since we were able to use the text inverse token)!

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -184,27 +225,40 @@ export default function LabeledField(props: Props) {
...
+    },
     required: {
-        color: color.red,
+        color: semanticColor.status.critical.foreground,
     },
     lightRequired: {
         color: color.fadedRed,
     },
+    textWordBreak: {
+        overflowWrap: "break-word",
```
beaesguerra, line 262: I added the `overflowWrap` style to all WB Label* components we use in LabeledField so that overflowing text can be wrapped!

I was wondering, should the WB typography components all handle overflowing text? I was thinking we could add it to all typography components, but that would impact a lot of places. Or maybe not everything should wrap the same way? Or some things might need truncation or an ellipsis. Let me know if anyone has any thoughts on this!

marcysutton: Is there a concept of utility classes we could use to make this kind of style more portable? That could add flexibility while keeping things consistent (and requiring less CSS repetition).

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -148,14 +174,29 @@ export default function LabeledField(props: Props) {
...
+                        aria-label={labels.errorIconAriaLabel}
+                    />
+                    <LabelSmall
+                        style={[
+                            styles.textWordBreak,
+                            styles.errorMessage,
+                            light ? styles.lightError : styles.error,
+                        ]}
+                        role="alert"
+                        id={errorId}
```
beaesguerra, line 194: The `aria-describedby` on the field uses this id for the error message. I'm going to update the PR so that it points to the whole error section so that the aria-label on the error icon is included!

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
beaesguerra: We could create a shared stylesheet like what we have for [typography styles](https://github.com/Khan/wonder-blocks/blob/b8f0fcf91e04bfe8be931b739cea49c6d846d2aa/packages/wonder-blocks-typography/src/util/styles.ts#L20).  Not sure which wonder-blocks package it would go in, maybe `core`? Or a new package called something like `wonder-blocks-styles`? 

I would also like to use something like utility classes or mixins, especially around common state styles (disabled state, hover, focus, etc) and other styles like the `sr-only` styling (related issue: WB-1753) 

I think @jandrade was also considering the concept of utility classes in the CSS Framework investigation!

marcysutton: Ooh, yes! That sounds awesome. Either `core` or a new package would make sense to me! I'll be curious to hear @jandrade's thoughts on utility classes, too.

jandrade: I like the idea of `wonder-blocks-styles` a lot :). Specially that a lot of places in `webapp` could benefit from this as well.

`core` could work too, but I feel that, as most of the packages depend on it, it would mean that for every change in the style utils, then all the packages would need to be bumped, and sometimes doesn't make sense due to the current API surface of that package.... as core currently does all sorts of things (e.g. HOCs, hooks, components, styling utils). I've thinking for a while to split `core` into smaller parts, and maybe this could be a good opportunity for moving some of that complexity to the right packages (like `addStyle` could be moved to `wb-styles`, for example).

As for utility classes as a concept, I think we probably don't need to go at the atomic level for now (like `tailwind`), but composed utility classes could work for the current state of Wonder Blocks. Generic things like this example, + other things like form field styles (@beaesguerra suggested this idea a while ago!), etc, could be useful for avoiding repetition.

As for atomic CSS, I've investigated this methodology as a possible approach to take for our CSS styling in the future. You can check `PandaCSS` in this document I wrote some time ago: https://khanacademy.atlassian.net/wiki/spaces/FRONTEND/pages/2966650967/Experimenting+with+CSS+Modules+and+Panda+CSS

Happy to share more details and discuss about it at any moment.

marcysutton: Awesome. `wb-styles` and decoupling from core makes a lot of sense! @beaesguerra what do you to implement for this PR to be considered done? Definitely don't want to bikeshed your work on error icons for broader wish list items.

beaesguerra: Created WB-1776 for the wonder-blocks-style package! I won't include as part of the LabeledField work. Feel free to add any more details or ideas for things we'd like to include!

For this PR, I am still looking into the aria-describedby issue for the error message (from the previous [comment](https://github.com/Khan/wonder-blocks/pull/2344#discussion_r1796197516)). It is proving to be less straightforward than I thought because [VoiceOver doesn't announce aria-describedby info on elements with role="alert"](https://discussions.apple.com/thread/255163557?sortBy=rank) 🙃 so I am investigating alternatives

I'm thinking I'll leave this PR as is to focus on styling and then make updates in the following validation PR to address a11y issues with the errors

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
```diff
@@ -126,49 +145,6 @@ describe("LabeledField", () => {
         expect(error).toBeInTheDocument();
     });
 
-    it("LabeledField adds the correctly formatted id to label's htmlFor", () => {
```
beaesguerra, line 129: Removed since this is a duplicate of this test: https://github.com/Khan/wonder-blocks/blob/2b809b7d46afbf8bfab02559afb5778ad3723992/packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx#L436-L453

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
```diff
@@ -126,49 +145,6 @@ describe("LabeledField", () => {
...
-            />,
-            defaultOptions,
-        );
-
-        // Assert
-        const label = screen.getByTestId(`${testId}-label`);
-        expect(label).toHaveAttribute("for", `${id}-field`);
-    });
-
-    it("LabeledField adds the correctly formatted id to error's id", () => {
```
beaesguerra, line 150: Removed since this is a duplicate for this test (the "error" case): https://github.com/Khan/wonder-blocks/blob/2b809b7d46afbf8bfab02559afb5778ad3723992/packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx#L260-L291

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -141,29 +172,51 @@ export default function LabeledField(props: Props) {
...
             <React.Fragment>
                 <Strut size={spacing.small_12} />
-                <LabelSmall
-                    style={light ? styles.lightError : styles.error}
-                    role="alert"
-                    id={errorId}
-                    testId={testId && `${testId}-error`}
+                <View
+                    style={styles.errorSection}
+                    id={errorSectionId}
```
beaesguerra, line 180: I made a few more changes!
- Since we added the error icon with an aria-label that serves as a prefix to the error message, the field's aria-describedby attribute should include this error section element since it contains both the icon and error message
- Replacing `role="alert"` with `aria-live="assertive"`
  - This was moved to the error section so that it would include both the icon and message
  - VoiceOver + Safari would not read out the error when focused on if the element referenced by the `aria-describedby` had `role="alert"` (here's a related [bug report](https://discussions.apple.com/thread/255163557?sortBy=rank) and a [a11y support test report](https://a11ysupport.io/tests/tech__aria__aria-describedby-with-role-alert)). We can use `aria-live="assertive"` instead
- With these changes, screen readers will read out the error message properly when an invalid field is focused on (tested with Safari + VO, Chrome + NVDA, Firefox + NVDA, see recording in PR description)

Some notes on fields that get an error on blur:
- We use `aria-live="assertive"` on the error message so that it is immediately announced and the user can address the issue before submitting the form
- This behavior isn't consistent across browsers/screen readers unfortunately 😢 : 
  -  aria-live=assertive (and the prev role=alert implementation) would not announce the error message on blur in VO + Safari (can be reproduced in this [W3 on focus change example](https://www.w3.org/WAI/tutorials/forms/notifications/#on-focus-change)
  - aria-live=assertive (and the prev role=alert implementation) is not necessary for NVDA+Chrome to properly announce the error message on blur
  - aria-live=assertive is needed for NVDA + Firefox to properly announce the error message on blur

I'll document more about the accessibility of the error handling/validation and will put together examples in the upcoming validation PR! We can also iterate and improve the a11y of this to address browser inconsistencies (see https://github.com/Khan/wonder-blocks/pull/2339#discussion_r1797439504)

marcysutton: Thanks for this write-up, @beaesguerra! The varying behavior of these alerts made me wonder about how they are rendered. I saw the span element for the error in Storybook has a `block` CSS display style, so that's good to rule out. 

Out of curiosity, does it behave any differently if you append an `aria-live=assertive` / `role=alert` element on page load, and then inject the error message content when the field is marked invalid (potentially at different times based on user interaction)?

beaesguerra: > if you append an aria-live=assertive / role=alert element on page load

Does this mean having the live region there always? If so, the live region will still be there if there is no error message content since the contents are conditionally rendered. I did have to make sure the element with `aria-live` persisted in order for it to be announced with nvda + firefox (this is why I removed returning null if there is no error)!

Let me know if I am misunderstanding your question! Happy to try out other things :)

marcysutton: Yes, that's what I meant. Thanks for clarifying! When does the `aria-live` element itself get rendered (as opposed to the error message contents)? That's the part I'm curious about: how early it gets added to the page. It probably won't affect Safari and Voiceover but you never know 🤔

beaesguerra: The `aria-live` element gets rendered when the `LabeledField` component gets rendered! Then the error message contents will get rendered when there is a value in the `error` prop (the `error` prop value can be set any time: on blur, after form submission, etc). 

The [W3 on focus changes validation example](https://www.w3.org/WAI/tutorials/forms/notifications/#on-focus-change) also has the `aria-live` element there before the error message is shown and has the same issue with Safari + VoiceOver!
image

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -141,29 +172,51 @@ export default function LabeledField(props: Props) {
...
+                    testId={testId && `${testId}-error-section`}
```
jandrade, line 181: **suggestion:** This made me realize that we shouldn't use `testId=${testId}-error` in the error label (below) and instead just use a single testId for the error itself. This would also simplify testing the error state.

 ```suggestion
                    testId={testId && `${testId}-error`}
```

File: packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx [hunk 8]
```diff
@@ -141,29 +172,51 @@ export default function LabeledField(props: Props) {
...
+                                aria-label={labels.errorIconAriaLabel}
+                            />
+                            <LabelSmall
+                                style={[
+                                    styles.textWordBreak,
+                                    styles.errorMessage,
+                                    light ? styles.lightError : styles.error,
+                                ]}
+                                id={errorId}
+                                testId={testId && `${testId}-error`}
```
jandrade, line 202: Related to my previous comment about simplifying the error section + testId.

```suggestion
                                testId={testId && `${testId}-error`}
```

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -141,29 +172,51 @@ export default function LabeledField(props: Props) {
...
+                                role="img"
+                                aria-label={labels.errorIconAriaLabel}
+                            />
+                            <LabelSmall
+                                style={[
+                                    styles.textWordBreak,
+                                    styles.errorMessage,
+                                    light ? styles.lightError : styles.error,
+                                ]}
+                                id={errorId}
```
jandrade, line 201: **question:** I wonder if this should be moved to the `View` above. That way, the field could be announced as "some description.... **Error:** the error message" instead of just "some description.... the error message". wdyt?

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -141,29 +172,51 @@ export default function LabeledField(props: Props) {
...
-                <LabelSmall
-                    style={light ? styles.lightError : styles.error}
-                    role="alert"
-                    id={errorId}
-                    testId={testId && `${testId}-error`}
+                <View
+                    style={styles.errorSection}
+                    id={errorSectionId}
+                    testId={testId && `${testId}-error-section`}
+                    aria-live="assertive"
```
jandrade, line 182: **question:** I wonder if we should include `aria-atomic=true` here. I'm thinking on the situation where there are more than 1 validation rules. For example, using an email validation:

1. Field is empty
2. Format is invalid
3. Using the invalid email

In cases like this, the error could persist, but the contents of the error region could change, and ussing `aria-atomic` could help Screen Readers to announce these changes in the DOM. Wdyt?

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -65,6 +71,20 @@ type Props = {
...
+     * The object containing the custom labels used inside this component.
+     *
+     * This is useful for internationalization. Defaults to English.
+     */
+    labels?: LabeledFieldLabels;
+};
+
+type LabeledFieldLabels = {
+    errorIconAriaLabel: string;
+};
```
jandrade, line 84: **suggestion:** This probably needs to be exported so consumers can use the type if they need it (it's probably fine as it is b/c the component instance should be able to infer the types, but just making sure that this case will be covered).

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
```diff
@@ -14,6 +14,25 @@ const defaultOptions = {
...
+    const getField = () => screen.getByRole("textbox");
+    const getError = () => screen.getByText(error);
+    const getErrorSection = () => {
+        // eslint-disable-next-line testing-library/no-node-access
+        const el = getError().parentElement; // The error section is the parent of the alert.
+        if (!el) {
+            throw Error("Error section in LabeledField not found");
+        }
+        return el;
+    };
```
jandrade, line 34: **suggestion:** Related to the error/error-section comment. This is one of the parts that makes me think that we could simplify that code by just only using `error` as the section. That way we wouldn't need to add this complexity to the test and instead we could still test things by checking for `Error: ${errorMessage}`.

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
jandrade: This context is amazing Bea! I'd suggest adding a short comment in the code summarizing why we got to this conclusion for using `assertive` as it would help us contextualize for any future changes.

jandrade (commented):
This is looking great! I'd like to get your thoughts on the error/error-section part before approving, as you have more context on the change.

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
beaesguerra: Good question! The `aria-describedby` is using `errorSectionId` instead of `errorId` so that the "Error:" prefix on the icon aria-label is included in the announcement! 

https://github.com/Khan/wonder-blocks/blob/2b809b7d46afbf8bfab02559afb5778ad3723992/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx#L216-L219

That said, I think having both errorSectionId and errorId can be confusing, and we probably don't need both! I will keep `errorId` and move it to the `View` wrapping the icon and message :)

beaesguerra: I updated things so that there is only `errorId` on the element wrapping the icon and message! The unit tests were also updated to rely on testIds more to access the error section!

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
beaesguerra: Good call! I added in `aria-atomic=true`! I added a Story for an example where the error message changes:  https://5e1bf4b385e3fb0020b7073c-fmfsnacksw.chromatic.com/?path=/story/packages-labeledfield--changing-errors

From testing, without this set, an updated error message is still announced, however, the "Error:" prefix from the icon isn't. Setting `aria-atomic` makes it so both are announced when the error message changes :)

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
```diff
@@ -353,7 +355,22 @@ describe("LabeledField", () => {
                 ["label", getLabel],
                 ["description", getDescription],
                 ["field", getField],
-                ["error", getError],
+                [
+                    "error",
+                    () => {
+                        // In order to get the error section (icon + message)
```
beaesguerra, line 361: Let me know if there is a better way to target the error section without using a test id (since this is a test for when the testId isn't set)!

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
beaesguerra: I updated most of the tests so that we can use the error test id to get the error section! There was one test I kept this logic for since it was explicitly testing when test ids weren't being used https://github.com/Khan/wonder-blocks/pull/2344/files#r1807113422 . Let me know what you think!

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
beaesguerra: Added! Great suggestion 👍

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
beaesguerra: Also, I wasn't able to check for `Error: ${errorMessage}` explicitly in the tests because the `Error:` prefix is an aria-label for the icon so it isn't included as part of the element text. It would be neat though if we could do that though, let me know though if there's a way to achieve this!

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
jandrade: I'm not sure if getting the element that matches the text would work in that case.

e.g.
`screen.getByText(/error: the error message/i)`

jandrade (approved):
LGTM! Thanks for all the details you've added here (+ the investigation on how to announce this to Screen Readers properly) 🚢

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
beaesguerra: It's unable to getByText like this because the "error:" prefix isn't rendered as text, it's the aria-label for the error icon! So the only way I've found to target the error section is to get the parent element of the error message

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
```diff
@@ -64,7 +75,7 @@ describe("LabeledField", () => {
         );
 
         // Assert
-        expect(screen.getByRole("alert")).toBeInTheDocument();
+        expect(screen.getByText(error)).toBeInTheDocument();
```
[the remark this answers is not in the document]
beaesguerra: This will test to make sure the provided error message is rendered in the document! Let me know if I am misunderstanding what you mean!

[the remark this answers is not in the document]
beaesguerra: We discussed offline whether we could use `getByLabelText` to target "Error: the error message". Circling back, getByLabelText is only able to get the prefix part unfortunately! Will keep this test as is to check that the error message is being rendered. There is another test checking that the error icon has an aria-label: 

https://github.com/Khan/wonder-blocks/blob/e05d1f7c9cf9e848c06710b1eb287114f7b2d377/packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx#L211-L229

PR #2345, opened 2024-10-15
WB-1671: Dropdown: use `combobox` role in all openers

jandrade:
## Summary:

Addreses an issue in WB dropdowns where we were using the `button` role in the
opener, which was causing issues with screen readers. This change updates the
role to `combobox` that is more appropriate for the dropdown opener.

Issue: https://khanacademy.atlassian.net/browse/WB-1671

## Test plan:

Navigate to the dropdowns in the Storybook and verify that the role of the
opener is `combobox`.

More details TBD

### DIFF
--- a/.changeset/tasty-rockets-mix.md
+++ b/.changeset/tasty-rockets-mix.md
@@ -2,5 +2,6 @@
 "@khanacademy/wonder-blocks-dropdown": major
 ---
 
-Update dropdown openers for SingleSelect and MultiSelect to use `role="combobox"` instead of `button`.
-SingleSelect and MultiSelect now also require a paired `<label>` element or `aria-label` attribute for accessibility.
+1. Updates dropdown openers for SingleSelect and MultiSelect to use `role="combobox"` instead of `button`.
+2. SingleSelect and MultiSelect now also require a paired `<label>` element or `aria-label` attribute for accessibility. They no longer fall back to text content for labeling, as those contents are now used as combobox values.
+3. Changes the type names for custom label objects from `Labels` to `LabelsValues` and `SingleSelectLabels` to `SingleSelectLabelsValues`, respectively.
--- a/.changeset/tasty-rockets-mix.md
+++ b/.changeset/tasty-rockets-mix.md
@@ -3,5 +3,5 @@
 ---
 
 1. Updates dropdown openers for SingleSelect and MultiSelect to use `role="combobox"` instead of `button`.
-2. SingleSelect and MultiSelect now also require a paired `<label>` element or `aria-label` attribute for accessibility. They no longer fall back to text content for labeling, as those contents are now used as combobox values.
+2. SingleSelect and MultiSelect should have a paired `<label>` element or `aria-label` attribute for accessibility. They no longer fall back to text content for labeling, as those contents are now used as combobox values.
 3. Changes the type names for custom label objects from `Labels` to `LabelsValues` and `SingleSelectLabels` to `SingleSelectLabelsValues`, respectively.
--- a/.changeset/tasty-rockets-mix.md
+++ b/.changeset/tasty-rockets-mix.md
@@ -3,3 +3,4 @@
 ---
 
 Update dropdown openers for SingleSelect and MultiSelect to use `role="combobox"` instead of `button`.
+SingleSelect and MultiSelect now also require a paired `<label>` element or `aria-label` attribute for accessibility.
--- a/.changeset/tasty-rockets-mix.md
+++ b/.changeset/tasty-rockets-mix.md
@@ -1,5 +1,5 @@
 ---
-"@khanacademy/wonder-blocks-dropdown": patch
+"@khanacademy/wonder-blocks-dropdown": major
 ---
 
-Update dropdown openers to use `role="combobox"` instead of `button`.
+Update dropdown openers for SingleSelect and MultiSelect to use `role="combobox"` instead of `button`.
--- a/.changeset/tasty-rockets-mix.md
+++ b/.changeset/tasty-rockets-mix.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-dropdown": patch
+---
+
+Update dropdown openers to use `role="combobox"` instead of `button`.
--- a/__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
@@ -352,7 +352,7 @@ export const Controlled: StoryComponentType = {
  * following arguments:
  *  - `eventState`: lets you customize the style for different states, such as
  *    pressed, hovered and focused.
- *  - `content`: Passes the menu label defined in the parent component. This value
+ *  - `text`: Passes the menu label defined in the parent component. This value
  *    is passed using the placeholder prop set in the ActionMenu component.
  *  - `opened`: Whether the dropdown is opened.
  *
@@ -366,7 +366,7 @@ export const Controlled: StoryComponentType = {
 export const CustomOpener: StoryComponentType = {
     name: "With custom opener",
     args: {
-        opener: ({focused, hovered, pressed, content}: any) => (
+        opener: ({focused, hovered, pressed, text}: any) => (
             <LabelLarge
                 onClick={() => {
                     console.log("custom click!!!!!");
@@ -380,7 +380,7 @@ export const CustomOpener: StoryComponentType = {
                 ]}
                 role="button"
             >
-                {content}
+                {text}
             </LabelLarge>
         ),
     } as Partial<typeof ActionMenu>,
--- a/__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
@@ -352,7 +352,7 @@ export const Controlled: StoryComponentType = {
  * following arguments:
  *  - `eventState`: lets you customize the style for different states, such as
  *    pressed, hovered and focused.
- *  - `text`: Passes the menu label defined in the parent component. This value
+ *  - `content`: Passes the menu label defined in the parent component. This value
  *    is passed using the placeholder prop set in the ActionMenu component.
  *  - `opened`: Whether the dropdown is opened.
  *
@@ -366,7 +366,7 @@ export const Controlled: StoryComponentType = {
 export const CustomOpener: StoryComponentType = {
     name: "With custom opener",
     args: {
-        opener: ({focused, hovered, pressed, text}: any) => (
+        opener: ({focused, hovered, pressed, content}: any) => (
             <LabelLarge
                 onClick={() => {
                     console.log("custom click!!!!!");
@@ -380,7 +380,7 @@ export const CustomOpener: StoryComponentType = {
                 ]}
                 role="button"
             >
-                {text}
+                {content}
             </LabelLarge>
         ),
     } as Partial<typeof ActionMenu>,
--- a/__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
+++ b/__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
@@ -7,61 +7,40 @@ import {LabeledField} from "@khanacademy/wonder-blocks-labeled-field";
 
 <Meta of={MultiSelectAccessibilityStories} />
 
-export const MultiSelectAccessibility = () => (
-    <View>
-        <LabeledField
-            label={<strong>Associated label element</strong>}
-            field={
-                <MultiSelect
-                    selectedValues={["one"]}
-                >
-                    <OptionItem
-                        label="First element"
-                        aria-label="First element, selected"
-                        value="one"
-                    />
-                    <OptionItem
-                        label="Second element"
-                        aria-label="Second element, unselelected"
-                        value="two"
-                    />
-                </MultiSelect>
-            }
-        />
-    </View>
-);
+# Accessibility
 
-# MultiSelect Accessibility
+## Using `LabeledField` with `MultiSelect`
 
-## Labeling the `MultiSelect` opener and option items
+To associate a `MultiSelect` with another visible element (e.g. a `<label>`),
+wrap it in a `LabeledField` component. The label will apply to the `MultiSelect`
+opener. With `LabeledField`, you can supply label text (or a JSX node)
+using the `label` prop to generate a paired `<label>` element. It comes with
+field validation and other features baked in!
 
-If you need to associate this component with another element (e.g. `<label>`),
-make sure to pass the `aria-labelledby` or `id` props to the `MultiSelect` component.
+If for some reason you can't use `LabeledField` for a visible label, you can still
+make `MultiSelect` accessible in a screen reader by associating it with `<label for="">`.
+Pass the `id` of the `MultiSelect` to the `for` attribute.
 
-An `id` on the MultiSelect would pair with `<label for="id">`, while `aria-labelledby`
-on the MultiSelect could associate it with the `id` of any other element. By using
-one of these methods, the `opener` will pair with the associated label and expose
-that label's text for its accessible name.
-
-Also, if you need screen readers to understand relevant information on
-option items, you can use `aria-label` on each item. e.g. You can use it to let
-screen readers know the current selected/unselected status of the item when it
-receives focus.
+Alternatively, you can create an accessible name for `MultiSelect` using `aria-labelledby`.
+Put `aria-labelledby` on `MultiSelect` pointing to the `id` of any other element.
+It won't give you the same enhanced click target as a paired `<label>`, but it still
+helps to create a more accessible experience.
 
 <Canvas of={MultiSelectAccessibilityStories.UsingAriaAttributes} />
 
-## Using `aria-label` for the opener
+## Using `aria-label` for the opener and/or child options
 
 A visible label with `<LabeledField>` is preferred. However, for specific cases
 where the `MultiSelect` is not paired with a `LabeledField` or other visible
 `<label>` element, you **must** supply an `aria-label` attribute for an
 accessible name on the opener.
 
-This will ensure the `MultiSelect` has a name that describes its purpose.
+This will ensure the `MultiSelect` as a whole has a name that describes its purpose.
 
-For example, an `aria-label` for `MultiSelect` in a geography quiz could be
-"Countries in North America" while its value would be one or more of the selected
-options, such as "USA" or "3 selected". It might also have a placeholder value of
-"Select a country", which would go away when the user selected an option.
+Also, if you need screen readers to understand relevant information on
+option items, you can use `aria-label` on each item. e.g. You can use it to let
+screen readers know the current selected/unselected status of the item when it
+receives focus. This can be useful when the options contain icons or other information
+that would need to be omitted from the visible label.
 
 <Canvas of={MultiSelectAccessibilityStories.UsingOpenerAriaLabel} />
--- a/__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
+++ b/__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
@@ -3,35 +3,31 @@ import * as MultiSelectAccessibilityStories from './multi-select.accessibility.s
 
 import {OptionItem, MultiSelect} from "@khanacademy/wonder-blocks-dropdown";
 import {View} from "@khanacademy/wonder-blocks-core";
-import {LabelLarge} from "@khanacademy/wonder-blocks-typography";
+import {LabeledField} from "@khanacademy/wonder-blocks-labeled-field";
 
 <Meta of={MultiSelectAccessibilityStories} />
 
 export const MultiSelectAccessibility = () => (
     <View>
-        <LabelLarge
-            tag="label"
-            id="label-for-multi-select"
-            htmlFor="unique-multi-select"
-        >
-            Associated label element
-        </LabelLarge>
-        <MultiSelect
-            aria-labelledby="label-for-multi-select"
-            id="unique-multi-select"
-            selectedValues={["one"]}
-        >
-            <OptionItem
-                label="First element"
-                aria-label="First element, selected"
-                value="one"
-            />
-            <OptionItem
-                label="Second element"
-                aria-label="Second element, unselelected"
-                value="two"
-            />
-        </MultiSelect>
+        <LabeledField
+            label={<strong>Associated label element</strong>}
+            field={
+                <MultiSelect
+                    selectedValues={["one"]}
+                >
+                    <OptionItem
+                        label="First element"
+                        aria-label="First element, selected"
+                        value="one"
+                    />
+                    <OptionItem
+                        label="Second element"
+                        aria-label="Second element, unselelected"
+                        value="two"
+                    />
+                </MultiSelect>
+            }
+        />
     </View>
 );
 
@@ -56,9 +52,10 @@ receives focus.
 
 ## Using `aria-label` for the opener
 
-For cases where the `MultiSelect` is not paired with a `LabeledField` or other
-visible `<label>` element, you **must** supply an `aria-label` attribute
-for an accessible name on the opener.
+A visible label with `<LabeledField>` is preferred. However, for specific cases
+where the `MultiSelect` is not paired with a `LabeledField` or other visible
+`<label>` element, you **must** supply an `aria-label` attribute for an
+accessible name on the opener.
 
 This will ensure the `MultiSelect` has a name that describes its purpose.
 
--- a/__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
+++ b/__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
@@ -1,0 +1,70 @@
+import {Meta, Story, Canvas} from "@storybook/blocks";
+import * as MultiSelectAccessibilityStories from './multi-select.accessibility.stories';
+
+import {OptionItem, MultiSelect} from "@khanacademy/wonder-blocks-dropdown";
+import {View} from "@khanacademy/wonder-blocks-core";
+import {LabelLarge} from "@khanacademy/wonder-blocks-typography";
+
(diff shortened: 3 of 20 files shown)


jandrade:
NOTE: unchecking the `wonder-blocks` team as I want to run some tests in `webapp` first.

marcysutton:
I rebased this one to bring it up to date and I'm proposing we merge it to deal with a number of screen reader issues for Dropdowns. Essentially, we need the `combobox` role to specify both a label and a value for these components. The one caveat is that VoiceOver and Safari are very buggy with comboboxes in React. I plan to add additional screen reader support in a later PR using the new Announcer in #2362.

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -569,6 +569,7 @@ const MultiSelect = (props: Props) => {
                             disabled={isDisabled}
                             id={uniqueOpenerId}
                             aria-controls={dropdownId}
+                            placeholder={noneSelected}
```
beaesguerra, line 572: Should the placeholder be `menuText` so the aria-label changes when the selected values change? Otherwise, it'll always have an aria-label for no items selected. `menuText` isn't always a string though since it can be custom jsx! Or maybe the select opener wouldn't need an aria-label since the component should be used with a `<label>`? 🤔 What do you think!

Here's a  screenshot that shows the aria-label with "0 items" while the component shows "3 items":
Screenshot 2025-01-09 at 4 05 35 PM

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -38,6 +38,8 @@ type SelectOpenerProps = AriaProps & {
      * of this component. A placeholder has more faded text colors and styles.
      */
     isPlaceholder: boolean;
+
+    placeholder?: string;
```
beaesguerra, line 42: nit: What do you think about renaming the `placeholder` prop to `ariaLabel` so it is named after what it's used for? We also have the `isPlaceholder` prop so I was wondering how they were related (they're not 😅). Or adding prop jsdocs could help with understanding!

.changeset/tasty-rockets-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-dropdown": patch
+---
+
+Update dropdown openers to use `role="combobox"` instead of `button`.
```
beaesguerra, line 5: (very) nit: The ActionMenu also uses DropdownOpener and it still uses role="button"!

```suggestion
Update dropdown openers for SingleSelect and MultiSelect to use `role="combobox"` instead of `button`.
```

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -173,8 +176,9 @@ export default class SelectOpener extends React.Component<
                 aria-haspopup="listbox"
                 data-testid={testId}
                 id={id}
+                role="combobox"
```
beaesguerra, line 179: Side note: Since our selects now use `role=combobox`, we'll be able to treat SingleSelect/MultiSelect more like fields! I created WB-1807 for applying aria-required on these components since it's only valid on `role=combobox` and not `role=button` 😄

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
```diff
@@ -77,7 +77,7 @@ describe("MultiSelect", () => {
             const {userEvent} = doRender(uncontrolledMultiSelect);
 
             // Act
-            await userEvent.click(await screen.findByRole("button"));
+            await userEvent.click(await screen.findByRole("combobox"));
```
beaesguerra, line 80: When we make internal changes that break existing tests, do we consider it a major breaking change since it could break tests in consuming projects? In this case, I'm thinking it's not since the role isn't part of the external API of the component! I'm curious to hear what others think though!

(I might have asked this before but wanted to double check!)

beaesguerra (commented):
Overall, looks great and improves the accessibility of our components :accessibility:  🎉 
 Had some comments and questions!

.changeset/tasty-rockets-mix.md
marcysutton: Seems very important to me! Good call!

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
marcysutton: That's a really good question. I do think it should be a `major` breaking change! Initially I thought it would be a minor change (and I think this PR already had a changeset with `patch` indicated?). But I asked Chat GPT about it to counter any of my own biases and got a really good response:

> Changing an ARIA role from button to combobox in an open source design system component would generally be considered a major change according to Semantic Versioning (SemVer).
> 
> Here's why:
> - **Breaking change:** The ARIA role defines how assistive technologies and user agents interpret the element. Changing the role from `button` to `combobox` could fundamentally alter how the component behaves for users relying on assistive technologies like screen readers. For example:
>     - A `button` role indicates a simple button for triggering an action, whereas a `combobox` role indicates an interactive control for selecting a value, often with a dropdown, which could change the expected behavior of the component significantly.
>    - This change would affect accessibility, user experience, and potentially even the visual appearance or interaction model if other associated attributes (like `aria-expanded`, `aria-controls`, etc.) also need to be modified.
> - **Backward compatibility:** If users of the component are relying on the previous behavior (button), changing the ARIA role to `combobox` could break their existing implementations. This would be a backward-incompatible change.

packages/wonder-blocks-dropdown/src/components/dropdown-opener.tsx
```diff
@@ -51,15 +51,21 @@ type Props = Partial<Omit<AriaProps, "aria-disabled">> & {
      * If the dropdown has an error.
      */
     error?: boolean;
+    /**
+     * The role of the opener.
+     */
+    role?: "combobox" | "button";
```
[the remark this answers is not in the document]
marcysutton: Makes sense to me, especially after looking at usage! I wouldn't want to require every consuming author to set a role as it should be handled for them internally. But it looks like they would use a custom opener in that case, and not this component directly.

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
marcysutton: LOVE that! And yes, `combobox` makes it into a real form field with a value! (even though VoiceOver is buggy on that part!)

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: Good catch @beaesguerra! I don't think that `aria-label` is necessarily the same as a placeholder for a combobox. This text about the number of items is more of a value than a label. I'm guessing the intent was to use a placeholder as a label in the absence of a separate label element (or associated label, however it's marked up). But we have to be careful with `aria-label` to not override an associated label, if it's paired with `<label for="$idMatchingThisCombobox">`. I'm looking into it!

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
marcysutton: Good idea! I see the other `isPlaceholder` prop is for styling. I mentioned in my other comment that we have to be careful with `aria-label` to ensure the correct label is computed, so maybe this other prop can play a role in that logic.

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
beaesguerra: cc: @jandrade in case you have thoughts on how this is normally handled in WB! 

If this is a major breaking change, we'll have to make sure to coordinate the perseus + webapp updates since [WB is a peer dependency in perseus](https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/3548610656/Perseus+and+Wonder+Blocks) and we have to be more careful with the order of releases for breaking changes!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: @jandrade or @beaesguerra do you know if the MultiSelect component is used with an explicit label (`<label for="thisSelectComponentWithMatchingID">`? I don't see a story for it like there is on [SingleSelect](https://khan.github.io/wonder-blocks/?path=/docs/packages-dropdown-singleselect-accessibility--docs).

I'll dig into this more -- but I'm wondering if we have any way to conditionally add an `aria-label` to the opener for placeholders and not clash with other labeling mechanisms. Any input you can add would be helpful!

beaesguerra: It should be used with a label, specifically LabeledField 😄 Here's an example in Storybook: https://5e1bf4b385e3fb0020b7073c-fnsnwmxvwb.chromatic.com/?path=/story/packages-labeledfield--fields

jandrade: Agreed, we should use `LabeledField` (or `label` on very specific cases), but there might be places where we are not using labels and `aria-label` could be useful. For example, the course selector in the teacher dashboard.

You'll need to have "Teacher" access (available with your @khanacademy.org account), then go to `/teacher/dashboard`, click on any class and use the dropdown on the top left.

Screenshot 2025-01-14 at 2 59 58 PM

NOTE: This could be a case where we could benefit from having an actual label and this could be validated with the TX team.

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
jandrade: The point about changing how Screen Readers could change based on these changes is very compelling. I've tended not to mark as `major` these kind of changes as the public API does not change, but as it affects the way this component is announced by SRs, then that makes this a great candidate for a `major` change. For the consumer unit tests changing, I don't consider it a breaking change as the bundled code does not change (or the code that is being shipped and consumed by the end user).

@beaesguerra makes another great point about coordinating with the Perseus team as I know they use the dropdown package in their widgets.

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: Oh yes, `LabeledField` 🤦‍♀️! Thank you so much for this context. I think the challenge is at the Combobox level, where it won't know whether it's being used by labeled field. Fortunately I think we can leverage the Accessible Name Computation spec to prioritize the LabeledField label over an `aria-label` (in the event both are present). We can also instruct consumers when and when not to use `aria-label` in docs!

.changeset/tasty-rockets-mix.md
marcysutton: Added another note to the changelog about labeling!

marcysutton:
I've updated this PR with a few things:
- Using `aria-label` when applied. not falling back to placeholder content ("none selected", etc.)
- Stories were updated to show how to use `<label>` vs. `aria-label`
- A couple of internal renames:
    - typing of `Labels` to `LabelsValues` since that object contains both things
    - a `text` variable in openers to `content`, since it also potentially contains JSX nodes. This one isn't strictly necessary and could be omitted. But it helped my understanding of the code when I was attempting to make automatic `aria-label` happen (it required pure text).

The one outstanding thing I'm not sure how to handle is labeling of a custom opener. My initial attempt didn't work, as I'm not totally understanding that API just yet! Seems like it could be done as a follow-up and not block this work.

packages/wonder-blocks-dropdown/src/index.ts
```diff
@@ -21,4 +21,4 @@ export {
     Listbox,
 };
 
-export type {Labels, SingleSelectLabels};
+export type {LabelsValues, SingleSelectLabelsValues};
```
beaesguerra, line 24: The type renames will also be breaking changes since we export them. If we want to keep the rename, we'll also have to update existing imports and mention it in the changeset!

packages/wonder-blocks-dropdown/src/components/dropdown-opener.tsx
```diff
@@ -51,15 +51,21 @@ type Props = Partial<Omit<AriaProps, "aria-disabled">> & {
...
 
 type DefaultProps = {
     disabled: Props["disabled"];
+    role: Props["role"];
 };
 
 class DropdownOpener extends React.Component<Props> {
     static defaultProps: DefaultProps = {
         disabled: false,
+        role: "combobox",
```
beaesguerra, line 68: Since the `role` prop is required, we don't need to provide it in the default props!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -557,7 +557,7 @@ const MultiSelect = (props: Props) => {
                             onBlur={onOpenerBlurValidation}
                             disabled={isDisabled}
                             ref={handleOpenerRef}
-                            text={menuText}
+                            content={menuContent}
                             opened={open}
```
beaesguerra, line 561: Do we need to apply the `aria-label` to the DropdownOpener too for when a custom opener is provided? Here's an [example of using a custom opener](https://5e1bf4b385e3fb0020b7073c-kblofxoolb.chromatic.com/?path=/story/packages-dropdown-singleselect--custom-opener)!

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -38,6 +38,11 @@ type SelectOpenerProps = AriaProps & {
      * of this component. A placeholder has more faded text colors and styles.
      */
     isPlaceholder: boolean;
+    /**
+     * A label to expose on the opener, in the absence of an associated label element or `aria-labelledby`.
+     * If no `ariaLabel` prop is provided, the opener `aria-label` will fall back to `placeholder` (if present).
```
beaesguerra, line 43: I think this part can be removed since there is no `placeholder` prop anymore!

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -183,7 +190,7 @@ export default class SelectOpener extends React.Component<
                 <LabelMedium style={styles.text}>
                     {/* Note(tamarab): Prevents unwanted vertical
                                 shift for empty selection */}
-                    {children || "\u00A0"}
+                    {children || <span aria-hidden="true">&nbsp;</span>}
```
beaesguerra, line 193: I'm curious about this update! Was it causing issues?

packages/wonder-blocks-dropdown/src/components/single-select.tsx
```diff
@@ -453,7 +454,7 @@ const SingleSelect = (props: Props) => {
                             onClick={handleClick}
                             disabled={isDisabled}
                             ref={handleOpenerRef}
-                            text={menuText}
+                            content={menuText}
                             opened={open}
```
beaesguerra, line 458: Same question here about applying the aria-label to custom opener!

beaesguerra (commented):
Left some comments and questions! With the label types rename and the `text` -> `content` renaming, they will be code breaking changes - wanted to get your thoughts on this!

__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
```diff
@@ -366,7 +366,7 @@ export const Controlled: StoryComponentType = {
 export const CustomOpener: StoryComponentType = {
     name: "With custom opener",
     args: {
-        opener: ({focused, hovered, pressed, text}: any) => (
+        opener: ({focused, hovered, pressed, content}: any) => (
```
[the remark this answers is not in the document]
marcysutton: Good call! I think I'm inclined to revert that one to limit the scope of this change. It's going to be impactful enough just having to label so many component instances! Curious to hear @jandrade's take on it too, though.

__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
```diff
@@ -0,0 +1,70 @@
...
+For cases where the `MultiSelect` is not paired with a `LabeledField` or other
+visible `<label>` element, you **must** supply an `aria-label` attribute
+for an accessible name on the opener.
+
+This will ensure the `MultiSelect` has a name that describes its purpose.
+
+For example, an `aria-label` for `MultiSelect` in a geography quiz could be
+"Countries in North America" while its value would be one or more of the selected
+options, such as "USA" or "3 selected". It might also have a placeholder value of
+"Select a country", which would go away when the user selected an option.
```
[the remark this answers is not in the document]
marcysutton: Yes, I will add that guidance! Maybe there's a better example I could highlight for this scenario, if you have any ideas...

__docs__/wonder-blocks-dropdown/single-select.accessibility.mdx
```diff
@@ -36,16 +36,36 @@ export const SingleSelectAccessibility = () => (
...
-Also, if you need screen readers to understand any relevant information on every
-option item, you can use `aria-label` on each item. e.g. You can use it to let
+Also, if you need screen readers to understand relevant information on
+option items, you can use `aria-label` on each item. e.g. You can use it to let
 screen readers know the current selected/unselected status of the item when it
 receives focus.
 
 <Canvas of={SingleSelectAccessibilityStories.UsingAriaAttributes} />
+
+## Using `aria-label` for the opener
```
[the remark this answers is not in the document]
marcysutton: That's what I'm not sure about! I'll have to investigate custom openers in a follow-up. It wasn't immediately obvious to me where an `aria-label` would go in the public API for custom openers.

packages/wonder-blocks-dropdown/src/components/dropdown-opener.tsx
```diff
@@ -51,15 +51,21 @@ type Props = Partial<Omit<AriaProps, "aria-disabled">> & {
...
     error?: boolean;
+    /**
+     * The role of the opener.
+     */
+    role: "combobox" | "button";
 };
 
 type DefaultProps = {
     disabled: Props["disabled"];
+    role: Props["role"];
```
marcysutton, line 62: @beaesguerra or @kevinb-khan out of curiosity, do you know why this object access pattern is used for DefaultProps?

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
marcysutton: Yes -- the word "space" was being announced in VoiceOver!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: Yes we do! That might be a good huddle topic, since I wasn't exactly sure how to make that work in my first attempt. I have some stashed changes I could pick up, but I figured they could happen in a follow-up.

__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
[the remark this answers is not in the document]
marcysutton: I restored the API name but left a few test changes and one internal variable. Makes sense to limit the scope of the PR and make it easier on everyone!

__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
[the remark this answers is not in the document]
beaesguerra: Hmm an example I can think of where labels would be extra helpful is if there are multiple MultiSelect components. When there are selected values in each, all a user would see is "2 selected", "3 selected", etc!

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
beaesguerra: Ah good catch! Thank you!

packages/wonder-blocks-dropdown/src/index.ts
beaesguerra: @marcysutton any thoughts about this other rename breaking change!

__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
```diff
@@ -3,35 +3,31 @@ import * as MultiSelectAccessibilityStories from './multi-select.accessibility.s
...
-                value="one"
-            />
-            <OptionItem
-                label="Second element"
-                aria-label="Second element, unselelected"
-                value="two"
-            />
-        </MultiSelect>
+        <LabeledField
+            label={<strong>Associated label element</strong>}
```
beaesguerra, line 13: Do we need the label wrapped in the `<strong>` tag? I'm wondering if we should use the default styling in our examples so we can avoid one-offs!

marcysutton: We probably don't need it! I was trying to make it match how it looked before.

__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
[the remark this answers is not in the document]
marcysutton: Thanks @beaesguerra! I guess I'm still wondering about the opposite situation: where an `aria-label` would be satisfactory without the visible label. It's probably more likely for `SingleSelect` though.

packages/wonder-blocks-dropdown/src/index.ts
marcysutton: I think this one is more important to communicate the nature of the object, since it now contains values. I can add a note to the changelog!

__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
marcysutton: Do these examples even get used somehow? I don't think they are actually rendered. Maybe something broke along the way? I'm considering removing them since the examples come from the `*.accessibility.stories` files.

__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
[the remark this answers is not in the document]
marcysutton: I think I'll just delete that from the `MultiSelect` doc and keep a relevant example on the `SingleSelect` one.

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: This is fixed and added to the PR!

packages/wonder-blocks-dropdown/src/components/single-select.tsx
marcysutton: It's in there now, good call!

__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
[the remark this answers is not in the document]
beaesguerra: Good question! Hmmm I'm not sure, maybe the use case where the select is used in the middle of a sentence?

__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
beaesguerra: I'm also a bit unsure when we should create both `.accessibility.mdx` files and `.accessibility.stories.ts` files. I'm okay with combining them and keeping the docs in `.accessibility.stories.ts` files instead!

cc: @jandrade in case you have thoughts on this!

.changeset/tasty-rockets-mix.md
```diff
@@ -0,0 +1,7 @@
+---
+"@khanacademy/wonder-blocks-dropdown": major
+---
+
+1. Updates dropdown openers for SingleSelect and MultiSelect to use `role="combobox"` instead of `button`.
+2. SingleSelect and MultiSelect now also require a paired `<label>` element or `aria-label` attribute for accessibility. They no longer fall back to text content for labeling, as those contents are now used as combobox values.
```
beaesguerra, line 6: small nit: Might be worth rephrasing the "required" part a bit since we aren't changing the API to enforce this (aria-label is still optional and we don't have anything in the component checking if there is an associated label element). Maybe: 
```suggestion
2. SingleSelect and MultiSelect should have a paired `<label>` element or `aria-label` attribute for accessibility. They no longer fall back to text content for labeling, as those contents are now used as combobox values.
```

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
```diff
@@ -931,6 +968,53 @@ describe("MultiSelect", () => {
...
+                    <OptionItem label="item 3" value="3" />
+                </MultiSelect>,
+            );
+
+            // Act
+            const opener = await screen.findByRole("combobox");
+
+            // Assert
+            expect(opener).toHaveAccessibleName("Search button");
+        });
```
beaesguerra, line 1016: Thanks for adding these test cases!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -251,6 +253,7 @@ const MultiSelect = (props: Props) => {
         shortcuts = false,
         style,
         className,
+        "aria-label": ariaLabel,
```
beaesguerra, line 256: (non-blocking) docs: One thing I realized with the aria attributes is that they don't automatically show up in the Storybook component docs 😢 : https://5e1bf4b385e3fb0020b7073c-oodutnedcb.chromatic.com/?path=/docs/packages-dropdown-multiselect--docs 

We could explicitly add it to the story argTypes so they show up in the table! This can also be part of other work, since this is common to all components that extend the `AriaProps` type

beaesguerra (approved):
Looks great! 🎉 This will address many accessibility issues!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: That is a great idea! I can open an issue for it, unless there already is one!

PR #2346, opened 2024-10-15
Include lint and test jobs for node-ci-push workflow

beaesguerra:
## Summary 
- lint and test jobs from node-ci-pr.yml are copied over to node-ci-push.yml
- this is done so that feature branches can be updated with the latest changes in main (there are currently branch protection rules on feature/* branches that require these status checks to pass)

Issue: XXX-XXXX

### DIFF
--- a/.github/workflows/node-ci-push.yml
+++ b/.github/workflows/node-ci-push.yml
@@ -68,3 +68,100 @@ jobs:
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
         files: ./coverage/coverage-final.json
+
+  lint:
+    name: Lint
+    needs: prime_cache_primary
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        node-version: [20.x]
+    steps:
+      - uses: actions/checkout@v4
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Install & cache node_modules
+        uses: Khan/actions@shared-node-cache-v2
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Get All Changed Files
+        uses: Khan/actions@get-changed-files-v2
+        id: changed
+
+      - id: js-ts-files
+        name: Find .js, .ts changed files
+        uses: Khan/actions@filter-files-v1
+        with:
+          changed-files: ${{ steps.changed.outputs.files }}
+          extensions: '.js,.jsx,.ts,.tsx'
+
+      - id: eslint-reset
+        uses: Khan/actions@filter-files-v1
+        name: Files that would trigger a full eslint run
+        with:
+          changed-files: ${{ steps.changed.outputs.files }}
+          files: '.eslintrc.js,yarn.lock,.eslintignore'
+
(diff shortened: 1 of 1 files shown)


.github/workflows/node-ci-push.yml
```diff
@@ -68,3 +68,100 @@ jobs:
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
         files: ./coverage/coverage-final.json
+
```
beaesguerra, line 71: These changes are based on chatting with Lilli on how to work with [Intermediate Integration Branches](https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/2918416504/Understand+our+Branching+Strategy#Example-2---Intermediate-Integration-Branch). These changes will make it so `main` will include these checks and we can include commits from `main` into feature branches. Please let me know if there's another way to address this so that we're able to merge the latest changes from main into feature branches!

jandrade:
@beaesguerra as we discussed offline. It would be worth exploring using reusable workflows to avoid repeating the jobs in two separate workflows.

.github/actions/lint/action.yml
```diff
@@ -0,0 +1,67 @@
...
+      with:
+        full-trigger: ${{ steps.eslint-reset.outputs.filtered }}
+        full: yarn lint:ci .
+        limited-trigger: ${{ steps.js-ts-files.outputs.filtered }}
+        limited: yarn lint:ci {}
+
+    - name: Typecheck
+      if: always() # always run this check until we update the eslint config
+      # if: steps.js-ts-files.outputs.filtered != '[]' || steps.typecheck-reset.outputs.filtered != '[]'
+      shell: bash
```
beaesguerra, line 54: The `action.yml` files contain the same steps as before with the addition of `shell: bash` on the steps that had errors around requiring the `shell` config

.github/workflows/node-ci-push.yml
```diff
@@ -35,36 +36,29 @@ jobs:
...
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        node-version: [20.x]
+        shard: ["1/2", "2/2"]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/test
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
beaesguerra, line 64: One other difference between composite actions and reusing workflows is that [secrets can only be accessed by actions using an input or environment variable](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#accessing-your-secrets), while [reusable workflows can define secrets](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow) 

Here, we pass the codecov token using an env variable. Let me know if there are any concerns with this!

.github/workflows/node-ci-push.yml
```diff
@@ -35,36 +36,29 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
-
-  coverage:
```
beaesguerra, line 39: The coverage job can be removed since the test action covers test coverage too!

beaesguerra:
We will be moving forward with option 2 using reusable workflows. See https://github.com/Khan/wonder-blocks/pull/2348 for more details!

PR #2347, opened 2024-10-15
TextField and TextArea: add `error` prop

beaesguerra:
## Summary:
Adds an `error` prop to TextField and TextArea components so they can be put in an error state declaratively. This is useful for validation that happens after a form submission.

Related to the LabeledField work:
- By having an error prop, the LabeledField component will set the `error` prop to `true` if it is provided with an error message (will do this in a separate PR for LabeledField work)
- The error prop is consistent with other fields such as the SingleSelect, MultiSelect, and Combobox components. I'll be updating SearchField in another PR (with other updates)!
- These changes can be released separately from the LabeledField changes since they are incremental changes to existing components

Issue: WB-1777

### TextField
- Setting the `error` prop puts the component in an error state (`aria-invalid` is set to `true` and error styling is applied) (`?path=/story/packages-form-textfield--error`)
- Validation continues to put the component in an error state if the value is not valid (`?path=/story/packages-form-textfield--error-from-validation`)
- Required prop continues to put the component in an error state if a value is cleared (`?path=/story/packages-form-textfield--required`)

### TextArea
- Setting the `error` prop puts the component in an error state (`aria-invalid` is set to `true` and error styling is applied) (`?path=/story/packages-form-textarea--error`)
- Validation continues to put the component in an error state if the value is not valid (`?path=/story/packages-form-textarea--error-from-validation`)
- Required prop continues to put the component in an error state if a value is cleared (`?path=/story/packages-form-textarea--required`)

### DIFF
--- a/__docs__/wonder-blocks-form/text-area-variants.stories.tsx
+++ b/__docs__/wonder-blocks-form/text-area-variants.stories.tsx
@@ -10,8 +10,6 @@ import {TextArea} from "@khanacademy/wonder-blocks-form";
 /**
  * The following stories are used to generate the pseudo states for the
  * TextArea component. This is only used for visual testing in Chromatic.
- *
- * Note: Error state is not shown on initial render if the TextArea value is empty.
  */
 export default {
     title: "Packages / Form / TextArea / All Variants",
@@ -40,7 +38,7 @@ const states = [
     },
     {
         label: "Error",
-        props: {validate: () => "Error"},
+        props: {error: true},
     },
 ];
 const States = (props: {
--- a/__docs__/wonder-blocks-form/text-area.stories.tsx
+++ b/__docs__/wonder-blocks-form/text-area.stories.tsx
@@ -10,7 +10,7 @@ import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
 import Button from "@khanacademy/wonder-blocks-button";
 import {LabelSmall, LabelLarge} from "@khanacademy/wonder-blocks-typography";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
-import {View} from "@khanacademy/wonder-blocks-core";
+import {PropsFor, View} from "@khanacademy/wonder-blocks-core";
 
 import TextAreaArgTypes from "./text-area.argtypes";
 
@@ -60,9 +60,9 @@ const styles = StyleSheet.create({
     },
 });
 
-const ControlledTextArea = (args: any) => {
+const ControlledTextArea = (args: PropsFor<typeof TextArea>) => {
     const [value, setValue] = React.useState(args.value || "");
-    const [error, setError] = React.useState<string | null>(null);
+    const [error, setError] = React.useState<string | null | undefined>(null);
 
     const handleChange = (newValue: string) => {
         setValue(newValue);
@@ -77,7 +77,11 @@ const ControlledTextArea = (args: any) => {
                 onValidate={setError}
             />
             <Strut size={spacing.xxSmall_6} />
-            {error && <LabelSmall style={styles.error}>{error}</LabelSmall>}
+            {(error || args.error) && (
+                <LabelSmall style={styles.error}>
+                    {error || "Error from error prop"}
+                </LabelSmall>
+            )}
         </View>
     );
 };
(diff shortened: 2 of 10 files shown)


__docs__/wonder-blocks-form/text-area-variants.stories.tsx
```diff
@@ -10,8 +10,6 @@ import {TextArea} from "@khanacademy/wonder-blocks-form";
 /**
  * The following stories are used to generate the pseudo states for the
  * TextArea component. This is only used for visual testing in Chromatic.
- *
- * Note: Error state is not shown on initial render if the TextArea value is empty.
```
beaesguerra, line 14: Since we can explicitly put the component in an error state, the variants stories can properly show the error styling in all cases now!

packages/wonder-blocks-form/src/components/text-area.tsx
```diff
@@ -139,6 +139,10 @@ type TextAreaProps = AriaProps & {
      * Called right after the textarea is validated.
      */
     onValidate?: (errorMessage?: string | null | undefined) => unknown;
+    /**
+     * Whether the textarea is in an error state.
+     */
+    error?: boolean;
```
jandrade, line 145: I feel that we could get into a tricky situation here.

Have you found a situation where a component instance needs to use both `error` and `validate`?

I'm thinking if we need to add more guardrails to prevent a case where these two props could collide.

It's tricky from the API perspective that the validation function would pass but the user would still see an error, and I wonder if this case is expected or not (and if not, then we should prevent from this to happen from the dev perspective).

Here's an example:
https://5e1bf4b385e3fb0020b7073c-kwxacxtfdq.chromatic.com/iframe.html?args=error:!true&id=packages-form-textarea--error-from-validation&viewMode=story

Screenshot 2024-10-16 at 11 23 04 AM

The email pattern is now valid, but the input is still showing its `error` state.

I wonder if/how we need to cover this case. wdyt?

beaesguerra: Great question! I think this scenario is behaving as I would have expected (remaining in an error state after the email is in a valid format because the error prop is set to true). An example use case for this could be a sign up form where even if the email is valid, there could still be an error after form submission like the email provided is already associated with an account. 

The workaround I used for backend validation errors before was to keep track of whether there is an error in the component state, check for that error in the validate function, and then also set the `key` on the component to whether there was an error so that the component would re-mount to trigger the validate function. Here is an [example from webapp](https://github.com/Khan/webapp/pull/24613/files#r1717640619) where I ran into this issue! 

Part of the confusion could also be there is no error message in the story. The error message isn't part of the component, but it is shown in the story for example purposes. I'll update the story so another error message is shown if the `error` control is set! 

Let me know what you think or if you have some ideas on other alternatives!

jandrade: oh I see, thanks for the context.

Given your example, I still see that the user could get into an state where they use an email that is not associated with an account and they use valid email. Should the input update to its valid state? or should it remain using the error state?

beaesguerra: The input would return to a valid state when the `error` prop is set to false (since in that case the email is fine to use).

I'll put together another story that has both `error` and `validate` props configured with logic so we can see how it will behave!

jandrade: > The input would return to a valid state when the error prop is set to false (since in that case the email is fine to use).

That's the point I'm trying to get into.... this could lead into an "[make impossible states impossible](https://kentcdodds.com/blog/make-impossible-states-impossible)" situation. We would need to make sure that both `error` and `validate` are in sync and there might be cases where it would be confusing for the developer implementing a feature with these characteristics.

I wonder if there's a way to refactor this API to make it more predictive (even if it means changing `validate/onValidate`).

beaesguerra: Thanks for the feedback, Juan! From our pairing session, we decided that:
- it makes sense to support both `error` and `validate` props in the TextField and TextArea components since there are use cases for using both
- we will need documentation around when to use these props
- we will leave validation up to the field components instead of having LabeledField pass it down. This means LabeledField will only have an error prop for the error message and the validate prop will stay on TextField/TextArea
- it will be helpful to have a shared utility/hook for the shared logic around validation: when to trigger validation, support for async validation, etc (will be explored in another PR). 

I've added some more documentation around when to use `error` vs `validate` props, and have added examples for when both are used ([TextField example](https://5e1bf4b385e3fb0020b7073c-cvpbgtsyrn.chromatic.com/?path=/story/packages-form-textfield--error-from-prop-and-validation) and [TextArea example](https://5e1bf4b385e3fb0020b7073c-cvpbgtsyrn.chromatic.com/?path=/story/packages-form-textarea--error-from-prop-and-validation))! I'll update the implementation spec as well!

Let me know what you think!

__docs__/wonder-blocks-form/text-area.stories.tsx
```diff
@@ -178,6 +204,83 @@ export const Error: StoryComponentType = {
...
+ * 1. It starts with an invalid email. The error message shown is the message returned
+ * by the `validate` function prop
+ * 2. Once the email is fixed to `test@test.com`, the validation error message
+ * goes away since it is a valid email.
+ * 3. When the Submit button is pressed, another error message is shown (this
+ * simulates backend validation).
+ * 4. When you enter any other email address, the error message is
+ * cleared.
+ */
+export const ErrorFromPropAndValidation = (args: PropsFor<typeof TextArea>) => {
```
jandrade, line 223: **praise:** This example + documentation is awesome 👏

__docs__/wonder-blocks-form/text-area.stories.tsx
```diff
@@ -158,14 +162,36 @@ export const ReadOnly: StoryComponentType = {
...
+ *
  * Note that we will internally set the correct `aria-invalid` attribute to the
  * `textarea` element:
- * - `aria-invalid="true"` if there is an error message.
- * - `aria-invalid="false"` if there is no error message.
+ * - `aria-invalid="true"` if there is an error.
+ * - `aria-invalid="false"` if there is no error.
  */
-export const Error: StoryComponentType = {
+export const ErrorFromValidation: StoryComponentType = {
```
jandrade, line 194: **question:** Do you think we should disable snapshots for this story? or maybe the `Error` one, or both? I'm not sure if we are testing different visual things and/if this is already covered in some way by the "All variants" stories.

__docs__/wonder-blocks-form/text-field.stories.tsx
```diff
@@ -419,50 +423,145 @@ function ErrorRender() {
...
  * `input` element:
- * - aria-invalid="true" if there is an error message.
- * - aria-invalid="false" if there is no error message.
+ * - aria-invalid="true" if there is an error.
+ * - aria-invalid="false" if there is no error.
  */
-export const Error: StoryComponentType = {
+export const ErrorFromValidation: StoryComponentType = {
     render: ErrorRender,
 };
```
jandrade, line 482: same question here to understand if both stories need to include visual snapshots.

jandrade (approved):
Looks awesome! this is a great improvement 👏. Just left a couple of very small  questions/suggestions for some stories before landing, but this looks really solid overall 🚀

__docs__/wonder-blocks-form/text-area.stories.tsx
beaesguerra: Good idea! The "All variants" stories covers the error state already, I've disabled snapshots for the error related stories for TextField and TextArea!

beaesguerra:
Update: This PR is good to merge! My next set of changes will also update TextField and TextArea so I plan to release these changes with that (I should've created a feature branch for these, but will use a stacked branch approach now!)

PR #2348, opened 2024-10-17
Include lint and test jobs for node-ci-push workflow (using reusable workflow)

beaesguerra:
## Summary:
- Lint and test jobs from [node-ci-pr.yml](https://github.com/Khan/wonder-blocks/blob/a32b0779acc9ed8dc742955160cc7748052296e8/.github/workflows/node-ci-pr.yml#L65-L160) are also run in the `node-ci-push.yml` workflow
  - this is done so that feature branches can be updated with the latest changes in main (there are currently branch protection rules on feature/* branches that require these status checks to pass)
- Lint and test jobs are refactored into [reusable workflows](https://docs.github.com/en/actions/sharing-automations/reusing-workflows) so that both `node-ci-push` and `node-ci-pr` can use the same config

See https://github.com/Khan/wonder-blocks/pull/2346 for an alternative where we use composite actions instead to share configuration

Update: We decided to use reusable workflows so that the logs are easier to read for the steps and we can use the `secrets` config (which isn't supported by composite actions). The branch rules have been updated to look for `Lint / Lint` and `Test / Test` checks.

Pros:
- shared configuration for `node-ci-push` and `node-ci-pr`
- easy to see the [logs](https://github.com/Khan/wonder-blocks/actions/runs/11393035774/job/31700470290?pr=2348) for the different steps:
Screenshot 2024-10-17 at 3 21 04 PM

Cons:
- since we are using another workflow, the names of the jobs are nested (ie. `Node CI (PR) / Lint / Lint` instead of just `Node CI (PR) / Lint` ) and the github checks will need to be updated (let me know though if there's a way around this!)
Screenshot 2024-10-17 at 3 16 53 PM

Issue: WB-1778

## Test plan:
- Confirm linting and tests run in CI
- Confirm steps properly fail when there are linting errors, typescript errors, and failing tests
image


.github/workflows/node-ci-push.yml
```diff
@@ -35,36 +36,14 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
+  lint:
+    name: Lint
+    needs: prime_cache_primary
+    uses: ./.github/workflows/node-ci-lint.yml
 
-  coverage:
```
beaesguerra, line 39: This job can be removed since the Test workflow also gathers coverage!

.github/workflows/node-ci-pr.yml
```diff
@@ -65,99 +65,14 @@ jobs:
...
-      # Testing and coverage
-      - name: Run jest tests with coverage
-        run: yarn coverage:ci --shard ${{ matrix.shard }}
-      - name: Upload Coverage
-        uses: codecov/codecov-action@v4
-        with:
-          token: ${{ secrets.CODECOV_TOKEN }}
-          files: ./coverage/coverage-final.json
+    uses: ./.github/workflows/node-ci-test.yml
+    secrets:
```
beaesguerra, line 74: Here, we are able to pass tokens [using secrets](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow)

jandrade (approved):
This is a great improvement! thanks for the detailed description and iterating on the problem 👏 🚀

beaesguerra:
Added some notes around these findings here: https://khanacademy.atlassian.net/wiki/x/AYGUyQ

PR #2349, opened 2024-10-18
WB-1655: Add mediaQuery breakpoints to wonder-blocks-tokens

marcysutton:
Adds initial media query tokens for breakpoint sizes in wonder-blocks-tokens. Breakpoint values are derived from internal docs:

  | Wonder Blocks (Web) | Design at (Figma)
-- | -- | --
small | < 682px | 320px width (iPhone SE, 1st-gen)
medium | 682 – 1023px | 768px width (iPad portrait, many generations)
large | 1024px – infinity | 1366px width (Chromebook)

The structure of this new `mediaQueries` object mimics the version in [webapp](https://github.com/Khan/wonder-blocks/blob/d12df6e73471a79f9a11ad4e6e5e23e0985e0144/packages/wonder-blocks-tokens/src/tokens/mediaQueries.ts) (internal) pretty closely, but with values documented internally in Confluence: (see "Future Breakpoints"):  https://khanacademy.atlassian.net/wiki/spaces/WB/pages/2099970518/Layout+Breakpoints

Screenshot of the new docs, note there aren't examples of each one like the other tokens:
Screenshot 2024-10-18 at 3 47 51 PM

I did some initial testing in Wonder Blocks to ensure the type definition works and paired with @kevinbarabash to see if we could make it any more dynamic. But I think simple is fine for now as it's a bit of dark magic how these media query fragments turn into CSS Properties in Aphrodite. 

My next step will be to update all `@media` viewport sizing in WB to use these new breakpoints!

Jira issue: https://khanacademy.atlassian.net/browse/WB-1655


packages/wonder-blocks-tokens/src/tokens/media-queries.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * A default set of media queries to use for different screen sizes.
+ *
+ * Breakpoint documentation: https://khanacademy.atlassian.net/wiki/spaces/WB/pages/2099970518/Layout+Breakpoints
+ *
+ */
+
+export const pureXsMax = "567px";
```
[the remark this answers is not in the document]
beaesguerra: The individual token values would be useful for things like the [Storybook breakpoint config](https://github.com/Khan/wonder-blocks/blob/1e2d70bf62ccd08c2c6406f1bd137ca8004eb8cc/.storybook/preview.tsx#L14-L43)!

beaesguerra (commented):
Excited to have tokens for breakpoints 😄 Left a few questions I had!

__docs__/wonder-blocks-tokens/tokens-media-queries.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+import TokenTable from "../../.storybook/components/token-table";
+
+import {View} from "@khanacademy/wonder-blocks-core";
+import * as tokens from "@khanacademy/wonder-blocks-tokens";
+
+import ComponentInfo from "../../.storybook/components/component-info";
+import packageConfig from "../../packages/wonder-blocks-tokens/package.json";
+
+<Meta title="Packages / Tokens / Media Queries" />
```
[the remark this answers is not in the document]
marcysutton: Great idea, I'll add that!

packages/wonder-blocks-tokens/src/tokens/media-queries.ts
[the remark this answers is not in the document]
marcysutton: That is a fantastic idea. It had occurred to me...it makes total sense! I'm looking at the types for this change again today so I'll work that in too. Thanks, @beaesguerra!

types/aphrodite.d.ts
```diff
@@ -1,6 +1,29 @@
 declare module "aphrodite" {
     import * as React from "react";
 
+    const xs = "@media screen and (max-width: 567px) /* breakpoints.xs */";
```
marcysutton, line 4: I really don't like having to hard code these values in here. But I tried _many_ different approaches to typing the tokens and this was the only thing that worked!

packages/wonder-blocks-tokens/src/tokens/media-queries.ts
```diff
@@ -0,0 +1,31 @@
...
+export const pureSmMin = "568px";
+export const pureSmMax = "681px";
+export const pureMdMin = "682px";
+export const pureMdMax = "1023px";
+export const pureLgMin = "1024px";
+
+export const mediaQueries = {
```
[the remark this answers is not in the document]
marcysutton: I renamed this export to `breakpoints` so it would be consistent with webapp, i.e. `breakpoints.sm`!

packages/wonder-blocks-tokens/src/tokens/media-queries.ts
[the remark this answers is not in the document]
marcysutton: Done!

packages/wonder-blocks-tokens/src/tokens/media-queries.ts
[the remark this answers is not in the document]
jandrade: Actually, these are technically media queries, and the tokens in `pureWidth` are closer to breakpoints.

FWIW we have use the term `mediaQueries` in `webapp` as well.

https://github.com/Khan/webapp/blob/251ca8abf4207012ddf63ff0b2430f3445277c3e/services/static/javascript/shared-styles-package/media-queries.ts#L47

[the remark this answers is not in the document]
marcysutton: I guess my reasoning was that there are other types of media queries, not just widths. So if we wanted to also support other types of media query utilities, breakpoints would be a good name for this.

packages/wonder-blocks-tokens/src/tokens/media-queries.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+export const pureWidths = {
+    xsMax: 567,
+    smMin: 568,
+    smMax: 681,
+    mdMin: 682,
+    mdMax: 1023,
+    lgMin: 1024,
+} as const;
```
jandrade, line 15: **suggestion:** I feel that the `pureWidths` naming is too open. It would be better if we could give it a name that relates to media queries / breakpoints.

Related to your comment about media queries vs breakpoints, how do you feel about exporting this as an object that contains media query breakpoints and width breakpoints (or something like that)?

e.g.
```
export const breakpoint = {
  width: pureWidths,
  mediaQuery (or mq): the object with all the mq breakpoints
};
```

We use this approach with other tokens already, so this could bring some consistency to the token naming structure.

packages/wonder-blocks-tokens/src/tokens/media-queries.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const breakpoints = {
```
jandrade, line 17: **suggestion:** We tend to use singular for defining tokens, so in this case it would be better something like `breakpoint`, `mediaQuery` or `mq`.

You can see some existing examples here:
https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-tokens/src/tokens/border.ts
https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-tokens/src/tokens/typography.ts

types/aphrodite.d.ts
jandrade: Agreed, that's one of the things where TS could improve for sure!

jandrade (commented):
Thanks for working through this. I'd like to get your thoughts on some suggestions to get this breakpoint tokens more in line with the other WB tokens.

packages/wonder-blocks-tokens/src/tokens/media-queries.ts
marcysutton: Works for me, I love objects! 😂

packages/wonder-blocks-tokens/src/tokens/media-queries.ts
marcysutton: Done!

packages/wonder-blocks-tokens/src/tokens/media-queries.ts
```diff
@@ -0,0 +1,38 @@
...
+    xsMax: 567,
+    smMin: 568,
+    smMax: 681,
+    mdMin: 682,
+    mdMax: 1023,
+    lgMin: 1024,
+} as const;
+
+const mediaQuery = {
+    // Named
```
kevinb-khan, line 18: There should probably be a comment here telling people if they make changes here they need to also update aphrodite.d.ts.

types/aphrodite.d.ts
```diff
@@ -1,6 +1,31 @@
 declare module "aphrodite" {
     import * as React from "react";
 
+    const xs =
+        "@media screen and (max-width: 567px) /* breakpoint.mediaQuery.xs */";
+    const sm =
+        "@media screen and (min-width: 568px) and (max-width: 681px) /* breakpoint.mediaQuery.sm */";
```
kevinb-khan, line 7: There should probably be a comment here telling people if they make changes here they also need to update tokens/media-queries.ts.

kevinb-khan (commented):
One thing you may want to consider for a potential future PR is using code generation to keep tokens/media-queries.ts and types/aphrodite.d.ts.

types/aphrodite.d.ts
marcysutton: Good call!

packages/wonder-blocks-tokens/src/tokens/media-queries.ts
```diff
@@ -0,0 +1,38 @@
...
+    xs: `@media screen and (max-width: ${width.xsMax}px) /* breakpoint.mediaQuery.xs */`,
```
jandrade, line 19: Is `/* breakpoint.mediaQuery.xs */` useful for IDE auto completion or something else?

I was looking at the Docs and that info appears duplicated in the table.

Screenshot 2024-11-04 at 12 02 33 PM

Again, if this is helpful for DX reasons, you can surely leave it as it is.

marcysutton:
> One thing you may want to consider for a potential future PR is using code generation to keep tokens/media-queries.ts and types/aphrodite.d.ts.

@kevinb-khan good idea. Do you know of an example of that being done in Wonder Blocks? Could the media-queries.ts file be the source of truth to generate the type definition, or would it have to be the other way around? Some of this will need to change when we move away from Aphrodite, anyway.

jandrade (approved):
Looks great! thank you so much for implementing the changes and iterating on the Types 🚀

__docs__/wonder-blocks-tokens/tokens-media-queries.mdx
```diff
@@ -0,0 +1,76 @@
...
+import {breakpoint} from "@khanacademy/wonder-blocks-tokens";
+const styles = {
+    [breakpoint.mediaQuery.sm]: {
+        flexDirection: "column"
+    }
+};
+```
+
+### Pure widths
+
```
jandrade, line 38: **suggestion:** It would be nice adding a short paragraph explaining why this could be useful.

packages/wonder-blocks-tokens/src/tokens/media-queries.ts
marcysutton: Good question! It was a trick to avoid conflicts for typing of strings. I included it based on this comment in webapp from @kevinb-khan: https://github.com/Khan/webapp/blob/55c4465017a829d53bdc6d00d94d6d93e3b5dfe4/services/static/javascript/shared-styles-package/media-queries.ts#L26

> We attach a unique comment string to each of the media queries in this file. That way, if another media query has the same constraints as one of these queries, their strings won't be identical, and they therefore won't conflict.

Admittedly, the typing for this was really hard to get right and I still wasn't that happy with it in the end. It would be one of the first things I'd want to refactor for the Aphrodite successor.

__docs__/wonder-blocks-tokens/tokens-media-queries.mdx
marcysutton: Done!

__docs__/wonder-blocks-tokens/tokens-media-queries.mdx
```diff
@@ -0,0 +1,82 @@
...
+/>
+
+<TokenTable
+    columns={[
+        {
+            label: "Width Token",
+            cell: (row) => <code>breakpoint.width.{row.label}</code>,
+        },
+        {
+            label: "Value",
```
beaesguerra, line 77: Since the values are numbers now instead of a string with "px", we can specify the unit in the column label! 

```suggestion
            label: "Value (px)",
```

beaesguerra (approved):
Looks great Marcy! :)

PR #2350, opened 2024-10-23
Enable eslint-plugin-jsx-a11y rules

beaesguerra:
## Summary:
As part of the hackathon project for accessibility automation, one of the ideas explored is how we can use linting rules from eslint-plugin-jsx-a11y to catch some accessibility errors. We already had the plugin installed in Wonder Blocks, though we didn't have it configured in our eslint config yet!

These changes include: 
- Updating to the latest eslint-plugin-jsx-a11y version
- Using the `eslint-plugin-jsx-a11y` plugin
- Extending the strict jsx-a11y linting rules 
- Addressing some jsx-a11y lint errors
  - Added eslint disable comments for focus trap and focus manager components (created WB-1788 and WB-1789 for further investigation)
- Using the same plugin config/rules proposed in the [ADR](https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/3394600994/ADR+781+Enabling+more+lint+rules+for+accessibility)
- [ ] More lint errors need to be addressed now that the updated config was able to detect more issues

By enabling these rules, future changes to Wonder Blocks will be checked against these rules!

## Related Notes:
- PR for enabling more a11y rules in webapp: https://github.com/Khan/webapp/pull/26561
- PR for disabling a11y errors in webapp: https://github.com/Khan/webapp/pull/26840

Issue: XXX-XXXX

## Test plan:

### DIFF
--- a/.changeset/beige-laws-shop.md
+++ b/.changeset/beige-laws-shop.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/wonder-blocks-popover": patch
+"@khanacademy/wonder-blocks-modal": patch
+---
+
+Update eslint tooling
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,10 +7,12 @@ module.exports = {
         // This config includes rules from storybook to enforce story best
         // practices
         "plugin:storybook/recommended",
+        "plugin:jsx-a11y/recommended",
     ],
     plugins: [
         "import",
         "jest",
+        "jsx-a11y",
         "jsdoc",
         "promise",
         "monorepo",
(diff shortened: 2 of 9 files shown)


marcysutton:
This looks great, @beaesguerra! I tested it out locally and it's working well with the additional rules. Seems like a great change to me!

beaesguerra:
@marcysutton Thanks for taking an early look! I updated the eslint config some more so it's the same as the proposed config for webapp. It found more errors which I need to address still!

beaesguerra:
Closing this PR for now, will open a new one to enable a11y rules in Wonder Blocks for WB-229. We will add the a11y rules to the shared Khan eslint config and will extend that in WB, see [ADR#781](https://khanacademy.atlassian.net/wiki/x/IoBVyg) for more details!

PR #2352, opened 2024-10-24
PoC: Add `makeStyled("tag")` that provides an object with named component (i.e., StyledSpan)

nedredmond:
@beaesguerra is exploring using more a11y rules in eslint, but there is an issue-- `addStyle` allows components to have any arbitrary name, so it's impossible to map native elements to custom components.

We can enforce consistent naming by using some newer TS features to force return types to have specific names, like `StyledSpan` or `StyledDiv`.

For example, `makeStyled("div")` yields `{StyledDiv}`!

We can deprecate `addStyle` in favor of `makeDiv` and migrate uses in `webapp` in order to allow eslint configs to be effective.

### DIFF
--- a/packages/wonder-blocks-core/src/util/add-style.tsx
+++ b/packages/wonder-blocks-core/src/util/add-style.tsx
@@ -5,15 +5,46 @@ import {processStyleList} from "./util";
 
 import type {StyleType} from "./types";
 
+type StyledProps = {
+    className?: string;
+    style?: StyleType;
+    children?: React.ReactNode;
+} & Omit<React.ComponentProps<"div">, "style">; // Removes the 'style' prop from the original component
+
+type StyledElement<
+    T extends keyof JSX.IntrinsicElements,
+    Props extends StyledProps,
+> = React.ForwardRefExoticComponent<
+    React.PropsWithoutRef<Props> & React.RefAttributes<IntrinsicElementsMap[T]>
+>;
+
+type NamedStyledType<
+    T extends keyof JSX.IntrinsicElements,
+    Props extends StyledProps,
+> = {
+    [K in T as `Styled${Capitalize<T & string>}`]: StyledElement<T, Props>;
+};
+
+export const makeStyled = <
+    T extends keyof JSX.IntrinsicElements,
+    Props extends StyledProps,
+>(
+    Component: T,
+    defaultStyle?: StyleType,
+): NamedStyledType<T, Props> => {
+    const styledComponentName = `Styled${Component.charAt(
+        0,
+    ).toUpperCase()}${Component.slice(1)}` as `Styled${Capitalize<T & string>}`;
+    return {
+        [styledComponentName]: addStyle(Component, defaultStyle),
+    } as NamedStyledType<T, Props>;
+};
+
 export default function addStyle<
     // We extend `React.ComponentType<any>` to support `addStyle(Link)` with
     // react-router's `Link` component.
     T extends React.ComponentType<any> | keyof JSX.IntrinsicElements,
-    Props extends {
-        className?: string;
-        style?: StyleType;
-        children?: React.ReactNode;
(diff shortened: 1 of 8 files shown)


packages/wonder-blocks-core/src/util/add-style.tsx
```diff
@@ -5,6 +5,40 @@ import {processStyleList} from "./util";
 
 import type {StyleType} from "./types";
 
+type StyledProps<T extends keyof JSX.IntrinsicElements> = {
+    className?: string;
+    style?: StyleType;
+    children?: React.ReactNode;
+} & Omit<React.ComponentProps<T>, "style">;
+
+interface StyledTag<T extends keyof JSX.IntrinsicElements> {
```
marcysutton, line 14: Some mind-bending typing here! Can you add code comments to help consumers understand what it does?

nedredmond: I commente the function itself below since that's the only exported part -- these types wouldn't have any other consumers, but I can certainly comment for other devs' edification!

marcysutton: That's a good point! I suppose the workflow I'm thinking of is when devs dig in more to understand what the code does. Thanks @nedredmond!

nedredmond: @marcysutton Do you think I should prepare this for a real merge? I made it a draft as an illustration of a solution but I'll add a `changeset` if you'd like to pull it into wonder-blocks. It _will_ mark `addStyle` as deprecated!

kevinb-khan (commented):
Interesting way of enforcing a coding style rule.  I probably would've created an eslint for this, but this is so much easier to implement than that would be.  Nice work!

nedredmond:
@kevinbarabash please see the discussion about this solution [here](https://khanacademy.slack.com/archives/C4PE1QM5Y/p1730295908289759); @jandrade believes this is not a simpler solution than eslint.

Should I close this?

jandrade:
> @kevinbarabash please see the discussion about this solution [here](https://khanacademy.slack.com/archives/C4PE1QM5Y/p1730295908289759); @jandrade believes this is not a simpler solution than eslint.
> 
> Should I close this?

@nedredmond I'd like to clarify my thinking in that thread. I don't believe that writing an ESLint rule is easier, it is a different level of abstraction and I agree that we would need to share that rule in different repos (adding complexity). In that regard, using TypeScript is better in terms of simplicity.

Regarding simplicity in solving the problem (that is being able to map existing `addStyle` calls to force a particular naming convention), I was referring to just rename the existing `addStyle` call sites to use the naming convention, and optionally write the ESLint rule to enforce that. I think that is optional because we are going to deprecate/sunset `addStyle` when we migrate to a different CSS approach.

My main thinking around using this approach vs leaving `addStyle` intact is API design. The goal for this function is to return a single JSX component, and returning an object, even if it is to enforce naming, it is not strictly correct from the API perspective. I feel that this adds a different level of complexity due to the function now returning an object instead of a single reference. This helps enforcing naming, but it is not bulletproof, as for example, you can still use the object as the identifier (instead of destructuring) and that would change the mapping defined for the ESLint a11y plugin:

```
const StyledComponent = makeStyled("li");
<StyledComponent.StyledLi />
```

I have to add that it's a really neat approach as it offers all the TS benefits like type checking and autocompletion, and it's great that it's so flexible, but IMO this approach works when having to deal with more complex return types, or when it makes sense to return an object and enforce certain patterns. I would really love that there was a way to enforce identifier namings in TS (not associated to objects).

My recommendation would be to close this PR, but I'd like to know if other folks have a different opinion on this.

@nedredmond as @marcysutton mentioned you are really a wizard! and regardless of the decision, please keep experimenting and proposing these kind of ideas in our codebases.

PR #2354, opened 2024-11-05
Update label in SingleSelect and MultiSelect

daniellewhyte:
## Summary:
When you pass in a JSX Element as a label to `OptionItem`, the SelectOpener is labeled with an empty string. This PR updates SelectOpener in the `SingleSelect` and `MultiSelect` components to return the JSX as a label in that case. 

This change is being made to unblock supporting TEX in the Perseus Dropdown widget. 

Issue: LIT-1425

## Test plan:
- Added new stories
  - https://5e1bf4b385e3fb0020b7073c-xhxyrfwkfd.chromatic.com/?path=/story/packages-dropdown-singleselect--custom-option-item-with-node-label
  - https://5e1bf4b385e3fb0020b7073c-xhxyrfwkfd.chromatic.com/?path=/story/packages-dropdown-multiselect--custom-option-items-with-node-label
- Unit tests pass
- Installed npm snapshot and tested against my branch in Perseus ([PR here](https://github.com/Khan/perseus/pull/1810))
- Test in webapp and storybook to ensure no regressions


daniellewhyte:
@Khan/wonder-blocks This is now ready for review.

__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
```diff
@@ -650,3 +655,51 @@ export const CustomOptionItems: StoryComponentType = {
         ),
     ],
 };
+
+/**
+ * This example illustrates how a JSX Element can appear as the label when only
+ * one option is selected and `labelAsText` is undefined.
+ * **Note** this is only supported for SingleSelect and MultiSelect, not Combobox.
+ */
+export const CustomOptionItemsWithoutTextLabel: StoryComponentType = {
```
marcysutton, line 664: Maybe it's just me, but I found the "without text label" part of these new stories a little confusing! 😅  

Would something like "with node label" or "with child element as label" reflect the unique aspect of these updates?

__docs__/wonder-blocks-dropdown/single-select.stories.tsx
```diff
@@ -884,6 +888,55 @@ export const CustomOptionItems: StoryComponentType = {
     },
 };
 
+/**
+ * This example illustrates how a JSX Element can appear as the label if
+ * `labelAsText` is undefined.
+ * **Note** this is only supported for SingleSelect and MultiSelect, not Combobox.
+ */
+export const CustomOptionItemWithoutTextLabel: StoryComponentType = {
```
marcysutton, line 896: Same as above. Would renaming these stories with something like "With element child as label" work for you? It would make it easier to understand them when scanning through Storybook.

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
```diff
@@ -150,7 +150,7 @@ describe("MultiSelect", () => {
 
             // Assert
             expect(await screen.findByRole("button")).toHaveTextContent(
-                "1 student",
```
marcysutton, line 153: Wow, that's a pretty confusing test result looking at this older code. Where did "student" even come from? Your change seems way more intuitive.

__docs__/wonder-blocks-dropdown/single-select.stories.tsx
```diff
@@ -884,6 +888,55 @@ export const CustomOptionItems: StoryComponentType = {
...
+                    onToggle={handleToggle}
+                    opened={opened}
+                >
+                    {currencies.map((currency, index) => (
+                        <OptionItem
+                            key={index}
+                            value={String(index)}
+                            horizontalRule="full-width"
+                            label={
+                                <span>
```
marcysutton, line 924: This tested well in VoiceOver! I couldn't get my Windows VM to load the local Storybook environment for NVDA but I'm pretty sure it will be fine.

marcysutton (commented):
This is looking solid to me but I'd be curious to hear what @jandrade and @beaesguerra have to say. These components are pretty complex!

Great job, @daniellewhyte!

__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
daniellewhyte: Yes, good suggestion! just pushed an update

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
beaesguerra: I was wondering the same thing! Looks like "1 student" was coming from the `defaultLabels` that are passed in!
https://github.com/Khan/wonder-blocks/blob/5eedd04c9ea17bdf7141efb38ed144475da0ad4b/packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx#L28-L35

__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
```diff
@@ -650,3 +655,51 @@ export const CustomOptionItems: StoryComponentType = {
...
+        const handleChange = (selectedValues: Array<string>) => {
+            setSelectedValues(selectedValues);
+        };
+
+        const handleToggle = (opened: boolean) => {
+            setOpened(opened);
+        };
+
+        return (
+            <MultiSelect
```
beaesguerra, line 680: I played around with the changes and found if we set `isFilterable` in this story, the filtering functionality doesn't work as expected! However, if we add a `labelAsText` prop to the `OptionItem` components with the content (ie. the `locale` in this example), filtering works as expected. 

I'm wondering, should we use `labelAsText` prop when we use a node label? And if so, would that solve for the original issue where the select opener is labeled with an empty string? Let me know though if there's some context I'm missing!

cc: @jandrade , I think he was running into similar things when he worked on the combobox!

beaesguerra (commented):
Thanks for looking into this, @daniellewhyte 😄 

I noticed a bug when we set `isFilterable` in the `CustomOptionItemsWithNodeLabel` story and had follow up questions!

__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
daniellewhyte: That's a good callout, and the proposed solution does work solve the original issue.

daniellewhyte: @beaesguerra Actually I don't think this works. If you add `labelAsText` to the option item, you see plain text as the label for the menu. 
Screenshot 2024-11-13 at 9 07 33 AM

That said, my specific use case does not use filtering.

jandrade: I understand that the specific use case in this PR doesn't need filtering, but I worry that adding this feature while introducing a known issue might be counterproductive if there's another call site that needs both things (custom options + filtering).

I'd really recommend applying the fix/change in this PR before landing this change.

I was thinking that it might be solved by changing the filter logic to use `.toString()` after the `getSelectOpenerLabel` invocation. I'm not fully sure if this is the right solution but it might be worth trying it out for both `MultiSelect` and `SingleSelect`.

jandrade (changes requested):
This is looking great, thanks! I'm requesting changes to include the new helper in the filtering feature to prevent regressions in the future.

Let me know if you have more questions, happy to pair and chat about it.

daniellewhyte:
Reopened PR using an approached suggested by Bea when we huddled a while back. Some context here: https://khanacademy.slack.com/archives/C4PE1QM5Y/p1732564950057359

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -91,6 +91,13 @@ type DefaultProps = Readonly<{
      * Whether to display shortcuts for Select All and Select None.
      */
     shortcuts: boolean;
+    /**
+     * When false, the SelectOpener can show a Node as a label. When true, the
+     * SelectOpener will use a string as a label. If using custom OptionItems, a
+     * plain text label can be provided with the `labelAsText` prop.
+     * Defaults to true.
+     */
+    showLabelAsText: boolean;
```
beaesguerra, line 100: super minor: What do you think about naming the prop `showOpenerLabelAsText` so it helps communicate that it's used for the opener? (same for SingleSelect!)

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -227,6 +234,7 @@ export default class MultiSelect extends React.Component<Props, State> {
         light: false,
         shortcuts: false,
         selectedValues: [],
+        showLabelAsText: true,
```
beaesguerra, line 237: 👍 Great idea defaulting to true for existing usage!

.changeset/thirty-shirts-leave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-dropdown": patch
```
beaesguerra, line 2: This will be a `minor` update since a new prop was added!

```suggestion
"@khanacademy/wonder-blocks-dropdown": minor
```

beaesguerra (commented):
This is working great, thanks for the fix 🎉  custom node is shown for the selected value and filtering works as expected with the option `labelAsText` prop 😄

Left a few minor comments around naming and versioning!

beaesguerra (approved):
Looks great! Thanks for working on this improvement 🎉 🚀

jandrade (approved):
This looks great! thanks Danielle for iterating on this and trying out different solutions 🚀

PR #2356, opened 2024-11-06
Refactor TextField to be a function component

beaesguerra:
## Summary:
Refactor TextField to be a function component. 

This will make it easier to share validation logic from the TextArea instantValidation PR https://github.com/Khan/wonder-blocks/pull/2355 so that TextField will also have the `instantValidation` prop functionality (I'll be working on this next)!

Issue: WB-1781

## Test plan:
Tests pass and TextField looks and functions the same way (`/?path=/docs/packages-form-textfield--docs`)

### DIFF
--- a/packages/wonder-blocks-form/src/components/text-field.tsx
+++ b/packages/wonder-blocks-form/src/components/text-field.tsx
@@ -48,7 +48,7 @@ type CommonProps = AriaProps & {
      * This `disabled` prop will also set the `readonly` attribute to prevent
      * typing in the field.
      */
-    disabled: boolean;
+    disabled?: boolean;
     /**
      * Provide a validation for the input value.
      * Return a string error message or null | void for a valid input.
@@ -114,7 +114,7 @@ type CommonProps = AriaProps & {
     /**
      * Change the default focus ring color to fit a dark background.
      */
-    light: boolean;
+    light?: boolean;
     /**
      * Custom styles for the input.
      */
@@ -138,7 +138,7 @@ type CommonProps = AriaProps & {
 };
 
 type OtherInputProps = CommonProps & {
-    type: "text" | "password" | "email" | "tel";
+    type?: "text" | "password" | "email" | "tel";
 };
 
 // Props that are only available for inputs of type "number".
--- a/packages/wonder-blocks-form/src/components/text-field.tsx
+++ b/packages/wonder-blocks-form/src/components/text-field.tsx
@@ -204,7 +204,7 @@ const TextField = (props: PropsWithForwardRef) => {
         }
     });
 
-    const maybeValidate: (newValue: string) => void = (newValue) => {
+    const maybeValidate = (newValue: string) => {
         if (validate) {
             const maybeError = validate(newValue) || null;
             setErrorMessage(maybeError);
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-form/src/components/text-field.tsx
```diff
@@ -44,7 +48,7 @@ type CommonProps = AriaProps & {
      * This `disabled` prop will also set the `readonly` attribute to prevent
      * typing in the field.
      */
-    disabled: boolean;
+    disabled?: boolean;
```
beaesguerra, line 51: I made `disabled`, `light`, and `type` props optional to address typescript errors: https://github.com/Khan/wonder-blocks/actions/runs/11713064887/job/32625094179

These [3 props used to have default props defined](https://github.com/Khan/wonder-blocks/pull/2356/files#diff-3f6913d896b8a5fb2855cf22cd787696ab51bb9833024d46344da4b4de3dbeccL179-L183) in the class component. I'm curious, how did the types work before if they weren't marked as optional? 🤔

packages/wonder-blocks-form/src/components/text-field.tsx
```diff
@@ -159,100 +163,85 @@ type FullNumericInputProps = CommonProps & NumericInputProps;
...
+        onKeyDown,
+        onChange,
+        onFocus,
+        onBlur,
+        // Should only include Aria related props
+        ...otherProps
+    } = props;
+    // Ensures error is updated on unmounted server-side renders
+    const [errorMessage, setErrorMessage] = React.useState(
+        (props.validate && props.value !== "" && props.validate(props.value)) ||
```
beaesguerra, line 197: This is something TextArea doesn't do currently, but can be something I add when we share validation logic between TextField and TextArea (if it makes sense to keep)! Currently TextArea [initializes the errorMessage to null](https://github.com/Khan/wonder-blocks/blob/e1c98df51ec152dbe80bd153d0e980caab436083/packages/wonder-blocks-form/src/components/text-area.tsx#L218)

For my own understanding: When I was converting this component, this logic is equivalent to setting the state errorMessage in the [constructor](https://github.com/Khan/wonder-blocks/pull/2356/files#diff-3f6913d896b8a5fb2855cf22cd787696ab51bb9833024d46344da4b4de3dbeccL187-L189), while the [useOnMountEffect](https://github.com/Khan/wonder-blocks/pull/2356/files#diff-3f6913d896b8a5fb2855cf22cd787696ab51bb9833024d46344da4b4de3dbeccR202-R204) logic is equivalent to the [componentDidMount](https://github.com/Khan/wonder-blocks/pull/2356/files#diff-3f6913d896b8a5fb2855cf22cd787696ab51bb9833024d46344da4b4de3dbeccL197-L201) logic. Are both needed? What is the use case for unmounted server-side renders?

jandrade: This was probably added for preventing hydration issues (SSR not being in sync with CSR). Here's more context about that: https://github.com/Khan/wonder-blocks/pull/1131#discussion_r643972638

jandrade: Note that we don't use SSR right now, but this would be turned on in the future, so maybe that's a feature we should consider?

jandrade (approved):
Looks good! thanks for the cleanup :shipit:

packages/wonder-blocks-form/src/components/text-field.tsx
beaesguerra: Thanks for the context! I will include it in the shared validation logic!

marcysutton (approved):
Great work, Bea!

packages/wonder-blocks-form/src/components/text-field.tsx
kevinb-khan: I believe the TypeScript types for class-based components combine the type of the props passed to `React.Component<>` with the inferred types for `static defaultProps` to determine what the final props should be for the component.

kevinb-khan: I created a TS playground illustrating the difference in types when there is (or isn't) a static `defaultProps` field: https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcARFDvmQNwBQdMAnmFnAArFjoC8cA3nTjC4BCBABccVDCjAAdgHMANEJEAjZFCnyAriHVYoqgL706uADZp0AYRJgAjHCwAPGFnkATdNjwwAOntwCHlPGAAeTghuAD4BNWEZZBhgXDgvLAJkXUsYaO44PkERUtFxKTIACyxLSwgyVVKTRLhKbyMACgBKBLKRShhdKHk4PTr6ZroWi2tUOwcAJhd3Tx9MKkDgyDD5SILUeJKBta7e4-7B4dHxy0mRFpncUJkUIrgI7edNKB5+AFYTHAAPSxejPeSvdTvT4OZxiCB-MgAd2gli8ZCBPz+gJBYIYENerhh22W2IBQNB4Je8CYJKW5UR-BRaIxWK0OMp+KAA.

PR #2357, opened 2024-11-08
Add `instantValidation` prop for TextField & share logic with TextArea

beaesguerra:
## Summary:
- Adds `instantValidation` prop to `TextField` component
- Create `useFieldValidation` hook to share validation logic between `TextField` and `TextArea`
- `TextArea` is updated to use `useFieldValidation` hook.
- Both `TextArea` and `TextField` will now validate on initialization if `value` is not empty and it is not disabled
- Stories: Use utility functions for validate functions

Issue: WB-1781

## Test plan:
- TextField
  - Instant Validation docs are reviewed (`/?path=/docs/packages-form-textfield--docs#instant%20validation`)
  - Instant Validation works as expected (see docs for expected behaviour) (`/?path=/story/packages-form-textfield--instant-validation`)
- TextArea
  - Instant Validation docs are reviewed (`/?path=/docs/packages-form-textarea--docs#instant%20validation`)
  - Instant Validation works as expected (see docs for expected behaviour) (`/?path=/story/packages-form-textarea--instant-validation`)

### DIFF
--- a/__docs__/wonder-blocks-form/form-utilities.ts
+++ b/__docs__/wonder-blocks-form/form-utilities.ts
@@ -15,7 +15,6 @@ export const validateEmail = (value: string) => {
  * @param value the phone number to validate
  * @returns An error message if there is a validation error
  */
-
 export const validatePhoneNumber = (value: string) => {
     const telRegex = /^\(?([0-9]{3})\)?[-. ]?([0-9]{3})[-. ]?([0-9]{4})$/;
     if (!telRegex.test(value)) {
--- a/__docs__/wonder-blocks-form/form-utilities.ts
+++ b/__docs__/wonder-blocks-form/form-utilities.ts
@@ -1,0 +1,24 @@
+/**
+ * Checks if a value is a valid email and returns an error message if it is invalid.
+ * @param value the email to validate
+ * @returns An error message if there is a validation error
+ */
+export const validateEmail = (value: string) => {
+    const emailRegex = /^[^@\s]+@[^@\s.]+\.[^@.\s]+$/;
+    if (!emailRegex.test(value)) {
+        return "Please enter a valid email";
+    }
+};
+
+/**
+ * Checks if a value is a valid phone number and returns an error message if it is invalid.
+ * @param value the phone number to validate
+ * @returns An error message if there is a validation error
+ */
+
+export const validatePhoneNumber = (value: string) => {
+    const telRegex = /^\(?([0-9]{3})\)?[-. ]?([0-9]{3})[-. ]?([0-9]{4})$/;
+    if (!telRegex.test(value)) {
+        return "Invalid US telephone number";
+    }
+};
--- a/__docs__/wonder-blocks-form/text-field.argtypes.ts
+++ b/__docs__/wonder-blocks-form/text-field.argtypes.ts
@@ -187,6 +187,19 @@ export default {
         },
     },
 
+    instantValidation: {
+        description:
+            "If true, TextField is validated as the user types (onChange). If false, it is validated when the user's focus moves out of the field (onBlur). It is preferred that instantValidation is set to `false`, however, it defaults to `true` for backwards compatibility with existing implementations.",
+        table: {
+            type: {
+                summary: "boolean",
+            },
+        },
+        control: {
+            type: "boolean",
+        },
+    },
+
     /**
      * Number-specific props
      */
--- a/__docs__/wonder-blocks-form/text-field.stories.tsx
+++ b/__docs__/wonder-blocks-form/text-field.stories.tsx
@@ -398,6 +398,32 @@ Telephone.parameters = {
     },
 };
 
+const ControlledTextField = (args: PropsFor<typeof TextField>) => {
+    const [value, setValue] = React.useState(args.value || "");
+    const [error, setError] = React.useState<string | null | undefined>(null);
+
+    const handleChange = (newValue: string) => {
+        setValue(newValue);
+    };
+
+    return (
+        <View>
+            <TextField
+                {...args}
+                value={value}
+                onChange={handleChange}
+                onValidate={setError}
+            />
+            <Strut size={spacing.xxSmall_6} />
+            {(error || args.error) && (
+                <LabelSmall style={styles.errorMessage}>
+                    {error || "Error from error prop"}
+                </LabelSmall>
+            )}
+        </View>
+    );
+};
+
 function ErrorRender(args: PropsFor<typeof TextField>) {
     const [value, setValue] = React.useState("khan");
     const [errorMessage, setErrorMessage] = React.useState<any>();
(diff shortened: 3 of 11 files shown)


__docs__/wonder-blocks-form/form-utilities.ts
```diff
@@ -0,0 +1,24 @@
+/**
```
beaesguerra, line 1: Adding validate utilities for the form components, thanks for the suggestion! https://github.com/Khan/wonder-blocks/pull/2355#discussion_r1833554428

packages/wonder-blocks-form/src/components/__tests__/text-area.test.tsx
```diff
@@ -977,8 +977,9 @@ describe("TextArea", () => {
                 expect(textArea).toHaveAttribute("aria-invalid", "false");
             });
 
-            it("should call the validate function when it is first rendered", async () => {
+            it("should call the validate function twice when it is first rendered (once on initialization, once after mount)", async () => {
```
beaesguerra, line 980: The updated and new TextArea tests are to address the new logic where the [errorMessage state is initialized using the validate function](https://github.com/Khan/wonder-blocks/pull/2357/files#diff-0766026139a14b159f212620adffa729b119cb7059eed30b3917aac1ea913875R72-R78)

__docs__/wonder-blocks-form/text-field.argtypes.ts
```diff
@@ -187,6 +187,19 @@ export default {
         },
     },
 
+    instantValidation: {
```
beaesguerra, line 190: Wasn't able to get the autodocs working with TextField, we have WB-1765 to remove prop descriptions from *.argtypes.ts files!

packages/wonder-blocks-form/src/components/__tests__/text-field.test.tsx
```diff
@@ -740,4 +746,901 @@ describe("TextField", () => {
         const input = await screen.findByRole("textbox");
         expect(input).toHaveAttribute("aria-invalid", "false");
     });
+
+    describe("Validation", () => {
```
beaesguerra, line 750: These validation tests are the same as the validation tests added for TextArea in [PR#2344](https://github.com/Khan/wonder-blocks/pull/2355/files#diff-0b9f5a7a93ed66290b63ac82420db318d5dafd407741b5013def25790bedffd7), but renamed to work with TextField.

I was thinking about whether we needed these tests + the 1TextArea1 tests + the 1useFieldValidation1 tests. I think it doesn't hurt to verify all of these places, especially since there are lots of edge cases. One difference is `TextField` and `TextArea` will test for the `aria-invalid` attribute, while the `useFieldValidation` tests check the `errorMessage`

 Let me know though if anyone has thoughts on this!

beaesguerra:
This is the last PR for the TextField + TextArea updates! I'll merge + release the stacked PRs once this is approved :)

__docs__/wonder-blocks-form/form-utilities.ts
```diff
@@ -0,0 +1,24 @@
...
+        return "Please enter a valid email";
+    }
+};
+
+/**
+ * Checks if a value is a valid phone number and returns an error message if it is invalid.
+ * @param value the phone number to validate
+ * @returns An error message if there is a validation error
+ */
+
```
jandrade, line 18: **nit(really super nit):** There's an extra line here. I wouldn't bother about this, but I'm not sure if that extra line messes up with the info that we would see in the IDE (when trying to call this fn).

__docs__/wonder-blocks-form/text-field.argtypes.ts
jandrade: hmm yeah that's probably because `TextField` uses union types, which is a react-docgen limitation as it cannot infer those kind of complex prop types.

For reference: https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-form/src/components/text-field.tsx#L149

__docs__/wonder-blocks-form/text-field.stories.tsx
```diff
@@ -398,6 +385,32 @@ Telephone.parameters = {
...
+const ControlledTextField = (args: PropsFor<typeof TextField>) => {
+    const [value, setValue] = React.useState(args.value || "");
+    const [error, setError] = React.useState<string | null | undefined>(null);
+
+    const handleChange = (newValue: string) => {
+        setValue(newValue);
+    };
+
+    return (
+        <View>
```
jandrade, line 397: **question:** Is this `View` needed for layout purposes? Not sure if it can be removed, but it's fine if you find it useful for a particular reason.

packages/wonder-blocks-form/src/components/__tests__/text-area.test.tsx
```diff
@@ -977,8 +977,9 @@ describe("TextArea", () => {
...
                 // Arrange
+                // Act
                 const validate = jest.fn();
```
jandrade, line 983: **nit:** This mock would go in the `// Arrange` section instead.

For context: https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/98402353/Unit+Testing+Best+Practices#Arrange-Act-Assert

packages/wonder-blocks-form/src/components/__tests__/text-area.test.tsx
```diff
@@ -988,10 +989,48 @@ describe("TextArea", () => {
...
+                        onChange={() => {}}
+                        validate={() => errorMessage}
+                        onValidate={onValidate}
+                    />,
+                    defaultOptions,
+                );
 
                 // Assert
-                expect(validate).toHaveBeenCalledExactlyOnceWith("text");
+                expect(onValidate).toHaveBeenCalledExactlyOnceWith(
```
jandrade, line 1013: **praise:** TIL about `toHaveBeenCalledExactlyOnceWith`! I previously used `toHaveBeenCalledOnce` but it was removed from the API some time ago, so it's great to see that we can still use something like that :).

packages/wonder-blocks-form/src/components/__tests__/text-field.test.tsx
jandrade: Good question. I agree that increasing coverage is good in these cases. I feel that we are really trying to cover edge cases (not just following a coverage number), so keeping these tests make sense.

As for the `TextField` + `TextArea` tests, we have a `consistency-tests` folder that we use when we want to test similar behaviors for common components. I wonder if it would make sense to move these tests to that folder of if it's definitely better to keep them with the other tests. how do you feel about this? (tbh it's totally fine if we keep them separated)

cc @marcysutton in case you have opinions as well.

Reference: https://github.com/Khan/wonder-blocks/tree/main/consistency-tests/__tests__

packages/wonder-blocks-form/src/components/__tests__/text-field.test.tsx
```diff
@@ -740,4 +746,901 @@ describe("TextField", () => {
...
+        describe("validate prop", () => {
+            it("should be in an error state if the initial value is not empty and not valid", async () => {
+                // Arrange
+                render(
```
jandrade, line 754: **suggestion:** This would go in `Act` as it is the part of the test that triggers the assertion.

packages/wonder-blocks-form/src/hooks/use-field-validation.ts
```diff
@@ -0,0 +1,146 @@
...
+        if (disabled) {
+            return;
+        }
+        if (validate) {
+            const error = validate(newValue) || null;
+            setErrorMessage(error);
+            if (onValidate) {
+                onValidate(error);
+            }
+        } else if (required) {
```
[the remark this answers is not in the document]
jandrade: Agreed, early return would be really handy here.

Would it mean that WB-1794 would be used for integrating these changes in `webapp` later? I wonder if you have considered using a snapshot release to verify if those changes will affect `webapp` right now.

packages/wonder-blocks-form/src/hooks/use-field-validation.ts
jandrade, line 1: **praise:** Yesss!!!! 👏.

This made me think if we should consider exposing this to external consumers. I'm not sure if `webapp` could benefit from this hook, or if it would be better to keep this internally and suggest devs to continue using the WB components directly?

jandrade (commented):
This is incredible Bea 👏. I have some suggestions/questions for you to consider, and also I'd like to ask a couple of questions to get your thougths:

1. How do we feel about exposing `useFieldValidation` to consumers?
2. Do you think this hook could be reused by other WB form elements? If that's the case, what would you think about moving this hook to a more central place so the other WB packages wouldn't have to import `wb-form`? (maybe it makes sense to keep it here, but I'd like to hear your thoughts about it).

packages/wonder-blocks-form/src/components/__tests__/text-field.test.tsx
beaesguerra: Oh cool, I didn't know we had consistency tests in WB!

I think for this case, since this validation logic is specific to these 2 components, we can keep them separated. If we were to have a third component with the same validation logic, then I think we could move the common tests there!

__docs__/wonder-blocks-form/form-utilities.ts
beaesguerra: Updated! VS code was still showing the docs with the new line, but cleaned it up anyways!

__docs__/wonder-blocks-form/text-field.stories.tsx
beaesguerra: Without it, the `Strut` doesn't provide spacing as expected!

packages/wonder-blocks-form/src/components/__tests__/text-field.test.tsx
beaesguerra: I cleaned up the Arrange/Act comments! I see now the initial render can be both the Arrange and Act part, thanks for letting me know!

packages/wonder-blocks-form/src/hooks/use-field-validation.ts
beaesguerra: Good question! I'm not sure if there's a use case for using the hook in webapp so for now the hook can be used internally and devs should use the WB components! If there's a need for it, we could expose it externally!

> Do you think this hook could be reused by other WB form elements? If that's the case, what would you think about moving this hook to a more central place so the other WB packages wouldn't have to import wb-form? (maybe it makes sense to keep it here, but I'd like to hear your thoughts about it).

Another good question! I don't think other WB components would use this hook, it is pretty specific to input and textarea elements. For example, SingleSelect and MultiSelect will have different validation logic since it won't support the `instantValidation` prop. I would like to avoid making the hook super flexible since there would be more edge cases to support! 

Let me know what you think or if there are some places you think it would be helpful!

packages/wonder-blocks-form/src/hooks/use-field-validation.ts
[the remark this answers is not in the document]
beaesguerra: I think WB-1794 can be worked on separately since it's a pre-existing bug in both TextField and TextArea! I plan to release and integrate this current stack of PRs after this PR (it's adding up 😅). I was aiming to make only additive and backwards compatible changes to support the LabeledField work so that there's minimal changes to behaviour (with the exception to the [error initialization](https://github.com/Khan/wonder-blocks/pull/2357#discussion_r1835134581) since TextField already had this behaviour).

I ran the webapp unit tests with the snapshot of the current changes and things are passing! I'll also spot check some places manually before merging just to be sure! With WB-1794, I'd like to check more thoroughly in case there's some logic relying on the bug

packages/wonder-blocks-form/src/components/__tests__/text-field.test.tsx
jandrade: All great points! and agree on SearchField as well.

packages/wonder-blocks-form/src/hooks/use-field-validation.ts
jandrade: all of this make sense, thanks for the clarification :)

jandrade (approved):
Looks great to me! Thanks for addressing all my questions. Awesome work 🎉 🚢

PR #2358, opened 2024-11-13
PhosphorIcon: default role to img when aria-label is provided

SeanMcP:
## Summary:
When an `aria-label` is provided, there should be a `role` applied to the element to give it meaning to browsers and screenreaders. This PR defaults the `role` to `"img"` in `PhosphorIcon` when an `aria-label` is provided.

I'm not sure what our definitions are for "major", "minor", and "patch" releases. _I_ would consider this a `minor` release since we are changing the behavior (even if it is slight), but it is backwards compatible so `patch` might be fine.

Issue: WB-1795

## Test plan:
- Tests pass
- Axe is happy

### DIFF
--- a/.changeset/dirty-dragons-fix.md
+++ b/.changeset/dirty-dragons-fix.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-icon": minor
+---
+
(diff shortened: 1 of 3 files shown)


jeresig (approved):
Great - thank you! This does feel like a minor change to me, agreed.

PR #2361, opened 2024-11-15
Update `Toolbar` to use CSS grid

timmcca-be:
## Summary:
This PR makes it so elements passed to the `title` of a `Toolbar` can set `max-width: 100%` to ensure the title never overlaps `leftContent` or `rightContent`.

Issue: AX-315

## Test plan:
Check out the `Toolbar` stories

### DIFF
--- a/packages/wonder-blocks-toolbar/src/components/toolbar.tsx
+++ b/packages/wonder-blocks-toolbar/src/components/toolbar.tsx
@@ -137,8 +137,9 @@ const sharedStyles = StyleSheet.create({
     container: {
         border: `1px solid ${color.offBlack16}`,
         flex: 1,
-        flexDirection: "row",
-        justifyContent: "space-between",
+        display: "grid",
+        gridTemplateColumns:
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-toolbar/src/components/toolbar.tsx
```diff
@@ -81,54 +81,39 @@ export default function Toolbar({
...
+                        </LabelSmall>
+                    )}
                 </View>
             )}
+            {title && typeof title !== "string" && (
+                <View style={sharedStyles.titles}>{title}</View>
+            )}
+            {!title && (
+                <View style={leftContent ? sharedStyles.spacer : undefined} />
+            )}
```
jandrade, line 114: **question:** I wonder if this could be solved by applying `gap: spacing.small_12` in the parent container. Or is this need to be flexible enough for responsiveness?

jandrade (approved):
This is a really great improvement, thanks a lot for proposing this solution 👏

PR #2362, opened 2024-11-15
Announcer: Part 1

marcysutton:
The initial implementation for a live region component! I'm getting this draft PR up so I can test with the remote URL.

Issue: https://khanacademy.atlassian.net/browse/WB-1768

Outstanding questions/work areas:
- Testing in more ATs, particularly in webapp and on mobile devices.
- More integration with React: usage for JSX, debouncing with custom duration in continuously re-rendering components like the Video Player with Clarifications, etc.

## Test Plan

Play with the Story with screen readers turned on
    1. VoiceOver on OSX
    2. VoiceOver on iOS
    3. NVDA on Windows
    4. JAWS on Windows


packages/wonder-blocks-announcer/src/index.ts
```diff
@@ -0,0 +1,334 @@
...
+};
+
+/**
+ * Public API method to send screen reader messages.
+ * @param {string} message The message to send.
+ * @param {PolitenessLevel} level Polite or assertive announcements
+ * @param {number} removalDelay Optional duration to remove a message after sending. Defaults to 5000ms.
+ * @returns {string} IDREF for targeted live region element
+ */
+export function sendMessage({
```
jandrade, line 26: **suggestion:** I understand that this is the first iteration. I'd suggest moving this into a separate file so the code is more isolated (and tests can be isolated as you do it already).

I'd suggest doing something similar with `clearMessages`.

You can see how other packages follow this structure (for example, wb-core).

packages/wonder-blocks-announcer/src/index.ts
```diff
@@ -0,0 +1,334 @@
...
+    announcer = Announcer.getInstance();
+
+    if (typeof jest === "undefined") {
+        setTimeout(() => {
+            return announcer?.announce(message, level, removalDelay);
+        }, 100);
+    } else {
+        // If we are in a test environment, announce without waiting
+        return announcer.announce(message, level, removalDelay);
+    }
```
jandrade, line 40: **suggestion:** I'd strongly recommend not using test checks in the prod code. Instead, you could probably use advanceTimersByTime in during test setup.

e.g.
```
// Act
do something to trigger the announcer
jest.advanceTimersByTime(100);
```

For reference: https://jestjs.io/docs/timer-mocks#advance-timers-by-time

You can see some existing examples in the codebase.

packages/wonder-blocks-announcer/src/index.ts
```diff
@@ -0,0 +1,334 @@
...
+    message,
+    level = "polite", // TODO: decide whether to allow other roles, i.e. role=`timer`
+    removalDelay,
+}: SendMessageProps): string {
+    announcer = Announcer.getInstance();
+
+    if (typeof jest === "undefined") {
+        setTimeout(() => {
+            return announcer?.announce(message, level, removalDelay);
+        }, 100);
```
jandrade, line 36: **question:** Is there a specific reason why we need to use this timeout? I wonder if there's another way to handle this case, but I don't think I have full context on that.

packages/wonder-blocks-announcer/src/index.ts
```diff
@@ -0,0 +1,334 @@
...
+            // Check if our top level element already exists
+            const announcerCheck = document.getElementById(topLevelId);
+
+            // Init new structure if the coast is clear
+            if (announcerCheck === null) {
+                this.init(topLevelId);
+            }
+            // The structure exists but references are lost, so help HMR recover
+            else {
+                this.rebootForHMR();
```
jandrade, line 81: **thought:** ohh I see, yeah this is annoying. I wonder if you could setup storybook to initiate Announcer outside of the stories themselves. Not sure if this would be useful:
https://storybook.js.org/docs/api/main-config/main-config-preview-body

I haven't tried it out yet, but I was thinking about creating an initial DOM element to attach all the announcer contents to it. Similar to how react uses `createRoot` (e.g. `<div id="root">`). I'm not sure if you have explored a similar solution to this or not, but it could be worth considering.

packages/wonder-blocks-announcer/src/index.ts
```diff
@@ -0,0 +1,334 @@
...
+    regionFactory: RegionFactory = {
+        count: 2,
+        aIndex: 0,
+        pIndex: 0,
+    };
+    dictionary: RegionDictionary = new Map();
+
+    private constructor() {
+        if (typeof document !== "undefined") {
+            const topLevelId = `wbAnnounce`;
```
jandrade, line 71: **suggestion:** This is a good candidate for a constant at the file level so this ID can be shared across the file. wdyt?

packages/wonder-blocks-announcer/src/index.ts
```diff
@@ -0,0 +1,334 @@
...
+     */
+    init(id: string) {
+        this.node = document.createElement("div");
+        this.node.id = id;
+        this.node.setAttribute("data-testid", id);
+
+        Object.assign(this.node.style, srOnly);
+
+        const aWrapper = this.createRegionWrapper("assertive");
+        this.createDuplicateRegions(aWrapper, "assertive");
```
jandrade, line 108: **question:** Could you please elaborate why we need to create duplicate regions here? I was looking at the story in the browser and I'm not sure what should expect from that.

packages/wonder-blocks-announcer/src/index.ts
```diff
@@ -0,0 +1,334 @@
...
+        return index;
+    }
+}
+
+export default Announcer;
+
+/**
+ * Styling for live region.
+ * TODO: move to wonder-blocks-style package.
+ * Note: This style is overridden in Storybook for testing.
```
jandrade, line 323: **praise:** Thanks for adding this note 🫶

packages/wonder-blocks-announcer/src/index.ts
```diff
@@ -0,0 +1,334 @@
...
+        announcer?.clear(id);
+    } else if (document) {
+        announcer?.clear();
+    }
+}
+
+/**
+ * Internal class to manage screen reader announcements.
+ */
+class Announcer {
```
jandrade, line 59: **suggestion:** I'd recommend moving this class into a separate file as well. Sometimes we use an `internal/` folder for things like this, but up to you to decide what folder is best.

Side question: Do you think that `Announcer` can be tested independently? or is it better to rely on the exposed APIs for unit testing (the ones that you already created)?

packages/wonder-blocks-announcer/src/index.ts
jandrade, line 1: **suggestion:** Related to some of the comments left in this file. For consistency, we recommend splitting this file into separate files and keep this `index.ts` file as a barrel file. To clarify, I'm 100% with you about avoiding barrel files, but in the WB case, it really helps to make more clear and explicit what's going to be public API and what's internal/private. You can check other packages to see how we usually use this approach.

jandrade (commented):
This is looking really good! Giving my initial review with some suggestions on how to structure the files to be more consistent with the rest of the repo and asking some questions to get a bit more context on certain parts. This is great progress 👏

packages/wonder-blocks-announcer/src/index.ts
marcysutton: It was for Safari and VoiceOver, but I can play around with it again to see if we really need it.

packages/wonder-blocks-announcer/src/index.ts
marcysutton: That's what I had at first, but it added cognitive load when working with the API. It makes the file a lot more scannable to locate this ID closer to the code it's used on!

packages/wonder-blocks-announcer/src/index.ts
marcysutton: It's an architectural decision to try and prevent messages from being dropped by AT, something I picked up back in the Angular.js days. Coupled with the debounce logic I'm working on, I think I'll keep it in for now to ensure unique messages are appended and always picked up by AT.

packages/wonder-blocks-announcer/src/index.ts
marcysutton: Good question about testing! It seems redundant to test the class itself, at least the parts that affect the DOM. But perhaps some unit testing for the singleton and some of the utility methods would be a good idea?

packages/wonder-blocks-announcer/src/index.ts
marcysutton: One consideration for this is I was using a shared variable for the announcer in both of those API methods. I will lose that variable access when moving to separate files. One solution would be to use a global on the `window` object. Do you have any objections or suggestions for how to work around that?

packages/wonder-blocks-announcer/src/index.ts
jandrade: totally agree! if you see that adding unit tests to the class add value, then go for it. I wouldn't want to add more tests just for the sake of increasing coverage 😄.

packages/wonder-blocks-announcer/src/index.ts
jandrade: fair enough!

packages/wonder-blocks-announcer/src/index.ts
jandrade: ah ok, that makes sense thanks! Do you think you could add a short comment explaining this context? so other folks could understand the why of this approach. I'm sure my future self will appreciate that :).

packages/wonder-blocks-announcer/src/index.ts
marcysutton: Oh my gosh @beaesguerra you are so right! That is a very easy solve for this. Thank you!

jandrade: ahh that makes sense. TBH using globals make me worry a bit, so I think keeping it scoped would be still preferred. how do you feel about moving these two methods into the same file? so the `announcer` variable can still be shared.

packages/wonder-blocks-announcer/src/Announcer.ts
```diff
@@ -0,0 +1,291 @@
...
+     * @param {number} removalDelay How long to wait before removing message
+     * @returns {string} IDREF for targeted element or empty string if it failed
+     */
+    announce(
+        message: string,
+        level: PolitenessLevel,
+        removalDelay?: number,
+    ): string {
+        if (!this.node) {
+            return "";
```
beaesguerra, line 170: If node isn't set at this point, could we trigger the initialization steps so that it can continue to announce the message?

packages/wonder-blocks-announcer/src/Announcer.ts
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    /**
+     * Create live region element for a given level
+     * @param {string} level Politeness level for grouping
+     * @param {number} index Incrementor for duplicate regions
+     * @param {string} role Role attribute for live regions, defaults to log
+     * @returns {HTMLElement} DOM element reference for live region
+     */
+    createRegion(level: PolitenessLevel, index: number, role = "log") {
```
beaesguerra, line 139: What do you think about moving the document related utilities (`createRegion`, `createDuplicateRegions`, `createRegionWrapper` into a separate utility file? They could be tested on their own since they don't rely on the announcer instance! Then everything else in the `Announcer` class can be around managing announcements without worrying how to update the DOM!

beaesguerra (commented):
Great work so far Marcy! Left some feedback and questions 😄

packages/wonder-blocks-announcer/src/index.ts
beaesguerra: +1 me (and future me) too! Especially around alternating index. It would be great to document why we do this and how the live regions are structured!

packages/wonder-blocks-announcer/src/send-message.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+/**
+ * Public API method to send screen reader messages.
+ * @param {string} message The message to send.
+ * @param {PolitenessLevel} level Polite or assertive announcements
+ * @param {number} removalDelay Optional duration to remove a message after sending. Defaults to 5000ms.
+ * @returns {string} IDREF for targeted live region element
+ */
+export function sendMessage({
```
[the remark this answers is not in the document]
marcysutton: I was thinking this too, and now is the easiest time to rename it! Thanks so much for the suggestion (and all your great feedback!)

packages/wonder-blocks-announcer/src/Announcer.ts
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    /**
+     * Create a wrapper element to group regions for a given level
+     * @param {string} level Politeness level for grouping
+     * @returns {HTMLElement} Wrapper DOM element reference
+     */
+    createRegionWrapper(level: PolitenessLevel) {
+        const wrapper = document.createElement("div");
+        wrapper.id = `wbAWrap-${level}`;
```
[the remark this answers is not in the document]
marcysutton: Yes, it was a prefix for Wonder Blocks Announcer! The full id for these ends up like: `wbAWrap-polite`. It's pretty arbitrary though, so I could change it. I debated on whether to keep these wrapper elements at all, as they aren't really being used for anything.

packages/wonder-blocks-announcer/src/Announcer.ts
marcysutton: Good idea! I'll see what I can do with those...

packages/wonder-blocks-announcer/src/Announcer.ts
```diff
@@ -0,0 +1,291 @@
...
+
+        const pWrapper = this.createRegionWrapper("polite");
+        this.createDuplicateRegions(pWrapper, "polite");
+        this.node.appendChild(pWrapper);
+
+        document.body.prepend(this.node);
+    }
+    /**
+     * Recover in the event regions get lost
+     * This happens in Storybook when saving a file:
```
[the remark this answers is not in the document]
marcysutton: Yes, potentially! I'll do some testing on it.

packages/wonder-blocks-announcer/src/Announcer.ts
```diff
@@ -0,0 +1,291 @@
...
+
+        const aWrapper = this.createRegionWrapper("assertive");
+        this.createDuplicateRegions(aWrapper, "assertive");
+        this.node?.appendChild(aWrapper);
+
+        const pWrapper = this.createRegionWrapper("polite");
+        this.createDuplicateRegions(pWrapper, "polite");
+        this.node.appendChild(pWrapper);
+
+        document.body.prepend(this.node);
```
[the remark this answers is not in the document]
marcysutton: Good call. I could append it to the end of the document instead!

__docs__/wonder-blocks-announcer/announcer.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+ *  </button>
+ * </div>
+ * ```
+ */
+export default {
+    title: "Packages / Announcer",
+    component: AnnouncerExample,
+    decorators: [
+        (Story): React.ReactElement<React.ComponentProps<typeof View>> => (
+            <View style={styles.example}>
```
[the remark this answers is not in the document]
beaesguerra: Ahh gotcha, let's leave it as is! Thanks for trying it out 😄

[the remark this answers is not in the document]
marcysutton: Sounds good. Thanks for the suggestion, as it helped me learn a bit more about the decorators in our Storybook setup!

packages/wonder-blocks-announcer/src/index.ts
marcysutton: This is a good callout! I tried using the Jest functionality but it presented two challenges: 1) it became overly verbose to call before every method. and 2) the tests failed as the setTimeout was still being called.

Instead, I'm proposing a `timeoutDelay` parameter to adjust the announcement timeout duration (in addition to the `removalDelay`). That way the code can check if it's set to 0 and skip the initial timeout that way, rather than referencing Jest.

jandrade: cool! dependency injection FTW! (kinda)

packages/wonder-blocks-announcer/src/index.ts
marcysutton: Yes, this does seem important for Safari in particular! I made it configurable so it can be manipulated at runtime if necessary (especially for tests).

packages/wonder-blocks-announcer/src/index.ts
marcysutton: Good call. I added some comments to this and the types!

packages/wonder-blocks-announcer/src/index.ts
marcysutton: Makes sense to me! I made some updates so I can keep working on unit tests and the debounce functionality.

packages/wonder-blocks-announcer/src/Announcer.ts
marcysutton: Yes, great idea! I renamed the HMR method to "reattachNodes" and called it from this part of the code.

packages/wonder-blocks-announcer/src/Announcer.ts
```diff
@@ -0,0 +1,291 @@
...
+    node: HTMLElement | null = null;
+    regionFactory: RegionFactory = {
+        count: 2,
+        aIndex: 0,
+        pIndex: 0,
+    };
+    dictionary: RegionDictionary = new Map();
+
+    private constructor() {
+        if (typeof document !== "undefined") {
```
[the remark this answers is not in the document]
marcysutton: Good question! I think this version is safer if there isn't a document context, which is what I'm guarding against. I updated the check in `clearMessages` to match this one so it won't blow up (e.g. if JSDOM isn't loaded, or we are in SSR again for some reason)! The if condition in that function checks for an element on the document object with optional chaining, so *I believe* it has a bit more type safety built in.

packages/wonder-blocks-announcer/src/Announcer.ts
```diff
@@ -0,0 +1,291 @@
...
+    RegionDef,
+} from "../types/Announcer.types";
+
+const TIMEOUT_DELAY = 5000;
+
+/**
+ * Internal class to manage screen reader announcements.
+ */
+class Announcer {
+    private static _instance: Announcer;
```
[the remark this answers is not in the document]
marcysutton: Yes, probably. I'm a little surprised TS doesn't complain about it as-is!

packages/wonder-blocks-announcer/src/index.ts
marcysutton: I ended up removing this with the move to async! It wasn't needed in my testing for Safari/Voiceover.

marcysutton:
@jandrade @beaesguerra this is ready for another review! I added quite a few more tests. Yay!

In testing the debounce logic more thoroughly, I realized the first iteration was going to be unpredictable and hard to use. So I refactored it a bit. Some highlights:

1. I added a `debounceThreshold` parameter to allow callers to specify how long to wait before announcing another message. This is intended for repeatedly-rendering components like the [ClarificationNotifications component for the video player](https://github.com/Khan/webapp/blob/master/services/static/javascript/clarifications-package/official.tsx) which are re-rendered every 1/2 second. This seemed like a reasonable way to handle it without having to compare strings. (We might still need to do that so it's not annoying to screen reader users, but I want to try this approach first.)
2. The debounce function now returns the first message rather than the last, so it will be announced immediately (other implementations use the last callback rather than the first).
3. I got rid of the `removalDelay` parameter as it wasn't that useful to begin with, and it made the entire thing super confusing. Elements are still removed after a period time, I just changed it to not be configurable rather than trying to maintain `removalDelay` and `debounceThreshold` which both use setTimeouts. 
  -  The removalDelay is internally set to 5000 ms + the default debounce wait time of 250ms. If a custom debounceThreshold is passed in (say 2000ms), the internal removalDelay would updated to a longer duration for removal to avoid a race condition (2000ms + 5000ms).

My next step is to test this more in webapp and see if it works as I expected!

packages/wonder-blocks-announcer/src/index.ts
marcysutton: An initial timeout might still be needed to get reliable announcements! The async code does seem to be messing with Voiceover in both Chrome and Safari, especially on the first couple of messages. When I test in Storybook, polite messages are getting dropped almost entirely and only some assertive messages are making it through. So I'll figure out how to work it back in..probably in the `announce` method of the Announcer class.

packages/wonder-blocks-announcer/src/Announcer.ts
```diff
@@ -0,0 +1,257 @@
...
+     * This happens in Storybook or other HMR environments when saving a file:
+     * Announcer exists, but it loses the connection to DOM element Refs
+     */
+    reattachNodes() {
+        const announcerCheck = document.getElementById(`wbAnnounce`);
+        if (announcerCheck !== null) {
+            this.node = announcerCheck;
+            const regions = Array.from(
+                announcerCheck.querySelectorAll<HTMLElement>(
+                    "[id^='wbARegion'",
```
nishasy, line 106: I think this line is missing a `]`

marcysutton: Good call! I wish there was some more static typing for those strings...

packages/wonder-blocks-announcer/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "types": "dist/index.d.ts",
+  "author": "",
+  "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@khanacademy/wonder-blocks-core": "^7.0.1"
```
jandrade, line 19: **suggestion:** You don't need to include `core` as dependency as it seems that it is not going to be used in the bundle. If this is for testing and/or storybook, you can include it as a `devDependency`.

packages/wonder-blocks-announcer/types/Announcer.types.ts
```diff
@@ -0,0 +1,37 @@
...
+export type PolitenessLevel = "polite" | "assertive";
+
+/*
+RegionFactory: A config for creating duplicate region elements.
+- Count is the total number for each level.
+- aIndex references the index of the last-used assertive log element.
+- pIndex references the index of the last-used polite log element.
+*/
+export type RegionFactory = {
+    count: number;
```
jandrade, line 15: **question:** Is count changed at some point? or is it always a static value? I wonder if you can get this number by counting the keys in this factory object instead? (I don't think this is going to change at all, so this approach works too if you find it simpler).

packages/wonder-blocks-announcer/src/__tests__/util/dom.test.ts
jandrade, line 1: **suggestion:** I'd recommend moving this file is inside `src/util/__tests__` so the tests are closer to the code. Same would apply to `util.test.ts`

packages/wonder-blocks-announcer/src/util/dom.ts
```diff
@@ -0,0 +1,82 @@
...
+    level: PolitenessLevel,
+    regionCount: number,
+    dictionary: RegionDictionary,
+): HTMLElement[] {
+    const result = new Array(regionCount).fill(0).map((el, i) => {
+        const region = createRegion(level, i, dictionary);
+        wrapper.appendChild(region);
+        return region;
+    });
+    return result;
```
jandrade, line 36: **question:** Is the return value only needed for tests? I see that it is not used in the code, so I'm wondering if that return would be useful in the `Announcer` class in some way?

packages/wonder-blocks-announcer/src/__tests__/util/dom.test.ts
```diff
@@ -0,0 +1,133 @@
...
+    createRegionWrapper,
+    createDuplicateRegions,
+    createRegion,
+    removeMessage,
+} from "../../util/dom";
+
+jest.useFakeTimers();
+jest.spyOn(global, "setTimeout");
+
+describe("Announcer utility functions", () => {
```
jandrade, line 12: **suggestion:** It would be great using AAA in these tests: https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/98402353/Unit+Testing+Best+Practices#Arrange-Act-Assert

packages/wonder-blocks-announcer/src/__tests__/util/dom.test.ts
```diff
@@ -0,0 +1,133 @@
...
+    describe("createRegionWrapper", () => {
+        test("it creates a polite region wrapper element", () => {
+            const element = createRegionWrapper("polite");
+
+            expect(element.tagName).toBe("DIV");
```
jandrade, line 17: **suggestion:** This assertion could be dropped as we can test that the element exists with the `id` we expect (`DIV` would be an implementation detail here that the test doesn't need to be aware of).
```suggestion
           
```

packages/wonder-blocks-announcer/src/__tests__/util/dom.test.ts
```diff
@@ -0,0 +1,133 @@
...
+            expect(element.id).toEqual("wbAWrap-polite");
+        });
+
+        test("it creates an assertive region wrapper element", () => {
+            const element = createRegionWrapper("assertive");
+
+            expect(element.tagName).toBe("DIV");
```
jandrade, line 24: Related to the previous comment

```suggestion
            
```

packages/wonder-blocks-announcer/src/__tests__/util/util.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    test("a single message", async () => {
+        // ARRANGE
+        const callback = jest.fn((message: string) => message);
+        const debounced = debounce(callback, 300);
+
+        // ACT
+        const resultPromise = debounced("Hello, World!");
+
+        // ASSERT
+        expect(resultPromise).toBeInstanceOf(Promise);
```
jandrade, line 15: **question:** Is this assertion required for this test? I wonder if just using the expectation below is a good signal to see if the test works as expected?

packages/wonder-blocks-announcer/src/__tests__/util/dom.test.ts
```diff
@@ -0,0 +1,133 @@
...
+                "assertive",
+                2,
+                dictionary,
+            );
+
+            expect(regionList.length).toBe(2);
+            expect(regionList[0].id).toBe("wbARegion-assertive0");
+            expect(regionList[1].id).toBe("wbARegion-assertive1");
+            expect(dictionary.size).toBe(2);
+        });
```
jandrade, line 62: **suggestion:** I think this test could be dropped as the actual functionality is tested in the previous test.

packages/wonder-blocks-announcer/src/__tests__/util/dom.test.ts
```diff
@@ -0,0 +1,133 @@
...
+            expect(region.getAttribute("aria-live")).toBe("assertive");
+            expect(region.getAttribute("role")).toBe("log");
+            expect(dictionary.size).toBe(1);
+        });
+
+        test("it allows the role to be overridden", () => {
+            const dictionary = new Map();
+            const region = createRegion("polite", 0, dictionary, "timer");
+
+            expect(region.getAttribute("aria-live")).toBe("polite");
```
jandrade, line 90: **suggestion:** You could drop this assertion as it was already tested in a previous test, and instead only assert the part described in the test (which would be the role).

```suggestion

```

packages/wonder-blocks-announcer/src/__tests__/util/dom.test.ts
```diff
@@ -0,0 +1,133 @@
...
+
+        test("it creates an assertive Live Region element", () => {
+            const dictionary = new Map();
+            const region = createRegion("assertive", 0, dictionary);
+
+            expect(region.id).toBe("wbARegion-assertive0");
+            expect(region.getAttribute("aria-live")).toBe("assertive");
+            expect(region.getAttribute("role")).toBe("log");
+            expect(dictionary.size).toBe(1);
+        });
```
jandrade, line 84: **suggestion:** Given another look at these tests that only change the politeness level. Maybe you could also add dynamic tests to avoid repeating the whole test structure?

```
test.each(["polite", "assertive"])("it creates an %s Live Region element", (politenessLevel) => {
            // Arrange
            const dictionary = new Map();

            // Act
            const region = createRegion(politenessLevel, 0, dictionary);

            // Assert
            expect(region.id).toBe(`wbARegion-${politenessLevel}0`);
            expect(region.getAttribute("aria-live")).toBe("politenessLevel");
            expect(region.getAttribute("role")).toBe("log");
            expect(dictionary.size).toBe(1);
        });
```

jandrade: Another suggestion would be to split the assertions into multiple tests, so it would help understand better which assertion/test is exactly failing (if that's ever the case), which follows the AAA pattern.

packages/wonder-blocks-announcer/src/Announcer.ts
```diff
@@ -0,0 +1,257 @@
...
+
+        // For each level, we create at least two live region elements.
+        // This is to work around AT occasionally dropping messages.
+        const aWrapper = createRegionWrapper("assertive");
+        createDuplicateRegions(
+            aWrapper,
+            "assertive",
+            this.regionFactory.count,
+            this.dictionary,
+        );
```
jandrade, line 81: **thought:**  One thing that I noticed while playing with Storybook is that if I change the text in the Story and click the "submit" button, sometimes the announcement is not read by Voice Over. I'm not sure where to put this, but saw that sometimes was announced in one of the duplicate regions.

jandrade (commented):
Thanks for this new iteration. I'm adding some extra suggestions around the code (specially around tests). I'll keep digging into it and will integrate with the Combobox component. Thanks for all this progress!!

packages/wonder-blocks-announcer/src/Announcer.ts
```diff
@@ -0,0 +1,257 @@
...
+} from "../types/Announcer.types";
+
+import {
+    createRegionWrapper,
+    createDuplicateRegions,
+    removeMessage,
+} from "./util/dom";
+import {alternateIndex, debounce} from "./util/util";
+
+const REMOVAL_TIMEOUT_DELAY = 5000;
```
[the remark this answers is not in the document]
marcysutton: It's admittedly a bit of a random number... long enough to let the platform accessibility API read out a message of unknown length before it's removed, and also longer than the default initial timeout. Initially I made this configurable by the user but I removed it due to added complexity to maintain.

packages/wonder-blocks-announcer/src/__tests__/util/util.test.ts
marcysutton: Makes sense to me!

packages/wonder-blocks-announcer/src/__tests__/util/dom.test.ts
marcysutton: The `aria-live` attribute is the part actually changing the politeness level, so I think I'll keep that in. I am refactoring it to be dynamic and less repetitive, though! Thanks for the suggestion.

packages/wonder-blocks-announcer/src/Announcer.ts
marcysutton: Hmm, that is odd. I noticed this if I change the politeness level to Assertive but not for different text. The default Storybook frame also seems to impact reliability of messages, so I tend to test it in a standalone window. I think I'll opt for integration testing to observe these kinds of issues since I've spent so long on it already!

packages/wonder-blocks-announcer/src/Announcer.ts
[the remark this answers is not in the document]
beaesguerra: **idea**: What if we determined the delay dynamically based on the number of words in the announcement multiplied by something like 500ms? That way long messages are accounted for! This might also account for translations that might be longer, though we would be assuming approximately how long each word is read!

packages/wonder-blocks-announcer/src/clear-messages.ts
```diff
@@ -0,0 +1,15 @@
+import Announcer from "./announcer";
+
+/**
+ * Public API method to clear screen reader messages after sending.
```
beaesguerra, line 4: **nit:** _very_ minor - I think we can remove the "Public API method" part in the comment since this comment block will be shown in code hints in text editors! This is more useful for us so maybe it could be a file comment?

packages/wonder-blocks-announcer/src/announcer.ts
```diff
@@ -0,0 +1,257 @@
...
+     * @returns {Promise<string>} Promise that resolves with an IDREF for targeted element or empty string if it failed
+     */
+    async announce(
+        message: string,
+        level: PolitenessLevel,
+        debounceThreshold: number = DEFAULT_WAIT_THRESHOLD,
+    ): Promise<string> {
+        // Locking variable to prevent simultaneous updates
+        if (this.stateLock) {
+            return Promise.resolve(""); // Return early if state is locked
```
beaesguerra, line 136: **question**: If `announce` is called and `stateLock` is true, is there a waiting/re-try mechanism for the message?

beaesguerra (commented):
Great progress Marcy! Left some questions and comments 😄

packages/wonder-blocks-announcer/src/index.ts
marcysutton: I did some more testing and added the initial timeout back in for Safari. I also included an option to set the length to 0 so that tests can skip it and work properly.

packages/wonder-blocks-announcer/src/Announcer.ts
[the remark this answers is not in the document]
marcysutton: It's an interesting thought. I'm hesitant to go that route as screen reader speech rates are configurable, so it's a moving target. Perhaps we could mark that down for a future improvement if it comes up in production?

packages/wonder-blocks-announcer/src/util/util.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/**
+ * Keep announcements from happening too often by limiting callback execution by time.
+ * Anytime the announcer is called repeatedly, this can slow down the results.
+ * @param {Function} callback Announce method to call with argments
+ * @param {number} wait Length of time to wait before calling callback again
+ * @returns {string} idRef of targeted live region element
+ */
+export function debounce(callback: (...args: any[]) => string, wait: number) {
```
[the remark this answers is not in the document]
marcysutton: Good call out! This one needs to return a promise with a string so it's pretty specific internally.

packages/wonder-blocks-announcer/src/clear-messages.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Clears all regions by default.
+ * @param {string} id Optional id of live region element to clear.
+ */
+export function clearMessages(id?: string) {
+    const announcer = Announcer.getInstance();
+    if (id && document?.getElementById(id)) {
+        announcer?.clear(id);
```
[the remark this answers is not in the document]
marcysutton: Good call, we don't need it!

marcysutton:
Alright, the debounce refactor is in! (it wasn't actually working when I tested it) I had to change it to allow a configurable debounce wait parameter with the simple `announceMessage` API. This means you can just call it with options without having to configure anything separately (helpful when it's being called in multiple places, so configurations don't interfere with each other).

@jandrade I also fixed that styling issue in Storybook that you were seeing with the Combobox. I added an `addBodyClass` decorator that can be used for anything...kind of a nice general improvement!

__docs__/wonder-blocks-announcer/announcer.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+    },
+} as Meta<typeof AnnouncerExample>;
+
+/**
+ * This is an example of a live region with all the options set to their default
+ * values and the `message` argument set to some example text.
+ */
+export const SendMessage: StoryComponentType = {
+    args: {
+        message: "Here is some example text.",
```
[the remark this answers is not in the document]
marcysutton: So I played around with adding Date.now() or Math.random() to the default message, and it caches it so nothing changes with repeated clicks. The message can be changed manually for this purpose. Any ideas on how to change it dynamically in the way you're imagining?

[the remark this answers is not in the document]
beaesguerra: I think to have dynamic messages each time the button is clicked, we would need to update it in the `AnnouncerExample` component. When we configure the story args here, it will only initialize the storybook control values for the first render

For example, if we add `Date.now()` to the message when we call `announceMessage` in the `AnnouncerExample` component, it should be a different message each time!

beaesguerra (commented):
This is looking great, I tested the story out with VoiceOver + Safari, NVDA + Chrome/Firefox and the message is announced as I expected! Thanks for your hard work on this!

I had a few suggestions around stories for other scenarios and documentation (non-blocking). I'm looking forward to seeing the announcer get used in context! 

(Will let Juan be the final approver though since he may have feedback/insights on how it was integrating it with combobox!)

__docs__/wonder-blocks-announcer/announcer.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+        level: "polite",
+    },
+};
```
beaesguerra, line 109: We can disable this story for chromatic since we don't need visual regression tests for it! https://www.chromatic.com/docs/disable-snapshots/#with-storybook

jandrade:
@marcysutton I've been testing the latest changes in `Combobox` and I'm still having some trouble announcing messages correctly when the dismiss/clear button is pressed.

The issue is that the button is pressed, the message is queued to be announced, but because the `input[combobox]` is focused again, then it causes the `listbox` to open and prioritizes the announcement for that (e.g. `Expanded, completion selected`). I'm attaching a screenshot here to illustrate this case:


https://github.com/user-attachments/assets/5d767a50-c573-45dc-a186-ac666ffee94d

I'll try to put a new PR with my current progress.

packages/wonder-blocks-announcer/src/__tests__/announce-message.test.tsx
```diff
@@ -0,0 +1,172 @@
...
+        button.click();
+
+        // ASSERT: expect live regions to exist
+        const politeRegion1 = screen.queryByTestId("wbARegion-polite0");
+        const politeRegion2 = screen.queryByTestId("wbARegion-polite1");
+        expect(politeRegion1).toHaveAttribute("aria-live", "polite");
+        expect(politeRegion1).toHaveAttribute("id", "wbARegion-polite0");
+        expect(politeRegion2).toHaveAttribute("aria-live", "polite");
+        expect(politeRegion2).toHaveAttribute("id", "wbARegion-polite1");
+    });
```
nedredmond, line 69: Question: Can we make these `aria-atomic` so that they read the whole message instead of browsers/srs trimming similar starting text? Or is that handled in a different way?

marcysutton: We can definitely try it! I wasn't noticing differences with the current implementation. But I will caveat that I'm also planning to experiment with a queue of messages. So `aria-atomic` might not make sense in that implementation!

__docs__/wonder-blocks-announcer/announcer.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+    argTypes: {
+        level: {
+            control: "radio",
+            options: ["polite", "assertive"],
+            defaultValue: "polite",
+        },
+        timeoutDelay: {
+            control: "number",
+            type: "number",
+            description: "(milliseconds)",
```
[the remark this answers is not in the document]
marcysutton: Good ideas on docs! I will plan to tackle this in a future PR.

__docs__/wonder-blocks-announcer/announcer.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+import ComponentInfo from "../../.storybook/components/component-info";
+import packageConfig from "../../packages/wonder-blocks-announcer/package.json";
+
+const AnnouncerExample = ({
+    message = "Clicked!",
+    level,
+    debounceThreshold,
+}: AnnounceMessageProps) => {
+    return (
+        <Button
```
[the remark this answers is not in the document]
marcysutton: Love these ideas. I will tackle them in a future PR also! Thanks Bea!

packages/wonder-blocks-announcer/src/__tests__/announce-message.test.tsx
marcysutton: I made a note to look into this in a future PR, since I want to do some more cross-platform testing anyway!

PR #2363, opened 2024-11-18
SearchField: Add validation props and refine states

beaesguerra:
## Summary:
To make the SearchField component more consistent with other form field components like TextField and TextArea, we add validation related props to the component. See [LabeledField Implementation Spec](https://khanacademy.atlassian.net/wiki/spaces/WB/pages/2261352551/LabeledField+Implementation+Spec) for more details!

The changes include:
- Add error, instantValidation, validate, onValidate props to SearchField
- Minor tweaks
  - Hide clear icon button if the field is disabled
  - Refine magnifying glass icon styling to make it [match Figma component](https://www.figma.com/design/VbVu3h2BpBhH80niq101MHHE/%F0%9F%92%A0-Main-Components?node-id=13840-8605&t=83MBW2IFJ6nx42vp-4) more (smaller, bold icon, spacing, update disabled state)
- Storybook:
  - Add error and validation stories for SearchField
  - Add variant stories for SearchField states

Issue: WB-1761

## Test plan:
- Review Storybook docs for Error and Validation stories
  - `?path=/docs/packages-searchfield--docs#error`
  - `?path=/docs/packages-searchfield--docs#validation`
- Review prop documentation for SearchField (`?path=/docs/packages-searchfield--docs`)
- Review new Search Field All Variants stories (`?path=/docs/packages-searchfield-all-variants--docs`)
- Review Chromatic diff for styling updates (left some comments in the UI Tests!)

### DIFF
--- a/__docs__/wonder-blocks-search-field/search-field-variants.stories.tsx
+++ b/__docs__/wonder-blocks-search-field/search-field-variants.stories.tsx
@@ -1,0 +1,167 @@
+import * as React from "react";
+import {StyleSheet} from "aphrodite";
+import type {Meta, StoryObj} from "@storybook/react";
+
+import {View} from "@khanacademy/wonder-blocks-core";
+import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {LabelLarge, LabelMedium} from "@khanacademy/wonder-blocks-typography";
+import SearchField from "@khanacademy/wonder-blocks-search-field";
+
+/**
+ * The following stories are used to generate the pseudo states for the
+ * SearchField component. This is only used for visual testing in Chromatic.
+ */
+export default {
+    title: "Packages / SearchField / All Variants",
+    parameters: {
+        docs: {
+            autodocs: false,
+        },
+    },
+} as Meta;
+
+type StoryComponentType = StoryObj<typeof SearchField>;
+
+const longText =
+    "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.";
+const longTextWithNoWordBreak =
+    "Loremipsumdolorsitametconsecteturadipiscingelitseddoeiusmodtemporincididuntutlaboreetdoloremagnaaliqua";
+
+const states = [
+    {
+        label: "Default",
+        props: {},
+    },
+    {
+        label: "Disabled",
+        props: {disabled: true},
+    },
+    {
+        label: "Error",
+        props: {error: true},
+    },
+];
+const States = (props: {
+    light: boolean;
+    label: string;
+    value?: string;
+    placeholder?: string;
+}) => {
+    return (
+        <View
+            style={[props.light && styles.darkDefault, styles.statesContainer]}
+        >
+            <LabelLarge style={props.light && {color: color.white}}>
+                {props.label}
+            </LabelLarge>
+            <View style={[styles.scenarios]}>
+                {states.map((scenario) => {
+                    return (
+                        <View style={styles.scenario} key={scenario.label}>
+                            <LabelMedium
+                                style={props.light && {color: color.white}}
+                            >
+                                {scenario.label}
+                            </LabelMedium>
+                            <SearchField
+                                value=""
+                                onChange={() => {}}
+                                {...props}
+                                {...scenario.props}
+                            />
+                        </View>
+                    );
+                })}
+            </View>
+        </View>
+    );
+};
+
+const AllVariants = () => (
+    <View>
+        {[false, true].map((light) => {
+            return (
+                <React.Fragment key={`light-${light}`}>
+                    <States light={light} label="Default" />
+                    <States light={light} label="With Value" value="Text" />
+                    <States
+                        light={light}
+                        label="With Value (long)"
+                        value={longText}
+                    />
+                    <States
+                        light={light}
+                        label="With Value (long, no word breaks)"
+                        value={longTextWithNoWordBreak}
+                    />
+                    <States
+                        light={light}
+                        label="With Placeholder"
+                        placeholder="Placeholder text"
+                    />
+                    <States
+                        light={light}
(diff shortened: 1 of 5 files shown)


packages/wonder-blocks-search-field/src/components/search-field.tsx
```diff
@@ -167,8 +195,14 @@ const SearchField: React.ForwardRefExoticComponent<
                 <View onClick={onClick} style={[styles.inputContainer, style]}>
                     <PhosphorIcon
                         icon={magnifyingGlassIcon}
-                        size="medium"
-                        color={color.offBlack64}
+                        size="small"
```
beaesguerra, line 198: Applying some styling updates to make the search field look more like the Figma component! It also addresses the disabled and light states

Before (left) and After (right):
image

jandrade (commented):
This is looking great! left a couple of nit/non-blocking suggestions before giving final approval.

__docs__/wonder-blocks-search-field/search-field-variants.stories.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+const styles = StyleSheet.create({
+    darkDefault: {
+        backgroundColor: color.darkBlue,
+    },
+    statesContainer: {
+        padding: spacing.medium_16,
+    },
+    scenarios: {
+        display: "flex",
```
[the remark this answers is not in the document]
beaesguerra: Removed! (🦸 super-nit 😄)

__docs__/wonder-blocks-search-field/search-field.stories.tsx
```diff
@@ -66,15 +70,26 @@ const Template = (args: any) => {
...
+                onChange={handleChange}
+                onKeyDown={(e) => {
+                    action("onKeyDown")(e);
+                    handleKeyDown(e);
+                }}
+                onValidate={setErrorMessage}
+            />
+            {(errorMessage || args.error) && (
+                <>
+                    <Strut size={spacing.xSmall_8} />
```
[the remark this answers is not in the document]
beaesguerra: That makes sense! I wonder if it would be helpful to add a `gap` or `space` prop to `View`. What do you think?

Also, is it preferred we inline the aphrodite styles or create styles in a stylesheet for things like this? 
```
<View style={{ gap: spacing.xSmall_8 }} />
```
or
```
<View style={styles.container} />
...
  container: {
    gap: spacing.xSmall_8,
  }
});
```
Is one more performant? One nice thing about inlining the styles is that if the styling gets removed, it's removed! Unlike the other way, there's a chance we forget to remove the unused style from the stylesheet! This may also change depending on our css approach, curious if anyone has thoughts on this :)

packages/wonder-blocks-search-field/src/components/search-field.tsx
```diff
@@ -146,13 +174,14 @@ const SearchField: React.ForwardRefExoticComponent<
...
     const maybeRenderClearIconButton: () => React.ReactElement = () => {
-        if (!value.length) {
+        if (!value.length || disabled) {
             return null;
         }
 
         return (
             <IconButton
                 icon={xIcon}
+                size="xsmall"
```
beaesguerra, line 184: I also updated the dismiss icon styling (bold icon, smaller, spacing!). Thanks for catching this @jandrade! 

Since IconButton is smaller now, this addresses the issue where the focus outline on the dismiss button can be cropped out if an ancestor element has `overflow: hidden` (see [Chromatic comment](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=673baf9294168df34de5619b))
image

The hit area seems too small though for accessibility. I wonder if all sizes of the IconButton component should all have the same minimum hit area?

jandrade: I think we should probably use `size="small"` instead, as it keeps the same icon size (16x16) but the hit area is slightly bigger (32 instead of 24), which would improve a11y in that regard. We'd still fail to achieve AAA, but I think it would be a good compromise for the mid-term.

And I like the suggestion about revisiting the hit areas of all `IconButton` variants. I think this is something worth exploring and documenting in the WB audit that we are doing for the new experience.

beaesguerra: `size="small"` for IconButton uses a [24x24 icon](https://khan.github.io/wonder-blocks/?path=/docs/packages-iconbutton--docs#sizes), which is different from `size="small"` for PhosphorIcon! I'll update this to `size="small"` and have the icon be a bit bigger so it is more usable!

beaesguerra: I'll also update it back to the regular icon weight, I think when it's bold and 24x24, the dismiss icon is too prominent! 
image image

jandrade: Thanks for trying this out. The last iteration makes sense to me.

For context, I got a big confused as I saw that the designs apparently used an icon of 24x24 but in reality the icon was 16x16

Screenshot 2024-11-20 at 2 48 50 PM

https://www.figma.com/design/VbVu3h2BpBhH80niq101MHHE/%F0%9F%92%A0-Main-Components?node-id=13843-9211&t=nNY8Jl9Ah253UTqj-4

jandrade (approved):
Looks awesome! thanks for adding these changes and for getting more consistency across our form fields 👏 🚀

packages/wonder-blocks-search-field/src/components/search-field.tsx
beaesguerra: Ahh yes that is confusing! I added a note in WB-1792 to document tips for developers around icons once we work with design to make the translation easier!

I also updated the [audit spreadsheet](https://docs.google.com/spreadsheets/d/1Wcnv-JKnyT-w2MG-dAZsS5PacV8IoUqYbzNYQq-mxJY/edit?gid=0#gid=0) to include notes about the IconButton hit area and IconButton/PhosphorIcon size consistency!

__docs__/wonder-blocks-search-field/search-field.stories.tsx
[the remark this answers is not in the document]
marcysutton: The inline approach sure reminds me of Tailwind! It puts the styles closer to their usage. One downside is the code is less portable, whereas classes can be centralized and reused.

marcysutton (approved):
This seems like a really great change, nice work @beaesguerra!

PR #2364, opened 2024-11-18
WB-1779: Add startIcon prop to Combobox

jandrade:
## Summary:

Adds a new `startIcon` prop (optional) to the Combobox component. This prop
will allow to optinally add an icon to the left of the input field.

This use case will be helpful for the AI tools case where we want to use a
search icon.

Issue: https://khanacademy.atlassian.net/browse/WB-1779

## Test plan:

Verify that the `Start icon` docs make sense and that the example works as
expected.

/?path=/docs/packages-dropdown-combobox--docs#start%20icon

Screenshot 2024-11-18 at 3 21 00 PM

### DIFF
--- a/__docs__/wonder-blocks-dropdown/combobox.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/combobox.stories.tsx
@@ -6,7 +6,7 @@ import {StyleSheet} from "aphrodite";
 import * as React from "react";
 import magnifyingGlassIcon from "@phosphor-icons/core/regular/magnifying-glass.svg";
 
-import {LabelLarge} from "@khanacademy/wonder-blocks-typography";
+import {LabelLarge, LabelMedium} from "@khanacademy/wonder-blocks-typography";
 import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
 import {Checkbox} from "@khanacademy/wonder-blocks-form";
 import {Combobox, OptionItem} from "@khanacademy/wonder-blocks-dropdown";
@@ -440,8 +440,39 @@ export const Error: Story = {
  * In this example, we show how this is done by setting the `startIcon` prop.
  */
 export const StartIcon: Story = {
+    render: function Render(args: PropsFor<typeof Combobox>) {
+        const [_, updateArgs] = useArgs();
+
+        return (
+            <View style={{gap: spacing.medium_16}}>
+                <LabelMedium>Default:</LabelMedium>
+                <Combobox
+                    {...args}
+                    onChange={(newValue) => {
+                        updateArgs({value: newValue});
+                        action("onChange")(newValue);
+                    }}
+                />
+                <LabelMedium>Disabled:</LabelMedium>
+                <Combobox
+                    {...args}
+                    disabled={true}
+                    onChange={(newValue) => {
+                        updateArgs({value: newValue});
+                        action("onChange")(newValue);
+                    }}
+                />
+            </View>
+        );
+    },
     args: {
         children: items,
-        startIcon: <PhosphorIcon icon={magnifyingGlassIcon} size="medium" />,
+        startIcon: (
+            <PhosphorIcon
+                icon={magnifyingGlassIcon}
+                size="medium"
+                color={color.offBlack64}
+            />
+        ),
     },
 };
--- a/__docs__/wonder-blocks-dropdown/combobox.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/combobox.stories.tsx
@@ -4,10 +4,10 @@ import {Meta, StoryObj} from "@storybook/react";
 import {expect, userEvent, within} from "@storybook/test";
 import {StyleSheet} from "aphrodite";
 import * as React from "react";
-import magnifyingGlassIcon from "@phosphor-icons/core/regular/magnifying-glass.svg";
+import magnifyingGlassIcon from "@phosphor-icons/core/bold/magnifying-glass-bold.svg";
 
 import {LabelLarge, LabelMedium} from "@khanacademy/wonder-blocks-typography";
-import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {color, semanticColor, spacing} from "@khanacademy/wonder-blocks-tokens";
 import {Checkbox} from "@khanacademy/wonder-blocks-form";
 import {Combobox, OptionItem} from "@khanacademy/wonder-blocks-dropdown";
 import {PhosphorIcon} from "@khanacademy/wonder-blocks-icon";
@@ -433,11 +433,16 @@ export const Error: Story = {
 };
 
 /**
- * With start icon, you can customize the icon that appears at the beginning of
+ * With `startIcon`, you can customize the icon that appears at the beginning of
  * the Combobox. This is useful when you want to add a custom icon to the
  * component.
  *
- * In this example, we show how this is done by setting the `startIcon` prop.
+ * **NOTE:** When `startIcon` is set, we set some default values for the icon:
+ * - `size`: "small"
+ * - `color`: `semanticColor.icon.default`
+ *
+ * You can customize the size and color of the icon by passing the `size` and
+ * `color` props to the `PhosphorIcon` component.
  */
 export const StartIcon: Story = {
     render: function Render(args: PropsFor<typeof Combobox>) {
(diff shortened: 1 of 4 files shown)


packages/wonder-blocks-dropdown/src/components/combobox.tsx
```diff
@@ -132,6 +133,14 @@ type Props = {
      */
     // TODO(WB-1740): Add support to `inline` and `both` values.
     autoComplete?: "none" | "list" | undefined;
+
+    /**
+     * The icon (wrapped by a button) that allows opening the listbox widget
```
beaesguerra, line 138: I think the prop docs need to be updated for `startIcon`!

beaesguerra (commented):
Changes look good, I noticed the prop docs needed to be updated though!

I also have some questions/things we can consider (doesn't have to be part of this PR!): 

1. I was also working on a similar magnifying icon in the SearchField component in PR https://github.com/Khan/wonder-blocks/pull/2363! I noticed the styling in the code component looked a bit different from the Figma component so I updated the styles while I was there 😄 . Since both Combobox and SearchField use TextField, it would be great if we could consolidate how the start icon gets rendered! There were several considerations for styling the icon like the `disabled` and `light` props. One idea on how we could consolidate the styling would be to add a `startIcon` prop to `TextField` and it could provide the color, sizing, spacing. This could be useful for `TextField` on its own too! What do you think?
    - I was also wondering if a `endIcon` prop would also be useful for things like the Chevron icon, though it is an icon button so maybe the prop should support icon buttons too. Another use case would be the clear icon button in SearchField, or a show/hide icon button on a password field! The Combobox can potentially have 2 end icon buttons though for the clear and expand buttons, so perhaps it would take a React.Node instead of limiting it to Icon/IconButton!

2. For the combobox, would it be helpful for users to always have the magnifying glass icon there by default so there is a hint that they can type to search? Then teams can customize the icon for specific use cases if needed!

jandrade:
> I also have some questions/things we can consider (doesn't have to be part of this PR!):
> 
> 1. I was also working on a similar magnifying icon in the SearchField component in PR [SearchField: Add validation props and refine states #2363](https://github.com/Khan/wonder-blocks/pull/2363)! I noticed the styling in the code component looked a bit different from the Figma component so I updated the styles while I was there 😄 . Since both Combobox and SearchField use TextField, it would be great if we could consolidate how the start icon gets rendered! There were several considerations for styling the icon like the `disabled` and `light` props. One idea on how we could consolidate the styling would be to add a `startIcon` prop to `TextField` and it could provide the color, sizing, spacing. This could be useful for `TextField` on its own too! What do you think?
>    
>    * I was also wondering if a `endIcon` prop would also be useful for things like the Chevron icon, though it is an icon button so maybe the prop should support icon buttons too. Another use case would be the clear icon button in SearchField, or a show/hide icon button on a password field! The Combobox can potentially have 2 end icon buttons though for the clear and expand buttons, so perhaps it would take a React.Node instead of limiting it to Icon/IconButton!
> 2. For the combobox, would it be helpful for users to always have the magnifying glass icon there by default so there is a hint that they can type to search? Then teams can customize the icon for specific use cases if needed!

@beaesguerra Good points, here are my thoughts:

1. a. The `disabled` case makes a lot of sense. As for `light`, I think I are going to use it less, specially with the new experience coming later. One thing that I have been noticing in WB is that historically we have tend to allow passing `PhosphorIcon` instances as it provides more flexibility to implement different styles, and that's part of why I used that approach here. That means that the user can provide more control over `color, sizing, spacing` and even styling. I like the idea of centralizing that in `TextField`, but the only thing that gets me thinking is current usage, as I'm not sure if our form fields are using icons currently (specially text fields). This would be a great subject to bring to our WB cubby meeting to see what designers think. As for the API, how do you envision `startIcon` would work?

1.b. The original goal was to include `endIcon` with `Combobox` as well, but I decided to drop it to try to be consistent with the chevron icon button and prevent introducing any possible a11y issues by giving too much flexibility to that icon button. I think that in the end icon case, we should enforce our comboboxes to always look the same way (similar to what we do with other dropdowns like `SingleSelect` and `MultiSelect`).

2. I'd suggest to not use that icon by default as I think it could be confusing when the combobox it is in its selected state (for both single and multiple selection), and instead allowing to include it for very specific cases. For context on this API, I added this magnifying icon as a specific request for one of the TX flows: https://www.khanacademy.org/teacher/khanmigo-tools/lesson-plan?platform=KhanAcademy.

packages/wonder-blocks-dropdown/src/components/combobox.tsx
jandrade: oh oops, changing it! (I copy pasted from the impl. spec so I'm going to update it there as well.

jandrade: Fixed! btw I also modified the new story to include both the default and disabled state.

I'm thinking about creating an All Variants story to capture all the Combobox styles, but I'd do it in a separate PR. wdyt?

beaesguerra:
> 1a. As for light, I think I are going to use it less, specially with the new experience coming later.

👍  That simplifies things, great!

> 1a. As for the API, how do you envision startIcon would work?

`startIcon` in TextField would work similar to `startIcon` in Combobox in this PR! TextField would inject the color/size props into the PhosphorIcon. If we wanted it to be more flexible (ie. any svg can be passed in), we could inject styles to it, though I'm not sure how we could pass in aphrodite styles if the provided element isn't compatible with aphrodite 🤔 I like the approach of it expecting a PhosphorIcon! 

1.b. Oops I should've clarified! I was thinking about an `endIcon` prop for `TextField` since both SearchField and Combobox included things at the end of the field! We can see though if there's a need for this in TextField since this is lower priority :) I agree that Combobox doesn't need to have this flexibility since it should always have the expansion/clear buttons!

> I'd suggest to not use that icon by default as I think it could be confusing when the combobox it is in its selected state (for both single and multiple selection), and instead allowing to include it for very specific cases. For context on this API, I added this magnifying icon as a specific request for one of the TX flows: https://www.khanacademy.org/teacher/khanmigo-tools/lesson-plan?platform=KhanAcademy.

Got it, thanks for the context! Something else I'm thinking about is if there are things we can do to make sure icon usage is consistent! For example, when I updated SearchField, I updated it to use the `bold` icon to have it match Figma more closely (previously it was `regular`, similar to what's in Combobox in this PR! Let me know if I should change it back). This is also probably a bigger question related to how we use the design system in Figma too! Let me know if anyone has thoughts on this :) we can also solve for this later in the upcoming work!

packages/wonder-blocks-dropdown/src/components/combobox.tsx
```diff
@@ -477,6 +486,24 @@ export default function Combobox({
...
+    /**
+     * Renders the start icon if provided.
+     */
+    const maybeRenderStartIcon = () => {
+        if (!startIcon) {
+            return null;
+        }
+
+        const startIconElement = React.cloneElement(startIcon, {
+            ...startIcon.props,
```
beaesguerra, line 498: **suggestion:** (no changes necessary) What do you think about providing a default size and color if the `color` and `size` props aren't defined on the `startIcon`? That way, teams could pass in the PhosphorIcon without having to worry about what color and size to provide, and they can if they need to override it!

packages/wonder-blocks-dropdown/src/components/combobox.tsx
```diff
@@ -682,6 +711,12 @@ const styles = StyleSheet.create({
         border: `1px solid ${color.red}`,
         color: color.offBlack,
     },
+    /**
+     * Start icon styles
+     */
+    iconDisabled: {
```
beaesguerra, line 717: **nit:** I think this isn't used!

__docs__/wonder-blocks-dropdown/combobox.stories.tsx
```diff
@@ -428,3 +431,48 @@ export const Error: Story = {
...
+                        updateArgs({value: newValue});
+                        action("onChange")(newValue);
+                    }}
+                />
+            </View>
+        );
+    },
+    args: {
+        children: items,
+        startIcon: (
```
beaesguerra, line 470: Let's consolidate on the magnifying icon styling! In [SearchField](https://github.com/Khan/wonder-blocks/pull/2363/files#r1847290920), I updated the following to match the design more closely:
- `size=small`
- `color=semanticColor.icon.primary`
- use the bold icon instead `@phosphor-icons/core/bold/magnifying-glass-bold.svg`

Let me know though if we prefer to stick with the existing styles, I can update SearchField to change it back to how it was! (I was assuming Figma was the source of truth for styling, but that may not be the case!)

beaesguerra (approved):
Looks good, left some minor comments! Thanks for the discussion RE: icons in TextField! As discussed in the cubby meeting, we can solve for that later and see if we need it :)

packages/wonder-blocks-dropdown/src/components/combobox.tsx
jandrade: Love it!

marcysutton (approved):
This is looking great, @jandrade! Nice work!

__docs__/wonder-blocks-dropdown/combobox.stories.tsx
```diff
@@ -433,29 +433,73 @@ export const Error: Story = {
...
         const [_, updateArgs] = useArgs();
 
         return (
             <View style={{gap: spacing.medium_16}}>
-                <LabelMedium>Default:</LabelMedium>
+                <LabelMedium>With default size and color:</LabelMedium>
                 <Combobox
                     {...args}
+                    startIcon={
+                        <PhosphorIcon icon={magnifyingGlassIcon} size="small" />
```
beaesguerra, line 457: **suggestion:** We can remove `size="small"` here so we can confirm that the default size is being applied!

beaesguerra (approved):
Looks great! 🚀

__docs__/wonder-blocks-dropdown/combobox.stories.tsx
jandrade: Good catch! I'll fix it before landing.

PR #2365, opened 2024-11-18
SearchField: Remove -field postfix from id

beaesguerra:
# Summary:
Previously, the SearchField component was appending "-field" to the end of a provided id prop. We remove this post-fix and now use the `id` prop as is

This change is needed so when SearchField is used with LabeledField, LabeledField will set the id of the field element and wire it up with the label element. Since LabeledField works with a variety of components, the id shouldn't be modified by the field.

Issue: WB-1771

## Test plan:
- Confirm that the id on the search field is the same value as the id prop (`/?path=/story/packages-searchfield--default&args=id:test123`)
Screenshot 2024-11-18 at 3 08 19 PM
- Verify with the npm snapshot if there will be any breaking unit/e2e tests in webapp due to the internal id change

### DIFF
--- a/.changeset/cuddly-fireants-cheat.md
+++ b/.changeset/cuddly-fireants-cheat.md
@@ -1,5 +1,5 @@
 ---
-"@khanacademy/wonder-blocks-search-field": minor
+"@khanacademy/wonder-blocks-search-field": patch
 ---
 
 SearchField: Remove "-field" that was automatically appended to the id.
--- a/.changeset/cuddly-fireants-cheat.md
+++ b/.changeset/cuddly-fireants-cheat.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-search-field": minor
+---
+
+SearchField: Remove "-field" that was automatically appended to the id.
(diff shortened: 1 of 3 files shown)


beaesguerra:
Checks look okay when I tried the [npm snapshot in webapp](https://github.com/Khan/webapp/pull/27385)!

.changeset/cuddly-fireants-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-search-field": minor
```
jandrade, line 2: **nit:** This would probably account for a `patch` change, but I also understand if you want to keep it as `minor` as the API is being affected in some way (via the renderer HTML). Feel free to keep it or change it as you prefer.

jandrade (approved):
🎉

.changeset/cuddly-fireants-cheat.md
beaesguerra: Ah yes thanks for catching that! I think `patch` makes sense since the external api (props) hasn't changed

PR #2367, opened 2024-11-20
[🔥AUDIT🔥] Fix GitHub workflow to check for release PRs

jandrade:
🖍 _This is an audit!_ 🖍

## Summary:

There's an issue when a new Release Pull Request is created.

The problem is that one of the workflow job steps is failing because it cannot
find the base commit of the PR and it defaults to `00000000....00` instead.

https://github.com/Khan/wonder-blocks/actions/runs/11937708116/job/33275169358#step:5:23

This has been temporarily solved by pushing an empty commit to the PR branch
which causes the branch/PR to have clean its state and have a base commit.

### Possible cause:

My theory is that the PR/branch cannot determine the base commit because of the
way GH configures the COMMIT_SHA for this type of event (creating a PR/branch in
this case).

> Last commit on the created branch or tag.

My thinking is that the base commit is not being fetched because the workflow
automatically creates a new branch so there's no commit to compare it to.

For reference: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#create

### Solution:

This PR fixes the issue by adding a `fetch-depth` value to the `checkout` action in
the CI workflow. This will allow the workflow to fetch all the tags and commits
from the repository, so the base commit can be found.

Issue: XXX-XXXX

## Test plan:

Verify that there are no PR check errors the next time the "Release PR" is
created (unfortunately, this is the only way to test these type of changes).


beaesguerra (approved):
Looks good to me! Thanks for investigating!

To test if it works sooner, we could add an empty changeset using `yarn changeset --empty` so that when this PR gets merged, a changeset release PR is opened!

jandrade:
@beaesguerra thanks for the suggestion, but I now remember that creating empty changesets won't generate a new release PR. It was still worth trying :).

https://github.com/Khan/wonder-blocks/actions/runs/11940885877/job/33284531353#step:5:14

beaesguerra:
@jandrade ah good to know, thanks for trying it out!

marcysutton:
Looks like Changeset PRs are still failing: https://github.com/Khan/wonder-blocks/actions/runs/11977720772/job/33397027198?pr=2368

jandrade:
> Looks like Changeset PRs are still failing: https://github.com/Khan/wonder-blocks/actions/runs/11977720772/job/33397027198?pr=2368

@marcysutton yeah :(. I have another theory and it might be an issue with the `get-changed-files` custom action that seems to have an issue with reusable workflows. I'll try another solution later this week.

PR #2370, opened 2024-11-23
Convert MultiSelect and SingleSelect to functional components

beaesguerra:
## Summary:
Convert MultiSelect and SingleSelect to functional components to support the validation work to support LabeledField.

I was trying to add in the validation logic without refactoring too much since we might replace these components with Combobox. It was more complicated to account for different cases so I am refactoring it first to make it easier!

Issue: WB-1782

## Test plan:
- SingleSelect continues to work as expected
- MultiSelect continues to work as expected

(would appreciate help with extra testing in case I missed something. These components were a bit more complicated to convert to FC!)


.changeset/mean-sheep-shout.md
```diff
@@ -0,0 +1,5 @@
+---
```
beaesguerra, line 1: PR is a WIP - I'll be reviewing and testing more carefully since these components were a bit more complicated!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -171,33 +171,6 @@ type Props = AriaProps &
...
-    /**
-     * The object containing the custom labels used inside this component.
-     */
-    labels: Labels;
-    /**
-     * The DOM reference to the opener element. This is mainly used to set focus
-     * to this element, and also to pass the reference to Popper.js.
-     */
-    openerElement?: HTMLElement;
-}>;
```
beaesguerra, line 199: `open`, `searchText`, `lastSelectedValues` and `openerElement` have `useState` equivalents
https://github.com/Khan/wonder-blocks/blob/e02ada1e89fd30c08b34fabba5658e2547541640/packages/wonder-blocks-dropdown/src/components/multi-select.tsx#L223-L238

`labels` doesn't need to be in state and combines the default labels and prop labels: 
https://github.com/Khan/wonder-blocks/blob/e02ada1e89fd30c08b34fabba5658e2547541640/packages/wonder-blocks-dropdown/src/components/multi-select.tsx#L220-L221

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -217,74 +190,77 @@ type State = Readonly<{
  * </MultiSelect>
  * ```
  */
-export default class MultiSelect extends React.Component<Props, State> {
-    labels: Labels;
-
-    static defaultProps: DefaultProps = {
```
beaesguerra, line 223: Default props are set when we deconstruct props in the functional component

https://github.com/Khan/wonder-blocks/blob/e02ada1e89fd30c08b34fabba5658e2547541640/packages/wonder-blocks-dropdown/src/components/multi-select.tsx#L194-L218

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -217,74 +190,77 @@ type State = Readonly<{
...
-
-        this.state = {
-            open: false,
-            searchText: "",
-            lastSelectedValues: [],
-            // merge custom labels with the default ones
-            labels: {...defaultLabels, ...props.labels},
-        };
-        // merge custom labels with the default ones
-        this.labels = {...defaultLabels, ...props.labels};
```
beaesguerra, line 243: `labels` on the class wasn't being used so it can be removed. We combine default labels and prop labels:

https://github.com/Khan/wonder-blocks/blob/e02ada1e89fd30c08b34fabba5658e2547541640/packages/wonder-blocks-dropdown/src/components/multi-select.tsx#L220-L221

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -217,74 +190,77 @@ type State = Readonly<{
...
+         * Used to sync the `opened` state when this component acts as a controlled
+         * component
+         */
+        if (disabled) {
             // open should always be false if select is disabled
-            open: props.disabled
-                ? false
-                : typeof props.opened === "boolean"
-                ? props.opened
-                : state.open,
```
beaesguerra, line 260: The logic for `getDerivedStateFromProps` is translated to a useEffect:
- If `disabled` prop is set to `true`, we update the local `open` state to `false`.
- If the `opened` prop changes, we update the local `open` state to the value of the prop
- Otherwise, don't change local `open` state

https://github.com/Khan/wonder-blocks/blob/32b4f84272bf6c7b98e904fb9a87260f6ae467fd/packages/wonder-blocks-dropdown/src/components/multi-select.tsx#L240-L248

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -217,74 +190,74 @@ type State = Readonly<{
...
-        };
-    }
-
-    componentDidUpdate(prevProps: Props) {
-        if (this.props.labels !== prevProps.labels) {
-            // eslint-disable-next-line react/no-did-update-set-state
-            this.setState({
-                labels: {...this.state.labels, ...this.props.labels},
-            });
```
beaesguerra, line 269: Labels will stay up to date since it will combine the default labels and the prop labels

https://github.com/Khan/wonder-blocks/blob/32b4f84272bf6c7b98e904fb9a87260f6ae467fd/packages/wonder-blocks-dropdown/src/components/multi-select.tsx#L220-L221

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -449,79 +417,50 @@ export default class MultiSelect extends React.Component<Props, State> {
...
         dropdownId: string,
     ):
         | React.ReactElement<React.ComponentProps<typeof DropdownOpener>>
-        | React.ReactElement<React.ComponentProps<typeof SelectOpener>> {
-        const {
-            id,
-            light,
-            opener,
-            testId,
-            // the following props are being included here to avoid
```
beaesguerra, line 502: Since we deconstruct the props at the start of the function, we don't need to again in the renderOpener function. 
https://github.com/Khan/wonder-blocks/blob/32b4f84272bf6c7b98e904fb9a87260f6ae467fd/packages/wonder-blocks-dropdown/src/components/multi-select.tsx#L194-L218

The deconstruction previously is almost the same. The FC will also extract `disabled`, `error`, `children`, and `dropdownId` so they won't be included in `sharedProps` when it is spread. The only one that matters to the SelectOpener is `error`, which is why we also explicitly set the `error` prop: https://github.com/Khan/wonder-blocks/pull/2370/files#diff-f7d0f4de4f6447c1f878dd6c51e6646adae276658cf9d7f29fcea7f630b17430R484

(changes for the `disabled` doesn't matter since we were already explicitly setting it https://github.com/Khan/wonder-blocks/pull/2370/files#diff-f7d0f4de4f6447c1f878dd6c51e6646adae276658cf9d7f29fcea7f630b17430R485)

packages/wonder-blocks-dropdown/src/components/single-select.tsx
```diff
@@ -48,16 +48,16 @@ type DefaultProps = Readonly<{
      * Whether this dropdown should be left-aligned or right-aligned with the
      * opener component. Defaults to left-aligned.
      */
-    alignment: "left" | "right";
```
beaesguerra, line 51: Most of the changes in SingleSelect are similar to the ones annotated in MultiSelect. Will highlight specific changes related to SingleSelect

packages/wonder-blocks-dropdown/src/components/single-select.tsx
```diff
@@ -225,99 +208,99 @@ type State = Readonly<{
...
             filter: defaultLabels.filter,
             noResults: defaultLabels.noResults,
             someResults: defaultLabels.someSelected,
         },
-    };
-
-    constructor(props: Props) {
-        super(props);
-
-        this.selectedIndex = 0;
```
beaesguerra, line 249: We use a ref for this now: 
https://github.com/Khan/wonder-blocks/blob/32b4f84272bf6c7b98e904fb9a87260f6ae467fd/packages/wonder-blocks-dropdown/src/components/single-select.tsx#L212

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
jandrade: yes! one benefit of switching to hooks. These kind of lifecycle events get more simpler.

jandrade (approved):
This is awesome!! thanks for all the detailed notes. I did a lot of manual testing and can confidently say that works as expected (also tests passing is a good signal) :shipit: 👏

PR #2371, opened 2024-11-26
[🔥AUDIT🔥] Fix ngettext not returning plural translations.

jeresig:
🖍 _This is an audit!_ 🖍

## Summary:
This bug is in prod right now, unfortunately. Some strings are appearing untranslated.

Issue: FEI-6007

## Test plan:
This was actually wrong in the tests!! So I fixed the test in the process.

### DIFF
--- a/packages/wonder-blocks-i18n/src/functions/i18n.ts
+++ b/packages/wonder-blocks-i18n/src/functions/i18n.ts
@@ -210,7 +210,7 @@ export const ngettext: ngettextOverloads = (
         typeof singular === "object"
             ? singular
             : {
-                  lang: "en",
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-i18n/src/functions/i18n.ts
```diff
@@ -210,7 +210,7 @@ export const ngettext: ngettextOverloads = (
         typeof singular === "object"
             ? singular
             : {
-                  lang: "en",
+                  lang: getLocale(),
```
somewhatabstract, line 213: **note:** Aha! I wondered about this in the previous audit.

PR #2374, opened 2024-11-27
[🔥AUDIT🔥] Fix up plural form fallback when no translation is present.

jeresig:
🖍 _This is an audit!_ 🖍

## Summary:
I may have introduced this in my last fix, actually - but this now helps us ensure that we don't accidentally use the other language pluralization rules when we are falling back to English strings.

Issue: FEI-6007

## Test plan:
The tests pass!

### DIFF
--- a/packages/wonder-blocks-i18n/src/functions/__tests__/i18n.test.ts
+++ b/packages/wonder-blocks-i18n/src/functions/__tests__/i18n.test.ts
@@ -142,6 +142,17 @@ describe("i18n", () => {
             expect(result).toMatchInlineSnapshot(`"arrrr mateys"`);
         });
 
+        it("ngettext should handle missing translations", () => {
+            // Arrange
+            jest.spyOn(Locale, "getLocale").mockImplementation(() => "ru");
+
+            // Act
+            const result = ngettext("Singular", "Plural", 0);
+
+            // Assert
+            expect(result).toMatchInlineSnapshot(`"Plural"`);
+        });
+
         it("doNotTranslate should not translate", () => {
             // Arrange
             loadTranslations(TEST_LOCALE, {
--- a/packages/wonder-blocks-i18n/src/functions/i18n-store.ts
+++ b/packages/wonder-blocks-i18n/src/functions/i18n-store.ts
@@ -5,7 +5,10 @@
  */
 import FakeTranslate from "./i18n-faketranslate";
 import {getLocale} from "./locale";
+import {allPluralForms} from "./plural-forms";
 import {PluralConfigurationObject} from "./types";
+
+type Language = keyof typeof allPluralForms;
 
 // The cache of strings that have been translated, by locale.
 const localeMessageStore = new Map<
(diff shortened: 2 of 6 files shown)


packages/wonder-blocks-i18n/src/functions/__tests__/i18n.test.ts
```diff
@@ -142,6 +142,17 @@ describe("i18n", () => {
...
+        it("ngettext should handle missing translations", () => {
+            // Arrange
+            jest.spyOn(Locale, "getLocale").mockImplementation(() => "ru");
+
+            // Act
+            const result = ngettext("Singular", "Plural", 0);
+
+            // Assert
+            expect(result).toMatchInlineSnapshot(`"Plural"`);
+        });
```
somewhatabstract, line 154: **praise:** Thanks for adding an additional test to cover this!

packages/wonder-blocks-i18n/src/functions/i18n-store.ts
```diff
@@ -18,6 +21,20 @@ const localeMessageStore = new Map<
 // Create a fake translate object to use if we can't find a translation.
 const {translate: fakeTranslate} = new FakeTranslate();
 
+/*
+ * Return the ngettext position that matches the given number and lang.
+ *
+ * Arguments:
+ *  - num: The number upon which to toggle the plural forms.
+ *  - lang: The language to use as the basis for the pluralization.
```
somewhatabstract, line 29: **suggestion:** Use JSDoc syntax so that it just works with IDEs and tooling that support it.

```suggestion
 * 
 * @param num The number upon which to toggle the plural forms.
 * @param lang The language to use as the basis for the pluralization.
 * @return The position that matches the given number and language.
```

packages/wonder-blocks-i18n/src/functions/i18n-store.ts
```diff
@@ -18,6 +21,20 @@ const localeMessageStore = new Map<
 // Create a fake translate object to use if we can't find a translation.
 const {translate: fakeTranslate} = new FakeTranslate();
 
+/*
+ * Return the ngettext position that matches the given number and lang.
```
somewhatabstract, line 25: **suggestion:** This doesn't quite fit our conventions for this type of comment. I might consider a tweak.

```suggestion
 * Get the ngettext position that matches a given number and lang.
```

somewhatabstract: Oh, I see this was moved code. Well, the point stands, but I'm not surprised this audit didn't change it.

packages/wonder-blocks-i18n/src/functions/i18n-store.ts
somewhatabstract: Oh, I see this was moved code. Well, the point stands, but I'm not surprised this audit didn't change it.

packages/wonder-blocks-i18n/src/functions/i18n.ts
```diff
@@ -210,7 +207,7 @@ export const ngettext: ngettextOverloads = (
         typeof singular === "object"
             ? singular
             : {
-                  lang: getLocale(),
+                  lang: "en",
```
somewhatabstract, line 210: **question:** This is using `"en"` because the tooling works from that locale as the source locale?

packages/wonder-blocks-i18n/src/functions/i18n.ts
```diff
@@ -245,20 +238,6 @@ const formatNumber = (num: number): string => {
...
- * Arguments:
- *  - num: The number upon which to toggle the plural forms.
- *  - lang: The language to use as the basis for the pluralization.
- */
-export const ngetpos = function (num: number, lang?: Language): number {
-    const pluralForm = (lang && allPluralForms[lang]) || allPluralForms["en"];
-    const pos = pluralForm(num);
-    // Map true to 1 and false to 0, keep any numeric return value the same.
-    return pos === true ? 1 : pos ? pos : 0;
-};
```
somewhatabstract, line 260: **note:** I see now that this is where the function was original defined before being moved to `i18n-store.ts`.

PR #2378, opened 2024-12-02
Add validation props to MultiSelect and share logic with SingleSelect

beaesguerra:
## Summary:
- Refactor SingleSelect validation logic to useSelectValidation hook
- Adding validation related props to MultiSelect: validate, onValidate, required. (error prop was already supported)
  - Make sure aria-invalid is set if it is in an error state

Issue: WB-1782

## Test plan:
- MultiSelect docs are reviewed `?path=/docs/packages-dropdown-multiselect--docs`
- Validation works as expected in MultiSelect (see docs for more details on validation behaviour):
  - Error (`?path=/story/packages-dropdown-multiselect--error`)
  - Required (`?path=/story/packages-dropdown-multiselect--required`)
  - Error from Validation (`?path=/story/packages-dropdown-multiselect--error-from-validation`)
- MultiSelect continues to work as expected (including keyboard interactions)
- SingleSelect continues to work as expected (including validation)

### DIFF
--- a/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
@@ -271,36 +271,11 @@ export const CustomStylesOpened: StoryComponentType = {
     ],
 };
 
-const ErrorWrapper = (args: any) => {
-    const [selectedValues, setSelectedValues] = React.useState<string[]>([]);
-    const [opened, setOpened] = React.useState(false);
-    const [error, setError] = React.useState(true);
-
-    return (
-        <>
-            <LabelMedium style={{marginBottom: spacing.xSmall_8}}>
-                Select at least 2 options to clear the error!
-            </LabelMedium>
-            <MultiSelect
-                {...args}
-                error={error}
-                onChange={(values) => {
-                    setSelectedValues(values);
-                    setError(values.length < 2);
-                }}
-                onToggle={setOpened}
-                opened={opened}
-                selectedValues={selectedValues}
-            >
-                {items}
-            </MultiSelect>
-        </>
-    );
-};
-
 const ControlledMultiSelect = (args: PropsFor<typeof MultiSelect>) => {
     const [opened, setOpened] = React.useState(false);
-    const [selectedValues, setSelectedValues] = React.useState<string[]>([]);
+    const [selectedValues, setSelectedValues] = React.useState<string[]>(
+        args.selectedValues || [],
+    );
     const [errorMessage, setErrorMessage] = React.useState<
         null | string | void
     >(null);
@@ -395,9 +370,9 @@ export const Required: StoryComponentType = {
  *
  * Validation is triggered:
  * - On mount if the `value` prop is not empty and it is not required
- * - When an option is selected
+ * - When the dropdown is closed after updating the selected values
  *
- * Validation errors are cleared when a valid value is selected. The component
+ * Validation errors are cleared when the value is updated. The component
  * will set aria-invalid to "false" and call the onValidate prop with null.
  */
 export const ErrorFromValidation: StoryComponentType = {
--- a/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
@@ -129,14 +129,14 @@ const styles = StyleSheet.create({
 });
 
 const items = [
-    <OptionItem label="Mercury" value="1" key={1} />,
-    <OptionItem label="Venus" value="2" key={2} />,
-    <OptionItem label="Earth" value="3" disabled key={3} />,
-    <OptionItem label="Mars" value="4" key={4} />,
-    <OptionItem label="Jupiter" value="5" key={5} />,
-    <OptionItem label="Saturn" value="6" key={6} />,
-    <OptionItem label="Neptune" value="7" key={7} />,
-    <OptionItem label="Uranus" value="8" key={8} />,
+    <OptionItem label="Mercury" value="mercury" key={1} />,
+    <OptionItem label="Venus" value="venus" key={2} />,
+    <OptionItem label="Earth" value="earth" disabled key={3} />,
+    <OptionItem label="Mars" value="mars" key={4} />,
+    <OptionItem label="Jupiter" value="jupiter" key={5} />,
+    <OptionItem label="Saturn" value="saturn" key={6} />,
+    <OptionItem label="Neptune" value="neptune" key={7} />,
+    <OptionItem label="Uranus" value="uranus" key={8} />,
 ];
 
 const Template = (args: any) => {
@@ -289,7 +289,7 @@ const ControlledMultiSelect = (args: PropsFor<typeof MultiSelect>) => {
                 selectedValues={selectedValues}
                 onChange={setSelectedValues}
                 validate={(values) => {
-                    if (values.includes("5")) {
+                    if (values.includes("jupiter")) {
                         return "Don't pick jupiter!";
                     }
                 }}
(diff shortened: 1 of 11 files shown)


packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
```diff
@@ -2098,4 +2100,919 @@ describe("MultiSelect", () => {
...
+                await userEvent.tab();
+
+                // Assert
+                expect(onValidate).toHaveBeenCalledExactlyOnceWith(
+                    "This field is required.",
+                );
+            });
+        });
+
+        describe("validate and required props", () => {
```
beaesguerra, line 2944: SingleSelect doesn't have these tests since we don't have a way to clear/unselect the selected value

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -478,12 +547,16 @@ const MultiSelect = (props: Props) => {
...
             <IDProvider id={id} scope="multi-select-opener">
                 {(uniqueOpenerId) => {
+                    const hasAriaRequired = ariaRequired || !!required;
                     return opener ? (
                         <DropdownOpener
                             id={uniqueOpenerId}
+                            error={hasError}
                             aria-controls={dropdownId}
                             aria-haspopup="listbox"
+                            aria-required={hasAriaRequired}
```
beaesguerra, line 557: I'm going to remove setting `aria-required` since the opener is a button and aria-required isn't allowed on buttons 😅 Found this in failing jest-axe tests while testing the snapshot in webapp (I'll add some jest-axe tests for the dropdown components too!) 

If we switch these to use role=combobox instead, [aria-required would be supported](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-required#associated_roles)! (WB-1671)

beaesguerra: Removed setting aria-required and added jest-axe tests! I did the same thing for SingleSelect in the [SingleSelect validation PR](https://github.com/Khan/wonder-blocks/pull/2376/commits/7f9b154c4ef875f3b47281343af4b92b04e6e56b)

I created WB-1807 to set `aria-required` once WB-1671 is done!

packages/wonder-blocks-dropdown/src/hooks/use-select-validation.tsx
```diff
@@ -0,0 +1,145 @@
...
+        handleValidation(newValue);
+    };
+
+    const onSelectedValuesChangeValidation = () => {
+        // When selected values change, clear the error message
+        // This is so errors aren't shown to the user while they update the
+        // selected values. Note that this should only be called when there are
+        // multiple values that can be selected. onSelectionValidation should be
+        // used whenever the user is done selecting a value or values.
+        setErrorMessage(null);
```
jandrade, line 132: **thought:** I wonder if we want to keep this behavior here. I was testing this [here](https://5e1bf4b385e3fb0020b7073c-njsnevcwlb.chromatic.com/?path=/story/packages-dropdown-multiselect--error-from-validation), and I'm not sure if it would make sense to reset the error state to later show it again when the listbox is closed. How do you feel about adding a check here to conditionally reset the error?

 

https://github.com/user-attachments/assets/593e152d-786d-42a0-aa4f-ff8f85bf07ad

jandrade (commented):
This looks really solid!. This is my first pass, but wanted to share for now specially for getting your thoughts on a question I have in the `useSelectValidation` hook.

__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
```diff
@@ -271,42 +271,129 @@ export const CustomStylesOpened: StoryComponentType = {
...
                 {items}
             </MultiSelect>
-        </>
+            {(errorMessage || args.error) && (
+                <LabelMedium
+                    style={{color: semanticColor.status.critical.foreground}}
+                >
+                    {errorMessage || "Error from error prop"}
+                </LabelMedium>
+            )}
```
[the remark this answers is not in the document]
beaesguerra: Yes! Will go through and update the form field stories to use LabeledField (LabeledField is still in a feature branch not in master yet so I'll work on integrating it all next!)

__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
```diff
@@ -271,42 +271,129 @@ export const CustomStylesOpened: StoryComponentType = {
...
+ * will set aria-invalid to "false" and call the onValidate prop with null.
+ */
+export const ErrorFromValidation: StoryComponentType = {
+    render: (args: PropsFor<typeof MultiSelect>) => {
+        return (
+            <View style={{gap: spacing.xSmall_8}}>
+                <LabelMedium htmlFor="multi-select" tag="label">
+                    Validation example (try picking jupiter)
+                </LabelMedium>
+                <ControlledMultiSelect {...args} id="multi-select">
```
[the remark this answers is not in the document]
beaesguerra: Configured the argTypes for `selectedValues`! I also changed the `value` of the items to strings (ie. `jupiter`) because when entering an array like `["1"]` in the Storybook controls, the quotations would get stripped in the url parameters and it becomes `[1]`, which means the item values don't match up with the selectedValues!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
beaesguerra: @jandrade FYI! I realize now that this thread doesn't show in the PR diff since it's on code I removed 😅

__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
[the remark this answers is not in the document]
jandrade: That makes sense. I tested the story in Storybook and looks great!

__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
```diff
@@ -271,42 +271,139 @@ export const CustomStylesOpened: StoryComponentType = {
...
+                >
+                    {items}
+                </ControlledMultiSelect>
+            </View>
+        );
+    },
+    parameters: {
+        chromatic: {
+            // Disabling because this doesn't test anything visual.
+            disableSnapshot: true,
```
jandrade, line 404: **question:** Now that you added the on mount example, I wonder if we want to take a snapshot here. If it's covered in a separate story, it's fine to keep it as it is.

packages/wonder-blocks-dropdown/src/hooks/use-select-validation.tsx
jandrade: would calling `validate()` here make sense? Although, that would mean that we are calling `instantValidation` here.

__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
beaesguerra: Good point! The error state styling is also covered by the All Variants stories, though it uses the `error` prop instead. There are unit tests that verify it's in an error state but won't check for styling. I've removed disabling the stories with the mount validation example!

packages/wonder-blocks-dropdown/src/hooks/use-select-validation.tsx
jandrade: After chatting offline, we decided to keep things as they are. Thanks for talking through this @beaesguerra .

jandrade (approved):
LGTM 🚀

beaesguerra:
The parent pull-request ([#2376](https://github.com/Khan/wonder-blocks/pull/2376)) has been merged into `main`, but this branch (`dropdown-multi-select-validation`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #2383, opened 2024-12-06
WB-1796: Change modal default focus to be the Close (X) button

jandrade:
## Summary:

Today, the default for a modal is to focus the first focusable element, but we
exclude the top close button. What this often means is that the first button in
the footer area of the modal is focused.

This PR changes the default focus to be the top close button. This is a more
consistent experience as it allows to focus in the first focusable element in
the dialog content area.

Also had to modify the `IconButton` styles to make sure the focus ring in close
button is visible when the focus is applied programmatically.

Note that we could use `node.focus({focusVisible: true})` to make the focus
visible, but this is not supported in all browsers. So, we are using the
`:focus` pseudo-class to style the focus ring in these cases.

Issue: https://khanacademy.atlassian.net/browse/WB-1796

## Test plan:

Navigate to the `ModalLauncher` docs: /?path=/docs/packages-modal-modallauncher--__docs__

1. Open the modal by clicking the "Open Modal" button.
2. Verify that the dismiss button is focused when the modal is opened.

Screenshot 2024-12-06 at 5 35 38 PM

Also verify that the focus state works as expected in the `IconButton` stories.

### DIFF
--- a/packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
+++ b/packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
@@ -322,15 +322,6 @@ const _generateStyles = (
         outlineColor: disabledStrokeColor,
     };
 
-    const focusStyles = {
-        outlineWidth: theme.border.width.default,
-        outlineColor: defaultStrokeColor,
-        outlineOffset: 1,
-        outlineStyle: "solid",
-        borderRadius: theme.border.radius.default,
-        ...kindOverrides[":focus-visible"],
-    };
-
     const newStyles = {
         default: {
             height: pixelsForSize,
@@ -362,11 +353,16 @@ const _generateStyles = (
                     backgroundColor: "transparent",
                 },
             },
-            // :focus -> Provide basic, default focus styles to support
-            // programmatic focus (e.g. element.focus()).
-            ":focus": focusStyles,
+
             // :focus-visible -> Provide focus styles for keyboard users only.
-            ":focus-visible": focusStyles,
+            ":focus-visible": {
+                outlineWidth: theme.border.width.default,
+                outlineColor: defaultStrokeColor,
+                outlineOffset: 1,
+                outlineStyle: "solid",
+                borderRadius: theme.border.radius.default,
+                ...kindOverrides[":focus-visible"],
+            },
             ":active": {
                 color: light ? activeInverseColor : activeColor,
                 outlineWidth: theme.border.width.default,
@@ -388,7 +384,6 @@ const _generateStyles = (
             // For order reference: https://css-tricks.com/snippets/css/link-pseudo-classes-in-order/
             ":hover": {...disabledStatesStyles, outline: "none"},
             ":active": {...disabledStatesStyles, outline: "none"},
-            ":focus": disabledStatesStyles,
             ":focus-visible": disabledStatesStyles,
         },
     } as const;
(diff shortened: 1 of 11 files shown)


packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
```diff
@@ -353,29 +362,11 @@ const _generateStyles = (
...
-                outlineOffset: 1,
-                outlineStyle: "solid",
-                borderRadius: theme.border.radius.default,
-                ...kindOverrides[":focus-visible"],
-            },
+            // :focus -> Provide basic, default focus styles to support
+            // programmatic focus (e.g. element.focus()).
+            ":focus": focusStyles,
+            // :focus-visible -> Provide focus styles for keyboard users only.
+            ":focus-visible": focusStyles,
```
jandrade, line 369: **note:** I refactored this to use `:focus` for when we use `.focus()` and `:focus-visible` for when the focus happens via keyboard.

packages/wonder-blocks-modal/src/util/find-focusable-nodes.ts
```diff
@@ -2,8 +2,7 @@
  * List of elements that can be focused
  * @see https://www.w3.org/TR/html5/editing.html#can-be-focused
  */
-const FOCUSABLE_ELEMENTS =
-    'a[href], details, input, textarea, select, button:not([aria-label^="Close"])';
+const FOCUSABLE_ELEMENTS = "a[href], details, input, textarea, select, button";
```
jandrade, line 5: **note:** This is the main change in this PR. By removing that custom selector, we can include the dismiss button in the collection.

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
```diff
@@ -353,29 +362,11 @@ const _generateStyles = (
...
-                outlineWidth: theme.border.width.default,
-                outlineColor: defaultStrokeColor,
-                outlineOffset: 1,
-                outlineStyle: "solid",
-                borderRadius: theme.border.radius.default,
-                ...kindOverrides[":focus-visible"],
-            },
+            // :focus -> Provide basic, default focus styles to support
+            // programmatic focus (e.g. element.focus()).
+            ":focus": focusStyles,
```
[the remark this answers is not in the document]
marcysutton: Normally it could be an issue for `:focus` to persist on click as sighted users sometimes complain about it...but in this case clicking the button closes the modal, and the entire thing goes away. I agree that it might make sense to limit this change to the modal close buttons, and not `icon-button-core`. That would require quite a bit more testing!

packages/wonder-blocks-modal/src/themes/default.ts
```diff
@@ -15,6 +15,8 @@ const theme = {
     },
     border: {
         radius: tokens.border.radius.medium_4,
+        width: tokens.border.width.thin,
+        color: tokens.color.blue,
```
jandrade, line 19: **note:** Added these two new component tokens to be used in the dismiss button for the focus ring

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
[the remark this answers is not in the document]
jandrade: ok, I've applied the changes, let me know what you think.

beaesguerra (approved):
Looks good! Thanks for cleaning up the safari 14 focus styles too!

PR #2386, opened 2024-12-11
Remove wonder-blocks-i18n

jandrade:
## Summary:

Now that we have adopted Lingui, we no longer need our custom i18n solution.
This PR removes the wonder-blocks-i18n package.

Issue: XXX-XXXX

## Test plan:

Verify that all checks pass


jeresig (approved):
Woohoo! Thank you!

PR #2387, opened 2024-12-11
[wb1812.0.fixrenderstate] Make sure RenderState.Root is an internal concept only

somewhatabstract:
## Summary:
This addresses an issue where `useRenderState` would return `RenderState.Root` during initial render to components nested inside our `RenderStateRoot` component, when they should only ever see `Initial` or `Standard`.

I suspect this was happening due to hook execution order and render order. React component render functions are executed from the inside out, and I suspect that because of this, the initial call to `useRenderState` is using the default context value, because the root component hasn't rendered it's value yet - a general flaw in hook-based context access.

However, for everything but `RenderStateRoot` and `InitialFallback` components, code doesn't need to know if it's the root render or not. It's an easier API if they just always see `Initial` or `Standard`. This change makes that happen. The only components that need to know are those that need to render the actual context, and our consumers don't need to do that.

This addresses an issue where the `useUniqueId` hook and it's initial render fallback version would throw an error on initial render. It's a precursor PR to unblock folks while we work on replacing our unique ID stuff with `useId`.

Issue: WB-1812

## Test plan:
`yarn test`
`yarn typecheck`

I also checked our Wonder Blocks consumers to verify that they don't care about the "root" render state (as they shouldn't).

## Release Info:
This is a major release of Core because we are changing the `RenderState` enum to remove a value, and changing the behavior of the `useRenderState` hook. In real terms, consumers should be unaffected, but changing exports like this should be a major release so following protocol.

### DIFF
--- a/packages/wonder-blocks-core/src/components/render-state-context.ts
+++ b/packages/wonder-blocks-core/src/components/render-state-context.ts
@@ -1,13 +1,57 @@
 import * as React from "react";
 
+/**
+ * The possible states of rendering.
+ *
+ * This is used to determine if we are rendering our initial, hydrateable state
+ * or not. Initial renders must be consistent between the server and client so
+ * that hydration will succeed.
+ *
+ * We use a render state like this instead of a simple check for being mounted
+ * or not, or some other way of each component knowing if it is rendering itself
+ * for the first time so that we can avoid cascading initial renders where each
+ * component has to render itself and its children multiple times to reach a
+ * stable state. Instead, we track the initial render from the root of the tree
+ * and switch everything accordingly so that there are fewer additional renders.
+ */
 export enum RenderState {
+    /**
+     * The initial render, either on the server or client.
+     */
     Initial = "initial",
+
+    /**
+     * Any render after the initial render. Only occurs on the client.
+     */
     Standard = "standard",
 }
 
+/**
+ * The internal states of rendering.
+ *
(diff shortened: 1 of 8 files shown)


packages/wonder-blocks-core/src/components/render-state-context.ts
```diff
@@ -1,6 +1,11 @@
 import * as React from "react";
 
 export enum RenderState {
+    Initial = "initial",
+    Standard = "standard",
+}
+
+export enum RenderStateInternal {
```
jandrade, line 8: **nit:** This is explicit in the context of this PR, but I think it would be useful adding a comment in the code describing why we have two enums and what the differences are.

packages/wonder-blocks-core/src/hooks/use-unique-id.ts
```diff
@@ -50,12 +44,6 @@ export const useUniqueIdWithoutMock = (
     const renderState = useRenderState();
     const idFactory = useRef<IIdentifierFactory | null | undefined>(null);
 
-    if (renderState === RenderState.Root) {
-        throw new Error(
-            "Components using useUniqueIdWithoutMock() should be descendants of <RenderStateRoot>",
-        );
-    }
-
```
jandrade, line 58: **praise:** It is great seeing how this got simplified!!

jandrade (approved):
This looks great! thanks for simplifying the public API 🚢

PR #2388, opened 2024-12-11
[wb1812.1.deprecate] Mark ID stuff as deprecated

somewhatabstract:
## Summary:
This is the first step in removing our custom ID utilities in favour of using the `useId` hook from React's API. This PR is not going to be merged/released on its own. Another PR will be coming to add an `Id` component that wraps `useId` so that there's an easier path for consumers to migrate.

The difficult bit is moving from the provider architecture that gives an ID factory, to the `useId`/`Id` approach. However, long term, it should simplify things greatly.

### Upgrade note:
Upgrading to this will cause deprecation notices in consuming code. If a lint rule is set up to block deprecated usage (which it probably is), then those errors will have to be suppressed. A find/replace may help in some circumstances, but this may also be a manual process. The alternative, of course, is to actually modify stuff to the new approach.

Issue: WB-1812

## Test plan:
`yarn test`
`yarn lint`

### DIFF
--- a/.changeset/light-goats-serve.md
+++ b/.changeset/light-goats-serve.md
@@ -1,9 +1,12 @@
 ---
 "@khanacademy/wonder-blocks-core": major
 "@khanacademy/wonder-blocks-search-field": patch
+"@khanacademy/wonder-blocks-accordion": patch
 "@khanacademy/wonder-blocks-dropdown": patch
 "@khanacademy/wonder-blocks-popover": patch
+"@khanacademy/wonder-blocks-testing": patch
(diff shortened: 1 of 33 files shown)


jandrade (approved):

somewhatabstract:
The parent pull-request ([#2387](https://github.com/Khan/wonder-blocks/pull/2387)) has been merged into `main`, but this branch (`wb1812.1.deprecate`) now has conflicts with the new base branch. These conflicts must be resolved before checks can complete on this pull-request.

PR #2389, opened 2024-12-11
[wb1812.2.idcomponent] Add the Id component

somewhatabstract:
## Summary:
This adds a simple CaaF (children-as-a-function) component to generate unique IDs as a stand-in for the `useId` hook. This is useful for cases where one needs to generate unique IDs for a component, but cannot use the hook without a larger refactor.

Issue: WB-1812

## Test plan:
`yarn test`
`yarn start:storybook` to check for the added docs
`yarn typecheck`


jeresig (approved):
Looks simple enough, thank you!

PR #2391, opened 2024-12-11
[wb1812.3.migratewb] Migrate Wonder Blocks off old id providers

somewhatabstract:
## Summary:
This is the last piece in the first batch of work. This migrates all Wonder Blocks components off our old ID providers and onto the new `Id` component.

There are also some documentation tweaks to make the deprecation clearer in our stories, since that's some primary documentation for folks.

With this PR, we can cut a release and then begin updating consumer repos accordingly. First, to include this update, then to migrate them off the old ways and onto the new.

### Release process:
Once this small stack of PRs are landed and released, the following PRs need to be updated with that release, then landed/deployed:

- Perseus: https://github.com/Khan/perseus/pull/2007
- Webapp
  - https://github.com/Khan/webapp/pull/28105
  - https://github.com/Khan/webapp/pull/28127

Issue: WB-1812

## Test plan:
`yarn test`
`yarn typecheck`
`yarn start:storybook`


packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -546,7 +545,7 @@ const MultiSelect = (props: Props) => {
         const menuText = getMenuText(allChildren);
 
         const dropdownOpener = (
-            <IDProvider id={id} scope="multi-select-opener">
+            <Id id={id}>
```
jandrade, line 548: **nit:** This is now a function component, so you could use `useId` directly as well. I understand if you want to keep things simple for this migration.

packages/wonder-blocks-dropdown/src/components/single-select.tsx
```diff
@@ -444,7 +443,7 @@ const SingleSelect = (props: Props) => {
             : placeholder;
 
         const dropdownOpener = (
-            <IDProvider id={id} scope="single-select-opener">
+            <Id id={id}>
```
jandrade, line 446: Same suggestion about `useId` (no changes required).

jandrade (approved):
Looks great to me! thanks for the quick refactor :shipit:

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
somewhatabstract: Yeah, I considered that, but the component seems so complex; this felt cleaner and a little easier. :shrug:

jandrade: makes total sense :)

packages/wonder-blocks-dropdown/src/components/combobox.tsx
```diff
@@ -684,7 +680,6 @@ export default function Combobox({
                                         testId ? `${testId}-listbox` : undefined
                                     }
                                     aria-label={labels.listbox}
-                                    aria-labelledby={ids.get("input")}
```
somewhatabstract, line 687: **question:** @jandrade seems like this was a bug? There's already an `aria-label` attribute, and setting `aria-labelledby` to an ID that isn't attached to any element feels like it'll just confuse screen reader tech and browsers.

somewhatabstract: For the record, there was no bug here, I don't think. We're using label and labelledby, which may be an oversight, but it should be fine as-is.

PR #2397, opened 2024-12-16
LabeledField: integrate with field components and fixes

beaesguerra:
## Summary:
Putting LabeledField work together with field updates and addressing some issues. Fixes include:

TextField and TextArea:
- Set `aria-required` based on `required` prop
- Always clear error state onChange if instantValidation is false so that external error messages set using the `error` prop can still be cleared (This is useful for clearing any backend errors when the input value changes)

LabeledField updates:
- Let `required` prop be a boolean or string so it can be passed down to the field prop
- Conditionally set spacing above the error message depending on if there is an error message
- Set `required`, `error` and `light` props for LabeledField and field component if it is set on either LabeledField or field component
- Rename `error` prop to `errorMessage` (This technically isn't a breaking change since the batch of labeled field work hasn't been released yet, labeled field work is currently in the `feature/labeled-field` branch)

Stories/docs:
- Add docs on best practices for forms and fields
- Add docs around LabeledTextField deprecation in favour of LabeledField + TextField
- Update LabeledField examples and docs. Includes examples for validation and moving focus to the first field that has an error message after form submission

Issue: WB-1783

## Test plan:
1. Review new docs:
- LabeledField (`?path=/docs/packages-labeledfield--docs`)
- LabeledTextField (`/?path=/docs/packages-form-labeledtextfield-deprecated--docs`)
- Form Best Practices (`?path=/docs/packages-form-overview--docs`)

2. Test LabeledField with Field Validation (including screen reader usability)
Confirm validation works as expected
- LabeledField Required story (`?path=/story/packages-labeledfield--required`)
- LabeledField Validation story (`?path=/story/packages-labeledfield--validation`)


beaesguerra:
Cleaning up commits, will create a different PR

PR #2398, opened 2024-12-16
[wb1812.4.delete] Delete the custom identifier generation API

somewhatabstract:
## Summary:
This deletes the components and hooks that we had for generating hydration-safe identifiers. We are now using the built-in `useId` hook internally, and expect our consumers to do the same (or the `Id` component we provide for more complex cases).

This should not be released until the following PRs are merged and deployed to webapp:
 - https://github.com/Khan/webapp/pull/28161
 - https://github.com/Khan/webapp/pull/28156

Issue: WB-1812

## Test plan:
`yarn test`
`yarn typecheck`


beaesguerra (approved):
Looks good to me! 🚀

PR #2399, opened 2024-12-16
LabeledField: integrate with field components and fixes

beaesguerra:
## Summary:
Putting LabeledField work together with field updates and addressing some issues. Fixes include:

TextField and TextArea:
- Set `aria-required` based on `required` prop
- Always clear error state onChange if instantValidation is false so that external error messages set using the `error` prop can still be cleared (This is useful for clearing any backend errors when the input value changes)

LabeledField updates:
- Let `required` prop be a boolean or string so it can be passed down to the field prop
- Conditionally set spacing above the error message depending on if there is an error message
- Set `required`, `error` and `light` props for LabeledField and field component if it is set on either LabeledField or field component
- Rename `error` prop to `errorMessage` (This technically isn't a breaking change since the batch of labeled field work hasn't been released yet, labeled field work is currently in the `feature/labeled-field` branch)

Stories/docs:
- Add docs on best practices for forms and fields
- Add docs around LabeledTextField deprecation in favour of LabeledField + TextField
- Update LabeledField examples and docs. Includes examples for validation and moving focus to the first field that has an error message after form submission

Issue: WB-1783

## Test plan:
1. Review new docs:
- LabeledField (`?path=/docs/packages-labeledfield--docs`)
- LabeledTextField (`/?path=/docs/packages-form-labeledtextfield-deprecated--docs`)
- Form Best Practices (`?path=/docs/packages-form-overview--docs`)

2. Test LabeledField with Field Validation: Confirm validation works as expected
- LabeledField Required story (`?path=/story/packages-labeledfield--required`)
  - Tabbing through fields without entering anything
  - Submitting without filling in fields
  - Inputting/selecting a value and removing it
- LabeledField Validation story (`?path=/story/packages-labeledfield--validation`)
  - Triggering validation errors
  - Clearing validation errors when selected value changes
  - Submitting the form to see an example for backend errors

Note: Did some preliminary testing with LabeledField and new validation features. Different screen reader and browser combinations had different behaviours around announcing errors so I created WB-1842 for more testing + documentation.


.changeset/smart-grapes-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-labeled-field": patch
```
beaesguerra, line 2: Note: Although renaming a prop is technically a breaking change, I marked this as a patch change since the LabeledField component isn't released yet (it's in the feature/labeled-field branch so the previous prop hasn't been available for consumers. Let me know if there are any concerns with this!

__docs__/wonder-blocks-form/accessibility.mdx
```diff
@@ -1,34 +0,0 @@
-import {Meta, Story, Canvas} from "@storybook/blocks";
```
beaesguerra, line 1: I combined this page with the Overview page so guidelines are together

packages/wonder-blocks-form/src/components/__tests__/text-area.test.tsx
```diff
@@ -1519,10 +1560,11 @@ describe("TextArea", () => {
                         />,
                         defaultOptions,
                     );
-
-                    // Act
                     const field = screen.getByRole("textbox");
                     await userEvent.type(field, "test");
+                    handleValidate.mockReset(); // Reset mock before leaving the field
```
beaesguerra, line 1565: Since we now clear the error state onChange when `instantValidation=false` (see `use-field-validation.ts` changes), `onValidate` is called each time a user types a character. We reset the mocks now before the action so we can more clearly check the `onValidate` calls triggered by the action

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -27,17 +22,38 @@ type Props = {
...
-     * The message for the error element.
+     * The message for the error element. If there is a message, it will also
+     * set the `error` prop on the `field` component.
      *
      * Note: Since the error icon has an aria-label, screen readers will
      * prefix the error message with "Error:" (or the value provided to the
      * errorIconAriaLabel in the `labels` prop)
      */
-    error?: React.ReactNode;
+    errorMessage?: React.ReactNode;
```
beaesguerra, line 56: Renaming to `errorMessage` so it is clear it's the message. (Form components like TextField, TextArea, etc have `error` boolean props so it could be confusing if this one continuted to be named `error` when it doesn't expect a boolean)

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -103,23 +120,27 @@ export default function LabeledField(props: Props) {
...
+    const generatedUniqueId = React.useId();
+    const uniqueId = id ?? `${generatedUniqueId}-labeled-field`;
     const labelId = `${uniqueId}-label`;
     const descriptionId = `${uniqueId}-description`;
     const fieldId = `${uniqueId}-field`;
     const errorId = `${uniqueId}-error`;
 
+    const isRequired = !!required || !!field.props.required;
+    const hasError = !!errorMessage || !!field.props.error;
+    const isLight = light || field.props.light;
```
beaesguerra, line 136: We also check the field's props to see if LabeledField is required, is in an error state, or is in light mode. 

This will make migration easier too so that we don't need to move `light` or `required` props from the field to the LabeledField component. The `errorMessage` prop would need to be set for LabeledField still since fields don't have this prop currently

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -27,17 +22,38 @@ type Props = {
...
+     *
+     * Boolean:
+     * True/false indicating whether this field is required. Please do not pass
+     * in `true` if possible - pass in the error string instead.
+     * If `true` is passed, and a `validate` prop is not passed, that means
+     * there is no corresponding message and the default untranlsated message
+     * will be used.
      */
-    required?: boolean;
+    required?: boolean | string;
```
beaesguerra, line 47: We update the `required` prop type to be the same as field components so it can be passed down

.changeset/smart-grapes-serve.md
jandrade: sounds good to me, thanks for the clarification!

__docs__/wonder-blocks-form/_overview_.mdx
```diff
@@ -1,10 +1,50 @@
...
 />
 
 # Form
 
-`wonder-blocks-form` provides building blocks form Form components, including
+`wonder-blocks-form` provides building blocks for Form components, including
 TextField, TextArea, Choice, Checkbox, RadioButton, etc.
+
+
+## Best Practices
```
jandrade, line 14: **praise:** Love this new consolidated section!

__docs__/wonder-blocks-form/_overview_.mdx
```diff
@@ -1,10 +1,50 @@
...
+
+### Error Validation
+
+- For fields like `TextField`, `TextArea`, and `SearchField` prefer setting
+`instantValidation=false`. This makes it so validation occurs on blur after a
+user is done interacting with a field.
+- Avoid disabling form submission buttons. There could be exceptions if the
+button is for one field.
+- If there are errors after a form is submitted, programatically move the user's
+focus to the first field with an error
```
jandrade, line 50: **thought:** I wonder if we should include static images at some point to illustrate how this could look like. No changes necessary, just food for thought.

__docs__/wonder-blocks-form/accessibility.stories.tsx
```diff
@@ -6,7 +6,7 @@ import {TextArea} from "@khanacademy/wonder-blocks-form";
 import {spacing} from "@khanacademy/wonder-blocks-tokens";
 
 export default {
-    title: "Packages / Form / Accessibility",
+    title: "", // Empty title so that these examples don't show in the sidebar. Examples are referenced in the .mdx file
```
jandrade, line 9: **praise:** Neat! I didn't know about this.

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
jandrade: **question:** Would this work in the case where we use custom fields? I'm not sure if this will fail of a custom field doesn't contain one of these props.

jandrade (approved):
This looks incredible! Just have a question about the logic for overriding the `light`, `required`, `error` props, but the rest looks really solid, great job 👏 🚀

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
beaesguerra: Good question! 

If LabeledField and a custom field doesn't have the required/error/light props set, then these values will be `false` or `undefined`! (no issues here)

If a custom field doesn't support these props, then they won't be applied to the component. From testing out LabeledField with a plain `<input>` element, I get warnings around props not being supported or are the wrong type 

image

Note: There is documentation around how LabeledField works best with field components that accept `error` and `required` props (I'll update this to include the light prop also) (`?path=/docs/packages-labeledfield--docs#fields`): 
image

One idea is we could see if we can refine the `field` type so that the component must have these props supported.

What do you think?

__docs__/wonder-blocks-form/_overview_.mdx
beaesguerra: I included a Canvas of the LabeledField validation story so that there is an example of validation and focusing on the first field with an error after submission! `?path=/docs/packages-form-overview--docs`

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
beaesguerra: I added a story that shows [LabeledField with a plain input element](https://5e1bf4b385e3fb0020b7073c-ybogfhqdoy.chromatic.com/?path=/story/packages-labeledfield--with-non-wb), let me know what you think!

__docs__/wonder-blocks-form/_overview_.mdx
beaesguerra: Also shared these docs with the design team to see if they have any feedback or would like to add any other guidelines! The content now is based on our cubby meeting discussion a while back

__docs__/wonder-blocks-form/_overview_.mdx
jandrade: Nice, thanks!

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
jandrade: > One idea is we could see if we can refine the field type so that the component must have these props supported.

I think it is fine not to enforce those props in `field` instances (at least not for now). I'm sure there will be cases where we still have reusable UI components in webapp that could be combined with `LabeledField` (e.g. `DatePicker`, `TimePicker`, some deprecated text field/area components). If we get to a point where we can get rid of those deprecated versions, then it would be nice to enforce those props.

Maybe these errors/warnings could go away using the nullish coalescing operator (??)

e.g.
```
const isRequired = !!required || !!field.props.required ?? false;
```

__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
```diff
@@ -308,3 +564,18 @@ export const Scenarios = (args: PropsFor<typeof LabeledField>) => {
         </View>
     );
 };
+
+/**
+ * Here is an example where LabeledField is used with a custom element.
```
jandrade, line 569: **suggestion:** What do you think about adding some NOTE here for folks to be aware that this is available but that we still suggest using WB fields as much as they can?

__docs__/wonder-blocks-form/accessibility.stories.tsx
beaesguerra: Disregard this! It wasn't working as expected, the story was still there in another section if I scrolled down 😅 I found if I named these stories the same as the Overview title, it isn't shown in the sidebar

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
beaesguerra: > Maybe these errors/warnings could go away using the nullish coalescing operator (??)

This doesn't address it unfortunately! The `required` prop isn't as much of an issue since it is a supported html attribute. It's more of the other props since those are specific to WB components! I was trying to see if there's a way for us to check at runtime if a component supports certain props before we set it. I couldn't find anything around this (let me know though if we have a way to do this!)

Some other options:
1. LabeledField only sets html attributes. LabeledField works best though when the field components can handle their own required/error/light states and their testId. 
2. Use a renderField prop instead so that the props are passed as an argument and consumers could decide how to apply the props to their custom element 
3. We support both the current `field` prop and a `renderCustomField` prop (see #3)
4. We leave things as is with the console warnings and custom components need to support the prop 

Let me know what you think or if you have other ideas!

jandrade: Ahhh got it, thanks for clarifying. I'd say 4 would be best in this case as we still offer support to WB fields (making a good case for trying to use these components).

jandrade (approved):
Looks great! thanks for clarifying the new props part. The current approach looks good to me 🚀

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -133,15 +154,15 @@ export default function LabeledField(props: Props) {
...
                     tag="label"
                     htmlFor={fieldId}
                     testId={testId && `${testId}-label`}
                     id={labelId}
                 >
                     {label}
-                    {required && requiredIcon}
+                    {isRequired && requiredIcon}
                 </LabelMedium>
                 <Strut size={spacing.xxxSmall_4} />
```
jandrade, line 167: **thought:** Looking at the stories, this got me thinking..... we should probably refactor this at some point to use `gap` instead of `Strut` (as we have been discussing previously). No need to make any changes here, but this would be a really nice improvement in the future (for all WB components).

beaesguerra: Good catch! I'll update this in a follow up PR :) Thanks!

PR #2404, opened 2024-12-19
WB-1808: Button - Use CSS pseudo-classes for styling states (hover, focus, etc)

jandrade:
## Summary:

Changing the way we style different states. We are going to rely on CSS
pseudo-classes for styling states (`:hover`, `:focus-visible`).

Note that we have to keep some `ClickableBehavior` states for programmatic focus
and preserve active/pressed overrides.

Also took the opportunity to modify the primary variant to use `outline +
outlineOffset` instead of `boxShadow`.

Issue: https://khanacademy.atlassian.net/browse/WB-1808

## Test plan:

Navigate to `/?path=/docs/packages-button--docs` and verify that the buttons are
styled correctly.

Screenshot 2024-12-19 at 5 28 22 PM


__docs__/wonder-blocks-button/button.stories.tsx
```diff
@@ -27,26 +26,6 @@ import ComponentInfo from "../../.storybook/components/component-info";
 import ButtonArgTypes from "./button.argtypes";
 import {ThemeSwitcherContext} from "@khanacademy/wonder-blocks-theming";
 
-/**
```
jandrade, line 30: **note:** Got rid of this b/c it's duplicated. The source of truth is the code in `button.tsx`.

__docs__/wonder-blocks-button/button.stories.tsx
```diff
@@ -203,6 +135,11 @@ Variants.parameters = {
             story: "There are three kinds of buttons: `primary` (default), `secondary`, and `tertiary`.",
         },
     },
+    chromatic: {
+        // We already have screenshots of other stories that cover more of
+        // the button states
+        disableSnapshot: true,
+    },
```
jandrade, line 142: **note:** I'm disabling a bunch of snapshots now that we can centralize visual regression tests in the `All variants` stories.

packages/wonder-blocks-button/src/__tests__/__snapshots__/custom-snapshot.test.tsx.snap
jandrade, line 1: **note:** Removed this to now use Chromatic snapshots instead of unit tests.

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -84,12 +84,12 @@ const ButtonCore: React.ForwardRefExoticComponent<
             kind !== "tertiary" &&
                 !disabled &&
                 (pressed
-                    ? buttonStyles.active
-                    : (hovered || focused) && buttonStyles.focus),
+                    ? buttonStyles.pressed
+                    : focused && buttonStyles.focused),
```
jandrade, line 88: **note:** Changed the style names to match what's returned from `clickable-behavior`.

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -362,47 +368,45 @@ export const _generateStyles = (
...
+                    activePressedStyling,
             },
-            focus: {
-                // This assumes a background of white for the regular button and
-                // a background of darkBlue for the light version. The inner
-                // box shadow/ring is also small enough for a slight variation
-                // in the background color not to matter too much.
-                boxShadow: `0 0 0 1px ${boxShadowInnerColor}, 0 0 0 3px ${
-                    light ? theme.color.bg.primary.default : color
-                }`,
```
jandrade, line 383: **note:** Took the opportunity to switch from `boxShadow` to `outline`. This could help solving this ticket: https://khanacademy.atlassian.net/browse/WB-1712

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -466,40 +478,36 @@ export const _generateStyles = (
...
-            },
-            active: {
-                color: light ? fadedColor : activeColor,
-                ":after": {
-                    height: theme.size.height.tertiaryHover,
-                    background: light ? fadedColor : activeColor,
+                [":hover:not([aria-disabled=true])" as any]: {
+                    textUnderlineOffset: theme.font.offset.default,
+                    textDecoration: "underline",
+                    textDecorationThickness: theme.size.underline.hover,
```
jandrade, line 502: **note:** Also simplified the `tertiary` hover styles to rely on underlines instead of using the `:after` hack.

__docs__/wonder-blocks-button/button.stories.tsx
```diff
@@ -539,6 +481,11 @@ Size.parameters = {
...
 };
 
 export const Spinner: StoryComponentType = () => (
```
beaesguerra, line 491: Would it be helpful to include the spinner state in All Variants as well?

packages/wonder-blocks-button/src/themes/default.ts
```diff
@@ -25,8 +30,6 @@ const theme = {
             primary: {
                 default: tokens.color.white,
                 disabled: tokens.color.offBlack32,
-                // used in boxShadow
-                inverse: tokens.color.darkBlue,
```
beaesguerra, line 29: **question:** When we remove things from a theme, is it considered a breaking change? It seems like it isn't since we don't export the theme from the package explicitly, though thought I'd double check! 

I was also wondering if we needed to remove this from other themes, but I see now that the khanmigo theme doesn't have these set since we merge the default theme with the khanmigo theme!

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -362,47 +368,45 @@ export const _generateStyles = (
...
     let newStyles: Record<string, CSSProperties> = {};
     if (kind === "primary") {
-        const boxShadowInnerColor: string = light
-            ? theme.color.bg.primary.inverse
-            : theme.color.bg.primary.default;
+        const focusStyling = {
+            outlineColor: light ? theme.color.bg.primary.default : color,
+            outlineOffset: 2,
+            outlineStyle: "solid",
+            outlineWidth: 2,
```
beaesguerra, line 375: (For my own learning): When do we include values in a theme vs setting the style directly?

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -416,6 +420,25 @@ export const _generateStyles = (
...
+        };
+
+        const activePressedStyling = {
+            background: light ? activeColor : secondaryActiveColor,
+            color: light ? fadedColor : activeColor,
+            borderColor: "transparent",
+            outlineColor: light ? fadedColor : activeColor,
+            outlineStyle: "solid",
+            outlineWidth: theme.border.width.focused,
+        };
```
beaesguerra, line 440: It'll be so nice when we can share state styling across different components 😄

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -429,34 +452,23 @@ export const _generateStyles = (
                 borderWidth: theme.border.width.secondary,
                 paddingLeft: padding,
                 paddingRight: padding,
+                // TODO(WB-1844): Change this when we get final designs for
+                // hover.
+                [":hover:not([aria-disabled=true])" as any]: focusStyling,
+                [":focus-visible:not([aria-disabled=true])" as any]:
+                    focusStyling,
+                [":active:not([aria-disabled=true])" as any]:
```
beaesguerra, line 460: Is there something we can do to prevent having `as any`? 

I was trying to see if I ran into similar things when working on the SelectOpener and it looks like we don't type cast it explicitly because we set `newStyles` in SelectOpener to `Record<String, any>` instead of `Record<String, CSSProperties>` 😅 

https://github.com/Khan/wonder-blocks/blob/c58e3fa49e7eba8c4235d6e3c805a5226e1b42d2/packages/wonder-blocks-dropdown/src/components/select-opener.tsx#L294-L304

https://github.com/Khan/wonder-blocks/blob/c58e3fa49e7eba8c4235d6e3c805a5226e1b42d2/packages/wonder-blocks-dropdown/src/components/select-opener.tsx#L258

packages/wonder-blocks-button/src/__tests__/__snapshots__/custom-snapshot.test.tsx.snap
beaesguerra, line 1: 🎉 Thanks for removing snapshot tests now that we have Chromatic coverage!

beaesguerra (commented):
Yay more sticker sheets! 🎉 It looks great, I left some comments/questions! I wanted to see what your thoughts were on removing things from the theme before approving

__docs__/wonder-blocks-button/button.stories.tsx
jandrade: I thought about that one, but I decided to keep it separately as it could introduce flakyness to that snapshot. I'm mentioning that because this variant includes animation and I've seen that Chromatic sometimes takes screenshots at different times.

That said, maybe we could add it and see how it goes? wdyt?

packages/wonder-blocks-button/src/themes/default.ts
jandrade: Good question, and you are right. Right now, as theming works, this is only available internally, so it wouldn't be a breaking change.

The other themes are built based on the shape of the `default` object, and that means that if we remove a key that is used in a merged theme (e.g. `khanmigo`), then TS would complain about it :). I created that merge theme architecture having that in mind, so it would help us to have a more robust theming approach.

packages/wonder-blocks-button/src/components/button-core.tsx
jandrade: Yess!!! 1000% to that 🙏

packages/wonder-blocks-button/src/components/button-core.tsx
jandrade: ohhhh thanks for being curious. This should be part of the theme as it could allow us in the future to configure these (e.g. in Polaris). I'm going to use a theme variable for this.

packages/wonder-blocks-button/src/components/button-core.tsx
jandrade: ugh yeah that's very annoying tbh. This is one of the things I'd hope Aphrodite would have better support for. We could add types, but it's hard to calculate all the possible combinations/variants for these selectors.

I guess this is a good/solid point for switching to CSS selectors (aka css-modules). Well, you know that I worry about opening the door to specificity, but that's something that we'll have to evaluate when we make the definitive decision to switch our styling approach.

jandrade: Maybe we could do something related to typing 🤔 
https://www.typescriptlang.org/play/?#code/C4TwDgpgBACgzhArgEwPYGEA2BDOC5QC8UAUFFAD5QBEAXABaoBuEATtWZTbQGaoDGiOB3JU6fQXAC0TAJZxZAI0wQRXOtn7BZLagG5OJUJCgAxRADsts1BeyZ4SNFlz4inMQ1xrP8n9wtUYH0SI3BoAEFWWWwI4GBoxURgCAJCDygAcgBtbGjsKWR5bGUIZABdTIycvJipWQsme1kKqtEs3PypVggAR0RZHtaDclDjSPjE5IgAZQgVLVRWIigAAwASAG9HFAwcPFSAXy3zK21bex3nffxDgAotqJi4hKVpuEOASlWDEn5bODAKBNTAtAASzDYcwWwCWtCgLymKWhEEWy2IdEYLFYtECwDunTqRTgJRUFU+IRIAHoqVAALbyBQWADmUB4lmsF0wUDACF2Un4NygdwgADpmaKoAxIaxPn8AUCGiCWqYBEIUWj4Yi3sj5qjYeiAkECbUCkrmuT9EA

packages/wonder-blocks-button/src/components/button-core.tsx
jandrade: @beaesguerra I've updated the styles to use theme now.

__docs__/wonder-blocks-button/button.stories.tsx
beaesguerra: Good point! I was thinking it could be helpful to see it across the different themes/states/variants etc, though I understand snapshots for animated things can be flaky! I'm okay leaving it as a separate story!

packages/wonder-blocks-button/src/themes/default.ts
beaesguerra: Nice! Good to know this about our theming!

packages/wonder-blocks-button/src/components/button-core.tsx
beaesguerra: Oooh cool! That TS example looks promising! Looks like the ordering would matter. For example, `:not([aria-disabled]):hover` wouldn't meet the type requirements like `:hover:not([aria-disabled])`

beaesguerra (approved):
Looks great! 🚀

marcysutton:
Tested in Chrome, Safari, and Firefox on Mac, and Chrome and Firefox on Windows. The interaction states work great! Nice work @jandrade!

PR #2405, opened 2024-12-19
RadioGroup: Add flexible width to legend element to fill available space

jandrade:
## Summary:

Adds `width: 100%` to `legend` element to allow expanding it to fill the
available space.

Note that this applies to both `RadioGroup` and `CheckboxGroup`.

Issue: XXX-XXXX

## Test plan:

Navigate to `/?path=/docs/packages-form-radiogroup--docs#custom%20label`.

Verify that the legend element is now full-bleed.


beaesguerra (approved):
Looks good to me, thanks for the quick fix!

PR #2407, opened 2024-12-19
DetailCell: Add variant stories

beaesguerra:
## Summary:
Add scenario and variant stories for DetailCell. This will help us confirm styling for different states with visual regression tests.

For now, the color contrast issues haven't been addressed yet. Coordinating with design on what styling updates we want long term (see WB-1797 comments for more details!). Including these stories for now so we can have visual regression tests ready when we address the color issues!

Issue: WB-1797

## Test plan:
- Review stories for:
  - All Variants (`?path=/docs/packages-cell-detailcell-all-variants--docs`)
  - Scenarios (`?path=/story/packages-cell-detailcell--scenarios`)

### DIFF
--- a/__docs__/wonder-blocks-cell/detail-cell-variants.stories.tsx
+++ b/__docs__/wonder-blocks-cell/detail-cell-variants.stories.tsx
@@ -1,0 +1,122 @@
+import * as React from "react";
+import {StyleSheet} from "aphrodite";
+import type {Meta, StoryObj} from "@storybook/react";
+
+import {PropsFor, View} from "@khanacademy/wonder-blocks-core";
+import {spacing} from "@khanacademy/wonder-blocks-tokens";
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-cell/src/components/detail-cell.tsx
```diff
@@ -94,7 +94,7 @@ const DetailCell = function (props: DetailCellProps): React.ReactElement {
 
 const styles = StyleSheet.create({
     subtitle: {
-        color: color.offBlack64,
+        color: "#6D6F74",
```
beaesguerra, line 97: Hardcoding for now based on the thread in WB-1797. It still has a11y color warnings so will get feedback!

marcysutton (approved):
I see the color contrast errors you're talking about...glad to know those will be addressed! This looks good to me otherwise!

PR #2408, opened 2024-12-20
Tooling: Move Storybook reusable components to __docs__ to optimize Turbosnap

jandrade:
## Summary:

While investigating the performance of Chromatic/Turbosnap, I found that
chromatic was disabling Turbosnap of the storybook stories in a lot of builds
because apparently we were making changing changes in files inside the
`.storybook` folder. This is not the case as those builds didn't include changes
in such files, but this made me realize that we could optimize the performance
of Turbosnap by moving the reusable components from the `.storybook` folder to
the `__docs__` folder.

Issue: XXX-XXXX

## Test plan:

Verify that Storybook still works as expected and check the Chromatic build to
see if Turbosnap is enabled for the stories that were previously disabled.

### DIFF
--- a/__docs__/components/component-info.tsx
+++ b/__docs__/components/component-info.tsx
@@ -1,31 +1,26 @@
 import * as React from "react";
 
+import githubLogo from "@phosphor-icons/core/fill/github-logo-fill.svg";
 import Button from "@khanacademy/wonder-blocks-button";
 import {View} from "@khanacademy/wonder-blocks-core";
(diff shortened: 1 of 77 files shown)


beaesguerra (approved):
Changes look good! As [discussed](https://khanacademy.slack.com/archives/C4PE1QM5Y/p1734711502678399), there are other TurboSnap considerations, these are still a helpful improvement! Nice find!

PR #2412, opened 2024-12-20
[fei6062.releaseprotections] Update prepublish checks to add SNAPSHOT_RELEASE support

somewhatabstract:
## Summary:
This brings over the latest updates from Perseus, and edits them for Wonder Blocks. The changes are:

1. New `prepublishOnly` script in each package that blocks publishing non-snapshot packages when publishing snapshots
2. Updates 'publish-snapshot.sh' to set the `SNAPSHOT_RELEASE` env var
3. Updates `pre-publish-check-ci.js` and `pre-publish-utils.js` files to TypeScript
4. Makes `pre-publish-check-ci.ts` executable
5. Adds a check to `pre-publish-check-ci.ts` to ensure that each public package has the `prepublishOnly` script
6. Updates `pre-publish-utils.ts` and `pre-publish-check-ci.ts` to perform all checks before erroring, instead of quitting on the first failure. This is a quality of life change to make it easier on devs having to update multiple packages to conform

Long term, might be nice to have a `wonder-stuff-cli` or something that can encapsulate this stuff across our packages.

This change adds initial protection against accidentally publishing release package versions under a snapshot tag. It protects both automated runs and runs done by a user locally.

The next PR will add workflow changes to block the snapshot publish jobs from running when there is a release workflow in progress. This adds an additional layer of protection and also provides a nicer dev experience by adding a useful PR comment on why it failed.

### Methodology:
I diffed the Perseus and Wonder Blocks scripts to work out what was different, then updated things to TS. I then modified the publish script and workflows, as necessary. Finally, I used Copilot Edits in VSCode to update the package.json files with the new `prepublishOnly` script.

Issue: FEI-6062

## Test plan:
This stuff is working in Perseus already. I've tested the new executable nature of the pre-publish checks, and verified the `SNAPSHOT_RELEASE` behavior for packages by running `SNAPSHOT_RELEASE=1 yarn run prepublishOnly` in a few packages.


somewhatabstract:
The changeset error is too sensitive. These changes don't need releasing.

utils/publish/pre-publish-check-ci.ts
somewhatabstract, line 1: **note:** This isn't a big a change as it's making out.

utils/publish/pre-publish-utils.ts
somewhatabstract, line 1: **note:** This isn't as big a change as it's making out.

utils/publish/pre-publish-check-ci.ts
```diff
@@ -0,0 +1,49 @@
...
+import {
+    checkPrivate,
+    checkEntrypoints,
+    checkSource,
+    checkTypes,
+    checkPublishConfig,
+} from "./pre-publish-utils";
+
+// Get the root folder; use the LICENSE file as our root anchor.
+const rootDir = ancesdir(__dirname, "LICENSE");
```
somewhatabstract, line 23: **note:** Other than making it so all checks are run before exiting (this is copied from Perseus and tested), this is the biggest change that isn't in Perseus - using ancesdir to locate the root directory rather than hard-coding it. This way it'll work if the file gets moved around.

jandrade (approved):
Looks good! 🚀

PR #2413, opened 2024-12-20
Update useOnMountEffect to pass isMountedRef to callback

kevinb-khan:
## Summary:
Without this, consumers of useOnMountEffect have to keep track of whether the hosting component is mounted or not when running async code within the effect.

Issue: None

## Test plan:
- let CI run automated checks


marcysutton (approved):
This is super cool. Looks good to me!

PR #2414, opened 2024-12-20
[fei6062.releaseprotections.2] Fail snapshot runs if a release is happening

somewhatabstract:
## Summary:
This brings over workflow changes from Perseus that are intended to prevent trying to publish snapshots if a release is in progress. This is the first layer of protection against this, we also have a script solution that works both in automation and locally for devs that will kick in for cases where this workflow change doesn't catch it.

This change looks for in-progress releases (`release.yml` runs for "Version Packages" changes) and fails the snapshot job if it finds one. This then adds a comment to the PR to indicate why the snapshot failed.

Issue: FEI-6062

## Test plan:
This was already tested for Perseus and testing requires orchestrating a release job at the same time as the PR trying to do snapshots. I don't think we need to do that for this PR, but we can if folks prefer.


jandrade (approved):
This looks really solid, thanks! :shipit:

PR #2420, opened 2025-01-07
FEI-5533: Re-enable select keyboard tests for Dropdown and Clickable

marcysutton:
In working on [WB-1799](https://khanacademy.atlassian.net/browse/WB-1799) for SingleSelect, I found a bunch of [disabled tests](https://khanacademy.atlassian.net/browse/FEI-5533) that I started fixing to ensure keyboard interactions work. Fixing these tests required changes to Clickable, which is used in WB Dropdowns.

Issue: https://khanacademy.atlassian.net/browse/FEI-6114

PR highlights:

1. Re-enables some keyboard tests that were disabled with `userEvent@14`
2. Exposes a centralized `keys` object in `wonder-blocks-core` for use in `wonder-blocks-clickable`, `wonder-blocks-dropdown`, and any other modules that need it.
    1. Note: userEvent passes the literal key code string back, i.e. `Space` or `space`. I lower-cased the key names to make sure they match in the code. We could use linting to enforce this instead... I'm open to suggestions.
3. Keeps two tests for SingleSelect disabled:
    1.  "It should find and select an item using the keyboard": fixing this one broke other tests (the dropdown isn't open before typing so refs aren't populated, but explicitly opening it made filtering tests fail).
    2. "It should change the number of options after using the search filter" (Live Region test): this one fails for me when I enable it. I'll plan to revisit it as part of the Announcer implementation.

## Test Plan

Run the tests using `yarn test`. 

There is one pre-existing lint failure for me locally in `expect-render-error.d.ts`, but it seems unrelated to these changes.

[WB-1799]: https://khanacademy.atlassian.net/browse/WB-1799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/mean-cherries-press.md
+++ b/.changeset/mean-cherries-press.md
@@ -1,7 +1,7 @@
 ---
-"@khanacademy/wonder-blocks-clickable": minor
-"@khanacademy/wonder-blocks-dropdown": minor
-"@khanacademy/wonder-blocks-core": minor
+"@khanacademy/wonder-blocks-clickable": major
+"@khanacademy/wonder-blocks-dropdown": major
+"@khanacademy/wonder-blocks-core": major
 ---
 
-Fixes keyboard tests in Dropdown and Clickable with specific key events. We now check event.key instead of event.which or event.keyCode to match the keys returned from Testing Library/userEvent.
+Fixes keyboard tests in Dropdown and Clickable with specific key events. We now check `event.key` instead of `event.which` or `event.keyCode` to remove deprecated event properties and match the keys returned from Testing Library/userEvent.
--- a/.changeset/mean-cherries-press.md
+++ b/.changeset/mean-cherries-press.md
@@ -1,7 +1,7 @@
 ---
-"@khanacademy/wonder-blocks-clickable": patch
-"@khanacademy/wonder-blocks-dropdown": patch
-"@khanacademy/wonder-blocks-core": patch
+"@khanacademy/wonder-blocks-clickable": minor
+"@khanacademy/wonder-blocks-dropdown": minor
+"@khanacademy/wonder-blocks-core": minor
 ---
 
-Fixes keyboard tests in Dropdown and Clickable
+Fixes keyboard tests in Dropdown and Clickable with specific key events. We now check event.key instead of event.which or event.keyCode to match the keys returned from Testing Library/userEvent.
--- a/.changeset/mean-cherries-press.md
+++ b/.changeset/mean-cherries-press.md
@@ -1,0 +1,7 @@
+---
+"@khanacademy/wonder-blocks-clickable": patch
+"@khanacademy/wonder-blocks-dropdown": patch
+"@khanacademy/wonder-blocks-core": patch
+---
+
+Fixes keyboard tests in Dropdown and Clickable
--- a/__docs__/wonder-blocks-dropdown/single-select.accessibility.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/single-select.accessibility.stories.tsx
@@ -57,3 +57,25 @@ export const UsingAriaAttributes = {
     render: SingleSelectAccessibility.bind({}),
     name: "Using aria attributes",
 };
+
+const SingleSelectKeyboardSelection = () => {
+    const [selectedValue, setSelectedValue] = React.useState("");
+    return (
+        <View>
+            <SingleSelect
+                placeholder="Choose"
+                onChange={setSelectedValue}
+                selectedValue={selectedValue}
+            >
+                <OptionItem label="apple" value="apple" />
+                <OptionItem label="orange" value="orange" />
+                <OptionItem label="pear" value="pear" />
+            </SingleSelect>
+        </View>
+    );
+};
+
+export const UsingKeyboardSelection = {
+    render: SingleSelectKeyboardSelection.bind({}),
+    name: "Using the keyboard",
+};
--- a/__docs__/wonder-blocks-dropdown/single-select.accessibility.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/single-select.accessibility.stories.tsx
@@ -58,6 +58,7 @@ export const UsingAriaAttributes = {
     name: "Using aria attributes",
 };
 
+// This story exists for debugging automated unit tests.
 const SingleSelectKeyboardSelection = () => {
     const [selectedValue, setSelectedValue] = React.useState("");
     return (
@@ -78,4 +79,7 @@ const SingleSelectKeyboardSelection = () => {
 export const UsingKeyboardSelection = {
     render: SingleSelectKeyboardSelection.bind({}),
     name: "Using the keyboard",
+    parameters: {
+        chromatic: {disableSnapshot: true},
+    },
 };
--- a/packages/wonder-blocks-clickable/src/components/clickable-behavior.ts
+++ b/packages/wonder-blocks-clickable/src/components/clickable-behavior.ts
@@ -556,13 +556,12 @@ export default class ClickableBehavior extends React.Component<
         if (onKeyDown) {
             onKeyDown(e);
         }
-        // Allow tests to use mixed case commands ("Space" or "space")
-        const keyCode = e.key.toLowerCase();
+        const keyName = e.key;
         const {triggerOnEnter, triggerOnSpace} =
             getAppropriateTriggersForRole(role);
         if (
-            (triggerOnEnter && keyCode === keys.enter.toLowerCase()) ||
-            (triggerOnSpace && keyCode === keys.space.toLowerCase())
+            (triggerOnEnter && keyName === keys.enter) ||
+            (triggerOnSpace && keyName === keys.space)
         ) {
             // This prevents space from scrolling down. It also prevents the
             // space and enter keys from triggering click events. We manually
@@ -570,7 +569,7 @@ export default class ClickableBehavior extends React.Component<
             // handleKeyUp instead.
             e.preventDefault();
             this.setState({pressed: true});
-        } else if (!triggerOnEnter && keyCode === keys.enter.toLowerCase()) {
+        } else if (!triggerOnEnter && keyName === keys.enter) {
             // If the component isn't supposed to trigger on enter, we have to
             // keep track of the enter keydown to negate the onClick callback
             this.enterClick = true;
@@ -583,17 +582,17 @@ export default class ClickableBehavior extends React.Component<
             onKeyUp(e);
         }
 
-        const keyCode = e.key.toLowerCase();
+        const keyName = e.key;
         const {triggerOnEnter, triggerOnSpace} =
             getAppropriateTriggersForRole(role);
         if (
-            (triggerOnEnter && keyCode === keys.enter.toLowerCase()) ||
-            (triggerOnSpace && keyCode === keys.space.toLowerCase())
+            (triggerOnEnter && keyName === keys.enter) ||
+            (triggerOnSpace && keyName === keys.space)
         ) {
             this.setState({pressed: false, focused: true});
 
             this.runCallbackAndMaybeNavigate(e);
-        } else if (!triggerOnEnter && keyCode === keys.enter) {
+        } else if (!triggerOnEnter && keyName === keys.enter) {
             this.enterClick = false;
         }
     };
--- a/packages/wonder-blocks-clickable/src/components/clickable-behavior.ts
+++ b/packages/wonder-blocks-clickable/src/components/clickable-behavior.ts
@@ -1,4 +1,5 @@
 import * as React from "react";
+import {keys} from "@khanacademy/wonder-blocks-core";
 
 // NOTE: Potentially add to this as more cases come up.
 export type ClickableRole =
@@ -227,11 +228,6 @@ const disabledHandlers = {
     onTouchCancel: () => void 0,
     onKeyDown: () => void 0,
     onKeyUp: () => void 0,
-} as const;
-
-const keyCodes = {
-    enter: 13,
-    space: 32,
 } as const;
 
 const startState: ClickableState = {
@@ -560,13 +556,13 @@ export default class ClickableBehavior extends React.Component<
         if (onKeyDown) {
             onKeyDown(e);
         }
-
-        const keyCode = e.which || e.keyCode;
+        // Allow tests to use mixed case commands ("Space" or "space")
+        const keyCode = e.key.toLowerCase();
         const {triggerOnEnter, triggerOnSpace} =
             getAppropriateTriggersForRole(role);
         if (
-            (triggerOnEnter && keyCode === keyCodes.enter) ||
-            (triggerOnSpace && keyCode === keyCodes.space)
+            (triggerOnEnter && keyCode === keys.enter.toLowerCase()) ||
+            (triggerOnSpace && keyCode === keys.space.toLowerCase())
         ) {
             // This prevents space from scrolling down. It also prevents the
             // space and enter keys from triggering click events. We manually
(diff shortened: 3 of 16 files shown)


packages/wonder-blocks-clickable/src/components/clickable-behavior.ts
```diff
@@ -560,21 +556,21 @@ export default class ClickableBehavior extends React.Component<
...
-        const keyCode = e.which || e.keyCode;
+        // Allow tests to use mixed case commands ("Space" or "space")
+        const keyCode = e.key.toLowerCase();
         const {triggerOnEnter, triggerOnSpace} =
             getAppropriateTriggersForRole(role);
         if (
-            (triggerOnEnter && keyCode === keyCodes.enter) ||
-            (triggerOnSpace && keyCode === keyCodes.space)
+            (triggerOnEnter && keyCode === keys.enter.toLowerCase()) ||
+            (triggerOnSpace && keyCode === keys.space.toLowerCase())
```
kevinb-khan, line 565: If you're normalizing the key name that we're getting from the event to be lowercase, why not have the constants in `keys` be lowercase to begin with?

packages/wonder-blocks-clickable/src/components/clickable-behavior.ts
```diff
@@ -560,21 +556,21 @@ export default class ClickableBehavior extends React.Component<
         if (onKeyDown) {
             onKeyDown(e);
         }
-
-        const keyCode = e.which || e.keyCode;
+        // Allow tests to use mixed case commands ("Space" or "space")
+        const keyCode = e.key.toLowerCase();
```
kevinb-khan, line 560: This variable should probably be called `keyName` or just `key` to avoid confusing with `e.keyCode` which is a number.

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -107,21 +107,27 @@ export default class SelectOpener extends React.Component<
...
 
     handleKeyUp: (e: React.KeyboardEvent) => void = (e) => {
-        const keyCode = e.key;
+        const keyCode = e.key.toLowerCase();
         // On key up for Enter and Space, trigger the click handler
-        if (keyCode === "Enter" || keyCode === " ") {
+        if (
+            keyCode === keys.enter.toLowerCase() ||
+            keyCode.toLowerCase() === keys.space.toLowerCase()
+        ) {
```
kevinb-khan, line 130: The variable `keyCode` should probably be renamed to `key` or `keyName` to avoid confusion with `event.keyCode` which is a number.

packages/wonder-blocks-dropdown/src/util/constants.ts
```diff
@@ -8,7 +8,7 @@ import {ComboboxLabels} from "./types";
 export const keys = {
     escape: "Escape",
     tab: "Tab",
-    space: " ",
+    space: "Space",
     up: "ArrowUp",
     down: "ArrowDown",
 } as const;
```
kevinb-khan, line 14: Can uses of this be replaced with the `keys` object exported by `@khanacademy/wonder-blocks-core`?

kevinb-khan (approved):
LGTM.  Thanks for figuring out why our keyboard tests were broken.  I left a few suggestions in the comments (nothing blocking).

packages/wonder-blocks-clickable/src/components/clickable-behavior.ts
```diff
@@ -560,21 +556,21 @@ export default class ClickableBehavior extends React.Component<
         if (onKeyDown) {
             onKeyDown(e);
         }
-
-        const keyCode = e.which || e.keyCode;
+        // Allow tests to use mixed case commands ("Space" or "space")
```
beaesguerra, line 559: It seems interesting to modify the component logic to accommodate the casing used in tests!

It looks like [user-event will also use KeyboardEvent.key](https://testing-library.com/docs/user-event/keyboard) (or KeyboardEvent.code). I wonder if it's enough to solve the issue in tests by using the `keys` constant as well so we're testing with the proper keys that the browser uses (which you've updated already!). Let me know what you think! I'm still learning more about user-event so let me know if I am missing something!

packages/wonder-blocks-dropdown/src/components/__tests__/single-select.test.tsx
```diff
@@ -304,36 +320,52 @@ describe("SingleSelect", () => {
...
+                const firstItem = await screen.findByRole("option", {
+                    name: /item 1/,
+                });
+                expect(firstItem).toHaveFocus();
+
                 // Act
                 await userEvent.keyboard("{enter}");
 
-                // Assert
+                // // Assert
```
beaesguerra, line 338: nit: extra `//`!
```suggestion
                // Assert
```

packages/wonder-blocks-dropdown/src/components/__tests__/single-select.test.tsx
```diff
@@ -304,36 +320,52 @@ describe("SingleSelect", () => {
...
                 // Arrange
                 const {userEvent} = doRender(uncontrolledSingleSelect);
                 await userEvent.tab();
                 await userEvent.keyboard("{enter}"); // open
 
+                // Ensure first option is focused, not the opener
+                const firstItem = await screen.findByRole("option", {
+                    name: /item 1/,
+                });
+                expect(firstItem).toHaveFocus();
```
beaesguerra, line 333: In general, it is suggested to have only one expectation in a unit test: https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/98402353/Unit+Testing+Best+Practices#One-Expectation-Per-Test-Case 

I am getting used this too 😅 Could we move this to another test that checks that the first option has focus?

beaesguerra (commented):
Thanks for diving into this, Marcy! Enabling the tests again can help build our confidence in our components, especially with these more complicated components 😄 I've left some comments and questions!

.changeset/mean-cherries-press.md
```diff
@@ -0,0 +1,7 @@
+---
+"@khanacademy/wonder-blocks-clickable": patch
+"@khanacademy/wonder-blocks-dropdown": patch
+"@khanacademy/wonder-blocks-core": patch
+---
+
+Fixes keyboard tests in Dropdown and Clickable
```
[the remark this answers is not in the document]
marcysutton: Really great idea. Thanks @beaesguerra!

packages/wonder-blocks-dropdown/src/util/constants.ts
marcysutton: Yes, good call!

__docs__/wonder-blocks-dropdown/single-select.accessibility.stories.tsx
```diff
@@ -57,3 +57,25 @@ export const UsingAriaAttributes = {
...
+            >
+                <OptionItem label="apple" value="apple" />
+                <OptionItem label="orange" value="orange" />
+                <OptionItem label="pear" value="pear" />
+            </SingleSelect>
+        </View>
+    );
+};
+
+export const UsingKeyboardSelection = {
```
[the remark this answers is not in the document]
marcysutton: Yes, good ideas. Primarily this is a mechanism for double-checking the fixture used in a lot of the keyboard tests. It was helpful for debugging, but isn't really needed for much else!

packages/wonder-blocks-dropdown/src/util/constants.ts
```diff
@@ -8,7 +8,7 @@ import {ComboboxLabels} from "./types";
 export const keys = {
     escape: "Escape",
     tab: "Tab",
-    space: " ",
+    space: "Space",
```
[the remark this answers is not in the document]
marcysutton: It was weird that `" "` wasn't matching for a space any longer. I only changed it because userEvent passed back the key name, in mixed casing! I considered doing an OR statement so that either one would match. One of them would have to be named something else, though...probably the literal space since the rest are represented by key names?

packages/wonder-blocks-dropdown/src/components/__tests__/single-select.test.tsx
marcysutton: Good point, that is something to get used to! My reasoning was that readability/understanding of the next assertion depended on this one. But I can definitely change it!

packages/wonder-blocks-dropdown/src/util/constants.ts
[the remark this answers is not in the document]
marcysutton: Here's the event object under test when passing `userEvent.keyboard('{space}')` as was written in these tests, with some irrelevant parts stripped out for brevity:

```    
SyntheticBaseEvent {
      _reactName: 'onKeyDown',
      type: 'keydown',
      target: <ref *1> HTMLButtonElement {
        '__reactFiber$dkck3rhjopo': FiberNode {
          tag: 5,
          key: null,
          elementType: 'button',
          type: 'button',
        },
        '__reactProps$dkck3rhjopo': {
          'aria-label': undefined,
          'aria-selected': 'false',
          'aria-checked': undefined,
          role: 'option',
          'aria-current': undefined,
          children: [Object],
          'data-testid': undefined,
          onClick: [Function: handleClick],
          onMouseEnter: [Function: handleMouseEnter],
          onMouseLeave: [Function: handleMouseLeave],
          onMouseDown: [Function: handleMouseDown],
          onMouseUp: [Function: handleMouseUp],
          onTouchStart: [Function: handleTouchStart],
          onTouchEnd: [Function: handleTouchEnd],
          onTouchCancel: [Function: handleTouchCancel],
          onKeyDown: [Function: handleKeyDown],
          onKeyUp: [Function: handleKeyUp],
          onFocus: [Function: handleFocus],
          onBlur: [Function: handleBlur],
          tabIndex: undefined,
          rel: undefined,
          type: 'button',
          'aria-disabled': false,
          className: '',
          style: [Object]
        },
      currentTarget: <ref *1> HTMLButtonElement {
        '__reactFiber$dkck3rhjopo': FiberNode {
          tag: 5,
          key: null,
          elementType: 'button',
          type: 'button',
        },
      eventPhase: 3,
      bubbles: true,
      cancelable: true,
      timeStamp: 1736555344807,
      defaultPrevented: false,
      isTrusted: false,
      view: null,
      detail: 0,
      key: 'space',
      code: 'Unknown',
      location: 0,
      ctrlKey: false,
      shiftKey: false,
      altKey: false,
      metaKey: false,
      repeat: false,
      locale: undefined,
      getModifierState: [Function: modifierStateGetter],
      charCode: 0,
      keyCode: 0,
      which: 0,
      isDefaultPrevented: [Function: functionThatReturnsFalse],
      isPropagationStopped: [Function: functionThatReturnsFalse]
    }
```

The `key: 'space'` part is what I'm matching against. I suppose the test would have to pass a literal space character to match on that? I was trying to get the existing tests to pass, and they had lowercase key names in them (`'{space}'`)!

packages/wonder-blocks-clickable/src/components/clickable-behavior.ts
marcysutton: Added a bunch of details in my previous comment!

packages/wonder-blocks-clickable/src/components/clickable-behavior.ts
marcysutton: Good question. It would have to be normalized somewhere in the event of mixed casing. The tests already had lowercase key names in them, but our constants are in capital case...so it makes sense why they would fail to match. I could change the constants to all be lowercase, but that will require a lottttt of testing for every key event (Enter, Escape, ArrowDown, ArrowUp, etc.). I'm actually thinking I could just update these tests to use capital case and go with that. It might limit the scope of this a bit better!

marcysutton: Update: I did make some updates including re-enabling more tests! But I limited it to `wonder-blocks-dropdown` and `wonder-blocks-clickable`. There are still more in other packages that could be updated later!

packages/wonder-blocks-dropdown/src/util/constants.ts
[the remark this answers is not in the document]
beaesguerra: Hmmm I wonder if we need to update how we simulate the space press key event in our tests: https://github.com/testing-library/user-event/issues/972#issuecomment-1154774419 (`{space}` or `{Space}` isn't shown as an example)

Looking up `{space}` in the React Testing Library docs shows documentation for user events v13: https://testing-library.com/docs/user-event/v13/#special-characters so it could be related to why the tests broke after upgrading to v14!

By updating the tests, it will be more similar to what happens in the browser! Testing things out in Storybook, I found that the space key functionality in the [Clickable example](https://5e1bf4b385e3fb0020b7073c-vlzxexlbss.chromatic.com/?path=/story/packages-clickable-clickablebehavior--wrapping-button) is not showing the pressed styles as [expected](https://khan.github.io/wonder-blocks/?path=/story/packages-clickable-clickablebehavior--wrapping-button) since the browser sets the `key` to ` ` instead of `Space`:


https://github.com/user-attachments/assets/53fe05ee-228b-44ad-b3b1-b2e566bc7e5c

.changeset/mean-cherries-press.md
```diff
@@ -0,0 +1,7 @@
+---
+"@khanacademy/wonder-blocks-clickable": minor
+"@khanacademy/wonder-blocks-dropdown": minor
+"@khanacademy/wonder-blocks-core": minor
```
beaesguerra, line 4: Related to our [thread about versioning in the other PR](https://github.com/Khan/wonder-blocks/pull/2345#discussion_r1909588990), would this also be a breaking change if consumer projects were relying on the `event.which/event.keyCode` functionality?

marcysutton:
I did a bit of an audit as part of creating subtasks and found a few more tests that could be enabled! There are 5 outstanding tests in WB Dropdown that I've indicated in the subtask issue: https://khanacademy.atlassian.net/browse/FEI-6114

.changeset/mean-cherries-press.md
marcysutton: Great question. I think yes, and this could be released along with the other breaking changes. So it's good timing!

marcysutton: Hmm, now I'm rethinking `major` on this and wondering if `minor` is more appropriate. The move to `event.key` is internal, and not part of the public API. Curious to hear your thoughts @beaesguerra!

packages/wonder-blocks-dropdown/src/util/constants.ts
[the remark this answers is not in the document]
marcysutton: Great catch, Bea! This prompted me to check the tests again and we can fire the literal string space instead! I made some updates to use `" "` instead of `"{Space}"` and ClickableBehavior should work again.

.changeset/mean-cherries-press.md
beaesguerra: Looks like I marked the [similar event.key change](https://github.com/Khan/wonder-blocks/pull/2376/files#diff-d3528e7b0e4be61064f2f91918d8d6f3e76c3585340c38ad444a3b8b43014a83R1-R5) for the SelectOpener as a `patch` change 😅 I think it is fine that we don't mark it as `major` since it is internal. It would be good to double check these areas in current usage though just to be sure there wasn't functionality relying on that! 

And since this would be released with the other combobox changes, it'll be part of the new major version as you mentioned!

beaesguerra (approved):
Changes look good, thanks for fixing this issue and re-enabling a bunch of tests! 🎉

Will leave it up to you on the versioning since this will be combined with the other combobox changes!

marcysutton:
Thanks @beaesguerra, I think I'll merge it so we can get this shipped! :shipit:

PR #2422, opened 2025-01-10
WB-1847: Dropdown - Update SelectOpener to match design specs

jandrade:
## Summary:

The dropdown opener currently uses a darkBlue/transparent background and that
could cause accessibility issues (color contrast). These styles are not matching
the current Figma specs, which use a white background.

Also, took the opportunity to switch from `color` to `semanticColor` in most of
the cases of that component.

### Specific style changes:

- Light version: Changed from `darkBlue` to a `white` background, and also changed the `disabled` state to use disabled semantic colors instead of blue shades.
- Switched from `border` to `outline` when focused/active for better consistency across the library.
- Modified focus/active outlines to be inset (located inside the box).
- Fixed the pressed/placeholder state to use an `activeBlue` text color.

### New `semanticColor` tokens

Additionally, I've added a couple of new tokens in `semanticColor` to include
some missing colors that are defined in Figma.

https://www.figma.com/design/VbVu3h2BpBhH80niq101MHHE/%F0%9F%92%A0-Main-Components?node-id=17450-482&t=jCX4iO74AL9fsTwh-4

Screenshot 2025-01-10 at 12 44 33 PM

Issue: https://khanacademy.atlassian.net/browse/WB-1847

## Test plan:

Navigate to `/?path=/story/packages-dropdown-singleselect--light` and verify that
the SelectOpener component now has a white background.

Also verify all the variants in `/?path=/docs/packages-dropdown-singleselect-all-variants--docs`

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/select-opener.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/select-opener.tsx
@@ -150,8 +150,8 @@ export default class SelectOpener extends React.Component<
         const iconColor = light
             ? "currentColor"
             : disabled
-            ? tokens.color.offBlack32
-            : tokens.color.offBlack64;
+            ? semanticColor.action.disabled.default
+            : semanticColor.icon.primary;
 
         const style = [
             styles.shared,
@@ -229,15 +229,9 @@ const styles = StyleSheet.create({
     },
 
     caret: {
-        minWidth: 16,
+        minWidth: tokens.spacing.medium_16,
     },
 });
-
-// These values are default padding (16 and 12) minus 1, because
-// changing the borderWidth to 2 messes up the button width
-// and causes it to move a couple pixels. This fixes that.
-const adjustedPaddingLeft = tokens.spacing.medium_16 - 1;
-const adjustedPaddingRight = tokens.spacing.small_12 - 1;
 
 const stateStyles: Record<string, any> = {};
 
@@ -253,9 +247,9 @@ const _generateStyles = (
     }
 
     let newStyles: Record<string, any> = {};
-    if (light) {
-        const borderWidth = tokens.border.width.hairline;
+    const borderWidth = tokens.border.width.hairline;
 
+    if (light) {
         const focusHoverStyling = {
             borderColor: semanticColor.border.inverse,
             outlineColor: error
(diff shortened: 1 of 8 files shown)


packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -146,15 +147,11 @@ export default class SelectOpener extends React.Component<
 
         const stateStyles = _generateStyles(light, isPlaceholder, error);
 
-        // The icon colors are kind of fickle. This is just logic
-        // based on the zeplin design.
         const iconColor = light
-            ? disabled || error
-                ? "currentColor"
-                : tokens.color.white
```
jandrade, line 154: **note:** this was no longer needed, so I got rid of this (now `light` uses a similar style compared to the default version).

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -232,16 +229,10 @@ const styles = StyleSheet.create({
...
-        minWidth: 16,
+        minWidth: tokens.spacing.medium_16,
     },
 });
 
-// These values are default padding (16 and 12) minus 1, because
-// changing the borderWidth to 2 messes up the button width
-// and causes it to move a couple pixels. This fixes that.
-const adjustedPaddingLeft = tokens.spacing.medium_16 - 1;
-const adjustedPaddingRight = tokens.spacing.small_12 - 1;
```
jandrade, line 243: **note:** This hack is no longer needed now that we use `outline` instead of `border`.

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -353,12 +385,15 @@ const _generateStyles = (
...
             disabled: {
-                background: tokens.color.offWhite,
-                borderColor: tokens.color.offBlack16,
-                color: tokens.color.offBlack64,
+                background: semanticColor.action.disabled.secondary,
+                borderColor: semanticColor.border.primary,
+                color: semanticColor.text.secondary,
                 cursor: "not-allowed",
                 ":focus-visible": {
-                    boxShadow: `0 0 0 1px ${tokens.color.white}, 0 0 0 3px ${tokens.color.offBlack32}`,
```
jandrade, line 361: **note:** Also changed this `box-shadow` to `outline` for consistency.

packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
```diff
@@ -9,10 +9,12 @@ export const semanticColor = {
         primary: {
             default: color.blue,
             active: color.activeBlue,
+            pressing: color.fadedBlue,
```
beaesguerra, line 12: **question (non-blocking):** In our tokens meeting last week, we were chatting about `active` vs `pressed` naming and decided to go with the `pressed` (or is it `press`) naming since it doesn't have to match the CSS names. I see we have `active` and `pressing` token names from the original semantic colors in Figma. Do we want to start moving towards that, or is that a later thing we want to handle?

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -256,45 +247,60 @@ const _generateStyles = (
...
-            paddingLeft: adjustedPaddingLeft,
-            paddingRight: adjustedPaddingRight,
+            borderColor: semanticColor.border.inverse,
+            outlineColor: error
+                ? semanticColor.status.critical.foreground
+                : semanticColor.action.primary.default,
+            // Provides an outline around the button to match the border width
+            outlineOffset: -(tokens.border.width.thin + borderWidth),
+            outlineStyle: "solid",
+            outlineWidth: tokens.border.width.thin,
```
beaesguerra, line 261: The [outline in Figma](https://www.figma.com/design/VbVu3h2BpBhH80niq101MHHE/%F0%9F%92%A0-Main-Components?node-id=13778-9969&t=Wu9bIjKLCX40Kj3V-4) visually is 3px wide instead of 2! In Figma, it's shown with a 2px border and a shadow with 1px spread (I think it's fine we've simplified it to using only an outline!). We don't have a border width token for 3px though. Would this be a one-off, a new token, or an update to the designs so it uses 2px instead? Let me know what you think!

image

The [error outline when light=true](https://www.figma.com/design/VbVu3h2BpBhH80niq101MHHE/%F0%9F%92%A0-Main-Components?node-id=13992-9501&t=Wu9bIjKLCX40Kj3V-4) is 2px though!
image

beaesguerra (commented):
Thanks for updating our components to match the designs more and use more semantic tokens 😄 Left some comments and questions in the PR and in Chromatic!

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
jandrade: hmmm yeah, tbh that's a bit confusing. In Figma, they are using border: inset, 2px for the normal outline, but I see it translated as `3px`. I'll bring this topic today to our WB sync meeting.

packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
jandrade: Thanks for bringing this up! As we discussed in the WB cubby meeting, I'm going to rename this to `press.foreground` and `active` will change to `press.background`.

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
jandrade: Also, based on the last WB cubby discussion, I'll use the same focus that we use with TextField/TextArea for better consistency.

jandrade:
@beaesguerra I'm closing this PR as I'm doing bigger changes, namely:

- #2428
- #2430 
- TBD that will include any missing styles in `SelectOpener`

PR #2424, opened 2025-01-10
Clear field error state if the field is cleared

beaesguerra:
## Summary:
Previously, if a MultiSelect component had shortcuts enabled, if it was in an error state, the "select all" and "select none" options do not clear the error state. Changing the selected values using these shortcuts now clear the error.

Issue: WB-1846

#### Select none
1. Navigate to `?path=/story/packages-dropdown-multiselect--error-from-validation` and confirm that the shortcuts control is enabled
2. Press the "Select none" option on the field that has an error already
3. Verify that the error message is cleared, even after blurring the field (since Jupiter is not selected)

#### Select all
1. Navigate to `?path=/story/packages-dropdown-multiselect--error-from-validation` and confirm that the shortcuts control is enabled
2. Press the "Select all" option on the field that has an error already
3. Verify that the error message is cleared. After blurring the field, the error message should return since Jupiter is selected


packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -357,10 +357,12 @@ const MultiSelect = (props: Props) => {
             .filter((option) => !!option && !option.props.disabled)
             .map((option) => option.props.value);
         onChange(selected);
+        onSelectedValuesChangeValidation();
```
beaesguerra, line 360: Since the values change when the `Select none` / `Select all` options are picked, we call the `onSelectedValuesChangeValidation` handler

packages/wonder-blocks-form/src/components/__tests__/labeled-text-field.test.tsx
```diff
@@ -729,10 +729,10 @@ describe("Required LabeledTextField", () => {
             "test-labeled-text-field-field",
         );
         textField.focus();
-        await userEvent.clear(textField);
```
beaesguerra, line 732: Since we now clear the error state when the field is cleared, this test case can be updated with an example where we update the input value instead of clear it

packages/wonder-blocks-form/src/hooks/use-field-validation.ts
```diff
@@ -141,6 +142,17 @@ export const useFieldValidation = ({
         }
     });
 
+    React.useEffect(() => {
+        // If it is not required, clear the error when the value is now empty
+        if (!required && prevValue.current !== value && value.length === 0) {
```
jandrade, line 147: **question:** I wonder if you need to use `trim()` here to cover the case where the user only types spaces (e.g. `"   "`)

jandrade (approved):
This looks great! thanks for including all these tests to cover that 🚢

packages/wonder-blocks-form/src/hooks/use-field-validation.ts
beaesguerra: Great question! We don't need to use `trim` here to cover the `" "` case since the error should only be cleared if the field is completely empty so any validation errors are still shown!

beaesguerra:
I'm testing the npm snapshot in `webapp` and found that there are some [tests that are failing](https://github.com/Khan/webapp/actions/runs/12798241629/job/35681847163?pr=28690) because they rely on the error message to be shown still after the TextField is cleared. I'm going to remove the updates to `use-field-validation` so we can preserve the existing functionality. It was originally added to address the case where pressing the "clear button" on the SearchField didn't immediately clear the error state. This isn't necessary though since once the user updates the field, the error state will get cleared anyways (unless the new value is invalid).

beaesguerra:
I will release and deploy these changes with the LabeledField feature branch!

jandrade (approved):
Thanks for including the snapshot release results. These recent changes make sense! :shipit:

PR #2425, opened 2025-01-11
LabeledField: Make sure custom required message is shown

beaesguerra:
## Summary:
Makes sure custom required messages passed into `LabeledField` or the `field` prop are displayed

Issue: WB-1848

## Test plan:
Confirm in `?path=/story/packages-labeledfield--required` that the error message shows the custom required message

### DIFF
--- a/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
+++ b/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
@@ -250,7 +250,7 @@ export default function LabeledField(props: Props) {
             ]
                 .filter(Boolean)
                 .join(" "),
-            required: isRequired,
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -250,7 +250,7 @@ export default function LabeledField(props: Props) {
             ]
                 .filter(Boolean)
                 .join(" "),
-            required: isRequired,
```
beaesguerra, line 253: Since we were passing in a boolean, the default required error message was always used 😅

jandrade (approved):
LGTM! nice fix 🚀

PR #2426, opened 2025-01-14
LabeledField: Add a story for custom jsx for props

beaesguerra:
## Summary:
The LabeledField accepts a ReactNode for the `label`, `description`, and `errorMessage` props already. This adds a story and documentation for this functionality!

Issue: WB-1785

## Test plan:
Review Custom story docs in `?path=/docs/packages-labeledfield--docs#custom`

### DIFF
--- a/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
+++ b/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
@@ -598,3 +598,30 @@ export const WithNonWb = {
         },
     },
 };
+
+/**
+ * Custom ReactNode elements can be used for the `label`, `description`, and
+ * `error` props. Ideally, the styling of LabeledField should not be overridden.
+ * If there is a specific use case where the styling needs to be overridden,
+ * please reach out to the Wonder Blocks team!
+ */
+export const Custom = {
+    args: {
+        label: (
+            <span>
+                <b>Label</b> <i>using</i> <u>JSX</u>
+            </span>
+        ),
+        description: (
+            <span>
+                <b>Description</b> <i>using</i> <u>JSX</u>
+            </span>
+        ),
+        field: <TextField value="" onChange={() => {}} />,
+        errorMessage: (
+            <span>
+                <b>Error</b> <i>using</i> <u>JSX</u>
+            </span>
+        ),
+    },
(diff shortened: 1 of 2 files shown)


__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
```diff
@@ -598,3 +598,30 @@ export const WithNonWb = {
         },
     },
 };
+
+/**
+ * Custom ReactNode elements can be used for the `label`, `description`, and
+ * `error` props. Ideally, the styling of LabeledField should not be overridden.
+ * If there is a specific use case where the styling needs to be overridden,
+ * please reach out to the Wonder Blocks team!
```
jandrade, line 606: **praise:** This is great! thanks for calling out reaching out to us!

**suggestion:** How do you feel about including a message mentioning that this is useful for things like toggletips? or including icons that represent meaning? This is kinda tricky as I'm a firmly believer that we should use `description` (or inlining content) instead of relying on Tooltips. wdyt?

beaesguerra: I'm thinking we can leave it open-ended for now and see how often teams reach out about toggletips with fields. If it is a common pattern, I think we should create a separate prop for supporting the toggle tip so that it can be part of the API so it is consistent (ex: the same icon and styling/spacing is used for the hint). 

Another thing to note is the contents of the `label` prop are wrapped with a `<label>` tag. I'm not sure if the toggle tip should be inside the `label` tag or outside, since clicking on the label will also focus on the associated field. And screen readers will read out the associated label when the field is interacted with, so the aria-label for the toggletip trigger will be included as part of that. 

Inlining content in the description would simplify a lot of these things!


Another possible use case for the custom elements if adding the `(optional)` tag to the label. I think it isn't necessary to support this since we have the required indicator and we should stick with one of these approaches!

As always, let me know what you think!

jandrade: I like that!

jandrade (approved):
Yay!! this is an awesome wrap up. Amazing job for creating all of this for LabeledField 👏 💯 🚢 :shipit: 🚀

PR #2428, opened 2025-01-15
[WB-1849] Rework action semantic color tokens

jandrade:
## Summary:

Reworked `semanticColor` actions to use a new structure. Every `action` now
includes `default`, `hover` and `press` states, and each state includes
`border`, `background` and `foreground` tokens.

By doing this change, we'll offer a more consistent and predictable styling
direction for our interactive components.

Issue: https://khanacademy.atlassian.net/browse/WB-1849

## Test plan:

Navigate to `/?path=/docs/foundations-using-color--docs` and check that the
`Action` section includes all the new tokens and the structure is readable and
makes sense.

### DIFF
--- a/packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
+++ b/packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
@@ -1,18 +1,87 @@
 import {color} from "./color";
 
+const border = {
+    primary: color.fadedOffBlack16,
+    subtle: color.fadedOffBlack8,
+    strong: color.fadedOffBlack50,
+    inverse: color.white,
+};
+
 export const semanticColor = {
     /**
      * For buttons, links, and controls to communicate the presence and meaning
      * of interaction.
      */
     action: {
-        primary: {
-            default: color.blue,
-            active: color.activeBlue,
+        // For primary actions
+        progressive: {
+            default: {
+                border: "transparent",
+                background: color.blue,
+                foreground: color.white,
+            },
+            hover: {
+                border: color.blue,
+                background: color.blue,
+                foreground: color.white,
+            },
+            press: {
+                border: color.activeBlue,
+                background: color.activeBlue,
+                foreground: color.white,
+            },
+        },
+        // Inverse is meant for use on secondary controls.
+        progressiveInverse: {
+            default: {
+                border: border.strong,
+                background: color.white,
+                foreground: color.blue,
+            },
+            hover: {
+                border: color.blue,
+                background: color.white,
+                foreground: color.blue,
+            },
+            press: {
+                border: color.activeBlue,
+                background: color.fadedBlue,
+                foreground: color.activeBlue,
+            },
         },
         destructive: {
-            default: color.red,
-            active: color.activeRed,
+            default: {
+                border: "transparent",
+                background: color.red,
+                foreground: color.white,
+            },
+            hover: {
+                border: color.red,
+                background: color.white,
+                foreground: color.red,
+            },
+            press: {
+                border: color.activeRed,
+                background: color.activeRed,
+                foreground: color.white,
+            },
+        },
+        destructiveInverse: {
+            default: {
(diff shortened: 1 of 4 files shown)


packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
```diff
@@ -1,18 +1,87 @@
...
+            hover: {
+                border: color.blue,
+                background: color.blue,
+                foreground: color.white,
+            },
+            press: {
+                border: color.activeBlue,
+                background: color.activeBlue,
+                foreground: color.white,
+            },
```
beaesguerra, line 32: Should the `focus` state also have tokens associated with it?

packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
```diff
@@ -1,18 +1,87 @@
...
+                background: color.white,
+                foreground: color.red,
+            },
+            press: {
+                border: color.activeRed,
+                background: color.fadedRed,
+                foreground: color.activeRed,
+            },
         },
         disabled: {
```
beaesguerra, line 86: Should the tokens associated with the `disabled` state also have the backgorund, foreground, border structure?

packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
```diff
@@ -1,18 +1,87 @@
...
     /**
      * For buttons, links, and controls to communicate the presence and meaning
      * of interaction.
      */
     action: {
-        primary: {
-            default: color.blue,
-            active: color.activeBlue,
+        // For primary actions
+        progressive: {
```
beaesguerra, line 17: For my own understanding, what was the reason for using the `progressive`/`progressiveInverse` terminology instead of `primary`/`secondary`? 

And should we start using this terminology for component `kind` variants too?

beaesguerra (approved):
Looks good! Had some non-blocking questions I wanted to get your thoughts on! 

I noticed in the docs some of the token names wrap (you mentioned this also!). I confirmed that copy + pasting it still copies the name without the line break though!
image

packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
jandrade: `focus` will have its own separate state, that is directly allocated in the `semanticColor.border.focus` token. This is something Caitlyn and I have discussed and the goal is that any focus on any component should use a blue outline, so that's the reason why I'm not including it here.

packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
jandrade: It might be in the future, but for now I'm leaving them out of every specific semantic global token and I'm defining them inside `Button` for now. Buttons need simplification and before doing that, I prefer to keep it this way. FWIW Design also has this structure, but this could change in the next weeks.

packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
beaesguerra: Sounds good! There is also the disabled focus outline!

jandrade:
@beaesguerra I made some changes based on your feedback/questions (also after discussing offline).

- Changed the structure to use `filled` and `outlined` to represent `primary/semantic` values (this might change in the future depending on the visuals approved for polaris).
- Changed the structure of the "Using color" page to include an example of how all these new tokens work in their context + modified the layout to make it more readable.

https://5e1bf4b385e3fb0020b7073c-xynstbxfft.chromatic.com/?path=/docs/foundations-using-color--docs

Screenshot 2025-01-21 at 5 17 51 PM

Let me know what you think, and thanks again for the suggestions!

beaesguerra (approved):
Looks great! Thanks for making those updates! 🎉

PR #2431, opened 2025-01-16
[WB-1853] Button: Refactor theme to support semanticColor tokens

jandrade:
## Summary:

Now that we have `semanticColor` tokens available, this PR is responsible for
refactoring the existing `Button` code to use these tokens instead of the
`color` primitive tokens. Also, it modifies the theme structure to follow
closer the semantic color tokens structure.

Also added a `light.destructive.disabled` variant to the Dark story as we didn't
include a snapshot for it.

Issue: WB-1853

## Test plan:

Navigate to `/?path=/docs/packages-button--docs` and verify that the button
hasn't changed.


packages/wonder-blocks-button/src/components/__tests__/button-with-icon.test.tsx
```diff
@@ -228,50 +228,4 @@ describe("button with icon", () => {
         expect(icon).toBeInTheDocument();
         expect(icon).toHaveAttribute("aria-hidden", "true");
     });
-
```
jandrade, line 231: **note:** These tests no longer apply as tertiary buttons don't need any changes for the Khanmigo theme.

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -182,7 +182,7 @@ const ButtonCore: React.ForwardRefExoticComponent<
                                 sharedStyles.endIconWrapperTertiary,
                             (focused || hovered) &&
                                 kind !== "primary" &&
-                                sharedStyles.iconWrapperSecondaryHovered,
+                                buttonStyles.iconWrapperHovered,
```
jandrade, line 185: I've moved this to the function where we generate style colors dynamically to be able to support `default` (`progressive`) and `destructive` colors.

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -344,48 +340,48 @@ export const _generateStyles = (
...
     let newStyles: Record<string, CSSProperties> = {};
     if (kind === "primary") {
+        const colorToAction = light
+            ? buttonColor === "destructive"
+                ? "destructiveLight"
+                : "progressiveLight"
+            : buttonColor === "destructive"
+            ? "destructive"
+            : "progressive";
+        const themeColorAction = theme.color[colorToAction];
```
jandrade, line 361: **note:** Now that the theme tries to cover all the different aspects of the button states (hover/focus, press, disabled, light), I'm simplifying the logic a bit here to avoid using ternary operators when defining the CSS properties below.

e.g.
Instead of: `bg: light ? theme.color.light.background : somethingActionEtc.background`

We now use `bg: themeColorAction.default.background`.

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -344,48 +340,48 @@ export const _generateStyles = (
...
 
         newStyles = {
             default: {
-                background: light ? theme.color.bg.primary.default : color,
-                color: light ? color : theme.color.text.inverse,
-                paddingLeft: padding,
-                paddingRight: padding,
+                background: themeColorAction.default.background,
+                color: themeColorAction.default.foreground,
+                paddingInline: padding,
```
jandrade, line 384: **note:** Switching to `Inline` here for better RTL support (it shouldn't affect the current styles at all).

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -397,61 +393,50 @@ export const _generateStyles = (
...
-                    ? theme.color.border.secondary.inverse
-                    : secondaryBorderColor,
+                background: themeColorAction.default.background,
+                color: themeColorAction.default.foreground,
+                borderColor: themeColorAction.default.border,
                 borderStyle: "solid",
                 borderWidth: theme.border.width.secondary,
-                paddingLeft: padding,
-                paddingRight: padding,
+                paddingInline: padding,
```
jandrade, line 439: **note:** These are the main changes of the secondary button (resting state). I don't understand why Chromatic is detecting so many changes with the secondary variant.

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -510,13 +506,18 @@ export const _generateStyles = (
...
             disabled: {
-                color: light ? fadedColor : theme.color.text.disabled,
+                // NOTE: This is an special case to handle the light color
+                color: themeColorAction.disabled.border,
                 cursor: "default",
+                ":focus-visible": {
+                    outlineColor: themeColorAction.disabled.foreground,
+                    outlineStyle: "solid",
+                    outlineWidth: theme.border.width.disabled,
+                },
```
jandrade, line 516: **note:** Adding CSS pseudo-state here.

packages/wonder-blocks-button/src/themes/default.ts
jandrade, line 1: **note:** I've changed the `color` tokens structure here to follow a similar approach to the `semanticColor` actions. Also added a `disabled` category to every variant b/c `Button` has a very particular disabled states (I hope we could really simplify this in the future).

jandrade, line 1: **note:** I've changed the `color` tokens structure here to follow a similar approach to the `semanticColor` actions. Also added a `disabled` category to every variant b/c `Button` has particular disabled states (I hope we could really simplify this in the future).

packages/wonder-blocks-button/src/themes/default.ts
```diff
@@ -1,101 +1,289 @@
...
-                inverse: tokens.color.fadedRed,
+        /**
+         * Primary
+         */
+        filled: {
+            // kind=primary / color=default / light=false
+            progressive: {
+                // filled
+                ...semanticColor.action.filled.progressive,
+                disabled: {
```
beaesguerra, line 20: Would it be helpful to have the `disabled` object as part of the semantic color tokens? (I might have asked this when we were pairing but I can't remember 😅)

beaesguerra (commented):
I left some questions around the semantic token usage! I'm still wrapping my head around it + themes so I'll reach out as well! 

I also left a comment in Chromatic around border color changes in light mode, wanted to confirm it was intended!

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -510,13 +490,18 @@ export const _generateStyles = (
             focused: focusStyling,
             pressed: activePressedStyling,
             disabled: {
-                color: light ? fadedColor : theme.color.text.disabled,
+                // NOTE: This is an special case to handle the light color
+                color: themeColorAction.disabled.border,
```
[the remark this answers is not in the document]
jandrade: Nice catch! It was a mistake from my side. I'm now using the correct token.

beaesguerra (approved):
Looks good! 🎉

PR #2435, opened 2025-01-20
[wb-1797] Address cell color contrast a11y issues

beaesguerra:
## Summary:
Updating state styles based on the new designs to address color contrast issues: https://www.figma.com/design/qSKXmIvmsNPKFlQNP5cztQ?node-id=4337-8746#1091493830

Styles to highlight:
- If a cell is selected (active: true), there is no hover and pressed state since clicking on the cell will bring the user to the same page. It continues to be focusable
- Press state and selected states now have left visual indicators of different widths so that color is not the only indicator
- The disabled focus outline is now using the disabled color token
- See changeset for more specific details on the style changes

A new primitive color token is also added: `fadedOffBlack72`. This is the new value for `semanticColor.text.secondary`. This is a darker gray that has better color contrast with a `fadedBlue8` background

Issue: WB-1797

## Test plan:
- Confirm DetailCell and CompactCell have no color contrast issues and the styling of the states are expected
  - DetailCell `?path=/docs/packages-cell-detailcell-all-variants--docs`
  - CompactCell `?path=/docs/packages-cell-compactcell-all-variants--docs`
- Testing the snapshot with webapp: https://prod-znd-250120-28792-a2bc7f4.khanacademy.org/


packages/wonder-blocks-cell/src/components/detail-cell.tsx
```diff
@@ -94,7 +94,7 @@ const DetailCell = function (props: DetailCellProps): React.ReactElement {
 
 const styles = StyleSheet.create({
     subtitle: {
-        color: color.offBlack64,
+        color: "#6D6F74", // TODO: use token
```
beaesguerra, line 97: I'm getting feedback from design on if we can use a token for the new subtitle color (or if this is a one-off): https://www.figma.com/design/qSKXmIvmsNPKFlQNP5cztQ?node-id=19669-5006#1098834936

packages/wonder-blocks-cell/src/components/detail-cell.tsx
beaesguerra, line 1: One thing I noticed about DetailCell is that it is used by OptionItem. The updated styles aren't propagated to the OptionItems in the dropdown component because OptionItem [overrides a lot of the styles](https://github.com/Khan/wonder-blocks/blob/d8c91db1a7cc06492e5dd8f74e4351fde7a5905f/packages/wonder-blocks-dropdown/src/components/option-item.tsx#L312-L445). Is it expected that DetailCell and OptionItem have different styling?

beaesguerra:
Converting this PR to a draft while I get some feedback on introducing a new token for the cell subtitle text (a darked `fadedOffBlack` value. I put together a [story to show potential alternatives](https://5e1bf4b385e3fb0020b7073c-eyzdtwcxxx.chromatic.com/?path=/story/grays--default) and will remove it once finalized!

__docs__/wonder-blocks-cell/detail-cell.stories.tsx
```diff
@@ -4,7 +4,7 @@ import {MemoryRouter, Route, Switch} from "react-router-dom";
 import type {Meta, StoryObj} from "@storybook/react";
 
 import {View} from "@khanacademy/wonder-blocks-core";
-import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {border, color, spacing} from "@khanacademy/wonder-blocks-tokens";
```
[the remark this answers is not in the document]
beaesguerra: I will do this in a follow up PR and release the changes together!

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -309,28 +330,42 @@ const styles = StyleSheet.create({
...
             borderRadius: spacing.xxxSmall_4,
         },
+        [":focus-visible[aria-disabled=true]:after" as any]: {
+            borderColor: semanticColor.action.disabled.default,
+        },
 
         // hover + enabled
         [":hover[aria-disabled=false]" as any]: {
-            background: color.offBlack8,
+            background: color.fadedBlue8,
```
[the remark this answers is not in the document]
beaesguerra: I'll switch to using semantic tokens in a follow up PR and include this! Thanks for the suggestion!

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -342,17 +343,25 @@ const styles = StyleSheet.create({
...
-                width: border.width.thin,
-                backgroundColor: semanticColor.surface.emphasis,
-            },
         },
+        // press + enabled + not currently selected (active prop: false)
+        // Using the first child to apply the left bar indicator on the pressed
+        // state because setting the styles on the clickable element
+        // directly causes issues since overflow must be hidden for cases where
+        // the border is rounded
+        [":active[aria-disabled=false]:not([aria-current=true]) > *:first-child" as any]:
```
beaesguerra, line 352: This selector 😅 Let me know if there's a better way to do this! 

More context:
- This addresses the issue caught in Chromatic where providing a maxHeight to the MultiSelect caused the OptionItems to overlap. This was introduced when adding `overflow: hidden` to the root of the cell
- We need the `overflow: hidden` because there are custom overrides in webapp that sets the border-radius on DetailCell and we want to make sure the new left bar indicator stays within the cell
- To fix this, the `overflow: hidden` and the left bar indicator `:before` styling is moved to the `innerWrapper` of the cell, instead of the root
- Since the pressed styling (`:active`) is on the root element, we use `> *:first-child` to access to innerWrapper

Alternatives I considered:
- keep track of pressed state using JS and apply pressed styles directly to the innerWrapper (won't be able to capture :active state in visual regression tests)
- using an alternative way to apply the left bar indicator where we don't have to set overflow: hidden
  - the left bar curves with the border radius in other alternatives I explored: https://jsfiddle.net/1dghqx0t/24/

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -345,11 +348,13 @@ const styles = StyleSheet.create({
...
-        // directly causes issues since overflow must be hidden for cases where
-        // the border is rounded
-        [":active[aria-disabled=false]:not([aria-current=true]) > *:first-child" as any]:
+        // We apply the left bar indicator styles on the inner-wrapper element
+        // instead of the clickable element directly because we need to hide the
+        // left bar overflow when custom cell styles apply a border-radius. We
+        // have overflow: hidden on the inner wrapper instead of the clickable element
+        // because setting it on the clickable element causes issues with existing
+        // cases.
+        [":active[aria-disabled=false]:not([aria-current=true]) .inner-wrapper" as any]:
```
beaesguerra, line 357: Continuing https://github.com/Khan/wonder-blocks/pull/2435#discussion_r1926129345 : Discussed with @jandrade , we can use a class selector on the inner wrapper element directly (instead of using `:first-child`) and remove the `>` to reduce specificity so it is less brittle if the structure changes!

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -118,7 +123,11 @@ function CellInner(props: CellCoreProps): React.ReactElement {
                 // custom styles
                 style,
                 horizontalRuleStyles,
+                active && styles.activeInnerWrapper,
             ]}
+            // Set className so we can set styles on the inner wrapper directly
+            // when the clickable element is pressed
+            className="inner-wrapper"
```
jandrade, line 130: **praise:** Thanks for adding this comment explaining the why we do this for this very specific case.

jandrade (approved):
Looks really great! thanks for all the details you've put into this. 🚀

beaesguerra:
@jandrade I made a small update to add `height: 100%` to the inner wrapper so that it will always be the same height as the parent (I came across this while testing the snapshot in webapp!).
image


I also updated the branch with `main`. Chromatic now shows the diffs for the change in the secondary text color! Let me know if you have any questions!

__docs__/wonder-blocks-cell/detail-cell.stories.tsx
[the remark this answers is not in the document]
beaesguerra: Follow up PR for using semantic tokens in CellCore: https://github.com/Khan/wonder-blocks/pull/2448 !

jandrade (approved):
Looks great! thanks for the extra context on the recent changes.

beaesguerra:
Now that the last batch of WB changes are updated in webapp, I updated this PR with the latest changes and re-tested the cell changes in webapp (https://github.com/Khan/webapp/pull/29021). The Chromatic snapshots need to be approved again since the secondary text token was updated in this PR and there are more things using semantic tokens now!

PR #2440, opened 2025-01-22
[WB-1814.2] Refactor TextField and TextArea to use semantic colors

jandrade:
## Summary:

- Form: Migrated `color` instances to use `semanticColor` internally in
TextField and TextArea.

- Tokens: Added `border.focus` semantic color token to use for the focus outline

Issue: WB-1814

## Test plan:

Navigate to the Storybook for `TextField` and `TextArea` components and verify
that the colors are looking as expected.

- /?path=/docs/packages-form-textarea--docs
- /?path=/docs/packages-form-textfield--docs

### DIFF
--- a/packages/wonder-blocks-form/src/components/field-heading.tsx
+++ b/packages/wonder-blocks-form/src/components/field-heading.tsx
@@ -3,7 +3,7 @@ import {StyleSheet} from "aphrodite";
 
 import {View, addStyle, StyleType} from "@khanacademy/wonder-blocks-core";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
-import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {semanticColor, spacing} from "@khanacademy/wonder-blocks-tokens";
 import {LabelMedium, LabelSmall} from "@khanacademy/wonder-blocks-typography";
 
 type Props = {
@@ -136,15 +136,15 @@ export default class FieldHeading extends React.Component<Props> {
 
 const styles = StyleSheet.create({
     label: {
-        color: color.offBlack,
+        color: semanticColor.text.primary,
     },
     description: {
-        color: color.offBlack64,
+        color: semanticColor.text.secondary,
     },
     error: {
-        color: color.red,
+        color: semanticColor.status.critical.foreground,
     },
     required: {
-        color: color.red,
+        color: semanticColor.status.critical.foreground,
     },
 });
--- a/packages/wonder-blocks-form/src/components/text-area.tsx
+++ b/packages/wonder-blocks-form/src/components/text-area.tsx
@@ -7,7 +7,12 @@ import {
     addStyle,
     View,
 } from "@khanacademy/wonder-blocks-core";
-import {border, color, font, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {
+    border,
+    font,
+    semanticColor,
+    spacing,
+} from "@khanacademy/wonder-blocks-tokens";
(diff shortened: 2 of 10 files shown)


packages/wonder-blocks-form/src/components/field-heading.tsx
```diff
@@ -136,15 +136,15 @@ export default class FieldHeading extends React.Component<Props> {
...
-        color: color.offBlack64,
+        color: semanticColor.text.secondary,
     },
     error: {
-        color: color.red,
+        color: semanticColor.status.critical.foreground,
     },
     required: {
-        color: color.red,
+        color: semanticColor.status.critical.foreground,
```
beaesguerra, line 148: So clean and straightforward 🤩

packages/wonder-blocks-form/src/components/text-area.tsx
```diff
@@ -312,45 +317,45 @@ const styles = StyleSheet.create({
...
     default: {
-        background: color.white,
-        border: `1px solid ${color.offBlack50}`,
-        color: color.offBlack,
+        background: semanticColor.surface.primary,
+        border: `${border.width.hairline}px solid ${semanticColor.border.strong}`,
+        color: semanticColor.text.primary,
         "::placeholder": {
-            color: color.offBlack64,
+            color: semanticColor.text.secondary,
```
beaesguerra, line 324: Same comment here around if we want to have a separate token for placeholder text color! (comment from other PR: https://github.com/Khan/wonder-blocks/pull/2438#discussion_r1927547218)

packages/wonder-blocks-form/src/components/text-area.tsx
```diff
@@ -312,45 +317,45 @@ const styles = StyleSheet.create({
...
+            outline: `${border.width.thin}px solid ${semanticColor.action.disabled.default}`,
+            outlineOffset: -3,
         },
     },
     error: {
-        background: color.fadedRed8,
-        border: `1px solid ${color.red}`,
-        color: color.offBlack,
+        background: semanticColor.status.critical.background,
+        border: `${border.width.hairline}px solid ${semanticColor.status.critical.foreground}`,
```
beaesguerra, line 351: Would it make sense to have `semanticColor.status.critical.border` here and in the focus-visible state?

packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
```diff
@@ -4,6 +4,7 @@ const border = {
     primary: color.fadedOffBlack16,
     subtle: color.fadedOffBlack8,
     strong: color.fadedOffBlack50,
+    focus: color.blue,
```
beaesguerra, line 7: 🎉

packages/wonder-blocks-form/src/components/text-field.tsx
beaesguerra, line 1: Same comments from TextArea for TextField since they have similar styles!

packages/wonder-blocks-form/src/components/text-area.tsx
jandrade: Discussed with Design and they are good with this approach.

packages/wonder-blocks-form/src/components/text-area.tsx
jandrade: As mentioned in other PRs, the goal is to keep `status` as it is. I'll create a ticket to also simplify the `action` tokens once we have something more defined a more stable visual direction for the new experience.

PR #2441, opened 2025-01-23
[WB-1814.3] Refactor Switch to use semantic colors

jandrade:
## Summary:

Continuing with the semantic colors migration, this PR covers the `Switch`
component.

- Refactored the `Switch` themes to use semantic colors.
- Added `All Variants` story to switch.
- Disabled chromatic snapshots for existing stories (now using All Variants).

### Implementation plan:

1. #2439
2. #2440
3. Switch (current PR)
4. Accordion, Banner, BirthdayPicker
5. IconButton
6. Clickable, Link
7. Modal
8. Popover, Tooltip
9. Pill, Toolbar

Issue: WB-1814

## Test plan:

- Navigate to /?path=/story/packages-switch-switch-all-variants--default
- Verify that the styles are correct.


__docs__/wonder-blocks-switch/switch.stories.tsx
```diff
@@ -109,22 +113,13 @@ export const Controlled: StoryComponentType = {
         await userEvent.tab();
         await userEvent.tab();
 
-        expect(switchWithIcon).toHaveStyle(
-            "background-color: rgba(33, 36, 44, 0.5)",
-        );
-        expect(switchWithIcon).toHaveStyle(
-            "outline: 2px solid rgb(24, 101, 242)",
-        );
```
jandrade, line 117: **note:** Removed these assertions now that we fully check this via Chromatic snapshots

packages/wonder-blocks-switch/src/components/switch.tsx
```diff
@@ -155,6 +155,10 @@ const themedSharedStyles: ThemedStylesFn<SwitchThemeContract> = (theme) => ({
         ":hover": {
             outline: "none",
         },
+        ":focus-within": {
+            outline: `solid ${theme.size.width.small}px ${theme.color.outline.default}`,
+            outlineOffset: theme.size.offset.default,
+        },
```
jandrade, line 161: **note:** I noticed that this was missing to display the focus outline on disabled switches.

packages/wonder-blocks-switch/src/themes/default.ts
```diff
@@ -1,29 +1,35 @@
...
             switch: {
-                off: tokens.color.offBlack50,
-                disabledOff: tokens.color.offBlack32,
-                activeOff: tokens.color.offBlack64,
-                on: tokens.color.blue,
-                disabledOn: tokens.color.fadedBlue,
-                activeOn: tokens.color.activeBlue,
+                off: action.default.border,
+                disabledOff: semanticColor.action.disabled.default,
+                activeOff: semanticColor.text.secondary,
```
beaesguerra, line 14: Would it make sense to use the primitive token directly here or a different semantic token? If we were to change the secondary text color value, would we expect things that are not text to change as well? 

Same question about using the border color for `theme.color.bg.switch.off`!

beaesguerra (approved):
Looks good! Left a non-blocking question about token use! 😄

__docs__/wonder-blocks-switch/switch-variants.stories.tsx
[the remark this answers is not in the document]
beaesguerra: Chromatic only showed one test and it wasn't a diff for some reason! It looked good to me so I approved it anyways! https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=679a4e2a99ecb0356d5a8663

packages/wonder-blocks-switch/src/themes/default.ts
jandrade: That's an excellent question! I think using primitive token would make sense here for now.

Side note: I wonder if we should change `icon.primary` to use the new `offBlack72` token. wdyt? should we raise this to Caitlyn?

beaesguerra: I was wondering something similar! At one point, I came across text with an icon in `webapp`. The text now uses `text.secondary` (offBlack72), so I thought I'd use `icon.secondary` for the icon to match it. For icons, `secondary` is darker than `primary`! Do we want to swap them so they are consistent with the text tokens (ie. use primary text with primary icon color, use secondary test with secondary icon color)? Also, I wonder what the use cases are for using the secondary icon color (currently `offBlack`), maybe we don't need both?

I think this is something we can bring up with Caitlyn for the token work!

PR #2443, opened 2025-01-28
Fix bug in popover child tab indices

nedredmond:
`focus-manager.tsx` is overriding tabindex settings on all of its children. Anything that needs to implement roving focus and restrict tabbing, like a `radiogroup`, is not possible since tabindex is always reset to 0 when focus is inside of the popover.

This fixes that issue by allowing the original tabindex to remain when internal focus is not disabled.

### DIFF
--- a/packages/wonder-blocks-popover/src/components/__tests__/focus-manager.test.tsx
+++ b/packages/wonder-blocks-popover/src/components/__tests__/focus-manager.test.tsx
@@ -177,4 +177,50 @@ describe("FocusManager", () => {
         // Assert
         expect(focusableElementInside).toHaveAttribute("tabIndex", "-1");
     });
+
+    it("should keep the original focusability of the internal elements when tabindex is set explicitly", async () => {
+        // Arrange
+        const externalNodes = (
+            <div>
+                <button>Open popover</button>
+            </div>
+        );
+        render(externalNodes);
+
+        // get the anchor reference to be able pass it to the FocusManager
+        const anchorElementNode = await screen.findByRole("button", {
+            name: "Open popover",
+        });
+
+        render(
+            <FocusManager anchorElement={anchorElementNode}>
+                <div>
+                    <button tabIndex={0}>first focusable element inside</button>
+                    <button tabIndex={-1}>
+                        second focusable element inside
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-popover/src/components/__tests__/focus-manager.test.tsx
```diff
@@ -177,4 +177,50 @@ describe("FocusManager", () => {
...
+        );
+        const thirdFocusableElementInside = await screen.findByText(
+            "third focusable element inside",
+        );
+
+        // Assert
+        expect(firstFocusableElementInside).toHaveAttribute("tabIndex", "0"); // explicit set to 0
+        expect(secondFocusableElementInside).toHaveAttribute("tabIndex", "-1"); // explicit set to -1
+        expect(thirdFocusableElementInside).toHaveAttribute("tabIndex", "0"); // not set, should default to 0
+    });
```
jandrade, line 225: **suggestion:** Could you please split these assertions into 3 separate tests? This would follow our AAA approach and it would simplify understanding when there's a specific failure in that test case.

jandrade (commented):
The implementation makes total sense! I'd like to have some minor refactor in the tests before giving a final approval. Thanks for fixing this! (also, this would need to be a `patch` version).

packages/wonder-blocks-popover/src/components/__tests__/focus-manager.test.tsx
nedredmond: @jandrade I am having issues with authentication in git cli with WB specifically!

I can't push the changeset or change this 😢
I was only able to make the PR through the GitHub CLI, not Our Lovely CLI or GitHub Desktop

nedredmond: Was there a change to protections in this repo?

nedredmond: fixed it, updated

jandrade (approved):
Looks great! thanks for the quick fix + modifying the tests 💯 🏅 🚀

PR #2446, opened 2025-01-29
[WB-1814.4] Refactor Accordion, Banner, BirthdayPicker to use semantic colors

jandrade:
## Summary:

Next step is to refactor the `Accordion`, `Banner`, and `BirthdayPicker`
components to use semantic colors.

### Implementation plan:

1. #2439
2. #2440
3. #2441
4. Accordion, Banner, BirthdayPicker (current PR)
5. IconButton
6. Clickable, Link
7. Modal
8. Popover, Tooltip
9. Pill, Toolbar

Issue: WB-1814

## Test plan:

### DIFF
--- a/packages/wonder-blocks-banner/src/components/banner.tsx
+++ b/packages/wonder-blocks-banner/src/components/banner.tsx
@@ -313,6 +313,8 @@ const styles = StyleSheet.create({
         // the base color needs to be hard-coded as white for the
         // intended pastel background color to show up correctly
         // on dark backgrounds.
+        // TODO(WB-1865): Verify if we can change this to use semanticColor
+        // status tokens.
         backgroundColor: semanticColor.surface.primary,
     },
     containerInner: {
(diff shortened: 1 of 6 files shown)


packages/wonder-blocks-banner/src/components/banner.tsx
```diff
@@ -313,7 +313,7 @@ const styles = StyleSheet.create({
         // the base color needs to be hard-coded as white for the
         // intended pastel background color to show up correctly
         // on dark backgrounds.
-        backgroundColor: color.white,
+        backgroundColor: semanticColor.surface.primary,
```
beaesguerra, line 316: Ohh interesting, we don't use the semantic background colors for the background of the banner (we rely on opacity). Do we want to use the semantic background tokens?

beaesguerra (approved):
Looks good! Left a non-blocking question around the banner background color!

packages/wonder-blocks-banner/src/components/banner.tsx
jandrade: Good question! tbh I'm not sure how this will be handled with the Polaris theme. I'm going to revisit this once we add theming support to banner. I'll add a TODO comment to make sure we track this.

PR #2449, opened 2025-01-30
[WB-1814.5] Refactor IconButton to use semantic colors

jandrade:
## Summary:

Next step is to refactor the `IconButton` component to use semantic colors.

- Reworked the theme structure to make it closer to the semanticColor structure.
- Moved all the color-related CSS properties to a separate function (`getStylesByKind`).
- Added `border` to all the states to prepare the work for Polaris as these
icon buttons will look similar to the Button ones.

### Implementation plan:

1. #2439
2. #2440
3. #2441
4. #2446
5. IconButton (current PR)
6. Clickable, Link
7. Modal
8. Popover, Tooltip
9. Pill, Toolbar

Issue: WB-1814

## Test plan:

URL: `/?path=/story/packages-iconbutton-all-variants--default`


packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
```diff
@@ -186,86 +190,134 @@ const sharedStyles = StyleSheet.create({
...
 function getStylesByKind(
-    kind: "primary" | "secondary" | "tertiary",
-    theme: IconButtonThemeContract,
-    color: string,
+    buttonColor: ButtonColor,
+    disabled: boolean,
+    kind: Kind,
     light: boolean,
-    buttonColor: string,
+    theme: IconButtonThemeContract,
```
jandrade, line 224: **note:** Just reordered the arguments alphabetically.

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
```diff
@@ -186,86 +190,134 @@ const sharedStyles = StyleSheet.create({
...
 ) {
-    switch (kind) {
```
jandrade, line 196: **note:** Changed from `switch` to `if` as constants cannot be redefined within switch statements. I did this as I wanted to be able to reuse `const themeVariant` etc for better readability.

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
```diff
@@ -186,86 +190,134 @@ const sharedStyles = StyleSheet.create({
...
+        ":focus-visible": {},
+        ":active": {},
+        disabled: {},
+    };
 }
 
 const _generateStyles = (
-    buttonColor = "default",
-    kind: "primary" | "secondary" | "tertiary",
+    buttonColor: ButtonColor = "default",
```
jandrade, line 312: **note:** Same here related to reordering args alphabetically.

packages/wonder-blocks-icon-button/src/themes/default.ts
jandrade, line 1: **note:** Modified the structure to make it really closer to `semanticColor.action`. Note that this time used `primary`, `secondary`, etc instead of `filled, `outlined` to be able to map the keys to the props defined in `IconButton` (e.g. `kind="primary"`). Still undecided with which approach to take. Any thoughts are completely welcome!

__docs__/wonder-blocks-icon-button/icon-button-variants.stories.tsx
```diff
@@ -16,11 +16,7 @@ import IconButton from "@khanacademy/wonder-blocks-icon-button";
  */
 export default {
     title: "Packages / IconButton / All Variants",
-    parameters: {
-        docs: {
-            autodocs: false,
-        },
-    },
+    tags: ["!autodocs"],
```
marcysutton, line 19: I didn't know you could do this! Nice!

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
```diff
@@ -101,6 +104,7 @@ const IconButtonCore: React.ForwardRefExoticComponent<
     const renderInner = (router: any): React.ReactNode => {
         const buttonStyles = _generateStyles(
             color,
+            !!disabled,
```
marcysutton, line 107: **question:** Is the boolean coercion of `disabled` meant to handle an undefined case?

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
```diff
@@ -186,86 +190,134 @@ const sharedStyles = StyleSheet.create({
...
+
+/**
+ * Returns the action type based on the button color and disabled state.
+ *
+ * This is useful to determine which token variant to use for the button, which
+ * is based on the theme structure.
+ */
+function getActionType(buttonColor: ButtonColor, disabled: boolean) {
+    const actionType =
+        buttonColor === "destructive" ? "destructive" : "progressive";
```
marcysutton, line 210: **question:** could you use nullish coalescing here? (out of curiosity)

```
const actionType = buttonColor === "destructive" ?? "progressive";
```

packages/wonder-blocks-icon-button/src/themes/default.ts
```diff
@@ -1,109 +1,157 @@
...
+};
+
+/**
+ * The color styles shared between all the button kinds.
+ *
+ * NOTE: This will change when we create the new polaris theme as IconButton
+ * will apply the same styles as Button.
+ */
+const baseColorStates = {
+    ...semanticColor.action.outlined,
```
marcysutton, line 23: **question:** what case does this spread token cover? Could you add a comment for it?

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
```diff
@@ -18,6 +18,9 @@ import {
     IconButtonThemeContract,
 } from "../themes/themed-icon-button";
 
+type Kind = "primary" | "secondary" | "tertiary";
+type ButtonColor = "default" | "destructive";
```
marcysutton, line 22: **question:** Where does `progressive` come in? Should that be included here among the options?

marcysutton (approved):
This looks like a really robust overhaul! I left some questions on things I wasn't understanding 100%. But I think it's going to be way easier to theme with this in place. Nice work!

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
```diff
@@ -186,86 +190,134 @@ const sharedStyles = StyleSheet.create({
...
     theme: IconButtonThemeContract,
     themeName: string,
 ) => {
-    const color: string =
-        buttonColor === "destructive"
-            ? theme.color.stroke.critical.default
-            : theme.color.stroke.action.default;
-
-    const buttonType = `${color}-${kind}-${light}-${size}-${themeName}`;
+    const buttonType = `${buttonColor}-d_${disabled}-${kind}-l_${light}-${size}-${themeName}`;
```
beaesguerra, line 320: **question**: What does the `-d_` mean in the buttonType name?

beaesguerra (approved):
Looks good! Left a question around naming!

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
jandrade: That's correct, this is mostly required b/c that prop is optional.

packages/wonder-blocks-icon-button/src/themes/default.ts
jandrade: Good point! I'll add it.

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
jandrade: ahh great question..... so the thing is that historically we have called the "blue" color version as `default`, but this is changing with Polaris. This is a change that it is being proposed in Figma (which makes sense IMO). So we have to make that mapping (`default` -> `progressive`), but the goal is to align on the naming later. I'll add a note in the code to clarify that.

jandrade: I created a ticket to address this later. I'll add the TODO comment to point to it. Thanks again for raising this question.

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
jandrade: In this case it wouldn't work because the left-hand operand would return `true` if the condition is fulfilled.

```
const actionType = buttonColor === "destructive" ?? "progressive";
// when `buttonColor=destructive` => actionType=true
// when `buttonColor=default` => actionType="progressive"
```

Apologies if I'm being too literal 😅, but maybe I'm missing something in the expression (this could still use nullish coalescing here, which would be cool).

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
jandrade: Disabled, but that's just to be able to cache this string properly.

This is mostly for debugging purposes if needed.

PR #2451, opened 2025-01-31
[🔥AUDIT🔥] Storybook: Add remark-gfm plugin to add GFM MDX support

jandrade:
🖍 _This is an audit!_ 🖍

## Summary:

After the Storybook v8 upgrade, we lost the ability to use GFM in MDX files.
This PR adds the `remark-gfm` plugin to the MDX loader to re-enable GFM support.

For more context, GFM (GitHub Flavored Markdown) is a superset of the original
Markdown specification that adds additional features like tables, task lists,
and strikethrough text.

https://storybook.js.org/docs/writing-docs/mdx#markdown-tables-arent-rendering-correctly

Issue: WB-XXXX

## Test plan:

Navigate to any docs page that renders markdown tables and verify that they are
rendered correctly.

/?path=/docs/packages-clickable-clickable-accessibility--docs

BEFORE | AFTER
Screenshot 2025-01-31 at 3 44 47 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,6 +115,7 @@
     "playwright": "^1.49.0",
     "prettier": "^2.8.1",
     "react-refresh": "^0.14.0",
+    "remark-gfm": "^4.0.0",
     "rollup": "^2.79.1",
     "rollup-plugin-babel": "^4.4.0",
     "rollup-plugin-node-externals": "^7.1.2",
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -1,3 +1,4 @@
+import remarkGfm from "remark-gfm";
 import type {StorybookConfig} from "@storybook/react-vite";
 
 const config: StorybookConfig = {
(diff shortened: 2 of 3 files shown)


jandrade:
Here is how it looks like in the PR Storybook site: https://5e1bf4b385e3fb0020b7073c-rscxtgmwsv.chromatic.com/?path=/docs/packages-clickable-clickable-accessibility--docs

package.json
```diff
@@ -115,6 +115,7 @@
     "playwright": "^1.49.0",
     "prettier": "^2.8.1",
     "react-refresh": "^0.14.0",
+    "remark-gfm": "^4.0.0",
```
somewhatabstract, line 118: **question:** Was there a decision record associated with this 3rd party dep? If so, it may be useful to link that in this PR.

somewhatabstract (commented):
LGTM! Thanks for fixing this!

PR #2453, opened 2025-02-03
[WB-1870] Adopt pnpm in Wonder Blocks

jandrade:
## Summary:

- Switched to pnpm.
- Udpated most of the devDependencies to the latest version. Except `eslint` as
the latest version is not compatible with our current setup.
- Fixed TS errors in tests.
- Fixed stories, lint, prettier, etc.
- Updated rollup config to use pnpm (for builds).

Issue: WB-1870

## Test plan:

Verify that all the checks pass and that the Storybook site works fine.


.github/actions/shared-node-cache/action.yml
jandrade, line 1: **note:** I created this custom action based on the one @jeresig created in the main repo (see `./github/actions/shared-node-cache/action.yml`).

A bunch of files are now using this internal action instead of the external one to avoid breaking other repos that still rely on `yarn`.

.github/workflows/node-ci-pr.yml
```diff
@@ -49,29 +49,12 @@ jobs:
...
-        with:
-            changed-files: ${{ steps.changed.outputs.files }}
-            files: packages/ # Only look for changes in packages
-            globs: "!(**/__tests__/**), !(**/dist/*), !(**/*.test.{ts,tsx})" # Ignore test files
-            matchAllGlobs: true # All globs must match (disjunction is the default)
-            conjunctive: true # Only return files that match all of the above
-
-      - name: Verify changeset entries
-        uses: Khan/changeset-per-package@v1.0.2-pre
+      - uses: Khan/actions@check-for-changeset-v1
```
jandrade, line 52: **note:** Reverted back to use `Khan/actions@check-for-changeset-v1` as it doesn't rely on a package manager to resolve dependency changes.

The reason why `Khan/changeset-per-package` is no longer viable with `pnpm` is that this action relies heavily on `yarn`: https://github.com/Khan/changeset-per-package/blob/main/src/main.ts#L35

One thing to take into consideration is that the WB repo never got issues with `check-for-changeset`, so reverting back to it should be safe.

.storybook/.babelrc.js
jandrade, line 1: **note:** Updated some packages, and that needed to update the plugin references in here.

.storybook/main.ts
```diff
@@ -25,9 +24,21 @@ const config: StorybookConfig = {
...
+        // Merge custom configuration into the default config
+        const {mergeConfig} = await import("vite");
+
+        return mergeConfig(config, {
+            // Prevent Vite from inlining phosphor-icons
+            build: {
+                assetsInlineLimit: 0,
+            },
+        });
+    },
```
jandrade, line 41: **note:** I tried to define `assetsInlineLimit` directly in `vite.config.ts` but Storybook seems to have an issue reading that file, so I had to move that part here to avoid having issues with SVG rendering in Storybook.

__docs__/wonder-blocks-accordion/accordion.stories.tsx
```diff
@@ -464,7 +464,7 @@ export const SingleSection: StoryComponentType = {
         return (
             <Accordion>
                 {[
-                    <AccordionSection header="First section">
+                    <AccordionSection header="First section" key={0}>
```
jandrade, line 467: **note:** Added `key` in a bunch of story examples to prevent a browser warning where a collection of nodes need a specific key/id in React.

build-settings/rollup.config.mjs
```diff
@@ -41,7 +41,7 @@ const createConfig = (pkgName) => {
                 presets,
                 plugins,
                 exclude: "node_modules/**",
-                runtimeHelpers: true,
+                babelHelpers: "runtime",
```
jandrade, line 44: **note:** This API changed with `rollup` v4, so I just had to make it compatible.

package.json
```diff
@@ -5,132 +5,133 @@
...
-    "test": "yarn run jest && yarn run test:common",
+    "test": "pnpm run jest && pnpm run test:common",
     "typecheck": "tsc",
     "typewatch": "tsc --noEmit --watch --incremental",
-    "add:devdepbysha": "bash -c 'yarn add -W --dev \"git+https://git@github.com/Khan/$0.git#$1\"'"
+    "add:devdepbysha": "bash -c 'pnpm add -W --dev \"git+https://git@github.com/Khan/$0.git#$1\"'"
   },
   "author": "",
   "license": "MIT",
   "devDependencies": {
```
jandrade, line 39: **note:** Used `pnpm u -i --latest` to update a bunch of dev dependencies in the repo. I tried to preserve or update to some specific versions in some cases to be compatible with webapp (e.g. `babel`).

packages/wonder-blocks-accordion/package.json
```diff
@@ -29,6 +30,6 @@
     "react": "18.2.0"
   },
   "devDependencies": {
-    "@khanacademy/wb-dev-build-settings": "^2.0.0"
+    "@khanacademy/wb-dev-build-settings": "workspace:*"
```
jandrade, line 33: **note:** modified this in all WB packages to manage `build-settings` as a workspace alias.

https://pnpm.io/workspaces#referencing-workspace-packages-through-aliases

beaesguerra (approved):
Looks good! I left some questions for my own learning 😄 Thanks for switching us to pnpm in WB 🎉

packages/wonder-blocks-accordion/package.json
jandrade: That's an excellent question! I'm not sure why I didn't try that, but I don't see why not. My only concern would be that maybe changesets would need to find a way to figure out how to bump to the proper versions. I'll try to see what happens.

packages/wonder-blocks-data/src/util/__tests__/serializable-in-memory-cache.test.ts
```diff
@@ -23,6 +22,7 @@ describe("SerializableInMemoryCache", () => {
 
         it("should throw if the cloning fails", () => {
             // Arrange
+            const WSCore = require("@khanacademy/wonder-stuff-core");
```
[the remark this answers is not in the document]
jandrade: TBH I changed devDeps and at some point those tests started to break. I don't recall changing anything related to Jest, so this was surprising to me. I applied this solution as I encountered similar situations in webapp and/or WB in the past and that solved the issue.

My guess is that `pnpm` is not able to import the same dependency that is used in the implementation, so the mock is never called. I switched to this approach to ensure that we always import/require the right dep. It might have to do with the fact that Jest still uses CJS for testing (where as the code uses ES Modules).

packages/wonder-blocks-accordion/package.json
jandrade: @beaesguerra I was looking at monorepos with similar tooling and it seems that it should work (in theory). But in order to test that properly, I'm going to land this into `main`, so we can get the "Check build sizes" workflow step running, then I'll experiment with it in a separate PR to see if something breaks and/or the bundled versions are affected. If that works, it would be really great as we would reduce the times we'd need to update `package.json` files, or it would simplify such updates. I'll keep you posted about this, and thanks for the suggestion/question!

beaesguerra: Sounds like a good plan, thanks for looking into this!

.storybook/main.ts
jeresig: Ohh interesting!

jeresig (approved):
Changes all look reasonable to me - glad you were able to find a good solution here!

.github/actions/shared-node-cache/action.yml
```diff
@@ -0,0 +1,33 @@
+name: 'Shared Node Cache'
+description: 'Install & cache our npm dependencies'
+
+# The inputs this action expects.
+inputs:
+  node-version:
+    description: 'Node version to use'
+    required: false
+    default: '20.x'
+  cache-bust:
```
jeremywiebe, line 10: This appears unused... or does a value passed to it do enough to bust the cache?

PR #2455, opened 2025-02-03
[timeouthooks.bugfix] Fix clear function on useTimeout and useInterval

somewhatabstract:
## Summary:
This fixes a bug I found in the API returned by the `useTimeout` and `useInterval` hooks. The `clear` function returned by these hooks was not properly honouring the clear policy explicitly passed to the hook options. This PR fixes that issue.

## Test plan:
`yarn test`
`yarn start:storybook` and visit the `useTimeout` and `useInterval` stories. Go to the OnDemandAndResolveOnClear story and see that clicking the Clear button will invoke the corresponding action as it should - this was not the case, previously.

### DIFF
--- a/__docs__/wonder-blocks-timing/use-interval.stories.tsx
+++ b/__docs__/wonder-blocks-timing/use-interval.stories.tsx
@@ -8,6 +8,7 @@ import {
     ClearPolicy,
     SchedulePolicy,
 } from "@khanacademy/wonder-blocks-timing";
+import {Body} from "@khanacademy/wonder-blocks-typography";
 
 export default {
     title: "Packages / Timing / useInterval",
@@ -21,13 +22,23 @@ export default {
 
 export const Immediately = () => {
     const [callCount, setCallCount] = React.useState(0);
+    const [intervalSet, setIntervalSet] = React.useState(false);
     const callback = React.useCallback(() => {
         setCallCount((callCount) => callCount + 1);
     }, []);
     const interval = useInterval(callback, 1000);
+    useInterval(() => {
+        // Need to update on an interval as the returned `isSet` value is not
+        // driven by React state.
+        setIntervalSet(interval.isSet);
+    }, 100);
     return (
         <View>
-            <View>isSet = {interval.isSet.toString()}</View>
+            <Body>
+                Interval should fire every second until cleared. Setting the
+                interval again resets the interval.
+            </Body>
+            <View>isSet = {String(intervalSet)}</View>
             <View>callCount = {callCount}</View>
             <View style={{flexDirection: "row"}}>
                 <Button onClick={() => interval.set()}>Set interval</Button>
@@ -39,6 +50,7 @@ export const Immediately = () => {
 
 export const OnDemandAndResolveOnClear = () => {
     const [callCount, setCallCount] = React.useState(0);
+    const [intervalSet, setIntervalSet] = React.useState(false);
     const callback = React.useCallback(() => {
         setCallCount((callCount) => callCount + 1);
     }, []);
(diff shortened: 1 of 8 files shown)


packages/wonder-blocks-timing/src/hooks/use-interval.ts
```diff
@@ -87,14 +89,14 @@ export function useInterval(
             set: () => {
                 intervalRef.current?.set();
             },
-            clear: (policy?: ClearPolicy) => {
+            clear: (policy: ClearPolicy | undefined = clearPolicy) => {
                 intervalRef.current?.clear(policy);
             },
```
kevinb-khan, line 94: **Question:** Do we need the `| undefined`.  I believe it's only necessary to support someone explicitly passing `undefined` as an argument to `clear`.

packages/wonder-blocks-timing/src/hooks/use-timeout.ts
```diff
@@ -87,14 +89,16 @@ export function useTimeout(
             set: () => {
                 timeoutRef.current?.set();
             },
-            clear: (policy?: ClearPolicy) => {
+            clear: (policy: ClearPolicy | undefined = clearPolicy) => {
+                // Note that we default to the clear policy passed to the hook
+                // so that this works as folks might expect.
```
kevinb-khan, line 94: Same concern as use-interval.ts.

packages/wonder-blocks-timing/src/hooks/use-interval.ts
```diff
@@ -31,7 +31,9 @@ import Interval from "../util/interval";
  * scheduling the interval, use `SchedulePolicy.OnDemand`.
  * @returns An `IInterval` API for interacting with the given interval. This
  * API is a no-op if called when not mounted. This means that any calls prior
- * to mounting or after unmounting will not have any effect.
+ * to mounting or after unmounting will not have any effect. This API is
+ * not reactive, so do not deconstruct the return value, but instead
+ * dereference it at the time of use.
```
kevinb-khan, line 36: **Question:** What happens if you destruct the return value?

__docs__/wonder-blocks-timing/use-timeout.mdx
```diff
@@ -33,58 +33,14 @@ called when the timeout is cleared.
...
-            <View>isSet = {isSet.toString()}</View>
-            <View>callCount = {callCount}</View>
-            <View style={{flexDirection: "row"}}>
-                <Button onClick={() => set()}>Set timeout</Button>
-                <Button onClick={() => clear()}>Clear timeout</Button>
-            </View>
-        </View>
-    );
-};
-```
```
kevinb-khan, line 64: I guess readers can look at the source code for the story.

__docs__/wonder-blocks-timing/use-interval.stories.tsx
```diff
@@ -39,16 +50,28 @@ export const Immediately = () => {
...
+    }, 100);
     return (
         <View>
-            <View>isSet = {interval.isSet.toString()}</View>
+            <Body>
+                Interval will not start until set is explicitly invoked.
+                Interval should fire every second until cleared. Clearing the
+                interval will invoke the interval action one more time. Setting
+                the interval again resets the interval.
+            </Body>
```
kevinb-khan, line 73: **Praise:** Thanks for adding this.

kevinb-khan (approved):
LGTM.  Thanks for fixing this issue.  I have a few questions in the comments but nothing blocking. 🎉

packages/wonder-blocks-timing/src/hooks/use-interval.ts
somewhatabstract: Yes, we do because `clearPolicy` could be `undefined.

packages/wonder-blocks-timing/src/hooks/use-timeout.ts
somewhatabstract: Yes, we do because clearPolicy could be `undefined.

packages/wonder-blocks-timing/src/hooks/use-interval.ts
somewhatabstract: Because it captures the values as they are at that point. The object returned isn't reactive so changes won't trigger re-renders which may leave effects using stale values. This is especially problematic for say `isSet`, which is a boolean.

__docs__/wonder-blocks-timing/use-timeout.mdx
somewhatabstract: Yeah, it's already included. I don't know why we're duplicating it. It's inlined right with there.

packages/wonder-blocks-timing/src/hooks/use-interval.ts
somewhatabstract: We can probably update the API make `isSet` a function instead of a getter, but that's a bigger change. Right now, these hooks return our schedule interfaces like `ITimeout` and `IInterval`. Changing the return values that way means either redefining the interface for all cases, or make these hooks return something that isn't compliant to that interface.

__docs__/wonder-blocks-timing/use-timeout.mdx
somewhatabstract: That's what the `sourceState="shown"` does. Seems silly that we duplicated the code - perhaps `sourceState` wasn't always a thing?

image

PR #2456, opened 2025-02-04
[birthday-picker-combobox] Add labels to birthday picker

marcysutton:
## Summary:
Updating Birthday Picker to use aria-labels for accessible names. The label placeholders were not applied in this way after WB Dropdowns were changed to use `role="combobox"`.

Issue: https://khanacademy.atlassian.net/browse/WB-1872

## Test plan:

### DIFF
--- a/packages/wonder-blocks-birthday-picker/src/components/birthday-picker.tsx
+++ b/packages/wonder-blocks-birthday-picker/src/components/birthday-picker.tsx
@@ -306,6 +306,7 @@ export default class BirthdayPicker extends React.Component<Props, State> {
         const minWidth = this.getMonthYearWidth(monthYearOnly);
         return (
             <SingleSelect
+                aria-label={this.labels.month}
                 aria-invalid={!!this.state.error}
                 error={!!this.state.error}
                 disabled={disabled}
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-birthday-picker/src/components/birthday-picker.tsx
```diff
@@ -306,6 +306,7 @@ export default class BirthdayPicker extends React.Component<Props, State> {
         const minWidth = this.getMonthYearWidth(monthYearOnly);
         return (
             <SingleSelect
+                aria-label={this.labels.month}
```
beaesguerra, line 309: Nice improvements! 

Do you know what the labelling behaviour is like when the BirthdayPicker is wrapped by a `<legend>` element in a `<fieldset>`? I'm wondering because the [BirthdayPicker wrapper in webapp](https://github.com/Khan/webapp/blob/9f1ec10df5fd724cc77e601b04c444a062dfba08/services/static/javascript/components/birthday-picker/birthday-picker.tsx#L106-L107)       uses fieldset + legend to provide a visual label to the birthday picker too (since it's a group of inputs)!

beaesguerra (approved):
Looks good, thanks for the labelling improvement! Left a non-blocking question around labelling outside of the BirthdayPicker component 😄

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.tsx
marcysutton: Good question! It's kind of a description for the group more than a label, since the dropdowns aren't actually labeled with that legend content. The accessible name for each field starting with this PR is "month", "day", or "year", and the legend describes them all.

PR #2460, opened 2025-02-05
[WB-1870.2] pnpm: Switch to workspace protocol to handle dependency versions

jandrade:
## Summary:

Following up with the pnpm migration, this PR attempts to switch to the
workspace protocol for handling dependency versions. This will allow us to let
`pnpm` resolve the dependencies for us, and we can avoid the need to manually
update the versions in the `package.json` files via changesets when the publish
to npm is done.

More context: https://khanacademy.slack.com/archives/C4PE1QM5Y/p1738792765903199

Issue: WB-1870

## Test plan:

Verify bundle sizes and that the packages are working as expected.

After landing this change, verify that the "Version packages" PR is working
again.

### DIFF
--- a/packages/wonder-blocks-pill/package.json
+++ b/packages/wonder-blocks-pill/package.json
@@ -17,11 +17,12 @@
     "access": "public"
   },
   "dependencies": {
-    "@khanacademy/wonder-blocks-clickable": "^6.0.0",
-    "@khanacademy/wonder-blocks-core": "^12.0.0",
-    "@khanacademy/wonder-blocks-link": "^7.0.8",
-    "@khanacademy/wonder-blocks-tokens": "^4.1.0",
-    "@khanacademy/wonder-blocks-typography": "^3.0.6"
+    "@babel/runtime": "^7.24.5",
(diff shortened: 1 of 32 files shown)


packages/wonder-blocks-pill/package.json
```diff
@@ -17,11 +17,12 @@
     "access": "public"
   },
   "dependencies": {
-    "@khanacademy/wonder-blocks-clickable": "^6.0.0",
-    "@khanacademy/wonder-blocks-core": "^12.0.0",
-    "@khanacademy/wonder-blocks-link": "^7.0.8",
-    "@khanacademy/wonder-blocks-tokens": "^4.1.0",
-    "@khanacademy/wonder-blocks-typography": "^3.0.6"
+    "@babel/runtime": "^7.24.5",
```
beaesguerra, line 20: Referencing our thread in Slack to close the loop! https://khanacademy.slack.com/archives/C4PE1QM5Y/p1738795295547179?thread_ts=1738792765.903199&cid=C4PE1QM5Y 

Because we add this dependency, the bundle size for the pill package has changed!

beaesguerra (approved):
Looks good! 🎉

PR #2461, opened 2025-02-06
[WB-1809] Link: Use CSS pseudo-classes for styling states

jandrade:
## Summary:

Updates `Link` to use CSS pseudo-classes (:hover, :focus-visible, :active)
instead of JS driven states (hovered, focused, pressed).

Also added the `All Variants` story to the `Link` documentation for visual
regression testing.

Issue: WB-1809

## Test plan:

Navigate to `/?path=/story/packages-link-link-all-variants--default` and verify
that the `Link` component behaves as expected when hovered, focused, and
pressed.


packages/wonder-blocks-link/src/components/__tests__/link.test.tsx
```diff
@@ -556,4 +556,19 @@ describe("Link", () => {
             expect(icon).toHaveStyle({maskImage: "url(plus-bold.svg)"});
         });
     });
+
+    test("should have title attribute", () => {
```
jandrade, line 560: **note:** This was previously an interaction test, but I moved it to unit test due to the nature of the test. Another benefit is that unit tests are faster for these kind of assertions (faster in the sense that JSDOM is faster than a real browser).

__docs__/wonder-blocks-link/link.stories.tsx
```diff
@@ -34,806 +29,412 @@ export default {
...
+            >
+                external Primary link
+            </Link>
+            , whereas this is an inline{" "}
+            <Link href="#" kind="secondary" inline={true}>
+                Secondary link
+            </Link>
+            , and an inline{" "}
+            <Link
+                href="https://www.procatinator.com/"
```
beaesguerra, line 341: TIL about procatinator, stuffonmycat, and cat-bounce 😸

__docs__/wonder-blocks-link/link.stories.tsx
```diff
@@ -34,806 +29,412 @@ export default {
                 version={packageConfig.version}
             />
         ),
+        chromatic: {
+            // These stories are being tested in link-variants.stories.tsx
+            disableSnapshot: true,
```
beaesguerra, line 34: **suggestion:** What do you think about disabling snapshots for the specific stories that are duplicated in the variants stories? There are some stories that I think would still be helpful to have for visual regression tests like:
- Opens in a new tab
- Start and end icons
- Inline 
- Inline light
- With Typography
- With Style
- With Title (?)
- Right To Left With Icons

Another option would be to include these scenarios in the All Variants stories too so we have coverage for those cases in the different states as well!

beaesguerra (approved):
Looks good! Left a suggestion around visual regression tests and a suggestion in Chromatic about documenting the Secondary Light variant!

__docs__/wonder-blocks-link/link.stories.tsx
jandrade: ohhh good call about this! I forgot to include the icons in all variants 🤦.

I left `Typography` out on purpose mostly because I felt that these are implementation details that are already tested in typography stories, and all the states are inherited from link.

Also left the `With Style` out b/c it is a very specific case that is out of the possible variants of the link. I'm going to keep the snapshot for now.

`inline` is covered by all variants.

What do you think about this plan?

- Add `startIcon`, `endIcon`, `opens in new tab` to the variants matrix in `All variants`.
- Do this in a follow up PR so it's easier to review. I don't expect UI changes as the icons inherit the color from the link state (aka `currentcolor`).

I'm totally open to suggestions :)

beaesguerra: > Also left the With Style out b/c it is a very specific case that is out of the possible variants of the link. I'm going to keep the snapshot for now.

I find the custom styles in the `WithStyle` stories not too important, but it is helpful to have so we know that custom styles are working properly since we don't cover it in unit tests! Same with the Typography story, it helps confirm that the component works with Typography components!

> inline is covered by all variants.

I find the `Inline` story helpful because it shows the Link used in context! Then we can make sure that any new styling changes don't change the layout (it could be harder to tell in the AllVariants stories where the Link is isolated on it's own!). 

> What do you think about this plan?

This plan sounds good!

jandrade: All fair points! I'll add them back. Thanks for the great discussion 🫶

jandrade: Here are the changes (in upcoming PR): https://github.com/Khan/wonder-blocks/pull/2464/commits/fb9d5c77c13dbafdf7b7632c6de489a832ce9458

jandrade: And the commit re-enabling the stories you mentioned in this thread: https://github.com/Khan/wonder-blocks/pull/2464/commits/bf9bc52d69d445b73c06172950ba50f51c3df65f

PR #2463, opened 2025-02-06
[WB-1874] Remove `kind` prop now that `secondary` kind is being deprecated

jandrade:
## Summary:

Now that we are moving to Polaris, we’ve noticed that the Link.secondary variant
is not used very often (~0.2% of total Link usage). We are going to remove it to
simplify the styling of Links and consolidate on a more consistent pattern
across our sites.

This PR removes the `kind` prop from the Link component and updates the
documentation to reflect this change.

Also removes the `kind="primary"` from the Link instance in `Banner` component.

Issue: WB-1874

## Test plan:

Navigate to `/?path=/docs/packages-link--docs` and verify that the Link
component does not have the `kind` prop anymore.

Also navigate to `/?path=/story/packages-link-link-all-variants--default` and
verify that the "kind" variants are not available anymore. Only "default" and
"inline".

### DIFF
--- a/.changeset/moody-pianos-return.md
+++ b/.changeset/moody-pianos-return.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-link": major
+---
+
+Remove `kind` prop as the `secondary` variant is now dropped.
--- a/__docs__/wonder-blocks-link/link.stories.tsx
+++ b/__docs__/wonder-blocks-link/link.stories.tsx
@@ -40,24 +40,13 @@ export default {
 type StoryComponentType = StoryObj<typeof Link>;
 
 /**
- * By default the link is a `primary` link.
+ * By default the link uses a blue color.
  */
 export const Default: StoryComponentType = {
     args: {
         href: "/",
         children: "The quick brown fox jumps over the lazy dog.",
     },
-};
-
-/**
- * This is a `secondary` link. This links to the top of the page.
- */
-export const Secondary: StoryComponentType = {
-    render: () => (
-        <Link href="#" kind="secondary">
-            The quick brown fox jumps over the lazy dog.
-        </Link>
-    ),
 };
 
 /**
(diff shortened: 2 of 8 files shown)


.changeset/moody-pianos-return.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-link": major
```
beaesguerra, line 2: Does this have an impact on Perseus?

__docs__/wonder-blocks-link/link.stories.tsx
```diff
@@ -40,7 +40,7 @@ export default {
 type StoryComponentType = StoryObj<typeof Link>;
 
 /**
- * By default the link is a `primary` link.
+ * By default the link uses a blue color.
```
beaesguerra, line 43: Would it be helpful to reference the token name in docs instead of `blue` since it might be a different color for other themes?

packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -170,24 +169,15 @@ const sharedStyles = StyleSheet.create({
...
         return styles[buttonType];
     }
 
-    if (kind === "secondary" && light) {
-        throw new Error("Secondary Light links are not supported");
-    }
-
-    if (visitable && kind !== "primary") {
-        throw new Error("Only primary link is visitable");
-    }
```
beaesguerra, line 188: 🧹  Glad we can simplify things!

beaesguerra (approved):
Looks good! Left some non-blocking questions!

__docs__/wonder-blocks-link/link.stories.tsx
jandrade: hmm great question! how about adding a semantic meaning instead?

Something in the lines of: `uses a color that communicates the presence and meaning of interaction` (extracted from the semantic colors docs).

beaesguerra: This sounds good!

.changeset/moody-pianos-return.md
jandrade: No, I checked in the Perseus repo and they don't use this variant.

PR #2464, opened 2025-02-06
[WB-1814.6] Refactor Link to use semantic colors

jandrade:
Next step is to refactor the `Link` component to use semantic colors.

- Reworked the theme structure to make it closer to the semanticColor structure.
- Updated stories to use the new semantic colors.

### Implementation plan:

1. #2439
2. #2440
3. #2441
4. #2446
5. #2449
6. Link (current PR)
7. Modal
8. Popover, Tooltip
9. Pill
10. Clickable, Toolbar

Issue: WB-1814

## Test plan:

URL: `/?path=/story/packages-link-link-all-variants--default`

### DIFF
--- a/__docs__/wonder-blocks-link/link.stories.tsx
+++ b/__docs__/wonder-blocks-link/link.stories.tsx
@@ -364,7 +364,7 @@ export const InlineLight: StoryComponentType = {
             </Link>
             , whereas this is an inline{" "}
             <Link href="#link" visitable={true} inline={true} light={true}>
-                Visitable link (Primary only)
+                Visitable link
             </Link>{" "}
             and an{" "}
             <Link
@@ -374,9 +374,9 @@ export const InlineLight: StoryComponentType = {
                 light={true}
                 target="_blank"
             >
-                external Visitable link (Primary only)
+                external Visitable link
             </Link>
-            . Secondary light links are not supported.
+            .
         </Body>
     ),
     parameters: {
--- a/__docs__/wonder-blocks-link/link.stories.tsx
+++ b/__docs__/wonder-blocks-link/link.stories.tsx
@@ -6,7 +6,7 @@ import type {Meta, StoryObj} from "@storybook/react";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {PhosphorIcon} from "@khanacademy/wonder-blocks-icon";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
-import {color, semanticColor, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {semanticColor, spacing} from "@khanacademy/wonder-blocks-tokens";
 import {
     Body,
     HeadingSmall,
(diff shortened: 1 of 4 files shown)


__docs__/wonder-blocks-link/link.stories.tsx
```diff
@@ -449,12 +420,42 @@ export const Navigation: StoryComponentType = () => (
  * documentation](/story/button-navigation-callbacks--before-nav-callbacks&viewMode=docs)
  * for details.
  */
-Navigation.parameters = {
-    docs: {
-        description: {
-            story: ``,
-        },
-    },
+export const Navigation: StoryComponentType = {
```
jandrade, line 423: **note:** I messed up CSF here (from my previous PR), so I'm fixing it here.

packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -167,6 +173,63 @@ const sharedStyles = StyleSheet.create({
...
+        default: {
+            rest: {
+                foreground: action.default.foreground,
+            },
+            hover: {
+                foreground: action.hover.foreground,
+            },
+            focus: {
+                border: semanticColor.border.focus,
+                foreground: action.hover.foreground,
```
beaesguerra, line 196: Maybe this is an example where `action.focus.foreground` would be useful! If we change hover styles, would we expect focus styles to change too?

packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -167,6 +173,63 @@ const sharedStyles = StyleSheet.create({
...
+            },
+            press: {
+                foreground: action.press.foreground,
+            },
+            restVisited: {
+                foreground: color.purple,
+            },
+            pressVisited: {
+                foreground: mix(color.offBlack32, color.purple),
```
beaesguerra, line 205: For my own learning, how do we know when to use the color utils like `mix`?

packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -167,6 +173,63 @@ const sharedStyles = StyleSheet.create({
     },
 });
 
+const action = semanticColor.action.outlined.progressive;
+
+// NOTE: This color is only used here.
+const pink = "#fa50ae";
```
beaesguerra, line 179: It'll be nice when we have the full color palette so we won't have to hardcode one-off colors!

beaesguerra (approved):
Code changes look good! Left some non-blocking questions in the PR, and one comment in Chromatic I'd like to get your thoughts on!

packages/wonder-blocks-link/src/components/link-core.tsx
jandrade: TBH this shouldn't be used too much these days as most of the mixed colors are codified as primitive colors. This was originally introduced to mix colors in the past due to the limitation in our previous primitive colors, but now that we include more shades, this shouldn't be needed anymore (mostly).

PR #2466, opened 2025-02-07
[rollup-rollback] Rollback rollup upgrade

jandrade:
## Summary:

I recently upgraded to Rollup v4 when I did the pnpm upgrade. This caused the
cjs build to fail in `webapp`. I am rolling back to the previous version of
Rollup to fix the build.

The cause of the build failure is that Rollup v4 is not resolving properly
default exports in the `wb-cell` package. This is causing the build to fail
in tests as the CJS build is not able to resolve this properly.

Issue: "none"

## Test plan:

Run the Snapshot release in webapp and verify that tests no longer break.


package.json
```diff
@@ -119,7 +119,7 @@
     "prettier": "^3.4.2",
     "react-refresh": "^0.16.0",
     "remark-gfm": "^4.0.0",
-    "rollup": "^4.32.1",
+    "rollup": "^2.79.2",
     "rollup-plugin-node-externals": "^8.0.0",
```
beaesguerra, line 123: question: Do we need to restore the versions for the rollup related dependencies as well? https://github.com/Khan/wonder-blocks/pull/2453/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L119-L121

jandrade: In theory, rolling back `rollup` should be good enough. I tested this change manually and it seemed to work just fine.

beaesguerra: Thanks for confirming!

beaesguerra (approved):
Looks good to me!

marcysutton:
Yay, thanks for shipping this so fast @jandrade! It appears to have addressed the test failures that were blocking us in webapp. 💯

PR #2468, opened 2025-02-10
[WB-1814.7] Refactor Modal to use semantic colors

jandrade:
## Summary:

Next step is to refactor the `Modal` package to use semantic colors.

- Reworked the theme structure to split tokens by sub-component.
- Updated stories to use the new semantic colors.
- Adapted some components to use theming instead of hardcoded colors.

### Implementation plan:

1. #2439
2. #2440
3. #2441
4. #2446
5. #2449
6. #2464
7. Modal (current PR)
8. Popover, Tooltip
9. Dropdown
10. Clickable, Pill, Toolbar

Issue: WB-1814

## Test plan:

Verify that the Modal Chromatic snapshots are mostly unchanged.

URL: `/?path=/docs/packages-modal-onepanedialog--docs&globals=viewport:desktop`

### DIFF
--- a/packages/wonder-blocks-modal/src/components/modal-backdrop.tsx
+++ b/packages/wonder-blocks-modal/src/components/modal-backdrop.tsx
@@ -1,14 +1,20 @@
 import * as React from "react";
 import * as ReactDOM from "react-dom";
-import {StyleSheet} from "aphrodite";
-
-import {color} from "@khanacademy/wonder-blocks-tokens";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {ModalLauncherPortalAttributeName} from "../util/constants";
 
 import {findFocusableNodes} from "../util/find-focusable-nodes";
 
 import type {ModalElement} from "../util/types";
+import {
+    ThemedStylesFn,
+    withScopedTheme,
+    WithThemeProps,
+} from "@khanacademy/wonder-blocks-theming";
+import {
+    ModalDialogThemeContext,
+    ModalDialogThemeContract,
+} from "../themes/themed-modal-dialog";
 
 type Props = {
     children: ModalElement;
@@ -23,7 +29,7 @@ type Props = {
      * Test ID used for e2e testing.
      */
     testId?: string;
-};
+} & WithThemeProps;
 
 /**
  * A private component used by ModalLauncher. This is the fixed-position
@@ -35,7 +41,7 @@ type Props = {
  * and adding an `onClose` prop that will call `onCloseModal`. If an
  * `onClose` prop is already provided, the two are merged.
  */
-export default class ModalBackdrop extends React.Component<Props> {
+class ModalBackdrop extends React.Component<Props> {
     componentDidMount() {
         // eslint-disable-next-line import/no-deprecated
         const node: HTMLElement = ReactDOM.findDOMNode(this) as any;
@@ -137,7 +143,7 @@ export default class ModalBackdrop extends React.Component<Props> {
 
         return (
             <View
-                style={styles.modalPositioner}
+                style={this.props.wbThemeStyles.modalPositioner}
                 onMouseDown={this.handleMouseDown}
                 onMouseUp={this.handleMouseUp}
                 testId={testId}
@@ -149,7 +155,7 @@ export default class ModalBackdrop extends React.Component<Props> {
     }
 }
 
-const styles = StyleSheet.create({
+const themedStylesFn: ThemedStylesFn<ModalDialogThemeContract> = (theme) => ({
     modalPositioner: {
         position: "fixed",
         left: 0,
@@ -171,6 +177,11 @@ const styles = StyleSheet.create({
         //     now!
         overflow: "auto",
 
-        background: color.offBlack64,
+        background: theme.backdrop.color.background,
     },
 });
+
+export default withScopedTheme(
+    themedStylesFn,
+    ModalDialogThemeContext,
+)(ModalBackdrop);
--- a/packages/wonder-blocks-modal/src/components/modal-backdrop.tsx
+++ b/packages/wonder-blocks-modal/src/components/modal-backdrop.tsx
@@ -1,20 +1,20 @@
 import * as React from "react";
 import * as ReactDOM from "react-dom";
+
 import {View} from "@khanacademy/wonder-blocks-core";
-import {ModalLauncherPortalAttributeName} from "../util/constants";
-
-import {findFocusableNodes} from "../util/find-focusable-nodes";
-
-import type {ModalElement} from "../util/types";
 import {
     ThemedStylesFn,
     withScopedTheme,
     WithThemeProps,
 } from "@khanacademy/wonder-blocks-theming";
+
 import {
     ModalDialogThemeContext,
     ModalDialogThemeContract,
 } from "../themes/themed-modal-dialog";
+import {ModalLauncherPortalAttributeName} from "../util/constants";
+import {findFocusableNodes} from "../util/find-focusable-nodes";
+import type {ModalElement} from "../util/types";
 
 type Props = {
     children: ModalElement;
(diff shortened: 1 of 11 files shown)


packages/wonder-blocks-modal/src/components/modal-backdrop.tsx
jandrade, line 1: **note:** this file didn't support theming yet, so I used the HOC available for that.

packages/wonder-blocks-modal/src/components/modal-footer.tsx
jandrade, line 1: **note:** Also applied theming here, but using the hook in this case.

packages/wonder-blocks-modal/src/themes/default.ts
jandrade, line 1: **note:** Restructured the theme contract to split it by component, then separate tokens by semantic category:

e.g.
```
// theme.component.category.property
theme.header.color.background
theme.footer.spacing.gap
```

packages/wonder-blocks-theming/src/index.ts
```diff
@@ -5,6 +5,7 @@ export {default as useScopedTheme} from "./hooks/use-scoped-theme";
 export {
     default as withScopedTheme,
     type WithThemeProps,
+    type WithoutTheme,
```
jandrade, line 8: **note:** I had to expose this in order to use `withScopedTheme` in other WB packages, otherwise TS would complain b/c it needed that type for inferring the correct types in the HOC consumer (ModalBackdrop in this case).

packages/wonder-blocks-modal/src/components/modal-footer.tsx
beaesguerra: It's great we have the option to apply theming both ways, nice job 😄

packages/wonder-blocks-modal/src/components/modal-header.tsx
```diff
@@ -153,41 +153,40 @@ const small = "@media (max-width: 767px)";
 
 const themedStylesFn: ThemedStylesFn<ModalDialogThemeContract> = (theme) => ({
     header: {
-        boxShadow: `0px 1px 0px ${theme.color.shadow.default}`,
+        boxShadow: `0px 1px 0px ${theme.header.color.border}`,
```
beaesguerra, line 156: (for later): Do you think box shadow values should belong in the theme?

packages/wonder-blocks-modal/src/components/modal-panel.tsx
```diff
@@ -180,20 +180,20 @@ const themedStylesFn: ThemedStylesFn<ModalDialogThemeContract> = (theme) => ({
         // programmatic focus. This is a workaround to make sure the focus
         // outline is visible when this control is focused.
         ":focus": {
-            outlineWidth: theme.border.width,
-            outlineColor: theme.border.color,
+            outlineWidth: theme.root.border.width,
+            outlineColor: theme.panel.color.border,
```
beaesguerra, line 184: Would it make sense for this to be `theme.root.color.border`? Or for the outlineWidth/borderRadius to be `theme.panel.border.width` and `theme.panel.border.radius` so using root or panel styles is consistent?

packages/wonder-blocks-modal/src/themes/default.ts
```diff
@@ -1,36 +1,69 @@
...
+            width: border.width.thin,
         },
-        shadow: {
-            default: tokens.color.offBlack16,
+    },
+    /**
+     * Building blocks
+     */
+    backdrop: {
+        color: {
```
beaesguerra, line 28: For the theme component tokens with building blocks, do you think it's helpful to keep the additional separation between `color`,`spacing`, etc? Curious what you think!

packages/wonder-blocks-modal/src/themes/default.ts
```diff
@@ -1,36 +1,69 @@
...
+            titlePaddingRightMd: spacing.medium_16,
+            titlePaddingRightSm: spacing.xLarge_32,
+        },
+    },
+    panel: {
+        color: {
+            border: semanticColor.border.focus,
+        },
+        spacing: {
+            paddingBlockEnd: spacing.xLarge_32,
```
beaesguerra, line 61: For token names, what do you think about naming the tokens so that they aren't specific to how it is used in css? That way, if we were to change how we apply the spacing value in the component, we wouldn't have to rename the token!

packages/wonder-blocks-modal/src/themes/default.ts
```diff
@@ -1,36 +1,69 @@
...
-    color: {
-        bg: {
-            inverse: tokens.color.darkBlue,
+    /**
+     * Shared tokens
+     */
+    root: {
+        // TODO(WB-1852): Remove light variant.
+        color: {
+            inverse: {
```
beaesguerra, line 14: Is there a reason we don't need the background and foreground defined for the default case (non-inverse)?

beaesguerra (approved):
Looks good, left some non-blocking questions!

packages/wonder-blocks-modal/src/components/modal-header.tsx
```diff
@@ -153,41 +153,40 @@ const small = "@media (max-width: 767px)";
...
         display: "flex",
         flexDirection: "column",
         minHeight: 66,
-        padding: `${theme.spacing.header.medium}px ${theme.spacing.header.large}px`,
+        padding: `${theme.header.spacing.paddingBlockMd}px ${theme.header.spacing.paddingInlineMd}px`,
```
marcysutton, line 160: Could we put the `px` unit into a variable so it could be easily replaced later (for `rem` or `em`)?

packages/wonder-blocks-modal/src/components/modal-header.tsx
```diff
@@ -153,41 +153,40 @@ const small = "@media (max-width: 767px)";
...
+        marginBottom: theme.header.spacing.gap,
     },
 
     title: {
         // Prevent title from overlapping the close button
-        paddingRight: theme.spacing.header.small,
+        paddingRight: theme.header.spacing.titlePaddingRightMd,
         [small]: {
-            paddingRight: theme.spacing.header.large,
+            paddingRight: theme.header.spacing.titlePaddingRightSm,
```
marcysutton, line 183: The padding appears to have gone from `large` to `sm`, is that intentional?

packages/wonder-blocks-modal/src/themes/default.ts
```diff
@@ -1,36 +1,69 @@
...
+            border: semanticColor.border.primary,
+            secondary: semanticColor.text.secondary,
+        },
+        spacing: {
+            paddingBlockMd: spacing.large_24,
+            paddingInlineMd: spacing.xLarge_32,
+            paddingInlineSm: spacing.medium_16,
+            gap: spacing.xSmall_8,
+            titlePaddingRightMd: spacing.medium_16,
+            titlePaddingRightSm: spacing.xLarge_32,
```
marcysutton, line 53: Should this be `titlePaddingRightXl`?

packages/wonder-blocks-modal/src/themes/default.ts
```diff
@@ -1,36 +1,69 @@
...
+        },
+    },
+    panel: {
```
marcysutton, line 56: I like how you grouped these styles by component!

marcysutton (commented):
Left you a few additional questions! Thanks for doing all of this theming work, @jandrade!

packages/wonder-blocks-modal/src/themes/default.ts
jandrade: I think it makes sense in this case as it can be renamed to `gap` (not directly related to the css prop). In other cases, it might be trickier where tokens are directly associated to specific properties (e.g. the `header` padding).

PR #2469, opened 2025-02-10
[combobox-dropdown-docs] Update docs with labeling examples

marcysutton:
## Summary:
Primarily motivated by API changes to SingleSelect/MultiSelect as combobox, but also improving labeling violations on the Component gallery page.

Issue: https://khanacademy.atlassian.net/browse/WB-1824

## Test plan:
1. Open PR in Storybook: Component gallery, SingleSelect docs, MultiSelect docs
2. Verify there are labels on all componente examples
3. Verify docs make sense


__docs__/wonder-blocks-dropdown/single-select.argtypes.ts
```diff
@@ -16,15 +28,15 @@ const argTypes: ArgTypes = {
     labels: {
         control: {type: "object"},
         description:
-            "The object containing the custom labels used inside this component.",
+            "The object containing the custom values used inside this component.",
```
beaesguerra, line 31:
```suggestion
            "The object containing the custom label values used inside this component.",
```
Minor suggestion to make it more specific to avoid confusion since the `OptionItem` component has a `value` prop! (same for MultiSelect!)

__docs__/wonder-blocks-dropdown/single-select.argtypes.ts
```diff
@@ -16,15 +28,15 @@ const argTypes: ArgTypes = {
...
         table: {
             type: {summary: "Labels"},
         },
     },
     showOpenerLabelAsText: {
         control: {type: "boolean"},
-        description: `When false, the SelectOpener can show a Node as a label. When true, the
-         SelectOpener will use a string as a label. If using custom OptionItems, a
+        description: `When false, the SelectOpener can show a Node as a value. When true, the
+         SelectOpener will use a string as a value. If using custom OptionItems, a
```
beaesguerra, line 39: I find this description a bit confusing (it might just be me!). Do you have any ideas on how we could make it easier to understand? 

Maybe something like this?
```suggestion
        description: `When false, the SelectOpener can show a Node as the selected value. When true, the
         SelectOpener will use a string as the selected value. If using custom OptionItems, a
```

(same for MultiSelect)

marcysutton: Such a great call out. This part IS uniquely confusing since the API changed from "labels" to "values". I'll also admit this particular prop was confusing to me from the get-go. 😅 

I think `selected value` works. It might not totally fix the confusing prop name, though?

__docs__/wonder-blocks-dropdown/single-select.argtypes.ts
marcysutton: `label values` works for me! It's also confusing that an `OptionItem`'s content becomes a value in the opener...

__docs__/wonder-blocks-dropdown/single-select.argtypes.ts
beaesguerra: I think the docs will help! In the context of form fields that have labels, it is confusing 😆 I think renaming the prop will be a bigger lift since we'll want to make this prop consistent with all the components that have translations too!

__docs__/wonder-blocks-dropdown/multi-select.argtypes.ts
```diff
@@ -17,15 +29,15 @@ const argTypes: ArgTypes = {
     labels: {
         control: {type: "object"},
         description:
-            "The object containing the custom labels used inside this component.",
+            "The object containing the custom values used inside this component.",
```
beaesguerra, line 32: Same thing from the singleselect can be applied to the multiselect docs!
```suggestion
            "The object containing the custom label values used inside this component.",
```

__docs__/wonder-blocks-dropdown/multi-select.argtypes.ts
```diff
@@ -17,15 +29,15 @@ const argTypes: ArgTypes = {
...
         table: {
             type: {summary: "Labels"},
         },
     },
     showOpenerLabelAsText: {
         control: {type: "boolean"},
-        description: `When false, the SelectOpener can show a Node as a label. When true, the
-         SelectOpener will use a string as a label. If using custom OptionItems, a
+        description: `When false, the SelectOpener can show a Node as a value. When true, the
+         SelectOpener will use a string as a value. If using custom OptionItems, a
```
beaesguerra, line 40: Same thing from the singleselect docs!
```suggestion
        description: `When false, the SelectOpener can show a Node as a selected value. When true, the
         SelectOpener will use a string as a selected value. If using custom OptionItems, a
```

beaesguerra (approved):
I noticed the MultiSelect docs can be updated with the same recent changes in the SingleSelect docs! Other than that, looks good to go 😄  thanks for updating the docs!

PR #2472, opened 2025-02-12
[WB-1814.9] Refactor Clickable and Pill to use semantic colors

jandrade:
## Summary:

- Migrate Clickable and Pill to use semantic colors
- Add `All Variants` story to Pill

### Implementation plan:

1. #2439
2. #2440
3. #2441
4. #2446
5. #2449
6. #2464
7. #2468
8. #2470
9. Clickable, Pill (current PR)
10. Dropdown

Issue: WB-1814

## Test plan:

Navigate to the `Clickable` and `Pill` stories in Storybook and verify that the
colors are correct.

PR #2470, opened 2025-02-11
[WB-1814.8] Refactor Popover, Tooltip, Toolbar to use semantic colors

jandrade:
## Summary:

Next step is to refactor the `Popover`, `Tooltip` and `Toolbar` packages to use
semantic colors.

### Implementation plan:

1. #2439
2. #2440
3. #2441
4. #2446
5. #2449
6. #2464
7. #2468
8. Popover, Tooltip, Toolbar (current PR)
9. Dropdown
10. Clickable, Pill

Issue: WB-1814

## Test plan:

Verify that the Modal Chromatic snapshots are mostly unchanged.

PR #2474, opened 2025-02-17
[WB-1814.10] Refactor wonder-blocks-dropdown to use semantic colors

jandrade:
## Summary:

Last PR that migrates the remaining components to use semantic colors.

- Added `All Variants` stories to `ActionItem` and `OptionItem`.
- Refactored `ActionItem` and `OptionItem` to use semantic colors.
- Also refactored `ActionMenu`, `SingleSelect`, `MultiSelect` and `Combobox` to
use semantic colors.
- Simplified the styles in `ActionMenuOpenerCore` to use a single static
StyleSheet instance (instead of generating another one in runtime).

### Implementation plan:

1. #2439
2. #2440
3. #2441
4. #2446
5. #2449
6. #2464
7. #2468
8. #2470
9. #2472
10. Dropdown (current PR)

Issue: WB-1814

## Test plan:

Verify that all the dropdown stories are still working as expected.


packages/wonder-blocks-dropdown/src/components/action-item.tsx
```diff
@@ -189,30 +198,20 @@ const styles = StyleSheet.create({
...
-        ["@media not (hover: hover)" as any]: {
-            // Revert the hover styles to the default/resting state (mobile
-            // only).
-            [":hover[aria-disabled=false]" as any]: {
-                color: white,
-                background: offBlack,
-            },
+            color: theme.hover.foreground,
+            background: theme.hover.background,
         },
```
jandrade, line 209: **note:** I removed this as it is no longer needed. I checked other popular design systems in the industry and all of them preserve the hover state on mobile devices, so we just let the browsers decide how to apply these styles from now on.

packages/wonder-blocks-dropdown/src/components/action-menu-opener-core.tsx
```diff
@@ -139,53 +141,19 @@ const sharedStyles = StyleSheet.create({
...
-    spinner: {
-        position: "absolute",
+    press: {
+        color: semanticColor.action.outlined.progressive.press.foreground,
     },
 });
-
-const styles: Record<string, any> = {};
-
-const _generateStyles = (localColor: string) => {
```
jandrade, line 152: **note:** This was overhead, so I simplified these styles by moving them to the `StyleSheet` definition above.

__docs__/wonder-blocks-dropdown/option-item-variants.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+    {
+        name: "Default",
+        props: {},
+    },
+    {
+        name: "Disabled",
+        props: {disabled: true},
+    },
+    {
+        name: "Custon label",
```
beaesguerra, line 28: nit: typo!
```suggestion
        name: "Custom label",
```

__docs__/wonder-blocks-dropdown/option-item-variants.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+        testId: "",
+        horizontalRule: "none",
+        leftAccessory: null,
+        rightAccessory: null,
+    },
+    decorators: [
+        (Story): React.ReactElement<React.ComponentProps<typeof View>> => (
+            <View
+                style={{
+                    background: semanticColor.surface.secondary,
```
beaesguerra, line 71: What do you think about using the Storybook background control to set the background instead? That way, if we need to, we can change the background color when testing to see how it looks with other backgrounds! https://storybook.js.org/docs/essentials/backgrounds

packages/wonder-blocks-dropdown/src/components/action-item.tsx
```diff
@@ -175,6 +173,17 @@ export default class ActionItem extends React.Component<ActionProps> {
...
 
+const theme = {
+    hover: {
+        background: semanticColor.action.filled.progressive.hover.background,
+        foreground: semanticColor.action.filled.progressive.hover.foreground,
+    },
+    press: {
+        background: semanticColor.action.filled.progressive.press.background,
+        foreground: semanticColor.action.filled.progressive.press.foreground,
+    },
```
beaesguerra, line 184: I would have thought that this would use the `outlined` tokens since in it's rest state, it looks more like the `outlined` styles than `filled` (it is still a bit different though)! How do we decide if we should use the `filled` or `outlined` styles?

packages/wonder-blocks-dropdown/src/components/action-menu-opener-core.tsx
```diff
@@ -122,8 +119,13 @@ const sharedStyles = StyleSheet.create({
             WebkitTapHighlightColor: "rgba(0,0,0,0)",
         },
     },
+    default: {
+        background: "none",
```
beaesguerra, line 123: Should a background of `none` be tokenized?

packages/wonder-blocks-dropdown/src/components/action-menu-opener-core.tsx
```diff
@@ -122,8 +119,13 @@ const sharedStyles = StyleSheet.create({
...
+        color: semanticColor.action.outlined.progressive.default.foreground,
+    },
     disabled: {
-        cursor: "auto",
+        color: semanticColor.action.disabled.default,
+        cursor: "default",
```
beaesguerra, line 128: Do we use the `not-allowed` cursor on `disabled` states?

packages/wonder-blocks-dropdown/src/components/checkbox.tsx
```diff
@@ -69,16 +67,16 @@ const styles = StyleSheet.create({
         borderRadius: 3,
         borderWidth: 1,
         borderStyle: "solid",
-        borderColor: offBlack50,
+        borderColor: semanticColor.border.strong,
```
beaesguerra, line 70: Out of curiosity, should the Checkbox in the dropdown always have the same style as the Checkbox component from the form package? Would it make sense to share styles between the 2? Or let them each evolve on their own? 

If they should always be the same styles, another alternative could be to use the Checkbox component from the form package directly!

packages/wonder-blocks-dropdown/src/components/combobox.tsx
```diff
@@ -695,24 +695,24 @@ const styles = StyleSheet.create({
...
+        background: semanticColor.surface.primary,
+        border: `1px solid ${semanticColor.border.focus}`,
     },
     disabled: {
-        background: color.offWhite,
-        border: `1px solid ${color.offBlack16}`,
-        color: color.offBlack64,
+        background: semanticColor.action.disabled.secondary,
+        border: `1px solid ${semanticColor.border.primary}`,
+        color: semanticColor.text.secondary,
```
beaesguerra, line 710: Do you think it would be helpful to follow the background/foreground/outline structure for the primary and secondary disabled action tokens? So that the 3 tokens are always used together!

packages/wonder-blocks-dropdown/src/components/option-item.tsx
```diff
@@ -301,16 +301,65 @@ export default class OptionItem extends React.Component<OptionProps> {
...
 
+const theme = {
+    root: {
+        default: {
+            background: semanticColor.surface.primary,
+            foreground: semanticColor.text.primary,
+        },
+        hover: {
+            background:
+                semanticColor.action.filled.progressive.hover.background,
```
beaesguerra, line 320: Same question about filled vs outlined tokens!

packages/wonder-blocks-dropdown/src/components/option-item.tsx
```diff
@@ -301,16 +301,65 @@ export default class OptionItem extends React.Component<OptionProps> {
...
+                semanticColor.action.filled.progressive.hover.foreground,
+        },
+        press: {
+            background:
+                semanticColor.action.filled.progressive.press.background,
+            foreground:
+                semanticColor.action.filled.progressive.press.foreground,
+        },
+        disabled: {
+            background: "transparent",
```
beaesguerra, line 331: Should we create a color token for `transparent`?

packages/wonder-blocks-dropdown/src/components/option-item.tsx
```diff
@@ -301,16 +301,65 @@ export default class OptionItem extends React.Component<OptionProps> {
...
+            foreground: semanticColor.text.disabled,
+        },
+    },
+    checkbox: {
```
beaesguerra, line 335: Same question about if checkbox styles need to stay in sync with the Checkbox component from the form package!

beaesguerra (approved):
Yay last one! Thanks for all of your work on updating our components to use semantic color tokens 😄 I left some non-blocking questions in the PR!

packages/wonder-blocks-dropdown/src/components/combobox.tsx
jandrade: Could you please elaborate on this? would it be to add `disabled` in outlined and filled?

FYI I moved these tokens for now to a `theme` object that I'll later move to a dropdown theme file.

beaesguerra: Sure! For other action colors, we have (for example)`semanticColor.action.filled.progressive.default` which has
- border
- background
- foreground

We define the 3 properties for other states and variants. I'm wondering if it would be helpful to have border, background, and foreground for the disabled actions too.

```
// Before (shows part of the semantic color token structure)
const action = {
  disabled: {
    default: color.fadedOffBlack32,
    secondary: color.offWhite,
  }
}

// After (default and secondary are expanded)
const action = {
  disabled: {
    default: {
      background: color.fadedOffBlack32,
      foreground: semanticColor.text.inverse,
      border: semanticColor.border.primary,
    },
    secondary: {
      background: color.offWhite,
      foreground: semanticColor.text.secondary,
      border: semanticColor.border.primary,
    },
}
```

Then applying the disabled action tokens would look like: 
```
const styles = {
  disabled: {
        background: semanticColor.action.disabled.secondary.background,
        border: `1px solid ${semanticColor.action.disabled.secondary.border}`,
        color: semanticColor.action.disabled.secondary.foreground,
  }
}
```

Grouping the foreground, background and border together at the semantic tokens level could help make sure the correct values are used together so that they have enough color contrast (and for consistency)!

jandrade: Ohh I see, thanks! yeah that makes a lot of sense :)

In fact, Design is taking a similar approach. I'm going to discuss with Caitlyn how we are going to align with these tokens and add this change in a separate PR. I'll add a TODO comment for now.

Screenshot 2025-02-20 at 1 45 55 PM

jandrade: Here's the ticket: https://khanacademy.atlassian.net/browse/WB-1889

beaesguerra (approved):
Looks good to me! Thanks for adding the TODOs :)

PR #2477, opened 2025-02-20
Enable View to accept all tag names

nedredmond:
Enables View to accept all React-compatible tag names.

This will allow Views to accept `main` and `header`, for instance.

Tests added.


beaesguerra (approved):
Looks great, thanks for making the View component more flexible!

jandrade (approved):
Looks great! thanks for simplifying the code! and for applying this change 🚀

PR #2480, opened 2025-02-24
[WB-1856] Focus: Define general pattern for focus indicators in all components

jandrade:
## Summary:

- Tokens: Renamed `border.focus` to `focus.outer` and added a new `focus.inner` token.
- Switch all components to use a new `focus.outer` semanticColor token.
- Consolidated all the interactive elements to use the same global semantic focus token. This means that `error` and `disabled` states now use the blue outline when focused.

Issue: WB-1856

## Test plan:

Verify that all components have the correct focus indicator color.

### DIFF
--- a/.changeset/chilled-otters-hang.md
+++ b/.changeset/chilled-otters-hang.md
@@ -2,4 +2,4 @@
 "@khanacademy/wonder-blocks-tokens": major
 ---
 
-Rename `semanticColor.border.focus` to `semanticColor.focus.outer`
+Rename `semanticColor.border.focus` to `semanticColor.focus.outer`, and add `focus.inner`.
--- a/.changeset/chilled-otters-hang.md
+++ b/.changeset/chilled-otters-hang.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": major
+---
+
+Rename `semanticColor.border.focus` to `semanticColor.focus.outer`
--- a/packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
+++ b/packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
@@ -4,7 +4,6 @@ const border = {
     primary: color.fadedOffBlack16,
     subtle: color.fadedOffBlack8,
     strong: color.fadedOffBlack50,
-    focus: color.blue,
     inverse: color.white,
 };
 
@@ -148,6 +147,11 @@ export const semanticColor = {
      * -Strong for when 3:1 contrast is a priority (ex. form elements)
      */
     border: border,
+
+    focus: {
+        outer: color.blue,
+        inner: color.white,
+    },
     /**
      * Default icon colors that change in context (like actions).
      */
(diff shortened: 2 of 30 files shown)


.changeset/chilled-otters-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": major
+---
+
+Rename `semanticColor.border.focus` to `semanticColor.focus.outer`
```
beaesguerra, line 5: `focus.inner` was added as well!

packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
```diff
@@ -148,6 +147,11 @@ export const semanticColor = {
      * -Strong for when 3:1 contrast is a priority (ex. form elements)
      */
     border: border,
+
+    focus: {
+        outer: color.blue,
+        inner: color.white,
```
beaesguerra, line 153: I have a few questions around naming!
- What is `focus.inner` used for?
- Is there a reason why outer/inner naming is used instead of primary and inverse like the other tokens? 
- Would it be helpful to follow the background/foreground/border pattern for the focus colors too? The structure would be consistent with the other tokens and if we need to change global focus styling to include a background or foreground color, we could add that in (we wouldn't need to define background and foreground now)! This could be a later problem if it comes to it, but thought I'd ask!

beaesguerra (commented):
Thanks for making the focus indicators more consistent 😄 Left some non-blocking questions in the PR, and some comments in Chromatic that I wanted to confirm with you around hover states!

packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
jandrade: 1/2. I think `inner` is mostly useful for Design purposes. I know that Caitlyn is using both to define the outline ring using `boxShadow`, so `inner` is the white color immediate to the element's border, then `outer` is the `blue` color that appears outside of the white bg.

In code we are taking another approach thanks to `outlineOffset` as we can use `outline` + `outlineOffset` to apply the same behavior Figma is trying to do with `boxShadow`.

It's not going to be used for now, but left it at the moment for consistency reasons with the Design tokens in Figma. I might delete it later if we don't need it.

3. I don't think background and foreground apply here as these tokens are strictly used for the focus indicator in all interactive components. Now that I see, I forgot to add this new category in the docs, I'll add it.

jandrade: After chatting offline, We are planning on switching from `outline` to `boxShadow`, specially for supporting better light variants (dark mode in the future).... this will happen later in a separate PR. (cc @beaesguerra in case I forgot to add something).

beaesguerra: We'll also keep the `focus.outer` and `focus.inner` tokens and we'll create a styles package that will use the semantic tokens and have all the focus styling together!

beaesguerra (approved):
Looks great! 🎉

PR #2482, opened 2025-02-27
[WB-1844] Button: Create separate styles for hover and focus

jandrade:
## Summary:

This PR aligns with the new Design aproach in Figma, were we want to define
separate styles for focus and hover states. From now on, the focus state will be
a global style, and the hover state will be a local style (replicating what we
currently have).

Figma: https://www.figma.com/design/VbVu3h2BpBhH80niq101MHHE/%F0%9F%92%A0-Main-Components?node-id=19724-1290&t=7yCkqQqjqnJeqCVc-4

Issue: WB-1844

## Test plan:

Navigate to the Button docs and check the button styles for hover and focus
states look as expected.

- hover: the button should have a different outline color based on the color
(destructive vs default).

- focus: the button should always have a blue outline (destructive, default,
disabled).

### DIFF
--- a/.changeset/smooth-kiwis-reflect.md
+++ b/.changeset/smooth-kiwis-reflect.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-button": patch
+---
+
+Split hover and focus into separate styles. Use default outline ring color for all variants (blue)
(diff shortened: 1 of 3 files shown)


jaredly (approved):
looks reasonable

beaesguerra (approved):
Looks good! 🚀

PR #2483, opened 2025-02-28
[🔥AUDIT🔥] Tokens: Rename action tokens: `filled` -> `primary`, `outlined` -> `secondary`.

jandrade:
🖍 _This is an audit!_ 🖍

## Summary:

This is so we can have a more consistent naming convention for our action tokens
(also defined in Figma). This change is backwards compatible, as we are just
renaming the tokens.

Issue: "none"

## Test plan:

Navigate to the `foundations > Using color` section in the docs and verify that
the action tokens are now named `primary` and `secondary` instead of `filled`
and `outlined`.

### DIFF
--- a/.changeset/gorgeous-wolves-vanish.md
+++ b/.changeset/gorgeous-wolves-vanish.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": major
+---
+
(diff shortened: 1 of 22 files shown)


beaesguerra (approved):
Looks good!

PR #2485, opened 2025-03-03
WB-1889: Rework action.disabled tokens

jandrade:
## Summary:

- Modified the `semanticColor` structure to include `disabled` tokens on
every `action` variant (`primary` and `secondary`).
- Updated the codebase to to make use of the new `disabled` tokens.

Issue: WB-1889

## Test plan:

Verify that the disabled state of the components is consistent with the
design system.

### DIFF
--- a/.changeset/rare-balloons-beam.md
+++ b/.changeset/rare-balloons-beam.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": major
+---
+
+Remove `action.disabled` object in favor of `action.primary.disabled` and `action.secondary.disabled`
--- a/packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
+++ b/packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
@@ -48,6 +48,12 @@ export const semanticColor = {
                     foreground: color.white,
                 },
             },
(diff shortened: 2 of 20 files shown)


packages/wonder-blocks-tokens/src/tokens/semantic-color.ts
```diff
@@ -85,10 +91,12 @@ export const semanticColor = {
                     foreground: color.activeRed,
                 },
             },
-        },
-        disabled: {
-            default: color.fadedOffBlack32,
-            secondary: color.offWhite,
```
beaesguerra, line 91: Do we need to mark the changeset as a `major` change since we changed the structure for the semantic tokens?

beaesguerra (commented):
Changes look good! Thanks for making the structure more consistent 😄 Left a question around versioning!

.changeset/rare-balloons-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": major
```
jandrade, line 2: @beaesguerra I changed the versioning to `major` as you suggested.

beaesguerra (approved):
Looks good! 🚀

PR #2486, opened 2025-03-03
Update monthYearOnly to use last day of month

dmcalpin:
## Summary:
BirthdayPicker currently use the first day of the month for the date if `monthYearOnly` prop is set to `true`. But for most use-cases developers want to use the last day of the month so that a user who is still 17, for example, is not considered 18, if their birthday is anything after the 1st of the month.

This update modifies the behavior of the component to pass the last day of the month to `onChange` if `monthYearOnly` is set to `true`.

Issue: PG-5959

## Test plan:
- Modified unit tests to prove last day of month is returned when the prop is set

### DIFF
--- a/packages/wonder-blocks-birthday-picker/src/components/birthday-picker.tsx
+++ b/packages/wonder-blocks-birthday-picker/src/components/birthday-picker.tsx
@@ -46,9 +46,9 @@ type Props = {
     /**
      * Whether we want to hide the day field.
      *
-     * **NOTE:** We will set the day to the _first_ day of the _selected_ month
+     * **NOTE:** We will set the day to the _last_ day of the _selected_ month
      * if the day field is hidden. Please make sure to modify the passed date
-     * value to fit different needs (e.g. if you want to set the _first_ day of
+     * value to fit different needs (e.g. if you want to set the _last_ day of
      * the _following_ month instead).
      */
     monthYearOnly?: boolean;
@@ -65,11 +65,6 @@ type Props = {
      * Additional styles applied to the dropdowns.
      */
     dropdownStyle?: StyleType;
-    /**
-     * Use the last day of the month as the day value instead of the first
-     * when `monthYearOnly` is true
-     */
-    useLastDayOfMonth?: boolean;
 };
 
 type State = {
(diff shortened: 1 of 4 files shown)


packages/wonder-blocks-birthday-picker/src/components/birthday-picker.tsx
```diff
@@ -175,7 +180,7 @@ export default class BirthdayPicker extends React.Component<Props, State> {
      * Calculates the initial state values based on the default value.
      */
     getStateFromDefault(): State {
-        const {defaultValue, monthYearOnly} = this.props;
+        const {defaultValue, monthYearOnly, useLastDayOfMonth} = this.props;
         const initialState: State = {
             month: null,
             day: monthYearOnly ? "1" : null,
```
jandrade, line 186: This could be simplified to this:

```suggestion
            day: null,
```

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.tsx
```diff
@@ -189,9 +194,13 @@ export default class BirthdayPicker extends React.Component<Props, State> {
         // If a default value was provided then we use moment to convert it
         // into a date that we can use to populate the
         if (defaultValue) {
-            const date = moment(defaultValue);
+            let date = moment(defaultValue);
 
             if (date.isValid()) {
+                if (monthYearOnly && useLastDayOfMonth) {
+                    date = date.endOf("month");
+                }
```
jandrade, line 202:
```suggestion
                if (monthYearOnly) {
                    date = date.endOf("month");
                }
```

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.tsx
```diff
@@ -65,6 +65,11 @@ type Props = {
      * Additional styles applied to the dropdowns.
      */
     dropdownStyle?: StyleType;
+    /**
+     * Use the last day of the month as the day value instead of the first
+     * when `monthYearOnly` is true
+     */
+    useLastDayOfMonth?: boolean;
```
jandrade, line 72: **suggestion:** It would be better to move the logic to `monthYearOnly` so we can avoid adding an extra prop to simplify the API here.

jandrade (commented):
The "last day" logic looks fine. I'm requesting changes to keep this new logic contained within the `monthYearOnly` prop to simplify the API.

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.tsx
dmcalpin: sounds good

beaesguerra (approved):
Changes look good! Could you update the PR title and description to reflect the updated changes since we're not using the useLastDayOfMonth prop anymore? This will make sure our git history more accurate 😄  Thank you!

PR #2487, opened 2025-03-03
[WB-1887] Add new sizing tokens for ThunderBlocks

jandrade:
## Summary:

Add sizing primitive tokens to support the new tokens with rem units that will
be used by the thunderblocks/classroom theme.

This is done to offer a more flexible way to define sizes in the components
that use these tokens by setting the size in rem units.

Also modified the Storybook styles to use 50%/8px as the base font size to
better match the rem units.

The tokens were defined here: https://khanacademy.atlassian.net/wiki/spaces/WB/pages/3683483678/REMs

Issue: https://khanacademy.atlassian.net/browse/WB-1887

## Test plan:

Navigate to the docs and see that the new sizing tokens are available.

/?path=/docs/packages-tokens-sizing--docs

Also verify that the font sizes in the storybook are preserved.

### DIFF
--- a/__docs__/wonder-blocks-tokens/tokens-sizing.mdx
+++ b/__docs__/wonder-blocks-tokens/tokens-sizing.mdx
@@ -1,0 +1,71 @@
+import {Meta, Source} from "@storybook/blocks";
+
+import TokenTable from "../components/token-table";
+
+import {View} from "@khanacademy/wonder-blocks-core";
+import * as tokens from "@khanacademy/wonder-blocks-tokens";
+
+import ComponentInfo from "../components/component-info";
+import packageConfig from "../../packages/wonder-blocks-tokens/package.json";
+
+<Meta title="Packages / Tokens / Sizing" />
+
+# Sizing
+
+<ComponentInfo name={packageConfig.name} version={packageConfig.version} />
+
+All the available sizing values that can be used for margin, padding, width,
+height, border-width, etc.
+
+## Usage
+
+You can use these sizes directly by importing `sizing` from the
+`wonder-blocks-tokens` package and accessing the named property like so:
+`sizing.size_200`.
+
+Note that these values are represented in rem units. Using rem units allows the
+sizes to scale with the font size of the root element, which is set to 50%/8px
+in our Design System. This means that the sizes will scale with the user's
+browser settings.
+
+```js
+import {sizing} from "@khanacademy/wonder-blocks-tokens";
+const styles = {padding: spacing.size_200};
+```
+
+## Tokens
+
+<TokenTable
+    columns={[
+        {
+            label: "Token",
+            cell: (row) => <code>sizing.{row.label}</code>,
+        },
+        {
+            label: "Base unit multiplier",
+            cell: (row) => row.value.replace("rem", "x"),
+        },
+        {
+            label: "Value",
+            cell: "value",
+        },
+        {
+            label: "Pixels (equivalent)",
+            cell: (row) => parseFloat(row.value.replace("rem", "") * 8),
+        },
+        {
+            label: "Example",
+            cell: (row) => (
+                <View
+                    style={{
+                        backgroundColor: tokens.color.purple,
+                        height: row.value,
+                    }}
+                >
+                    &nbsp;
+                </View>
+            ),
+        },
+    ]}
+    tokens={tokens.sizing}
+/>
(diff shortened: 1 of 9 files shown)


__docs__/wonder-blocks-tokens/tokens-sizing.mdx
```diff
@@ -0,0 +1,71 @@
...
+<Meta title="Packages / Tokens / Sizing" />
+
+# Sizing
+
+<ComponentInfo name={packageConfig.name} version={packageConfig.version} />
+
+All the available sizing values that can be used for margin, padding, width,
+height, border-width, etc.
+
+## Usage
```
beaesguerra, line 20: Would it be worth mentioning that the units are include in the token value? This is different from the current spacing tokens which are the numeric value only! 

Kinda related, do we think that the rem numeric value would be used on it's own? I was thinking maybe if consumers need to calculate something (like total width minus border and padding or something), but that should still work with units if the css `calc` function is used. Let me know if you know of other examples!

packages/wonder-blocks-tokens/src/tokens/sizing.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * The baseline for the size tokens.
+ */
+const baseline = 8;
+
+/**
+ * Converts a number (px) to a rem value.
+ */
+function convertUnitsToRem(value: number): string {
```
beaesguerra, line 9: Would this utility be helpful for consumers? Should we export it?

This reminds me of the idea @marcysutton shared around providing utilities for converting values when we were brainstorming around the sizing tokens!

__docs__/wonder-blocks-tokens/tokens-sizing.mdx
beaesguerra: Sidenote: I like that the unit is included in the token! Then we can have:
```
style: {
  padding: `${sizing.size_050} ${sizing.size_100}`,
}
```
Instead of
```
style: {
  padding: `${sizing.size_050}rem ${sizing.size_100}rem`,
}
```

beaesguerra (approved):
Looks great! Left some questions 😄

packages/wonder-blocks-tokens/src/tokens/sizing.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+/**
+ * Tokens that define the sizing of elements.
+ *
+ * These tokens are multiples of 8px, the baseline for the design system.
+ *
+ * These tokens can be used for:
+ * - Margin and padding
+ * - Component sizes
```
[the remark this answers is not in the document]
jandrade: oh Typography for sure! I'm not too sure about images as there are very open.

__docs__/wonder-blocks-tokens/tokens-sizing.mdx
jandrade: I think keeping always the `rem` unit would be preferred. In your example, `calc` uses `rem` properly when making calculations.

packages/wonder-blocks-tokens/src/tokens/sizing.ts
jandrade: I don't see why not! In that case, I think it would be better to move this to the `util/` folder, wdyt?

Side-note: maybe the function should be renamed to `convertPxToRem()` as the input unit is always px.

jandrade: hmmm I take it back. I'd like to see first how people are using these tokens. I feel that we should encourage teams to try to only use these tokens directly, and if we expose these kind of utils, we would be opening the door to use any kind of arbitrary values.

beaesguerra: Fair enough! We can add the utilities as needed 😄

__docs__/wonder-blocks-tokens/tokens-sizing.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+You can use these sizes directly by importing `sizing` from the
+`wonder-blocks-tokens` package and accessing the named property like so:
+`sizing.size_200`.
+
+```js
+import {sizing} from "@khanacademy/wonder-blocks-tokens";
+const styles = {padding: spacing.size_200};
```
beaesguerra, line 33:
```suggestion
const styles = {padding: sizing.size_200};
```
😄

beaesguerra (approved):
Looks good! Thanks for updating the docs with examples and a spacing to sizing mapping 😄 🚀

PR #2488, opened 2025-03-04
[routeradapterupdate] Rework the catch all path error

somewhatabstract:
## Summary:
This update is intended to make the error case when using path matching be more informative, and less confusing. The intention is to give a clearer signal to the developer about why things didn't work as they expected.

Made this a minor update because:
- Though it could potentially be considered a breaking change, it is only changing the error message thrown in the fail condition, which folks shouldn't be relying on anyway.
- It's a bit more than a patch fix, imho.

Issue: WB-XXXX

## Test plan:
`pnpm jest packages/wonder-blocks-testing-core`

### DIFF
--- a/packages/wonder-blocks-testing-core/src/harness/adapters/router.tsx
+++ b/packages/wonder-blocks-testing-core/src/harness/adapters/router.tsx
@@ -127,10 +127,10 @@ const MaybeWithRoute = ({
                 path="*"
                 render={() => {
                     throw new Error(
-                        `The current location ${actualLocation.pathname} ` +
-                            `does not match the configured path ${path}. ` +
+                        `The current location '${actualLocation.pathname}' ` +
+                            `does not match the configured path '${path}'. ` +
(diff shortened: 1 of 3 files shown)


jeresig (approved):
Thank you Jeff, appreciate the change!

packages/wonder-blocks-testing-core/src/harness/adapters/router.tsx
```diff
@@ -13,90 +19,101 @@ type MemoryRouterProps = JSX.LibraryManagedAttributes<
...
-                     *
-                     * If the path does not match the location, then the route
-                     * will not render the component.
-                     */
-                    path?: string;
-                }
-          >
-        | string;
+type Config =
+    | Readonly<
```
somewhatabstract, line 23: **note:** All this reformatting is because I moved the comment to where it was always supposed to be. I think some automated tooling had detached it from it's properly location.

PR #2490, opened 2025-03-04
Revert recent BirthdayPicker picker changes

jandrade:
## Summary:

Reverts this PR: #2486

The reason for reverting is to keep the current behavior of the BirthdayPicker
component, which is to default to the first day of the month when no day is
selected. This is to release other WB packages separately, and let this change
be released separately.

Issue: "none"

## Test plan:

N/A


somewhatabstract:
:thumbsup:

PR #2491, opened 2025-03-05
[dmcalpin/birthday-picker-update] Update monthYearOnly to use last day of month

dmcalpin:
## Summary:
BirthdayPicker currently use the first day of the month for the date if monthYearOnly prop is set to true. But for most use-cases developers want to use the last day of the month so that a user who is still 17, for example, is not considered 18, if their birthday is anything after the 1st of the month.

This update modifies the behavior of the component to pass the last day of the month to onChange if monthYearOnly is set to true.

Issue: PG-5959

## Test plan:
- Modified unit tests to prove last day of month is returned when the prop is set

### DIFF
--- a/.changeset/unlucky-ladybugs-talk.md
+++ b/.changeset/unlucky-ladybugs-talk.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-birthday-picker": minor
+---
+
+Updated BirthdayPicker to use last day of month when monthYearOnly is true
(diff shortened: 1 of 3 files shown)


beaesguerra (approved):
Looks good! Thanks for re-creating this PR! 

Once you merge in this PR, a `Versions PR` will get created. That can be approved and you can merge that PR to release the new WB packages (it should only include your changes!)

PR #2495, opened 2025-03-06
[WB-1891]: Integrating Announcer into SingleSelect and MultiSelect

marcysutton:
## Summary:
Integrating WB Announcer into SingleSelect and MultiSelect for updating combobox values. This work is intended to work around VoiceOver/Safari bugs where the value isn't read out properly in an opener.

Issue: WB-1891

## Test plan:
1. Review SingleSelect and MultiSelect stories in Safari and VO
2. Ensure live region announcements occur first-time through
3. Ensure announcement messages are relevant to the UX (not stale)


packages/wonder-blocks-dropdown/src/components/single-select.tsx
```diff
@@ -316,6 +324,10 @@ const SingleSelect = (props: Props) => {
     });
     const hasError = error || !!errorMessage;
 
+    React.useEffect(() => {
+        initAnnouncer();
```
beaesguerra, line 328: This is great that we get to test out how the announcer is used by other things! I have a few questions 😄 
- Do all of the consumers of the announcer have to initialize the announcer?
- Would there be any issues if a consumer didn't initialize the announcer? Would there be any issues if the announcer is initialized multiple times? (left a related comment in the initAnnouncer function!)
- Could we call `initAnnouncer` when `announceMessage` is used when the announcer isn't initialized yet? I'm wondering if `initAnnouncer` should be private to the announcer package so consumers don't have to worry about setting it up! 
  - If we need to initialize the announcer even earlier, could we initialize the announcer in a central place so that we don't have to initialize it when we use it? Maybe in something like `RenderStateRoot` or something similar? I'm not sure the best place to put it! I was thinking of `RenderStateRoot` because it's high up in the render tree ([webapp example](https://github.com/Khan/webapp/blob/e654d4b3bfc878550333b79e118b2982afa4b231/services/static/javascript/shared/app-shell/app-shell.tsx#L93-L100))!

packages/wonder-blocks-announcer/src/init-announcer.ts
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Utility to inject Announcer on page load.
+ * It can be called from useEffect or elsewhere to improve ARIA Live Region performance on the first announcement.
+ * @returns {Announcer} The Announcer instance created.
+ */
+export function initAnnouncer(props?: InitAnnouncerProps): Announcer {
+    const announcer = Announcer.getInstance();
+    if (props?.debounceThreshold !== undefined) {
+        announcer.updateWaitThreshold(props?.debounceThreshold);
```
beaesguerra, line 15: Since there can only be one instance of the announcer, could it cause issues if different consumers called `initAnnouncer` with different `debounceThreshold` values? The last thing to call `initAnnouncer` would override the previous value

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
```diff
@@ -1667,7 +1668,14 @@ describe("MultiSelect", () => {
...
 
     describe("a11y > Live region", () => {
+        beforeEach(() => {
+            initAnnouncer({debounceThreshold: 0});
+        });
+
         it("should announce the number of options when the listbox is open", async () => {
+            // const mockDateNow = jest.spyOn(Date, "now");
+            // mockDateNow.mockImplementation(() => 0);
+
```
beaesguerra: **nit:** Can this be removed? Same with the commented out lines with mockDateNow in the test!

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
```diff
@@ -1724,9 +1745,15 @@ describe("MultiSelect", () => {
             await userEvent.click(textbox);
             await userEvent.paste("ear");
 
+            // wait to avoid getting caught in the Announcer debounce
+            jest.advanceTimersByTime(250);
```
beaesguerra, line 1743: Is 250 a specific number that needs to be used? I see we also advance timers by 10!

I wonder if it would be helpful to add some docs around how to write unit tests for things using the announcer so other teams know how to create tests for it! Things like how to look for the announcer in the DOM, how long to advance timers, and any other internal details that would impact how it gets tested. Or maybe we could provide a testing utility to encapsulate these details? Just a thought!

beaesguerra (commented):
Exciting to see the announcer be used in places! I left some questions I had so far! I'll do some testing as well!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -547,7 +564,14 @@ const MultiSelect = (props: Props) => {
...
         const {noneSelected} = labels;
 
-        const menuContent = getMenuTextOrNode(allChildren);
+        const menuTextOrNode = getMenuTextOrNode(allChildren);
+        const [openerStringValue, openerContent] =
+            maybeExtractStringFromNode(menuTextOrNode);
+
+        if (openerStringValue) {
+            // opener value changed, so let's announce it
+            handleAnnouncement(openerStringValue);
```
beaesguerra, line 573: If I'm understanding correctly, this will announce the message on every render, even when the opener value hasn't changed (for example, every time the dropdown is opened/closed). Would it make sense to only announce it when the opener value changes? We could do that by calling the announcer in a `useEffect` with `selectedValues`, or we could call it whenever we trigger the `onChange` event prop for the MultiSelect! What do you think?

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -547,7 +564,14 @@ const MultiSelect = (props: Props) => {
         | React.ReactElement<React.ComponentProps<typeof SelectOpener>> => {
         const {noneSelected} = labels;
 
-        const menuContent = getMenuTextOrNode(allChildren);
+        const menuTextOrNode = getMenuTextOrNode(allChildren);
+        const [openerStringValue, openerContent] =
+            maybeExtractStringFromNode(menuTextOrNode);
```
beaesguerra, line 569: To simplify things, could we use the `getLabel` utility directly to get the announcement message? It looks like we updated `getMenuTextOrNode` and added `maybeExtractStringFromNode` to get the `openerStringValue`, which looks to always be the return value from `getLabel`! 

Let me know if I'm missing something though! It always takes me longer to understand the menu opener with all the flexibility it supports 😅

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -607,6 +631,10 @@ const MultiSelect = (props: Props) => {
     const filteredItems = getMenuItems(allChildren);
     const isDisabled = numEnabledOptions === 0 || disabled;
 
+    if (open && isFilterable) {
+        handleAnnouncement(labels.someSelected(filteredItems.length));
```
beaesguerra, line 635: Would it make sense to trigger this announcement when an event occurs or when the data changes?

beaesguerra (commented):
I left more questions that applies to both single and multiselect! 😄

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
marcysutton: Yes, I noticed that too and cleaned it up! Good call!

__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
```diff
@@ -611,7 +611,7 @@ const VirtualizedMultiSelect = function (props: Props): React.ReactElement {
     const [selectedValues, setSelectedValues] = React.useState<Array<string>>(
         [],
     );
-    const [opened, setOpened] = React.useState(props.opened || false);
```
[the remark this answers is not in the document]
marcysutton: This was an artifact of testing, and I cleaned it up!

packages/wonder-blocks-announcer/src/__tests__/announce-message.test.tsx
```diff
@@ -159,7 +159,7 @@ describe("Announcer.announceMessage", () => {
         expect(message1Region).toHaveTextContent(message1);
 
         expect(setTimeout).toHaveBeenNthCalledWith(
-            1,
+            2,
```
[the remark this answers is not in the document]
marcysutton: Yes, I'll add a comment! This was due to the leading edge / trailing edge change in the debounce function. I think the first setTimeout is the initial one for Safari timing, so this configurable one in the test with specific duration/etc. comes second.

packages/wonder-blocks-dropdown/src/components/single-select.tsx
marcysutton: Good questions! All consumers don't _have_ to initialize the Announcer, as it already initializes itself when you call `announceMessage`. There were two reasons I added this `initAnnouncer` option: 
- To make the debounceThreshold (and potentially other options) configurable in tests when you aren't handling the public Announcer functions directly (like in `SingleSelect` tests, where it's called internally)
- To improve reliability on the first call in VoiceOver and Safari -- which I should definitely document if we keep it in there! It injects the live regions earlier so that screen readers can pick them up every time. 

The initAnnouncer function would have to be public in order for consumers to call it. 

Initializing it in a central place on page load also makes a lot of sense to me, assuming it doesn't have a negative impact on performance or something. I can look into initializing it in `RenderStateRoot` in my webapp integration work! It would always return the same Announcer instance no matter how it's initialized, so we could probably do both.

packages/wonder-blocks-announcer/src/init-announcer.ts
marcysutton: Yes, that's a good point. It could change on every call.. but I think that is what you would want. Maybe we could play with it more in webapp and see if that becomes an issue?

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
marcysutton: I love that idea! The `250` number is the default debounceThreshold. The 10 number came up when I was using a 0 value in tests... setTimeouts with a value of 0 weren't _quite_ immediate in the JavaScript event loop so I had to use 10! I can write a doc on testing this so the testing path is more well-trodden.

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: I debated on whether to gate this with opener value changes, and ultimately kept it this way because of the bugs in Safari and VoiceOver. It will always announce the value to counteract the clipped and stale value that is announced regardless of the visual state of it!

I also played with setting state for these values and using useEffect and got thrown into an infinite loop. This was the only way I could get it to work with the amount of re-rendering going on.

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: I wish we could... unless I'm missing something, `getLabel` returns a node sometimes that is the wrong type for a live region (which requires a string). I needed a string label consistently, so that's why I had to `maybeExtractStringFromNode`.

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: I looked into that too...this code was the biggest culprit in triggering infinite loops as it's in the main component thread. The filtered items aren't tracked with state, so I'd have to find a safe way to do that. The change event for this doesn't do any filtering on children. I could try moving it back into dropdown-core, but I hoped to keep all announcement handling centralized in `SingleSelect` and `MultiSelect`, respectively. I'm open to suggestions as this was really tricky!

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
beaesguerra: Thinking about this more, do want consumers of the announcer to mock out the announcer instead so they don't have to worry about the implementation details? Their tests could be around making sure the announcer is called with the correct message! Then the announcer's unit tests are what makes sure it's putting the content in the live region correctly. What do you think?

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
beaesguerra: `getLabel` will always return a string 😄 
https://github.com/Khan/wonder-blocks/blob/78c353586dc24c9b9aed3b34d3d2a56f9ac85d2d/packages/wonder-blocks-dropdown/src/util/helpers.ts#L62-L72 

`getSelectOpenerLabel` returns a string or node!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
beaesguerra: Ahh I see, this is what you were mentioning the other day! I'll poke around and get back to you 😄

packages/wonder-blocks-dropdown/src/components/single-select.tsx
beaesguerra: Sounds good! For context, [RenderStateRoot](https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-core/src/components/render-state-root.tsx) is a WB Core component! We have a [storybook decorator](https://github.com/Khan/wonder-blocks/blob/78c353586dc24c9b9aed3b34d3d2a56f9ac85d2d/.storybook/preview.tsx#L102-L110) to wrap our stories with the RenderStateRoot component. Looking at the RenderStateRoot implementation more, maybe we should have a different component to initialize the announcer so we can separate the concerns! Or we can call `initAnnouncer` directly when Storybook or the app first mounts!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: Oh yes, you're right! Drowning in complexity a bit here... I'll test it out! Simplification is always a good idea when we can do it.

packages/wonder-blocks-dropdown/src/components/single-select.tsx
marcysutton: Ooh, thanks for this! I actually might need it for some target node customization I want to do in Storybook. I need to inject the Announcer into a more precise place within the docs examples so it doesn't overlap story content when shown visually. It's also helping me look more closely at the lifecycle of the initAnnouncer function, as it appears to be called much later than I intended. I'll check this out!

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
marcysutton: That's a great suggestion. I'd love to make the tests easier to work with, and mocking could be a really helpful way to do that!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: Ah, I remember now... it's the complexity from `getMenuTextOrNode` in `MultiSelect` that made this difficult, as there are multiple cases for selected items in a switch statement that return a string OR a node. I wanted to leverage that logic without having to iterate through it all again!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
beaesguerra: Tinkered around a bit in SingleSelect and was able to reduce the announcement to only when the selectedValue/children changes and when the filtered items length changes! https://github.com/Khan/wonder-blocks/compare/combobox-announcer-v2...combobox-tinkering?expand=1 

(Note: I duplicated some logic to figure out the `selectedOption` within the `useEffect` to avoid additional renders. We could memoize that logic since other things use that logic too!)

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
beaesguerra: Ahh gotcha, that makes sense!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: Thanks so much for this Bea!! I appreciate you taking the time to tinker with it. I'm testing it out in Safari and VO and it is working so far... I'll keep playing with it and a MultiSelect version.

__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
```diff
@@ -44,3 +44,18 @@ receives focus. This can be useful when the options contain icons or other infor
 that would need to be omitted from the visible label.
 
 <Canvas of={MultiSelectAccessibilityStories.UsingOpenerAriaLabel} />
+
+## Automatic screen reader announcements in `MultiSelect`
```
beaesguerra, line 48: Would it be helpful to document the specific Safari/VoiceOver issues and the related Safari bug tickets? So we know that's the existing behaviour and it is to be expected! (Same comment for SingleSelect)

__docs__/wonder-blocks-dropdown/multi-select.accessibility.stories.tsx
```diff
@@ -129,3 +131,42 @@ export const UsingCustomOpenerAriaLabel = {
...
+
+const MultiSelectWithVisibleAnnouncer = () => {
+    // React.useEffect(() => {
+    //     // Inject Announcer into the Storybook iframe instead of body
+    //     const storybookRoot = document.getElementById("storybook-docs");
+    //     if (storybookRoot) {
+    //         console.log("announcer init");
+    //         // initAnnouncer({targetNode: storybookRoot});
+    //     }
+    // }, []);
```
beaesguerra, line 147: How did this go? Did using the Storybook decorator at the root help with the issues you were seeing?

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
```diff
@@ -1683,12 +1683,7 @@ describe("MultiSelect", () => {
             };
 
             const {userEvent} = doRender(
-                <MultiSelect
-                    onChange={jest.fn()}
-                    isFilterable={true}
```
beaesguerra, line 1684: Does including `isFilterable` result in a different behaviour? Would a test covering both cases be helpful?

packages/wonder-blocks-dropdown/src/components/single-select.tsx
```diff
@@ -460,23 +476,20 @@ const SingleSelect = (props: Props) => {
...
+            const menuStringOrNode = getSelectOpenerLabel(
                 showOpenerLabelAsText,
                 selectedItem.props,
             );
-            // For Custom Option Items with Node Labels, we have to extract
-            // strings to announce
-            const [label, node] = maybeExtractStringFromNode(menuOpenerLabel);
+            // We only need the guaranteed node for SingleSelect here
+            // As the string label for the Announcer is in a useEffect above
+            const [, node] = maybeExtractStringFromNode(menuStringOrNode);
```
beaesguerra, line 488: Since we don't need the `label` from `maybeExtractStringFromNode` here anymore, could we go back to using the return value from `getSelectOpenerLabel` for the `menuContent`?

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
beaesguerra: Did moving the announcement to a `useEffect` similar to the SingleSelect have the expected behaviour?

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -607,6 +631,12 @@ const MultiSelect = (props: Props) => {
     const filteredItems = getMenuItems(allChildren);
     const isDisabled = numEnabledOptions === 0 || disabled;
 
+    React.useEffect(() => {
+        if (open) {
```
beaesguerra, line 635: The equivalent logic in SingleSelect doesn't check for `open`, do we want to include it there too? If I'm understanding correctly, this would make sure the message isn't announced when the length of filteredItems changes while the MultiSelect is closed!

beaesguerra (commented):
Thanks for iterating on this Marcy! Left some questions 😄 

I'm also curious if @jandrade has any thoughts on the announcer integration since he was going to integrate it with the Combobox component!

packages/wonder-blocks-dropdown/src/components/single-select.tsx
marcysutton: I'll have to play with the types more, or create a separate helper function from `getSelectOpenerLabel`. `SingleSelect` and `MultiSelect` use the same helper functions but this change would require a different type signature. If I use that value from `getSelectOpenerLabel` for `menuContent` in `SingleSelect` as written, it has a type failure:
```
Type 'OpenerStringOrNode' is not assignable to type 'OptionLabel'.
      Type '{ [key: string]: string | Element; }' is not assignable to type 'OptionLabel'.
```

So I'll probably opt for a separate helper function to limit the complexity. I'm not entirely sure why MultiSelect is so much more complicated in this precise way.

__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
marcysutton: I added some notes!

packages/wonder-blocks-dropdown/src/components/single-select.tsx
marcysutton: I got it simplified quite a bit and pushed an update!

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
marcysutton: It should have the same behavior for this scenario! There isn't any logic for isFilterable around announcing the number of options when the listbox is open.

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: I believe so! The last I checked it was working great. I'm still working on initializing the Announcer so I'll do another round of screen reader testing then!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: Done!

packages/wonder-blocks-dropdown/src/components/single-select.tsx
marcysutton: This code has been updated!

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
marcysutton: The Dropdown tests now use a mocked Announcer, so the internal details aren't important anymore. I can write some test how-tos in Confluence...that seemed like the best place for it!

__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
```diff
@@ -84,6 +84,15 @@ export default {
             default: "offWhite",
         },
     },
+    decorators: [
+        (Story) => {
+            React.useEffect(() => {
```
beaesguerra, line 89: Would it be helpful to initialize the announcer for all stories? Or on a per component level? If we want to have it for all stories, we could configure a global decorator.

If we'd like to keep it on a per component level, should we initalize the announcer for the BirthdayPicker too since it uses SingleSelect?

__docs__/wonder-blocks-dropdown/multi-select.accessibility.mdx
```diff
@@ -51,6 +51,18 @@ that would need to be omitted from the visible label.
...
 
+This integration works around 2 bugs in VoiceOver and Safari on Mac OSX 14 and 15
+where the combobox opener value is cut off and cached incorrectly. The value is
+buggy when announced, differing from its current visual presentation and DOM content.
+
+Bugs filed in WebKit include:
+
+1. AX: combobox button value text clipped https://bugs.webkit.org/show_bug.cgi?id=285047
+2. AX: VoiceOver does not perceive changes to combobox value in an opener
+https://bugs.webkit.org/show_bug.cgi?id=286828
```
beaesguerra, line 62: Thanks for including a summary of the buggy behaviour!

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
```diff
@@ -1668,8 +1671,17 @@ describe("MultiSelect", () => {
...
     describe("a11y > Live region", () => {
-        beforeEach(() => {
-            initAnnouncer({debounceThreshold: 0});
+        let announceMessageSpy: any;
+
+        beforeAll(() => {
+            announceMessageSpy = jest.spyOn(
+                require("@khanacademy/wonder-blocks-announcer"),
+                "announceMessage",
+            );
```
beaesguerra, line 1680: If I'm understanding correctly, we spy on the `announceMessage` function, which is also mocked out using `jest.mock(...)`. For my own learning, are both needed? Does one cover specific things the other doesn't?

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
```diff
@@ -1691,17 +1703,13 @@ describe("MultiSelect", () => {
...
             // Act
             await userEvent.click(opener);
 
-            const announcer = screen.getByTestId("wbAnnounce");
-            const announcementText =
-                await within(announcer).findByText("3 schools");
-
             // Assert
-            expect(announcementText).toBeInTheDocument();
+            await expect(announceMessageSpy).toHaveBeenCalledWith({
```
beaesguerra, line 1710: Yay it's so much easier to test when we mock out the announcer 🎉

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -547,7 +571,16 @@ const MultiSelect = (props: Props) => {
...
-        const menuContent = getMenuTextOrNode(allChildren);
+        const openerContent = getMenuTextOrNode(allChildren);
+        const openerStringValue = maybeGetOpenerStringValue(
+            allChildren,
+            openerContent,
+        );
+
+        if (openerStringValue) {
+            // opener value changed, so let's announce it
+            handleAnnouncement(openerStringValue);
```
beaesguerra, line 582: Did we want to move calling `handleAnnouncement` in the MultiSelect to a `useEffect` so it doesn't happen on each render? Similar to what was done in SingleSelect!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -607,6 +640,12 @@ const MultiSelect = (props: Props) => {
     const filteredItems = getMenuItems(allChildren);
     const isDisabled = numEnabledOptions === 0 || disabled;
 
+    React.useEffect(() => {
+        if (open) {
+            handleAnnouncement(someSelected(filteredItems.length));
```
beaesguerra, line 645: I'm realizing now that the announcement for when the number of filtered items changes is very similar to when the the selected values are updated!
- When filtered items change, the message is: `numSelectedValues === 1 ? "1 item" : ${numSelectedValues} items`
- When selected values change, the message is the opener text. If one item is selected, the opener text is the label of the item. If multiple are selected, the opener text is also in the format `x items`

It could be difficult to tell the difference between these 2 cases. This could be a future improvement though since we're working with the existing labels!

beaesguerra (commented):
Thanks for iterating on this! I had a few more questions - most are not blocking, though I wanted to get your input first on using a `useEffect` for announcements in `MultiSelect` too!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: The challenge is that getting this string value depends on the `openerContent`, which is calculated in the `renderOpener` function. There is a switch statement going through various cases of selected items that dictates what the opener content will be. The bulk of the complexity happens when `selectedValues.length === 1`, where the selected item can be a string or a node. That is the code that runs in this renderOpener function that I'd have to refactor again to go into `useEffect`.

```
const renderOpener = (
        allChildren: React.ReactElement<
            React.ComponentProps<typeof OptionItem>
        >[],
        isDisabled: boolean,
        dropdownId: string,
    ):
        | React.ReactElement<React.ComponentProps<typeof DropdownOpener>>
        | React.ReactElement<React.ComponentProps<typeof SelectOpener>> => {
        const {noneSelected} = labels;

        const openerContent = getMenuTextOrNode(allChildren);
        const openerStringValue = maybeGetOpenerStringValue(
            allChildren,
            openerContent, // <-- this dependency is calculated on the line above
        );

        if (openerStringValue) {
            // opener value changed, so let's announce it
            handleAnnouncement(openerStringValue);
        }
```

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: Yes, good point! This does depend on the labels, which have to be localized. It would be helpful to tackle that in a later phase if possible.

packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
marcysutton: I mocked it and spied on it so I could replace the implementation of `announceMessage`. But it turned out I didn't need to go that far for the tests to pass. I removed the mocking part and it still works!

__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
marcysutton: Good call. I moved it to a global decorator and added a `targetElement` prop so the Live Regions could be injected in a more precise place in Storybook than `document.body`!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: Alright, I pushed a change. It required a few `useCallback` and `useMemo` updates, but seems to work okay as far as I can tell!

.storybook/preview.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from "react";
 import wonderBlocksTheme from "./wonder-blocks-theme";
 import {Decorator} from "@storybook/react";
 import {semanticColor} from "@khanacademy/wonder-blocks-tokens";
+import {initAnnouncer} from "../packages/wonder-blocks-announcer";
```
marcysutton, line 5: Note: This wasn't working as a package import, I believe because `wonder-blocks-announcer` hasn't yet been published.

__docs__/wonder-blocks-dropdown/combobox.stories.tsx
```diff
@@ -341,7 +341,7 @@ export const ControlledMultilpleCombobox: Story = {
         await userEvent.keyboard("{Enter}");
 
         // Assert
-        expect(canvas.getByRole("log")).toHaveTextContent(
+        expect(canvas.getAllByRole("log")[0]).toHaveTextContent(
```
marcysutton, line 344: Note: This started to fail with the Storybook global decorator, as there were multiple `role="log"` elements at that point!

.storybook/preview.tsx
somewhatabstract: It should work even without it being published. Is it listed in the appropriate tsconfig and package.json's?

somewhatabstract: I think when I had this issue before, I just had to do a local build for things to start working right.

marcysutton: Good tip! I'd looked around but missed a `tsconfig-build.json` file...I _think_ I got it to work with an update. No more red squigglies.

__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
marcysutton: One thing to note, the global decorator means the Announcer is injected into every story. There are a lot of Chromatic changes as a result (non-visual as it is screen-reader hidden, but still)!

.storybook/preview.tsx
```diff
@@ -115,14 +120,18 @@ const withThemeSwitcher: Decorator = (
         return (
             <RenderStateRoot>
                 <ThemeSwitcherContext.Provider value={theme}>
-                    <Story />
+                    <div ref={containerRef}>
```
beaesguerra, line 123: Could you try setting `display: inline-block` on the wrapping div with the ref? I think the Chromatic snapshots were updated because the default `display: block` styling on the div makes it so the story takes up the full width (and that's why there's extra spacing on the right side in stories now, which causes the components to look smaller in the snapshots!)  

Default `block`
Screenshot 2025-03-20 at 9 58 57 AM

Trying with `inline-block`
image

.storybook/preview.tsx
```diff
@@ -115,14 +120,18 @@ const withThemeSwitcher: Decorator = (
...
+                        <Story />
+                    </div>
                 </ThemeSwitcherContext.Provider>
             </RenderStateRoot>
         );
     }
     return (
         <ThemeSwitcherContext.Provider value={theme}>
-            <Story />
+            <div ref={containerRef}>
```
beaesguerra, line 132: Idea: What do you think about making a different decorator for the announcer so that it is separated from the theme switcher? (not blocking)

.storybook/preview.tsx
marcysutton: Actually this didn't seem to work. The red squigglies are back for me even after a `pnpm install` and `pnpm build` 😭

.storybook/preview.tsx
marcysutton: Yes, good catch! I converted it to a span. Does that work?

.storybook/preview.tsx
marcysutton: Good call. I pushed a change with a separate decorator!

.storybook/preview.tsx
beaesguerra: Looks like `span` also takes up the full width! 
image

.storybook/preview.tsx
somewhatabstract: Have you tried running a local build so the built files exist?

somewhatabstract: Is WB on pnpm? It may be that `pnpm i` needs to be run, too.

.storybook/preview.tsx
```diff
@@ -97,39 +98,55 @@ const parameters = {
...
-            </RenderStateRoot>
-        );
-    }
     return (
-        <ThemeSwitcherContext.Provider value={theme}>
+        // inline-block keeps wrapper from impacting story widths
+        <span ref={containerRef} style={{display: 'inline-block'}}>
             <Story />
-        </ThemeSwitcherContext.Provider>
+        </span>
```
jandrade, line 143: I might miss a lot of context in the PR discussions.... have you tried other solutions to this to avoid changing the story widths? by looking at chromatic, it looks like this is still happening.

marcysutton: Maybe we can think through a different solution or whether we need this at all! Here's why it was added:

The Announcer injects the live region wrapper into `document.body` by default. In Storybook, that puts it outside of the DIV elements containing stories so it has an odd leaking effect. I added an API to customize the target element, using a React ref to set the targetElement scope. Is there another way we can inject the Announcer into the story canvas?

I also tried using DOM selectors to target the Storybook elements (`#storybook-root` and `#storybook-docs`) but I noticed which one is used depends on how you're viewing stories (the default Storybook UI or as a standalone story in its own window). One DIV or the other ends up hidden and empty, so it's hard to determine which one to target from `preview.tsx`.

Some stories go full width, and some don't. If we do need this `ref` solution, maybe each story could include a wrapper that sets the right display value (`block`, `inline-block`, `flex`, etc.)?

marcysutton: Actually, I tried the DOM query method again without a new ref element and it seems fine in both the Storybook UI and standalone stories. So maybe that will work after all! I pushed a change for testing.

.storybook/preview.tsx
```diff
@@ -127,9 +116,32 @@ const withThemeSwitcher: Decorator = (
...
+      }, [addBodyClass]);
+    return (
+        <Story />
+    );
+};
+
 const preview: Preview = {
     parameters,
-    decorators: [withThemeSwitcher],
+    decorators: [withThemeSwitcher, withAnnouncer],
```
beaesguerra, line 144: Just a heads up that I also added some storybook decorators that are in `main` now so there might be a merge conflict when the feature branch is updated with `main`!

beaesguerra (approved):
Looks great, thanks for iterating and optimizing things! Exciting to have our first components using the announcer 🎉

.storybook/preview.tsx
beaesguerra: Approved the Chromatic tests! There were some diffs in the changes because I had approved some of the snapshots that were wider during the last review. They're restored now to the original widths 🎉  Thanks for looking into this!

PR #2497, opened 2025-03-11
WB-1797 Fix ActionItem height

beaesguerra:
## Summary:
Addressing the styling issue where ActionItem contents can get cropped with the new Cell changes (found in https://github.com/Khan/webapp/pull/29021#pullrequestreview-2675406111). This was happening because we set [`overflow: hidden` now in CellCore](https://github.com/Khan/wonder-blocks/pull/2435/files#diff-0b4c2beea72229dd5e2a29d5053ba87703ec2c51882604d29f36995815a299d5R248-R250) to prevent the left bar indicator from overflowing when the border radius is set on a cell. 

Remove explicitly setting the height of action items so the contents don't overflow. We can inspect the current implementation and see the cell contents are overflowing the inner wrapper container

| Before | After |
| --- | --- |
| image | image |

Issue: WB-1797

## Test plan:
- Review Chromatic diffs
- Review Custom Action Item Multi Line story (`?path=/story/packages-dropdown-actionitem--custom-action-item-multi-line`)
- Check ActionItem use in webapp

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/action-item.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/action-item.tsx
@@ -223,7 +223,6 @@ const styles = StyleSheet.create({
     },
     shared: {
         minHeight: DROPDOWN_ITEM_HEIGHT,
-        height: DROPDOWN_ITEM_HEIGHT,
     },
 
     label: {
--- a/packages/wonder-blocks-dropdown/src/components/action-item.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/action-item.tsx
@@ -2,7 +2,11 @@ import * as React from "react";
 import {StyleSheet} from "aphrodite";
 
 import {CompactCell} from "@khanacademy/wonder-blocks-cell";
-import {semanticColor, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {
+    semanticColor,
+    sizing,
+    spacing,
+} from "@khanacademy/wonder-blocks-tokens";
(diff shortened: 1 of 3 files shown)


beaesguerra:
I'll do some testing with the npm snapshot in webapp!

packages/wonder-blocks-dropdown/src/components/action-item.tsx
```diff
@@ -223,7 +223,6 @@ const styles = StyleSheet.create({
     },
     shared: {
         minHeight: DROPDOWN_ITEM_HEIGHT,
-        height: DROPDOWN_ITEM_HEIGHT,
```
jandrade, line 226: Now I remember why I added this..... according to the Design spec, single-line items are expected to have a `40px` height, and in this case these would be `44px`. I would recommend looking at `OptionItem` to see how this is approached there (I noticed that those items remain at 40px on single line).

beaesguerra: Good catch! OptionItem uses 10px for vertical padding. Updated ActionItem to do the same using our new sizing tokens :) 
https://github.com/Khan/wonder-blocks/blob/b592ba98dc35a16e4b36d7f1af8a7fcab67d4f7d/packages/wonder-blocks-dropdown/src/components/option-item.tsx#L478-L479

jandrade (approved):
Looks great! thanks for iterating on the solution 🎉 🚢

beaesguerra:
Tested with webapp!

Custom ActionItem content no longer gets cropped 
image

ActionItems with one line of text continue to be 40px in height!
image

PR #2500, opened 2025-03-11
Set up NavigationTabs and NavigationTabItem

beaesguerra:
## Summary:
These changes set up the `wonder-blocks-tabs` package and the `NavigationTabs` and `NavigationTabItem` components.

So far, these components implement the html semantics and base styles. Will be implementing the props and addressing more styling edge cases in upcoming PRs.

Issue: WB-1898

## Test plan:
Review the docs for:
- NavigationTabs (`?path=/docs/packages-tabs-navigationtabs-navigationtabs--docs`)
- NavigationTabItem (`?path=/docs/packages-tabs-navigationtabs-navigationtabitem--docs`)

Review base styling for the components

### DIFF
--- a/__docs__/wonder-blocks-tabs/navigation-tab-item.stories.tsx
+++ b/__docs__/wonder-blocks-tabs/navigation-tab-item.stories.tsx
@@ -1,0 +1,27 @@
+import * as React from "react";
+import type {Meta, StoryObj} from "@storybook/react";
+import ComponentInfo from "../components/component-info";
+import packageConfig from "../../packages/wonder-blocks-form/package.json";
+import {NavigationTabItem} from "@khanacademy/wonder-blocks-tabs";
+
+/**
+ *
+ */
+export default {
+    title: "Packages / Tabs / NavigationTabs / NavigationTabItem",
+    component: NavigationTabItem,
+    parameters: {
+        componentSubtitle: (
+            <ComponentInfo
+                name={packageConfig.name}
+                version={packageConfig.version}
+            />
+        ),
+    },
+} as Meta<typeof NavigationTabItem>;
+
+type StoryComponentType = StoryObj<typeof NavigationTabItem>;
+
+export const Default: StoryComponentType = {
+    args: {},
+};
--- a/__docs__/wonder-blocks-tabs/navigation-tab-item.stories.tsx
+++ b/__docs__/wonder-blocks-tabs/navigation-tab-item.stories.tsx
@@ -1,7 +1,7 @@
 import * as React from "react";
 import type {Meta, StoryObj} from "@storybook/react";
 import ComponentInfo from "../components/component-info";
-import packageConfig from "../../packages/wonder-blocks-form/package.json";
+import packageConfig from "../../packages/wonder-blocks-tabs/package.json";
 import {NavigationTabItem} from "@khanacademy/wonder-blocks-tabs";
 import Link from "@khanacademy/wonder-blocks-link";
 
@@ -37,5 +37,11 @@ type StoryComponentType = StoryObj<typeof NavigationTabItem>;
 export const Default: StoryComponentType = {
     args: {
         children: <Link href="#link">Navigation tab item</Link>,
+    },
+    parameters: {
+        chromatic: {
+            // Disabling because it's covered by All Variants
+            disableSnapshot: true,
+        },
     },
 };
--- a/__docs__/wonder-blocks-tabs/navigation-tab-item.stories.tsx
+++ b/__docs__/wonder-blocks-tabs/navigation-tab-item.stories.tsx
@@ -5,10 +5,6 @@ import packageConfig from "../../packages/wonder-blocks-tabs/package.json";
 import {NavigationTabItem} from "@khanacademy/wonder-blocks-tabs";
 import Link from "@khanacademy/wonder-blocks-link";
 
-/**
- * A component for a tab item in NavigationTabs. It is used with a Link
- * component.
- */
 export default {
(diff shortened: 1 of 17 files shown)


__docs__/wonder-blocks-tabs/navigation-tab-item.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+    },
+} as Meta<typeof NavigationTabItem>;
+
+type StoryComponentType = StoryObj<typeof NavigationTabItem>;
+
+export const Default: StoryComponentType = {
+    args: {
+        children: <Link href="#link">Navigation tab item</Link>,
+    },
+};
```
jandrade, line 41: **suggestion:** It seems that this case would be already covered by `All Variants`, so you could disable the chromatic snapshot.

__docs__/wonder-blocks-tabs/navigation-tabs.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+    <NavigationTabItem>
+        <Link href="#link-3">Navigation tab item 3</Link>
+    </NavigationTabItem>,
+];
+
+export const Default: StoryComponentType = {
+    args: {
+        children: navigationTabItems,
+    },
+};
```
jandrade, line 49: **suggestion:** It seems that this case would be already covered by `All Variants`, so you could disable the chromatic snapshot.

__docs__/wonder-blocks-tabs/navigation-tab-item.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+import type {Meta, StoryObj} from "@storybook/react";
+import ComponentInfo from "../components/component-info";
+import packageConfig from "../../packages/wonder-blocks-form/package.json";
+import {NavigationTabItem} from "@khanacademy/wonder-blocks-tabs";
+import Link from "@khanacademy/wonder-blocks-link";
+
+/**
+ * A component for a tab item in NavigationTabs. It is used with a Link
+ * component.
+ */
```
jandrade, line 11: **thought:** This comment block could probably be included in the code implementation directly, so devs could see it when using intellisense in their IDEs. (I'm not sure if there's a limitation with Storybook).

Same question applies to the `NavigationTabs` comment block.

jandrade (commented):
Taking a first pass on this, this is taking great shape! Thanks for adding a11y unit tests from the start 👏

packages/wonder-blocks-tabs/src/components/navigation-tab-item.tsx
```diff
@@ -0,0 +1,59 @@
...
+        paddingInline: 0,
+        position: "relative",
+        ":hover": {
+            textDecoration: "none",
+            [":after" as any]: {
+                ...underlineStyles,
+                backgroundColor:
+                    semanticColor.action.primary.progressive.hover.border,
+            },
+        },
```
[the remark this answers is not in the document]
beaesguerra: Great question! I didn't know `textUnderlineOffset` and `textDecorationThickness` were a thing! Gave it a try and the underline is only applied under the text and not the icons unfortunately! I'll add a comment in the styles too so we know why we don't use `textDecoration` instead!

Open to suggestions though if you have ideas!

image

packages/wonder-blocks-tabs/src/components/navigation-tab-item.tsx
```diff
@@ -0,0 +1,59 @@
...
+    },
```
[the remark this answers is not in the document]
beaesguerra: Thanks for noticing this! I will iterate on the styles in another PR to address other cases and will add that then!

__docs__/wonder-blocks-tabs/navigation-tab-item.stories.tsx
beaesguerra: Good call! I added docs to the component js docs too so it shows up using intellisense (with an example 😄)! 

image

I kept this block around so it also shows in the Storybook docs page!

beaesguerra: Update: Nevermind! Storybook extracts the component docs from the jsdocs! We can let Storybook generate the description for the component :)

packages/wonder-blocks-tabs/src/components/navigation-tab-item.tsx
```diff
@@ -0,0 +1,81 @@
...
+        position: "relative",
+        ":hover": {
+            textDecoration: "none",
+            [":after" as any]: {
+                // We use :after to apply underline styles instead of textDecoration
+                // so the underline is shown under icons in Link too
+                ...underlineStyles,
+                backgroundColor:
+                    semanticColor.action.primary.progressive.hover.border,
+            },
```
jandrade, line 78: I'd like to propose another approach to get your thoughts on:

Do you think `boxShadow` could work in this case?

e.g.
```
boxShadow: inset 0 -${sizing.size_050} 0 0 ${semanticColor.action.primary.progressive.hover.border}
```

If you think that works, this could simplify the `:after` approach. Also would play nice with the idea we have for changing global focus to use `box-shadow` instead of `outline` :)

jandrade: @beaesguerra After discussing offline, we have found that this needs to be revisited in the next PR (or the PR that tackles styling). Probably `boxShadow` would still be suitable based on the latest design specs.

jandrade (approved):
Looks good to me! looking forward to see more progress on this component 🚢

PR #2501, opened 2025-03-12
[WB-1896] Add ThunderBlocks theme

jandrade:
## Summary:

Adds the new ThunderBlocks theme to support the new classroom experience.

Issue: https://khanacademy.atlassian.net/browse/WB-1896

## Test plan:

TBD


packages/wonder-blocks-tokens/src/tokens/color-thunderblocks.ts
```diff
@@ -0,0 +1,119 @@
+import {fade} from "../util/utils";
```
marcysutton, line 1: **question:** Should we put theme tokens into directories to organize them? Now that I'm looking into Typography tokens, I'm noticing this folder could get a bit crowded.

packages/wonder-blocks-tokens/src/tokens/color-thunderblocks.ts
```diff
@@ -0,0 +1,119 @@
...
+
+// Base colors
+const black = "#191918";
+const white = "#FFFFFF";
+
+export const color = {
```
marcysutton, line 7: **question:** I realize this is a draft PR...but can you add some comments or docs on how this works when you get a chance? I'm a little confused after looking at `color.ts` vs. `color-thunderblocks.ts`. The object entries are SO different, it's a bit difficult to understand the overlap! Does that only matter when you get to the theme merging part?

jandrade: I'm planning to add all of that later, but as you mentioned, this PR is very drafty, so I'll make sure to mark it as ready for review when I cover all the required parts. I need to work on other PRs first (and land them) before coming back to this one.

jandrade:
Took a slight different approach here: #2562 2562

packages/wonder-blocks-tokens/src/tokens/color-thunderblocks.ts
jandrade: @marcysutton Here's how the structure would look like: https://github.com/Khan/wonder-blocks/pull/2562

Let's continue the convo in that PR, thanks for the suggestions! :)

PR #2505, opened 2025-03-13
Switch to pnpm catalogs

jandrade:
## Summary:

Switching to pnpm catalogs to pin dependency versions across packages.

This is helpful for ensuring that all packages are using the same version of a
dependency, and for making it easier to update dependencies across all packages.

https://pnpm.io/catalogs

Issue: "none"

## Test plan:

Verify that the builds are still passing.

Also verify that the components work as expected in Storybook.

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,25 +129,23 @@
     "vitest": "^3.0.4"
   },
   "dependencies": {
-    "@babel/runtime": "^7.24.5",
-    "@khanacademy/wonder-stuff-core": "^1.5.4",
-    "@phosphor-icons/core": "^2.0.2",
-    "@popperjs/core": "^2.11.8",
-    "aphrodite": "^1.2.5",
-    "moment": "2.29.4",
-    "node-fetch": "^2.6.7",
-    "react": "18.2.0",
-    "react-dom": "18.2.0",
-    "react-popper": "^2.3.0",
-    "react-router": "5.3.4",
-    "react-router-dom": "5.3.4",
-    "react-window": "^1.8.11"
+    "@babel/runtime": "catalog:",
+    "@khanacademy/wonder-stuff-core": "catalog:",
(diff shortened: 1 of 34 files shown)


package.json
```diff
@@ -129,25 +129,23 @@
...
+    "react-popper": "catalog:",
+    "react-router": "catalog:",
+    "react-router-dom": "catalog:",
+    "react-window": "catalog:"
   },
   "resolutions": {
     "@types/react": "18",
-    "@types/react-dom": "18",
-    "strip-ansi": "6.0.1",
-    "strip-ansi-explanation": "There's an issue with strip-ansi v7 which causes conflicts with the Khan/changeset-per-package action"
```
jandrade, line 150: **note:** This is no longer necessary as we are not using `Khan/changeset-per-package"

pnpm-workspace.yaml
```diff
@@ -1,7 +1,31 @@
...
+    # Testing/data
+    node-fetch: ^2.6.7
+    # React
+    "react": 18.2.0
+    "react-dom": 18.2.0
+    # React Router
+    react-router: 5.3.4
+    react-router-dom: 5.3.4
+    # Virtualization
+    react-window: ^1.8.11
```
jandrade, line 31: **note:** This is moved from `package.json` to this place so we can centralize versions across in a single source of truth.

jandrade:
Here's the snapshot release in `webapp` to confirm that these changes won't break anything: https://prod-znd-250313-30365-dc0be24.khanacademy.org/

somewhatabstract (approved):
LGTM

Thanks for making this change!

jeremywiebe (commented):
Nice. I have wondered if we'll want separate catalogs for things like `react` vs other packages. But this feels like a good first step (all in one catalog).

PR #2506, opened 2025-03-13
Add `styles` prop to `LabeledField` and switch to sizing tokens

beaesguerra:
## Summary:
- Add `styles` prop to support custom stying for the label, description, and error elements
- Use CSS padding instead of `<Strut>` for spacing
- Migrated to `sizing` tokens instead of `spacing`

Issue: WB-1900

## Test plan:

- Review new Custom Styles story (`?path=/story/packages-labeledfield--custom-styles`)
- Existing styling should not be changed (no visual changes for existing stories should be detected in Chromatic, especially for `?path=/story/packages-labeledfield--scenarios`)

### DIFF
--- a/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
+++ b/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
@@ -3,7 +3,6 @@ import {StyleSheet} from "aphrodite";
 import WarningCircle from "@phosphor-icons/core/bold/warning-circle-bold.svg";
 
 import {View, addStyle, StyleType} from "@khanacademy/wonder-blocks-core";
-import {Strut} from "@khanacademy/wonder-blocks-layout";
 import {semanticColor, spacing} from "@khanacademy/wonder-blocks-tokens";
 import {LabelMedium, LabelSmall} from "@khanacademy/wonder-blocks-typography";
 import {PhosphorIcon} from "@khanacademy/wonder-blocks-icon";
@@ -143,7 +142,13 @@ export default function LabeledField(props: Props) {
         return (
             <React.Fragment>
                 <LabelMedium
-                    style={[styles.textWordBreak, styles.label]}
+                    style={[
+                        styles.textWordBreak,
+                        styles.label,
+                        description
+                            ? styles.labelWithDescription
+                            : styles.labelWithNoDescription,
+                    ]}
                     tag="label"
                     htmlFor={fieldId}
                     testId={testId && `${testId}-label`}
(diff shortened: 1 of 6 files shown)


packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -58,6 +57,15 @@ type Props = {
      * Custom styles for the labeled field container.
      */
     style?: StyleType;
```
beaesguerra, line 59: We can remove the `style` prop later and move it to the `styles` object with a key for `root`. This would be a breaking change so I created https://khanacademy.atlassian.net/browse/WB-1903 for that work!

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -253,15 +271,22 @@ const styles = StyleSheet.create({
     label: {
         color: semanticColor.text.primary,
     },
+    labelWithDescription: {
+        paddingBottom: sizing.size_050,
```
nedredmond, line 275: should these be `paddingBlockEnd`?

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -253,15 +271,22 @@ const styles = StyleSheet.create({
...
+        paddingBottom: sizing.size_150,
     },
     errorSection: {
         flexDirection: "row",
-        gap: spacing.xSmall_8,
+        gap: sizing.size_100,
     },
     errorSectionWithContent: {
-        paddingTop: spacing.small_12,
+        paddingTop: sizing.size_150,
```
nedredmond, line 289: `paddingBlockStart`?

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
beaesguerra: Yesss, thanks for catching that!

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
jandrade: I was thinking about that.... Do you think you could make that change in this PR? I checked in `webapp` and the `style` prop is only used once. I feel that keeping `style` and `styles` around would be a bit confusing, so it would be nicer if we could make the change right away to make the API cleaner from start.

beaesguerra: Sure, I can include it in these changes!

PR #2507, opened 2025-03-18
[DO NOT REVIEW]: Upgrade to ReactRouter v6

jandrade:
## Summary:

This is a WIP PR to upgrade to React Router v6. This PR is not meant for review,
but rather to enable WB to work with the new CSS View Transition API.

Issue: "none"

## Test plan:

Verify that the ReactRouter-related examples still work, for example:

`/?path=/story/packages-button--with-router`


jandrade:
No longer needed

PR #2508, opened 2025-03-20
NavigationTabs Styling

beaesguerra:
## Summary:

The following changes address various styling requirements and includes adding some tools for testing in Storybook

Props
- NavigationTabs supports `styles` prop and NavigationTabs supports `style` prop for custom styling

Styles
- States for current, press, hover, and focus (using a box shadow) are implemented
- Semantic color and sizing tokens are used (colors may need to be revisited once Thunder Blocks theme is implemented)
- Used box-shadow for the underline. This doesn't conflict with the box-shadow used for the focus outline of the link because the underline is on the parent `<li>` element instead of the `<a>`
- Long labels within the tab won't wrap. No max-width is set at the time to better support translated strings
- The navigation tabs will horizontally scroll if the tabs don't fit on the screen
- Styling is tested with Link, anchor tags, and IconButton
- Handles RTL

Storybook
- Added a decorator and toolbar control to set the language direction
- Added a decorator and toolbar control to set specific zoom presets
- New constants file for text to use for testing different cases (`text-for-testing.ts`)
- New ScenariosLayout component for laying out various scenarios in a list
- Making AllVariants responsive and supporting a list only mode (useful for testing zoom for all variants)

Issue: WB-1898

## Test plan:
- Review stories and docs for NavigationTabs and NavigationTabItem
- Testing rtl, zoom, different browsers, responsiveness


__docs__/wonder-blocks-tabs/navigation-tab-item-variants.stories.tsx
```diff
@@ -143,8 +201,42 @@ export const Press: Story = {
...
+            </div>
+        </div>
+    ),
+    globals: {
+        zoom: "400%",
+    },
+    parameters: {
+        chromatic: {
+            // Disabling because Chromatic crops the story when zoom is used
+            disableSnapshot: true,
```
beaesguerra, line 232: I disabled the Zoom stories in Chromatic because the stories would get cropped in the snapshots 😢 here is an [example in Chromatic](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=67dc6d67c5e1fc7d3d894464). I tried some things to address this but couldn't get it working. Let me know if you have other ideas!
- setting the `viewport` config for Chromatic for that story
- adding a delay for the snapshot
- setting width: 100% on the decorator

image

packages/wonder-blocks-tabs/src/components/navigation-tab-item.tsx
```diff
@@ -89,35 +103,55 @@ export const NavigationTabItem = React.forwardRef(function NavigationTabItem(
...
-    right: 0,
-    height: sizing.size_050,
-};
-
 const styles = StyleSheet.create({
     root: {
         listStyle: "none",
         display: "inline-flex",
+        [":has(a:hover)" as any]: {
+            boxShadow: `inset 0 -${sizing.size_050} 0 0 ${semanticColor.action.secondary.progressive.hover.foreground}`,
```
beaesguerra, line 111: I mostly used the secondary action semantic tokens. We may need to revisit this though once Thunder Blocks theming is in place. ([related Figma thread](https://www.figma.com/design/EuFu0U7gqc1koXm8ZhlOLp?node-id=5507-1646#1181055789))

.storybook/preview.tsx
```diff
@@ -143,6 +176,48 @@ const preview: Preview = {
...
+            },
+        },
+        zoom: {
+            description: "Preset zoom level",
+            toolbar: {
+                title: "Zoom Presets",
+                icon: "zoom",
+                items: [
+                    {
+                        // undefined so the there is no zoom value set
```
marcysutton, line 206: **nit:** extra word?

```suggestion
                        // undefined so there is no zoom value set
```

__docs__/components/all-variants.tsx
```diff
@@ -31,59 +34,125 @@ type Props = {
...
+                                                border: `${border.width.hairline}px dashed ${semanticColor.border.primary}`,
+                                            },
+                                        ]}
+                                    >
+                                        {children(
+                                            {
+                                                ...row.props,
+                                                ...col.props,
+                                            },
+                                            `${row.name} ${col.name}`,
```
marcysutton, line 93: Oh interesting, I just did something super similar in #2509!

__docs__/wonder-blocks-tabs/navigation-tab-item-variants.stories.tsx
```diff
@@ -143,8 +201,36 @@ export const Press: Story = {
...
+                        <View style={styles.container}>
+                            <NavigationTabItem {...args} {...props} />
+                        </View>
+                    )}
+                </AllVariants>
+            </div>
+        </div>
+    ),
+    globals: {
+        zoom: "400%",
```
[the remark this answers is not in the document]
marcysutton: That is so useful!!

__docs__/wonder-blocks-tabs/navigation-tab-item-variants.stories.tsx
marcysutton: Weird. Maybe try `min-width: 100vw`, and/or some borders to see if you can determine which top-level element is being cut off?

__docs__/wonder-blocks-tabs/navigation-tabs-variants.stories.tsx
```diff
@@ -141,6 +227,58 @@ export const Active: Story = {
...
+                    )}
+                </AllVariants>
+            </div>
+        </div>
+    ),
+    parameters: {
+        a11y: {
+            config: {
+                rules: [
+                    // Disabling warning: "Element's background color could not
```
marcysutton, line 263: I'm curious about this one, as I wrote part of the overlapping logic in the color-contrast rule! It looks like the Axe extension doesn't surface incompletes anymore, and I don't see any Needs Review items in Accessibility Insights either. Maybe Storybook is using an older axe-core API version?

__docs__/wonder-blocks-tabs/navigation-tabs.stories.tsx
```diff
@@ -49,3 +60,286 @@ export const Default: StoryComponentType = {
...
+        return (
+            <StyledDiv style={styles.pageStyle}>
+                <StyledHeader style={styles.headerStyle}>
+                    <img
+                        src="/logo.svg"
+                        width="40px"
+                        alt="Wonder Blocks logo"
+                    />
+                    <SingleSelect
+                        placeholder="Placeholder"
```
marcysutton, line 314: This one needs a label with the new API!

```suggestion
                        aria-label="Dropdown"
                        placeholder="Placeholder"
```

marcysutton (approved):
Great work @beaesguerra! I left you a few really tiny tweaks. Excited to see this component in the wild!

__docs__/components/all-variants.tsx
```diff
@@ -31,59 +34,125 @@ type Props = {
...
+                                    Column: {column.name}, Row: {row.name}
+                                </LabelLarge>
+
+                                <div
+                                    style={{
+                                        padding: sizing.size_100,
+                                        marginBlock: sizing.size_100,
+                                        border: `${border.width.hairline}px dashed ${semanticColor.border.primary}`,
+                                    }}
+                                >
```
jandrade, line 122: **suggestion:** It would be great switching to `View` (or `StyledDiv` if view doesn't fit) to prevent inlining the CSS in the html.

```suggestion
                                <View
                                    style={{
                                        padding: sizing.size_100,
                                        marginBlock: sizing.size_100,
                                        border: `${border.width.hairline}px dashed ${semanticColor.border.primary}`,
                                    }}
                                >
```

Expanding on this, the styles could be moved to the `styles` object as well.

__docs__/wonder-blocks-tabs/navigation-tab-item-variants.stories.tsx
```diff
@@ -96,17 +145,26 @@ const meta = {
...
     component: NavigationTabItem,
     render: (args) => (
-        <AllVariants rows={rows} columns={columns}>
-            {(props) => (
-                <View style={styles.container}>
-                    <NavigationTabItem {...args} {...props} />
-                </View>
-            )}
-        </AllVariants>
+        <div>
```
jandrade, line 148: **suggestion:** You could also use a fragment to avoid adding extra divs.

```suggestion
        <>
```

__docs__/wonder-blocks-tabs/navigation-tab-item-variants.stories.tsx
```diff
@@ -143,8 +201,42 @@ export const Press: Story = {
     parameters: {pseudo: {hover: true, active: true}},
 };
 
+export const Zoom: Story = {
+    render: (args) => (
+        <div>
```
jandrade, line 206:
```suggestion
        <>
```

__docs__/wonder-blocks-tabs/navigation-tab-item.stories.tsx
```diff
@@ -108,3 +115,104 @@ export const ChildrenRenderFunction: StoryComponentType = {
...
+ * Custom styles can be set for the NavigationTabItem.
+ *
+ * For custom link styling, prefer applying the styles to the `Link` component.
+ * Note: The `NavigationTabItem` will also set styles to the `Link` child
+ * component.
+ *
+ * If there is a specific use case where the styling needs to be
+ * overridden, please reach out to the Wonder Blocks team!
+ */
+export const CustomStyle: StoryComponentType = {
```
jandrade, line 129: **suggestion:** By looking at the example, this could be disabled in Chromatic (unless you expect to introduce more visual changes).

jandrade (commented):
Thanks for working on this and adding all these a11y features, it is going to be REALLY useful in general. I've left some comments for you to consider before giving a final approval.

__docs__/wonder-blocks-tabs/navigation-tab-item-variants.stories.tsx
```diff
@@ -96,17 +145,26 @@ const meta = {
...
-        </AllVariants>
+        <div>
+            <AllVariants rows={rows} columns={columns}>
+                {(props) => (
+                    <View style={styles.container}>
+                        <NavigationTabItem {...args} {...props} />
+                    </View>
+                )}
+            </AllVariants>
+            <div dir="rtl">
```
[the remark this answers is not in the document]
beaesguerra: Haha yes I was referencing your work for the AllVariants with rtl!

 > For later.... it would be nice defining a general pattern in all-variants.tsx to include RTL by default, but we can define that later if needed.

While working on this, I attempted to implement this because I thought it'd be nice to include the rtl cases by default! I then realized I had to change the labels for rtl so it wasn't using English. Thinking back, this probably could be solved by passing a boolean to the children function for whether or not it is rtl. Here's the commit I tried it in https://github.com/Khan/wonder-blocks/pull/2508/commits/4d233cebd34d8c99eb8157d672a59da2e1b8f970 .

On a related note, I was also wondering if we wanted the `AllVariants` component to cover the hover, focus, press states too in one story instead having one story for each state combination. We could do this by wrapping each table with an id, and then specifying specific selectors when we configure the pseudo states add on:
```
parameters: {
      pseudo: {
          hover: [
              `#${stateElementIds.hover} *`,
              `#${stateElementIds.hoverAndFocus} *`,
          ],
          focusVisible: [
              `#${stateElementIds.focus} *`,
              `#${stateElementIds.hoverAndFocus} *`,
          ],
          active: [`#${stateElementIds.press} *`],
      },
  },
```

While it's possible, I'm not sure if it would be helpful to include it all in one sticker sheet. It would save snapshot amounts though since we'd have ~5 snapshots per component due to the different states!

Curious to hear if anyone has thoughts on this! Happy to iterate on this in another PR :)

packages/wonder-blocks-tabs/src/components/navigation-tab-item.tsx
beaesguerra: Good catch!! Updated the hover underline to be 2px tall. I also made sure the current + hover state remains the same at 4px!

__docs__/components/all-variants.tsx
beaesguerra: Nice!! I was also addressing a11y warnings in the new stories! I see in #2509 we add `aria-label` to the props. I ended up adding a new argument for the `name` so it could be used for a unique `aria-label`. I was thinking not all components may support an `aria-label` at the root level so providing it as an arg keeps it flexible! What do you think?

__docs__/wonder-blocks-tabs/navigation-tabs-variants.stories.tsx
beaesguerra: It was showing as an incomplete in the Storybook add on! Any tips on how to address this? 

image

I also see it in the labels for the story! Might be related to setting the CSS zoom property?
image

__docs__/components/all-variants.tsx
beaesguerra: Thanks for the reminder! Updated :)

packages/wonder-blocks-tabs/src/components/navigation-tabs.tsx
```diff
@@ -32,8 +32,16 @@ type Props = AriaProps & {
      * tabs.
      */
     "aria-labelledby"?: string;
+    /**
+     * Custom styles for the elements in NavigationTabs.
+     */
+    styles?: {
+        root?: StyleType;
+        list?: StyleType;
```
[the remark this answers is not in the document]
beaesguerra: Added more docs!

__docs__/wonder-blocks-tabs/navigation-tab-item.stories.tsx
beaesguerra: It is verifying that the style prop works in the visual regression tests (very subtly using spacing 😅). I updated the custom style to have a background color instead so it is more noticable!

__docs__/wonder-blocks-tabs/navigation-tab-item-variants.stories.tsx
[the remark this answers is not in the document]
jandrade: oh merging pseudo states would be worth experimenting, and then we could see if there's benefit from that change. how does that sound?

__docs__/wonder-blocks-tabs/navigation-tab-item-variants.stories.tsx
beaesguerra: Thanks for the suggestions! 
- Setting `min-width: 100vw` made it so the NavigationTabs are no longer horizontally scrollable since it allows the container to expand. We want to be able to observe the scrolling behaviour!
- Related to that, I tried setting `width: 100vw` and `maxWidth: 100vw`. The snapshot shows more of the contents, though we lose the horizontal scrolling in the NavigationTabs (it's put on the story instead). [Chromatic snapshot](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=67ddc970032bfa82f8d939b1)
- Setting `viewports` didn't work. [Chromatic with viewports config](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=67ddc5cc6d557f9e26ef4ba5)
- Setting `defaultViewport` didn't work too (the snapshot was less wide too for some reason) [Chromatic with defaultViewport](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=67ddc7f2b3eb368b88187ce4)

I went back to disabling the Zoom stories 😅 I think at least the stories are functional and show the expected behaviour when you try it out in Storybook! One other idea is we could reach out to the Chromatic team to see if they've seen this issue before. (I think it's a Chromatic issue since the story renders as expected and it's only the snapshot that is cropped!). @jandrade how do you normally contact Chromatic for support?

__docs__/wonder-blocks-tabs/navigation-tab-item-variants.stories.tsx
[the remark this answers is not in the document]
beaesguerra: I experimented with merging the pseudo states and rtl in the same story: https://github.com/Khan/wonder-blocks/pull/2511

Left some comments in the draft PR too, feel free to provide feedback in that PR!

__docs__/wonder-blocks-tabs/navigation-tab-item-variants.stories.tsx
jandrade: That's via their website. You go to the chromatic site, go to the dashboard, then click on the "Chat" button on the bottom right.

https://github.com/user-attachments/assets/7c7166f0-737a-4788-beb5-a97a85eb2923

jandrade (approved):
LGTM! Thank you for iterating on the Zoom story. Hopefully the Chromatic team will now how to fix that (or apply a patch to this issue). 🚀

PR #2509, opened 2025-03-20
Address numerous a11y issues in WB Stories

marcysutton:
## Summary:
[misc-a11y-issues] WB-1838: Add labels to toolbar icons
[misc-a11y-issues] WB-1837: Add labels to switches
[misc-a11y-issues] WB-1833: Add labels to radios
[misc-a11y-issues] WB-1832: Fix ids on Choices
[misc-a11y-issues] WB-1831: Add labels to checkboxes
[misc-a11y-issues] WB-1823: Add aria-label to listbox
[misc-a11y-issues] WB-1822: Fix actionItem role issue
[misc-a11y-issues] WB-1821: Fix tab role issue
[misc-a11y-issues] WB-1820: Fix label in button story
[misc-a11y-issues] WB-1819: Fix banner spinner label
[misc-a11y-issues] WB-1826: add listbox to options

Parent Issue: WB-1817

## Test plan:

1. Open stories in tickets
2. Verify there are no violations in Storybook accessibility pane


somewhatabstract (approved):
LGTM. Thanks for getting these fixed up!

__docs__/wonder-blocks-form/choice.stories.tsx
```diff
@@ -83,7 +104,6 @@ export const Default: StoryComponentType = {
     render: (args) => <ChoiceWrapper {...args} />,
     args: {
         label: "Pineapple (Control)",
-        value: "pineapple",
```
beaesguerra, line 86: Do we need the `value` arg set so the `value` is set on the Pineapple choice on line 57?

https://github.com/Khan/wonder-blocks/blob/b42a7b6ba25e40530c1b7c9854f1bcb75e07ec74/__docs__/wonder-blocks-form/choice.stories.tsx#L57

__docs__/wonder-blocks-dropdown/option-item.stories.tsx
```diff
@@ -55,7 +55,9 @@ export default {
     decorators: [
         (Story): React.ReactElement<React.ComponentProps<typeof View>> => (
             <View style={styles.example}>
-                <Story />
+                <div role="listbox" aria-label="Example">
```
beaesguerra, line 58: Something I've been doing to address a11y warnings is to disable specific rules if we know it is expected.

 For example, for the `NavigationTabItem` stories, because it renders a `<li>` without a `<ul>` (since that's provided by `NavigationTabs`), we expect a related warning about that, but there should be no warnings when `NavigationTabs` and `NavigationTabItem` are used together! Here's an [example of disabling the `listitem` rule](https://github.com/Khan/wonder-blocks/pull/2500/files#diff-4f9d8a9178620afc2da5e873abed4bf2d97c9f7851030b313d503fbb59706679R18-R27) when it's expected

Do we have a preference for disabling expected rules or wrapping it with additional things to make sure it's semantically correct? Curious to see if others have thoughts on this!

beaesguerra (approved):
Yay looks great! Thanks for addressing these issues in our stories 🎉  Left some questions!

__docs__/wonder-blocks-form/choice.stories.tsx
marcysutton: Yes, good catch! Those repeated controls were broken before (one label for two inputs)...I had to go through each prop piece by piece to fix it and missed putting one back!

__docs__/wonder-blocks-dropdown/option-item.stories.tsx
marcysutton: That's a really good question. My personal opinion is that it would be better to render an accessible example even if the component is incomplete by itself. The code could document those details so it's clear!

beaesguerra: Sounds good! I'll update those stories, thanks for the feedback!

__docs__/wonder-blocks-form/choice.stories.tsx
marcysutton: I added this value back in!

PR #2510, opened 2025-03-20
[WB-1871] IconButton: Rename `color` prop to `actionType`

jandrade:
## Summary:

We are trying to get more alignment with design, and part of that is that icon
buttons will use an `actionType` prop, which supports `progressive` and
`destructive`.

This PR includes the following API changes:

- `color` => `actionType`.
- `color=default` => `actionType=progressive`.

NOTE: `progressive` is now the default value.

Issue: https://khanacademy.atlassian.net/browse/WB-1871

## Test plan:

Navigate to the IconButton docs and verify that the prop is now `actionType`
instead of `color`.

/?path=/docs/packages-iconbutton--docs#actiontype

### DIFF
--- a/__docs__/wonder-blocks-icon-button/icon-button.argtypes.ts
+++ b/__docs__/wonder-blocks-icon-button/icon-button.argtypes.ts
@@ -2,18 +2,13 @@ import type {ArgTypes} from "@storybook/react";
 import {IconMappings} from "../wonder-blocks-icon/phosphor-icon.argtypes";
 
 export default {
-    color: {
+    actionType: {
         control: {
             type: "select",
         },
-        description: "Color of the icon button.",
-        options: ["default", "destructive"],
         table: {
-            type: {
-                summary: `"default" | "destructive"`,
-            },
             defaultValue: {
-                summary: `"default"`,
+                summary: `"progressive"`,
             },
         },
     },
@@ -21,25 +16,18 @@ export default {
         control: {
             type: "boolean",
         },
-        description: "Whether the icon button is disabled.",
     },
     icon: {
         control: {
             type: "select",
         },
-        description: "A Phosphor icon asset (imported as a static SVG file).",
         options: IconMappings as any,
     },
     kind: {
         control: {
             type: "select",
         },
-        description: "The kind of the icon button.",
-        options: ["primary", "secondary", "tertiary"],
         table: {
-            type: {
-                summary: `"primary" | "secondary" | "tertiary"`,
-            },
             defaultValue: {
                 summary: `"primary"`,
             },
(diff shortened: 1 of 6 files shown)


__docs__/wonder-blocks-icon-button/icon-button.argtypes.ts
jandrade, line 1: **note:** Simplified argTypes here to rely on the code comments directly (instead of duplicating the description here).

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
```diff
@@ -207,40 +202,24 @@ type ActionType =
...
-    buttonColor: ButtonColor,
+    actionType: IconButtonActionType = "progressive",
     disabled: boolean,
     kind: Kind,
     light: boolean,
     theme: IconButtonThemeContract,
 ) {
-    let actionType: ActionType = getActionType(buttonColor, disabled);
-    const themeVariant = theme.color[kind][actionType];
+    let actionTypeOrDisabled: ActionType = disabled ? "disabled" : actionType;
```
jandrade, line 212: **note:** This replaces the logic of `getActionType()` now that we have a correct mapping. This is just to add support to the `disabled` variant.

marcysutton (approved):
This looks good to me, but I'll be curious to hear @beaesguerra's thoughts! I think it will clarify how color is used in themes. It was a bit confusing to me before with the levels of tokens and configuration objects.

My only other thought is: should there be tests for these changes, or does Storybook cover it?

jandrade:
> My only other thought is: should there be tests for these changes, or does Storybook cover it?

@marcysutton Great question! We have been moving a lot of unit tests that covered styling to Visual regression testing with Storybook + Chromatic, and in this case it would be covered by these tests:

https://5e1bf4b385e3fb0020b7073c-zqttvbhhap.chromatic.com/?path=/story/packages-iconbutton-all-variants--default

jandrade:
> * Are there plans to also update perseus before webapp since it's a breaking change?

That's correct. I'll chat with the LEMS team before landing this change (sometime next week).

> * Will we still also be updating the icon semantic tokens (`primary` to `neutral`, getting rid of the `secondary` semantic icon token, etc)? And having the `kind` prop values match those values? (related to the [thread](https://khanacademy.slack.com/archives/C07CA2YR0BZ/p1741897551684599?thread_ts=1738357739.268909&cid=C07CA2YR0BZ) from last week!)

Yes, but that will happen in an upcoming PR: https://khanacademy.atlassian.net/browse/WB-1902

I'm going to add an "Implementation plan" to this PR and the stacked ones to make the plan clearer.

Thanks for asking these questions!

PR #2511, opened 2025-03-21
Experimenting with single AllVariants story for all pseudo states

beaesguerra:
## Summary:

Issue: WB-XXXX

## Test plan:

### DIFF
--- a/__docs__/components/all-variants-states.tsx
+++ b/__docs__/components/all-variants-states.tsx
@@ -1,10 +1,7 @@
 import * as React from "react";
 import {StyleSheet} from "aphrodite";
 import {AllVariants, type Props as AllVariantsProps} from "./all-variants";
-import {
-    HeadingLarge,
-    HeadingMedium,
-} from "@khanacademy/wonder-blocks-typography";
+import {HeadingMedium} from "@khanacademy/wonder-blocks-typography";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {sizing} from "@khanacademy/wonder-blocks-tokens";
 
@@ -19,11 +16,6 @@ type Props = Omit<AllVariantsProps, "children"> & {
      * The states to render.
      */
     states?: State[];
-
-    /**
-     * Whether to show the RTL version of the component. Defaults to true.
-     */
-    showRtl?: boolean;
 };
 
 type State = {name: string; id: string};
@@ -57,14 +49,7 @@ export const defaultStates: State[] = [
 ];
 
 export const AllVariantsStates = (props: Props) => {
-    const {
-        children,
-        rows,
-        columns,
-        layout,
-        states = defaultStates,
-        showRtl = true,
-    } = props;
+    const {children, rows, columns, layout, states = defaultStates} = props;
 
     const renderState = (state: State, isRtl: boolean = false) => (
         <View key={state.id} id={state.id} style={styles.state}>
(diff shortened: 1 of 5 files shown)


__docs__/components/all-variants-states.tsx
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * Whether to show the RTL version of the component. Defaults to true.
+     */
+    showRtl?: boolean;
+};
+
+type State = {name: string; id: string};
+
+export const commonStates = {
+    base: {name: "Base", id: "base"},
```
jandrade, line 32: **thought:** How do you feel about changing this to `Rest` to be consistent on how Design uses these states in Figma? Would love to hear your thoughts on why `base` could apply better here (if that's the case).

__docs__/components/all-variants-states.tsx
```diff
@@ -0,0 +1,104 @@
...
+    hover: [
+        `#${commonStates.hover.id} *`,
+        `#${commonStates.hoverAndFocus.id} *`,
+    ],
+    focusVisible: [
+        `#${commonStates.focus.id} *`,
+        `#${commonStates.hoverAndFocus.id} *`,
+    ],
+    active: [`#${commonStates.press.id} *`],
+};
```
jandrade, line 49: **praise:** Ohh this is neat! Very smart solution :). this is using this feature right? https://github.com/chromaui/storybook-addon-pseudo-states?tab=readme-ov-file#targeting-specific-elements

jandrade (commented):
This is a super interesting experiment! thanks for looking into this 🫶 

> Do we want to combine all the pseudo states into one story?
> Is it managable to review visual changes in Chromatic with these larger stories?

I'm seeing that in some cases there's a lot of information in the snapshot, so I'm not sure if that would make it harder to review and/or find small changes. How do you feel about that?

> Do we want to include rtl for all the pseudo states?

IMO yes! we should include `rtl` as a first-class citizen in our visual regression tests.

> What about the approach for zoom?

Where you able to fix the viewport issue?

__docs__/components/all-variants-states.tsx
beaesguerra: Yes! Then we can target specific sections with the selector styles 😄

__docs__/components/all-variants-states.tsx
beaesguerra: Rest makes sense! Will update :)

beaesguerra:
@jandrade 

> I'm seeing that in some cases there's a lot of information in the snapshot, so I'm not sure if that would make it harder to review and/or find small changes. How do you feel about that?

I find when there are many small changes in the same snapshot, it takes longer for me to review all the details (that might just be because we're making more changes right now to support polaris!). With that said, this set up with all the states doesn't account for different themes yet! Would we want it to? Or are we still exploring using `modes`? 

Thinking about themes, it might be easier to see all the states for one theme in one snapshot. So each component would have a stickersheet for each theme.

Let me know what you think! If we want to combine the states into one story, I'll clean up this branch and open a separate PR for it!

> > What about the approach for zoom?
> 
> Where you able to fix the viewport issue?

I reached out to Chromatic and they suggested adding a delay. That didn't fix things so they escalated it to their engineers and will follow up!

beaesguerra:
Closing since this has been iterated on in https://github.com/Khan/wonder-blocks/pull/2550 !

PR #2520, opened 2025-03-25
NavigationTabs stories updates

beaesguerra:
## Summary:

Minor updates to the NavigationTabs stories:
- Circling back to wrap NavigationTabItem with a `<ul>` element instead of disabling expected a11y rules ([discussed in the a11y fixes PR](https://github.com/Khan/wonder-blocks/pull/2509#discussion_r2008067305)). It was expected because the NavigationTabs component provides the `<ul>` element for it
- Use the logo with text image instead for header example (the logo asset can't be found in the deployed version, but the logo with text image should be available since we use that for the logo in Storybook)
- Update comments

Issue: WB-XXXX

## Test plan:
- Confirm the NavigationTabItem stories don't have a11y warnings using the Storybook a11y add on


somewhatabstract (approved):
LGTM

PR #2521, opened 2025-03-25
NavigationTabs: Add `animated` prop to enable transition animation

beaesguerra:
## Summary:

If the `animated` prop is set to `true`, the current tab underline will transition to the tab item that is selected

We achieve this animation by moving the current indicator from NavigationTabItem to NavigationTabs. This is so that the indicator can move fluidly to the other tab items. In order to know what size and position to use for the underline, we dynamically determine the width and position of the current tab. We also use a ResizeObserver to update the width and position in case the sizing of the tabs changes. This also addresses the initial layout shift that occurs when the Aphrodite styles are first initialized.

Issue: WB-1906

## Test plan:
`?path=/story/packages-tabs-navigationtabs-navigationtabs--interactive`
- Confirm setting `animated` to `false` disables the transition and the current tab is still shown
- Confirm setting `animated` to `true` enables the current tab item transition

`?path=/story/packages-tabs-navigationtabs-navigationtabs-navigationtabs-all-variants--default`

Manually confirm the styling for NavigationTabs states (these were generating flaky snapshots in Chromatic and have been temporarily disabled. I reached out to the Chromatic team for support and created WB-1917 as a follow up!)

- hover
- focus
- hover + focus
- active
- zoom


__docs__/components/all-variants.tsx
```diff
@@ -87,7 +87,7 @@ export function AllVariants(props: Props) {
                                     >
                                         {children(
                                             {
-                                                ariaLabel: `${row.name} ${col.name}`,
+                                                "aria-label": `${row.name} ${col.name}`,
```
beaesguerra, line 90: Was getting a warning that `ariaLabel` isn't a prop!

__docs__/wonder-blocks-tabs/navigation-tabs-variants.stories.tsx
```diff
@@ -211,16 +211,34 @@ export default meta;
 export const Default: Story = {};
 
 export const Hover: Story = {
-    parameters: {pseudo: {hover: true}},
+    parameters: {
+        pseudo: {hover: true},
+        chromatic: {
+            // Need different delay value for Chromatic to detect the styles
+            delay: 500,
```
beaesguerra, line 224: The Hover, Focus, Hover + Focus, and Active stories for NavigationTabs have been flaky 😢 

Sometimes the state styles show up in the Chromatic snapshot, sometimes they don't. Let me know if you have any ideas on how to address this, the delays don't seem to help with consistency, but thought I'd try it and see if others have ideas!

Luckily though, the AllVariants stories for NavigationTabItem is more stable and covers the different states too. I'm thinking of disabling these stories in Chromatic if we can't find a way to make the NavigationTabs AllVariants more stable and then reaching out to the Chromatic team (I'm still in contact with them for the zoom chromatic issue too)

Examples of inconsistencies:
- In [build 6289](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=67e5c2f0952ca1fff129eb5e), focus and active styles aren't shown
- In [build 6287](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=67e5bf1e4c6e47d15ededc73), hover story and focus story don't have the styling shown, but the hover + focus story does 🙃

marcysutton (commented):
Cool work on this @beaesguerra! My main question is around `prefers-reduced-motion`. I see a transition in the NavigationTabs styles. It seems like this should just be turned off internally if the user has their motion preference set to `reduce` so every consumer doesn't have to do that work. But I'm curious to hear your take on it!

__docs__/wonder-blocks-tabs/accessibility.mdx
```diff
@@ -11,6 +11,9 @@ component, especially if there are multiple `<nav>` elements on the page.
     - If there is already a visual label for the tabs, set the `aria-labelledby`
     prop to the id of the visual label
     - Otherwise, set the `aria-label` prop.
+- When enabling animations in the component using the `animated` prop, consider
```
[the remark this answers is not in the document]
beaesguerra: Great question! The `animated` prop approach follows what we have for the [Accordion](https://khan.github.io/wonder-blocks/?path=/docs/packages-accordion-accessibility--docs#considerations-for-the-user) component!

From what I understand, we handle it this way because users can also configure a setting in webapp (`/settings/account`). When we use this setting in webapp, we also [check for the os setting](https://github.com/Khan/webapp/blob/10fe1ebd792e6650c8332384605cab2a567847f2/services/static/javascript/data-access/internal/prefer-reduced-motion.ts#L18-L25)!

image


cc: @jandrade in case you have more context on how we handle animations in WB!

__docs__/wonder-blocks-tabs/navigation-tabs-variants.stories.tsx
beaesguerra: I found that you can re-run Chromatic tests on the platform, and once you do, you can debug using `Traces` and step through what the test sees. Here's an example build from a re-run: https://www.chromatic.com/build?appId=5e1bf4b385e3fb0020b7073c&number=6289
- In this re-run, the Focus story is flaky and is shown earlier in the replay and then disappears 😢 I'll try again to see if we can find why it disappears after!
- The [Trace for the focus story](https://trace.playwright.dev/?trace=https://snapshots.chromatic.com/snapshots/5e1bf4b385e3fb0020b7073c-67e5c2f0952ca1fff129e8b5/trace-3367fcc5.zip?assetToken=eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJyZXNvdXJjZUtleSI6IjVlMWJmNGIzODVlM2ZiMDAyMGI3MDczYy02N2U1YzJmMDk1MmNhMWZmZjEyOWU4YjUiLCJleHAiOjE3NDMxODI4ODQuOTcyLCJmaWxlIjoidHJhY2UtMzM2N2ZjYzUuemlwIn0.pV3z3gNr7X5tcgBqlMZyJ_IL1U-osy0xhdO1VfB3j4XGB9QiRcfp2XbaAPnkQe50lWMyn2MsWa1ln6mTnpRC-7R-TxPRs0ACQalKuqDTfOco3tAqyOnZQPynjhW6cAwT2V1ZT7k1KlogHftwAk0Q7fN2a7tiun0ZK9MyPVz6oA_59p3U-mCg4i6EOyEn4UKYWh7MlIUSYwXA8x0hPSIOVmSeXkvhLv_4VbscDOkMPBPKY-nKAbMMSLw6z1sw6KZpj4vwOD4x53cbjVbdZvDldCh6eN-D_0YHsTfsV-NrL5287Zgdxe0iktczG7A_glUXQZ12qpjcdatX30wNX3x6c1rFWyKALGw5APYKYVbKeFCHAeKivI1SLpbrg1E2BycvLhOLHT3JGmFlk4SN7cDKL_GB-whSiisKANMqN1VYhypS0prbI7KBYutPbE34wvCiRfLl-mIH7MtGN7Lv1Ve4MyYjujI7N8mjQLZ_TMSrNO2GQb7QKFH3zNCvXXTU0yuRgcGmBIcuamZgADlzGg0VGowswBASI-Yw-y7GeExvK1iEaZSsthARmzjn15_VnMUKlKn_K4zYwVXCe_vGxbh-vcl3n96B3tCjhSbZzAlgqFkLMkpMA0bDlaZH2aknKCyUvgiBZfGUcgoJJ3Gm8tFo8h4m6hk6VYW7Pe77egmQBtQ)

__docs__/wonder-blocks-tabs/accessibility.mdx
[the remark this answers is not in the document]
jandrade: ahh yeah, we use that `animated` approach b/c we have to check if the user has disabled motion at the OS level or at the site level (as Bea described).

I remember debating at some point if `animated` should be `false` or `true` by default, but I don't remember if we landed on a decision.

@marcysutton @beaesguerra do you have any thoughts/suggestions about what the default state should be?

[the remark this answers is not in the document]
marcysutton: Ah, that makes sense! Good to know there are levers at the site level for this setting. Maybe we could document that a bit more so consumers are aware of why it is set up this way? 

Generally the accessibility recommendation for animation is to disable it by default, so there is less of a risk of harming someone. That might be a good idea here since we aren't tapping into reduced motion in the library itself.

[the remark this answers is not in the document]
beaesguerra: Defaulting `animated` to `false` makes sense to me! 

I elaborated on this more in the accessibility docs for NavigationTabs. I also updated our [component guidelines](https://khanacademy.atlassian.net/wiki/spaces/WB/pages/3186294919/Guidelines+for+Web+Wonder+Blocks+Components#General-Props) for the `animated` prop!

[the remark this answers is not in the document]
beaesguerra: It would be nice to have something so consumers don't need to configure the `animated` prop every time! I could see us using the Settings provider for other things too (like theme or maybe injecting the announcer?)

At first, I was hesitant of having design system components rely on context outside of the component because then they can no longer be used in isolation in the application or in tests!
- One mitigation for this could be defaulting the animations to false if there is no context provider! (as opposed to throwing an error if the context can't be found. Or maybe we do want to throw an error to make sure it gets configured?)
- Since we are informed about the consuming projects (perseus and webapp), we can make sure the settings provider in place when we introduce the change. With more frontend apps and libs, there would be more to keep up with. Would all the apps and libs need to have the settings provider?

Some other ideas/thoughts:

1. We could have a shared component in webapp that sets the `animated` prop based on `usePrefersReducedMotion()`. This would be similar to how we have shared components for WB in webapp to provide things like translations for the `labels` prop
- If we decide to go with the context approach, maybe the WBSettings could also receive translations for WB components so that common labels (like the aria-label for a modal close button or clear button) don't need to be set each time

2. Revisit the prefers reduced motion site setting - I'm not sure about this one, but if we find that we don't need the site setting because the OS setting is sufficient, we wouldn't need the `animated` props and we can rely on CSS to enable/disable styles! 

3. Do you know how other animations (like the page transitions) will be handled in polaris?

__docs__/wonder-blocks-tabs/navigation-tabs-variants.stories.tsx
beaesguerra, line 1: See related thread around Chromatic + pseudo state stories: https://github.com/Khan/wonder-blocks/pull/2521#discussion_r2017632459

__docs__/wonder-blocks-tabs/accessibility.mdx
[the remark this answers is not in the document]
jandrade: Now I remember that this was discussed in the past: https://khanacademy.slack.com/archives/C4PE1QM5Y/p1708721034153189

@somewhatabstract provided a great insight in this regard, which would be a combination of a WB hook + singleton to centralize the logic for setting/updating the prefers-reduced-motion value.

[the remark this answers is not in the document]
marcysutton: These are great ideas! Here are some comments that I meant to send yesterday:
- Making animations default to false if there is no context provider (and therefore no error on basic setup) makes total sense. You could make animations opt-in this way.
- Any time we can avoid using React context seems advantageous...for the same reasons I used a singleton class for the Announcer! But perhaps there are unique details about this state tracking piece that would make more sense to use Context.
- The site setting for reducing motion is useful in browsers where the OS-level setting isn't available. At this point it's so [well supported](https://caniuse.com/?search=prefers-reduced-motion) that it doesn't seem as critical...but that's just my personal opinion!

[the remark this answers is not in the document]
beaesguerra: Thanks for the discussion!

I created WB-1918 for creating a way to check if animations should be used in a component! That way we can explore this more separately! Any concerns if we include the `animated` prop for NavigationTabs for now? Or thoughts on the priority of the shared animated setting work?

__docs__/wonder-blocks-tabs/navigation-tabs-variants.stories.tsx
```diff
@@ -211,20 +211,52 @@ export default meta;
...
 
 export const Hover: Story = {
-    parameters: {pseudo: {hover: true}},
+    parameters: {
+        pseudo: {hover: true},
+        chromatic: {
+            // TODO(WB-1917): Temporarily disabled since snapshots with
+            // NavigationTabs and pseudo states are flaky. We still get state
+            // style coverage in NavigationTabItem AllVariants stories
+            disable: true,
```
jandrade, line 220: Maybe this should be `disableSnapshot` instead? I'm seeing the snapshots still being included here: https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=67e71a51097d57f7d1951deb

__docs__/wonder-blocks-tabs/navigation-tabs.stories.tsx
```diff
@@ -301,7 +302,16 @@ export const HeaderWithNavigationTabsExample: StoryComponentType = {
                 margin: `-${headerVerticalSpacing} 0`,
             },
         });
-
+        const [currentTab, setCurrentTab] = React.useState(0);
+        const tabs = Array(4)
```
jandrade, line 306: **praise:** Nice way of refactoring this story!

__docs__/wonder-blocks-tabs/navigation-tabs.stories.tsx
```diff
@@ -343,4 +344,100 @@ export const HeaderWithNavigationTabsExample: StoryComponentType = {
...
+    },
+};
+
+const transitionStyle = "left 0.3s ease, width 0.3s ease";
+
+/**
+ * The `animated` prop can be set to `true` to animate the current underline
+ * indicator.
+ */
+export const Interactive: StoryComponentType = {
```
jandrade, line 358: **thought:** How do you feel about changing the story title to something like `Animated` (or something else)? I feel that interactive doesn't fully convey the intention of the story.

__docs__/wonder-blocks-tabs/navigation-tabs.stories.tsx
```diff
@@ -343,4 +344,100 @@ export const HeaderWithNavigationTabsExample: StoryComponentType = {
...
+    args: {
+        animated: true,
+    },
+};
+
+const transitionStyle = "left 0.3s ease, width 0.3s ease";
+
+/**
+ * The `animated` prop can be set to `true` to animate the current underline
+ * indicator.
```
jandrade, line 356: **suggestion:** It would be good mentioning here, or at some point that animations are disabled by default, but that these can be turned on by using this prop.

packages/wonder-blocks-tabs/src/components/navigation-tabs.tsx
```diff
@@ -81,8 +94,97 @@ export const NavigationTabs = React.forwardRef(function NavigationTabs(
...
+
+            // When calculating, we divide by the zoom factor so the underline
+            // is proportional to the rest of the component
+            const left = (tabRect.left - parentRect.left) / zoomFactor; // Get position relative to parent
+            const width = tabRect.width / zoomFactor; // Use bounding width
+
+            setUnderlineStyle({
+                left,
+                width,
+            });
```
jandrade, line 149: **question:** Have you considered using `translateX(left)` for better performance? TBH I'm not sure if at this point there are any perf. differences between `left` and `transform`.

jandrade (commented):
This is taking great shape. The logic for calculating the underline is really solid!

__docs__/wonder-blocks-tabs/navigation-tabs.stories.tsx
beaesguerra: I've added it to the story docs for the `Animated` and `AnimationsDisabled` stories! It is also mentioned in the `animated` prop docs that it defaults to `false`

packages/wonder-blocks-tabs/src/components/navigation-tabs.tsx
beaesguerra: Thanks for the reminder!! I updated it to transition the `transform` property instead of `left`

jandrade (approved):
Looks really great! the logic for the animation is really neat! great job 👏 🚀

PR #2522, opened 2025-03-26
[WB-1911] Popover: Remove "emphasized" and "color" variants

jandrade:
## Summary:

Removed the `emphasized` props from `PopoverContent` and `color` prop from
`PopoverContentCore` as these are not currently used and no longer recommended.

Issue: https://khanacademy.atlassian.net/browse/WB-1911

## Test plan:

Navigate to the Popover docs and verify that the `emphasized` and `dark` stories
are no longer present.

`/?path=/docs/packages-popover-popovercontentcore--docs`
`/?path=/docs/packages-popover-popovercontent--docs`

### DIFF
--- a/packages/wonder-blocks-popover/src/components/popover-content-core.tsx
+++ b/packages/wonder-blocks-popover/src/components/popover-content-core.tsx
@@ -105,18 +105,6 @@ const styles = StyleSheet.create({
         overflow: "hidden",
         justifyContent: "center",
     },
-    /**
-     * Theming
-     */
-    blue: {
-        backgroundColor: semanticColor.surface.emphasis,
-        color: semanticColor.text.inverse,
-    },
-
-    darkBlue: {
-        backgroundColor: semanticColor.surface.inverse,
-        color: semanticColor.text.inverse,
-    },
 
     /**
      * elements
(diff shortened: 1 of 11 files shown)


packages/wonder-blocks-popover/src/components/popover-content-core.tsx
```diff
@@ -75,25 +68,20 @@ export default class PopoverContentCore extends React.Component<Props> {
...
             style,
             testId,
         } = this.props;
 
         return (
             <View
                 testId={testId}
-                style={[
-                    styles.content,
-                    color !== "white" && styles[color],
```
somewhatabstract, line 88: **question:** If `styles[color]` isn't being used here, are there any unused styles now to be removed?

somewhatabstract (approved):
LGTM

Just one question inline but nothing to get in the way of landing.

packages/wonder-blocks-popover/src/components/popover-content-core.tsx
jandrade: oh great point, yeah I should remove unused styles! thanks for catching that :)

PR #2523, opened 2025-03-26
Update Announcer docs

marcysutton:
## Summary:
The docs needed updating for usage in webapp and elsewhere. The API functions were outdated and poorly documented. This update will hopefully help with adoption!

Issue: FEI-6331

## Test plan:
1. Open docs in Storybook
2. Verify for accuracy and readability


__docs__/wonder-blocks-announcer/announcer.stories.tsx
```diff
@@ -36,29 +36,116 @@ const AnnouncerExample = ({
...
+ * To test the Announcer, turn on VoiceOver for Mac/iOS or NVDA on Windows and play with the [React example below](#react-announcer-example).
+ *
+ * By default, the `wbAnnouncer` container element is rendered _visually hidden_.
+ *
+ * > In Storybook, the Live Regions appended by Announcer are shown visually on the
+ * right side of the screen for debugging (see red outlined boxes). This can be
+ * controlled with a Storybook parameter, `addBodyClass: "showAnnouncer"`.
+ *
+ * ## API functions
+ * ### announceMessage()
```
beaesguerra, line 63: One suggestion would be to include the argument in the function signature so that it's clear it takes an options argument (since it's one object, rather than multiple parameters)!

```suggestion
 * ### announceMessage(options)
```

__docs__/wonder-blocks-announcer/announcer.stories.tsx
```diff
@@ -36,29 +36,116 @@ const AnnouncerExample = ({
...
+ *          }
+ *      }, [someState]);
+ *
+ *      return (
+ *          <div>Some content</div>
+ *      )
+ * }
+ * ```
+ *
+ * ### initAnnouncer()
```
beaesguerra, line 114: Same suggestion around including the args for `initAnnouncer` heading too!

__docs__/wonder-blocks-announcer/announcer.stories.tsx
marcysutton: Ooh, good call! And I forgot the required `message` option. Those are both in there now!

PR #2524, opened 2025-03-27
[WB-1902.1] IconButton: Update IconButton focus state to use box-shadow instead of outline

jandrade:
## Summary:

This PR updates the `IconButton` component to use a box-shadow for the focus
state instead of an outline. This change is made to improve accessibility and
visual consistency across the application.

Also, taking a spin at modifying the states to make them look closer to
ThunderBlocks.

Issue: WB-1902

## Test plan:

Verify that the IconButton component's focus state now uses a box-shadow instead
of an outline.


beaesguerra (commented):
Code changes look good! Left some questions in Chromatic that I'd like to get your thoughts on!

beaesguerra (approved):
Thanks for making these improvements!

PR #2525, opened 2025-03-28
[WB-1776] Styles: Add wonder-blocks-styles package

jandrade:
## Summary:

Created the `wonder-blocks-styles` package to include the focus styles for
Wonder Blocks (more styles will come later!).

This package will be used by other Wonder Blocks packages to ensure consistent
styles across the library. Also to provide a single source of truth for focus
styles in consumer applications.

Issue: https://khanacademy.atlassian.net/browse/WB-1776

## Test plan:

Navigate to `/?path=/docs/packages-styles-overview--docs` and verify that the
overview page looks fine.

Also navigate to `/?path=/docs/wonder-blocks-styles-styles--focus-styles` and
verify that the focus styles are applied correctly in both backgrounds (light
and dark).

### DIFF
--- a/.changeset/rotten-queens-sneeze.md
+++ b/.changeset/rotten-queens-sneeze.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-styles": minor
+---
+
+Add `styles` package. Create `focusStyles` for focus indicators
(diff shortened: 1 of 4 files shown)


jandrade:
I messed up this PR, so I recreated it in a separate one: #2527 2527

PR #2529, opened 2025-03-31
[WB-1902.2] IconButton: Update IconButton focus state to use box-shadow + outline for better a11y

jandrade:
## Summary:

In #2524, we updated the IconButton focus state to use a box-shadow instead of
an outline. This change was made to improve accessibility and provide a more
modern look to the component.

Turns out that the box-shadow is not 100% accessible for all users, as it
doesn't work well on Windows High Contrast mode. This PR modifies the focus
style to use the `focus` style from `wonder-blocks-styles`, which is a
combination of a box-shadow and an outline. This should provide a more
consistent experience across different platforms and improve accessibility.

Issue: WB-1902

## Test plan:

Navigate to the IconButton component in Storybook and verify that the focus
state looks correct.

`/?path=/story/packages-iconbutton-all-variants--focus&globals=viewport:desktop;pseudo.focusVisible:!true`

Also verify that the close button in the Modal component has the correct focus
state.

`/?path=/story/packages-modal-onepanedialog--with-launcher`

### DIFF
--- a/packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
+++ b/packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
@@ -8,6 +8,7 @@ import {isClientSideUrl} from "@khanacademy/wonder-blocks-clickable";
 import {PhosphorIcon} from "@khanacademy/wonder-blocks-icon";
 import {useScopedTheme} from "@khanacademy/wonder-blocks-theming";
 
+import {focusStyles} from "@khanacademy/wonder-blocks-styles";
 import type {
     IconButtonActionType,
     IconButtonSize,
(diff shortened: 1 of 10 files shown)


packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
```diff
@@ -234,9 +235,7 @@ function getStylesByKind(
                 borderStyle: "solid",
                 borderWidth: theme.border.width.hover,
             },
-            ":focus-visible": {
-                boxShadow: light ? theme.focus.inverse : theme.focus.default,
```
beaesguerra, line 238: Glad we can simplify the focus styling for the light/inverse variants!

beaesguerra (approved):
Looks good! Looking forward to using the shared styles more! 🚀

PR #2530, opened 2025-04-01
[WB-1852.1] IconButton: Remove `light` variant

jandrade:
## Summary:

This PR removes the `light` variant from the IconButton component and replaces
it with a new `actionStyles.inverse` style. This change is part of a larger
effort to standardize the styling of interactive elements across WB. The `light`
variant was previously used in the `Popover` and `Moda`l components. The new
`actionStyles.inverse` style provides a way to apply almost the same styling as
the `light` variant, but allow us to avoid having to maintain a separate `light`
variant in the IconButton component (which won't be supported in Thunder
Blocks and new styles).

Key changes:

- Styles: Added a new `actionStyles.inverse` style to apply to interactive
(action) components.
- IconButton: Removed the `light` variant and replaced it with
`actionStyles.inverse`.
- Modal and Popover: Updated dismiss button (CloseButton) to use
`actionStyles.inverse` instead of the now deprecated `IconButton.light` variant.

Issue: WB-1852

### IconButton

Navigate to the `IconButton` docs and verify that all the dark stories are no
longer present. Also verify that the "All variant" snapshots remove the `light`
variant.

- /?path=/docs/packages-iconbutton--docs
- /?path=/story/packages-iconbutton-all-variants--default

### Popover and Modal

Verify that the `Popover` and `Modal` components are still working as expected
and that the dismiss button (CloseButton) is using the new
`actionStyles.inverse` style.

- Popover: /?path=/docs/packages-popover-popovercontent--docs#with%20illustration
- Modal: /?path=/story/packages-modal-building-blocks-modalheader--dark&globals=viewport:desktop;

### Styles

Verify that the new `actionStyles.inverse` style is working as expected in the
`ActionStyles` docs.

- /?path=/docs/packages-styles-action-styles--docs

### DIFF
--- a/__docs__/wonder-blocks-styles/action-styles-variants.stories.tsx
+++ b/__docs__/wonder-blocks-styles/action-styles-variants.stories.tsx
@@ -52,12 +52,18 @@ export default {
                                 icon={info}
                                 style={actionStyles.inverse}
                             />
+                            <IconButton
+                                kind="primary"
+                                disabled
+                                icon={info}
+                                style={actionStyles.inverse}
+                            />
                         </View>
                     ),
                 },
             },
             {
-                name: "Using Clickable",
+                name: "Using Clickable (no border set)",
                 props: {
                     children: (
                         <Clickable
@@ -70,7 +76,7 @@ export default {
                 },
             },
             {
-                name: "Using Clickable with initial border",
+                name: "Using Clickable with border",
                 props: {
                     children: (
                         <Clickable
--- a/__docs__/wonder-blocks-styles/action-styles-variants.stories.tsx
+++ b/__docs__/wonder-blocks-styles/action-styles-variants.stories.tsx
@@ -100,6 +100,9 @@ export default {
                         <StyledButton
                             style={[
                                 {
+                                    // NOTE: Swapping the colors is intentional
+                                    // here to show the inverse version of the
+                                    // button.
                                     border: `1px solid ${semanticColor.status.success.background}`,
                                     backgroundColor:
                                         semanticColor.status.success.foreground,
--- a/__docs__/wonder-blocks-styles/action-styles-variants.stories.tsx
+++ b/__docs__/wonder-blocks-styles/action-styles-variants.stories.tsx
@@ -1,0 +1,155 @@
+import * as React from "react";
+import type {Meta, StoryObj} from "@storybook/react";
+import info from "@phosphor-icons/core/regular/info.svg";
+import {ScenariosLayout} from "../components/scenarios-layout";
+import {addStyle, View} from "@khanacademy/wonder-blocks-core";
+import {
+    border,
+    semanticColor,
+    spacing,
+} from "@khanacademy/wonder-blocks-tokens";
+import Clickable from "@khanacademy/wonder-blocks-clickable";
+import {actionStyles} from "@khanacademy/wonder-blocks-styles";
+import IconButton from "@khanacademy/wonder-blocks-icon-button";
+
+const StyledButton = addStyle("button");
+
+/**
+ * The following stories are used to generate the pseudo states for these
+ * stories. This is only used for visual testing in Chromatic.
+ */
+export default {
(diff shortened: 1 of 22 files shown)


packages/wonder-blocks-icon-button/src/themes/default.ts
```diff
@@ -37,7 +22,10 @@ const baseColorStates = {
             border: "transparent",
             background: "transparent",
         },
-        focus: focusOutline,
+        hover: {
```
beaesguerra, line 25: Question for later: Do we think global hover styles would be helpful too (similar to the global focus styles)? Or more on a component level basis?

From what I've noticed, we have a variety of hover styles: outline around the element, underline, left bar indicator. We could eventually have different hover styles in the styles package (ex: `hoverOutline`, `hoverUnderline`, etc). What do you think!

packages/wonder-blocks-icon-button/src/themes/default.ts
```diff
@@ -62,63 +53,14 @@ const baseColorStates = {
...
-        press: disabledLightStates,
-    },
 };
 
 const theme = {
     color: {
-        primary: {
-            ...baseColorStates,
-            // Only primary supports the light variants.
-            // TODO(WB-1852): Remove light variants.
```
beaesguerra, line 83: Nice! 🧹

__docs__/wonder-blocks-styles/action-styles-variants.stories.tsx
```diff
@@ -0,0 +1,155 @@
...
+            {
+                name: "Using an HTML element",
+                props: {
+                    children: (
+                        <StyledButton
+                            style={[
+                                {
+                                    border: `1px solid ${semanticColor.status.success.background}`,
+                                    backgroundColor:
+                                        semanticColor.status.success.foreground,
```
beaesguerra, line 99: Should `foreground` be used for `color` and `background` used for `backgroundColor`?

__docs__/wonder-blocks-styles/action-styles-variants.stories.tsx
```diff
@@ -0,0 +1,155 @@
...
+ */
+export default {
+    title: "Packages / Styles / Action Styles / Action Styles - All Variants",
+    parameters: {
+        docs: {
+            autodocs: false,
+        },
+    },
+    render: () => {
+        const scenarios = [
```
beaesguerra, line 29: Would it helpful to have scenarios that include the disabled state (and disabled + focus state)?

beaesguerra (commented):
Nice to see IconButton styling get simplified 😄 Left some questions in the PR and Chromatic!

__docs__/wonder-blocks-styles/action-styles-variants.stories.tsx
beaesguerra: A comment sounds good to me! So we know it's intentionally swapped 😄

packages/wonder-blocks-modal/src/components/modal-panel.tsx
```diff
@@ -126,16 +126,21 @@ export default function ModalPanel({
...
         >
             {closeButtonVisible && (
                 <CloseButton
-                    light={!light}
+                    light={isInverse}
                     onClick={onClose}
-                    style={styles.closeButton}
+                    style={[
+                        styles.closeButton,
+                        isInverse && actionStyles.inverse,
```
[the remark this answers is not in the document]
jandrade: now that I realize, I'm removing that in the next PR. Would it be ok if I apply this change there?

packages/wonder-blocks-icon-button/src/themes/default.ts
jandrade: Yeah!! and wait for the next PR to see how this theme file gets simplified 😄

beaesguerra (approved):
Looks good! Thanks for making these updates 🧹

PR #2526, opened 2025-03-28
[WB-1914] Update wonder-blocks-tokens to use Base 10

marcysutton:
## Summary:
This change updates the sizing tokens in wonder-blocks-tokens to use Base 10 rather than Base 8. We decided on this change as a team to work around some minimum issues in browsers where JavaScript returned the incorrect font-size on a page.

Issue: WB-1914

## Test plan:
1. Reference Base 10 scale in https://khanacademy.atlassian.net/wiki/spaces/WB/pages/3683483678/Sizing
2. Double-check math of values used
3. Ensure there are no lint errors with tokens used in WB
4. To-do: update usage in webapp (FEI-6229)

PR #2532, opened 2025-04-02
[WB-1927]: Update wonder-blocks-typography to use rems

marcysutton:
## Summary:

Update wonder-blocks-typography component styles to use REMs for font-sizing and line-heights. This PR follows https://github.com/Khan/wonder-blocks/pull/2526 where the base font size was changed to 62.5%.

Issue: WB-1927, FEI-6534

## Test plan:
1. Open Chromatic build for this PR.
2. Look for font-sizes and line-heights to stay the same at the default size, with a few exceptions:
    - `BodySerifBlock` now has a line-height of `24px` (it was `28px` before)
    - `BodySerif` now has a font-size of `16px` and line-height of `20px` to leverage t-shirt sizing (it was `18px` / `22px` before)
    - `BodyMonospace` now has a font-size of `16px` and line-height of `20px` (it was `17px` / `22px` before)
3. Double-check `TextArea` min-height and `Button-Core` line-heights are computed with actual values
4. Increase/decrease font-size in Chrome browser settings (Settings > Appearance > Font Size) and watch it change size.

Adjustments for REM sizing and reflow in webapp are covered under individual font-sizing issues: https://khanacademy.atlassian.net/issues/?jql=labels%20%3D%20%22webapp-font-sizing-rems%22


packages/wonder-blocks-tokens/src/tokens/sizing.ts
```diff
@@ -32,11 +32,15 @@ export const sizing = {
     size_120: pxToRem(12),
     size_140: pxToRem(14),
     size_160: pxToRem(16),
+    size_170: pxToRem(17),
```
marcysutton, line 35: Note: I added these extra tokens after testing in webapp. There are some additional font-sizes on the LOHP and other places that I wanted to include in WB for minimal impact on teams. There are likely to be more, as there is a very large surface area for font-sizing in components. We could try to convince teams to use the standard tokens, though it will be a lot of work to do so!

marcysutton:
I got this PR ready to review by decoupling other work from it, namely updating REMs in webapp! The direct work that will follow this PR is updating webapp with the `wonder-blocks-` packages. Hard-coded pixel font-sizing _and_ theming will be whole other workstreams.

packages/wonder-blocks-tokens/src/tokens/sizing.ts
jandrade: Actually, I see it the other way around. We should avoid adding these tokens to provide a more consistent experience, so I would strongly recommend not adding these tokens here and instead just changing the call sites to use WB tokens and/or leaving those as one-offs. The signal about those tokens being used is that those values were introduced when the design language was different around that time, and it would be best not to introduce these legacy tokens to WB.

packages/wonder-blocks-typography/src/util/styles.ts
```diff
@@ -23,132 +24,132 @@ const styles: StyleDeclaration = StyleSheet.create({
         fontFamily: SansFamily,
         fontWeight: Black,
         [desktop]: {
-            fontSize: 36,
-            lineHeight: "40px",
+            fontSize: sizing.size_360,
+            lineHeight: sizing.size_400,
```
jandrade, line 28: **suggestion:** This is something that I never did when creating the [font tokens](https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-tokens/src/tokens/font.ts), but I think it would be better to change these styles to use `font` tokens directly, then change those tokens to map to the `sizing` tokens instead. That way consumers that use the font tokens would benefit from this change as well.

jandrade (changes requested):
Thanks for introducing this change! As you are modifying these components, I think it would be great refactoring the `typography` styles to use `font` tokens instead, and then modifying `font` tokens to use the new `sizing` values.

packages/wonder-blocks-typography/src/util/styles.ts
marcysutton: Cool, makes sense! I'll take a look.

packages/wonder-blocks-tokens/src/tokens/sizing.ts
beaesguerra: For the one-offs, would we want to export the `pxToRem` utility so that custom values can scale in rems too?

static/sb-styles/preview.css
```diff
@@ -10,7 +10,7 @@ html {
     padding: 0.8rem;
 }
 .sbdocs.sbdocs-wrapper {
-    padding: 3.2rem 20px;
+    padding: 0 20px 3.2rem 20px;
```
beaesguerra, line 13: Do we use `px` instead of `rem` for the horizontal padding here so that it's fixed? 

This isn't specific to typography rems but got me thinking, would we want to provide a way for devs to use the `px` version of our tokens?

beaesguerra (commented):
I agree with Juan's feedback! Left some questions too 😄

packages/wonder-blocks-tokens/src/tokens/sizing.ts
marcysutton: Alright, I'll remove them from this PR and do something else for call-sites in webapp.

For the `pxToRem` utility, how would that be used if we did export it? Isn't that an internal tool for WB?

static/sb-styles/preview.css
marcysutton: Good question! I had to override the top padding so it didn't get overlapped. But it had pixels before, so I kept them there. I will switch them to rems!

packages/wonder-blocks-tokens/src/tokens/sizing.ts
marcysutton: Ah, in updating the font / sizing tokens I discovered 3 of these added values are in WB: `17px`, `22px`, and `26px`

- `HeadingMedium` at a small viewport size: font-size `22px` and line-height `26px`

```
  HeadingMedium: {
        [mobile]: {
            fontSize: sizing.size_220,
            lineHeight: sizing.size_260,
        },
    },
```

- `BodyMonospace`: font-size `17px` and line-height `22px` (this one is used a LOT in webapp)

```
BodyMonospace: {
        fontSize: sizing.size_170,
        lineHeight: sizing.size_220,
    },
```

- `Body`: line-height `22px` (also used a lot)

- `BodySerifBlock`: font-size `22px` but it's only used on an `sat-landing-package` / `summary-module.tsx` in webapp. Is that product used anymore?

beaesguerra: > For the pxToRem utility, how would that be used if we did export it? Isn't that an internal tool for WB?

Right now it is internal to WB since we don't export it!

If we were to export `pxToRem`, then consumers can import it and use it to convert px values to rem. Consumers should use the sizing tokens where they can, but if there are one off values not in the WB scale, they can convert it to rems using the utility which will use the same WB baseline (10). The utility would make it so if we have to change the rem baseline, the one-offs using the utility wouldn't need to change since it takes into account the WB defined baseline.

That said, exposing `pxToRem` could encourage more one-off values too. We don't have to export the utility, but thought I'd share the thought!

packages/wonder-blocks-tokens/src/tokens/font.ts
```diff
@@ -1,30 +1,46 @@
...
-        xLarge: 24,
-        large: 20,
-        medium: 16,
-        small: 14,
-        xSmall: 12,
+        xxxLarge: sizing.size_360,
+        xxLarge: sizing.size_280,
+        xLarge: sizing.size_240,
+        large: sizing.size_200,
+        medium: sizing.size_160,
```
marcysutton, line 27: This one will need to change in a TB theme file: Foundations uses `sizing.size_180` for `medium`: https://www.figma.com/design/9seilz0fBgo3QUj6eXM0yR/%E2%9A%A1%EF%B8%8F-Foundations?node-id=52-3&t=NFXovVlCFbYhQHvm-0

.changeset/clever-timers-whisper.md
```diff
@@ -0,0 +1,6 @@
+---
+"@khanacademy/wonder-blocks-typography": minor
+"@khanacademy/wonder-blocks-tokens": minor
```
marcysutton, line 3: Should these be majors as some of the font-sizes are changing? Or just minor?

packages/wonder-blocks-typography/src/util/styles.ts
```diff
@@ -1,154 +1,143 @@
...
     Title: {
         ...common,
-        fontFamily: SansFamily,
-        fontWeight: Black,
-        [desktop]: {
-            fontSize: 36,
-            lineHeight: "40px",
+        fontFamily: font.family.sans,
+        fontWeight: font.weight.black,
+        [breakpoint.mediaQuery.xl]: {
```
marcysutton, line 14: I did a bonus move of updating these breakpoints to use tokens. I could separate it out into a different PR if that is preferred!

packages/wonder-blocks-typography/src/util/styles.ts
```diff
@@ -1,154 +1,143 @@
...
+            fontSize: font.size.xLarge,
+            lineHeight: font.lineHeight.xLarge,
         },
-        [mobile]: {
-            fontSize: 22,
-            lineHeight: "26px",
+        [breakpoint.mediaQuery.mdOrSmaller]: {
+            fontSize: font.size.large,
+            lineHeight: font.lineHeight.large,
```
marcysutton, line 53: Note: `sizing.size_260` (`26px`) isn't part of our REM token scale, so I adjusted it to `font.lineHeight.large` (`24px`)

packages/wonder-blocks-typography/src/util/styles.ts
```diff
@@ -1,154 +1,143 @@
...
+        fontWeight: font.weight.bold,
+        [breakpoint.mediaQuery.xl]: {
+            fontSize: font.size.xLarge,
+            lineHeight: font.lineHeight.xLarge,
         },
-        [mobile]: {
-            fontSize: 22,
-            lineHeight: "26px",
+        [breakpoint.mediaQuery.mdOrSmaller]: {
+            fontSize: font.size.large,
```
marcysutton, line 52: **Note:** `sizing.size_220` isn't part of our REM token scale, so I adjusted it to `font.size.large` (`20px`)

packages/wonder-blocks-typography/src/util/styles.ts
```diff
@@ -1,154 +1,143 @@
...
     },
     BodySerifBlock: {
         ...common,
-        fontFamily: SerifFamily,
-        fontWeight: Regular,
-        fontSize: 22,
-        lineHeight: "28px",
+        fontFamily: font.family.serif,
+        fontWeight: font.weight.regular,
+        fontSize: font.size.large,
```
marcysutton, line 76: **Note:** like `HeadingMedium` at `< 1023px` wide, `22px` isn't part of our REM scale. I adjusted it to `font.size.large` (`20px`).

packages/wonder-blocks-typography/src/util/styles.ts
```diff
@@ -1,154 +1,143 @@
...
     BodyMonospace: {
         ...common,
-        fontFamily: InconsolataFamily,
-        fontWeight: Regular,
-        fontSize: 17,
-        lineHeight: "22px",
+        fontFamily: font.family.mono,
+        fontWeight: font.weight.regular,
+        fontSize: font.size.medium,
+        lineHeight: font.lineHeight.medium,
```
marcysutton, line 91: **Note:** `22px` also isn't on our REM scale so I converted this one to `font.lineHeight.medium` (`20px`)

packages/wonder-blocks-typography/src/util/styles.ts
```diff
@@ -1,154 +1,143 @@
...
     Body: {
         ...common,
-        fontFamily: SansFamily,
-        fontWeight: Regular,
-        fontSize: 16,
-        lineHeight: "22px",
+        fontFamily: font.family.sans,
+        fontWeight: font.weight.regular,
+        fontSize: font.size.medium,
+        lineHeight: font.lineHeight.medium,
```
marcysutton, line 98: **Note:** this line-height was updated to `font.lineHeight.medium` (`20px`) as `22px` isn't on our REM scale.

packages/wonder-blocks-typography/src/util/styles.ts
```diff
@@ -1,154 +1,143 @@
...
     },
     BodySerif: {
         ...common,
-        fontFamily: SerifFamily,
-        fontWeight: Regular,
-        fontSize: 18,
-        lineHeight: "22px",
+        fontFamily: font.family.serif,
+        fontWeight: font.weight.regular,
+        fontSize: font.size.medium,
```
marcysutton, line 83: **Note:** While `sizing.size_180` exists, it isn't part of our t-shirt sizing scale. Therefore I updated this to `font.size.medium` (`16px`) rather than have a one-off. `BodySerif` is only used in the WB docs as far as I can tell!

packages/wonder-blocks-typography/src/util/styles.ts
```diff
@@ -1,154 +1,143 @@
...
+        lineHeight: font.size.medium,
```
marcysutton, line 84: **Note:** Same as other entries - `22px` isn't on our REM scale, so I updated the line-height to `font.size.medium` (`20px`) here.

packages/wonder-blocks-tokens/src/tokens/sizing.ts
marcysutton: Added some tokens with necessary font-size changes here to avoid one-offs. I think we could keep chatting about `pxToRem` and tackle it in a future PR, potentially?

.changeset/clever-timers-whisper.md
jandrade: IMO it should be `minor` if only the values change. If the `font` structure changes (public API), then that could be considered as a breaking change.

TBH this is the kind of situation were any option could work :)

packages/wonder-blocks-tokens/src/tokens/font.ts
jandrade: Agreed! you might want to look at the CSS vars refactor that I've been doing + the TB theming PR, as probably most of this stuff will get moved to the `theme/` folder to convert these tokens as CSS variables + create the `tb` version of these tokens. We can chat later about this!

packages/wonder-blocks-typography/src/util/styles.ts
jandrade: **praise:** This is great! thanks for adding it 👏

packages/wonder-blocks-typography/src/util/styles.ts
jandrade: I think it would be better to keep it as one off for now, because it would use the same value defined for `HeadingSmall` below, and this would make these to headings look the same in small/medium screens.

```suggestion
            fontSize: `22px`,
```

packages/wonder-blocks-typography/src/util/styles.ts
jandrade: Same suggestion about keeping it as one-off for now to avoid having the same `lineHeight` defined in `HeadingSmall`.

packages/wonder-blocks-typography/src/util/styles.ts
jandrade: This one should be fine as there are no different sizes for `BodySerif`.

packages/wonder-blocks-typography/src/util/styles.ts
jandrade: This would probably affect a lot the vertical rhythm in a lot of places in the website. I'd suggest checking with Caitlyn in case she has thoughts about this change.

.changeset/clever-timers-whisper.md
marcysutton: Alright, I'll leave it as-is!

packages/wonder-blocks-typography/src/util/styles.ts
marcysutton: Alright, I checked in via Slack! Thanks for the suggestion!

packages/wonder-blocks-typography/src/util/styles.ts
```diff
@@ -49,8 +49,8 @@ const styles: StyleDeclaration = StyleSheet.create({
             lineHeight: font.lineHeight.xLarge,
         },
         [breakpoint.mediaQuery.mdOrSmaller]: {
-            fontSize: font.size.large,
-            lineHeight: font.lineHeight.large,
+            fontSize: "22px",
+            lineHeight: "26px",
```
jandrade, line 53: Now that I think, this could be better left as rem unit (still one off for now) so it is consistent with the other new rem tokens? wdyt?

Sorry for going back and forth with this.

marcysutton: Oh yeah, that totally makes sense! Thank you for the suggestion. I'll update it!

packages/wonder-blocks-typography/src/util/styles.ts
```diff
@@ -1,154 +1,143 @@
...
     Body: {
         ...common,
-        fontFamily: SansFamily,
-        fontWeight: Regular,
-        fontSize: 16,
-        lineHeight: "22px",
+        fontFamily: font.family.sans,
+        fontWeight: font.weight.regular,
+        fontSize: font.size.medium,
+        lineHeight: "2.2rem",
```
marcysutton, line 98: **Note:** I reset this one to 2.2rem (22px) to avoid unexpected layout shifts.

packages/wonder-blocks-tokens/src/tokens/font.ts
```diff
@@ -1,30 +1,46 @@
...
-        xLarge: 28,
-        large: 24,
-        medium: 20,
-        small: 18,
-        xSmall: 16,
+        xxxLarge: sizing.size_400,
+        xxLarge: sizing.size_320,
+        xLarge: sizing.size_280,
+        large: sizing.size_240,
+        medium: sizing.size_200,
```
jandrade, line 36: **issue:** It looks like this change could be breaking `TextArea`: https://github.com/Khan/wonder-blocks/blob/e63adea876333afa4ebba54da0fcdfdab9ce7236/packages/wonder-blocks-form/src/components/text-area.tsx#L347

That style will need to be changed for a different value.

jandrade, line 36: **issue:** Another part that is breaking the styles is here: https://github.com/Khan/wonder-blocks/blob/38c926cc45c36f7ccbc60413b101a38c96993338/packages/wonder-blocks-button/src/components/button-core.tsx#L270-L280

The `px` unit needs to be removed.

Also the `Button` theme needs to change in order to avoid computing the value here (now that we use rem units): https://github.com/Khan/wonder-blocks/blob/38c926cc45c36f7ccbc60413b101a38c96993338/packages/wonder-blocks-button/src/components/button-core.tsx#L270-L280

jandrade (commented):
This is looking great. I just found a couple of places that would need to be updated before landing (found that in the Chromatic snapshots).

https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=68017455ec38d6ca6f11b063

packages/wonder-blocks-tokens/src/tokens/font.ts
marcysutton: Good catch! Since we're moving to REM units for font-sizing/line-heights, should this `minHeight` value adjust along with them or stay in pixels?

.changeset/clever-timers-whisper.md
beaesguerra: Since we are using `sizing` tokens now for the `font.size` and `font.lineHeight` tokens, the token values will be strings now including the `rem` unit, instead of the numeric value. Would this be a breaking change since something like `fontSize: ${font.size.small}px` would need to be updated so it doesn't become `font-size: 1.4rempx`?

(Or, depending on if the css variable changes are released first, it'd become `font-size: var(--wb-token-name)px`)

beaesguerra (commented):
Left a question around whether the changes are breaking changes since we're using semantic tokens (with units in the value) now within the font tokens! https://github.com/Khan/wonder-blocks/pull/2532#discussion_r2049714313

packages/wonder-blocks-tokens/src/tokens/font.ts
marcysutton: I also took a first swing at the `TextArea` change by converting the sizing token with `remToPx` (a motivated use case for exporting it, along with `pxToRem`!). 

This raises a question: does the `TextArea` theme really need those computed values, like this:

```
minHeight: `${VERTICAL_SPACING_PX * 2 + remToPx(font.lineHeight.medium) + 2}px`
```

These seem like they could be simplified, as `VERTICAL_SPACING_PX` is always `10` and there isn't really anything dynamic going on when I look at it closely. I would suggest we add the values up and convert to REM units. If needed, this `minHeight` / `padding` in TextArea could be abstracted into theme configuration.

marcysutton: Good call! I removed those `px` units. There was also a hard-coded pixel value in the `Button` theme that I converted to REM sizing, so they could still be computed. This prompted me to check the codebase for other pixel units in template literals, and I found one in the docs!

.changeset/clever-timers-whisper.md
marcysutton: Great question. That does seem breaking to me as it will have to be updated at the call-sites to prevent an issue. I see numerous instances of `px` added after `font.size` and `font.lineHeight` tokens in webapp!

packages/wonder-blocks-button/src/themes/default.ts
```diff
@@ -339,12 +339,12 @@ const theme = {
...
-            large: 18,
+            large: "1.8rem",
         },
         lineHeight: {
-            small: tokens.font.lineHeight.small + textUnderlineOffset,
-            default: tokens.font.lineHeight.medium + textUnderlineOffset,
-            large: tokens.font.lineHeight.medium + 2 + textUnderlineOffset,
+            small: tokens.font.lineHeight.xMedium,
+            default: tokens.font.lineHeight.large,
+            large: tokens.font.lineHeight.large,
```
marcysutton, line 347: So this has taken some fiddling as you can't compute REM units by adding them together. I wonder if we can land on specific values instead of adding magic numbers together, since the `textUnderlineOffset` is constant. The extra `2` in `font.lineHeight.large` is very much a one-off, so I opted to remove it.

One alternative idea would be to create a REM addition function that removes the `rem` from the units, adds them together, and puts `rem` back on the end. Or we could create a one-off rem value in this file?

jandrade: I thin one-off (including comment) makes total sense. I had to do something similar in a separate PR for the same reason.

packages/wonder-blocks-tokens/src/tokens/font.ts
marcysutton: This was still broken so I refactored it a little, namely adding a flag to remToPx to optionally leave off the unit. I also moved those sizing utility functions into a `util` folder so they could be tested, and separated out the color-utils!

packages/wonder-blocks-button/src/themes/default.ts
marcysutton: I added a one-off unit in there!

packages/wonder-blocks-tokens/src/util/sizing-utils.ts
```diff
@@ -0,0 +1,42 @@
...
+    includeUnit?: true,
+    baseline?: number,
+): string;
+export function remToPx(
+    value: string,
+    includeUnit: false,
+    baseline?: number,
+): number;
+
+export function remToPx(
```
marcysutton, line 24: @jandrade I made a change to these CSS utilities to make them easier to use in compute functions. Can you take a look? I used an overload and a param for including the unit, but it could be a different function instead if that's preferred. I figured having one export that's configurable would be nice for ease of use.

packages/wonder-blocks-button/src/themes/default.ts
```diff
@@ -339,12 +339,13 @@ const theme = {
...
             // NOTE: This token is specific to this button size.
-            large: 18,
+            large: "1.8rem",
         },
         lineHeight: {
-            small: tokens.font.lineHeight.small + textUnderlineOffset,
-            default: tokens.font.lineHeight.medium + textUnderlineOffset,
-            large: tokens.font.lineHeight.medium + 2 + textUnderlineOffset,
+            small: tokens.font.lineHeight.xMedium,
+            default: tokens.font.lineHeight.large,
```
marcysutton, line 346: `default` button line-height is now `sizing.size_240`.  (2.4rem / 24px)

Previously it used `tokens.font.lineHeight.medium` (`20px`) + `textUnderlineOffset` (`4`) = `24px` so it should be equivalent.

packages/wonder-blocks-tokens/src/util/sizing-utils.ts
marcysutton: Update: I simplified this function back to the way it was, since the approach I took wasn't going to work with CSS variables! I kept the function in a separate utils file to make it more portable and testable.

.changeset/clever-timers-whisper.md
jandrade, line 1: **suggestion:** it might be good adding `patch` changes to `button` and `form` now that there are `sizing` uses in there.

jandrade (approved):
This looks great!! thanks so much for making all the final changes and improving the utils in the tokens package 👏 🚀 

The only thing to figure out now is fixing the "Unit tests" check, but the rest looks really solid.

.changeset/clever-timers-whisper.md
marcysutton: Good call! I was just looking at the changeset to finalize it. It seems like `wonder-blocks-tokens` might be a major based on the unit output changing, where `wonder-blocks-typography` could be a minor?

PR #2536, opened 2025-04-03
Tabs: Set up keyboard navigation and activationMode prop

beaesguerra:
### Tabs:
Implement the focus management and keyboard interactions from the [ARIA APG Tabs Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/)
- Add keyboard support for Tabs (left arrow, right arrow, home, and end keys)
- Add `activationMode` prop to choose between manual and automatic activation (defaults to `manual`)
- Make the tabpanel focusable if there are no focusable elements in it
- Make only the active tab focusable

### Core:
- Copy focus utilities to core so more components can use them
- Add more key constants for keyboard events

Issue: WB-1909

## Test plan:

- Test the keyboard interactions and focus management in the following stories:
  - `?path=/story/packages-tabs-tabs-tabs--manual-activation`
  - `?path=/story/packages-tabs-tabs-tabs--automatic-activation`
  - `?path=/story/packages-tabs-tabs-tabs--with-focusable-panel-content`
- Also test with screen readers
- Review a11y docs `?path=/docs/packages-tabs-tabs-accessibility--docs`

## Implementation plan
- Set up base component https://github.com/Khan/wonder-blocks/pull/2528
- Keyboard interactions https://github.com/Khan/wonder-blocks/pull/2536
- Remaining props
- Styles

### DIFF
--- a/__docs__/wonder-blocks-tabs/tabs.stories.tsx
+++ b/__docs__/wonder-blocks-tabs/tabs.stories.tsx
@@ -4,6 +4,9 @@ import ComponentInfo from "../components/component-info";
 import packageConfig from "../../packages/wonder-blocks-form/package.json";
 import {TabItem, Tabs} from "@khanacademy/wonder-blocks-tabs";
 import argTypes from "./tabs.argtypes";
+import Button from "@khanacademy/wonder-blocks-button";
+import Link from "@khanacademy/wonder-blocks-link";
+import {TextField} from "@khanacademy/wonder-blocks-form";
 
 const tabs: TabItem[] = [
     {label: "Tab 1", id: "tab-1", panel: <div>Tab contents 1</div>},
@@ -91,10 +94,41 @@ export const AutomaticActivation: StoryComponentType = {
  */
 export const WithFocusablePanelContent: StoryComponentType = {
     args: {
+        selectedTabId: "tab-wb-button",
         tabs: [
             {
+                label: "Content with WB Button",
+                id: "tab-wb-button",
+                panel: (
+                    <div>
+                        Tab contents with button{" "}
+                        <Button>Focusable Button</Button>
+                    </div>
+                ),
+            },
+            {
+                label: "Content with WB Link",
+                id: "tab-wb-link",
+                panel: (
+                    <div>
+                        Tab contents with link{" "}
+                        <Link href="#link">Focusable Link</Link>
+                    </div>
+                ),
+            },
+            {
+                label: "Content with WB TextField",
+                id: "tab-wb-textfield",
+                panel: (
+                    <div>
+                        Tab contents with WB TextField{" "}
+                        <TextField value="" onChange={() => {}} />
+                    </div>
+                ),
+            },
+            {
                 label: "Content with button",
-                id: "tab-1",
+                id: "tab-button",
                 panel: (
                     <div>
                         Tab contents with button{" "}
@@ -104,7 +138,7 @@ export const WithFocusablePanelContent: StoryComponentType = {
             },
             {
                 label: "Content with link",
-                id: "tab-2",
+                id: "tab-link",
                 panel: (
                     <div>
                         Tab contents with link{" "}
@@ -114,7 +148,7 @@ export const WithFocusablePanelContent: StoryComponentType = {
             },
             {
                 label: "Content with input",
-                id: "tab-3",
+                id: "tab-input",
                 panel: (
                     <div>
                         Tab contents with input <input type="text" />
@@ -123,7 +157,7 @@ export const WithFocusablePanelContent: StoryComponentType = {
             },
             {
                 label: "Content with no focusable elements",
-                id: "tab-4",
+                id: "tab-no-focusable-elements",
                 panel: <div>No focusable elements. Tab panel is focusable</div>,
             },
         ],
--- a/__docs__/wonder-blocks-tabs/tabs.stories.tsx
+++ b/__docs__/wonder-blocks-tabs/tabs.stories.tsx
@@ -49,3 +49,44 @@ type StoryComponentType = StoryObj<typeof Tabs>;
 export const Default: StoryComponentType = {
     args: {},
 };
+
+export const WithFocusablePanelContent: StoryComponentType = {
+    args: {
+        tabs: [
+            {
+                label: "Content with button",
+                id: "tab-1",
+                panel: (
+                    <div>
+                        Tab contents with button{" "}
+                        <button>Focusable Button</button>
+                    </div>
+                ),
+            },
+            {
+                label: "Content with link",
+                id: "tab-2",
+                panel: (
+                    <div>
+                        Tab contents with link{" "}
+                        <a href="#link">Focusable Link</a>
+                    </div>
+                ),
+            },
+            {
+                label: "Content with input",
+                id: "tab-3",
+                panel: (
+                    <div>
+                        Tab contents with input <input type="text" />
+                    </div>
+                ),
+            },
+            {
+                label: "Content with no focusable elements",
+                id: "tab-4",
+                panel: <div>No focusable elements. Tab panel is focusable</div>,
+            },
+        ],
+    },
+};
--- a/__docs__/wonder-blocks-tabs/tabs.stories.tsx
+++ b/__docs__/wonder-blocks-tabs/tabs.stories.tsx
@@ -85,9 +85,9 @@ export const AutomaticActivation: StoryComponentType = {
 
 /**
  * When a tab panel has focusable elements, pressing `Tab` from the tablist
- * will move focus to the first focusable element in the tab. If there are no
- * focusable elements in the active panel, the tabpanel will be focused on
- * instead.
+ * will move focus to the first focusable element in the tab panel. If there
+ * are no focusable elements in the active tab panel, the tab panel will be
+ * focused instead.
  */
 export const WithFocusablePanelContent: StoryComponentType = {
     args: {
--- a/__docs__/wonder-blocks-tabs/tabs.stories.tsx
+++ b/__docs__/wonder-blocks-tabs/tabs.stories.tsx
@@ -92,7 +92,7 @@ export const AutomaticActivation: StoryComponentType = {
  * are no focusable elements in the active tab panel, the tab panel will be
  * focused instead.
  */
-export const WithFocusablePanelContent: StoryComponentType = {
+export const WithFocusableContent: StoryComponentType = {
     args: {
         selectedTabId: "tab-wb-button",
         tabs: [
@@ -122,7 +122,11 @@ export const WithFocusablePanelContent: StoryComponentType = {
                 panel: (
                     <div>
                         Tab contents with WB TextField{" "}
-                        <TextField value="" onChange={() => {}} />
+                        <TextField
+                            value=""
+                            onChange={() => {}}
+                            aria-label="Focusable TextField"
+                        />
                     </div>
                 ),
             },
@@ -151,7 +155,8 @@ export const WithFocusablePanelContent: StoryComponentType = {
                 id: "tab-input",
                 panel: (
                     <div>
-                        Tab contents with input <input type="text" />
+                        Tab contents with input{" "}
+                        <input type="text" aria-label="Focusable input" />
                     </div>
                 ),
             },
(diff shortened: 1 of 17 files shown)


packages/wonder-blocks-tabs/src/components/tab-panel.tsx
```diff
@@ -13,20 +16,55 @@ type Props = {
...
 };
 
+const StyledDiv = addStyle("div");
 /**
  * A component that has `role="tabpanel"` and is used to represent a tab panel
  * in a tabbed interface.
  */
-export const TabPanel = React.forwardRef(function TabPanel(
-    props: Props,
-    ref: React.ForwardedRef<HTMLDivElement>,
```
beaesguerra, line 24: Removed support for forwarding a ref so we can use a ref internally to figure out if the tab panel has focusable elements in it. `TabPanel` is an internal component anyways used by `Tabs`, so consumers won't need to set a ref on the tab panel (if they needed to, they could put a ref on the contents of the panel, which they provide!)

packages/wonder-blocks-core/src/util/focus.ts
```diff
@@ -0,0 +1,20 @@
+/**
```
beaesguerra, line 1: Copied this over to the core package from the [popover utilities](https://github.com/Khan/wonder-blocks/blob/42a238515fa5cbaa1eb460f21819bcb9413223de/packages/wonder-blocks-popover/src/util/util.ts) so that we can use it in the Tabs component.

Created https://khanacademy.atlassian.net/browse/WB-1920 for consolidating the focus utilities into the core package (the modal package also has focus utilities!)

packages/wonder-blocks-core/src/util/focus.ts
```diff
@@ -0,0 +1,20 @@
...
+ * List of elements that can be focused
+ * @see https://www.w3.org/TR/html5/editing.html#can-be-focused
+ */
+const FOCUSABLE_ELEMENTS =
+    'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])';
```
marcysutton, line 6: I realize this probably already existed...but should `[contenteditable]` go on this list? And maybe the utility could all interactive elements for `tabindex="-1"` and filter them out if present?

packages/wonder-blocks-tabs/src/components/__tests__/tab-panel.test.tsx
```diff
@@ -125,5 +146,59 @@ describe("TabPanel", () => {
...
+                            {element}
+                        </TabPanel>,
+                    );
+                    const tabPanel = await screen.findByRole("tabpanel");
+
+                    // Act
+                    await userEvent.tab();
+
+                    // Assert
+                    expect(tabPanel).not.toHaveFocus();
```
marcysutton, line 199: Should this assert which element does have focus, to ensure the contents are focusable? Maybe in a separate set of tests?

packages/wonder-blocks-tabs/src/components/__tests__/tab.test.tsx
```diff
@@ -176,5 +197,39 @@ describe("Tab", () => {
...
+                // Arrange
+                render(
+                    <Tab {...props} selected={false}>
+                        Tab
+                    </Tab>,
+                );
+                const tab = await screen.findByRole("tab");
+
+                // Act
+                await userEvent.keyboard("{tab}");
```
marcysutton, line 228: I'm realizing the word `tab` can mean so many different things! 😅

Should we standardize on `userEvent.tab()` vs. `userEvent.keyboard("{tab}")` or `userEvent.keyboard("{Tab}")`?

packages/wonder-blocks-tabs/src/components/tab-panel.tsx
```diff
@@ -13,20 +16,55 @@ type Props = {
...
+            tabIndex={hasFocusableElement ? undefined : 0}
+        >
             {children}
-        </div>
+        </StyledDiv>
     );
+};
+
+const styles = StyleSheet.create({
+    hidden: {
```
marcysutton, line 67: This would be a good candidate for `wonder-blocks-styles`! It could also be defined with the HTML `hidden` attribute (e.g. `<div hidden>`).

packages/wonder-blocks-tabs/src/components/tabs.tsx
```diff
@@ -86,8 +96,90 @@ export const Tabs = React.forwardRef(function Tabs(
...
+                onTabSelected(tabId);
+            }
+        },
+        [onTabSelected, selectedTabId],
+    );
+
+    const handleKeyInteraction = React.useCallback(
+        (tabId: string) => {
+            // Move focus to the tab
+            const tabElement = document.getElementById(getTabId(tabId));
```
marcysutton, line 121: I'm trying to think through a way of getting the tabElement without relying on DOM methods. Maybe create an array of refs and filter based on a matching id? The DOM method might just be easier and more performant, though...

packages/wonder-blocks-tabs/src/components/tabs.tsx
```diff
@@ -86,8 +96,90 @@ export const Tabs = React.forwardRef(function Tabs(
...
+    );
+
+    const handleKeyDown = React.useCallback(
+        (event: React.KeyboardEvent<HTMLButtonElement>) => {
+            const currentIndex = tabs.findIndex(
+                (tab) => tab.id === focusedTabId.current,
+            );
+
+            switch (event.key) {
+                case keys.left: {
```
marcysutton, line 149: This could be a future enhancement, but do you think the tabs will ever stack vertically on small viewport sizes? Or will they always lay horizontally? I'm wondering: should the ArrowUp / ArrowDown keys do anything?

marcysutton (approved):
This is great work, Bea! I left a few comments and questions, nothing blocking.

packages/wonder-blocks-tabs/src/components/tabs.tsx
```diff
@@ -86,8 +96,90 @@ export const Tabs = React.forwardRef(function Tabs(
...
+        (tabId: string) => {
+            if (tabId !== selectedTabId) {
+                // Select the tab only if it's not already selected
+                onTabSelected(tabId);
+            }
+        },
+        [onTabSelected, selectedTabId],
+    );
+
+    const handleKeyInteraction = React.useCallback(
```
jandrade, line 118: **issue(blocking):** Probably related to @marcysutton's comment below....

This logic seems to be breaking the keyboard logic when there are multiple `Tabs` components within the same page.

If you test this behavior in https://5e1bf4b385e3fb0020b7073c-tbnxlswopz.chromatic.com/?path=/docs/packages-tabs-tabs-tabs--docs, you'll see that if you focus in a `Tabs` instance down in the page and press the arrow keys, then the focus will be moved to the next `Tab` element in the first `Tabs` instance.

__docs__/wonder-blocks-tabs/tabs.stories.tsx
```diff
@@ -49,3 +49,89 @@ type StoryComponentType = StoryObj<typeof Tabs>;
...
+    },
+};
+
+/**
+ * When a tab panel has focusable elements, pressing `Tab` from the tablist
+ * will move focus to the first focusable element in the tab panel. If there
+ * are no focusable elements in the active tab panel, the tab panel will be
+ * focused instead.
+ */
+export const WithFocusablePanelContent: StoryComponentType = {
```
jandrade, line 92: **question:** I'm not sure if I'm missing something with this story but when I test it, the focus is not moved to the active tab panel that contains a focusable element.

__docs__/wonder-blocks-tabs/tabs.stories.tsx
```diff
@@ -49,3 +49,89 @@ type StoryComponentType = StoryObj<typeof Tabs>;
...
+    args: {
+        tabs: [
+            {
+                label: "Content with button",
+                id: "tab-1",
+                panel: (
+                    <div>
+                        Tab contents with button{" "}
+                        <button>Focusable Button</button>
+                    </div>
```
jandrade, line 102: **suggestion:** It would be nice using WB components to start testing how focus behaviors would be handled with our own custom implementations. Probably good to still keep a tab with native elements.

packages/wonder-blocks-core/src/util/focus.ts
beaesguerra: Thanks for the suggestion! I made a note of this in WB-1920 so we can make the utility more robust and test it works as expected with the other components that would use this utility too! 

@marcysutton Could you elaborate on the second part? I'll clarify it in the ticket as well:

> And maybe the utility could all interactive elements for tabindex="-1" and filter them out if present?

packages/wonder-blocks-tabs/src/components/__tests__/tab.test.tsx
beaesguerra: Good call! Went with using `userEvent.tab()`!

packages/wonder-blocks-tabs/src/components/tab-panel.tsx
beaesguerra: Updated to use the `hidden` attribute instead (TIL!). I also made a note in https://khanacademy.atlassian.net/browse/WB-1776 that it'd be helpful to also have global `display: none` styles

packages/wonder-blocks-tabs/src/components/tabs.tsx
beaesguerra: Good question! I asked design about the responsive behaviour of tabs and we are having them be horizontally scrollable ([Slack thread](https://khanacademy.slack.com/archives/C07CA2YR0BZ/p1740082550098109?thread_ts=1740071221.168509&cid=C07CA2YR0BZ)).

The `ArrowUp` and `ArrowDown` keys don't do anything based on the [ARIA APG Tabs Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/):

> If the tab list is horizontal, it does not listen for Down Arrow or Up Arrow so those keys can provide their normal browser scrolling functions even when focus is inside the tab list.

I've seen some implementations where the DownArrow will move focus to the tabpanel though. I decided to stick with the ARIA APG pattern instead. Let me know though if you have any suggestions!

packages/wonder-blocks-tabs/src/components/tabs.tsx
beaesguerra: Updated! I was originally using refs, but then I realized I had the unique tab id already so I thought I'd use that! 😅 I forgot that I'm using the same tab id for the stories and that they all get put on the same page on the Storybook docs page. Switched it back to using `refs` so it is more robust!

__docs__/wonder-blocks-tabs/tabs.stories.tsx
beaesguerra: Added tabs with WB components! While testing these updates, I also fixed a bug where we needed to reset the focusTabId to the selected tab whenever focus moves out of the tablist!

__docs__/wonder-blocks-tabs/tabs.stories.tsx
jandrade: Ahh I see, thanks for clarifying :)

jandrade (approved):
Looks and works great!! thanks for applying the recent changes 🚢

packages/wonder-blocks-core/src/util/focus.ts
marcysutton: Sorry for the delay on this... my question was around whether elements removed from the tab order with `tabIndex="-1"` should still be considered focusable. For example:

```html
<a href="#" tabIndex="-1"></a>
<button tabIndex="-1"></button>
```
These elements are technically focusable with JavaScript, but not manually with the keyboard. 

Also, hidden inputs:
```
<input type="hidden" />
```

Curious to hear your thoughts on the tabIndex part in particular!

beaesguerra: Great points! I added these considerations to the ticket as well.

Hmmm I'm not sure about the tabIndex part yet, I think it will depend on how we use the utility. For the Tabs component, I think we shouldn't include elements with tabIndex=-1 since users can't focus on it manually with the keyboard. So if there's nothing inside a tab panel that the user can focus on themselves, then the tabpanel should be made focusable so it's in the tab order! Let me know what you think 😄

marcysutton: Makes sense! Looking at this line of code now, I see how it would depend on the way the utility is used (as you said). Like, are those selectors exclusively checked or does the `[tabindex]:not([tabindex="-1"])` part apply to the other selectors, in the end? Lots of test permutations to write for that kind of utility... it would be a really handy thing to have in our toolbox!

PR #2539, opened 2025-04-05
Upgrade WB to using react-router-dom-v5-compat.

jeresig:
## Summary:

This goes along with this PR: https://github.com/Khan/webapp/pull/31190 it switches all of the components that do routing over to using react-router-dom v5 + react-router-dom-v5-compat, enabling them to work in both v5 and v6 while we're in this transitory period.

Issue: FEI-6356

## Test plan:
I've confirmed that all tests are passing. I've also confirmed that they worked in-product, see the other PR for more details.

### DIFF
--- a/packages/wonder-blocks-button/src/components/button-core.tsx
+++ b/packages/wonder-blocks-button/src/components/button-core.tsx
@@ -238,7 +238,7 @@ const themedSharedStyles: ThemedStylesFn<ButtonThemeContract> = (theme) => ({
         border: "none",
         borderRadius: theme.border.radius.default,
         cursor: "pointer",
-        //outline: "none",
+        outline: "none",
         textDecoration: "none",
         boxSizing: "border-box",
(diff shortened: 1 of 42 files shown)


jandrade (commented):
This looks great!! thanks for working on this so quickly 🚀

jandrade (approved):

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -250,7 +238,7 @@ const themedSharedStyles: ThemedStylesFn<ButtonThemeContract> = (theme) => ({
         border: "none",
         borderRadius: theme.border.radius.default,
         cursor: "pointer",
-        outline: "none",
+        //outline: "none",
```
somewhatabstract, line 241: **question:** Why is this commented out?

somewhatabstract (approved):
Thanks for updating this! 

LGTM though there's possibly a commented out style that needs to be put back?

PR #2540, opened 2025-04-07
re-enable snapshots to see if they work in chromatic now

beaesguerra:
## Summary:

Testing to see if snapshots are still flaky

Issue: WB-XXXX

## Test plan:


beaesguerra:
Addressed the flakiness by making the tab indicator styles inline, see https://github.com/Khan/wonder-blocks/pull/2568#discussion_r2056598708

PR #2545, opened 2025-04-08
Tabs: Styling

beaesguerra:
## Summary:

- Implement styling for the Tabs component
- Add support for `styles` prop
- Implement the selected tab indicator animation with `animated` prop
  - Refactored NavigationTabs tab indicator logic into a hook so it can be shared

Issue: WB-1908

## Test plan:

- styles look as expected
  - Scenarios (including scrolling behaviour) (`?path=/story/packages-tabs-tabs--scenarios`)
  - Tab All Variants (`?path=/story/packages-tabs-tabs-subcomponents-tab-all-variants--default`)
  - Tabs All Variants (states + zoom) (`?path=/story/packages-tabs-tabs-tabs-all-variants--default`)
- `animated` prop works as expected (`?path=/story/packages-tabs-tabs--animated`)
- `styles` prop works as expected (`?path=/story/packages-tabs-tabs--custom-styles`)
- NavigationTabs animations continues to work as expected

## Implementation Plan
- Set up base component https://github.com/Khan/wonder-blocks/pull/2528
- Keyboard interactions https://github.com/Khan/wonder-blocks/pull/2536
- Remaining props https://github.com/Khan/wonder-blocks/pull/2542
- Styles https://github.com/Khan/wonder-blocks/pull/2545

### DIFF
--- a/__docs__/wonder-blocks-tabs/tabs-variants.stories.tsx
+++ b/__docs__/wonder-blocks-tabs/tabs-variants.stories.tsx
@@ -109,6 +109,7 @@ const generateRows = (rtl: boolean = false) => [
                     ),
                     id: "tab-1",
                     panel: rtl ? rtlText : "Tab 1 Contents",
+                    "aria-label": "Tab 1",
                 },
                 {
                     label: (
@@ -119,6 +120,7 @@ const generateRows = (rtl: boolean = false) => [
                     ),
                     id: "tab-2",
                     panel: rtl ? rtlText : "Tab 2 Contents",
+                    "aria-label": "Tab 2",
                 },
             ],
             selectedTabId: "tab-1",
@@ -188,4 +190,30 @@ export const HoverFocus: Story = {
 
 export const Press: Story = {
     parameters: {pseudo: {hover: true, active: true}},
+};
+
+export const Zoom: Story = {
+    render: (args) => (
+        <>
+            <AllVariants rows={rows} columns={columns} layout="list">
+                {(props) => (
+                    <View>
+                        <Tabs {...args} {...props} />
+                    </View>
+                )}
+            </AllVariants>
+            <div dir="rtl">
+                <AllVariants rows={rtlRows} columns={columns} layout="list">
+                    {(props) => (
+                        <View>
+                            <Tabs {...args} {...props} />
+                        </View>
+                    )}
+                </AllVariants>
+            </div>
+        </>
+    ),
+    globals: {
+        zoom: "400%",
+    },
 };
--- a/__docs__/wonder-blocks-tabs/tabs-variants.stories.tsx
+++ b/__docs__/wonder-blocks-tabs/tabs-variants.stories.tsx
@@ -1,0 +1,191 @@
+import type {Meta, StoryObj} from "@storybook/react";
+import * as React from "react";
+
+import {Tabs} from "@khanacademy/wonder-blocks-tabs";
+import {AllVariants} from "../components/all-variants";
+import {PhosphorIcon} from "@khanacademy/wonder-blocks-icon";
+import {IconMappings} from "../wonder-blocks-icon/phosphor-icon.argtypes";
+import {addStyle, View} from "@khanacademy/wonder-blocks-core";
+import {rtlText} from "../components/text-for-testing";
+
+const StyledDiv = addStyle("div");
+
+const generateRows = (rtl: boolean = false) => [
+    {
+        name: "Default",
+        props: {
+            tabs: [
+                {
+                    label: rtl ? rtlText : "Tab 1",
+                    id: "tab-1",
+                    panel: rtl ? rtlText : "Tab 1 Contents",
+                },
+                {
+                    label: rtl ? rtlText : "Tab 2",
+                    id: "tab-2",
+                    panel: rtl ? rtlText : "Tab 2 Contents",
+                },
+                {
+                    label: rtl ? rtlText : "Tab 3",
+                    id: "tab-3",
+                    panel: rtl ? rtlText : "Tab 3 Contents",
+                },
+            ],
+            selectedTabId: "tab-1",
+        },
+    },
+    {
+        name: "With Icons",
+        props: {
+            tabs: [
+                {
+                    label: (
+                        <StyledDiv
+                            style={{
+                                display: "flex",
+                                alignItems: "center",
+                                // TODO: Update to use spacing tokens
+                                gap: "4px",
+                            }}
+                        >
+                            <PhosphorIcon icon={IconMappings.cookie} />
+                            {rtl ? rtlText : "Tab 1"}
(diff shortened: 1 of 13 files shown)


packages/wonder-blocks-tabs/src/components/tabs.tsx
```diff
@@ -174,10 +212,34 @@ export const Tabs = React.forwardRef(function Tabs(
...
+    const tablistRef = React.useRef<HTMLDivElement>(null);
+
+    /**
+     * Determine if the tab is active if it has aria-selected="true".
+     */
+    const isTabActive = React.useCallback((tabElement: Element): boolean => {
+        return tabElement.ariaSelected === "true";
+    }, []);
+
+    const {indicatorProps, updateUnderlineStyle} = useTabIndicator({
```
beaesguerra, line 227: This uses the same logic as NavigationTabs for the selected tab indicator animation. 

I created https://khanacademy.atlassian.net/browse/WB-1924 for adding animations to the tab panel. This is something that won't be needed for NavigationTabs since that relies on link navigation.

__docs__/wonder-blocks-tabs/tabs-variants.stories.tsx
```diff
@@ -0,0 +1,219 @@
...
+                    panel: rtl ? rtlText : "Tab 2 Contents",
+                },
+                {
+                    label: (
+                        <StyledDiv
+                            style={{
+                                display: "flex",
+                                alignItems: "center",
+                                // TODO: Update to use spacing tokens
+                                gap: "4px",
```
jandrade, line 84: **suggestion:** I guess this is possible now that Marcy has landed the new spacing tokens :).

__docs__/wonder-blocks-tabs/tabs-variants.stories.tsx
```diff
@@ -0,0 +1,219 @@
...
+ * component. This is only used for visual testing in Chromatic.
+ */
+const meta = {
+    title: "Packages / Tabs / Tabs / Tabs - All Variants",
+    component: Tabs,
+    render: (args) => (
+        <>
+            <AllVariants rows={rows} columns={columns}>
+                {(props) => (
+                    <View>
```
jandrade, line 154: **suggestion:** This `View` could be removed if it is not necessary.

__docs__/wonder-blocks-tabs/tabs.stories.tsx
```diff
@@ -10,16 +11,78 @@ import Link from "@khanacademy/wonder-blocks-link";
...
+            selectedTabId={selectedTabId}
+            onTabSelected={setSelectedTabId}
+            tabs={props.tabs}
+        />
+    );
+}
+
 export default {
-    title: "Packages / Tabs / Tabs / Tabs",
+    title: "Packages / Tabs / Tabs",
```
jandrade, line 83: yay 🎉

__docs__/wonder-blocks-tabs/tabs.stories.tsx
```diff
@@ -10,16 +11,78 @@ import Link from "@khanacademy/wonder-blocks-link";
...
     component: Tabs,
+    subcomponents: {Tab},
```
jandrade, line 85: **praise:** Ohh this makes total sense! nice add.

__docs__/wonder-blocks-tabs/tabs.stories.tsx
```diff
@@ -276,3 +396,178 @@ export const PanelCaching: StoryComponentType = {
...
+        return (
+            <ScenariosLayout scenarios={scenarios}>
+                {(props) => <ControlledTabs {...args} {...props} />}
+            </ScenariosLayout>
+        );
+    },
+    args: {
+        animated: true,
+    },
+};
```
jandrade, line 573: **thought:** (no change required).... this is really useful for testing/internal purposes. How do you feel about creating a separate stories file to cover this? I'm not sure if we should include this as part of the Docs page, or if it is still useful for devs to understand/test. wdyt?

packages/wonder-blocks-tabs/src/components/navigation-tabs.tsx
```diff
@@ -107,81 +107,22 @@ export const NavigationTabs = React.forwardRef(function NavigationTabs(
...
-                    indicatorIsReady.current = true;
-                }
-            }
-        });
-
-        observer.observe(listRef.current);
+    const isTabActive = React.useCallback(
+        (navTabItemElement: Element): boolean => {
+            // Check the link inside the NavigationTabItem to see if it has aria-current="page"
+            return navTabItemElement.children[0]?.ariaCurrent === "page";
```
jandrade, line 113: **question:** Unrelated to this PR..... Would there be a case where the first element is NOT a Link? I don't think so, but want to confirm if this could be an edge case.

packages/wonder-blocks-tabs/src/components/tab.tsx
```diff
@@ -65,8 +76,60 @@ export const Tab = React.forwardRef(function Tab(
...
+            right: 0,
+            bottom: `-${bottomSpacing}`,
+        },
+        // Only apply hover styles to tabs that are not selected
+        [":hover:not([aria-selected='true'])" as any]: {
+            color: semanticColor.action.secondary.progressive.hover.foreground,
+            [":after" as any]: {
+                height: border.width.hairline,
+                backgroundColor:
+                    semanticColor.action.secondary.progressive.hover.foreground,
```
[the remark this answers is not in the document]
jandrade: This could be helpful as a summarized comment in the code.

FWIW I think it would be fine using `box-shadow` directly as we use a combination of `box-shadow` + `outine` for the global focus online. So using this approach would be only overriding the inner `white` ring, but the outer `blue` would remain there.

Do you think it would make sense to keep it this way?

jandrade (approved):
This looks great! I left some comments for you to consider, but nothing blocking that I could think of. Great stuff 👏 💯

__docs__/wonder-blocks-tabs/tabs.stories.tsx
beaesguerra: I was able to remove one level because I created a "Subcomponents" level for the `Tab` component instead!

__docs__/wonder-blocks-tabs/tabs.stories.tsx
beaesguerra: I will update these stories in a follow up to use the new `Testing/Snapshot` and `StateSheet` structure 😄

packages/wonder-blocks-tabs/src/components/tab.tsx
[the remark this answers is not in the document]
beaesguerra: I think it makes sense to keep the underline styling separate from the focus outline since the inner white ring is useful for dark backgrounds! I left a comment in the code too so we remember why we are using `:after` instead.

PR #2546, opened 2025-04-09
docs(changeset): Update Link styles to use global focus outline

jandrade:
## Summary:
[tb-link] Add TB theming support to Link

Issue: WB-XXXX

## Test plan:


jandrade:
`focusStyles.focus` breaks on links with multiple lines.

PR #2547, opened 2025-04-10
Update base in tokens docs

marcysutton:
## Summary:
The base for calculating pixel equivalents was still listed as `8`, so the token docs were outdated. This PR updates that hard-coded value to `10`! 

Issue: WB-1914

## Test plan:
1. Visit the token docs page /?path=/docs/packages-tokens-sizing--docs
2. Look at the values in the "Pixels (equivalent)" column
3. The values should be base 10, such as sizing.size_180 / 1.8x / 1.8rem / 18


beaesguerra (approved):
Thanks for updating this!

PR #2552, opened 2025-04-11
IconButton: Fix primary border on hover+focus and press+focus

jandrade:
## Summary:

Fixes the `IconButton.primary` border/outline to always use the global focus
outline color (blue).

This issue was found while working on #2550.

Also simplified the css selectors to make it easier to override (if ever needed).

Issue: "none"

## Test plan:

Verify that the `IconButton` snapshot tests reflect this change.

`/?path=/story/packages-iconbutton-testing-iconbutton-snapshots--state-sheet-story`


packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
```diff
@@ -266,15 +268,18 @@ const _generateStyles = (
...
         },
         disabled: {
             cursor: "not-allowed",
             ...disabledStatesStyles,
             // NOTE: Even that browsers recommend to specify pseudo-classes in
             // this order: link, visited, focus, hover, active, we need to
             // specify focus after hover to override hover styles. By doing this
-            // we are able to remove the hover border when the button is
-            // disabled.
+            // we are able to restet the border/outline styles to the default
```
beaesguerra, line 281: small typo 😄 

```suggestion
            // we are able to reset the border/outline styles to the default
```

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
```diff
@@ -225,9 +226,12 @@ const _generateStyles = (
             /**
              * States
              *
-             * Defined in the following order: hover, focus, active.
+             * Defined in the following order: hover, active, focus.
+             *
+             * This is important as we want to give more priority to the
+             * :focus-visible styles.
              */
-            ":hover:not([aria-disabled=true])": {
```
beaesguerra, line 230: Nice! 🧹 Will we need to update the overrides in webapp that use `:hover:not([aria-disabled=true])`?

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
```diff
@@ -225,9 +226,12 @@ const _generateStyles = (
             /**
              * States
              *
-             * Defined in the following order: hover, focus, active.
+             * Defined in the following order: hover, active, focus.
+             *
+             * This is important as we want to give more priority to the
```
beaesguerra, line 231: How do we feel about increasing css specificity so we don't have to rely on the order of css declarations? Could we have specific styles for `:active:focus-visible` and `:hover:focus-visible`?

beaesguerra (approved):
Looks good! Left some questions in the PR and wanted to get your thoughts on one of the Chromatic snapshots before approving them!

packages/wonder-blocks-icon-button/src/components/icon-button-core.tsx
jandrade: I would prefer to avoid that at the DS level at least, as consumers would need to use the same selector (or a higher one) if for some reason they need to override styles (similar to the `":hover:not([aria-disabled=true])"` case). I think keeping it simple at the WB level would be beneficial for consumers so they can rely on more resilient styles and selectors from the start. This is something that we'll need to revisit for sure once we convert WB to CSS modules.

beaesguerra (approved):
Looks good!

PR #2556, opened 2025-04-15
[WB-1926] Switch: Remove khanmigo theme, style updates

jandrade:
## Summary:

- Removes `khanmigo` theme as it is no longer used.
- Adds `className` prop to `Switch` component to allow custom styling of the
switch in Storybook (Chromatic), and to prepare for a future CSS modules
integration.
- Modifies the focus outline to use the global `focusStyles.focus` style from
`wonder-blocks-styles`.

Other changes:

- Storybook: Fixes an issue with `StateSheet` where the `aria-label` prop was
not being passed down correctly to the `children` of the `StateSheet`.

Issue: WB-1926

## Test plan:

Navigate to the `Switch` component in Storybook and verify that stories
look and work correct.

`/?path=/docs/packages-switch--docs`

Also verify the Chromatic snapshots for the `Switch` component now that we use
StateSheet to display all the possible states of the Switch in a single snapshot.

`/story/packages-switch-testing-switch-snapshots--state-sheet-story`


packages/wonder-blocks-switch/src/components/switch.tsx
```diff
@@ -238,6 +244,10 @@ const _generateStyles = (
                         !disabled && clickable
                             ? theme.color.bg.switch.activeOff
                             : undefined,
+                    ...focusStyles.focus[":focus-visible"],
+                    outline: clickable
+                        ? focusStyles.focus[":focus-visible"].outline
+                        : "none",
```
jandrade, line 250: **note:** I noticed that the `press` state was not including the outline, so I added it here.

packages/wonder-blocks-switch/src/themes/themed-switch.tsx
```diff
@@ -19,7 +18,6 @@ export type SwitchThemeContract = typeof defaultTheme;
  */
 const themes: Themes<SwitchThemeContract> = {
     default: defaultTheme,
-    khanmigo: khanmigoTheme,
```
jandrade, line 22: **note:** I'm just removing the `km` theme here for now to reduce impact on the PR, but the goal later is to move this logic to `wb-tokens` + css variables.

.changeset/flat-bottles-flash.md
```diff
@@ -0,0 +1,7 @@
+---
+"@khanacademy/wonder-blocks-switch": minor
```
beaesguerra, line 2: Would removing support for the `khanmigo` theme be a breaking change? If so, should this be a `major` version?

packages/wonder-blocks-switch/src/components/switch.tsx
```diff
@@ -142,22 +149,16 @@ const themedSharedStyles: ThemedStylesFn<SwitchThemeContract> = (theme) => ({
...
         borderRadius: theme.border.radius.small,
         flexShrink: 0,
-        ":hover": {
-            outlineOffset: theme.size.offset.default,
-        },
-        ":focus-within": {
-            outline: `solid ${theme.size.width.small}px ${theme.color.outline.default}`,
-            outlineOffset: theme.size.offset.default,
-        },
+        ":focus-within": focusStyles.focus,
```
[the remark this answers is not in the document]
beaesguerra: When I tab into the switch, I no longer see the focus outline: https://5e1bf4b385e3fb0020b7073c-ubevqgopjf.chromatic.com/?path=/story/packages-switch--default (It looks correct though in the StateSheet snapshot!)

Since `focusStyles.focus` includes the `:focus-visible` selector, should we be using `focusStyles.focus[":focus-visible"` for the `:focus-within` styles? Or is it intended that we use both `:focus-within` and `:focus-visible`?

beaesguerra (commented):
Left some comments, StateSheet snapshot looks good!

.changeset/flat-bottles-flash.md
jandrade: I considered that, but in this case it is not a breaking change per se (yet) as the `wb-theming` package still supports the `khanmigo` theme underneath. This is one of the cases where this implementation doesn't break the component directly.

packages/wonder-blocks-switch/src/components/switch.tsx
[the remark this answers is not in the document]
jandrade: ohhh oops greeeat catch! 😅

beaesguerra (approved):
Looks good! 🚀

PR #2557, opened 2025-04-15
[WB-1644.1] Modify ThemeSwitcher to use a data-wb-theme attribute for theming

jandrade:
## Summary:

This PR builds on top of #2481.

It includes the following changes:

- Theming:
  - Exported `THEME_DATA_ATTRIBUTE` to centralize how the theme is set in
Wonder Blocks.
  - Updated `ThemeSwitcher` to use THEME_DATA_ATTRIBUTE [data-wb-theme]

- Tokens:
  - Refactored build script to export CSS variables for other themes using
  `THEME_DATA_ATTRIBUTE`.

- Internal:
  - Updated root `package.json` to install all the WB packages to avoid having
  to use vite/rollup aliases. This was preventing to import the
  `@ka/wonder-blocks-tokens/styles.css` file correctly in Storybook.
  - Added `Theming > ThemeSwitcher` docs to explain how to use the
  `ThemeSwitcher` component.

Issue: WB-1644

## Test plan:

Navigate to `?path=/docs/packages-theming-themeswitcher--docs`.

Verify that the docs look correct and that the `ThemeSwitcher` component
works correctly:

1. Open devTools and inspect the example.
2. Verify that there's a `<div data-wb-theme="default">` element in the
   DOM.
3. Click on the "Switch theme" button and verify that the `data-wb-theme`
attribute changes to `khanmigo`.

NOTE: The switch theme button won't actually change the styles as this is going
to be added in a future PR (when we introduce the ThunderBlocks theme).

Screenshot 2025-04-15 at 3 34 00 PM


.storybook/preview.tsx
```diff
@@ -12,10 +12,7 @@ import {RenderStateRoot} from "../packages/wonder-blocks-core/src";
 import {Preview} from "@storybook/react";
 
 // Import the Wonder Blocks CSS variables
-// NOTE: External consumers should import the CSS variables from the
-// wonder-blocks-tokens package directly.
-// e.g. import "@khanacademy/wonder-blocks-tokens/styles.css";
-import "../node_modules/@khanacademy/wonder-blocks-tokens/dist/css/index.css";
+import "@khanacademy/wonder-blocks-tokens/styles.css";
```
jandrade, line 15: **note:** I fixed this by removing `resolve.alias` in vite.config.ts below.

__docs__/wonder-blocks-theming/theme-switcher.mdx
jandrade, line 1: **note:** I noticed that I never added docs to this new component, so I'm adding some WIP docs for now. These docs will make more sense as soon as I integrate the TB theme (added a TODO ticket to specify that).

package.json
jandrade, line 1: **note:** Added all the WB deps here to avoid having to use `resolve.alias` with Vite. This allow us to resolve CSS files properly!

packages/wonder-blocks-theming/src/utils/constants.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * The data attribute that scopes the theme to a specific element.
+ * This is used to apply the theme to a specific element(s) in the DOM.
+ *
+ * NOTE: The -wb- prefix is used to avoid conflicts with other libraries, and
+ * scope the theme to Wonder Blocks.
+ */
+export const THEME_DATA_ATTRIBUTE = "data-wb-theme";
```
jandrade, line 8: **note:** Centralizing this here, so we can ensure that we use the right selector in `wb-tokens` and `wb-theming`.

__docs__/wonder-blocks-theming/theme-switcher.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+<Meta of={ThemeSwitcherStories} />
+
+# `ThemeSwitcher`
+
+`ThemeSwitcher` is a component that wraps a children with a reference to the
+selected theme. This uses a combination of custom data attributes
+(`data-wb-theme`) and CSS variables to apply the theme to the children. The
+default CSS variables are defined in the `:root` selector, and the
+theme-specific CSS variables are defined in the `data-wb-theme` selector. For
```
marcysutton, line 12: Maybe you're already planning for this, but it would be helpful to see what the `data-wb-theme` selector looks like for the non-default theme. Is it a value assigned to `data-wb-theme`, like `[data-wb-theme="classroom"]{}`?

For a part 2, is there a quick and easy way for readers to look at `@khanacademy/wonder-blocks-tokens/styles.css`? Do you have to install the package, import the file and then open it, or can we link to it somewhere in the docs?

jandrade: > Maybe you're already planning for this, but it would be helpful to see what the data-wb-theme selector looks like for the non-default theme. Is it a value assigned to data-wb-theme, like [data-wb-theme="classroom"]{}?

That's correct! I'm going to be adding that in https://khanacademy.atlassian.net/browse/WB-1896

> For a part 2, is there a quick and easy way for readers to look at @khanacademy/wonder-blocks-tokens/styles.css? Do you have to install the package, import the file and then open it, or can we link to it somewhere in the docs?

The css file should come for free when this wb-tokens version is installed in consumers. One thing that I could do is to add the css var equivalent in all the tokens pages so folks can see how to use these tokens in a `*.css` context. For now this appears in the `Using color` page as the reference to the CSS var: `/?path=/docs/foundations-using-color--docs`

Screenshot 2025-04-15 at 5 01 58 PM

I'll try to modify that so it includes both the CSS variable and the JS token.

Open to any other ideas around documenting this approach :)

jandrade:
> This looks good to me! It will be nice to lean on regular CSS instead of React Context for this.
> 
> My one remaining question is related to the Vite change: will it still pick up changes to local packages with the removal of `resolve.alias`?

@marcysutton That's an excellent point. I rerun things locally and it's not picking up local changes. I've messed up my local env. at this point for all the different things I'm building/changing right now 😅. So I'll have to somewhat revert this `resolve.alias` change to make it work with css files as well.

__docs__/wonder-blocks-theming/theme-switcher.mdx
jandrade: Not all of them, as it would be good to discourage use of some of these tokens (e.g. `spacing` and `color`).

jandrade: To expand, these are the ones that I have contemplated to add CSS vars support so far:
- semanticColor 👍
- sizing 👍 
- border.radius 👍
- border.width 👍
- color 👎 (folks should try to use `semanticColor` instead).
- thunderblocks-color (name TBD) 👎 same as `color`.
- font 👍  (this will be done as soon as @marcysutton creates the new typography tokens to support both OG and TB).

Any thoughts on this?

jandrade: This should work now :). I've modified the `TokenTable` SB component to optionally transform the CSS var into a more readable data so we can present the info in the table.

https://github.com/Khan/wonder-blocks/pull/2557/commits/36619ff1b2b78be3a693c5ff74f411cac8c73c7c

Screenshot 2025-04-15 at 6 24 20 PM

PR #2559, opened 2025-04-16
Add toggleMode to accordion and accordion section

maddy531:
## Summary:
This PR updates the Accordion and AccordionSection so that sections can be nested and exposes an internal toggle handler. 

Updates:
- Nesting is now supported.
- The toggle handler is exposed for the consumer.
- They can programmatically control open/close states.

## Screenshots:

https://github.com/user-attachments/assets/4de6c907-8f88-47f7-afb8-3500759175b2

Issue: CLASS-11631

## Test plan:

### DIFF
--- a/__docs__/wonder-blocks-accordion/accordion.stories.tsx
+++ b/__docs__/wonder-blocks-accordion/accordion.stories.tsx
@@ -8,7 +8,6 @@ import {
 } from "@khanacademy/wonder-blocks-accordion";
 import Button from "@khanacademy/wonder-blocks-button";
 import {View} from "@khanacademy/wonder-blocks-core";
-import {Strut} from "@khanacademy/wonder-blocks-layout";
 import * as tokens from "@khanacademy/wonder-blocks-tokens";
 import {LabelLarge} from "@khanacademy/wonder-blocks-typography";
 import {
@@ -21,6 +20,7 @@ import ComponentInfo from "../components/component-info";
 import packageConfig from "../../packages/wonder-blocks-accordion/package.json";
 
 import AccordionArgtypes from "./accordion.argtypes";
+import type {AccordionToggleModeStatus} from "../../packages/wonder-blocks-accordion/src/components/accordion";
 
 /**
  * An accordion displays a vertically stacked list of sections, each of which
@@ -162,7 +162,9 @@ export const CaretPositions: StoryComponentType = {
             <View>
                 {/* Left-to-right */}
                 <View style={styles.sideBySide}>
-                    <View style={styles.fullWidth}>
+                    <View
+                        style={(styles.fullWidth, styles.caretExamplePadding)}
+                    >
                         <LabelLarge>
                             Caret position: end, language direction: left to
                             right
@@ -171,7 +173,6 @@ export const CaretPositions: StoryComponentType = {
                             {exampleSections}
                         </Accordion>
                     </View>
-                    <Strut size={tokens.spacing.xLarge_32} />
                     <View style={styles.fullWidth}>
                         <LabelLarge>
                             Caret position: start, language direction: left to
@@ -185,7 +186,9 @@ export const CaretPositions: StoryComponentType = {
 
                 {/* Right-to-left */}
                 <View style={[styles.sideBySide, styles.rtl]}>
-                    <View style={styles.fullWidth}>
+                    <View
+                        style={(styles.fullWidth, styles.caretExamplePadding)}
+                    >
                         <LabelLarge>
                             Caret position: end, language direction: right to
                             left
@@ -204,7 +207,7 @@ export const CaretPositions: StoryComponentType = {
                             </AccordionSection>
                         </Accordion>
                     </View>
-                    <Strut size={tokens.spacing.xLarge_32} />
+
                     <View style={styles.fullWidth}>
                         <LabelLarge>
                             Caret position: start, language direction: right to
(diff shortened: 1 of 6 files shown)


__docs__/wonder-blocks-accordion/accordion.stories.tsx
```diff
@@ -684,4 +694,177 @@ const styles = StyleSheet.create({
...
+                            caretPosition="start"
+                        >
+                            This is the content of section 3.
+                        </AccordionSection>
+                        <AccordionSection
+                            header={`Nested Section 1`}
+                            key={`nested-section-1`}
+                            caretPosition="start"
+                        >
+                            <AccordionSection
```
nishasy, line 841: Can you add a small margin around the inner nested sections so it's visually more clear that they're nested?

packages/wonder-blocks-accordion/src/components/accordion-section-header.tsx
```diff
@@ -17,7 +17,7 @@ type Props = {
     // Unique ID for this section's button.
     id: string;
     // Header content.
-    header: string | React.ReactElement;
+    header: string | React.ReactElement | TemplateStringsArray;
```
marcysutton, line 20: **question:** I'm curious, are there specific cases where `TemplateStringsArray` becomes necessary?

marcysutton (changes requested):
Thank you _so_ much for contributing this PR! The toggleMode functionality is going to be super useful. I had some feedback primarily on nested accordions and tests!

packages/wonder-blocks-accordion/src/components/accordion.tsx
```diff
@@ -73,6 +83,26 @@ type Props = AriaProps & {
...
+     * Called whenever the accordion’s overall expansion state may have changed,
+     * for example after the user toggles a single section or a “Expand All”
+     * / “Collapse All” action. The callback receives an object with booleans
+     * `anyExpanded` and `anyCollapsed`, indicating whether at least one section
+     * is open or closed, respectively.
+     */
+    onToggleModeComplete?: (status: {
+        anyExpanded: boolean;
+        anyCollapsed: boolean;
+    }) => void;
```
jandrade, line 97: **question:** Is there a specific use case for this callback? By looking at the Story, I don't follow how this could be used in an instance.

packages/wonder-blocks-accordion/src/components/accordion.tsx
```diff
@@ -161,6 +212,19 @@ const Accordion = React.forwardRef(function Accordion(
...
+        const anyClosed = newSectionsOpened.some((sectionOpen) => !sectionOpen);
+
+        if (anyOpen && anyClosed) {
+            onToggleModeComplete?.({anyExpanded: true, anyCollapsed: true});
+        } else {
+            onToggleModeComplete?.({
+                anyExpanded: anyOpen,
+                anyCollapsed: anyClosed,
+            });
+        }
```
jandrade, line 226: **suggestion:** If this callback still needs to be exposed to the consumer, I think it would be best from the API perspective to return the same value of `toggleMode`. That way we could let the consumer know why kind of toggle is currently used.

```suggestion
        // Detect partial vs all open vs all closed sections
        const currentToggleMode = anyOpen && anyClosed ? internalToggleMode : "none";
        onToggleModeComplete?.(currentToggleMode);
```

By using this approach, it is easier for consumers to understand only a single type of input/output. instead of having to understand one input (union type) and a different output (object with booleans).

packages/wonder-blocks-accordion/src/components/accordion.tsx
```diff
@@ -73,6 +83,26 @@ type Props = AriaProps & {
...
+    onToggleModeComplete?: (status: {
+        anyExpanded: boolean;
+        anyCollapsed: boolean;
+    }) => void;
+
+    /**
+     * Uniformly expands or collapses all sections:
+     * `expand-all` opens them all.
+     * `collapse-all` closes them all.
+     * `none` leaves their state user-controlled.
```
jandrade, line 103: **suggestion:** This is great! just an extra `-` would help formatting the docs even more :)

```suggestion
     * - `expand-all` opens them all.
     * - `collapse-all` closes them all.
     * - `none` leaves their state user-controlled.
```

Screenshot 2025-05-01 at 10 25 17 AM

packages/wonder-blocks-accordion/src/components/accordion.tsx
```diff
@@ -13,6 +13,16 @@ export type AccordionCornerKindType =
...
+    anyExpanded: boolean;
+    anyCollapsed: boolean;
+};
+
+export type AccordionToggleMode =
+    | "expand-all"
+    | "collapse-all"
+    | "none"
+    | undefined;
 type Props = AriaProps & {
```
jandrade, line 26: **nit:** it would be nice adding a line break (suuper nit 😄) 

```suggestion
    | undefined;
    
type Props = AriaProps & {
```

packages/wonder-blocks-accordion/src/components/accordion.tsx
```diff
@@ -147,6 +188,16 @@ const Accordion = React.forwardRef(function Accordion(
...
         children.length <= LANDMARK_PROLIFERATION_THRESHOLD;
 
+    // Updates the sectionsOpened state based on the toggleMode
+    // when the toggleMode prop changes to either "expand-all" or "collapse-all"
+    React.useEffect(() => {
+        if (toggleMode === "expand-all") {
+            setSectionsOpened(Array(children.length).fill(true));
+        } else if (toggleMode === "collapse-all") {
+            setSectionsOpened(Array(children.length).fill(false));
+        }
```
jandrade, line 198: **suggestion:** Thanks for adding the comment above! I think this could be simplified to:

```suggestion
        if (toggleMode !== "none") {
            setSectionsOpened(Array(children.length).fill(toggleMode === "expand-all"));
        }
```

PR #2561, opened 2025-04-16
[WB-1644.2] Convert border and sizing tokens to CSS variables

jandrade:
## Summary:

This PR migrates the `border` and `sizing` tokens to internally use CSS
variables. It also updates the documentation to reflect these changes.

Also, refactored the logic for generating CSS variables by moving all those
tokens to a `theme/default.ts` file, then converting all the tokens in there to
CSS variables.

Issue: https://khanacademy.atlassian.net/browse/WB-1644

## Test plan:

Navigate to the `border` and `sizing` docs and verify that the CSS variables are now included:

- `/?path=/docs/packages-tokens-border--docs`
- `/?path=/docs/packages-tokens-sizing--docs`

Also inspect the styles in the browser and check that the CSS variables are
being used in places where `sizing` are `border` tokens are used.

### DIFF
--- a/.changeset/clean-ravens-happen.md
+++ b/.changeset/clean-ravens-happen.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": minor
+---
+
+Export `border` and `sizing` as CSS variables. Map JS tokens to css var counterparts
--- a/__docs__/wonder-blocks-tokens/tokens-border.mdx
+++ b/__docs__/wonder-blocks-tokens/tokens-border.mdx
@@ -4,7 +4,7 @@ import TokenTable from "../components/token-table";
 
 import Banner from "@khanacademy/wonder-blocks-banner";
 import {View} from "@khanacademy/wonder-blocks-core";
-import * as tokens from "@khanacademy/wonder-blocks-tokens";
+import {border, color, sizing} from "@khanacademy/wonder-blocks-tokens";
 
 <Meta title="Packages / Tokens / Border" />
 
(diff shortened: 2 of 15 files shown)


__docs__/wonder-blocks-tabs/navigation-tabs.stories.tsx
```diff
@@ -299,7 +299,7 @@ export const HeaderWithNavigationTabsExample: StoryComponentType = {
             navigationTabsRoot: {
                 // set margin to negative value of header vertical spacing so
                 // that selected indicator lines up with header border
-                margin: `-${headerVerticalSpacing} 0`,
+                margin: `calc(${headerVerticalSpacing}, *-1) 0`,
```
beaesguerra, line 302: Is the comma after the token intended? I think this might be why there's an unexpected Chromatic diff for this story!

beaesguerra (approved):
Noticed a typo in one of the stories that might be causing the Chromatic diff. Other than that, looks good! 

Would it be helpful to document somewhere how to negate WB token values? And how the token values are css variables now instead of specific values? I wonder if a lint rule could be helpful to make sure tokens are used properly as css variables now!

__docs__/wonder-blocks-tabs/navigation-tabs.stories.tsx
jandrade: that's why, thanks for noticing that :)

jandrade:
> Would it be helpful to document somewhere how to negate WB token values? And how the token values are css variables now instead of specific values? I wonder if a lint rule could be helpful to make sure tokens are used properly as css variables now!

@beaesguerra These are great points! I'm going to iterate on these ideas in separate PRs.

https://khanacademy.atlassian.net/browse/WB-1813

PR #2562, opened 2025-04-16
[WB-1896] Tokens: Add `thunderblocks` theme to Wonder Blocks

jandrade:
## Summary:

Adds the new `thunderblocks` theme to support the new classroom experience.

This new theme is based on the existing default theme (OG) but it uses its own
primitive colors.

Also added theming support in Chromatic modes, so that the new theme can be
used for visual regression testing. Note that this was only added to the
IconButton snapshots for now to verify that the new theme works as expected.

More changes will come in future PRs to start using all the correct tokens in
all the required components (including `sizing`, `borders`, `offsets`,
`typography`, etc).

Issue: WB-1896

## Test plan:

Verify that the new theme is available in the Storybook via the "Theme" toolbar.

Verify that the component colors change when the theme is switched to
`thunderblocks`.

`/?path=/docs/foundations-using-color--docs&globals=theme:thunderblocks`

Screenshot 2025-04-16 at 6 01 45 PM

Also verify that the `IconButton` snapshots now include the `thunderblocks` theme.

Screenshot 2025-04-16 at 6 02 29 PM


packages/wonder-blocks-tokens/src/theme/semantic/internal/primitive-color-thunderblocks.ts
```diff
@@ -0,0 +1,127 @@
...
+    white_90: fade(white, 0.9),
+    white_80: fade(white, 0.8),
+    white_70: fade(white, 0.7),
+    white_60: fade(white, 0.6),
+    white_50: fade(white, 0.5),
+    white_40: fade(white, 0.4),
+    white_30: fade(white, 0.3),
+    white_20: fade(white, 0.2),
+    white_10: fade(white, 0.1),
+    white_01: fade(white, 0.01),
```
jandrade, line 126: **note:** This might change based on a convo with Caitlyn.

packages/wonder-blocks-tokens/src/theme/semantic/internal/primitive-color-thunderblocks.ts
```diff
@@ -0,0 +1,109 @@
+import {fade} from "../../../util/utils";
+
+// Base colors
+const black = "#191918";
```
marcysutton, line 4: My one thought on working through some font theming is: should the default theme be put into these theme folders to keep similar objects near each other in the codebase? I think it could help reduce some cognitive complexity for theming.

marcysutton:
@jandrade FYI I think this one might need to be updated with `main` to get the latest changes for `color-utils`!

PR #2564, opened 2025-04-17
[🔥AUDIT🔥] Tokens: Export TS types correctly in package.json so consumers can use them correctly

jandrade:
🖍 _This is an audit!_ 🖍

## Summary:

This change updates the `package.json` file in the `wonder-blocks-tokens`
package to ensure that TypeScript types are exported correctly. This allows
consumers of the package to use the types without any issues.

I noticed this issue when trying to use a snapshot release in Webapp. The cause
is that the `package.json`'s `field` setting was modified to support exporting
CSS files, and the change was not including the type definitions for this
package.

Issue: "none"

## Test plan:

Verify that checks pass.

In the Snapshot release, verify that the types are correctly exported and can be
used in the site properly.

### DIFF
--- a/packages/wonder-blocks-tokens/package.json
+++ b/packages/wonder-blocks-tokens/package.json
@@ -5,7 +5,8 @@
   "exports": {
     ".": {
       "import": "./dist/es/index.js",
-      "require": "./dist/index.js"
+      "require": "./dist/index.js",
+      "types": "./dist/index.d.ts"
     },
     "./styles.css": "./dist/css/index.css"
   },
--- a/.changeset/odd-turkeys-train.md
+++ b/.changeset/odd-turkeys-train.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": patch
+---
+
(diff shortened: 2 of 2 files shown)


jandrade:
Confirmed in consumer app that the types now work:

Screenshot 2025-04-17 at 1 13 26 PM

packages/wonder-blocks-tokens/package.json
```diff
@@ -5,7 +5,8 @@
   "exports": {
     ".": {
       "import": "./dist/es/index.js",
-      "require": "./dist/index.js"
+      "require": "./dist/index.js",
+      "types": "./dist/index.d.ts"
```
somewhatabstract, line 9: **suggestion:** This looks OK. I think long term, the `types` export and the `types` field should probably be deprecated, and instead, the `index.d.ts` file should be copied to be next to both export types so that TypeScripts type discovery just works as intended.

somewhatabstract (commented):
No issues with this change, but I think there's a better, long term approach that we should use. See inline comment.

PR #2565, opened 2025-04-17
Deprecate fixtures framework

somewhatabstract:
## Summary:
This marks the fixtures framework as deprecated. The fixtures framework is not
intended to be used for new code, but that is happening in Webapp.

I am hoping that showing the deprecated strikethrough will discourage folks from more new use so we can move towards removing this and not maintaining it anymore.

I do not believe webapp currently lints against deprecated code use, so this should not create any new errors in webapp when it gets updated.

Issue: FEI-5273

## Test plan:
`pnpm typecheck`
`pnpm lint`

### DIFF
--- a/.changeset/clever-kids-talk.md
+++ b/.changeset/clever-kids-talk.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-testing-core": patch
+---
+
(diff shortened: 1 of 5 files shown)


jeresig (approved):
Ah great, thank you Jeff!

PR #2568, opened 2025-04-22
Use StateSheet for tabs package stories & address snapshot flakiness

beaesguerra:
## Summary:

Improvements to stories:
- Use StateSheet component for states for Tabs and NavigationTabs components
- Update StateSheet so the state labels are with the children component. This is so that the labels are closer to the children (especially in responsive views)
- Move `Scenarios` stories to snapshot testing dir
- Rename `AllVariants` stories to `Testing / <Component> - Snapshots`
- Restructure NavigationTabs stories so that NavigationTabItem is in subcomponents directory and we can reduce nesting

Component updates
- Update NavigationTabs styles so that the current tab styling doesn't change when interacted with (specifically hover + pressed states)
- Use updated border tokens in Tabs
- Move the current tab indicator to use inline styles on a `div` instead of a `View` component. Since the positioning values are dynamic (`translateX` and `width` values), using inline styles will prevent the creation of new classes from aphrodite styles whenever the current tab changes or the tabs are resized. As far as I can tell, this addresses the flakiness in Chromatic snapshots! (After debugging some more, it seems the flakiness would occur whenever we updated the aphrodite styles used for the tab indicator)

Issue: WB-1917

## Test plan:

- Verify NavigationTabs and Tabs stories look as expected in Storybook and Chromatic
- `animated` prop continues to work as expected in Tabs and NavigationTabs


packages/wonder-blocks-tabs/src/hooks/use-tab-indicator.ts
```diff
@@ -112,31 +110,33 @@ export const useTabIndicator = (props: Props) => {
...
         role: "presentation",
     };
 
     return {indicatorProps, updateUnderlineStyle};
 };
 
-const styles = StyleSheet.create({
+// Styles for the tab indicator. We use the styles as inline styles instead of
+// aphrodite styles so that new classes aren't generated each time the tabs
+// resize.
```
beaesguerra, line 136: Switching to inline styles for the indicator which has dynamic styles also made the Chromatic snapshots more consistent for the StateSheet stories!

I think the issue was that the pseudo-states add on would add a class for the state, but it would get overridden when we updated the Aphrodite styles

jandrade: This makes perfect sense, thanks for this perf. improvement!

packages/wonder-blocks-tabs/src/hooks/use-tab-indicator.ts
```diff
@@ -112,31 +110,33 @@ export const useTabIndicator = (props: Props) => {
...
+const styles = {
     currentUnderline: {
-        position: "absolute",
+        position: "absolute" as const,
```
jandrade, line 139: **question:** Is there a specific reason why you had to include the `as const` here?

__docs__/wonder-blocks-tabs/navigation-tab-item-testing-snapshots.stories.tsx
```diff
@@ -143,72 +148,68 @@ type Story = StoryObj<typeof NavigationTabItem>;
...
+                        <View
+                            style={styles.container}
+                            tag="ul"
+                            className={className}
+                        >
                             <NavigationTabItem {...args} {...props} />
                         </View>
                     )}
-                </AllVariants>
+                </StateSheet>
```
jandrade, line 185: **thought:** I mentioned this in Chromatic... a future optimization would be to migrate this to a RTL mode, but that's something that we can totally tackle in the future :).

jandrade (approved):
Looks great overall! thanks for improving `StateSheet`, it's way more readable 👏 🚀

packages/wonder-blocks-tabs/src/hooks/use-tab-indicator.ts
beaesguerra: Yes, good question! Without `as const`, there was a TS error because it was treating `"absolute"` as a string. The error was:

> Type 'string' is not assignable to type 'Position | undefined'.

Adding `as const` makes it so it's treated as the literal `"absolute"` string

PR #2569, opened 2025-04-23
Tabs: Add `mountAllPanels` prop.

beaesguerra:
## Summary:

- Add `mountAllPanels` prop to `Tabs`. It defaults to `false`
- When it is false, panels are only mounted the first time it is visited. It stays mounted so it doesn't need to be remounted when it is visited again.
- When it is true, all panels are mounted when the Tabs first render. This is useful for SEO purposes if we want the unvisited tab panels to be crawlable (context: https://github.com/Khan/wonder-blocks/pull/2542#discussion_r2033952892)

Issue: WB-1923

## Test plan:
- Confirm the PanelCaching story behaves the same way as before (`[?path=/docs/packages-tabs-tabs--docs](http://localhost:6061/?path=/docs/packages-tabs-tabs--docs#panel%20caching)`)
- Confirm the new MountAllPanels story includes all tab panels in the DOM (even if they aren't visited yet) (`http://localhost:6061/?path=/docs/packages-tabs-tabs--docs#mount%20all%20panels`)
  - Confirm that screen readers only read out the panel content for the selected tab

### DIFF
--- a/.changeset/fair-ants-hunt.md
+++ b/.changeset/fair-ants-hunt.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tabs": minor
+---
+
+Tabs: Add `mountAllPanels` prop. Defaults to `false` so tab panels are only rendered if they are visited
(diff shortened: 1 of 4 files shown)


jandrade (approved):
Looks great!!, thanks for the improvement! 🚢

PR #2572, opened 2025-04-25
[WB-1930] Internal: Switch repo to use SWC (instead of babel)

jandrade:
## Summary:

- Updates Storybook to use SWC
- Switch Jest to use SWC
- Switch ESLint to use SWC
- Switch Rollup to use SWC
- Removes `@babel/runtime` and remaining babel instances

Issue: WB-1930

## Test plan:

Verify that all the checks pass.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -400,6 +400,10 @@ importers:
       react:
         specifier: 'catalog:'
         version: 18.2.0
+    devDependencies:
+      '@khanacademy/wb-dev-build-settings':
+        specifier: workspace:*
+        version: link:../../build-settings
 
   packages/wonder-blocks-banner:
     dependencies:


package.json
```diff
@@ -39,40 +39,33 @@
...
     "@khanacademy/eslint-plugin": "^3.1.1",
-    "@khanacademy/wonder-stuff-testing": "^3.0.5",
     "@khanacademy/wonder-blocks-announcer": "workspace:*",
     "@khanacademy/wonder-blocks-core": "workspace:*",
     "@khanacademy/wonder-blocks-link": "workspace:*",
-    "@khanacademy/wonder-blocks-tokens": "workspace:*",
     "@khanacademy/wonder-blocks-theming": "workspace:*",
-    "@rollup/plugin-babel": "^6.0.4",
+    "@khanacademy/wonder-blocks-tokens": "workspace:*",
+    "@khanacademy/wonder-stuff-testing": "^3.0.5",
```
beaesguerra, line 51: Should this also use `workspace:*`? (This was pre-existing!)

.changeset/quick-elephants-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wb-dev-build-settings": major
```
beaesguerra, line 2: Since most of the package's `package.json` dependencies were updated to remove `@babel/runtime`, do we need a changeset for those packages so a new version of it is released? 

I think all the packages will get patch updated anyways since they use `wb-dev-build-settings` and it's a major change, but wanted to check!

package.json
```diff
@@ -119,23 +110,23 @@
...
+    "jscodeshift": "catalog:",
     "npm-package-json-lint": "^8.0.0",
     "playwright": "^1.50.1",
     "prettier": "^3.4.2",
     "react-refresh": "^0.16.0",
     "remark-gfm": "^4.0.0",
     "rollup": "^2.79.2",
     "rollup-plugin-node-externals": "^8.0.0",
-    "storybook": "^8.5.2",
+    "storybook": "^8.6.12",
```
beaesguerra, line 121: Thanks for updating us to the latest storybook! 🫶

beaesguerra (approved):
Looks good to me! Left some non-blocking questions 😄

.changeset/quick-elephants-yawn.md
jandrade: That's how I thought about it :). All packages should get `patch` updates from the `major` change.

jeremywiebe (approved):
Looks right to me.  :)

packages/wonder-blocks-announcer/package.json
```diff
@@ -16,7 +16,6 @@
     "access": "public"
   },
   "dependencies": {
-    "@babel/runtime": "catalog:",
```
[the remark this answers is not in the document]
jandrade: Oh good call! otherwise this package won't be bumped. I'll add it.

jeresig (approved):
This is looking great! So glad you were able to get this all working.

.swcrc
```diff
@@ -0,0 +1,16 @@
...
+        },
+        "transform": {
+            "react": {
+                "runtime": "automatic"
+            }
+        },
+        "externalHelpers": true,
+        "experimental": {}
+    }
+}
```
jeremywiebe, line 16: @jandrade  Circling back to this PR as something came up in Perseus. SWC supports polyfilling for older browsers, but you have to configure it. We're going to be enabling the [browser support](https://swc.rs/docs/configuration/supported-browsers) in Perseus shortly, but you might need to do that here also for WB. We had a regression slip into production where the use of `Array.at` isn't supported in Safari 15.

PR #2575, opened 2025-04-30
[WB-1931] Link: Add `viewTransition` prop to enable the View Transitions API

jandrade:
## Summary:

Adds a `viewTransition` prop to support ReactRouter's v6 `Link.viewTransition`
prop which supports the View Transitions API for transitions between pages.

This includes a similar change in `ClickableBehavior` to support the same prop
for allowing passing the `viewTransition` prop to ReactRouter correctly.

Also patches the `react-router-dom-v5-compat` package to support the
`useViewTransition` hook, which is used in the `Link` stories.

Issue: https://khanacademy.atlassian.net/browse/WB-1931

## Test plan:

Navigate to the `Link` storybook and check the "View Transition" story. Click
the link and check that the view transition occurs correctly between the two
pages.

`/?path=/story/packages-link--view-transition`


static/sb-styles/preview.css
```diff
@@ -125,4 +126,14 @@ html {
...
+}
+
+/**
+ * View Transitions
+ */
+::view-transition-group(*),
+::view-transition-old(*),
+::view-transition-new(*) {
+    animation-timing-function: linear(0, 0.03 1.3%, 0.113 2.7%, 0.84 11.5%, 0.981 14.7%, 1.049 18.2%, 1.061 20%, 1.061 22.2%, 1.011 32.3%, 0.997 39.3%, 1);
+    animation-duration: 1s;
```
beaesguerra, line 138: I'm curious about these styles! Is this something we'll need to set up in `webapp` as well? And for storybook in other repos?

packages/wonder-blocks-link/src/components/link.tsx
```diff
@@ -123,6 +123,16 @@ type CommonProps = AriaProps & {
...
+
+    /**
+     * An optional prop that enables a
+     * [https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API](View
+     * Transition) for this navigation by wrapping the final state update in
+     * `document.startViewTransition()`.
+     *
+     * @see https://reactrouter.com/6.30.0/components/link#viewtransition
+     */
+    viewTransition?: boolean;
```
beaesguerra, line 135: How do we see the `prefers-reduced-motion` app setting fitting in with the `viewTransition`?  Would this prop be set based on that setting (similar to the `animated` prop)?

beaesguerra (approved):
This is so cool! 🤩 I'll play around with it for `NavigationTabs` since it uses `Link` components! 

Left some non-blocking questions!

static/sb-styles/preview.css
jandrade: This is an example for us to showcase how View Transitions works, but it would be up for consumers to define their own view transitions.

packages/wonder-blocks-link/src/components/link.tsx
jandrade: Good question! this should be probably handled at the consumer level (probably using the `prefers-reduced-motion` mq when defining the view transition animations).

It's kind of annoying to do it this way, but it is a limitation of the VT API as they only allow applying styles globally.

static/sb-styles/preview.css
jandrade: Maybe we could explore further how to disable this in WB with the idea that we have been considering around a global context/object to determine when to animate.

packages/wonder-blocks-link/src/components/link.tsx
beaesguerra: Would it be helpful to add documentation around the preferred way of handling it? (We can add it later on too when we learn more from integrating it into places 😄)

marcysutton: That was my question too! We'll want to have a solid pattern for reducing motion.

static/sb-styles/preview.css
marcysutton: Maybe that could be part of `wonder-blocks-styles` to centralize some animations for consistency?

beaesguerra:
@jandrade I played around with Link + NavigationTabs and documented the findings in an experiment PR: https://github.com/Khan/wonder-blocks/pull/2582 , let me know what you think!

PR #2576, opened 2025-04-30
Make sure that the tooltip bubble is using the same ID as the aria-describedby attribute on the anchor

somewhatabstract:
## Summary:
When I migrated from the old ID generation to the new `useId`-based generation, I broke the `aria-describedby` attribute on the anchor. This caused screen readers to not read out the tooltip content when the anchor was focused.

This change updates things so that the tooltip bubble ID and the ID used in the `aria-describedby` attribute are the same.

I have also added test cases to ensure that we detect regressions.

Issue: WB-1932

## Test plan:
`pnpm test`
`pnpm typecheck`

### DIFF
--- a/packages/wonder-blocks-tooltip/src/components/__tests__/tooltip.test.tsx
+++ b/packages/wonder-blocks-tooltip/src/components/__tests__/tooltip.test.tsx
@@ -176,7 +176,7 @@ describe("Tooltip", () => {
             const result = await screen.findByRole("tooltip");
 
             // Assert
-            expect(result).toHaveAttribute("id", "tooltip-1");
+            expect(result).toHaveAttribute("id", "tooltip-1-aria-content");
         });
 
         describe("text-only anchor", () => {
@@ -212,8 +212,31 @@ describe("Tooltip", () => {
                 // Assert
                 expect(result).toHaveAttribute(
                     "aria-describedby",
-                    "tooltip-2-anchor-aria-content",
+                    "tooltip-2-aria-content",
                 );
+            });
+
+            test("id provided, aria-describedby id matches bubble id", async () => {
+                // Arrange
+                const ue = userEvent.setup({
+                    advanceTimers: jest.advanceTimersByTimeAsync,
+                });
+                render(
+                    <View>
+                        <Tooltip id="tooltip-2" content="Content">
+                            Anchor
+                        </Tooltip>
+                    </View>,
+                );
+
+                // Act
+                const node = await screen.findByText("Anchor");
+                await ue.hover(node);
+                await act(() => jest.runOnlyPendingTimersAsync());
+                const tooltip = await screen.findByRole("tooltip");
+
+                // Assert
+                expect(node).toHaveAttribute("aria-describedby", tooltip.id);
             });
 
             test("no id provided, attaches aria-describedby", async () => {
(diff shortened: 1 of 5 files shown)


packages/wonder-blocks-tooltip/src/components/tooltip-anchor.tsx
```diff
@@ -53,9 +53,10 @@ type Props = {
      */
     onActiveChanged: (active: boolean) => unknown;
     /**
-     * Optional unique id.
+     * Required aria-describedby id.
+     * This ID will reference the text in the tooltip bubble.
      */
-    id?: string | undefined;
+    "aria-describedby": string;
```
somewhatabstract, line 59: **note:** Made this explicitly being the `aria-describedby` attribute because that's all it was used for anyway.

packages/wonder-blocks-tooltip/src/components/tooltip.tsx
```diff
@@ -217,7 +217,7 @@ export default class Tooltip extends React.Component<Props, State> {
         }
     }
 
-    _renderPopper(bubbleId: string): React.ReactNode {
+    _renderPopper(ariaContentId: string): React.ReactNode {
```
somewhatabstract, line 220: **note:** Some clearer naming.

packages/wonder-blocks-tooltip/src/components/tooltip.tsx
```diff
@@ -269,20 +269,22 @@ export default class Tooltip extends React.Component<Props, State> {
         const shouldBeVisible =
             popperHost && (active || activeBubble) && shouldAnchorExist;
 
+        const ariaContentId = `${uniqueId}-aria-content`;
```
somewhatabstract, line 272: **note:** ID created in one place and sent clearly to the places that use it.

packages/wonder-blocks-tooltip/src/components/__tests__/tooltip.test.tsx
```diff
@@ -212,10 +212,33 @@ describe("Tooltip", () => {
                 // Assert
                 expect(result).toHaveAttribute(
                     "aria-describedby",
-                    "tooltip-2-anchor-aria-content",
+                    "tooltip-2-aria-content",
                 );
             });
 
+            test("id provided, aria-describedby id matches bubble id", async () => {
```
somewhatabstract, line 219: **note:** Added some test cases to verify that the bubble and anchor are using the same ID so we avoid regressions.

beaesguerra (approved):
Looks good to me! Thanks for the fix 😄 

I mentioned in the [Slack thread](https://khanacademy.slack.com/archives/C4PE1QM5Y/p1746052649553119) that I came across this issue while testing tooltips with tabs too! I like your approach of renaming the prop to `aria-describedby` so it is clear (I'll pull in your changes once it's landed!)

One other thing I noticed while I was testing with tabs was there's another a11y warning for when the tooltip is not opened and the anchor references an id that isn't in the DOM yet (here is an [example](https://5e1bf4b385e3fb0020b7073c-ckserrhzoh.chromatic.com/?path=/story/packages-tooltip-tooltip--tooltip-on-buttons)). I fixed this in https://github.com/Khan/wonder-blocks/pull/2574#discussion_r2068973708 too in case you had any thoughts on that!

somewhatabstract:
> One other thing I noticed while I was testing with tabs was there's another a11y warning for when the tooltip is not opened and the anchor references an id that isn't in the DOM yet (here is an [example](https://5e1bf4b385e3fb0020b7073c-ckserrhzoh.chromatic.com/?path=/story/packages-tooltip-tooltip--tooltip-on-buttons)). I fixed this in https://github.com/Khan/wonder-blocks/pull/2574#discussion_r2068973708 too in case you had any thoughts on that!

I am in two minds about this. On the one hand, it's annoying to see a warning. On the other, it's annoying to modify this just to avoid the warning when there's no real functional impact. As an aside to give personal context; I'm all for fixing things when it makes functional sense, but I'm not so keen on doing things just to satisfy automated testing if manual testing works - the tools should help us do what is necessary, not have us doing non-productive busy work just to keep them quiet.

I don't feel strongly either way though so I think getting rid of the warning makes sense if it doesn't introduce any new issue (which it shouldn't).

PR #2581, opened 2025-05-02
[WB-1899.3] Simplify Button theme and styles

jandrade:
## Summary:

Updates Button default theme to simplify its use and have more consistency with
the `IconButton` styles. Also makes some changes to get this button closer to
the design specs.

Figma: https://www.figma.com/design/VbVu3h2BpBhH80niq101MHHE/%F0%9F%92%A0-Main-Components?node-id=19724-1290&t=f0f3JDOmhwq6FUr8-4

### Implementation plan:

1. #2579
2. #2580
3. **Simplify theme and styles to use `semanticColor` tokens directly. (CURRENT)**
4. Add `actionType=neutral`.
4. Refactor `ButtonCore` to make it an unstyled component.

Issue: WB-1899

## Test plan:

Navigate to the `Button` component in Storybook and verify that all the buttons
are working as expected. Use the Figma specs as reference.


packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -251,30 +249,30 @@ const themedSharedStyles: ThemedStylesFn<ButtonThemeContract> = (theme) => ({
...
+        lineHeight: theme.root.font.lineHeight.small,
     },
     largeText: {
-        fontSize: theme.font.size.large,
-        lineHeight: theme.font.lineHeight.large,
+        fontSize: theme.root.font.size.large,
+        lineHeight: theme.root.font.lineHeight.large,
     },
     textWithFocus: {
         position: "relative", // allows the tertiary button border to use the label width
```
jandrade, line 278: **note:** All changes in this section are related to creating the `root` section in the `theme` object.

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -331,170 +329,147 @@ export const _generateStyles = (
     }
 
     const padding =
-        size === "large" ? theme.padding.xLarge : theme.padding.large;
+        size === "large" ? theme.root.padding.large : theme.root.padding.medium;
```
jandrade, line 332: **note:** I cleaned up the `padding` tokens a bit, now there are less sizes for the `root` object.

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -331,170 +329,147 @@ export const _generateStyles = (
...
 
-    const disabledState = theme.color[kind].disabled;
+    const borderWidthKind = theme.root.border.width[kind];
```
jandrade, line 334: **note:** The following lines are mostly copied from `icon-button-core.tsx` with some minor changes (I'll add more notes with that).

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -331,170 +329,147 @@ export const _generateStyles = (
...
+
+                // tertiary-specific styles
+                ...(kind === "tertiary"
+                    ? {
+                          textDecoration: "underline",
+                          textUnderlineOffset: theme.root.font.offset.default,
+                          textDecorationThickness:
+                              theme.root.sizing.underline.hover,
+                      }
+                    : undefined),
```
jandrade, line 390: **note:** This section is different from `IconButton` as the `Button.tertiary` kind uses underline (and IconButton doesn't).

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -331,170 +329,147 @@ export const _generateStyles = (
...
+
+                // tertiary-specific styles
+                ...(kind === "tertiary"
+                    ? {
+                          textDecoration: "underline",
+                          textUnderlineOffset: theme.root.font.offset.default,
+                          textDecorationThickness:
+                              theme.root.sizing.underline.press,
+                      }
+                    : undefined),
```
jandrade, line 427: **note:** Similar change here to support `textDecoration` in regular buttons.

packages/wonder-blocks-button/src/themes/default.ts
jandrade, line 1: **note:** The main changes in this file are:

- `root`: Contains the tokens applied to the `button` element.
- `icon`: icon-specific tokens (`startIcon` and `endIcon`).

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -331,170 +329,148 @@ export const _generateStyles = (
...
+            borderStyle: "solid",
+            borderWidth: borderWidthKind.default,
+            borderColor: themeVariant.default.border,
+            background: themeVariant.default.background,
+            color: themeVariant.default.foreground,
+
+            /**
+             * States
+             *
+             * Defined in the following order: hover, active, focus.
```
beaesguerra, line 360: Should the comment say: `focus, hover, active`? I see the focusStyles are applied first!

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -331,170 +329,148 @@ export const _generateStyles = (
...
-                    textDecorationThickness: theme.size.underline.hover,
-                },
-                ":focus-visible": focusStyling,
-                ":active": activePressedStyling,
+        },
+        disabled: {
+            cursor: "not-allowed",
+            ...disabledStatesStyles,
+            // NOTE: Even that browsers recommend to specify pseudo-classes in
+            // this order: link, visited, focus, hover, active, we need to
```
beaesguerra, line 438: I think there seems to be a mismatch between the comment and the order of styles (comment says: `focus, hover, active`, and the order is `hover, active, focus`. Let me know if I am misunderstanding something though!

beaesguerra (commented):
Thank you for simplifying the theme and styles! I left some questions in Chromatic and in the PR!

packages/wonder-blocks-button/src/components/button-core.tsx
jandrade: I've reverted back the order to have focus defined last.

packages/wonder-blocks-button/src/components/button-core.tsx
jandrade: Good point, changing it!

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -331,170 +329,163 @@ export const _generateStyles = (
...
+                          boxShadow: `inset 0 0 0 ${borderWidthKind.press} ${themeVariant.press.border}, ${focusStyles.focus[":focus-visible"].boxShadow}`,
+                      },
+                  }
+                : {}),
+        },
+        disabled: {
+            cursor: "not-allowed",
+            ...disabledStatesStyles,
+            // NOTE: Even that browsers recommend to specify pseudo-classes in
+            // this order: link, visited, hover, focus, active, we need to
```
beaesguerra, line 467: Should the comment be in this order so it matches the implementation? 
```suggestion
            // this order: link, visited, hover, active, focus, we need to
```

beaesguerra (commented):
Thanks for fixing the disabled + pressed/hover styles! I noticed one more difference in Chromatic that I wanted to get your thoughts on!

packages/wonder-blocks-button/src/components/button-core.tsx
jandrade: That comment is intended to present how browsers order these pseudo-classes. According to the article in the comments, that's how it would work (but I'm not 100% sure if that's true).

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -145,7 +145,7 @@ export const semanticColor = {
             },
 
             disabled: {
-                border: border.primary,
+                border: color.fadedOffBlack32,
```
beaesguerra, line 148: Do we need a changeset file for the tokens update?

beaesguerra (approved):
Looks good!

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
jandrade: Added! as a `patch` change as the public API won't change (just a fix).

PR #2582, opened 2025-05-02
Experimenting with view transition api and Tabs/NavigationTabs

beaesguerra:
## Summary:

Experimenting with the view transition api for the panel/page animation when a tab in Tabs or NavigationTabs is selected. This won't be merged in! (Note: The PR includes the `link-view-transition` and `feature/tabs` changes. The tabs feature branch is shown in the diff. See the latest commits prefixed with `view-transition-tabs` for the new changes)

Let me know if anyone has any feedback or suggestions around these findings!

Issue: WB-1924

#### NavigationTabs and Links

`?path=/story/packages-tabs-navigationtabs--with-link-view-transition`

- Initially tried it using the new `viewTransition` prop in WB `Link`. However, I was only able to get it working where the transitions would occur in one direction
- We can use [view transition types](https://developer.chrome.com/docs/web-platform/view-transitions/same-document#view-transition-types) to be able to control which animation to use
- React Router doesn't support view transition types, though there's an open PR to add it in `https://github.com/remix-run/react-router/pull/12991`
- As a workaround, I was able to get it sliding in/out both directions by using the `document.startViewTransition` directly with the `types` option. I hooked it into the `Link`'s `beforeNav` functionality

#### Tabs

`?path=/story/packages-tabs-tabs--animated`
- We can use `startViewTransition` when we update the selected tab id. We can use it with the `types` option so we can decide on the forwards/backwards animation
- We have 2 options:
  - Including the view transition in the component ([diff](https://github.com/Khan/wonder-blocks/pull/2582/files/f0ce4e73fac2888c9091ac57efa7dcade3a4e1a5..641665f520c79a8430c64a36027a59f87d38acb8))
    - Pros: consistent transitions by default
    - Cons: less flexibility (though we can let consumers override/disable the behaviour)
  - Let the consumers add the view transition for the tab panel ([diff](https://github.com/Khan/wonder-blocks/pull/2582/files/f0ce4e73fac2888c9091ac57efa7dcade3a4e1a5..fb78d3694302b033499b72c565ed3097d35c2a3c#diff-c66992f223cbc77b8c6c5da6b3e7f8f075fd884fc630699df7265f1f7621e757))
    - Pros: more flexibility for the consumers
    - Cons: more things consumers would have to handle (ie. rtl animation), can be inconsistent across usage 

#### General considerations / Things to look into more
- Browser support
  - no support for firefox, safari < 18 (caniuse: [View Transitions](https://caniuse.com/view-transitions), [:active-view-transition-type](https://caniuse.com/mdn-css_selectors_active-view-transition-type))
  - mitigation: we can implement a fallback where the view transition isn't used if it is not supported in the browser. Should behave the same except there is no transition
- Scoping
  - Currently, I've got these working in WB storybook. The css is defined in the Storybook `preview.css` file. Is there a way to encapsulate the styles for the component? Or should we leave it up to the consumers to set up the panel/page animations?
  - How can we support multiple instances of Tabs/NavigationTabs on the same page? 
- Reduced motion 
  - We currently use an `animated` prop in `Tabs` to enable/disable the tab indicator animation.
    - Should this also control the animation for the tab panel transition? Or would we want separate props to control them differently? 
- RTL
  - Tabs: Need to update keyboard navigation implementation to account for rtl (ex: Pressing the right arrow doesn't mean go to the next tab in rtl) (WB-1933)
  - Confirm animation direction in rtl vs ltr


packages/wonder-blocks-tabs/src/components/tabs.tsx
```diff
@@ -254,12 +255,24 @@ export const Tabs = React.forwardRef(function Tabs(
...
+                newTabIndex > currentTabIndex ? "forwards" : "backwards";
+            startViewTransition({
+                update: () => {
+                    if (tabId !== selectedTabId) {
+                        // Select the tab only if it's not already selected
+                        onTabSelected(tabId);
+                    }
+                },
+                types: [direction],
+            });
```
jandrade: **suggestion:** I don't think we should use the View Transitions API with the `Tabs` component. For `NavigationTabs` it makes total sense (as the page is changing to a different route), but for `Tabs`, this should be handled by the consumer (and this would probably happen with `motion.dev`).

beaesguerra: Gotcha! So we only want to use the the View Transitions API with route changes?

For `Tabs`, do we want to include the `motion.dev` transition styles in the WB component, or let the `Tabs` consumer configure it? I can experiment with `motion.dev` next!

jandrade: I think it would be best (at least for now) to let consumers define that. We can revisit later what would be the best strategy from the WB perspective.

beaesguerra:
See https://github.com/Khan/wonder-blocks/pull/2591 for exploration using CSS animations for Tabs

PR #2584, opened 2025-05-06
[🔥AUDIT🔥] [WB-1936] Disable Storybook tests check

jandrade:
🖍 _This is an audit!_ 🖍

## Summary:

The Storybook test addon started failing in our Chromatic/SB GitHub workflow.
After chatting with the Storybook team, we haven’t figured out the cause of this
problem, so we have to temporarily disable these tests until we get a fix later.

Issue: https://khanacademy.atlassian.net/browse/WB-1936

## Test plan:

Verify that the Chromatic/SB GitHub workflow passes and that the Storybook
tests are skipped.


.github/workflows/chromatic-build.yml
somewhatabstract, line 1: **note:** Disappointing but at least it got things unstuck.

somewhatabstract (commented):
Thanks for raising a ticket to track getting these additional tests activated again. I hope Chromatic works out what the issue was.

PR #2586, opened 2025-05-07
[WB-1895.2] Tokens: Use transparent background for secondary.disabled state

jandrade:
## Summary:

Fix `semanticColor.action.secondary.disabled.background` to match the Design
specs (transparent instead of offWhite).

For reference, take a look at the `Button` specs in the OG library.

Issue: https://khanacademy.atlassian.net/browse/WB-1895

## Test plan:

Verify that the snapshots now use the `transparent` color for the
`secondary.disabled.background` variants (e.g. `Button`, `IconButton`,
`actionStyles`).


packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -129,7 +129,7 @@ export const semanticColor = {
 
             disabled: {
                 border: border.primary,
-                background: color.offWhite,
+                background: "transparent",
```
beaesguerra, line 132: Would it be helpful to have a color token for `transparent`?

beaesguerra (approved):
Looks good! 🚀

PR #2588, opened 2025-05-07
[WB-1937] IconButton not opening with the `Space` key (fix)

jandrade:
## Summary:
- Fixes an issue with the IconButton component not triggering `onClick` when the
 `Space` key is pressed.
- Adds a story to demonstrate the issue and the fix.

Issue: https://khanacademy.atlassian.net/browse/WB-1937

## Test plan:

Navigate to `?path=/story/packages-dropdown-actionmenu--opening-modal&globals=viewport:desktop`

- Open the ActionMenu using the `Space` key.
- Verify that the ActionMenu opens as expected.
- Navigate to the "Open modal" action item by using the arrow keys.
- Press the `Space` key to open the modal.
- Verify that the modal opens as expected.
- Tab through the Modal dialog and verify that the focus is trapped within the modal.
- Press the `Escape` key to close the modal.

### DIFF
--- a/.changeset/sweet-pianos-end.md
+++ b/.changeset/sweet-pianos-end.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/wonder-blocks-icon-button": patch
+"@khanacademy/wonder-blocks-dropdown": patch
+---
+
+Fixes an issue with the IconButton component not triggering `onClick` when the `Space` key is pressed.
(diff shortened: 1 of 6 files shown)


beaesguerra (approved):
Looks good to me, thanks for the fix!

marcysutton:
Thanks for the fix and the additional tests @jandrade !

PR #2589, opened 2025-05-08
[WB-1899.4] Button: Add neutral actionType

jandrade:
## Summary:

Updates `Button` to include a new `actionType` called `neutral`. This is to
support the upcoming ThunderBlocks theme.

Also fixes the `neutral.secondary.press.background` value to be closer to the
progressive and destructive backgrounds.

### Implementation plan:

1. #2579
2. #2580
3. #2581
4. **Add `actionType=neutral`. (CURRENT)**
4. Refactor `ButtonCore` to make it an unstyled component.

Issue: WB-1899

## Test plan:

Navigate to the `Button` component in Storybook and verify that the new `neutral`
action type is available and looks as expected.

Take a look at the `IconButton` stories for reference.

`/?path=/docs/packages-button--docs`

Screenshot 2025-05-08 at 5 22 10 PM

PR #2590, opened 2025-05-09
[WB-1899.5] Refactor `ButtonCore` to make it an unstyled component

jandrade:
## Summary:

Splits `ButtonCore` to keep styles there and creates a new `ButtonUnstyled` to
handle the semantics of the Button HTML elements.

This will allow us to create more "Button" components like the upcoming TB
`ActivityButton` and `DynamicActionButton` that will use the same logic as
`Button` but with different styles.

### Implementation plan:

1. #2579
2. #2580
3. #2581
4. #2589
5. **Refactor `ButtonCore` to make it an unstyled component.(CURRENT)**

Issue: WB-1899

## Test plan:

Verify that the `Button` component still works and looks as expected.


__docs__/wonder-blocks-button/button.stories.tsx
```diff
@@ -29,7 +29,7 @@ import {ThemeSwitcherContext} from "@khanacademy/wonder-blocks-theming";
 import {LabeledField} from "@khanacademy/wonder-blocks-labeled-field";
 
 export default {
-    title: "Packages / Button",
+    title: "Packages / Button / Button",
```
jandrade, line 32: **note:** This is so we can have `Packages / Button / ActivityButton` in the near future.

__docs__/wonder-blocks-button/button.stories.tsx
```diff
@@ -671,36 +671,3 @@ WithRouter.parameters = {
...
-        );
-    },
-    parameters: {
-        chromatic: {
-            // We already have screenshots of other stories that cover more of
-            // the button states
-            disableSnapshot: true,
-        },
-    },
-};
```
jandrade, line 706: **note:** This approach is no longer recommended, so I'm getting rid of the docs right away (this will be replaced by DAB).

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -188,65 +175,31 @@ const ButtonCore: React.ForwardRefExoticComponent<
...
 });
 
 export default ButtonCore;
 
 const themedSharedStyles: ThemedStylesFn<ButtonThemeContract> = (theme) => ({
     shared: {
-        position: "relative",
-        display: "inline-flex",
-        alignItems: "center",
-        justifyContent: "center",
```
jandrade, line 230: **note:** Moved the "reset" styles to `ButtonUnstyled`.

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -274,9 +227,6 @@ const themedSharedStyles: ThemedStylesFn<ButtonThemeContract> = (theme) => ({
         fontSize: theme.root.font.size.large,
         lineHeight: theme.root.font.lineHeight.large,
     },
-    textWithFocus: {
-        position: "relative", // allows the tertiary button border to use the label width
-    },
```
jandrade, line 279: **note:** This is no longer relevant/necessary, so I just got rid of it.

packages/wonder-blocks-button/src/components/button.tsx
```diff
@@ -5,198 +5,19 @@ import type {
...
     ChildrenProps,
 } from "@khanacademy/wonder-blocks-clickable";
-import type {AriaProps, StyleType} from "@khanacademy/wonder-blocks-core";
-import type {PhosphorIconAsset} from "@khanacademy/wonder-blocks-icon";
 import {Link, useInRouterContext} from "react-router-dom-v5-compat";
 import ButtonCore from "./button-core";
 import ThemedButton from "../themes/themed-button";
-
-// Variants
-export type ButtonActionType = "progressive" | "destructive" | "neutral";
```
jandrade, line 15: **note:** Moved all these props to `util/button.types.ts` so these can be reused by the upcoming `*Button` variants.

packages/wonder-blocks-button/src/components/button.tsx
```diff
@@ -240,65 +60,43 @@ const Button: React.ForwardRefExoticComponent<
...
-                onClick={onClick}
-                safeWithNav={safeWithNav}
-                target={target}
-                rel={rel}
-            >
-                {renderProp}
-            </ClickableBehavior>
-        );
-    }
+    const extraClickableProps = beforeNav ? {beforeNav} : {target};
```
jandrade, line 63: **note:** Simplified the if/else on the left with this.

packages/wonder-blocks-button/src/util/button.types.ts
jandrade, line 1: **note:** As mentioned, this was just moved from `button.tsx` (no changes at all).

packages/wonder-blocks-button/src/components/button-unstyled.tsx
```diff
@@ -0,0 +1,105 @@
...
+const StyledLink = addStyle(Link);
+
+type Props = Omit<ButtonProps, "children"> & {
+    /**
+     * The button content.
+     */
+    children: React.ReactNode;
+};
+
+const ButtonUnstyled: React.ForwardRefExoticComponent<
```
beaesguerra, line 20: Do you think it would be helpful to have visual regression tests for ButtonUnstyled on it's own?

packages/wonder-blocks-button/src/components/button.tsx
```diff
@@ -240,65 +60,43 @@ const Button: React.ForwardRefExoticComponent<
...
+
+    return (
+        <ClickableBehavior
+            disabled={spinner || disabled}
+            href={href}
+            role="button"
```
beaesguerra, line 69: Do we want it to have `role="link"` if it is a link and has a `href`? (Previously, ClickableBehavior always had `role="button"` so this was pre-existing!)

Or do we need to set the role? We choose between a `<button>` or `<a>` in the `ButtonUnstyled` component! 

Here is the behaviour where an anchor tag is used with `role=button`:
image

beaesguerra (approved):
Looks good to me! Left some non-blocking questions 😄

packages/wonder-blocks-button/src/components/button-unstyled.tsx
jandrade: I think it should be fine to leave it as it is as the styles are fairly simple.

packages/wonder-blocks-button/src/components/button.tsx
jandrade: That makes total sense, although I'd prefer to hold on this change for a future PR. I'd like to avoid introducing any unintended behaviors with this change 😬. I'll create a ticket to track this.

jandrade: Added as a TODO comment: https://khanacademy.atlassian.net/browse/WB-1940

PR #2591, opened 2025-05-09
[WIP] Tabs animation

beaesguerra:
Progress so far for animating the panels in tabs. 

Remaining tasks:
- Cross browser testing
- Allowing the panel animation to be customized (using style prop?). Also include an example in Storybook on how to do this
- Polish timing
- TS type to support keyframes in aphrodite
- Slide out animation after first render + first time new tab is selected is not working- More testing
- Example for how to add animations for NavigationTabs

Creating a draft PR for reference since we will be de-prioritizing animations for now and will revisit later

Issue: WB-1924

## Test plan:

### DIFF
--- a/__docs__/components/example-table.tsx
+++ b/__docs__/components/example-table.tsx
@@ -62,6 +62,6 @@ const styles = StyleSheet.create({
     },
     headerCell: {
         borderBottom: `${border.width.medium} solid ${semanticColor.border.primary}`,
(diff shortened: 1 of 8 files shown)


beaesguerra:
I created a draft PR for reference for the Tabs animations work so far and documented remaining tasks for when we revisit animations later on!

cc: @jandrade @marcysutton

PR #2592, opened 2025-05-09
Add "active" prop to ActionItem

nedredmond:
Adds "active" prop to ActionItem so that the underlying link or button can get an `aria-current` value

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/action-item.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/action-item.tsx
@@ -96,6 +96,11 @@ type ActionProps = {
      * Optional right accessory to display in the `ActionItem` element.
      */
     rightAccessory?: CompactCellProps["rightAccessory"];
+
+    /**
+     * Optional; applies aria-current to the cell.
+     */
+    active?: CompactCellProps["active"];
 };
 
 type DefaultProps = {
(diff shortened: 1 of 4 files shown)


packages/wonder-blocks-dropdown/src/components/action-item.tsx
```diff
@@ -96,6 +96,11 @@ type ActionProps = {
      * Optional right accessory to display in the `ActionItem` element.
      */
     rightAccessory?: CompactCellProps["rightAccessory"];
+
+    /**
+     * Optional; applies aria-current to the cell.
+     */
+    active?: CompactCellProps["active"];
```
beaesguerra, line 103: **suggestion:** [`action-item.argtypes.tsx`](https://github.com/Khan/wonder-blocks/blob/main/__docs__/wonder-blocks-dropdown/action-item.argtypes.tsx) can be updated so the [prop docs in Storybook](https://5e1bf4b385e3fb0020b7073c-acktdierau.chromatic.com/?path=/docs/packages-dropdown-actionitem--docs) says it's a boolean!

We've found that we have to explicitly define the arg types for inferred types so that they're documented accurately! (It's able to infer the description from the JS docs for the prop so you don't have to duplicate the description in the `argtypes`!)


image

beaesguerra (approved):
Looks good, left a suggestion for documentation! Thanks for making this improvement!

packages/wonder-blocks-dropdown/src/components/action-item.tsx
nedredmond: I wonder why it's not able to infer the type properly? Usually can! I'll add it

PR #2596, opened 2025-05-13
[WB-1941] Experimenting with semanticColor.core tokens (border)

jandrade:
## Summary:

- Creates a new `semanticColor.core` category.
- Moves `border` tokens to the new category.
- Replaces border call sites to support new structure.

Issue: https://khanacademy.atlassian.net/browse/WB-1941

## Test plan:

TBD

### DIFF
--- a/packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
+++ b/packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
@@ -1,13 +1,34 @@
 import {color} from "../../tokens/color";
 
-const border = {
-    primary: color.fadedOffBlack16,
-    subtle: color.fadedOffBlack8,
-    strong: color.fadedOffBlack50,
-    inverse: color.white,
-    progressive: color.blue,
-    destructive: color.red,
+const core = {
+    progressive: {
+        subtle: color.fadedBlue8,
+        default: color.blue,
+        strong: color.activeBlue,
+    },
+    destructive: {
+        subtle: color.fadedRed8,
+        default: color.red,
+        strong: color.activeRed,
+    },
+    neutral: {
+        subtle: color.fadedOffBlack8,
+        default: color.fadedOffBlack16,
+        strong: color.fadedOffBlack50,
+        // what should we do with faddedOffBlack72?
+    },
 };
+
+const border = core;
+
+// const border = {
+//     primary: color.fadedOffBlack16,
+//     subtle: color.fadedOffBlack8,
+//     strong: color.fadedOffBlack50,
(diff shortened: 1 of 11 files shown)


packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -1,12 +1,72 @@
...
+        subtle: "transparent",
+        default: color.fadedOffBlack32,
+        strong: color.fadedOffBlack50,
+    },
+    inverse: {
+        subtle: color.fadedOffBlack8,
+        default: color.offWhite,
+        strong: color.white,
+    },
+    // States
```
beaesguerra, line 31: Would we also have core semantic tokens for `info` and `critical`?

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -1,12 +1,72 @@
...
+        strong: mix(color.offBlack32, color.gold),
+    },
+    success: {
+        subtle: color.fadedGreen8,
+        default: color.green,
+        strong: mix(color.offBlack32, color.green),
+    },
+};
+
+const background = {
```
beaesguerra, line 44: I'm taking a look at the new structure with the context of the Badge component!

If I'm understanding correctly, the Badge should use the semantic color status tokens. And those tokens use the core tokens! Is that right?

Should the status tokens use these background tokens instead? How do we decide when we should use the core tokens directly vs the abstraction?

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -245,23 +305,23 @@ export const semanticColor = {
      */
     status: {
         critical: {
-            background: color.fadedRed8,
-            foreground: color.red,
+            background: core.destructive.subtle,
```
beaesguerra, line 308: Should we have `critical` and `notice` core tokens so that it is separate from `destructive` and `progressive` tokens?

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -1,12 +1,72 @@
...
+};
+
+const background = {
+    neutral: core.neutral,
+    disabled: core.disabled,
+    progressive: core.progressive,
+    destructive: core.destructive,
+};
 
 const border = {
```
beaesguerra, line 51: Would the icon and text tokens also follow the new structure?

PR #2599, opened 2025-05-14
[WB-1942] Adapt CSS build tools to export CSS vars from multiple packages

jandrade:
## Summary:

- `Internal`: Adds Turborepo to Wonder Blocks to initially help with building
CSS variables.

- `Internal`: Add build-css script to merge vars.css files generated by Wonder
Blocks packages.

 - `Tokens`: Exposes `mapValuesToCssVars` so other WB packages can switch to
 theming with CSS vars instead of relying on React Context.

 - `IconButton`: switches IconButton to use CSS variables, instead of relying on
 ReactContext.

### Setup instructions

- Run `pnpm start` as usual to start Storybook.
- In a separate terminal tab, run `pnpm dev` to start Turborepo and watch for CSS vars changes.
- Make a change in `semantic-color.ts` or `wb-icon-button/src/theme/default.ts` and check how the component changes in the browser.

Issue: https://khanacademy.atlassian.net/browse/WB-1928

## Test plan:

Verify that the `IconButton` styles still look as expected.

Also verify that the new build script `pnpm dev` reacts to token changes properly.

https://github.com/user-attachments/assets/5cd48409-6f13-4a66-814e-88dd694eafaa


packages/wonder-blocks-icon-button/src/theme/index.ts
jandrade, line 1: **note:** This is an important piece to get theming working. It reuses `mapValuesToCssVars` to convert the existing JS tokens to references to CSS variables.

packages/wonder-blocks-icon-button/src/theme/index.ts
```diff
@@ -0,0 +1,4 @@
+import {mapValuesToCssVars} from "@khanacademy/wonder-blocks-tokens";
+import themeDefault from "./default";
+
+export default mapValuesToCssVars(themeDefault, "--wb-c-icon-button-");
```
jandrade, line 4: **note:** Setting the prefix is a temporary change. It would be ideal not having to manually configure that on every package (It opens the door to incorrect naming between this and what's bundled in the CSS file). Hoping to find a nice solution that could work with any package. Open to suggestions!

packages/wonder-blocks-tokens/bin/cli.js
jandrade, line 1: **note:** This bin file is created so we can generate the CSS vars from any package. Take a look at this package.json file to see how it works:
https://github.com/Khan/wonder-blocks/pull/2599/files#diff-ff7ca2c6eb92ad0898b315dbc2be00bb991aa88f07a0d4fbfa6de7c131dc7a5fR21

packages/wonder-blocks-tokens/package.json
```diff
@@ -29,5 +29,6 @@
   },
   "devDependencies": {
     "@khanacademy/wb-dev-build-settings": "workspace:*"
-  }
+  },
+  "bin": "./bin/cli.js"
```
jandrade, line 33: **note:** Adding this here so this command can be exposed and consumed by other dependent packages.

packages/wonder-blocks-tokens/src/build/generate-css-variables.ts
jandrade, line 1: **note:** Made some changes to enable generating CSS vars from different packages in parallel (using turborepo).

turbo.json
```diff
@@ -0,0 +1,15 @@
+{
+    "$schema": "https://turborepo.com/schema.json",
+    "tasks": {
+        "dev": {
+            "cache": false
+        },
+        "//#dev:css": {
```
jandrade, line 7: **note:** This is the turbo entrypoint. It is called when running `pnpm dev` (configured in the root `package.json` file).

turbo.json
```diff
@@ -0,0 +1,15 @@
+{
```
jeresig, line 1: I'm finding the structure of the `turbo.json` file (and the commands in the `package.json` files) to be confusing. I think it may be better to think about things as to how they would be structured, ideally:

1. Each package would have a "build" script that generates its own `dist/` folder. Ideally, the build step would be building the full package, not just part of it (the CSS) but I realized that would be a bigger refactor. So for now, I recommend naming it: `build:css` for the packages that have it.
2. The package's inputs would be defined inside the `build:css` task in `turbo.json`. Maybe that's: `"inputs": ["src/theme/**.ts"],`
3. The outputs of this package should also be defined, probably something like: `"outputs": ["dist/css/vars.css"]`
4. The build task should then be cacheable as it'll have clearly defined inputs and outputs.
5. The build task would also have `"dependsOn": ["^build:css"]`
6. I would remove the root `dev:css` task from turbo. You can then just run `turbo watch build:css` and it should re-build only the packages that change, when they change. So you can update `pnpm dev` to run this command for you. If you make this change I think you can also remove all of the dependencies that you had to add to the root `package.json` file, as I think those were only needed due to this root task with its `"dependsOn": ["^dev"]`.

Hope this helps, happy to chat about this more! Caching is a super-critical part of Turborepo, so we really want to enable this, if possible - it's what helps to speed up the incremental builds!

jandrade: @jeresig Thanks again for your feedback and for pairing. I've implemented most of the changes you suggested:

1. Changing from `dev:css` to `build:css` (in root and packages).
2. Modifying the `inputs` field in package targets to enable caching.
3. Modifying the root `build:css` task to make it more performant.

build-settings/build-css.ts
```diff
@@ -0,0 +1,49 @@
...
+
+if (fs.existsSync(indexFile)) {
+    // clear the file if it exists
+    fs.writeFileSync(indexFile, "");
+}
+
+// Generate the output inside dist/css/index.css
+fs.writeFileSync(indexFile, allContent, {
+    flag: "w+",
+});
```
jeresig, line 46: It seems like this would be equivalent to always just overwriting the file with `allContent`?

jeresig (approved):
Changes look good to me - thanks so much for working on this!

PR #2600, opened 2025-05-14
Implement base Badge component

beaesguerra:
## Summary:
- Implement Badge component with base props and props for the label and icon
- The icon can be any ReactElement, so it supports PhosphorIcons or custom icons

Issue: WB-1935

## Test plan:
- Badge works as expected `?path=/docs/packages-badge-badge--docs`
- Docs are reviewed

## Implementation Plan:
- Will implement StatusBadge, GemBadge, StreakBadge, and DueBadge components next. They will use the Badge component and override the default styles


__docs__/components/all-variants.tsx
```diff
@@ -141,7 +141,6 @@ export function AllVariants(props: Props) {
                                         >
                                             {children({
                                                 props: {
-                                                    "aria-label": ariaLabel,
```
beaesguerra, line 144: Removing including `aria-label` in the props since it is not always valid to have an `aria-label` depending on the element. This addresses the a11y warning: `Ensure ARIA attributes are not prohibited for an element's role`.

We can explicitly set the aria-label for components in specific stories using the `name` that is also provided to the `children` render function

packages/wonder-blocks-badge/src/components/badge.tsx
```diff
@@ -1,5 +1,151 @@
...
+                </StyledSpan>
+            )}
+            {label}
+        </StyledDiv>
+    );
+});
+
+export {Badge};
+
+const badgeTokens = {
```
beaesguerra, line 87: Setting up a structure for tokens related to badges. Let me know if anyone has any feedback on this - I was trying to follow the [naming guide](https://khanacademy.atlassian.net/wiki/x/FwB47)!

This doesn't reflect the TB styling yet. We can make updates to the tokens used (or if we need to have component themes) once the TB theme is in place! We may also make more updates to use `semantic core` tokens

packages/wonder-blocks-badge/src/components/badge.tsx
```diff
@@ -1,5 +1,151 @@
...
+    icon: {
+        sizing: {
+            width: sizing.size_160,
+            height: sizing.size_160,
+        },
+        color: {
+            foreground: semanticColor.icon.primary,
+        },
+    },
+    iconOnly: {
```
beaesguerra, line 114: `iconOnly` styles... I'm not sure about where to place the padding token! For context, the badge has smaller padding when it is using only an icon. It's more of a variant of the `root` element, but I'm not sure how to structure it. I'm open to changing this!

packages/wonder-blocks-badge/src/components/badge.tsx
```diff
@@ -1,5 +1,151 @@
...
+    label?: string;
+    /**
+     * The icon to display in the badge. It can be a PhosphorIcon, a custom svg,
+     * or `img` element.
+     * Considerations:
+     * - If the icon conveys meaning, set the alt text on the icon being used
+     * - If the icon is an `img` element, it may need width: 100% and height: 100%
+     *   to render properly in the badge.
+     */
+    icon?: React.ReactElement;
```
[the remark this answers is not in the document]
jandrade: After chatting offline, this API won't change. Do you think it would make sense to add a TODO comment in the prop description to ensure that this is updated once the `Icon` component is available? (no need to add the comment if you think so)

packages/wonder-blocks-badge/src/components/badge.tsx
jandrade: How do you feel about using the `layout` category? (described in the naming guide).

jandrade (commented):
This is looking great! Left some comments to discuss further about the theming approach. Let me know what you think :)

packages/wonder-blocks-badge/src/components/badge.tsx
```diff
@@ -0,0 +1,151 @@
...
+const Badge = React.forwardRef<HTMLDivElement, BadgeProps>(function Badge(
+    props: BadgeProps,
+    ref: React.ForwardedRef<HTMLDivElement>,
+) {
+    const {icon, label, id, testId, styles: stylesProp, ...otherProps} = props;
+    if (!label && !icon) {
+        return <React.Fragment />;
+    }
+    return (
+        <StyledDiv
```
marcysutton, line 56: Looking at the semantics of the static Badge (non-interactive with no Tooltip/etc.), I wonder if there's any other option than a DIV. Maybe a `<strong>` tag, at least?

marcysutton (commented):
Added a couple of comments! This is going to be great!

packages/wonder-blocks-badge/src/components/__tests__/badge.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+    });
+
+    it("should not render anything if there is no label or icon", () => {
+        // Arrange
+        // Act
+        const {container} = render(<Badge />);
+
+        // Assert
+        expect(container).toBeEmptyDOMElement();
+    });
```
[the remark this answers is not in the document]
beaesguerra: Updated the prop types so that one of `label` or `icon` props should be defined! I also explicitly added the prop descriptions to `badge.argtypes.ts`!

packages/wonder-blocks-badge/src/components/badge.tsx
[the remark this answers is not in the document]
beaesguerra: I added a TODO in the prop docs! https://github.com/Khan/wonder-blocks/blob/37568f0c20014958ee84ffd12680666a9900001e/__docs__/wonder-blocks-badge/badge.argtypes.ts#L15

packages/wonder-blocks-badge/src/components/badge.tsx
beaesguerra: Good call! I was mixing up `layout` and `sizing`. I update the grouping for padding/gap properties to `layout`!

packages/wonder-blocks-badge/src/components/badge.tsx
beaesguerra: Thanks for the examples! I will set up the theming structure and token usage in a follow up PR!

__docs__/components/all-variants.tsx
beaesguerra: Good call! I updated the aria-label for OptionItemVariants!

https://github.com/Khan/wonder-blocks/blob/37568f0c20014958ee84ffd12680666a9900001e/__docs__/wonder-blocks-dropdown/option-item-variants.stories.tsx#L55

packages/wonder-blocks-badge/src/components/badge.tsx
beaesguerra: @marcysutton I think the semantics might be something we want to leave up to consumers since it'll depend on usage! 

For example, here's a design with a bunch of badges in a table! If all the badges have `<strong>`, it might lose emphasis/importance since there are many instances of it 😅 
image

To better support it though, we can also add a `tag` prop to the component so consumers can choose the underlying html element! What do you think?

packages/wonder-blocks-badge/src/components/badge.tsx
```diff
@@ -1,5 +1,165 @@
...
+ * a status, label, or count.
+ */
+const Badge = React.forwardRef<HTMLDivElement, BadgeProps>(function Badge(
+    props: BadgeProps,
+    ref: React.ForwardedRef<HTMLDivElement>,
+) {
+    const {icon, label, id, testId, styles: stylesProp, ...otherProps} = props;
+    if (!label && !icon) {
+        return <React.Fragment />;
+    }
```
jandrade, line 68: **thought:** I think this condition is no longer a case? at least not if we enforce the TS types.

packages/wonder-blocks-badge/src/components/badge.tsx
jandrade: Sounds great!

jandrade (approved):
Looks great! thanks for applying the changes 🎉 

Would love to hear from @marcysutton about the `tag` prop, which could be a great addition if it makes sense!

packages/wonder-blocks-badge/src/components/badge.tsx
beaesguerra: Oh I forgot to leave a comment! I left this logic in to account for if `label` is an empty string and icon isn't provided! (I updated the unit tests to cover this case instead of them both being not defined)

packages/wonder-blocks-badge/src/components/badge.tsx
marcysutton: Adding a `tag` prop makes a ton of sense to me!

beaesguerra: Added a `tag` prop and some documentation around how a `strong` tag can be used for emphasis 😄 https://5e1bf4b385e3fb0020b7073c-pggdaoumdz.chromatic.com/?path=/docs/packages-badge-badge--docs#tag

Thanks for the feedback!

beaesguerra:
@jandrade I created https://khanacademy.atlassian.net/browse/WB-1949 for investigating Chromatic vertical alignment issues. Related to the thread in the [Chromatic snapshot](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=6824f9d68ce4611579f88ace&activeElementId=comment-thread-6826315863a46bb3ef1e6306) (I'm no longer able to respond to the thread since there was a newer Chromatic build in this branch!)

PR #2602, opened 2025-05-14
WB-1913: Add Components for Thunderblocks

marcysutton:
## Summary:
Adds new Heading and BodyText components for Thunderblocks, along with Plus Jakarta Sans typeface

Also migrates existing typography tokens to theming

Issue: WB-1913, WB-1912

## Test plan:
1. Inspect Heading and BodyText components to ensure sizes and weights are accounted for
2. Ensure existing components work as-is
3. Check Theme toggle for effects (e.g. Classroom)


packages/wonder-blocks-tokens/src/theme/semantic/internal/primitive-font-thunderblocks.ts
```diff
@@ -0,0 +1,50 @@
...
+        xSmall: sizing.size_160,
+    },
+    weight: {
+        light: fontWeight.light,
+        medium: fontWeight.medium,
+        semi: fontWeight.semi,
+        bold: fontWeight.bold,
+        black: fontWeight.black,
+    },
+};
```
jandrade, line 50: **suggestion:** Oh I forgot to mention one thing that could be helpful around generating the TB theme for `font` tokens. You can use `mergeTheme` here so you only need to change the specific tokens to override + OG is used as a Theme Contract, ensuring that both themes use the same exact token structure.

You can see how that works in `semantic-color-thunderblocks.ts`.

```suggestion
import {mergeTheme} from "@khanacademy/wonder-blocks-theming";
import {font as defaultFont} from "./font";

export const font = mergeTheme(defaultFont, {
    family: {
        sans: fontFamily.sans,
        serif: fontFamily.serif,
        mono: fontFamily.mono,
    },
    size: {
        xxxLarge: sizing.size_360,
        xxLarge: sizing.size_280,
        xLarge: sizing.size_240,
        large: sizing.size_200,
        xMedium: sizing.size_180, // only used in TB
        medium: sizing.size_160,
        small: sizing.size_140,
        xSmall: sizing.size_120,
    },
    lineHeight: {
        xxxLarge: sizing.size_400,
        xxLarge: sizing.size_320,
        xLarge: sizing.size_280,
        large: sizing.size_240,
        xMedium: sizing.size_220,
        medium: sizing.size_200,
        small: sizing.size_180,
        xSmall: sizing.size_160,
    },
});
```

packages/wonder-blocks-tokens/src/util/types.ts
```diff
@@ -1,3 +1,22 @@
...
+
+type OGFontWeights = {
+    regular: number;
+    medium?: never;
+};
+
+type TBFontWeights = {
+    medium: number;
+    regular?: never;
+};
```
jandrade, line 20: **suggestion:** By using the OG font file as the contract, then these styles are no longer needed.

packages/wonder-blocks-typography/src/components/body-text.tsx
```diff
@@ -0,0 +1,42 @@
...
+        ...otherProps
+    }: Props,
+    ref,
+) {
+    // map props to theme defaults
+    const bodyTextStyles = {
+        fontSize: font.size[size],
+        fontWeight: font.weight[weight],
+        lineHeight: font.lineHeight[size],
+    };
```
jandrade, line 28: **praise:** This is neat!

packages/wonder-blocks-typography/src/util/styles.ts
```diff
@@ -27,6 +27,18 @@ const styles: StyleDeclaration = StyleSheet.create({
         fontSize: font.size.large,
         lineHeight: font.lineHeight.large,
     },
+    Heading: {
+        ...common,
+        fontFamily: font.family.sans,
```
jandrade, line 32: **question:** Would this mean that the Jakarta font would be used if the current global theme is `thunderblocks`?

static/jakarta.css
```diff
@@ -0,0 +1,15 @@
...
+    font-variation-settings: "wght" 400;
+    src: url(/static/fonts/jakarta-sans/PlusJakartaSans-VariableFont_wght.woff2) format('woff2');
+}
+
+@font-face {
+    font-family:  "Plus Jakarta Sans", serif;
+    font-style: 'italic';
+    font-weight: 400;
+    src: url(fonts/jakarta-sans/PlusJakartaSans-Italic-VariableFont_wght.woff2) format('woff2');
+}
```
jandrade, line 15: **issue:** It looks like the font is not being loaded by the browser.

Screenshot 2025-05-15 at 12 09 42 PM

jandrade (commented):
Thanks for working on this! I've added a first round of comments, mostly around the following topics:

1. Changing how the TB fonts are defined with `mergeTheme`.
2. Loading the Jakarta font in Storybook.
3. Changing the new stories to support CSF3.

Let me know if you have any questions about any of this!

__docs__/wonder-blocks-typography/body-text.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+When wrapped in \`<ThemeSwitcher theme="classroom">\`, this component will use the Plus Jakarta Sans font family.
+
+The \`size\` prop will select a font size token based on our [REM font sizing scale](/?path=/docs/packages-tokens-typography--docs&globals=theme:thunderblocks#size). A corresponding line-height token will be automatically selected from our [line-height scale](/?path=/docs/packages-tokens-typography--docs&globals=theme:thunderblocks#lineHeight).
+
+The \`weight\` prop will match a font weight token based on the available [font weights](/?path=/docs/packages-tokens-typography--docs&globals=theme:thunderblocks#weight).
+
+Heading text should utilize the \`Heading\` component.`;
+
+export default {
```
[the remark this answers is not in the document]
marcysutton: Good call! I looked around to see how to handle a large text description and only saw something similar to this! I'll update it to use JSDoc. 👍

__docs__/wonder-blocks-typography/body-text.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+    title: "Packages / Typography / BodyText",
```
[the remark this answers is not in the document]
marcysutton: I like that idea! I toyed around with putting "(New)" in the story name, but perhaps grouping them would make more sense. I think they would benefit with some kind of organization.

static/jakarta.css
marcysutton: Weird. They loaded on my machine but clearly didn't in that case! I'll take another look... maybe the font file paths are different on the remote server.

packages/wonder-blocks-typography/src/util/styles.ts
marcysutton: I believe so, yes! That seems to be how it is working now.

static/jakarta.css
marcysutton: Ah, I see it's loading a Local Font for me (thank you Font Book), which obscured the fact that the fonts aren't being loaded. I'll look into it some more!

__docs__/wonder-blocks-typography/body-text.stories.tsx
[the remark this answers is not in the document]
marcysutton: I added "Thunderblocks" to group these components together under `Typography`. Let me know if a different name is preferred! For the stage we're in, it seemed like the most obvious choice but we may want to change it later.

static/jakarta.css
marcysutton: I got the font to load! I'm not 100% sure what fixed it...maybe an incorrectly written `font-style` value. But fiddling around with the `@font-face` definition seems to have worked and it is now loading remotely with my local font removed!

packages/wonder-blocks-tokens/src/theme/primitive/font.ts
```diff
@@ -40,7 +43,9 @@ export const font = {
     },
     weight: {
         light: fontWeight.light,
+        medium: fontWeight.regular,
         regular: fontWeight.regular,
+        semi: fontWeight.regular, // is no semi in Lato, so this is a fallback
         bold: fontWeight.bold,
         black: fontWeight.black,
     },
```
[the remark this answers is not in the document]
marcysutton: Okay, I normalized the fonts in themes to use `font.weight.medium`! I wonder if that would be a breaking change for  `wonder-blocks-tokens` as it removes `font.weight.regular` from the API surface. 🤔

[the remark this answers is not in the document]
jandrade: Good point, agreed, it would be a breaking change in that case. FWIW this token is not used in Perseus. This is a separate convo we should have at some point with the LEMS team as they haven't used the wb-tokens package as we would expect.

__docs__/wonder-blocks-typography/body-text.stories.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+import ComponentInfo from "../components/component-info";
+import TypographyArgTypes from "./typography.argtypes";
+
+import {View} from "@khanacademy/wonder-blocks-core";
+import {BodyText} from "@khanacademy/wonder-blocks-typography";
+import {spacing} from "@khanacademy/wonder-blocks-tokens";
+
+/**
+New component for Thunderblocks body text.
```
beaesguerra, line 14: For my own understanding, should teams use the `BodyText` and `Heading` components in the classic experience? 

Also, how do teams know when they should use the `BodyText` or `Heading` components? Is it something they can inspect in Figma? Maybe we can document something around this so teams know to use these components instead of re-implementing the typography styles!

__docs__/wonder-blocks-typography/heading.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+        size: "large",
+        weight: "bold",
+    },
+};
+
+/**
+ * An example of the `Heading` component's `size` and `weight` prop combinations,
+ * mimicking the ones found in Figma Foundation specs.
+ */
+export const SizesAndWeights = {
```
beaesguerra, line 68: I find this story useful for visual regression tests since it has the different combinations! What do you think about enabling snapshots for this story instead of the [New Typography story](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=682b904ba1d77d7bd773cdd6) which only shows one variant of Heading/Body.

Same thing for the `BodyText SizesAndWeights` story! 

It would also be useful to have visual regression tests for both OG and TB themes!

__docs__/wonder-blocks-typography/typography.stories.tsx
```diff
@@ -73,34 +77,47 @@ export const ControlProps: StoryObj<typeof Title> = {
...
+            <BodySerif>BodySerif</BodySerif>
+            <BodyMonospace>BodyMonospace</BodyMonospace>
+            <Body>Body</Body>
+            <LabelLarge>LabelLarge</LabelLarge>
+            <LabelMedium>LabelMedium</LabelMedium>
+            <LabelSmall>LabelSmall</LabelSmall>
+            <LabelXSmall>LabelXSmall</LabelXSmall>
+            <Tagline>Tagline</Tagline>
+            <Caption>Caption</Caption>
+            <Footnote>Footnote</Footnote>
```
beaesguerra, line 118: It would also be great to have visual regression tests for the classic typography components for both OG and TB themes (to make sure nothing is looking wonky!)

beaesguerra (commented):
Looks great so far Marcy!! Left some comments and questions 😄

packages/wonder-blocks-tokens/src/theme/primitive/font.ts
```diff
@@ -40,7 +42,8 @@ export const font = {
...
     weight: {
         light: fontWeight.light,
-        regular: fontWeight.regular,
```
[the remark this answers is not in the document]
marcysutton: Yes, this will become a major!

__docs__/wonder-blocks-typography/typography.stories.tsx
```diff
@@ -73,34 +77,47 @@ export const ControlProps: StoryObj<typeof Title> = {
...
-        description: {
-            story: `These are all the available typography elements
-            with their names written out in their respective styles.`,
-        },
-    },
+/**
+These are all the available classic typography elements with their names written
+out in their respective styles.
+ */
+export const ClassicTypographyElements: StoryObj<any> = {
```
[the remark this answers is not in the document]
marcysutton: I'll have to tackle those much later on. A mapping guide would be helpful, I opened an issue for it here: https://khanacademy.atlassian.net/browse/WB-1955

As far as deprecation for the other typography components, I think we'd need to discuss it a bit before making that call! At the very least, it should be obvious that a) there are new components and b) how they work! The guide piece could cover that.

packages/wonder-blocks-typography/src/components/heading.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from "react";
+import {css} from "aphrodite";
+import {PropsFor, Text} from "@khanacademy/wonder-blocks-core";
+import {font} from "@khanacademy/wonder-blocks-tokens";
+import styles from "../util/styles";
+
+type Props = PropsFor<typeof Text> & {
+    size?: "small" | "medium" | "large" | "xLarge" | "xxLarge";
+    tag?: "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
```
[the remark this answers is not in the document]
marcysutton: I'm hesitant to add it so it doesn't get misused. Headings should be marked up as headings! One use case I considered was a `Heading` inside of a larger piece of content that _was_ marked up with a heading tag. But that seems confusing to me, and not super reliable in VoiceOver where content broken up with multiple line breaks or layout features gets split apart. I'd say let's wait and see if we need to support it.

packages/wonder-blocks-typography/src/index.ts
```diff
@@ -30,16 +33,17 @@ export type Heading =
 /**
  * Typography components for representing body text.
  */
-export type BodyText =
+export type BodyComponents =
```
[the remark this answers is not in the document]
marcysutton: Good call. I don't think these types are used anywhere else, as there is even a `Heading` class defined in webapp that would collide with this type definition. But I added it to the changeset anyway!

packages/wonder-blocks-typography/src/util/styles.ts
```diff
@@ -23,10 +23,22 @@ const styles: StyleDeclaration = StyleSheet.create({
...
+        // weight and size are determined by props
+        // lineHeight is determined by fontSize on REM scale
+    },
+    BodyText: {
+        ...common,
+        fontFamily: font.family.sans,
+        // weight and size are determined by props
+        // lineHeight is determined by fontSize on REM scale
```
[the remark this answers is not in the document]
marcysutton: I like the idea of having easily reusable defaults, as they can save time! I think I will opt to wait on that one to keep this work moving. I opened an issue for it here: https://khanacademy.atlassian.net/browse/WB-1956

__docs__/wonder-blocks-typography/body-text.stories.tsx
```diff
@@ -0,0 +1,106 @@
...
+[line-height scale](/?path=/docs/packages-tokens-typography--docs&globals=theme:thunderblocks#lineHeight).
+
+The `weight` prop will match a font weight token based on the available
+[font weights](/?path=/docs/packages-tokens-typography--docs&globals=theme:thunderblocks#weight).
+
+Heading text should utilize the `Heading` component.
+**/
+
+export default {
+    title: "Packages / Typography / Thunderblocks / BodyText",
```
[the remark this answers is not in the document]
marcysutton: I like that idea. There might be quite a few links we have to change, as I don't think Storybook handles 404s!

__docs__/wonder-blocks-typography/body-text.stories.tsx
marcysutton: To be honest, I don't know on the classic experience part. But we could chat about it and try to cover it in this docs issue: https://khanacademy.atlassian.net/browse/WB-1955

__docs__/wonder-blocks-typography/heading.stories.tsx
marcysutton: I adjusted the story settings for the Typography components and opened an issue for the theme visual regression tests here: https://khanacademy.atlassian.net/browse/WB-1957

__docs__/wonder-blocks-typography/body-text.stories.tsx
jandrade: The short answer is yes, teams should use these `BodyText` and `Heading` components in the classic experience.

They are going to use them as soon as we start integrating these components in the `Classroom` experience as there are some UI components that are shared between Classic and Classroom (the ones that live inside the `static/libs` folder.

In general, our strategy will be to create new components having in mind that these should always fallback to the classic experience.

__docs__/wonder-blocks-typography/body-text.stories.tsx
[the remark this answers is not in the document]
jandrade: It makes sense for now. If you can make that change later in a separate PR would be great! One thing that I personally like of using "New" instead of `ThunderBlocks` is that it signals that these are new components, but are not exclusive to a single theme, and these can be used in both themes/contexts.

__docs__/wonder-blocks-typography/heading.stories.tsx
jandrade: For visual regression tests, you can use the IconButton example as reference. It allow us to create snapshots for both themes using the same story:

https://github.com/Khan/wonder-blocks/blob/main/__docs__/wonder-blocks-icon-button/icon-button-testing-snapshots.stories.tsx#L27-L33

__docs__/wonder-blocks-typography/typography.stories.tsx
jandrade: ++ to this, you can use Chromatic modes for this (see my previous example). Adding to that, this is probably something that could be covered with a `typography-testing-snapshots` file that uses the `StateSheet` or `AllVariants` custom components.

packages/wonder-blocks-tokens/src/theme/semantic/internal/primitive-font-thunderblocks.ts
```diff
@@ -0,0 +1,52 @@
...
+    lineHeight: {
+        xxxLarge: sizing.size_400,
+        xxLarge: sizing.size_320,
+        xLarge: sizing.size_280,
+        large: sizing.size_240,
+        xMedium: sizing.size_220,
+        medium: sizing.size_200,
+        small: sizing.size_180,
+        xSmall: sizing.size_160,
+    },
```
jandrade, line 44: @marcysutton based on the Chromatic snapshots, it looks like there are differences between these tokens and how these are defined in Figma, causing the new typography components not to look like the Figma specs.

__docs__/wonder-blocks-typography/body-text.stories.tsx
marcysutton: Related to our longer discussion on Slack, we'll have to map the theme for these components for the Classic experience. TB uses a different size and line-height scale (Heading and Body styles, vs. global font tokens). I pushed an update to add component-level theming for these components, so there is a space to add the classic values. But they'll have to be determined first!

packages/wonder-blocks-tokens/src/theme/semantic/internal/primitive-font-thunderblocks.ts
marcysutton: Thank you for surfacing this @jandrade! It uncovered a nuance of the size and line-height scale that I'd missed. I refactored the Typography components to use theming and they *should* be correct now.

packages/wonder-blocks-tokens/src/theme/primitive/font.ts
```diff
@@ -20,27 +21,28 @@ export const font = {
         mono: fontFamily.mono,
     },
     size: {
-        xxxLarge: sizing.size_360,
-        xxLarge: sizing.size_280,
-        xLarge: sizing.size_240,
+        xxxlarge: sizing.size_360,
```
marcysutton, line 24: **Note:** I changed these to lowercase as we discussed for a consistent API across components.

packages/wonder-blocks-typography/src/theme/default.ts
```diff
@@ -0,0 +1,37 @@
+import {sizing} from "@khanacademy/wonder-blocks-tokens";
+
+// Font sizes and lineheights for Thunderblocks
+export default {
```
marcysutton, line 4: **Note:** I added component-level theming to handle the different sizes/line-heights for TB. But these should probably be in a `thunderblocks` theme, and actual `default` values need to be defined for Classic fallbacks.

__docs__/wonder-blocks-typography/typography.stories.tsx
marcysutton: I enabled Chromatic modes for this story and the SizesAndWeights stories. Do you think that is sufficient for covering the components for regressions? I could add an `AllVariants` snapshot for all of the components as a follow-up PR.

__docs__/wonder-blocks-typography/body-text.stories.tsx
marcysutton: Classic sizing has been added to this PR, so `BodyText` and `Heading` can be used outside of the Classroom theme!

__docs__/wonder-blocks-typography/heading.stories.tsx
marcysutton: I added the modes to this story, and I see what you mean! That is going to be helpful.

packages/wonder-blocks-typography/src/util/styles.ts
marcysutton: Here's an example to see the Jakarta font loaded for older components, unless overridden with something else (like a serif or monospace): https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=682fa8c1c9b6c708b4be5a70

__docs__/wonder-blocks-typography/body-text.stories.tsx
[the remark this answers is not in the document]
marcysutton: Note: I ended up keeping `BodyText (New)` and `Heading (New)` so they would stand out among the other components. Putting the new ones in a directory made them less obvious.

packages/wonder-blocks-tokens/src/theme/primitive/font.ts
marcysutton: **Note:** One additional change: I kept the older to-be-deprecated font API camelCase to limit the scope of these changes. The new structure is lowercase.

packages/wonder-blocks-tokens/src/theme/primitive/font.ts
```diff
@@ -0,0 +1,80 @@
...
+    heading: {
+        size: {
+            small: sizing.size_120, // HeadingXSmall
+            medium: sizing.size_200, // HeadingSmall and Tagline
+            large: sizing.size_240, // HeadingMedium
+            xlarge: sizing.size_280, // HeadingLarge
+            xxlarge: sizing.size_360, // Title
+        },
+        lineHeight: {
+            small: sizing.size_200, // HeadingXSmall
```
jandrade, line 73: Oops, my bad.... I calculated the wrong size initially.

```suggestion
            small: sizing.size_160, // HeadingXSmall
```

__docs__/wonder-blocks-typography/heading.stories.tsx
jandrade: For some reason that I don't understand, I'm not able to see these new snapshots being captured in Chromatic.

__docs__/wonder-blocks-typography/typography.stories.tsx
```diff
@@ -27,23 +29,27 @@ import packageConfig from "../../packages/wonder-blocks-typography/package.json"
...
 import TypographyArgTypes from "./typography.argtypes";
+import {allModes} from "../../.storybook/modes";
 
 // NOTE: Only for testing purposes.
 // eslint-disable-next-line import/no-unassigned-import
 import "./styles.css";
 
-const typographyDescription = `Typography. \`wonder-blocks-typography\`
+/**
+Typography. `wonder-blocks-typography`
```
jandrade, line 39: **praise:** Yay! thanks for making this change 💯

__docs__/wonder-blocks-typography/typography.stories.tsx
```diff
@@ -73,34 +78,62 @@ export const ControlProps: StoryObj<typeof Title> = {
...
+ */
+export const ClassicTypographyElements: StoryObj<any> = {
+    parameters: {
+        chromatic: {
+            modes: {
+                default: allModes.themeDefault,
+                thunderblocks: allModes.themeThunderBlocks,
+            },
         },
     },
```
jandrade, line 116: same question related to a previous comment.... I couldn't see the snapshot changes in the recent Chromatic build.

__docs__/wonder-blocks-typography/typography.stories.tsx
jandrade: Probably adding `AllVariants` would be helpful as we started standardizing on the `Testing > [component name] - Snapshots` (and `*-testing-snapshots.stories.tsx` file) to know when we want to cover different variants of our components in Chromatic. Follow-up PR sounds great!

packages/wonder-blocks-tokens/src/theme/primitive/font.ts
jandrade, line 73: Oops, my bad.... I calculated the wrong size initially.

```suggestion
            small: sizing.size_160, // HeadingXSmall
```

jandrade (commented):
Thanks for the recent changes, everything is working really great! I left some minor comments before giving final approval (I shared a wrong OG token value, and Chromatic snapshots not being captured in the build step).

Solid stuff @marcysutton 👏

jandrade (approved):
This looks awesome!!! Thanks so much for implementing this key part of the system! 🚀 🚢

PR #2603, opened 2025-05-14
Chromatic: Migrate All Variants stories to use StateSheet

jandrade:
## Summary:

Updates all the existing Chromatic stories that used the "All Variants"
snapshots to use the new StateSheet component.

This also includes the following changes:

- Added `rtl` and `dark` modes to the `modes.ts` file.
- Updated the `Link` snapshots file to include test `rtl` and `dark` modes.
- Refactored all the other stories to use `StateSheet`.

Issue: "none"

## Test plan:

Verify that the new `*-testing-snapshots.stories.tsx` files are looking as
expected and that the are still covering all the different visual states
of the components.


__docs__/wonder-blocks-link/link-variants.stories.tsx
```diff
@@ -1,154 +0,0 @@
...
-export const Focus: Story = {
-    parameters: {pseudo: {focusVisible: true}},
-};
-
-export const HoverFocus: Story = {
-    name: "Hover + Focus",
-    parameters: {pseudo: {hover: true, focusVisible: true}},
-};
-
-export const HoverVisited: Story = {
```
beaesguerra, line 115: **suggestion:** The current AllVariants story for Link includes Hover + Visited, Focus + Visited, Press + Visited states. It would be helpful to include those states in the StateSheet too so we don't have any visual regressions with those! (same with a visited state on it's own!)

__docs__/wonder-blocks-pill/pill-testing-snapshots.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+    {name: "Success", props: {kind: "success"}},
+    {name: "Warning", props: {kind: "warning"}},
+    {name: "Critical", props: {kind: "critical"}},
+];
+
+const rows = [
+    {
+        name: "Default",
+        props: {onClick: () => {}},
+    },
```
beaesguerra, line 25: **suggestion:** Can we also include the static kind? Similar to [what we have currently with AllVariants](https://khan.github.io/wonder-blocks/?path=/story/packages-pill-pill-all-variants--default). It'll be helpful until badges fully replace Pill!

__docs__/wonder-blocks-pill/pill-testing-snapshots.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+        name: "Medium",
+        props: {size: "medium"},
+    },
+    {
+        name: "Large",
+        props: {size: "large"},
+    },
+];
+
+export const Sizes: Story = {
```
beaesguerra, line 94: **suggestion:** Could we also show the sizes for both static and clickable types so we get coverage on the different combinations?

__docs__/wonder-blocks-link/link-testing-snapshots.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+    render: (args, {globals}) => {
+        const isRTL = globals.direction === "rtl";
+        const isDark =
+            globals.backgrounds?.value === semanticColor.surface.inverse;
+        const columnsPerMode = isRTL ? rtlColumns : columns;
+
+        return (
+            <StateSheet rows={rows} columns={columnsPerMode}>
+                {({props, className, name}) => (
+                    // eslint-disable-next-line jsx-a11y/anchor-has-content
```
beaesguerra, line 91: Would explicitly providing the children address the lint error?

beaesguerra (commented):
Thanks for updating things to use the StateSheet! ❤️ 

I left some non-blocking comments and suggestions. One thing I wanted to get your thoughts on before approving was the empty cells in the [OptionItem StateSheet](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=682510c7fc9b7e683b028439)!

__docs__/wonder-blocks-link/link-variants.stories.tsx
jandrade: ahhh good question! turns out that we removed support to the `:visited` state, so this story/variant doesn't apply anymore. I probably forgot to remove these stories.

__docs__/wonder-blocks-pill/pill-testing-snapshots.stories.tsx
jandrade: I didn't add it as the static pills would only be helpful with the `rest` state, which is already covered by the "clickable" ones (making them kinda duplicate visual tests). Do you think it would be still helpful to include them?

__docs__/wonder-blocks-link/link-testing-snapshots.stories.tsx
jandrade: hmmm I don't remember what happened, but let me confirm if that fixes the issue :)

__docs__/wonder-blocks-pill/pill-testing-snapshots.stories.tsx
jandrade: Kinda similar answer to the previous comment. The coverage would be the same as it would only be testing the `rest` state of clickable pills (which is shared with the static ones).

jandrade: To be fair, it makes sense to include clickable here as this is the Pill variant that will remain after releasing Badge.

__docs__/wonder-blocks-pill/pill-testing-snapshots.stories.tsx
beaesguerra: Yes please! 😄 More so we can confirm that there are no unintentional changes to Pill until the static parts are removed in favour of Badge. Especially since removing the static Pill will be a later task! (And I don't want to break things from the OG theme until then!)

__docs__/wonder-blocks-pill/pill-testing-snapshots.stories.tsx
beaesguerra: Similar to the other thread, would be helpful to have in the short term while Pill and Badge are still overlapping! We can remove the extra stories once the static functionality is removed from Pill 😄

I also made a note of this in https://khanacademy.atlassian.net/browse/WB-1950 !

__docs__/wonder-blocks-link/link-variants.stories.tsx
beaesguerra: Ahh I see that makes sense! We could keep the visited states around in the StateSheet so we'll know if we unintentionally apply visited styles. I'll leave it up to you!

beaesguerra (approved):
Thanks for updating the StateSheet stories! 

I added comments to the previous threads ([visited links](https://github.com/Khan/wonder-blocks/pull/2603#discussion_r2092055897), [pill statesheet](https://github.com/Khan/wonder-blocks/pull/2603#discussion_r2093711884)) on snapshots that could be helpful to keep, though they are suggestions and not required 😄

__docs__/wonder-blocks-link/link-testing-snapshots.stories.tsx
jandrade: That fixed it!

PR #2604, opened 2025-05-15
Export ActionTrigger type for Banner

MikeKlemarewski:
## Summary:
Typescript doesn't seem to infer action types correctly when passing a memoized array to the Banner component.
Exporting the ActionTrigger type should allow us to properly define the type.

### DIFF
--- a/.changeset/mighty-chairs-roll.md
+++ b/.changeset/mighty-chairs-roll.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-banner": major
+---
+
+Export ActionTrigger type
--- a/packages/wonder-blocks-banner/src/components/banner.tsx
+++ b/packages/wonder-blocks-banner/src/components/banner.tsx
@@ -42,7 +42,7 @@ type ActionTriggerCustom = {
     node: React.ReactNode;
 };
 
-export type ActionTrigger =
+type ActionTrigger =
     | ActionTriggerWithButton
     | ActionTriggerWithLink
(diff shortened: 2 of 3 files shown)


packages/wonder-blocks-banner/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@khanacademy/wonder-blocks-banner",
-  "version": "4.1.16",
+  "version": "4.1.17",
```
jandrade, line 3: Could you please revert this change? `pnpm changeset` should take care of this for us as soon as these changes are released to npm.

MikeKlemarewski:
Screenshot 2025-05-15 at 1 31 35 PM

I installed this version into Classroom and verified that I can get around any TS errors with the updated ReadonlyArray on the prop

jandrade (approved):
This makes total sense! thanks for fixing the types 🚀

.changeset/mighty-chairs-roll.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-banner": major
```
jandrade, line 2: forgot to mention, this would be a `patch` change.
```suggestion
"@khanacademy/wonder-blocks-banner": patch
```

jandrade: @MikeKlemarewski I missed the `major` part in changesets, sorry. You'll need to change that to become `patch`. This PR will need to be reverted.

PR #2608, opened 2025-05-16
Fix changeset to apply patch instead of major version bump

MikeKlemarewski:
## Summary:
My previous PR applied a major version change instead of a patch.
This should fix that.


jandrade (approved):
Thanks!

PR #2610, opened 2025-05-20
Adds `Icon` component to support custom icons.

beaesguerra:
## Summary:

- Add `Icon` component to support custom icons that are multicolored (PhosphorIcon component doesn't support icons that provide their own color fill). `PhosphorIcon` should still be used for single filled icons
- `Icon` works with `img` elements or inline `svg` elements. Pros for supporting inline svgs:
  - multicolored svgs can be themable by using css variables for path fills (will be implementing this in an upcoming `GemIcon` component, which will be used by an upcoming `GemBadge` component)
  - more flexibility around svgs (ex: could animate parts of an svg. This wouldn't be possible if the svg was used in an `img` element)
- `Icon` supports a `size` prop that matches the `PhosphorIcon` sizes
- Integrated Icon with the Badge component examples

Note: I created WB-1954 for tracking work to support the new Icon component in other WB components!

Issue: WB-1921

## Test plan:
`?path=/docs/packages-icon-icon--docs`
- The new Icon component looks as expected
- Icon docs are reviewed

`?path=/story/packages-badge-badge--custom-icons`
- Badge docs and examples around custom icons are reviewed and work as expected


packages/wonder-blocks-badge/src/components/badge.tsx
```diff
@@ -165,8 +165,13 @@ const styles = StyleSheet.create({
         color: badgeTokens.root.color.foreground,
     },
     icon: {
-        width: badgeTokens.icon.sizing.width,
-        height: badgeTokens.icon.sizing.height,
+        // Use minWidth and minHeight to ensure custom sized icons don't
```
beaesguerra, line 168: I created https://khanacademy.atlassian.net/browse/WB-1954 for tracking work to support the new Icon component in other WB components

beaesguerra:
I'm going to re-visit the Icon props API so that we can support inline svgs. This is helpful for multicolored themed icons (such as the gem/streak icons for the gem/streak badges!)

packages/wonder-blocks-badge/src/components/badge.tsx
```diff
@@ -165,8 +165,13 @@ const styles = StyleSheet.create({
...
+        // overflow the badge
```
beaesguerra, line 169: This is more for if PhosphorIcon or Icon components aren't used with Badge (as a backup)

beaesguerra:
Updated this PR to support inline svg icons (to better support theming). It's ready for review again!

__docs__/wonder-blocks-badge/badge.argtypes.ts
```diff
@@ -7,14 +7,15 @@ export default {
         mapping: IconMappings,
         table: {
             type: {
-                summary: "ReactElement",
+                // `icon` prop can be any ReactElement, however, we want to
+                // encourage the use of `PhosphorIcon` or `Icon` components.
```
jandrade, line 11: **praise:** Thanks for adding this note!

packages/wonder-blocks-icon/src/components/icon.tsx
```diff
@@ -0,0 +1,113 @@
...
+                // Make sure the children element takes up the width/height of
+                // the Icon component div
+                width: "100%",
+                height: "100%",
+            },
+        });
+
+        return (
+            <StyledDiv
+                style={[getSize(size), style]}
```
jandrade, line 82: **suggestion:** Would it make sense to simplify the size part by invoking the style directly here?

```suggestion
                style={[styles[size], style]}
```

jandrade (approved):
This looks great! thanks for working on this as it will unblock some Classroom-related components 🚀

packages/wonder-blocks-icon/src/components/icon.tsx
beaesguerra: Good question! I was thinking there was less type safety! For example, if a style for a size doesn't exist, there's no warning. Though I'm seeing now that if I access something on `styles` that doesn't exist, there's no warning either 🤔 We do get visual coverage from the Chromatic snapshots though!

Do you have any suggestions on how to make it more type safe?

PR #2614, opened 2025-05-22
Bump TypeScript target to ES2021

jeremywiebe:
## Summary:

This PR upgrades the TypeScript `target` to ES2021. The current minimum browser list at Khan Academy is: 
  * Chrome v128
  * Safari v15.6
  * Latest Edge
  * Latest Firefox 

All of these browsers fully support ES2021 (verified by looking through the linked features of https://caniuse.com/?search=ES2021). By bumping to ES2021 we give ourselves access to all ES2021 features without risking shipping code that is incompatible with our supported browser list.

This also cleans up the 'external' flag in our Rollup build (it referenced `@babel/preset-env which we no longer use and `core-js` is no longer a dependency of the project, further we don't use core-js because we don't need any polyfills for ES2021 support).

Related: https://github.com/Khan/perseus/pull/2511

Issue: "none"

## Test plan:

I ran `pnpm build; pnpm build:types` on `main` and on this branch and compared the `dist/` folder for each package. They were identical (ie. nothing was using `core-js` on `main` today and nothing changed). 

I ran `pnpm tsc` - no issues.

### DIFF
--- a/build-settings/rollup.config.mjs
+++ b/build-settings/rollup.config.mjs
@@ -27,16 +27,17 @@ const createConfig = (pkgName) => {
             },
         ],
         input: `packages/${pkgName}/src/index.ts`,
-        // We're using `builtIns: "usage"` with @babel/preset-env.
-        // This results in individual modules being imported from
-        // core-js directly.  `autoExternal` doesn't know how to
-        // deal with this so we manually externalize these imports.
-        external: [/core-js/],
         plugins: [
             swc({
                 swc: {
                     swcrc: true,
                     minify: true,
+                    // We do _not_ specify "env" here (a la @babel/preset-env)
+                    // because our TypeScrip compiler "target" is set to ES2021
+                    // which is compatible with all of Khan Academy's supported
+                    // browsers _and_ will protect us against using APIs that
+                    // aren't supported in this browser list).
+                    // "env": {...}
                 },
                 exclude: "node_modules/**",
             }),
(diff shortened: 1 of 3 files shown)


build-settings/rollup.config.mjs
```diff
@@ -27,16 +27,17 @@ const createConfig = (pkgName) => {
...
-        // core-js directly.  `autoExternal` doesn't know how to
-        // deal with this so we manually externalize these imports.
-        external: [/core-js/],
         plugins: [
             swc({
                 swc: {
                     swcrc: true,
                     minify: true,
+                    // We do _not_ specify "env" here (a la @babel/preset-env)
+                    // because our TypeScrip compiler "target" is set to ES2021
```
jandrade, line 36: super-nit(typo):
```suggestion
                    // because our TypeScript compiler "target" is set to ES2021
```

Thanks for adding this comment! :)

build-settings/rollup.config.mjs
```diff
@@ -27,16 +27,17 @@ const createConfig = (pkgName) => {
             },
         ],
         input: `packages/${pkgName}/src/index.ts`,
-        // We're using `builtIns: "usage"` with @babel/preset-env.
-        // This results in individual modules being imported from
-        // core-js directly.  `autoExternal` doesn't know how to
-        // deal with this so we manually externalize these imports.
-        external: [/core-js/],
```
jandrade, line 34: oh good finding! thanks for cleaning up.

jandrade (approved):
Looks great! 🚀

PR #2616, opened 2025-05-22
Add GemBadge and GemIcon components

beaesguerra:
## Summary:
- Adds `GemBadge` component. It uses the `Badge` component with a pre-defined style and icon using the new `Icon` component
  - Since the icon is pre-defined, it accepts a `showIcon` prop instead of an `icon` prop like the `Badge` and `StatusBadge` components
  - Will be updating it to semantic color tokens in an upcoming PR
- Adds `GemIcon` custom icon component. It can be used with the `Icon` component and renders an inline svg so that we can use semantic color tokens for the fill so that the multi-colored svg is themable

GemBadge:
Gem Badge component with a label and gem icon

GemIcon:
Gem Icon component

Issue: WB-1943

## Test plan:
- Review docs and components
  - Badge Overview (`?path=/docs/packages-badge-overview--docs`) 
  - GemBadge (`?path=/docs/packages-badge-gembadge--docs`)
  - Custom Icon Components (includes GemIcon) (`?path=/docs/packages-icon-custom-icon-components--docs`)

### DIFF
--- a/__docs__/wonder-blocks-badge/_overview_.mdx
+++ b/__docs__/wonder-blocks-badge/_overview_.mdx
@@ -17,9 +17,13 @@ about each type of badge, refer to the docs for that component!
 
 ## StatusBadge
 
+A badge that represents a status.
+
 <Canvas of={StatusBadgeStories.Kinds} />
 
 ## GemBadge
+
+ A badge that represents gems.
 
 <Canvas of={GemBadgeStories.Default} />
 
--- a/__docs__/wonder-blocks-badge/_overview_.mdx
+++ b/__docs__/wonder-blocks-badge/_overview_.mdx
@@ -1,6 +1,7 @@
 import {Meta, Canvas} from "@storybook/blocks";
 import * as StatusBadgeStories from "./status-badge.stories";
 import * as BadgeStories from "./badge.stories";
+import * as GemBadgeStories from "./gem-badge.stories";
 
 <Meta
     title="Packages / Badge / Overview"
@@ -17,6 +18,10 @@ about each type of badge, refer to the docs for that component!
 ## StatusBadge
 
 <Canvas of={StatusBadgeStories.Kinds} />
+
+## GemBadge
+
+<Canvas of={GemBadgeStories.Default} />
 
 ## Custom Badges
 
--- a/__docs__/wonder-blocks-icon/custom-icon-components.stories.tsx
+++ b/__docs__/wonder-blocks-icon/custom-icon-components.stories.tsx
@@ -1,0 +1,88 @@
+import * as React from "react";
+import type {Meta, StoryObj} from "@storybook/react";
+import ComponentInfo from "../components/component-info";
+import packageConfig from "../../packages/wonder-blocks-icon/package.json";
+import {GemIcon} from "@khanacademy/wonder-blocks-icon";
+import {View} from "@khanacademy/wonder-blocks-core";
+import {border, semanticColor, sizing} from "@khanacademy/wonder-blocks-tokens";
+
+/**
+ * Custom icon components that render an inline svg. Use with the `Icon`
+ * component to display the icon.
+ */
+export default {
+    title: "Packages / Icon / Custom Icon Components",
+    parameters: {
+        componentSubtitle: (
+            <ComponentInfo
+                name={packageConfig.name}
+                version={packageConfig.version}
+            />
+        ),
+        chromatic: {
+            disableSnapshot: true,
+        },
+    },
+    component: GemIcon,
+    argTypes: {
+        style: {
+            table: {
+                type: {
+                    summary: "StyleType",
+                },
+            },
+        },
+    },
+    decorators: [
+        (Story) => (
+            // Set a fixed size for the custom icons since they expand to the
+            // size of its parent container.
+            <View style={{width: sizing.size_400, height: sizing.size_400}}>
+                <Story />
+            </View>
+        ),
+    ],
+} as Meta<typeof GemIcon>;
+
+type StoryComponentType = StoryObj<typeof GemIcon>;
+
+export const AllCustomIcons: StoryComponentType = {
+    render: (args) => (
+        <View style={{gap: sizing.size_160}}>
+            <GemIcon {...args} />
+            {/* Add other custom icons here */}
+        </View>
+    ),
+    parameters: {
+        chromatic: {
+            // Include snapshots for all the custom icons
+            disableSnapshot: false,
+        },
+    },
+};
+
+/**
+ * Custom icons can be styled using the `style` prop.
+ */
+export const CustomIconsWithCustomStyle: StoryComponentType = {
+    ...AllCustomIcons,
+    args: {
+        style: {
+            backgroundColor: semanticColor.surface.secondary,
+            padding: sizing.size_040,
+            borderRadius: border.radius.radius_040,
+            border: `${border.width.thin} solid ${semanticColor.border.subtle}`,
+        },
+    },
+};
+
+/**
+ * Use the `GemIcon` component to represent gems.
+ */
+export const Gem: StoryComponentType = {
+    name: "GemIcon",
+    render: (args) => <GemIcon {...args} />,
+    args: {
+        "aria-label": "Gem",
+    },
+};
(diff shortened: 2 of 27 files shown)


packages/wonder-blocks-icon/src/hooks/use-svg-attributes.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Determines what attributes should be applied to an icon based on if there is
+ * an accessible label for the icon.
+ *
+ * @returns The attributes to apply to an svg element.
+ */
+export function useSvgAttributes(props: {
```
beaesguerra, line 7: Utility for the custom icon components (like GemIcon). It determines what attributes should be used depending on if `aria-label` or `aria-labelledby` is provided.

I've seen a variety of ways to mark svgs as presentation only or as images with meaning, so let me know if you have any feedback on how we'd like to mark our svg icons!

I decided on these attributes as they aligned with:
- an article around svg a11y: [Contextually Marking up accessible images and SVGs](https://www.scottohara.me/blog/2019/05/22/contextual-images-svgs-and-a11y.html)
   - Note: I didn't include the `focusable=false` attribute since we don't support IE:
  - >focusable="false" is also used to ensure Internet Explorer won’t allow the Tab key to navigate into the SVG.
- our [Icon Accessibility Docs](https://khan.github.io/wonder-blocks/?path=/docs/packages-icon-accessibility--docs)
  - So it is consistent with our current recommendations!

marcysutton: `role=presentation` is probably the most common method I've seen for marking SVGs as decorative. But `aria-hidden="true"` makes sense, too! I think the most meaningful choice is whatever works in supported ATs (especially VO with Safari, since it's the buggiest combo!)

So long, IE! 😂

packages/wonder-blocks-badge/src/components/gem-badge.tsx
```diff
@@ -0,0 +1,52 @@
...
+        iconAriaLabel?: string;
+    };
+} & BaseBadgeProps &
+    ShowIconProps;
+
+/**
+ * A badge that represents gems.
+ */
+const GemBadge = React.forwardRef<HTMLDivElement, Props>((props, ref) => {
+    const {label, showIcon, labels, ...otherProps} = props;
```
marcysutton, line 22: **Question:** are there both `label` and `labels` props here? Can they be combined into one object, or do they need to stay separate? I could see it being a little confusing to consumers to have both.

__docs__/wonder-blocks-badge/_overview_.mdx
```diff
@@ -18,6 +19,10 @@ about each type of badge, refer to the docs for that component!
 
 <Canvas of={StatusBadgeStories.Kinds} />
 
+## GemBadge
+
```
marcysutton, line 23: **Suggestion:** Maybe you're already tackling this in a future PR, but a short description of the GemBadge would be helpful here!

packages/wonder-blocks-badge/src/components/gem-badge.tsx
```diff
@@ -0,0 +1,52 @@
...
+     */
+    labels?: {
+        /** The alt text for the gem icon. */
+        iconAriaLabel?: string;
+    };
+} & BaseBadgeProps &
+    ShowIconProps;
+
+/**
+ * A badge that represents gems.
```
marcysutton, line 19: **Suggestion:** Since this shows up in the Storybook docs page, maybe it could briefly describe the purpose of a GemBadge?

__docs__/wonder-blocks-icon/custom-icon-components.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+import {GemIcon} from "@khanacademy/wonder-blocks-icon";
+import {View} from "@khanacademy/wonder-blocks-core";
+import {border, semanticColor, sizing} from "@khanacademy/wonder-blocks-tokens";
+
+/**
+ * Custom icon components that render an inline svg. Use with the `Icon`
+ * component to display the icon.
+ */
+export default {
+    title: "Packages / Icon / Custom Icon Components",
```
marcysutton, line 14: **Praise:** VERY cool. Thanks for adding these!

marcysutton (commented):
The GemBadge is so cute! Looking good so far. I left a few comments and questions!

packages/wonder-blocks-badge/src/components/__tests__/gem-badge.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+    it("should render the label", () => {
+        // Arrange
+        const label = "Badge label";
+        render(<GemBadge label={label} />);
+
+        // Act
+        const badge = screen.getByText(label);
+
+        // Assert
+        expect(badge).toBeInTheDocument();
```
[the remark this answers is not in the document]
beaesguerra: Updated to check for `toBeVisible` instead!

packages/wonder-blocks-badge/src/components/gem-badge.tsx
beaesguerra: Good question! I added the `labels` prop since we normally use that prop for an object of translated strings in WB components. And then the `label` is the main prop for adding text to the Badge (it didn't seem right to nest an important prop in the `labels` object!).

Since there's only one key in the `labels` prop, I removed the `labels` prop and added an `iconAriaLabel` prop instead so it is no longer nested! This is similar to how `SearchField` has a `clearAriaLabel` prop!

Let me know what you think!

__docs__/wonder-blocks-badge/_overview_.mdx
beaesguerra: Added!

packages/wonder-blocks-icon/src/hooks/use-svg-attributes.ts
beaesguerra: Thanks for the feedback! I was considering `role=presentation` as well! Then I saw our icon guidelines recommended `aria-hidden` and so did the svg a11y article mentioned earlier in this thread:

> Note that while some screen readers may ignore an SVG if it has no role or accessible name, other screen readers may still find the element and announce it as a “group” without an accessible name. It’s best to avoid these situations by always using aria-hidden="true" if the SVG is meant to be decorative.

 I confirmed that with Safari + VoiceOver, Chrome + NVDA, and Firefox + NVDA that the gem icon is not read out when aria-hidden is used!

packages/wonder-blocks-badge/src/components/gem-badge.tsx
```diff
@@ -0,0 +1,52 @@
...
+import {StyleSheet} from "aphrodite";
+import {GemIcon, Icon} from "@khanacademy/wonder-blocks-icon";
+import {Badge} from "./badge";
+import {BaseBadgeProps, ShowIconProps} from "../types";
+
+type Props = {
+    /**
+     * The alt text for the gem icon.
+     */
+    iconAriaLabel?: string;
```
jandrade, line 11: **question:** Do you think we should provide a default value for labelling this icon?

jandrade: If not, maybe this should be required so we could enforce fully accessible GemBadge components?

packages/wonder-blocks-icon/src/hooks/use-svg-attributes.ts
jandrade, line 1: **praise:** This is super useful!!

packages/wonder-blocks-icon/src/hooks/use-svg-attributes.ts
```diff
@@ -0,0 +1,29 @@
...
+        role: "img",
+    };
+
+    const attributes =
+        ariaLabel || ariaLabelledBy
+            ? iconMeaningAttributes
+            : presentationOnlyAttributes;
+
+    return attributes;
+}
```
jandrade, line 29: **question:** Do you see this being useful for other kind of icons? I wonder if this hook should be part of the `icon` public API?

jandrade (approved):
This is really great! ship it 🚀

packages/wonder-blocks-badge/src/components/gem-badge.tsx
beaesguerra: Great question! I don't think there should be a default value or be fully required in case the badge label is meaningful on it's own and the icon is decorative only. That said, it should be required if there is no label and it's an icon only badge! 

I updated the prop types to enforce this! So if `showIcon` is `true` and there is no `label`, `iconAriaLabel` will be a required prop. It is optional if both the icon and label are present

PR #2617, opened 2025-05-23
[WB-1941.2] Add core.background and core.foreground semantic color tokens

jandrade:
## Summary:

- Replaced some `action` call sites with `core.background` tokens.
- Added `core.background` and `core.foreground` tokens in `OG` and `TB`.
- Updated `press` action states to use `core.subtle` (`fadedBlue8`) instead of `fadedBlue`.

### Implementation plan:

1. #2613
**2. Add `core.background` and `core.foreground` semantic color tokens (CURRENT)**

Issue: WB-1941

## Test plan:

Verify that the Chromatic snapshots look as expected.

Note that `Button`, `IconButton`, `Action/OptionItem` and `Checkbox`/`Radio` will change as we changed the press background color.

https://khanacademy.slack.com/archives/C07CA2YR0BZ/p1747931528405019

### DIFF
--- a/__docs__/foundations-color.mdx
+++ b/__docs__/foundations-color.mdx
@@ -265,6 +265,55 @@ For all type to ensure contrast for legibility. Inverse text applies for dark
 backgrounds in light mode.
 
 <Banner
+    kind="warning"
+    layout="full-width"
+    text="These tokens are deprecated and will be removed in the future. Please use the semanticColor.core.foreground tokens instead."
+/>
+
+<ColorGroup colors={semanticColor.text} group="text" />
+
+### Core
+
+Core colors are used for the most common elements in the UI. These colors are
+used for backgrounds, borders, and text. They are the foundation of our color
+system and are used to create a consistent look and feel across the product.
+
+### Border
+
+Borders define structure for elements. Generally borders for component elements
+would use `core.border.neutral.subtle`, `core.border.neutral.default` for when
+3:1 contrast is a priority (ex. form elements) and `core.border.neutral.strong`
+for 4.5:1 contrast.
+
+<View style={styles.gridCompact}>
+    <ActionColorGroup
+        category={semanticColor.core.border}
+        group="core.border"
+        includeExample={false}
+    />
+</View>
+
+### Background
+
+Background colors are used to define the background of elements. This includes
+the background of components like buttons, banners, and other elements that
+require a background color.
+
+<View style={styles.gridCompact}>
+    <ActionColorGroup
+        category={semanticColor.core.background}
+        group="core.background"
+        includeExample={false}
+    />
+</View>
+
+### Foreground
+
+Foreground colors are used to define the foreground of elements. This includes
+the text and icons that are displayed on top of the background color. This is
+important for ensuring that the text and icons are legible and accessible.
+
+<Banner
     kind="success"
     layout="full-width"
     text="Make sure to use these colors consistently to ensure readability and accessibility. This includes checking contrast ratios for text and background colors."
@@ -283,21 +332,10 @@ backgrounds in light mode.
     ]}
 />
 
-<ColorGroup colors={semanticColor.text} group="text" />
-
-### Core
-
-#### Border
-
-Borders define structure for elements. Generally borders for component elements
-would use `core.border.neutral.subtle`, `core.border.neutral.default` for when
-3:1 contrast is a priority (ex. form elements) and `core.border.neutral.strong`
-for 4.5:1 contrast.
-
 <View style={styles.gridCompact}>
     <ActionColorGroup
-        category={semanticColor.core.border}
-        group="border"
+        category={semanticColor.core.foreground}
+        group="core.foreground"
         includeExample={false}
     />
 </View>
--- a/__docs__/foundations-color.mdx
+++ b/__docs__/foundations-color.mdx
@@ -267,7 +267,7 @@ backgrounds in light mode.
 <Banner
     kind="warning"
     layout="full-width"
-    text="These tokens are deprecated and will be removed in the future. Please use the semanticColor.core.foreground tokens instead."
+    text="These text tokens are deprecated and will be removed in the future. Please use the semanticColor.core.foreground tokens for text instead."
 />
 
 <ColorGroup colors={semanticColor.text} group="text" />
@@ -278,7 +278,7 @@ Core colors are used for the most common elements in the UI. These colors are
 used for backgrounds, borders, and text. They are the foundation of our color
 system and are used to create a consistent look and feel across the product.
 
-### Border
+#### Border
 
 Borders define structure for elements. Generally borders for component elements
 would use `core.border.neutral.subtle`, `core.border.neutral.default` for when
@@ -293,7 +293,7 @@ for 4.5:1 contrast.
     />
 </View>
 
-### Background
+#### Background
 
 Background colors are used to define the background of elements. This includes
 the background of components like buttons, banners, and other elements that
@@ -307,7 +307,7 @@ require a background color.
     />
 </View>
 
-### Foreground
+#### Foreground
 
 Foreground colors are used to define the foreground of elements. This includes
 the text and icons that are displayed on top of the background color. This is
--- a/packages/wonder-blocks-form/src/components/radio-core.tsx
+++ b/packages/wonder-blocks-form/src/components/radio-core.tsx
@@ -102,19 +102,22 @@ const _generateStyles = (checked: Checked, error: boolean) => {
     // state.
     const colorAction = semanticColor.action.secondary[actionType];
 
+    const coreType = error ? "critical" : "instructive";
+    const colorCore = semanticColor.core.background[coreType];
+
     // The different states that the component can be in.
     const states = {
         // Resting state (unchecked)
         unchecked: {
             border: semanticColor.input.default.border,
-            background: colorAction.default.background,
+            background: colorCore.subtle,
         },
         checked: {
             // NOTE: This is a special case where the border is the same color
             // as the foreground. This should change as soon as we simplify the
             // existing `action` tokens.
             border: colorAction.default.foreground,
-            background: colorAction.default.background,
+            background: colorCore.subtle,
         },
         // Form validation error state
         error: {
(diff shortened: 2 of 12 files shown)


__docs__/foundations-color.mdx
```diff
@@ -264,6 +264,55 @@ the UI.
 For all type to ensure contrast for legibility. Inverse text applies for dark
 backgrounds in light mode.
 
+<Banner
+    kind="warning"
+    layout="full-width"
+    text="These tokens are deprecated and will be removed in the future. Please use the semanticColor.core.foreground tokens instead."
```
marcysutton, line 270: **Suggestion:** in the rendered page, it was a little unclear to me that only the Text tokens are deprecated. Perhaps the warning message could be more descriptive:

```suggestion
    text="These Text tokens are deprecated and will be removed in the future. Please use the semanticColor.core.foreground tokens for text instead."
```

__docs__/foundations-color.mdx
```diff
@@ -264,6 +264,55 @@ the UI.
...
+
+<ColorGroup colors={semanticColor.text} group="text" />
+
+### Core
+
+Core colors are used for the most common elements in the UI. These colors are
+used for backgrounds, borders, and text. They are the foundation of our color
+system and are used to create a consistent look and feel across the product.
+
+### Border
```
marcysutton, line 281: **Suggestion:** I think this might need to be structured underneath Core, as it's one of the Core sections. Is that right?

```suggestion
#### Border
```

Same for Background and Foreground.

marcysutton (commented):
That's a lot of token changes! Looks good to me (and I'm so glad to have Chromatic for this). I had a couple docs suggestions but that's about it!

__docs__/foundations-color.mdx
jandrade: hmmm wait, this was working before (the TOC addon), but turns out that this broke at some point with SB8: https://github.com/storybookjs/storybook/issues/26521

I'll change preview.tsx to include h4s so we can have both things  (semantics + side nav).

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -36,8 +36,76 @@ const core = {
...
+        },
+        critical: {
+            subtle: color.fadedRed8,
+            default: color.red,
+            strong: color.activeRed,
+        },
+        success: {
+            subtle: color.fadedGreen8,
+            default: color.green,
+            strong: color.activeGreen,
```
beaesguerra, line 63: The doc uses the hex directly (#00672C), does the doc need to be updated with `activeGreen`?

I'm treating the doc as the source of truth since it's the shared resource, but let me know if that's not the case!

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -36,8 +36,76 @@ const core = {
...
+        },
+        warning: {
+            subtle: color.fadedGold8,
+            default: color.gold,
+            strong: color.activeGold,
```
beaesguerra, line 68: This is different from the doc which uses `fadedGold8`, `fadedGold24`, and `gold`

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -36,8 +36,76 @@ const core = {
...
+        },
+    },
+
+    /**
+     * Used for text and icons.
+     */
+    foreground: {
+        instructive: {
+            subtle: color.blue,
+            default: color.activeBlue,
```
beaesguerra, line 83: The default foreground colors are different from what's defined in the doc (I see there's an ongoing thread around that too though!)

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -64,11 +132,15 @@ const surface = {
...
+/**
+ * TODO(WB-1941): Remove text once we have migrated to the new core.foreground
+ * tokens.
+ */
 const text = {
-    primary: color.offBlack,
-    secondary: color.fadedOffBlack72,
-    disabled: color.fadedOffBlack32,
-    inverse: color.white,
+    primary: core.foreground.neutral.strong,
```
beaesguerra, line 140: I found the previous semantic text tokens intuitive to use since `text.primary` should be used for most text, and `text.secondary` for subtle text!

Since we will be getting rid of these text tokens, consumers will be using the core tokens directly for text (if I'm understanding correctly!). Based on the semantic naming, I thought that `core.foreground.neutral.default` would be the primary text color that's most used and `core.foreground.neutral.subtle` would be the secondary text color. And `core.foreground.neutral.strong` would be for emphasized text or something like that! Here, I see that `strong` is the default primary text.

This might just be my own assumptions though, but was curious what you (and others) think about this!

__docs__/foundations-color.mdx
beaesguerra, line 1: **praise:** Great docs!

packages/wonder-blocks-form/src/components/radio-core.tsx
```diff
@@ -60,7 +60,7 @@ const disabledChecked = {
     height: size / 2,
     width: size / 2,
     borderRadius: border.radius.radius_full,
-    backgroundColor: semanticColor.action.primary.disabled.background,
+    backgroundColor: semanticColor.core.border.disabled.strong,
```
beaesguerra, line 63: Nice! I'm glad we're adding the core semantic colors. It makes a lot of sense to me to use the core tokens instead of the action tokens for things like forms!

packages/wonder-blocks-switch/src/themes/default.ts
```diff
@@ -5,30 +5,28 @@ import {
...
                 off: semanticColor.icon.inverse,
             },
             icon: {
-                on: action.default.foreground,
-                disabledOn: action.press.background,
-                off: action.default.border,
+                on: semanticColor.icon.action,
+                // NOTE: We use border to match the switch color.
+                disabledOn: semanticColor.core.border.instructive.subtle,
+                off: semanticColor.core.border.neutral.default,
```
beaesguerra, line 29: Should we be using the semantic core background tokens for the `color.bg` theme tokens? I see we use a mix of core `background` and `border` tokens for the backgrounds! 

Or is this more because this is for the Classic theme which doesn't map to the new semantics?

packages/wonder-blocks-switch/src/themes/default.ts
```diff
@@ -5,30 +5,28 @@ import {
...
-                on: action.default.foreground,
-                disabledOn: action.press.background,
-                activeOn: action.press.foreground,
+                on: semanticColor.core.background.instructive.default,
+                disabledOn: semanticColor.core.border.instructive.subtle,
+                activeOn: semanticColor.core.background.instructive.strong,
             },
             slider: {
                 on: semanticColor.icon.inverse,
                 off: semanticColor.icon.inverse,
```
beaesguerra, line 23: Will `semanticColor.icon` also be replaced by core tokens?

packages/wonder-blocks-tabs/src/components/navigation-tab-item.tsx
```diff
@@ -111,10 +111,10 @@ const styles = StyleSheet.create({
         listStyle: "none",
         display: "inline-flex",
         [":has(a:hover)" as any]: {
-            boxShadow: `inset 0 calc(${sizing.size_020}*-1) 0 0 ${semanticColor.action.secondary.progressive.hover.foreground}`,
```
beaesguerra, line 114: This makes sense! I wasn't sure which semantic token to use for the tabs components and the secondary action had the most similar styles!

Will we also be removing the action tokens in favour of the core tokens?

beaesguerra (commented):
Thanks for implementing more of the core tokens! Left some questions and comments! 😄 

Were there expected Chromatic changes? None are showing up for me but it might be because they have been approved already!

packages/wonder-blocks-switch/src/themes/default.ts
jandrade: yeah, that would probably be the case, but something to work on later.

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -36,8 +36,76 @@ const core = {
             strong: color.fadedOffBlack32,
         },
         inverse: {
-            subtle: color.offBlack16,
-            default: color.offBlack8,
+            subtle: color.fadedOffBlack16,
+            default: color.fadedOffBlack8,
```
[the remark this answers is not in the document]
jandrade: hmm yeah good point, I see that it's meant to use the rgba version, I'll revert this change.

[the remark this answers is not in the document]
jandrade: wait, I see that the docs use `HEX` so the token name is incorrect in the doc, I'll change it there and keep this change in code.

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -36,8 +36,76 @@ const core = {
...
+        },
+    },
+    background: {
+        instructive: {
+            subtle: color.fadedBlue8,
+            default: color.blue,
+            strong: color.activeBlue,
+        },
+        neutral: {
+            subtle: color.fadedOffBlack8,
```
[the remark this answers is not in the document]
jandrade: I'm going to keep it as it is right now in code due to some existing limitations in the OG code. We can revisit this later once we get more alignment between TB and OG.

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
jandrade: Ah good point, I'll change it.

beaesguerra (approved):
Looks great! Thanks for making it easier to use semantic color tokens 😄

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
jandrade: That said, I'm going to keep these tokens for now, but we'll see how these can be used in the system and how/if we use them in product, and we can get back to implement any required changes based on usage.

PR #2619, opened 2025-05-26
[WB-1915.1]: Add Chonky (ActivityIconbutton) component

jandrade:
## Summary:

- Tokens: Adds a new `chonky` semantic color.
- IconButton: Adds a new `ActivityIconButton` component that uses the `chonky` semantics.
- Also updates the docs to include this new component in the wb-icon-button package.

NOTE: The next PR will include the `label` prop (see it as `Visible Label On` in figma).

Issue: WB-1915

## Test plan:

- IconButton overview docs: `/?path=/docs/packages-iconbutton-overview--docs`
- ActivityIconButton docs: Navigate to `/?path=/docs/packages-iconbutton-activityiconbutton--docs&globals=theme:thunderblocks`

Verify that the docs look correct and that the `ActivityIconButton` component
renders properly.

Also verify that the visual snapshot tests look as expected in Figma.

https://www.figma.com/design/e9qdyiDUBZ3rDabP9S7ulO/%E2%9A%A1%EF%B8%8F-Handshake?node-id=825-14254&m=dev


marcysutton (commented):
What a delightful component! I left a few questions about labeling, but it looks good otherwise.

__docs__/wonder-blocks-icon-button/activity-icon-button.stories.tsx
```diff
@@ -0,0 +1,185 @@
...
+        );
+    },
+};
+
+/**
+ * By default, the icon buttons do not have accessible names. The `aria-label`
+ * prop must be used to explain the function of the button. Remember to keep the
+ * description concise but understandable.
+ */
+export const WithAriaLabel: StoryComponentType = {
```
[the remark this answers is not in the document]
jandrade: That's correct, it will be the next PR.

__docs__/wonder-blocks-icon-button/activity-icon-button.stories.tsx
```diff
@@ -0,0 +1,185 @@
...
+                href="/"
+                target="_blank"
+                onClick={(e) => action("clicked")(e)}
+            />
+        );
+    },
+};
+
+/**
+ * By default, the icon buttons do not have accessible names. The `aria-label`
```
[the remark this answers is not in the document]
jandrade: I'm going to make this work for both cases in the next PR (good observation!)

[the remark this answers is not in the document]
marcysutton: Awesome, sounds good!

marcysutton (approved):
I saw the tiny border issue that was mentioned in Slack, but that seems non-blocking. You could try small non-zero values for the box-shadow, like `0.001rem` vs `0.0`?

The states otherwise looked enough like the Figma versions that it makes sense to me to keep this moving!

packages/wonder-blocks-icon-button/src/components/activity-icon-button.tsx
```diff
@@ -0,0 +1,261 @@
...
+            padding: {
+                block: sizing.size_140,
+                inline: sizing.size_200,
+            },
+        },
+        shadow: {
+            y: {
+                rest: "6px",
+                hover: "8px",
+                press: sizing.size_0,
```
beaesguerra, line 135: Would it make sense to have the shadow `y` values as tokens as well? (for later!)

And is there a reason why we use `px` instead of `rem`? It would be helpful to include a comment here so we remember that it is intentional!

packages/wonder-blocks-icon-button/src/components/activity-icon-button.tsx
```diff
@@ -0,0 +1,261 @@
...
+            borderRadius: theme.root.border.radius,
+            paddingBlock: theme.root.layout.padding.block,
+            paddingInline: theme.root.layout.padding.inline,
+            // theming
+            background: themeVariant.background[kind].rest,
+            border: `${borderWidthKind.rest} solid ${themeVariant.border[kind].rest}`,
+            // Gives the button a "chonky" look and feel.
+            boxShadow: `0 ${theme.root.shadow.y.rest} 0 0 ${themeVariant.shadow[kind].rest}`,
+            // motion
+            transition: "all 0.12s ease-out",
```
beaesguerra, line 249: (For later) Do you think we'll tokenize motion values later on? (like duration and type of transitions)

packages/wonder-blocks-icon-button/src/components/activity-icon-button.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+            ["@media not (hover: hover)" as any]: {
```
beaesguerra, line 251: I was curious about how it looks in mobile so I tried it out! I found that after clicking one chonky button, when I click another chonky button, the first button de-presses. Is that intended?

https://github.com/user-attachments/assets/93d047a9-d852-47cc-b611-50dcb5cd9641

beaesguerra (approved):
Nice work, the chonky buttons are so fun to press! Left some questions (nothing blocking!)

packages/wonder-blocks-icon-button/src/components/activity-icon-button.tsx
jandrade: ohh I guess it's kinda expected.... it's one of those odd things that happen with `Clickable/ClickableBehavior`. Unfortunately, we cannot get rid of that component yet (due to the `beforeNav/safeNav` dependency.

packages/wonder-blocks-icon-button/src/components/activity-icon-button.tsx
jandrade: I had to use `px` to prevent a Safari issue (not being able to animate properly with `rem`). I'll add a comment.

packages/wonder-blocks-icon-button/src/components/activity-icon-button.tsx
jandrade: I think it will depend with how animations look better with each component. I agree that it might be worth tokenizing later, but we'll see what patterns we can land as soon as we are adding more motion to WB components.

packages/wonder-blocks-icon-button/src/components/activity-icon-button.tsx
jandrade: FYI I'm going to add the comment in the follow up PR (#2632) for easier landing.

PR #2623, opened 2025-05-28
Adjust BodyText and Heading docs and defaults

marcysutton:
## Summary:
While integrating the new WB Typography components into webapp, I found a few things needing updates for a smooth transition.

Issue: WB-1966

## Test plan:
1. Ensure tests pass
2. Test integration into webapp

### DIFF
--- a/packages/wonder-blocks-typography/src/components/body-text.tsx
+++ b/packages/wonder-blocks-typography/src/components/body-text.tsx
@@ -5,16 +5,16 @@ import styles from "../util/styles";
 
 type Props = PropsFor<typeof Text> & {
     size?: "xsmall" | "small" | "medium";
-    weight?: "medium" | "semi" | "bold";
+    weight?: "light" | "medium" | "semi" | "bold";
 };
 
 const BodyText = React.forwardRef(function BodyText(
     {
-        size = "small",
-        weight = "semi",
+        size = "medium",
+        weight = "medium",
         style,
         children,
-        tag = "p",
+        tag = "span",
         ...otherProps
     }: Props,
     ref,
--- a/packages/wonder-blocks-typography/src/components/body-text.tsx
+++ b/packages/wonder-blocks-typography/src/components/body-text.tsx
@@ -5,7 +5,7 @@ import styles from "../util/styles";
 
 type Props = PropsFor<typeof Text> & {
     size?: "xsmall" | "small" | "medium";
-    weight?: "light" | "medium" | "semi" | "bold";
+    weight?: "medium" | "semi" | "bold";
 };
 
 const BodyText = React.forwardRef(function BodyText(
--- a/packages/wonder-blocks-typography/src/components/body-text.tsx
+++ b/packages/wonder-blocks-typography/src/components/body-text.tsx
@@ -14,7 +14,7 @@ const BodyText = React.forwardRef(function BodyText(
         weight = "medium",
         style,
         children,
-        tag = "span",
+        tag = "p",
         ...otherProps
     }: Props,
     ref,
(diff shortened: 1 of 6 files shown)


packages/wonder-blocks-typography/src/components/heading.tsx
```diff
@@ -3,33 +3,42 @@ import {PropsFor, Text} from "@khanacademy/wonder-blocks-core";
 import {font} from "@khanacademy/wonder-blocks-tokens";
 import styles from "../util/styles";
 
+const tagMap = {
+    xxlarge: "h1",
+    xlarge: "h2",
+    large: "h3",
+    medium: "h4",
+    small: "h4",
+} as const;
```
jandrade, line 12: **praise:** This is great!!

packages/wonder-blocks-typography/src/components/heading.tsx
```diff
@@ -3,33 +3,42 @@ import {PropsFor, Text} from "@khanacademy/wonder-blocks-core";
...
-    ref,
-) {
+const Heading = React.forwardRef(function Heading(props: Props, ref) {
+    const {size, weight = "bold", style, children, tag, ...otherProps} = props;
+
+    // Determine final size for style purposes
+    const finalSize = size ?? "large";
+
+    // Resolve tag: prefer explicit `tag`, fallback to size-based tag if size was provided, or "h2" if neither
+    const resolvedTag = tag ?? (size ? tagMap[size] : "h2");
```
jandrade, line 29: **question/thought:** I wonder if there's documentation around this. If not, I think it would be helpful for devs to understand what the default tags are for each size, and also understand how to override this.

packages/wonder-blocks-typography/src/components/body-text.tsx
```diff
@@ -5,16 +5,16 @@ import styles from "../util/styles";
...
 const BodyText = React.forwardRef(function BodyText(
     {
-        size = "small",
-        weight = "semi",
+        size = "medium",
+        weight = "medium",
         style,
         children,
-        tag = "p",
+        tag = "span",
```
jandrade, line 17: **question:** Is there a particular reason why this changed to `span`? if so, would be a good idea to document the why?

jandrade (approved):
This looks great! just left some comments around docs/comments. Thanks for the quick follow up on this 🚢

packages/wonder-blocks-typography/src/components/body-text.tsx
marcysutton: Initially I changed it to `span` to match the expectations of `Body`. But seeing how that tag is used, I think it would be better to create paragraphs by default and document the change.

packages/wonder-blocks-typography/src/components/body-text.tsx
```diff
@@ -5,16 +5,16 @@ import styles from "../util/styles";
 
 type Props = PropsFor<typeof Text> & {
     size?: "xsmall" | "small" | "medium";
-    weight?: "medium" | "semi" | "bold";
+    weight?: "light" | "medium" | "semi" | "bold";
```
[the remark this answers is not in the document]
marcysutton: There is an instance of `light` in Classroom, which is why I added it here! Otherwise devs will have to use `font.weight.light` and miss out on these props.

packages/wonder-blocks-typography/src/components/heading.tsx
marcysutton: Good idea! Docs will be helpful for this. I'll add them now.

packages/wonder-blocks-typography/src/components/body-text.tsx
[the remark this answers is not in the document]
marcysutton: Removed this change for now!

m9ee1:
> ## Summary:
> While integrating the new WB Typography components into webapp, I found a few things needing updates for a smooth transition.
> 
> Issue: WB-1966
> 
> ## Test plan:
> 1. Ensure tests pass
> 2. Test integration into webapp
``````

PR #2625, opened 2025-05-28
[WB-1859.1] Button theming prep work

jandrade:
## Summary:

- Removes the `khanmigo` theme from the `Button` component.
- Switches `Button` theming to use CSS variables.

This is a preparatory change for the upcoming Button theming work, which will
allow to use a new `thunderblocks` theme for the Button component.

Issue: WB-1859

## Test plan:

Navigate to the Button storybook and verify that all stories render correctly.
Ensure that the snapshots look correct and that there are no visual regressions.

### DIFF
--- a/__docs__/wonder-blocks-button/button-testing-snapshots.stories.tsx
+++ b/__docs__/wonder-blocks-button/button-testing-snapshots.stories.tsx
@@ -21,7 +21,6 @@ export default {
         chromatic: {
             modes: {
                 default: allModes.themeDefault,
-                khanmigo: allModes.themeKhanmigo,
             },
         },
     },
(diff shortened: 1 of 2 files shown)


.changeset/metal-stingrays-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-button": patch
```
beaesguerra, line 2: Should this be a `major` version change because we removed the `khanmigo` theme from `Button`? Or is it a patch because we've already migrated off of the khanmigo theme usage from before? Thought I'd check!

beaesguerra (approved):
Left a question around versioning! Other than that, looks good to me! 😄

.changeset/metal-stingrays-enjoy.md
jandrade: The latter, I used `patch` b/c I'm removing the khanmigo button in the webapp PR https://github.com/Khan/webapp/pull/32895. I intended to make that `webapp` change first, that way we don't have to deal with breaking changes :). Hope that makes sense.

PR #2626, opened 2025-05-28
Supporting Badge with Tooltip

beaesguerra:
## Summary:

- Document and add examples for using Badge with Tooltips. 
- `role` needs to be set on `Badge` in order for the tooltip content to be read out by screen readers (since Tooltip sets aria-describedby on the anchor)
- Sincle the Badge would be interactive when used with a Tooltip, we use the global focus styles 

image

Issue: WB-1952

## Test plan:
`?path=/story/packages-badge-badge--badge-with-tooltip`
- The Badge should be focusable when used with a Tooltip
- The tooltip content should be read out by screen readers when the Badge is focused
  - Tested with NVDA+Chrome, NVDA+Firefox, VO+Safari. When the tooltip is open, the tooltip content is read out with the Badge label
  - Note: Created WB-1967 for broader tooltip SR improvements so that it works by default with any component without consumers having to set the role. It's also for addressing general tooltip issues like how VO doesn't always announce tooltips when it is toggled
- The Badge should use the global focus outline style 
- Docs are reviewed (`?path=/docs/packages-badge-badge--docs#badge-with-tooltip`)

`?path=/story/packages-badge-testing-snapshots--all-badges-scenarios`
- It works for all Badge types

### DIFF
--- a/__docs__/wonder-blocks-badge/badge-testing-snapshots.stories.tsx
+++ b/__docs__/wonder-blocks-badge/badge-testing-snapshots.stories.tsx
@@ -1,5 +1,6 @@
 import * as React from "react";
 import type {Meta, StoryObj} from "@storybook/react";
+import {StyleSheet} from "aphrodite";
 import ComponentInfo from "../components/component-info";
 import packageConfig from "../../packages/wonder-blocks-form/package.json";
 import {
@@ -321,13 +322,13 @@ export const AllBadgesScenarios: StoryComponentType = {
                 }}
             >
                 <Heading>Badges</Heading>
-                <View style={{gap: sizing.size_240, flexDirection: "row"}}>
+                <View style={styles.badgesContainer}>
                     {badges.map((badge, index) => (
                         <View key={index}>{badge}</View>
                     ))}
                 </View>
                 <Heading>Badges with Tooltip</Heading>
-                <View style={{gap: sizing.size_240, flexDirection: "row"}}>
+                <View style={styles.badgesContainer}>
                     {badges.map((badge, index) => (
                         <Tooltip
                             content="Tooltip"
@@ -335,7 +336,7 @@ export const AllBadgesScenarios: StoryComponentType = {
                             key={index}
                             placement="bottom"
                         >
-                            {badge}
+                            {React.cloneElement(badge, {role: "button"})}
                         </Tooltip>
                     ))}
                 </View>
@@ -343,3 +344,10 @@ export const AllBadgesScenarios: StoryComponentType = {
         );
     },
 };
+
+const styles = StyleSheet.create({
+    badgesContainer: {
+        gap: sizing.size_240,
+        flexDirection: "row",
+    },
+});
--- a/__docs__/wonder-blocks-badge/badge-testing-snapshots.stories.tsx
+++ b/__docs__/wonder-blocks-badge/badge-testing-snapshots.stories.tsx
@@ -11,7 +11,11 @@ import {
 } from "@khanacademy/wonder-blocks-badge";
 import {IconMappings} from "../wonder-blocks-icon/phosphor-icon.argtypes";
 import {Icon, PhosphorIcon} from "@khanacademy/wonder-blocks-icon";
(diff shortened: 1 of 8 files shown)


__docs__/wonder-blocks-badge/badge-testing-snapshots.stories.tsx
```diff
@@ -41,6 +47,15 @@ export default {
...
 type StoryComponentType = StoryObj<typeof Badge>;
 
+const statusKinds = ["info", "success", "warning", "critical"] as const;
+
+const states = [
+    commonStates.rest,
+    // Include snapshots for focus states to ensure focus styles are applied
+    // correctly. The Badge is not interactive by default, but becomes interactive
+    // when used with another component like `Tooltip`.
+    commonStates.focus,
```
jandrade, line 57: **praise:** love this simple and effective approach!

__docs__/wonder-blocks-badge/badge-testing-snapshots.stories.tsx
```diff
@@ -78,7 +93,7 @@ export const StateSheetStory: StoryComponentType = {
         ];
 
         const statusRows = [
-            ...["info", "success", "warning", "critical"].map((kind) => ({
+            ...statusKinds.map((kind) => ({
                 name: kind,
                 props: {kind},
             })),
```
jandrade, line 99: **suggestion:** Now that I look at this, I wonder if you could just do a direct map call here.

```
const statusRows = statusKinds.map((kind) => ...
```

__docs__/wonder-blocks-badge/badge-testing-snapshots.stories.tsx
```diff
@@ -280,3 +286,68 @@ export const Scenarios: StoryComponentType = {
...
+        return (
+            <View
+                style={{
+                    gap: sizing.size_240,
+                    // Include end padding to ensure tooltips are included in the
+                    // snapshot
+                    paddingBlockEnd: sizing.size_880,
+                }}
+            >
+                <Heading>Badges</Heading>
```
jandrade, line 324: **praise:** yay! it's great to see the new`Heading` component being used here.

packages/wonder-blocks-badge/src/components/__tests__/badge-types.test.tsx
```diff
@@ -79,6 +79,17 @@ describe("Badge types", () => {
...
+                it("should set the role if provided", () => {
+                    // Arrange
+                    render(<Component role="button" label="Badge label" />);
+
+                    // Act
+                    const badge = screen.getByText("Badge label");
+
+                    // Assert
+                    expect(badge).toHaveAttribute("role", "button");
+                });
```
jandrade, line 92: **suggestion/thought:** I wonder if it would be useful adding a test to check that the badge can be focused when `role=button` is set.

packages/wonder-blocks-badge/src/components/badge.tsx
jandrade, line 1: **question:** How is `tabIndex` added to the element?

beaesguerra: Good question! The attribute is added to the element by the `TooltipAnchor` component:  

https://github.com/Khan/wonder-blocks/blob/d9fd9f9bc7f9325b04b44b3bef7518035cd397be/packages/wonder-blocks-tooltip/src/components/tooltip-anchor.tsx#L194

I was trying to infer when this happens in the `Badge` component, but realized it doesn't receive the tabIndex as a prop since it is set on the element directly. I was considering using a `ref` to check the `tabIndex` of the underlying element, but that would prevent us from exposing the ref (since an element can only have one ref. Let me know though if there's a way to forward a ref and use ref values at the same time!). So for now, consumers can explicitly set the `role=button` on the `Badge` when they use it with a `Tooltip`!

jandrade: ahh good observation, that makes sense :)

jandrade (approved):

PR #2629, opened 2025-05-29
[WB-1859.3] Add thunderblocks theme to Button component

jandrade:
## Summary:

The last `Button/TB` PR. This one integrates the `thunderblocks` theme to the
`Button` component.

Issue: WB-1859

## Test plan:

Navigate to the `Button` component in Storybook and select the `Thunderblocks` theme
to see the new styles applied.

Ensure that all button states (default, hover, active, disabled) are visually
consistent with the design specifications.

Figma: https://www.figma.com/design/e9qdyiDUBZ3rDabP9S7ulO/%E2%9A%A1%EF%B8%8F-Handshake?node-id=13-1258&m=dev


packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -385,6 +382,7 @@ export const _generateStyles = (
                   }
                 : {}),
         },
+        pressed: pressStyles,
```
jandrade, line 385: **note:** I noticed that the `press` styles were not being applied with keyboard..... probably messed up this styles at some point during the refactor, so I'm bringing this back.

packages/wonder-blocks-button/src/components/button-icon.tsx
```diff
@@ -17,10 +19,17 @@ export function ButtonIcon({
     style?: StyleType;
     testId?: string;
 }) {
+    // We set the icon size based on the theme object. This is necessary
+    // because the icon size could change based on the theme.
+    const iconStyle = {
+        width: theme.icon.sizing[size],
+        height: theme.icon.sizing[size],
+    };
```
jandrade, line 27: **note:** I did a similar implementation with `IconButton`, and it is meant to support different icon sizes per button size.

packages/wonder-blocks-button/src/theme/default.ts
jandrade, line 1: **note:** I changed the component-level tokens structure to support the thunderblocks styles.

PR #2635, opened 2025-05-30
Adjust layout default for BodyText with zero margins

marcysutton:
## Summary:
There were concerns about `BodyText` changing the default tag from `span` to `p`. The `p` was selected based on usage of the `Body`, and Label components, which most often wrapped lines of text. To ease the transition, this change zeroes out the margins for `BodyText` so it doesn't introduce a layout shift when switching from another component. All of the Typography components already had `display: block` on them.

The `tag` prop is the mechanism to specify a different tag than the default, such as `div`, `span`, `label`, or `li`.

Issue: WB-1966

## Test plan:
1. Review usage of `BodyText`, especially margins
2. Review the docs: /?path=/docs/packages-typography-bodytext-new--docs

### DIFF
--- a/packages/wonder-blocks-typography/src/util/styles.ts
+++ b/packages/wonder-blocks-typography/src/util/styles.ts
@@ -36,6 +36,7 @@ const styles: StyleDeclaration = StyleSheet.create({
     BodyText: {
         ...common,
         fontFamily: font.family.sans,
+        margin: 0,
         // weight and size are determined by props
         // lineHeight is determined by fontSize on REM scale
     },
--- a/.changeset/metal-pianos-compete.md
+++ b/.changeset/metal-pianos-compete.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-typography": minor
+---
+
+Removes default margins from BodyText component, which defaults to an HTML paragraph tag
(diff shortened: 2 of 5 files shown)


packages/wonder-blocks-typography/src/util/styles.ts
```diff
@@ -36,6 +36,7 @@ const styles: StyleDeclaration = StyleSheet.create({
     BodyText: {
         ...common,
         fontFamily: font.family.sans,
+        margin: 0,
```
beaesguerra, line 39: Should we do the same for the `Heading` styles as well?

beaesguerra (approved):
Looks good! Left a question about whether we'll want to remove the margin for Heading too!

packages/wonder-blocks-typography/src/util/styles.ts
marcysutton: Good question! I thought about it but decided to keep the Heading styles the same since the original components didn't zero out margins. @jandrade what do you think?

jandrade: FYI `HeadingX` components (including the new `Heading`) should already set `marginInline: 0` by default: https://github.com/Khan/wonder-blocks/blob/58075352c336b1c3c85ddeccba906058ca579c49/packages/wonder-blocks-core/src/components/text.tsx#L23-L28

PR #2636, opened 2025-05-30
[WB-1975] Add ActivityButton component

jandrade:
## Summary:

This PR introduces the `ActivityButton` component, which is a chonky button
designed to be used in activity cards. It is a large button that can be used to
navigate to different activities or sections within the application.

Issue: https://khanacademy.atlassian.net/browse/WB-1975

## Test plan:

Review the new docs and snapshots to confirm that everything is correct:

- Docs: `/?path=/docs/packages-button-activitybutton--docs&globals=theme:thunderblocks`
- Snapshots: `/?path=/story/packages-button-testing-snapshots-activitybutton--state-sheet-story&globals=theme:thunderblocks`
- Figma: https://www.figma.com/design/e9qdyiDUBZ3rDabP9S7ulO/%E2%9A%A1%EF%B8%8F-Handshake?node-id=3134-18450&t=JgNeWIreUJbeM5ow-4


__docs__/wonder-blocks-button/activity-button.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+type Story = StoryObj<typeof ActivityButton>;
+
+/**
+ * Minimal activity button. The only props specified in this example are `icon`
+ * and `onClick`.
+ */
+export const Default: Story = {
+    args: {
+        // icon: magnifyingGlass,
```
beaesguerra: Should this be commented out since the story description says the `icon` is specified in this example? Or the description should be updated?

__docs__/wonder-blocks-button/activity-button.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+        onClick: (e: React.SyntheticEvent) => {
+            action("clicked")(e);
+        },
+    },
+} as Meta<typeof ActivityButton>;
+
+type Story = StoryObj<typeof ActivityButton>;
+
+/**
+ * Minimal activity button. The only props specified in this example are `icon`
```
marcysutton: Should the mention of `icon` go on the next story since this one is text only?

packages/wonder-blocks-button/src/components/activity-button.tsx
```diff
@@ -0,0 +1,417 @@
...
+        spinner,
+        startIcon,
+        endIcon,
+        waiting: _,
+        ...restProps
+    } = props;
+
+    const buttonStyles = _generateStyles(actionType, disabledProp, kind);
+
+    const disabled = spinner || disabledProp;
```
beaesguerra: Do we want to include the `spinner` in the `StateSheet`? And should the `spinner` prop be included in the prop docs for `ActivityButton`?

image

__docs__/wonder-blocks-button/activity-button.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+        children: "Search",
+        startIcon: magnifyingGlass,
+        actionType: "progressive",
+        disabled: false,
+        kind: "primary",
+    },
+};
+
+/**
+ * In this example, we have `primary`, `secondary`, `tertiary` and `disabled`
```
marcysutton: Would it be helpful to mark the default kind so the prop can be left off? 

```suggestion
 * In this example, we have `primary` (default), `secondary`, `tertiary` and `disabled`
```

packages/wonder-blocks-button/src/components/activity-button.tsx
```diff
@@ -0,0 +1,417 @@
...
+                {/* NOTE: Using a regular className to be able to use descendant selectors to account for the hover and press states */}
+                <View style={chonkyStyles} className="chonky">
+                    {startIcon && (
+                        // We use the `PhosphorIcon` component to render the icon.
+                        // It is a wrapper around the Phosphor icons library.
+                        <PhosphorIcon
+                            size="medium"
+                            color="currentColor"
+                            icon={startIcon}
+                            style={styles.icon}
```
beaesguerra: Do we want to support custom styling for sub-elements within the ActivityButton? For example, if `styles.icon` or `styles.label` need to be overridden?

__docs__/wonder-blocks-button/activity-button.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+                <ActivityButton {...args} kind="secondary" />
+                <ActivityButton {...args} kind="tertiary" />
+                <ActivityButton {...args} disabled={true} />
+            </View>
+        );
+    },
+};
+
+const kinds = ["primary", "secondary", "tertiary"] as const;
+const actionTypes = ["progressive", "neutral"] as const;
```
marcysutton: Perhaps this is broader than this one component, but I've found kinds and actionTypes to be a little confusing. Is there a way to explain the differences or purpose of the two taxomonies? I think inline on relevant stories would be the most helpful at this stage, although a more general explanation would be welcome somewhere too.

Edited to add: I see now, the `Button` docs explain these more fully but `ActivityButton` is missing the same info!

ActivityButton docs in Storybook

packages/wonder-blocks-button/src/components/activity-button.tsx
beaesguerra: The docs for the other props would be helpful too! And if the `kind` Storybook control only allowed the supported kinds instead of a text field (similar to how the `actionType` control only allows `progressive` or `neutral` values)

packages/wonder-blocks-button/src/components/activity-button.tsx
```diff
@@ -0,0 +1,417 @@
...
+                    {endIcon && (
+                        // We use the `PhosphorIcon` component to render the icon.
+                        // It is a wrapper around the Phosphor icons library.
+                        <PhosphorIcon
+                            size="medium"
+                            color="currentColor"
+                            icon={endIcon}
+                            style={styles.icon}
+                            aria-hidden="true"
+                        />
```
beaesguerra: The designs only show a `startIcon` option, should the code component also only support the `startIcon` too?

packages/wonder-blocks-button/src/components/activity-button.tsx
```diff
@@ -0,0 +1,417 @@
...
+
+import {ButtonUnstyled} from "./button-unstyled";
+
+type Props = Omit<ButtonProps, "actionType" | "size"> & {
+    /**
+     * The action type of the button. This determines the visual style of the
+     * button.
+     *
+     * - `progressive` is used for actions that move the user forward in a flow.
+     * - `neutral` is used for buttons that indicate a neutral action.
```
beaesguerra: I don't see the `neutral` action type in the Figma component! Should we support `neutral`, or is there a missing Figma variant?

packages/wonder-blocks-button/src/components/activity-button.tsx
```diff
@@ -0,0 +1,417 @@
...
+                    rest: border.width.thin,
+                    hover: border.width.thin,
+                    press: border.width.thin,
+                },
+                secondary: {
+                    rest: border.width.thin,
+                    hover: border.width.thin,
+                    press: border.width.thin,
+                },
+                tertiary: {
```
beaesguerra, line 211: Do we need to support the `tertiary` type? It is also not in the Figma component!

__docs__/wonder-blocks-button/activity-button.stories.tsx
jandrade: ohh good catch! I missed that one (I had to do a similar change in `ActivityIconButton`). I'll add that info.

packages/wonder-blocks-button/src/components/activity-button.tsx
jandrade: now that I think, I'm going to remove `spinner` prop for now, and we can put it back if needed. AFAIK designs don't account for this case, and we are moving towards a different GraphQL loading model (relying on React Router 6) so this pattern might not be useful anymore.

packages/wonder-blocks-button/src/components/activity-button.tsx
jandrade: hmmm that's a great question.... I feel that we want to have more consistency with these new components, so I'm not sure if we need to allow custom styles right now. cc @caitlynmayers in case you have thoughts.

packages/wonder-blocks-button/src/components/activity-button.tsx
jandrade: Asked the question here: https://www.figma.com/design/e9qdyiDUBZ3rDabP9S7ulO?node-id=3134-18450&m=dev#1283841177

packages/wonder-blocks-button/src/components/activity-button.tsx
jandrade: we want to support it. From @caitlynmayers:

> An aside about the hierarchies — across all buttons, the hierarchy structure is this:
> - Primary: default state has solid fill high contrast color button, no visible border
> - Secondary: default state has visible border with lighter fill, in general lighter contrast
> - Tertiary: default state has no fill in the button (transparent) and no visible border

packages/wonder-blocks-button/src/components/activity-button.tsx
jandrade: Good point! I'll remove it for now :). If we need it later, then it will be easy to adapt.

packages/wonder-blocks-button/src/components/activity-button.tsx
jandrade: Ended up adding the `styles` prop just in case. I followed the pattern created in recent components. Also added a new story (including in Chromatic) to explain how it works: https://5e1bf4b385e3fb0020b7073c-zdwizetqsr.chromatic.com/?path=/docs/packages-button-activitybutton--docs#with-custom-styles

Totally open to feedback! (including the example 😅)

packages/wonder-blocks-button/src/components/activity-button.tsx
jandrade: We are going with both `startIcon` and `endIcon`.

packages/wonder-blocks-button/src/components/activity-button.tsx
jandrade: Removed `neutral`.

packages/wonder-blocks-button/src/components/activity-button.tsx
beaesguerra: Thanks for adding an example too! I struggle with picking what custom styles to use in the examples since we want to show that things can be customized, but we also don't want to showcase something we don't want consumers to do 😅 I will often use color or outlines (like you have!) to customize styles!

One thing I noticed is the custom outline style doesn't show on the start and end icons! It's interesting because the values are set, though it isn't visible in the component 🤔 
image

jandrade: oh right, good call! this is down to how icons are set via `mask-image` (it doesn't allow showing outline/border/background as everything is a mask).

There are two options here:

a) Adding a wrapper element, then apply custom styles here.
b) Keep things as they are, and don't bother about this (IMO icons shouldn't be customized in terms of colors, well... only `color` if something).

Do you have a preference?

__docs__/wonder-blocks-button/button-shared.argtypes.ts
```diff
@@ -0,0 +1,108 @@
+import {ArgTypes} from "@storybook/react";
+import {IconMappings} from "../wonder-blocks-icon/phosphor-icon.argtypes";
+
+export default {
```
beaesguerra, line 4: I noticed the description for the `type` prop renders weird because the description includes a `<form>` tag (This seems to be pre-existing though!)

image

beaesguerra (approved):
Looks good! Thanks for iterating and looking into what features we want to continue to include in the component 😄

packages/wonder-blocks-button/src/components/activity-button.tsx
beaesguerra: That makes sense! I think if we remove the `outline` on the `startIcon`/`endIcon` styles then we should be fine! So we don't think the custom styling isn't working because we can't see the outline. And the alignSelf shows that the custom styles are being used! 

One potential other option for custom styles on the icons is changing the width/height!

jandrade: I'm going to remove the outline. I understand the part for changing the size, but I want to prevent devs from thinking that's ok 😅. I might create a scenario for that case as it is more meant for internal purposes :).

PR #2639, opened 2025-06-02
Badge text truncation

beaesguerra:
## Summary:

- Set max width of badge text to ~30 characters
- Add a way to provide styles to the label in the badge using the `styles.label` prop

Issue: WB-1971

## Test plan:

- Confirm badge label is truncated (`?path=/story/packages-badge-testing-snapshots--scenarios`)
- Review docs around truncation (`?path=/docs/packages-badge-badge--docs#badge-truncation`)


__docs__/wonder-blocks-badge/due-badge.stories.tsx
```diff
@@ -172,48 +172,3 @@ export const CustomIcons: StoryComponentType = {
         },
     },
 };
-
```
beaesguerra, line 175: To reduce duplication, I moved the `CustomStyle` story for `DueBadge` and `StatusBadge` to `badge-testing-snapshots` so we can test all badge types for custom styles: `?path=/story/packages-badge-testing-snapshots--all-badges-scenarios`

image

The base Badge component continues to have docs around how to use the `styles` prop: `?path=/docs/packages-badge-badge--docs#custom-styles`

image


It's also included in each badge component prop docs!

packages/wonder-blocks-badge/src/components/__tests__/badge-types.test.tsx
```diff
@@ -19,49 +19,57 @@ describe("Badge types", () => {
...
 
             // Act
-            const badge = screen.getByText("Badge label");
+            const {container} = render(
+                <Component ref={ref} label="Badge label" />,
+            );
 
             // Assert
-            expect(ref.current).toBe(badge);
+            // eslint-disable-next-line testing-library/no-node-access -- explicitly check the root element
```
beaesguerra: Because we wrap the label text in a `span` now, `screen.getByText` returns that `span` instead of the root element of the badge component. We now use `container.firstChild` to get the root element of the badge components

jandrade: Have you tried using `getByText(container, /badge label/i)` instead?

beaesguerra: Gave that a try and it didn't work!

I checked to see if there is anything else from react testing library that we can use to get the root element of a component and only found guidance for using [container.firstChild](https://testing-library.com/docs/react-testing-library/api/#container-1):
> Tip: To get the root element of your rendered element, use container.firstChild.

Digging it into more, the `no-node-access` lint rule supports an [allowContainerFirstChild](https://github.com/testing-library/eslint-plugin-testing-library/blob/984f245ff1222231a64f57fb91e376e61e7219fd/docs/rules/no-node-access.md?plain=1#L69-L86) option for this exact use case! 

I configured this option so we can get rid of the `eslint-disable-next-line testing-library/no-node-access`, let me know what you think! Open to other suggestions!

__docs__/wonder-blocks-badge/due-badge.stories.tsx
marcysutton: It's so beautiful 😍😂

packages/wonder-blocks-badge/src/components/__tests__/badge-types.test.tsx
marcysutton: One alternative could be putting a `data-testid` on the element you need to target. Then the order of descendants won't matter! (There are definitely mixed opinions on test-ids, I personally think they're fine for many scenarios 🤷‍♀️ )

packages/wonder-blocks-badge/src/components/badge.tsx
```diff
@@ -119,6 +123,13 @@ const styles = StyleSheet.create({
         // it with another component like `Tooltip`
         ...focusStyles.focus,
     },
+    label: {
+        // Truncate the label after ~30 characters
+        maxWidth: "30ch",
```
marcysutton, line 128: What a cool unit!

packages/wonder-blocks-badge/src/components/__tests__/badge-types.test.tsx
jandrade: I feel that it would be better to keep the rule enabled in that case, so we still signal that using this approach is a escape hatch. Silencing the rule would give the wrong signal in tests where we would be allowing using this approach more.

If `getByText` didn't work and testId is not possible (as Marcy suggests), I think it would be better to leave the tests as you created them in the first place.

beaesguerra: The rule continues to be enabled, but it's configured to allow specifically `container.firstChild` since that is the recommended approach for accessing the root element of a component. 

I'll take a look at switching the tests to use the test id!

jandrade (approved):
Looks great! left a comment around testing for your consideration (nothing preventing from landing this PR). 🚀 :shipit:

packages/wonder-blocks-badge/src/components/__tests__/badge-types.test.tsx
beaesguerra: Thanks for the suggestions! I was able to use `getByTestId` instead 😄 For the unit tests specifically testing the `testId` prop, I continue to check the attribute on `container.firstChild` so we can confirm that functionality without using `getByTestId` (since we want to test it is on the right element)!

PR #2632, opened 2025-05-30
[WB-1915.2] ActivityIconbutton: Add `label` prop

jandrade:
## Summary:

- Adds a `label` prop to `ActivityIconButton` to be able to use visible icon descriptions.
- Modifies the types to expect one of `aria-label` or `label`.
- Adds new docs and snapshots to cover this variant.

Screenshot 2025-05-30 at 10 07 07 AM

Issue: WB-1915

## Test plan:

- Verify that the new Label docs look correct: `/?path=/docs/packages-iconbutton-activityiconbutton--docs#with-label`
- Verify that the `label` prop is rendered correctly in the snapshots: `/?path=/story/packages-iconbutton-testing-snapshots-activityiconbutton--state-sheet-visible-label-story&globals=theme:thunderblocks`
- Verify that the Scenarios story renders the `label` prop correctly: `/?path=/story/packages-iconbutton-testing-snapshots-activityiconbutton--scenarios&globals=;theme:thunderblocks`

PR #2640, opened 2025-06-03
Update badge tsconfig-build and add type tests

beaesguerra:
## Summary:
- Update root tsconfig-build.json so types are exported from the new package
- Also update tsconfig-build so there are no errors in the badge package
- Add typestest for badges

Issue: WB-XXXX

## Test plan:

- Confirm that types are generated in the badge package snapshot

### DIFF
--- a/.changeset/tame-bees-kick.md
+++ b/.changeset/tame-bees-kick.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-badge": minor
+---
+
+Update tsconfig-build for types


packages/wonder-blocks-badge/tsconfig-build.json
```diff
@@ -12,5 +12,6 @@
       {"path": "../wonder-blocks-styles/tsconfig-build.json"},
       {"path": "../wonder-blocks-icon/tsconfig-build.json"},
       {"path": "../wonder-blocks-styles/tsconfig-build.json"},
+      {"path": "../wonder-blocks-tooltip/tsconfig-build.json"},
```
beaesguerra, line 15: This is because we have tests for testing badges with Tooltip! Was getting errors related to the tooltip package without this

tsconfig-build.json
```diff
@@ -8,6 +8,7 @@
         {"path": "./packages/wonder-blocks-accordion/tsconfig-build.json"},
         {"path": "./packages/wonder-blocks-announcer/tsconfig-build.json"},
         {"path": "./packages/wonder-blocks-banner/tsconfig-build.json"},
+        {"path": "./packages/wonder-blocks-badge/tsconfig-build.json"},
```
beaesguerra, line 11: This makes it so types are included in the badge package! (Discovered this because auto-import wasn't working in webapp with the badge package snapshot)

packages/wonder-blocks-badge/src/components/__tests__/badge.typestest.tsx
```diff
@@ -0,0 +1,80 @@
+import * as React from "react";
```
beaesguerra, line 1: Added a typestest for testing different prop combinations! (Thanks Juan for the example in https://github.com/Khan/wonder-blocks/pull/2632#discussion_r2116848070!)

PR #2641, opened 2025-06-03
Upgrade to Storybook v9

jeremywiebe:
## Summary:

This PR upgrades Wonder-Blocks to Storybook v9. I ran `pnpm dlx @storybook/latest upgrade` and allowed it to run all of the migrations it thought it needed to. 

After that when I ran storybook, I noticed a CORS warning with Google Fonts so I fixed that also.  

Issue: "none"

## Test plan:

`rm -rf node_modules`
`pnpm install`
`pnpm start`
`pnpm test:storybook` - all tests pass

I checked around and everything seems to work


.storybook/manager-head.html
```diff
@@ -2,16 +2,13 @@
 <link rel="preload" href="./jakarta.css" as="style" />
 
 <link rel="stylesheet" href="./lato.css">
-<link
-    rel="preload"
-    href="https://fonts.googleapis.com/css?family=Inconsolata"
-/>
-<link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Inconsolata">
+<link rel="preload"    crossorigin="anonymous" href="https://fonts.googleapis.com/css?family=Inconsolata" as="style" />
+<link rel="stylesheet" crossorigin="anonymous" href="https://fonts.googleapis.com/css?family=Inconsolata">
```
jeremywiebe, line 6: This fixes CORS warnings I was seeing (at least in Safari)

.storybook/preview.tsx
```diff
@@ -73,29 +73,32 @@ function DocsContainerWithTheme({children, context, ...props}) {
...
-            {
+            offWhite: {
                 name: "offWhite",
                 value: semanticColor.surface.secondary,
             },
-        ],
+        },
+    },
+    initialGlobals: {
+        // 👇 Set the initial background color
```
jeremywiebe, line 98: This is how we do backgrounds in Storybook v9. 

https://storybook.js.org/docs/essentials/backgrounds

doctor-storybook.log
jandrade, line 1: **note:** This probably needs to be included to `.gitignore`.

packages/wonder-blocks-testing-core/src/fixtures/fixtures.tsx
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable import/no-deprecated */
 import * as React from "react";
-import {action} from "@storybook/addon-actions";
+import {action} from "storybook/actions";
```
jandrade: **thought:** I wonder if this change could affect and/or break `webapp` fixtures in any way (I think we are still using SB v8 there). cc @somewhatabstract in case you have thoughts.

packages/wonder-blocks-popover/src/components/__tests__/popover.test.tsx
```diff
@@ -5,7 +5,7 @@ import {userEvent, PointerEventsCheckLevel} from "@testing-library/user-event";
 import {View} from "@khanacademy/wonder-blocks-core";
 import Button from "@khanacademy/wonder-blocks-button";
 
-import {fireEvent} from "@storybook/test";
+import {fireEvent} from "storybook/test";
```
jandrade, line 8: I'm not sure why this is here 😅... maybe it was meant to be RTL's fireEvent? (unrelated to your PR)

.changeset/red-ducks-try.md
```diff
@@ -0,0 +1,7 @@
+---
+"@khanacademy/wonder-blocks-testing-core": patch
+"@khanacademy/wonder-blocks-popover": patch
```
jandrade: **nit:** `Popover` doesn't require a bump as tests were the only thing affected inside the package.

packages/wonder-blocks-testing/package.json
```diff
@@ -25,8 +25,7 @@
     "react": "catalog:",
     "react-dom": "catalog:",
     "react-router-dom": "catalog:",
-    "react-router-dom-v5-compat": "catalog:",
-    "storybook": "catalog:"
```
jandrade: **question:** Would this be expected?  I see that we still import `actions` in the fixtures code.

jandrade (commented):
This mostly looks great! I'd like to understand better if the `wb-testing` changes will keep working in `webapp`. Great work and thanks for setting up this @jeremywiebe 🫶

.changeset/red-ducks-try.md
jeremywiebe: Oh, you're right! I'll fix this.

packages/wonder-blocks-testing/package.json
jeremywiebe: `actions` are used in [`@khanacademy/wonder-blocks-testing-core`](https://github.com/Khan/wonder-blocks/blob/e2fb1a4fe96117015c1be4a1a06b4d8278bbeaab/packages/wonder-blocks-testing-core/src/fixtures/fixtures.tsx#L3) and that package still depends on `storybook`.

packages/wonder-blocks-testing-core/src/fixtures/fixtures.tsx
jeremywiebe: You're right.  I created a PR for ~~webapp (https://github.com/Khan/webapp/pull/33321)~~ frontend (https://github.com/Khan/frontend/pull/140) that adapts things. So once we land this PR we can do a WB release and bump that webapp PR immediately.

packages/wonder-blocks-popover/src/components/__tests__/popover.test.tsx
jeremywiebe: Yeah, this was wrong. I changed it to be `fireEvent` from RTL and things work!

.storybook/preview.tsx
```diff
@@ -74,25 +74,29 @@ function DocsContainerWithTheme({children, context, ...props}) {
...
         default: "baseDefault",
-        values: [
-            {
+        options: {
+            baseDefault: {
                 name: "baseDefault",
                 value: semanticColor.core.background.base.default,
             },
-            {
+            light: {
```
jandrade: **suggestion:** This probably needs to get renamed to `neutralStrong` (or the related stories to `light`) so the current snapshots are not affected by this change.

__docs__/wonder-blocks-accordion/accordion.stories.tsx
```diff
@@ -655,9 +655,9 @@ export const BackgroundColorExample: StoryComponentType = {
     },
 };
 
-BackgroundColorExample.parameters = {
+BackgroundColorExample.globals = {
     backgrounds: {
-        default: "neutralStrong",
+        value: "neutralStrong",
```
jandrade, line 660: **suggestion:** See my previous comment. This could be `light` or kept as it is if the global value gets renamed to `neutralStrong`. There are another two examples that fall into this issue.

pnpm-workspace.yaml
```diff
@@ -31,3 +31,5 @@ catalog:
     react-window: ^1.8.11
     # Codemods
     jscodeshift: ^17.3.0
+    # Storybook
+    storybook: ^9.1.8
```
jandrade, line 35: **praise:** Good call!

jandrade (approved):
Looks great overall! just left a comment to fix a few story snapshots. Thank you so much for getting us upgraded to the latest version 👏 🚀

.storybook/preview.tsx
jeremywiebe: I'll rename them to `neutralStrong` so that we don't pull forward a deprecated term.

__docs__/wonder-blocks-accordion/accordion.stories.tsx
jeremywiebe: I did rename it to `neutralStrong`.  I think it's working. :)

PR #2644, opened 2025-06-04
[WB-1958.2] Add theming support to Cell

jandrade:
## Summary:

- Refactors Cell structure by removing Strut in favor of CSS's gap property
- Moves the Cell tokens to a separate `theme` file and converts these component-level tokens to CSS variables (this is required here as the Cell styles differ in TB).
- Change old typography components with new ones (`Heading`, `BodyText`).

Issue: https://khanacademy.atlassian.net/browse/WB-1958

## Test plan:

Navigate to the Cell docs and verify that there are no noticeable visual
changes. The Cell should still render correctly with the same spacing and
alignment as before.

- `/?path=/story/packages-cell-testing-snapshots-compactcell--state-sheet-story`
- `/?path=/story/packages-cell-testing-snapshots-detailcell--state-sheet-story`

### DIFF
--- a/packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
+++ b/packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
@@ -333,20 +333,21 @@ const styles = StyleSheet.create({
         // have overflow: hidden on the inner wrapper instead of the clickable element
         // because setting it on the clickable element causes issues with existing
         // cases.
-        [":active:not([aria-current=true]) .inner-wrapper" as any]: {
-            position: "relative",
-            ":before": {
-                content: "''",
-                position: "absolute",
-                top: 0,
-                left: 0,
-                bottom: 0,
-                width: theme.root.border.width.default,
-                // We use the border token as this element acts like a border
-                // when the cell is pressed.
-                backgroundColor: theme.root.color.press.border,
+        [":active[aria-disabled=false]:not([aria-current=true]) .inner-wrapper" as any]:
+            {
+                position: "relative",
+                ":before": {
+                    content: "''",
+                    position: "absolute",
+                    top: 0,
+                    left: 0,
+                    bottom: 0,
+                    width: theme.root.border.width.default,
+                    // We use the border token as this element acts like a border
+                    // when the cell is pressed.
+                    backgroundColor: theme.root.color.press.border,
+                },
             },
-        },
     },
 
     active: {
--- a/packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
+++ b/packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
@@ -5,7 +5,6 @@ import type {StyleType} from "@khanacademy/wonder-blocks-core";
 
 import Clickable from "@khanacademy/wonder-blocks-clickable";
 import {View} from "@khanacademy/wonder-blocks-core";
-import {Strut} from "@khanacademy/wonder-blocks-layout";
 import {
     border,
     color,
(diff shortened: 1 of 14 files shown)


packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -229,61 +218,22 @@ const CellCore = (props: CellCoreProps): React.ReactElement => {
...
-        },
-    },
-};
-
 const styles = StyleSheet.create({
     wrapper: {
-        background: cellTokens.root.default.background,
-        color: cellTokens.root.default.foreground,
+        background: semanticColor.surface.primary,
+        color: semanticColor.core.foreground.neutral.strong,
```
jandrade, line 224: **note:** Most of the color tokens where refactored to use `semanticColor` directly, that way we don't have to recreate some tokens as CSS variables.

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -345,16 +294,15 @@ const styles = StyleSheet.create({
         /**
          * States
          */
-        // disabled
-        // NOTE: We use `aria-disabled` instead of `disabled` because we want
-        // to allow the cell to be focusable even when it's disabled.
-        [":hover[aria-disabled=true]" as any]: {
-            cursor: "not-allowed",
+        ":hover": {
```
jandrade, line 297: **note:** Took the opportunity to simplify the selectors here by removing the `[aria-disabled]` combinator and moving the required styles to the `.disabled` key below. (I've been doing this with other components to simplify styling + specificity).

packages/wonder-blocks-cell/src/components/internal/common.ts
```diff
@@ -1,34 +1,8 @@
...
-    detailCellPadding: {
-        paddingVertical: spacing.medium_16,
-        paddingHorizontal: spacing.medium_16,
-    },
-
-    /**
-     * The horizontal spacing between the left and right accessory.
-     */
-    accessoryHorizontalSpacing: spacing.medium_16,
-} as const;
```
jandrade, line 31: **note:** Moved all of this to tokens in `theme/default.ts`.

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -310,8 +259,8 @@ const styles = StyleSheet.create({
             top: 0,
             left: 0,
```
beaesguerra, line 260: I realize now that because we took the `:before` styling approach for the left bar indicator, it doesn't support RTL correctly (this is pre-existing!). I created https://khanacademy.atlassian.net/browse/WB-1981 for this!

packages/wonder-blocks-cell/src/util/types.ts
```diff
@@ -31,6 +31,10 @@ export type AccessoryStyle = {
      * To horizontally align the accessory.
      */
     alignItems?: "flex-start" | "flex-end" | "center";
+    /**
+     * To set spacing between child elements.
+     */
+    gap?: number | string;
```
beaesguerra, line 37: This is a neat way to limit the custom styles that are allowed in WB components!

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
jandrade: Great catch, and thanks for creating the ticket!

packages/wonder-blocks-cell/src/util/types.ts
jandrade: 100% I totally forgot about that part, and took me a bit to figure out why some style overrides weren't working 😅

PR #2645, opened 2025-06-04
Set up snapshots for Banner

beaesguerra:
## Summary:

Prep work for theming the Banner component! 
- Set up statesheet and scenarios for banner
- Disable snapshots for most of the existing banner stories that are covered by the testing snapshots

Issue: WB-1865

## Test plan:

Review snapshots and confirm they cover the different combinations of Banner configuration


__docs__/components/text-for-testing.ts
```diff
@@ -3,3 +3,5 @@ export const longText =
     "Lorem ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor incididunt ut labore et dolore magna aliqua";
 export const longTextWithNoWordBreak =
     "Loremipsumdolorsitametconsecteturadipiscingelitseddoeiusmodtemporincididuntutlaboreetdoloremagnaaliqua";
+export const reallyLongText =
```
jandrade, line 6: **note:** 😄 love the descriptive name here!

__docs__/wonder-blocks-banner/banner-testing-snapshots.stories.tsx
```diff
@@ -0,0 +1,272 @@
...
+                                icon={IconMappings.cookie}
+                            />
+                            <Banner
+                                {...props}
+                                text={`Floating. ${longText}`}
+                                layout="floating"
+                            />
+                        </View>
+                    )}
+                </StateSheet>
```
jandrade: **suggestion:** It took me a moment to understand the differences between this `StateSheet` and the one below. Do you think these could be split into different stories?

I also notice that the first one only uses `Rest` (which makes sense), so I wonder if it would be better just to use `AllVariants` for this instead? wdyt?

__docs__/wonder-blocks-banner/banner.stories.tsx
```diff
@@ -212,6 +216,12 @@ DarkBackground.parameters = {
     backgrounds: {
         default: "darkBlue",
     },
+    parameters: {
+        chromatic: {
+            // Testing snapshots don't cover dark background case
+            disableSnapshot: false,
+        },
+    },
```
jandrade: **question:** Do you think that this story is helpful at this point? I'm not sure if we are adding Banners on dark backgrounds.

__docs__/wonder-blocks-banner/banner-testing-snapshots.stories.tsx
beaesguerra: Great suggestions! I updated it to use `AllVariants` for the examples that only need the rest state. And moved it out of the StateSheet into an AllVariants story!

__docs__/wonder-blocks-banner/banner.stories.tsx
beaesguerra: Great question! I kept it around for backwards compatibility for the Classic theme, esp after seeing that the banner styling is done in a specific way to support dark backgrounds (I plan to simplify the styling in the next PR so we don't need to use opacity!) 

https://github.com/Khan/wonder-blocks/blob/db3174f41d7461ba7761e0cf3f15f9da095d4fdf/packages/wonder-blocks-banner/src/components/banner.tsx#L317-L323

Once I integrate the new Banner changes in webapp, I'll confirm if we're using it on any dark backgrounds and see if we can remove this story!

jandrade: sounds great, Thanks!

PR #2647, opened 2025-06-04
[🔥AUDIT🔥] fix image path in docs

beaesguerra:
🖍 _This is an audit!_ 🖍

## Summary:

Some images in the Icon stories in the deployed instance of Storybook are not rendering as expected. 

image

It renders as expected when the image path doesn't start with `/`

image

Issue: WB-XXXX

## Test plan:
The icons should all render an image (`?path=/docs/packages-icon-icon--docs`)


beaesguerra:
Going to re-create this! Made an additional commit to address other places with the same issue

PR #2649, opened 2025-06-05
[🔥AUDIT🔥] [fix] ActivityButton: Fix focus + press style

jandrade:
🖍 _This is an audit!_ 🖍

## Summary:

This bug was introduced after adding support to programmatic focus, which
caused the focus outline to appear when the button was pressed.

This change ensures that the focus outline is only shown when the button
is focused.

Issue: "none"

## Test plan:

Navigate to `/?path=/docs/packages-button-activitybutton--docs`

Click/press on the button and verify that the focus outline does not appear, but
only appears when it is focused.

BEFORE

https://github.com/user-attachments/assets/30d93700-b0e8-43af-bfac-a0bd3cc031ce

AFTER

https://github.com/user-attachments/assets/89b90f87-52b9-434e-b918-018fbcb71393

### DIFF
--- a/.changeset/angry-waves-fail.md
+++ b/.changeset/angry-waves-fail.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-button": patch
+---
+
+Fixes an issue where the focus outline was included when the button was pressed"


somewhatabstract:
:thumbsup:

PR #2652, opened 2025-06-05
Fix other image links in docs

beaesguerra:
## Summary:

Went through where we use `` elements in stories and updated the ones that fail to render in the deployed Storybook docs: https://khan.github.io/wonder-blocks/?path=/docs/overview--docs

Note: These were rendering as expected in Storybook locally and Chromatic.

Issue: WB-1977

## Test plan:

Images should render as expected locally and in the Chromatic SB instance. Verify after deployment that the main docsite renders it properly too


marcysutton (approved):
Thanks for fixing these!

PR #2653, opened 2025-06-05
Set up banner tokens structure and update styles

beaesguerra:
## Summary:
- Set up banner tokens structure and define layout, sizing, color values (will set up the thunderblocks theming in the next PR)
- Updated styles
  - replace `LabelSmall` with `BodyText`
  - use `sizing` tokens instead of `spacing`
  - refactor styling for the background color so we can use semantic color tokens instead of also relying on opacity

Issue: WB-1865

## Test plan:
- Review Banner stories and snapshots
  - The component should look the same in Classic theme. 
  - Layout is now using rem values instead of px (since we use `sizing` tokens now)

### DIFF
--- a/packages/wonder-blocks-banner/src/components/banner.tsx
+++ b/packages/wonder-blocks-banner/src/components/banner.tsx
@@ -296,9 +296,7 @@ const Banner = (props: Props): React.ReactElement => {
                 />
                 <View style={styles.labelAndButtonsContainer}>
                     <View style={styles.labelContainer}>
-                        <BodyText size="small" tag="span">
-                            {text}
-                        </BodyText>
+                        <BodyText size="small">{text}</BodyText>
                     </View>
                     {actions && (
                         <View style={styles.actionsContainer}>
--- a/packages/wonder-blocks-banner/src/components/banner.tsx
+++ b/packages/wonder-blocks-banner/src/components/banner.tsx
@@ -170,6 +170,19 @@ const getBannerKindStyle = (kind: BannerKind) => {
     }
 };
 
+const getBannerIconKindStyle = (kind: BannerKind) => {
+    switch (kind) {
+        case "success":
+            return styles.successIcon;
+        case "info":
+            return styles.infoIcon;
+        case "warning":
+            return styles.warningIcon;
+        case "critical":
+            return styles.criticalIcon;
+    }
+};
+
 /**
  * Banner. A banner displays a prominent message and related optional actions.
  * It can be used as a way of informing the user of important changes.
@@ -257,6 +270,7 @@ const Banner = (props: Props): React.ReactElement => {
     const valuesForKind = getValuesForKind(kind);
 
     const bannerKindStyle = getBannerKindStyle(kind);
+    const bannerIconKindStyle = getBannerIconKindStyle(kind);
 
     return (
         <View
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-banner/src/components/banner.tsx
```diff
@@ -302,40 +323,102 @@ const Banner = (props: Props): React.ReactElement => {
...
+        borderInlineStartWidth: bannerTokens.root.border.width.inlineStart,
+        borderInlineEndWidth: bannerTokens.root.border.width.inlineEnd,
+        borderBlockStartWidth: bannerTokens.root.border.width.blockStart,
+        borderBlockEndWidth: bannerTokens.root.border.width.blockEnd,
         width: "100%",
-        // Because of the background color's opacity value,
-        // the base color needs to be hard-coded as white for the
-        // intended pastel background color to show up correctly
-        // on dark backgrounds.
-        // TODO(WB-1865): Verify if we can change this to use semanticColor
```
beaesguerra, line 321: We are able to use semantic colors for the background, without having to set the opacity!

packages/wonder-blocks-banner/src/components/banner.tsx
```diff
@@ -302,40 +321,102 @@ const Banner = (props: Props): React.ReactElement => {
...
-        opacity: 0.08,
+const bannerTokens = {
+    root: {
+        border: {
+            radius: {
+                default: border.radius.radius_0,
+                floating: border.radius.radius_040,
+            },
+            width: {
+                inlineStart: sizing.size_060, // uses rem so the border indicator scales to font size
```
jandrade, line 332: **praise:** Neat! thanks for adding the comment.

packages/wonder-blocks-banner/src/components/banner.tsx
```diff
@@ -302,40 +321,102 @@ const Banner = (props: Props): React.ReactElement => {
...
+    },
+};
+
+const styles = StyleSheet.create({
     containerOuter: {
-        borderInlineStartWidth: spacing.xxSmall_6,
+        borderInlineStartWidth: bannerTokens.root.border.width.inlineStart,
+        borderInlineEndWidth: bannerTokens.root.border.width.inlineEnd,
+        borderBlockStartWidth: bannerTokens.root.border.width.blockStart,
+        borderBlockEndWidth: bannerTokens.root.border.width.blockEnd,
```
jandrade, line 407: **question:** Is this change done to allow for more flexible borders in the upcoming TB theme?

packages/wonder-blocks-banner/src/components/banner.tsx
```diff
@@ -372,19 +451,49 @@ const styles = StyleSheet.create({
...
+    infoIcon: {
+        color: bannerTokens.icon.color.info,
+    },
+    warningIcon: {
+        color: bannerTokens.icon.color.warning,
+    },
+    criticalIcon: {
+        color: bannerTokens.icon.color.critical,
+    },
 });
```
jandrade, line 497: **praise:** Yes! this is better than computing the styles dynamically. Thanks for refactoring this part to make it future proof :)

jandrade (approved):
Looks great!! looking forward to seeing how this integrates with the TB theme 🚢

packages/wonder-blocks-banner/src/components/banner.tsx
beaesguerra: Yes exactly! The TB banner has borders on all the sides, where as Classic only has a border on the left!
image

PR #2655, opened 2025-06-06
Switch: Add support for Thunderblocks theming

marcysutton:
## Summary:
[WB-1973-switch] Remove redundant variables
[WB-1973-switch] Add css variables script to package.json
[WB-1973-switch] Update theming away from React Context

Issue: WB-1973

## Test plan:

1. Review stories for Switch: /?path=/docs/packages-switch--docs&globals=theme:thunderblocks
2. Ensure TB changes are reflected with theme switcher

### Thunderblocks:
State sheet for Switch and Thunderblocks theme

### OG:
State sheet for Switch and Classic theme


packages/wonder-blocks-switch/src/theme/default.ts
```diff
@@ -0,0 +1,70 @@
...
+                },
+                icon: {
+                    on: semanticColor.icon.action,
+                    disabledOn: semanticColor.core.border.instructive.subtle,
+                    off: semanticColor.core.border.neutral.default,
+                    disabledOff: semanticColor.icon.disabled,
+                },
+            },
+        },
+    },
```
jandrade: **suggestion:** It looks like TB doesn't override any colors, so I'd recommend to remove these tokens from this file and move it directly to `switch.tsx` instead.

That way we don't create more CSS variables that can directly map to existing `semanticColor` css vars :)

If you want, you can keep this structure within the `switch.tsx` file.

e.g.
```
const color = {
  bg: { ...
}

function Switch() {
  ...
}

const Styles = StyleSheet.create({
  someKey: {
    background: color.bg.slider.on
```

packages/wonder-blocks-switch/src/theme/default.ts
```diff
@@ -0,0 +1,70 @@
+import {
+    border,
+    color,
+    semanticColor,
+    spacing,
+} from "@khanacademy/wonder-blocks-tokens";
+
+export default {
+    switch: {
```
jandrade: **suggestion:** This `switch` key could probably removed and instead use the second level as the first one. Now I remember that I have to do the same cleanup in `iconButton` 😅.

For context, I was originally going to add themes to multiple components in the `icon-button` package, but turned out that `IconButton` was the only one that needed component level tokens. In `Switch` is the same case, as this is the only component in the package.

packages/wonder-blocks-switch/src/components/switch.tsx
```diff
@@ -3,17 +3,9 @@ import {CSSProperties, StyleSheet} from "aphrodite";
...
-    useScopedTheme,
-    useStyles,
-} from "@khanacademy/wonder-blocks-theming";
 import {useId} from "react";
 import {focusStyles} from "@khanacademy/wonder-blocks-styles";
-import ThemedSwitch, {
-    SwitchThemeContext,
-    SwitchThemeContract,
-} from "../themes/themed-switch";
+import {theme as switchTheme} from "../theme/index";
```
jandrade: **suggestion:** This could just change to:

```suggestion
import {theme} from "../theme/index";
```

The you could remove `const theme ...` in line 46.

packages/wonder-blocks-switch/src/theme/default.ts
```diff
@@ -0,0 +1,70 @@
...
+        root: {
+            border: {
+                radius: {
+                    default: border.radius.radius_120,
+                    hover: border.radius.radius_120,
+                    press: border.radius.radius_120,
+                },
+            },
+            sizing: {
+                height: spacing.large_24,
```
jandrade: **suggestion:** It would be great changing this to use `sizing` tokens in places where it makes sense :) (including the harcoded values below).

packages/wonder-blocks-switch/src/theme/default.ts
```diff
@@ -0,0 +1,70 @@
...
+                transition: "transform 0.15s ease-in-out",
+            },
+        },
+        icon: {
+            position: {
+                top: spacing.xxxSmall_4,
+                left: spacing.xxxSmall_4,
+            },
+            transform: {
+                default: `translateX(${spacing.medium_16}px)`,
```
jandrade: **suggestion:** This would require a sligther change if you do the `sizing` refactor (if it makes sense ofc).

```suggestion
                default: `translateX(${sizing.size_160})`,
```

jandrade (commented):
The changes look good overall! Left some comments for you to consider before giving final approval (some are related to things I also have to change in IconButton).

packages/wonder-blocks-switch/src/theme/default.ts
marcysutton: @jandrade I'd like to import the styles from another file to make this one less gigantic. Do you have a preferred naming convention for that?

marcysutton: One other question -- should I leave the script stuff in `package.json` if we aren't doing component-level theming?

packages/wonder-blocks-switch/src/theme/index.ts
```diff
@@ -0,0 +1,8 @@
+import {mapValuesToCssVars} from "@khanacademy/wonder-blocks-tokens";
+import themeDefault from "./default";
+
+// Export the raw theme for direct use in components
+export const theme = themeDefault;
+
+// Export the CSS variables for use in CSS
+export default mapValuesToCssVars(themeDefault, "--wb-c-switch-");
```
marcysutton, line 5: Do we need this file?

packages/wonder-blocks-switch/src/theme/thunderblocks.ts
```diff
@@ -0,0 +1,16 @@
+import {mergeTheme} from "@khanacademy/wonder-blocks-theming";
+import {border} from "@khanacademy/wonder-blocks-tokens";
+import defaultTheme from "./default";
+
+export default mergeTheme(defaultTheme, {
```
marcysutton, line 5: Probably don't need this one either!

packages/wonder-blocks-switch/src/theme/default.ts
jandrade: to clarify, component-level theming is still useful for some tokens, for example `root.sizing`, `root.border.radius` tokens might still need component-level tokens as these could change between OG and TB.

packages/wonder-blocks-switch/src/theme/thunderblocks.ts
jandrade: I think we still need this file as `border.radius` changes between themes.

packages/wonder-blocks-switch/src/theme/index.ts
jandrade: See my other comment re `root.border.radius`. We would still need this.

packages/wonder-blocks-switch/src/theme/default.ts
marcysutton: okay, I'll put that stuff back in! Thanks Juan!

marcysutton: I did put those colors in a `base-styles.ts` file so the switch component file itself wouldn't be so large. That name convention allowed it to take other styles than colors, if necessary. Let me know what you think!

packages/wonder-blocks-switch/src/theme/thunderblocks.ts
```diff
@@ -0,0 +1,17 @@
...
+    root: {
+        border: {
+            radius: {
+                default: border.radius.radius_080,
+                hover: border.radius.radius_080,
+            },
+        },
+        sizing: {
+            width: sizing.size_440,
```
marcysutton, line 13: I'm not seeing the width of the switch overridden in Storybook. I'll keep looking for it.

packages/wonder-blocks-switch/src/theme/index.ts
```diff
@@ -0,0 +1,8 @@
+import {mapValuesToCssVars} from "@khanacademy/wonder-blocks-tokens";
+import themeDefault from "./default";
+
+// Export the raw theme for direct use in components
+export const theme = themeDefault;
```
jandrade: **suggestion:** This will not work with theming, as this will use the static theme. The one below (default export) works as that function is the one that is in charge of transforming `themeDefault` to point to CSS variables instead of raw values.

packages/wonder-blocks-switch/src/theme/default.ts
beaesguerra: I have something similar for [Banner theming](https://github.com/Khan/wonder-blocks/pull/2656/files#diff-d8a6fed396abe006106fb096c54d6db4802989f655d48c4379fee25a270e7014R322) since only some properties needed to be themable! I ended up keeping a tokens structure in the component file, and it would either use global tokens, or tokens from the imported `theme`! 

And I noticed in @jandrade's [Cell theming PR](https://github.com/Khan/wonder-blocks/pull/2644/files#diff-0b4c2beea72229dd5e2a29d5053ba87703ec2c51882604d29f36995815a299d5L232), we remove the tokens structure from the component file!

Do we want to consolidate on how we're approaching the static component tokens?

jandrade: In the Cell case, I moved that structure to component-level tokens as I found out that most of those tokens would change between OG and TB).

You can see the TB component-level tokens here for example: https://github.com/Khan/wonder-blocks/pull/2657/files#diff-5821627caf6a61e59e52cd25b3730da9e8bfb6c7157f5e117a158ca40152ee5c

jandrade: As for how to approach static components tokens, my personal preference is to favor colocation over separation. My reasons for that is it helps me understand the relation between the style and the token (and its value).... now that I think, that's one factor why I ended up removing the `cellTokens` object in the PR 😄.

As I mentioned, that's a personal preference, but I don't feel strongly about that, so I'm fine with either option we choose.

How do you both feel about each option?

marcysutton: I ended up moving the base styles into `switch.tsx` because there weren't very many. But I generally do want to avoid component files getting too long simply for the sake of colocation. The Switch component file is only 285 lines so it's not too bad. But anything over 300 starts to get too complex to hold in one's head at one time! For this case, I think it's fine as-is though.

beaesguerra: I'm good with colocating the styles for component files that aren't too long! I'll update the banner PR to remove the `bannerTokens` structure so we can use the tokens or theme tokens directly! 

Thanks for the discussion 😄

jandrade: 🤔 one benefit of colocating tokens is that it makes less complex migrate from Aphrodite to CSS modules using AST/codemods.

For example, having the tokens live in the same file will simplify statically mapping and extracting the styles to the relevant tokens. Having them in a separate file might require an extra step of figuring out which file includes those variables and I'm not sure if AST is smart enough to calculate that, so that might require extra JS logic to do that.

I'm now thinking that component-level tokens (defined in `theme/*`) might be really interesting to migrate from Aphrodite to CSS modules 😅, we'll see what happens when we have to do that migration.

packages/wonder-blocks-switch/src/components/switch.tsx
```diff
@@ -54,7 +47,81 @@ type Props = Pick<
...
-const SwitchCore = React.forwardRef(function SwitchCore(
+const focusStylesObject = focusStyles.focus[":focus-visible"];
+
+const baseStyles = {
+    color: {
+        bg: {
+            switch: {
+                off: semanticColor.core.border.neutral.default,
+                disabledOff: semanticColor.core.border.disabled.strong,
+                activeOff: color.fadedOffBlack64,
```
jandrade: **suggestion:** This should change to a semantic color instead.

Based on the current colors, it looks like `semanticColor.core.border.neutral.strong` would be the best bet here as the `rest/default` state uses `neutral.default`. What do you think?

jandrade: To clarify, this color would slightly change in OG as currently is `offBlack64` and it would change to `ob72`, and it would use an existing TB color instead of hard-coding an OG one in both themes.

marcysutton: Makes sense! I updated it to use semanticColor.

jandrade (approved):
Looks great! thanks for all the improvements + cleanup. Really solid stuff!! 👏 🚀

PR #2656, opened 2025-06-06
Add thunderblocks theme to Banner

beaesguerra:
## Summary:
- Set up Thunderblocks styling for Banner
- Remove `layout` functionality. The Banner component continues to accept this prop, however, the prop is not used. The banner will always have rounded corners (we decided on this with design). Created WB-1979 for removing the `layout` prop completely since it's a breaking change
- Simplify spacing and alignment
- Use bold icons in the Banner (dismiss icon and default banner kind icons)

Classic:
image

Thunderblocks:
image

Issue: WB-1866

## Test plan:
- Review the Banner stories for both Classic and TB
  - `?path=/docs/packages-banner--docs#layouts&globals=theme:thunderblocks`
  - `?path=/docs/packages-banner--docs#layouts&globals=theme:default`


__docs__/wonder-blocks-banner/banner-testing-snapshots.stories.tsx
```diff
@@ -47,48 +47,61 @@ const rows = [
...
+                    title: "Button",
+                    onClick: action("onClick"),
+                },
+                {
+                    title: "Link",
+                    href: "/",
+                    type: "link",
+                },
+            ],
+            text: "Example banner text.",
```
beaesguerra, line 66: Added another example where the default text styling is used. The other example with the inline link uses a React node for the text so it doesn't show the default typography

__docs__/wonder-blocks-banner/banner-testing-snapshots.stories.tsx
```diff
@@ -164,8 +177,8 @@ const scenarios = [
         props: {
             text: reallyLongText,
             actions: [
-                {title: "Link 1", href: "/"},
-                {title: "Link 2", href: "/"},
+                {title: "Link 1", href: "/", type: "link"},
```
beaesguerra, line 180: Missed setting the `type` 😅 It was rendering a button prior to this change

packages/wonder-blocks-banner/src/components/banner.tsx
```diff
@@ -445,7 +477,15 @@ const styles = StyleSheet.create({
         justifyContent: "center",
     },
     link: {
-        fontSize: font.body.size.small,
+        fontFamily: bannerTokens.link.font.family,
```
beaesguerra: Styling the link based on the [banner examples with links](https://www.figma.com/design/EuFu0U7gqc1koXm8ZhlOLp/%E2%9A%A1%EF%B8%8F-Components?node-id=6962-1163&m=dev ) 

I think we may need to check the Link in general in terms of theming. The colors look correct, though it uses the Lato font still and the focus outline is different (there may be other differences too, those were the ones I've noticed so far!). If this makes sense, I can create a follow up ticket for theming the Link component! 

I included typography styling for the links in the banner anyways so that it follows the banner examples. The links in the examples have an underline on the links, whereas the Link component doesn't have an underline by default! 

Let me know though if this should be approached in another way!

__docs__/wonder-blocks-banner/banner.argtypes.ts
```diff
@@ -36,7 +36,6 @@ export default {
     kind: {
         control: {type: "select"},
         defaultValue: "info",
-        description: "Determines the color and icon of the banner.",
```
beaesguerra, line 39: Removed prop descriptions so that it uses the prop jsdocs (less duplication)!

__docs__/wonder-blocks-banner/banner.stories.tsx
```diff
@@ -122,6 +122,9 @@ export const Kinds: StoryComponentType = {
 };
 
 /**
+ * DEPRECATED: The `layout` prop is deprecated and will be removed in a future
```
beaesguerra, line 125: The layout prop is still accepted so it isn't a breaking change, but it isn't being used. Created https://khanacademy.atlassian.net/browse/WB-1979 for removing the `layout` prop in our cleanup epic!

Next, I'm going to test this PR snapshot in `webapp` and check existing uses of `Banner` to confirm that ignoring the `layout` prop is fine

packages/wonder-blocks-banner/src/components/banner.tsx
```diff
@@ -274,12 +274,7 @@ const Banner = (props: Props): React.ReactElement => {
 
     return (
         <View
-            style={[
-                styles.containerOuter,
-                layout === "floating" && styles.floatingLayout,
-                layout === "full-width" && styles.fullWidthLayout,
```
beaesguerra, line 280: Here is where we stop using the layout prop. We set the border radius in the `containerOuter` styles now

packages/wonder-blocks-banner/src/components/banner.tsx
```diff
@@ -296,7 +291,10 @@ const Banner = (props: Props): React.ReactElement => {
                 />
                 <View style={styles.labelAndButtonsContainer}>
                     <View style={styles.labelContainer}>
-                        <BodyText size="small">{text}</BodyText>
```
beaesguerra, line 299: Using a `StyledDiv` for now since we're unable to override BodyText default styles currently (WB-1978). The font size changes between Classic and TB so we use the component theme tokens (the css variables) instead

packages/wonder-blocks-banner/src/components/banner.tsx
```diff
@@ -324,41 +322,44 @@ const Banner = (props: Props): React.ReactElement => {
 const bannerTokens = {
     root: {
         border: {
-            radius: {
-                default: border.radius.radius_0,
-                floating: border.radius.radius_040,
```
beaesguerra, line 329: We no longer need to support the different banner layouts so we can flatten this structure! We use the `theme` because we use different radius values for Classic and TB

packages/wonder-blocks-banner/src/components/banner.tsx
```diff
@@ -324,41 +322,44 @@ const Banner = (props: Props): React.ReactElement => {
...
-            },
+            radius: theme.root.border.radius,
```
beaesguerra: All the banner token values that are replaced with `theme.*` means the value is different in TB (and we can't use semantic tokens to address the difference)! Let me know if you have any feedback on this approach

packages/wonder-blocks-banner/src/components/banner.tsx
```diff
@@ -445,7 +477,15 @@ const styles = StyleSheet.create({
...
-        fontSize: font.body.size.small,
+        fontFamily: bannerTokens.link.font.family,
+        fontSize: bannerTokens.link.font.size,
+        fontWeight: bannerTokens.link.font.weight,
+        lineHeight: bannerTokens.link.font.lineHeight,
+        textDecoration: bannerTokens.link.font.decoration,
+        textUnderlineOffset: bannerTokens.link.font.underlineOffset,
+    },
+    button: {
+        marginInline: bannerTokens.button.layout.marginInline,
```
beaesguerra: We set a negative inline margin in TB to account for the padding in the tertiary Button. This makes the text in the Button line up with the banner text when there are multiple lines of text

packages/wonder-blocks-banner/src/theme/thunderblocks.ts
```diff
@@ -0,0 +1,61 @@
...
+            info: semanticColor.feedback.info.subtle.icon,
+            success: semanticColor.feedback.success.subtle.icon,
+            warning: semanticColor.feedback.warning.subtle.icon,
+            critical: semanticColor.feedback.critical.subtle.icon,
+        },
+    },
+    button: {
+        layout: {
+            // Set the margin to negative to offset the padding of the button
+            marginInline: `calc(${sizing.size_080} * -1)`,
```
beaesguerra, line 49: I tried importing the component tokens from the button so that this could stay in sync. I wasn't able to though since we don't export the component tokens. Do we want to be able to use other component tokens in other packages?

packages/wonder-blocks-banner/src/components/banner.tsx
```diff
@@ -93,7 +88,10 @@ type Props = {
      */
     kind?: BannerKind;
     /**
-     * Determines the edge style of the Banner.
+     * (DEPRECATED) Determines the edge style of the Banner.
+     *
+     * This prop is deprecated and will be removed in a future release.
+     * Currently, it has no effect on the component.
```
jandrade, line 94: **suggestion:** You could also set the `@deprecated` JSDoc tag, that way the IDE will show this info to consumers.

packages/wonder-blocks-banner/src/components/banner.tsx
jandrade: makes sense, thanks for adding the comment!

packages/wonder-blocks-banner/src/components/banner.tsx
jandrade: that's the way I've approached this as well :)

packages/wonder-blocks-banner/src/components/banner.tsx
```diff
@@ -374,6 +375,12 @@ const bannerTokens = {
         layout: {
             margin: sizing.size_080,
         },
+        font: {
+            size: theme.label.font.size,
+            weight: font.weight.semi,
+            lineHeight: font.body.lineHeight.small,
+            family: font.family.sans,
```
jandrade: **question:** Do you need to set `fontFamily` here? I wonder if this can be let unset and let the page/document inject that globally?

packages/wonder-blocks-banner/src/theme/default.ts
```diff
@@ -0,0 +1,60 @@
...
+                success: semanticColor.core.border.success.default,
+                warning: semanticColor.core.border.warning.default,
+                critical: semanticColor.core.border.critical.default,
+            },
+            foreground: {
+                info: semanticColor.core.foreground.neutral.strong,
+                success: semanticColor.core.foreground.neutral.strong,
+                warning: semanticColor.core.foreground.neutral.strong,
+                critical: semanticColor.core.foreground.neutral.strong,
+            },
```
jandrade: **question/thought:** Is it possible at this point to reuse the `feedback` tokens instead of `core` ones here? I thought that `feedback` would be reused between Badges and Banners.

packages/wonder-blocks-banner/src/theme/thunderblocks.ts
jandrade: I think it is fine to keep it this way for now. This could be a good case for starting semantic non-color tokens at some point, and move this instead somewhere inside the `wb-tokens` package that could be consumed here and in `wb-button/wb-icon-button`.

packages/wonder-blocks-banner/src/components/banner.tsx
beaesguerra: Without setting `fontFamily`, the TB theme uses "Lato" still! 

Might be because we set the font to `Lato` on the body? https://github.com/Khan/wonder-blocks/blob/2b98a60f8d77f89aebfd8ed0b8eea9e537a86c1e/.storybook/preview-head.html#L24 https://github.com/Khan/wonder-blocks/blob/2b98a60f8d77f89aebfd8ed0b8eea9e537a86c1e/.storybook/manager-head.html#L15

Is it preferred to omit setting the `fontFamily` in the components so that it uses the global styles instead?

cc: @marcysutton in case you have thoughts on how the default font should work in Storybook with theming!

packages/wonder-blocks-banner/src/theme/default.ts
beaesguerra: Good question! I updated the banner styling so that the root foreground is no longer themed. The `feedback` tokens handles setting the font color to `offBlack` instead of the status colors! 

The `feedback` tokens are reused between Badges and Banners for Thunderblocks. I kept the theming for the border and icon colors because the usage in Classic is different! 

**Border:**
This is what it looks like when we use the feedback border tokens in classic:
image
The feedback border tokens in the classic theme work well for Badges though:
image
I think this is more because we're using `border` for the left indicator in Classic, rather than a border that goes all around the banner

**Icons:**
- We use theming for the icon color because the classic design uses gray icons, while TB uses colored icons via the feedback color tokens!

jandrade: ah I see, that makes total sense! thanks for all the context.

__docs__/wonder-blocks-banner/banner.stories.tsx
beaesguerra: I checked in perseus and frontend to see if there's anything specific relying on the full-width layout and didn't find anything conflicting so ignoring `layout` should be fine. It helps that we weren't exposing a `style/styles` prop (I'll work on adding support for that after these changes!)

packages/wonder-blocks-banner/src/components/banner.tsx
```diff
@@ -321,102 +321,28 @@ const Banner = (props: Props): React.ReactElement => {
...
+        // Stop the square corners of the inner container from
+        // flowing out of the rounded corners of the outer container.
+        overflow: "hidden",
+        padding: sizing.size_160,
+        paddingInlineStart: theme.root.layout.paddingInlineStart,
     },
     containerInner: {
         flexDirection: "row",
-        padding: bannerTokens.root.layout.padding,
+        gap: theme.root.layout.gap,
```
jandrade, line 340: yay!!

jandrade (approved):
Everything looks great! thanks for doing all the extra things (refactoring, styling, modernizing, etc) 👏 🚀

packages/wonder-blocks-banner/src/components/banner.tsx
beaesguerra: For reference later on, we discussed how we want to use static component tokens in https://github.com/Khan/wonder-blocks/pull/2655#discussion_r2132852528 !

PR #2657, opened 2025-06-09
[WB-1959] Add thunderblocks theme to CompactCell and DetailCell

jandrade:
## Summary:

- Updates `CompactCell` and `DetailCell` to support the `thunderblocks` theme.
- Refactors internal typography instances to support both themes.

Issue: WB-1959

## Test plan:

Navigate to the `CompactCell` and `DetailCell` components in both the
`thunderblocks` and `default` themes to ensure that the styles look correct.

- CompactCell OG: `/?path=/docs/packages-cell-compactcell--docs&globals=theme:default`
- CompactCell TB: `/?path=/docs/packages-cell-compactcell--docs&globals=theme:thunderblocks`
- DetailCell OG: `/?path=/docs/packages-cell-detailcell--docs&globals=theme:default`
- DetailCell TB: `/?path=/docs/packages-cell-detailcell--docs&globals=theme:thunderblocks`


packages/wonder-blocks-cell/src/theme/thunderblocks.ts
jandrade, line 1: **note:** Use this for reference: https://www.figma.com/design/EuFu0U7gqc1koXm8ZhlOLp/%E2%9A%A1%EF%B8%8F-Components?node-id=5353-941&m=dev

packages/wonder-blocks-cell/src/theme/default.ts
```diff
@@ -7,7 +12,11 @@ export default {
                 default: border.width.medium,
                 selected: border.width.thick,
             },
-            radius: border.radius.radius_040,
+            radius: {
+                default: border.radius.radius_0,
+                focus: border.radius.radius_040,
+                press: border.radius.radius_0,
+            },
```
beaesguerra, line 19: (A thought for later!) - Several of our components change border radius on focus, I wonder if it would make sense to make global semantic border radius tokens for the different states so we can change the behaviour for all those components in one place. Or maybe it makes sense to make it different depending on the component!

beaesguerra (commented):
Looks good so far! I compared the statesheets with the designs and left some comments in Chromatic!

packages/wonder-blocks-cell/src/theme/default.ts
jandrade: That's a great thought! I feel that we have consolidated a single focus radius token in TB, but I'm not fully sure if we have that in OG. If we have it, then it makes absolute sense to create that more globally.

.changeset/kind-rings-tan.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-cell": minor
```
beaesguerra, line 2: Do we want to include the tokens package too? (might be fine without too if you have other PRs in the feature branch that will mark the `tokens` package!)

beaesguerra (approved):
Approved the snapshots! Thanks for doing a follow up to simplify styling 😄

.changeset/kind-rings-tan.md
jandrade: yep, thanks for the reminder! the `patch` will go in the next PR. btw Do you think it should be `patch` or `minor`?

beaesguerra: Hmmm good question. I was thinking `minor` because it adds new css variables.. though the css variables aren't really a part of the public api right now (for example, we don't include the component css variables in the docs)! So maybe a `patch`!

If we have a separate package for the css variables for all packages, I think that would make more sense for that to be a minor update!

PR #2658, opened 2025-06-09
Update code snippet for running deploy_wonder_blocks

beaesguerra:
## Summary:

Updates the deploy_wonder_blocks code snippet now that we're using the frontend repo

Issue: WB-XXXX

## Test plan:
Confirm that the updated command works as expected


jandrade (approved):
Good catch! 🚢

PR #2659, opened 2025-06-10
[WB-1868.1] Dropdown: use Cell default styles

jandrade:
## Summary:

This is a prep PR to simplify some of the dropdown styles before adding theming support to it.

- Simplifies the `OptionItem` and `ActionItem` styles to just use the default CellCore styles.
- Simplifies the `OptionItem` structure so the same layout can be reused by `Single/MultiSelect` and `Listbox`.
- Adds a `tabIndex` prop to `Cell` so the component can work with different focus modes (needed for visual focus on Listbox/Combobox).
- Updates the "check" icon color to be `blue` instead of `offBlack`. 

### Implementation plan: 

1. #2659
2. Add theming support to wonder-blocks-dropdown package

Issue: https://khanacademy.atlassian.net/browse/WB-1868

## Test plan:

Navigate to the Dropdown docs and verify that the stories look correct.


packages/wonder-blocks-dropdown/src/components/combobox.tsx
```diff
@@ -639,7 +639,13 @@ export default function Combobox({
                     referenceElement={rootNodeRef?.current as HTMLElement}
                 >
                     {(isReferenceHidden) => (
-                        <>
+                        <View
+                            onMouseDown={(e) => {
+                                // Prevents the combobox from losing focus when clicking
+                                // on the option item.
+                                e.preventDefault();
+                            }}
```
beaesguerra, line 647: Is the equivalent logic needed for keyboard interactions?

packages/wonder-blocks-dropdown/src/components/option-item.tsx
```diff
@@ -8,16 +8,12 @@ import {
     border,
     sizing,
 } from "@khanacademy/wonder-blocks-tokens";
-import {LabelMedium, LabelSmall} from "@khanacademy/wonder-blocks-typography";
+import {LabelMedium} from "@khanacademy/wonder-blocks-typography";
```
beaesguerra, line 11: Could we use the BodyText component? (not sure if you had plans for that later!)

packages/wonder-blocks-dropdown/src/components/option-item.tsx
```diff
@@ -143,8 +139,6 @@ type DefaultProps = {
     selected: OptionProps["selected"];
 };
 
-const StyledLi = addStyle("li");
```
beaesguerra, line 146: Did we intend to remove using the `li` semantics for combobox option items?

beaesguerra (commented):
It's looking good, the styling improvements are great! Left some questions in the PR and chromatic!

packages/wonder-blocks-dropdown/src/components/option-item.tsx
jandrade: good question! I had plans to add it in the next PR as this one was already including a bunch of changes, but I can add it here if you prefer.

packages/wonder-blocks-dropdown/src/components/option-item.tsx
jandrade: Yeah! I saw that it had this structure:

```
listbox
  > li
  > il
```

I changed it to use the correct semantics:
```
  listbox
    > option
    > option
```

Let me know if you see something off.

packages/wonder-blocks-dropdown/src/components/combobox.tsx
jandrade: Yes, I just moved that from `option-item` to this component directly to simplify the implementation (forgot to add a note here).

https://github.com/Khan/wonder-blocks/pull/2659/files#diff-38e10b1fd4a491d9c1d9de42989fb83e7da8ad1a88171f0976b73847b03ba2d7L281-L285

packages/wonder-blocks-dropdown/src/components/option-item.tsx
beaesguerra: In another PR is fine too :)

beaesguerra (approved):
Looks great! 🚀

PR #2661, opened 2025-06-11
[WB-1959.3] Cell layout refactor

jandrade:
## Summary:

Following up with the Cell theming support, this PR simplifies the layout and structure of the cell component.

Issue: https://khanacademy.atlassian.net/browse/WB-1959

## Test plan:

Navigate to the Cell docs and verify that the layout looks correct.

Also check the option items in dropdown examples.

### DIFF
--- a/__docs__/wonder-blocks-cell/detail-cell.stories.tsx
+++ b/__docs__/wonder-blocks-cell/detail-cell.stories.tsx
@@ -4,13 +4,8 @@ import {MemoryRouter} from "react-router-dom";
 import {CompatRouter, Route, Routes} from "react-router-dom-v5-compat";
 import type {Meta, StoryObj} from "@storybook/react";
 
-import {PropsFor, View} from "@khanacademy/wonder-blocks-core";
-import {
-    border,
-    color,
-    semanticColor,
-    spacing,
-} from "@khanacademy/wonder-blocks-tokens";
+import {View} from "@khanacademy/wonder-blocks-core";
+import {color, semanticColor, spacing} from "@khanacademy/wonder-blocks-tokens";
 import {PhosphorIcon} from "@khanacademy/wonder-blocks-icon";
 
 import {DetailCell} from "@khanacademy/wonder-blocks-cell";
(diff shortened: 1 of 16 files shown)


packages/wonder-blocks-cell/src/components/internal/__tests__/cell-core.test.tsx
```diff
@@ -125,39 +125,6 @@ describe("CellCore", () => {
         ).toBeInTheDocument();
     });
 
-    it("should pass an style to the top node", () => {
```
jandrade, line 128: **note:** Removed these tests as we now cover styles via Chromatic snapshots (also as rootStyle is going away).

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -93,32 +94,16 @@ function CellInner(props: CellCoreProps): React.ReactElement {
...
-                // custom styles
-                style,
-                horizontalRuleStyles,
-                active && styles.activeInnerWrapper,
-            ]}
-            // Set className so we can set styles on the inner wrapper directly
-            // when the clickable element is pressed
-            className="inner-wrapper"
-        >
+        <>
```
jandrade, line 106: **note:** This is one of the main changes here. I was able to get rid of this `inner-wrapper` container now that the styles are more aligned from the design and a11y perspective.

jandrade:
This PR is ready for review. I've tested this changes in the `frontend` repo here: https://github.com/Khan/frontend/pull/358

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -300,68 +272,62 @@ const styles = StyleSheet.create({
...
-        ":active": {
-            background: semanticColor.core.background.instructive.subtle,
-            borderRadius: theme.root.border.radius.press,
-        },
+
         // focus (only visible when using keyboard navigation)
         ":focus-visible": {
             borderRadius: theme.root.border.radius.focus,
+            outline: focusStyles.focus[":focus-visible"].outline,
+            outlineOffset: `calc(${theme.root.border.width.default} * -1)`,
```
beaesguerra, line 286: Is there a reason why we don't use the global focus styles as is with the box-shadow too?

packages/wonder-blocks-cell/src/util/types.ts
```diff
@@ -103,13 +103,6 @@ export type CellProps = {
      * A custom role for the cell.
      */
     role?: ClickableRole;
-    /**
-     * Optional custom styles applied to the top node.
-     *
-     * _NOTE:_ This is the top node of the cell, not the cell container. If
-     * possible, try to use this prop carefully and use `style` instead.
-     */
-    rootStyle?: StyleType;
```
beaesguerra, line 112: Thanks for simplifying how consumers can provide custom styles!

beaesguerra: Do you think it'd be helpful to combine the `style` and `rightAccessoryStyle` and `leftAccessoryStyle` props into one `styles` prop? (Can be for later on since we're focusing on theming!)

__docs__/wonder-blocks-cell/detail-cell.stories.tsx
```diff
@@ -381,18 +380,18 @@ export const CustomRootStyle = {
     render(args: PropsFor<typeof DetailCell>) {
         return (
             <View style={{gap: spacing.large_24}}>
-                Active (with rootStyle prop):
+                Active
                 <DetailCell
                     {...args}
-                    rootStyle={{borderRadius: border.radius.radius_120}}
+                    style={{borderRadius: border.radius.radius_120}}
```
beaesguerra: The border radius doesn't seem to get applied to the component! https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=684c8f1e5b6b9675df7fee18

beaesguerra (commented):
Thanks for simplifying the styling in Cell 😄 I left some comments in the PR and Chromatic!

packages/wonder-blocks-cell/src/util/types.ts
jandrade: Agreed, I created this ticket for that specific reason: https://khanacademy.atlassian.net/browse/WB-1990

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
jandrade: The reason is because we can't use offset with boxShadow, and I had to use it with `outline` to prevent the overlap issue we have with cells in content pages. I could try with `inset` but I'm not fully sure if that will work.

I'm thinking that for TB, I'm not going to use the `offset` and instead let cell use global `focusStyles` instead (cells have different uses between experiences).

__docs__/wonder-blocks-cell/detail-cell.stories.tsx
jandrade: hmmm this is another tricky case as the snapshot is testing that the radius changes in `:active`, but in here we are overriding the radius for the default/rest state.

I had to change how radius works in Cell due to some differences in how that works in TB.

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
jandrade: ok I got it working with `box-shadow`'s inset + adding a thicker spread :) (added a comment to mention this). It doesn't use the global token, but it works in a similar way to make it work for this very specific use case.

beaesguerra (commented):
Thanks for the updates! Reviewed the new snapshots and wanted to confirm 2 more things with you before approving!
- [Classic cell press border radius](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=68506f496dd27c384ca395f5) (shown in multiple stories)
- [TB OptionItem focus outline](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=68506f496dd27c384ca3965a)

beaesguerra (approved):
Looks good! 🚀 Thanks for making these improvements!

PR #2662, opened 2025-06-11
Add theming to Checkbox/Radio

marcysutton:
## Summary:
Adds theming support to Checkbox and Radio, including the Thunderblocks theme.

Notable differences in TB include:
- Increased hover and target click areas (4px)
- 4px border-radius instead of 3px
- Semantic colors: Component/Input/Selected blue

Issue: WB-1972, WB-1974

## Test plan:
1. Review Test sheets for Checkbox and Radio
2. Test clicking of padded wrapper around both component types
3. Compare to Figma designs
   - TB: https://www.figma.com/design/EuFu0U7gqc1koXm8ZhlOLp/%E2%9A%A1%EF%B8%8F-Components?node-id=2048-52&t=B2xxLXv7GkDySMja-0
   - OG: https://www.figma.com/design/VbVu3h2BpBhH80niq101MHHE/%F0%9F%92%A0-Main-Components?node-id=15197-6798&t=zmIZ176DtJ5rmwyU-0

### DIFF
--- a/.changeset/dull-experts-deny.md
+++ b/.changeset/dull-experts-deny.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-form": minor
+---
+
+Add Thunderblocks theme to Checkbox and Radio components
--- a/packages/wonder-blocks-dropdown/package.json
+++ b/packages/wonder-blocks-dropdown/package.json
@@ -28,7 +28,8 @@
     "@khanacademy/wonder-blocks-search-field": "workspace:*",
     "@khanacademy/wonder-blocks-timing": "workspace:*",
     "@khanacademy/wonder-blocks-tokens": "workspace:*",
-    "@khanacademy/wonder-blocks-typography": "workspace:*"
+    "@khanacademy/wonder-blocks-typography": "workspace:*",
+    "@khanacademy/wonder-blocks-form": "workspace:*"
   },
   "peerDependencies": {
     "@phosphor-icons/core": "catalog:",
--- a/packages/wonder-blocks-form/src/components/checkbox-core.tsx
+++ b/packages/wonder-blocks-form/src/components/checkbox-core.tsx
@@ -30,10 +30,12 @@ function mapCheckedToAriaChecked(value: Checked): AriaChecked {
     }
 }
 
-// The checkbox size
-const size = spacing.medium_16;
-// The check icon size
-const checkSize = spacing.small_12;
+const baseStyles = {
+    // The checkbox size
+    size: spacing.medium_16,
+    // The check icon size
+    checkSize: spacing.small_12,
+};
 
 const StyledInput = addStyle("input");
 
@@ -83,8 +85,8 @@ const CheckboxCore = React.forwardRef(function CheckboxCore(
                 sharedStyles.checkboxIcon,
                 // The check icon is smaller than the checkbox, as per design.
                 {
-                    width: checkSize,
-                    height: checkSize,
+                    width: baseStyles.checkSize,
+                    height: baseStyles.checkSize,
                 },
             ]}
         />
@@ -132,10 +134,10 @@ const sharedStyles = StyleSheet.create({
     },
 
     default: {
-        height: size,
-        width: size,
-        minHeight: size,
-        minWidth: size,
+        height: baseStyles.size,
+        width: baseStyles.size,
+        minHeight: baseStyles.size,
+        minWidth: baseStyles.size,
         margin: 0,
         outline: "none",
         boxSizing: "border-box",
@@ -155,7 +157,7 @@ const sharedStyles = StyleSheet.create({
         position: "absolute",
         pointerEvents: "none",
         // This margin is to center the check icon in the checkbox.
-        margin: (size - checkSize) / 2,
+        margin: (baseStyles.size - baseStyles.checkSize) / 2,
     },
 });
 
--- a/packages/wonder-blocks-form/src/components/checkbox-core.tsx
+++ b/packages/wonder-blocks-form/src/components/checkbox-core.tsx
@@ -73,8 +73,8 @@ const CheckboxCore = React.forwardRef(function CheckboxCore(
         <PhosphorIcon
             color={
                 disabled
-                    ? semanticColor.icon.disabled
-                    : semanticColor.icon.inverse
+                    ? theme.icon.disabled.foreground
+                    : theme.icon.default.foreground
             }
             icon={checked ? checkIcon : minusIcon}
             size="small"
@@ -195,19 +195,13 @@ const _generateStyles = (
         },
         // Form validation error state
         error: {
-            border: semanticColor.choice.error.border,
-            background: semanticColor.choice.error.background,
+            border: theme.choice.error.border,
+            background: theme.choice.error.background,
         },
         // Disabled state
-        disabled: {
-            border: semanticColor.choice.disabled.border,
-            background: semanticColor.choice.disabled.background,
-        },
+        disabled: theme.choice.disabled,
         // Disabled and checked state
-        disabledChecked: {
-            border: semanticColor.choice.disabledChecked.border,
-            background: semanticColor.choice.disabledChecked.background,
-        },
+        disabledChecked: theme.choice.disabledChecked,
     };
 
     let stateStyles: Record<string, any> = {};
@@ -271,7 +265,7 @@ const _generateStyles = (
             inputWrapper: {
                 ":hover input:not([disabled])": {
                     backgroundColor: error
-                        ? (semanticColor.choice as any).error.background
+                        ? theme.choice.error.background
                         : colorAction.hover.background,
                     outline: `${border.width.medium} solid ${colorAction.hover.border}`,
                     outlineOffset: -1,
@@ -286,7 +280,7 @@ const _generateStyles = (
                 // TODO(WB-1864): Use focusStyles.focus
                 ":focus-visible": {
                     backgroundColor: error
-                        ? (semanticColor.choice as any).error.background
+                        ? theme.choice.error.background
                         : colorAction.hover.background,
                     outline: `${border.width.medium} solid ${semanticColor.focus.outer}`,
                     outlineOffset: -1,
--- a/packages/wonder-blocks-form/src/components/checkbox-core.tsx
+++ b/packages/wonder-blocks-form/src/components/checkbox-core.tsx
@@ -10,6 +10,7 @@ import {addStyle} from "@khanacademy/wonder-blocks-core";
 import {PhosphorIcon} from "@khanacademy/wonder-blocks-icon";
 import checkIcon from "@phosphor-icons/core/bold/check-bold.svg";
 import minusIcon from "@phosphor-icons/core/bold/minus-bold.svg";
+import theme from "../theme/index";
 
 import type {ChoiceCoreProps, Checked} from "../util/types";
 
@@ -139,16 +140,15 @@ const sharedStyles = StyleSheet.create({
         outline: "none",
         boxSizing: "border-box",
         borderStyle: "solid",
-        borderWidth: border.width.thin,
-        // TODO(WB-1864): Use the correct token once TB is updated.
-        borderRadius: 3,
+        borderWidth: theme.root.border.width.default,
+        borderRadius: theme.checkbox.border.radius.default,
     },
 
     disabled: {
         cursor: "auto",
         backgroundColor: semanticColor.input.disabled.background,
         borderColor: semanticColor.input.disabled.border,
-        borderWidth: border.width.thin,
+        borderWidth: theme.root.border.width.default,
     },
 
     checkboxIcon: {
--- a/packages/wonder-blocks-form/src/components/checkbox-core.tsx
+++ b/packages/wonder-blocks-form/src/components/checkbox-core.tsx
@@ -63,7 +63,7 @@ const CheckboxCore = React.forwardRef(function CheckboxCore(
         stateStyles.default,
     ];
 
-    const wrapperStyle = [theme.inputWrapper, stateStyles.inputWrapper];
+    const wrapperStyle = [sharedStyles.inputWrapper, stateStyles.inputWrapper];
 
     const checkboxIcon = (
         <PhosphorIcon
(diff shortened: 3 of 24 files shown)


packages/wonder-blocks-dropdown/package.json
```diff
@@ -28,7 +28,8 @@
     "@khanacademy/wonder-blocks-search-field": "workspace:*",
     "@khanacademy/wonder-blocks-timing": "workspace:*",
     "@khanacademy/wonder-blocks-tokens": "workspace:*",
-    "@khanacademy/wonder-blocks-typography": "workspace:*"
+    "@khanacademy/wonder-blocks-typography": "workspace:*",
+    "@khanacademy/wonder-blocks-form": "workspace:*"
```
beaesguerra, line 32: Ohhh it makes sense that we need `wonder-blocks-form` in `wonder-blocks-dropdown` since we import the `TextField` component in the dropdown package. It is odd that it is only showing an error now since it was importing it before already 🤔

.changeset/dull-experts-deny.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-form": minor
```
beaesguerra, line 2: We can include the tokens package to the changeset too to be sure that it gets updated with the new css variables! (related https://github.com/Khan/wonder-blocks/pull/2660#issuecomment-2967880943)

packages/wonder-blocks-form/src/components/radio-core.tsx
```diff
@@ -30,63 +32,88 @@ const StyledInput = addStyle("input");
...
+};
+
 const disabledChecked = {
     position: "absolute",
-    top: size / 4,
-    left: size / 4,
-    height: size / 2,
-    width: size / 2,
-    borderRadius: border.radius.radius_full,
+    top: `calc(${baseStyles.size} * .25 + ${theme.inputWrapper.padding})`,
```
beaesguerra: Can we add a comment for why we multiply by .25 😄

packages/wonder-blocks-form/src/theme/default.ts
```diff
@@ -0,0 +1,28 @@
+import {border} from "@khanacademy/wonder-blocks-tokens";
+
+export default {
+    inputWrapper: {
+        padding: "0px",
+        margin: "0px",
```
beaesguerra: Could we use `sizing.size_0` for spacing? Do we also want to group this into a `layout` group following the [component token structure](https://khanacademy.atlassian.net/wiki/spaces/WB/pages/3967287319/Component+tokens+structure)?

```
inputWrapper: {
  layout: {
    padding: sizing.size_0,
    margin: sizing.size_0,
  }
}
```

beaesguerra (commented):
Nice work so far! Took a first pass at comparing the snapshots with the TB designs and left some comments in the PR and Chromatic. The TB radio button doesn't show up in the Chromatic diffs for some reason, so I didn't leave comments in Chromatic for that component (there are some visual differences though, like the disabled radio button colors!)

beaesguerra:
Thanks for making these updates! Left some comments in Chromatic!

packages/wonder-blocks-form/src/components/radio-core.tsx
```diff
@@ -22,170 +27,162 @@ const StyledInput = addStyle("input");
...
+            },
+            default: {
+                backgroundColor: checkedStyles.rest.background,
+                borderColor: checkedStyles.rest.border,
+                // borders need to render in pixels for consistent size
+                borderWidth: `calc(${baseStyles.radio.sizing.size} / 4)`,
+
+                ":focus-visible:not([disabled])": {
+                    // Use the global focus style
+                    ...focusStyles.focus,
```
jandrade: **suggestion:** I think you'll need to target the `:focus-visible` pseudo-class here.

```suggestion
                    // Use the global focus style
                    ...focusStyles.focus[":focus-visible],
```

Also this should work a bit better (no need to spread the object):

`":focus-visible:not([disabled])": focusStyles.focus[":focus-visible]`

packages/wonder-blocks-form/src/components/radio-core.tsx
```diff
@@ -22,170 +27,162 @@ const StyledInput = addStyle("input");
...
-                // TODO(WB-1864): Revisit hover, press tokens
-                ":hover": {
-                    backgroundColor: error
-                        ? states.error.background
-                        : colorAction.hover.background,
-                    outline: `${border.width.medium} solid ${colorAction.hover.border}`,
-                    outlineOffset: -1,
+                ":focus-visible:not([disabled])": {
+                    // Use the global focus style
+                    ...focusStyles.focus,
```
jandrade: Same suggestion around `:focus-visible` here.

packages/wonder-blocks-form/src/theme/default.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export default {
+    inputWrapper: {
+        padding: sizing.size_0,
+        margin: sizing.size_0,
+    },
+    checkbox: {
+        border: {
+            radius: {
+                default: "3px",
```
jandrade: **question:** What happens if you use a valid `border_radius` token here? would it change too much the layout?

packages/wonder-blocks-form/src/util/styles.ts
```diff
@@ -0,0 +1,210 @@
...
+        default: {
+            foreground: semanticColor.core.foreground.inverse.strong,
+        },
+        disabled: {
+            // the semanticColor isn't working for this icon.
+            // our approach changed to use the TB theme in Classic,
+            // but the icon foreground color is themed and no longer works.
+            // semanticColor for TB: core.border.neutral.subtle (too dark in classic)
+            foreground: "#CBCBCD",
+        },
```
jandrade: **suggestion:** If this is for the icon, would `core.foreground.neutral.subtle` work instead? I wonder if this token is worth moving to the `theme` folder.

packages/wonder-blocks-form/src/components/radio-core.tsx
marcysutton: Oddly it is working without `focus-visible`! Clicking a checkbox doesn't persist a focus style, while the focus style is visible with the keyboard.

packages/wonder-blocks-form/src/theme/default.ts
marcysutton: I wanted to keep the Classic style as close as possible, and it had a `3px` radius. https://github.com/Khan/wonder-blocks/blob/038f9a91992205ca71de1999a28b800dea57f4f3/packages/wonder-blocks-form/src/components/checkbox-core.tsx#L144

But now that you mention it, we went with the TB style for other characteristics. This was one of the few differences I kept between themes. TB has a `4px` radius. Should we check with Design on that one or just change it? I don't think a border-radius will really affect the layout. It might not even be that noticeable.

packages/wonder-blocks-form/src/util/styles.ts
marcysutton: I do think we need to use a different color, as the token in Figma is too dark on the background color.

Bea and I both thought this was too hard to see at the default size (TB screenshot):
Disabled checkbox screenshot with gray checkmark on gray

This is with `core.foreground.neutral.subtle` on Classic. I think the checkmark looks better:
Disabled checkbox screenshot with lighter gray checkmark

I committed that change, thank you for the suggestion!

marcysutton: And now that I see it, that Label text is suuuper light in Classic too. Disabled components are exempt from contrast requirements, but it still gets into the realm of usability. I think we need some consistency across disabled components and labels, particularly for text colors....but for text sizes too. Maybe we could do an overall polishing effort to make sure compound forms look good with components used together?

beaesguerra: > And now that I see it, that Label text is suuuper light in Classic too

Maybe the disabled label text color is something we can theme for these components? I find the [current disabled text styling in Classic](https://khan.github.io/wonder-blocks/?path=/story/packages-form-radiogroup--basic) easier to read! 
image

> Maybe we could do an overall polishing effort to make sure compound forms look good with components used together

Agreed! The placement of the error message for `CheckboxGroup` and `RadioGroup` is different from how we place error messages under the field in `LabeledField`. Something we can make more consistent as well!

packages/wonder-blocks-form/src/components/radio-core.tsx
marcysutton: Ah thanks, I missed one!

packages/wonder-blocks-form/src/components/checkbox-group.tsx
```diff
@@ -129,28 +128,44 @@ const CheckboxGroup = React.forwardRef(function CheckboxGroup(
...
-                    {description}
-                </LabelSmall>
-            )}
-            {errorMessage && (
-                <LabelSmall style={styles.error}>{errorMessage}</LabelSmall>
-            )}
-            {(label || description || errorMessage) && (
-                <Strut size={spacing.small_12} />
-            )}
+            <View
```
marcysutton: We might want to go about this a different way... To remove `Strut`, I wrapped the affected elements in a `View` and put styling on it instead. But it broke the direct descendant behavior for `fieldset > legend`. I worked around this by explicitly adding an `aria-labelledby` attribute onto the fieldset, so it would have an accessible name exposed. But this might be a case where `Strut` is better suited for the problem.

packages/wonder-blocks-form/src/util/styles.ts
marcysutton: I think this is best suited for a follow-up issue on disabled colors as a whole. Here's why: the Classic token uses `core.foreground.disabled.default`, which has a different value per theme:

- OG: `core.foreground.disabled.default` is `default: color.fadedOffBlack32`
- TB: `core.foreground.disabled.default` is `default: color.gray_50`

Maybe I could hard-code a primitive value in the component theme? Or is there a better way?

packages/wonder-blocks-form/src/components/checkbox-group.tsx
jandrade: what if you move the gap style to the `fieldset` element directly? that way, there's no need to define an intermediate wrapper so the fieldset > legend connection is not affected (FWIW that fieldset instance uses `flex`)

packages/wonder-blocks-form/src/theme/default.ts
jandrade: I'd say that changing to `4px` should be a safe change, but better double-checking with design as you suggest.

packages/wonder-blocks-form/src/util/styles.ts
beaesguerra: I think we want to avoid using primitive values directly! One we way we can approach it is to have a component theme where different semantic tokens are used for OG and TB. 

I'm doing something similar for the label in LabeledField:

https://github.com/Khan/wonder-blocks/blob/1175c085cb814bf39ef6cfdaa42893f0ad8159ce/packages/wonder-blocks-labeled-field/src/theme/default.ts#L19-L21

https://github.com/Khan/wonder-blocks/blob/1175c085cb814bf39ef6cfdaa42893f0ad8159ce/packages/wonder-blocks-labeled-field/src/theme/thunderblocks.ts#L21-L23

packages/wonder-blocks-form/src/components/checkbox-core.tsx
```diff
@@ -60,69 +55,86 @@ const CheckboxCore = React.forwardRef(function CheckboxCore(
...
 
     const defaultStyle = [
         sharedStyles.inputReset,
         sharedStyles.default,
-        !disabled && stateStyles.default,
-        disabled && sharedStyles.disabled,
+        stateStyles.default,
     ];
 
+    const wrapperStyle = [theme.inputWrapper, stateStyles.inputWrapper];
```
beaesguerra: What do you think about explicitly creating an aphrodite style for the `inputWrapper` css properties? In this case, `theme.inputWrapper` has valid CSS properties so it works out! But if we add other tokens to that structure it may not work. What comes to mind for example is if we add a `foreground` property to that object, it won't map to a css property!

packages/wonder-blocks-form/src/theme/default.ts
```diff
@@ -0,0 +1,18 @@
+import {border, sizing} from "@khanacademy/wonder-blocks-tokens";
+
+export default {
+    inputWrapper: {
+        padding: sizing.size_0,
+        margin: sizing.size_0,
+    },
```
beaesguerra: Since this theme is for the `form` package, I am also adding to this structure for shared [component theme tokens for TextField and TextArea](https://github.com/Khan/wonder-blocks/pull/2684/files#diff-3aa38bf91cd8d103ad68b72a96098cef44b1e7b7d1ce759a82558cd30755dd7cR4)! Since `inputWrapper` will only apply to radio/checkboxes, would it make sense to create a group for it? Something like:

```tsx
export default {
  choice: { // <-- new choice group
    inputWrapper: {
      layout: { // <-- also new, to group the layout properties together
        padding: sizing.size_0,
        margin: sizing.size_0,
      }
    }
  },
  checkbox: { ... },
  field: { ... }, // <-- component theme tokens for TextField and TextArea
}
```
Let me know what you think!

marcysutton: That makes total sense! Thanks Bea!

packages/wonder-blocks-form/src/components/checkbox-group.tsx
marcysutton: The problem I ran into was fieldset potentially containing multiple elements: `StyledLegend`,  `description`, and `errorMessage`. A gap would apply between all of these, rather than at the end. I wanted to group them and apply a margin to the container. I just pushed an update that wraps the children below in a View, and optionally puts `marginBlockStart` at the top of that instead!

packages/wonder-blocks-form/src/components/checkbox-core.tsx
marcysutton: Good call. I added that!

packages/wonder-blocks-form/src/components/checkbox-group.tsx
marcysutton: Okay, that didn't work -- I couldn't keep the flex properties on the children and put a margin on it. I opted to restore `Strut` and kick the can down the road!

beaesguerra (approved):
Looks good to me! 🚀

PR #2663, opened 2025-06-11
Add `styles` prop to `Banner` so that the `root` element can be styled

beaesguerra:
## Summary:
- Allows the `root` element of the `Banner` component to be styled using a `styles` prop
- This supports the use case where consumers may need to apply layout properties to the banner so it works with existing layouts (for example, setting `flex-shrink: 0` on the banner so it doesn't shrink)
- Currently, we only support custom styling for the `root` element. This may change later on if we decide to support the customization of other elements in the component

- Stories:
  - stories clean up: remove use of color tokens, spacing tokens, and strut usage
  - set up story that needs to have custom styles supported

Issue: WB-1980

## Test plan:
- Confirm that the custom styles story renders as expected (banner does not shrink) (`/?path=/docs/packages-banner--docs#with-custom-styles`)
- Review prop docs and Custom Styles docs for Banner (`/?path=/docs/packages-banner--docs#with-custom-styles`)
- Note: I've pre-accepted the new story in Chromatic so that we can see what the visual difference is once we support the `styles` prop

### DIFF
--- a/__docs__/wonder-blocks-banner/banner.stories.tsx
+++ b/__docs__/wonder-blocks-banner/banner.stories.tsx
@@ -5,14 +5,8 @@ import magnifyingGlass from "@phosphor-icons/core/regular/magnifying-glass.svg";
 
 import Button from "@khanacademy/wonder-blocks-button";
 import {View} from "@khanacademy/wonder-blocks-core";
-import {Strut} from "@khanacademy/wonder-blocks-layout";
 import Link from "@khanacademy/wonder-blocks-link";
-import {
-    color,
-    spacing,
-    semanticColor,
-    sizing,
-} from "@khanacademy/wonder-blocks-tokens";
+import {semanticColor, sizing} from "@khanacademy/wonder-blocks-tokens";
 import {LabelSmall} from "@khanacademy/wonder-blocks-typography";
 import Banner from "@khanacademy/wonder-blocks-banner";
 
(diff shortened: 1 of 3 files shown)


__docs__/wonder-blocks-banner/banner.stories.tsx
```diff
@@ -643,13 +634,54 @@ export const RightToLeftMultiline: StoryComponentType = {
...
+};
+
 const styles = StyleSheet.create({
     example: {
         alignItems: "center",
         justifyContent: "center",
     },
     container: {
         width: "100%",
+        gap: sizing.size_160,
```
jaredly, line 684: nice

jaredly (approved):

__docs__/wonder-blocks-banner/banner.stories.tsx
```diff
@@ -135,8 +132,10 @@ export const Kinds: StoryComponentType = {
  * added around the floating banner so that it will not touch its outline.
  */
 export const Layouts: StoryComponentType = () => {
-    const borderStyle = {border: `2px solid ${color.fadedPurple24}`} as const;
-    const floatingContainerStyle = {padding: spacing.xSmall_8} as const;
+    const borderStyle = {
+        border: `2px solid ${semanticColor.core.border.inverse.strong}`,
```
jandrade: **nit:** Unrelated to your PR (don't bother changing it if you don't want to)

```suggestion
        border: `${border.width.medium} solid ${semanticColor.core.border.inverse.strong}`,
```

jandrade (approved):
LGTM! great improvement!! 🚢

PR #2665, opened 2025-06-11
Support theming in Pill

beaesguerra:
## Summary:
Update styling in the Pill component: 
- Use semantic color tokens instead of primitive color tokens
- Use new typography components
- Use sizing tokens
- Use global focus styling

Note: There is no Figma component for Pills with the TB theme. These changes are more for making sure that Pills are using the correct tokens and that it looks fine in TB. WB-1950 and WB-1951 are follow up tasks to remove the overlap between Badge and Pills:
  - Pill `info`, `success`, `warning`, `error` kinds will be removed in favour of `StatusBadge`
  - Pill `neutral` kind that are static will be removed in favour of `Badge`
    - existing `neutral` interactive Pills will need to be replaced (with either the new Filter component or a proper input element)
  - Remove support for `transparent` since this kind doesn't seem to be used much

Issue: WB-1876

## Test plan:
- Confirm that the Pill looks fine in Classic and TB themes
  - `?path=/story/packages-pill-testing-snapshots-pill--sizes&globals=;theme:thunderblocks`
  - `?path=/story/packages-pill-testing-snapshots-pill--sizes&globals=;theme:default` 
- Same for the sizes story
  - `?path=/story/packages-pill-testing-snapshots-pill--sizes&globals=;theme:thunderblocks` 
  -  `?path=/story/packages-pill-testing-snapshots-pill--sizes&globals=;theme:default`

| Classic | Thunderblocks |
| --- | --- |
| image | image | 
| image | image |


__docs__/wonder-blocks-pill/pill-testing-snapshots.stories.tsx
```diff
@@ -107,7 +107,7 @@ export const Sizes: Story = {
                                     kind={kind}
                                     key={index}
                                 >
-                                    {kind}, {props.size}
+                                    {`${kind}, ${props.size}`}
```
beaesguerra, line 110: This updates the `children` prop so it is treated as a string, so we can see the typography styles applied by the component based on the size

https://github.com/Khan/wonder-blocks/blob/345fc42a49ca27b4748e9ec2b9a99eed781ee11b/packages/wonder-blocks-pill/src/components/pill.tsx#L80-L107

packages/wonder-blocks-pill/src/components/pill.tsx
```diff
@@ -88,14 +88,22 @@ const PillInner = (props: {
     }
 
     if (size === "small") {
-        return <LabelXSmall>{props.children}</LabelXSmall>;
+        return (
+            <BodyText size="xsmall" tag="span">
```
beaesguerra, line 92: Using `span` tags so it is equivalent to what was being rendered before

packages/wonder-blocks-pill/src/components/pill.tsx
```diff
@@ -4,17 +4,17 @@ import type {StyleDeclaration} from "aphrodite";
...
-    LabelXSmall,
-} from "@khanacademy/wonder-blocks-typography";
+import {BodyText} from "@khanacademy/wonder-blocks-typography";
 import type {StyleType, AriaProps} from "@khanacademy/wonder-blocks-core";
 import type {ClickableRole} from "@khanacademy/wonder-blocks-clickable";
-import * as tokens from "@khanacademy/wonder-blocks-tokens";
+import {
+    semanticColor,
+    border,
+    color,
```
beaesguerra, line 13: I left a `NOTE` on the one remaining place we use the `color` tokens. It will get removed with the WB-1950 work!

packages/wonder-blocks-pill/src/components/pill.tsx
```diff
@@ -223,56 +228,57 @@ const _generateColorStyles = (clickable: boolean, kind: PillKind) => {
     }
 
     let backgroundColor;
+    let textColor;
 
     switch (kind) {
         case "accent":
-            backgroundColor = semanticColor.status.notice.foreground;
+            backgroundColor = semanticColor.core.background.instructive.default;
```
beaesguerra, line 235: One thing to note is we won't need to create specific themed component tokens for pills, using semantic colors is enough (especially since we'll be moving away from pills)

packages/wonder-blocks-pill/src/components/pill.tsx
```diff
@@ -104,8 +112,11 @@ const PillInner = (props: {
  *
  * **Note:** Before using the `Pill` component, please see if a component from the
  * [Badge Package](/?path=/docs/packages-badge-overview--docs&globals=theme:default)
- * can be used instead. For example, prefer using the `StatusBadge` component
- * instead of a `Pill` to indicate a status.
+ * can be used instead.
+ *
+ * For example, prefer using the `StatusBadge` component instead of a `Pill` to
+ * indicate a status. Or, use the `Badge` component instead of a `Pill` with
+ * `kind="neutral"`.
```
jandrade, line 119: **praise:** This is really good docs!

jandrade: **suggestion:** In fact, I think this would be great to include in the Pill docs in Storybook :)

jandrade (approved):
All the changes look good for me! Thanks for making this work in the short term 🚀

packages/wonder-blocks-pill/src/components/pill.tsx
beaesguerra: The component docs are included in the Storybook Pill docs! (Let me know if you were suggesting another way to document it!)

image

jandrade: oh I didn't see that this was the component description 🤦. This looks great!

PR #2666, opened 2025-06-11
Update styling of BodyText and Heading components to allow overrides

marcysutton:
## Summary:
The BodyText and Heading components used dynamic styles to match the fontSize, fontWeight, and lineHeight via component props. But the way the styles were inlined meant they couldn't be overridden.

This PR puts all of the style combinations into a styles.ts file but still matches them via props. There are also stories showing the styles being overridden (specifically the lineHeight).

Issue: WB-1978

## Test plan:
1. Review stories for style combinations and custom styles
   - BodyText: /?path=/docs/packages-typography-bodytext-new--docs&globals=theme:thunderblocks
   - Heading: /?path=/docs/packages-typography-heading-new--docs&globals=theme:thunderblocks

### DIFF
--- a/__docs__/wonder-blocks-typography/body-text.stories.tsx
+++ b/__docs__/wonder-blocks-typography/body-text.stories.tsx
@@ -15,7 +15,7 @@ import {
     LabelLarge,
     Body,
 } from "@khanacademy/wonder-blocks-typography";
-import {spacing, sizing} from "@khanacademy/wonder-blocks-tokens";
+import {font, spacing, sizing} from "@khanacademy/wonder-blocks-tokens";
 import {allModes} from "../../.storybook/modes";
 
 /**
@@ -233,6 +233,7 @@ const styles = StyleSheet.create({
     },
     customStyle: {
         fontSize: sizing.size_280,
+        fontWeight: font.weight.bold,
         lineHeight: sizing.size_320,
         marginTop: sizing.size_200,
     },
--- a/__docs__/wonder-blocks-typography/body-text.stories.tsx
+++ b/__docs__/wonder-blocks-typography/body-text.stories.tsx
@@ -15,7 +15,7 @@ import {
     LabelLarge,
     Body,
 } from "@khanacademy/wonder-blocks-typography";
-import {spacing} from "@khanacademy/wonder-blocks-tokens";
+import {spacing, sizing} from "@khanacademy/wonder-blocks-tokens";
 import {allModes} from "../../.storybook/modes";
 
 /**
@@ -198,14 +198,6 @@ export const ClassicConversionGuide = {
  * An example of overriding `BodyText` component's styling.
  */
 export const CustomStyling = {
-    parameters: {
-        chromatic: {
-            modes: {
-                default: allModes.themeDefault,
-                thunderblocks: allModes.themeThunderBlocks,
-            },
-        },
-    },
     render: () => (
         <View>
             <BodyText>
@@ -240,7 +232,8 @@ const styles = StyleSheet.create({
         margin: 0,
     },
     customStyle: {
-        lineHeight: "3rem",
-        marginTop: "2rem",
+        fontSize: sizing.size_280,
+        lineHeight: sizing.size_320,
+        marginTop: sizing.size_200,
     },
 });
--- a/__docs__/wonder-blocks-typography/body-text.stories.tsx
+++ b/__docs__/wonder-blocks-typography/body-text.stories.tsx
@@ -115,11 +115,11 @@ export const SizesAndWeights = {
     render: () => (
         <View style={styles.grid}>
             <View style={styles.row}>
-                <BodyText size="xsmall" weight="bold">
-                    xSmall bold
-                </BodyText>
                 <BodyText size="xsmall" weight="medium">
                     xSmall medium
+                </BodyText>
+                <BodyText size="xsmall" weight="bold">
+                    xSmall bold
                 </BodyText>
                 <div />
             </View>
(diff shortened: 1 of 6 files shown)


__docs__/wonder-blocks-typography/body-text.stories.tsx
```diff
@@ -199,4 +236,11 @@ const styles = StyleSheet.create({
     row: {
         display: "contents",
     },
+    classic: {
+        margin: 0,
+    },
+    customStyle: {
+        lineHeight: "3rem",
```
beaesguerra: Can we also customize `fontWeight` and `fontSize` in the example so we can have visual regression tests confirming those can override the default styles? (overriding fontSize was what I needed for theming the banner because classic and tb use different font sizes!

__docs__/wonder-blocks-typography/heading.stories.tsx
```diff
@@ -227,4 +270,8 @@ const styles = StyleSheet.create({
     row: {
         display: "contents",
     },
+    customStyle: {
```
beaesguerra, line 265: Same question about adding custom styles for other typography properties!

packages/wonder-blocks-typography/src/components/body-text.tsx
```diff
@@ -1,13 +1,25 @@
...
 import styles from "../util/styles";
 
 type Props = PropsFor<typeof Text> & {
     size?: "xsmall" | "small" | "medium";
     weight?: "medium" | "semi" | "bold";
 };
 
+// List style combinations for matching with props
+const styleMapping = {
+    "xsmall-medium": styles.BodyTextXSmallMedium,
```
beaesguerra: It is unfortunate that `medium` can be used for both size and weight (`BodyTextMediumMedium` 😅)!  What do you think about referencing medium as `MediumWeight` so it can be differentiated from size? Or something else! 

Ex: `styles.BodyTextXSmallMediumWeight` or something like that. `styles.BodyTextXSmallSizeMediumWeight` would be very clear, but long 😅 

I think the `styleMapping` keys are fine since it's internal to WB. I'm thinking more of the typography styles since that'll be something consumers can use too!

__docs__/wonder-blocks-typography/body-text.stories.tsx
```diff
@@ -161,29 +161,66 @@ export const ClassicConversionGuide = {
...
                 <BodyText>BodyText</BodyText>
             </View>
         </View>
     ),
 };
 
+/**
+ * An example of overriding `BodyText` component's styling.
+ */
+export const CustomStyling = {
```
beaesguerra, line 200: Do you think it's helpful to have Chromatic tests for this story for both TB and Classic themes? I was thinking one of them should be enough since overriding works the same way for both themes. Will leave it up to you!

__docs__/wonder-blocks-typography/heading.stories.tsx
```diff
@@ -134,6 +134,9 @@ export const SizesAndWeights = {
                 <Heading size="small" weight="semi">
                     Small semibold
                 </Heading>
+                <Heading size="small" weight="medium">
```
beaesguerra, line 137: I think the `AllVariants` component would be helpful in this example to show all the different size and weight combinations! (same for BodyText!)

__docs__/wonder-blocks-typography/heading.stories.tsx
```diff
@@ -213,10 +229,37 @@ export const ClassicConversionGuide = {
...
+    render: () => (
+        <View>
+            <Heading>
+                Text to show the default styling based on props. If we add more
+                text here, it will run on multiple lines.
+            </Heading>
+            <Heading style={styles.customStyle}>
+                A lot of text that runs on multiple lines, with custom styling.
+                We really like ice cream. What flavor is your favorite? That’s
+                not ice cream, it’s sorbet!
```
beaesguerra, line 245: 🍨 😄

packages/wonder-blocks-typography/src/components/body-text.tsx
marcysutton: Good call! I've also noticed that the styles object doesn't seem to have any typing. If I change those typography styles, the references in `BodyText` and `Heading` don't throw any errors. 🤔  A separate issue, but it also relates to usage!

VSCode showing BodyText component and Typography styles

__docs__/wonder-blocks-typography/heading.stories.tsx
marcysutton: If it's alright, I'll defer that work to https://khanacademy.atlassian.net/browse/WB-1957!

__docs__/wonder-blocks-typography/body-text.stories.tsx
```diff
@@ -199,4 +228,12 @@ const styles = StyleSheet.create({
     row: {
         display: "contents",
     },
+    classic: {
+        margin: 0,
+    },
+    customStyle: {
```
beaesguerra, line 234: suggestion: `fontWeight` would be another helpful property to include in the custom style so we can have visual regression tests confirming it can be overridden :) same for the heading custom style story! (not blocking!)

packages/wonder-blocks-typography/src/components/body-text.tsx
beaesguerra: Oooh good point! It would be worth exploring if we can explicitly create a type for the typography styles so that it can help catch invalid styles! (Can be for later!)

__docs__/wonder-blocks-typography/body-text.stories.tsx
marcysutton: Good idea, I added those to the PR!

PR #2667, opened 2025-06-12
[🔥AUDIT🔥] Tokens: Move some TB colors out of OG primitives

jandrade:
🖍 _This is an audit!_ 🖍

## Summary:

Moves some TB colors out of the OG primitives, so we ensure that these colors
are only available via `semanticColor`.

This is to prevent that consumers of the OG primitives can use these colors
directly, which is not the intention.

Issue: "none"

## Test plan:

N/A


packages/wonder-blocks-tokens/src/tokens/color.ts
```diff
@@ -58,6 +58,11 @@ const baseColors: ColorType = {
 const fadedColorWithWhite = (color: string, alpha: number) =>
     mix(fade(color, alpha), baseColors.white);
 
+/**
+ * OG color primitives used for the Classic experience.
+ *
+ * @deprecated Use `semanticColor` from `@khanacademy/wonder-blocks-tokens` instead.
```
jandrade, line 64: Screenshot 2025-06-12 at 11 49 47 AM

cc @beaesguerra

PR #2670, opened 2025-06-17
[WB-1982] Add learning category to semanticColor tokens

jandrade:
## Summary:

- Adds a new `learning` category in `semanticColor`.
- Updates `yellow_10` and `gray_20` TB primitives:
  - `yellow_10` is now `#966B00`
  - `gray_20` is now `#717279`
- Updates badge components to use the new learning tokens.
- Updates documentation to reflect these changes.
- Adds banners to mark `action` and `input` categories as deprecated.

Issue: https://khanacademy.atlassian.net/browse/WB-1982

## Test plan:

Navigate to the colors docs and verify that everything looks correct:

`/?path=/docs/foundations-using-color--docs#learning`

Verify that the snapshots for the badge components don't change.

Verify that the places that use `yellow_10` and `gray_20` (via semantic colors)
are updated correctly in Chromatic.


__docs__/foundations-color.mdx
```diff
@@ -75,7 +75,12 @@ how to use these colors depending on the context.
     />
 </View>
 
-### Action
+### Action (deprecated)
```
beaesguerra, line 78: Thanks for marking the deprecated tokens!

beaesguerra: (for later) I wonder if it would be worth re-organizing the token docs so that all the deprecated tokens are at the bottom!

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
```diff
@@ -657,4 +612,108 @@ export const semanticColor = mergeTheme(defaultSemanticColor, {
...
+                blue: color.cyan_10,
+                gold: color.yellow_10,
+                green: color.green_20,
+                gray: color.gray_10,
+                grayH: color.gray_05,
+                grayI: color.black_100,
+                purple: color.blue_05,
+                purpleD: color.blue_10,
+                pink: color.magenta_10,
+                red: color.red_20,
```
beaesguerra, line 627: These semantic colors are named after specific colors, which we've avoided for the other semantic tokens so the concept is the same for different themes! In this case, if we had a dark theme, would these semantic tokens named after colors still apply? For example, would the `learning.math.foreground.blue` still represent a blue color in all themes?

This can be something we evolve for later too!

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
```diff
@@ -657,4 +612,108 @@ export const semanticColor = mergeTheme(defaultSemanticColor, {
...
+                attempted: {
+                    subtle: color.yellow_50,
+                    strong: color.yellow_10,
+                },
+                complete: {
+                    strong: color.green_50,
+                },
+            },
+        },
+        shadow: {
```
beaesguerra, line 705: Out of curiosity, what does `shadow` refer to? `box-shadow`?

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -220,6 +175,119 @@ export const semanticColor = {
...
+                blue: "#20628F",
+                gold: "#946700",
+                green: "#447A53",
+                gray: "#5D5F66",
+                grayH: "#3B3D45",
+                grayI: "#21242C",
+                purple: "#594094",
+                purpleD: "#8351E8",
+                pink: "#B25071",
+                red: "#D92916",
```
beaesguerra: Do we want to use `thunderBlocksColor` for these values so it stays in sync with the TB primitive colors? (Similar to how the gems/due/streak tokens use the TB primitives!

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
jandrade: I thought about the same, but this is a very specific use case as those are the colors expected by MathJax: https://github.com/Khan/mathjax-renderer/blob/main/src/ts/tex-input-config.ts#L26-L35

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
jandrade: Yeah, It's for the color prop in `box-shadow`.

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
jandrade: We need to keep these colors to use the same colors we define in the classic experience.

https://github.com/Khan/mathjax-renderer/blob/main/src/ts/tex-input-config.ts#L26-L35

In fact, I added the wrong `blue` value, I'm going to fix it.

beaesguerra (approved):
Looks great! 🚀

PR #2676, opened 2025-06-20
[WB-1867.1] Modal setup: Remove Khanmigo theme from modal

jandrade:
## Summary:

- Removes the `khanmigo` theme from modal in favor of CSS vars.
- Uses `sizing` instead of `spacing`.

### Implementation plan

1. #2676
2. #2677
3. #2678

## Test plan:

correctly.


__docs__/wonder-blocks-modal/modal-dialog.stories.tsx
```diff
@@ -300,16 +294,7 @@ export const WithDarkPanel: StoryComponentType = {
...
-                                    actionStyles.inverse,
-                                    // NOTE: This is a workaround while we move
-                                    // this button to DAB and remove the
-                                    // `Button.khanmigo` theme.
-                                    {
-                                        background: "transparent",
-                                        ":hover": {background: "transparent"},
-                                    },
-                                ]}
+                                style={[actionStyles.inverse]}
```
jandrade, line 297: **note:** FYI This story is removed in the next PR.

packages/wonder-blocks-modal/src/theme/default.ts
jandrade, line 1: **note:** This only includes a conversion of React Context based theming to CSS vars. I'm going to refactor/simplify this file in the last stacked PR.

packages/wonder-blocks-modal/src/components/modal-content.tsx
```diff
@@ -48,9 +38,9 @@ ModalContent.isComponentOf = (instance: any): boolean => {
  * TODO(WB-1655): Change this to use the theme instead (inside themedStylesFn).
  * e.g. `[theme.breakpoints.small]: {...}`
  */
-const small = "@media (max-width: 767px)";
+const small = "@media (max-width: 767px)" as any;
```
beaesguerra, line 41: I wonder if we can use the media query tokens for this! https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-tokens/src/tokens/media-queries.ts

beaesguerra (approved):
Looks good! Thanks for removing the last khanmigo themed component!

packages/wonder-blocks-modal/src/components/modal-content.tsx
jandrade: I initially thought about this, but ended up not changing it as there's no current WB query equivalent. This is something that we surely need to address once we migrate our overall MQ strategy to use these official WB tokens.

PR #2677, opened 2025-06-20
[WB-1867.2] Remove `light` prop from Modal package.

jandrade:
## Summary:

Removes the `light` prop from modal as it is no longer needed.

### Implementation plan

1. #2676
2. #2677
3. #2678

## Test plan:

correctly.

Verify that the "Dark" stories are no longer present in the docs.

### DIFF
--- a/__docs__/wonder-blocks-modal/modal-dialog.stories.tsx
+++ b/__docs__/wonder-blocks-modal/modal-dialog.stories.tsx
@@ -253,66 +253,6 @@ export const WithLauncher: StoryComponentType = {
     },
 };
 
-/**
- * This example shows how a modal dialog/panel can be created with a dark
- * background. The `light` prop of the `<ModalPanel>` element is set to `false`
- * to create a dark background.
- */
-export const WithDarkPanel: StoryComponentType = {
-    render: (args) => (
-        <View style={styles.previewSizer}>
-            <View style={styles.modalPositioner}>
-                <ModalDialog
-                    {...args}
-                    aria-labelledby="modal-title-0"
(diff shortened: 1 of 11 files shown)


__docs__/wonder-blocks-modal/modal-dialog.stories.tsx
```diff
@@ -87,6 +86,7 @@ export default {
             modes: {
                 small: allModes.small,
                 large: allModes.large,
+                thunderblocks: allModes.themeThunderBlocks,
```
beaesguerra: Now that we include snapshots for thunderblocks too, there are many Chromatic snapshots for modals! Do you think there are things we could do to have fewer? Maybe by combining some stories! (not blocking, but a consideration for later, especially if/when we support other themes!)

beaesguerra (approved):
Thanks for simplifying things by removing the `light` prop! 🧹

__docs__/wonder-blocks-modal/modal-dialog.stories.tsx
jandrade: Agreed! I think that there are some duplicate snapshots between `ModalDialog`, `ModalHeader`, `ModalPanel` stories, I'll see which one of those make sense to remove. We can optimize even more later as you suggested :).

jandrade: Done here: https://github.com/Khan/wonder-blocks/pull/2677/commits/4e26d2c654b1ca18e178bd0c855d93620fbf8e8e

This should reduce the tests count by 16 (`761` ▶️ `745`)

PR #2678, opened 2025-06-20
[WB-1867.3] Add ThunderBlocks theme to modal

jandrade:
## Summary:

Last PR to add Thunderblocks theme to modal components.

- Adds ThunderBlocks theme to modal package.
- Updates typography to use BodyText and Heading.
- Removes Strut uses (including stories).

Figma: https://www.figma.com/design/EuFu0U7gqc1koXm8ZhlOLp/%E2%9A%A1%EF%B8%8F-Components?node-id=2213-141&m=dev

### Implementation plan

1. #2676
2. #2677
3. #2678

## Test plan:

correctly.

Make sure that the typography elements are looking as expected.

### DIFF
--- a/__docs__/wonder-blocks-modal/modal-dialog.stories.tsx
+++ b/__docs__/wonder-blocks-modal/modal-dialog.stories.tsx
@@ -3,9 +3,8 @@ import {StyleSheet} from "aphrodite";
 import type {Meta, StoryObj} from "@storybook/react";
 import Button from "@khanacademy/wonder-blocks-button";
 import {View} from "@khanacademy/wonder-blocks-core";
-import {Strut} from "@khanacademy/wonder-blocks-layout";
 import {BodyText, Heading} from "@khanacademy/wonder-blocks-typography";
-import {semanticColor, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {semanticColor, sizing} from "@khanacademy/wonder-blocks-tokens";
 
 import {
     ModalLauncher,
@@ -125,15 +124,14 @@ export const Default: StoryComponentType = {
                 >
                     <ModalPanel
                         content={
-                            <>
+                            <View style={{gap: sizing.size_240}}>
                                 <Heading size="xxlarge" id="modal-title-0">
                                     Modal Title
                                 </Heading>
-                                <Strut size={spacing.large_24} />
                                 <BodyText id="modal-desc-0">
                                     Here is some text in the modal.
                                 </BodyText>
-                            </>
+                            </View>
                         }
                     />
                 </ModalDialog>
@@ -190,15 +188,14 @@ export const WithAboveAndBelow: StoryComponentType = {
                     >
                         <ModalPanel
                             content={
-                                <>
+                                <View style={{gap: sizing.size_240}}>
                                     <Heading size="xxlarge" id="modal-title-2">
                                         Modal Title
                                     </Heading>
-                                    <Strut size={spacing.large_24} />
                                     <BodyText>
                                         Here is some text in the modal.
                                     </BodyText>
-                                </>
+                                </View>
                             }
                         />
                     </ModalDialog>
@@ -228,13 +225,12 @@ export const WithLauncher: StoryComponentType = {
             >
                 <ModalPanel
                     content={
-                        <>
+                        <View style={{gap: sizing.size_240}}>
                             <Heading size="xxlarge" id="modal-title-3">
                                 Modal Title
                             </Heading>
-                            <Strut size={spacing.large_24} />
                             <BodyText>Here is some text in the modal.</BodyText>
-                        </>
+                        </View>
                     }
                 />
             </ModalDialog>
--- a/__docs__/wonder-blocks-modal/modal-dialog.stories.tsx
+++ b/__docs__/wonder-blocks-modal/modal-dialog.stories.tsx
@@ -4,7 +4,7 @@ import type {Meta, StoryObj} from "@storybook/react";
 import Button from "@khanacademy/wonder-blocks-button";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {Strut} from "@khanacademy/wonder-blocks-layout";
-import {Body, Title} from "@khanacademy/wonder-blocks-typography";
+import {BodyText, Heading} from "@khanacademy/wonder-blocks-typography";
 import {semanticColor, spacing} from "@khanacademy/wonder-blocks-tokens";
 
 import {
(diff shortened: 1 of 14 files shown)


.storybook/preview.tsx
```diff
@@ -132,6 +133,13 @@ const withThemeSwitcher: Decorator = (
     Story,
     {globals: {theme}, parameters: {enableRenderStateRootDecorator}},
 ) => {
+    React.useEffect(() => {
+        if (theme) {
```
beaesguerra, line 137: Will we be removing the `ThemeSwitcher` related things later on? The components that were for the original theming mechanism!

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
```diff
@@ -119,6 +119,7 @@ const core = {
     },
     shadow: {
         transparent: transparent,
+        instructive: `color-mix(in srgb, ${color.blue_10} 20%, ${transparent})`,
```
beaesguerra: Will this token also be added in Figma? The color tokens in the handshake file doesn't include this token! Though it seems the transparent token for TB has a blueish tint? 

https://www.figma.com/design/e9qdyiDUBZ3rDabP9S7ulO/%E2%9A%A1%EF%B8%8F-Handshake?node-id=3040-15249&m=dev

beaesguerra (commented):
Left some questions in Chromatic and in the PR! 😄

.storybook/preview.tsx
jandrade: I don't think so. We'll need it just in case someone needs to override the theme down in the tree for any reason.

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
jandrade: oh good catch! turns out that I was using the incorrect tokens and incorrect values. I fixed that by moving it to `core.shadow.transparent`, and using the correct value in the modal component-level `dialog.shadow.default` token. It should now match the designs as expected. Let me know what you think.

__docs__/wonder-blocks-modal/modal-dialog.stories.tsx
```diff
@@ -121,13 +120,14 @@ export const Default: StoryComponentType = {
...
                         content={
-                            <>
-                                <Title id="modal-title-0">Modal Title</Title>
-                                <Strut size={spacing.large_24} />
-                                <Body id="modal-desc-0">
+                            <View style={{gap: sizing.size_240}}>
+                                <Heading size="xxlarge" id="modal-title-0">
+                                    Modal Title
+                                </Heading>
+                                <BodyText id="modal-desc-0">
```
marcysutton, line 127: Such a clean switchover! Love to see it!

__docs__/wonder-blocks-modal/modal-footer.stories.tsx
```diff
@@ -43,12 +42,11 @@ const customViewports = {
...
     <>
-        <Body>
+        <BodyText>
             {`Let's make this body content long in order
 to test scroll overflow.`}
-        </Body>
-        <br />
-        <Body>
+        </BodyText>
+        <BodyText>
```
marcysutton, line 49: Should we add any margins or a gap here to replace the `<br />` tags?

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
```diff
@@ -118,7 +118,7 @@ const core = {
         },
     },
     shadow: {
-        transparent: transparent,
+        transparent: `color-mix(in srgb, ${color.blue_10} 20%, ${transparent})`,
```
marcysutton: What does this line do? Could you add a comment?

.changeset/gold-yaks-grab.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": minor
+---
+
+Add core.shadow token. Export modal tokens.
```
beaesguerra: Do we need to update the changeset since we don't add a new shadow token anymore?

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
beaesguerra: +1! I'm still not sure when we should use color utilities in code 😄 

I see in Figma it's set to `#252368` `20%`, though `color.blue_10` is `#363498`

image

beaesguerra (commented):
Thanks for making these updates! Left some questions in Chromatic around letting custom styles override the default title styles!

.changeset/gold-yaks-grab.md
jandrade: This was to make sure that I would include the CSS vars generated for WB Modal for all the 3 stacked PRs. But turns out that I didn't land this last PR with the other two, so I messed up the release (and probably this is not needed anymore).

__docs__/wonder-blocks-modal/modal-footer.stories.tsx
jandrade: that's being added to the parent container (in all the stories).

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
jandrade: ohhh it should be `color_blue_05` instead.

`color-mix` is the official CSS way of generating alpha colors. I couldn't use the WB `fade()` function as it doesn't work with CSS variables (same happens when concatenating WB tokens with the alpha channel).

BEFORE:

```
const fadedBlue20 = fade(color.blue, 0.2) // e.g. fade(#00f, 0.2) = rgba(0, 0, 255, 0.2)
```

This doesn't work anymore because the semantic color variable gets converted to a reference to a CSS variable, so we have to use CSS functions as a workaround.

I'll add a note there.

jandrade: Added a comment explaining this

.changeset/gold-yaks-grab.md
beaesguerra: If we keep the changeset, could we remove the `Add core.shadow token` part since we no longer add the `instructive` shadow token?

beaesguerra (approved):
Looks good, nice work!

PR #2681, opened 2025-06-24
Fix variable font rendering in Firefox

marcysutton:
## Summary:
The font-face definition for Plus Jakarta Sans wasn't working for variable weights in Firefox. This change ensures all the supported font-weights will render properly.

Issue: WB-1996

## Test plan:
1. Test WB Site in Firefox: /?path=/docs/packages-typography-bodytext-new--docs&globals=theme:thunderblocks
2. Compare rendered font weights to live site: https://khan.github.io/wonder-blocks/?path=/docs/packages-typography-bodytext-new--docs


marcysutton:
> this means that you'll have to apply a similar fix in frontend right?

@jandrade yep! This was a quickie for the WB site, and it's a nice validation of the fix. I figure I can make a similar update in `frontend` when I integrate the `BodyText` and `Heading` changes.

PR #2687, opened 2025-06-26
Ensure tablist doesn't disappear

nedredmond:
In certain situations, the tabs component can be vertically compressed, which causes the tablist to disappear since it's flex-shrinking.

Adding `flexShrink: 0` prevents that issue.

https://github.com/user-attachments/assets/02f2bcdc-a19b-406e-9c56-fe9793c9831e


beaesguerra (approved):
Looks good to me! Thanks for the fix!

PR #2692, opened 2025-06-27
Reverts CheckRadio changes to clear release queue

marcysutton:
## Summary:
Checkbox and Radio theming was landed to main before other changes could go out. This PR reverts the merge commit so we can release them when needed.

Issue: WB-XXXX

## Test plan:


jandrade (approved):
This should be good enough! thanks Marcy for being comprehensive with this, really appreciate it 🫶

PR #2689, opened 2025-06-26
Include readonly and error icon in fields

beaesguerra:
## Summary:

- Form
  - Add an error icon and read only icon for TextField and TextArea in the TB theme. It is not shown in the classic theme
  - In order to support this new icon in `TextField`, we now wrap the `input` element with a `div`. This changes the DOM structure, we will need to test to make sure there are no layout changes in `frontend`/`perseus` because of this change. `TextArea` already had a wrapping div so we were able to add styles to that existing `div`. 
  - TextField: A `styles` prop was also added so that this newly introduced wrapping `div` can be styled (most likely for layout purposes)
 
- SearchField
  - Since the TextField in TB has an error icon now, we shift the `dismiss` icon button in the SearchField so that it doesn't overlap with the error icon or field value 

- Dropdown Combobox
  - Minor UI modifications since it uses TextField to address visual diffs that were original found in Chromatic

image

Issue: WB-1863

Figma designs: https://www.figma.com/design/EuFu0U7gqc1koXm8ZhlOLp/%E2%9A%A1%EF%B8%8F-Components?node-id=5302-354&m=dev

## Test plan:

- Review TextField and TextArea error and readonly states. Confirm icons are shown in TB, but not in OG
- Combobox, and SearchField stories to make sure things look as expected since they use TextField
- Test with the snapshot package in `frontend` and `perseus` to check for visual layout changes in the TextField component

PR #2693, opened 2025-06-27
LabeledField: Show the error icon for TB theme

beaesguerra:
## Summary:

We now show the error icon in the `LabeledField` error message in the Thunderblocks theme too.

We were going to move the error icon and add a readonly icon inside the field in TB in PR https://github.com/Khan/wonder-blocks/pull/2689, however, this would require a change in the TextField DOM structure so that we can place icons in the field (we would need to wrap the underlying `input` element). This is a more involved change since this would mean testing and updating many instances of TextField in the other repos, to make sure layouts and custom styles don't break. 

Design has agreed with keeping the error icon with the message! Adding to the LabeledField component is more flexible, and will also be more consistent with the different form fields (for example, we wouldn't also need to add error/readonly icons to dropdown components too). Design is also working on updating some of the TB designs, which can come later in code once finalized

Issue: WB-1863

## Test plan:

- Confirm that LabeledField in the TB theme has the error icon
- Confirm error and readonly states look as expected (without the icon in the field)

I'll do some integration testing with `frontend` still to make sure there are no unexpected changes!

### DIFF
--- a/.changeset/heavy-dolphins-lie.md
+++ b/.changeset/heavy-dolphins-lie.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": patch
+---
+
+Update LabeledField component css variables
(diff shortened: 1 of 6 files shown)


marcysutton (approved):
This looks good to me!

PR #2697, opened 2025-06-27
Add shared feedback tokens to semanticColor

marcysutton:
## Summary:
There are some shared feedback tokens for Toasts to add to semanticColor. This PR introduces those colors for use in Toast components.

Issue: WB-1986

## Test plan:
1, Review foundations-color to see how colors are structured
2. Review naming and structure to ensure it aligns with Figma

### DIFF
--- a/__docs__/foundations-color.mdx
+++ b/__docs__/foundations-color.mdx
@@ -204,6 +204,22 @@ and include an icon or label to provide more context and clarity."
     group="feedback.critical.subtle"
 />
 
+#### Shared Actions
+
+##### Subdued
+
+<ColorGroup
+    colors={semanticColor.feedback.shared.actions.subdued}
+    group="feedback.shared.actions.subdued"
+/>
+
+##### Strong
+
+<ColorGroup
+    colors={semanticColor.feedback.shared.actions.strong}
+    group="feedback.shared.actions.strong"
+/>
+
 ### Focus
 
 For focus states on interactive elements. This is meant to be used globally for
--- a/__docs__/foundations-color.mdx
+++ b/__docs__/foundations-color.mdx
@@ -204,20 +204,20 @@ and include an icon or label to provide more context and clarity."
     group="feedback.critical.subtle"
 />
 
-#### Shared Actions
-
-##### Subdued
-
-<ColorGroup
-    colors={semanticColor.feedback.shared.actions.subdued}
-    group="feedback.shared.actions.subdued"
-/>
+#### Shared
 
 ##### Strong
 
 <ColorGroup
-    colors={semanticColor.feedback.shared.actions.strong}
-    group="feedback.shared.actions.strong"
+    colors={semanticColor.feedback.shared.strong}
+    group="feedback.shared.strong"
+/>
+
+##### Actions
+
+<ColorGroup
+    colors={semanticColor.feedback.shared.actions}
+    group="feedback.shared.actions"
 />
 
 ### Focus
--- a/__docs__/foundations-color.mdx
+++ b/__docs__/foundations-color.mdx
@@ -177,6 +177,13 @@ and include an icon or label to provide more context and clarity."
     group="feedback.info.subtle"
 />
 
+#### Strong
+
+<ColorGroup
+    colors={semanticColor.feedback.info.strong}
+    group="feedback.info.strong"
+/>
+
 #### Success
 
 ##### Subtle
@@ -205,13 +212,6 @@ and include an icon or label to provide more context and clarity."
 />
 
 #### Shared
-
-##### Strong
-
-<ColorGroup
-    colors={semanticColor.feedback.shared.strong}
-    group="feedback.shared.strong"
-/>
 
 ##### Actions
 
--- a/__docs__/foundations-color.mdx
+++ b/__docs__/foundations-color.mdx
@@ -177,7 +177,7 @@ and include an icon or label to provide more context and clarity."
     group="feedback.info.subtle"
 />
 
-#### Strong
+##### Strong
 
 <ColorGroup
     colors={semanticColor.feedback.info.strong}
@@ -193,6 +193,13 @@ and include an icon or label to provide more context and clarity."
     group="feedback.success.subtle"
 />
 
+##### Strong
+
+<ColorGroup
+    colors={semanticColor.feedback.success.strong}
+    group="feedback.success.strong"
+/>
+
 #### Warning
 
 ##### Subtle
@@ -200,6 +207,13 @@ and include an icon or label to provide more context and clarity."
 <ColorGroup
     colors={semanticColor.feedback.warning.subtle}
     group="feedback.warning.subtle"
+/>
+
+##### Strong
+
+<ColorGroup
+    colors={semanticColor.feedback.warning.strong}
+    group="feedback.warning.strong"
 />
 
 #### Critical
@@ -210,14 +224,11 @@ and include an icon or label to provide more context and clarity."
     colors={semanticColor.feedback.critical.subtle}
     group="feedback.critical.subtle"
 />
-
-#### Shared
-
-##### Actions
+##### Strong
 
 <ColorGroup
-    colors={semanticColor.feedback.shared.actions}
-    group="feedback.shared.actions"
+    colors={semanticColor.feedback.critical.strong}
+    group="feedback.critical.strong"
 />
 
 ### Focus
--- a/packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
+++ b/packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
@@ -817,5 +817,15 @@ export const semanticColor = {
                 text: core.foreground.critical.strong,
             },
         },
+        shared: {
+            actions: {
+                subdued: {
+                    foreground: core.foreground.instructive.default,
+                },
+                strong: {
+                    foreground: core.foreground.inverse.strong,
+                },
+            },
+        },
     },
 };
--- a/packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
+++ b/packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
@@ -157,6 +157,12 @@ const surface = {
     overlay: color.offBlack64,
 };
 
+const sharedFeedbackStrongTokens = {
+    background: core.background.neutral.strong,
+    border: core.border.neutral.strong,
+    text: core.foreground.inverse.strong,
+};
+
 /**
  * TODO(WB-1941): Remove text once we have migrated to the new core.foreground
  * tokens.
@@ -793,9 +799,7 @@ export const semanticColor = {
                 text: core.foreground.instructive.strong,
             },
             strong: {
-                background: core.background.neutral.strong,
-                border: core.border.neutral.strong,
-                text: core.foreground.inverse.strong,
+                ...sharedFeedbackStrongTokens,
                 icon: core.foreground.instructive.subtle,
             },
         },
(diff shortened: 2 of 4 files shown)


packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -817,5 +817,16 @@ export const semanticColor = {
                 text: core.foreground.critical.strong,
             },
         },
+        shared: {
+            strong: {
+                background: core.background.neutral.strong,
+                border: core.border.neutral.strong,
+                text: core.foreground.inverse.strong,
+            },
```
beaesguerra: I was under the impression that the `shared` category in Figma was a workaround for sharing Figma variables! So I thought the `strong` tokens would get embedded into the rest of the object. For example: 
```
 feedback: {
        info: {
            subtle: {
                background: core.background.instructive.subtle,
                border: core.border.instructive.subtle,
                icon: core.foreground.instructive.default,
                text: core.foreground.instructive.strong,
            },
            // -- start of new tokens --
            strong: {
                background: core.background.neutral.strong, // <-- this is from shared category
                border: core.border.neutral.strong, // <-- this is from shared category
                text: core.foreground.inverse.strong, // <-- this is from shared category
                icon: core.foreground.instructive.subtle, // <-- this is under Feedback > Info > Strong > Foreground
            },
            // -- end of new tokens --
        },
        warning: {...}, // etc
}
```
Note that the `icon` value changes based on the kind of feedback (which we'll need to handle either way!) 

This would also make it easy for consumers to use either subtle or strong feedback tokens since they have the same structure!

What do you think of this approach?

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -817,5 +817,16 @@ export const semanticColor = {
...
+            actions: {
+                subdued: core.foreground.instructive.default,
+                strong: core.foreground.inverse.strong,
+            },
```
beaesguerra: From what I understand, the `actions` tokens are used for [the links in feedback components ](https://www.figma.com/design/EuFu0U7gqc1koXm8ZhlOLp/%E2%9A%A1%EF%B8%8F-Components?node-id=6875-2059&m=dev) (Banners and Toasts). I think it makes sense to keep `actions` in the `shared` category!

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
marcysutton: Makes sense! I hadn't seen that context. It wasn't exactly clear where those should go!

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -817,5 +823,11 @@ export const semanticColor = {
                 text: core.foreground.critical.strong,
             },
         },
+        shared: {
+            actions: {
+                subdued: core.foreground.instructive.default,
+                strong: core.foreground.inverse.strong,
+            },
```
jandrade: I feel that this category is redundant for two reasons:

1. These actions should be WB `Link` instances, which already have their own set of predefined tokens and styles.
2. We already have `semanticColor.link` to cover these cases, and if we need to support the `strong` case, we can use the `core` tokens directly.

As I can see in Figma, those tokens are more meant for internal use, and I think that it would be better to keep them that way.

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -792,6 +792,12 @@ export const semanticColor = {
                 icon: core.foreground.instructive.default,
                 text: core.foreground.instructive.strong,
             },
+            strong: {
+                background: core.background.neutral.strong,
+                border: core.border.neutral.strong,
+                text: core.foreground.inverse.strong,
+                icon: core.foreground.instructive.subtle,
+            },
```
beaesguerra: We'll want this for all the feedback types (`success`, `warning`, `critical`)! 
Screenshot 2025-07-02 at 3 37 26 PM

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
marcysutton: What is the default Link color, though -- isn't it blue? The toast component has a dark background with white text, which is atypical from the regular page colors (excluding a future dark mode). Should we skip having a feedback actions category and just use core tokens directly? I'm trying to figure out the right level of abstraction.

Screenshot 2025-07-03 at 1 24 20 PM

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
marcysutton: Thanks for this! I hadn't seen those icon entries. I see why the shared colors are useful though, as the rest of the tokens are copied from `feedback.info.strong` after your last suggestion. The icons are the only differences. Is this the structure you have in mind?

```
feedback: {
        info: {
            subtle: {
                background: core.background.instructive.subtle,
                border: core.border.instructive.subtle,
                icon: core.foreground.instructive.default,
                text: core.foreground.instructive.strong,
            },
            strong: {
                background: core.background.neutral.strong,
                border: core.border.neutral.strong,
                text: core.foreground.inverse.strong,
                icon: core.foreground.instructive.subtle,
            },
        },
        success: {
            subtle: {
                background: core.background.success.subtle,
                border: core.border.success.subtle,
                icon: core.foreground.success.default,
                text: core.foreground.success.strong,
            },
            strong: {
                background: core.background.neutral.strong,
                border: core.border.neutral.strong,
                text: core.foreground.inverse.strong,
                icon: core.foreground.success.subtle,
            },
        },
        warning: {
            subtle: {
                background: core.background.warning.subtle,
                border: core.border.warning.default,
                icon: core.foreground.warning.default,
                text: core.foreground.warning.strong,
            },
            strong: {
                background: core.background.neutral.strong,
                border: core.border.neutral.strong,
                text: core.foreground.inverse.strong,
                icon: core.foreground.warning.subtle,
            },
        },
        critical: {
            subtle: {
                background: core.background.critical.subtle,
                border: core.border.critical.subtle,
                icon: core.foreground.critical.default,
                text: core.foreground.critical.strong,
            },
            strong: {
                background: core.background.neutral.strong,
                border: core.border.neutral.strong,
                text: core.foreground.inverse.strong,
                icon: core.foreground.critical.subtle,
            },
        },```

File: packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts (hunk 3 above)
Comment (jandrade) [reply packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts]: > Should we skip having a feedback actions category and just use core tokens directly?

I think this should be handled directly by the `Link` component internally (using the tokens defined for links). In the inverse case, this could probably work: https://khan.github.io/wonder-blocks/?path=/docs/packages-styles-action-styles--docs

I feel that these `actions` tokens are not covering `hover and press` states (only accounts for `rest/default`. Another point is that these colors are already abstracted out by the existing `link` semanticColor tokens, and in the inverse case, it can be covered by using `<Link style={actionStyles.inverse} />` (inverse should cover `darkMode` via a new `semantic-color-dark.ts` file in the future).

File: packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts (hunk 4 above)
Comment (beaesguerra) [reply packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts]: Yes! Seeing it now though, if we wanted to consolidate things, we could create an internal shared object for the `strong` styles with `background`, `border`, and `text`, and then spread it in each of the feedback types. I think the important part is that consumers can access the subtle and strong values the same way! (ex: `feedback.info.subtle.background` and `feedback.info.strong.background`) 

```tsx
const sharedFeedbackStrongTokens = { // <-- doesn't need to be exported
  background: core.background.neutral.strong,
  border: core.border.neutral.strong,
  text: core.foreground.inverse.strong,
};

export const semanticColor = {
  feedback: {
    info: {
      subtle: { ... }, // the existing subtle tokens
      strong: {
        ...sharedFeedbackStrongTokens,
       icon: core.foreground.instructive.subtle,
      },
    }
  },
  // ...
}
```

marcysutton: Implemented! The `sharedFeedbackStrongTokens` object is a bit removed from its usage in the file, as it had to be hoisted outside of the `semanticColor` object. 

I wonder if it would make sense to assemble `semanticColor` from smaller objects as it grows in size. Then non-exported consts like this could live closer to their exported part of the `semanticColor` structure, if that makes sense.

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
marcysutton: Okay, I removed that section. I'll play with the snapshot release from this PR in `frontend` and see if I can get it to work with `Link` styles! Good call-out about the pseudo states.

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
beaesguerra: That makes sense to me! We do this for some of the other semantic tokens like `core`, or `surface`. I think the only gotcha is we have to remember to include it in the exported object, otherwise the default values will be used instead!

__docs__/foundations-color.mdx
```diff
@@ -204,6 +211,15 @@ and include an icon or label to provide more context and clarity."
     group="feedback.critical.subtle"
 />
 
+#### Shared
```
beaesguerra: I think we won't need these docs anymore since we'll use the default link styling for actions!

__docs__/foundations-color.mdx
```diff
@@ -177,6 +177,13 @@ and include an icon or label to provide more context and clarity."
     group="feedback.info.subtle"
 />
 
+#### Strong
+
+<ColorGroup
+    colors={semanticColor.feedback.info.strong}
+    group="feedback.info.strong"
+/>
```
beaesguerra, line 185: We can add the color groups for the other strong feedback types as well for documentation!

marcysutton: good call! Thank you!

beaesguerra (approved):
Looks good to me! 🚀

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
jandrade: Sounds great, thanks :)

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
jandrade, line 1: @marcysutton I think you'll need to make the same change in `semantic-color-thunderblocks.ts` so the right TB color primitives get propagated correctly.

marcysutton: That's a good call out! The `feedback` values look identical between the two `semanticColor` theme files. I was under the impression that only unique values needed to be redefined in the thunderblocks theme. Is that not the case for this file?

jandrade: Unfortunately that's not the case. The object structure gets copied, but in order to reflect the new colors, you'll need to explicitly override those tokens in that category, otherwise it will use the OG colors even for TB (or any other theme that doesn't override them).

marcysutton: Thanks for the info. I added them to the TB theme!

jandrade (approved):
LGTM! 🚀

PR #2699, opened 2025-06-27
Use new typography styles for TextField and TextArea

beaesguerra:
## Summary:

- Using the new typography styles for TextField and TextArea
- Also updated the shared typography styles so it included the base Heading/BodyText styles so that the different typography styles can be used on it's own
  - Also added stories so we can have visual regression tests for the typography styles we export 

Issue: WB-1863

## Test plan:

Confirm the styling of the typography in TextField and TextArea

Confirm the styling in the new typography styles story for Classic and TB themes (`?path=/story/packages-typography--typography-styles&globals=;theme:thunderblocks`)

### DIFF
--- a/__docs__/wonder-blocks-typography/typography.stories.tsx
+++ b/__docs__/wonder-blocks-typography/typography.stories.tsx
@@ -2,9 +2,9 @@ import * as React from "react";
 import {css, StyleSheet} from "aphrodite";
 import type {Meta, StoryObj} from "@storybook/react";
 
-import {View} from "@khanacademy/wonder-blocks-core";
+import {addStyle, View} from "@khanacademy/wonder-blocks-core";
 import {OptionItem, SingleSelect} from "@khanacademy/wonder-blocks-dropdown";
-import {color, spacing} from "@khanacademy/wonder-blocks-tokens";
+import {color, sizing, spacing} from "@khanacademy/wonder-blocks-tokens";
 import {
     Title,
     Heading,
@@ -24,6 +24,7 @@ import {
     Tagline,
     Caption,
     Footnote,
+    styles as typographyStyles,
 } from "@khanacademy/wonder-blocks-typography";
 import packageConfig from "../../packages/wonder-blocks-typography/package.json";
 
@@ -69,6 +70,8 @@ export default {
     },
     argTypes: TypographyArgTypes,
 } as Meta<typeof ComponentInfo>;
+
+const StyledDiv = addStyle("div");
 
 export const ControlProps: StoryObj<typeof Title> = {
     render: (args) => <Title {...args} />,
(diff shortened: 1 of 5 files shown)


packages/wonder-blocks-typography/src/util/styles.ts
```diff
@@ -6,6 +6,21 @@ const common = {
     display: "block",
 } as const;
 
+const Heading = {
```
beaesguerra, line 9: Moved these to the top so that they can be referenced in the other styles!

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
```diff
@@ -726,7 +726,7 @@ export const semanticColor = mergeTheme(defaultSemanticColor, {
         default: {
             border: core.border.neutral.default,
             background: surface.primary,
-            foreground: core.foreground.neutral.default,
+            foreground: core.foreground.neutral.strong,
```
beaesguerra, line 729: I can't remember why I changed it to `neutral.default` in a previous PR 😅  `neutral.strong` is what's used for the input field value! 
image

packages/wonder-blocks-typography/src/util/styles.ts
```diff
@@ -27,130 +42,142 @@ const styles: StyleDeclaration = StyleSheet.create({
...
     },
-    Heading: {
-        ...common,
-        fontFamily: font.family.sans,
-        // weight and size are matched by props, using the combinations below
-        // lineHeight is determined by fontSize on REM scale
-    },
+    Heading,
     HeadingSmallBoldWeight: {
+        ...Heading,
```
beaesguerra, line 47: When I first swapped to use the typography styles in the fields, I noticed the font family wasn't getting applied. This is why I added the base styles to the different `Heading*`/`BodyText*` styles!

Let me know @marcysutton if you have any concerns with this!

marcysutton: Makes sense! Thanks Bea!

__docs__/wonder-blocks-typography/typography.stories.tsx
```diff
@@ -432,3 +435,32 @@ LineHeight.parameters = {
...
 };
+
+/**
+ * The following shows the typography styles available.
+ *
+ * ```
+ *     import { styles as typographyStyles } from "@khanacademy/wonder-blocks-typography";
+ * ```
+ */
+export const TypographyStyles: StoryObj = {
```
marcysutton, line 446: Very cool! Thanks for adding this.

marcysutton (approved):
Thanks for the Typography changes! I added a few questions in Chromatic. Otherwise it looks good to go!

PR #2703, opened 2025-06-30
Change core background success strong to green 20

beaesguerra:
## Summary:

- Updating core background success strong token value to match updated Figma token value for TB. 
- Also updating learning.foreground.progress.complete strong to green 10

Issue: WB-2010

## Test plan:
- Confirm green 20 is used for the core background success strong token
  - Note: in the docs `?path=/docs/foundations-using-color--docs`, there is a bug with the hex code for the colors. If the color swatch is inspected, it should be using the correct color (`#3C6D4A`)

image

image

### DIFF
--- a/packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
+++ b/packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
@@ -701,7 +701,7 @@ export const semanticColor = mergeTheme(defaultSemanticColor, {
                     strong: color.yellow_10,
                 },
                 complete: {
-                    strong: color.green_50,
+                    strong: color.green_10,
                 },
             },
         },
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
```diff
@@ -701,7 +701,7 @@ export const semanticColor = mergeTheme(defaultSemanticColor, {
                     strong: color.yellow_10,
                 },
                 complete: {
-                    strong: color.green_50,
+                    strong: color.green_10,
```
beaesguerra, line 704: Updated the learning complete token too! @jandrade

jandrade: Thanks!

jandrade (approved):

PR #2709, opened 2025-07-08
Add `readOnlyMessage` prop for `LabeledField`

beaesguerra:
## Summary:

set the readOnly prop on the field if the message is set. This is similar to setting the error prop when there is an error message
[wb-2011-labeled-field] add custom styles readonly case
[wb-2011-labeled-field] add jsx example for read only message prop
[wb-2011-labeled-field] add coverage for id and test id props for readonly section
[wb-2011-labeled-field] allow both read only and error message text to be shown
[wb-2011-labeled-field] support custom styling for readonly message
[wb-2011-labeled-field] add tests for readonly
[wb-2011-labeled-field] rename component token so it can be used for helper text
[wb-2011-labeled-field] testing things with sr
[wb-2011-labeled-field] fix disabled styling on description
[wb-2011-labeled-field] fix wrapping in helper text read only
[wb-2011-labeled-field] set aria-describedby on the field to the id of the readOnlyMessageId
[wb-2011-labeled-field] handle if both errorMessage and readOnlyMessage are set
[wb-2011-labeled-field] add more testing stories
[wb-2011-labeled-field] theme the gap between helper text and icon
[wb-2011-labeled-field] Add readOnlyMessage prop

Issue: WB-2011

## Test plan:

### DIFF
--- a/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
+++ b/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
@@ -382,7 +382,11 @@ export const Fields: StoryComponentType = {
                 <Heading>Disabled</Heading>
                 <AllFields {...args} disabled />
                 <Heading>Read Only</Heading>
-                <AllFields {...args} readOnly={true} textValue={"Value"} />
+                <AllFields
+                    {...args}
+                    textValue={"Value"}
+                    readOnlyMessage="Read Only"
+                />
             </View>
         );
     },
--- a/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
+++ b/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
@@ -372,7 +372,7 @@ export const Fields: StoryComponentType = {
                 <AllFields
                     {...args}
                     textValue={"Value"}
-                    readOnlyMessage="Read Only"
+                    readOnlyMessage="Message about why it is read only"
                 />
             </View>
         );
--- a/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
+++ b/__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
@@ -76,7 +76,6 @@ const AllFields = (
         shouldValidateInStory?: boolean;
         showSubmitButtonInStory?: boolean;
         disabled?: boolean;
-        readOnly?: boolean;
         textValue?: string;
     },
 ) => {
@@ -84,7 +83,6 @@ const AllFields = (
         shouldValidateInStory,
         showSubmitButtonInStory,
         disabled,
-        readOnly,
         textValue,
         ...args
     } = storyArgs;
@@ -245,7 +243,6 @@ const AllFields = (
                         }
                         instantValidation={false}
                         disabled={disabled}
-                        readOnly={readOnly}
                     />
                 }
             />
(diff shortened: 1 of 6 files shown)


.changeset/chilly-adults-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": patch
```
beaesguerra, line 2: I renamed and reorganized some of the component theme tokens, should this be a breaking change? I used `patch` here since the component css variables aren't part of the public api, but curious what others think!

packages/wonder-blocks-labeled-field/src/theme/thunderblocks.ts
```diff
@@ -9,7 +9,7 @@ export default mergeTheme(defaultTheme, {
                 labelWithDescription: sizing.size_100,
                 labelWithNoDescription: sizing.size_100,
                 description: sizing.size_100,
-                errorSectionWithContent: sizing.size_100,
```
beaesguerra, line 12: Renamed this so we can use the same token for the read only message too! I'm thinking this might need to get restructured after adding support for the other helper text (will see in the next PR!)

__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
```diff
@@ -76,15 +76,13 @@ const AllFields = (
         shouldValidateInStory?: boolean;
         showSubmitButtonInStory?: boolean;
         disabled?: boolean;
-        readOnly?: boolean;
```
beaesguerra, line 79: This was originally used for setting the readOnly prop on the field. LabeledField sets the readOnly prop on the field now so this isn't needed!

__docs__/wonder-blocks-labeled-field/labeled-field.stories.tsx
```diff
@@ -265,86 +262,76 @@ const AllFields = (
                         }
                         instantValidation={false}
                         disabled={disabled}
-                        readOnly={readOnly}
                     />
                 }
             />
 
-            {/* SingleSelect does not support readonly state */}
```
beaesguerra, line 273: Decided to include SingleSelect and MultiSelect for the readonly state so we can make sure all the fields have consistent styling!

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -310,7 +376,6 @@ const styles = StyleSheet.create({
         marginTop: sizing.size_010, // This vertically aligns the icon with the text
     },
     errorMessage: {
-        minWidth: sizing.size_0, // This enables the wrapping behaviour on the error message
```
beaesguerra, line 313: Moved this to `helperTextMessage` so it can be more generic!

packages/wonder-blocks-labeled-field/src/theme/thunderblocks.ts
```diff
@@ -30,25 +30,33 @@ export default mergeTheme(defaultTheme, {
...
+                foreground: semanticColor.core.foreground.disabled.strong,
+            },
         },
     },
     error: {
         color: {
             foreground: semanticColor.core.foreground.critical.default,
         },
         font: {
-            size: font.body.size.xsmall,
```
beaesguerra, line 40: Moved the common properties between the error message and read only message to the `helperText` section

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -52,8 +52,21 @@ type Props = {
...
+    /**
+     * The helpful text message to display when the field is read only.
+     *
+     * Use the `labels.readOnlyAriaLabel` prop to set the `aria-label` for
+     * the read only icon.
+     *
+     * If both `errorMessage` and `readOnlyMessage` are provided, the `readOnlyMessage`
+     * is displayed first.
+     */
+    readOnlyMessage?: React.ReactNode;
```
jandrade, line 69: I wonder if it would be better to expand this new prop not only to readOnly cases but to any case where we want to use a "helper" message. I'm not sure if you have discussed this with Design, but I can see it as a way to provide more flexibility to this element. Wdyt?

beaesguerra: Good question! In the next PR, I plan to introduce props for more generic helper text:
- `elementBeforeFieldStart`
- `elementBeforeFieldEnd`
- `elementAfterFieldStart`
- `elementAfterFieldEnd`

(Open to naming suggestions! Was trying to avoid top/bottom and left/right naming)

image

I think for the `readOnlyMessage` and `errorMessage`, it would be nice for us to provide support for that out of the box because those I think this pattern will be more common. Then we can control the styling and what icon is provided, and also the `aria-live` region for the error message (I remember this was tricky to get right!). These are also both included in the [design state sheet for input fields](https://www.figma.com/design/EuFu0U7gqc1koXm8ZhlOLp/%E2%9A%A1%EF%B8%8F-Components?node-id=5302-354&m=dev)

image

The new props I plan on introducing are more for customized helper text.

Let me know what you think!

beaesguerra: Here is the other PR where more props are added to LabeledField to support more flexible helper text elements: https://github.com/Khan/wonder-blocks/pull/2710 !

jandrade: Thanks for the context!

If you check in the Figma State sheet, you'll see that the ones on the right are for RTL purposes, so we would only expect to include helpers on the bottom left (by default) of the `LabeledField` component.

I think our intention from the design perspective is to enforce/align on one way to present these helper texts so the experience is the same for all forms across the experiences. We can check with Design to see what they think about this, but my initial thinking is that it would be better to only allow one `helperText` message per LabeledField and it's up to us (WB) to define the layout for it. Similar to how `label` and `description` work already.

beaesguerra: In the [Handshake file](https://www.figma.com/design/e9qdyiDUBZ3rDabP9S7ulO/%E2%9A%A1%EF%B8%8F-Handshake?node-id=4098-6293&m=dev), the Figma component supports Start/End Helper text above and below the field so it accounts for helper text on all corners!

image 

Here's an example from Caitlyn of how it could be used:
image


It does feel a bit off to support `errorMessage`, `readOnlyMessage` and `elementAfterFieldStart` since they could technically all be set and all be shown (I've documented that it is up to the consuming apps to handle the logic of when to show what! [Helper Text docs](https://5e1bf4b385e3fb0020b7073c-fytwzghyze.chromatic.com/?path=/docs/packages-labeledfield--docs#helper-text) from the helper text PR) 

An alternative is to only have a generic `elementAfterFieldStart` prop, and then create separate `FieldErrorMessage` and `FieldReadOnlyMessage` components that can be placed in this prop so that it can handle the styling, icon and aria-live region for the error message. It would still be possible to place multiple things in `elementAfterFieldStart`.

One drawback of this alternative is that the LabeledField component wouldn't specifically know if it is in an error or readOnly state. Right now, if an `errorMessage` is provided, we apply error styling to some parts of LabeledField and pass in `error=true` into the field component. If a `readOnlyMessage` is provided, we pass in `readOnly=true` into the field component too. If we only have the generic `elementAfterFieldStart`, we would need to add these as props explicitly! 

Let me know what you think!

.changeset/chilly-adults-kiss.md
marcysutton: That's a good question. I would think yes, as it could break styling for CSS variable consumers (if there were any). But curious to hear from @jandrade!

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
```diff
@@ -482,6 +639,60 @@ describe("LabeledField", () => {
...
+                    defaultOptions,
+                );
+
+                // Act
+                const field = screen.getByRole("textbox");
+
+                // Assert
+                expect(field).toHaveAttribute(
+                    "aria-describedby",
+                    `${id}-description ${id}-error ${id}-read-only-message`,
```
marcysutton: This is cool!

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
beaesguerra: I paired with Juan earlier today to chat more about this! Here is a proof of concept for having a more flexible LabeledField API and composable helper text components: https://github.com/Khan/wonder-blocks/pull/2712 . Let me know what you all think !

beaesguerra: Thanks all for the feedback and discussion in our sync today (related [Figma thread](https://www.figma.com/design/e9qdyiDUBZ3rDabP9S7ulO?node-id=4073-4053#1325683925))! To summarize, these are the kinds of helper text we want to support:
- helper text above the field (✅ this is what `description` currently does)  
- helper text under the field (**new** - will implement next)
- read only helper text under the field (**new** - this is implemented in this PR! I think the `readOnlyMessage` prop make sense to keep)
- error message under the field, after all other helper text under the field (✅ this is the current `errorMessage` prop)

For the character count pattern, we decided with design to not include official support for it right now. It's something we can standardize later on once we have more information about patterns that would use it. Until then, requirements around character count can be included in the description, and exceeding any limits should result in an error message that is shown/announced to the user. 

Next steps for the helper text implementation:
- I'll keep this PR open since we want to standardize the pattern for read only helper text using the `readOnlyMessage` prop
- I'll close out the PR that adds helper text to all corners of the field https://github.com/Khan/wonder-blocks/pull/2710
- I'll close out the POC for supporting flexible elements https://github.com/Khan/wonder-blocks/pull/2712
  - I learned some other things while prototyping this POC! I was able to simplify LabeledField more by creating sub-components for the types of helper text. I might do this small refactor anyways so it is easier to maintain the different states, we will see! 
- Implement the helper text under the field (I'm thinking of adding a `additionalHelperMessage` prop, open to suggestions for naming!) 
- Replace the `required` prop with a `contextLabel` prop for the visible `required` or `optional` labels

.changeset/chilly-adults-kiss.md
jandrade: I'd go with `patch` in this case as component-level tokens are "private" for consumers. Supporting component-level CSS variables directly is a side effect of our current build setup, and consumers (Perseus) are aware that this is not publicly supported right now.

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
```diff
@@ -75,6 +78,64 @@ describe("LabeledField", () => {
...
+                readOnlyMessage={readOnlyMessage}
+            />,
+            defaultOptions,
+        );
+
+        // Assert
+        expect(screen.getByText(readOnlyMessage)).toBeInTheDocument();
+    });
+
+    it("LabeledField renders the read only message text if there is an error message", () => {
```
jandrade: **nit:** this means that it would read both, right? how do you feel about changing the test name/title to state that.

e.g.
```suggestion
    it("LabeledField renders the read only message text even if there is an error message", () => {
```

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
```diff
@@ -175,6 +258,31 @@ describe("LabeledField", () => {
             // Assert
             expect(errorIcon).toHaveAttribute("aria-label", "Error:");
         });
+
+        it("Should set an aria-label on the read only icon if provided", () => {
```
jandrade, line 262: **praise:** Nice test!

packages/wonder-blocks-labeled-field/src/components/__tests__/labeled-field.test.tsx
jandrade: +++

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -98,7 +114,8 @@ type Props = {
 };
 
 export type LabeledFieldLabels = {
-    errorIconAriaLabel: string;
+    errorIconAriaLabel?: string;
+    readOnlyAriaLabel?: string;
```
jandrade: **question:** How do you feel about using the same naming structure for both keys here?

e.g.
```
errorIconAriaLabel?: string;
readOnlyIconAriaLabel?: string;
```

Or renaming the error one, so both are consistent:

```
errorAriaLabel?: string;
readOnlyAriaLabel?: string;
```

beaesguerra: Good catch! I've updated it to `readOnlyIconAriaLabel` for consistency 😄

packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -251,22 +272,54 @@ export default function LabeledField(props: Props) {
         return React.cloneElement(field, {
             id: fieldId,
             "aria-describedby": [
-                errorMessage && errorId,
                 description && descriptionId,
+                readOnlyMessage && readOnlyMessageId,
+                errorMessage && errorId,
```
beaesguerra, line 277: I made a small update to the order of the aria-describedby ids so that the error message is last (similar to the visual order)

PR #2712, opened 2025-07-10
LabeledField POC: Composable helper text and generic slot props

beaesguerra:
## Summary:

Creating a proof of concept for generic slot props for above and below the field based on https://github.com/Khan/wonder-blocks/pull/2709#discussion_r2198106793

This PR includes the following components:
- FlexibleLabelField : For prototyping purposes, this is the LabeledField component (if we like this approach, these changes would be made to the existing component)
- Helper text components to be used with the LabeledField component
  - HelperText
  - ErrorHelperText
  - ReadOnlyHelperText 

I think this approach is more composable and flexible, though it is less opinionated 

Related stories:
- LabeledField usage `?path=/docs/packages-labeledfield-flexiblelabeledfieldtesting--docs&globals=theme:thunderblocks`
- Types of Helper Text `?path=/docs/packages-labeledfield-helpertext--docs&globals=theme:thunderblocks`

Note: This PR is a POC for the LabeledField api. It doesn't completely follow the design details right now and doesn't handle different text wrapping cases. It does wire up the `label` `htmlFor` attribute and the field's `aria-describedby` attribute

Issue: WB-2011

## Test plan:

### DIFF
--- a/__docs__/wonder-blocks-labeled-field/flexible-labeled-field.stories.tsx
+++ b/__docs__/wonder-blocks-labeled-field/flexible-labeled-field.stories.tsx
@@ -31,7 +31,7 @@ export default {
 export const Props = {
     args: {
         label: "Label",
-        field: <TextField value="Value" onChange={() => {}} />,
+        field: <TextField value="Field" onChange={() => {}} />,
         helperTextAbove: <HelperText message="Helper text above" />,
         helperTextBelow: <HelperText message="Helper text below" />,
     },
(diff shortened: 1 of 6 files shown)


__docs__/wonder-blocks-labeled-field/flexible-labeled-field.stories.tsx
```diff
@@ -0,0 +1,143 @@
+import * as React from "react";
```
beaesguerra, line 1: These stories show example usage on how the Helper Text components are used with LabeledField. 

(Created FlexibleLabeledField temporarily for prototyping. If we decide to go with this approach, the changes will be made to the LabeledField component)

beaesguerra: I've left code comments starting with a `NOTE:` in this file to highlight some important things/differences

__docs__/wonder-blocks-labeled-field/helper-text.stories.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from "react";
```
beaesguerra, line 1: This shows the different types of helper text components. We can create more later on if more common types of helper text come up.

packages/wonder-blocks-labeled-field/src/components/error-helper-text.tsx
```diff
@@ -0,0 +1,44 @@
...
+import WarningCircle from "@phosphor-icons/core/bold/warning-circle-bold.svg";
+import {PhosphorIcon} from "@khanacademy/wonder-blocks-icon";
+import {HelperText} from "./helper-text";
+import theme from "../theme";
+
+type Props = {
+    message?: React.ReactNode;
+    id?: string;
+    testId?: string;
+    show: boolean;
```
beaesguerra, line 11: There is a mandatory `show` prop for the `ErrorHelperText` component because even if there is no error, we want this component to be mounted so that the aria-live region is ready for when there is an error. Open to other suggestions on how to handle a persistent aria-live region!

packages/wonder-blocks-labeled-field/src/components/flexible-labeled-field.tsx
```diff
@@ -0,0 +1,67 @@
+import {View} from "@khanacademy/wonder-blocks-core";
```
beaesguerra, line 1: Created a new component for prototyping purposes. If we go with this approach, these changes would be made to the LabeledField component

packages/wonder-blocks-labeled-field/src/components/error-helper-text.tsx
```diff
@@ -0,0 +1,44 @@
...
+            // instead of role=alert because Safari + VoiceOver would
+            // not read out the error when focused on if the element
+            // referenced by the aria-describedby had role="alert".
+            aria-live="assertive"
+            // We add aria-atomic=true so that any updates to the error
+            // is announced
+            aria-atomic="true"
+        />
+    );
+};
```
beaesguerra, line 44: Overall, I also like breaking out the helper text into smaller components to simplify the LabeledField component. We can also export it and they can be re-used in other places too if there are custom implementations of things that would have error text.

We can break out the helper text into smaller components regardless of which approach we take!

packages/wonder-blocks-labeled-field/src/components/flexible-labeled-field.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const hasError = field.props.error;
+    const isDisabled = field.props.disabled;
+
+    function renderField() {
+        return React.cloneElement(field, {
+            id: fieldId,
+            "aria-describedby": [helperTextAboveId, helperTextBelowId]
+                .filter(Boolean)
+                .join(" "),
+        });
```
beaesguerra, line 30: In LabeledField, we would also set the `error` and `readOnly` props on the `field` here. Now, since we don't know if there is an `errorMessage` or `readOnlyMessage`, we leave it up to the consumer to set those props on the field properly. 

I like this because it narrows down what the LabeledField component does

packages/wonder-blocks-labeled-field/src/components/flexible-labeled-field.tsx
```diff
@@ -0,0 +1,67 @@
...
+import {StyleSheet} from "aphrodite";
+import * as React from "react";
+
+type Props = {
+    id?: string;
+    testId?: string;
+    label: React.ReactNode;
+    field: React.ReactElement;
+    helperTextAbove?: React.ReactNode;
+    helperTextBelow?: React.ReactNode;
```
beaesguerra, line 13: What naming is preferred for this prop?
- `helperTextAbove`/`helperTextBelow`
- `helperTextBefore`/`helperTextAfter`
- `helperTextStart`/`helperTextEnd`

Something else?

beaesguerra:
Discussed with the team, we won't be supporting the end helper text anymore! See discussion notes for more details: https://github.com/Khan/wonder-blocks/pull/2709#discussion_r2201807093

I may apply some of the work done in this POC around creating smaller components for the different kinds of helper text to help with maintainability, we will see how it all comes together!

PR #2714, opened 2025-07-15
LabeledField: Add `additionalHelperMessage` prop

beaesguerra:
## Summary:
[wb-2018-additional-helper-message] simplify styles
[wb-2018-additional-helper-message] share styles between error and readonly icon
[wb-2018-additional-helper-message] rename some styles withing labeledfield
[wb-2018-additional-helper-message] Add tests
[wb-2018-additional-helper-message] Add stories
[wb-2018-additional-helper-message] handle disabled styling for the helper text
[wb-2018-additional-helper-message] add stories
[wb-2018-additional-helper-message] add additionalHelperMessage prop to LabeledField

Issue: WB-2018

## Test plan:

PR #2715, opened 2025-07-17
LabeledField: Remove `required` prop

beaesguerra:
## Summary:

Removing the `required` prop from `LabeledField` in favour of the new upcoming `contextLabel` prop. The asterisk that marks a field as required will be replaced with a context label (either "required" or "optional"). Consumers can pass in the `required` prop to the underlying field component used with the LabeledField

Issue: WB-2017

## Test plan:
- Confirm there is no `required` prop on the `LabeledField` anymore
- Confirm the LabeledField docs don't include references to the `required` prop for `LabeledField` (Note: The form field components will still support a `required` prop)
- Confirm the LabeledField Required story continues to work as expected (pressing "Submit" while all the fields are empty shows the custom error message provided to the field) `?path=/story/packages-labeledfield--required`

## Implementation plan for LabeledField updates:
- Support `readOnlyMessage` (released and deployed!) https://github.com/Khan/wonder-blocks/pull/2709
- Support `additionalHelperMessage` https://github.com/Khan/wonder-blocks/pull/2714
- Remove `required` prop (this PR) https://github.com/Khan/wonder-blocks/pull/2715
- Add `contextLabel` prop (next)

### DIFF
--- a/.changeset/fifty-bananas-explode.md
+++ b/.changeset/fifty-bananas-explode.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-labeled-field": major
+---
+
+LabeledField: Remove `required` prop. The upcoming `contextLabel` prop should be used instead for visual labels indicating if a field is required or optional. Field components will need to have their `required` prop set explicitly
--- a/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
@@ -204,12 +204,12 @@ export const WithLabeledField: StoryComponentType = {
                         selectedValues={value}
                         onChange={setValue}
                         onValidate={setErrorMessage}
+                        required={true}
                     >
                         {optionItems}
                     </MultiSelect>
                 }
                 description="Description"
-                required={true}
                 errorMessage={errorMessage}
             />
         );
--- a/__docs__/wonder-blocks-labeled-field/accessibility.mdx
+++ b/__docs__/wonder-blocks-labeled-field/accessibility.mdx
@@ -30,8 +30,8 @@ prop is not provided, a unique id will be auto-generated!
 
 Since the label, description, and error message are associated with the field,
 they are read out with the field when it is interacted with. If the field component
-handles the `error` and `required` props and sets the `aria-invalid` and `aria-required`
-attributes, then this information is also announced.
+handles the `error` prop and sets the `aria-invalid` attribute, then this
+information is also announced.
 
 ### Error messages
 
--- a/__docs__/wonder-blocks-search-field/search-field.stories.tsx
+++ b/__docs__/wonder-blocks-search-field/search-field.stories.tsx
@@ -139,7 +139,6 @@ export const WithLabeledField: StoryComponentType = {
                     />
                 }
                 description="Description"
-                required={true}
                 errorMessage={errorMessage}
             />
         );
--- a/.changeset/strange-squids-try.md
+++ b/.changeset/strange-squids-try.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": patch
+---
+
+Update LabeledField component theme css variables (required tokens are no longer needed)
--- a/__docs__/wonder-blocks-dropdown/single-select.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/single-select.stories.tsx
@@ -225,12 +225,12 @@ export const WithLabeledField: StoryComponentType = {
                         selectedValue={value}
                         onChange={setValue}
                         onValidate={setErrorMessage}
+                        required={true}
                     >
                         {optionItems}
                     </SingleSelect>
                 }
                 description="Description"
-                required={true}
                 errorMessage={errorMessage}
             />
         );
--- a/__docs__/wonder-blocks-form/labeled-text-field.stories.tsx
+++ b/__docs__/wonder-blocks-form/labeled-text-field.stories.tsx
@@ -108,11 +108,11 @@ export const MigrationToLabeledField: StoryComponentType = {
                 <LabeledField
                     label="Using LabeledField with TextField (recommended)"
                     description={description}
-                    required={required}
                     errorMessage={textFieldErrorMessage}
                     field={
                         <TextField
                             {...args}
+                            required={required}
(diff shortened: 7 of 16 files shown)


__docs__/wonder-blocks-search-field/search-field.stories.tsx
```diff
@@ -139,7 +139,6 @@ export const WithLabeledField: StoryComponentType = {
                     />
                 }
                 description="Description"
-                required={true}
```
beaesguerra, line 142: The `required` prop isn't moved to the `SearchField` because the `SearchField` doesn't support a `required` prop

__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
```diff
@@ -204,12 +204,12 @@ export const WithLabeledField: StoryComponentType = {
                         selectedValues={value}
                         onChange={setValue}
                         onValidate={setErrorMessage}
+                        required={true}
```
beaesguerra, line 207: Moving the `required` prop to the field directly instead of on `LabeledField`

.changeset/fifty-bananas-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-labeled-field": major
+---
+
+LabeledField: Remove `required` prop. The upcoming `contextLabel` prop should be used instead for visual labels indicating if a field is required or optional. Field components will need to have their `required` prop set explicitly
```
marcysutton, line 5: My first reaction was about whether this change would be confusing or harder to get right by consumers. Is there a technical reason why removing `required` from the outer component makes the most sense? Was it conflicting with `contextLabel`? I'm wondering if a higher-level `required` prop could enable some easy defaults with the context labeling.

__docs__/wonder-blocks-labeled-field/accessibility.mdx
```diff
@@ -30,8 +30,8 @@ prop is not provided, a unique id will be auto-generated!
 
 Since the label, description, and error message are associated with the field,
 they are read out with the field when it is interacted with. If the field component
-handles the `error` and `required` props and sets the `aria-invalid` and `aria-required`
-attributes, then this information is also announced.
+handles the `error` prop and sets the `aria-invalid` attribute, then this
```
marcysutton, line 33: This part of the docs is a little confusing to me. Is the textfield input the component with the `required` prop, or the parent `LabeledField` component?

marcysutton (commented):
I left a few high level questions!

.changeset/fifty-bananas-explode.md
beaesguerra: Good question! 

The `required` prop on the `LabeledField` component was used for:
- adding an `*` after the label
- [injecting the `required` prop on the form field component](https://github.com/Khan/wonder-blocks/blob/73d435acbaf41d02447fe771d1052b745f6a18dc/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx#L281). This was so that a consumer didn't have to pass in a `required` prop into both `LabeledField` and the form field component (TextField, TextArea, etc) 

For more context, the `required?: boolean | string` prop on form field components (`TextField`, `TextArea`, `MultiSelect`, `SingleSelect`) is used for:
- setting the `required` attribute on the element
- validation logic - when a required field is cleared, the `onValidate` prop on fields will be called with an error message. The error message is either a default error message, or the `required` prop string value: https://github.com/Khan/wonder-blocks/blob/73d435acbaf41d02447fe771d1052b745f6a18dc/packages/wonder-blocks-form/src/hooks/use-field-validation.ts#L123-L131

With the upcoming `contextLabel` prop, it is more flexible. Consumers pass in a translated string to indicate if a field is `required` or `optional`. Since it is more flexible, it could be used for other things too (in the [design](https://www.figma.com/design/e9qdyiDUBZ3rDabP9S7ulO/%E2%9A%A1%EF%B8%8F-Handshake?node-id=4076-1365&m=dev), this text is only referred to as `Context`). Because of this flexibility, we are no longer able to inject the form field `required` prop since we don't know if the `contextLabel` is being used to mark the field as required. And since we are also no longer using the `*` to mark the LabeledField as required too, the `required` prop won't be needed anymore!

Consumers will have to set the `required` prop directly on form field components instead for validation and setting the `required` attribute. I think this makes the separation between `LabeledField` and the form field clearer, since the form field components already receive the other validation props (`validate`,`onValidate`, `instantValidation`) instead of the `LabeledField` component. And the `LabeledField` component can focus on providing labels and helper text for the provided form field. In hindsight, I think we probably didn't need to have the `required` prop on the `LabeledField` too. We could've determined if we needed to show the `*` based only on if the `field` has it's `required` prop set!

With all of that, these new changes shouldn't make it more difficult for consumers to use since they were already using the `required` prop directly on form fields before there was a `LabeledField` component. The new functionality allows for a visible label, which we didn't have before! Let me know what you think 😄

__docs__/wonder-blocks-labeled-field/accessibility.mdx
beaesguerra: The `TextField` (and other form field components) have the `required` prop for validation! 

This was saying that the `field` prop needed to support `error` and `required` props because `LabeledField` would inject these props into the field. We only do this for the `error` prop now!

.changeset/fifty-bananas-explode.md
marcysutton: Makes sense! Thank you so much for laying that all out in detail!

beaesguerra:
Thanks for the review @marcysutton !

> I left one question in Chromatic about descriptions for disabled fields. Maybe we should wrap them in a fieldset that can be marked as disabled?

Great suggestion, responded in the [Chromatic thread](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=6876be878deb00f54437297b) 😄 If we prefer using `fieldset` + `legend` for `LabeledField`, I can create a separate ticket for updating the semantics! Let me know what you think!

marcysutton:
> If we prefer using fieldset + legend for LabeledField, I can create a separate ticket for updating the semantics! 

I think it would be the best way to associate all the helper text items with a disabled field, as well as solve the automated test errors. Agreed that a separate ticket would be a good way to go, especially since it would require some experimentation!

PR #2717, opened 2025-07-17
Create initial FlexibleDialog (part 1)

marcysutton:
## Summary:
[WB-1883-modal] Add longer docs description
[WB-1883-modal] Add tests for labeling
[WB-1883-modal] Add title render prop
[WB-1883-modal] Add basic flexible modal

Issue: WB-1883

## Test plan:
1. Check out FlexibleDialog stories: /?path=/docs/packages-modal-flexibledialog--docs&globals=viewport:desktop
2. Ensure tests pass

## Implementation plan:
1. Introduce initial flexible dialog (this PR)
2. Introduce DrawerLauncher (future PR)


packages/wonder-blocks-modal/src/components/flexible-content.tsx
```diff
@@ -0,0 +1,73 @@
...
+    /** Optional styling to apply to the contents. */
+    style?: StyleType;
+    /** Optional id reference to apply to the first heading for labeling the dialog. */
+    titleId?: string;
+};
+
+/**
+ * The FlexibleDialog main content
+ */
+function FlexibleContent(props: Props) {
```
marcysutton: We might not need this file. I copied it over from `OnePaneDialog`/`ModalDialog`/etc., but other than naming the files are identical.

packages/wonder-blocks-modal/src/components/__tests__/flexible-dialog.test.tsx
```diff
@@ -0,0 +1,236 @@
...
+            );
+
+            // Act
+            const dialog = screen.getByRole("dialog");
+
+            // Assert
+            expect(dialog).toHaveAccessibleName("Accessible name");
+        });
+
+        it("applies aria-labelledby from titleId and title component", () => {
```
marcysutton, line 181: I might add some more tests around `aria-labelledby` and a `title` component with its own ID, to ensure those connections can be made.

packages/wonder-blocks-modal/src/components/flexible-footer.tsx
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * **Usage**
+ *
+ * ```js
+ * <FlexibleFooter>
+ *     <Button onClick={() => {}}>Submit</Button>
+ * </FlexibleFooter>
+ * ```
+ */
+export default function FlexibleFooter({children}: Props) {
```
marcysutton: I'm not sure if we need this either. It's unclear whether we need a footer, or whether consumers could add their own.

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
```diff
@@ -0,0 +1,231 @@
...
+        });
+    }, [title, content, footer, scrollOverflow]);
+
+    const mainContent = renderMainContent();
+
+    const defaultBackgroundStyle = {
+        backgroundColor: semanticColor.surface.primary,
+    };
+
+    const combinedBackgroundStyles = {
```
marcysutton, line 117: This is where a custom background image/color can be passed in to the component where it displays. It was hard-coded to `surface.primary` before and couldn't be overridden from up at the `FlexibleDialog` level.

packages/wonder-blocks-modal/src/components/modal-dialog.tsx
```diff
@@ -36,6 +36,11 @@ type Props = {
      * Test ID used for e2e testing.
      */
     testId?: string;
+    /**
+     * The accessible name of dialog.
+     * See WCAG 2.1: 4.1.2 Name, Role, Value
+     */
+    "aria-label"?: string;
```
marcysutton, line 43: Note: `aria-label` is supported in dialogs, so I added it here since `FlexibleDialog` builds on top. I didn't mess with `aria-labelledby` being required to limit the scope here, but it really should require one or the other.

__docs__/wonder-blocks-modal/flexible-dialog.argtypes.ts
```diff
@@ -0,0 +1,152 @@
+import type {ArgTypes} from "@storybook/react";
+
+export default {
+    title: {
+        control: {type: "text"},
+        description: `The main heading of the modal, for labeling the dialog.
+            If string content, an ID will be generated for the heading for aria-labelledby on the dialog.
+            If a node, an ID will be applied to a DIV wrapping the node.
+            Required unless using aria-label or aria-labelledby.`,
```
jandrade: **suggestion:** I think you can get rid of all the `description` keys in this file. For context, we had to keep this with the other modal argtypes due to some limitations with a prev. Storybook version, but that got fixed after v8.

__docs__/wonder-blocks-modal/flexible-dialog.argtypes.ts
```diff
@@ -0,0 +1,152 @@
+import type {ArgTypes} from "@storybook/react";
+
+export default {
+    title: {
+        control: {type: "text"},
+        description: `The main heading of the modal, for labeling the dialog.
```
beaesguerra: Is it possible for Storybook to infer the prop descriptions based on the jsdocs for the props? If it does, we can remove the description in the argtypes so we only need to update descriptions in one place 😄 

I've found that Storybook docs sometimes doesn't infer more complicated types though so I think the other configurations for the `control` and `table` make sense!

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
```diff
@@ -0,0 +1,211 @@
...
+     */
+    content:
+        | React.ReactElement<PropsFor<typeof ModalContent>>
+        | ((slots: RenderProps) => React.ReactNode)
+        | React.ReactNode;
+    /**
+     * The optional background styles for the panel.
+     * If not provided, defaults to semanticColor.surface.primary background color.
+     */
+    backgroundStyles?: BackgroundStyles;
```
beaesguerra: For components that have multiple elements within it that can be styled, we've been using a `styles` prop with different keys for the elements that can get styled! 

Would it make sense here to combine the `style` and `backgroundStyles` props into: 
```
styles?: {
  panel?: StyleType;
  background?: StyleType;
}
```

I see though that `style` and `backgroundStyles` are applied to the same element though! So I wonder if only support a `style` prop would suffice?

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+    return (
+        <View
+            style={[styles.wrapper, combinedBackgroundStyles, style]}
+            testId={testId && `${testId}-panel`}
+        >
+            {closeButtonVisible && (
+                <CloseButton
+                    onClick={onClose}
+                    style={[styles.closeButton]}
```
beaesguerra: Should we support custom styling for the close button? There could be a case where we want the close icon to have an inverse color if a full width background image is darker! If so, it would make sense to have a `styles` prop instead of `style` so that different elements can be styled!

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
```diff
@@ -0,0 +1,211 @@
...
+    backgroundStyles?: BackgroundStyles;
+    /**
+     * When true, the close button is shown; otherwise, the close button is not shown.
+     */
+    closeButtonVisible: boolean;
+    /**
+     * Should the contents of the panel become scrollable should they
+     * become too tall?
+     */
+    scrollOverflow: boolean;
```
beaesguerra: Is there a use case where we wouldn't want the panel to be scrollable if the contents become too tall? I'm wondering if this behaviour should always be true! (I see this is something in ModalPanel too, so it's pre-existing there!)

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
```diff
@@ -0,0 +1,211 @@
...
+     * If you're using `ModalLauncher`, you should not use this prop!
+     * Instead, to listen for when the modal closes, add an `onClose` handler
+     * to the `ModalLauncher`.  Doing so will throw an error.
+     */
+    onClose?: () => unknown;
+    /**
+     * Test ID used for e2e testing.
+     *
+     * In this case, this `testId` comes from the `testId` prop defined in the
+     * Dialog variant (e.g. OnePaneDialog).
```
beaesguerra: Should it reference FlexibleDialog in the example?

```suggestion
     * Dialog variant (e.g. FlexibleDialog).
```

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
```diff
@@ -0,0 +1,211 @@
...
+    backgroundPosition?: string;
+    /**
+     * How the background image should be sized.
+     */
+    backgroundSize?: string;
+    /**
+     * How the background image should fit within its container.
+     */
+    objectFit?: "fill" | "contain" | "cover" | "none" | "scale-down";
+};
```
jandrade: **suggestion:** It would be probably better to let the consumer decide which styles to apply to the panel. The custom background in the GEM celebration modal is more of a very rare case where they need to pass a custom bg.

packages/wonder-blocks-modal/src/index.ts
```diff
@@ -13,5 +15,7 @@ export {
     ModalPanel,
     ModalLauncher,
     OnePaneDialog,
+    FlexibleDialog,
+    FlexiblePanel,
```
jandrade: **suggestion:** This probably doesn't need to be exported as consumers should be able to create these kind of modal dialogs with just `FlexibleDialog`.

```suggestion

```

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
```diff
@@ -0,0 +1,211 @@
...
+    const renderMainContent = React.useCallback((): React.ReactNode => {
+        const contentNode =
+            typeof content === "function" ? (
+                content({title})
+            ) : (
+                <>
+                    {title}
+                    {content}
+                </>
+            );
```
jandrade, line 92: **praise:** This API is neat! love how the `title` can be used in a different layout.

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
```diff
@@ -0,0 +1,225 @@
...
+type AriaLabelledByOnly = {
+    title?: never;
+    "aria-label"?: never;
+    "aria-labelledby": string;
+};
+
+/**
+ * One of the labeling methods is required:
+ */
+type AccessibleLabelProps =
```
beaesguerra: I recently learned we have `*.typestest.tsx` tests in WB for TypeScript types! It's helpful for asserting when there should/shouldn't be a TS errors based on prop combinations. I think it'd be helpful for this component too! I did something similar for [badges](https://github.com/Khan/wonder-blocks/blob/main/packages/wonder-blocks-badge/src/components/__tests__/badge.typestest.tsx)!

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
```diff
@@ -0,0 +1,225 @@
...
+    | AriaLabelOnly
+    | AriaLabelledByOnly;
+
+type Props = AccessibleLabelProps & {
+    /**
+     * The main heading of the modal, for labeling the dialog.
+     * If string content, an ID will be generated for the heading for aria-labelledby on the dialog.
+     * If a node, an ID will be applied to a DIV wrapping the node.
+     */
+    title?: React.ReactNode | string;
```
beaesguerra: Does `title` need to be included here if it's part of `AccessibleLabelProps`?

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
```diff
@@ -0,0 +1,225 @@
...
+     * The content of the modal. Supports a render prop for placing the title in a slot.
+     */
+    content:
+        | React.ReactElement<PropsFor<typeof FlexiblePanel>>
+        | ((slots: RenderProps) => React.ReactNode);
+    /**
+     * The background styles for the modal panel.
+     * If not provided, defaults to semanticColor.surface.primary background color.
+     */
+    backgroundStyles?: BackgroundStyles;
```
jandrade: **suggestion:** I'd like to propose the following change to be more inline with the rest of the WB stying APIs.

We could remove `backgroundStyles` and `style`, and instead create a single `styles` object that could contain two keys: (1) to target the dialog itself, (2) to target the panel.

e.g.
```
styles?: {
  root?: StyleType // points to the top node of the component, in this case the dialog node
  panel?: StyleType // replaces bgStyles and leaves the object more open to accept any style (this offers even more flexibility, which would be our end goal here).
}
```

https://khanacademy.atlassian.net/wiki/spaces/WB/pages/3186294919/Guidelines+for+Web+Wonder+Blocks+Components?focusedCommentId=3298197505

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
```diff
@@ -0,0 +1,225 @@
...
+    above?: React.ReactNode;
+    /**
+     * When set, provides a component that will render content below the bottom of the modal;
+     * when not set, no additional content is shown below the modal.
+     * This prop is passed down to the ModalDialog.
+     *
+     * NOTE: Devs can customize this content by rendering the component assigned to this prop with custom styles,
+     * such as by wrapping it in a View.
+     */
+    below?: React.ReactNode;
```
jandrade: **suggestion:** These two props can be removed. For context, these were added a few years ago to support these cases but Design + Product never used them, so this is no longer necessary.

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
```diff
@@ -0,0 +1,225 @@
...
+     */
+    title?: React.ReactNode | string;
+    /**
+     * An optional id parameter for the main heading. If one is not provided,
+     * an ID will be generated.
+     */
+    titleId?: string;
+    /**
+     * The content of the modal. Supports a render prop for placing the title in a slot.
+     */
```
beaesguerra, line 29: I think it would be worth mentioning in the description that in addition to supporting a render prop, a `FlexiblePanel` should be used with `FlexibleDialog`!

I found that TS can't make children components type safe at this time so `React.ReactElement<PropsFor<typeof FlexiblePanel>>` doesn't enforce the content element to be a `FlexiblePanel` component 🥲 ([Related Slack thread](https://khanacademy.slack.com/archives/C028T2HDX3P/p1741210473344549))

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
```diff
@@ -0,0 +1,225 @@
...
+        </Id>
+    );
+};
+
+const componentStyles = StyleSheet.create({
+    dialog: {
+        width: "93.75%",
+        maxWidth: 576,
+        height: "81.25%",
+        maxHeight: 624,
```
jandrade: **thought:** One regular comment that we get from devs is that it would be nice if the height would be given by the dialog contents, so this could be a good opportunity to to that (we should still provide a width by default).

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
```diff
@@ -0,0 +1,225 @@
...
+    style,
+    closeButtonVisible = true,
+    testId,
+    titleId,
+    role,
+    backgroundStyles,
+    ...accessibilityProps
+}: Props): React.ReactElement => {
+    return (
+        <Id id={titleId}>
```
beaesguerra: Since `FlexibleDialog` is a functional component already, can we use the `React.useId` hook? 😄

__docs__/wonder-blocks-modal/flexible-dialog.stories.tsx
```diff
@@ -0,0 +1,539 @@
...
+    </View>
+);
+
+CelebrationModal.parameters = {
+    docs: {
+        description: {
+            story: `A FlexibleDialog with full-bleed background image and custom contents.`,
+        },
+    },
+};
```
jandrade: **suggestion:** these could be moved to story definition as JS doc blocks to be more inline with Storybook's CSFv3. (same applies to all stories in this file).

__docs__/wonder-blocks-modal/flexible-dialog.stories.tsx
```diff
@@ -0,0 +1,539 @@
...
+
+export const WithAboveAndBelow: StoryComponentType = () => {
```
jandrade: **suggestion:** Based on the related comment, this story could be removed.

__docs__/wonder-blocks-modal/flexible-dialog.stories.tsx
jandrade, line 1: **suggestion:** I feel that it would be super valuable adding an example passing `title` in the render function as I think developers will find this feature super useful.

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
```diff
@@ -0,0 +1,225 @@
...
+            {(uniqueId) => {
+                const headingId = titleId || uniqueId;
+
+                const renderedTitle =
+                    title == null ? null : typeof title === "string" ? (
+                        <Heading id={headingId}>{title}</Heading>
+                    ) : (
+                        <View id={headingId} testId="title-heading-wrapper">
+                            {title}
+                        </View>
```
jandrade: **thought:** I wonder if instead of creating an extra wrapper/div, this could be approached using `React.cloneElement` to augment the component and pass the `headingId` as an extra prop?

jandrade (commented):
This is looking really good! thanks for creating this :). I left some comments to get your thoughts on. Let me know what you think.

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
beaesguerra: Is there a use case for supporting `above` and `below` props in `FlexibleDialog`? I'm wondering if consumers need to use those props, should they use the `ModalDialog` kind instead? What do you think?

packages/wonder-blocks-modal/src/components/__tests__/flexible-dialog.test.tsx
```diff
@@ -0,0 +1,253 @@
...
+            );
+
+            // Act
+            const modal = screen.getByRole("dialog");
+
+            // Assert
+            expect(modal).toHaveAccessibleDescription(/cool dialog/i);
+        });
+    });
+});
```
beaesguerra, line 245: Would it be helpful to also include jest-axe tests? We have `toHaveNoA11yViolations` matcher for running axe checks on components!

__docs__/wonder-blocks-modal/flexible-dialog.stories.tsx
```diff
@@ -0,0 +1,539 @@
+/* eslint-disable no-alert */
```
beaesguerra: Could we use a `eslint-disable-next-line` comment instead so it is locallzed to the call site where it is intentionally used?

__docs__/wonder-blocks-modal/flexible-dialog.stories.tsx
```diff
@@ -0,0 +1,539 @@
...
+import packageConfig from "../../packages/wonder-blocks-modal/package.json";
+
+import ComponentInfo from "../components/component-info";
+import FlexibleDialogArgTypes from "./flexible-dialog.argtypes";
+import {allModes} from "../../.storybook/modes";
+
+import celebrationBg from "../../static/celebration_bg.svg";
+import celebrationChest from "../../static/celebration-chest.svg";
+
+const customViewports = {
```
beaesguerra, line 20: I wonder if this is something we can standardize so we can reuse it for other components too!

__docs__/wonder-blocks-modal/flexible-dialog.stories.tsx
```diff
@@ -0,0 +1,539 @@
...
+                occaecat cupidatat non proident, sunt in culpa qui officia
+                deserunt mollit anim id est.`}
+                </BodyText>
+            </>
+        ),
+    },
+};
+
+const celebrationBgStyle = {
+    backgroundColor: "#FCE6F7", // fallback color
```
beaesguerra: Is there a semantic token we can use for the example?

__docs__/wonder-blocks-modal/flexible-dialog.stories.tsx
```diff
@@ -0,0 +1,539 @@
...
+        <View style={styles.modalPositioner}>
+            <FlexibleDialog
+                title={
+                    <Heading
+                        size="xxlarge"
+                        weight="bold"
+                        id="gem-challenge-completed-modal-heading"
+                        tag="h2"
+                    >
+                        Congrats Rainier McCheddarton!
```
beaesguerra: ❤️

__docs__/wonder-blocks-modal/flexible-dialog.stories.tsx
```diff
@@ -0,0 +1,539 @@
...
+};
+
+const celebrationBgStyle = {
+    backgroundColor: "#FCE6F7", // fallback color
+    backgroundImage: `url(${celebrationBg})`,
+    backgroundRepeat: "no-repeat",
+    backgroundSize: "cover",
+} as const;
+
+export const CelebrationModal: StoryComponentType = () => (
```
beaesguerra: Can we rename this story to be more generic? Maybe ModalWithBackgroundImage or something so that it can be applied to other use cases!

__docs__/wonder-blocks-modal/flexible-dialog.stories.tsx
```diff
@@ -0,0 +1,539 @@
...
+                    below={<View style={belowStyle} />}
+                />
+            </View>
+        </View>
+    );
+};
+
+WithAboveAndBelow.parameters = {
+    docs: {
+        description: {
```
beaesguerra: I find it easier to provide the story description as jsdocs for the story so you don't need to escape certain characters!

__docs__/wonder-blocks-modal/flexible-dialog.stories.tsx
```diff
@@ -0,0 +1,539 @@
...
+            story: `A FlexibleDialog can have custom styles via the
+            \`style\` prop. Here, the modal has a \`maxWidth: 1000\` and
+            \`color: Color.blue\` in its custom styles.`,
+        },
+    },
+};
+
+export const MultiStepModal: StoryComponentType = () => {
```
beaesguerra: The spacing in the MultiStepModal and WithLauncher stories is a bit off! For these use cases, should consumers use the ModalDialog component instead?

beaesguerra (commented):
Great work so far! Left some comments and questions in the PR and Chromatic 😄

__docs__/wonder-blocks-modal/flexible-dialog.stories.tsx
jandrade: ohhh I missed this, soo cool 😄

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
marcysutton: I think we need `styles` to support the background and `closeButton`!

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
marcysutton: Added!

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
marcysutton: I ended up deleting the `scrollOverflow` prop since I also removed the fixed height! I figure we can add it back in at a later date if it's needed.

__docs__/wonder-blocks-modal/flexible-dialog.stories.tsx
marcysutton: I removed the MultiStep story and fixed the WithLauncher story!

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
marcysutton: Great idea. Updated!

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+            }}
+        </Id>
+    );
+};
+
+const componentStyles = StyleSheet.create({
+    dialog: {
+        width: "93.75%",
+        maxWidth: 576,
+        height: "81.25%",
```
beaesguerra: Resurfacing Juan's comment RE: flexible height! https://github.com/Khan/wonder-blocks/pull/2717#discussion_r2223091282

This would fix the large empty space under the content!  I wonder if we should have kept the `max-height` instead? 

Image

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+    | AriaLabelOnly
+    | AriaLabelledByOnly;
+
+type Props = AccessibleLabelProps & {
+    /**
+     * The main heading of the modal, for labeling the dialog.
+     * If string content, an ID will be generated for the heading for aria-labelledby on the dialog.
+     * If a node, an ID will be applied to a DIV wrapping the node.
+     */
+    title?: React.ReactElement | string;
```
beaesguerra: Does the `title` prop need to be redefined here? Would it override/conflict with how `title` is configured as part of `AccessibleLabelProps`? 

Types tests can help verify this! Related comment: https://github.com/Khan/wonder-blocks/pull/2717/files#r2223077582

beaesguerra (commented):
Thanks for making those updates! Left some comments! There were some comments from the last review that haven't been addressed yet (most were minor, but was curious what your thoughts were!). 

There is also the a11y warning in Chromatic still!

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
marcysutton: The `FlexiblePanel` is used internally, so it wouldn't be something for consumers. Is it still worth adding a comment?

__docs__/wonder-blocks-modal/flexible-dialog.stories.tsx
marcysutton: Yes, good call! I see that code is repeated in a bunch of the modal stories. I'd prefer to do that as a follow-up, so I created an issue here: https://khanacademy.atlassian.net/browse/WB-2053

beaesguerra: Thank you! ❤️

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
beaesguerra: Ah right, good point! We don't need to mention it in the comment :) 

Should the type be `React.ReactElement` only without the `PropsFor<typeof FlexiblePanel>` part to avoid confusion?

beaesguerra: Also, related question! I see the render function returns a `React.ReactNode` element. Do you think it should be the same type as when a render function is not used? (ie. Should they both be `ReactElement` or `ReactNode`? It looks like ReactNode supports more types!) 

I'm still figuring out when to use which type, so I'm curious what others think about this!

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
marcysutton: Thanks for the nudge! This is a bit of a tricky one... I updated the component so it doesn't have the gap for short contents. But to allow longer contents to scroll, I had to add some constraints. `maxHeight` alone wasn't fixing the gap problem, so I included `height: "auto"` to unset the default `height: "100%"` from `ModalDialog`. One remaining issue I see is the scrollbar sits outside of the ModalDialog and its default background color, which looks a little odd with the border radius. But maybe it's okay since the styles can be redefined?

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
marcysutton: Yes, good call! I actually uncovered a bug with this change that was caught by the unit tests!

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
marcysutton: No, it shouldn't have been redefined... I pushed an update and fiddled with the types a bit!

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
marcysutton: Love this! I added a `typestest` file!

packages/wonder-blocks-modal/src/components/__tests__/flexible-dialog.test.tsx
marcysutton: I added this! I'm hesitant to add a bunch of these for all the label combinations for perf reasons (axe is a bit heavy for unit tests). Hopefully the types cover it, though.

__docs__/wonder-blocks-modal/flexible-dialog.stories.tsx
marcysutton: I think this was an artifact from `OnePaneDialog` stories. I removed it!

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
marcysutton: Here are some screenshots:

Short contents:
Screenshot 2025-07-29 at 4 43 47 PM

Long contents:
Screenshot 2025-07-29 at 4 43 55 PM

Long contents scrolled:
Screenshot 2025-07-29 at 4 44 23 PM

Short contents, small viewport:
Screenshot 2025-07-29 at 4 45 27 PM

Long contents, small viewport:
Screenshot 2025-07-29 at 4 44 51 PM

marcysutton:
@beaesguerra thanks for the review! Somehow I missed some of those comments as they weren't highlighted in the main PR view. So I appreciate the nudge to take another look!

The default story was failing without an accessible name but also not showing a type error for missing props. I opted to just add one of the labeling methods to that story rather than fiddle with Storybook typing further. The rest of the stories do properly complain if they are missing one of the labeling methods!

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
marcysutton: @jandrade mentioned that design requested those in ModalDialog but never used them, so they were safe to remove. I think it's a case of "we can add it back in it we need it"!

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
marcysutton: I played around with it, and a simpler type seemed to work for all the use cases I'd expect. So I went with it! As long as `content` is wrapped in a fragment, it's pretty flexible. You can do stuff like this:

```
<FlexibleDialog
    title={"Hey, Bagley!"}
    content={({title}) => <>{title}</>}
/>
```

And this:
```
<FlexibleDialog
    aria-label="Hey, Bagley!"
    content={<>some text</>}
/>
```

And then any other marked up content... but you can't pass a boolean, which is one difference between `React.ReactNode` and `React.ReactElement`. Similarly, I had to change the `title` prop type to `React.ReactElement | string` for the `renderedTitle` function to work properly, by excluding booleans!

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+type RenderProps = {
+    title: React.ReactNode | string;
+};
+
+type Props = {
+    /**
+     * The main heading of the FlexiblePanel. Used to label the dialog.
+     */
+    title?: React.ReactNode | string;
```
marcysutton, line 20: Interestingly, this one needed to stay as `React.ReactNode` to work with the renderedTitle from `FlexibleDialog`.

__docs__/wonder-blocks-modal/flexible-dialog.stories.tsx
```diff
@@ -251,6 +299,114 @@ export const WithStyle: StoryComponentType = {
...
+export const WithLongContents: StoryComponentType = {
+    render: () => (
+        <View style={styles.previewSizer}>
+            <View style={styles.modalPositioner}>
+                <FlexibleDialog
+                    title={<Heading>Hello, world!</Heading>}
+                    content={
+                        <>
+                            <BodyText>
+                                {`Lorem ipsum dolor sit amet, consectetur adipiscing
```
beaesguerra, line 315: Tip: We also have lorem ipsum constants for common text to test with! https://github.com/Khan/wonder-blocks/blob/main/__docs__/components/text-for-testing.ts

packages/wonder-blocks-modal/src/components/__tests__/flexible-dialog.typestest.tsx
beaesguerra, line 1: Nice! Do we want to add any cases where we expect there to be an error? Maybe when there is no title/aria-label/aria-labelledby?

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
beaesguerra: This makes sense! Thanks for testing the different scenarios and adding a story for long content! 😄

beaesguerra (approved):
Looks great! Thanks for working on this! 🚀

PR #2718, opened 2025-07-23
[WB-2002] Remove deprecated theming APIs and docs + khanmigo theme

jandrade:
## Summary:

Now that the khanmigo theme is no longer used, we are removing it from the
theming package and Storybook.

This also includes removing deprecated theming APIs and documentation as we have
moved to a different theming approach (using custom tooling + CSS variables).

NOTE: The `ThemeSwitcherContext` is being kept for internal purposes, as it is
required for presenting the ThunderBlocks values properly in Storybook (upcoming
PR with that change).

Issue: https://khanacademy.atlassian.net/browse/WB-2002

## Test plan:

Verify that no errors are thrown when running Storybook, and that the
ThunderBlocks theme is still available in the toolbar.


.changeset/flat-peaches-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-theming": major
+---
+
+Remove most of deprecated APIs that use React Context in favor of CSS variables + custom tooling"
```
beaesguerra: nit: extra `"` character!
```suggestion
Remove most of deprecated APIs that use React Context in favor of CSS variables + custom tooling
```

.changeset/flat-peaches-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-theming": major
```
beaesguerra, line 2: Will there be anything that needs to be removed in `perseus` and/or `frontend` due to the breaking change?

jandrade: Good question!

- I checked in `perseus` and we don't use those APIs.
- I removed all the uses in `frontend` a while ago to make sure that the upgrade would be simpler.

PR #2720, opened 2025-07-23
Do not try to focus hidden elements in `FocusTrap`

timmcca-be:
## Summary:
This prevents issues when the last element in a modal is hidden. Without this PR, we would try to focus it when shift+tabbing back from the close button, and focus would exit the trap.

Ideally, we'd use [`checkVisibility`](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility), but Safari 15.6 doesn't support it per [caniuse](https://caniuse.com/mdn-api_element_checkvisibility).

## Test plan:
- Run the new test
- Check out the reverted commit in this PR, which includes a story which can be used for manually testing (I don't think it's worth keeping this story around, but I wanted to see it working firsthand)

### DIFF
--- a/.changeset/flat-rivers-reflect.md
+++ b/.changeset/flat-rivers-reflect.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-modal": patch
+---
+
+Do not try to focus hidden elements in FocusTrap
--- a/packages/wonder-blocks-modal/src/components/__tests__/focus-trap.test.tsx
+++ b/packages/wonder-blocks-modal/src/components/__tests__/focus-trap.test.tsx
@@ -7,9 +7,17 @@ import {Choice, RadioGroup} from "@khanacademy/wonder-blocks-form";
 
 import FocusTrap from "../focus-trap";
 
+// jsdom's Element does not implement checkVisibility, so we assign a placeholder
+// to it here, then mock the implementation in the tests.
+Element.prototype.checkVisibility = () => true;
+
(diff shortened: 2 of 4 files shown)


.changeset/flat-rivers-reflect.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-modal": patch
```
timmcca-be, line 2: I think it's reasonable to assume nobody is depending on the broken focus trap behavior, so I think this is a pretty pure non-breaking bugfix

packages/wonder-blocks-modal/src/components/__tests__/focus-trap.test.tsx
```diff
@@ -94,7 +94,30 @@ describe("FocusTrap", () => {
...
         ).toHaveFocus();
     });
+
+    it("does not move focus to hidden elements", async () => {
+        // Arrange
+        render(
+            <FocusTrap>
+                <Button>button 1</Button>
+                <Button>button 2</Button>
+                <Button id="hidden-element" style={{display: "none"}}>
```
beaesguerra, line 107: **suggestion:** It would also be helpful to have a test for a Button with `visibility: "hidden"`!

beaesguerra (approved):
Thanks for the fix, and including a commit with a story for us to test with! The changes look good to me, I left a suggestion for more test coverage! 

Also curious if @jandrade or @marcysutton has thoughts on the PR too in case they have more context!

PR #2721, opened 2025-07-23
[WB-2042.1] Remove semanticColor.border category

jandrade:
## Summary:

Removes the `semanticColor.border` category as it is deprecated, as we have
created the `semanticColor.core.border` category instead.

Next PR will remove `semanticColor.text`.

Issue: https://khanacademy.atlassian.net/browse/WB-2042

## Test plan:

Verify that the stories in `wonder-blocks-badge` look as expected. There might be a slight change with the Base Badge component, but that should be ok.

### DIFF
--- a/.changeset/silver-icons-itch.md
+++ b/.changeset/silver-icons-itch.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": major
+---
+
+Removes semanticColor.border category as it is deprecated (core.border is the replacement)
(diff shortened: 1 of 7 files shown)


.changeset/silver-icons-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": major
```
beaesguerra, line 2: Should there be a changeset for the `badge` package too?

jandrade: Good point! added the changeset.

PR #2725, opened 2025-07-23
[WB-1940] Button: Set link role when button passes an href value

jandrade:
## Summary:

 Sets `role="link"` in Button when the `href` prop is included.

Issue: https://khanacademy.atlassian.net/browse/WB-1940

## Test plan:

Navigate to `/?path=/story/packages-button-button--with-router` and verify that
the button has `role="link"` when the `href` prop is passed.

Vefify that the unit tests pass.


marcysutton (approved):
This makes a ton of sense! There are some failing consistency tests to fix, and I anticipate there will be some failing tests in `frontend` to fix as well. But that shouldn't be too big of a lift.

Generally, I wish as an organization we were more opinionated about Links that look like links and Buttons that look like buttons. It makes it more clear for users of voice navigation software and sighted screen readers as well. But that is a much broader discussion than this change, which is still helpful!

.changeset/lovely-beers-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-button": patch
```
beaesguerra: Would changing the underlying role be a `major` change? The external API doesn't change, though this could technically break existing unit tests, maybe e2e tests too? Curious to hear if anyone has thoughts on this!

jandrade: Technically it wouldn't be a breaking change in the sense that it doesn't affect the bundled code for the library or for its consumers, only tooling would be affected (external tooling in this case). But you make a great point about breaking tooling! so I'll make sure to see what happens via snapshot releases and get back to this PR later.

marcysutton: I'm testing this PR in `frontend` and tests definitely do need to be updated! 

I also checked whether attributes like `aria-disabled` will apply with a `link` role instead of `button`, since that is used in the codebase. It looks like it will work fine as the attribute is [supported](https://www.w3.org/TR/wai-aria-1.2/#aria-disabled) and exposed for link roles.

marcysutton:
@beaesguerra I pushed an update to this PR so that links and buttons render without explicit roles. I don't think we need to bolt on roles at all anymore, since we aren't changing them based on `href`. This change adheres to the [Rules of ARIA](https://www.w3.org/TR/using-aria) more to not use ARIA when we don't need it.

.changeset/lovely-beers-bow.md
beaesguerra: Good to know, thanks for looking into that!

One thing to note is that it looks like when the Button has an `href` and is `disabled`, the `href` isn't used and the underlying element is a `button` with `aria-disabled` set! https://5e1bf4b385e3fb0020b7073c-hddxxturku.chromatic.com/?path=/story/packages-button-button--default&args=disabled:!true (I also set the href control to test this out)

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -156,6 +156,8 @@ const ButtonCore: React.ForwardRefExoticComponent<
     return (
         <ButtonUnstyled
             {...restProps}
+            // Let buttons and links have their implicit roles
+            role={undefined}
```
beaesguerra: From Juan's hand off message in our Slack chat:
> - We need be able to pass a custom role when the role prop is set. There are a couple of places in frontend where this is unexpectedly used: https://github.com/Khan/frontend/blob/juan/apps/writing-coach/src/pages/essay-history.tsx#L223-L238
> - Almost all the unit tests are passing. The only failures that exist are related to the missing role logic in the WB PR :point_up:. These tests should pass once the WB fix is in place and the snapshot release is updated in the frontend integration PR.


Here's what that component looks like. The buttons with role=tab are on the right side to represent each version:
image

Do you think that using `role=tab` is the right pattern in this case? If so, we'd need to support the `role` prop. If we don't want consumers to be able to override the role, then we could remove support for that prop. `role` is being included in the props from the `AriaProps` type: https://github.com/Khan/wonder-blocks/blob/d23c9c5f0ab6a355edade690ed03de5f31225b56/packages/wonder-blocks-core/src/util/types.ts#L14-L17

@marcysutton What do you think?

marcysutton: Thanks for the context! Maybe I could be more precise and not bolt on a role in the case of links and buttons only. The original changes in this PR oddly seem to cause some `menuitem` and `tab` tests to fail, which I'll look into some more. 🤔  I am beginning to think this _is_ a breaking change because of the widespread effect of it.

marcysutton: I pushed an update! I do think the role flexibility is warranted. I just wanted to prevent bolting on roles that we don't need in the process.

packages/wonder-blocks-button/src/components/button.tsx
```diff
@@ -58,13 +59,17 @@ const Button = React.forwardRef(function Button(
     );
 
     const extraClickableProps = beforeNav ? {beforeNav} : {target};
+    // Invoke link or button clickable behavior
+    const roleForEvents = href ? "link" : "button";
+    // prevent redundant roles for links and buttons, while allowing other roles like `tab` or `menuitem`
+    const renderedRole =
+        role === "link" || role === "button" ? undefined : role;
```
beaesguerra, line 66: Do we want to include similar logic for the `ActivityButton` too?

marcysutton: Yes, good call! I looked around the codebase and didn't see any other similar usage. But if you know of anything else I should investigate, let me know! The failing `menuitem` tests in my draft `frontend` [integration PR](https://github.com/Khan/frontend/pull/2457) are a bit odd, so I'll look more closely at those.

beaesguerra: Thanks for looking into that! I think that's everything to fix 👍 I confirmed that the components in the icon button package don't set unnecessary roles when it's a button or anchor element!

.changeset/lovely-beers-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-button": major
```
beaesguerra, line 2: Would also be worth checking how these changes will impact `perseus`!

marcysutton: Good thinking! I don't see any failing tests in Perseus with this snapshot release, and there don't seem to be any `<Button>` components with `href` attributes either. So it might not have an impact. But I'll alert the team anyway!

PR #2727, opened 2025-07-23
LabeledField: minor styling updates

beaesguerra:
## Summary:
Stories:
- Add max width to testing stories so that visual regression tests capture the text wrapping behaviour

While checking these scenarios, I addressed some styling issues:
- Only render the context label wrapper if there is a `contextLabel` prop. This prevents unnecessary spacing beside the label due to the `gap` styling on the parent
- Prevent the context label from shrinking unnecessarily. The context label should not shrink if it is less than 30% of the width (This is covered by the "With long label and short context label" scenario)

Issue: WB-XXXX

## Test plan:
- Review the scenarios to make sure text wrapping in LabeledField works as expected 
  - There should be no gap beside the label if there is no `contextLabel`
  - A short context label should not wrap if the label is long 
  - `?path=/story/packages-labeledfield-testing-snapshots-labeledfield--scenarios&globals=theme:thunderblocks`

### DIFF
--- a/__docs__/wonder-blocks-labeled-field/labeled-field-testing-snapshots.stories.tsx
+++ b/__docs__/wonder-blocks-labeled-field/labeled-field-testing-snapshots.stories.tsx
@@ -346,7 +346,7 @@ const scenarios = [
 export const Scenarios = (args: PropsFor<typeof LabeledField>) => {
     const [textFieldValue, setTextFieldValue] = React.useState("");
     return (
-        <View style={{maxWidth: "450px"}}>
+        <View style={{maxWidth: "475px"}}>
             <ScenariosLayout
                 scenarios={scenarios}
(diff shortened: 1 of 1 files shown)


packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx
```diff
@@ -168,20 +168,22 @@ export default function LabeledField(props: Props) {
...
+                            styles.helperText,
+                            styles.contextLabel,
+                            isDisabled && styles.disabledHelperText,
+                            hasError ? styles.contextLabelWithError : undefined,
+                            stylesProp?.contextLabel,
+                        ]}
+                    >
+                        {contextLabel}
+                    </BodyText>
+                )}
```
jandrade, line 186: **praise:** Great catch!

PR #2728, opened 2025-07-24
Ensure styles are generated before rendering in `ActivityIconButton`

timmcca-be:
## Summary:
This fixes an unusual bug. If all of these events happened at the same time:

- the button is pressed
- the button's action type switches to one that has not been rendered before
- the button loses focus

The button would render briefly without any styling, then awkwardly transition to its normal state.

This fix feels pretty clunky, but I'm struggling to think of something better. I tried using `flushToStyleTag` from aphrodite after generating the styles, but it did not work. I also considered pre-computing the styles for all combinations of kind and action type, but that would add a lot of noise to our stylesheets on initial render (but maybe that's okay!).

## Test plan:
Check out the commit in this PR with the story added for manual testing, click the button, and verify that it doesn't exhibit the behavior from [this video](https://khanacademy.slack.com/archives/C0917FL3B47/p1753291369069999)

### DIFF
--- a/packages/wonder-blocks-icon-button/src/components/activity-icon-button.tsx
+++ b/packages/wonder-blocks-icon-button/src/components/activity-icon-button.tsx
@@ -87,7 +87,7 @@ export const ActivityIconButton: React.ForwardRefExoticComponent<
 
     const [pressed, setPressed] = React.useState(false);
 
-    const latestButtonStyles = _generateStyles(actionType, !!disabled, kind);
+    const latestButtonStyles = _generateStyles(actionType, kind);
     // we only want to update the button styles the render cycle after they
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-icon-button/src/components/activity-icon-button.tsx
```diff
@@ -194,14 +204,13 @@ const theme = {
     },
 };
 
-const styles: Record<string, any> = {};
+const styles: Record<string, StyleDeclaration> = {};
 
 const _generateStyles = (
     actionType: ActivityIconButtonActionType = "progressive",
-    disabled: boolean,
```
timmcca-be, line 201: This was not actually used in computing the styles, so I removed it

PR #2729, opened 2025-07-24
Add onPaste prop to TextArea and TextField

nishasy:
## Summary:
Add the `onPaste` prop to TextArea and TextField.

Adding this for a specific case in Perseus: In the Radio widget editor,
content authors can paste image URLs in markdown to display images. This
is hard to read, so I'm making a UI that automatically reads the pasted
text and converts it into an easily readable placeholder, and the actual
image can be edited in a separate UI.

Issue: none

## Test plan:
`pnpm jest packages/wonder-blocks-form/src/components/__tests__/text-area.test.tsx`
`pnpm jest packages/wonder-blocks-form/src/components/__tests__/text-field.test.tsx`

### DIFF
--- a/packages/wonder-blocks-form/src/components/__tests__/text-area.test.tsx
+++ b/packages/wonder-blocks-form/src/components/__tests__/text-area.test.tsx
@@ -608,6 +608,27 @@ describe("TextArea", () => {
             // Assert
             expect(handleOnBlur).toHaveBeenCalledOnce();
         });
+
+        it("should call the onPaste prop when text is pasted into the textarea", async () => {
+            // Arrange
+            const handleOnPaste = jest.fn();
+            render(
+                <TextArea
+                    value=""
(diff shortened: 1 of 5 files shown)


packages/wonder-blocks-form/src/components/__tests__/text-area.test.tsx
```diff
@@ -608,6 +608,27 @@ describe("TextArea", () => {
...
+                defaultOptions,
+            );
+
+            // Act
+            const textArea = await screen.findByRole("textbox");
+            textArea.focus();
+            await userEvent.paste("test");
+
+            // Assert
+            expect(handleOnPaste).toHaveBeenCalledOnce();
```
beaesguerra, line 630: **suggestion**: It would also be helpful to have a test to make sure `handleOnPaste` is not called when the TextArea and TextField are disabled!

PR #2731, opened 2025-07-24
Fix TB css variable values in color foundation docs

beaesguerra:
## Summary:

Fixing the issue where the Thunderblocks hex values are not updated in the docs when you switch themes. Switching the examples showing the tokens to a proper story makes it respond correctly to the theme in Storybook

Documentation updates:
- Convert color groups usage in the Color Foundations doc to stories and include them using Canvas components
- Use the data theme attribute for getting the computed style for the css variables
- Update theme name label 

Issue: WB-2044

## Test plan:

Navigate to `?path=/docs/foundations-using-color--docs` 
- Confirm that the hex values in Classic and TB are different for `core.border` and `core.background`
- Confirm that the structure is the same as the existing Color Foundations docs 

Confirm there are Chromatic snapshots for the colors in OG and TB


__docs__/components/tokens-util.ts
```diff
@@ -84,7 +84,8 @@ export function maybeGetCssVariableInfo(value: string) {
     if (value.startsWith("var(--")) {
         const cssVariable = value.slice(4, -1);
         const rawValue = getComputedStyle(
-            document.documentElement,
+            document.querySelector("[data-wb-theme]") ||
```
beaesguerra, line 87: This is so we get the computed styles from the element that has the theme data attribute. It defaults to `document.documentElement` in case it's not there

__docs__/foundations-color.mdx
```diff
@@ -88,27 +89,15 @@ would use `core.border.neutral.subtle`, `core.border.neutral.default` for when
 3:1 contrast is a priority (ex. form elements) and `core.border.neutral.strong`
 for 4.5:1 contrast.
 
-<View style={styles.gridCompact}>
-    <ActionColorGroup
```
beaesguerra, line 92: I found that the React components were not properly reactive when embedded in a `.mdx` file 🥲  I moved the jsx to a story and reference that story using a `Canvas`. 

Wanted to get more feedback on this approach before doing the same for the rest of the semantic color tokens.

I think the token tables will also need something similar so that the values can be updated when the theme changes

__docs__/wonder-blocks-tokens/semantic-color.stories.tsx
```diff
@@ -0,0 +1,106 @@
...
+            <ActionColorGroup
+                category={category}
+                group={group}
+                includeExample={false}
+            />
+        </View>
+    );
+};
+
+export const SemanticColors = {
```
beaesguerra, line 35: This story puts everything together too. Enabled chromatic for this story so we can have visual regression tests for when semantic colors change (I believe we don't have any test coverage on this right now, other than when it changes and it's used in other components)!

__docs__/components/color.tsx
```diff
@@ -27,7 +27,7 @@ type Props = {
     /**
      * The type of color group to display.
      */
-    variant: Variant;
+    variant?: Variant;
```
beaesguerra, line 30: We provide a default value so this can be optional!

__docs__/wonder-blocks-tokens/semantic-color.stories.tsx
```diff
@@ -0,0 +1,200 @@
...
+                thunderblocks: allModes.themeThunderBlocks,
+            },
+        },
+    },
+};
+
+/**
+ * All the semantic colors for visual regression testing.
+ */
+export const SemanticColors = () => {
```
beaesguerra: This story puts everything together too. Enabled chromatic for this story so we can have visual regression tests for when semantic colors change (I believe we don't have any test coverage on this right now, other than when it changes and it's used in other components)!

.storybook/preview.tsx
```diff
@@ -232,12 +232,12 @@ const preview: Preview = {
                     {
                         value: "default",
                         icon: "circlehollow",
-                        title: "Wonder Blocks (default)",
+                        title: "Legacy (Classic Wonder Blocks)",
```
beaesguerra: Aligning the theme labels with the tokens doc Caitlyn is working on. I've let the underlying ids the name! Do we want to update the underlying theme ids too? (This can be a later thing, I can create a ticket for it!)

jandrade: Good point! I missed this in the doc. I asked a question about this: https://khanacademy.atlassian.net/wiki/spaces/WB/pages/3683221530/Color?focusedCommentId=4230512641

I'm not sure if we should call it `Legacy` at this point (maybe too early), but I'd love to hear your thoughts about this (and Caitlyn's as well).

__docs__/components/tokens-util.ts
jandrade: **praise:** this is so simple and right to the point 🥇

__docs__/foundations-color.mdx
jandrade: This makes total sense! thanks for improving it :)

__docs__/wonder-blocks-tokens/semantic-color.stories.tsx
jandrade: Thanks for this! I left a comment in Chromatic related to this. Let me know what you think.

.storybook/preview.tsx
beaesguerra: I changed the `title` back to what it was until we clarify the naming of the default WB theme!

__docs__/wonder-blocks-tokens/semantic-color.stories.tsx
beaesguerra: Removed this story! I'll add chromatic tests for the token tables once they're moved to stories. Thanks for the feedback!

jandrade (approved):
Looks great! thanks for fixing this. This will be super helpful for engineers :) 🚀

PR #2732, opened 2025-07-24
[WB-2007] Dropdown: Align SelectOpener with the last Figma designs

jandrade:
## Summary:

- Updates `SelectOpener` to match design specs. This means that the opener now
shares styling with the other form field components like `TextField`,
`TextArea`, etc.
- Simplifies the style generation logic by removing the dynamic style generation
function and moving that to a static object.

Issue: https://khanacademy.atlassian.net/browse/WB-2007

## Test plan:

Navigate to the dropdown docs in Storybook and verify that the opener looks as
expected.

- MultiSelect: `/?path=/docs/packages-dropdown-multiselect--docs`
- SingleSelect: `/?path=/docs/packages-dropdown-singleselect--docs`

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/select-opener.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/select-opener.tsx
@@ -143,17 +143,17 @@ export default class SelectOpener extends React.Component<
             ...sharedProps
         } = this.props;
 
-        const stateStyles = _generateStyles(isPlaceholder, error);
-
         const iconColor = disabled
             ? semanticColor.core.foreground.disabled.default
             : theme.opener.color.icon;
 
         const style = [
             styles.shared,
-            stateStyles.default,
-            disabled && stateStyles.disabled,
-            !disabled && this.state.pressed && stateStyles.press,
+            styles.default,
+            disabled && styles.disabled,
+            error && styles.error,
+            isPlaceholder && styles.placeholder,
+            !disabled && this.state.pressed && styles.press,
         ];
 
         return (
@@ -196,6 +196,12 @@ export default class SelectOpener extends React.Component<
     }
 }
 
+const pressStyling = {
+    // Use box shadow to make the border in the press state look thicker
+    // without changing the border
+    boxShadow: `0 0 0 ${border.width.thin} ${semanticColor.input.default.border}`,
+};
+
 const styles = StyleSheet.create({
     // TODO: Dedupe with Button styles
     shared: {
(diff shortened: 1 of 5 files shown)


packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -231,99 +237,45 @@ const styles = StyleSheet.create({
...
-        boxShadow: `inset 0 0 0 ${border.width.thin} ${action.press.border}`,
-        borderRadius: theme.opener.border.radius.press,
-    };
-
-    const currentState = error
-        ? semanticColor.input.error
-        : semanticColor.input.default;
-
-    const disabledStatesStyles = {
+    /**
```
jandrade, line 240: **note:** I based most of the following changes on the `TextField` styles to have general consistency with fields.

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -143,17 +143,17 @@ export default class SelectOpener extends React.Component<
...
 
         const style = [
             styles.shared,
-            stateStyles.default,
-            disabled && stateStyles.disabled,
-            !disabled && this.state.pressed && stateStyles.press,
+            styles.default,
+            disabled && styles.disabled,
+            error && styles.error,
+            isPlaceholder && styles.placeholder,
```
jandrade, line 155: **note:** Added this here now that the `styles` object is statically generated (instead of dynamic).

packages/wonder-blocks-dropdown/src/theme/default.ts
```diff
@@ -27,7 +27,6 @@ export default {
             },
             radius: {
                 rest: border.radius.radius_040,
-                press: border.radius.radius_040,
```
jandrade, line 30: **note:** This is no longer used.

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -231,99 +235,54 @@ const styles = StyleSheet.create({
...
-
-    const disabledStatesStyles = {
+    /**
+     * Theming
+     */
+    default: {
+        background: semanticColor.input.default.background,
+        border: `${border.width.thin} solid ${semanticColor.input.default.border}`,
+        color: semanticColor.input.default.foreground,
+        "::placeholder": {
```
beaesguerra: Since the select opener is a button instead of an `input` element, the `::placeholder` styles don't get applied! 

For example, in the disabled state, the placeholder is still using input placeholder default token, instead of the disabled token 
Image

They're the same value anyways but thought I'd mention it! Applies to other places in the file that uses `::placeholder`!

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -143,17 +143,17 @@ export default class SelectOpener extends React.Component<
...
             ? semanticColor.core.foreground.disabled.default
             : theme.opener.color.icon;
 
         const style = [
             styles.shared,
-            stateStyles.default,
-            disabled && stateStyles.disabled,
-            !disabled && this.state.pressed && stateStyles.press,
+            styles.default,
+            disabled && styles.disabled,
```
beaesguerra, line 153: For consistency, the selects also have a [read only state](https://www.figma.com/design/EuFu0U7gqc1koXm8ZhlOLp/%E2%9A%A1%EF%B8%8F-Components?node-id=9492-1546&t=3aHtZciAj58GVKsM-4) 

We can create a separate ticket for this later on if you prefer!

beaesguerra (approved):
Looks good! Left a few comments on things I noticed, nothing blocking though :)

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
jandrade: ohh good point! let me fix it.

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
jandrade: I created https://khanacademy.atlassian.net/browse/WB-2046 to track that for later.

PR #2734, opened 2025-07-25
Fix mobile hover reset for `Button`

timmcca-be:
## Summary:
Rather than (incorrectly) undoing the hover effect if the current device doesn't support hovering, this PR flips it to only apply the hover effect if the current device does support hovering.

I also improved the TypeScript coverage, which uncovered a lot of references to styles that are never actually defined. I cleaned those up too.

Question: is it intended that we're now using CSS to detect hover/focus states, instead of the ones from `Clickable` that get passed in from props? Before this change, the code was trying to do both, but since the expected styles were never defined, the props had no effect. This PR just makes that explicit.

Issue: AX-1550

## Test plan:
- Start Storybook
- Go to `http://localhost:6061/?path=/story/packages-button-button--kinds` in Chrome
- Enable touch simulation
- Press a primary button
- Verify that it doesn't disappear

### DIFF
--- a/.changeset/famous-dolphins-protect.md
+++ b/.changeset/famous-dolphins-protect.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-button": patch
+---
+
+Fix mobile hover reset for Button


packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -61,9 +59,7 @@ const ButtonCore: React.ForwardRefExoticComponent<
         endIcon && sharedStyles.withEndIcon,
         buttonStyles.default,
         disabled && buttonStyles.disabled,
-        // apply focus effect only to default and secondary buttons
-        !disabled &&
-            (pressed ? buttonStyles.pressed : focused && buttonStyles.focused),
```
timmcca-be, line 66: `buttonStyles.focused` is not defined

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -78,12 +74,6 @@ const ButtonCore: React.ForwardRefExoticComponent<
                 size === "large" && sharedStyles.largeText,
                 labelStyle,
                 spinner && sharedStyles.hiddenText,
-                // apply press/hover effects on the label
-                kind === "tertiary" &&
-                    !disabled &&
-                    (pressed
-                        ? [buttonStyles.hover, buttonStyles.active]
-                        : hovered && buttonStyles.hover),
```
timmcca-be, line 86: `buttonStyles.hover` and `buttonStyles.active` are not defined. ~~Maybe it was meant to be `buttonStyles.pressed` instead of active?~~ I tested the `pressed` style on the label and it looked very wrong! I think removing this outright is correct

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -136,7 +126,7 @@ const ButtonCore: React.ForwardRefExoticComponent<
                 <View
                     testId={testId ? `${testId}-end-icon-wrapper` : undefined}
                     style={[
-                        styles.endIcon,
+                        sharedStyles.endIcon,
```
timmcca-be, line 129: Fixing types also brought this to light

PR #2737, opened 2025-07-29
ButtonCore: Fix passing in invalid attributes

beaesguerra:
## Summary:

The recent button changes in https://github.com/Khan/wonder-blocks/pull/2734/files#diff-ede6855e27a1ecfe88d6f2e68ba1979e36ea896776f7e023078b353df3b84b94L36-L37 would cause tests to fail in `frontend` due to unexpected `console.error` calls.

This was because `hovered` and `focused` props were being included with `restProps`, which is applied to the underlying button element. These aren't valid html attributes so an error was logged.

To fix this, we go back to pulling out the `focused` and `hovered` props from `restProps`. Since they are unused, we disable the unused variable lint rules for these lines

Note: We continue to have `focused` and `hovered` props because `Button` still uses `ClickableBehaviour` and it passes in state information. Long term we want to move away from using `ClickableBehaviour`, but rely on it in some places to handle browsers inconsistencies ([more details about ClickableBehaviour](https://github.com/Khan/wonder-blocks/pull/2244#discussion_r1628440285))

Issue: WB-2051

## Test plan:

Test that using the npm snapshot for this PR in frontend results in no unit test failures

### DIFF
--- a/.changeset/healthy-crabs-refuse.md
+++ b/.changeset/healthy-crabs-refuse.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-button": patch
+---
+
+ButtonCore: Prevent `focused` and `hovered` props from being applied to the underlying button element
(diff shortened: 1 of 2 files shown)


beaesguerra:
Thanks @marcysutton! I also updated the PR description to include why we keep the props due to the use of `ClickableBehaviour` 😄

beaesguerra:
All the checks are passing in the `frontend` integration PR: https://github.com/Khan/frontend/pull/2406 ! Will land and get the button updates ready for release!

PR #2743, opened 2025-08-01
Add theming to Link, use BodyText in Button

marcysutton:
## Summary:
This PR addresses font theming in WB Link, and updates internal typography usage in Button. 

Longer version: I investigated why Link wasn't using the themed font by comparing it to WB Button. Button used Label typography styles to wrap its text content, which set a themed font on a child `span`. I updated Button to use BodyText. Rather than wrapping Link's contents in an extra span, I elected to add a `font` declaration and account for font weight differences with component tokens.

Issue: WB-2052

## Test plan:
1. Verify Link has Lato font family and regular weight in OG
2. Verify Link has Jakarta font family and bold weight in TB
3. Verify Button has no visual changes in either theme, including color
4. Check font-size scaling for larger sizes in chrome://settings > Appearance > Font Size

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -22,7 +22,7 @@
 <style>
     body {
         font-family: "Lato", sans-serif;
-        font-size: 16px!important;
+        font-size: 1.6rem!important;
     }
 
     /* Overrides the ArgTypes table tabs to avoid some z-index issues with
--- a/__docs__/wonder-blocks-link/link-testing-snapshots.stories.tsx
+++ b/__docs__/wonder-blocks-link/link-testing-snapshots.stories.tsx
@@ -22,7 +22,7 @@ const generateCells = (rtl: boolean = false) => [
     {
         name: "Primary",
         props: {
-            children: rtl ? rtlText : "This is a Link",
+            children: rtl ? rtlText : "This is my Link",
         },
     },
     {
@@ -61,10 +61,6 @@ type Story = StoryObj<typeof Link>;
  */
 const meta = {
     title: "Packages / Link / Testing / Snapshots / Link",
-    args: {
-        children: "This is a Link",
-        href: "https://www.khanacademy.org",
-    },
     parameters: {
         chromatic: {
             modes: {
--- a/__docs__/wonder-blocks-pill/pill.stories.tsx
+++ b/__docs__/wonder-blocks-pill/pill.stories.tsx
@@ -213,8 +213,13 @@ export const Variants: StoryComponentType = {
 export const WithTypography: StoryComponentType = () => (
     <Pill size="large">
         <BodySerif>
-            This is a {<Link href="#link">link example</Link>} inside the text
-            of a pill.
+            This is a{" "}
+            {
+                <Link href="#link" style={{fontFamily: "inherit"}}>
+                    link example
+                </Link>
+            }{" "}
+            inside the text of a pill.
         </BodySerif>
     </Pill>
 );
--- a/__docs__/wonder-blocks-pill/pill.stories.tsx
+++ b/__docs__/wonder-blocks-pill/pill.stories.tsx
@@ -213,13 +213,8 @@ export const Variants: StoryComponentType = {
 export const WithTypography: StoryComponentType = () => (
     <Pill size="large">
         <BodySerif>
-            This is a{" "}
-            {
-                <Link href="#link" style={{fontFamily: "inherit"}}>
-                    link example
-                </Link>
-            }{" "}
-            inside the text of a pill.
+            This is a {<Link href="#link">link example</Link>} inside the text
+            of a pill.
         </BodySerif>
     </Pill>
 );
--- a/packages/wonder-blocks-icon/src/components/phosphor-icon.tsx
+++ b/packages/wonder-blocks-icon/src/components/phosphor-icon.tsx
@@ -3,7 +3,7 @@ import {StyleSheet} from "aphrodite";
 
 import {addStyle, AriaProps, StyleType} from "@khanacademy/wonder-blocks-core";
 
-import {viewportPixelsForSize, viewportRemsForSize} from "../util/icon-util";
+import {viewportPixelsForSize} from "../util/icon-util";
 import {IconSize, PhosphorIconAsset} from "../types";
 
 // We use a span instead of an img because we want to use the mask-image CSS
@@ -52,14 +52,6 @@ type Props = Pick<AriaProps, "aria-hidden" | "aria-label" | "role"> & {
      * - `string`: an import referencing an arbitrary SVG file.
      */
     icon: PhosphorIconAsset | string;
-    /**
-     * Optional sizing unit for width and height: pixels or REMs.
-     *
-     * Defaults to px for backwards compatibility.
-     *
-     * A REM unit will scale with the user's font-size, so it is preferred.
-     */
-    sizingUnit?: "px" | "rem";
 };
 
 /**
@@ -99,16 +91,12 @@ export const PhosphorIcon = React.forwardRef(function PhosphorIcon(
         testId,
         className,
         role,
-        sizingUnit = "px",
         ...sharedProps
     } = props;
 
-    const iconSize =
-        sizingUnit === "rem"
-            ? viewportRemsForSize(size)
-            : viewportPixelsForSize(size);
+    const pixelSize = viewportPixelsForSize(size);
     const classNames = `${className ?? ""}`;
-    const iconStyles = _generateStyles(color, iconSize);
+    const iconStyles = _generateStyles(color, pixelSize);
 
     return (
         <StyledSpan
@@ -137,7 +125,7 @@ const dynamicStyles: Record<string, any> = {};
 /**
  * Generates the visual styles for the icon.
  */
-const _generateStyles = (color: string, size: string | number) => {
+const _generateStyles = (color: string, size: number) => {
     const iconStyle = `${color}-${size}`;
     // The styles are cached to avoid creating a new object on every render.
     if (styles[iconStyle]) {
--- a/packages/wonder-blocks-icon/src/components/phosphor-icon.tsx
+++ b/packages/wonder-blocks-icon/src/components/phosphor-icon.tsx
@@ -57,7 +57,7 @@ type Props = Pick<AriaProps, "aria-hidden" | "aria-label" | "role"> & {
      *
      * Defaults to px for backwards compatibility.
      *
-     * REM units will scale with the user's font-size, and it is preferred.
+     * A REM unit will scale with the user's font-size, so it is preferred.
      */
     sizingUnit?: "px" | "rem";
 };
--- a/packages/wonder-blocks-icon/src/components/phosphor-icon.tsx
+++ b/packages/wonder-blocks-icon/src/components/phosphor-icon.tsx
@@ -3,7 +3,7 @@ import {StyleSheet} from "aphrodite";
 
 import {addStyle, AriaProps, StyleType} from "@khanacademy/wonder-blocks-core";
 
-import {viewportPixelsForSize} from "../util/icon-util";
+import {viewportPixelsForSize, viewportRemsForSize} from "../util/icon-util";
 import {IconSize, PhosphorIconAsset} from "../types";
 
 // We use a span instead of an img because we want to use the mask-image CSS
@@ -52,6 +52,14 @@ type Props = Pick<AriaProps, "aria-hidden" | "aria-label" | "role"> & {
      * - `string`: an import referencing an arbitrary SVG file.
      */
     icon: PhosphorIconAsset | string;
+    /**
+     * Optional sizing unit for width and height: pixels or REMs.
+     *
+     * Defaults to px for backwards compatibility.
+     *
+     * REM units will scale with the user's font-size, and it is preferred.
+     */
+    sizingUnit?: "px" | "rem";
 };
 
 /**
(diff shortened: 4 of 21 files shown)


__docs__/wonder-blocks-pill/pill.stories.tsx
```diff
@@ -213,8 +213,13 @@ export const Variants: StoryComponentType = {
 export const WithTypography: StoryComponentType = () => (
     <Pill size="large">
         <BodySerif>
-            This is a {<Link href="#link">link example</Link>} inside the text
-            of a pill.
+            This is a{" "}
+            {
+                <Link href="#link" style={{fontFamily: "inherit"}}>
```
beaesguerra: Would it make sense to set the `fontFamily` for `Link` in the Classic theme to `inherit` so that all existing Links don't need to be updated to inherit font properties? Or is this more a one off? I'm wondering if there are similar use cases that exist across the platform!

packages/wonder-blocks-link/src/theme/default.ts
```diff
@@ -1,9 +1,12 @@
...
+import {border} from "@khanacademy/wonder-blocks-tokens";
 
 export default {
     root: {
         font: {
-            weight: font.weight.regular,
+            weight: "inherit",
+        },
+        border: {
+            offset: border.width.medium,
```
beaesguerra: Would a sizing token be appropriate for the `offset` token? Since the border width token is being used for something other than the border width! The sizing token would be a rem value though 🤔 Curious to hear what you think!

packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -10,6 +10,7 @@ import {
     semanticColor,
     border,
 } from "@khanacademy/wonder-blocks-tokens";
+import {focusStyles} from "@khanacademy/wonder-blocks-styles";
```
beaesguerra, line 13: We'll need to include the styles package as a dependency in the link package, and update the `tsconfig-build.json` file!

beaesguerra (commented):
Thanks for making the styling updates! Left a few more comments in the PR and Chromatic to get your thoughts on 😄

__docs__/wonder-blocks-pill/pill.stories.tsx
marcysutton: Yes, it worked! I think making it automatic in Classic makes sense. TB won't have that capability, though!

packages/wonder-blocks-tokens/src/theme/primitive/border.ts
```diff
@@ -24,4 +24,8 @@ export const border = {
         medium: remToPx(sizing.size_020),
         thick: remToPx(sizing.size_040),
     },
+    offset: {
```
beaesguerra: Do you think the border offset token would be helpful as design tokens too in Figma? If so, this is something we can coordinate with Caitlyn on!

Also, do we want to document the new border offset tokens? Or use it internally only for now? (can be done separately!)

beaesguerra: Also, since we use it for setting the text underline offset, I'm wondering if the underline offset token should be a token related to `font`, rather than `border`! Borders don't really have offsets, but they could apply to outlines! What do you think?

beaesguerra (commented):
Wanted to get your thoughts on the new border offset tokens! There's also one remaining comment in Chromatic about overlapping focus outlines!

packages/wonder-blocks-tokens/src/theme/primitive/border.ts
marcysutton: I like the idea of a `font.textDecoration.offset` token or similar! I floated the idea with Caitlyn in Slack. I'll hold off to document them until we know the API for sure.

.storybook/preview-head.html
```diff
@@ -22,7 +22,7 @@
 <style>
     body {
         font-family: "Lato", sans-serif;
-        font-size: 16px!important;
+        font-size: 1.6rem!important;
```
marcysutton, line 25: The PX unit was preventing font scaling with REMs in Links! Surprised I only noticed it now.

__docs__/wonder-blocks-link/link-testing-snapshots.stories.tsx
```diff
@@ -61,10 +61,6 @@ type Story = StoryObj<typeof Link>;
  */
 const meta = {
     title: "Packages / Link / Testing / Snapshots / Link",
-    args: {
```
marcysutton, line 64: I noticed these args were unused when I added a descender to the Primary cell content. Rather than confusing anyone else, I opted to remove unused args.

packages/wonder-blocks-icon/src/components/phosphor-icon.tsx
```diff
@@ -52,6 +52,14 @@ type Props = Pick<AriaProps, "aria-hidden" | "aria-label" | "role"> & {
...
      */
     icon: PhosphorIconAsset | string;
+    /**
+     * Optional sizing unit for width and height: pixels or REMs.
+     *
+     * Defaults to px for backwards compatibility.
+     *
+     * A REM unit will scale with the user's font-size, so it is preferred.
+     */
+    sizingUnit?: "px" | "rem";
```
beaesguerra: Are there any cases where an icon should be in `px` instead of `rem`? If not, I think all icons can be in `rem` and we won't need this prop! That can be handled separately though since switching to rems may require more testing for existing use cases !

packages/wonder-blocks-icon/src/components/phosphor-icon.tsx
```diff
@@ -91,12 +99,16 @@ export const PhosphorIcon = React.forwardRef(function PhosphorIcon(
         testId,
         className,
         role,
+        sizingUnit = "px",
         ...sharedProps
     } = props;
 
-    const pixelSize = viewportPixelsForSize(size);
+    const iconSize =
```
beaesguerra: On another note, there is also the `Icon` component in the icon package for custom icons. It's using `sizing` tokens so it is using rems already so that's handled already!

packages/wonder-blocks-link/src/theme/default.ts
```diff
@@ -1,13 +1,14 @@
...
         font: {
             family: "inherit",
             weight: "inherit",
-        },
-        border: {
-            offset: border.offset.medium,
+            textDecoration: {
+                thickness: font.textDecoration.thickness,
+                offset: font.textDecoration.offset,
+            },
```
beaesguerra: Since the thickness and offset values are the same in default and TB, we can use the global tokens directly in `Link` without having component themed tokens! This will reduce the number of component css variables we have!

packages/wonder-blocks-tokens/src/theme/primitive/font.ts
```diff
@@ -78,4 +78,8 @@ export const font = {
             xxlarge: sizing.size_400, // Title
         },
     },
+    textDecoration: {
```
beaesguerra, line 81: Nice! Thanks for sorting out the new tokens! 😄 

Do we want to include the term `underline` somewhere in the token name? Related to [Caitlyn's comment](https://khanacademy.slack.com/archives/C07CA2YR0BZ/p1755005715448969?thread_ts=1754955327.064349&cid=C07CA2YR0BZ)

packages/wonder-blocks-tokens/src/theme/primitive/border.ts
```diff
@@ -12,6 +12,7 @@ export const border = {
      */
     radius: {
         radius_0: remToPx(sizing.size_0),
+        radius_010: remToPx(sizing.size_010),
```
beaesguerra, line 15: Do we want an equivalent in Figma for radius_010? 

Image

packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -217,14 +220,14 @@ const _generateStyles = (inline: boolean, light: boolean) => {
...
-        borderRadius: border.radius.radius_040,
+        borderRadius: border.radius.radius_010,
         outlineOffset: border.width.medium,
     };
 
     const pressStyling = {
         color: variant.press.foreground,
         textDecoration: "underline currentcolor solid",
-        textUnderlineOffset: theme.root.border.offset,
+        textUnderlineOffset: theme.root.font.textDecoration.offset,
```
beaesguerra: The text underline properties can use the new text decoration global tokens directly since it's the same in both themes!

packages/wonder-blocks-tokens/src/theme/primitive/font.ts
```diff
@@ -78,4 +78,8 @@ export const font = {
...
+        offset: sizing.size_020,
+        thickness: sizing.size_010,
+    },
```
beaesguerra, line 84: It would be helpful to have a separate changeset for the `tokens` package for the newly added text decoration and border radius tokens so it gets included in the changelog 😄

beaesguerra (commented):
Thanks for iterating on this and sorting out the new text decoration tokens! I had a few more questions about the new updates!

packages/wonder-blocks-link/src/theme/default.ts
marcysutton: Oh yes, good call!

packages/wonder-blocks-tokens/src/theme/primitive/border.ts
marcysutton: I'm seeing that token already in the Foundations Figma file:
Screenshot 2025-08-12 at 1 54 52 PM

marcysutton: I see the Figma file has a token from `sizing`, I posed it to Caitlyn since I'm a read-only consumer of the Figma files!

packages/wonder-blocks-tokens/src/theme/primitive/font.ts
marcysutton: There is a changelog already for `wonder-blocks-tokens` - I can add a mention about the textDecoration stuff. Should it be a totally separate changeset entry?

packages/wonder-blocks-tokens/src/theme/primitive/font.ts
marcysutton: Yes, updated!

packages/wonder-blocks-icon/src/components/phosphor-icon.tsx
marcysutton: I'm not sure. I mostly didn't want this to snowball into a larger change since the original scope of work was related to font theming, not icon sizing. If it didn't look so bad with font scaling inside of Links I would have tackled it separately!

marcysutton: Okay, I think I will opt to pull this change out into its own PR so we can keep the font work moving and not deal with the scope creep. Thank you for asking about it! Here's an issue for PhosphorIcon sizing: https://khanacademy.atlassian.net/browse/WB-2057

packages/wonder-blocks-tokens/src/theme/primitive/font.ts
beaesguerra: I think a separate one makes sense! 

The changeset description is going to show up in each of the packages changelog! So adding something like "Added new border radius and text decoration tokens" to the existing changeset file will get added to the `link` and `breadcrumbs` packages too, which is less relavent!

https://github.com/Khan/wonder-blocks/blob/2290ea4afd21e39dbda69a517d6ac3272c48afb0/.changeset/lazy-dolphins-stare.md?plain=1#L1-L7

I've been thinking more about our changeset descriptions and changelogs since I think they will probably get utilized more now that other teams (ie. assessments) will be managing their own WB dependencies later on.

Let me know if you have any thoughts on this!

packages/wonder-blocks-tokens/src/theme/primitive/border.ts
beaesguerra: If you have dev mode, you should be able to see the tokens interface too! https://www.figma.com/design/9seilz0fBgo3QUj6eXM0yR/%E2%9A%A1%EF%B8%8F-Foundations?node-id=0-1&p=f&vars=1&m=dev

packages/wonder-blocks-tokens/src/theme/primitive/font.ts
```diff
@@ -78,4 +78,8 @@ export const font = {
             xxlarge: sizing.size_400, // Title
         },
     },
+    textDecoration: {
+        underlineOffset: sizing.size_020,
+        thickness: sizing.size_010,
```
beaesguerra, line 83: Do we want this to be `underlineThickness` to match `underlineOffset`?

beaesguerra (approved):
Looks good! Thanks for iterating on this and making sure we have things aligned with design 🚀 There are a few suggestions remaining but none are blocking!

packages/wonder-blocks-tokens/src/theme/primitive/font.ts
marcysutton: I considered it, but decided to keep them as-is based on the actual CSS property names! The thickness property is `textDecorationThickness`, so it seemed more intuitive to leave it that way. Here is a screenshot:

screenshot of restInline styles

packages/wonder-blocks-tokens/src/theme/primitive/font.ts
marcysutton: That totally makes sense. I separated the entries, thanks for suggesting it!

PR #2744, opened 2025-08-04
Create new Modal DrawerLauncher component

marcysutton:
## Summary:

Creates new `DrawerLauncher` and `DrawerDialog` components with three alignments: `inlineStart`, `inlineEnd`, and `blockEnd`. Ensures dialogs can be used with RTL writing mode.

Issue: WB-2048

## Test plan:
1. Ensure FlexibleDialog can be used standalone
2. Ensure DrawerLauncher with DrawerDialog can be positioned on left, right, and bottom
3. Ensure DrawerLauncher can be used with RTL writing mode
4. Ensure tests pass

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -179,15 +179,7 @@ const withLanguageDirection: Decorator = (Story, context) => {
         };
     }, [context.globals.direction]);
 
-    if (context.globals.direction === "rtl") {
-        return (
-            <div dir="rtl">
-                <Story />
-            </div>
-        );
-    } else {
-        return <Story />;
-    }
+    return <Story />;
 };
 
 /**
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -179,7 +179,13 @@ const withLanguageDirection: Decorator = (Story, context) => {
         };
     }, [context.globals.direction]);
 
-    return <div key={context.globals.direction}><Story /></div>;
+    if (context.globals.direction === "rtl") {
+        return (
+            <div dir="rtl">
+                <Story />
+            </div>);
+    }
+    return <Story />;
 };
 
 /**
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -179,7 +179,7 @@ const withLanguageDirection: Decorator = (Story, context) => {
         };
     }, [context.globals.direction]);
 
-    return <Story />;
+    return <div key={context.globals.direction}><Story /></div>;
 };
 
 /**
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -162,9 +162,23 @@ const withThemeSwitcher: Decorator = (
 };
 
 /**
- * Wraps a story with `<div dir="rtl">` so it is shown in rtl mode.
+ * Sets the dir attribute on document.documentElement and wraps story with dir div.
+ * This ensures portaled content (like modals) can detect the direction.
  */
 const withLanguageDirection: Decorator = (Story, context) => {
+    React.useEffect(() => {
+        if (context.globals.direction === "rtl") {
+            document.documentElement.setAttribute("dir", "rtl");
+        } else {
+            document.documentElement.setAttribute("dir", "ltr");
+        }
+
+        // Cleanup on unmount
+        return () => {
+            document.documentElement.removeAttribute("dir");
+        };
+    }, [context.globals.direction]);
+
     if (context.globals.direction === "rtl") {
         return (
             <div dir="rtl">
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -183,9 +183,11 @@ const withLanguageDirection: Decorator = (Story, context) => {
         return (
             <div dir="rtl">
                 <Story />
-            </div>);
+            </div>
+        );
+    } else {
+        return <Story />;
     }
-    return <Story />;
 };
 
 /**
--- a/__docs__/wonder-blocks-modal/drawer-dialog.stories.tsx
+++ b/__docs__/wonder-blocks-modal/drawer-dialog.stories.tsx
@@ -1,0 +1,404 @@
+import * as React from "react";
+import {StyleSheet} from "aphrodite";
+import type {Meta, StoryObj} from "@storybook/react";
+
+import Button from "@khanacademy/wonder-blocks-button";
+import {View} from "@khanacademy/wonder-blocks-core";
+import {ActionMenu, ActionItem} from "@khanacademy/wonder-blocks-dropdown";
+import {RadioGroup, Choice} from "@khanacademy/wonder-blocks-form";
+import {LabeledField} from "@khanacademy/wonder-blocks-labeled-field";
+import {spacing} from "@khanacademy/wonder-blocks-tokens";
+import {BodyText, Title} from "@khanacademy/wonder-blocks-typography";
+
+import {DrawerDialog, DrawerLauncher} from "@khanacademy/wonder-blocks-modal";
+import packageConfig from "../../packages/wonder-blocks-modal/package.json";
+
+import ComponentInfo from "../components/component-info";
+import {allModes} from "../../.storybook/modes";
+import {reallyLongText} from "../components/text-for-testing";
+import DrawerDialogArgTypes from "./drawer-dialog.argtypes";
+
+/**
+ * `DrawerDialog` is the modal content component designed specifically for use with `DrawerLauncher`.
+ * It provides a consistent drawer interface with proper animations, positioning, and accessibility features.
+ *
+ * **IMPORTANT**: This component should only be used with `DrawerLauncher`. Using it with other
+ * modal launchers may result in incorrect animations, positioning, and styling.
+ *
+ * The component automatically receives alignment, animation, and timing props from `DrawerLauncher`
+ * via React Context, eliminating the need for manual prop passing in nested components.
+ */
+export default {
+    title: "Packages / Modal / DrawerDialog",
+    component: DrawerDialog,
+    parameters: {
+        componentSubtitle: "Modal dialog designed for DrawerLauncher",
+        docs: {
+            description: {
+                component:
+                    "A drawer modal dialog that slides in from the specified alignment (inline start/end or block end). Automatically integrates with DrawerLauncher for proper animations and positioning.",
+            },
+        },
+        chromatic: {
+            // Drawer stories are behavior-based, disable visual regression testing
+            disableSnapshot: true,
+        },
+    },
+    argTypes: DrawerDialogArgTypes,
+} satisfies Meta<typeof DrawerDialog>;
+
+type StoryComponentType = StoryObj<typeof DrawerDialog>;
+
+const styles = StyleSheet.create({
+    content: {
+        padding: spacing.medium_16,
+    },
+    section: {
+        marginBottom: spacing.medium_16,
+    },
+    form: {
+        display: "flex",
+        flexDirection: "column",
+        gap: spacing.medium_16,
+    },
+});
+
+/**
+ * Basic drawer dialog with simple content. This shows the default structure
+ * and styling of a drawer dialog.
+ */
+export const Default: StoryComponentType = {
+    render: () => (
+        <DrawerLauncher
+            alignment="inlineEnd"
+            modal={
+                <DrawerDialog
+                    title="Default Drawer"
+                    content={
+                        <View style={styles.content}>
+                            <BodyText>
+                                This is a basic drawer dialog with simple text
+                                content. The drawer slides in from the inline
+                                end (right in LTR mode).
+                            </BodyText>
+                        </View>
+                    }
+                />
+            }
+        >
+            {({openModal}) => (
+                <Button onClick={openModal}>Open Default Drawer</Button>
+            )}
+        </DrawerLauncher>
+    ),
+};
+
+/**
+ * Drawer with rich content including form elements and actions.
+ * Demonstrates how to create more complex drawer interfaces.
+ */
+export const WithFormContent: StoryComponentType = {
+    render: () => (
+        <DrawerLauncher
+            alignment="inlineStart"
+            modal={
+                <DrawerDialog
+                    title="Settings"
+                    content={
+                        <View style={styles.content}>
+                            <View style={styles.form}>
+                                <View style={styles.section}>
+                                    <BodyText size="medium">
+                                        Preferences
+                                    </BodyText>
+                                    <RadioGroup
+                                        groupName="theme"
+                                        onChange={() => {}}
+                                        selectedValue="light"
+                                    >
+                                        <Choice
+                                            label="Light theme"
+                                            value="light"
+                                        />
+                                        <Choice
+                                            label="Dark theme"
+                                            value="dark"
+                                        />
+                                        <Choice label="Auto" value="auto" />
+                                    </RadioGroup>
+                                </View>
+
+                                <View style={styles.section}>
+                                    <ActionMenu
+                                        menuText="More actions"
+                                        testId="action-menu"
+                                    >
+                                        <ActionItem label="Reset settings" />
+                                        <ActionItem label="Export data" />
+                                        <ActionItem label="Delete account" />
+                                    </ActionMenu>
+                                </View>
+                            </View>
+                        </View>
+                    }
+                />
+            }
+        >
+            {({openModal}) => (
+                <Button onClick={openModal}>Open Settings</Button>
+            )}
+        </DrawerLauncher>
+    ),
+};
+
+/**
+ * Drawer positioned at the bottom of the screen (block end).
+ * Useful for mobile-friendly interfaces and action sheets.
+ */
+export const BottomDrawer: StoryComponentType = {
+    render: () => (
+        <DrawerLauncher
+            alignment="blockEnd"
+            modal={
+                <DrawerDialog
+                    title="Actions"
+                    content={
+                        <View style={styles.content}>
+                            <BodyText style={styles.section}>
+                                Choose an action from the options below:
+                            </BodyText>
+                            <View style={styles.form}>
+                                <Button kind="primary">Primary Action</Button>
+                                <Button kind="secondary">
+                                    Secondary Action
+                                </Button>
+                                <Button kind="tertiary">Cancel</Button>
+                            </View>
+                        </View>
+                    }
+                />
+            }
+        >
+            {({openModal}) => (
+                <Button onClick={openModal}>Open Bottom Drawer</Button>
+            )}
+        </DrawerLauncher>
+    ),
+};
+
+/**
+ * Drawer using render prop pattern for content. This allows placing
+ * the title element in a custom location within the content.
+ */
+export const WithRenderProp: StoryComponentType = {
+    render: () => (
+        <DrawerLauncher
+            alignment="inlineEnd"
+            modal={
+                <DrawerDialog
+                    title="Render Prop Example"
+                    content={({title}) => (
+                        <View style={styles.content}>
+                            {title}
+                            <BodyText style={styles.section}>
+                                This content uses a render prop to customize how
+                                the title is positioned within the drawer
+                                content.
+                            </BodyText>
+                            <BodyText>
+                                The title element is passed as a prop to the
+                                render function, allowing for flexible layout
+                                arrangements.
+                            </BodyText>
+                        </View>
+                    )}
+                />
+            }
+        >
+            {({openModal}) => (
+                <Button onClick={openModal}>Open Render Prop Drawer</Button>
+            )}
+        </DrawerLauncher>
+    ),
+};
+
+/**
+ * Drawer without visible close button. Useful when the content
+ * provides its own close mechanism.
+ */
+export const NoCloseButton: StoryComponentType = {
+    render: () => (
+        <DrawerLauncher
+            alignment="inlineEnd"
+            modal={
+                <DrawerDialog
+                    title="Confirm Action"
+                    closeButtonVisible={false}
+                    content={
+                        <View style={styles.content}>
+                            <BodyText style={styles.section}>
+                                Are you sure you want to delete this item? This
+                                action cannot be undone.
+                            </BodyText>
+                            <View style={styles.form}>
+                                <Button kind="primary">Delete</Button>
+                                <Button kind="secondary">Cancel</Button>
+                            </View>
+                        </View>
+                    }
+                />
+            }
+        >
+            {({openModal}) => (
+                <Button onClick={openModal}>Open Confirmation</Button>
+            )}
+        </DrawerLauncher>
+    ),
+};
+
+/**
+ * Drawer with long scrollable content. Demonstrates how the drawer
+ * handles content overflow.
+ */
+export const WithScrollableContent: StoryComponentType = {
+    render: () => (
+        <DrawerLauncher
+            alignment="inlineStart"
+            modal={
+                <DrawerDialog
+                    title="Terms of Service"
+                    content={
+                        <View style={styles.content}>
+                            <BodyText>{reallyLongText}</BodyText>
+                            <BodyText>{reallyLongText}</BodyText>
+                            <BodyText>{reallyLongText}</BodyText>
+                        </View>
+                    }
+                />
+            }
+        >
+            {({openModal}) => (
+                <Button onClick={openModal}>Open Long Content</Button>
+            )}
+        </DrawerLauncher>
+    ),
+};
+
+/**
+ * Drawer nested inside wrapper components. Demonstrates how the
+ * context-based prop passing works with nested component structures.
+ */
+export const WithNestedComponents: StoryComponentType = {
+    render: () => {
+        const NestedWrapper = ({children}: {children: React.ReactNode}) => (
+            <View style={{padding: spacing.small_12}}>{children}</View>
+        );
+
+        return (
+            <DrawerLauncher
+                alignment="blockEnd"
+                modal={
+                    <NestedWrapper>
+                        <DrawerDialog
+                            title="Nested Dialog"
+                            content={
+                                <View style={styles.content}>
+                                    <BodyText>
+                                        This drawer is nested inside wrapper
+                                        components but still receives the proper
+                                        alignment and animation props via React
+                                        Context.
+                                    </BodyText>
+                                </View>
+                            }
+                        />
+                    </NestedWrapper>
+                }
+            >
+                {({openModal}) => (
+                    <Button onClick={openModal}>Open Nested Drawer</Button>
+                )}
+            </DrawerLauncher>
+        );
+    },
+};
+
+/**
+ * Shows all drawer alignment options side by side for comparison.
+ */
+export const AllAlignments: StoryComponentType = {
+    parameters: {
+        chromatic: {
+            disableSnapshot: false,
+            modes: {
+                large: allModes.large,
+                medium: allModes.medium,
+            },
+        },
+    },
+    render: () => (
+        <View
+            style={{display: "flex", gap: spacing.medium_16, flexWrap: "wrap"}}
+        >
+            <DrawerLauncher
+                alignment="inlineStart"
+                modal={
+                    <DrawerDialog
+                        title="Inline Start"
+                        content={
+                            <View style={styles.content}>
+                                <BodyText>
+                                    Slides in from inline start (left in LTR)
+                                </BodyText>
+                            </View>
+                        }
+                    />
+                }
+            >
+                {({openModal}) => (
+                    <Button onClick={openModal}>Inline Start</Button>
+                )}
+            </DrawerLauncher>
+
+            <DrawerLauncher
+                alignment="inlineEnd"
+                modal={
+                    <DrawerDialog
+                        title="Inline End"
+                        content={
+                            <View style={styles.content}>
+                                <BodyText>
+                                    Slides in from inline end (right in LTR)
+                                </BodyText>
+                            </View>
+                        }
+                    />
+                }
+            >
+                {({openModal}) => (
+                    <Button onClick={openModal}>Inline End</Button>
+                )}
+            </DrawerLauncher>
+
+            <DrawerLauncher
+                alignment="blockEnd"
+                modal={
+                    <DrawerDialog
+                        title="Block End"
+                        content={
+                            <View style={styles.content}>
+                                <BodyText>
+                                    Slides in from block end (bottom)
+                                </BodyText>
+                            </View>
+                        }
+                    />
+                }
+            >
+                {({openModal}) => (
+                    <Button onClick={openModal}>Block End</Button>
+                )}
+            </DrawerLauncher>
+        </View>
+    ),
+};
--- a/__docs__/wonder-blocks-modal/drawer-dialog.stories.tsx
+++ b/__docs__/wonder-blocks-modal/drawer-dialog.stories.tsx
@@ -6,14 +6,11 @@ import Button from "@khanacademy/wonder-blocks-button";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {ActionMenu, ActionItem} from "@khanacademy/wonder-blocks-dropdown";
 import {RadioGroup, Choice} from "@khanacademy/wonder-blocks-form";
-import {LabeledField} from "@khanacademy/wonder-blocks-labeled-field";
 import {spacing} from "@khanacademy/wonder-blocks-tokens";
-import {BodyText, Title} from "@khanacademy/wonder-blocks-typography";
+import {BodyText} from "@khanacademy/wonder-blocks-typography";
 
 import {DrawerDialog, DrawerLauncher} from "@khanacademy/wonder-blocks-modal";
-import packageConfig from "../../packages/wonder-blocks-modal/package.json";
 
-import ComponentInfo from "../components/component-info";
 import {allModes} from "../../.storybook/modes";
 import {reallyLongText} from "../components/text-for-testing";
 import DrawerDialogArgTypes from "./drawer-dialog.argtypes";
@@ -29,16 +26,10 @@ import DrawerDialogArgTypes from "./drawer-dialog.argtypes";
  * via React Context, eliminating the need for manual prop passing in nested components.
  */
 export default {
-    title: "Packages / Modal / DrawerDialog",
+    title: "Packages / Modal / DrawerLauncher / DrawerDialog",
     component: DrawerDialog,
     parameters: {
         componentSubtitle: "Modal dialog designed for DrawerLauncher",
-        docs: {
-            description: {
-                component:
-                    "A drawer modal dialog that slides in from the specified alignment (inline start/end or block end). Automatically integrates with DrawerLauncher for proper animations and positioning.",
-            },
-        },
         chromatic: {
             // Drawer stories are behavior-based, disable visual regression testing
             disableSnapshot: true,
--- a/__docs__/wonder-blocks-modal/drawer-dialog.stories.tsx
+++ b/__docs__/wonder-blocks-modal/drawer-dialog.stories.tsx
@@ -11,7 +11,6 @@ import {BodyText} from "@khanacademy/wonder-blocks-typography";
 
 import {DrawerDialog, DrawerLauncher} from "@khanacademy/wonder-blocks-modal";
 
-import {allModes} from "../../.storybook/modes";
 import {reallyLongText} from "../components/text-for-testing";
 import DrawerDialogArgTypes from "./drawer-dialog.argtypes";
 
@@ -143,10 +142,10 @@ export const WithFormContent: StoryComponentType = {
 };
 
 /**
- * Drawer positioned at the bottom of the screen (block end).
- * Useful for mobile-friendly interfaces and action sheets.
+ * Drawer with action-focused content layout. Demonstrates a common
+ * pattern for presenting multiple action options to the user.
  */
-export const BottomDrawer: StoryComponentType = {
+export const WithActionList: StoryComponentType = {
     render: () => (
         <DrawerLauncher
             alignment="blockEnd"
@@ -170,9 +169,7 @@ export const BottomDrawer: StoryComponentType = {
                 />
             }
         >
-            {({openModal}) => (
-                <Button onClick={openModal}>Open Bottom Drawer</Button>
-            )}
+            {({openModal}) => <Button onClick={openModal}>Show Actions</Button>}
         </DrawerLauncher>
     ),
 };
@@ -190,6 +187,7 @@ export const WithRenderProp: StoryComponentType = {
                     title="Render Prop Example"
                     content={({title}) => (
                         <View style={styles.content}>
+                            <BodyText size="xsmall">Eyebrow</BodyText>
                             {title}
                             <BodyText style={styles.section}>
                                 This content uses a render prop to customize how
@@ -263,6 +261,11 @@ export const WithScrollableContent: StoryComponentType = {
                             <BodyText>{reallyLongText}</BodyText>
                             <BodyText>{reallyLongText}</BodyText>
                             <BodyText>{reallyLongText}</BodyText>
+                            <BodyText>{reallyLongText}</BodyText>
+                            <BodyText>{reallyLongText}</BodyText>
+                            <BodyText>{reallyLongText}</BodyText>
+                            <BodyText>{reallyLongText}</BodyText>
+                            <BodyText>{reallyLongText}</BodyText>
                         </View>
                     }
                 />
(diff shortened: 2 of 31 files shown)


__docs__/wonder-blocks-modal/drawer-launcher.argtypes.ts
```diff
@@ -0,0 +1,64 @@
+import type {ArgTypes} from "@storybook/react";
+
+export default {
```
beaesguerra, line 3: There seems to be an issue with the autogenerated docs for DrawerLauncher in Storybook! The description from the jsdocs aren't showing up, and some props like the `animated` prop aren't included in the props table 🤔 

Image

marcysutton: Yes, I'm not sure why exactly. I think it might be caused by the `withActionScheduler` HOC wrapping `DrawerLauncher`. I got the jsdoc description to show up once in development, and then it went away. I'll keep playing with it!

marcysutton:
@beaesguerra thank you so much for that review.... I actually had to do some heavy refactoring to fix the styling and RTL mode. I moved the styles into `DrawerLauncher` so that `FlexibleDialog` can still be used standalone, and now the concerns  should be located in the correct spots. I also added some visual snapshots for the end states since they kept getting messed up in the process. It doesn't capture the actual animations though...just where they land!

__docs__/wonder-blocks-modal/drawer-launcher.argtypes.ts
beaesguerra: Ah I see! We may need to explicitly set the prop descriptions for the arg types then so it shows up in the docs! I see this is what we do for ModalLauncher too!

__docs__/wonder-blocks-modal/drawer-launcher.stories.tsx
```diff
@@ -0,0 +1,726 @@
...
+        disableSnapshot: true,
+    },
+};
+
+/**
+ *
+ * An drawer with less content. Uses the `alignment` prop to slide in from the
+ * right in LTR writing mode and left in RTL writing mode.
+ */
+export const WithShortContent: StoryComponentType = {
```
beaesguerra, line 262: Should the width of the FlexibleDialog in the drawer be smaller when there is short content? 

Image

__docs__/wonder-blocks-modal/drawer-launcher.stories.tsx
```diff
@@ -0,0 +1,726 @@
...
+                    </View>
+                }
+            />
+        );
+
+        return (
+            <DrawerLauncher
+                alignment={args.alignment}
+                modal={modalInitialFocus}
+                initialFocusId="text-field-to-be-focused-field"
```
beaesguerra: The behaviour in this example doesn't seem to match the behaviour described - the initial focus id isn't set to the text field! I think this is because LabeledField provides it's own `id` to the field so that it can be wired up. 

One way to fix this is to move the id for `LabeledField` instead!

__docs__/wonder-blocks-modal/flexible-dialog.stories.tsx
```diff
@@ -274,24 +265,17 @@ export const WithStyle: StoryComponentType = {
...
+                            <BodyText>{reallyLongText}</BodyText>
                         </>
                     }
                     styles={{
                         root: {
                             color: semanticColor.status.notice.foreground,
                             maxWidth: 1000,
                         },
+                        panel: {
+                            backgroundColor: "red",
```
beaesguerra: Could we use a semantic color token as an example?

packages/wonder-blocks-styles/src/styles/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const zindexDropdown = 1000;
+export const zindexPopover = 1010;
+export const zindexTooltip = 1030;
+export const zindexFixedNavbar = 1030;
+export const zindexHeader = 1060;
+export const zindexModalBackdrop = 1070;
+export const zindexModal = 1080;
+export const zindexAboveModal = 1090;
```
beaesguerra: Would it make sense to have these as tokens instead? These are more constant values rather than css styles that can be used directly! What do you think?

Also, these values are probably something we want to align on with design so that it's represented the same way in Figma too!

packages/wonder-blocks-modal/src/util/types.ts
```diff
@@ -11,3 +11,11 @@ import * as React from "react";
...
  */
 export type ModalElement = React.ReactElement | null;
+
+/**
+ * The position of the modal, with logical support for RTL (Right-to-Left).
+ * `inlineStart` is left-aligned in LTR,
+ * `inlineEnd` is right-aligned in LTR,
+ * `blockEnd` is bottom-aligned.
+ */
+export type DrawerAlignment = "inlineStart" | "inlineEnd" | "blockEnd";
```
beaesguerra, line 21: Nice! I like that we are using logical naming!

packages/wonder-blocks-modal/src/components/__tests__/drawer-launcher.test.tsx
```diff
@@ -0,0 +1,819 @@
...
+
+const exampleModal = (
+    <FlexibleDialog
+        title="drawer launcher test"
+        content={<div data-modal-child />}
+    />
+);
+
+describe("DrawerLauncher", () => {
+    window.scrollTo = jest.fn();
```
beaesguerra: Do we need to restore `window.scrollTo` after this test suite runs?

packages/wonder-blocks-modal/src/components/__tests__/flexible-panel.test.tsx
beaesguerra, line 1: One thing to note is we also have Chromatic verifying styling/the position of things so we've been slowly moving away from testing for specific styles in jest tests!

packages/wonder-blocks-modal/src/components/__tests__/modal-launcher.test.tsx
```diff
@@ -104,9 +104,7 @@ describe("ModalLauncher", () => {
         expect(onCloseMock).toHaveBeenCalled();
     });
 
-    // TODO(FEI-5533): Key press events aren't working correctly with
-    // user-event v14. We need to investigate and fix this.
-    test.skip("Pressing Escape closes the modal", async () => {
+    test("Pressing Escape closes the modal", async () => {
```
beaesguerra, line 113: Thanks for fixing this!

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
```diff
@@ -141,27 +175,32 @@ FlexiblePanel.defaultProps = {
...
+            height: "100%",
+            width: "100%",
+        },
+
+        closeButton: {
+            position: "absolute",
+            // If writing direction is RTL, position the button on the left side
+            left: isRtl ? theme.closeButton.layout.gapRight : "unset",
+            // If writing direction is LTR, position the button on the right side
+            right: !isRtl ? theme.closeButton.layout.gapRight : "unset",
```
beaesguerra: Could we use `inset-inline-start`/ `inset-inline-end` instead of `left` and `right` CSS properties so it'll handle the ltr/rtl logic? Then we can rely on CSS and remove the MutationObserver logic!

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
```diff
@@ -80,6 +80,41 @@ export default function FlexiblePanel({
...
+
+        // Start observing document root for `dir` changes after initial render
+        observer.observe(document.documentElement, {
+            attributes: true,
+            attributeFilter: ["dir"],
+            subtree: true,
+        });
+
+        return () => observer.disconnect();
+    }, []);
```
beaesguerra: idea (for later): Would it be helpful to have a reusable `isRtl` hook utility? It could be used for other WB components and in other frontend codebases too! 

update: We may not need this logic if we can use logical css properties here instead! The utility may be useful for other cases though! For example, for tabs, we also search for the closest `[dir=rtl]` so we can determine what the next tab is when we use arrow keys. This would be something we couldn't achieve with CSS alone!

__docs__/wonder-blocks-modal/drawer-launcher-testing-snapshots.stories.tsx
```diff
@@ -0,0 +1,463 @@
...
+);
+
+export default {
+    title: "Packages / Modal / Testing / Snapshots / DrawerLauncher",
+    component: DrawerLauncher,
+    parameters: {
+        chromatic: {
+            modes: {
+                default: allModes.themeDefault,
+                "default rtl": allModes["themeDefault rtl"],
```
beaesguerra, line 56: Thanks for adding stories for the different cases!

I wonder if there's a way to reduce the number of snapshots generated for the testing stories! Each story in this file generates 6 snapshots and there's 14 stories so in total there's 84 new snapshots. I think there's some duplicated snapshots too since there are both the modes and individual stories that cover rtl and screen sizes!

For example, these 2 snapshots test the same thing (the rtl behaviour when alignment is `inline-start`)
Image

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
marcysutton: Yes, we should make a reusable utility. I think we need to support both bottom-up and top-down logic, as both seem necessary! I.e. look up the DOM tree for the closest element ancestor, or look at container elements from a sibling (or just from the top-down). Here is an issue: https://khanacademy.atlassian.net/browse/WB-2060

__docs__/wonder-blocks-modal/drawer-launcher-testing-snapshots.stories.tsx
marcysutton: Ah, I didn't see those toggles! I agree, that is too many stories. I will clean them up!

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
marcysutton: Yes, that worked! Good suggestion!

packages/wonder-blocks-modal/src/components/__tests__/flexible-panel.test.tsx
marcysutton: I have mixed feelings about it for this particular component: it isn't used directly, so it doesn't have stories of its own. Perhaps the functionality could be covered by the FlexibleDialog or DrawerLauncher stories, but then it wouldn't have any of its own test coverage. These tests seemed like a way to prevent breakages for an internal component (and not rely on Chromatic to uncover a failure).

packages/wonder-blocks-styles/src/styles/constants.ts
marcysutton: Good question. Maybe this concept would fit in with the other [elevation tokens](https://khanacademy.atlassian.net/browse/WB-1878)? These constants were copied from the `frontend` repo, where they are used a lot. I know z-index wars have been a huge challenge in the past and we try to avoid them by using stacking contexts. But I think there are still going to be times when z-index is unavoidable, and tokens make a ton of sense. Perhaps that could be done in a follow-up?

__docs__/wonder-blocks-modal/drawer-launcher.stories.tsx
marcysutton: As discussed, this drawer probably shouldn't be used when there is short content. I included it in the stories to ensure the styling was functional independent of the content, so I will opt to leave it as-is (with a comment)!

__docs__/wonder-blocks-modal/drawer-launcher.stories.tsx
marcysutton: Good find, I included a fix!

packages/wonder-blocks-styles/src/styles/constants.ts
marcysutton: Removed these constants for now, as the component itself didn't need `z-index`. I did include a `styles` escape hatch on the DrawerLauncher so it can be styled if necessary.

packages/wonder-blocks-modal/src/components/__tests__/flexible-panel.test.tsx
beaesguerra: Thanks for the context, that sounds fine to me!

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
```diff
@@ -189,9 +157,13 @@ const getComponentStyles = ({isRtl}: {isRtl: boolean}) => {
...
             // If writing direction is RTL, position the button on the left side
-            left: isRtl ? theme.closeButton.layout.gapRight : "unset",
+            insetInlineStart: isRtl
+                ? theme.closeButton.layout.gapRight
+                : "unset",
             // If writing direction is LTR, position the button on the right side
-            right: !isRtl ? theme.closeButton.layout.gapRight : "unset",
+            insetInlineEnd: !isRtl
+                ? theme.closeButton.layout.gapRight
+                : "unset",
```
beaesguerra: Thinking about this more, since `insetInlineStart/End` handles rtl, we may not need to change the values based on `isRtl`. `insetInlineEnd` could always have the `gapRight` value, and `insetInlineStart = unset`.

This seems to be the case already for some reason! When I change the writing direction in Storybook, the `insetInline` values don't swap!

| LTR | RTL |
| --- | --- |
| Image | Image |

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
```diff
@@ -304,7 +326,7 @@ const DrawerLauncher = (props: Props) => {
             {(opened && !isExiting) || (opened && isExiting && animated)
                 ? ReactDOM.createPortal(
                       <div dir={direction}>
```
beaesguerra: Interesting! I see we need to add the `dir` attribute to the portal because it gets rendered outside of the storybook-root (which includes the div with the `dir` set by the withLanguageDirection SB decorator). 

Image

I wonder if this wrapping `<div dir={direction}>` is needed in applications because ideally, the `dir` would be set higher in the DOM! The issue might be that our Storybook language direction decorator doesn't properly simulate an application environment! 

It's also interesting that we don't set a direction for other things using portals. I think that ideally we wouldn't need to set `dir` on a bunch of things so that there is only one source of truth higher up in the DOM.

What do you think?

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
```diff
@@ -0,0 +1,549 @@
...
+                    />
+                    <ScrollDisabler />
+                </>
+            )}
+        </ModalContext.Provider>
+    );
+};
+
+/** A component that, when mounted, calls `onClose` when Escape is pressed. */
+function DrawerLauncherKeypressListener({onClose}: {onClose: () => unknown}) {
```
beaesguerra, line 348: Cool! This functionality seems like it could be useful for other components that also close when `Escape` is pressed! 

My gut feeling was that this would be a helpful hook that is reusable! I think though it's because I assume components always need to return elements 😅 How do we decide when we should have reusable logic in a component vs a hook?

(No changes here necessary, I was curious about when to use this approach!)

Update: I see this is what `ModalLauncher` has too, so that's probably why!

__docs__/wonder-blocks-modal/drawer-launcher-testing-snapshots.stories.tsx
```diff
@@ -0,0 +1,164 @@
...
+                small: allModes.small,
+                large: allModes.large,
```
beaesguerra: I think we can remove the `large` mode since the `default`, `default rtl`, and `thunderblocks` modes are all at large screen sizes too! This can reduce our num of snapshots even more!

beaesguerra (approved):
Looks good! Left a few non-blocking questions in the PR, and some comments in Chromatic! Thanks for working on this and looking into z-indexes more!

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
```diff
@@ -0,0 +1,549 @@
...
+            if (typeof controlledOpened === "boolean" && !onClose) {
+                // eslint-disable-next-line no-console
+                console.warn("'onClose' should be used with 'opened'");
+            }
+            if (typeof controlledOpened !== "boolean" && !children) {
+                // eslint-disable-next-line no-console
+                console.warn("either 'children' or 'opened' must be set");
+            }
+        }
+    }, [controlledOpened, children, onClose]);
```
beaesguerra: Do you think we could structure the prop types to make sure invalid prop combinations can never be used together? I haven't seen runtime prop validation in WB components, and I'm not sure if logs are monitored by teams during development so these messages might be missed! It would be nice if we can avoid these prop combinations earlier during build time with typechecks!

It is nice though how clear these warning messages are, compared to TS errors 😅 What do you think?

Update: I see it's pre-existing in modal-launcher!

beaesguerra (commented):
Adding another question, for some reason it wasn't included in the last review!

Also noting that some prop docs are missing still!
image

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
marcysutton: I totally agree that these messages would probably get missed. TS errors would be a lot more appropriate! I did copy them over from ModalLauncher, but I'll see if I can get some feedback from typing instead. I have a similar thing happening with validation for the new `DrawerDialog` to ensure it is the only one used with `DrawerLauncher`. Runtime messages aren't a good fit for development workflows!

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
marcysutton: Maybe that code could be abstracted in a future follow-up!

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
marcysutton: I think we're going to have to do some deeper thinking on RTL support in `frontend`, so I'm going to go with a minimal solution for this with a plan to revisit later. I saw in `frontend` that the RTL logic comes from `RequestInfo`. Have you tackled integrating that into WB components yet? Maybe for Tabs? 

For now, I'm trying out an internal hook for looking at the `dir` attribute and I'm modifying `preview.tsx` for Storybook to append the `dir` higher in the DOM. I figure we can ship a core hook later on.

.storybook/preview.tsx
```diff
@@ -162,9 +162,23 @@ const withThemeSwitcher: Decorator = (
...
+
+        // Cleanup on unmount
+        return () => {
+            document.documentElement.removeAttribute("dir");
+        };
+    }, [context.globals.direction]);
+
     if (context.globals.direction === "rtl") {
         return (
             <div dir="rtl">
```
marcysutton, line 184: Maybe we don't need both attributes set? I didn't want to mess with any other stories so I left it in here for now.

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
marcysutton: Removed these runtime warnings in favor of stronger typing!

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
marcysutton: I filed an issue for it: https://khanacademy.atlassian.net/browse/WB-2065

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
marcysutton: Here's the issue for an RTL utility: https://khanacademy.atlassian.net/browse/WB-2060

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
marcysutton: Good call, this was a really easy fix! And I made the same change in `ModalPanel` for the [RTL issue](https://khanacademy.atlassian.net/browse/WB-2064?atlOrigin=eyJpIjoiNDFlZmZjODFhMTBjNDQwOThkODNjNmRmY2FhYTdjY2YiLCJwIjoiamlyYS1zbGFjay1pbnQifQ) that came in today.

__docs__/wonder-blocks-modal/drawer-launcher.argtypes.ts
```diff
@@ -0,0 +1,113 @@
...
+            required: true,
+        },
+    },
+
+    styles: {
+        control: {type: undefined},
+        table: {
+            type: {summary: "{container?: StyleType}"},
+        },
+        description: "Optional styles for the DrawerLauncher container",
```
marcysutton, line 29: These descriptions were necessary to show up...maybe another issue with the Scheduler HOC?

__docs__/wonder-blocks-modal/flexible-dialog.argtypes.ts
```diff
@@ -62,11 +65,15 @@ export default {
     "aria-label": {
         control: {type: "text"},
         table: {type: {summary: "string"}},
+        description:
+            "An accessible name for the dialog if there is no main heading.",
```
marcysutton, line 69: These descriptions were also necessary. Maybe it's due to the advanced typing in the component.

packages/wonder-blocks-modal/src/hooks/use-direction-detection.ts
```diff
@@ -0,0 +1,135 @@
...
+ * Simplified version that just returns the current direction without a ref.
+ * Useful for detecting the overall page direction.
+ *
+ * This function performs a direct DOM query on each render and does not use
+ * state or observers.
+ *
+ * @param options - Configuration options
+ * @returns The detected direction ("ltr" or "rtl")
+ */
+export function usePageDirection(
```
marcysutton: This hook is a first-go for this component, with the intent to create and export a super reusable hook later in https://khanacademy.atlassian.net/browse/WB-2060.

packages/wonder-blocks-modal/src/index.ts
```diff
@@ -15,5 +20,9 @@ export {
     ModalLauncher,
     OnePaneDialog,
     FlexibleDialog,
+    DrawerLauncher,
+    DrawerDialog,
     maybeGetPortalMountedModalHostElement,
 };
+
+export type {DrawerAlignment, DrawerDialogStyles};
```
marcysutton, line 28: Exporting these types seems useful to consumers, so I kept them in here.

packages/wonder-blocks-modal/src/util/types.ts
marcysutton: I simplified the names, as I kept getting them wrong! "is it insetInlineStart or inlineInsetStart?" These are much easier to remember.

static/sb-styles/preview.css
```diff
@@ -12,6 +12,8 @@ html {
 
 .sbdocs.sbdocs-wrapper {
     padding: 0 2rem 3.2rem 2rem;
+    /* create stacking context for docs so modals sit on top without z-index */
+    position: relative;
```
marcysutton, line 16: In the docs pages, the Storybook copy buttons were rendering on top of the modals due to a `z-index` set by Storybook. This positioning fixes that issue without having to use a z-index on the modals by creating a stacking context.

packages/wonder-blocks-modal/src/hooks/use-direction-detection.ts
```diff
@@ -0,0 +1,90 @@
...
+ * // With explicit direction (e.g., from RequestInfo)
+ * const MyRTLComponent = ({requestInfo}) => {
+ *     const direction = useDirectionDetection(undefined, {
+ *         direction: requestInfo.isRTL ? "rtl" : "ltr"
+ *     });
+ *     return <div>Explicit direction: {direction}</div>;
+ * };
+ * ```
+ */
+export function useDirectionDetection(
```
marcysutton, line 60: Re-adding my comment since I cleaned up the redundancy... this is a first go-around for a hook to ship with this modal. I intend to move it into `wonder-blocks-core` as a follow-up! That work will be covered under https://khanacademy.atlassian.net/browse/WB-2060

__docs__/wonder-blocks-modal/drawer-dialog.stories.tsx
```diff
@@ -0,0 +1,395 @@
...
+            )}
+        </DrawerLauncher>
+    ),
+};
+
+/**
+ * Drawer nested inside wrapper components. Demonstrates how the
+ * context-based prop passing works with nested component structures.
+ */
+export const WithNestedComponents: StoryComponentType = {
```
beaesguerra: The dialog when it comes from the bottom doesn't take up the full width if it's nested! Is this expected? I was assuming it'd have the same behaviour as when it isn't nested! Image

Would also be helpful to include a test case like this in Chromatic!

.storybook/preview.tsx
beaesguerra: I think we can remove this wrapper then since we're setting the `dir` on the document! (Can be done in a follow up!)

Tabs have some snapshots in rtl that was relying on the `direction` context so we can see if anything changes with it removed (I'm assuming there'll be no changes but we'll see!)

packages/wonder-blocks-modal/src/hooks/use-direction-detection.ts
beaesguerra: Are there tests we could add for the hook? We can do it as part of WB-2060 also!

packages/wonder-blocks-modal/src/hooks/use-direction-detection.ts
```diff
@@ -0,0 +1,90 @@
+import * as React from "react";
+
+type Direction = "ltr" | "rtl";
+
+interface DirectionDetectionOptions {
+    /**
+     * Explicit direction to use, bypassing DOM detection.
+     * Useful when direction is known from external sources like RequestInfo.
+     */
+    direction?: Direction;
```
beaesguerra, line 10: I'm curious what the source of truth is for the writing direction! I'm assuming the DOM is, since all logical CSS properties will abide by that. I'm wondering if the utility should also only rely on the DOM so that things can't get out of sync (for example, RequestInfo says something different from the `dir` attribute in the DOM). 

Do you have more context on the use case for supporting explicitly setting the direction from external sources?

(This can also be something we address in WB-2060 since this is currently an internal utility)

packages/wonder-blocks-modal/src/hooks/use-direction-detection.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    // If we have an elementRef and it has a current element, use element-based detection
+    if (elementRef?.current) {
+        // Look up the DOM tree for the nearest element with a dir attribute
+        const elementWithDir = elementRef.current.closest("[dir]");
+
+        if (elementWithDir) {
+            const dirValue = elementWithDir.getAttribute("dir");
+            return dirValue === "rtl" ? "rtl" : "ltr";
```
beaesguerra: if `dirValue` is not `rtl` and not `ltr`, do we want to use `defaultDirection`?

packages/wonder-blocks-modal/src/hooks/use-direction-detection.ts
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    // Fall back to document-level detection
+    const documentDir = document.documentElement.getAttribute("dir");
```
beaesguerra: In the classic experience (https://en-x-rtl.khanacademy.org/) , the `dir` attribute is set on the `body` element, rather than the `document`. Can we include a fall back to body-level detection? 

Image

packages/wonder-blocks-modal/src/components/modal-panel.tsx
```diff
@@ -152,7 +152,8 @@ const styles = StyleSheet.create({
 
     closeButton: {
         position: "absolute",
-        right: theme.closeButton.layout.gapRight,
+        // insetInlineEnd supports both RTL and LTR layouts
+        insetInlineEnd: theme.closeButton.layout.gapRight,
```
beaesguerra, line 156: 🎉 thanks for fixing this!

On a related note, `frontend` has a lint rule enabled for requiring logical css properties. Would be helpful to enable in WB! Created https://khanacademy.atlassian.net/browse/WB-2066 to enable this rule in WB 😄 

Image

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
```diff
@@ -27,13 +44,12 @@ type Props = Readonly<{
...
+     * Using other modal types may result in incorrect animations, positioning,
+     * and styling behavior.
      */
-    modal:
-        | ModalElement
-        | ((props: {
-              closeModal: () => void;
-              styles?: FlexibleDialogStyles;
-          }) => ModalElement);
+    modal: DrawerModalElement | DrawerModalFunction;
```
beaesguerra, line 72: minor naming suggestion: I wonder if this prop should be renamed to `dialog`, since consumers pass a DrawerDialog into it (or a function that returns something that includes DrawerDialog)

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
```diff
@@ -84,41 +100,49 @@ type Props = Readonly<{
...
+    opened?: never;
+};
+
+/**
+ * DrawerLauncher props - enforces proper controlled/uncontrolled usage at the type level
+ */
+type Props = ControlledProps | UncontrolledProps;
 
 // Set a default timing duration for animations and focus management. Also used for tests.
 export const DEFAULT_TIMING_DURATION = 400;
```
beaesguerra: It would be helpful to denote that this is in milliseconds! Either in a comment or in the naming like `DEFAULT_TIMING_DURATION_MS` or something like that!

packages/wonder-blocks-modal/src/components/drawer-dialog.tsx
```diff
@@ -0,0 +1,336 @@
...
+    /**
+     * Optional flag to determine whether the dialog is closing, to fine-tune animations.
+     */
+    isExiting?: boolean; // should this be in this layer?
+    /**
+     * Whether to include animation in the `DrawerLauncher` and child components.
+     * This should be false if the user has `prefers-reduced-motion` opted in.
+     * Defaults to `true`.
+     */
+    animated?: boolean;
```
beaesguerra: Do we need these props on both DrawerDialog and DrawerLauncher components? Since DrawerDialog uses the DrawerContext, we may not need the duplicated props anymore! Ideally, consumers shouldn't have to think about which component they need to configure!

packages/wonder-blocks-modal/src/util/drawer-context.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export interface DrawerContextProps {
+    alignment?: DrawerAlignment;
+    animated?: boolean;
+    isExiting?: boolean;
+    timingDuration?: number;
+}
+
+export const DrawerContext = React.createContext<DrawerContextProps | null>(
+    null,
```
beaesguerra: Could we set default values for the context data?

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
```diff
@@ -248,66 +247,52 @@ const DrawerLauncher = (props: Props) => {
...
-                ...modalElement.props,
-                ...additionalProps,
-                styles: mergedStyles,
-            });
+        const drawerDialogProps = {
+            alignment,
+            animated,
+            isExiting,
+            timingDuration,
         };
```
beaesguerra: Should this object of dialog props be memoized to avoid creating a new object on every render?

packages/wonder-blocks-modal/src/components/drawer-backdrop.tsx
```diff
@@ -0,0 +1,260 @@
...
+    /**
+     * Whether to include animation in the `DrawerBackdrop` component.
+     * This should be false if the user has `prefers-reduced-motion` opted in.
+     * Defaults to `true`.
+     */
+    animated: boolean;
+    /**
+     * Whether the drawer is exiting. Used to trigger exit animations.
+     */
+    isExiting?: boolean;
```
beaesguerra: Would it make sense for DrawerBackdrop to use the Drawer context too so it doesn't need these props?

packages/wonder-blocks-modal/src/components/drawer-dialog.tsx
```diff
@@ -0,0 +1,336 @@
...
+    /**
+     * Called when the close button is clicked.
+     *
+     * If you're using `ModalLauncher`, you probably shouldn't use this prop!
+     * Instead, to listen for when the modal closes, add an `onClose` handler
+     * to the `ModalLauncher`.  Doing so will result in a console.warn().
```
beaesguerra: We can remove the docs around the console warnings since you've used the prop types to prevent certain combinations!

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
```diff
@@ -0,0 +1,405 @@
...
+import ScrollDisabler from "./scroll-disabler";
+import type {DrawerAlignment, ModalElement} from "../util/types";
+import {DrawerContext} from "../util/drawer-context";
+import ModalContext from "./modal-context";
+import type {DrawerDialogStyles} from "./drawer-dialog";
+
+/**
+ * A more restrictive type for DrawerLauncher that encourages the use of DrawerDialog.
+ * While we still allow ModalElement for backwards compatibility, the type documentation
+ * and runtime warnings encourage proper usage.
```
beaesguerra: We can remove the docs around the runtime warnings since you've used the prop types to prevent certain combinations!

beaesguerra (commented):
Thanks for iterating on this after testing the snapshot in `frontend`! Left some questions 😄

__docs__/wonder-blocks-modal/drawer-dialog.stories.tsx
marcysutton: Oh weird! It worked from the DrawerLauncher stories. I wonder if it even makes sense to include those (Cursor generated them!). I wanted to include examples of how to use DrawerDialog generally, but nesting demonstrations might be better suited to DrawerLauncher. I'm going to opt to remove the story since it isn't that useful!

packages/wonder-blocks-modal/src/hooks/use-direction-detection.ts
marcysutton: Yes! I have some stashed for `wonder-blocks-core`. I was going to elect to add it in the follow-up!

marcysutton: Actually I changed my mind. I will include some tests!

packages/wonder-blocks-modal/src/hooks/use-direction-detection.ts
marcysutton: I'm not sure. I'll add this to the ticket so we can look at it then!

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
marcysutton: I think I'm going to elect to keep it as a `modal` to signal that the component being passed in _is_ essentially the modal dialog! It gets sorta confusing when you're in the collection of elements to discern what each piece is responsible for. The child passed to this prop is top-level component responsible for the modal dialog that users see.

packages/wonder-blocks-modal/src/components/drawer-dialog.tsx
marcysutton: Ah, good point! I cleaned those up in DrawerDialog and DrawerBackdrop, and moved default values into DrawerContext.

.storybook/preview.tsx
marcysutton: I removed the DIV in favor of the dynamic attribute setting!

static/sb-styles/preview.css
marcysutton: Oddly the copy buttons showed on top for me after coming back to the code to address your comments. Adding a `z-index: 0` to this wrapper fixed it. Something to keep an eye on...

packages/wonder-blocks-modal/src/hooks/use-direction-detection.ts
marcysutton: Yes, I changed it to use the defaultDirection. Good suggestion!

packages/wonder-blocks-modal/src/components/drawer-backdrop.tsx
marcysutton: Absolutely! Good call!

.storybook/preview.tsx
```diff
@@ -162,18 +162,24 @@ const withThemeSwitcher: Decorator = (
...
+            document.documentElement.setAttribute("dir", "ltr");
+        }
+
+        // Cleanup on unmount
+        return () => {
+            document.documentElement.removeAttribute("dir");
+        };
+    }, [context.globals.direction]);
+
+    return <Story />;
```
beaesguerra: We no longer need to wrap the Story with a `<div dir="rtl">`, but we'll need stories to re-render when the direction changes! I think we can achieve that by setting a `key` on `<Story />` to `context.globals.direction` so it re-renders when the value changes!

While testing `Tabs`, I noticed the active styling wasn't properly updating like it used to! (Tabs Story ([before](https://khan.github.io/wonder-blocks/?path=/story/packages-tabs-tabs--default&globals=direction:ltr)/[after](https://5e1bf4b385e3fb0020b7073c-jcopircsoa.chromatic.com/?path=/story/packages-tabs-tabs--default&globals=direction:ltr))). The issue can be reproduced by changing the writing direction using the control! 

https://github.com/user-attachments/assets/6cf3752a-a1d3-4156-a356-6a4ce380f3ba

packages/wonder-blocks-modal/src/hooks/use-direction-detection.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    // If we have an elementRef and it has a current element, use element-based detection
+    if (elementRef?.current) {
+        // Look up the DOM tree for the nearest element with a dir attribute
+        const elementWithDir = elementRef.current.closest("[dir]");
+
+        if (elementWithDir) {
+            const dirValue = elementWithDir.getAttribute("dir");
+            return dirValue === "rtl" ? "rtl" : defaultDirection;
```
beaesguerra: If `dirValue=ltr` and `defaultDirection=rtl`, this will always return `rtl`! 

Could do something like this:
```suggestion
            return dirValue === "rtl" || dirValue === "ltr" ? dirValue : defaultDirection;
```

packages/wonder-blocks-modal/src/hooks/use-direction-detection.ts
```diff
@@ -0,0 +1,92 @@
...
+    // Fall back to document-level detection
+    const documentDir =
+        document.documentElement.getAttribute("dir") ||
+        document.body.getAttribute("dir");
+    if (documentDir === "rtl") {
+        return "rtl";
+    }
+
+    // Fall back to default
+    return defaultDirection;
```
beaesguerra: Similar case here where if `defaultDirection="rtl"`, the function will always return `rtl`! 


```suggestion
    if (documentDir === "rtl" || documentDir === "ltr") {
        return documentDir;
    }

    // Fall back to default
    return defaultDirection;
```

packages/wonder-blocks-modal/src/util/drawer-context.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// Default values for the drawer context - using centralized defaults from DrawerLauncher
+const defaultDrawerContextValue: DrawerContextProps = {
+    animated: DEFAULT_DRAWER_ANIMATED,
+    isExiting: DEFAULT_DRAWER_IS_EXITING,
+    timingDuration: DEFAULT_DRAWER_TIMING_DURATION_MS,
+};
+
+export const DrawerContext = React.createContext<DrawerContextProps | null>(
```
beaesguerra: If you pass in `defaultDrawerContextValue` into `createContext`, it'll become the initial default value! Then the type can also always be `React.createContext<DrawerContextProps>` since the value will never be null!

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
```diff
@@ -0,0 +1,408 @@
...
+        }),
+        [alignment, animated, isExiting, timingDuration],
+    );
+
+    const renderModal = React.useCallback(() => {
+        // Handle function-based modals
+        if (typeof modal === "function") {
+            const renderedModal = modal({
+                closeModal: handleCloseModal,
+                ...drawerDialogProps,
```
beaesguerra: Do the `drawerDialogProps` need to be passed in as function props still since the `DrawerDialog` component can get the props from the DialogContext now?

packages/wonder-blocks-modal/src/util/drawer-context.ts
```diff
@@ -1,18 +1,46 @@
...
+export const DEFAULT_DRAWER_BACKDROP_DISMISS_ENABLED = true;
+
+/** Default value for internal exit animation state. */
+export const DEFAULT_DRAWER_IS_EXITING = false;
+
 export interface DrawerContextProps {
     alignment?: DrawerAlignment;
     animated?: boolean;
     isExiting?: boolean;
     timingDuration?: number;
```
beaesguerra: Since we return `defaultDrawerContextValue` merged with `context`, could these values be all non-optional (without the `?`)? 

That way, consumers of the context don't need to handle the defaults since the hook/context will handle that logic, and we won't need to export the default values!

.storybook/preview.tsx
marcysutton: Ah, good find! I'm going to punt on this change until later and restore the wrapper DIV with the document attribute also. The key isn't fixing the Tabs style issue for me, even with a dynamic `dir` attribute on that wrapper element. 

In debugging Tabs I also found the Popover CloseButton isn't updating with RTL. It wasn't obvious before, because it gets injected outside of the Story like Modals do! I filed an issue to split that work off: https://khanacademy.atlassian.net/browse/WB-2069

packages/wonder-blocks-modal/src/hooks/use-direction-detection.ts
marcysutton: Thanks Bea!

packages/wonder-blocks-modal/src/util/drawer-context.ts
```diff
@@ -0,0 +1,46 @@
...
+    timingDuration: DEFAULT_DRAWER_TIMING_DURATION_MS,
+};
+
+export const DrawerContext = React.createContext<DrawerContextProps>(
+    defaultDrawerContextValue,
+);
+
+export const useDrawerContext = (): DrawerContextProps => {
+    const context = React.useContext(DrawerContext);
+    return context
```
beaesguerra: Since `context` can no longer be null, can we return `context` directly?

beaesguerra (approved):
Nice work! I'm excited to have an official drawer component! 🎉 

Left a non-blocking question !

packages/wonder-blocks-modal/src/util/drawer-context.ts
marcysutton: Seems to work! Good call!

PR #2745, opened 2025-08-05
Adds a `locale` prop to `BirthdayPicker` to better support i18n

beaesguerra:
## Summary:

Previously, we were getting the list of months in short form from `moment`. When `moment` was removed from WB in favour of the Temporal API in https://github.com/Khan/wonder-blocks/pull/2549, we started using the Intl.DateTimeFormat api with `navigator.language` from the browser

Unfortunately, `navigator.language` isn't always the same as the user's language setting. To address this, we introduce a `locale` prop so that consuming applications can set the locale to use. If this isn't provided, it will default to using `navigator.language` to get the month names

Issue: WB-1939

## Test plan:

1. Confirm the BirthdayPicker Locale story shows localized labels for the month names and selected month value (?path=/story/packages-birthdaypicker--locale)
2. Using the npm snapshot in frontend, confirm that using the `locale` prop fixes the localization issue

### DIFF
--- a/packages/wonder-blocks-birthday-picker/src/components/birthday-picker.tsx
+++ b/packages/wonder-blocks-birthday-picker/src/components/birthday-picker.tsx
@@ -65,6 +65,11 @@ type Props = {
      * Additional styles applied to the dropdowns.
      */
     dropdownStyle?: StyleType;
+    /**
+     * The locale to use for the month names. If not provided, the browser's
+     * `navigator.language` value will be used.
+     */
+    locale?: string;
 };
 
 type State = {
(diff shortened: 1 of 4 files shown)


beaesguerra:
Marking this as "Ready for review" so that I can test with the npm snapshot in `frontend`!

beaesguerra:
Set up the `frontend` PR and got the integration changes ready. Translated month names are working as expected: https://github.com/Khan/frontend/pull/2653

These changes are ready for review!

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.tsx
```diff
@@ -355,9 +360,12 @@ export default class BirthdayPicker extends React.Component<Props, State> {
     }
 
     monthsShort(): string[] {
-        const format = new Intl.DateTimeFormat(navigator.language, {
-            month: "short",
-        }).format;
+        const format = new Intl.DateTimeFormat(
+            this.props.locale || navigator.language,
```
marcysutton: What do you think about using nullish coalescing here (`??`) instead of logical or to handle the fallback? It might be a little safer if `this.props.locale` is falsy.

marcysutton (approved):
Thanks for working on this @beaesguerra!

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.tsx
beaesguerra: Updated, thanks Marcy!

PR #2750, opened 2025-08-12
Ensure openers in MultiSelect and SingleSelect set aria-required

marcysutton:
## Summary:
SingleSelect and MultiSelect were missing programmatic `required` states even though context messages were set. In this PR, I ensure SingleSelect and MultiSelect render with `aria-required="true"` if an error message is passed in OR if `aria-required` is passed in.

Issue: WB-2047

## Test plan:
1. Ensure tests pass
2. Review rendered DOM for required SingleSelect/MultiSelect and ensure `aria-required="true"` is present


beaesguerra (approved):
Thanks for the fix! 🚀

PR #2754, opened 2025-08-13
Update Tab and NavigationTabItem font color

beaesguerra:
## Summary:
Update font color for Tabs and NavigationTabs so that it matches the designs 

Figma: https://www.figma.com/design/EuFu0U7gqc1koXm8ZhlOLp/%E2%9A%A1%EF%B8%8F-Components?node-id=2156-77&m=dev

Issue: WB-2026

## Test plan:

Confirm font color in Tabs and NavigationTabs matches the designs
- `?path=/docs/packages-tabs-navigationtabs--docs`
- `?path=/docs/packages-tabs-tabs--docs`

### DIFF
--- a/.changeset/tender-items-move.md
+++ b/.changeset/tender-items-move.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tabs": patch
+---
+
+Update font color in Tabs and NavigationTabs to `core.foreground.neutral.subtle`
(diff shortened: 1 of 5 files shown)


marcysutton (approved):
Looks good!

beaesguerra:
> I did notice one issue in Chromatic for a custom style that didn't pass contrast ratios. It is _so_close at 4.46:1! Curious if we can adjust it to pass.

@marcysutton Updated! It was an example to show that custom styles can be applied to NavigationTabItem. I used setting the border as an example instead! 

I also disabled the snapshot for that individual story and included it in the existing Scenarios story so that we could reduce our number of snapshots!

PR #2757, opened 2025-08-14
Banner: Support hiding the icon

beaesguerra:
## Summary:

- Banner: Adds support for hiding the icon by setting the `icon` prop to `none`

Issue: WB-2029

## Test plan:

Confirm that the icon is not shown when `icon="none"`(`/?path=/story/packages-banner--with-no-icon`)


marcysutton (approved):
Looks good!

PR #2758, opened 2025-08-15
WB-2050(part1): Tokens update - Add knockout tokens and replace `foreground.inverse.*` and `border.inverse.*` usage

beaesguerra:
## Summary:

- Introducing `border.knockout.default` and `foreground.knockout.default` tokens. This will replace `border.inverse.*` and `foreground.inverse.*` tokens to align with the recent design token changes (the inverse tokens will be removed in a future update once existing usage of these tokens have moved to the new token structure)
- Updating existing usage of `border.inverse.*` and `foreground.inverse.*` tokens in WB

Token Mapping:

| Before | After |
| --- | --- |
| `foreground.inverse.strong` | `foreground.knockout.default` |
| `border.inverse.strong` | `border.knockout.default` |

For `foreground.inverse.subtle/default` and `border.inverse.subtle/default` the mapping will either map to a `neutral` or `disabled` value depending on the use case

Issue: WB-2050
Design: [Core Tokens](https://www.figma.com/design/e9qdyiDUBZ3rDabP9S7ulO/%E2%9A%A1%EF%B8%8F-Handshake?node-id=3040-15158&m=dev)

## Test plan:

- Confirm the `foreground.knockout.default` and `border.knockout.default` tokens exist (`?path=/docs/foundations-using-color--docs`)
- Confirm the `foreground.inverse.*` and `border.inverse.*` tokens have not been removed
- Most things should look the same. For any components that have any changes, confirm they are intentional (I noted the expected changes in Chromatic)

## Implementation plan

1. Add knockout tokens (https://github.com/Khan/wonder-blocks/pull/2758)
2. Add background base and overlay tokens
3. Update usage (in WB and other repos)
- Replace foreground and border inverse tokens usage in other repos
- Replace surface tokens with background base tokens
- Use background.base.default token for white instead of background.neutral.subtle
4. Update background.neutral.subtle to light gray
5. Remove surface and inverse tokens

### DIFF
--- a/.changeset/silent-bikes-cheer.md
+++ b/.changeset/silent-bikes-cheer.md
@@ -1,5 +1,5 @@
 ---
-"@khanacademy/wonder-blocks-tokens": major
+"@khanacademy/wonder-blocks-tokens": minor
 ---
 
-Adds `foreground.knockout.default` and `border.knockout.default` semantic core colors. Removes `foreground.inverse.*` and `border.inverse.*` in favour of the new knockout tokens. Tokens that were previously using this have been updated to use other core tokens
+Adds `foreground.knockout.default` and `border.knockout.default` semantic core colors. Tokens that were previously using `foreground.inverse.*` and `border.inverse.*` have been updated to use the knockout token or other core tokens
--- a/packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
+++ b/packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
@@ -38,6 +38,11 @@ const core = {
             default: color.gray_70,
             strong: color.gray_60,
         },
+        inverse: {
+            subtle: color.gray_60,
+            default: color.gray_90,
+            strong: color.white_100,
+        },
         knockout: {
             default: color.white_100,
         },
@@ -108,6 +113,11 @@ const core = {
             subtle: color.gray_60,
             default: color.gray_50,
             strong: color.gray_40,
+        },
+        inverse: {
+            subtle: color.gray_60,
+            default: color.gray_90,
+            strong: color.white_100,
         },
         knockout: {
             default: color.white_100,
--- a/packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
+++ b/packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
@@ -41,6 +41,11 @@ const core = {
             default: color.fadedOffBlack16,
             strong: color.fadedOffBlack32,
         },
+        inverse: {
+            subtle: color.fadedOffBlack16,
+            default: color.fadedOffBlack8,
+            strong: color.white,
+        },
         knockout: {
             default: color.white,
         },
@@ -111,6 +116,11 @@ const core = {
             subtle: color.fadedOffBlack16,
             default: color.fadedOffBlack32,
             strong: color.fadedOffBlack50,
+        },
+        inverse: {
+            subtle: color.fadedOffBlack32,
+            default: color.offWhite,
+            strong: color.white,
         },
         knockout: {
             default: color.white,


packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -342,7 +342,7 @@ const styles = StyleSheet.create({
     disabled: {
         background: semanticColor.surface.primary,
         borderRadius: theme.root.border.radius.default,
-        color: semanticColor.core.foreground.inverse.subtle,
+        color: theme.root.color.disabled.foreground,
```
beaesguerra, line 345: The disabled colors for TB and Classic are different so I set up theme tokens for the color!

packages/wonder-blocks-cell/src/theme/default.ts
```diff
@@ -77,7 +80,7 @@ export default {
         sizing: {
             height: sizing.size_020,
         },
-        shadow: `inset 0px -1px 0px ${semanticColor.core.border.inverse.default}`,
+        shadow: `inset 0px -1px 0px ${semanticColor.core.border.neutral.subtle}`,
```
beaesguerra, line 83: This causes a visual diff, from the comment in the [Chromatic diff](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=689f6be2bac67cac4cb11e34): 
> The cell divider was using `border.inverse.default (fadedOffBlack8)`. `fadedOffBlack8` isn't used for a semantic border token so as an alternative, `border.neutral.subtle (fadedOffBlack16)` is used instead!

Semantically, it makes sense to be using `border.neutral.subtle` for dividers!

packages/wonder-blocks-form/src/util/styles.ts
```diff
@@ -176,7 +176,7 @@ export const colorStates: StyleMap = {
             default: {
                 rest: {
                     border: semanticColor.input.checked.background,
-                    background: semanticColor.core.border.inverse.strong,
+                    background: semanticColor.core.border.knockout.default, // TODO(WB-2050): use white background
```
beaesguerra, line 179: In one of the next PRs, there'll be a new `background.base.default` token which is `white`. Will update to that new token later on so we don't have to use a border token for a background here!

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -725,7 +721,7 @@ export const semanticColor = {
         rest: core.foreground.instructive.subtle,
         hover: core.foreground.instructive.subtle,
         press: core.foreground.instructive.default,
-        disabled: core.foreground.inverse.subtle,
+        disabled: core.foreground.disabled.default,
```
beaesguerra, line 734: `foreground.inverse.subtle` was `fadedOffBlack32`, `foreground.disabled.default` is also `fadedOffBlack32`! No value changes here!

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -283,59 +279,59 @@ export const semanticColor = {
...
-                    foreground: core.foreground.inverse.strong,
+                    foreground: core.foreground.knockout.default,
                 },
             },
 
             disabled: {
                 border: core.border.disabled.strong,
                 background: core.border.disabled.strong,
-                foreground: core.foreground.inverse.default,
+                foreground: core.foreground.knockout.default, // TODO check this is okay
```
beaesguerra: `core.foreground.inverse.default` was `offWhite`, `core.foreground.knockout.default`is `white`. This is a slight since we don't include `offWhite` in any core foreground colors

Image

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -283,59 +279,59 @@ export const semanticColor = {
...
-                    foreground: core.foreground.inverse.strong,
+                    foreground: core.foreground.knockout.default,
                 },
             },
 
             disabled: {
                 border: core.border.disabled.strong,
                 background: core.border.disabled.strong,
-                foreground: core.foreground.inverse.default,
+                foreground: core.foreground.knockout.default,
```
beaesguerra, line 344: `core.foreground.inverse.default` was `offWhite`, `core.foreground.knockout.default`is `white`. This is a slight change since we don't include `offWhite` in any core foreground colors. Since it's a slight change, it doesn't show in the diffs in Chromatic!

Image

Note: TB is not impacted by this

Image

packages/wonder-blocks-link/src/components/link-core.tsx
```diff
@@ -193,13 +193,10 @@ const states = {
         // TODO(WB-1852): Remove light variant.
         inverse: {
             rest: {
-                foreground: semanticColor.core.foreground.inverse.strong,
+                foreground: semanticColor.core.foreground.knockout.default,
             },
             hover: {
-                foreground: semanticColor.core.foreground.inverse.strong,
-            },
-            focus: {
```
beaesguerra, line 201: This was removed since we use the global focus styling for links now!

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
```diff
@@ -558,7 +554,7 @@ export const semanticColor = mergeTheme(defaultSemanticColor, {
         rest: core.foreground.instructive.default,
         hover: core.foreground.instructive.strong,
         press: core.foreground.instructive.strong,
-        disabled: core.foreground.inverse.subtle,
+        disabled: core.foreground.disabled.subtle,
```
beaesguerra, line 567: `core.foreground.inverse.subtle` in TB was `gray_60`, ` core.foreground.disabled.subtle` is `gray_60` too! No value changes here!

beaesguerra:
Putting this back in draft mode! I realized Perseus is using the inverse css variable tokens. To simplify the release dependencies with perseus and frontend, I'm going to change my approach by releasing the new tokens in a new minor version, switching to the new tokens in other repos, then removing the tokens we want to replace in a new major version! Will make some minor updates in this PR to support that!

.changeset/silent-bikes-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": minor
```
marcysutton, line 2: What happens when there are multiple changesets for the same package in a release? (e.g. tokens) Should there be one?

beaesguerra: Good question! They'll get combined, so it's fine to have multiple changesets! 

In this case, since there's both a `minor` and `patch` update to the `tokens` package, the version bump will be a new minor version, and both changeset descriptions will be included in the changelog. The descriptions will be grouped by change type (`minor`/`patch`)!

Since the `few-carrots-cough.md` changeset also includes an update to `wonder-blocks-cell`, the changelog for the cell package will include the same description as the one for the patch change in `wonder-blocks-tokens`.

Here's an example from a recent Versions PR where there was a minor and patch changeset to the banner package: https://github.com/Khan/wonder-blocks/pull/2760

PR #2762, opened 2025-08-21
Allow ActivityIconButton to take an arbitrary element as the icon argument

MikeKlemarewski:
## Summary:
I'm working on a design that adds an ActivityIconButton with Khanmigo's face as the icon in the button. This requires us to be able to pass something other than a PhosphorIcon as the `icon` prop to ActivityIconButton. This updates the prop to take an arbitrary ReactElement as well.

https://github.com/user-attachments/assets/0ed9da33-b9ea-46f3-a4ea-ef31b0e44899

## Test plan:
- Make sure tests pass
- Run storybook and check all the ActivityIconButton stories

### DIFF
--- a/__docs__/wonder-blocks-icon-button/activity-icon-button-testing-snapshots.stories.tsx
+++ b/__docs__/wonder-blocks-icon-button/activity-icon-button-testing-snapshots.stories.tsx
@@ -181,6 +181,7 @@ export const Scenarios: Story = {
                                 </Icon>
                             }
                             aria-label="Khanmigo"
+                            kind="secondary"
                         />
                     ),
                 },


packages/wonder-blocks-icon-button/src/components/activity-icon-button.tsx
```diff
@@ -46,6 +46,11 @@ type Props = BaseIconButtonProps &
          * - `neutral` is used for buttons that indicate a neutral action.
          */
         actionType?: ActivityIconButtonActionType;
+
+        /**
+         * A Phosphor icon asset (imported as a static SVG file), or an element.
+         */
+        icon: PhosphorIconAsset | React.ReactElement;
```
MikeKlemarewski, line 53: Alternatively, instead of `ReactElement`, we might be able to enforce us only passing our own wonderblocks `<Icon />` type. LMK what y'all think.

beaesguerra: Using `React.ReactElement` is fine here! We also use `React.ReactElement` for the `icon` prop type in the StatusBadge component (which supports both PhosphorIcon and Icon components!)

It would be nice if we could use TS to enforce what component can be used for the prop. TS doesn't support it right now though (see [related thread](https://khanacademy.slack.com/archives/C028T2HDX3P/p1741210473344549))

__docs__/wonder-blocks-icon-button/activity-icon-button.stories.tsx
```diff
@@ -213,3 +216,32 @@ export const WithLabel: StoryComponentType = {
...
+                style={{
+                    gap: sizing.size_160,
+                    flexDirection: "row",
+                    alignItems: "flex-start",
+                }}
+            >
+                <ActivityIconButton
+                    icon={
+                        <Icon size="medium">
+                            <img alt="Khanmigo" src={khanmigoIcon} />
```
beaesguerra: Since the img has `alt` set and the button has an `aria-label`, the "Khanmigo" description is read out twice when a screen reader reads out everything. The alt can be set to an empty string so only the button aria-label will be read out! This is the behaviour with the ActivityIconButton when a PhosphorIcon is used 

Here's a recording of VoiceOver with and without the img alt. `Khanmigo, image` is only announced before the `alt` content was removed:

https://github.com/user-attachments/assets/6bed9582-286b-4410-bd03-b2d8ab18186b

cc: @marcysutton in case you have thoughts on having both an `alt` on the icon img and an `aria-label` on the `button` element!

__docs__/wonder-blocks-icon-button/activity-icon-button.stories.tsx
beaesguerra, line 1: Could you also add a scenario for an example showing the custom icon in the testing snapshots Scenarios story: 

https://github.com/Khan/wonder-blocks/blob/750c3e8a99388fb073e893ee8ed3acb33db46e07/__docs__/wonder-blocks-button/activity-button-testing-snapshots.stories.tsx#L98

This is so that this use case can get included in our visual regression tests in Chromatic to make sure this functionality doesn't break!

beaesguerra (commented):
Thanks for working on this Mike, this looks great! Left some comments for updating the a11y in the documented example and including coverage for custom icons in our visual regression tests!

__docs__/wonder-blocks-icon-button/activity-icon-button.stories.tsx
marcysutton: Yep, we only need one! The ActivityButton can get its accessible name from either the icon alt _or_ the `aria-label`.

__docs__/wonder-blocks-icon-button/activity-icon-button.stories.tsx
MikeKlemarewski: Sure thing. Added! Is there anything else I need to update? Or do I just add the scenario?

beaesguerra: Yup that's all! That story gets captured as a snapshot and we can see the visual diffs in Chromatic (it'll check for both classic and thunderblocks themes): https://www.chromatic.com/build?appId=5e1bf4b385e3fb0020b7073c&number=7643 

Could you also update the custom icon scenario to use the `secondary` kind so it's easier to see the Khanmigo icon? I'll approve the Chromatic snapshots after!

MikeKlemarewski: Oh yup!

PR #2763, opened 2025-08-21
Add onMouseDown / onMouseDown / onMouseLeave to `ActivityButton` and `Button`

marcysutton:
## Summary:
Adds support for mouse events for advanced animation purposes

Issue: WB-2006

## Test plan:
1. Review new `ActivityButton` story: /?path=/story/packages-button-activitybutton--press-duration-tracking&globals=viewport:desktop
2. Review new `Button` story: `/?path=/story/packages-button-button--press-duration-tracking`
3. Ensure tests pass

### DIFF
--- a/.changeset/dirty-toys-dance.md
+++ b/.changeset/dirty-toys-dance.md
@@ -2,4 +2,4 @@
 "@khanacademy/wonder-blocks-button": minor
 ---
 
-Add onMouseUp, onMouseDown, and onMouseLeave to ActivityButton to track duration of press before release
+Add onMouseUp, onMouseDown, and onMouseLeave to ActivityButton and Button to track duration of press before release
--- a/.changeset/dirty-toys-dance.md
+++ b/.changeset/dirty-toys-dance.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-button": minor
+---
+
+Add onMouseDown, onMouseDown, and onMouseLeave to ActivityButton to track duration of press before release
--- a/.changeset/dirty-toys-dance.md
+++ b/.changeset/dirty-toys-dance.md
@@ -2,4 +2,4 @@
 "@khanacademy/wonder-blocks-button": minor
 ---
 
-Add onMouseDown, onMouseDown, and onMouseLeave to ActivityButton to track duration of press before release
+Add onMouseUp, onMouseDown, and onMouseLeave to ActivityButton to track duration of press before release
--- a/__docs__/wonder-blocks-button/activity-button.stories.tsx
+++ b/__docs__/wonder-blocks-button/activity-button.stories.tsx
@@ -229,12 +229,20 @@ export const PressDurationTracking: Story = {
             setLastEvent(eventName);
         };
 
+        const baseActions = {
+            onMouseDown: action("onMouseDown"),
+            onMouseUp: action("onMouseUp"),
+            onMouseLeave: action("onMouseLeave"),
+            onClick: action("onClick"),
+            onMouseEnter: action("onMouseEnter"),
+        };
+
         const handleMouseDown = (e: React.MouseEvent) => {
             const startTime = Date.now();
             setPressStartTime(startTime);
             setPressDuration(null);
             logEvent("Mouse Down - Press Started");
-            action("onMouseDown")(e);
+            baseActions.onMouseDown?.(e);
         };
 
         const handleMouseUp = (e: React.MouseEvent) => {
@@ -243,7 +251,12 @@ export const PressDurationTracking: Story = {
                 setPressDuration(duration);
                 logEvent("Mouse Up - Press Completed", duration);
             }
-            action("onMouseUp")(e);
+            baseActions.onMouseUp?.(e);
+        };
+
+        const handleMouseEnter = (e: React.MouseEvent) => {
+            logEvent("Mouse Enter");
+            baseActions.onMouseEnter?.(e);
         };
 
         const handleMouseLeave = (e: React.MouseEvent) => {
@@ -253,12 +266,12 @@ export const PressDurationTracking: Story = {
                 logEvent("Mouse Leave - Press Abandoned", duration);
             }
             setPressStartTime(null);
-            action("onMouseLeave")(e);
+            baseActions.onMouseLeave?.(e);
         };
 
         const handleClick = (e: React.SyntheticEvent) => {
             logEvent("Click - Action Executed");
-            action("onClick")(e);
+            baseActions.onClick?.(e);
         };
 
         const resetTracking = () => {
@@ -284,6 +297,7 @@ export const PressDurationTracking: Story = {
                     <ActivityButton
                         {...args}
                         startIcon={clock}
+                        onMouseEnter={handleMouseEnter}
                         onMouseDown={handleMouseDown}
                         onMouseUp={handleMouseUp}
                         onMouseLeave={handleMouseLeave}
--- a/__docs__/wonder-blocks-button/activity-button.stories.tsx
+++ b/__docs__/wonder-blocks-button/activity-button.stories.tsx
@@ -4,9 +4,11 @@ import type {Meta, StoryObj} from "@storybook/react";
 
 import magnifyingGlass from "@phosphor-icons/core/regular/magnifying-glass.svg";
 import caretRight from "@phosphor-icons/core/regular/caret-right.svg";
+import clock from "@phosphor-icons/core/regular/clock.svg";
 
 import {action} from "@storybook/addon-actions";
 import {View} from "@khanacademy/wonder-blocks-core";
+import {BodyText} from "@khanacademy/wonder-blocks-typography";
 import Button, {ActivityButton} from "@khanacademy/wonder-blocks-button";
 
 import ComponentInfo from "../components/component-info";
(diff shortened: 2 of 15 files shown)


packages/wonder-blocks-button/src/components/activity-button.tsx
```diff
@@ -176,26 +179,53 @@ export const ActivityButton = React.forwardRef(function ActivityButton(
...
-                    tabIndex={tabIndex}
-                    ref={ref}
-                >
-                    {children}
-                </ActivityButtonCore>
-            )}
+            {(state: ClickableState, restChildProps: ChildrenProps) => {
+                // Wrap onMouseLeave to call custom handler before internal handler
+                // Only call custom handler if button is not disabled and handler exists
+                const wrappedOnMouseLeave =
```
marcysutton: I got help from Cursor with this, and asked why it added the wrapped event. It's due to `ClickableBehavior`:

>The wrapper exists because ClickableBehavior treats onMouseLeave fundamentally differently than onMouseDown/onMouseUp:
✅ onMouseDown/onMouseUp: Direct prop support with automatic user handler calls
❌ onMouseLeave: No prop support, internal handler only, signature mismatch
Without the wrapper: User's onMouseLeave handlers would simply be ignored, making the feature request impossible to fulfill properly.
The wrapper is the cleanest architectural solution that doesn't require modifying the core ClickableBehavior component while ensuring both user callbacks and internal state management work correctly.

I suppose an alternative would be to add support in ClickableBehavior so other buttons could use it. I'm hesitant to add mouse event support more broadly so it doesn't create mouse-only interactions. I also think `Clickable` needs to be revisited for the markup it allows...maybe it could be part of that effort?  https://khanacademy.atlassian.net/browse/WB-2062

packages/wonder-blocks-button/src/util/button.types.ts
```diff
@@ -140,6 +140,21 @@ export type BaseButtonProps =
...
+
+        /**
+         * Respond to a raw "mouseup" event.
+         */
+        onMouseUp?: (e: React.MouseEvent) => unknown;
+
+        /**
+         * Respond to a raw "mouseleave" event.
+         */
+        onMouseLeave?: (e: React.MouseEvent) => unknown;
```
beaesguerra: Since these props are added to `BaseButtonProps`, the Button component will also have it's props updated too! Do we want to add support for these event handlers in the `Button` component too? If not, we could add these types to the `ActivityButtonProps` type only!

packages/wonder-blocks-button/src/components/activity-button.tsx
beaesguerra: Interesting! My instinct was to update `ClickableBehavior` to support `onMouseLeave` so if other things needed it, we could use the default browser behaviour!

> I'm hesitant to add mouse event support more broadly so it doesn't create mouse-only interactions.

This is a very good point! I wonder if we should also expose `onKeyDown` and `onKeyUp` so that the equivalent confetti functionality is available for keyboard users too. What do you think? (Could be done separately if we decide to do this!)

One idea that would be interesting for our WB components is to have event handlers for `onInteractionStart` and `onInteractionEnd` so that internally, we can wire them up to work with both mouse and key events (prop naming could be something else)! That way, consumers can pass in their functions without having to set up both mouse and key events, and without having to handle both space and enter keys (assuming that's the expected behaviour)! 

I agree that Clickable needs to be revisited, since we were starting to move away from using it!

pnpm-lock.yaml
beaesguerra, line 1: Interesting that the lock file changed without any dependency changes! Are these changes from running `pnpm install` locally? I'm curious how often we should be doing this! I feel like I run `pnpm i` often, but the lockfile isn't always updated 🤔

.changeset/dirty-toys-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-button": minor
+---
+
+Add onMouseDown, onMouseDown, and onMouseLeave to ActivityButton to track duration of press before release
```
beaesguerra:
```suggestion
Add onMouseUp, onMouseDown, and onMouseLeave to ActivityButton to track duration of press before release
```

__docs__/wonder-blocks-button/activity-button.stories.tsx
```diff
@@ -180,3 +182,183 @@ export const ReceivingFocusProgrammatically: Story = {
...
+        };
+
+        const handleMouseLeave = (e: React.MouseEvent) => {
+            if (pressStartTime) {
+                const duration = Date.now() - pressStartTime;
+                setPressDuration(duration);
+                logEvent("Mouse Leave - Press Abandoned", duration);
+            }
+            setPressStartTime(null);
+            action("onMouseLeave")(e);
```
beaesguerra: I find the SB actions add-on super helpful for testing out event handlers manually! We can also configure it for all stories for a component so that the action logs show up in the other stories too (similar to how the `onClick` one is set up currently for Activity Button)

https://github.com/Khan/wonder-blocks/blob/83620b48f0f9f49800b67f642f9d6135361a6574/__docs__/wonder-blocks-button/activity-button.stories.tsx#L41-L47

beaesguerra (commented):
Nice! Left some questions for you 😄

.changeset/dirty-toys-dance.md
marcysutton: Oh funny, I obviously missed this typo! Thank you!

pnpm-lock.yaml
marcysutton: I think that this was an artifact from my system when switching branches, I will see if I can remove it from the PR!

__docs__/wonder-blocks-button/activity-button.stories.tsx
marcysutton: Good to know! The global args aren't triggering anything for mouse enter/leave/up/down events, unlike these inline versions. I'll keep playing with it, but I'll probably leave them inline in the story.

beaesguerra: It's fine to keep how it's configured in this story, since you've set it up to call `action` in the event handlers already (this story overrides those args)! It's more so the actions show up in the other stories like the [ActivityButton Default](https://5e1bf4b385e3fb0020b7073c-bflgijpprj.chromatic.com/?path=/story/packages-button-activitybutton--default) story!

packages/wonder-blocks-button/src/components/activity-button.tsx
marcysutton: > One idea that would be interesting for our WB components is to have event handlers for onInteractionStart and onInteractionEnd so that internally, we can wire them up to work with both mouse and key events

I like this idea! I feel like the raw events would still be important for fine-tuned experiences. There might be use cases where `onKeyDown` and `onMouseDown` have different needs, and an API to invoke both might not be what you want. I'll hold off on that for now and we could consider it...maybe as part of `onKeyDown`/etc. we could explore a "do both" API too? That way it could be opt-in! Here is an issue for it: https://khanacademy.atlassian.net/browse/WB-2072

packages/wonder-blocks-button/src/util/button.types.ts
marcysutton: I did end up adding the event handlers to `Button`! That makes sense to me. I looked into whether they could go in `ButtonUnstyled`, but it got complicated with `ClickableBehavior`. Maybe that's an area where we could simplify things later on.

__docs__/wonder-blocks-button/button-shared.argtypes.ts
```diff
@@ -44,6 +44,42 @@ export default {
             },
         },
     },
+    onMouseDown: {
+        action: "onMouseDown",
```
beaesguerra, line 48: Cool! TIL you can configure actions in the arg types!

packages/wonder-blocks-button/src/components/button.tsx
```diff
@@ -74,29 +78,61 @@ const Button = React.forwardRef(function Button(
             onClick={onClick}
             safeWithNav={safeWithNav}
             rel={rel}
+            onMouseDown={onMouseDown}
+            onMouseUp={onMouseUp}
```
beaesguerra, line 80: Since ButtonCore/ButtonUnstyled supports the mouse events as props, does it need to be passed into ClickableBehaviour as well?

packages/wonder-blocks-button/src/components/button.tsx
```diff
@@ -74,29 +78,61 @@ const Button = React.forwardRef(function Button(
...
+                                  "mouseleave",
+                                  {
+                                      bubbles: true,
+                                      cancelable: true,
+                                  },
+                              ) as unknown as React.MouseEvent;
+                              onMouseLeave(syntheticEvent);
+                              restChildProps.onMouseLeave();
+                          }
+                        : restChildProps.onMouseLeave;
```
beaesguerra: Do we still need to wrap the mouse events when we're able to pass the events directly into the underlying button element in `ButtonUnstyled`? If so, it'd be helpful to include a comment explaining why! (It's a bit unclear to me since it seems we pass in event handlers into both ClickableBehaviour and ButtonCore!)

marcysutton: This code is intended to combine the events from `ClickableBehavior` and `Button`:

>We need to wrap mouse events because we have TWO sets of handlers that need to execute:
> 1. ClickableBehavior's internal handlers (restChildProps.onMouseEnter/onMouseLeave)
>     - These manage button state (hover, focus, pressed) for styling and accessibility
> 2. User-provided custom handlers (onMouseEnter/onMouseLeave props)
>      - These allow consumers to add custom behavior on mouse interactions

>Even though ButtonUnstyled directly passes handlers to the DOM element, we still need this wrapping layer to ensure BOTH sets of handlers execute. ButtonCore and ButtonUnstyled are just styling/rendering layers - they don't add their own event logic, they just pass our wrapped handlers through.

packages/wonder-blocks-button/src/components/button.tsx
marcysutton: Yes -- I think this line is passing it into `ClickableBehavior`! It's just hard to tell with the diff.

packages/wonder-blocks-button/src/components/activity-button.tsx
marcysutton: Your instinct was right, it was `ClickableBehavior` that needed the events! It was a very minimal amount of wiring, as that code already handled the events. I just had to invoke the user events in the same handlers, and ouila!

packages/wonder-blocks-button/src/components/__tests__/button.test.tsx
```diff
@@ -103,6 +103,40 @@ describe("Button", () => {
...
+            target: button,
+            keys: "[MouseLeft>][/MouseLeft]",
+        });
+        await userEvent.unhover(button);
+
+        // Assert
+        expect(onMouseEnter).toHaveBeenCalledTimes(1);
+        expect(onMouseDown).toHaveBeenCalledTimes(1);
+        expect(onMouseUp).toHaveBeenCalledTimes(1);
+        expect(onMouseLeave).toHaveBeenCalledTimes(1);
```
beaesguerra: From the [Unit Testing Best Practices doc](https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/98402353/Unit+Testing+Best+Practices#Assert-One-Requirement-Per-Test), there's one guideline for asserting one requirement per test! 

An alternative way of testing all these event handlers without duplicating a bunch of code is by using `test.each` so it can iterate through each event handler and create a separate test for each!

packages/wonder-blocks-clickable/src/components/__tests__/clickable-behavior.test.tsx
```diff
@@ -81,6 +81,28 @@ describe("ClickableBehavior", () => {
...
+                    const label = labelForState(state);
+                    return <button {...childrenProps}>{label}</button>;
+                }}
+            </ClickableBehavior>,
+        );
+        const button = await screen.findByRole("button");
+        await userEvent.hover(button);
+        expect(onMouseEnter).toHaveBeenCalled();
+        await userEvent.unhover(button);
+        expect(onMouseLeave).toHaveBeenCalled();
```
beaesguerra: Same comment about having one assertion per test!

packages/wonder-blocks-button/src/components/__tests__/button.test.tsx
marcysutton: Ah yes, good call on `test.each`! I will add that in.

PR #2759, opened 2025-08-15
WB-2050(pt2): Tokens update - Add semantic color core background base and overlay tokens

beaesguerra:
## Summary:

Add base and overlay background tokens

These will be used instead of the `surface` tokens

Issue: WB-2050
Design: [Colors](https://www.figma.com/design/e9qdyiDUBZ3rDabP9S7ulO/%E2%9A%A1%EF%B8%8F-Handshake?node-id=3040-15158&m=dev)

## Test plan:

Confirm base and overlay background tokens match the Figma tokens in TB and Classic `?path=/docs/foundations-using-color--docs`

## Implementation plan 

1. Add knockout tokens (https://github.com/Khan/wonder-blocks/pull/2758)
2. Add background base and overlay tokens (https://github.com/Khan/wonder-blocks/pull/2759)
3. Update usage (in WB and other repos)
- Replace foreground and border inverse tokens usage in other repos
- Replace surface tokens with background base tokens
- Use background.base.default token for white instead of background.neutral.subtle
4. Update background.neutral.subtle to light gray
5. Remove surface and inverse tokens

PR #2766, opened 2025-08-22
Use new background tokens in WB components

beaesguerra:
## Summary:

- Moves away from the `semanticColor.surface.*` tokens in favour of the new `semanticColor.core.background.base/overlay` tokens
- Also moves away from `semanticColor.core.background.neutral.subtle` usage in TB that expects the value to be `white`  (the remaining usage of `background.neutral.subtle` is fine to change to light gray later on for the TB theme) 

Token mapping: 

| Previous token | New token [1] | 
| --- | --- |
| `semanticColor.surface.primary` | `semanticColor.core.background.base.default` |
| `semanticColor.surface.secondary` | `semanticColor.core.background.base.subtle` |
| `semanticColor.surface.emphasis` | `semanticColor.core.background.base.strong` |
| `semanticColor.surface.inverse` [2] | `semanticColor.core.background.neutral.strong` |
| `semanticColor.surface.overlay` | `semanticColor.core.overlay.default` |
| `semanticColor.core.background.neutral.subtle`[3] | `semanticColore.core.background.base.default` |

[1] For places where this direct mapping isn't used, there was probably a more appropriate token to use. I've annotated the PR with these use cases. Let me know if you have any questions! 

[2] In the Classic theme, `surface.inverse` is `darkBlue`. We don't have a semantic token for `darkBlue` since this is a pattern we want to start moving away from so we map existing usage to `background.neutral.strong` where it makes sense (see related https://khanacademy.slack.com/archives/C07CA2YR0BZ/p1755209242348059?thread_ts=1755201560.160039&cid=C07CA2YR0BZ). Annotated the PR with any exceptions to this mapping

[3] If `semanticColor.core.background.neutral.subtle` is expected to be `white` in the ThunderBlocks theme, it was updated to `semanticColore.core.background.base.default`. This is because the background neutral subtle value will change in a following PR

Issue: WB-2050

Design: [Colors](https://www.figma.com/design/e9qdyiDUBZ3rDabP9S7ulO/%E2%9A%A1%EF%B8%8F-Handshake?node-id=3040-15158&m=dev)

## Test plan:

1. Confirm there are no unexpected changes to components (I've commented on the expected changes in Chromatic as well)
2. Confirm `surface.` tokens are no longer used in WB
3. Confirm remaining use of `background.neutral.subtle` is okay to change to light gray in the TB themes
4. Confirm new token usage makes sense

## Implementation Plan

1. Add knockout tokens (https://github.com/Khan/wonder-blocks/pull/2758)
2. Add background base and overlay tokens (https://github.com/Khan/wonder-blocks/pull/2759)
5. Update usage (in WB and other repos) (WB (this PR): https://github.com/Khan/wonder-blocks/pull/2766) 
- Replace foreground and border inverse tokens usage in other repos
- Replace surface tokens with background base tokens
- Use background.base.default token for white instead of background.neutral.subtle
6. Remove surface and inverse tokens  
7. Update background.neutral.subtle to light gray

### DIFF
--- a/__docs__/wonder-blocks-clickable/clickable.stories.tsx
+++ b/__docs__/wonder-blocks-clickable/clickable.stories.tsx
@@ -336,7 +336,7 @@ const styles = StyleSheet.create({
         textAlign: "center",
     },
     dark: {
-        backgroundColor: semanticColor.surface.inverse,
+        backgroundColor: semanticColor.core.background.neutral.strong,
         color: semanticColor.core.foreground.knockout.default,
         padding: spacing.xSmall_8,
     },
--- a/__docs__/wonder-blocks-clickable/clickable.stories.tsx
+++ b/__docs__/wonder-blocks-clickable/clickable.stories.tsx
@@ -353,8 +353,8 @@ const styles = StyleSheet.create({
         padding: spacing.large_24,
     },
     disabled: {
-        color: semanticColor.core.foreground.knockout.default,
-        backgroundColor: semanticColor.surface.overlay,
+        color: semanticColor.action.primary.disabled.foreground,
+        backgroundColor: semanticColor.action.primary.disabled.background,
     },
     button: {
         maxWidth: 150,
--- a/__docs__/wonder-blocks-pill/pill.stories.tsx
+++ b/__docs__/wonder-blocks-pill/pill.stories.tsx
@@ -245,7 +245,8 @@ export const WithStyle: StoryComponentType = () => {
 
         ":active": {
             outlineColor: tokens.semanticColor.core.border.neutral.subtle,
-            backgroundColor: tokens.semanticColor.surface.overlay,
+            backgroundColor:
+                tokens.semanticColor.core.background.neutral.default,
         },
     };
 
--- a/__docs__/wonder-blocks-pill/pill.stories.tsx
+++ b/__docs__/wonder-blocks-pill/pill.stories.tsx
@@ -234,7 +234,7 @@ WithTypography.parameters = {
 
 export const WithStyle: StoryComponentType = () => {
     const customStyle = {
-        backgroundColor: tokens.semanticColor.surface.inverse,
+        backgroundColor: tokens.semanticColor.core.background.neutral.strong,
         color: tokens.semanticColor.core.foreground.knockout.default,
         paddingLeft: tokens.spacing.xxLarge_48,
         paddingRight: tokens.spacing.xxLarge_48,
--- a/__docs__/wonder-blocks-tokens/semantic-color.stories.tsx
+++ b/__docs__/wonder-blocks-tokens/semantic-color.stories.tsx
@@ -87,8 +87,6 @@ export const SemanticColors = () => (
                 cell: (row) => (
                     <View
                         style={{
-                            background:
-                                semanticColor.core.background.neutral.subtle,
                             padding: sizing.size_060,
                         }}
                     >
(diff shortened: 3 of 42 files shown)


packages/wonder-blocks-toolbar/src/components/toolbar.tsx
```diff
@@ -144,7 +144,7 @@ const sharedStyles = StyleSheet.create({
     },
     // TODO(WB-1852): Remove light variant.
     dark: {
-        background: semanticColor.surface.inverse,
+        background: color.darkBlue,
```
beaesguerra, line 147: Intentionally using the primitive `darkBlue` color to defer removing support for the light variant in Toolbar in WB-1852!

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-thunderblocks.ts
```diff
@@ -264,17 +264,17 @@ export const semanticColor = mergeTheme(defaultSemanticColor, {
             neutral: {
                 default: {
                     border: core.border.neutral.subtle,
-                    background: core.background.neutral.subtle,
```
beaesguerra, line 267: One of the upcoming breaking changes is TB will have a light gray value instead of white for `core.background.neutral.subtle`. We move to `core.background.base.default` (white) now to make this move easier later on!

__docs__/wonder-blocks-tokens/semantic-color.stories.tsx
```diff
@@ -87,8 +87,6 @@ export const SemanticColors = () => (
                 cell: (row) => (
                     <View
                         style={{
-                            background:
-                                semanticColor.core.background.neutral.subtle,
```
beaesguerra, line 91: Removed the additional background color for the color docs! Sharing from the related [Chromatic comment](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=68a8c406fcaa973a83bdd04b):

> I removed the gray box around the color box!
>
> I was checking the instances of semanticColor.background.neutral.subtle  since the value will be updated from white to light gray in TB. 
>
> I find the colors easier to see without the gray background so I opted to remove the background box color altogether!

__docs__/wonder-blocks-pill/pill.stories.tsx
```diff
@@ -245,7 +245,8 @@ export const WithStyle: StoryComponentType = () => {
 
         ":active": {
             outlineColor: tokens.semanticColor.core.border.neutral.subtle,
-            backgroundColor: tokens.semanticColor.surface.overlay,
```
beaesguerra, line 248: This wasn't being used as an overlay so we can used a strong neutral background token instead!

__docs__/wonder-blocks-clickable/clickable.stories.tsx
```diff
@@ -353,8 +353,8 @@ const styles = StyleSheet.create({
         padding: spacing.large_24,
     },
     disabled: {
-        color: semanticColor.core.foreground.knockout.default,
-        backgroundColor: semanticColor.surface.overlay,
+        color: semanticColor.action.primary.disabled.foreground,
+        backgroundColor: semanticColor.action.primary.disabled.background,
```
beaesguerra, line 357: Realized the action tokens would be more suitable for this example!

marcysutton: Do the action tokens seem stable enough to use? I fear we'd have to change it again if the abstraction didn't stick around...

beaesguerra: I used the action tokens since it's what we use for Buttons right now, and this Clickable Disabled example is similar to Buttons! I'm not sure if we have plans to revisit action tokens, but if we do, we can update this the same way we update other components using the action tokens!

PR #2768, opened 2025-08-25
Fix misc. dev errors and warnings in TS 5.9.x

marcysutton:
## Summary:
Every time I work in the WB codebase, there are a couple of "problems" related to TypeScript, Jest DOM, or odd testing quirks that I have to look past. This PR attempts to fix some of those issues with configuration and tweaks!

Update: These errors are dependent on the TypeScript version! I saw them in VSCode when it selected the TS version (5.9.2). They aren't present with the Workspace version (5.7.3). So it isn't urgent, but will have to be solved eventually for future TS upgrades.

Errors in VS Code

Issue: WB-2073

## Test plan:
1. Ensure tests pass
2. Ensure there are no "workspace problems" anymore


consistency-tests/__tests__/clickables.test.tsx
```diff
@@ -184,10 +190,18 @@ describe.each`
...
+        // @ts-expect-error [FEI-5019] - TS2740 - Type '{ assign: Mock<any, any, any>; }' is missing the following properties from type 'Location': ancestorOrigins, hash, host, hostname, and 8 more.
+        window.location = {assign: jest.fn(), origin: "http://localhost"};
 
         // Act
         render(
-            <Component onClick={jest.fn()} testId="clickable-component-test-id">
+            <Component
+                onClick={jest.fn()}
+                testId="clickable-component-test-id"
+                href="#" // Required for Link component
```
marcysutton, line 203: I don't love putting `href` on all of these variants since they probably aren't all links. But I didn't want to hack on anything too involved, either!

beaesguerra: If we set the `href` prop, the rendered component will always be a link! It looks like this test case is for other interactive components too. Maybe we can conditionally set `href`? 

Also, do you know how the test was passing before without setting `href`?

consistency-tests/__tests__/clickables.test.tsx
```diff
@@ -54,6 +54,8 @@ describe.each`
     });
 
     afterEach(() => {
+        // @ts-expect-error [FEI-5019] - TS2339 - Property 'mockClear' does not exist on type '(url: string | URL) => void'.
```
beaesguerra, line 57: I see we have other similar `@ts-expect-error`s in the repo so this is probably fine! I wonder if there's an alternative way to address these TS issues 🤔 

Interesting to see we have `@ts-expect-error`s like this in `frontend` too!

beaesguerra (commented):
Thanks for taking this on! Left some comments on things I wanted to get your thoughts on (mainly adding an `href` prop to the clickables test!)

marcysutton:
I think I'll elect to close this one since it isn't an issue for our workspace version of TypeScript. If we decide to fix these in the future, we can start from this PR!

PR #2771, opened 2025-08-28
Tokens: Update ThunderBlocks background neutral subtle value from white to gray_80

beaesguerra:
## Summary:

-  Update `semanticColor.core.background.neutral.subtle` from `white` to `gray_80` to align with design
- I've gone through WB, frontend, and perseus to convert any token usage that was relying on a `white` value for this token in the TB theme (I'll confirm again when I integrate these changes into frontend)

Issue: WB-2050

## Test plan:

1. Confirm `semanticColor.core.background.neutral.subtle` in TB is `gray_80`
2. Confirm any visual changes are expected


.changeset/afraid-geese-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": major
```
jandrade: **question:** Could you please clarify why this would be a major change? As far as I can tell, the token value is changing but the semantic color structure is still the same.

beaesguerra: Good question! That's right, there is no breaking change to the external API! I was considering this as a visual/functional breaking change. I found in `frontend` that some places were using `core.background.neutral.subtle` for the `white` value in the TB theme, so this token value change would cause unexpected results without intervention! 

To mitigate this change, I went through frontend and perseus to change the places that were expecting a white value in TB. I did this early last week so I'll check again when I integrate the new WB tokens packages! 

For reference, here are the related PRs for migrating the background neutral subtle token:
- [frontend PR](https://github.com/Khan/frontend/pull/3293) for updating `core.background.neutral.subtle` usage
- [perseus PR](https://github.com/Khan/perseus/pull/2835)

I also plan to release this with the removal of the surface and border/foreground inverse tokens so this will get put together with other breaking changes too!

jandrade: I understand your point and agree that this requires changes in consumers. That said, I  think is best to keep this as a `patch` change as it ends up affecting the visual styles of the consuming apps (as you mentioned in your reply), but it doesn't affect the behavior of it. Having to change consumers is something that would be mostly expected regardless of the version type.

You can see this case documented here: https://khanacademy.atlassian.net/wiki/spaces/WB/pages/2217640195/Versioning+and+releasing+Wonder+Blocks+packages#Patch-changes-0.0.1 (see "Styling changes").

beaesguerra: Oh cool, I have not read these docs, thanks for sharing! Updated to `patch`!

As mentioned, I'll release this with the other token changes so this will end up as part of a new major version. At least the changeset description associated with these changes will be included under the `patch` notes!

jandrade: thanks! I appreciate your context, just recommended this path to keep the versioning somewhat lean (imagine if we would marking more styling changes as breaking changes, then the API would grow very quickly 😄). Thanks again!

jandrade (approved):
Looks great! 🚀

PR #2773, opened 2025-09-02
DrawerDialog: Add support for zeroed out padding

marcysutton:
## Summary:
The DrawerDialog needs an update to support zeroed-out padding where borders and content go all the way to the edge. There is padding set in the `ModalContent` component, a nested child deep in the structure. In this PR, plumbing is created to allow a `styles.content` object to override the padding in `ModalContent`.

Issue: WB-2079

## Test plan:
1. Review stories for DrawerDialog, FlexibleDialog, and Modal Testing snapshot stories
2. Ensure ModalContent has no unexpected layout shifts in OG modals
3. Determine whether media query usage in ModalContent can be deferred until later (WB-2080)

### DIFF
--- a/__docs__/wonder-blocks-modal/drawer-dialog.stories.tsx
+++ b/__docs__/wonder-blocks-modal/drawer-dialog.stories.tsx
@@ -23,6 +23,16 @@ import DrawerDialogArgTypes from "./drawer-dialog.argtypes";
  *
  * The component automatically receives alignment, animation, and timing props from `DrawerLauncher`
  * via React Context, eliminating the need for manual prop passing in nested components.
+ *
+ * ### Custom styling
+ *
+ * You can optionally pass in the `styles` prop to override various parts of a DrawerDialog.
+ *
+ * - `styles.root` -  The outermost container of the dialog itself: alignment styles, box shadow, minWidth, maxWidth, width, height, maxHeight, etc.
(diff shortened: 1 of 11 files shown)


__docs__/wonder-blocks-modal/drawer-dialog.stories.tsx
```diff
@@ -33,6 +43,55 @@ export default {
             // Drawer stories are behavior-based, disable visual regression testing
             disableSnapshot: true,
         },
+        docs: {
+            description: {
+                component: `\`DrawerDialog\` is the modal content component designed specifically for use with \`DrawerLauncher\`.
+It provides a consistent drawer interface with proper animations, positioning, and accessibility features.
+
+**IMPORTANT**: This component should only be used with \`DrawerLauncher\`. Using it with other
+modal launchers may result in incorrect animations, positioning, and styling.
```
beaesguerra, line 52: Good call adding this to the docs! 

Should the DrawerLauncher also only be used with DrawerDialog? Would that be helpful to document for the DrawerLauncher component too?

beaesguerra (approved):
Looks good! 🚀

__docs__/wonder-blocks-modal/drawer-dialog.stories.tsx
marcysutton: Good call, updated!

PR #2778, opened 2025-09-04
[WB-2059] Sync shadow color tokens with Figma tokens

jandrade:
## Summary:

- Tokens:
  - Updated `core.shadow.transparent` to include `low`, `mid` and `high` tokens. Renames `shadow.chonky.progressive` to `shadow.chonky.instructive`.
  - Updated `core.shadow.chonky.progressive` with `*.instructive`
- Dropdown and Modal: Updates boxShadow to use `core.shadow.transparent.{low, high}` instead of the previous `core.shadow.transparent` token.
- Popover and Tooltip: Updates boxShadow to use `core.shadow.transparent.low` instead of a primitive `color` token.

Issue: https://khanacademy.atlassian.net/browse/WB-2059

## Test plan:

Navigate to the Popover and Tooltip examples and verify that the box shadow now uses the updated shadow color token.

In OG, the new value should be `offBlack16` instead of `offBlack8`.

### DIFF
--- a/packages/wonder-blocks-dropdown/src/theme/default.ts
+++ b/packages/wonder-blocks-dropdown/src/theme/default.ts
@@ -17,7 +17,7 @@ export default {
             },
         },
         shadow: {
-            default: `0 ${sizing.size_080} ${sizing.size_080} 0 ${semanticColor.core.shadow.transparent}`,
+            default: `0 ${sizing.size_080} ${sizing.size_080} 0 ${semanticColor.core.shadow.transparent.low}`,
         },
     },
     opener: {
--- a/packages/wonder-blocks-popover/src/components/popover-content-core.tsx
+++ b/packages/wonder-blocks-popover/src/components/popover-content-core.tsx
@@ -5,7 +5,6 @@ import type {AriaProps, StyleType} from "@khanacademy/wonder-blocks-core";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {
     border,
-    color,
     semanticColor,
     spacing,
 } from "@khanacademy/wonder-blocks-tokens";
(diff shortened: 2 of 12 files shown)


packages/wonder-blocks-dropdown/src/theme/default.ts
```diff
@@ -17,7 +17,7 @@ export default {
             },
         },
         shadow: {
-            default: `0 ${sizing.size_080} ${sizing.size_080} 0 ${semanticColor.core.shadow.transparent}`,
+            default: `0 ${sizing.size_080} ${sizing.size_080} 0 ${semanticColor.core.shadow.transparent.low}`,
```
jandrade, line 20: **note:** This PR only changes the shadow color value. The next PR will include the rest of the box-shadow values (via the new `elevation` tokens category).

beaesguerra: Ohh interesting, I see we use sizing tokens for the box shadow values. I wonder if we should be using px values instead (similar to border width). Or maybe it's expected to scale with text, I'm not sure! Something we can explore more with the `elevation` tokens work 😄

packages/wonder-blocks-popover/src/components/popover-content-core.tsx
```diff
@@ -5,7 +5,6 @@ import type {AriaProps, StyleType} from "@khanacademy/wonder-blocks-core";
 import {View} from "@khanacademy/wonder-blocks-core";
 import {
     border,
-    color,
```
beaesguerra, line 8: Nice! 🧹

beaesguerra (approved):
Looks good to me! 🚀

PR #2783, opened 2025-09-05
[WB-1878] Add elevation tokens

jandrade:
## Summary:

- Adds elevation tokens to `wonder-blocks-tokens`.
- Updates components to use elevation tokens instead of hardcoded box shadows.
- Updates Tooltip and Popover to use the correct elevation tokens.

Issue: WB-1878

## Test plan:

Navigate to `dropdown`, `modal`, `popover`, and `tooltip` stories in Storybook
to ensure they render correctly with the new elevation tokens.

### DIFF
--- a/__docs__/wonder-blocks-tokens/elevation.stories.tsx
+++ b/__docs__/wonder-blocks-tokens/elevation.stories.tsx
@@ -1,0 +1,100 @@
+import * as React from "react";
+import {Title, Subtitle, Description, Stories} from "@storybook/blocks";
+import {Meta} from "@storybook/react";
+import {View} from "@khanacademy/wonder-blocks-core";
+import TokenTable from "../components/token-table";
+import {
+    elevation,
+    semanticColor,
+    sizing,
+} from "@khanacademy/wonder-blocks-tokens";
+import ComponentInfo from "../components/component-info";
+import packageConfig from "../../packages/wonder-blocks-tokens/package.json";
+import {Code} from "../components/code";
+
+/**
+ * The `elevation` tokens are used to define the box-shadow of an element.
+ *
+ * Note: We don't define box-shadow colors here, and instead we recommend you to
+ * rely on the `semanticColor.core.shadow` tokens to build on top of that.
+ *
+ * ## Usage
+ *
+ * ```ts
+ *   import {elevation} from "@khanacademy/wonder-blocks-tokens";
+ *   const styles = {
+ *       elevatedContainer: {
+ *           boxShadow: `{elevation.low} ${semanticColor.core.shadow.transparent.low}`,
+ *       },
+ *   };
+ * ```
+ */
+export default {
+    title: "Packages /Tokens / Elevation",
+    parameters: {
+        docs: {
+            // Use a custom page so the SB <Primary> component is not rendered
+            // <Primary> renders the first story as part of the description,
+            // which isn't necessary for the token pages.
+            page: () => (
+                <>
+                    <Title />
+                    <Subtitle />
+                    <Description />
+                    <Stories title="Tokens" />
+                </>
+            ),
+        },
+        componentSubtitle: (
+            <ComponentInfo
+                name={packageConfig.name}
+                version={packageConfig.version}
+            />
+        ),
+    },
+    tags: ["!dev"],
+} as Meta;
+
+type Row = {label: string; css: string; value: string};
+const baseColumns = (kind: string) => [
+    {
+        label: "Token",
+        cell: (row: Row) => <Code>{`elevation.${row.label}`}</Code>,
+    },
+    {
+        label: "CSS Variable",
+        cell: (row: Row) => <Code>{row.css}</Code>,
+    },
+    {
+        label: "Value",
+        cell: "value",
+    },
+];
+
+export const Elevation = {
+    render: () => (
+        <TokenTable
+            columns={[
+                ...baseColumns("elevation"),
+                {
+                    label: "Example",
+                    cell: (row) => (
+                        <View
+                            style={{
+                                backgroundColor:
+                                    semanticColor.core.background.instructive
+                                        .default,
+                                boxShadow: `var(${row.css}) ${semanticColor.core.shadow.transparent.low}`,
+                                width: sizing.size_480,
+                                height: sizing.size_480,
+                            }}
+                        >
+                            &nbsp;
+                        </View>
+                    ),
+                },
+            ]}
+            tokens={elevation}
+        />
+    ),
+};
--- a/packages/wonder-blocks-dropdown/src/theme/default.ts
+++ b/packages/wonder-blocks-dropdown/src/theme/default.ts
@@ -1,5 +1,6 @@
 import {
     border,
+    elevation,
     font,
     semanticColor,
     sizing,
@@ -17,7 +18,7 @@ export default {
             },
         },
         shadow: {
-            default: `0 ${sizing.size_080} ${sizing.size_080} 0 ${semanticColor.core.shadow.transparent.low}`,
+            default: `${elevation.mid} ${semanticColor.core.shadow.transparent.mid}`,
         },
     },
     opener: {
--- a/packages/wonder-blocks-dropdown/src/theme/thunderblocks.ts
+++ b/packages/wonder-blocks-dropdown/src/theme/thunderblocks.ts
@@ -2,6 +2,7 @@ import {mergeTheme} from "@khanacademy/wonder-blocks-theming";
 
 import {
     border,
+    elevation,
     font,
     semanticColor,
     sizing,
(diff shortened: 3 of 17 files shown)


packages/wonder-blocks-dropdown/src/theme/default.ts
```diff
@@ -17,7 +18,7 @@ export default {
             },
         },
         shadow: {
-            default: `0 ${sizing.size_080} ${sizing.size_080} 0 ${semanticColor.core.shadow.transparent.low}`,
+            default: `${elevation.mid} ${semanticColor.core.shadow.transparent.mid}`,
```
jandrade: **note:** Turns out that OG uses a different value than TB, so I kept them separate.

packages/wonder-blocks-popover/src/components/popover-content-core.tsx
```diff
@@ -104,8 +105,7 @@ const styles = StyleSheet.create({
         borderRadius: border.radius.radius_040,
         border: `solid 1px ${semanticColor.core.border.neutral.subtle}`,
         backgroundColor: semanticColor.core.background.base.default,
-        // TODO(WB-1878): Use `elevation` token.
-        boxShadow: `0 ${spacing.xSmall_8}px ${spacing.xSmall_8}px 0 ${semanticColor.core.shadow.transparent.low}`,
+        boxShadow: `${elevation.mid} ${semanticColor.core.shadow.transparent.mid}`,
```
jandrade: **note:** Popover and Tooltip are using `mid` in OG, and `low` in TB. We haven't themed these two components yet, so that's probably something we need to reconcile soon. We might end up just using the same values for both themes, so I'll check with Caitlyn about this once she's back.

packages/wonder-blocks-dropdown/src/theme/thunderblocks.ts
```diff
@@ -20,7 +21,7 @@ export default mergeTheme(defaultTheme, {
             },
         },
         shadow: {
-            default: `0 ${sizing.size_020} ${sizing.size_020} 0 ${semanticColor.core.shadow.transparent.low}`,
+            default: `${elevation.low} ${semanticColor.core.shadow.transparent.low}`,
```
jandrade: **question:** I'm wondering if we should merge the current elevation values + the color in a single token as I'm seeing that all the cases are sharing `low` in both `elevation` and `semanticColor.core.shadow.transparent` categories. I'd appreciate any opinions on this.

__docs__/wonder-blocks-tokens/elevation.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+                name={packageConfig.name}
+                version={packageConfig.version}
+            />
+        ),
+    },
+    tags: ["!dev"],
+} as Meta;
+
+type Row = {label: string; css: string; value: string};
+const baseColumns = (kind: string) => [
```
beaesguerra: It looks like the `kind` argument isn't used! Is it something we can remove?

packages/wonder-blocks-dropdown/src/theme/thunderblocks.ts
beaesguerra: I think it would be nice to have tokens that combines the values already! One idea is, we'll have:
- the semantic color shadow tokens (pre-existing)
- the elevation tokens (this PR)
- boxShadow tokens (new) - that puts together the elevation + shadow tokens

To use the tokens, consumers could then do something like: 

```ts
const styles = {
  container: {
    boxShadow: boxShadow.low,
  },
};
```

and the box shadow tokens would look like: 
```ts
export const boxShadow = {
  low: `${elevation.low} ${semanticColor.core.shadow.transparent.low`,
 // ...
};
```

What do you think?

packages/wonder-blocks-modal/src/components/modal-header.tsx
```diff
@@ -137,7 +137,6 @@ const small = "@media (max-width: 767px)";
 
 const styles = StyleSheet.create({
     header: {
-        // TODO(WB-1878): Move this to an `elevation` theme token.
         boxShadow: `0px 1px 0px ${semanticColor.core.border.neutral.subtle}`,
```
beaesguerra, line 140: Would it make sense to use an elevation token here?

packages/wonder-blocks-popover/src/components/popover-content-core.tsx
beaesguerra: Sounds good! I wonder if we want to choose `low` for now since we have been favouring TB styles over OG. (Something we can decide later too!)

packages/wonder-blocks-tokens/src/theme/primitive/elevation.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const elevation = {
+    /**
+     * Elevation values for box shadows.
+     *
+     * NOTE: We use fixed values for the box shadows instead of using the
+     * regular `sizing` tokens to ensure that the shadows are consistent across
+     * different root font sizes.
+     */
+    low: `${sizing.size_0} ${SIZE_20_PX} ${SIZE_20_PX} ${sizing.size_0}`,
```
beaesguerra: Do we want to convert 0rem to px to so they're all the same unit? It probably doesn't make a difference, but it was something I noticed in the Chromatic snapshot!

packages/wonder-blocks-modal/src/components/modal-header.tsx
jandrade: I don't think so, this is a very specific use case in ModalHeader (it's the kind of style that attempts to replace `border` but without taking layout space).

packages/wonder-blocks-tokens/src/theme/default.ts
```diff
@@ -14,6 +15,9 @@ import {border} from "./primitive/border";
  */
 export default {
     border,
+    // We need to pass semanticColor to boxShadow to have access to the
+    // correct shadow colors.
+    boxShadow: boxShadow(semanticColor),
```
beaesguerra, line 20: Thanks for adding a comment! I was wondering why the values were returned from a function!

__docs__/wonder-blocks-tokens/box-shadow.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+    boxShadow,
+    semanticColor,
+    sizing,
+} from "@khanacademy/wonder-blocks-tokens";
+import ComponentInfo from "../components/component-info";
+import packageConfig from "../../packages/wonder-blocks-tokens/package.json";
+import {Code} from "../components/code";
+
+/**
+ * The `boxS` tokens are used to define the box-shadow of an element. This
```
beaesguerra:
```suggestion
 * The `boxShadow` tokens are used to define the box-shadow of an element. This
```

__docs__/wonder-blocks-tokens/box-shadow.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+                version={packageConfig.version}
+            />
+        ),
+    },
+    tags: ["!dev"],
+} as Meta;
+
+type Row = {label: string; css: string; value: string};
+
+export const Elevation = {
```
beaesguerra: Should the title of the story be `boxShadow` since it's the box shadow tokens?

Image

__docs__/wonder-blocks-tokens/box-shadow.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+ *   import {boxShadow} from "@khanacademy/wonder-blocks-tokens";
+ *   const styles = {
+ *       elevatedContainer: {
+ *           boxShadow: boxShadow.low,
+ *       },
+ *   };
+ * ```
+ */
+export default {
+    title: "Packages /Tokens / boxShadow",
```
beaesguerra: What do you think about having the casing match the other stories? For example, we have `Semantic Color` docs instead of `semanticColor`. Keeping this casing and changing the Semantic Color is fine with me too, as long as our naming is consistent!

```suggestion
    title: "Packages /Tokens / Box Shadow",
```

Image

__docs__/wonder-blocks-tokens/box-shadow.stories.tsx
```diff
@@ -0,0 +1,94 @@
+import * as React from "react";
```
beaesguerra, line 1: We could also add a link to these tokens in the Token Overview page! https://5e1bf4b385e3fb0020b7073c-kxahbsrdhm.chromatic.com/?path=/docs/packages-tokens-overview--docs&globals=theme:thunderblocks

packages/wonder-blocks-dropdown/src/theme/default.ts
```diff
@@ -17,7 +18,7 @@ export default {
             },
         },
         shadow: {
-            default: `0 ${sizing.size_080} ${sizing.size_080} 0 ${semanticColor.core.shadow.transparent.low}`,
+            default: boxShadow.mid,
```
beaesguerra, line 21: Nice!!

PR #2785, opened 2025-09-05
Create LayerRoot to integrate Announcer and Modal together

marcysutton:
## Summary:
The Announcer wasn't working from a modal context, such as in modal-level Toasts or other content announcements. This was because Modal had the `aria-modal` property set, while the Announcer was rendered outside of that layer. This PR moves Modals and the Announcer into a new LayerRoot context where `aria-modal` can be turned on and off to create the same context.

Issue: WB-2019

## Test plan:
1. Ensure tests pass
2. Review naming and structure of LayerRoot: should it live in wonder-blocks-announcer?
3. Determine if all necessary use cases for Modals have been covered, including custom building blocks


.changeset/brave-clocks-sparkle.md
```diff
@@ -0,0 +1,6 @@
+---
+"@khanacademy/wonder-blocks-announcer": minor
+"@khanacademy/wonder-blocks-modal": minor
```
marcysutton, line 3: Does minor make sense, or is this a big enough change to warrant a major?

packages/wonder-blocks-announcer/src/util/manage-layer-root.ts
```diff
@@ -0,0 +1,126 @@
...
+    layerRoot.setAttribute("aria-modal", (modalReferenceCount > 0).toString());
+}
+
+/**
+ * Gets the current modal state of the layer root
+ * @returns {boolean} Whether the layer root is currently in modal mode
+ */
+export function getLayerRootModalState(): boolean {
+    const layerRoot = getLayerRootElement();
+    return layerRoot?.getAttribute("aria-modal") === "true";
```
marcysutton, line 86: Sadly I think this entire approach won't work due to the requirements for `aria-modal`: it has to be on a `dialog` or `alertdialog` role, not on an ancestor 😭 

I didn't realize it until opening the PR!

marcysutton:
The approach in this PR didn't work because of the constraints for `aria-modal`, so I'm going to opt to close it and try another way! Stay tuned...

PR #2786, opened 2025-09-11
Allow ActionMenu to support any valid popper placement

MikeKlemarewski:
## Summary:
In some niche scenarios, we may want to position an ActionMenu somewhere other than below the opener. To support this, we can update the `alignment` prop to allow any valid popper placement as a value.

## Test plan:
- Ensure tests pass
- View storybook and make sure the ActionMenu stories are rendering correctly

### DIFF
--- a/__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
@@ -395,6 +395,24 @@ export const CustomOpener: StoryComponentType = {
  */
 export const WithPopperPlacement: StoryComponentType = {
     name: "With popper placement",
+    render: function Render(args) {
+        const [opened, setOpened] = React.useState(false);
+
+        React.useEffect(() => {
+            setOpened(true);
+        }, []);
+
+        return (
+            <ActionMenu
+                menuText="Betsy Appleseed"
+                {...args}
+                opened={opened}
+                onToggle={setOpened}
+            >
+                {actionItems.map((actionItem, index) => actionItem)}
+            </ActionMenu>
+        );
+    },
     args: {
         alignment: "right-start",
         opener: ({text}: any) => (
--- a/__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
@@ -389,6 +389,21 @@ export const CustomOpener: StoryComponentType = {
 };
 
 /**
+ * Sometimes you may want to align the dropdown somewhere besides below the
+ * opener. In these cases, you can specify any valid popper placement as the
+ * alignment.
+ */
+export const WithPopperPlacement: StoryComponentType = {
+    name: "With popper placement",
+    args: {
+        alignment: "right-start",
+        opener: ({text}: any) => (
+            <Button endIcon={IconMappings.caretRight}>{text}</Button>
+        ),
+    } as Partial<typeof ActionMenu>,
+};
+
+/**
  * You can use the `lang` attribute to specify the language of the action
  * item(s). This is useful if you want to avoid issues with Screen Readers
  * trying to read the proper language for the rendered text.
--- a/packages/wonder-blocks-dropdown/src/components/action-menu.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/action-menu.tsx
@@ -46,8 +46,9 @@ type Props = AriaProps &
          */
         selectedValues?: Array<string>;
         /**
-         * Whether this menu should be left-aligned or right-aligned with the
-         * opener component. Defaults to left-aligned.
+         * The alignment of the menu component in relation to the opener
+         * component. Defaults to "left", which is below the opener and left
+         * aligned. Any valid Popper placement is also supported.
          */
         alignment: "left" | "right" | Placement;
         /**
--- a/packages/wonder-blocks-dropdown/src/components/action-menu.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/action-menu.tsx
@@ -7,6 +7,7 @@ import {
     type StyleType,
 } from "@khanacademy/wonder-blocks-core";
 import {sizing} from "@khanacademy/wonder-blocks-tokens";
+import {Placement} from "@popperjs/core";
 import DropdownOpener from "./dropdown-opener";
 import ActionItem from "./action-item";
 import OptionItem from "./option-item";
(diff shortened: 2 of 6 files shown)


__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
```diff
@@ -388,6 +388,21 @@ export const CustomOpener: StoryComponentType = {
...
 };
 
+/**
+ * Sometimes you may want to align the dropdown somewhere besides below the
+ * opener. In these cases, you can specify any valid popper placement as the
+ * alignment.
+ */
+export const WithPopperPlacement: StoryComponentType = {
+    name: "With popper placement",
+    args: {
```
beaesguerra, line 416: Could this story default to being open? That way, the [Chromatic snapshot](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=68c33050087979ffca43ee7c) can cover that the popper is placed correctly!

packages/wonder-blocks-dropdown/src/components/dropdown-popper.tsx
```diff
@@ -36,7 +37,7 @@ type Props = {
      * Whether this menu should be left-aligned or right-aligned with the
      * reference component. Defaults to left-aligned.
      */
-    alignment?: "left" | "right";
+    alignment?: "left" | "right" | Placement;
```
beaesguerra, line 40: @jandrade Any concerns with supporting any placement from popper? I'm curious if you have context on if there was a reason we were only supporting left / right alignment (which mapped to `bottom-start` and `bottom-end`)!

packages/wonder-blocks-dropdown/src/components/action-menu.tsx
```diff
@@ -48,7 +49,7 @@ type Props = AriaProps &
          * Whether this menu should be left-aligned or right-aligned with the
          * opener component. Defaults to left-aligned.
          */
-        alignment: "left" | "right";
+        alignment: "left" | "right" | Placement;
```
beaesguerra, line 53: The jsdocs for the `alignment` prop can be updated since any placement would be supported!

beaesguerra: The argtypes for the prop docs need to be updated as well! https://5e1bf4b385e3fb0020b7073c-tbncjvdlap.chromatic.com/?path=/docs/packages-dropdown-actionmenu--docs

Image

__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
```diff
@@ -388,6 +388,21 @@ export const CustomOpener: StoryComponentType = {
...
+        alignment: "right-start",
```
beaesguerra, line 417: Neat! Popperjs handles language direction so the dropdown swaps sides when switching to RTL! (Consumers would need to update the opener button icon as needed)

Image

packages/wonder-blocks-dropdown/src/components/dropdown-popper.tsx
jandrade: No concerns at all, I think this makes total sense, as it allow us to adapt our UIs better for different constrains.

__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
MikeKlemarewski: https://github.com/user-attachments/assets/5f5f164f-0a87-41b4-8867-cb79fba4345f

I've updated this to default to open, but I think there may be a but with the ActionMenu component. It seems to render incorrectly if the default `opened` state is set to `true`. I'm guessing it's because the button size hasn't been computed yet so the popper places the menu incorrectly. Let me know if this is a known issue and if there's a workaround.

__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
```diff
@@ -388,6 +388,35 @@ export const CustomOpener: StoryComponentType = {
...
 
+/**
+ * Sometimes you may want to align the dropdown somewhere besides below the
+ * opener. In these cases, you can specify any valid popper placement as the
+ * alignment.
+ */
+export const WithPopperPlacement: StoryComponentType = {
+    name: "With popper placement",
+    render: function Render(args) {
+        const [opened, setOpened] = React.useState(true);
```
beaesguerra: A possible work around for the [issue](https://github.com/Khan/wonder-blocks/pull/2786/files#r2353774963):

```suggestion
        const [opened, setOpened] = React.useState(false);

        React.useEffect(() => {
            setOpened(true);
        }, []);
```

This makes it so it is opened after the button size has been computed!

__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
beaesguerra: Interesting! I think this is a different issue, I created https://khanacademy.atlassian.net/browse/WB-2087 for this!

I suggested a possible workaround for this PR: https://github.com/Khan/wonder-blocks/pull/2786/files#r2353799801

__docs__/wonder-blocks-dropdown/base-select.argtypes.ts
```diff
@@ -4,12 +4,29 @@ const argTypes: ArgTypes = {
...
+            "left-start",
+            "left-end",
+            "auto",
+            "auto-start",
+            "auto-end",
+        ],
         table: {
             category: "Layout",
-            type: {summary: `"left" | "right"`},
+            type: {summary: `"left" | "right" | Placement`},
```
beaesguerra: One more small request for docs! Could you apply these changes in [action-menu.argtypes.ts](https://github.com/Khan/wonder-blocks/blob/main/__docs__/wonder-blocks-dropdown/action-menu.argtypes.ts)? `base-select.argtypes.ts` is used for `SingleSelect` and `MultiSelect` components too, and those components currently only support the `left | right` placement options!

MikeKlemarewski: Ahhh, gotcha! I didn't see that other file. I'll move them.

__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
beaesguerra: The [snapshot](https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=68c9ed49f24107424574c36d) looks great, thanks for fixing that! 

Image

beaesguerra (approved):
Looks great! Thanks for contributing again to Wonder Blocks! 🚀

PR #2788, opened 2025-09-16
[WB-1852.3] Remove Link.light and Clickable.light variants

jandrade:
## Summary:

- Removes the `light` prop from `Clickable` and `Link` now that we can use `actionStyles.inverse` instead.
- This change also updates the `Link` component stories to use the actionStyles.inverse
style instead of the light prop for dark background links.
- Adds a `Link` example in the `actionStyles` statesheet snapshot.

Issue: https://khanacademy.atlassian.net/browse/WB-1852

## Test plan:

- Verify that all stories render correctly with the new styles applied.
- Check that the actionStyles.inverse style is applied to all dark background links.
- Ensure that the light prop is no longer used in any stories.

### DIFF
--- a/.changeset/poor-gorillas-marry.md
+++ b/.changeset/poor-gorillas-marry.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-clickable": major
+---
+
+Removes light prop from Clickable as it is no longer used/recommended. Please use `actionStyles.inverse` instead.
--- a/__docs__/wonder-blocks-clickable/clickable.argtypes.ts
+++ b/__docs__/wonder-blocks-clickable/clickable.argtypes.ts
@@ -40,21 +40,6 @@ export default {
     /**
      * States
      */
-    light: {
-        description:
-            "Whether the Clickable is on a dark colored background. Sets the default focus ring color to white, instead of blue. Defaults to false.",
-        defaultValue: false,
-        type: {
-            name: "boolean",
-            required: true,
-        },
-        table: {
-            category: "States",
-            type: {
-                summary: "boolean",
-            },
-        },
-    },
     disabled: {
         description: "Disables or enables the child; defaults to false",
         defaultValue: false,
--- a/__docs__/wonder-blocks-clickable/clickable.stories.tsx
+++ b/__docs__/wonder-blocks-clickable/clickable.stories.tsx
@@ -22,7 +22,6 @@ export default {
     args: {
         testId: "",
         disabled: false,
-        light: false,
         hideDefaultFocusRing: false,
     },
     decorators: [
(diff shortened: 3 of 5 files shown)


__docs__/wonder-blocks-link/link.stories.tsx
```diff
@@ -315,50 +294,6 @@ export const Inline: StoryComponentType = {
...
 };
 
-/**
- * Inline links include an underline to distinguish them from the surrounding
- * text. If the link is on a dark background, set the `light` prop to true for
- * it to be appropriately visible.
- *
- * **NOTE:** Secondary light links are not supported.
- */
-export const InlineLight: StoryComponentType = {
```
jandrade, line 325: **note:** I moved this example to the `actionStyles` scenarios.

__docs__/wonder-blocks-link/link.stories.tsx
```diff
@@ -196,8 +180,7 @@ export const StartAndEndIcons: StoryComponentType = {
                     startIcon={
                         <PhosphorIcon icon={IconMappings.plusCircleBold} />
                     }
-                    light={true}
-                    style={styles.standaloneLinkWrapper}
+                    style={[styles.standaloneLinkWrapper, actionStyles.inverse]}
```
beaesguerra, line 183: The active styling for the actionStyles.inverse style is different from before, is this intended?

| Before | After | 
| --- | --- |
| Image | Image |

beaesguerra (commented):
Nice to see we can simplify styles for Clickable and Link 🧹 ! Left some questions in the PR and Chromatic around the active styling for Links that use the inverse action styles!

__docs__/wonder-blocks-link/link.stories.tsx
jandrade: Yes, that's intended and it aligns with how we approach link styles in ThunderBlocks.

TB Figma for reference: https://www.figma.com/design/EuFu0U7gqc1koXm8ZhlOLp/%E2%9A%A1%EF%B8%8F-Components?node-id=5089-229&m=dev

jandrade:
I forgot to share that there's a parallel frontend PR converting Link.light instances to use `actionStyles`: https://github.com/Khan/frontend/pull/3761

__docs__/wonder-blocks-link/link.stories.tsx
beaesguerra: Showing the underline for active styling makes sense to me! I was unsure about the font color, since it no longer changes when pressed. The link styles in the TB designs shows the font color changing on press states, however, the designs are showing the normal link styling, not the inverse styles (which is expected since we want to move away from inverse styling!). At least there is a slight background change though so that there is some other visual change for users to know it was interacted with! I'll approve the changes and the related snapshots!

jandrade: >The link styles in the TB designs shows the font color changing on press states

I'm seeing that `hover` and `press` share the same font color. Do you see something different? AFAIK `rest` is a different color, but I'm not sure if I'm missing something.

beaesguerra: With the inverse styles, the `rest` font color and the `press` font color are now the same (white). Previously, they were different, and the TB styles have different font colors for `rest` and `press` states! (Let me know if you want to pair to chat, maybe I'm missing something also haha!)

PR #2789, opened 2025-09-16
Make useTabIndicator more reliable

beaesguerra:
## Summary:
[fix-tabs] set up mutation observer
[fix-tabs] add story to reproduce icon issue
[fix-tabs] rename to resize observer

Issue: WB-XXXX

## Test plan:

### DIFF
--- a/packages/wonder-blocks-tabs/src/components/navigation-tabs.tsx
+++ b/packages/wonder-blocks-tabs/src/components/navigation-tabs.tsx
@@ -110,16 +110,11 @@ export const NavigationTabs = React.forwardRef(function NavigationTabs(
         [],
     );
 
-    const {indicatorProps, updateUnderlineStyle} = useTabIndicator({
+    const {indicatorProps} = useTabIndicator({
         animated,
         tabsContainerRef: listRef,
         isTabActive,
     });
-
-    React.useEffect(() => {
-        // Update the underline style when children (tabs) change
-        updateUnderlineStyle();
-    }, [children, updateUnderlineStyle]);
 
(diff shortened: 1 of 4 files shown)


beaesguerra:
Putting the PR in review so I can test the npm snapshot in `frontend`! Will update the PR once I've verified the fix

packages/wonder-blocks-tabs/src/components/navigation-tabs.tsx
```diff
@@ -110,17 +110,12 @@ export const NavigationTabs = React.forwardRef(function NavigationTabs(
...
-    const {indicatorProps, updateUnderlineStyle} = useTabIndicator({
+    const {indicatorProps} = useTabIndicator({
         animated,
         tabsContainerRef: listRef,
         isTabActive,
     });
 
-    React.useEffect(() => {
-        // Update the underline style when children (tabs) change
-        updateUnderlineStyle();
```
beaesguerra, line 121: Since we have a MutationObserver that watches for descendant element and attribute changes, we no longer need to watch the props to trigger `updateUnderlineStyle`! This also makes it so width calculations happen after the DOM elements have changed, rather than when data has changed

kevinb-khan:
I think it might be worth add a high-level comment in packages/wonder-blocks-tabs/src/hooks/use-tab-indicator.ts letting people know that the animated active tab indicator is what's driving the need for a hook to compute its size. 🙏 Don't worry about adding a comment as part of this PR, it could be a follow up PR when you have some spare time. 🙂

PR #2794, opened 2025-09-18
Tabs: Set type on tabs to prevent form submission

beaesguerra:
## Summary:

Set `type=button` in the Tab component to prevent form submission when Tabs are used within a form element

Issue: WB-2089

## Test plan:
1. Confirm that tabs have type="button" set `?path=/story/packages-tabs-tabs--default`

Screenshot 2025-09-18 at 3 01 04 PM


jandrade (approved):
Makes total sense! thanks for the quick fix :) 🚀

PR #2797, opened 2025-09-22
[WB-1990] Consolidate Cell style props into a single styles prop

jandrade:
## Summary:

- `Cell`: Removes `contentStyle`, `leftAccessoryStyle`, `rightAccessoryStyle`
and `style` to consolidate it into a single `styles` prop object that contains
all keys.

- `Dropdown`: Updates Dropdown components internally to adapt to the new Cell
`styles` prop.

Issue: https://khanacademy.atlassian.net/browse/WB-1990

## Test plan:

Navigate to the `Cell` and `Dropdown` stories and verify that everything looks
correctly and that the styles are being applied correctly.

### DIFF
--- a/__docs__/wonder-blocks-cell/compact-cell.stories.tsx
+++ b/__docs__/wonder-blocks-cell/compact-cell.stories.tsx
@@ -170,18 +170,20 @@ export const CompactCellAccessoryStyles: StoryComponentType = {
             leftAccessory={
                 <PhosphorIcon icon={IconMappings.article} size="medium" />
             }
-            leftAccessoryStyle={{
-                minWidth: spacing.xxLarge_48,
-                alignSelf: "flex-start",
-                alignItems: "flex-start",
-            }}
             rightAccessory={
                 <PhosphorIcon icon={IconMappings.caretRightBold} size="small" />
             }
-            rightAccessoryStyle={{
-                minWidth: spacing.large_24,
-                alignSelf: "flex-end",
-                alignItems: "flex-end",
+            styles={{
+                leftAccessory: {
+                    minWidth: spacing.xxLarge_48,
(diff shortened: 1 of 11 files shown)


__docs__/wonder-blocks-cell/compact-cell.stories.tsx
```diff
@@ -221,8 +223,12 @@ export const CompactCellHorizontalRules: StoryComponentType = {
 };
 
 /**
- * Cell components can also adapt to different visual needs. One example of this
- * can be done by passing a custom style object to the `style` prop.
+ * `Compact` can be used with custom styles. The following parts can be
```
beaesguerra:
```suggestion
 * `CompactCell` can be used with custom styles. The following parts can be
```

packages/wonder-blocks-cell/src/util/types.ts
```diff
@@ -14,29 +14,6 @@ export type HorizontalRuleVariant = "full-width" | "inset" | "none";
...
-    alignSelf?: "flex-start" | "flex-end" | "center";
-    /**
-     * To horizontally align the accessory.
-     */
-    alignItems?: "flex-start" | "flex-end" | "center";
-    /**
-     * To set spacing between child elements.
-     */
-    gap?: number | string;
-};
```
beaesguerra, line 38: Thanks for switching it over to use `StyleType` instead! In general, do we want to prefer using `StyleType` for styles, rather than only accepting a subset of the CSS property types?

jandrade: I think so, specially as we are moving towards CSS modules where we need to make styles more flexible. With CSS modules (or CSS syntax itself) it will be harder to enforce subsets of CSS properties in CSS syntax.

PR #2799, opened 2025-09-22
Create initial Card component

marcysutton:
## Summary:
Adds a new component package for Cards

Implementation spec: https://khanacademy.atlassian.net/wiki/spaces/WB/pages/4353884294/Cards+Implementation+Spec

Issue: WB-1886

## Test plan:
1. Review props, structure, and usage 
2. Review docs and stories

## Implementation plan:
1. Add basic component (this PR)
2. Incorporate other variants from the `frontend` repo https://khanacademy.atlassian.net/browse/FEI-6310
3. Create a reusable close button https://khanacademy.atlassian.net/browse/WB-2090


__docs__/wonder-blocks-card/card.argtypes.tsx
```diff
@@ -0,0 +1,56 @@
+import type {ArgTypes} from "@storybook/react";
+
+export default {
+    children: {
+        control: {type: "text"},
+        description:
```
beaesguerra: Was Storybook able to get the prop docs from the comments? I find most of the time it is able to, unless the prop types are more complex!

__docs__/wonder-blocks-card/card.stories.tsx
```diff
@@ -0,0 +1,306 @@
...
+                </View>
+                <Button kind="secondary">Learn more about gems</Button>
+                <Button kind="tertiary">View previous challenges</Button>
+            </View>
+        </Card>
+    ),
+};
+
+/**
+ * Cards also have the option to display a "close" button that can dismiss the card from the DOM.
```
beaesguerra, line 130: **question**: Once a card is dismissed, where should a user's focus move to? Is it something we want to handle in the component (maybe the next/previously focused thing)? Or should we leave focus management up to consumers? If so, we could document guidelines on where focus should go next!

packages/wonder-blocks-card/src/components/card.tsx
```diff
@@ -0,0 +1,184 @@
...
+    /**
+     * The content for the card.
+     */
+    children: React.ReactNode;
+    /**
+     * A set of localizable labels for this component, including a dismiss button
+     * and the card itself, if marked as an HTML region.
+     */
+    labels?: {
+        dismissButtonAriaLabel?: string;
```
beaesguerra: idea: Could we make `dismissButtonAriaLabel` required if `onDismiss` is provided? That way, consumers know they need to pass in a translated string for the close button!

__docs__/wonder-blocks-card/card.stories.tsx
```diff
@@ -0,0 +1,306 @@
...
+        onDismiss: () => {},
+        labels: {dismissButtonAriaLabel: "Dismiss"},
+    },
+    render: (args) => {
+        const CardWithRef = () => {
+            const wrapperRef = React.useRef<HTMLDivElement>(null);
+            const cardRef = React.useRef<HTMLDivElement>(null);
+            const onDismiss = () => {
+                // Remove the card from the DOM
+                cardRef.current?.remove();
```
beaesguerra, line 141: The card doesn't get removed for me when I click the close button! https://5e1bf4b385e3fb0020b7073c-macgfgvdbn.chromatic.com/?path=/story/packages-card-card--with-dismiss-button

As an alternative to removing the element from the DOM, maybe we could use state to not render the card once the close button is pressed!

__docs__/wonder-blocks-card/card.stories.tsx
```diff
@@ -0,0 +1,306 @@
...
+                <Heading>Card 1</Heading>
+            </Card>
+            <Card tag="li" styles={{root: styles.card}}>
+                <Heading>Card 2</Heading>
+            </Card>
+        </View>
+    ),
+};
+
+export const InAStack: StoryComponentType = {
```
beaesguerra, line 185: Cool example!

__docs__/wonder-blocks-card/card.stories.tsx
```diff
@@ -0,0 +1,306 @@
...
+                <Card styles={{root: styles.card}}>
+                    <Heading>Card 4</Heading>
+                    <BodyText>This is a basic card.</BodyText>
+                </Card>
+            </View>
+        </>
+    ),
+};
+
+export const WithStyleProps: StoryComponentType = {
```
beaesguerra, line 244: Do we want to add documentation around styling cards? I think we prefer consumers to use the dedicated props (borderRadius, backgroundColor, etc) over `styles` where possible so only a subset of options are available!

packages/wonder-blocks-card/src/components/card.tsx
```diff
@@ -0,0 +1,184 @@
...
+     * Optional styles to be applied to the root element and the dismiss button.
+     */
+    styles?: {
+        root?: StyleType;
+        dismissButton?: StyleType;
+    };
+    /**
+     * A ref that will be passed to the root element (i.e. the card container).
+     */
+    ref?: React.Ref<any>;
```
beaesguerra: I'm wondering if a specific type could be used for the ref, maybe `React.Ref<HtmlElement>`?

packages/wonder-blocks-card/src/components/card.tsx
```diff
@@ -0,0 +1,184 @@
...
+     * The content for the card.
+     */
+    children: React.ReactNode;
+    /**
+     * A set of localizable labels for this component, including a dismiss button
+     * and the card itself, if marked as an HTML region.
+     */
+    labels?: {
+        dismissButtonAriaLabel?: string;
+        cardAriaLabel?: string;
```
beaesguerra: When should an aria label for the card be provided? Is this something we could document guidelines for? (I haven't looked into what the best practices for cards are yet!)

packages/wonder-blocks-card/src/components/card.tsx
```diff
@@ -0,0 +1,184 @@
...
+     * The padding inside the card, as a string identifier that matches a sizing token.
+     * This can be one of:
+     * - `"size_0"`, matching `sizing.size_0`: no padding, useful for cards that need to have content flush with the edges.
+     * - `"size_160"`, matching `sizing.size_160`: moderate padding, useful for cards that need to have some space between the content and the edges.
+     * - `"size_240"`, matching `sizing.size_240`: more padding, useful for cards that need to have more space between the content and the edges.
+     *
+     * Default: `"size_160"`
+     */
+    padding?: "size_0" | "size_160" | "size_240";
+};
```
beaesguerra: **thought**: When the props are named after CSS properties, it could be confusing when to know to use these props instead of custom styles! This reminded me of how we started moving away from specific typed styles that limits what properties could be styled ([related comment](https://github.com/Khan/wonder-blocks/pull/2797#discussion_r2370267139))

As another approach, I wonder if we could have more semantic names for the styling options? Some ideas:
- `backgroundColor` => `kind` and the options could be `standard` or `subtle`/`neutral`
  - I'm always weary about naming things `default` because if the default thing ever changes, then we get into a weird state where something called "default" is no longer actually the default!)
  - I was initially thinking of `variant`, though `kind` is more consistent with other components like Buttons
- `padding`: the options could be `none`, `small`, `medium` (or something like that)
  - We could also name the prop `paddingSize` or something so it is different from the css property name
- `borderRadius` => the options could be `small`, `medium` (or something like that)
  - Could also name the prop to `radius` or something so it's also different from the css property name

By doing this, we also decouple the component API from the implementation. For example, if we wanted to support `radius_160` instead of `radius_120`, we could more easily change the value we use in the component without having to release breaking changes (and migrate existing usage 😅 )!

We probably want Caitlyn's input on this too since we could sync what properties/options are available in the Figma component too 😄 cc: @jandrade too in case you have thoughts on this!

packages/wonder-blocks-card/src/components/card.tsx
```diff
@@ -0,0 +1,184 @@
...
+ */
+
+const Card = React.forwardRef(function Card(
+    props: Props,
+    ref: React.ForwardedRef<any>,
+) {
+    const {
+        styles,
+        labels,
+        tag,
```
beaesguerra, line 154: Should we provide a default tag? `div`?

__docs__/wonder-blocks-card/card.argtypes.tsx
```diff
@@ -0,0 +1,56 @@
+import type {ArgTypes} from "@storybook/react";
+
+export default {
```
beaesguerra, line 3: Can we set up controls for the props that can only be specific values? (backgroundColor, padding, borderRadius, etc!) 

Image

packages/wonder-blocks-card/src/components/card.tsx
```diff
@@ -0,0 +1,184 @@
...
+            {children}
+        </View>
+    );
+});
+
+const getComponentStyles = ({
+    backgroundColor,
+    borderRadius,
+    padding,
+}: AllowedStyleProps) => {
```
beaesguerra: Something I started to do in WB was to create explicit styles for variants instead of creating dynamic styles since _I think_ we won't be able to do things as dynamically with css modules. My hope too is that explicitly laying out the styles will make it easier to code mod to css modules when the time comes! 

(These are my assumptions though, curious is either you or @jandrade have thoughts on this too!)

packages/wonder-blocks-card/src/components/dismiss-button.tsx
```diff
@@ -0,0 +1,61 @@
...
+            testId={testId}
+        />
+    );
+};
+
+const componentStyles = StyleSheet.create({
+    root: {
+        position: "absolute",
+        // insetInlineEnd supports both RTL and LTR layouts
+        insetInlineEnd: sizing.size_080,
```
beaesguerra, line 50: 🎉

beaesguerra (commented):
Yayy a card component! Looking good so far, I left some questions and comments 😄

__docs__/wonder-blocks-card/card.argtypes.tsx
marcysutton: Yes, good call!

__docs__/wonder-blocks-card/card.stories.tsx
marcysutton: Added a docs comment!

packages/wonder-blocks-card/src/components/card.tsx
marcysutton: Yep, I added it to the typing!

__docs__/wonder-blocks-card/card.stories.tsx
marcysutton: Ah, it was an issue with `args` conflicting with the inline props. I cleaned those up and it works now!

__docs__/wonder-blocks-card/card.stories.tsx
marcysutton: Yes, I added some more docs!

packages/wonder-blocks-card/src/components/card.tsx
marcysutton: I added some typings and docs for this. Basically, it's when a Document structure role is used that allows `aria-label` (`section` or `figure` seemed the most appropriate for cards).

packages/wonder-blocks-card/src/components/card.tsx
marcysutton: Made some updates! I ended up using `backgroundColorStyle` over `kind` because it seemed more clear for something limited to background-color -- lower cognitive load for understanding! When I do another revision for `backgroundImage`, it could change a bit too... stay tuned.

packages/wonder-blocks-card/src/components/card.tsx
marcysutton: It has a default of `div` already, from the underlying `View` tag! It's commented in the docs too.

packages/wonder-blocks-card/src/components/card.tsx
marcysutton: I think I'll opt to do a refactor later on so this work can keep moving. Good call, though!

packages/wonder-blocks-card/src/components/card.tsx
```diff
@@ -0,0 +1,277 @@
...
+    backgroundColorStyle?: "base-subtle" | "base-default";
+    /**
+     * The border radius of the card, as a string identifier that matches a border.radius token.
+     * This can be one of:
+     * - `"radius_080"`, matching `border.radius.radius_080`.
+     * - `"radius_120"`, matching `border.radius.radius_120`.
+     *
+     * Default: `"radius_080"`
+     */
+    borderRadiusStyle?: "small" | "medium";
```
jandrade: **suggestion:** It would be better not to include the `...Style` suffix at the end of these props. It would be more consistent with `paddingSize` + `elevation` props.

packages/wonder-blocks-card/src/components/card.tsx
```diff
@@ -0,0 +1,277 @@
...
+});
+
+const getComponentStyles = ({
+    backgroundColorStyle,
+    borderRadiusStyle,
+    paddingSize,
+    elevation,
+}: StyleOnlyProps) => {
+    // Map prop values to tokens
+    const styleMap = {
```
jandrade, line 198: **praise:** Neat approach!

packages/wonder-blocks-card/src/components/card.tsx
```diff
@@ -0,0 +1,277 @@
...
+      }
+    | {
+          tag?: Exclude<keyof JSX.IntrinsicElements, "section" | "figure">;
+          labels?: Record<string, any>;
+      };
+
+/**
+ * Combined props - these two requirements work independently
+ */
+type ConditionalProps = BaseCardProps & TagProps & DismissProps;
```
jandrade: **suggestion:** It feel that this could just be renamed as `Props` at this point (not sure if you are planning to augment this type in a next PR).

packages/wonder-blocks-card/src/components/card.tsx
```diff
@@ -0,0 +1,277 @@
...
+ * A callback function to handle dismissing the card. When this prop is present,
+ * a dismiss button with an X icon will be rendered.
+ *
+ * When `onDismiss` is provided, `labels.dismissButtonAriaLabel` must also be
+ * provided for accessibility and localization.
+ */
+type DismissProps =
+    | {
+          onDismiss: (e?: React.SyntheticEvent) => void;
+          labels: {dismissButtonAriaLabel: string} & Record<string, any>;
```
jandrade, line 55: **suggestion:** Using discriminated union types is a great approach for enforcing a11y here!

I think this would be a great opportunity to add a `*.typestest.tsx` file here to ensure that these types work as expected.

packages/wonder-blocks-card/src/components/card.tsx
```diff
@@ -0,0 +1,277 @@
...
+type StyleOnlyProps = {
+    /**
+     * The background color of the card, as a string identifier that matches a semanticColor token.
+     * This can be one of:
+     * - `"base-subtle"` `semanticColor.core.background.base.subtle`: a light gray background.
+     * - `"base-default"`, matching `semanticColor.core.background.base.default`: a white background.
+     *
+     * Default: `"base-default"`
+     */
+    backgroundColorStyle?: "base-subtle" | "base-default";
```
jandrade: **thought:** One more thing to consider is that this will support images in the future, so it would be better to expand the naming a bit so this prop doesn't get too limited from start (and we avoid having to introduce breaking changes in the future).

packages/wonder-blocks-card/src/components/card.tsx
marcysutton: Totally makes sense, thanks for pointing it out!

packages/wonder-blocks-card/src/components/card.tsx
marcysutton: Good call, I added one!

packages/wonder-blocks-card/src/components/card.tsx
marcysutton: Good thinking. I changed it to `background`, with a plan to add support for images with the same prop. TBD...

__docs__/wonder-blocks-card/card.argtypes.tsx
```diff
@@ -0,0 +1,131 @@
...
+    children: {
+        control: {type: "text"},
+        table: {
+            type: {summary: "React.ReactNode"},
+        },
+        type: {name: "other", value: "React.ReactNode", required: true},
+    },
+    backgroundColorStyle: {
```
beaesguerra: `backgroundColorStyle` and `borderRadiusStyle` prop names in the argtypes can be updated to the new names!

packages/wonder-blocks-card/src/components/card.tsx
```diff
@@ -0,0 +1,269 @@
...
+ * | `small` | `sizing.size_160` |
+ * | `medium` | `sizing.size_240` |
+ *
+ * Other styles can be applied via the `styles` prop, which accepts an object with `root` and `dismissButton` keys for custom styling of the card container and dismiss button respectively.
+ *
+ * ### Accessibility
+ * - When using `tag="section"` or `tag="figure"`, the `labels.cardAriaLabel` prop is required to provide an accessible name for the card.
+ * - When the `onDismiss` prop is provided, a dismiss button will be rendered. In this case, the `labels.dismissButtonAriaLabel` prop is required to provide an accessible label for the dismiss button.
+ * - Cards cannot be interactive elements themselves (e.g. buttons or links), but they can contain interactive elements as children.
+ */
```
beaesguerra: What do you think about moving these docs to a story description? Since the component documentation is long, the props table is no longer near the top! 
https://5e1bf4b385e3fb0020b7073c-uhnlrtvxlk.chromatic.com/?path=/docs/packages-card-card--docs

packages/wonder-blocks-card/src/components/card.tsx
beaesguerra: @marcysutton When should `section` or `figure` be used for cards? Might also be helpful to document what to consider when making that decision!

beaesguerra (approved):
The Card component looks great! 🚀 Thanks for shaping the types to encourage accessibility best practices! ❤️ 

Left a few non-blocking comments around docs!

packages/wonder-blocks-card/src/components/card.tsx
marcysutton: Hm, I'm not sure what this meant exactly! Maybe it was referring to a specific story? Maybe there could be an `Overview.mdx` file or something more editorial? I think I'll opt to leave it for this PR and revisit in an update.

__docs__/wonder-blocks-card/card.stories.tsx
```diff
@@ -0,0 +1,312 @@
...
+/**
+ * The most basic Card has simple contents as `children`.
+ */
+export const DefaultCard: StoryComponentType = {
+    render: () => (
+        <Card styles={{root: styles.card}}>
+            <Heading>Some Contents</Heading>
+            <BodyText>This is a basic card.</BodyText>
+        </Card>
+    ),
```
jandrade: **suggestion:** It would be great if this story could use the `args` object (instead of `render:`), that way we can interact with it in the `Controls` panel.

e.g.
```
    args: {
        styles: {{root: styles.card}},
        children: (
            <>
                    <Heading>Some Contents</Heading>
                    <BodyText>This is a basic card.</BodyText>
                </>
            ),
       }
```

We tend to use the first story as an interactive playground, so it makes total sense to keep the other ones intact :)

jandrade (approved):
Looks really great! thanks for adding this component to WB! 👏 🚀

__docs__/wonder-blocks-card/card.stories.tsx
marcysutton: Great idea, thanks @jandrade!

PR #2800, opened 2025-09-22
TextArea: Support dynamic content sizing

beaesguerra:
## Summary:

Implements the following changes to the TextArea component:
- `autoResize (defaults to false)` (new) - enables the textarea element to grow/shrink based on the content 
- `maxRows` (new) - limits the number of rows shown before becoming scrollable. This prevents the textarea from infinitely expanding
- `resizeType` (deprecated) - we mark this prop as deprecated in favour of the `autoResize` prop. We don't remove this prop yet so that this change isn't a breaking change. It'll be removed in a future version

Context around decisions made for this implementation: 
- Using JS to calculate the needed content height when the content changes or a resize observer is triggered
  - Explored using [field-sizing](https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing) for dynamic content sizing. The drawbacks were:
    - it is experimental still and not fully supported by all browsers. We would need the JS logic anyways as a fallback. It will be less maintenance for us to have one implementation for the behaviour, rather than switching implementations based on browser support!
    - `field-sizing` makes it so the width of an element is also dynamic, not just the height. This would be a more involved change since it would have an impact on the current TextArea usage. We only need dynamic content sizing vertically, which we can control with the custom implementation
- `autoResize` defaulting to `false`
  - I think long term, we want `autoResize` to default to `true`. I propose we let it be something opted into so that this initial release isn't a breaking change. Also, since we are manually updating the height when content changes, it is less performant than when the height is static and the TextArea becomes scrollable. Let me know though if anyone has any thoughts on this! 

Issue: WB-1843

## Test plan:
1. Confirm existing TextArea features work as expected `/?path=/docs/packages-form-textarea--docs`
2. Confirm TextArea auto-resizing works as expected (`/?path=/story/packages-form-textarea--auto-resize` and `?path=/story/packages-form-testing-snapshots-textarea--scenarios`). Some test cases:
  - adding text or new lines should cause the textarea to grow (up until the maxRows point)
  - if textarea has reached the max number of rows, the textarea should be scrollable
  - removing text should cause the textarea to shrink
  - pasting in text or removing a chunk of text should cause the textarea to grow/shrink accordingly
  - changing the zoom level or screen size should cause the textarea to grow/shrink accordingly
  - similar test cases for placeholder content, instead of the textarea value content

### DIFF
--- a/packages/wonder-blocks-form/src/components/__tests__/text-area.test.tsx
+++ b/packages/wonder-blocks-form/src/components/__tests__/text-area.test.tsx
@@ -213,21 +213,6 @@ describe("TextArea", () => {
             expect(textArea).toHaveClass(className);
         });
 
-        it("should set the rows attribute when the rows prop is provided", async () => {
-            // Arrange
-            const rows = 10;
-            render(
-                <TextArea value="Text" onChange={() => {}} rows={rows} />,
-                defaultOptions,
-            );
-
-            // Act
-
-            // Assert
-            const textArea = await screen.findByRole("textbox");
-            expect(textArea).toHaveAttribute("rows", `${rows}`);
-        });
-
         it("should set the spellcheck attribute when spellCheck prop is set to true", async () => {
             // Arrange
             render(
--- a/packages/wonder-blocks-form/src/components/text-area.tsx
+++ b/packages/wonder-blocks-form/src/components/text-area.tsx
@@ -331,7 +331,7 @@ const TextArea = React.forwardRef<HTMLTextAreaElement, TextAreaProps>(
                             : {
                                   // Dynamically set the height if field-sizing is
                                   // not supported
-                                  height: `${height}px`,
+                                  height: `calc(${height}px + 2px)`,
                               },
                         style,
                     ]}
--- a/packages/wonder-blocks-form/src/components/text-area.tsx
+++ b/packages/wonder-blocks-form/src/components/text-area.tsx
@@ -212,7 +212,7 @@ const StyledTextarea = addStyle("textarea");
  * Height = (number of rows * line height) + (2 * vertical padding) + (2 * border width)
  */
 function getHeightForNumberOfRows(rows: number) {
-    return `calc(${rows}lh + (2 * ${theme.field.layout.paddingBlock}) + (2 * ${border.width.thin}))`;
+    return `calc((${rows} * ${font.body.lineHeight.medium}) + (2 * ${theme.field.layout.paddingBlock}) + (2 * ${border.width.thin}))`;
 }
 
 /**
--- a/packages/wonder-blocks-form/src/components/text-area.tsx
+++ b/packages/wonder-blocks-form/src/components/text-area.tsx
@@ -184,6 +184,7 @@ type TextAreaProps = AriaProps & {
      */
     required?: boolean | string;
     /**
+     * @deprecated This prop is deprecated in favour of the `autoResize` prop.
      * Specifies the resizing behaviour for the textarea. Defaults to both
      * behaviour. For more details, see the [CSS resize property values MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/resize#values)
      */
--- a/packages/wonder-blocks-form/src/components/text-area.tsx
+++ b/packages/wonder-blocks-form/src/components/text-area.tsx
@@ -221,6 +221,33 @@ function getHeightForNumberOfRows(rows: number) {
 }
 
 /**
+ * Calculate the height of a textarea element.
+ * @param textArea - The textarea element.
+ * @returns The height of the textarea element.
+ */
+function getTextAreaHeight(textArea: HTMLTextAreaElement) {
+    // Save the original style properties
+    const originalHeight = textArea.style.height;
+    const originalOverflow = textArea.style.overflow;
+
+    // Force the textarea to shrink by setting height to 0 and hiding overflow
+    textArea.style.setProperty("height", "0px", "important");
+    textArea.style.setProperty("overflow", "hidden", "important");
+    // Now get the actual scrollHeight needed for the content
+    const newHeight = textArea.scrollHeight;
+
+    // Restore the original styles (remove !important)
+    textArea.style.removeProperty("height");
+    textArea.style.removeProperty("overflow");
+    if (originalHeight) {
+        textArea.style.height = originalHeight;
+    }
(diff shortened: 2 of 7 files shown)


packages/wonder-blocks-form/src/components/text-area.tsx
```diff
@@ -296,7 +406,6 @@ const TextArea = React.forwardRef<HTMLTextAreaElement, TextAreaProps>(
                     autoComplete={autoComplete}
                     name={name}
                     autoFocus={autoFocus}
-                    rows={rows}
```
beaesguerra, line 299: We no longer use the `rows` attribute since setting the `height` will override it

packages/wonder-blocks-form/src/components/__tests__/text-area.test.tsx
```diff
@@ -213,21 +213,6 @@ describe("TextArea", () => {
             expect(textArea).toHaveClass(className);
         });
 
-        it("should set the rows attribute when the rows prop is provided", async () => {
```
beaesguerra, line 216: This test is no longer needed since we don't set the rows attribute anymore since a dynamic height overrides it

packages/wonder-blocks-form/src/components/text-area.tsx
```diff
@@ -181,14 +184,75 @@ type TextAreaProps = AriaProps & {
      */
     required?: boolean | string;
     /**
+     * @deprecated This prop is deprecated in favour of the `autoResize` prop.
```
beaesguerra, line 187: Deprecating the `resizeType` prop so that we can start moving away from it in favour of the `autoResize` prop.

We don't remove the prop yet so we can switch to the `autoResize` prop before releasing the breaking change!

packages/wonder-blocks-form/src/components/text-area.tsx
```diff
@@ -181,14 +184,75 @@ type TextAreaProps = AriaProps & {
...
+    textArea.style.setProperty("overflow", "hidden", "important");
+
+    // Get the scrollHeight needed for the content. We account for the border
+    // width so the scrollbar is not shown.
+    const borderTop = style.borderTopWidth;
+    const borderBottom = style.borderBottomWidth;
+    const newHeight = `calc(${textArea.scrollHeight}px + ${borderTop} + ${borderBottom})`;
+
+    // Restore the original styles (remove !important)
+    textArea.style.removeProperty("height");
```
marcysutton: I didn't observe any obvious jank testing in Storybook, but I wondered if setting/removing/setting style properties in succession would cause layout thrashing. Maybe it happens so fast it isn't noticeable?

packages/wonder-blocks-form/src/components/text-area.tsx
marcysutton: Glad to see that `rows` is still used for the Textarea API, though! It would be hairy to expose `height` directly to consumers.

marcysutton (approved):
This looks really solid! I checked out [field-sizing on Caniuse](https://caniuse.com/mdn-css_properties_field-sizing) and `resize` on MDN and support really is lacking for those...so this makes a lot of sense! I also tested out the stories and they are super smooth. Looks good to me!

packages/wonder-blocks-form/src/components/text-area.tsx
beaesguerra: Good question! I haven't observed anything like that while testing! Since it's setting styles on the DOM, it won't trigger any React re-renders until we update `height` in the component state. I plan to do more testing using the snapshot in `frontend`, and will keep an eye out for any layout shifts!

I'm also open to other suggestions on how to measure the content height for the textarea! Looking online, another possible solution was to have a hidden textarea element with mirrored content so we could get the measurements of that. I felt it was unexpected to render to textarea elements so I went with this approach. Let me know if anyone has other ideas on how to achieve this :)

beaesguerra:
Did some integration testing with frontend and updated the PR to fix some issues! For more details around testing in frontend, see https://github.com/Khan/frontend/pull/4019 . The PR and Chromatic snapshots are ready for review again!

Note: The `autoResize` prop defaults to `false`. Teams can decide when/where this should be enabled! When these changes are released and integrated into `frontend`, it'll only include the package version updates.

I created https://khanacademy.atlassian.net/browse/WB-2095 to remove the deprecated `resizeType` prop (which will be a breaking change). Depending on the use case, we can decide then if certain usage should be updated to use `autoResize` instead!

packages/wonder-blocks-form/src/components/text-area.tsx
```diff
@@ -271,8 +366,65 @@ const TextArea = React.forwardRef<HTMLTextAreaElement, TextAreaProps>(
...
 
+        useOnMountEffect(() => {
+            // If autoResize is false, we don't need to set up the resize observer
+            if (!autoResize) {
+                return;
+            }
+            // We use a ref to the container to get the child textarea element
+            // so that consumers can pass in a ref to the textarea element
+            // directly.
+            const ref = textAreaContainerRef.current?.children[0];
```
jandrade, line 377: **praise:** Thanks for adding this comment. Really helps to understand why you use this query here.

packages/wonder-blocks-form/src/components/text-area.tsx
```diff
@@ -181,14 +184,93 @@ type TextAreaProps = AriaProps & {
...
 };
 
 const StyledTextarea = addStyle("textarea");
 
+/**
+ * Calculate the height of x number of rows, including padding and border.
+ *
+ * Height = (number of rows * line height) + (2 * vertical padding) + (2 * border width)
+ */
+function getHeightForNumberOfRows(rows: number) {
```
jandrade, line 220: **thought:** Something to consider for later..... I wonder how we could approach things like this with CSS modules.

jandrade (approved):
Looks great to me! 🚀 :shipit:

packages/wonder-blocks-form/src/components/text-area.tsx
beaesguerra: Good question! One idea is we could have the tokens available in JS too so we could set inline styles on elements. If we decide to have tokens available in different forms (JS, css variables), we could consider using something like [Style Dictionary](https://styledictionary.com/) to export into different formats. I'm not yet sure how it'd work with how we do theming, more to explore and consider for later! 😄

PR #2808, opened 2025-10-01
[WB-2101] Fix left-aligned dropdown components to be right-aligned

jandrade:
## Summary:

- Cell: Updates styles to support RTL.
- Tooling: Adds a `decorator` field to ScenariosLayout.
- Docs: adds RTL stories for Cell and SingleSelect.

Issue: https://khanacademy.atlassian.net/browse/WB-2101

## Test plan:

Navigate to the docs and verify that the new RTL stories render correctly.

- `/?path=/story/packages-dropdown-singleselect--right-to-left&globals=backgrounds.value:baseDefault;backgrounds.grid:!undefined;;direction:rtl`

### DIFF
--- a/.changeset/happy-trains-swim.md
+++ b/.changeset/happy-trains-swim.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-cell": patch
+---
+
+Updates styles to support RTL
--- a/__docs__/components/scenarios-layout.tsx
+++ b/__docs__/components/scenarios-layout.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import * as React from "react";
 import type {StrictArgs} from "@storybook/react-vite";
 import {StyleSheet} from "aphrodite";
 import {StyleType, View} from "@khanacademy/wonder-blocks-core";
@@ -6,7 +6,11 @@ import {LabelLarge} from "@khanacademy/wonder-blocks-typography";
 import {sizing} from "@khanacademy/wonder-blocks-tokens";
 
 type Props = {
-    scenarios: {name: string; props: StrictArgs}[];
+    scenarios: {
+        name: string;
+        props: StrictArgs;
+        decorator?: React.ReactElement;
+    }[];
     children: (props: any, name: string) => React.ReactNode;
     styles?: {
         root?: StyleType;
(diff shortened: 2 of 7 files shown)


packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -238,7 +238,7 @@ const styles = StyleSheet.create({
         // Hide overflow so that if custom styling applies a border radius, the
         // left visual indicator for press/active states does not overflow
         overflow: "hidden",
-        textAlign: "left",
+        textAlign: "start",
```
jandrade, line 241: **note:** Here you can see how it is fixed:

Original snapshot: https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=68dd938ef1dde9d96297ecd0

Most recent snapshot: https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=68dd960da7c6724a55b2d2e5

__docs__/components/scenarios-layout.tsx
```diff
@@ -23,10 +27,16 @@ export const ScenariosLayout = (props: Props) => {
...
+                const {decorator: Decorator} = scenario;
                 return (
                     <View key={scenario.name} style={styles.scenario}>
                         <LabelLarge>{scenario.name}</LabelLarge>
-                        {children(scenario.props, scenario.name)}
+                        {Decorator
+                            ? React.cloneElement(Decorator, {
+                                  children: renderer,
+                              })
+                            : renderer}
```
jandrade, line 39: **note:** Added this so we can add decorators/wrappers to the scenarios for specific cases.

.changeset/happy-trains-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-cell": patch
+---
+
+Updates styles to support RTL
```
beaesguerra, line 5: I think the changeset got created twice for some reason!

__docs__/components/scenarios-layout.tsx
beaesguerra: Neat! I've been setting the language dir for whole stories, it makes sense to support adding it only for specific scenarios!

__docs__/wonder-blocks-cell/detail-cell-testing-snapshots.stories.tsx
```diff
@@ -179,6 +179,11 @@ export const Scenarios: Story = {
                     rightAccessory: undefined,
                 },
             },
+            {
+                name: "Right to Left",
+                props: {...defaultProps, title: longTextWithNoWordBreak},
+                decorator: <div dir="rtl" />,
+            },
```
beaesguerra, line 186: Image

thought (for later): I wonder if there's something we can do so that chevron left/right arrow icons will automatically respond to the language direction.

One idea is to componentize the arrow icon so that it'll automatically flip based on the `dir`, so consumers don't have to repeat this logic everywhere! We could also componentize any other icons that we expect to change with language direction. They could live in `wonder-blocks-icon`, similar to the `GemIcon` and `StreakIcon` components.

beaesguerra (approved):
Looks great! Thanks for updating ScenariosLayout to be more flexible 🎉

PR #2809, opened 2025-10-01
Adding Get Started docs for set up instructions

beaesguerra:
## Summary:

Setting up documentation around how to set up Wonder Blocks in an application.

Note: I've created WB-2102 for writing instructions for importing fonts!

Issue: WB-1892

## Test plan:

- Review the Get Started docs `?path=/docs/get-started--docs`

I set up a code sandbox to figure out what steps are needed to set up WB. https://codesandbox.io/p/sandbox/wb-set-up-4k4n6v


__docs__/_get-started_.mdx
beaesguerra, line 1: Let me know if I'm missing anything, or feel free to update this as well!

__docs__/_get-started_.mdx
```diff
@@ -0,0 +1,187 @@
...
+1. Set the root font size to `62.5%` so that `1rem` is equal to `10px`. Many of
+the Wonder Blocks components and typography use rems based on this unit.
+
+```css
+html {
+  font-size: 62.5%;
+}
+```
+
+2. More details to come on how to import fonts!
```
beaesguerra, line 56: I was taking a look at how the fonts were set up for thunderblocks in webapp: https://github.com/Khan/webapp/pull/32961

I saw we added the related font files and defined the css font faces in webapp/frontend. I'm wondering if it would make sense to include those in WB so that consumers don't need to set these things up in their project. And we could modify any fonts internally within WB. Let me know what you think!

For now, I've included a placeholder for these steps and created WB-2102 to fill in more details about importing fonts!

__docs__/_get-started_.mdx
```diff
@@ -0,0 +1,187 @@
+import {Meta} from "@storybook/blocks";
+
+<Meta title="Get Started" />
+
+# Get Started
+
+## Design
+
+The design system components are available in Figma:
+- <a href="https://www.figma.com/design/EuFu0U7gqc1koXm8ZhlOLp/%E2%9A%A1%EF%B8%8F-Components?node-id=1-2">Wonder Blocks Component Library (Thunderblocks)</a>
```
beaesguerra: Using anchor tags for links so that they don't open in the Storybook iframe

__docs__/_get-started_.mdx
```diff
@@ -0,0 +1,187 @@
...
+
+#### Typography
+1. Set the root font size to `62.5%` so that `1rem` is equal to `10px`. Many of
+the Wonder Blocks components and typography use rems based on this unit.
+
+```css
+html {
+  font-size: 62.5%;
+}
+```
```
beaesguerra, line 54: Kinda related to the other comment: https://github.com/Khan/wonder-blocks/pull/2809/files#r2396094476

I wonder if we could also include setting the root font size in a base WB stylesheet so that if we ever need to change this, we can change it in one place within WB!

__docs__/_get-started_.mdx
jandrade: One thing that I've noticed is that sometimes we get requests from external people to see our WB Figma files. I wonder if we should instead remove these links and move them to another place (maybe Confluence?).

__docs__/_get-started_.mdx
jandrade: This got me thinking. Do you think it would make sense to keep this guide here? or would it be better to move it to Confluence?

I'm not sure if it's hard for product engs to find these pages in Confluence and if it is better fit to keep these docs in Storybook instead.

For example, we have these simple instructions here: https://khanacademy.atlassian.net/wiki/spaces/WB/pages/2217541905/Installing+Web+Wonder+Blocks+in+your+local+environment, but I'm not sure if people have checked that page and if it would make sense to expand it if we see fit for it.

How do you feel about all of this? would love to hear your thoughts.

beaesguerra: My assumption was that WB documentation is easier to find in Storybook since that is where devs go to find information on how to use components or tokens! I feel like the type of docs in our Confluence space (except for the release notes) are more about how to work on WB! 

One thing I like about having technical documentation in Storybook is that the docs are searchable with the rest of the codebase! If we have to make updates to docs based on a code change, sometimes we are able to find related docs when searching in the repo, while searching and updating docs in Confluence is an additional step. For example, in these set up instructions, I added a step for installing peer dependencies which includes `aphrodite`. When we migrate away from `aphrodite`, we can search for the term `"aphrodite"` and it'd be a reminder we can update the getting started docs too!
 
What are your thoughts on this? @marcysutton too in case you have thoughts on where docs like this should live! 

Sidenote: I'm able to view analytics in Confluence and see 9 unique viewers all time on the doc you shared! Though, they are instructions for setting up the WB repo, rather than installing WB in another project! I'd be curious to know if we have access to any analytics for our Storybook docs!

__docs__/_get-started_.mdx
beaesguerra: Good question! We can create a Confluence page about the Figma files and link to that instead!

beaesguerra: I created a Wonder Blocks Resources page in Confluence so we can link to that instead! https://khanacademy.atlassian.net/wiki/spaces/WB/pages/4396089552/Resources

I updated Figma links in the docs to this page instead 😄 Thanks for the suggestion!

__docs__/_get-started_.mdx
jandrade: Good point! I think it would be safe to do it with our current sites. But I'm not fully sure how impactful would be with other places that combine WB with other styles/components. Maybe it would be better to let the consumer decide where to put that override (e.g. which selector).

__docs__/_get-started_.mdx
jandrade: Thanks for the info! I think that's a totally fair observation.

README.md
```diff
@@ -44,7 +44,7 @@ please refer to the [internal documentation](https://khanacademy.atlassian.net/w
 
 Please note – before contributing ensure that any design changes you are wanting
 to make are reflected in the [Wonder Blocks project in
-Figma](https://www.figma.com/file/VbVu3h2BpBhH80niq101MHHE/Wonder-Blocks).
+Figma](https://khanacademy.atlassian.net/wiki/spaces/WB/pages/4396089552/Resources#Figma).
```
jandrade, line 47: **praise:** Thanks for updating this!!

__docs__/_get-started_.mdx
jandrade: This is a valid point! I think it would be a good idea to encapsulate fonts within WB :).

__docs__/_get-started_.mdx
jandrade: **question:** How do you feel about adding the CodeSandbox link here so folks still can see the steps in an IDE?

__docs__/_get-started_.mdx
beaesguerra: > Maybe it would be better to let the consumer decide where to put that override (e.g. which selector).

If I'm understanding correctly, consumers will need the `font-size: 62.5%` on the root html element in order for the correct rem values to work! Or is it fine for the root font-size to be a different value, since the system should scale according to it?

It is interesting though that the base rem value is defined outside of WB so WB isn't the thing that defines that 1rem is 10px (the apps or Storybook preview config determines it!)

(There aren't any issues with how things are now so this can be a consideration for later!)

__docs__/_get-started_.mdx
beaesguerra: I'm weary of adding the CodeSandbox link to the docs since it would be an additional thing we'd need to maintain! I set it up mostly as a draft for testing 😄 

It has crossed my mind though that it would be nice to have a WB environment to easily try out components (whether it be in CodeSandbox or something else!). I'm not sure what the proper way to do that is yet. I used my GitHub account on CodeSandbox to make the draft project (I was able to install WB packages since they are public on npm!)

PR #2811, opened 2025-10-02
FlexibleDialog: Fix positioning at middle breakpoint size

marcysutton:
## Summary:
Removes a media query breakpoint that was causing positioning to be lost at in-between sizes

Before:
Modal dialog in Storybook positioned at top of screen

After:

Modal dialog in storybook positioned in middle of screen

Issue: WB-2105

## Test plan:
1. Review FlexibleDialog at various screen widths
2. Ensure dialog is positioned in the middle of the screen
3. Ensure DrawerLauncher / DrawerDialog aren't affected, since they are built on top of FlexibleDialog

### DIFF
--- a/.changeset/brave-books-attend.md
+++ b/.changeset/brave-books-attend.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-modal": patch
+---
+
+Removes mobile media query in FlexibleDialog causing incorrect positioning
--- a/packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
+++ b/packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
@@ -194,12 +194,6 @@ const componentStyles = StyleSheet.create({
         maxHeight: "100vh",
         maxWidth: 576,
         width: "93.75%",
-
-        [breakpoint.mediaQuery.sm]: {
-            height: "100vh",
-            maxHeight: "100vh",
-            width: "100%",
-        },
     },
 });
 
--- a/packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
+++ b/packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
@@ -3,7 +3,7 @@ import {StyleSheet} from "aphrodite";
 import type {StyleType} from "@khanacademy/wonder-blocks-core";
 import {View} from "@khanacademy/wonder-blocks-core";
 
-import {breakpoint, semanticColor} from "@khanacademy/wonder-blocks-tokens";
+import {semanticColor} from "@khanacademy/wonder-blocks-tokens";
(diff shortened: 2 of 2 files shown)


packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
```diff
@@ -194,12 +194,6 @@ const componentStyles = StyleSheet.create({
         maxHeight: "100vh",
         maxWidth: 576,
         width: "93.75%",
-
-        [breakpoint.mediaQuery.sm]: {
-            height: "100vh",
-            maxHeight: "100vh",
-            width: "100%",
-        },
```
maggieliu123, line 202: Curious: Why was this media query gate breaking adding the shadow outline?

.changeset/brave-books-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-modal": patch
+---
+
+Removes mobile media query in FlexibleDialog causing incorrect positioning
```
maggieliu123, line 5: Have never worked in the WB repo before! What is brave books attend? Is it like a commit message tracker kind of thing?

maggieliu123 (commented):
Thank you Marcy! I left some questions to clarify my own understanding 😄

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
marcysutton: I'm not actually seeing the box-shadow, myself -- so I'm not sure! Removing this fixed the layout, so I went with it 🤷‍♀️

.changeset/brave-books-attend.md
marcysutton: It's an auto-generated filename for the changeset! It will get merged and the file deleted, eventually!

packages/wonder-blocks-modal/src/components/flexible-dialog.tsx
jandrade: fwiw that box-shadow only shows up with the TB theme, so maybe that's why it didn't appear at first.

marcysutton: Oh that's likely it! Related: I feel like Thunderblocks could be the default theme in Storybook at this point.

marcysutton: I can answer this question now: the `box-shadow` is on the outermost `FlexibleDialog` container element, the one that sets the height/min-height. The background color is on a child element, specifically `FlexiblePanel` https://github.com/Khan/wonder-blocks/blob/6aff82047ebf0bb93dc6af8406dc9fa322da72b1/packages/wonder-blocks-modal/src/components/flexible-panel.tsx#L113

PR #2967, opened 2026-02-19
[Experiment] Storybook-to-Markdown Skill Convertor for Wonder Blocks

jeresig:
## Summary:
This is an experiment to see if we can convert all of our storybook files to passable markdown so that it can then be accessed by a skill (that is then baked into this repo so that folks can install it by running: `npx skills @khan/wonder-blocks`).

This change was 99.9% created by Claude Code using Opus 4.6, my only change was adding in the `.gitattributes` bit to mark the output as generated files.

I used these prompts to produce this output:

* Create a Node script that is able to convert our existing Storybook documentation into a series of well-formatted Markdown files. It doesn't have to preserve functionality but it should preserve content (e.g. the code of the examples should still exist).
* Looking at a file like @docs/dropdown/listbox.md I see output like: { children: items, } for when there isn't a render function. It should instead use the default component (Listbox in this case) and display this as <Listbox>{items}</Listbox>. Additionally, there are examples where there is use of {...args} (shold be removed) and Storybook specific methods like action(...) (should be removed). Any other args should be added into the code examples, for example: <Listbox value="pear">
* Update docs-to-markdown.mjs so that it outputs the files to skills/wonder-blocks/docs/ and have it generate a skills/wonder-blocks/SKILL.md file following this specification: https://agentskills.io/specification.md It should provide a short description of how these files should be used and links to each of the component overview files. The component overviews should also be updated to include links to all of their related sub-files. Files should be generous in linking to related markdown files.
* Add a Github workflow that runs @scripts/docs-to-markdown.mjs and commits the results to the repo (making sure that any renamed or deleted files are handled correctly).

Now, is this perfect? Definitely not. I have some ideas on how we could improve the code output (making a linter to require all variables to be inline inside all Storybook examples, and then have Claude tackle that work). But I'm wondering if this might be "good enough" for folks to get experimenting?

I also added in a Github workflow so that the files will get regenerated whenever they land on `main`. I'm open to other suggestions here!

Issue: FEI-XXXX

## Test plan:

A cool side effect of this is that you can now browse the docs using tools like Obsidian!

Screenshot 2026-02-19 at 4 06 37 PM

Screenshot 2026-02-19 at 4 06 56 PM


scripts/docs-to-markdown.mjs
```diff
@@ -0,0 +1,1956 @@
+#!/usr/bin/env node
```
jeresig, line 1: This is the new script - I have not reviewed this at all.

scripts/docs-to-markdown.mjs
```diff
@@ -0,0 +1,1956 @@
...
+    });
+}
+
+// ─── Step 1: File Discovery ──────────────────────────────────────────────────
+
+async function discoverFiles() {
+    const storyFiles = await fg("__docs__/**/*.stories.tsx", {
+        cwd: ROOT,
+        ignore: [
+            "__docs__/**/*-testing-snapshots.stories.tsx",
```
jandrade, line 99: **suggestion:** We should also ignore `playtesting` stories. We use them just to do some internal manual testing (and some interactive testing). For example: https://github.com/Khan/wonder-blocks/blob/main/__docs__/wonder-blocks-tabs/tabs-testing-playtesting.stories.tsx

scripts/docs-to-markdown.mjs
```diff
@@ -0,0 +1,1956 @@
...
+            "__docs__/components/**",
```
jandrade, line 100: **suggestion:** We should also ignore any files that include the `deprecated` word. For example: `tokens/deprecated-color...`

jandrade: I'm also going to be experimenting with some visual badges (which uses Storybook tags) to be able to filter out deprecated components.

https://storybook.js.org/blog/storybook-tags/

skills/wonder-blocks/docs/switch/best-practices.md
jandrade, line 1: For some reason, the tool skipped creating the Switch Docs page.

skills/wonder-blocks/SKILL.md
jandrade, line 1: This is really cool! 👏

jandrade (commented):
This looks promising, and definitively worth experimenting. I left some initial comments to exclude some pages before giving final approval.

I'm going to be creating a PR to improve a bit the documentation around deprecated components that I'd like to land before this PR gets merged so we don't generate unwanted MD files.

Thanks for working on this!

jandrade:
@jeresig do you think this would still be useful? or is it ok to close the PR?

jeresig:
@jandrade I'll close this! It seems like we have other ways forward at the moment - we can always dust this off, if need be.

PR #2977, opened 2026-02-25
Tokens: Change `background.neutral.strong` in Thunderblocks

beaesguerra:
## Summary:

Now that `semanticColor.core.background.base.strong` is updated to a dark color in https://github.com/Khan/wonder-blocks/pull/2970, we can update `semanticColor.core.background.neutral.strong` in the `thunderblocks` theme from `black_100` to `gray_10`.

Note: If current usage in projects are expecting the value to be `black_100`, it should be updated to use `background.base.strong` instead for the TB theme

Issue: WB-2271

## Test plan:

Confirm token change is expected and changes in Chromatic are expected


jandrade:
Here's the associated `frontend` PR with the files that need to be changed to reflect the WB update: https://github.com/Khan/frontend/pull/10520

beaesguerra (commented):
Looks good to me, including the Chromatic snapshots! Thanks for taking this on @jandrade ! 😄

jandrade (approved):

PR #2990, opened 2026-03-25
[TUT-3774] Add autoUpdate prop to Popover component

tony-dinh:
## Summary:
This PR exposes the `autoUpdate` prop on the `Popover` component, forwarding it through to the underlying `TooltipPopper`.

- Adds optional `autoUpdate?: boolean` to `Popover`'s props interface with a JSDoc comment
- Forwards the prop to `TooltipPopper`, which uses a `MutationObserver` to reposition the popover when the anchor element changes (e.g. due to content reflow or DOM mutations)

Issue: https://khanacademy.atlassian.net/browse/TUT-3774

Test plan:

- Verify tests and linter check passes

### DIFF
--- a/.changeset/chatty-rocks-cry.md
+++ b/.changeset/chatty-rocks-cry.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-popover": minor
+---
+
+Add autoUpdate prop to Popover component
--- a/__docs__/wonder-blocks-banner/banner-testing-snapshots.stories.tsx
+++ b/__docs__/wonder-blocks-banner/banner-testing-snapshots.stories.tsx
@@ -2,9 +2,10 @@ import * as React from "react";
 import {action} from "storybook/actions";
 import type {Meta, StoryObj} from "@storybook/react-vite";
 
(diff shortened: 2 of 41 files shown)


.changeset/chatty-rocks-cry.md
tony-dinh, line 1: Image

jandrade (commented):
The implementation looks good! but it would be really helpful to include the following parts in the PR to improve documentation around this:

1. Adding `autoUpdate` in the Popover argTypes object, so it can be included in the docs page ([reference](https://github.com/Khan/wonder-blocks/blob/badad6ee26ede04c822e14f4f41c4935129d033a/__docs__/wonder-blocks-tooltip/tooltip.argtypes.ts#L43)).
2. Including a story to show how this could work. You could also take the Tooltip reference: https://5e1bf4b385e3fb0020b7073c-wngavcwftd.chromatic.com/?path=/docs/packages-tooltip-tooltip--docs#auto-update-2

Thanks for working on this!

jandrade (approved):
Looks great! thanks for adding the docs :shipit:

PR #2991, opened 2026-03-25
[WB-2281] Add useAnimationFrame and useActionScheduler hooks to wonder-blocks-timing

tony-dinh:
## Summary:
This PR adds two new hooks to `@khanacademy/wonder-blocks-timing` that complete the hook-based API for the timing package.

- Adds `useAnimationFrame`: a hook wrapping `requestAnimationFrame`/`cancelAnimationFrame` that follows the same proxy-ref and policy pattern as the existing `useTimeout` and `useInterval` hooks — supports `SchedulePolicy`, `ClearPolicy`, and `ActionPolicy`
- Adds `useActionScheduler`: a hook-based alternative to the `withActionScheduler` HOC that returns a stable `IScheduleActions` API and automatically calls `disable()` on unmount, allowing components to schedule timeouts, intervals, and animation frames without needing to wrap in a HOC
- Exports both hooks from the package's public index

Issue: https://khanacademy.atlassian.net/browse/WB-2281

Test plan:

- [ ] Run `pnpm jest packages/wonder-blocks-timing --no-coverage --no-watchman`


.changeset/real-teachers-draw.md
tony-dinh, line 1: Image

__docs__/wonder-blocks-timing/use-animation-frame.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+With the default `SchedulePolicy.Immediately`, the frame fires automatically on
+mount — useful for deferring a one-time DOM read/write to just before the first
+paint:
+
+<Canvas sourceState="shown" of={UseAnimationFrameStories.Immediately} />
+
+## Animation loop (OnDemand + Resolve on clear)
+
+The most common pattern — a loop that starts on demand and fires one final
```
somewhatabstract: **question:** I don't think this is an animation loop - we don't have that. It's only firing once - we're not firing repeatedly. I think these AI generated docs are really wrong. Take a look at the base docs for the `schedule.animationFrame` to see the actual behaviors.

packages/wonder-blocks-timing/src/hooks/use-animation-frame.ts
```diff
@@ -0,0 +1,100 @@
...
+ * `ActionPolicy.Reset`.
+ * @param options.clearPolicy Determines how the request is cleared when the
+ * component is unmounted or the request is recreated. By default, the
+ * request is cancelled immediately. If you want to let the request run to
+ * completion, use `ClearPolicy.Resolve`. This policy is also used as the
+ * default when calling `clear()` manually with no argument. Pass an explicit
+ * `ClearPolicy` to `clear(policy)` to override it for a specific call.
+ * @param options.schedulePolicy Determines when the request is scheduled.
+ * By default, the request is made immediately. If you want to delay
+ * scheduling the request, use `SchedulePolicy.OnDemand`.
```
somewhatabstract: **suggestion:** I don't think we should put this here; it should be in comments on the type fields instead.

somewhatabstract (changes requested):
The implementation looks pretty solid, but the docs seem quite wrong. They are describing behaviors that are not present (nor should be, at least for this initial version). Can you update them to be correct - it's likely you can mostly borrow from existing docs to make something consistent.

__docs__/wonder-blocks-timing/use-animation-frame.mdx
tony-dinh: Ah this section of the doc is describing a specific `use-animation-frame` story. I was having trouble coming up with an idea to make the story useful (isSet is never really visibly `true` since it's not a reactive state and we're polling for its value) so Claude came up with using an animation loop to show that specific use-case.

The meta documentation for this hook states:
> - Unlike `useTimeout` and `useInterval`, each frame fires **once**. To create a continuous animation loop, call `set()` again inside the action. 

but if this adds confusion, I can remove this story!

somewhatabstract: I think the story can stay but the docs shouldn't talk about the stories - they should be talking about the actual API

packages/wonder-blocks-timing/src/util/types.ts
```diff
@@ -131,14 +131,37 @@ export interface IAnimationFrame {
...
 /**
  * Options for the hook variants of our scheduling APIs.
+ *
+ * Note: in hook-based APIs (`useAnimationFrame`, `useTimeout`, `useInterval`),
+ * `clearPolicy` is additionally used as the default when calling `clear()`
+ * manually with no argument on the returned API. Pass an explicit
+ * `ClearPolicy` to `clear(policy)` to override it for a specific call.
  */
 export type HookOptions = Options & {
+    /**
```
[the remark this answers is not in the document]
tony-dinh: Hm Claude seems to be onto something here.. I think it's right that the existing timing API (use-interval and use-timeout) comments are incorrect 🤔 

Probably a good idea to remove their comments and rely on the JSDoc on the shared type instead.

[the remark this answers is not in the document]
somewhatabstract: The `HookOptions` shouldn't declare how things behave, it should declare the purpose of the type. The implementations should declare the behavior with respect to certain values.

packages/wonder-blocks-timing/src/hooks/use-animation-frame.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Hook providing access to a scheduled animation frame request.
+ *
+ * @param action The action to be invoked when the animation frame fires.
+ * By default, this will not cause the request to restart if it changes.
+ * This makes it easier to use with inline lambda functions rather than
+ * requiring consumers to wrap their action in a `useCallback`. To change
+ * this behavior, see the `actionPolicy` option.
+ * @param options Options for the hook. See `HookOptions` for details.
```
somewhatabstract: **suggestion[nit]:** This should explain how the options are used by this specific hook. I know this may seem contradictory to my earlier comment, but there's a nuance.

The options themselves should have their general purpose described in the type - what they are intended to control and what the values may mean.

The behavior that the options drive should be explained here in the hook docs since they are specific to the hook implementation. I don't think that we need to document them using the `@param options.<field>` syntax, but rather just text.

packages/wonder-blocks-timing/src/util/types.ts
```diff
@@ -131,14 +131,37 @@ export interface IAnimationFrame {
  * Options for the scheduling APIs.
  */
 export type Options = {
+    /**
+     * Determines when the request is scheduled. By default, the request is
+     * made immediately on creation. Use `SchedulePolicy.OnDemand` to delay
+     * scheduling until `set()` is explicitly called.
```
somewhatabstract: **suggestion:** This shouldn't include implementation detail as that's specific to the code using the type, not the type itself.

```suggestion
     * Determines when the request is scheduled.
```

packages/wonder-blocks-timing/src/util/types.ts
```diff
@@ -131,14 +131,37 @@ export interface IAnimationFrame {
...
+     */
     schedulePolicy?: Policies.SchedulePolicy;
+    /**
+     * Determines how the request is cleared when the component is unmounted
+     * or the request is recreated. By default, the request is cancelled
+     * immediately. Use `ClearPolicy.Resolve` to invoke the action one final
+     * time when the request is cleared.
```
somewhatabstract: **suggestion:** This shouldn't include implementation detail as that's specific to the code using the type, not the type itself.


```suggestion
     * Determines how the request is cleared when the component is unmounted
     * or the request is recreated.
```

packages/wonder-blocks-timing/src/util/types.ts
```diff
@@ -131,14 +131,37 @@ export interface IAnimationFrame {
...
     clearPolicy?: Policies.ClearPolicy;
 };
 
 /**
  * Options for the hook variants of our scheduling APIs.
+ *
+ * Note: in hook-based APIs (`useAnimationFrame`, `useTimeout`, `useInterval`),
+ * `clearPolicy` is additionally used as the default when calling `clear()`
+ * manually with no argument on the returned API. Pass an explicit
+ * `ClearPolicy` to `clear(policy)` to override it for a specific call.
```
somewhatabstract: **suggestion:** Don't include this here. Implementation details don't belong on the types - and they're likely to get out-of-date really easily (they're already not in sync in this change as Claude points out).


```suggestion
```

packages/wonder-blocks-timing/src/util/types.ts
```diff
@@ -131,14 +131,37 @@ export interface IAnimationFrame {
...
  */
 export type HookOptions = Options & {
+    /**
+     * Determines how the action is handled when it changes between renders.
+     * By default (`ActionPolicy.Passive`), the action is replaced but the
+     * request is not reset — the updated action will be invoked when the
+     * request next fires. Use `ActionPolicy.Reset` to cancel the current
+     * request and start a new one whenever the action changes.
```
somewhatabstract: **suggestion:** Don't include implementation details in type docs.

```suggestion
```

somewhatabstract (changes requested):
This looks great. The docs just need some tidy-up...mostly just deleting implementation details from the type docs.

somewhatabstract (approved):
I think this is looking good. Claude has some comments; perhaps one or two warrant a change in comments or docs.

PR #2995, opened 2026-04-02
Add Tooltip stories for Links and Rich Text Content

marcysutton:
## Summary:
We got a question in Slack about whether the Tooltip component supports HTML content, which it does. I noticed there wasn't a story showing this working, so I added one. The apparent consumer code also had incorrect structuring with a Link, and I noticed there wasn't a Link example either. https://khanacademy.slack.com/archives/C8Z9DSKC7/p1775096784762589

Issue: WB-XXXX

## Test plan:
1. Review new stories: 
   - `/?path=/story/packages-tooltip-tooltip--with-link-anchor`
   - `/?path=/story/packages-tooltip-tooltip--with-rich-text-content`
   - `/?path=/story/packages-tooltip-tooltipcontent--rich-text-content`


__docs__/wonder-blocks-tooltip/tooltip-content.stories.tsx
```diff
@@ -89,3 +89,21 @@ CustomContent.parameters = {
         },
     },
 };
+
+/**
+ * To render rich text in tooltip content, pass a React element as `children`
+ * instead of a plain string. When a string is passed, it is wrapped in
+ * `LabelMedium` and rendered as plain text — HTML tags in a string will appear
```
beaesguerra: The comment can be updated!

```suggestion
 * `BodyText` and rendered as plain text — HTML tags in a string will appear
```

__docs__/wonder-blocks-tooltip/tooltip.stories.tsx
```diff
@@ -164,6 +169,54 @@ export const ComplexAnchorAndTitle: StoryComponentType = {
     },
 };
 
+/**
+ * Tooltips can be used with links as anchors.
+ * When a `Link` is the anchor element, set `forceAnchorFocusivity={false}`
```
beaesguerra, line 174: TIL about `forceAnchorFocusivity`! 

I noticed that the docs for the prop isn't showing up in the SB docs! Could you update that as well? 

Image

__docs__/wonder-blocks-tooltip/tooltip.stories.tsx
```diff
@@ -164,6 +169,54 @@ export const ComplexAnchorAndTitle: StoryComponentType = {
...
+ * prop instead of a plain string. When a string is passed it is rendered as
+ * plain text — HTML tags in a string will appear literally (e.g.
+ * `<i>text</i>`). Use inline HTML elements inside a typography component to
+ * control formatting.
+ */
+export const WithRichTextContent: StoryComponentType = {
+    render: function Render() {
+        return (
+            <Tooltip
+                content={
```
beaesguerra, line 220: I'm wondering if `content` needs to be wrapped with `TooltipContent` when passing in jsx! I see the type definition for the prop includes TooltipContent:

https://github.com/Khan/wonder-blocks/blob/cb10df80faad3afafc91e3a1f4874a35e9e89435/packages/wonder-blocks-tooltip/src/components/tooltip.tsx#L55-L57

It wasn't immediately clear to me in the docs how `TooltipContent` should be used, so I am glad we're adding more examples :) 

cc: @jandrade in case you have context on this!

__docs__/wonder-blocks-tooltip/tooltip.stories.tsx
```diff
@@ -173,7 +226,7 @@ export const AnchorInScrollableParent: StoryComponentType = {
         return (
             <View style={styles.scrollbox}>
                 <View style={styles.hostbox}>
-                    <Body>
+                    <BodyText>
```
beaesguerra, line 245: Thanks for updating typography components! :)

beaesguerra (commented):
Thanks for updating the docs, great idea :) 

I left some questions! One thing I wanted to verify is if TooltipContent should be in the example!

__docs__/wonder-blocks-tooltip/tooltip.stories.tsx
```diff
@@ -164,6 +169,54 @@ export const ComplexAnchorAndTitle: StoryComponentType = {
...
+        return (
+            <Tooltip
+                content="This link navigates to the Khan Academy homepage."
+                placement="top"
+                forceAnchorFocusivity={false}
+            >
+                <Link href="https://www.khanacademy.org">Khan Academy</Link>
+            </Tooltip>
+        );
+    },
```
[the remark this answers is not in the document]
beaesguerra: Cool! It caught something I noticed from checking the snapshots visually https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=69cea98f30ca01a1a71c9dd9

__docs__/wonder-blocks-tooltip/tooltip.stories.tsx
marcysutton: I did it this way to mimic the production code, so we could validate it works with the setup in `frontend`. It seems to work, so maybe the type definition isn't quite right?

__docs__/wonder-blocks-tooltip/tooltip.stories.tsx
marcysutton: Focusivity isn't exactly a word but it works! (I thought it would be "forceAnchorFocusability" when trying to write from memory)

Updated the argtype.

somewhatabstract: Yeah, I invented that term a long time ago. `Focusability` would probably have been a better choice - I can't tell you where my brain was, naming things is hard.

__docs__/wonder-blocks-tooltip/tooltip.stories.tsx
jandrade: ahh hmmm I think this is another case where the Flow to TS migration couldn't add full type safety support for these cases with `TooltipContent`. I mean, probably we were expecting to wrap components with `TooltipContent`, but TS is not able to fully enforce that.

beaesguerra: It's too bad we can't use TS to limit what components can be used for specific props 😢 (I remember asking about this too in the typescript channel ([related thread](https://khanacademy.slack.com/archives/C028T2HDX3P/p1741210473344549)))

I think keeping the type as `React.ReactElement<React.ComponentProps<typeof TooltipContent>>` would be a helpful hint for what should be passed in (rather than `any`). Maybe this is also another area where we can use custom lint rules to enforce what components can be used as props for specific component props like this

marcysutton: I ended up dropping the commit to change the type definition in favor of a later lint rule. Since this PR was created in response to a question, I'd like to defer further work on the Tooltip itself for a later date!

beaesguerra (approved):
🚀

PR #2997, opened 2026-04-03
Introduce lint rule: no-invalid-bodytext-parent

marcysutton:
## Summary:
Adds new lint rule for the BodyText Typography component to disallow certain parents:

### WB Components

- `Choice`, `Checkbox`, `Radio` — render a `<label>` internally; `<p>` can't be inside a `<label>`
- `Button`, `ActivityButton` — inline/button context; `<p>` can't be inside `<button>`
- `Heading`, `HeadingLarge`, `HeadingMedium`, `HeadingSmall`, `HeadingXSmall` — heading elements can't contain block-level children (question: should we include the older components?)

### HTML Elements

- `<button>` — inline/interactive element
- `<p>` — block element, can't contain another block element
- `<label>` — inline element
- `<h1>`–`<h6>` — heading elements

### Styled/custom variants
- `StyledButton`
- `StyledP`
- `StyledLabel`

### BodyText itself

- `<BodyText>` nested inside another `<BodyText>` — both default to `<p>`, which would be `<p><p>` (invalid HTML). Exception: if the outer BodyText has a block-container `tag` prop (`div`, `section`, `article`, etc.), it's fine.

## Questions:
- Naming. Do we like the lint rule `no-invalid-bodytext-parent`?
- The test cases for the lint rule itself. Do we want to change/add/remove anything?

Issue: WB-2247

## Test plan:
1. Ensure tests pass
2. Run the lint rule in WB and frontend (In progress)

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,6 +108,7 @@
     "@types/react-window": "^1.8.8",
     "@typescript-eslint/eslint-plugin": "8.17.0",
     "@typescript-eslint/parser": "8.17.0",
+    "@typescript-eslint/rule-tester": "^8.46.3",
     "@vitejs/plugin-react-swc": "^3.9.0",
     "@vitest/browser": "^4.0.15",
     "@vitest/browser-playwright": "^4.0.15",
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     "@types/react-window": "^1.8.8",
     "@typescript-eslint/eslint-plugin": "8.17.0",
     "@typescript-eslint/parser": "8.17.0",
-    "@typescript-eslint/rule-tester": "^8.46.3",
+    "@typescript-eslint/rule-tester": "^8.58.0",
     "@vitejs/plugin-react-swc": "^3.9.0",
     "@vitest/browser": "^4.0.15",
     "@vitest/browser-playwright": "^4.0.15",
--- a/package.json
+++ b/package.json
@@ -108,7 +108,6 @@
     "@types/react-window": "^1.8.8",
     "@typescript-eslint/eslint-plugin": "8.17.0",
     "@typescript-eslint/parser": "8.17.0",
-    "@typescript-eslint/rule-tester": "^8.58.0",
     "@vitejs/plugin-react-swc": "^3.9.0",
     "@vitest/browser": "^4.0.15",
     "@vitest/browser-playwright": "^4.0.15",
--- a/packages/eslint-plugin-wonder-blocks/src/rules/__tests__/no-invalid-bodytext-parent.test.ts
+++ b/packages/eslint-plugin-wonder-blocks/src/rules/__tests__/no-invalid-bodytext-parent.test.ts
@@ -1,0 +1,331 @@
+/**
+ * Tests for the no-invalid-bodytext-parent ESLint rule.
+ *
+ * Uses ESLint's RuleTester to verify that the rule correctly flags BodyText
+ * components nested inside elements that cannot contain block-level content,
+ * and that the auto-fix correctly adds tag="span".
+ */
+import {RuleTester} from "eslint";
+import {noInvalidBodyTextParent} from "../no-invalid-bodytext-parent";
+
+// @types/eslint@9 expects flat config but this project uses eslint@8 (eslintrc format).
+// The runtime API works correctly; the cast silences the version mismatch.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const ruleTester = new RuleTester({
+    parserOptions: {
+        ecmaVersion: 2020,
+        ecmaFeatures: {jsx: true},
+        sourceType: "module",
+    },
+} as any);
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+ruleTester.run("no-invalid-bodytext-parent", noInvalidBodyTextParent as any, {
+    // ------------------------------------------------------------------ //
+    // VALID — BodyText with an inline tag is allowed everywhere;          //
+    //         BodyText without a tag is fine in non-restricted contexts.  //
+    // ------------------------------------------------------------------ //
+    valid: [
+        // Standalone BodyText — no parent context at all
+        {code: `<BodyText>Hello</BodyText>`},
+
+        // BodyText inside plain block containers
+        {code: `<div><BodyText>Hello</BodyText></div>`},
+        {code: `<section><BodyText>Hello</BodyText></section>`},
+        {code: `<article><BodyText>Hello</BodyText></article>`},
+
+        // BodyText with an inline tag inside WB form components
+        {
+            code: `<Choice label={<BodyText tag="span">Title</BodyText>} />`,
+        },
+        {
+            code: `<Checkbox label={<BodyText tag="span">Label</BodyText>} />`,
+        },
+        {
+            code: `<Radio label={<BodyText tag="span">Label</BodyText>} />`,
+        },
+
+        // BodyText with an inline tag inside WB Button
+        {code: `<Button><BodyText tag="span">Click</BodyText></Button>`},
+        {
+            code: `<ActivityButton><BodyText tag="span">Click</BodyText></ActivityButton>`,
+        },
+
+        // BodyText with an inline tag inside HTML button / StyledButton
+        {code: `<button><BodyText tag="span">Click</BodyText></button>`},
+        {
+            code: `<StyledButton><BodyText tag="span">Click</BodyText></StyledButton>`,
+        },
+
+        // BodyText with an inline tag inside HTML p / StyledP
+        {code: `<p><BodyText tag="span">Inline text</BodyText></p>`},
+        {
+            code: `<StyledP><BodyText tag="span">Inline text</BodyText></StyledP>`,
+        },
+
+        // BodyText with an inline tag inside HTML label / StyledLabel
+        {code: `<label><BodyText tag="span">Label text</BodyText></label>`},
+        {
+            code: `<StyledLabel><BodyText tag="span">Label text</BodyText></StyledLabel>`,
+        },
+
+        // BodyText with an inline tag inside another BodyText
+        {
+            code: `<BodyText><BodyText tag="span">Inline nested</BodyText></BodyText>`,
+        },
+
+        // Inner BodyText (no inline tag) inside outer BodyText with a
+        // block-container tag — the outer renders as a div, which can
+        // contain <p> elements.
+        {
+            code: `<BodyText tag="div"><BodyText>Nested para</BodyText></BodyText>`,
+        },
+        {
+            code: `<BodyText tag="section"><BodyText>Nested para</BodyText></BodyText>`,
+        },
+        {
+            code: `<BodyText tag="article"><BodyText>Nested para</BodyText></BodyText>`,
+        },
+
+        // BodyText with an inline tag inside WB Heading components
+        {
+            code: `<Heading><BodyText tag="sup">Superscript</BodyText></Heading>`,
+        },
+        {
+            code: `<HeadingLarge><BodyText tag="span">Sub</BodyText></HeadingLarge>`,
+        },
+        {
+            code: `<HeadingMedium><BodyText tag="code">Code</BodyText></HeadingMedium>`,
+        },
+
+        // BodyText with an inline tag inside HTML heading elements
+        {code: `<h1><BodyText tag="sup">Sup</BodyText></h1>`},
+        {code: `<h2><BodyText tag="span">Sub</BodyText></h2>`},
+
+        // BodyText with other inline tag variants
+        {code: `<button><BodyText tag="code">Code</BodyText></button>`},
+        {code: `<label><BodyText tag="strong">Bold</BodyText></label>`},
+        {code: `<p><BodyText tag="em">Emphasis</BodyText></p>`},
+
+        // BodyText deeper inside a non-restricted wrapper that itself is
+        // inside a restricted element — only the direct parent is checked.
+        {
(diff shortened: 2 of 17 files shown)


marcysutton:
@claude review once

package.json
```diff
@@ -108,6 +107,7 @@
     "@types/react-window": "^1.8.8",
     "@typescript-eslint/eslint-plugin": "8.17.0",
     "@typescript-eslint/parser": "8.17.0",
+    "@typescript-eslint/rule-tester": "^8.46.3",
```
beaesguerra: This is set in the `package.json` for the `eslint-plugin-wonder-blocks` package, so I _think_ it isn't needed at the root level!

packages/eslint-plugin-wonder-blocks/src/rules/index.ts
```diff
@@ -1,7 +1,9 @@
 import noCustomTabRole from "./no-custom-tab-role";
+import noInvalidBodyTextParent from "./no-invalid-bodytext-parent";
 
 const rules = {
     "no-custom-tab-role": noCustomTabRole,
+    "no-invalid-bodytext-parent": noInvalidBodyTextParent,
```
beaesguerra, line 8: Do you think this rule should be added to the `recommended` or `strict` config?

packages/eslint-plugin-wonder-blocks/src/rules/__tests__/no-invalid-bodytext-parent.test.ts
```diff
@@ -0,0 +1,343 @@
...
+        },
+    ],
+
+    // ------------------------------------------------------------------ //
+    // INVALID — BodyText without an inline tag inside restricted parents  //
+    // ------------------------------------------------------------------ //
+    invalid: [
+        // --- WB Form: Choice ---
+        {
+            code: `<Choice label={<BodyText>Title</BodyText>} />`,
```
beaesguerra, line 134: Do we want to handle <BodyText> used with LabeledField props too (`label`, `descripton`, `errorMessage`, `contextLabel`, `readOnlyMessage`, `additionalHelperMessage`)?

beaesguerra: Same with props for RadioGroup and CheckboxGroup (`description` and `errorMessage` props)!

packages/eslint-plugin-wonder-blocks/src/rules/__tests__/no-invalid-bodytext-parent.test.ts
```diff
@@ -0,0 +1,343 @@
...
+        },
+        {
+            code: `<HeadingXSmall><BodyText>Sub text</BodyText></HeadingXSmall>`,
+            errors: [{messageId: "nestedInHeading"}],
+            output: `<HeadingXSmall><BodyText tag="span">Sub text</BodyText></HeadingXSmall>`,
+        },
+
+        // --- HTML heading elements ---
+        {
+            code: `<h1><BodyText>Sub text</BodyText></h1>`,
```
beaesguerra, line 309: Do we want test cases for StyledH1 elements?

packages/eslint-plugin-wonder-blocks/src/rules/__tests__/no-invalid-bodytext-parent.test.ts
```diff
@@ -0,0 +1,343 @@
+/**
```
beaesguerra, line 1: Lots of test coverage, nice!!

packages/eslint-plugin-wonder-blocks/src/rules/no-invalid-bodytext-parent.ts
```diff
@@ -0,0 +1,354 @@
...
+    "figcaption",
+    "details",
+    "summary",
+]);
+
+/**
+ * Wonder Blocks form components that internally render a <label> or similar
+ * inline-context element around their content.
+ */
+const WB_FORM_COMPONENTS = new Set(["Choice", "Checkbox", "Radio"]);
```
beaesguerra: Could LabeledField, RadioGroup and CheckboxGroup be added here?

packages/eslint-plugin-wonder-blocks/src/rules/no-invalid-bodytext-parent.ts
```diff
@@ -0,0 +1,354 @@
...
+    "HeadingLarge",
+    "HeadingMedium",
+    "HeadingSmall",
+    "HeadingXSmall",
+]);
+
+/**
+ * HTML heading element names.
+ */
+const HTML_HEADING_ELEMENTS = new Set(["h1", "h2", "h3", "h4", "h5", "h6"]);
```
beaesguerra: I wonder if we could add StyledH1, StyledH2, etc. here!

packages/eslint-plugin-wonder-blocks/src/rules/__tests__/no-invalid-bodytext-parent.test.ts
```diff
@@ -0,0 +1,343 @@
...
+        // BodyText with other inline tag variants
+        {code: `<button><BodyText tag="code">Code</BodyText></button>`},
+        {code: `<label><BodyText tag="strong">Bold</BodyText></label>`},
+        {code: `<p><BodyText tag="em">Emphasis</BodyText></p>`},
+
+        // BodyText deeper inside a non-restricted wrapper that itself is
+        // inside a restricted element — only the direct parent is checked.
+        {
+            code: `<button><div><BodyText>Nested deeper</BodyText></div></button>`,
+        },
```
beaesguerra, line 120: Good to know!

packages/eslint-plugin-wonder-blocks/src/rules/no-invalid-bodytext-parent.ts
```diff
@@ -0,0 +1,354 @@
...
+ */
+function getAttributeStringValue(
+    openingElement: TSESTree.JSXOpeningElement,
+    attributeName: string,
+): string | null {
+    const attr = openingElement.attributes.find(
+        (a): a is TSESTree.JSXAttribute =>
+            a.type === "JSXAttribute" &&
+            a.name.type === "JSXIdentifier" &&
+            (a.name as TSESTree.JSXIdentifier).name === attributeName,
```
beaesguerra: I was curious if there was something we could do to avoid type casting (`as TSESTree.JSXIdentifier`). It seems like we don't need the type casting! 


```suggestion
            a.name.name === attributeName,
```

Same with the other places where we use `as`. Let me know if I am missing something!

packages/eslint-plugin-wonder-blocks/src/rules/no-invalid-bodytext-parent.ts
```diff
@@ -0,0 +1,354 @@
...
+export default createRule<Options, MessageIds>({
+    name: "no-invalid-bodytext-parent",
+    meta: {
+        type: "problem",
+        docs: {
+            description:
+                "Disallow BodyText with a block-level tag (default <p>) inside elements that cannot contain block-level content",
+            recommended: true,
+        },
+        fixable: "code",
```
beaesguerra, line 103: Cool! Our first fixable custom wb lint rule!

packages/eslint-plugin-wonder-blocks/src/rules/no-invalid-bodytext-parent.ts
```diff
@@ -0,0 +1,354 @@
...
+    | "nestedInParagraph"
+    | "nestedInLabel"
+    | "nestedInBodyText"
+    | "nestedInHeading";
+
+/**
+ * Inline tags that make BodyText safe to nest inside inline/interactive
+ * contexts (e.g., inside a button, label, or paragraph element).
+ */
+const INLINE_BODY_TEXT_TAGS = new Set([
```
beaesguerra: I like how you've made it easy to configure different elements/components!

packages/eslint-plugin-wonder-blocks/src/rules/__tests__/no-invalid-bodytext-parent.test.ts
```diff
@@ -0,0 +1,343 @@
...
+            errors: [
+                {
+                    messageId: "nestedInFormComponent",
+                    data: {componentName: "Choice"},
+                },
+            ],
+            output: `<Choice label={<BodyText tag="span">Title</BodyText>} />`,
+        },
+        {
+            code: `<Choice description={<BodyText>Description</BodyText>} />`,
```
beaesguerra, line 144: I see we don't need to define what props `BodyText` might be used with (`label` or `description` props)! Does that mean it'll check all props for where `BodyText` without an inline tag shouldn't be used ?

packages/eslint-plugin-wonder-blocks/src/rules/no-invalid-bodytext-parent.ts
```diff
@@ -0,0 +1,354 @@
+import {ESLintUtils, TSESLint, TSESTree} from "@typescript-eslint/utils";
+
+import type {WonderBlocksPluginDocs} from "../types";
+
+const createRule = ESLintUtils.RuleCreator<WonderBlocksPluginDocs>(
+    (name) =>
+        `https://github.com/Khan/wonder-blocks/blob/main/packages/eslint-plugin-wonder-blocks/docs/${name}.md`,
```
beaesguerra, line 16: Could you add some docs for the rule in `eslint-plugin-wonder-blocks/docs/`?  Lint errors will include this link with the error message! 

Here is an [example](https://github.com/Khan/wonder-blocks/blob/main/packages/eslint-plugin-wonder-blocks/docs/no-custom-tab-role.md) for a rule doc! 

Then we can reference that doc so it shows up in the WB docs. Here is an example of how to do that: https://github.com/Khan/wonder-blocks/blob/main/__docs__/tools/eslint-plugin-wonder-blocks/no-custom-tab-role.mdx

packages/eslint-plugin-wonder-blocks/src/rules/__tests__/no-invalid-bodytext-parent.test.ts
```diff
@@ -0,0 +1,343 @@
...
+            errors: [{messageId: "nestedInLabel"}],
+            output: null,
+        },
+        {
+            code: `<Button><BodyText tag={getTag()}>Click</BodyText></Button>`,
+            errors: [{messageId: "nestedInButton"}],
+            output: null,
+        },
+    ],
+});
```
beaesguerra, line 368: Could you add some of these as examples in the `demo` project? It will help us verify that the lint rules are being integrated and triggered properly!

packages/eslint-plugin-wonder-blocks/src/rules/no-invalid-bodytext-parent.ts
beaesguerra: Please also add the rule to the table in the package README!  https://github.com/Khan/wonder-blocks/blob/1b5b57f32a3162b3c8b1b98204043c6789ac486b/packages/eslint-plugin-wonder-blocks/README.md?plain=1#L67-L73 

(the package README is already set up to be used in the WB SB docs 😄)

beaesguerra (commented):
Great, thanks for working on our next custom lint rule! I left some questions and notes on how we can document the lint rule 😄 Let me know if you have any questions! 

On a side note, it would be neat if we had a tool that could set up the templating for a new lint rule (the files for docs, tests, demo example, etc!). Maybe I'll explore how we could do that next! It would be nice for setting up new packages and components too!

packages/eslint-plugin-wonder-blocks/src/rules/index.ts
marcysutton: I'm leaning toward `strict` since the rollout in classic apps could be rough. But I could run a test next week to decide?

packages/eslint-plugin-wonder-blocks/src/rules/no-invalid-bodytext-parent.ts
marcysutton: yes, good call! I added those!

packages/eslint-plugin-wonder-blocks/src/rules/__tests__/no-invalid-bodytext-parent.test.ts
marcysutton: Yes, I added them!

packages/eslint-plugin-wonder-blocks/src/rules/__tests__/no-invalid-bodytext-parent.test.ts
marcysutton: Yes, I believe so -- thanks to the AST, it walks up through any node references it finds. Any prop that takes a `BodyText` prop will be caught automatically!

packages/eslint-plugin-wonder-blocks/src/rules/__tests__/no-invalid-bodytext-parent.test.ts
beaesguerra: It would be helpful to have test coverage for LabeledField, RadioGroup and CheckboxGroup components too!

beaesguerra (approved):
The new rule and docs look good to me! Another new lint rule yay 😄

---

I landed a fix for using the workspace dependency in https://github.com/Khan/wonder-blocks/pull/3014 so merging in the latest changes from `main` should fix the lint check in this PR!

marcysutton:
Thanks @beaesguerra! Here is an integration PR to test out the changes. It seems to work great! https://github.com/Khan/frontend/pull/10509

PR #2998, opened 2026-04-07
Add Wonder Blocks skill to wb repo

beaesguerra:
## Summary:

Adding the WB skill from frontend to the wonder-blocks repo! 

I modified some of the guidance so it is better suited to the kind of work we do in Wonder Blocks.
I thought we would need to remove more things that wouldn't apply to the wb repo, but I think most of it could still be useful, especially for examples we create in stories. I put the modifications in a separate commit so
that it is easier to review what's unique to WB!

Issue: FEI-7619

## Test plan:

1. WB Skill should be available. Ask claude what skills it has access to
2. WB skill should be triggered with a prompt that includes UI changes, including in stories

image


.agents/skills/wonder-blocks/SKILL.md
```diff
@@ -11,8 +11,7 @@ description: >
   If the user is building any kind of form, layout, modal, button, dropdown, or
   typography treatment in a WB-enabled codebase, this skill applies — even if
   they don't explicitly say "Wonder Blocks". Do NOT trigger for debugging
-  TypeScript errors, writing tests, setting up Storybook stories, or fixing
```
beaesguerra: I updated the description so that the skill will get triggered when we create stories in WB. This is because our stories are examples of how consumers would use WB

jandrade:
@claude review once

.agents/skills/wonder-blocks/SKILL.md
```diff
@@ -0,0 +1,297 @@
+---
+name: wonder-blocks
+description: >
+  Implements user interfaces using the Wonder Blocks (WB) design system — Khan Academy's
+  React component library. Use this skill whenever the user asks you to build, modify,
+  or review UI components in a project that uses Wonder Blocks, even when similar
+  usage patterns already exist in the codebase; mentions any WB package
```
jandrade: **thought:** I wonder if we need to change the skill description here as we are building WB itself. Or my question is.... how we can trigger this skill in a way that help us build/update WB components that follow our own patterns, conventions and best practices?

.agents/skills/wonder-blocks/SKILL.md
```diff
@@ -0,0 +1,297 @@
...
+
+If the WB Storybook MCP is not available, refer to the type definitions for WB components
+to learn more about the API.
+
+> **IMPORTANT:** This skill is required even when similar patterns already exist in the
+> codebase. Do not skip it because you found a nearby file to copy from.
+
+---
+
+## Package quick-reference
```
jandrade, line 37: **question:** What happens when a new package/component is added? do we need to keep this up to date? Alternatively, would the Storybook MCP be better suited for this kind of information?

.agents/skills/wonder-blocks/SKILL.md
```diff
@@ -0,0 +1,297 @@
...
+
+```tsx
+import {breakpoint} from "@khanacademy/wonder-blocks-tokens";
+const styles = StyleSheet.create({
+    [breakpoint.mediaQuery.sm]: {
+        flexDirection: "column",
+    },
+});
+```
+
```
jandrade, line 172: **suggestion:** I wonder if we should include WB theming information here, so any new components could benefit from understanding our custom implementation.

.agents/skills/wonder-blocks/references/forms.md
```diff
@@ -0,0 +1,368 @@
...
+                {[
+                    {label: titleLabel, error: titleError},
+                    {label: descriptionLabel, error: descriptionError},
+                ]
+                    .filter(({error}) => error)
+                    .map((field) => (
+                        <StyledLi
+                            key={field.label}
+                            style={styles.bannerLi}
+                        >
```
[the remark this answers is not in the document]
beaesguerra: Updated the StyledUl/StyledLi reference. For the literal "X" in error count string feedback, I've left it as is since it is meant to be an example and will depend on the implementation!

jandrade (commented):
Thanks for working on this! I was going to fully review this PR, but have an important question to ask before continuing with the review. I understand that it would make sense to copy most of the existing WB skill, but I'm wondering where would make sense to add more specific stuff to internal WB development. I'd love to hear your thoughts about it!

.agents/skills/wonder-blocks/SKILL.md
beaesguerra: Good question - I updated the description a bit! 

For context, I'm adding the existing wb skill to the wb repo as a starting point (with slight modifications)! So that hopefully, AI agents can start following best practices within WB already for things like token usage, typography, etc, especially when we're updating docs or examples! 

I created WB-2302 for adding design system specific guidance around building/updating wb components, since there are different considerations when creating things for the design system (there were some things in [Guidelines for WB Components](https://khanacademy.atlassian.net/wiki/spaces/WB/pages/3186294919/Guidelines+for+Web+Wonder+Blocks+Components) that I didn't include in the existing WB skill since not all of it applies to consumers).

I wonder if we should put that guidance in a separate skill, since that context isn't always needed when working within the wb repo! It would be neat if the skill could be used to guide considerations for an implementation spec and the implementation of the component itself! 

Let me know what you think, happy to discuss this more :)

.agents/skills/wonder-blocks/SKILL.md
beaesguerra: We would need to keep it up to date if we want the quick reference to include the new package/component!

I think the Storybook MCP could be used as well, though my guess is it is easier for the AI agent to use this mapping, especially when the skill is loaded already! In theory, the SB mcp is useful for getting component specific docs and examples! 

I found that during my (limited) [testing with the wb skill + Storybook mcp](https://github.com/Khan/frontend/pull/9686#issuecomment-4172977263), sometimes the Storybook mcp wouldn't need to get used because there was enough context in the skill and existing code!

beaesguerra: On another note, it would be interesting to enable the wb sb mcp within the wonder-blocks repo! I'm curious if AI agents would prefer to use the mcp over checking the WB source code directly! 

What do you think? If there aren't any concerns, I can create a separate PR to enable the sb mcp in wb 😄

.agents/skills/wonder-blocks/SKILL.md
beaesguerra: I added this suggestion to WB-2302 since it is more relevant to building components! 

To clarify on the WB theming information that should be included, would this refer to the theming at the component level? (ex: the things in a [`theme/` directory for a package](https://github.com/Khan/wonder-blocks/tree/956f8bd9ae31171071607ed89125996074dd01af/packages/wonder-blocks-form/src/theme))

.agents/skills/wonder-blocks/SKILL.md
jandrade: oh yeah, I like the idea of enabling SB MCP here!

.agents/skills/wonder-blocks/SKILL.md
jandrade: Exactly

jandrade (approved):
Looks great to me! thanks for addressing my comments and to create tickets to follow up on that 🚀

.agents/skills/wonder-blocks/SKILL.md
beaesguerra: Here's the PR to enable the SB MCP in wb: https://github.com/Khan/wonder-blocks/pull/3007 !

PR #3001, opened 2026-04-08
remove mise.toml and update 'engines' in package.json

kevinb-khan:
## Summary:
The global mise config is now in place for all devs so the per-repo mise.toml is no longer necessary (until the next Node upgrade).

Issue: FEI-7566

## Test plan:
- node -v, see 24.13.1
- pnpm -v, see 10.1.0


jandrade (approved):
LGTM! 🚢

PR #3006, opened 2026-04-10
Fix gh-pages.yml by stripping out symlinks from Storybook build

kevinb-khan:
## Summary:
One of the things the old actions/upload-pages-artifact was doing was stripping symlinks from the files before creating the tarball artifact and the rest of the GH Pages upload process requires there to be no symlinks.  This PR fixes the issue by stripping out the symlinks before packaging.

Issue: None

## Test plan:
- land, watch the logs for the gh-pages.yml workflow


beaesguerra (approved):
Looks good to me!

PR #3007, opened 2026-04-10
Configure wb storybook mcp

beaesguerra:
## Summary:

Configuring the WB Storybook MCP similar to https://github.com/Khan/frontend/pull/9793 .

Note: This configures the WB Storybook MCP to `http://localhost:6061/mcp` so that it uses the local version instead of the hosted version. This is so that the mcp can be up to date with the developer's environment!

Issue: WB-2306

## Test plan:

Confirm that the SB mcp can be connected to
- Claude: Run `/mcp` to check what mcps are available (note: Claude may prompt you to approve the mcp server first)

Screenshot 2026-04-10 at 11 38 34 AM

- VS Code: Search for "MCP: List Servers" (cmd + Shift + P) and enable `wonder-blocks-storybook`

Screenshot 2026-04-10 at 11 36 28 AM

- Cursor: Search for "Cursor Settings: Tools & MCP" (cmd + Shift + P) and enable `wonder-blocks-storybook`

Screenshot 2026-04-10 at 11 35 27 AM


.cursor/mcp.json
```diff
@@ -0,0 +1,8 @@
+{
+    "mcpServers": {
+        "wonder-blocks-storybook": {
+            "type": "streamableHttp",
+            "url": "http://localhost:6061/mcp"
```
beaesguerra, line 5: I configured the WB SB MCP for the wb repo to use the local version! This assumes that the developer has SB running locally. 

I was thinking if the remote version of the mcp was configured, AI agents wouldn't be able to get new local changes in the dev environment. This assumes local changes are more relevant than what's in `main`. 

Another option I considered was configuring both the local and hosted instances (`wonder-blocks-storybook-local` and `wonder-blocks-storybook`).

Let me know what you think!

.gitignore
```diff
@@ -16,3 +16,6 @@ pnpm-debug.log
 # If you run `pnpm storybook doctor`, it creates this file, but definitely
 # doesn't need to be checked in.
 doctor-storybook.log
+
+# Ignore local Claude settings
+.claude/settings.local.json
```
beaesguerra, line 21: After setting up the WB SB MCP, Claude prompted me to approve the wb mcp. It updated `.claude/settings.local.json` with:

```
{
  "enabledMcpjsonServers": [
    "wonder-blocks-storybook"
  ]
}
```

This is probably something we want to leave up to the user so I've added it to the gitignore!

.mcp.json
```diff
@@ -0,0 +1,8 @@
+{
+    "mcpServers": {
+        "wonder-blocks-storybook": {
+            "type": "http",
```
beaesguerra, line 4: In the equivalent [frontend PR](https://github.com/Khan/frontend/pull/9793/changes#diff-d5d3368a61f8d02e085f7a4cb666ee471cf5383690a04b64c839c6ae2080ceb9R4), this was set to `streamableHttp`. For some reason, Claude wouldn't connect to the WB SB MCP in this repo unless `type` was set to `http`! I was checking if it's connected by using `claude mcp list`. 

cc: @marcysutton in case you ran into this!

.cursor/mcp.json
jandrade: That makes total sense to me

.mcp.json
jandrade: This is how it is internally configured in the SB repo as well: https://github.com/storybookjs/mcp/blob/main/apps/internal-storybook/.mcp.json

I would love to hear thoughts from Marcy too!

jandrade (approved):
Changes look great to me! 🚢

.mcp.json
marcysutton: Hmm, I don't have the same result. In `frontend` right now I can see the `wonder-blocks` MCP listed with `streamableHttp`. That's with both the desktop client and the command line. 

```
claude mcp list
claude.ai Google Cloud BigQuery: https://bigquery.googleapis.com/mcp - ! Needs
authentication
claude.ai Atlassian: https://mcp.atlassian.com/v1/mcp - ! Needs authentication
wonder-blocks: https://main--5e1bf4b385e3fb0020b7073c.chromatic.com/mcp (HTTP) - ✓
Connected
```

The MCP standard defines it as `streamableHttp`:https://modelcontextprotocol.io/specification/2025-11-25/basic/transports 

Claude Code apparently normalizes it to just "http" as a UX simplification. Internally it even does an explicit string replacement:

```
--transport streamable-http  →  --transport http
```

I'm curious why it didn't work for you!

beaesguerra: When I run `claude mcp list` in `frontend`, I get the same output as you! 

I noticed though that the connected wb mcp in `frontend` is named `wonder-blocks`, instead of `wonder-blocks-storybook` (the name defined in `.mcp.json`). When I run `claude mcp list` in this branch, it calls it `wonder-blocks-storybook`!

I'm wondering if maybe the project `.mcp.json` in `frontend` isn't being applied. I had set up the hosted SB MCP before the mcp changes were landed in `main` so maybe that's why I have a connected WB MCP with a different name! 🤔

beaesguerra: Ohh I asked Claude what MCPs I have access to in frontend and noticed a warning:

image

When I run `/doctor`, I get:
```
 MCP Config Diagnostics                                                    

  For help configuring MCP servers, see: https://code.claude.com/docs/en/mcp
                                                      
  [Failed to parse] Project config (shared via .mcp.json)                                          
  Location: /Users/beaesguerra/khan/frontend/.mcp.json
   └ [Error] mcpServers.wonder-blocks-storybook: Does not adhere to MCP server configuration schema
```

If you run `/doctor`, do you get the same output too?

marcysutton: Huh I didn't know about `/doctor`! I do get the same error even after removing the manual entry from `settings.local.json` (from before this landed for everyone). It suggests that changing the type to `http` will fix it. So yes, go for it!

beaesguerra: Here's a PR to update the config in frontend too! https://github.com/Khan/frontend/pull/10431

PR #3012, opened 2026-04-15
Set up WB lint rules in WB

beaesguerra:
## Summary:

Enabling the `strict` config from `@khanacademy/eslint-plugin-wonder-blocks` in the Wonder Blocks repo! 

The `strict` config currently includes the [`no-custom-tab-role`](https://khan.github.io/wonder-blocks/?path=/docs/tools-eslint-plugin-wonder-blocks-rules-no-custom-tab-role--docs) for now, so we address the lint errors that were detected after enabling the config. 

Issue: WB-2308

## Test plan:

1. Confirm lint checks pass

2. Confirm the wb lint rules can be detected when errors aren't addressed

When running `pnpm lint`:
Screenshot 2026-04-15 at 5 42 07 PM

In an IDE:
Screenshot 2026-04-15 at 5 41 36 PM


__docs__/wonder-blocks-clickable/accessibility.stories.tsx
```diff
@@ -87,36 +87,25 @@ export const DisabledState = {
...
-                </Clickable>
-            </div>
-            <View
-                id="panel-1"
-                role="tabpanel"
-                tabIndex={0}
-                aria-labelledby="tab-1"
-                style={styles.panel}
+            <Clickable
+                role="button"
```
beaesguerra, line 91: Updated the example to use a different `role` and aria attribute. Based on the docs, I think the intent of this example was to show that Clickable can be interacted with and have aria attributes set! (`?path=/docs/packages-clickable-clickable-accessibility--docs`)

package.json
```diff
@@ -44,7 +44,7 @@
     "@jest/globals": "^29.7.0",
     "@khanacademy/eslint-config": "^5.2.0",
     "@khanacademy/eslint-plugin": "^3.1.1",
-    "@khanacademy/eslint-plugin-wonder-blocks": "workspace:*",
+    "@khanacademy/eslint-plugin-wonder-blocks": "^0.2.0",
```
beaesguerra, line 47: Note: We use the released version of the package for linting the project, rather than the workspace version! This fixes this error in CI:
```sh
> wonder-blocks@0.0.1 lint:ci /home/github/work/wonder-blocks/wonder-blocks
> eslint --ext .ts --ext .js --ext .tsx --ext .jsx "."


Oops! Something went wrong! :(

ESLint: 8.57.1

ESLint couldn't find the plugin "@khanacademy/eslint-plugin-wonder-blocks".

(The package "@khanacademy/eslint-plugin-wonder-blocks" was not found when loaded as a Node module from the directory "/home/github/work/wonder-blocks/wonder-blocks".)

It's likely that the plugin isn't installed correctly. Try reinstalling by running the following:

    npm install @khanacademy/eslint-plugin-wonder-blocks@latest --save-dev

The plugin "@khanacademy/eslint-plugin-wonder-blocks" was referenced from the config file in ".eslintrc.js".

If you still can't figure out the problem, please stop by https://eslint.org/chat/help to chat with the team.

 ELIFECYCLE  Command failed with exit code 2.
```
https://github.com/Khan/wonder-blocks/actions/runs/24484355505/job/71555787574?pr=3012

This would mean we need to release changes to the eslint plugin first before updating the WB repo to use any new rules!

marcysutton: Interesting! Why does this work differently than other packages where we import from the workspace? Something to do with the tooling lifecycle? It does seem a little cumbersome to maintain.

marcysutton (approved):
Thanks for setting this up @beaesguerra! I left one question, but it's non-blocking.

package.json
beaesguerra: It's strange because using `workspace:*` works locally! I will dig into this more and follow up with my findings! I'll land the PR for now so I can get a blog post prepped for the frontend guild meeting later :) 

One benefit of having to release the plugin package first before using it in WB packages is that we don't need to address any existing errors as part of lint rule creation work so it'll be easier to break up the work into smaller PRs!

beaesguerra: I figured it out! We needed to build the packages before running the lint check so it would use the built version of the plugin :) Updated here: https://github.com/Khan/wonder-blocks/pull/3014

PR #3013, opened 2026-04-16
Add GitHub workflow step for running typetests

benchristel:
We now use TSTyche for type testing, which performs assertions beyond plain
typechecking.

Issue: LEMS-4034

## Test plan:

CI checks should pass. Type tests should run.


beaesguerra (approved):
Looks good to me, thank you! 😄

PR #3014, opened 2026-04-16
Use workspace:* for eslint-plugin-wonder-blocks dep in the project

beaesguerra:
## Summary:

Previously, the CI lint step would fail to find the plugin when we use the `workspace:*` version of the `eslint-plugin-wonder-blocks` package ([context](https://github.com/Khan/wonder-blocks/pull/3012#discussion_r3090058602)).

To fix this, we make sure we build the packages before the linting step, so that the plugin can be used.

This would mean that the wb repo will use the current version of the eslint plugin, rather than the last released version!

Issue: WB-2308

## Test plan:

Confirm checks pass.


.eslintignore
```diff
@@ -25,3 +25,6 @@ CODEOWNERS
 
 types/
 packages/eslint-plugin-wonder-blocks/demo
+
+# ignore auto-generated util type files
+utils/**/*.d.ts
```
beaesguerra, line 30: This ignores lint errors on auto-generated type files (this is in our gitignore as well!)

It addresses this error when we lint after building: https://github.com/Khan/wonder-blocks/actions/runs/24539136579/job/71741015598?pr=3014

jandrade (approved):
LGTM! :shipit:

PR #3015, opened 2026-04-16
BodyText linting part 2: no-invalid-bodytext-children

marcysutton:
## Summary:
This new lint rule will flag invalid children within BodyText:
- `viewChild` — always warns when `View` is a direct child of `BodyText`, regardless of `tag` prop
- `divChild` — warns when `<div>` is a direct child and `BodyText` renders as `<p>` (no `tag` or `tag="p"`)
- `paragraphChild` — warns when `<p>` or a default `<BodyText>` (renders as `<p>`) is a direct child
- `blockChild` — warns on other block-level HTML elements (`section`, `ul`, `h1`–`h6`, etc.) and WB heading components when `BodyText` renders as `<p>`
- Block-child checks are skipped when `BodyText` has an inline or block-container tag prop set

It also splits off a second lint rule for excessive child elements - `no-excessive-bodytext-children`, which can affect screen reader output:
- `tooManyChildren` — warns when direct JSX element children exceed a configurable threshold (default: 5)

Issue: WB-2248

## Test plan:
1. Ensure tests pass
2. Test in frontend: https://github.com/Khan/frontend/pull/10738

### DIFF
--- a/packages/eslint-plugin-wonder-blocks/demo/src/no-invalid-bodytext-children-example.tsx
+++ b/packages/eslint-plugin-wonder-blocks/demo/src/no-invalid-bodytext-children-example.tsx
@@ -1,0 +1,102 @@
+/**
+ * This file demonstrates the wonder-blocks ESLint rule:
+ * `@khanacademy/wonder-blocks/no-invalid-bodytext-children`
+ * Run `pnpm lint` in this directory to see the errors.
+ */
+
+import * as React from "react";
+
+import {View} from "@khanacademy/wonder-blocks-core";
+import {
+    Heading,
+    HeadingLarge,
+    BodyText,
+} from "@khanacademy/wonder-blocks-typography";
+
+// ✅ Valid: inline/phrasing-content children
+export function ValidExamples() {
+    return (
+        <>
+            {/* Plain text */}
+            <BodyText>Hello world</BodyText>
+
+            {/* Inline HTML elements */}
+            <BodyText>
+                Text with <strong>bold</strong> and <em>emphasis</em>.
+            </BodyText>
+
+            {/* block-container tag allows block children */}
+            <BodyText tag="div">
+                <div>block content is fine here</div>
+            </BodyText>
+
+            {/* Inner BodyText with inline tag */}
+            <BodyText>
+                Outer <BodyText tag="span">inline nested</BodyText>
+            </BodyText>
+        </>
+    );
+}
+
+// ❌ Invalid: View is never valid inside BodyText
+export function InvalidViewChild() {
+    return (
+        <BodyText>
+            <View>layout</View>
+        </BodyText>
+    );
+}
+
+// ❌ Invalid: block-level HTML elements inside a default BodyText (<p>)
+export function InvalidBlockChildren() {
+    return (
+        <>
+            <BodyText>
+                <div>block content</div>
+            </BodyText>
+            <BodyText>
+                <p>nested paragraph</p>
+            </BodyText>
+            <BodyText>
+                <section>section content</section>
+            </BodyText>
+        </>
+    );
+}
+
+// ❌ Invalid: BodyText nested inside BodyText (both default to <p>)
+export function InvalidNestedBodyText() {
+    return (
+        <BodyText>
+            <BodyText>nested</BodyText>
+        </BodyText>
+    );
+}
+
+// ❌ Invalid: WB Heading components render as block-level headings
+export function InvalidHeadingChildren() {
+    return (
+        <>
+            <BodyText>
+                <Heading>Title</Heading>
+            </BodyText>
+            <BodyText>
+                <HeadingLarge>Title</HeadingLarge>
+            </BodyText>
+        </>
+    );
+}
+
+// ❌ Invalid: too many direct child elements (default max: 5)
+export function InvalidTooManyChildren() {
+    return (
+        <BodyText>
+            <span>one</span>
+            <span>two</span>
+            <span>three</span>
+            <span>four</span>
+            <span>five</span>
+            <span>six</span>
+        </BodyText>
+    );
+}
--- a/packages/eslint-plugin-wonder-blocks/demo/src/no-invalid-bodytext-children-example.tsx
+++ b/packages/eslint-plugin-wonder-blocks/demo/src/no-invalid-bodytext-children-example.tsx
@@ -38,7 +38,16 @@ export function ValidExamples() {
     );
 }
 
-// ❌ Invalid: View is never valid inside BodyText
+// ✅ Valid: View with an inline tag is phrasing content
+export function ValidViewInlineTag() {
+    return (
+        <BodyText>
+            Read more <View tag="span">details here</View>.
+        </BodyText>
+    );
+}
+
+// ❌ Invalid: View defaults to <div> — block-level inside a <p>
 export function InvalidViewChild() {
     return (
         <BodyText>
@@ -79,16 +88,3 @@ export function InvalidHeadingChildren() {
     );
 }
 
-// ❌ Invalid: too many direct child elements (default max: 5)
-export function InvalidTooManyChildren() {
-    return (
-        <BodyText>
-            <span>one</span>
-            <span>two</span>
-            <span>three</span>
-            <span>four</span>
-            <span>five</span>
-            <span>six</span>
-        </BodyText>
-    );
-}
--- a/packages/eslint-plugin-wonder-blocks/demo/src/no-invalid-bodytext-children-example.tsx
+++ b/packages/eslint-plugin-wonder-blocks/demo/src/no-invalid-bodytext-children-example.tsx
@@ -64,14 +64,6 @@ export function InvalidBlockChildren() {
     );
 }
 
-// ❌ Invalid: BodyText nested inside BodyText (both default to <p>)
-export function InvalidNestedBodyText() {
-    return (
-        <BodyText>
-            <BodyText>nested</BodyText>
-        </BodyText>
-    );
-}
 
 // ❌ Invalid: WB Heading components render as block-level headings
 export function InvalidHeadingChildren() {
--- a/packages/eslint-plugin-wonder-blocks/docs/no-invalid-bodytext-children.md
+++ b/packages/eslint-plugin-wonder-blocks/docs/no-invalid-bodytext-children.md
@@ -1,0 +1,117 @@
+# no-invalid-bodytext-children
+
+Disallow block-level elements and excessive direct children inside `BodyText`.
+
+## Rule Details
+
+`BodyText` renders as a `<p>` element by default. HTML does not allow block-level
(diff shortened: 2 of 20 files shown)


marcysutton:
Ok I'm marking this one for review to see if we want to keep it. There weren't any violations in `frontend`, so I'm not sure it's adding much value! But since I already had it implemented I figured we could discuss.

Here's the integration PR: https://github.com/Khan/frontend/pull/10738

marcysutton:
The navigation tabs playtesting stories failed again even though I updated from `main` 🤔  Odd!

marcysutton:
@claude review once

beaesguerra:
> Ok I'm marking this one for review to see if we want to keep it. There weren't any violations in `frontend`, so I'm not sure it's adding much value! But since I already had it implemented I figured we could discuss.
> 
> Here's the integration PR: [Khan/frontend#10738](https://github.com/Khan/frontend/pull/10738)

I haven't reviewed the PR yet but I tried out the example in the demo project! I think it would be valuable to have still, so future instances can be prevented! 

One thing I noticed is that there are cases where it would get lint errors for both rules: 

image

Would it make sense to only report one of those rules in those cases? 

I also found it interesting that the error squiggly would be on the child element in some cases! I was expecting that the parent element would be the one with the lint rule violation because it's the element with invalid children elements. 

image


I'm also curious if there are any violations in the legacy apps. If not, we can add this rule to `recommended` so that instances of it can't be added in legacy apps too!

marcysutton:
Good call @beaesguerra -- I adjusted the division of responsibility between the two rules so they shouldn't conflict anymore. I didn't see any violations in `frontend` so maybe it could be `recommended`? I wonder if the parent one should be recommended too? 🤔

beaesguerra:
> Good call @beaesguerra -- I adjusted the division of responsibility between the two rules so they shouldn't conflict anymore. I didn't see any violations in `frontend` so maybe it could be `recommended`? I wonder if the parent one should be recommended too? 🤔

I think it makes to make them `recommended` yes! Especially when they aren't needing any migrations 😄 all `recommended` rules will get included in `strict` by default too!

@marcysutton Were you also able to check `perseus`? It would be great to upgrade the wb eslint plugin in perseus too when we introduce new rules!

packages/eslint-plugin-wonder-blocks/src/rules/jsx-utils.ts
```diff
@@ -131,3 +171,17 @@ export function getAttributeStringValue(
...
 }
+
+/**
+ * Returns true when the BodyText opening element has no tag prop or a tag prop
+ * of "p", meaning it will render as a <p> element.
+ *
+ * Block-container tags (div, section, …) and inline tags (span, …) both
+ * return false — only the default and explicit tag="p" cases return true.
+ */
+export function rendersAsParagraph(
```
beaesguerra: **small nit:** It would be helpful to include `BodyText` in the name of the function! Since `rendersAsParagraph` is in a shared utils file, it might not be expected that a `null` tag would mean the `openingElement` is a paragraph.

For example, if a `Heading` was passed into this with no tag, `rendersAsParagraph` would currently return `true`, which would be misleading since `Heading` doesn't default to a `p` tag!

packages/eslint-plugin-wonder-blocks/src/configs/strict.ts
```diff
@@ -5,6 +5,7 @@ export default {
     rules: {
         ...recommended.rules,
         "@khanacademy/wonder-blocks/no-custom-tab-role": "error",
+        "@khanacademy/wonder-blocks/no-invalid-bodytext-children": "error",
```
beaesguerra: Note: If we decide to put this in `recommended`, we can remove this line since the `strict` rules includes all the `recommended` rules!

packages/eslint-plugin-wonder-blocks/demo/src/no-invalid-bodytext-children-example.tsx
```diff
@@ -0,0 +1,94 @@
...
+            </BodyText>
+        </>
+    );
+}
+
+// ❌ Invalid: View is never valid inside BodyText
+export function InvalidViewChild() {
+    return (
+        <BodyText>
+            <View>layout</View>
```
beaesguerra, line 54: If `<View>` had `tag="span"`, would that be valid?

packages/eslint-plugin-wonder-blocks/docs/no-invalid-bodytext-children.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Rule Details
+
+`BodyText` renders as a `<p>` element by default. HTML does not allow block-level
+elements (e.g. `<div>`, `<section>`, headings) as children of `<p>`. Placing them
+inside a default `BodyText` produces invalid markup that browsers must silently
+repair, which can cause layout or accessibility issues.
+
+`View` always renders as `<div>` and is never valid as a child of `BodyText`,
+regardless of the `tag` prop.
```
beaesguerra: Interesting! This is related to my previous question. If `View` has a `tag` prop set to an inline element, and it's a child of `BodyText`, would that be valid?

packages/eslint-plugin-wonder-blocks/docs/no-invalid-bodytext-children.md
```diff
@@ -0,0 +1,116 @@
...
+
+This rule also warns when `BodyText` has more than 5 direct JSX element children
+(configurable). `BodyText` is for text content — complex layouts with many child
+elements belong in a `View` or other block container.
```
beaesguerra: I wonder if it would make sense to separate out the maximum number of direct JSX element children into it's own rule, so the 2 sets of logic and config can be independent. What do you think?

packages/eslint-plugin-wonder-blocks/src/rules/__tests__/jsx-utils.test.ts
```diff
@@ -80,3 +80,60 @@ describe("getAttributeStringValue", () => {
...
+        const node = parseOpeningElement(`<BodyText tag="div" />`);
+
+        // Act
+        const result = rendersAsParagraph(node);
+
+        // Assert
+        expect(result).toBe(false);
+    });
+
+    it("returns true when tag is a dynamic expression (unknown value treated as paragraph)", () => {
```
beaesguerra, line 132: It would be helpful to document this behaviour in the jsdocs for `rendersAsParagraph`! Currently, it states: 

```
...only the default and explicit tag="p" cases return true
```

packages/eslint-plugin-wonder-blocks/src/rules/__tests__/no-invalid-bodytext-children.test.ts
```diff
@@ -0,0 +1,229 @@
...
+        {
+            code: `<BodyText><h2>Heading</h2></BodyText>`,
+            errors: [{messageId: "blockChild", data: {childName: "h2"}}],
+        },
+
+        // ---------------------------------------------------------------- //
+        // WB Heading components (render as block-level headings)           //
+        // ---------------------------------------------------------------- //
+        {
+            code: `<BodyText><Heading>Title</Heading></BodyText>`,
```
beaesguerra: If WB Heading had a `tag`, would it be valid in this case?

packages/eslint-plugin-wonder-blocks/src/rules/__tests__/no-invalid-bodytext-children.test.ts
```diff
@@ -0,0 +1,229 @@
...
+            errors: [
+                {
+                    messageId: "tooManyChildren",
+                    data: {count: "6", max: "5"},
+                },
+                {messageId: "divChild"},
+            ],
+        },
+    ],
+});
```
beaesguerra, line 252: Neat, great set of tests!

beaesguerra (commented):
Going through and reviewing the code now that we decided to keep the rule! Left some questions 😄

packages/eslint-plugin-wonder-blocks/src/rules/__tests__/no-invalid-bodytext-children.test.ts
marcysutton: `Heading` can't ever be a valid child because it only accepts h1-h6 elements in the `tag` prop. Good question, though!

packages/eslint-plugin-wonder-blocks/demo/src/no-invalid-bodytext-children-example.tsx
marcysutton: Yes it would be valid! I've updated that one (View with tag=span).

packages/eslint-plugin-wonder-blocks/docs/no-invalid-bodytext-children.md
marcysutton: Yes, an inline tag prop would make it valid inside of `BodyText`. I was curious if CSS affected this and Claude claims it doesn't impact content model validity. I don't think that always holds up in a screen reader though -- `display` can definitely affect screen reader output. 

> CSS display does not affect HTML content model validity. The spec defines valid nesting based on element type, not visual presentation. The HTML parser sees element types at parse time, before any CSS applies. Setting display: inline on a `<div>` still produces an invalid `<div>` inside `<p>` — browsers silently repair it by implicitly closing the `<p>` before the `<div>`, regardless of styling.

For this iteration, I'm limiting it to looking at `tag` props since those impact what element gets rendered.

packages/eslint-plugin-wonder-blocks/docs/no-invalid-bodytext-children.md
marcysutton: Yes -- I split it off into a separate rule and included it in this PR!

packages/eslint-plugin-wonder-blocks/src/rules/no-invalid-bodytext-children.ts
```diff
@@ -0,0 +1,154 @@
...
+type Options = [];
+type MessageIds = "viewChild" | "divChild" | "paragraphChild" | "blockChild";
+
+export default createRule<Options, MessageIds>({
+    name: "no-invalid-bodytext-children",
+    meta: {
+        type: "problem",
+        docs: {
+            description:
+                "Disallow block-level elements and excessive children inside BodyText",
```
beaesguerra:
```suggestion
                "Disallow block-level elements inside BodyText",
```
Since they're separate rules now!

beaesguerra (approved):
Looks good to me! 2 more lint rules yay 🎉

packages/eslint-plugin-wonder-blocks/src/rules/no-invalid-bodytext-children.ts
marcysutton: Good call! Thank you!

PR #3017, opened 2026-04-20
Improve robustness of responsive tabs logic to avoid flaky sb tests

beaesguerra:
## Summary:

Previously, the Storybook interactions tests run in CI would be flaky for the ResponsiveNavigationTabs Toggle Icon interaction test. The test would sometimes fail because the layout wouldn't properly switch to the dropdown layout after the icons were toggled on. This was because we would check overflow when the tabs data changes with icons, rather than when the icons are rendered. To make it more robust, we check overflow once the tabs layout is rendered and when there are any changes to it or its descendants (using a MutationObserver).

Other improvements include:
- Triggering the `checkOverflow` logic only when the width changes in the ResizeObserver to prevent additional calls 
- Fixing edge cases where additional container padding/margin could cause continuous layout changes

Issue: WB-2311

## Test plan:

1. Unit tests pass
2. SB Interaction tests for ResponsiveTabs and ResponsiveNavigationTabs continue to pass (layout should change as expected based on triggers like changing the number of tabs, changing the tab labels, zoom level, if there are icons)
  - `?path=/story/packages-tabs-responsivetabs--interactive`
  - `?path=/story/packages-tabs-responsivenavigationtabs--interactive`

### DIFF
--- a/packages/wonder-blocks-tabs/src/hooks/use-responsive-layout.ts
+++ b/packages/wonder-blocks-tabs/src/hooks/use-responsive-layout.ts
@@ -89,8 +89,12 @@ export function useResponsiveLayout(
                     scrollableWrapper.clientWidth;
 
                 if (hasOverflow) {
-                    // Store the width before switching
-                    tabsWidthRef.current = scrollableWrapper.scrollWidth;
+                    // Store the container width needed to show tabs without
+                    // overflow. Accounts for any gap (padding/margin) between
+                    // the container and the scrollable element.
+                    tabsWidthRef.current =
+                        scrollableWrapper.scrollWidth +
+                        (container.clientWidth - scrollableWrapper.clientWidth);
                     setShowDropdown(true);
                 }
             }
--- a/packages/wonder-blocks-tabs/src/hooks/use-responsive-layout.ts
+++ b/packages/wonder-blocks-tabs/src/hooks/use-responsive-layout.ts
@@ -106,9 +106,10 @@ export function useResponsiveLayout(
     }, [showDropdown, elementWithOverflowRef, containerRef]);
 
     React.useEffect(() => {
-        // When tabsSignature changes and dropdown is shown, reset to tabs
-        // layout so we can re-measure with the new content. The MutationObserver
-        // handles checking for overflow once the new content is mounted.
+        // When tabsSignature changes and dropdown is shown, reset to horizontal tabs
+        // layout so we can re-measure with the new content (label, icon, number of tabs, etc).
+        // The MutationObserver handles checking for overflow once the new content
+        // is rendered.
         if (showDropdown) {
             tabsWidthRef.current = null;
             setShowDropdown(false);
--- a/packages/wonder-blocks-tabs/src/hooks/use-responsive-layout.ts
+++ b/packages/wonder-blocks-tabs/src/hooks/use-responsive-layout.ts
@@ -62,6 +62,9 @@ export function useResponsiveLayout(
     // we can switch back to the tabs view when the container width is wide enough.
     const tabsWidthRef = React.useRef<number | null>(null);
 
+    // Keep track of the container width to avoid unnecessary overflow checks when the container width is the same.
+    const containerWidthRef = React.useRef<number | null>(null);
+
     // Create a signature of the tabs to detect changes (tab added/removed, label changed, presence of tab icon)
     const tabsSignature = React.useMemo(
         () =>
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-tabs/src/hooks/use-responsive-layout.ts
```diff
@@ -89,8 +89,12 @@ export function useResponsiveLayout(
...
 
                 if (hasOverflow) {
-                    // Store the width before switching
-                    tabsWidthRef.current = scrollableWrapper.scrollWidth;
+                    // Store the container width needed to show tabs without
+                    // overflow. Accounts for any gap (padding/margin) between
+                    // the container and the scrollable element.
+                    tabsWidthRef.current =
+                        scrollableWrapper.scrollWidth +
+                        (container.clientWidth - scrollableWrapper.clientWidth);
```
beaesguerra, line 100: This wasn't related to the test flakiness, but an improvement to avoid an edge case where the layout constantly changes between the dropdown and horizontal tabs layout when the root container has padding/margin. It now accounts for additional spacing between the container and scrollable element

beaesguerra:
@claude review once

packages/wonder-blocks-tabs/src/hooks/use-responsive-layout.ts
```diff
@@ -106,19 +113,13 @@ export function useResponsiveLayout(
...
         if (showDropdown) {
-            // When tabsSignature changes and dropdown is shown, reset to
-            // tabs layout so we can re-measure and see if we can switch
-            // back
             tabsWidthRef.current = null;
             setShowDropdown(false);
-        } else {
-            // When tabsSignature changes and tabs layout is shown, check
-            // for overflow
-            checkOverflow();
```
beaesguerra, line 121: Previously, when `tabsSignature` changed when the tabs layout is shown, we would check for overflow. Now, this check for overflow happens in the MutationObserver so that we check for overflow after the DOM has updated

packages/wonder-blocks-tabs/src/hooks/use-responsive-layout.ts
```diff
@@ -130,8 +131,13 @@ export function useResponsiveLayout(
             return;
         }
 
+        // Check for overflow when the container size changes (screen size change, width change, etc)
         const resizeObserver = new ResizeObserver(() => {
-            checkOverflow();
+            // Only check for overflow if the container width has changed
+            if (containerWidthRef.current !== container.clientWidth) {
+                containerWidthRef.current = container.clientWidth;
+                checkOverflow();
```
beaesguerra, line 139: Not directly related to the SB test flakiness, but we can reduce the number of times we check for overflow by only checking when there is a change in width in the resize observer. 

Previously, `checkOverflow` would also be called whenever the layout changed since the height of the dropdown and horizontal tabs layout were different, so the resize observer would always trigger the checking logic.

packages/wonder-blocks-tabs/src/hooks/use-responsive-layout.ts
```diff
@@ -141,6 +147,32 @@ export function useResponsiveLayout(
...
+        // icons mounted, labels changed) in the horizontal tabs layout.
+        const mutationObserver = new MutationObserver(() => {
+            checkOverflow();
+        });
+
+        mutationObserver.observe(element, {
+            childList: true,
+            subtree: true,
+            characterData: true,
+        });
```
beaesguerra, line 169: This is where we set up the mutation observer so we check for overflow when a descendant has a change.

This is what should address the SB test flakiness since it will check for overflow after the icons are rendered

marcysutton (approved):
This looks good to me! I do wonder whether there will be noticeable performance impacts of a MutationObserver in a useEffect. But if it fixes the test flakiness and doesn't visibly thrash the layout, seems like a decent trade-off!

beaesguerra:
> This looks good to me! I do wonder whether there will be noticeable performance impacts of a MutationObserver in a useEffect. But if it fixes the test flakiness and doesn't visibly thrash the layout, seems like a decent trade-off!

@marcysutton The MutationObserver should only be triggered when the horizontal tabs layout changes descendant nodes (adds/removes tabs or adds/removes icons in tabs) or the tab labels change!

One thing to note is that the observer is watching the horizontal tabs element, rather than the container element (which contains both horizontal tabs and dropdown tabs). Because of this, the MutationObserver is set up and taken down when the layout changes so that we don't have to watch for unrelated changes when it is using the dropdown layout.

Realistically, most of the responsive tabs use cases are static, so the mutation observer is there to handle edge cases!

PR #3019, opened 2026-04-21
Update Storybook deps

beaesguerra:
## Summary:

Update Storybook dependencies to the latest, including the mcp add on.

Issue: WB-2301

## Test plan:

Storybook is updated to v10.3.5 `?path=/settings/about`

image


jandrade (approved):
Looks great! thanks for the upgrade 🎉 

FYI this new version includes this fix which allow us to hide the deprecated components from the docs via URL (if we want). This means that in the future we are going to be able to persist Design-related docs for certain scenarios (pending the design docs collab with the WB Design team).


https://5e1bf4b385e3fb0020b7073c-pzugevikmr.chromatic.com/?path=/docs/foundations-using-color--docs&tags=!deprecated

beaesguerra:
> FYI this new version includes this fix which allow us to hide the deprecated components from the docs via URL (if we want). This means that in the future we are going to be able to persist Design-related docs for certain scenarios (pending the design docs collab with the WB Design team).
> 
> https://5e1bf4b385e3fb0020b7073c-pzugevikmr.chromatic.com/?path=/docs/foundations-using-color--docs&tags=!deprecated

@jandrade When I open this url, I am able to see the deprecated components still 😅 not sure if this is a different bug! 

image

(Either way, it'll be helpful to be on the latest sb version, I'll land the PR!)

PR #3020, opened 2026-04-21
Static token docs for Storybook MCP

beaesguerra:
## Summary:

Based on the [Storybook AI Best Practices](https://storybook.js.org/docs/ai/best-practices#docs-content), the Storybook manifest for the mcp is limited to the information that is explicitly present in files. Since our current token docs dynamically render the token information, the token info is not available via the WB SB MCP. I confirmed this too by asking Claude how it'd get token information:

image

To address this, a `generate-token-docs` script was created that will take the tokens and generate corresponding mdx files for Storybook. The mdx includes a static table of the token values in the different themes. This should make it so the tokens are available via the mcp server!

This script is run when Storybook is built or started. The generated mdx files aren't committed to the repo since it is derived from the token values. The mdx files can be viewed in Storybook under `References`. 

image

We also remove the original token stories from the manifest in favour of the new reference files

Issue: WB-2301

## Test plan:

1. Review the static token docs and confirm that it includes the important tokens (ie.`/?path=/docs/references-token-map-border--docs`)
2. Connect to the Storybook MCP and verify that token details can be found using the mcp
Screenshot 2026-04-22 at 4 13 05 PM

3. Confirm the token map information is present in `/manifests/docs.json` and in `/manifests/components.html`

image

image


beaesguerra:
@claude review once

package.json
```diff
@@ -28,6 +30,7 @@
     "publish:check": "pnpm run lint:pkg-json && ./utils/publish/pre-publish-check-ci.ts",
     "publish:ci": "pnpm run publish:check && git diff --stat --exit-code HEAD && pnpm build && pnpm build:types && pnpm changeset publish",
     "start": "pnpm start:storybook",
+    "prestart:storybook": "pnpm build:token-docs",
```
beaesguerra, line 33: The token docs will be generated whenever we run `pnpm build` or `pnpm start`/`pnpm start:storybook`!

packages/wonder-blocks-tokens/src/build/generate-token-docs.ts
```diff
@@ -0,0 +1,286 @@
...
+        tokenPrefix: "font",
+        description:
+            "Typography tokens: family, weight, body, heading, and text-decoration.",
+        getTokens: (theme) => {
+            return {
+                ...theme.font,
+                // Exclude the legacy size tokens
+                size: {},
+                // Exclude the legacy lineHeight tokens
+                lineHeight: {},
```
beaesguerra, line 112: This is how we omit deprecated font tokens from the reference tables

packages/wonder-blocks-tokens/src/build/generate-token-docs.ts
```diff
@@ -0,0 +1,286 @@
...
+        tokenPrefix: "semanticColor",
+        description:
+            "Semantic colors are the recommended way to color UI. Prefer these over primitive `color` tokens.",
+        getTokens: (theme) => {
+            return {
+                ...theme.semanticColor,
+                // Exclude the action tokens
+                action: {},
+                // Exclude the legacy status tokens
+                status: {},
```
beaesguerra, line 129: Another example of omitting tokens from the reference files

packages/wonder-blocks-tokens/src/build/generate-token-docs.ts
```diff
@@ -0,0 +1,284 @@
+#!/usr/bin/env -S node -r @swc-node/register
```
beaesguerra, line 1: This script was mostly generated by claude! It takes the tokens for the different themes and then creates mdx files for each grouping

.gitignore
```diff
@@ -18,4 +18,8 @@ pnpm-debug.log
 doctor-storybook.log
 
 # Ignore local Claude settings
-.claude/settings.local.json
\ No newline at end of file
+.claude/settings.local.json
+
+# Generated by `pnpm build:token-docs` via the `prestart:storybook` /
+# `prebuild:storybook` hooks. Never checked in.
+__docs__/references/
```
beaesguerra, line 25: We put the reference docs in the gitignore so that they don't need to be maintained in git. The mdx files will be created whenever we build or start storybook!

packages/wonder-blocks-tokens/src/build/generate-token-docs.ts
```diff
@@ -0,0 +1,284 @@
...
+import fs from "node:fs";
+import path from "node:path";
+
+import themeDefault from "../theme/default";
```
marcysutton, line 6: Should we add a TODO comment to update this when we change the default theme to TB?

packages/wonder-blocks-tokens/src/build/generate-token-docs.ts
```diff
@@ -0,0 +1,284 @@
...
+const themes: ReadonlyArray<{
+    id: string;
+    label: string;
+    theme: ThemeObject;
+}> = [
+    {id: "default", label: "Default", theme: themeDefault},
+    {id: "thunderblocks", label: "Thunderblocks", theme: themeThunderblocks},
+];
+
+function flattenToMap(
```
marcysutton: Can we add some tests for this function and `renderThemedTable` so they don't break or change silently?

.gitignore
marcysutton: How do we guarantee stability if the files are excluded, though? Could we add a snapshot test or something?

marcysutton (commented):
Thanks for working on this, @beaesguerra! It's going to be super helpful!

As a sidenote: I didn't see Tokens Map in the Storybook `components.json`, but I did in the `components.html` manifest.

I left one question and a suggestion for tests, but that's about it!

.gitignore
beaesguerra: Yes great idea! I created snapshot tests for the content rendered in the `.mdx` files so we can detect if something is unintentionally changed!

packages/wonder-blocks-tokens/src/build/generate-token-docs.ts
beaesguerra: I added some unit tests for better coverage!

beaesguerra:
> As a sidenote: I didn't see Tokens Map in the Storybook `components.json`, but I did in the `components.html` manifest.

Thanks for the feedback @marcysutton ! You should see the Tokens Maps in `http://localhost:6061/manifests/docs.json` 😄 

`components.json` includes the documentation from CSF (component story format) files, and `docs.json` includes the documentation from `mdx` files! For more details, check out: https://storybook.js.org/docs/ai/manifests#docs-mdx-manifest . This is something we get now after updating to the latest Storybook version 😄 (This also makes it so our other mdx docs are available via the MCP now too!) 

And it seems `components.html` will show the contents from both `components.json` and `docs.json`!

marcysutton (approved):
Looks good! Thank you @beaesguerra!

PR #3021, opened 2026-04-22
🔷 Add tokenValue() runtime helper to wonder-blocks-tokens

jandrade:
## Summary

Add a new `tokenValue(token, element?)` helper to `@khanacademy/wonder-blocks-tokens` that resolves a semantic token to its concrete computed value at runtime (e.g. `#1865f2`). This is useful when handing WB tokens to third-party libraries that expect raw color values instead of CSS `var(...)` references. Because it reads `getComputedStyle`, the returned value automatically reflects whichever theme is active on the element (via the existing `data-wb-theme` CSS cascade).

Issue: [WB-2160](https://khanacademy.atlassian.net/browse/WB-2160)

## Test plan

- [ ] Run `pnpm start` and navigate to **Packages / Tokens / Utilities / tokenValue** in Storybook.
- [ ] Verify the demo renders two sections (`default theme` and `thunderblocks theme`) and that each shows a different raw hex/rgba value for `semanticColor.core.foreground.instructive.default`.
- [ ] Confirm unit tests pass: `pnpm run jest packages/wonder-blocks-tokens/src/util/__tests__/token-value.test.ts`.

## Review plan

1. 🚨 [`token-value.ts`](https://github.com/Khan/wonder-blocks/pull/3021#discussion_r3127758365)

This change is small and purely additive (no existing API surface is modified), so the rest of the diff — index re-exports, unit tests, MDX docs page, and changeset — is trivial risk.

[WB-2160]: https://khanacademy.atlassian.net/browse/WB-2160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/wonder-blocks-tokens/src/util/token-value.ts
```diff
@@ -0,0 +1,42 @@
+/**
```
jandrade, line 1: 🚨 **1 of 1:** Net-new public API in a shared lib consumed by multiple apps. The helper parses `var(--name)` (ignoring any `var()` fallback), reads `getComputedStyle(element).getPropertyValue(name).trim()`, and passes non-`var()` inputs through unchanged. No SSR guard — calling it outside a browser will throw, which is intentional per the ticket scope. cc: @Khan/wonder-blocks

jandrade:
@claude review once

.changeset/wb-2160-token-value.md
beaesguerra, line 1: Interesting, the changeset filename isn't the usual format of random words! Was there a reason why it's different in this case? I don't think the file name matters, but was curious haha!

__docs__/wonder-blocks-tokens/token-value.mdx
```diff
@@ -0,0 +1,124 @@
...
+    semanticColor,
+    sizing,
+    tokenValue,
+} from "@khanacademy/wonder-blocks-tokens";
+
+<Meta title="Packages / Tokens / Utilities / tokenValue" />
+
+# tokenValue()
+
+Wonder Blocks semantic tokens are exported as CSS `var(--...)` references so
```
beaesguerra:
```suggestion
Wonder Blocks tokens are exported as CSS `var(--...)` references so
```

Can `tokenValue` be used for other tokens too, not just semantic ones?

__docs__/wonder-blocks-tokens/token-value.mdx
```diff
@@ -0,0 +1,124 @@
...
+
+tokenValue(token: string, element?: Element): string;
+```
+
+## API
+
+| Argument  | TypeScript Type | Default                     | Description                                                                                                                           |
+| --------- | --------------- | --------------------------- | ------------------------------------------------------------------------------------------------------------------------------------- |
+| `token`   | `string`        | _Required_                  | A token value, typically a `var(--...)` reference from this package. Values that are not `var(...)` expressions are returned as-is.   |
+| `element` | `Element`       | `document.documentElement`  | The element whose computed style is read. Pass a descendant of a `[data-wb-theme]` element to resolve the value for that theme scope. |
```
beaesguerra: Should the default be `document.body`, since that's where we set `data-wb-theme` usually? From Classroom for example:

Image

packages/wonder-blocks-tokens/src/util/token-value.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Matches a CSS `var(--name)` expression and captures the custom property name.
+ * Anything after the first `,` (the `var()` fallback) is ignored.
+ */
+const cssVarRegex = /^\s*var\(\s*(--[^,)\s]+)/;
+
+/**
+ * Resolves the raw value of a Wonder Blocks token at runtime.
+ *
+ * Semantic tokens are exported as `var(--...)` strings so that theming works
```
beaesguerra, line 10:
```suggestion
 * Tokens are exported as `var(--...)` strings so that theming works
```

(Suggestion related to my question about if other tokens can be used with `token-value`!)

packages/wonder-blocks-tokens/src/util/token-value.ts
```diff
@@ -0,0 +1,42 @@
...
+ * ```ts
+ * import {semanticColor, tokenValue} from "@khanacademy/wonder-blocks-tokens";
+ *
+ * tokenValue(semanticColor.core.foreground.instructive.default);
+ * // => "#1865f2"
+ * ```
+ */
+export function tokenValue(
+    token: string,
+    element: Element = document.documentElement,
```
beaesguerra: Same question about defaulting to `document.body`! 

Also, I remember we were chatting about this utility (related [thread](https://khanacademy.slack.com/archives/C4PE1QM5Y/p1764177714721979?thread_ts=1764175339.646689&cid=C4PE1QM5Y)) and we were discussing using `closest()` to find the closest ancestor that defines the theme. I'm curious if you had some new insight on this! 

I'm curious how the utility would handle:
- nested themes
- theming in portalled components

__docs__/wonder-blocks-tokens/token-value.mdx
beaesguerra: Are there specific use cases where consumers should pass in an `element`? Or is it more for flexibility? If there are common cases, could we document that?

packages/wonder-blocks-tokens/src/util/__tests__/token-value.test.ts
```diff
@@ -0,0 +1,106 @@
+import {semanticColor} from "../../index";
```
beaesguerra: What do you think about adding some tests for the other tokens too 😄 (sizing, boxShadow, border radius/width, media query breakpoints, etc)

packages/wonder-blocks-tokens/src/util/__tests__/token-value.test.ts
```diff
@@ -0,0 +1,106 @@
...
+            "#1865f2",
+        );
+        const scoped = document.createElement("div");
+        scoped.style.setProperty(INSTRUCTIVE_FOREGROUND_VAR, "#ffff00");
+        document.body.appendChild(scoped);
+
+        // Act
+        const result = tokenValue(
+            semanticColor.core.foreground.instructive.default,
+            scoped,
```
beaesguerra, line 48: If the `scoped` element for some reason didn't have the css variable set, it would currently return `""`. Would it be expected that it would return the value of the css variable from the `documentElement` (or `body` if we change the default)?

beaesguerra (commented):
This utility will be helpful, thanks for adding it! I have some questions I wanted to get your thoughts on!

.changeset/wb-2160-token-value.md
jandrade: I was curious at first too :). Turns out Claude created the changesets file by itself without using the `pnpm changeset` command directly. This seems to be ok, but it's definitively better to use the usual commands to avoid any possible issues.

I'm going to fine tune the skill I'm using to create these PRs so it takes that command into account.

packages/wonder-blocks-tokens/src/util/__tests__/token-value.test.ts
jandrade: hmmm good question! I think it would make sense to fallback to the `:root` element.

__docs__/wonder-blocks-tokens/token-value.mdx
jandrade: good point, I'll change it!

packages/wonder-blocks-tokens/src/util/__tests__/token-value.test.ts
jandrade: Added unit tests for these cases, thanks for the suggestion!

__docs__/wonder-blocks-tokens/token-value.mdx
jandrade: I added new docs in the page to include this: `/?path=/docs/packages-tokens-utilities-tokenvalue--docs#overriding-the-element-argument`

Let me know what you think!

packages/wonder-blocks-tokens/src/util/token-value.ts
jandrade: This should work with nested themes via the `element` ref. I feel it would be the same with portalled components as far as the ref/element is passed in.

Is there something you think would be missing from this?

.storybook/preview.tsx
```diff
@@ -150,14 +158,15 @@ const parameters: Preview["parameters"] = {
...
-            setLocalTheme(theme);
-        }
-    }, [theme]);
+    // React.useEffect(() => {
+    //     if (theme) {
+    //         // Switch the body class based on the theme.
+    //         document.body.setAttribute(THEME_DATA_ATTRIBUTE, theme);
+    //         setLocalTheme(theme);
+    //     }
+    // }, [theme]);
```
beaesguerra: Did we want to put these changes back?

__docs__/wonder-blocks-tokens/token-value.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+tokenValue(token: string, element?: Element): string;
+```
+
+## API
+
+| Argument  | TypeScript Type | Default                    | Description                                                                                                                           |
+| --------- | --------------- | -------------------------- | ------------------------------------------------------------------------------------------------------------------------------------- |
+| `token`   | `string`        | _Required_                 | A token value, typically a `var(--...)` reference from this package. Values that are not `var(...)` expressions are returned as-is.   |
+| `element` | `Element`       | `document.documentElement` | The element whose computed style is read. Pass a descendant of a `[data-wb-theme]` element to resolve the value for that theme scope. |
```
beaesguerra, line 42: Should we update the documented `Default` value to reflect both default fallbacks? 

```
  document.querySelector("[data-wb-theme]") || document.documentElement
```

__docs__/wonder-blocks-tokens/token-value.mdx
```diff
@@ -0,0 +1,110 @@
...
+// Reads the value that would apply inside a specific theme scope.
+const instructiveInScope = tokenValue(
+    semanticColor.core.foreground.instructive.default,
+    elementInsideThunderblocksScope,
+);
+```
+
+## Overriding the `element` argument
+
+The `element` argument is optional and defaults to the nearest `[data-wb-theme]`
```
beaesguerra: Should this be "defaults to the _first_ `[data-wb-theme]`"? 

Since we use `document.querySelector`, it'll return the first instance found!

packages/wonder-blocks-tokens/src/util/token-value.ts
```diff
@@ -0,0 +1,45 @@
...
+ * through the CSS cascade. This helper reads the computed value of that custom
+ * property, returning the concrete string (e.g. a hex or `rgba()` value) that
+ * third-party libraries expecting raw color values can consume.
+ *
+ * If `token` is not a `var(...)` expression, it is returned unchanged.
+ *
+ * @param token A token string — typically a `var(--...)` reference exported
+ * from this package, but a raw value is accepted and returned as-is.
+ * @param element The element whose computed style should be read. Defaults
+ * to the nearest `[data-wb-theme]` element in the current document, falling
```
beaesguerra:
```suggestion
 * to the first `[data-wb-theme]` element in the current document, falling
```

Similar to the other comment, where document.querySelector will return the first element rather than the nearest!

__docs__/wonder-blocks-tokens/token-value.mdx
```diff
@@ -0,0 +1,110 @@
...
+element in the current document, falling back to `document.body`. Pass a
```
beaesguerra:
```suggestion
element in the current document, falling back to `document.documentElement`. Pass a
```

Did you mean `documentElement`? (This is what's currently implemented, though `document.querySelector("[data-wb-theme]")` will likely find the document.body first since that's where we put the data attribute!

__docs__/wonder-blocks-tokens/token-value.mdx
```diff
@@ -0,0 +1,110 @@
...
+```tsx
+import {semanticColor, tokenValue} from "@khanacademy/wonder-blocks-tokens";
+
+const instructive = tokenValue(
+    semanticColor.core.foreground.instructive.default,
+    elementInsideThunderblocksScope,
+);
+```
+
+## Non-semantic tokens
```
beaesguerra, line 99: Thanks for adding this!

__docs__/wonder-blocks-tokens/token-value.stories.tsx
```diff
@@ -0,0 +1,162 @@
...
+    value,
+}: {
+    color: string;
+    label: string;
+    value: string;
+}) => {
+    return (
+        <View style={styles.colorItem}>
+            {label}:
+            <Badge
```
beaesguerra, line 54: Neat 😄

packages/wonder-blocks-tokens/src/util/token-value.ts
beaesguerra: Ah! I was misunderstanding 😄 I see `getComputedStyle` will figure out what the value is based on the cascade, so it's fine that the property isn't directly on the element!

beaesguerra (commented):
Left some comments around leftover debugging code and clarification around the default element behaviour since the docs didn't reflect the implementation!

beaesguerra (approved):
Looks good! This utility will be helpful!

PR #3023, opened 2026-04-23
Update SB react docgen type

beaesguerra:
## Summary:

From the [Storybook AI Best Practices](https://storybook.js.org/docs/ai/best-practices#prop-type-extraction) doc, they recommend switching the docgen type to `react-docgen-typescript` for more accurate docs. They warn it is slower but more accurate. We can try this out for now and always switch back if we run into performance issues! The `manifests/components.html` page also recommended it:

Screenshot 2026-04-22 at 4 57 00 PM

Because of the docgen change, our component props that extend `AriaProps` now include all the aria attributes in the component props table:

Screenshot 2026-04-23 at 5 31 20 PM

To avoid overwhelming the docs with these attributes, we continue to hide them from the prop tables via argtypes. Components can override specific aria attributes as needed where relevant so they can be included for components that were explicitly defining these arg types.

image

We also do the same thing to omit event handler prop docs: 

Screenshot 2026-04-24 at 11 04 34 AM

After these changes, we should also be able to remove _some_ cases where we re-define prop descriptions/types due to more complicated types. I have left those cases as is for now and we can start removing duplicated descriptions/types in argtypes as we go along!

Issue: WB-2301

## Test plan:

1. Review prop docs for components to confirm it shows the expected props.
2. Aria prop argtypes that were defined before should still be available (example: `aria-label` prop docs should be included for IconButton still)
3. `manifests/components.html` states that `react-docgen-typescript` was used
Screenshot 2026-04-22 at 4 57 19 PM


beaesguerra:
@claude review once

__docs__/wonder-blocks-typography/typography.argtypes.ts
```diff
@@ -37,4 +42,12 @@ export default {
...
         table: {type: {summary: "string"}},
     },
+    // Hide internal data attribute props
+    "data-placement": {
+        table: {disable: true},
+    },
+    // Hide internal data attribute props
+    "data-modal-launcher-portal": {
+        table: {disable: true},
+    },
```
beaesguerra: This also omits the internal data attributes:

Image

https://github.com/Khan/wonder-blocks/blob/026fcc04be53f90fd255a20253807fe47913cc7f/packages/wonder-blocks-core/src/util/types.ts#L102-L105

beaesguerra:
@claude review once

__docs__/wonder-blocks-core/view.argtypes.ts
```diff
@@ -69,4 +74,12 @@ export default {
...
         },
     },
+    // Hide internal data attribute props
+    "data-placement": {
+        table: {disable: true},
+    },
+    // Hide internal data attribute props
+    "data-modal-launcher-portal": {
+        table: {disable: true},
+    },
```
beaesguerra: This omits the internal data attributes:

Image

https://github.com/Khan/wonder-blocks/blob/026fcc04be53f90fd255a20253807fe47913cc7f/packages/wonder-blocks-core/src/util/types.ts#L102-L105

jandrade: Cool!

I wonder if `@ignore` in the code comment does the same effect?

__docs__/wonder-blocks-form/checkbox.stories.tsx
```diff
@@ -8,12 +8,16 @@ import {Checkbox, CheckboxGroup, Choice} from "@khanacademy/wonder-blocks-form";
...
+import AriaArgTypes from "../wonder-blocks-core/aria.argtypes";
 import ComponentInfo from "../components/component-info";
 import Strut from "../../packages/wonder-blocks-layout/src/components/strut";
 
 export default {
     title: "Packages / Form / Checkbox",
     component: Checkbox,
+    argTypes: {
+        ...AriaArgTypes,
+    },
```
jandrade, line 20: **nit:** This can be probably passed in directly (no spread needed). It's fine as it is.


```suggestion
    argTypes: AriaArgTypes,
```

__docs__/wonder-blocks-typography/typography.argtypes.ts
jandrade: 🤔 this makes me wonder if we should omit those props in Typography components (probably out of scope for this PR)

jandrade (approved):
Looks great! thanks for improving Storybook docs :shipit:

__docs__/wonder-blocks-typography/typography.argtypes.ts
beaesguerra: I created https://khanacademy.atlassian.net/browse/WB-2320 for this!

__docs__/wonder-blocks-core/view.argtypes.ts
beaesguerra: Yes great question! This wasn't working so I configured `react-typescript-docgen` to ignore prop docs with the `@ignore` tag. So now any prop docs with this tag should no longer get included in the prop table (the pre-existing behaviour was properly omitting the ignored props so good catch!) 

I set `@ignore` on these internal props as well and removed the argtype overrides for those props!

__docs__/wonder-blocks-form/checkbox.stories.tsx
beaesguerra: I'll leave it as is so it's easy to configure other argTypes!

PR #3028, opened 2026-04-29
Update LLM guidance with agent skills

marcysutton:
## Summary:
Initially we added docs to .mdc files for Cursor, which I've been missing since switching to Claude. Claude breaks our test guidelines in particular and I keep having to prompt it with the basics - even with information in CLAUDE.md. To solve this, we should try including test guidance in skill files as that is the more recent format.

Issue: WB-XXXX

## Test plan:
1. Do a task in Claude involving tests and see if it still fails our guidance
2. Create stories and see if they follow guidelines


.agents/skills/storybook/SKILL.md
jandrade, line 1: **question:** Does the agents spec support this approach? I understand that skills (at least in Claude) would need to follow the `storybook/SKILL.md` structure, but maybe that has changed recently?

marcysutton: ooh good question! I think it does need to have that structure. I updated the filenames and references!

jandrade:
@claude review once

marcysutton:
Thanks for the Claude review @jandrade! I made those fixes and added a couple more to help Claude find the skills from the WB skill. They are properly populating in a chat:

Prompt: build a story for a button (intentionally vague). It missed the skill the first time, but then confirmed it found it

Claude prompt for writing a Wonder Blocks button story

Prompt for unit testing: successful

Claude prompt for writing unit tests

.agents/skills/wonder-blocks/SKILL.md
```diff
@@ -25,6 +25,8 @@ If the Figma MCP or WB Storybook MCP is used, use each for its purpose:
 - **Figma MCP** — design specs and layout intent. Start here.
 - **WB Storybook MCP** — source of truth for component APIs. Never guess prop names or token paths.
 - **This skill** — foundations and general best practices for using the design system.
+- **Writing stories** — follow the Storybook best practices skill for guidance on writing good stories as a form of documentation.
+- **Writing tests** — follow the unit testing best practices skill for guidance on writing good tests for your components.
```
beaesguerra: One thing to note is that the WB skill description includes:

```
Do NOT trigger for debugging TypeScript errors, writing tests, or fixing CI/lint issues in WB packages.
```

This is because a lot of the WB guidance isn't needed for writing tests! 

In `frontend` and `perseus`, the wb skill description includes not using the it when writing stories. I omitted that from the wb skill in `wonder-blocks` since we want our examples in the docs to follow WB best practices!

I'm wondering now though if we do want the WB skill to trigger when writing stories anywhere! I was assuming that WB shouldn't really be used in `frontend` stories since those stories would be using frontend components. 

Let me know if you have thoughts on this!

.cursor/rules/mdc-files.mdc
```diff
@@ -434,17 +434,18 @@ project-root/
...
-| Cursor File | Copilot File | Claude File |
-|-------------|--------------|-------------|
-| `.cursor/rules/general.mdc` | `.github/instructions/frontend-rules.instructions.md` | `CLAUDE.md` |
-| `.cursor/rules/storybook.mdc` | `.github/instructions/storybook.instructions.md` | `CLAUDE.md` (Storybook section) |
-| `.cursor/rules/unit-tests.mdc` | `.github/instructions/unit-tests.instructions.md` | `CLAUDE.md` (Jest section) |
+| Cursor File | Copilot File | Claude File | Agent Skills File |
+|-------------|--------------|-------------|-------------------|
+| `.cursor/rules/general.mdc` | `.github/instructions/frontend-rules.instructions.md` | `CLAUDE.md` | *(use `CLAUDE.md`)* |
+| `.cursor/rules/storybook.mdc` | `.github/instructions/storybook.instructions.md` | `CLAUDE.md` (Storybook section) | `.agents/skills/storybook/SKILL.md` |
+| `.cursor/rules/unit-tests.mdc` | `.github/instructions/unit-tests.instructions.md` | `CLAUDE.md` (Jest section) | `.agents/skills/unit-tests/SKILL.md` |
```
beaesguerra: Can we remove the existing agent rules files for storybook and unit tests since we have skills for them now?
- https://github.com/Khan/wonder-blocks/blob/main/.github/instructions/storybook.instructions.md
- https://github.com/Khan/wonder-blocks/blob/main/.github/instructions/storybook.instructions.md
- https://github.com/Khan/wonder-blocks/blob/main/.cursor/rules/storybook.mdc
- https://github.com/Khan/wonder-blocks/blob/main/.cursor/rules/unit-tests.mdc

I _think_ we shouldn't need instructions in both rules and skills! What do you think?

CLAUDE.md
```diff
@@ -212,324 +212,46 @@ packages/wonder-blocks-*/
...
-For stateful stories, combine actions with state updates:
-`action("onChange")(newValue); setValue(newValue);`
+- Title hierarchy: `"Packages / Button / Button"`, `"Packages / Button / Testing / Snapshots / ActivityButton"`
+- Default story must be interactive (works with controls panel)
+- Write stories for all prop combinations/states — these feed visual regression in Chromatic
+- Disable Chromatic snapshots (`chromatic: {disableSnapshot: true}`) for interaction/playtesting stories
+- Use `StateSheet` for pseudo-state grids (focus, hover, active, disabled); `ScenariosLayout` for edge cases (RTL, long text, overflow)
+- Use `play` functions for browser-specific behavior jsdom can't handle (scroll, layout, clipboard)
+- Use function declarations for render functions, not arrow functions: `render: function Render(args) {`
+- Separate snapshot stories (`tags: ["!autodocs"]`) from documentation stories
```
beaesguerra: Should this be in the Storybook skill?

CLAUDE.md
```diff
@@ -212,324 +212,46 @@ packages/wonder-blocks-*/
...
+- For thrown errors, wrap the call in an `underTest` function in the Act section
+- Use `userEvent` instead of `fireEvent` for all interactions
+- Query priority: `getByRole` → `getByLabelText` → `getByText` → `getByTestId`; never CSS selectors or direct node traversal
+- Never mock `console.error` — it hides real failures
+- Always use `jest.spyOn()` to create spies; only store the return value in a variable when asserting on it
+- Never mock outside of test cases
+- Don't test style-only props — use Storybook visual regression instead
+- Use `it.each` for multiple input/output combinations
+- Prefer one `expect` per test
+- Include `toHaveNoA11yViolations` tests; check `aria-disabled="true"`, not the `disabled` attribute
```
beaesguerra: Should this be in the unit test skill?

CLAUDE.md
```diff
@@ -212,324 +212,46 @@ packages/wonder-blocks-*/
...
-
-## Assertions
-
-- Use specific matchers: `toBe`, `toEqual`, `toHaveBeenCalledWith`
-- Use RTL matchers: `toBeInTheDocument()`, `toBeVisible()`, `toHaveAttribute()`
-- **Avoid Jest snapshots** (`.toMatchSnapshot()`) - use Chromatic + Storybook for visual regression
-- Prefer one expect per test when possible
+- Props: expected behavior when set; default values; `it.each` for combinations (skip style-only props)
+- Event handlers: triggered by correct conditions; called with correct arguments
+- Accessibility: roles/ARIA wired correctly, keyboard interactions, focus management, accessible names
```
beaesguerra: Should this be in the unit test skill too?

CLAUDE.md
```diff
@@ -586,24 +308,26 @@ Don't hesitate to bump major or minor versions when appropriate—following semv
...
-| Claude File | Cursor File | Copilot File |
-|-------------|-------------|--------------|
-| `CLAUDE.md` | `.cursor/rules/general.mdc` | `.github/instructions/frontend-rules.instructions.md` |
-| `CLAUDE.md` (Storybook section) | `.cursor/rules/storybook.mdc` | `.github/instructions/storybook.instructions.md` |
-| `CLAUDE.md` (Jest section) | `.cursor/rules/unit-tests.mdc` | `.github/instructions/unit-tests.instructions.md` |
+| Claude File | Cursor File | Copilot File | Agent Skills File |
+|-------------|-------------|--------------|-------------------|
+| `CLAUDE.md` | `.cursor/rules/general.mdc` | `.github/instructions/frontend-rules.instructions.md` | *(use `CLAUDE.md`)* |
+| `CLAUDE.md` (Storybook section) | `.cursor/rules/storybook.mdc` | `.github/instructions/storybook.instructions.md` | `.agents/skills/storybook/SKILL.md` |
+| `CLAUDE.md` (Jest section) | `.cursor/rules/unit-tests.mdc` | `.github/instructions/unit-tests.instructions.md` | `.agents/skills/unit-tests/SKILL.md` |
```
beaesguerra: We can update tables like these if we end up removing the storybook/unit test rule files!

beaesguerra: I'm glad the .`agents/skills` format is more agent agnostic than the rules files 😄

beaesguerra (commented):
Thanks for moving the Storybook and unit test guidance to skills! Left some comments/questions for you!

.agents/skills/wonder-blocks/SKILL.md
marcysutton: Hmmmm. Why _wouldn't_ the WB skill be relevant for writing tests or stories? My reasoning for these entries was if an agent is creating a Wonder Blocks component (as it's in this repo), it should create tests and stories following our guidelines. I think in `frontend`, the story part would at least be relevant since I see stories with raw buttons and things. Having skills cross-reference each other seems like a good way to network the best practices together. WYDT?

beaesguerra: Originally, I was thinking that the WB skill shouldn't get triggered when writing tests because there shouldn't be implementation logic in unit tests! Otherwise, those unit tests would be testing a custom implementation created in the test file, rather than the original component under test. I had a similar concern for triggering the WB skill in frontend for stories (ie. covering custom UI implementations in a story, rather than the component that would get used in the application) 

When I go through the details in the WB skill in this repo, I don't think any of it currently applies to tests (ex: there isn't really a need to use design tokens or typography components or other various WB components in unit tests). Let me know though if there is guidance in there that would help with tests!

On another note, there is a separate ticket for creating a skill for creating new WB components! I think for that, we would tie in the stories and unit tests skills https://khanacademy.atlassian.net/browse/WB-2302 since that would be included as part of creating a new component. Currently, this WB skill covers how to use WB! 

Since we are starting to add more skills, I'm curious to learn if chaining the skills is effective or if it overloads context. We can adapt things as needed :) We didn't have other skills yet when we first introduced the WB skill so I haven't tested it out yet!

.cursor/rules/mdc-files.mdc
marcysutton: Cursor appears to use skills, so I'd say we can remove the mdc files. Copilot still used its instructions so I think we should keep those.

.agents/skills/wonder-blocks/SKILL.md
marcysutton: I went ahead and removed those lines. Didn't seem that important right now!

.cursor/rules/mdc-files.mdc
beaesguerra: FWIW, I see copilot supports agent skills now! https://github.blog/changelog/2025-12-18-github-copilot-now-supports-agent-skills/ 

I haven't tested it out yet though! This is something we can remove later on too

beaesguerra (approved):
Skill changes look good, thanks for working on this! 

There's a remaining comment about removing duplicate guidance from the cursor and copilot rules files too, but that's not blocking!

PR #3031, opened 2026-05-01
⚠️ Soft-deprecate `spacing` tokens and ship migration codemod (PR 1 of 3)

jandrade:
Adds `@deprecated` JSDoc to the `spacing` primitive token and its `VALID_*` type exports so editors surface deprecation warnings. Introduces `wb-codemod/transforms/migrate-spacing-to-sizing.ts` — a best-effort jscodeshift transform that:

- rewrites `spacing.<key>` → `sizing.<mapped>` in style objects, JSX attributes, and other plain references;
- rewrites arithmetic involving `spacing.*` into a `calc(...)` template literal composed from `sizing` tokens; and
- leaves a `// TODO(spacing-migration):` comment on patterns it can't safely auto-migrate (computed access, spread, `VALID_*` type imports).

No runtime or visual changes. Step 1 of the 3-PR rollout in `plan.md`; internal migration (PR 2) and `spacing` export removal (PR 3) follow.

Reviewers: #frontend-infra-web

Issue: WB-2314

## Test plan

- [x] `pnpm typecheck`, `pnpm jest`, `pnpm build:storybook` all clean (verified locally).
- [ ] In an editor (VS Code), open any consumer that imports `spacing` from `@khanacademy/wonder-blocks-tokens` (e.g. `packages/wonder-blocks-accordion/src/components/accordion-section-header.tsx`) and confirm the hover shows the `@deprecated` JSDoc and a strikethrough on `spacing` and `VALID_*` types.
- [ ] (Optional) Dry-run the codemod against a heavy consumer: `node wb-codemod/bin/cli.js -t migrate-spacing-to-sizing -d -p packages/wonder-blocks-accordion` and skim the output for sanity.

## Review plan

Please review this risky change:

1. ⚠️ [`migrate-spacing-to-sizing.ts`](https://github.com/Khan/wonder-blocks/pull/3031#discussion_r3174883985)

### Common patterns

**4 sites in 1 file:** `@deprecated` JSDoc on the four `spacing` public-API exports, all sharing one message that points at `sizing` and the codemod path.

**1 transform:** New jscodeshift codemod that walks each top-level `BinaryExpression` containing a `spacing.X` and rewrites the whole arithmetic tree as a `calc(...)` template literal. Numeric literals get `px` for `+/-` and stay unitless for `*//`. Example:

```ts
// Before
const INNER_BORDER_RADIUS = spacing.small_12 - 1;

// After
[...]

PR #3032, opened 2026-05-04
⚠️ Migrate internal Wonder Blocks packages from `spacing` to `sizing` (PR 2 of 3)

jandrade:
PR 2 of 3 in the `spacing` → `sizing` migration (WB-2315, depends on PR #3031). Migrates every internal `packages/` use of the deprecated `spacing` primitive token to its `sizing` equivalent. Non-breaking; component packages bump **patch** only.

## What's in the diff

1. **Mechanical codemod output** (commit 1) — `node wb-codemod/bin/cli.js -t migrate-spacing-to-sizing packages` rewrites 18 files: simple member access (`spacing.medium_16` → `sizing.size_160`), arithmetic into `calc(...)` template literals, unary minus into `calc(-1 * sizing.X)`, and TODO-comments where it can't safely transform.

2. **Hand-fixes** (commit 2) — replaces internal `<Strut>` usages with CSS `gap` or direct `margin*` styles. Reason: `sizing.X` is a CSS variable string (`var(--wb-sizing-…)`) at runtime, not a number, so it can't be passed to `Strut.size: number`. Sites split as:
   - **gap on parent flex container** — birthday-picker (wrapper for selects + inline error row), tooltip-content (`withTitle` View).
   - **margins on adjacent elements** — field-heading (label/description/field-wrapper/error each get their own margin), checkbox-group / radio-group (new `firstChoiceMetaSpacing` style on the first Choice when meta is present).
   - **hardcoded JS numbers with token references in comments** for sites that genuinely need numbers — tooltip-tail SVG arithmetic constants, `MediaSpec.gutterWidth/marginWidth` and the `1120 + …` maxWidth math in `layout/specs.ts`, tooltip-popper `viewportPadding` (Popper.js option).
   - Resolves the codemod's TODO in `popover-content.tsx` by replacing the arithmetic in `width` with `calc(100% + ${sizing.size_480})` and switching negative margins from invalid `-${sizing.X}` to `calc(-1 * ${sizing.X})`.
   - Updates the tooltip-tail snapshot to capture the now-tokenized `stdDeviation`.

3. **Changeset** (commit 3) — patch bumps on all 10 touched component packages.

## Codemod fixes shipped on PR 1

PR 2 caught and fixed three real bugs in PR 1's codemod (pushed to `WB-2314`, not duplicated here):
- Node 24 type-stripping required `import type` for the type-only jscodeshift names.
- Returning `root.toSource()` for files we didn't touch caused recast to reformat unrelated files; now returns `file.source` directly.
- Unary-minus rewrite was producing invalid CSS (`-var(...)`) and string concatenation with `px` suffix in template literals would yield `"1.6rempx"`. Both now bail with TODO or use `calc(-1 * ...)`.

Issue: WB-2315

Depends on: #3031 (PR 1)

## Test plan

- [x] `pnpm typecheck`, `pnpm jest`, `pnpm build:storybook` all clean (verified locally).
- [ ] **Chromatic review** — the highest-risk part of this PR. Each `sizing.X` value matches the corresponding `spacing.X` at the default 10px baseline, so the **expected diff is zero**. Audit any diff carefully — likely root causes if any: (a) a `gap`/`margin` refactor that reflows differently than the original `<Strut>`, (b) a CSS-var fallback resolving differently in the test environment.
- [ ] In Storybook, smoke-test the affected components: BirthdayPicker (with and without error, with/without `monthYearOnly`), FieldHeading (label + description + error variants), CheckboxGroup / RadioGroup (with/without label/description/error), Tooltip with tail in all four placements, Popover with image illustration.

## Review plan

1. ⚠️ `wb-codemod/transforms/migrate-spacing-to-sizing.ts` (in PR #3031) — context-only, no further changes here, but the calc(-1 * …) fix is what makes negative margins valid. Worth a re-read of `flattenBinary` / `buildCalcTemplate` and the new `isInsideTemplateLiteralExpression` / `isUnsafeTemplateLiteralContext` helpers.
2. ⚠️ `packages/wonder-blocks-form/src/components/field-heading.tsx` — the field is now wrapped in a thin `<View>` for the `marginTop` slot. Adds one DOM element per FieldHeading. Visually equivalent.
3. ⚠️ `packages/wonder-blocks-tooltip/src/components/tooltip-tail.tsx` — `ARROW_WIDTH/HEIGHT/etc.` are hardcoded numbers (mirrored from sizing tokens) because the SVG point math needs JS numbers.
4. ⚠️ `packages/wonder-blocks-layout/src/util/specs.ts` — `MediaSpec.gutterWidth/marginWidth` similarly hardcoded; the `1120 + …` maxWidth arithmetic stays in JS.
5. 🔷 The remaining 7 packages — straight token swap, mostly mechanical from the codemod.


jandrade:
@claude review once

packages/wonder-blocks-tooltip/src/components/tooltip-tail.tsx
beaesguerra, line 1: There seems to be a regression in the tooltip tail styling: 

Image

I'm curious why this wasn't captured in Chromatic!

packages/wonder-blocks-accordion/src/components/accordion-section-header.tsx
```diff
@@ -213,16 +213,16 @@ const styles = StyleSheet.create({
         textAlign: "start",
     },
     headerString: {
-        paddingTop: spacing.medium_16,
-        paddingBottom: spacing.medium_16,
+        paddingBlockStart: sizing.size_160,
```
beaesguerra, line 216: Thanks for the conversion to logical CSS properties too!

packages/wonder-blocks-birthday-picker/src/components/birthday-picker.tsx
```diff
@@ -160,6 +160,15 @@ const defaultStyles = StyleSheet.create({
...
         },
     },
+    errorRow: {
+        flexDirection: "row",
+        placeItems: "center",
+        gap: sizing.size_040,
+        marginBlockStart: sizing.size_040,
+    },
+    errorText: {
+        color: semanticColor.status.critical.foreground,
```
beaesguerra: Could we also switch this to `semanticColor.core.foreground.critical.default` (`status` tokens are to be deprecated in https://khanacademy.atlassian.net/browse/WB-2041).  (This was pre-existing!)

packages/wonder-blocks-search-field/src/components/search-field.tsx
```diff
@@ -262,21 +258,21 @@ const styles = StyleSheet.create({
...
+        insetInlineEnd: sizing.size_040,
     },
     inputStyleReset: {
         display: "flex",
         flex: 1,
         width: "100%",
-        paddingLeft: spacing.xLarge_32,
-        paddingRight: spacing.large_24 + spacing.medium_16,
+        paddingInlineStart: sizing.size_320,
+        paddingInlineEnd: `calc(${sizing.size_240} + ${sizing.size_160})`,
```
beaesguerra: Could we use `sizing.size_400` here instead? 

```suggestion
        paddingInlineEnd: sizing.size_400,
```

I'm wondering if there's a category of additive spacing tokens that were workarounds to get a value that wasn't included in the set of spacing tokens 😅

packages/wonder-blocks-tooltip/src/components/__tests__/tooltip-tail.test.tsx
```diff
@@ -68,7 +68,7 @@ describe("TooltipTail", () => {
       >
         <fegaussianblur
           in="SourceAlpha"
-          stdDeviation="3"
+          stdDeviation="calc(var(--wb-sizing-size_060) / 2)"
```
beaesguerra: Would it make sense to round to the nearest sizing token? Or hardcoded to 3? Or maybe `stdValue` shouldn't be using a `sizing` token! What do you think?

packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx
```diff
@@ -140,7 +139,9 @@ const smallViewportModifier: SmallViewportModifier = {
 export default class TooltipPopper extends React.Component<Props, State> {
     static defaultProps: DefaultProps = {
         rootBoundary: "viewport",
-        viewportPadding: spacing.small_12,
+        // Popper.js's `padding` option expects a number of pixels. The matching
+        // sizing token is `sizing.size_120` (1.2rem = 12px at 10px baseline).
+        viewportPadding: 12,
```
beaesguerra: I wonder if it would make sense to use the upcoming `tokenValue` for something like this! So that values that need to be px can stay in sync with token values (if ever we need to change the sizing token values)

packages/wonder-blocks-layout/src/util/specs.ts
```diff
@@ -1,31 +1,39 @@
...
-const mediaDefaultSpecLargeMarginWidth = spacing.large_24;
+// Pixel values mirror the matching `sizing.size_*` tokens at the default 10px
+// baseline. They are hardcoded because `MediaSpec.gutterWidth` /
+// `marginWidth` are typed `number` and consumed by JS layout code, so the
+// `var(--wb-sizing-*)` strings that the runtime `sizing` token yields cannot
+// be used here.
+const SIZE_160_PX = 16; // sizing.size_160
+const SIZE_240_PX = 24; // sizing.size_240
+const SIZE_320_PX = 32; // sizing.size_320
+const SIZE_480_PX = 48; // sizing.size_480
```
beaesguerra: Would it make sense to use the upcoming `tokenValue` utility for these later on?

packages/wonder-blocks-tooltip/src/components/tooltip-tail.tsx
```diff
@@ -420,12 +431,14 @@ export default class TooltipTail extends React.Component<Props> {
...
-const ARROW_WIDTH = spacing.large_24;
-const ARROW_HEIGHT = spacing.small_12;
+// Layout constants for SVG arithmetic. Hardcoded because the polyline points
+// and the container's `width`/`height` styles need real JS numbers, not the
+// `var(--wb-sizing-*)` strings that the `sizing` token yields. Values mirror
+// the matching `sizing.size_*` tokens at the default 10px baseline.
+const DISTANCE_FROM_ANCHOR = 8; // sizing.size_080
+const MIN_DISTANCE_FROM_CORNERS = 8; // sizing.size_080
+const ARROW_WIDTH = 24; // sizing.size_240
+const ARROW_HEIGHT = 12; // sizing.size_120
```
beaesguerra, line 438: Same question about using the `tokenValue` util for px values!

beaesguerra (commented):
Thanks for working on this! Left some comments 😄 

--- 

I also manually reviewed the Catalog stories with my browser font size set to `Very large`  to see if we run into any issues in the components since more spacing is rem based now. This is what I noticed:
- ✅  The sizing change in SearchField was an improvement for large font sizes! 

| Before | After | 
| --- | --- |
| Image | Image |

- The Combobox ends up with overlapped elements (this was pre-existing!)

Image
Image

- Breadcrumbs: I'm wondering if we need to scale the size of the separator, I noticed it's pretty small despite the large base font size! (Also pre-existing) 

Image


Note: The Catalog stories don't cover all component cases and aren't included in Chromatic snapshots! I also only manually reviewed the default theme. If we wanted to increase our confidence in the rem changes, we could do a Chromatic run in another PR where we increase the base font-size in our Storybook to simulate increasing the browser font size and then see what the snapshots look like after applying the changes in this PR. Just an idea 😄

packages/wonder-blocks-tooltip/src/components/tooltip-tail.tsx
jandrade: yeah that's odd! I fixed it.

packages/wonder-blocks-tooltip/src/components/tooltip-tail.tsx
jandrade: I kept it that way to simplify the implementation for now. This `TooltipTail` file will be removed with the FloatingUI migration.

packages/wonder-blocks-tooltip/src/components/tooltip-popper.tsx
```diff
@@ -132,16 +134,16 @@ const smallViewportModifier: SmallViewportModifier = {
     fn: _modifyPosition,
 };
 
+const VIEWPORT_PADDING = parseInt(remToPx(tokenValue(sizing.size_120)));
```
beaesguerra, line 137: (For later), since 3rd parties might require px sizing values, I wonder if it would be worth documenting `remToPx` in our Storybook docs! I think it was intended to be internal only but seems like this is a valid use case! We can also leave it as is, and see if the need comes up!

packages/wonder-blocks-layout/src/util/specs.ts
```diff
@@ -1,17 +1,13 @@
...
-// Pixel values mirror the matching `sizing.size_*` tokens at the default 10px
-// baseline. They are hardcoded because `MediaSpec.gutterWidth` /
-// `marginWidth` are typed `number` and consumed by JS layout code, so the
-// `var(--wb-sizing-*)` strings that the runtime `sizing` token yields cannot
-// be used here.
-const SIZE_160_PX = 16; // sizing.size_160
-const SIZE_240_PX = 24; // sizing.size_240
-const SIZE_320_PX = 32; // sizing.size_320
-const SIZE_480_PX = 48; // sizing.size_480
+const SIZE_160_PX = parseInt(remToPx(tokenValue(sizing.size_160)));
```
beaesguerra, line 7: TIL `parseInt` can parse numbers out of strings with the unit included 😄

beaesguerra (approved):
Looks good to me!

PR #3035, opened 2026-05-05
I18n: Add Noto as fallback font in TB tokens.fontFamily

marcysutton:
## Summary:
For support in International, we adjust the fallback font for Thunderblocks to be Noto Sans.

Noto wasn't rendering as a fallback font for non-Latin languages here in WB. This change updates the TB token for fontFamily and adjusts how fonts load on the docs site. 

We also clean up warnings in the Chrome console about "The resource <URL> was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally."

Issue: WB-2235

## Test plan:
1. Review rendered fonts on the docs site for Non-Latin languages: `/?path=/story/packages-typography--noto-for-non-latin`
    - it wasn't actually rendering Noto in the default theme either, as far as I could tell
2. Check the Typography tokens for fontFamily: `/?path=/docs/packages-tokens-typography--docs&globals=theme:thunderblocks#font-family`
3. See the resource URL warnings go away in the browser console

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -1,17 +1,6 @@
-<!-- Preload WB fonts to prevent layout shifting due to font rendering -->
-<link rel="preload" href="fonts/lato/Lato-Regular.woff2" as="font" type="font/woff2" crossorigin />
-<link rel="preload" href="fonts/lato/Lato-Bold.woff2" as="font" type="font/woff2" crossorigin />
-<link rel="preload" href="fonts/lato/Lato-Black.woff2" as="font" type="font/woff2" crossorigin />
-<link rel="preload" href="fonts/lato/Lato-Italic.woff2" as="font" type="font/woff2" crossorigin />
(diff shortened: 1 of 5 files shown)


.storybook/preview-head.html
```diff
@@ -1,17 +1,6 @@
-<!-- Preload WB fonts to prevent layout shifting due to font rendering -->
-<link rel="preload" href="fonts/lato/Lato-Regular.woff2" as="font" type="font/woff2" crossorigin />
```
marcysutton, line 2: These weren't helping and were causing a ton of warnings in the browser console

Wonder Blocks docs site in Chrome DevTools

jandrade:
@marcysutton this mostly makes sense to me! do you have an integration branch in `frontend` for this by chance?

marcysutton:
@jandrade yep, here it is! https://github.com/Khan/frontend/pull/10892

jandrade (approved):
Looks great! thanks for the cleanup 🚀

PR #2973, opened 2026-02-23
Feature branch pull-request from `feature/wb-dark-mode` to `main`

beaesguerra:
## Summary:
This PR includes the following commits:
- Set up Storybook environment to support dark themes (#2953)
- Set up syl-dark tokens and theme (#2974)
- Update color swatch in docs to show transparency in colors (#2984)
- SYL Dark: Typography (global color) (#2983)
- SYL Dark - Badges (#3036)
- Minor Storybook updates (#3045)

<You can re-run git pull-request to update this message automatically with any newly-landed PRs.>

Issue: WB-2163

## Test plan:

PR #3036, opened 2026-05-06
SYL Dark - Badges

beaesguerra:
## Summary:

Reviewing the badge components in SYL Dark to confirm they look as expected, match the design specs, and have no a11y color violations. We also enable Chromatic snapshots for the syl-dark theme for the component.

- The semantic color core border neutral tokens were updated to align with design. This was to make it so the NeutralBadge with `showBorder=true` prop would show a visible border.
- The semantic color core background neutral strong token was also updated so that it works well with the foreground knockout token in `syl-dark`. This color combination was used for custom styles as an example for badges

Issue: WB-2166

## Test plan:

Review `?path=/story/packages-badge-testing-snapshots--state-sheet-story&globals=theme:syl-dark` and confirm things look as expected and there are no a11y violations

Review any visual changes in Chromatic due to the token change (note: Since we are gradually enabling the syl-dark snapshots, it should only catch changes in components that have been reviewed already)

image
image
image

### DIFF
--- a/__docs__/wonder-blocks-badge/badge.stories.tsx
+++ b/__docs__/wonder-blocks-badge/badge.stories.tsx
@@ -11,7 +11,6 @@ import {HeadingLarge} from "@khanacademy/wonder-blocks-typography";
 import badgeArgtypes, {iconArgType} from "./badge.argtypes";
 import {multiColoredIcon} from "../components/icons-for-testing";
 import Tooltip from "@khanacademy/wonder-blocks-tooltip";
-import {allThemeModes} from "../../.storybook/modes";
 
 export default {
     title: "Packages / Badge / Badge",
@@ -27,7 +26,6 @@ export default {
         chromatic: {
             // Disable snapshots since they're covered by the testing snapshots
             disableSnapshot: true,
-            modes: allThemeModes,
         },
     },
     render: (args: Omit<PropsFor<typeof Badge>, "icon"> & {icon: string}) => {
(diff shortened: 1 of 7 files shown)


__docs__/wonder-blocks-badge/badge.stories.tsx
```diff
@@ -27,7 +26,6 @@ export default {
         chromatic: {
             // Disable snapshots since they're covered by the testing snapshots
             disableSnapshot: true,
-            modes: themeModes,
```
beaesguerra, line 30: Since we disable snapshots for these stories, we don't need to set `modes`!

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-syl-dark.ts
```diff
@@ -15,9 +15,9 @@ const core = {
             strong: color.blue_50,
         },
         neutral: {
-            subtle: color.gray_10,
-            default: color.gray_20,
-            strong: color.gray_70,
+            subtle: color.gray_30,
+            default: color.gray_40,
+            strong: color.gray_60,
```
beaesguerra, line 20: Caitlyn updated these tokens in Figma so that the NeutralBadge would have a visible border in syl-dark when `showBorder=true`

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-syl-dark.ts
```diff
@@ -57,7 +57,7 @@ const core = {
         neutral: {
             subtle: color.gray_10,
             default: color.gray_20,
-            strong: color.black_100,
+            strong: color.gray_80,
```
beaesguerra, line 60: This fixes the `semanticColor.core.background.neutral.strong` + `semanticColor.core.foreground.knockout.default` combination in `syl-dark` (which we use as an example for custom styling for badges)

beaesguerra:
Note: I updated the feature branch with `main` so that it has the latest changes. Chromatic snapshots need to be reapproved there again: https://github.com/Khan/wonder-blocks/pull/2973 

This is why there are some unrelated snapshots for this PR!

beaesguerra:
@claude review once

marcysutton (approved):
This looks great to me! Thanks @beaesguerra!

PR #3037, opened 2026-05-08
Add require-logical-properties-for-rtl ESLint rule

maddy531:
## Summary:

This PR adds an ESLint rule to Wonder Blocks. It hasn't fixed any visible RTL issue in WB when I reviewed the components locally. It's mainly infrastructure to prevent future regressions. In a follow-up PR, I'll extract the lint rule into a published @khanacademy/eslint-plugin-wonder-blocks-rtl package and consume it from frontend (@beaesguerra you suggested this a while ago and I'm finally picking that up now).

Issue: CLASS-14218

- pnpm lint 
 - pnpm test 
 - pnpm typecheck 

## Test plan:


maddy531:
@claude review

packages/wonder-blocks-dropdown/src/components/checkbox.tsx
```diff
@@ -105,6 +105,6 @@ const styles = StyleSheet.create({
     disabledCheckFormatting: {
         position: "absolute",
         top: -1,
```
beaesguerra, line 107: Interesting! Should it have converted `top` to `insetBlockStart`?

packages/wonder-blocks-dropdown/src/components/separator-item.tsx
```diff
@@ -40,9 +40,9 @@ export default class SeparatorItem extends React.Component<{
 
 const styles = StyleSheet.create({
     separator: {
-        borderTop: `${border.width.thin} solid ${semanticColor.core.border.neutral.subtle}`,
+        borderBlockStart: `${border.width.thin} solid ${semanticColor.core.border.neutral.subtle}`,
         height: 1,
```
beaesguerra, line 44: The linter will convert `minHeight` to `minBlockSize`, is it intended to also cover `height` to `blockSize`? (And same for `width` to `inlineSize`)?

packages/wonder-blocks-modal/src/components/drawer-dialog.tsx
```diff
@@ -247,24 +247,24 @@ const getComponentStyles = ({
 
             // Override FlexibleDialog defaults for drawer usage
             height: "100%",
-            minHeight: "100vh",
+            minBlockSize: "100vh",
 
             // Use common widths for mininum/maximum
-            minWidth: breakpoint.width.xsMax,
+            minInlineSize: breakpoint.width.xsMax,
             maxWidth: breakpoint.width.smMax,
```
beaesguerra, line 254: Does `maxWidth` also need to be converted to `maxInlineSize`? (Probably same applies for `maxHeight` too!)

packages/wonder-blocks-switch/src/components/switch.tsx
```diff
@@ -108,7 +108,7 @@ const sharedStyles = StyleSheet.create({
     slider: {
         position: "absolute",
         top: theme.slider.position.top,
-        left: theme.slider.position.left,
+        insetInlineStart: theme.slider.position.left,
         height: theme.slider.sizing.height,
         width: theme.slider.sizing.width,
         borderRadius: theme.root.border.radius.default,
```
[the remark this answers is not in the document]
beaesguerra: This is a great catch! I wonder if we also need to lint against `translateX` use cases, so devs are reminded about this! It could be a separate rule, especially if we want the autofix to include negating translateX

https://github.com/user-attachments/assets/8dd42dd7-a2ad-43a9-a5b8-81c58406b92b

eslint-local-rules/require-logical-properties-for-rtl.js
```diff
@@ -0,0 +1,152 @@
...
+ * Ported from frontend/libs/eslint-plugin-static-service. Uses plain ESLint
+ * AST type checks instead of @babel/types so no extra dep is needed.
+ */
+
+// This list of replacements is based on the Logical Properties
+// Chart found here:
+// https://adrianroselli.com/2019/11/css-logical-properties.html
+const propertyReplacements = {
+    left: "insetInlineStart",
+    right: "insetInlineEnd",
```
beaesguerra, line 14: After reviewing the other changes, would it make sense to add `top` and `bottom` to this mapping?

eslint-local-rules/require-logical-properties-for-rtl.js
```diff
@@ -0,0 +1,152 @@
...
+    borderRight: "borderInlineEnd",
+    borderLeft: "borderInlineStart",
+    borderTop: "borderBlockStart",
+    borderBottom: "borderBlockEnd",
+    borderTopLeftRadius: "borderStartStartRadius",
+    borderTopRightRadius: "borderStartEndRadius",
+    borderBottomLeftRadius: "borderEndStartRadius",
+    borderBottomRightRadius: "borderEndEndRadius",
+    minWidth: "minInlineSize",
+    minHeight: "minBlockSize",
```
beaesguerra, line 32: Would it make sense to add `maxWidth`, `width`, `minHeight`, and `height` too?

eslint-local-rules/require-logical-properties-for-rtl.js
```diff
@@ -0,0 +1,152 @@
+/**
```
beaesguerra, line 1: We'll want to switch this to TypeScript and add tests when we move it to `eslint-plugin-wonder-blocks`!

package.json
```diff
@@ -126,6 +125,7 @@
     "eslint-plugin-jest": "28.11.0",
     "eslint-plugin-jsdoc": "^50.6.3",
     "eslint-plugin-jsx-a11y": "^6.10.2",
+    "eslint-plugin-local-rules": "^3.0.2",
```
beaesguerra, line 128: I'm curious about this dependency! I don't see `eslint-plugin-local-rules` in `frontend`, so I'm wondering if `eslint-plugin-local-rules` is needed!

__docs__/components/all-variants.tsx
```diff
@@ -192,7 +192,7 @@ export function AllVariants(props: Props) {
 const styles = StyleSheet.create({
     table: {
         borderCollapse: "collapse",
-        textAlign: "left",
+        textAlign: "start",
```
beaesguerra, line 195: I think this change is what's causing a lot of changes in our visual regression tests! We use this component frequently in our stories! Here's an example: https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=69fd3dbe62993b374b0cb954 

It looks like the text is getting centred now, I'm not sure why the specificity of this changed 🤔 One option to fix this is to move `textAlign: "start"` to the `cell` styles! 

When we fix the text alignment here, it'll reduce the changes in Chromatic so we can more easily tell if there were other things impacted!

beaesguerra (commented):
Thanks for working on this Maddy, I'm excited for this! 

I've left some comments! Once they're addressed, I'll review the Chromatic snapshots and do some more testing of the impacted components 😄

__docs__/components/gallery/tiles/phosphor-icon-tile.tsx
```diff
@@ -53,6 +53,6 @@ const localStyles = StyleSheet.create({
         alignItems: "end",
     },
     icon: {
-        marginRight: spacing.medium_16,
+        marginInlineEnd: spacing.medium_16,
```
beaesguerra, line 56: Also, FYI: @jandrade is working on a migration from `spacing` to `sizing` tokens, so someone will have merge conflicts depending on who lands their PR first!

packages/wonder-blocks-dropdown/src/components/checkbox.tsx
maddy531: yeah interesting it didn't get flagged!

eslint-local-rules/require-logical-properties-for-rtl.js
maddy531: yes!

eslint-local-rules/require-logical-properties-for-rtl.js
maddy531: Yup!

maddy531:
Close in favor of using the plugin route.

PR #3043, opened 2026-05-11
SYL Dark - CircularSpinner

beaesguerra:
## Summary:

Reviewing the CircularSpinner component in SYL Dark to confirm it looks as expected,
matches the design specs, and has no a11y color violations. We also enable Chromatic
snapshots for the syl-dark theme for the component.

Changes specific to this component: 
- Adds a `StateSheet` 
- Use semantic color tokens in CircularSpinner

Issue: WB-2166

## Test plan:

Review CircularSpinner in all themes `?path=/story/packages-progressspinner-testing-snapshots-circularspinner--state-sheet-story`

Confirm things look as expected and there are no a11y violations

Review any visual changes in Chromatic

### DIFF
--- a/packages/wonder-blocks-progress-spinner/src/components/circular-spinner.tsx
+++ b/packages/wonder-blocks-progress-spinner/src/components/circular-spinner.tsx
@@ -1,7 +1,7 @@
 import * as React from "react";
 import {StyleSheet} from "aphrodite";
 import {View, addStyle} from "@khanacademy/wonder-blocks-core";
-import {color} from "@khanacademy/wonder-blocks-tokens";
+import {semanticColor} from "@khanacademy/wonder-blocks-tokens";
 
 import type {AriaProps, StyleType} from "@khanacademy/wonder-blocks-core";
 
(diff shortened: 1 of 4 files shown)


packages/wonder-blocks-progress-spinner/src/components/circular-spinner.tsx
```diff
@@ -72,7 +67,9 @@ export default class CircularSpinner extends React.Component<Props> {
 
         const height = heights[size];
         const path = paths[size];
-        const color = light ? colors.light : colors.dark;
+        const color = light
```
jandrade, line 70: **thought:** Now that we have a notion of light and dark modes, it would be good to rename this `light` prop to something else in the future. Something like `inverse`, `knockout`, etc. Wdyt?

No change required as it is out of scope, but something worth considering for a future cleanup phase (in general).

jandrade (approved):
Looks great! thanks 💯

packages/wonder-blocks-progress-spinner/src/components/circular-spinner.tsx
beaesguerra: Yes good point! 

I created [WB-2344](https://khanacademy.atlassian.net/browse/WB-2344) for updating the `light` prop. And created an epic to keep track of [Post-Dark Mode Improvements](https://khanacademy.atlassian.net/browse/WB-2343)!

In Figma, this variant is controlled using a property called `Type`, which supports values `Default` or `Knockout`, so renaming to something similar would make sense for alignment! 

image

PR #3044, opened 2026-05-11
syl-dark tokens: Update `core.background.neutral.default` to `gray_30`

beaesguerra:
## Summary:

Update syl-dark token value for `core.background.neutral.default` from `gray_20` to `gray_30` so that it has high enough contrast when used with `core.foreground.knockout.default` 

Issue: WB-2166

## Test plan:

Review that updated token matches Figma. 

Review Chromatic snapshots

### DIFF
--- a/packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-syl-dark.ts
+++ b/packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-syl-dark.ts
@@ -56,7 +56,7 @@ const core = {
         },
         neutral: {
             subtle: color.gray_10,
-            default: color.gray_20,
+            default: color.gray_30,
             strong: color.gray_80,
         },
         critical: {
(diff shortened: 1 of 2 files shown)


packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-syl-dark.ts
```diff
@@ -56,7 +56,7 @@ const core = {
         },
         neutral: {
             subtle: color.gray_10,
-            default: color.gray_20,
+            default: color.gray_30,
```
beaesguerra, line 59: This token change was based on a discussion with Caitlyn ([related thread](https://khanacademy.slack.com/archives/C07CA2YR0BZ/p1778536281554769?thread_ts=1778530686.100199&cid=C07CA2YR0BZ))

jandrade (approved):
:shipit:

PR #3045, opened 2026-05-11
Minor Storybook updates

beaesguerra:
## Summary:

Addressing comments in https://github.com/Khan/wonder-blocks/pull/2973 as a follow up:
- Put back comment around base 10 unit
- Switch to new typography components 

Also adding `(WIP)` to the theme label in the SB control so that folks know that it's being worked on currently 

Issue: WB-XXXX

## Test plan:

Confirm docs look okay still (component gallery, theme control), general styling


.storybook/preview.tsx
```diff
@@ -268,7 +268,7 @@ const preview: Preview = {
                     {
                         value: "syl-dark",
                         icon: "moon",
-                        title: "Shape Your Learning - Dark",
+                        title: "Shape Your Learning - Dark (WIP)",
```
beaesguerra, line 279: Image

Adding (WIP) to the syl-dark label since this option will be available in our deployed SB instance!

marcysutton (approved):
Thanks for the quick updates!

PR #3049, opened 2026-05-13
⚠️ Make `tokenValue` safe to call when `document` is undefined

jandrade:
## Summary

Adds a `typeof document === "undefined"` guard to `tokenValue` so it returns an empty string (instead of throwing `ReferenceError: document is not defined`) in SSR / node-based test environments. Consumers like `wonder-blocks-layout/src/util/specs.ts` and `wonder-blocks-tooltip/src/components/tooltip-popper.tsx` call `tokenValue` at module load time, so without this guard any downstream test that imports those modules in a node test environment crashes.

Browser behavior is unchanged.

Issue: [WB-2160](https://khanacademy.atlassian.net/browse/WB-2160)

## Test plan

- [x] Existing `token-value.test.ts` unit tests pass (verified locally).
- [ ] Manual: open the TokenValue stories in Storybook (`http://localhost:6061/?path=/docs/wonder-blocks-tokens-tokenvalue--docs`) and confirm the resolved values still render correctly — browser behavior should be a no-op.
- [ ] Downstream: webapp's node-env tests that import `@khanacademy/wonder-blocks-layout` or `-tooltip` should stop throwing `ReferenceError: document is not defined`.

## Review plan

1. ⚠️ [`token-value.ts`](https://github.com/Khan/wonder-blocks/pull/3049#discussion_r3237017522)

[WB-2160]: https://khanacademy.atlassian.net/browse/WB-2160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/wonder-blocks-tokens/src/util/token-value.ts
+++ b/packages/wonder-blocks-tokens/src/util/token-value.ts
@@ -32,14 +32,22 @@ const cssVarRegex = /^\s*var\(\s*(--[^,)\s]+)/;
  * ```
  */
 export function tokenValue(token: string, element?: Element): string {
+    const match = token.match(cssVarRegex);
+    if (!match) {
+        return token;
+    }
+
+    // In non-browser environments (e.g. SSR or node-based tests), there is no
+    // document to read computed styles from. Return an empty string so callers
+    // can resolve values lazily at first render instead of crashing on import.
+    if (typeof document === "undefined") {
+        return "";
+    }
+
     const rootElement =
         element ||
         document.querySelector("[data-wb-theme]") ||
         document.documentElement;
 
-    const match = token.match(cssVarRegex);
-    if (!match) {
-        return token;
-    }
     return getComputedStyle(rootElement).getPropertyValue(match[1]).trim();
 }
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-tokens/src/util/token-value.ts
```diff
@@ -32,14 +32,22 @@ const cssVarRegex = /^\s*var\(\s*(--[^,)\s]+)/;
...
+    const match = token.match(cssVarRegex);
+    if (!match) {
+        return token;
+    }
+
+    // In non-browser environments (e.g. SSR or node-based tests), there is no
+    // document to read computed styles from. Return an empty string so callers
+    // can resolve values lazily at first render instead of crashing on import.
+    if (typeof document === "undefined") {
+        return "";
```
jandrade, line 44: ⚠️ **1 of 1:** Public utility in a shared lib used across many WB packages. The change is purely defensive — it adds a new code path for environments without a `document` and does not alter the existing browser path — but is still a behavior change in shared code (`ReferenceError` → `""`).

jandrade:
Confirmed that this fixes the tests issue in the Snapshot release here: https://github.com/Khan/frontend/pull/11190

cc @Khan/wonder-blocks

jandrade:
@claude review once

beaesguerra (approved):
Looks good to me!

PR #3054, opened 2026-05-18
Add new `no-raw-button` lint rule to eslint-plugin-wonder-blocks

marcysutton:
## Summary:
Adds a new lint rule to warn about raw <button> and StyledButton elements in favor of the WB Buttoncomponent. Typically these custom implementations are lacking design system defaults and accessibility requirements like global focus styles. 

Issue: WB-2340

## Test plan:
1. Ensure tests pass
2. Review changes to ensure they are what we want:
   - Using Button in tests, rather than disabling the rule and isolating deps
   - Updating guidance in action-styles and focus-styles
   - Disallowing custom buttons in Dropdown openers
   - Updating eslint-plugin-wonder-blocks demo to extend tsconfig-common

### DIFF
--- a/__docs__/wonder-blocks-clickable/clickable-behavior.stories.tsx
+++ b/__docs__/wonder-blocks-clickable/clickable-behavior.stories.tsx
@@ -2,7 +2,7 @@ import * as React from "react";
 import {StyleSheet} from "aphrodite";
 import type {Meta, StoryObj} from "@storybook/react-vite";
 
-import {View, addStyle} from "@khanacademy/wonder-blocks-core";
+import {View} from "@khanacademy/wonder-blocks-core";
 import {semanticColor, sizing} from "@khanacademy/wonder-blocks-tokens";
 
 import {getClickableBehavior} from "@khanacademy/wonder-blocks-clickable";
@@ -74,46 +74,6 @@ Default.parameters = {
     },
 };
 
-/**
- * This is an example of a `<ClickableBehavior>` wrapping a button. Since
- * buttons have a built in tabIndex, a tabIndex does not need to be added to
- * `<ClickableBehavior>` here.
- */
-export const WrappingButton: StoryComponentType = (args: any) => {
-    const ClickableBehavior = getClickableBehavior();
-    const StyledButton = addStyle("button");
-
-    return (
-        <ClickableBehavior {...args}>
-            {(state, childrenProps) => {
-                const {pressed, hovered, focused} = state;
-                return (
-                    <StyledButton
-                        style={[
-                            styles.clickable,
-                            styles.newButton,
-                            hovered && styles.hover,
-                            focused && styles.focus,
-                            pressed && styles.press,
-                        ]}
-                        {...childrenProps}
-                    >
-                        This is an element wrapped with ClickableBehavior
-                    </StyledButton>
-                );
-            }}
-        </ClickableBehavior>
-    );
-};
-
-WrappingButton.parameters = {
-    chromatic: {
-        // we don't need screenshots because this story only displays the
-        // resting/default state.
-        disableSnapshot: true,
-    },
-};
-
 export const WithTabIndex: StoryComponentType = () => {
     const ClickableBehavior = getClickableBehavior();
 
@@ -163,11 +123,6 @@ const styles = StyleSheet.create({
         cursor: "pointer",
         padding: sizing.size_160,
         textAlign: "center",
-    },
-    newButton: {
-        border: "none",
-        backgroundColor: actionCategory.default.background,
-        width: "100%",
     },
     hover: {
         textDecoration: "underline",
--- a/__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
@@ -14,6 +14,7 @@ import {BodyText} from "@khanacademy/wonder-blocks-typography";
 import {
     ActionItem,
     ActionMenu,
+    CustomOpener,
     OptionItem,
     SeparatorItem,
 } from "@khanacademy/wonder-blocks-dropdown";
@@ -27,7 +28,6 @@ import type {Item} from "../../packages/wonder-blocks-dropdown/src/util/types";
 import IconButton from "@khanacademy/wonder-blocks-icon-button";
 import {ModalLauncher, OnePaneDialog} from "@khanacademy/wonder-blocks-modal";
 import Button from "@khanacademy/wonder-blocks-button";
-import {focusStyles} from "@khanacademy/wonder-blocks-styles";
 import {StatusBadge} from "@khanacademy/wonder-blocks-badge";
 
 const actionItems: Array<Item> = [
@@ -147,19 +147,19 @@ const styles = StyleSheet.create({
      * Custom opener styles
      */
     customOpener: {
+        display: "inline-flex",
+        alignItems: "center",
+        gap: sizing.size_080,
+        height: sizing.size_400,
+        paddingInline: sizing.size_160,
+        border: `${border.width.thin} solid ${semanticColor.status.warning.foreground}`,
         borderInlineStart: `${border.width.thick} solid ${semanticColor.status.warning.foreground}`,
         borderRadius: border.radius.radius_040,
         background: semanticColor.status.warning.background,
         color: semanticColor.core.foreground.neutral.strong,
-        padding: sizing.size_160,
-    },
-    focused: focusStyles.focus[":focus-visible"],
-    hovered: {
-        textDecoration: "underline",
-        cursor: "pointer",
-    },
-    pressed: {
-        color: semanticColor.status.warning.foreground,
+        ":hover": {
+            filter: "brightness(0.95)",
+        },
     },
 });
 
@@ -347,42 +347,32 @@ export const Controlled: StoryComponentType = {
 };
 
 /**
- * In case you need to use a custom opener, you can use the opener property to
- * achieve this. In this example, the opener prop accepts a function with the
- * following arguments:
- *  - `eventState`: lets you customize the style for different states, such as
- *    pressed, hovered and focused.
- *  - `text`: Passes the menu label defined in the parent component. This value
- *    is passed using the placeholder prop set in the ActionMenu component.
- *  - `opened`: Whether the dropdown is opened.
+ * When you need a fully custom-styled opener, use `CustomOpener`. It provides
+ * a blank-slate `<button>` with the WB focus ring baked in and correct ref
+ * forwarding for the dropdown's focus management wiring.
  *
- * **Note:** If you need to use a custom ID for testing the opener, make sure to
- * pass the testId prop inside the opener component/element.
+ * The `opener` render prop receives `hovered`, `focused`, `pressed`, `text`,
+ * and `opened` values that can be passed to child content for conditional
+ * styling. Focus ring styles are handled automatically by `CustomOpener` via
+ * CSS — you do not need to apply `focusStyles` yourself.
  *
- * **Accessibility:** When a custom opener is used, the following attributes are
- * added automatically: `aria-expanded`, `aria-haspopup`, and `aria-controls`.
+ * **Note:** Pass `testId` directly to `CustomOpener` for e2e test targeting.
+ *
+ * **Accessibility:** When a custom opener is used, `aria-expanded`,
+ * `aria-haspopup`, and `aria-controls` are added automatically.
  */
-
-export const CustomOpener: StoryComponentType = {
+export const WithCustomOpener: StoryComponentType = {
     name: "With custom opener",
     args: {
-        opener: ({focused, hovered, pressed, text}: any) => (
-            <BodyText
-                weight="bold"
-                onClick={() => {
-                    console.log("custom click!!!!!");
-                }}
+        opener: ({text}: any) => (
+            <CustomOpener
                 testId="teacher-menu-custom-opener"
-                style={[
-                    styles.customOpener,
-                    focused && styles.focused,
-                    hovered && styles.hovered,
-                    pressed && styles.pressed,
-                ]}
-                role="button"
+                style={styles.customOpener}
             >
-                {text}
-            </BodyText>
+                <BodyText tag="span" weight="bold">
+                    {text}
+                </BodyText>
+            </CustomOpener>
         ),
     } as Partial<typeof ActionMenu>,
 };
--- a/__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
@@ -24,7 +24,10 @@ import actionMenuArgtypes from "./action-menu.argtypes";
 import ComponentInfo from "../components/component-info";
 import packageConfig from "../../packages/wonder-blocks-dropdown/package.json";
 
-import type {Item} from "../../packages/wonder-blocks-dropdown/src/util/types";
+import type {
+    Item,
+    OpenerProps,
+} from "../../packages/wonder-blocks-dropdown/src/util/types";
 import IconButton from "@khanacademy/wonder-blocks-icon-button";
 import {ModalLauncher, OnePaneDialog} from "@khanacademy/wonder-blocks-modal";
 import Button from "@khanacademy/wonder-blocks-button";
@@ -157,9 +160,18 @@ const styles = StyleSheet.create({
         borderRadius: border.radius.radius_040,
         background: semanticColor.status.warning.background,
         color: semanticColor.core.foreground.neutral.strong,
-        ":hover": {
-            filter: "brightness(0.95)",
-        },
+    },
+    customOpenerHovered: {
+        filter: "brightness(0.95)",
+    },
+    customOpenerPressed: {
+        filter: "brightness(0.9)",
+    },
+    customOpenerDisabled: {
+        color: semanticColor.core.foreground.neutral.subtle,
+        borderColor: semanticColor.core.border.neutral.subtle,
+        background: semanticColor.core.background.base.default,
+        cursor: "not-allowed",
     },
 });
 
@@ -363,17 +375,37 @@ export const Controlled: StoryComponentType = {
  */
 export const WithCustomOpener: StoryComponentType = {
     name: "With custom opener",
-    args: {
-        opener: ({text}: any) => (
-            <CustomOpener
-                testId="teacher-menu-custom-opener"
-                style={styles.customOpener}
+    render: function Render(args) {
+        const [opened, setOpened] = React.useState(false);
+        return (
+            <ActionMenu
+                {...args}
+                opened={opened}
+                onToggle={setOpened}
+                opener={({hovered, pressed, text}: OpenerProps) => (
+                    <CustomOpener
+                        testId="teacher-menu-custom-opener"
+                        styles={{
+                            root: [
+                                styles.customOpener,
+                                hovered && styles.customOpenerHovered,
+                                pressed && styles.customOpenerPressed,
+                                args.disabled && styles.customOpenerDisabled,
+                            ],
+                        }}
+                    >
+                        <BodyText tag="span" weight="bold">
+                            {text}
+                        </BodyText>
+                    </CustomOpener>
+                )}
             >
-                <BodyText tag="span" weight="bold">
-                    {text}
-                </BodyText>
-            </CustomOpener>
-        ),
+                {actionItems.map((actionItem, index) => actionItem)}
+            </ActionMenu>
+        );
+    },
+    args: {
+        disabled: false,
     } as Partial<typeof ActionMenu>,
 };
 
--- a/__docs__/wonder-blocks-dropdown/custom-opener-testing-snapshots.stories.tsx
+++ b/__docs__/wonder-blocks-dropdown/custom-opener-testing-snapshots.stories.tsx
@@ -1,0 +1,94 @@
+import * as React from "react";
+import {StyleSheet} from "aphrodite";
+import type {Meta, StoryObj} from "@storybook/react-vite";
+
+import caretDownIcon from "@phosphor-icons/core/regular/caret-down.svg";
+import {View} from "@khanacademy/wonder-blocks-core";
+import {CustomOpener} from "@khanacademy/wonder-blocks-dropdown";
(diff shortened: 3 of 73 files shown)


__docs__/wonder-blocks-clickable/clickable-behavior.stories.tsx
```diff
@@ -74,46 +74,6 @@ Default.parameters = {
     },
 };
 
-/**
```
marcysutton, line 77: Yeeting this example out of here to de-emphasize this pattern of ClickableBehavior!

__docs__/wonder-blocks-styles/action-styles.stories.tsx
```diff
@@ -5,27 +5,34 @@ import ComponentInfo from "../components/component-info";
...
+ *   inverse styles to a WB component, pass the style via the `style` prop.
  *
  * ```tsx
  * import {actionStyles} from "@khanacademy/wonder-blocks-styles";
  *
- * <StyledButton style={actionStyles.inverse}>
- *      Custom button
- * </StyledButton>
+ * // Applying to a WB component via style prop
+ * <Button style={actionStyles.inverse}>
```
marcysutton: I wasn't 100% sure what to do with this one, except to avoid confusion around custom components with `StyledButton`. It made sense to me to swap it out for a `Button` component. Maybe we need to rethink this utility a bit?

__docs__/wonder-blocks-styles/focus-styles.stories.tsx
```diff
@@ -3,24 +3,41 @@ import {Meta, StoryObj} from "@storybook/react-vite";
...
+ *   appearance on a WB component, pass the style via the `style` prop.
  *
  * ```tsx
  * import {focusStyles} from "@khanacademy/wonder-blocks-styles";
  *
- * <StyledButton style={focusStyles.focus}>
- *      Custom button
- * </StyledButton>
+ * // Applying to a WB component via style prop
+ * <IconButton
```
marcysutton: Same as `action-styles` -- these utils seem limited to internal use cases now that we're disallowing raw buttons.

packages/eslint-plugin-wonder-blocks/demo/tsconfig.json
```diff
@@ -1,30 +1,8 @@
...
-      "strictFunctionTypes": true,
-      "strictPropertyInitialization": true,
-      "strictBindCallApply": true,
-      "noImplicitAny": true,
-
-      /* Completeness */
-      "skipDefaultLibCheck": true, // it's safe to assume that built-in .d.ts files are correct
-      "skipLibCheck": false
+        "moduleResolution": "node",
+        "noImplicitAny": true
```
marcysutton, line 6: Note: The `IconButton` in the demo for this rule required some updates to `tsconfig`. There was quite a bit of overlap with the common config so I opted to change it a bit.

__docs__/wonder-blocks-dropdown/multi-select.accessibility.stories.tsx
```diff
@@ -80,13 +80,12 @@ const MultiSelectCustomOpenerLabeledField = () => {
...
-                                    icon={magnifyingGlassIcon}
-                                    size="medium"
-                                />
-                            </button>
+                        opener={() => (
+                            <IconButton
+                                icon={magnifyingGlassIcon}
+                                aria-label="Search"
+                                onClick={() => {}}
+                            />
```
beaesguerra, line 88: 🎉 nice! This is the type of things we want it to catch!

__docs__/wonder-blocks-styles/action-styles-variants.stories.tsx
```diff
@@ -111,15 +109,13 @@ export default {
                 },
             },
             {
-                name: "Using an HTML element",
```
beaesguerra, line 114: I wonder if we should leave this example as is, since it was intentionally testing an HTML element! These visual regression tests are helpful for validating that we still support these possible use cases, though we want to discourage the use of html elements! What do you think? 

cc: @jandrade in case you had thoughts on this since you worked on the action styles!

__docs__/wonder-blocks-styles/action-styles.stories.tsx
```diff
@@ -5,27 +5,34 @@ import ComponentInfo from "../components/component-info";
...
 
 /**
- * Styles that can be used to create reusable states for interactive elements.
+ * Styles that implement interactive states (hover, press, focus) for elements
+ * on inverse/dark backgrounds.
  *
- * NOTE: We recommend using existing Wonder Blocks interactive components (e.g.
- * `Button`, `IconButton`, `Link` etc.). These styles are meant to be used
- * as a last resort when our components do not meet your needs.
+ * `actionStyles` is used **internally** by Wonder Blocks components to ensure
```
beaesguerra: I see `actionStyles` is used externally in `frontend`! And I _think_ the actionStyles are meant to be used externally so this might be misleading! We do prefer consumers to use WB components directly though. 

cc: @jandrade in case you had thoughts on this!

__docs__/wonder-blocks-styles/action-styles.stories.tsx
beaesguerra: I wasn't sure too, I've left some questions in the comments and tagged Juan too since he may have more context around the actionStyles!

__docs__/wonder-blocks-styles/focus-styles.stories.tsx
```diff
@@ -154,20 +169,21 @@ export const Scenarios: Story = {
...
             },
             {
-                name: "Using an HTML element",
+                name: "Using Button",
                 props: {
                     children: (
-                        <StyledButton style={focusStyles.focus}>
-                            Custom button
-                        </StyledButton>
+                        <Button onClick={() => {}} style={focusStyles.focus}>
```
beaesguerra: Hmmm I wonder if we should also leave these as is to show that focusStyles work on elements that don't already have the focus styling!

beaesguerra (commented):
Thanks for working on this Marcy! I started reviewing and left some questions I wanted to get your thoughts on first since they were following a similar theme! I'll start a thread on Slack too so we can discuss more!

__docs__/wonder-blocks-styles/focus-styles.stories.tsx
jandrade: yeah, I'd keep it this way (using the current custom element) as `Button` already uses `focusStyles` internally.

__docs__/wonder-blocks-styles/action-styles-variants.stories.tsx
jandrade: To get back to this.... I feel that we should keep this for now until we figure out a better strategy around Classic+devadmin usage.

__docs__/wonder-blocks-styles/action-styles.stories.tsx
jandrade: That's correct, `actionStyles` are also used in `frontend` to work well over dark backgrounds, so we would need to keep this while we figure a long term solution from the Design/UX perspective.

marcysutton: Restored!

__docs__/wonder-blocks-styles/action-styles-variants.stories.tsx
marcysutton: Good call, I put the existing usage back in!

__docs__/wonder-blocks-styles/focus-styles.stories.tsx
marcysutton: Reset back to the way it was!

.changeset/stale-taxis-grow.md
```diff
@@ -0,0 +1,17 @@
...
+"@khanacademy/wonder-blocks-icon-button": minor
+"@khanacademy/wonder-blocks-announcer": minor
+"@khanacademy/wonder-blocks-clickable": minor
+"@khanacademy/wonder-blocks-dropdown": minor
+"@khanacademy/wonder-blocks-popover": minor
+"@khanacademy/wonder-blocks-button": minor
+"@khanacademy/wonder-blocks-modal": minor
+"@khanacademy/wonder-blocks-card": minor
+"@khanacademy/wonder-blocks-core": minor
+"@khanacademy/wonder-blocks-tabs": minor
```
beaesguerra: For the packages that were only updated with a comment, could we change them to a patch version since there are no changes to the API?

packages/wonder-blocks-dropdown/src/index.ts
```diff
@@ -19,6 +20,7 @@ export {
     MultiSelect,
     Combobox,
     Listbox,
+    CustomOpener,
```
beaesguerra, line 23: Can we create a separate changeset for adding a new CustomOpener component to the dropdown package?

__docs__/wonder-blocks-styles/focus-styles.stories.tsx
```diff
@@ -46,8 +68,8 @@ const StyledButton = addStyle("button");
  * the element is focused. This is used for accessibility purposes as it allows
  * the element to present a focus state on Windows High Contrast mode.
  *
- * In the example below, the focus style is applied to an `IconButton` component
- * and to a `button` element.
```
beaesguerra, line 50: Could we revert this comment?

__docs__/wonder-blocks-styles/focus-styles.stories.tsx
```diff
@@ -37,6 +53,12 @@ export default {
 
 type Story = StoryObj<any>;
 
+// The no-raw-button rule is disabled here intentionally: these stories
+// demonstrate focusStyles applied to an element that does NOT already include
+// them. Using a WB Button would obscure this because Button applies focusStyles
+// internally. This pattern should be rare in consumer code — prefer WB
+// interactive components which include focus styles out of the box.
+// eslint-disable-next-line @khanacademy/wonder-blocks/no-raw-button
```
beaesguerra: We can use the eslint disable comment formatting for providing context to why a rule is disabled! https://eslint.org/docs/latest/use/configure/rules#comment-descriptions

beaesguerra:
@claude review once

__docs__/wonder-blocks-styles/focus-styles.stories.tsx
```diff
@@ -11,16 +11,32 @@ import Clickable from "@khanacademy/wonder-blocks-clickable";
...
  *
- * <StyledButton style={focusStyles.focus}>
- *      Custom button
- * </StyledButton>
+ * // Applying to a WB component via style prop
+ * <IconButton
+ *     icon={myIcon}
+ *     aria-label="Action"
+ *     style={focusStyles.focus}
+ * />
```
beaesguerra: Can we remove this example? Consumers shouldn't need to apply focus styles to WB components that already have focus styles!

packages/eslint-plugin-wonder-blocks/src/configs/strict.ts
```diff
@@ -8,5 +8,6 @@ export default {
         "@khanacademy/wonder-blocks/no-excessive-bodytext-children": "error",
         "@khanacademy/wonder-blocks/no-invalid-bodytext-parent": "error",
         "@khanacademy/wonder-blocks/no-hardcoded-color": "error",
+        "@khanacademy/wonder-blocks/no-raw-button": "error",
```
beaesguerra, line 11: Do you have plans to also update the eslint-plugin in perseus too? (This can be done later if you prefer, I think the plugin is behind a couple versions in perseus, but it'll be something we want to keep up to date as we add more rules!)

packages/eslint-plugin-wonder-blocks/src/rules/no-raw-button.ts
```diff
@@ -0,0 +1,81 @@
...
+
+type Options = [];
+type MessageIds = "noRawButton";
+
+export default createRule<Options, MessageIds>({
+    name: "no-raw-button",
+    meta: {
+        docs: {
+            description:
+                'Disallow raw <button> elements and addStyle("button") components in favor of Wonder Blocks Button or IconButton.',
```
beaesguerra: What do you think about generalizing it to: 
```suggestion
                'Disallow raw <button> elements and addStyle("button") components in favor of Wonder Blocks components.',
```

Since we also have ActivityButton, ActivityIconButton, ConversationIconButton, and NodeIconButton components too!

packages/eslint-plugin-wonder-blocks/src/rules/no-raw-button.ts
```diff
@@ -0,0 +1,81 @@
...
+            recommended: true,
+        },
+        messages: {
+            noRawButton:
+                "Avoid '{{element}}'. Use Button or IconButton from Wonder Blocks for consistent focus styles, keyboard navigation, and design system behavior.",
```
beaesguerra: Same question about generalizing the listed WB components. 

Another consideration is custom tab implementations (`<button role="tab">`). We have another lint rule to catch that case too, which would recommend the ResponsiveTabs component!

packages/wonder-blocks-dropdown/src/components/custom-opener.tsx
```diff
@@ -0,0 +1,143 @@
...
+import {StyleSheet} from "aphrodite";
+
+import {addStyle} from "@khanacademy/wonder-blocks-core";
+import type {StyleType} from "@khanacademy/wonder-blocks-core";
+import {focusStyles} from "@khanacademy/wonder-blocks-styles";
+
+const StyledButton = addStyle("button");
+
+type Props = Omit<
+    React.ButtonHTMLAttributes<HTMLButtonElement>,
```
beaesguerra: One thing to flag is other button-like components in WB only support a subset of ButtonHTMLAttributes! So this will include more props than other WB components. I think long term, it could makes sense to support all native element attributes in WB components, however, I think we've also considered limiting the props to accept so we can fine tune what we want to allow! 

I think this is fine since this aims to replace native `<button>` elements, but thought I'd mention it! (cc: @jandrade in case you had thoughts on this too!)

packages/wonder-blocks-dropdown/src/components/custom-opener.tsx
```diff
@@ -0,0 +1,143 @@
...
+    "disabled" | "style"
+> & {
+    /**
+     * Content to render inside the button.
+     */
+    children: React.ReactNode;
+    /**
+     * Whether the opener is disabled.
+     *
+     * Internally, `aria-disabled` is used so the element remains focusable
```
beaesguerra, line 24: Nice!

packages/wonder-blocks-dropdown/src/components/custom-opener.tsx
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * ## Usage
+ *
+ * ```tsx
+ * import {SingleSelect, CustomOpener} from "@khanacademy/wonder-blocks-dropdown";
+ *
+ * <SingleSelect
+ *   placeholder="Choose an option"
+ *   opener={({hovered, focused, text}) => (
+ *     <CustomOpener style={styles.myOpener}>
```
beaesguerra: This kind of reminds me of Clickable! Is there a reason why CustomOpener should be used instead of Clickable? It might be worth adding that to the docs (and referencing CustomOpener from the Clickable docs too)

packages/wonder-blocks-dropdown/src/components/custom-opener.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+    return (
+        <StyledButton
+            {...restProps}
+            ref={ref as React.Ref<HTMLButtonElement>}
+            type="button"
+            aria-disabled={ariaDisabled}
+            onClick={disabled ? undefined : onClick}
+            data-testid={testId}
+            style={[styles.reset, disabled && styles.disabled, style]}
```
beaesguerra: Is it worth noting that consumers will also need to provide their own disabled styles to the custom opener too? (ie. greying out the opener using semantic color disabled tokens!)

packages/wonder-blocks-dropdown/src/components/custom-opener.tsx
beaesguerra, line 1: Would it make sense to have a StateSheet for CustomOpener? It will be helpful to verify that the focus style is being applied as expected! (And same with other state styles that are provided by consumers)

beaesguerra: And would it make sense to add a `custom-opener.stories.tsx` file too so we have the component and prop docs for CustomOpener?

packages/wonder-blocks-button/src/util/button.types.ts
```diff
@@ -148,6 +148,16 @@ export type BaseButtonProps =
...
+
+        /**
+         * Respond to a raw "blur" event.
+         */
+        onBlur?: (e: React.FocusEvent) => unknown;
+
+        /**
+         * Respond to a raw "focus" event.
+         */
+        onFocus?: (e: React.FocusEvent) => unknown;
```
beaesguerra, line 160: We can add a new changeset file for a minor change in the buttons package since button components support new props!

__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
```diff
@@ -347,42 +347,32 @@ export const Controlled: StoryComponentType = {
...
-                }}
+        opener: ({text}: any) => (
+            <CustomOpener
                 testId="teacher-menu-custom-opener"
-                style={[
-                    styles.customOpener,
-                    focused && styles.focused,
-                    hovered && styles.hovered,
-                    pressed && styles.pressed,
-                ]}
```
beaesguerra, line 381: Would it be helpful for the custom opener examples to show full implementations of how the custom opener should handle all states? (Disabled, hover, pressed, error)

These are the things that the default opener comes with, which is why it is preferred over custom implementations!

beaesguerra (commented):
Thanks Marcy! Left some questions for you!

packages/wonder-blocks-dropdown/src/components/custom-opener.tsx
marcysutton: Clickable enables some bad HTML patterns by allowing wrapping around anything, so I'd like to de-emphasize it in favor of dedicated components. That was my reasoning, anyway!

marcysutton: I'd also like to avoid scope creep by making changes to Clickable prerequisites to adopting this lint rule!

packages/eslint-plugin-wonder-blocks/docs/no-raw-button.md
```diff
@@ -0,0 +1,108 @@
...
+```tsx
+import {IconButton} from "@khanacademy/wonder-blocks-icon-button";
+import plusIcon from "@phosphor-icons/core/regular/plus.svg";
+
+<IconButton
+    icon={plusIcon}
+    aria-label="Add item"
+    onClick={handleClick}
+/>
+```
```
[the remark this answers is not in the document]
marcysutton: This one is funny to me because Claude itself gets the default Button import wrong all the time. Actually, it's probably the same issue. Will fix.

packages/eslint-plugin-wonder-blocks/src/configs/strict.ts
marcysutton: That would be done in a follow-up.

__docs__/wonder-blocks-dropdown/action-menu.stories.tsx
marcysutton: Yes, added some styles back in. I put them on a `styles.root` object so the child `BodyText` can be customized too!

packages/wonder-blocks-dropdown/src/components/custom-opener.tsx
marcysutton: Added!

packages/wonder-blocks-dropdown/src/components/custom-opener.tsx
marcysutton: Good flag! I did end up narrowing the type and props.

packages/wonder-blocks-button/src/util/button.types.ts
marcysutton: Added! I replaced the original changeset with a couple of targeted ones.

beaesguerra: Thanks for updating the changesets! I don't see a changeset for the new button props though!

__docs__/wonder-blocks-dropdown/custom-opener-testing-snapshots.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+import {BodyText} from "@khanacademy/wonder-blocks-typography";
+import {themeModes} from "../../.storybook/modes";
+import {defaultPseudoStates, StateSheet} from "../components/state-sheet";
+
+export default {
+    title: "Packages / Dropdown / Testing / Snapshots / CustomOpener",
+    tags: ["!autodocs", "!manifest"],
+    parameters: {
+        chromatic: {
+            modes: themeModes,
```
beaesguerra: Could you switch this to `allThemeModes`? That will include the new dark theme! (I will be removing `themeModes` soon after landing the dark mode changes!)

beaesguerra (commented):
Looks like the [CustomOpener story](https://5e1bf4b385e3fb0020b7073c-bfrpdofgln.chromatic.com/?path=/story/packages-dropdown-customopener--default) is rendering on 2 lines, is that expected? Might also be related to the new shift in Chromatic snapshots! 

Image

Only noticed a few more minor things:
- the missing changeset for the new button props
- enabling snapshots for all themes 

After that, should be good to go!

packages/wonder-blocks-button/src/util/button.types.ts
marcysutton: Ah, good catch! I missed one in the swirl of changesets. Added it!

__docs__/wonder-blocks-dropdown/custom-opener-testing-snapshots.stories.tsx
marcysutton: Yep, updated!

beaesguerra (approved):
Looks good to me! 🚀

PR #3058, opened 2026-05-19
SYL Dark - Cells

beaesguerra:
## Summary:

Reviewing the CompactCell and DetailCell components in SYL Dark to confirm they look as expected, match the design specs, and have no a11y color violations. We also enable Chromatic snapshots for the syl-dark theme for these components.

There were no changes made to the cell components. Only the stories were updated and snapshots were enabled for the syl-dark theme

Issue: WB-2166
Design: https://www.figma.com/design/04ptAsL2PE4e8KGfYewta3/bea-syl-dark?node-id=5358-29&m=dev

## Test plan:

Review DetailCell and CompactCell and confirm things look as expected and there are no a11y violations
- `?path=/story/packages-cell-testing-snapshots-compactcell--state-sheet-story&globals=theme:syl-dark` 
- `?path=/story/packages-cell-testing-snapshots-detailcell--state-sheet-story&globals=theme:syl-dark`


.changeset/slimy-laws-drop.md
```diff
@@ -0,0 +1,2 @@
+---
```
beaesguerra, line 1: Empty changeset since no changes were made to the components

marcysutton (commented):
Thanks for working on these! My biggest question is about the default state background color for cells. They are the same as the page background. In Figma, the canvas has a slightly different color. Do we know what the background color will be behind the cells, so they have a bit of separation?

beaesguerra:
> Thanks for working on these! My biggest question is about the default state background color for cells. They are the same as the page background. In Figma, the canvas has a slightly different color. Do we know what the background color will be behind the cells, so they have a bit of separation?

Thanks @marcysutton ! Good question, I updated the background for the snapshot stories! Looks like the background setting we had was misconfigured, so now we should be able to differentiate between the component and the background!

marcysutton (approved):
Yay, looks good!

PR #3059, opened 2026-05-20
🚨 [WB-2324.1] Phase 0 tooling for the CSS Modules migration

jandrade:
## Summary:
Wires the foundational PostCSS pipeline, cascade-layer order, Jest mocking,
and a throwaway Storybook spike so Phase 1 can start migrating real
components to CSS Modules without renegotiating infra. No public component
behavior changes; the only published surface change is `wonder-blocks-styles`
gaining wildcard CSS subpath exports.

Issue: WB-2324

## Test plan:

1. `pnpm start` → open Storybook at http://localhost:6061/ → navigate to
   **Tools / CSS Modules Spike** → confirm both stories ("Default",
   "WithBadge") render: rounded button using Wonder Blocks tokens, focus
   ring appears on Tab.
2. Inspect the spike story in DevTools: `.root` should have a hashed
   class name; the rules should be wrapped in `@layer shared`; the
   focus-visible outline should resolve `var(--wb-semanticColor-focus-outer)`.
3. `pnpm jest __docs__/css-modules-spike/spike.test.tsx` — confirms
   `identity-obj-proxy` returns class names as-is for `*.module.css`.
4. `pnpm jest` (no filter) on at least one component package — confirms
   the new module-mapper precedence didn't disturb existing tests.
5. `pnpm typecheck` — clean.

No Rollup CSS emission yet (Phase 1). No published-component behavior
changes to validate in a downstream consumer.

## Review plan:

1. 🚨 [`postcss.config.cjs`](https://github.com/Khan/wonder-blocks/pull/3059#discussion_r3283574456)
2. 🚨 [`vite.config.ts`](https://github.com/Khan/wonder-blocks/pull/3059#discussion_r3283639160)
3. 🚨 [`config/jest/test.config.js`](https://github.com/Khan/wonder-blocks/pull/3059#discussion_r3283639554)
4. 🚨 [`package.json`](https://github.com/Khan/wonder-blocks/pull/3059#discussion_r3283640020)
5. 🚨 [`packages/wonder-blocks-styles/package.json`](https://github.com/Khan/wonder-blocks/pull/3059#discussion_r3283640214)
6. ⚠️ [`.storybook/preview-head.html`](https://github.com/Khan/wonder-blocks/pull/3059#discussion_r3283640446)

### Common patterns:

**4 Files — global tooling additions for the CSS Modules pipeline:**

```ts
// vite.config.ts — alias regex tightened so subpaths fall through
- find: /^@khanacademy\/wonder-blocks(-.*)$/,
+ find: /^@khanacademy\/wonder-blocks(-[^/]*)$/,
```

```js
// config/jest/test.config.js — module.css first, plain .css second
moduleNameMapper: {
    "^@khanacademy/wonder-blocks-(.*)$": "<rootDir>/packages/wonder-blocks-$1/src/index.ts",
+   "\\.module\\.(css|less|scss|sass)$": "identity-obj-proxy",
    "\\.(css|less|scss|sass)$": "<rootDir>/config/jest/css.mock.js",
},
```

```js
// postcss.config.cjs (new) — order matters
plugins: [postcssImport(), postcssMixins()]
```

**1 File — wildcard CSS subpath exports on a published package (precedent in `wonder-blocks-tokens`):**

```jsonc
// packages/wonder-blocks-styles/package.json
+ "exports": {
+   ".": { "import": "./dist/es/index.js", ... },
+   "./*.css": "./src/styles/*.css"
+ }
```


postcss.config.cjs
```diff
@@ -0,0 +1,21 @@
+/**
```
jandrade, line 1: 🚨 **1 of 6:** New root-level PostCSS config picked up by Vite/Storybook for every CSS file. Pulls in `postcss-import` and `@csstools/postcss-mixins` from the workspace; downstream Rollup integration will reuse the same chain in Phase 1. cc: @Khan/frontend-infra-web

vite.config.ts
```diff
@@ -17,8 +17,13 @@ export default defineConfig({
                 ),
             },
             // Allow us to detect changes from local wonder-blocks packages.
+            // Anchored to `$` so subpath imports like
```
jandrade, line 20: 🚨 **2 of 6:** Workspace alias regex tightened from `^@khanacademy/wonder-blocks(-.*)$` to `^@khanacademy/wonder-blocks(-[^/]*)$`. Required so subpath imports like `@khanacademy/wonder-blocks-styles/focus-styles.css` fall through to Node resolution + the workspace symlink instead of being rewritten to `…/src/focus-styles.css`. Anything currently relying on the greedy rewrite for a subpath would break. cc: @Khan/frontend-infra-web

config/jest/test.config.js
```diff
@@ -45,6 +45,10 @@ const projectConfig = {
     moduleNameMapper: {
         "^@khanacademy/wonder-blocks-(.*)$":
             "<rootDir>/packages/wonder-blocks-$1/src/index.ts",
+        // Order matters: the module.css pattern must come before the generic
```
jandrade, line 45: 🚨 **3 of 6:** New `moduleNameMapper` entry maps `*.module.css` to `identity-obj-proxy` *before* the existing no-op `css.mock.js`. Order is load-bearing — both keys match `.module.css` and Jest takes the first match. cc: @Khan/frontend-infra-web

package.json
```diff
@@ -46,6 +46,7 @@
   "license": "MIT",
   "devDependencies": {
     "@changesets/cli": "^2.27.12",
+    "@csstools/postcss-mixins": "^1.0.0",
```
jandrade, line 49: 🚨 **4 of 6:** Adds four new external devDependencies — `postcss`, `postcss-import`, `@csstools/postcss-mixins`, `identity-obj-proxy`. First three are mainstream PostCSS plugins; `identity-obj-proxy` is the documented Jest pattern for CSS Modules. cc: @Khan/frontend-infra-web

packages/wonder-blocks-styles/package.json
```diff
@@ -18,6 +18,17 @@
   "main": "dist/index.js",
   "module": "dist/es/index.js",
   "types": "dist/index.d.ts",
+  "sideEffects": [
```
jandrade, line 21: 🚨 **5 of 6:** Adds `exports` map (with wildcard `./*.css → ./src/styles/*.css`) and `sideEffects: ["**/*.css"]`. Published-surface change: any consumer importing a subpath not previously covered by the implicit resolution will now resolve via the explicit map. Patterns after the `wonder-blocks-tokens` precedent. cc: @Khan/frontend-infra-web

.storybook/preview-head.html
```diff
@@ -14,6 +14,20 @@
 <link rel="stylesheet" href="./sb-styles/vars.css" />
 <link rel="stylesheet" href="./sb-styles/preview.css" />
 <style>
+    /*
```
jandrade, line 17: ⚠️ **6 of 6:** Declares the global cascade-layer order `@layer reset, tokens, shared, overrides;` for every Storybook preview. Authored CSS in `@layer shared` will lose to any unlayered rule, by design — verify existing Aphrodite-generated unlayered rules still win in stories. cc: @Khan/frontend-infra-web

.changeset/css-modules-phase-0-infra.md
beaesguerra, line 1: Would it make sense to put this in the PR description instead of a changeset file, since it isn't related to a changelog for packages to be released?

beaesguerra:
@claude review once

.changeset/quick-modules-spike.md
```diff
@@ -0,0 +1,19 @@
...
+pipeline that includes `postcss-import` and `@csstools/postcss-mixins` can
+write:
+
+```css
+@import "@khanacademy/wonder-blocks-styles/focus-styles.css";
+
+.my-component:focus-visible {
+    @apply --wb-focus-visible;
+}
+```
```
beaesguerra, line 16: Would it make sense to add these kind of docs to Storybook? 

Maybe https://khan.github.io/wonder-blocks/?path=/docs/packages-styles-focus-styles--docs for details on using the focus mixin!

beaesguerra: Out of curiosity, would we remove support for the js focus styles later on once the CSS migration is complete? Or do we plan to support both forms long term?

.changeset/quick-modules-spike.md
```diff
@@ -0,0 +1,19 @@
+---
+"@khanacademy/wonder-blocks-styles": minor
+---
+
+Ship a CSS-Mixins-1 focus mixin at
+`@khanacademy/wonder-blocks-styles/focus-styles.css`. Consumers with a PostCSS
+pipeline that includes `postcss-import` and `@csstools/postcss-mixins` can
```
beaesguerra, line 7: Would it be helpful to add instructions for consumers on setting up a PostCSS pipeline that includes postcss-import and `@csstools/postcss-mixins` in the Getting Started Docs? https://khan.github.io/wonder-blocks/?path=/docs/get-started--docs

.storybook/preview-head.html
```diff
@@ -14,6 +14,20 @@
...
+     * Earlier layers lose to later ones, so authoring order is also
+     * priority order. Anything unlayered wins over everything below.
+     *
+     *   reset     — future normalize / reset rules
+     *   tokens    — CSS custom properties from @khanacademy/wonder-blocks-tokens
+     *   shared    — every Wonder Blocks component (CSS Modules output is
+     *               wrapped in @layer shared by the build pipeline)
+     *   overrides — consumer-owned escape hatch
+     */
+    @layer reset, tokens, shared, overrides;
```
beaesguerra: CSS layers is something I am still learning more about! I have some questions around how we'll be using it: 
- Will consuming apps/Storybooks need to set this too?
  - I wonder if it would make sense for us to export a baseline css file from WB that includes the basic things for using WB (things like the layer order, the root font size, the default color token to use, etc). That way, we can centralize it in one place and keep consumers in sync (in both storybook and application environments). This can be something for later, let me know what you think! 
- Should we document this expected layer order? 
- I was trying to see if we had existing docs around our CSS layers and found this blog post: [Aphrodite important removed using layers in our LESS](https://khanacademy.atlassian.net/wiki/spaces/FRONTEND/blog/2024/07/31/3126919203/Aphrodite+important+removed+using+layers+in+our+LESS). What would the order look like with the `legacy` layer? Maybe it makes sense for applications to define their own layers!
- Based on the blog post, it looks like Perseus styles will be in the `shared` layer too. Could there be conflicting specificity if Perseus needs to override WB styles? Would it make sense to have a separate layer for wonder blocks?

__docs__/css-modules-spike/spike.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+        // lets className equality checks work in tests without running
+        // the real PostCSS pipeline.
+        expect(styles.root).toBe("root");
+        expect(styles.pill).toBe("pill");
+    });
+
+    it("renders the mocked class names onto the DOM", () => {
+        render(<Spike label="Spike button" badge="NEW" />);
+
+        expect(screen.getByRole("button")).toHaveClass("root");
```
beaesguerra, line 20: Interesting! Would this mean that jest snapshots in frontend won't need to be updated whenever we change styles for a wb component? Since the hash would no longer get included when jest runs?

__docs__/css-modules-spike/spike.module.css
```diff
@@ -0,0 +1,39 @@
...
+        background: var(--wb-semanticColor-core-background-base-default);
+        color: var(--wb-semanticColor-core-foreground-neutral-strong);
+        font-family: inherit;
+        font-size: var(--wb-sizing-size_160);
+        cursor: pointer;
+    }
+
+    .root:focus-visible {
+        @apply --wb-focus-visible;
+    }
```
beaesguerra, line 30: Neat! 

Can the css mixin include setting it on the `:focus-visible` pseudo-class? That way, it can be used consistently with `:focus-visible` rather than `:focus` (in most cases). This would be similar to how `focusStyles` includes the `:focus-visible` part!

__docs__/css-modules-spike/spike.module.css
```diff
@@ -0,0 +1,39 @@
...
+
+    .pill {
+        padding: 0 var(--wb-sizing-size_060);
+        border-radius: 999px;
```
beaesguerra: I wonder why Claude picked a 999px border radius 😅

__docs__/css-modules-spike/spike.module.css
```diff
@@ -0,0 +1,39 @@
...
+ * Phase 0 spike — proves the PostCSS pipeline works end-to-end:
+ *   - cross-package @import resolved through the workspace
+ *   - @apply expands a mixin defined in a sibling package at build time
+ *   - CSS Modules hashing yields stable, locally-scoped class names
+ *   - declared inside @layer shared so consumer .overrides can win
+ *   - token CSS variables resolve at runtime
+ */
+@import "@khanacademy/wonder-blocks-styles/focus-styles.css";
+
+@layer shared {
```
beaesguerra, line 11: Another question about CSS layers in general! Will we have tools/stylelint in place to encourage consumers to use the correct layers? For example, I'm assuming we don't want consumers modifying tokens or shared layers!

beaesguerra (approved):
Exciting! Overall looks good to me! I mostly had questions to better understand how we'll work with CSS modules and layers, and some suggestions for docs 😄

.changeset/quick-modules-spike.md
jandrade: That's a good observation! In a few PRs, I'm planning to create a shared PostCSS pipeline that can be reused by consumers as well. This shared package will probably also include some common Stylelint rules that apply to consumers as well.

I'll add docs soon as I have a better definition of this new shared package.

.changeset/quick-modules-spike.md
jandrade: re: docs -> I'm planning to iterate on this component later this week with this ticket: https://khanacademy.atlassian.net/browse/WB-2327

re: JS support -> Good question! for now, I'd say we'd need to keep both until we figure out what will remain from Classic + devadmin, but ideally, we should only support the CSS version at some point.

.storybook/preview-head.html
jandrade: oh thanks for pointing this out! so, I need to change the layer order here as there are things that don't really apply. We are going to match the layers in the document you shared (that is what we define in `frontend`), and that means that we'll only need to focus on the shared layer here, that is what our consumers expect to have WB styles defined.

__docs__/css-modules-spike/spike.module.css
jandrade: Consumers should not need to use/define any layers at all, that way any style defined at the consumer level will take more precedence than layered styles (e.g. wb, perseus styles).

__docs__/css-modules-spike/spike.module.css
jandrade: hmmm I remember changing this for a WB border token at some point 😢. Claude picked it again!

__docs__/css-modules-spike/spike.test.tsx
jandrade: great question! I'm not fully sure how this will work in frontend, but it will definitively depend on how we bundle the JS part of it and how it interoperates with the bundled CSS. I'll run some tests to verify how this would work.

__docs__/css-modules-spike/spike.module.css
jandrade: good question! It can definitively do it, but as there are some cases where we use this mixin outside of `:focus-visible`, then it will make it impossible to extract it from that key.

For example, how do we allow this?

```JS
 ":focus-visible:not([disabled])": focusStyles.focus[":focus-visible"],
```

In CSS, it would need to be something like:

```CSS
:focus-visible:not([disabled]) {
    @apply --wb-focus-visible()[":focus-visible"]; // ❌ INVALID SYNTAX
}
```

jandrade: maybe we can create two separate mixins to handle each case? wdyt?

For either option, I'll iterate on it in the upcoming `wb-styles` ticket.

beaesguerra: Sounds good, 2 separate mixins make sense! Maybe `wb-focus-styling` and `wb-apply-focus-styling` or something like that! 

Are we following a convention where mixin names should be prefixed with `--`, like css variable names?

packages/wonder-blocks-styles/package.json
```diff
@@ -18,6 +18,17 @@
...
+    "**/*.css"
+  ],
+  "exports": {
+    ".": {
+      "import": "./dist/es/index.js",
+      "require": "./dist/index.js",
+      "types": "./dist/index.d.ts"
+    },
+    "./*.css": "./src/styles/*.css"
+  },
```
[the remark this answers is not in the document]
jandrade: went with (b)

config/jest/test.config.js
```diff
@@ -45,6 +45,10 @@ const projectConfig = {
...
+        // CSS mock so identity-obj-proxy handles `styles.foo` lookups, while
+        // plain `*.css` imports keep using the no-op mock.
```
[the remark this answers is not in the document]
jandrade: fixed

jeresig (approved):
Looks reasonable to me, thank you!

__docs__/css-modules-spike/spike.module.css
jandrade: for the convention part, this is just something the community is pushing.

https://github.com/w3c/csswg-drafts/issues/9350#issuecomment-3079389971

TLTR: mostly used to avoid compatibility issues and follow a similar pattern for css variables/properties.

As for our internal convention, I was thinking to use another prefix, so we can separate this from our tokens: `--wb-styles-`

For example we could use it this way: `--wb-styles-focus` and `--wb-styles-focus-visible` (names still TBD), then in the near future we could follow with `--wb-styles-action-inverse` -> http://localhost:6061/?path=/docs/packages-styles-action-styles--docs.

PR #3061, opened 2026-05-21
Fix arrow key navigation from search field in filterable dropdowns

jandrade:
## Summary
Fixed arrow key navigation in filterable dropdowns when the search field has initial focus. Previously, pressing ArrowUp/ArrowDown while focused on the search field would not navigate to dropdown items.

## Changes
- **`focusPreviousItem()`**: Refactored to handle search field focus as the first condition. When the search field is focused, ArrowUp now moves focus to the last item. When at the first item, focus moves to the search field (if it exists) or wraps to the last item.
- **`focusNextItem()`**: Refactored similarly. When the search field is focused, ArrowDown now moves focus to the first item. When at the last item, focus moves to the search field (if it exists) or wraps to the first item.
- **Tests**: Added two new test cases:
  - Navigation from search field to first item with ArrowDown
  - Navigation from search field to last item with ArrowUp

## Implementation Details
The fix simplifies the navigation logic by checking `isSearchFieldFocused()` first, making the behavior more intuitive: the search field acts as a navigation point in the focus cycle rather than a barrier. The `enableTypeAhead` condition was removed as it was preventing proper navigation when the search field had focus.

https://claude.ai/code/session_019bAvtFYM2gG9E9yWXQymCm

### DIFF
--- a/.changeset/fix-filterable-dropdown-navigation.md
+++ b/.changeset/fix-filterable-dropdown-navigation.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-dropdown": patch
+---
+
+Fix arrow key navigation from search field in filterable dropdowns. When `isFilterable` is `true` and the dropdown opens with focus on the search field, pressing ArrowUp/ArrowDown now correctly moves focus to the dropdown items.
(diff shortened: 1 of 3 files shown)


marcysutton:
@claude review once

marcysutton (approved):
This looks good to me! It works in Windows screen readers too.

I'm not wholly familiar with `enableTypeahead` in `SingleSelect` to know how the code was supposed to work before, but it seems to be okay after this change! TY!

PR #3063, opened 2026-05-21
⚠️ [WB-2324.2] Phase 0.3 — Type generation + IDE autocomplete for CSS Modules

jandrade:
## Summary:
Adds the static-typing half of the CSS Modules migration tooling. `tsc --noEmit` now catches `styles.<missingClassName>` references, and VS Code / WebStorm get live autocomplete + Go-to-Definition on `*.module.css` imports.

Builds on top of `WB-2324` (PR #3059); no Rollup or runtime behavior changes.

Issue: WB-2324

## Test plan:

1. `pnpm typecheck` — passes; `prebuild`/`pretypecheck` regenerates `.d.ts` files first.
2. Introduce `styles.tpyo` in `__docs__/css-modules-spike/spike.tsx` → `pnpm typecheck` fails with `Property 'tpyo' does not exist on type '{ readonly pill: string; readonly root: string; }'. Did you mean 'root'?`.
3. Revert the typo → `pnpm typecheck` passes again.
4. `pnpm build` — runs `gen:css-types` first via the `prebuild` hook, then Rollup; build succeeds.
5. Open `__docs__/css-modules-spike/spike.tsx` in VS Code with the recommended extensions installed: `styles.` autocomplete shows `pill` and `root`; Go-to-Definition jumps to the matching CSS rule.

## Review plan:

1. ⚠️ [`package.json`](https://github.com/Khan/wonder-blocks/pull/3063#discussion_r3284623570)
2. ⚠️ [`tsconfig.json`](https://github.com/Khan/wonder-blocks/pull/3063#discussion_r3284623757)
3. 🔷 [`build-settings/gen-css-types.cjs`](https://github.com/Khan/wonder-blocks/pull/3063#discussion_r3284623889)

### Common patterns:

**2 Files — pre-script hooks ensuring `*.module.css.d.ts` exists before tsc / Rollup run:**

```jsonc
// package.json
+ "prebuild": "pnpm gen:css-types",
+ "gen:css-types": "node ./build-settings/gen-css-types.cjs",
...
+ "pretypecheck": "pnpm gen:css-types",
```

**1 File — gitignore the generated declarations (regenerated on every build):**

```
# Generated by `pnpm gen:css-types` (typed-css-modules) via the `prebuild` /
# `pretypecheck` hooks. Never checked in — fresh checkouts regenerate them.
**/*.module.css.d.ts
```


package.json
```diff
@@ -11,6 +11,8 @@
     "alex": "alex packages/",
     "dev": "pnpm build && turbo watch build:css",
     "build": "rollup -c ./build-settings/rollup.config.mjs && pnpm -r build && turbo build:css",
+    "prebuild": "pnpm gen:css-types",
```
jandrade, line 14: ⚠️ **1 of 3:** Adds `typed-css-modules` + `typescript-plugin-css-modules` devDeps and wires `gen:css-types` into both `prebuild` and `pretypecheck` (line 43). Hooks run for every `pnpm build` and every `pnpm typecheck`, so a regression in `build-settings/gen-css-types.cjs` would break CI. cc: @Khan/wonder-blocks

tsconfig.json
```diff
@@ -8,6 +8,11 @@
 
         "paths": {
             "@khanacademy/*": ["./packages/*/src"]
-        }
+        },
+
+        // Provides live IDE autocomplete and "Go to Definition" for
+        // `*.module.css` imports. Build-time type safety still comes from
+        // the generated `*.module.css.d.ts` files (see `gen:css-types`).
+        "plugins": [{"name": "typescript-plugin-css-modules"}]
```
jandrade, line 16: ⚠️ **2 of 3:** Adds the `typescript-plugin-css-modules` TS language-service plugin. Doesn't affect `tsc` output (plugins only influence the language service), but everyone using a TS-aware editor on this repo will start seeing the plugin's behavior — live autocomplete + Go-to-Definition for `styles.<className>`. cc: @Khan/wonder-blocks

build-settings/gen-css-types.cjs
```diff
@@ -0,0 +1,48 @@
+// Generate `*.module.css.d.ts` files next to every `*.module.css` source.
```
jandrade, line 1: 🔷 **3 of 3:** Root-level Node script that globs `packages/**/*.module.css` + `__docs__/**/*.module.css` (ignoring `node_modules`/`dist`) and invokes `typed-css-modules` to emit colocated `.d.ts` declarations. Camel-casing left disabled (`camelCase: false`) until 0.4 picks the project-wide naming convention — the spike's class names already happen to be camelCase, so this is a no-op for now. cc: @Khan/wonder-blocks

beaesguerra:
@claude review once

tsconfig.json
beaesguerra: Would it make sense to include instructions for setting up this ts plugin in the getting started docs? https://khan.github.io/wonder-blocks/?path=/docs/get-started--docs 

Or maybe somewhere else, since it isn't specific to WB?

beaesguerra (commented):
Nice! Confirmed type checking catches if I try using a style that doesn't exist: 

Image

---

~~In terms of DX, do you know if there's a way to make it so if I `cmd + click` on the import file name (ie. `./spike.module.css`), it'll take me to the css file, rather than the type declaration file? Before (in the WB-2324 branch), it gave me an option to choose the `.css` file. After, it'll take me to the `spike.module.css.d.ts` file directly.~~

| Before | After | 
| --- | --- |
| Image | Image |

It would be nice if `cmd + click` on the `styles` variable took the user to the `.css` file too!  

Update: Nvm! I had to install the recommended extensions and reload 😄 `cmd+click` on the file name gives me the option to go to the CSS file. `cmd+click` on the classname (the `root` part of `styles.root`), takes me to the css class definition, though `cmd+click` on the `styles` part takes me to the `assets.d.ts` file instead!

beaesguerra:
Another DX thing I noticed, what's the best way to identify what uses a `*.module.css` file? (besides searching for the filename, since the filename might not be unique?) 

With React components, I'm able to use the "Find File References" option. This option isn't available for the `.css` file though! 

| tsx file | css file |
| --- | --- |
| image |image |

build-settings/gen-css-types.cjs
```diff
@@ -0,0 +1,48 @@
...
+    if (files.length === 0) {
+        // eslint-disable-next-line no-console
+        console.log("ℹ️  No `*.module.css` files found; nothing to generate.");
+        return;
+    }
+
+    for (const file of files) {
+        const content = await creator.create(file);
+        await content.writeFile();
+    }
```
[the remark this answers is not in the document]
jandrade: @claude fix this

jeresig (commented):
I think I'm not understanding the changes in this PR. It seems like `typescript-plugin-css-modules` allows us to dynamically compute the types in Typescript - but then how are the generated type files used? They aren't checked in... are this distributed as part of the package? I thought we weren't exporting any CSS modules directly?

jandrade:
> I think I'm not understanding the changes in this PR. It seems like `typescript-plugin-css-modules` allows us to dynamically compute the types in Typescript - but then how are the generated type files used? They aren't checked in... are this distributed as part of the package? I thought we weren't exporting any CSS modules directly?

Good questions @jeresig! This is more for internal WB consumption as we are not planning to distribute these typedefs as part of the package.

- `typescript-plugin-css-modules` is included for integration with IDEs, so it allows autocompleting valid styles.
- `typed-css-modules` is the CLI that generate the type files so we can enforce that TS errors are caught during build-time, so we avoid generating component bundles where an invalid style could be introduced at build-time. This package is the one we use on `prebuild` and `pretypecheck` scripts 

My plan is to implement something similar in consumer repos in a way that folks can also use these tools to enforce that we are using valid css modules. I'm going to try to see if we can use a tool in `frontend` that doesn't create `.d.ts` files on disk, but open to any suggestions from your side.

jeresig (approved):
@jandrade Gotcha, ok that's helpful! So I think it may be worth confirming that in frontend when doing a build using Rspack that it doesn't error out anyway, as we may already have something that's able to catch failures. But it's possible that this would be "too late" and we'd want to catch it during typechecking anyway, as you noted.

PR #3064, opened 2026-05-21
Fix NVDA double-announcing modal content by removing role=log from live regions

maddy531:
## Summary

NVDA was announcing the close button and dialog heading twice on modal open, then reading the entire table inside.

WB's announcer injects 4 live regions inside role="dialog" aria-modal="true" elements on mount. These regions had both role="log" and an explicit aria-live attribute. Per [WAI-ARIA 1.2 §log](https://www.w3.org/TR/wai-aria-1.2/#log) and [WCAG ARIA23](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA23), role="log" carries an implicit aria-live="polite". On the assertive regions this is directly contradictory (role="log" → polite, aria-live="assertive" → interrupt). NVDA processes these as two separate live region signals and re-reads the dialog; JAWS and VoiceOver normalize them into one before announcing.

 Fix: 
 Remove role="log" from [createRegion](https://github.com/Khan/wonder-blocks/pull/3064/changes#diff-e9d4d52aff8fee6949bd08141400d6110027183111adfdabc336c05509405b9bL43) in [dom.ts](https://github.com/Khan/wonder-blocks/pull/3064/changes#diff-e9d4d52aff8fee6949bd08141400d6110027183111adfdabc336c05509405b9b)[dom.ts](https://github.com/Khan/wonder-blocks/pull/packages/wonder-blocks-announcer/src/util/dom.ts). Plain aria-live is sufficient.

Issue: CLASS-14343

## Test plan:

### DIFF
--- a/.changeset/happy-bobcats-tickle.md
+++ b/.changeset/happy-bobcats-tickle.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/wonder-blocks-announcer": patch
+"@khanacademy/wonder-blocks-modal": patch
+---
+
+Fix NVDA double-announcing modal content by omitting role="log" on live region elements
--- a/packages/wonder-blocks-announcer/src/__tests__/announce-message.test.tsx
+++ b/packages/wonder-blocks-announcer/src/__tests__/announce-message.test.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import {render, screen, waitFor} from "@testing-library/react";
+import {render, screen, waitFor, within} from "@testing-library/react";
 import Announcer, {REMOVAL_TIMEOUT_DELAY} from "../announcer";
 import {AnnounceMessageButton} from "./components/announce-message-button";
 import {announceMessage} from "../announce-message";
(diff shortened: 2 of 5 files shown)


marcysutton:
Thanks for the PR @maddy531! I don't actually think the double announcements are coming from WB Announcer. There is a Live Region Announcer in the modal but nothing is being appended to it as far I can tell. The Announcer exists in the modal _in case_ anything gets passed to `announceMessage`, but I'm not seeing any activity in those elements. If I comment out the modal layer Announcer hook, the double announcements still occur in NVDA. So it's definitely not the cause, even if there are redundant attributes.

I think we need to get to the bottom of the issue before making changes to the core component.

marcysutton:
I'm going to close this PR since it doesn't solve the intended issue. I verified this by testing in NVDA on Windows:
- The Announcer in Modal example still has this issue on the Chromatic site, with the change in this PR (meaning it didn't fix it): https://5e1bf4b385e3fb0020b7073c-fzyggurbme.chromatic.com/?path=/story/packages-announcer--announcer-in-modal
- The Modal examples themselves have this issue as well: https://5e1bf4b385e3fb0020b7073c-fzyggurbme.chromatic.com/?path=/story/packages-modal-modallauncher--default
    - I also tested this locally by commenting out the `useModalAnnouncer` layer hook, and had the same result.

The one condition I found was different in NVDA: if you send focus to a text input in a modal instead of a close button, it short-cuts the double announcements. But that's pretty situational and dependent on modal content.

I'll keep trying to track down the issue in our modals separately, since we do want to fix it for the close button pattern.

Thanks for the contribution!

PR #3066, opened 2026-05-21
Link: Accept the `state` prop from the React router Link component.

MikeKlemarewski:
## Summary:
Sometimes when we navigate, we want to set `state` on the React router navigation. We can do this with some hooks provided by React router, but if we are trying to use a Link to do the navigation then this gets tricky. React router's [Link component](https://reactrouter.com/api/components/Link#state) accempts a `state` prop. This exposes that same prop on our own Link component.

Issue: TUT-4070

## Test plan:
- Ensure tests pass

### DIFF
--- a/.changeset/twenty-worms-shake.md
+++ b/.changeset/twenty-worms-shake.md
@@ -1,0 +1,6 @@
+---
+"@khanacademy/wonder-blocks-clickable": minor
+"@khanacademy/wonder-blocks-link": minor
+---
+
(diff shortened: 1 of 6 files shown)


marcysutton (approved):
Thank you Mike! This looks good to me, but you might want to test the integration before landing it. You've probably done this before but here are some instructions just in case: https://github.com/Khan/wonder-blocks/pull/3066#issuecomment-4513454299

PR #3053, opened 2026-05-14
New WB lint rule: no-hardcoded-color

marcysutton:
## Summary:
Add a new lint rule to warn about hardcoded color values in RGB, RGBA, keyword, hsl, currentColor, hex, and more. This rule warns about hardcoded colors and warns to use semanticColor instead.

Note: this PR also disables the `require-logical-properties-for-rtl` rule that was rolled out without a timely integration, so other work in `frontend` could be unblocked. The rule is still available for opt-in usage, just not part of the recommended (automatic) lint rule config.

## Example lint rule failures

We have some fixes to make in WB itself for this rule to pass! (a sample of results below -- shortened for brevity)
```
 pnpm lint

> wonder-blocks@0.0.1 lint /Users/marcytodd/khan/wonder-blocks
> pnpm lint:ci .


> wonder-blocks@0.0.1 lint:ci /Users/marcytodd/khan/wonder-blocks
> eslint --ext .ts --ext .js --ext .tsx --ext .jsx "."


/Users/marcytodd/khan/wonder-blocks/__docs__/components/component-info.tsx
  41:32  error  Avoid hardcoded color value 'black'. Use a semantic color token from @khanacademy/wonder-blocks-tokens instead (e.g. semanticColor.core.foreground.primary) to support theming and dark mode  @khanacademy/wonder-blocks/no-hardcoded-color

/Users/marcytodd/khan/wonder-blocks/__docs__/wonder-blocks-dropdown/multi-select.stories.tsx
  116:17  error  Avoid hardcoded color value '1px solid grey'. Use a semantic color token from @khanacademy/wonder-blocks-tokens instead (e.g. semanticColor.core.foreground.primary) to support theming and dark mode  @khanacademy/wonder-blocks/no-hardcoded-color

/Users/marcytodd/khan/wonder-blocks/__docs__/wonder-blocks-form/checkbox-group.stories.tsx
  344:20  error  Avoid hardcoded color value 'solid 1px #CCC'. Use a semantic color token from @khanacademy/wonder-blocks-tokens instead (e.g. semanticColor.core.foreground.primary) to support theming and dark mode  @khanacademy/wonder-blocks/no-hardcoded-color
  354:23  error  Avoid hardcoded color value 'solid 1px #CCC'. Use a semantic color token from @khanacademy/wonder-blocks-tokens instead (e.g. semanticColor.core.foreground.primary) to support theming and dark mode  @khanacademy/wonder-blocks/no-hardcoded-color
```

Issue: WB-2339

[...]

PR #3069, opened 2026-05-22
🚨 [WB-2324.3] Phase 0.4 — Stylelint with token-strict-value + CI wiring

jandrade:
## Summary:
Adds CSS-side hygiene to the migration. `pnpm lint:css` is now an explicit CI step alongside ESLint and Typecheck. Highlights:

- **At-rule allow-list** lets `@apply`, `@mixin`, `@define-mixin`, `@include`, `@layer` pass through (matches `postcss.config.cjs`).
- **`selector-class-pattern`** enforces camelCase — `.root`, `.iconWrapper`, etc.
- **Token-required properties** (color / spacing / typography) raise a *warning* when authored as raw values; tightens to error once Phase 5 finishes.
- **`csstools/value-no-unknown-custom-properties`** rejects any `var(--foo)` that doesn't resolve to a token published by `@khanacademy/wonder-blocks-tokens` (reads `packages/wonder-blocks-tokens/dist/css/index.css`).
- **`prelint:css`** runs `turbo build:css` so the merged token CSS is fresh before stylelint reads it (CI already builds first; locally turbo's cache makes warm runs near-instant).

Builds on top of WB-2324.3 (PR #3063); no runtime behavior changes.

Issue: WB-2324

## Test plan:

1. `pnpm lint:css` — passes (0 errors, 1 pre-existing warning on `__docs__/wonder-blocks-typography/styles.css` for a hardcoded `color: red`).
2. Add a bogus at-rule body to `spike.module.css` (e.g. `@whatever-bogus-rule { color: red; }`): `pnpm lint:css` fails with `Unknown at-rule "@whatever-bogus-rule"` and warns about the raw color.
3. Change a `var(--wb-sizing-size_080)` reference to `var(--wb-totally-bogus-token)` in `spike.module.css`: `pnpm lint:css` fails with `Unexpected custom property "--wb-totally-bogus-token"`.
4. Rename a CSS Modules class to kebab-case (e.g. `.pill-badge`): `pnpm lint:css` fails with `Expected ".pill-badge" to match pattern "^[a-z]+([A-Z][a-z0-9]+)*$"`.
5. Revert all of the above → clean.
6. Watch CI on this branch: the new `Stylelint` step runs after `Eslint` and before `Typecheck` in `node-ci-lint.yml`.

## Review plan:

1. 🚨 [`.github/workflows/node-ci-lint.yml`](https://github.com/Khan/wonder-blocks/pull/3069#discussion_r3290160590)
2. ⚠️ [`.stylelintrc.cjs`](https://github.com/Khan/wonder-blocks/pull/3069#discussion_r3290242637)
3. ⚠️ [`package.json`](https://github.com/Khan/wonder-blocks/pull/3069#discussion_r3290242856)

### Common patterns:

**1 File — declarative at-rule allow-list paired with the PostCSS pipeline:**

```js
// .stylelintrc.cjs — must match plugin order in postcss.config.cjs
"at-rule-no-unknown": [
    true,
    {
        ignoreAtRules: [
            "apply", "mixin", "define-mixin", "include", "layer",
        ],
    },
],
```

**1 File — token-anchored `var()` validation:**

```js
// .stylelintrc.cjs
"csstools/value-no-unknown-custom-properties": [
    true,
    {
        importFrom: [
            "./packages/wonder-blocks-tokens/dist/css/index.css",
        ],
    },
],
```

**1 File — CI job step that mirrors the existing Eslint / Typecheck shape:**

```yaml
# .github/workflows/node-ci-lint.yml
- name: Stylelint
  if: always()
  run: pnpm lint:css
```


.github/workflows/node-ci-lint.yml
```diff
@@ -69,6 +69,10 @@ jobs:
                   limited-trigger: ${{ steps.js-ts-files.outputs.filtered }}
                   limited: pnpm lint:ci {}
 
+            - name: Stylelint
```
jandrade, line 72: 🚨 **1 of 4:** Adds a new `Stylelint` job step (`pnpm lint:css`) between `Eslint` and `Typecheck`. Runs on every PR; a regression in the stylelint config or a new false-positive rule would block unrelated work. `if: always()` so this step doesn't short-circuit on an upstream failure. The earlier `pnpm build` step on line 58 guarantees `packages/wonder-blocks-tokens/dist/css/index.css` is fresh by the time `csstools/value-no-unknown-custom-properties` reads it. cc: @Khan/frontend-infra-web

stylelint.config.ts
```diff
@@ -0,0 +1,92 @@
...
+//   * Catch CSS-side typos / structural issues that don't show up in JSX.
+//   * Enforce design-token usage for color/spacing/typography.
+//   * Co-exist with the PostCSS pipeline declared in `postcss.config.cjs`
+//     (postcss-import + @csstools/postcss-mixins).
+//
+// Phase 0 lands with permissive defaults; severity tightens to `error`
+// once Phase 5 migrates the last component (per WB-2324 0.4 plan).
+/** @type {import("stylelint").Config} */
+module.exports = {
+    extends: ["stylelint-config-standard", "stylelint-config-css-modules"],
```
jandrade, line 19: ⚠️ **2 of 4:** Full stylelint config for the migration. Extends `stylelint-config-standard` + `stylelint-config-css-modules`. Key choices:

* `at-rule-no-unknown` allow-lists `@apply`, `@mixin`, `@define-mixin`, `@include`, `@layer` so the PostCSS pipeline's at-rules pass — must stay in sync with `postcss.config.cjs`.
* `at-rule-disallowed-list: ['value']` — forbid CSS Modules `@value`; tokens from `@khanacademy/wonder-blocks-tokens` are the only sanctioned shared-value mechanism.
* `property-disallowed-list: ['composes']` — forbid CSS Modules `composes`; share styles via `@apply` mixins from `@khanacademy/wonder-blocks-styles` or className composition in JSX instead.
* `selector-class-pattern: ^[a-z]+([A-Z][a-z0-9]+)*$` enforces camelCase (matches the css-modules-v1 precedent and lets `styles.foo` work without Vite's `localsConvention` conversion).
* `custom-property-pattern: null` — WB tokens deliberately use `--wb-camelCase_underscore` and aren't kebab-case.
* `scale-unlimited/declaration-strict-value` at **warning** level for color/spacing/typography; escalates to error in Phase 5.
* `csstools/value-no-unknown-custom-properties` (error) reads the merged token CSS at `packages/wonder-blocks-tokens/dist/css/index.css` and rejects any `var(--foo)` that doesn't resolve to a published token.

**Follow-up (commit 1496477):** the `importFrom` path is now resolved via `path.resolve(__dirname, …)` so the rule finds the token CSS regardless of the cwd stylelint is invoked from — root `pnpm lint:css`, package-scoped runs, and the `vscode-stylelint` extension (which uses each `.css` file's own directory as cwd) all see the same absolute path.

**Follow-up (commit 69b6590):** two more guardrails:
* `max-nesting-depth: [2, {ignoreAtRules: ['layer', 'media', 'supports']}]` — keeps CSS Modules shallow. One level of `&`-nested state or descendant selectors is allowed inside a class (`.root { &:hover { … } }`); deeper Sass-style chains are rejected. The conventional `@layer shared { .root { … } }` wrapper is ignored so it doesn't trip the rule.
* `color-no-hex` (warning) — forbids hex literals (`#fff`, `#112233`) anywhere. Closes the side-channel `declaration-strict-value` doesn't cover: a hex on a property outside its list (`box-shadow`, `outline`, `fill`) would otherwise still slip through. Tightens to `error` in Phase 5.

cc: @Khan/frontend-infra-web

package.json
```diff
@@ -30,6 +30,8 @@
     "lint:watch": "esw --watch --config ./eslint/eslintrc packages",
     "lint": "pnpm lint:ci .",
     "lint:ci": "eslint --ext .ts --ext .js --ext .tsx --ext .jsx",
+    "lint:css": "stylelint '**/*.css'",
```
jandrade, line 33: ⚠️ **3 of 4:** Adds 5 external devDeps — `stylelint`, `stylelint-config-standard`, `stylelint-config-css-modules`, `stylelint-declaration-strict-value`, `stylelint-value-no-unknown-custom-properties` — plus two scripts. `lint:css` (line 33) is the new CI/dev command; `prelint:css` (line 34) runs `turbo build:css` so the token CSS at `packages/wonder-blocks-tokens/dist/css/index.css` is fresh before stylelint reads it. CI already runs `pnpm build` first so the hook is essentially a no-op there; locally turbo's cache makes warm runs near-instant. cc: @Khan/frontend-infra-web

.vscode/settings.json
```diff
@@ -23,5 +23,15 @@
...
     "js/ts.inlayHints.variableTypes.enabled": true,
 
-    "jest.jestCommandLine": "pnpm jest"
+    "jest.jestCommandLine": "pnpm jest",
+    // CSS Variable Autocomplete (`phoenisx.cssvar`). Points at the merged
+    // token CSS so authors get autocomplete + hover docs for every
+    // `--wb-*` custom property defined by `@khanacademy/wonder-blocks-tokens`.
+    // The file is generated by `pnpm build:css` (run automatically as
+    // part of `pnpm build` / `pnpm lint:css`); on a fresh checkout, run
+    // `pnpm build` once before expecting the popup.
```
jandrade: 🔷 **4 of 4:** IDE polish for the migration — wires up `phoenisx.cssvar` so authors get `var(--wb-…)` autocomplete + hover docs across `.css`, `.ts`, and `.tsx`. `cssvar.files` points at the merged token CSS at `packages/wonder-blocks-tokens/dist/css/index.css` (the same file stylelint reads), so a single source of truth drives both the lint error and the editor suggestion. The recommendation lives in `.vscode/extensions.json` alongside `stylelint.vscode-stylelint` for a one-prompt onboarding. Caveat: the popup is silent until `pnpm build` runs once after a fresh checkout — the token CSS is generated, not checked in. cc: @Khan/frontend-infra-web

stylelint.config.ts
```diff
@@ -0,0 +1,134 @@
...
+                    "layer",
+                ],
+            },
+        ],
+
+        // Forbid CSS Modules `@value`. Tokens from
+        // `@khanacademy/wonder-blocks-tokens` are the single source of
+        // truth for shared values — `@value` would create a second,
+        // parallel variable system that bypasses the design system.
+        "at-rule-disallowed-list": ["value"],
```
jandrade, line 48: Image

stylelint.config.ts
```diff
@@ -0,0 +1,134 @@
...
+
+        // Forbid CSS Modules `composes`. Compositional inheritance
+        // between modules is hard to trace at scale and conflicts with
+        // our `@apply` mixin pattern from
+        // `@khanacademy/wonder-blocks-styles`. Authors should compose
+        // class names in JS or share styles via mixins instead.
+        "property-disallowed-list": ["composes"],
```
jandrade, line 55: Image

stylelint.config.ts
```diff
@@ -0,0 +1,134 @@
...
+        ],
+
+        // Forbid hex literals (`#fff`, `#112233`). Pairs with
+        // `declaration-strict-value` above by closing the side-channel:
+        // a hex used in a property *not* on that list (e.g. `box-shadow`,
+        // `outline`, `fill`) would otherwise still slip through. Authors
+        // should reach for a `--wb-*` token via `var()` instead. WARN
+        // during the migration; tightens to `error` in Phase 5 alongside
+        // `declaration-strict-value`.
+        "color-no-hex": [true, {severity: "warning"}],
```
jandrade, line 186: Image

beaesguerra:
@claude review once

.changeset/css-modules-stylelint.md
```diff
@@ -0,0 +1,18 @@
+---
+---
+
+Phase 0.4 of the CSS Modules migration (WB-2324): Stylelint with token-strict
+values, PostCSS at-rule allow-list, and CI wiring.
+
+- Root `.stylelintrc.cjs` extending `stylelint-config-standard` +
```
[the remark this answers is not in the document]
beaesguerra: Maybe these details are better suited for the PR description?

__docs__/css-modules-spike/spike.module.css
```diff
@@ -6,18 +6,16 @@
  *   - declared inside @layer shared so consumer .overrides can win
  *   - token CSS variables resolve at runtime
  */
-@import "@khanacademy/wonder-blocks-styles/focus-styles.css";
+@import url("@khanacademy/wonder-blocks-styles/focus-styles.css");
```
beaesguerra: Interesting! Was there a specific reason why we needed to wrap the import in `url()`?

__docs__/wonder-blocks-typography/styles.css
```diff
@@ -1,6 +1,7 @@
-.custom-style {
-    border: 1px solid #000;
+.customStyle {
+    border: 1px solid var(--wb-semanticColor-core-border-neutral-strong);
```
beaesguerra: Should it have caught the hardcoded 1px border width?

stylelint.config.ts
```diff
@@ -0,0 +1,140 @@
+// Stylelint config for the CSS Modules migration (Phase 0.4 of WB-2324).
+//
+// Goals:
+//   * Catch CSS-side typos / structural issues that don't show up in JSX.
+//   * Enforce design-token usage for color/spacing/typography.
+//   * Co-exist with the PostCSS pipeline declared in `postcss.config.cjs`
+//     (postcss-import + @csstools/postcss-mixins).
+//
+// Phase 0 lands with permissive defaults; severity tightens to `error`
+// once Phase 5 migrates the last component (per WB-2324 0.4 plan).
```
beaesguerra, line 10: Do we want to keep the comments around the implementation/migration plan? The goals part is helpful though!

stylelint.config.ts
beaesguerra: Nice! So many helpful lint rules!

stylelint.config.ts
```diff
@@ -0,0 +1,140 @@
...
+
+        // Wonder Blocks tokens use a camelCase + underscore convention
+        // (`--wb-semanticColor-core-background-base`,
+        // `--wb-sizing-size_080`) that intentionally doesn't conform to
+        // stylelint-standard's kebab-case default. The token codegen
+        // owns the naming; authors should not invent new custom
+        // properties without going through the tokens package.
+        "custom-property-pattern": null,
+
+        // Reject `var(--anything-not-in-tokens)`. The token codegen
```
beaesguerra, line 83: Would it make sense for consuming clients (apps or libs) to create their own set of css variables? Or should they only be limited to wb css variables?

stylelint.config.ts
```diff
@@ -0,0 +1,140 @@
...
+        // failing CI. Tighten to `error` once Phase 5 component waves
+        // complete.
+        "scale-unlimited/declaration-strict-value": [
+            [
+                "/color/",
+                "background-color",
+                "border-color",
+                "padding",
+                "margin",
+                "font-size",
```
beaesguerra, line 159: Do we want to include other properties like `font-weight`, `line-height`, `border-width`, `border-radius` too?

I'm thinking about all our js tokens. We have box shadow and media query breakpoints too, but that might be trickier to use with this rule!

stylelint.config.ts
beaesguerra: Nice!! 

[color-named](https://stylelint.io/user-guide/rules/color-named/) would be helpful too! I'm not seeing an existing one for rgb values though, or other color formats! @marcysutton included several kinds in the [no-hardcoded-color eslint rule](https://github.com/Khan/wonder-blocks/pull/3053/changes#diff-3c28860db28f9cbb5c9dbab1bb616c94845bbec5cb3b8cf7f4c255c89aa71730R250-R262)!

stylelint.config.ts
```diff
@@ -0,0 +1,140 @@
+// Stylelint config for the CSS Modules migration (Phase 0.4 of WB-2324).
```
beaesguerra, line 1: It's great to have all these stylelint rules to help encourage WB usage! 

Would it make sense to have a stylelint plugin / config package for WB, similar to the eslint-plugin-wonder-blocks package? 

It would be nice to be able to centralize the WB related config in one place, so it is easy to configure and maintain in other projects. External docs for what rules are enabled would be great too. I also wonder if there's a way to have automated tests for the stylelint config so we can validate what it covers!

stylelint.config.ts
```diff
@@ -0,0 +1,140 @@
...
+        "csstools/value-no-unknown-custom-properties": [
+            true,
+            {
+                // Absolute path so the rule resolves the same regardless of
+                // the cwd stylelint is invoked from (root `pnpm lint:css`,
+                // editor / `vscode-stylelint`, or a package-scoped run).
+                importFrom: [
+                    path.resolve(
+                        __dirname,
+                        "packages/wonder-blocks-tokens/dist/css/index.css",
```
beaesguerra, line 99: Is there a way to omit certain tokens? For example, component level tokens that shouldn't be used by consumers!

We would want to include in the wb repo though since we use those tokens internally!

stylelint.config.ts
```diff
@@ -0,0 +1,140 @@
...
+        // Forbid hex literals (`#fff`, `#112233`). Pairs with
+        // `declaration-strict-value` above by closing the side-channel:
+        // a hex used in a property *not* on that list (e.g. `box-shadow`,
+        // `outline`, `fill`) would otherwise still slip through. Authors
+        // should reach for a `--wb-*` token via `var()` instead. WARN
+        // during the migration; tightens to `error` in Phase 5 alongside
+        // `declaration-strict-value`.
+        "color-no-hex": [true, {severity: "warning"}],
+    },
+};
```
beaesguerra, line 196: Do we have plans to add a stylelint rule to encourage using logical css properties for better rtl support?

stylelint.config.ts
beaesguerra: Or maybe we should import from `packages/wonder-blocks-tokens/dist/css/vars.css`. These don't seem to include the component tokens!

beaesguerra: I was also going to ask if we need to omit deprecated tokens like `spacing` and primitive `color`. Luckily, we don't use css variables for these! 

We do have the deprecated `font.size.*` tokens though! Which were replaced by `font.body/heading.size.*`

beaesguerra (commented):
Excited to have stylelint rules too to encourage WB usage! Left some questions 😄

stylelint.config.ts
beaesguerra: We might not need a lint rule for all the different color formats actually! Since we require tokens to be used, it'll be caught by the `scale-unlimited/declaration-strict-value` lint rule:

Image

stylelint.config.ts
```diff
@@ -0,0 +1,140 @@
...
+        "custom-property-pattern": null,
+
+        // Reject `var(--anything-not-in-tokens)`. The token codegen
+        // emits the canonical set of `--wb-*` custom properties into
+        // `packages/wonder-blocks-tokens/dist/css/index.css`; any
+        // reference outside that set is either a typo or an attempt to
+        // bypass the design system. Requires `pnpm build:css` (CI's
+        // build step already runs first; locally the `prelint:css` hook
+        // ensures the file is fresh).
+        "csstools/value-no-unknown-custom-properties": [
```
beaesguerra, line 90: I was able to bypass this rule by passing in a default value for a css variable that doesn't exist!


Image

beaesguerra, line 90: Do you know if there's a way to enable auto complete for the wb css variable names?

beaesguerra (commented):
Sharing some findings from testing out the stylelint rules locally!

.changeset/css-modules-stylelint.md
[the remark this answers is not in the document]
jandrade: TBH this is just claude adding the info that comes from the plan + implementation.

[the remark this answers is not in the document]
jandrade: @claude fix this

__docs__/css-modules-spike/spike.module.css
jandrade: I think this comes from the Stylelint standard config. Let me find the exact reason.

jandrade: That's correct, this comes from there: https://github.com/stylelint/stylelint-config-standard/blob/HEAD/index.js#L74

https://stylelint.io/user-guide/rules/import-notation/

That said, there's no effect on using url or plain string. I personally prefer the "string" notation, but I'm open to keep it wrapping `url()` if you think that's better.

__docs__/wonder-blocks-typography/styles.css
jandrade: not right now (I think), but it is definitively a good idea to caught that from start! I'll add that rule.

stylelint.config.ts
jandrade: ahh jinx see my related comment in the first PR :)

https://github.com/Khan/wonder-blocks/pull/3059#discussion_r3304111521

stylelint.config.ts
jandrade: these comments will be updated/removed by Claude once I tackle those phases in a future ticket :).

stylelint.config.ts
jandrade: IMO It should be open for consumers to decide. This part of the config will need to be revisited on the consumer side for sure.

stylelint.config.ts
jandrade: hmmm that shouldn't happen (in theory).... I'll take a look

jandrade: That should happen with this VSCode extension: https://github.com/Khan/wonder-blocks/pull/3069/changes#diff-a5de3e5871ffcc383a2294845bd3df25d3eeff6c29ad46e3a396577c413bf357R33-R35

Is it not working for you?

stylelint.config.ts
jandrade: ohh good point about `vars.css`! I totally forgot about that file :)

re: `spacing` -> That's the main reason why I decided to remove CSS vars support for spacing and other deprecated tokens. Good point about the `font.size.*` tokens. I'm close to clean up those tokens from consumers, so hopefully will get rid of those CSS vars.

stylelint.config.ts
jandrade: ohhh great point! Added `stylelint-use-logical` to cover that. Here's an example:

Image

__docs__/wonder-blocks-typography/styles.css
jandrade: Fixed!

stylelint.config.ts
jandrade: wait... now that I remember, we need to mark all the tokens that come from `index.css` as valid inside Wonder Blocks as we are going to use component tokens internally. We'll need to change to `vars.css` in consumer apps for sure!

stylelint.config.ts
jandrade: Added these properties! thanks for the suggestion.

stylelint.config.ts
jeresig: Yeah, I suspect that we're going to want to have similar linters set up in frontend too - so if there are common concerns we can either copy them over there or have a shared config package exported from WB that's re-used. That being said, I think it'll be easier to just copy over relevant configs. I suspect that you'll be able to be a lot stricter here compared with what we can get away with in frontend.

jeresig (approved):
Looks great, thank you!

__docs__/css-modules-spike/spike.module.css
beaesguerra: I prefer the string notation too, thanks for updating it!

stylelint.config.ts
```diff
@@ -0,0 +1,180 @@
...
+                 */
+                // radius
+                "border-radius",
+                // width
+                "border",
+                "border-width",
+                "border-top-width",
+                "border-bottom-width",
+                "border-left-width",
+                "border-right-width",
```
beaesguerra: Should we include the logical css property version for these? I'm not sure if we need both sets!

stylelint.config.ts
beaesguerra: Nice! cc: @maddy531 FYI, we'll also be including a css stylelint rule for logical css properties for the CSS module migration 😄

stylelint.config.ts
beaesguerra: Hmmm it isn't working for me! And I have the CSS Var Complete extension installed and tried reloading the IDE. Not sure why it is suggesting stylelint disable 🤔 


image

.changeset/css-modules-stylelint.md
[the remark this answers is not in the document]
jandrade: Claude never got to fix it, so I'm doing it manually 🤖

stylelint.config.ts
jandrade: ouch, I'll look into that!

jandrade: Fixed in the most recent commit

jandrade: Fixed! I had to switch to a different VSCode extension. It turns out that the one I originally installed recommended a different extension in the GH repo, so this one should work fine. I checked and it definitively makes sense to switch.

https://chatgpt.com/share/e/6a1d9b2b-0314-800f-b478-424bae2c2b4b

beaesguerra (approved):
Looks good to me! 🚀

PR #3070, opened 2026-05-22
SYL Dark - Tooltip

beaesguerra:
## Summary:

Reviewing the Tooltip component in SYL Dark to confirm it looks as expected, matches the design specs, and has no a11y color violations. We also enable Chromatic snapshots for the syl-dark theme for the component.

- Fixed the `TooltipTail` so that the `color` prop is optional. When no primitive color override is passed, the tail fill falls back to a semantic color token instead so it works across themes
- Use semantic color border token for the tooltip tail 

Issue: WB-2166
Design: https://www.figma.com/design/04ptAsL2PE4e8KGfYewta3/bea-syl-dark?node-id=5681-741&t=nbUJi6LBtja2ZGFs-4

Note: Currently, the Tooltip component supports a `backgroundColor` prop that only works with primitive colors. WB-2125 will introduce a new strong tooltip variant so that we can remove the `backgroundColor` and control which combinations of backgrounds and foregrounds can be used together in the tooltip

## Test plan:

Review the Tooltip across themes and confirm things look as expected and that there are no a11y violations related to color contrast `?path=/story/packages-tooltip-testing-snapshots-tooltip--state-sheet-story&globals=backgrounds.value:baseSubtle;backgrounds.grid:!false;theme:default`

### DIFF
--- a/__docs__/wonder-blocks-tooltip/tooltip-testing-snapshots.stories.tsx
+++ b/__docs__/wonder-blocks-tooltip/tooltip-testing-snapshots.stories.tsx
@@ -3,7 +3,7 @@ import {StyleSheet} from "aphrodite";
 import type {Meta, StoryObj} from "@storybook/react-vite";
 
 import {View} from "@khanacademy/wonder-blocks-core";
-import {sizing} from "@khanacademy/wonder-blocks-tokens";
+import {semanticColor, sizing} from "@khanacademy/wonder-blocks-tokens";
 import {BodyText} from "@khanacademy/wonder-blocks-typography";
 import Tooltip, {TooltipContent} from "@khanacademy/wonder-blocks-tooltip";
 
@@ -72,6 +72,16 @@ const rows = [
                     </BodyText>
                 </TooltipContent>
             ),
+        },
+    },
+    {
+        name: "Custom primitive background color",
+        props: {
+            content: "Custom background color",
+            backgroundColor: "darkBlue",
+            contentStyle: {
+                color: semanticColor.core.foreground.knockout.default,
+            },
         },
     },
 ];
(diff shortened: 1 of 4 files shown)


beaesguerra:
@claude review once

__docs__/wonder-blocks-tooltip/tooltip-testing-snapshots.stories.tsx
```diff
@@ -0,0 +1,137 @@
...
+            // in favour of a stroong tooltip variant.
+            //
+            // Note: This won't look right in syl-dark. We include this to ensure
+            // existing behaviour is not broken.
+            backgroundColor: "darkBlue",
+            contentStyle: {
+                color: semanticColor.core.foreground.knockout.default,
+            },
+        },
+    },
```
beaesguerra, line 91: Since the `backgroundColor` prop currently supports only primitive colors, this won't work in `syl-dark`. We include this case for now to make sure we don't break existing behaviour

We can remove support for `backgroundColor` once we add support for a new strong tool tip variant in https://khanacademy.atlassian.net/browse/WB-2125

__docs__/wonder-blocks-tooltip/tooltip-testing-snapshots.stories.tsx
jandrade, line 1: **thought:** This is great! now that we are covering this here, how do you feel about disabling most of the current stories in `tooltip.stories.tsx`?

jandrade (approved):
Looks great! Just left a minor open question for you to consider before landing 🚀

__docs__/wonder-blocks-tooltip/tooltip-testing-snapshots.stories.tsx
beaesguerra: I'll disable some existing tooltip snapshots in a follow up PR!

PR #3071, opened 2026-05-22
🚨 [WB-2324.4] Phase 0.1 — Rollup CSS emission + migrate Icon to CSS Modules

jandrade:
## Summary

Wires the build side of the CSS Modules pipeline and migrates the first real component (`wonder-blocks-icon`) as the end-to-end proof. Closes the last remaining sub-task of Phase 0.

Builds on top of [#3069](https://github.com/Khan/wonder-blocks/pull/3069) (Phase 0.4 — Stylelint). First PR in the chain that has visible runtime behavior changes.

Issue: WB-2324

## Test plan

- [x] `pnpm build` succeeds end-to-end.
- [x] `cat packages/wonder-blocks-icon/dist/index.css` shows the size rules wrapped in `@layer shared { … }` with hashed selectors (`.wbIcon_moduleSmall<hash>`, etc.).
- [x] `head -3 packages/wonder-blocks-icon/dist/es/index.js` shows `import "./index.css";` as the first line; `head -3 …/dist/index.js` shows `require("./index.css");` after `'use strict'`.
- [x] `pnpm jest packages/wonder-blocks-icon` passes — Icon tests assert behavior (ref forwarding, ARIA, accessibility), not class names.
- [x] `pnpm typecheck` passes; the new `icon.module.css.d.ts` is generated by the existing `pretypecheck` hook.
- [x] `pnpm lint:css` clean — the new `icon.module.css` passes `max-nesting-depth`, `color-no-hex`, `selector-class-pattern`, and `value-no-unknown-custom-properties` from PR #3069.
- [ ] (Manual) Open Storybook `Icon` stories: small / medium / large / xlarge variants render at the expected sizes, no visual regression vs. main.

## Review plan

1. 🚨 `build-settings/rollup.config.mjs`: adds `rollup-plugin-styles` to the plugin chain before `swc`. Emits per-package `dist/index.css` for any package that contains a `*.module.css` file. Injects a side-effect `import "./index.css"` (ESM) / `require("./index.css")` (CJS) at the top of those packages' JS bundles so npm consumers get styles automatically. SSR consumers that can't process CSS imports may need a loader / mock — standard webpack / Vite / Next.js setups handle this out of the box.
2. 🚨 `postcss.config.cjs`: adds a `wrap-in-layer` PostCSS plugin that moves every top-level rule into `@layer shared { … }`. Detects already-wrapped CSS and skips to avoid `@layer shared { @layer shared { … } }` nesting. Shared by Vite (Storybook) and Rollup (build) so authoring stays consistent.
3. ⚠️ `packages/wonder-blocks-icon/package.json`: adds `exports` map (`.` + `./css`) and `sideEffects: ["**/*.css"]`. Published-surface change patterned after `wonder-blocks-tokens`.
4. ⚠️ `packages/wonder-blocks-icon/src/components/icon.tsx`: migrates from `StyleSheet.create` + `getSize` to `className={styles[size]}`. The component still uses `addStyle("div")` so the `style` prop continues to accept Aphrodite `StyleType` values; only the *internal* size styles moved to CSS Modules.
5. 🔷 `packages/wonder-blocks-icon/src/components/icon.module.css` (new): the size classes themselves. Authors don't write `@layer shared { … }` here — the build wraps it.
6. 🔷 `__docs__/css-modules-spike/spike.module.css`: dropped the manual `@layer shared { … }` wrap since the build now does it. The PostCSS plugin's already-wrapped detection would have skipped re-wrapping anyway, but cleaning the source keeps authoring consistent across the codebase.

cc: @Khan/frontend-infra-web

### DIFF
--- a/.changeset/wonder-blocks-icon-css-modules.md
+++ b/.changeset/wonder-blocks-icon-css-modules.md
@@ -1,0 +1,16 @@
+---
+"@khanacademy/wonder-blocks-icon": patch
+---
+
+Internal: migrate the `Icon` component from Aphrodite to CSS Modules as
+the first real consumer of the new Rollup CSS pipeline (WB-2324 Phase
+0.1). The public component API is unchanged — same props, same DOM
+output. Consumers using `@khanacademy/wonder-blocks-icon` automatically
+pick up the bundled `dist/index.css` via a side-effect import in the
+JS entry; SSR consumers that can't process CSS imports may need a CSS
+loader / mock in their build (standard webpack / Vite / Next.js setups
+handle this out of the box).
+
+A new package subpath, `@khanacademy/wonder-blocks-icon/css`, exposes
+the extracted stylesheet for consumers that prefer to load it
+explicitly.
--- a/build-settings/rollup.config.mjs
+++ b/build-settings/rollup.config.mjs
@@ -1,6 +1,5 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 /* eslint-disable import/no-commonjs */
-import crypto from "crypto";
 import fs from "fs";
 import path from "path";
 import {nodeExternals} from "rollup-plugin-node-externals";
@@ -8,23 +7,11 @@ import swc from "@rollup/plugin-swc";
 import resolve from "@rollup/plugin-node-resolve";
 import styles from "rollup-plugin-styler";
 
-// Produce kebab-cased hashed selectors (`wb-<name>-<local>-<hash>`) instead
-// of styler's default, which routes the placeholder string through
-// `@rollup/pluginutils`'s `makeLegalIdentifier` and camel-cases away the
-// dashes. Hash is content-derived so the value stays stable across builds.
-const generateScopedName = (local, file, css) => {
-    const {base, name} = path.parse(file);
-    // `path.parse("icon.module.css").name` is `"icon.module"`; strip the
-    // `.module` suffix so selectors read as `wb-<component>-<local>-<hash>`
-    // (an embedded dot would otherwise render as a compound class anyway).
-    const safeName = name.replace(/\.module$/, "");
-    const hash = crypto
-        .createHash("sha256")
-        .update(`${base}:${css}`)
-        .digest("hex")
-        .slice(0, 8);
-    return `wb-${safeName}-${local}-${hash}`;
-};
+// Shared with Vite so Storybook and the published `dist` agree on the
+// hashed CSS Modules class names. Also dodges styler's default, which
+// routes the placeholder string through `@rollup/pluginutils`'s
+// `makeLegalIdentifier` and camel-cases away the dashes.
+import {generateScopedName} from "./css-modules-scoped-name.mjs";
 
 // Recursively check whether a package contains any `*.module.css` files
 // in its `src/` tree. Used to decide whether the package's JS bundle needs
@@ -33,7 +20,9 @@ const generateScopedName = (local, file, css) => {
 // themselves.
 const hasCssModules = (pkgName) => {
     const srcDir = path.join("packages", pkgName, "src");
-    if (!fs.existsSync(srcDir)) return false;
+    if (!fs.existsSync(srcDir)) {
+        return false;
+    }
     const entries = fs.readdirSync(srcDir, {recursive: true});
     return entries.some(
         (entry) => typeof entry === "string" && entry.endsWith(".module.css"),
--- a/build-settings/rollup.config.mjs
+++ b/build-settings/rollup.config.mjs
@@ -5,6 +5,21 @@ import path from "path";
 import {nodeExternals} from "rollup-plugin-node-externals";
 import swc from "@rollup/plugin-swc";
 import resolve from "@rollup/plugin-node-resolve";
+import styles from "rollup-plugin-styles";
+
+// Recursively check whether a package contains any `*.module.css` files
+// in its `src/` tree. Used to decide whether the package's JS bundle needs
+// a side-effect import of the extracted `dist/index.css`, so consumers
+// installing via npm don't have to wire `@khanacademy/wonder-blocks-*/css`
+// themselves.
+const hasCssModules = (pkgName) => {
+    const srcDir = path.join("packages", pkgName, "src");
+    if (!fs.existsSync(srcDir)) return false;
+    const entries = fs.readdirSync(srcDir, {recursive: true});
+    return entries.some(
+        (entry) => typeof entry === "string" && entry.endsWith(".module.css"),
+    );
+};
 
 const createConfig = (pkgName) => {
     const packageJsonPath = path.join("packages", pkgName, "package.json");
@@ -13,21 +28,48 @@ const createConfig = (pkgName) => {
     }
 
     const extensions = [".js", ".jsx", ".ts", ".tsx"];
+    const cssSideEffect = hasCssModules(pkgName);
 
     return {
         output: [
             {
                 file: `packages/${pkgName}/dist/es/index.js`,
                 format: "esm",
+                // Single CSS asset per package, regardless of how many
+                // `*.module.css` files were imported. Authoring layer-wrap
+                // happens in postcss.config.cjs (`@layer shared`).
+                assetFileNames: "index.css",
+                // Side-effect import of the extracted CSS so npm consumers
+                // get styles automatically when they import the package.
+                intro: cssSideEffect ? 'import "./index.css";' : "",
             },
             // TODO(FEI-5030): Stop building CJS modules
             {
                 file: `packages/${pkgName}/dist/index.js`,
                 format: "cjs",
+                assetFileNames: "index.css",
+                intro: cssSideEffect ? 'require("./index.css");' : "",
             },
         ],
         input: `packages/${pkgName}/src/index.ts`,
         plugins: [
+            // PostCSS + CSS Modules. Reads the root `postcss.config.cjs`
+            // chain (postcss-import → @csstools/postcss-mixins →
+            // wrap-in-layer) and hashes class names deterministically so
+            // they round-trip from Vite (Storybook) to Rollup (published
+            // dist) for the same source.
+            styles({
+                mode: "extract",
+                modules: {
+                    // Embeds the source file `[name]` + the local class
+                    // `[local]` and an 8-char content hash. Round-trips
+                    // deterministically across builds and stays readable
+                    // in DevTools.
+                    generateScopedName: "wb-[name]-[local]-[hash:8]",
+                    localsConvention: "camelCaseOnly",
+                },
+                sourceMap: true,
+            }),
             swc({
                 swc: {
                     swcrc: true,
--- a/build-settings/rollup.config.mjs
+++ b/build-settings/rollup.config.mjs
@@ -1,11 +1,30 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 /* eslint-disable import/no-commonjs */
+import crypto from "crypto";
 import fs from "fs";
 import path from "path";
 import {nodeExternals} from "rollup-plugin-node-externals";
 import swc from "@rollup/plugin-swc";
 import resolve from "@rollup/plugin-node-resolve";
-import styles from "rollup-plugin-styles";
+import styles from "rollup-plugin-styler";
+
+// Produce kebab-cased hashed selectors (`wb-<name>-<local>-<hash>`) instead
+// of styler's default, which routes the placeholder string through
+// `@rollup/pluginutils`'s `makeLegalIdentifier` and camel-cases away the
+// dashes. Hash is content-derived so the value stays stable across builds.
+const generateScopedName = (local, file, css) => {
+    const {base, name} = path.parse(file);
+    // `path.parse("icon.module.css").name` is `"icon.module"`; strip the
+    // `.module` suffix so selectors read as `wb-<component>-<local>-<hash>`
+    // (an embedded dot would otherwise render as a compound class anyway).
+    const safeName = name.replace(/\.module$/, "");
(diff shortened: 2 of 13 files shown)


build-settings/rollup.config.mjs
```diff
@@ -13,21 +28,48 @@ const createConfig = (pkgName) => {
...
+            // dist) for the same source.
+            styles({
+                mode: "extract",
+                modules: {
+                    // Embeds the source file `[name]` + the local class
+                    // `[local]` and an 8-char content hash. Round-trips
+                    // deterministically across builds and stays readable
+                    // in DevTools.
+                    generateScopedName: "wb-[name]-[local]-[hash:8]",
+                    localsConvention: "camelCaseOnly",
```
jandrade: **note:** Here's an example of how this would look like for consumers:

https://www.npmjs.com/package/@khanacademy/wonder-blocks-icon/v/0.0.0-PR3071-20260522210737?activeTab=code

`wbIcon_moduleMedium81fdcc3a`

beaesguerra:
@claude review once

packages/wonder-blocks-icon/src/components/icon.module.css
```diff
@@ -0,0 +1,23 @@
+/* Size variants for `Icon`. Tokens come from `@khanacademy/wonder-blocks-tokens`;
+ * the build wraps these rules in `@layer shared { … }` via the wrap-in-layer
```
beaesguerra, line 2: Good to know we don't need to explicitly wrap WB rules with the shared layer

packages/wonder-blocks-icon/src/components/icon.tsx
```diff
@@ -79,7 +60,8 @@ const Icon = React.forwardRef(
 
         return (
             <StyledDiv
-                style={[getSize(size), style]}
+                className={styles[size]}
```
beaesguerra: I was curious to see if the typecheck would catch if there was a mismatch between the class names and size names, and it did 🎉 

I tested this by renaming the `xlarge` typed string to `extraLarge` and also tried renaming the class name. When I renamed the class name though, I had to re-run `pnpm gen:css-types` in order to update the types for the stlyes (which makes sense!).

Would it make sense to include `pnpm gen:css-types` as part of the `pnpm dev` script so that the css types are updated automatically? (Let me know too if there's already something that should be doing that!)

packages/wonder-blocks-icon/package.json
```diff
@@ -19,6 +19,17 @@
...
+  "sideEffects": [
+    "**/*.css"
+  ],
+  "exports": {
+    ".": {
+      "import": "./dist/es/index.js",
+      "require": "./dist/index.js",
+      "types": "./dist/index.d.ts"
+    },
+    "./css": "./dist/index.css"
```
beaesguerra, line 31: Do consumers need to import the package's css file explicitly?

package.json
```diff
@@ -158,6 +158,7 @@
     "remark-gfm": "^4.0.0",
     "rollup": "^2.79.2",
     "rollup-plugin-node-externals": "^8.0.0",
+    "rollup-plugin-styles": "^4.0.0",
```
beaesguerra: Since I'm not too familiar with css/bundling tooling, I've been asking Claude what the new dependencies do for my own understanding! For this one, it mentioned:
```
  Note: the package is unmaintained (last release 2021); the active fork is
  rollup-plugin-styler, and many newer projects use @rollup/plugin-postcss or Vite's
  built-in CSS handling instead.
```

Are these alternatives worth considering?

build-settings/rollup.config.mjs
```diff
@@ -5,6 +5,21 @@ import path from "path";
 import {nodeExternals} from "rollup-plugin-node-externals";
 import swc from "@rollup/plugin-swc";
 import resolve from "@rollup/plugin-node-resolve";
+import styles from "rollup-plugin-styles";
+
+// Recursively check whether a package contains any `*.module.css` files
```
beaesguerra, line 16: Would we ever import a `*.module.css` file from another package? For example, if we had shared styles? I'm assuming no, and we would probably want to use a mixin for those cases?

build-settings/rollup.config.mjs
beaesguerra: Is the `module` part due to the file name? I wonder if it would make sense to exclude that part for brevity since all scoped names would include that? 

Though I see this name doesn't end up getting included in the rendered output! Why is that? I am still wrapping my head around it all! 😄 

Image

I asked Claude and it seems to be because Storybook runs vite, not Rollup!

beaesguerra: On another note, I like that we're prefixing the class name with `wb`! 

Assuming this is what would get rendered in applications, we can more easily determine if something is using WB or not! And we can also explore [visual measurements](https://www.productboard.com/blog/how-we-measure-adoption-of-a-design-system-at-productboard/#:~:text=Everything%20green%20is%20imported%20from%20a%20design%20system) of WB usage in apps!

.changeset/wonder-blocks-icon-css-modules.md
```diff
@@ -0,0 +1,16 @@
+---
+"@khanacademy/wonder-blocks-icon": patch
+---
+
+Internal: migrate the `Icon` component from Aphrodite to CSS Modules as
+the first real consumer of the new Rollup CSS pipeline (WB-2324 Phase
+0.1). The public component API is unchanged — same props, same DOM
+output. Consumers using `@khanacademy/wonder-blocks-icon` automatically
+pick up the bundled `dist/index.css` via a side-effect import in the
+JS entry; SSR consumers that can't process CSS imports may need a CSS
```
beaesguerra, line 10: Do consuming apps/libraries need to update anything in their project for these changes to work? If so, should this be considered a breaking change?

beaesguerra (commented):
First component to use CSS Modules, nice! Left some more questions 😄

.changeset/wonder-blocks-icon-css-modules.md
jandrade: Good question! this should be transparent for consumers :) as they are still importing the JS bundles that reference these CSS files.

build-settings/rollup.config.mjs
jandrade: that's correct, we should be using mixins that come from the `wb-styles` package.

packages/wonder-blocks-icon/src/components/icon.tsx
jandrade: It is already running as part of the `prebuild` script, which is invoked at some point by `dev` :)

packages/wonder-blocks-icon/package.json
jandrade: No need as the JS bundler takes care of it.

package.json
jandrade: I considered using `@rollup/plugin-postcss` at some point, but that's even older than the one currently picked in the PR.

I'm going to explore the other package + other ones. I'll keep you posted!

jandrade: Ended up switching to `rollup-plugin-styler`! thanks for looking into that. As a long term solution, I was considering switching from `rollup` to use vite/rolldown for bundling, so maybe we can use that instead, that way we don't have to customize this part as it is handled internally by default.

jandrade:
The integration with consumers can be seen here: https://github.com/Khan/frontend/pull/11572

This includes the WB `Icon` component using CSS modules. The mixins part (`--wb-focus-visible`) can only be integrated once the consumer repo defines a postcss pipeline (and includes a config file that is capable of processing the `postcss-mixins` plugin).

For reference on the config, see: https://github.com/Khan/wonder-blocks/blob/WB-2324.1/postcss.config.cjs#L76

Screenshot 2026-05-26 at 12 26 55 PM

cc @Khan/wonder-blocks @Khan/frontend-infra-web

build-settings/rollup.config.mjs
jandrade: I've modified the tooling to reuse the same prefix for both Vite (Storybook) and rollup (bundle).

Also removed the `module` part from the classname.

Screenshot 2026-05-27 at 10 27 37 AM

packages/wonder-blocks-icon/src/components/icon.tsx
```diff
@@ -77,9 +58,13 @@ const Icon = React.forwardRef(
             },
         });
 
+        // only allow the sizes defined in IconSize
+        const sizeStyles = styles[size] || styles.small;
+
         return (
             <StyledDiv
-                style={[getSize(size), style]}
+                className={sizeStyles}
```
jeresig, line 66: I wonder if you're going to want to add support for something like clsx https://www.npmjs.com/package/clsx to our addStyle logic so that we can handle cases like `className={["class1", condition && "class2"]}`.

jeresig (approved):
Looks good, thank you!

packages/wonder-blocks-icon/src/components/icon.tsx
jandrade: oh yeah, thanks for reminding me of that! I've considered `clsx` for that, but I'm also exploring other alternatives around that problem.

As a short term solution around the `addStyle` logic, my goal is to refactor the internals of that function so we can incrementally migrate to CSS modules using the existing `style` prop. You can see how this is going to work here: https://github.com/Khan/wonder-blocks/pull/3073/changes/b1cb814f64074ef206817ea5a1cc09061b800e9b#diff-56c6348fe785a54784ccb372f1a9f9f946e2520366035f7957f2b2177544fcd5

As for supporting libraries like `clsx`, I'm planning to write a new ADR to evaluate different tools: so far, I've contemplated `clsx`, `classnames` and `cva`.

You can see some research I quickly did with ChatGPT some weeks ago: https://chatgpt.com/share/e/69f3ab89-9440-800f-b0fc-a5ee94613fd5

.changeset/wonder-blocks-icon-css-modules.md
beaesguerra: Sounds good!

Will there be a major version release for something in WB? Since consuming apps would need to set up postcss/bundling before using new versions of certain packages? Maybe just the `styles` package? Or maybe not since the new css mixin is additive!

packages/wonder-blocks-icon/src/components/icon.tsx
beaesguerra: Oh cool! I was going to try it out again, but am running into errors now when running `pnpm dev`:

```
turbo 2.5.3

• Packages in scope: //, @khanacademy/eslint-plugin-wonder-blocks, @khanacademy/wb-codemod, @khanacademy/wb-dev-build-settings, @khanacademy/wonder-blocks-accordion, @khanacademy/wonder-blocks-announcer, @khanacademy/wonder-blocks-badge, @khanacademy/wonder-blocks-banner, @khanacademy/wonder-blocks-birthday-picker, @khanacademy/wonder-blocks-breadcrumbs, @khanacademy/wonder-blocks-button, @khanacademy/wonder-blocks-card, @khanacademy/wonder-blocks-cell, @khanacademy/wonder-blocks-clickable, @khanacademy/wonder-blocks-core, @khanacademy/wonder-blocks-data, @khanacademy/wonder-blocks-date-picker, @khanacademy/wonder-blocks-dropdown, @khanacademy/wonder-blocks-form, @khanacademy/wonder-blocks-grid, @khanacademy/wonder-blocks-icon, @khanacademy/wonder-blocks-icon-button, @khanacademy/wonder-blocks-labeled-field, @khanacademy/wonder-blocks-layout, @khanacademy/wonder-blocks-link, @khanacademy/wonder-blocks-modal, @khanacademy/wonder-blocks-pill, @khanacademy/wonder-blocks-popover, @khanacademy/wonder-blocks-progress-spinner, @khanacademy/wonder-blocks-search-field, @khanacademy/wonder-blocks-styles, @khanacademy/wonder-blocks-switch, @khanacademy/wonder-blocks-tabs, @khanacademy/wonder-blocks-testing, @khanacademy/wonder-blocks-testing-core, @khanacademy/wonder-blocks-theming, @khanacademy/wonder-blocks-timing, @khanacademy/wonder-blocks-tokens, @khanacademy/wonder-blocks-toolbar, @khanacademy/wonder-blocks-tooltip, @khanacademy/wonder-blocks-typography, eslint-plugin-wonder-blocks-demo
• Running build:css in 42 packages
• Remote caching disabled
  × Failed to connect to daemon.
  ╰─▶ unable to connect to daemon after 3 retries
```

I made sure to run `pnpm i` first to make sure deps are up to date!

.changeset/wonder-blocks-icon-css-modules.md
jandrade: My plan is the following:

1. Configure WB tooling
2. Start migrating WB components
3. Configure consumer repos tooling
4. Install WB versions with CSS modules support in consumer repos

#3 can be done before #2, so I think `minor/patch` versions would work transparently for consumer apps as I'd expect that at that point (#4) all the tooling/infra should be configured in advance. Agree about the `styles` package, as mixins are additive, and I'm planning to configure mixins in consumer repos as part of #3 as well.

packages/wonder-blocks-icon/src/components/icon.tsx
```diff
@@ -79,7 +60,8 @@ const Icon = React.forwardRef(
 
         return (
             <StyledDiv
-                style={[getSize(size), style]}
+                className={styles[size]}
+                style={style}
                 id={id}
```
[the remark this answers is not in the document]
jandrade: Fixed!

packages/wonder-blocks-icon/src/components/icon.tsx
jandrade: Just tried this again and it didn't happen to me, I'll keep an eye on it!

PR #3010, opened 2026-04-14
Install TSTyche and migrate typestests

benchristel:
## Summary:
In [ADR #923] we decided to use TSTyche for type testing in TypeScript code.

[ADR #923]: https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/4682383644/ADR+923+Adopt+tstyche+for+type+testing+in+TypeScript+code

This PR updates the typestests in WB to use TSTyche.

Issue: LEMS-4034
## Test plan:

`pnpm tstyche`

PR #3073, opened 2026-05-25
🚨 [WB-2325] Phase 1 — Extend processStyleList to accept CSS Module class names

jandrade:
## Summary:
Widen StyleType to also accept `string` (and nested arrays of strings) so
consumers can pass CSS Module identifiers alongside Aphrodite stylesheets,
inline objects, and falsy values.

Refactor processStyleList to sort each leaf into one of three buckets:
Aphrodite stylesheets (via the existing `_definition` symbol check) →
emitted through `css(...)`; plain strings → appended to `className`;
plain objects → forwarded as inline `style`. Non-string, non-object
truthy leaves are silently ignored.

Add a fast-path: when no Aphrodite stylesheet is present in the merge,
skip the `StyleSheet.create` rewrap (whose only purpose was to outweigh
Aphrodite's `!important`) and forward inline styles directly via the
`style` prop. This is the intentional behavioral change called out in
the changeset — DOM output for the no-Aphrodite case shifts from a
generated `inlineStyles_xxx` class to a real `style="..."` attribute.

Fix latent spread-of-StyleType bugs in Cell, DatePicker, and Modal that
the widened union now flags at type-check time. Behavior is preserved
or improved — no more accidental `{0: 'f', 1: 'o', ...}` results when a
non-object slipped into a spread.

Issue: WB-2325

## Test plan:

Verify that Chromatic snapshots have not changed. Focus on the `/?path=/docs/tools-css-modules-spike--docs` and see that there are no visual regressions introduced.


__docs__/css-modules-spike/spike.stories.tsx
```diff
@@ -22,9 +22,9 @@ export default {
     component: Spike,
     parameters: {
         chromatic: {
-            // Spike artifact, not a real component — no need to spend a
-            // Chromatic snapshot on it.
-            disableSnapshot: true,
+            // NOTE: This will be switched back to true once we start migrating
+            // real components.
+            disableSnapshot: false,
```
jandrade, line 27: **note:** I enabled these tests just to make sure that the changes in this PR won't affect the current spike examples.

Here's the baseline for those snapshots: https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=6a1ee965c9c21314969c4bb6

I've accepted the baselines, and these shouldn't change once I change the Spike component to use `addStyle`.

__docs__/css-modules-spike/spike.tsx
```diff
@@ -20,9 +22,25 @@ type Props = {
...
+    root: {
+        color: semanticColor.core.foreground.instructive.default,
+    },
+});
+
+const inlineStyles = {
+    root: {
+        backgroundColor: semanticColor.core.background.instructive.subtle,
+    },
+};
```
jandrade, line 48: **note:** These are temporary styles that I added to ensure that all kind of styles continue working as expected with `addStyle`.

jandrade:
I've created this snapshot PR to verify that this doesn't break consumers: https://github.com/Khan/frontend/pull/11912

cc @Khan/wonder-blocks

packages/wonder-blocks-cell/src/components/internal/cell-core.tsx
```diff
@@ -75,7 +75,7 @@ const RightAccessory = ({
                 styles.accessory,
                 styles.accessoryRight,
                 disabled && styles.accessoryDisabled,
-                {...style},
+                style,
```
jandrade, line 78: **note:** This is just a small refactor to make the new types happy. No styles should be affected by this change.

packages/wonder-blocks-date-picker/src/components/date-picker-overlay.tsx
```diff
@@ -127,11 +127,21 @@ const DatePickerOverlay = ({
...
+                    // Only spread `style` if it's a plain CSSProperties object;
+                    // StyleType also allows strings, arrays, and falsy values
+                    // which are not valid spreadable inputs here.
+                    const customStyle =
+                        style &&
+                        typeof style === "object" &&
+                        !Array.isArray(style)
+                            ? style
+                            : null;
+
```
jandrade, line 139: **note:** This is just a small refactor to make the new types happy. No styles should be affected by this change.

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
```diff
@@ -121,14 +121,13 @@ export default function FlexiblePanel({
...
-    };
-
     return (
         <View
-            style={[componentStyles.wrapper, combinedBackgroundStyles]}
+            style={[
+                componentStyles.wrapper,
+                defaultBackgroundStyle,
+                styles?.panel,
+            ]}
```
jandrade, line 130: **note:** This is just a small refactor to make the new types happy. No styles should be affected by this change.

postcss.config.cjs
```diff
@@ -17,9 +17,31 @@
 //   - Vite handles `*.module.css` natively (its built-in CSS Modules pass
 //     runs *after* this PostCSS chain).
 //   - Rollup uses `rollup-plugin-styler`'s `modules` option.
+const {createRequire} = require("node:module");
+const path = require("node:path");
 const postcssImport = require("postcss-import");
 const postcssMixins = require("@csstools/postcss-mixins");
 
+// postcss-import's default resolver (browserify/resolve) does not honor the
```
jandrade, line 25: **note:** The following changes were originally introduced in the `frontend` integration demo PR, and I'm bringing this back to WB so we can safely import CSS inter-depencencies

packages/wonder-blocks-core/src/util/add-style.js.flow
jandrade, line 1: **note:** Just getting rid of this unused file.

packages/wonder-blocks-core/src/util/__tests__/util.test.ts
jandrade, line 1: **note:** We didn't have tests for this util file before, so I instructed claude to help me generating tests that would cover the previous functionality (Aphrodite + inline styles), then to also include CSS modules support.

jandrade:
@claude review once

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
beaesguerra: Since the types have changed, should we mark these changes as a `major` version change since consumers may need to update how they pass in styles?

__docs__/css-modules-spike/spike.tsx
```diff
@@ -20,9 +25,24 @@ type Props = {
  */
 export function Spike({label, badge}: Props): React.ReactElement {
     return (
-        <button type="button" className={styles.root}>
+        <StyledButton
+            type="button"
+            style={[styles.root, aphroditeStyles.root, inlineStyles.root]}
```
beaesguerra, line 30: Neat!

packages/wonder-blocks-core/src/util/__tests__/add-style.test.tsx
```diff
@@ -89,6 +88,81 @@ describe("addStyle", () => {
         expect(classNames[1]).toEqual("foo");
     });
 
+    describe("CSS Modules", () => {
+        it("should forward a CSS module class name (string `style`) as a className", () => {
+            // Arrange
+            render(<StyledDiv style="module-class" data-testid="styled-div" />);
```
beaesguerra, line 94: Will we be moving away from `addStyle` long term? From what I understand, we were using addStyles so it would work with `aphrodite`! 

Would `<StyledDiv style="module-class" />` turn into `<div className="module-class" />` later on?

beaesguerra: I'm also curious if we'll want to support `className` props for WB components, and then the current `style`/`styles` props would be used for inline styles!

packages/wonder-blocks-core/src/util/__tests__/add-style.test.tsx
```diff
@@ -89,6 +88,81 @@ describe("addStyle", () => {
...
+
+            // Assert
+            expect(div).toHaveAttribute("class", "module-a module-b");
+        });
+
+        it("should append a CSS module class name after an aphrodite-generated class", () => {
+            // Arrange
+            render(
+                <StyledDiv
+                    style={[styles.foo, "module-class"]}
```
beaesguerra, line 152: Interesting! Does this mean css module styles will take precedence over aphrodite styles, assuming the rules are the same specificity?

packages/wonder-blocks-core/src/util/__tests__/add-style.test.tsx
```diff
@@ -89,6 +88,81 @@ describe("addStyle", () => {
...
+
+            // Act
+            const div = screen.getByTestId("styled-div");
+
+            // Assert — aphrodite-generated class first, then the CSS module class.
+            const classNames = div.className.split(" ");
+            expect(classNames[1]).toEqual("module-class");
+        });
+    });
+
```
beaesguerra, line 165: I'm curious about the use case:

```
<StyledDiv 
  style={[{ color: "red" }, "module-class"]}
/>
```

Would the `color: "red"` become an aphrodite style? Or an inline style? 

From what I understand, it'll become an inline style, but thought I'd check!

packages/wonder-blocks-core/src/util/add-styles.typestest.tsx
```diff
@@ -20,3 +20,25 @@ const list1 = <StyledUl />;
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const list2 = <StyledUl style={styles.customList} />;
+
+// Widened StyleType: a single CSS Module class name string is accepted.
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
beaesguerra, line 25: Oh interesting, I thought all our typestests were migrated to tsyche! Looks like this one might've been missed in https://github.com/Khan/wonder-blocks/pull/3010 

(Can be done separately since the existing tests are in this format already!)

beaesguerra (commented):
Nice, thanks for working on this! I left some questions 😄

packages/wonder-blocks-core/src/util/__tests__/add-style.test.tsx
jandrade: The idea is to keep `addStyle` as it still allow us to centralize how we handle styles in general, but that doesn't block folks from using `<div className=...`.

As for `style/styles`, the idea is to keep that around as folks are already aware with how to use that API, so they don't have to learn anything new (within KA).

`processStyleList` will continue being the responsible for adding the right styles to each prop regardless of how these are injected into the WB component instances.

packages/wonder-blocks-core/src/util/__tests__/add-style.test.tsx
jandrade: oh good question, this test could be misleading as this is not the case as we define CSS module styles in the `shared` @layer. Aphrodite styles are still part of the "main" layer, meaning that these take precedence over css modules styles.

packages/wonder-blocks-core/src/util/__tests__/add-style.test.tsx
jandrade: It will become an inline style. And btw, this is the main reason why we recommend folks to use the `StyleSheet.create()` wrapper instead of going to style plain objects (e.g. `const styles = {foo: {...}}`).

packages/wonder-blocks-core/src/util/add-styles.typestest.tsx
jandrade: yeah I thought about that too, but I didn't want to introduce more changes in this PR :)

packages/wonder-blocks-modal/src/components/flexible-panel.tsx
jandrade: This shouldn't have changed the styles, it is more like the object wrapper above (combinedBackgroundStyles) was unnecessary as `processStyleList` was already covering that for us, so this ended up being a micro-optimization that was caught by the refined types :)

beaesguerra: Ah gotcha! Did TypeScript catch this use case? Do we consider changes that could introduce TS errors breaking changes?

packages/wonder-blocks-core/src/util/__tests__/add-style.test.tsx
beaesguerra: Good to know! I was wondering if we would have to then support both `classNames` and `styles` props for wb components, so I'm glad we'll keep it to `style/styles`!

jandrade: maybe we can add `classNames` support if we prefer it, but for now it feels safer having to support only one API :)

jeresig (approved):
This seems like a great change to me - thanks for handling this!

PR #3078, opened 2026-05-28
Resolve hanging Run Storybook Test Playwright/chromium installs for CI

marcysutton:
## Summary:
I keep hitting timeouts/hangs in the Run Storybook Tests workflow in CI. Trying to cache Playwright so the runner doesn't hang, with these changes:

**Root problems fixed:**

1. **Silent hang on "Run Storybook tests"** — `chromatic.com` was missing from `extra-domains`, so Playwright couldn't reach the Chromatic-hosted Storybook URL. The step produced no output and ran until the 6-hour job timeout.

2. **2–6 hour Playwright browser download** — `playwright install chromium` downloads a 165 MiB binary. The primary CDN (`cdn.playwright.dev` → `storage.googleapis.com`) was blocked by the secure network action, forcing a fallback to `playwright.download.prss.microsoft.com` which is heavily throttled from these runners.

**Changes:**

- **`vitest.config.ts`** — Pass `executablePath` through the playwright provider's `launchOptions` when `PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH` is set. No effect on local development (env var won't be set).

- **`chromatic-pr.yml`** — Replace the Playwright browser download with a step that locates the pre-installed `google-chrome-stable` on `ubuntu-latest` and exports it as `PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH`. No download, no caching complexity. Add `chromatic.com` to `extra-domains` so the test step can reach the Storybook URL. Add `timeout-minutes: 15` as a safety net. Trim `extra-domains` from 11 entries down to just the npm registry, apt mirrors (fonts), and `chromatic.com` — no Playwright CDN domains needed since there's nothing to download. The Secure Network step stays at the top of the job (required by policy enforcement), so all steps that need external network access — npm install, apt font packages, and the Chromatic URL during tests — are listed in `extra-domains` rather than attempting to run before the filter.

Issue: WB-XXXX

## Test plan:
1. See if actions finish


marcysutton:
Closing in favor of https://github.com/Khan/wonder-blocks/pull/3079

PR #3080, opened 2026-05-28
Update the agent settings.

jeresig:
## Summary:
This is just bringing over the latest agent-settings which has a bunch of improvements. Noticed that a bunch of stuff was missing when working in the repo!

Issue: FEI-XXXX

## Test plan:


.claude/hooks/gh-filter.mjs
```diff
@@ -56,6 +65,7 @@ const ALLOWED_COMMANDS = [
     "pr status",
     "pr comment",
     "pr review",
+    "pr create",
```
jandrade, line 68: Good addition!

jandrade (approved):
Looks great! Thanks for the improvements :shipit:

PR #3081, opened 2026-05-28
Add links to tokenValue docs from semantic color token docs

beaesguerra:
## Summary:

Add links to the `tokenValue` docs from the semantic color docs, as [suggested by Ned](https://khanacademy.slack.com/archives/C4PE1QM5Y/p1779990860497309?thread_ts=1779990329.832639&cid=C4PE1QM5Y)! 

Issue: WB-2356

## Test plan:

Review the Using Color and Semantic Color Token docs:
- `?path=/docs/packages-tokens-semantic-color--docs#usage`
- `?path=/docs/foundations-using-color--docs#getting-raw-token-values`

### DIFF
--- a/__docs__/foundations-color.mdx
+++ b/__docs__/foundations-color.mdx
@@ -75,6 +75,7 @@ how to use these colors depending on the context.
     />
 </View>
 
+
 ### Core
 
 Core colors are used for the most common elements in the UI. These colors are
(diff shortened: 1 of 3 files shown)


__docs__/foundations-color.mdx
```diff
@@ -239,6 +240,15 @@ Standalone colors used only for communicating mastery.
...
 <Canvas of={SemanticColorGroups.Mastery} layout="fullscreen" />
 
+
+## Getting raw token values
+
+Semantic colors are exported as CSS `var(--...)` references so the active theme
+can switch values through the CSS cascade. This is great for styling, but less
+useful when something expects a concrete hex value (e.g. a third-party library).
+Use the [tokenValue](./?path=/docs/packages-tokens-utilities-tokenvalue--docs)
+utility to resolve a token to its computed raw value at runtime.
```
jandrade, line 250: Love this description!

One thing that I'm considering.... do you think we should include a Banner (or similar) to inform folks that this should only be used for those particular edge cases. I want to find a way to communicate this in a way that we don't overuse this API as it could break theming.

jandrade (approved):
Looks great! just a minor suggestion but the improvements look solid, thanks! 🚀

nedredmond (approved):
Thanks!

__docs__/foundations-color.mdx
beaesguerra: Added banners and docs for this, thanks for the suggestion!

PR #3084, opened 2026-05-29
SYL Dark - Selects

beaesguerra:
## Summary:

Reviewing the SelectOpener (SingleSelect and MultiSelect) components in SYL Dark to confirm they look as expected, match the design specs, and have no a11y color violations. We also enable Chromatic snapshots for the syl-dark theme for these components.

Changes include:
- Updating the disabled and readonly tokens for SYL Dark
  - Because of this, we also update the chonky button disabled shadow so it looks flat in syl-dark if the background is also `gray_05`
- Using `semanticColor.core.foreground.neutral.default` for the dropdown icon in all themes (this will be gray in all themes instead of blurple in tb/syl-dark). We remove the component level `theme.opener.color.icon` token for this
- Fixed the `placeholder` styling in selects when the field is disabled or readOnly
- Fixed the press + focused styling to work with syl-dark so that the white focus outline is included 

Issue: WB-2166
Design: https://www.figma.com/design/04ptAsL2PE4e8KGfYewta3/bea-syl-dark?node-id=8389-10450&t=48QoihVmhYQeteLi-4

## Test plan:

Review the following stories and confirm things look as expected and there are no a11y violations:
- `?path=/story/packages-dropdown-testing-snapshots-singleselect--state-sheet-story&globals=theme:syl-dark`
- `?path=/story/packages-dropdown-testing-snapshots-multiselect--state-sheet-story&globals=theme:syl-dark`

Review visual changes in Chromatic and make sure all changes are intentional

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/select-opener.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/select-opener.tsx
@@ -271,11 +271,17 @@ const styles = StyleSheet.create({
             // the focus behaviour more consistent with the other field components.
             ...focusStyles.focus[":focus-visible"],
         },
+        [":focus:active" as any]: {
+            boxShadow: `${PRESS_SHADOW}, ${focusStyles.focus[":focus-visible"].boxShadow}`,
+        },
     },
     error: {
         background: semanticColor.input.error.background,
         border: `${theme.opener.border.width.error} solid ${semanticColor.input.error.border}`,
         color: semanticColor.input.error.foreground,
+        [":focus:active" as any]: {
+            boxShadow: `${PRESS_SHADOW}, ${focusStyles.focus[":focus-visible"].boxShadow}`,
+        },
     },
     disabled: {
         background: semanticColor.input.disabled.background,
@@ -285,7 +291,9 @@ const styles = StyleSheet.create({
         ":active": {
             boxShadow: "none",
         },
-        ":focus": focusStyles.focus[":focus-visible"],
+        [":focus:active" as any]: {
+            boxShadow: focusStyles.focus[":focus-visible"].boxShadow,
+        },
     },
     press: {
         boxShadow: PRESS_SHADOW,
@@ -304,6 +312,8 @@ const styles = StyleSheet.create({
         ":active": {
             boxShadow: "none",
         },
-        ":focus": focusStyles.focus[":focus-visible"],
+        [":focus:active" as any]: {
+            boxShadow: focusStyles.focus[":focus-visible"].boxShadow,
+        },
     },
 });
--- a/packages/wonder-blocks-dropdown/src/components/select-opener.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/select-opener.tsx
@@ -155,7 +155,7 @@ export default class SelectOpener extends React.Component<
 
         const iconColor = disabled
             ? semanticColor.core.foreground.disabled.default
-            : theme.opener.color.icon;
+            : semanticColor.core.foreground.neutral.default;
 
         const style = [
             styles.shared,
--- a/packages/wonder-blocks-dropdown/src/components/select-opener.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/select-opener.tsx
@@ -166,7 +166,8 @@ export default class SelectOpener extends React.Component<
             isPlaceholder && disabled && styles.disabledPlaceholder,
             !disabled && !readOnly && this.state.pressed && styles.press,
             readOnly && styles.readOnly,
-            readOnly && isPlaceholder && styles.readOnlyPlaceholder,
+            // Re-apply placeholder styles for readOnly + placeholder combination
+            readOnly && isPlaceholder && styles.placeholder,
         ];
 
         const allowInteraction = !disabled && !readOnly;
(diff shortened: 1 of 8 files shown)


packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -155,16 +155,18 @@ export default class SelectOpener extends React.Component<
 
         const iconColor = disabled
             ? semanticColor.core.foreground.disabled.default
-            : theme.opener.color.icon;
+            : semanticColor.core.foreground.neutral.default;
```
beaesguerra, line 158: We can use the same token across all themes to match the Figma specs and remove the component specific token

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -262,20 +264,26 @@ const styles = StyleSheet.create({
...
+            boxShadow: PRESS_SHADOW,
+        },
         ":focus": {
             // Using :focus instead of :focus-visible to ensure the focus ring is
             // visible when the button is focused (even when clicked). This makes
             // the focus behaviour more consistent with the other field components.
             ...focusStyles.focus[":focus-visible"],
         },
-        ":active": {
+        [":focus:active" as any]: {
```
beaesguerra, line 277: Adding `:focus:active` styling to address issues with the focus + press states (there wasn't an explicit white focus outline included previously since the background color was white in light themes)

| Before | After |
| --- | --- |
| Image | Image |

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -155,16 +155,18 @@ export default class SelectOpener extends React.Component<
...
         const style = [
             styles.shared,
             styles.default,
             disabled && styles.disabled,
             error && styles.error,
             isPlaceholder && styles.placeholder,
+            isPlaceholder && disabled && styles.disabledPlaceholder,
             !disabled && !readOnly && this.state.pressed && styles.press,
             readOnly && styles.readOnly,
+            readOnly && isPlaceholder && styles.readOnlyPlaceholder,
```
beaesguerra: Add placeholder styling for reaconly and disabled states

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-syl-dark.ts
```diff
@@ -76,8 +76,9 @@ const core = {
         },
         disabled: {
             subtle: transparent,
-            default: color.gray_05,
-            strong: color.gray_20,
+            default: color.gray_10,
+            // There is no difference between subtle and default in dark mode, intentional to sync with design
+            strong: color.gray_10,
```
beaesguerra, line 81: We updated disabled tokens to fix the issue where the readonly styling wasn't visible in syl-dark ([Related discussion](https://khanacademy.slack.com/archives/C07CA2YR0BZ/p1780087039144839?thread_ts=1779826920.718159&cid=C07CA2YR0BZ))

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
```diff
@@ -297,11 +308,19 @@ const styles = StyleSheet.create({
     placeholder: {
         color: semanticColor.input.default.placeholder,
     },
+    disabledPlaceholder: {
+        color: semanticColor.input.disabled.placeholder,
+    },
+    readOnlyPlaceholder: {
+        color: semanticColor.input.default.placeholder,
+    },
```
jandrade: **suggestion/thought:** Are we planning to change this in the future? wondering if we can reuse the `placeholder` style instead?

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
marcysutton: Does this work cross-browser? I know focus point typically moves when you click, but I haven't tested the CSS pseudo selector combinations like this to see if `focus` always matches on `active`.

(Looks good in Chrome, Firefox, and Safari on desktop!)

jandrade: AFAIK it should work, but if that's not the case, maybe for this specific case, `:focus-visible:active` could work too?

jandrade (approved):
Looks good to me! I have a minor suggestion for you to consider before landing 🚀

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
beaesguerra: For context, we use `focus` instead of `focus-visible` so Selects have a similar behaviour to the other form fields when you press a field:
https://github.com/Khan/wonder-blocks/blob/29832f7f194134d4630ec68e2801b15ba8a15a32/packages/wonder-blocks-dropdown/src/components/select-opener.tsx#L265-L270

Here's what the difference between `:focus` and `:focus-visible` looks like!

https://github.com/user-attachments/assets/7c1a3ade-4c68-43b2-a939-14c824c7ff80

I tested Chrome on my Android phone and am able to see the white outline when I press and hold on the select:

https://github.com/user-attachments/assets/1fe2803a-3046-434c-a1b3-3668b61bd8c2

I tested Firefox and Safari too and only noticed that Safari only shows the press styling when I click and hold on the select opener. I think this is fine since it's based on browser implementation, and it still uses the expected press state styling. When I tab to it and hold down enter, it'll show both the press + focus styling: 



https://github.com/user-attachments/assets/a7ef5ce4-1800-4af7-9bd2-e793e1d16dd0

packages/wonder-blocks-dropdown/src/components/select-opener.tsx
beaesguerra: Reused the `placeholder` style!

PR #3085, opened 2026-05-29
[CLASS-14252] Remove the RTL lint checks that mostly cause false positives

maddy531:
## Summary

`require-logical-properties-for-rtl` originally shipped with six opt-in value-heuristic flags (`warnDirectionalTransforms`, `warnShadows`, `warnGradients`, `warnCursorDirections`, `warnBackgroundPositionXY`, `warnBackgroundPosition`) so we could A/B which ones actually caught RTL bugs. While escalating the rule from `warn` to `error` in the `frontend` repo (CLASS-14252), it became clear that most produced almost entirely false positives on RTL-safe code.

This PR removes the option flags, drops the checks that were pure noise, **narrows** the `translateX` check (so it keeps catching real bugs without the false positives), makes the `backgroundPosition` check unconditional, and tightens the `padding`/`margin`/`float`/`clear` auto-fixes. Net effect for consumers: far fewer spurious `eslint-disable` suppressions, so the rule can move to `error`.

## What changed

**Removed the low-signal checks** — directional `transformOrigin`, `boxShadow`/`textShadow`, `linear-gradient` direction, `cursor` resize, and `backgroundPositionX/Y`. None had a logical-property fix, and the heuristics structurally couldn't tell RTL-safe code (symmetric gradients, X-offset shadows already paired for RTL, block-axis `backgroundPositionY`) from a genuine bug. They're removed outright rather than left as disabled-by-default options, so no dead, un-fixable checks remain.

**Narrowed the `translateX` check rather than removing it** (per review feedback — it caught a real RTL bug before). The old check flagged *every* `translateX(...)`, which is what fired on the RTL-safe `translateX(-50%)` / `translateX(50%)` self-centering idiom and produced all of that idiom's false positives. It now **exempts `±50%` centering** (and skips `calc()`/`var()` values it can't analyze) while **still flagging px and non-±50% percentage offsets** — those are genuinely directional and don't mirror in RTL. Report-only, since there's no logical replacement.

**The rule now takes no options at all.** All flags are gone, including `warnBackgroundPosition` — the `backgroundPosition`/`background` directional check it used to gate is now unconditional, alongside `float`/`clear`/`direction`. In every audited case it only caught real bugs, so a config-level off-switch wasn't justified; per-line `eslint-disable` remains the escape hatch.

**Extended the `padding`/`margin` shorthand check** to 3-value (`A B C`) and 4-value (`A B C D`) forms, with autofix (was 2-value only). The 4-value form is the one that genuinely breaks in RTL — its left/right values differ and don't auto-mirror — and was previously not caught at all. Values containing functions (`calc()`, `var()`) are left alone since they can't be split safely.

**Added autofix for `float: left/right` and `clear: left/right`** → `inline-start`/`inline-end` (previously report-only). Those flow-relative values are now Baseline / widely available (Chrome/Edge 118, Safari 15, Firefox 55), so the fix the rule already recommended can be applied automatically.

**Kept and still enforced:** the property-name → logical-property fixes (`marginLeft` → `marginInlineStart`, etc., with autofix), the `textAlign: left/right` → `start/end` value swap, `float`/`clear`, `direction`, the `backgroundPosition`/`background` directional detection, and the narrowed `translateX` check.

## ⚠️ Breaking change (major)

The rule's schema previously used `additionalProperties: false`. Now that the schema is empty, any config that passes **any** option — `warnDirectionalTransforms`, `warnShadows`, `warnGradients`, `warnCursorDirections`, `warnBackgroundPositionXY`, or `warnBackgroundPosition` — will now error and must drop it. The `recommended` config enables the rule as a bare `"error"`.

## Evidence

Point-in-time audit of the four KA `frontend` PRs that were escalating this rule to `error` (CLASS-14252, since closed and consolidated into a single flip PR). They introduced **25 new `eslint-disable` comments** under the *old, broad* checks:

| Cause | Disables | Real RTL bug? |
|---|---|---|
| `translateX(-50%)` centering / counter-rotated orbit | 18 | ❌ No |
| Symmetric `transparent→white→transparent` gradient | 2 | ❌ No |
| X-offset `boxShadow` (paired LTR/RTL) | 2 | ❌ No |
| `backgroundPositionY: bottom` (block axis) | 1 | ❌ No |
| `backgroundPosition: bottom right/left` | 2 | ✅ Yes — caught by the check we keep |

All 18 translateX disables were exactly `translateX(-50%)` — the centering idiom. That's the key data point behind narrowing rather than removing: the **narrowed** check exempts `±50%`, so it would have flagged **0 of those 18** (no false-positive regression) while still catching a genuinely-directional `translateX(10px)`.

## Risk

For the fully-removed checks (`transformOrigin`, shadows, gradients, cursor, `backgroundPositionX/Y`), a genuinely directional value would no longer be linted — but across the 25 audited cases, zero were genuine, so the risk is empirically negligible. Directional `translateX` and `backgroundPosition` both stay covered.

## Test plan

- [ ] Rule unit tests cover: the removed options, the unconditional `backgroundPosition` check, `float`/`clear` autofix, the 2/3/4-value `padding`/`margin` shorthand expansion, and the narrowed `translateX` check (`translateX(-50%)`/`(50%)`/`calc(...)` pass; `translateX(10px)`/`(30%)`/`(-20px)` error). See `require-logical-properties-for-rtl.test.ts`.
- [ ] No test snapshots the `recommended` config flags (verified — only the rule's own test file references these names).
- [ ] Consumers see fewer `require-logical-properties-for-rtl` errors after upgrading.

Part of CLASS-14252.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


maddy531:
I wouldn't expect any UI changes, but seeing that I triggered one. Looking into it now.

packages/eslint-plugin-wonder-blocks/docs/require-logical-properties-for-rtl.md
```diff
@@ -62,42 +62,27 @@ StyleSheet.create({
...
 
 | Property | Warning | Default |
 | --- | --- | --- |
-| `float: "left"` / `"right"` | Use `inline-start` / `inline-end` | always on |
-| `clear: "left"` / `"right"` | Use `inline-start` / `inline-end` | always on |
 | `direction: "ltr"` / `"rtl"` | Don't force direction in styles; rely on container `dir` | always on |
-| `backgroundPosition` containing `left`/`right` | Use percentages or conditional assets | option `warnBackgroundPosition` |
-| `backgroundPositionX/Y` with directional keywords | Same | option `warnBackgroundPositionXY` |
-| `linear-gradient(to left/right/top/bottom, ...)` | Gradients are physical; verify or flip | option `warnGradients` |
-| `transform: "translateX(...)"` | translateX is directional | option `warnDirectionalTransforms` |
```
beaesguerra, line 80: I'm wondering about removing the check for `translateX`, since that part of the lint rule caught an important bug last time in Wonder Blocks! 

In the PR description, it included: 
|Cause | Disables | Real RTL bug?|
|-- | -- | --|
|translateX(-50%) centering / counter-rotated orbit | 18 | ❌ No|

Do you have more details around those cases? I'm assuming that they weren't a real RTL bug because it was translated 50%, which wouldn't be affected by rtl. But if it were to be a px value or a percent less than or greater than 50% it might be a RTL issue!

What do you think?

packages/eslint-plugin-wonder-blocks/docs/require-logical-properties-for-rtl.md
```diff
@@ -62,42 +62,27 @@ StyleSheet.create({
...
-        warnBackgroundPosition: true,        // default: true
-        warnBackgroundPositionXY: false,     // default: false
-        warnGradients: false,                // default: false
-        warnDirectionalTransforms: false,    // default: false
-        warnShadows: false,                  // default: false
-        warnCursorDirections: false,         // default: false
-    }],
-}
-```
+## Options
```
beaesguerra: I think we can remove the Options section now since there are none!

packages/eslint-plugin-wonder-blocks/src/configs/recommended.ts
```diff
@@ -2,10 +2,7 @@ export default {
     plugins: ["@khanacademy/wonder-blocks"],
     rules: {
         "@khanacademy/wonder-blocks/no-invalid-bodytext-children": "error",
-        // require-logical-properties-for-rtl is available in the plugin but
-        // intentionally not enabled in the recommended config. It was released
-        // without a timely integration plan for consumers and caused widespread
-        // failures. It will be re-enabled here once a migration path is ready.
-        // To opt in now: "@khanacademy/wonder-blocks/require-logical-properties-for-rtl": "error"
+        "@khanacademy/wonder-blocks/require-logical-properties-for-rtl":
+            "error",
```
beaesguerra, line 6: I tagged Marcy in our slack thread too, but cc'ing @marcysutton here again in case you have any concerns with including this rule in the plugin now!

beaesguerra (commented):
Overall, looks good! I wanted to get your thoughts around `translateX` before approving, and tagged Marcy for visibility!

packages/eslint-plugin-wonder-blocks/docs/require-logical-properties-for-rtl.md
maddy531: Oh I forgot about that, good point! Maybe this rule has to be narrowed rather than removed. I'll look into that now, excellent suggestion!

maddy531: Updated!

beaesguerra (approved):
Looks great, thank you for working on this @maddy531 !

PR #3086, opened 2026-05-29
Add snapshots for Combobox

beaesguerra:
## Summary:

Adding visual test coverage for Combobox. Will address the styling and add `syl-dark` coverage in a follow up PR!

To avoid new a11y violations in Storybook for the new stories, I also added support for an `aria-label` prop for the Combobox component so we can use it in the StateSheet. 

Note: There are some styling issues with different states. I'll fix the color related ones in the next PR, and created WB-2359 for other improvements (in the post clean up epic. Combobox is not used in `frontend` currently so it is lower priority currently!). Combobox also currently provides state styles programmatically so focus styles will not be shown in the StateSheet snapshot for now

Issue: WB-2166

## Test plan:

Review Chromatic snapshots. Confirm Combobox can receive an aria-label

### DIFF
--- a/.changeset/smooth-goats-beam.md
+++ b/.changeset/smooth-goats-beam.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-dropdown": minor
+---
+
+Combobox: Add support for `aria-label`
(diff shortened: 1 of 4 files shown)


jandrade (approved):
Looks great! thanks :shipit:

PR #3092, opened 2026-06-01
SYL Dark - Breadcrumbs

beaesguerra:
## Summary:

Reviewing the Breadcrumbs component in SYL Dark to confirm it looks as expected, matches the design specs, and has no a11y color violations. We also enable Chromatic snapshots for the syl-dark theme for the component.

- Use semanticColor token for separator styling to support syl-dark

Note: Created WB-2363 to fix the existing focus outline issue in all themes

Issue: WB-2166

## Test plan:

Review `?path=/story/packages-breadcrumbs-testing-snapshots-breadcrumbs--state-sheet-story&globals=theme:syl-dark` and confirm things look as expected and there are no a11y violations

### DIFF
--- a/__docs__/wonder-blocks-breadcrumbs/breadcrumbs.stories.tsx
+++ b/__docs__/wonder-blocks-breadcrumbs/breadcrumbs.stories.tsx
@@ -23,6 +23,10 @@ const meta: Meta<typeof Breadcrumbs> = {
                 version={packageConfig.version}
             />
         ),
+        chromatic: {
+            // Disable snapshots since they're covered by the testing snapshots
+            disableSnapshot: true,
+        },
     },
 };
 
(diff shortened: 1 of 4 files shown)


__docs__/wonder-blocks-breadcrumbs/breadcrumbs.stories.tsx
```diff
@@ -23,6 +23,10 @@ const meta: Meta<typeof Breadcrumbs> = {
                 version={packageConfig.version}
             />
         ),
+        chromatic: {
+            // Disable snapshots since they're covered by the testing snapshots
+            disableSnapshot: true,
+        },
```
jandrade, line 29: **praise:** Thanks for adding this!

jandrade (approved):
Looks good! it makes sense to follow up on a separate PR to fix the focus issue. 🚀

PR #3093, opened 2026-06-01
Cap storybook Vitest browser workers to avoid CI OOM

jandrade:
## Summary

The `Run Storybook tests` CI job (`vitest --project=storybook`, browser mode via Playwright/Chromium) has been **OOM-killed** (`exit 137` / `Killed`) partway through, after the process accumulates memory across hundreds of story files. Every test that ran *passed* — there was no assertion failure; the OS killed the process.

The failure is **reproducible** and not branch-specific: it reproduces whenever a bit more code lands (e.g. a `main` merge) because the run sits right at the self-hosted runner's memory ceiling. Recent history on `deploy/css-modules` alternates pass/fail with both failures on "merge main" commits, and unrelated branches OOM'd in the same window.

## Fix

Cap the `storybook` Vitest project to `maxWorkers: 2`. Browser mode spawns one Chromium context per worker (defaulting to ~CPU count), so bounding workers bounds peak memory with headroom. This trades a little wall-clock time for reliable CI.

This is a CI/tooling-only change — no impact on published packages (empty changeset).

## Context

Investigated while looking at the failing job on [#3073](https://github.com/Khan/wonder-blocks/pull/3073) (run [26774819028](https://github.com/Khan/wonder-blocks/actions/runs/26774819028)). A re-run failed identically, confirming it's a memory-ceiling issue rather than transient flakiness. Landing on `main` so every branch benefits. `maxWorkers` can be tuned up later if CI hosts gain memory.

## Test plan

- [ ] CI `Run Storybook tests` passes (no `exit 137`).

### DIFF
--- a/.changeset/cap-storybook-vitest-workers.md
+++ b/.changeset/cap-storybook-vitest-workers.md
@@ -1,0 +1,7 @@
+---
+---
+
+Cap the storybook Vitest browser project to `maxWorkers: 2` so the CI run
+stays under the runner's memory limit. The browser-mode suite had grown
+large enough to OOM-kill the runner (exit 137) at the default worker count.
+This is a CI/tooling-only change with no impact on published packages.
--- a/.changeset/cap-storybook-vitest-workers.md
+++ b/.changeset/cap-storybook-vitest-workers.md
@@ -1,7 +1,10 @@
 ---
 ---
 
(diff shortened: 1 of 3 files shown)


beaesguerra (approved):
Looks good to me!

jeresig (approved):

PR #3095, opened 2026-06-01
SYL Dark - Toolbar

beaesguerra:
## Summary:

Reviewing the Toolbar component in SYL Dark to confirm it looks as expected, matches the design specs, and has no a11y color violations. We also enable Chromatic snapshots for the syl-dark theme for the component.

- Use semantic color tokens for the dark variant (note: support for the dark variant is to be removed in WB-1852)

Issue: WB-2166

Design: https://www.figma.com/design/04ptAsL2PE4e8KGfYewta3/bea-syl-dark?node-id=8389-10552&t=48QoihVmhYQeteLi-4 (note there isn't a 1:1 mapping with Figma currently so for now, we make sure the current Toolbar use cases look okay in dark mode)

## Test plan:

Review `?path=/story/packages-toolbar-testing-snapshots-toolbar--scenarios&globals=theme:syl-dark` and confirm things look as expected and there are no a11y violations

### DIFF
--- a/__docs__/wonder-blocks-toolbar/toolbar.stories.tsx
+++ b/__docs__/wonder-blocks-toolbar/toolbar.stories.tsx
@@ -37,6 +37,10 @@ export default {
                 excludeDecorators: true,
             },
         },
+        chromatic: {
+            // Disable snapshots because it is covered by testing snapshots
+            disableSnapshot: true,
+        },
     },
     decorators: [
         (Story: any): React.ReactElement<React.ComponentProps<typeof View>> => (
@@ -189,6 +193,7 @@ Responsive.parameters = {
             medium: allModes.medium,
             large: allModes.large,
         },
+        disableSnapshot: false,
     },
 };
 
--- a/packages/wonder-blocks-toolbar/src/components/toolbar.tsx
+++ b/packages/wonder-blocks-toolbar/src/components/toolbar.tsx
@@ -5,7 +5,7 @@ import * as React from "react";
 import {StyleSheet} from "aphrodite";
 
 import {View} from "@khanacademy/wonder-blocks-core";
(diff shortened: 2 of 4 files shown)


packages/wonder-blocks-toolbar/src/components/toolbar.tsx
```diff
@@ -150,8 +150,8 @@ const sharedStyles = StyleSheet.create({
     },
     // TODO(WB-1852): Remove light variant.
     dark: {
-        background: color.darkBlue,
-        boxShadow: `0 1px 0 0 ${color.white64}`,
+        background: semanticColor.core.background.base.strong,
```
beaesguerra, line 153: This is `darkBlue` in the `default` theme still. This changes to `black_100` in `thunderblocks`.

It looks like there is only 1 remaining dark variant of the toolbar in `frontend` and it's used in the classic experience so this change should be okay. Let me know if there are any concerns with this! 

(cc: @jandrade in case you have some additional context on this, since you worked on color migrations!)

__docs__/wonder-blocks-toolbar/toolbar.stories.tsx
```diff
@@ -189,6 +193,7 @@ Responsive.parameters = {
             medium: allModes.medium,
             large: allModes.large,
         },
+        disableSnapshot: false,
```
jandrade, line 196: **praise:** Good call!

packages/wonder-blocks-toolbar/src/components/toolbar.tsx
jandrade: That makes total sense to me! I'm suspecting the only call site is the Dev toolbar, so it should be totally fine to make this change!

jandrade (approved):
:shipit:

PR #3096, opened 2026-06-01
Add StateSheets for styles package

beaesguerra:
## Summary:

Adding StateSheets for the styles package so we can have visual regression coverage for the shared styles.

The focus styles look fine in syl-dark, so no further changes will be made to that! 

The `actionStyles.inverse` has known issues in `syl-dark` when used with IconButton or Button. Since these styles are mostly used to handle legacy `light=true` patterns in the `default` theme, we will leave these as is. Long term, we want to remove support for these action styles. 

Issue: WB-2166

## Test plan:

Review action styles StateSheet and confirm it covers the same use cases as before
- `?path=/story/packages-styles-action-styles-testing-snapshots-action-styles--state-sheet-story`

Review Chromatic snapshots


__docs__/wonder-blocks-styles/action-styles-variants.stories.tsx
```diff
@@ -1,268 +0,0 @@
...
-};
-
-export const Press: StoryComponentType = {
-    parameters: {pseudo: {active: true}},
-};
-
-export const PressFocus: StoryComponentType = {
-    name: "Press + Focus",
-    parameters: {pseudo: {active: true, focusVisible: true}},
-};
```
beaesguerra, line 268: Combining these into a StateSheet so we can enable all themes for these different states

jandrade (approved):
LGTM! thanks for refactoring the stories to follow our guidelines! 🚢

PR #3101, opened 2026-06-04
SYL Dark - Combobox 

beaesguerra:
## Summary:

Reviewing the Combobox components in SYL Dark to confirm it looks okay and has no a11y color violations. We also enable Chromatic snapshots for the syl-dark theme for this component.

Changes include:
- Removing state styling for the internal IconButton and TextField components used 
- Note: I've created WB-2359 for improving the styles of Combobox (ie. using pseudoclasses for states, handling edge cases, etc). This is lower priority right now since these are pre-existing in all themes and Combobox isn't used in frontend yet! 

Issue: WB-2166

## Test plan:

Review the Combobox statesheet and confirm the additional state styles for the opener IconButton and TextField are no longer visible: `?path=/story/packages-dropdown-testing-snapshots-combobox--state-sheet-story`

Confirm the Combobox as a whole looks the same as before when interacting with it manually

### DIFF
--- a/.changeset/silent-mugs-guess.md
+++ b/.changeset/silent-mugs-guess.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-dropdown": patch
+---
+
+Combobox: Remove state styling for the dropdown button and the internal text field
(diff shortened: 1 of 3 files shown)


marcysutton (approved):
This looks good to me! Thanks for adding Combobox snapshots!

PR #3107, opened 2026-06-05
ActivityButton: Fix initial animation on Chrome

MikeKlemarewski:
## Summary:
Because Aphrodite lazily injects CSS, there could be a race in Chrome when applying the pressed styles for the first time and the animation that gets applied.

This adds the pressed styles to the main button styles so they're applied on first render. The pressed styles are now applied based on a data attributed that gets applied when the button is pressed.

https://github.com/user-attachments/assets/9d851371-0509-4b0d-8a3b-d0c7a45b47ce

## Test plan:
- pnpm start
- Open storybook in Chrome
- View the Activity Button story
- Reload the page
- Click the button and make sure it animates normally (without any flashing)

### DIFF
--- a/.changeset/breezy-deers-heal.md
+++ b/.changeset/breezy-deers-heal.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-button": patch
+---
+
+Update how pressed styles are applied to ActivityButton to avoid initial flash on Chrome.
--- a/packages/wonder-blocks-button/src/components/activity-button.tsx
+++ b/packages/wonder-blocks-button/src/components/activity-button.tsx
@@ -51,7 +51,6 @@ const ActivityButtonCore: React.ForwardRefExoticComponent<
     const sharedStyles = [
         buttonStyles.button,
         disabled && buttonStyles.disabled,
-        !disabled && pressed && buttonStyles.pressed,
         // Enables programmatic focus.
         !disabled && !pressed && focused && buttonStyles.focused,
         stylesProp?.root,
@@ -60,7 +59,6 @@ const ActivityButtonCore: React.ForwardRefExoticComponent<
     const chonkyStyles = [
         buttonStyles.chonky,
         disabled && buttonStyles.chonkyDisabled,
-        !disabled && pressed && buttonStyles.chonkyPressed,
         stylesProp?.box,
     ];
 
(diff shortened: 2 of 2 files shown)


marcysutton:
@claude review once

marcysutton (commented):
Interesting approach with `data-state`! The press action does look better.

One thing I'm seeing that we could potentially fix at the same time: the initial button shape loads in from the top left corner in a similar way. Do you see that too?

We'll want @jandrade to weigh in since he built this component and is doing the refactor to CSS Modules.

jandrade:
> Interesting approach with `data-state`! The press action does look better.
> 
> One thing I'm seeing that we could potentially fix at the same time: the initial button shape loads in from the top left corner in a similar way. Do you see that too?
> 
> We'll want @jandrade to weigh in since he built this component and is doing the refactor to CSS Modules.

I agree with @marcysutton's take. One thing that it would be important to state is that this would be a temporary fix while I get the CSS modules migration done as I would expect to avoid resorting to these selectors for this particular situation (as it is something specific to runtime CSS / Aphrodite).

@MikeKlemarewski If you end up landing this PR, could you please add a TODO comment pointing to this ticket https://khanacademy.atlassian.net/browse/WB-2328?. That way I can ensure to remove this custom styling once we migrate `ActivityButton` to CSS modules.

I would expect to release WB + CSS modules fully converted before the end of the month, so landing this PR would be a good temporary fix if you want to get this fixed before the migration is done.

MikeKlemarewski:
> > Interesting approach with `data-state`! The press action does look better.
> > One thing I'm seeing that we could potentially fix at the same time: the initial button shape loads in from the top left corner in a similar way. Do you see that too?
> > We'll want @jandrade to weigh in since he built this component and is doing the refactor to CSS Modules.
> 
> I agree with @marcysutton's take. One thing that it would be important to state is that this would be a temporary fix while I get the CSS modules migration done as I would expect to avoid resorting to these selectors for this particular situation (as it is something specific to runtime CSS / Aphrodite).
> 
> @MikeKlemarewski If you end up landing this PR, could you please add a TODO comment pointing to this ticket https://khanacademy.atlassian.net/browse/WB-2328?. That way I can ensure to remove this custom styling once we migrate `ActivityButton` to CSS modules.
> 
> I would expect to release WB + CSS modules fully converted before the end of the month, so landing this PR would be a good temporary fix if you want to get this fixed before the migration is done.

Sure thing. I'll add the comment and land it. I think it'll be good to have especially just incase the migration to CSS modules doesn't fix the issue.

marcysutton (approved):
Thanks @MikeKlemarewski!

PR #3108, opened 2026-06-08
🚨 [WB-2330] Phase 5 / Wave B — Migrate core View & Text to CSS Modules

jandrade:
## Summary:

Phase 5 / Wave B of the CSS Modules migration (WB-2330), scoped to
`wonder-blocks-core`, plus the `processStyleList` and tooling changes that
fell out of it.

**1. `View` / `Text` → CSS Modules.** Their internal Aphrodite
`StyleSheet.create` blocks move to colocated `view.module.css` /
`text.module.css`, authored inside the shared `@layer shared` cascade layer.
The class names route through the existing `addStyle` / `processStyleList`
pipeline, so consumer-provided Aphrodite styles, inline styles, and
`className` props continue to compose. Public API is unchanged.

**2. `processStyleList` — always wrap inline styles into a generated class.**
The Phase-1 fast-path (forwarding inline styles directly as a `style`
attribute when no Aphrodite was present) is removed. Inline-object styles are
now always merged and wrapped into a single generated Aphrodite class so they
keep outweighing Aphrodite's `!important`; `style` comes back empty (except in
snapshot mode, where everything is flattened into `style`). The className join
also now drops empty segments, so a class-only merge no longer emits a leading
space (e.g. `View` rendering `class=" default"`).

⚠️ **Precedence note:** because Aphrodite still emits `!important`, a
consumer-supplied Aphrodite declaration now wins over a conflicting inline
`style` value. Post-migration the only Aphrodite left in a typical merge is
consumer-supplied, so this trade-off is acceptable.

**3. Jest config.** Excludes the `.claude/worktrees/` repo copies from the
Haste map and from `--watch` (fixing module-name collisions and watch re-run
loops), and moves `prettierPath: null` into the per-project config (fixing the
Prettier 3 inline-snapshot error). Tooling-only.

Reviewers: #wonder-blocks

Issue: WB-2330

## Test plan:

All changes verified automatically:

- `pnpm jest packages/wonder-blocks-core` — full core suite passes
  (`util`, `add-style`, `view`, `text`).
- Downstream consumers pass: `wonder-blocks-modal`, `-layout`,
  `-typography`, `-clickable`, `-tooltip`.
- `pnpm typecheck`, `stylelint`, and `eslint` pass on the changed files.
- Jest config fixes verified directly: tests run without the
  `--modulePathIgnorePatterns` flag, an inline snapshot writes without the
  Prettier error, and `--showConfig` confirms `watchPathIgnorePatterns` is
  applied per project.

Visual parity (both themes) is covered by Chromatic in CI. Optional human
smoke check: open any component composing `View`/`Text` in Storybook and
confirm layout/typography is unchanged.

## Review plan:

1. 🚨 [`view.tsx`](https://github.com/Khan/wonder-blocks/pull/3108#discussion_r3375050752)
2. 🚨 [`text.tsx`](https://github.com/Khan/wonder-blocks/pull/3108#discussion_r3375050994)
3. 🚨 [`util.ts` (processStyleList)](https://github.com/Khan/wonder-blocks/pull/3108#discussion_r3382003666)
4. 🚨 [`config/jest/test.config.js`](https://github.com/Khan/wonder-blocks/pull/3108#discussion_r3382003949)

### Common patterns:

**2 Files:** Replace an Aphrodite `StyleSheet.create` block with a colocated CSS Module import; the resulting class-name string flows through the existing `addStyle`/`processStyleList` pipeline unchanged.

```diff
-import {StyleSheet} from "aphrodite";
-const styles = StyleSheet.create({
-    default: {
-        alignItems: "stretch",
-        display: "flex",
-        flexDirection: "column",
-        // ...
-    },
-});
+import styles from "./view.module.css";
 // styles.default is now a CSS Module class name (string), composed the
 // same way: addStyle(tag, styles.default) / processStyleList([...]).
```

**4 Test files:** Update assertions/snapshots to the new `processStyleList` and CSS-Module output — inline styles are wrapped into a generated class (`style` empty), and `View`/`Text` render a CSS-Module class instead of inlined Aphrodite styles.

```diff
-    class=""
-    style="...; display: flex; flex-direction: column; ...; width: 40px; ..."
+    class="default"
+    style="position: relative; pointer-events: none; ...; width: 40px; ..."
```


packages/wonder-blocks-core/src/components/view.tsx
```diff
@@ -1,30 +1,11 @@
-// WARNING: If you modify this file you must update view.js.flow.
 import * as React from "react";
```
jandrade, line 1: 🚨 **1 of 4:** `View` is the most widely consumed primitive in the design system; this swaps its default-style delivery from Aphrodite to a CSS Module class (`view.module.css`). Computed styles are identical and the change is fully tested (core suite + downstream layout/typography/clickable), but the blast radius is repo-wide. cc: @Khan/wonder-blocks

packages/wonder-blocks-core/src/components/text.tsx
```diff
@@ -1,33 +1,17 @@
-// WARNING: If you modify this file you must update text.js.flow.
 import * as React from "react";
```
jandrade, line 1: 🚨 **2 of 4:** Same as `View` for `Text`: the antialiasing / header-margin styles now come from a CSS Module (`text.module.css`) routed through the existing `processStyleList` pipeline. Public API and computed styles are unchanged. cc: @Khan/wonder-blocks

packages/wonder-blocks-core/src/util/util.ts
```diff
@@ -131,10 +123,13 @@ export function processStyleList(style?: StyleType): StyledExport {
...
 
+    // Join the aphrodite-generated class (empty when no aphrodite is present)
+    // with the CSS Module / plain class-name strings, dropping empties so we
+    // never emit a leading/trailing space.
     const aphroditeClassName = css(...aphroditeStyles);
-    const finalClassName = joinedClassNames
-        ? `${aphroditeClassName} ${joinedClassNames}`
-        : aphroditeClassName;
+    const finalClassName = [aphroditeClassName, joinedClassNames]
+        .filter(Boolean)
```
jandrade, line 131: 🚨 **3 of 4:** `processStyleList` is the shared styling pipeline behind every WB component that uses `addStyle` / `View` / `Text`. This removes the fast-path so inline-object styles are now always merged and wrapped into a generated Aphrodite class (`style` returns `{}`), and fixes the className join to drop empty segments (avoids a leading space like `class=" default"`). Behavior change with repo-wide reach — please sanity-check the precedence note: a consumer Aphrodite `!important` declaration now wins over a conflicting inline `style` value. cc: @Khan/wonder-blocks

config/jest/test.config.js
```diff
@@ -39,6 +39,20 @@ const projectConfig = {
         SNAPSHOT_INLINE_APHRODITE: true,
     },
     testMatch: ["<rootDir>/**/*.test.ts", "<rootDir>/**/*.test.tsx"],
+    // Ignore repo copies created under `.claude/worktrees/` (by Claude Code).
+    // Each worktree has its own `packages/**/package.json`, whose duplicate
+    // `name` fields otherwise collide in Jest's Haste module map ("The name
+    // `@khanacademy/wonder-blocks-core` was looked up in the Haste module
+    // map").
+    modulePathIgnorePatterns: ["/\\.claude/"],
```
jandrade, line 47: 🚨 **4 of 4:** Global Jest config. Two changes, all moved into the shared `projectConfig` so they apply per-project: (1) `modulePathIgnorePatterns`  excludes the `.claude/worktrees/` repo copies — fixing Haste module-name collisions; (2) `prettierPath: null` is set here (a project-level option, so the existing root-level copy was a no-op) to fix "Inline Snapshots are not supported when using Prettier 3.0.0 or above". Global config change — flagging for visibility; no runtime/published impact. cc: @Khan/wonder-blocks

packages/wonder-blocks-tooltip/src/components/__tests__/tooltip-tail.test.tsx
jandrade, line 1: **note:** Updated inline snapshot tests to use the new CSS mock used with CSS modules in Jest.

packages/wonder-blocks-core/src/util/__tests__/util.test.ts
jandrade, line 1: **note:** Updated tests to reflect the recent `processStyleList` change.

packages/wonder-blocks-core/src/util/util.ts
```diff
@@ -110,15 +111,6 @@ export function processStyleList(style?: StyleType): StyledExport {
...
 
-    // Fast-path: nothing to compete with Aphrodite's `!important`, so we can
-    // forward inline styles directly as `style` and skip the generated class.
-    if (aphroditeStyles.length === 0) {
-        return {
-            style: inlineStylesObject,
-            className: joinedClassNames,
-        };
-    }
-
```
jandrade, line 121: **note:** Removed this logic as I noticed that we needed to wrap/merge inline styles with Aphrodite in certain cases, otherwise this change would break consumers that tried to include inline style objects that used Aphrodite syntax to work.

Here's an example that follows that convention: https://github.com/Khan/wonder-blocks/blob/73c82e692d6abe1f1fb63cf7dc79502d65762181/__docs__/wonder-blocks-modal/modal-panel.stories.tsx#L250-L259

You can see how this could break styles in consumers: https://www.chromatic.com/test?appId=5e1bf4b385e3fb0020b7073c&id=6a26f8b90b20662bdccede1a

I've verified and `frontend` uses this approach in a lot of places, so that's why I'm removing this logic that I included in the previous PR.

beaesguerra:
@claude review once

beaesguerra (approved):
Looks good to me!

PR #3109, opened 2026-06-08
SYL Dark - DatePicker and BirthdayPicker

beaesguerra:
## Summary:

Reviewing the DatePicker and BirthdayPicker components in SYL Dark to confirm they look as expected, and have no a11y color violations. We also add StateSheet and Scenario stories and enable Chromatic snapshots for the syl-dark theme for both components, and move snapshot coverage to StateSheet and Scenarios stories

No specific changes were made to the components themselves. 

Issue: WB-2166
Design: No specific Figma component for these components, but it should look similar to the other [form fields](https://www.figma.com/design/04ptAsL2PE4e8KGfYewta3/bea-syl-dark?node-id=5302-354&m=dev) 

## Test plan:

Review the following and confirm things look as expected and there are no a11y violations:
- DatePicker StateSheet: `?path=/story/packages-datepicker-testing-snapshots-datepicker--state-sheet-story&globals=theme:syl-dark`
- BirthdayPicker StateSheet: `?path=/story/packages-birthdaypicker-testing-snapshots-birthdaypicker--state-sheet-story&globals=theme:syl-dark`


marcysutton (approved):
Looks good to me! I left one question about the icon overlap on the date picker input in Chromatic.

beaesguerra:
> Looks good to me! I left one question about the icon overlap on the date picker input in Chromatic.

@marcysutton Good question! The size of the input does depend on the size of the container! We could increase the size of the container, but I think we would still want to handle the case if the container isn't large enough for the field contents (especially with translated dates in mind!) 

I think adding padding to the end of the field to account for the icon and edge would help: 
image

This is a consideration for long placeholder text too! I added this detail to WB-2367 as well!
image

PR #3116, opened 2026-06-11
Add Helvetica and Corbel fallbacks to default sans font token

beaesguerra:
## Summary

Updates the `default` theme's `font.family.sans` token to include additional system font fallbacks.

**Before:**
```
Lato, "Noto Sans", sans-serif
```

**After:**
```
"Lato", "Noto Sans", "Helvetica", "Corbel", sans-serif
```

###  Why

Adds `Helvetica` and `Corbel` as fallback fonts for parity with the [current font value used in frontend](https://github.com/Khan/frontend/blob/b1193413e526b4b39a22646a41eaf8ad71749afd/libs/app-shell/src/styles/classic.css#L16). After this, we can use the css variable token for setting the body font!

Note: This only affects the `default` theme; the Thunderblocks `font.family.sans` token is unchanged. I created https://khanacademy.atlassian.net/browse/WB-2378 to re-evaluate fallback fonts if needed (for now, we are making this update for parity!) 

Issue: FEI-7881 

## Test Plan:

Confirm the `default` theme `font.family.sans` token is updated to include the fallback fonts. Confirm token is not updated in other themes

### DIFF
--- a/.changeset/fluffy-mangos-grow.md
+++ b/.changeset/fluffy-mangos-grow.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": patch
+---
+
+Add `Helvetica` and `Corbel` fallback fonts to the `default` theme's `font.family.sans` token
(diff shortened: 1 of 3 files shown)


jandrade (approved):
Looks great, thanks! 🚀

PR #3118, opened 2026-06-12
WB-2346: Deprecate opened prop in WB Modal to address focus management gotchas

marcysutton:
## Summary:
I've seen numerous PRs in `frontend` that either aim to address focus management issues or perpetuate them without realizing it. WB shouldn't allow a pattern that doesn't work for accessibility, and the `opened` prop is one of those. When WB Modals are conditionally rendered with the `opened` prop, the parent component unmounts and takes ModalLauncher with it. The component tree is detached before the focus management lifecycle can run, and no runtime code can fix it because it isn't mounted anymore.

This PR updates `ModalLauncher` and `DrawerLauncher` to support `modal={null}` as a first-class "closed" state, and deprecates the `opened` prop in favor of this pattern.

What changed:

- The `modal` prop now accepts `null` to indicate a closed state, keeping the launcher always mounted across open/close transitions
- The `opened` prop is deprecated with a runtime `console.warn` and `@deprecated` JSDoc, but remains fully functional — no existing consumers need to change code
- `DrawerLauncher` preserves exit animation content via a cached `ref` when modal transitions to `null`

Why: Conditionally rendering the launcher `({isOpen && <ModalLauncher>})` caused focus to be lost on unmount rather than returned to the triggering element. The new API avoids this pitfall by design.

Migration:

```jsx
// Deprecated (still works, but warns)
<ModalLauncher opened={isOpen} modal={<MyDialog />} onClose={...} />

// Recommended
<ModalLauncher modal={isOpen ? <MyDialog /> : null} onClose={...} />
```

Issue: WB-2346

## Test plan:
1. Ensure tests pass
2. Test stories for ModalLauncher and DrawerLauncher
3. Test integration in `frontend`

### DIFF
--- a/.changeset/forty-glasses-pretend.md
+++ b/.changeset/forty-glasses-pretend.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-modal": minor
+---
+
+Deprecate open prop in WB Modal to address focus management gotchas
--- a/__docs__/wonder-blocks-announcer/announcer.stories.tsx
+++ b/__docs__/wonder-blocks-announcer/announcer.stories.tsx
@@ -241,9 +241,8 @@ export const AnnouncerInModal: StoryComponentType = {
                     <Button onClick={handleOpen}>Open Modal to Test</Button>
                 )}
                 <ModalLauncher
-                    opened={isOpen}
                     onClose={handleClose}
-                    modal={ModalContent}
+                    modal={isOpen ? ModalContent : null}
                 />
             </View>
         );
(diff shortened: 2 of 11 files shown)


kevinb-khan:
@marcysutton and I had a good discussion about this in Slack: https://khanacademy.slack.com/archives/C4PE1QM5Y/p1781817885636749.

marcysutton:
The `opened` prop isn't the only way to conditionally render modal launchers: the `modal` prop could be used in the same way and cause focus management problems. I'm looking at lint rules on the consumer side instead, particularly around passing a hardcoded `opened={true}` to a launcher component. I'll close this PR as I don't think we want to remove `opened` after all.

PR #3122, opened 2026-06-16
ActivityIconButton: fit long translated labels (Body/Small + hyphenation)

jandrade:
## Summary:

Implements the `ActivityIconButton` label-fitting approach validated by the WB-2354 spike, so long translated labels fit in the limited width of the Classroom tools panel for the international markets launching July 6.

**Component (`activity-icon-button.tsx`)** — a `minor` change:
- The visible label now renders at **Body/Small** (`size="small"`).
- Label styles use **locale-aware hyphenation**: `hyphens: auto` with `wordBreak: normal` + `overflowWrap: break-word` (replacing `wordBreak: break-word`, which forced a hard break before hyphenation could kick in). Hyphenation follows the document/ancestor `lang`; long unbreakable words still fall back to breaking so nothing overflows the 2-line clamp.

**Stories** — the temporary spike file is replaced by a permanent regression snapshot folded into the existing ActivityIconButton testing snapshots: **`Label Fitting (Locale Matrix)`** renders the real component across the i18n team's Tier 1 launch reference set — 12 locales (incl. RTL Urdu/Sindhi and Indic scripts) × 9 Classroom-tools controls, with per-cell `lang`/`dir` and severity dots flagging the highest-risk strings. Captured as a single **ThunderBlocks** Chromatic snapshot (label fitting is layout, not theming). The severity dots are accessible: `role="img"` + `aria-label` so severity isn't color-only, plus a contrasting ring for non-text contrast.

This supersedes the original spike (which applied techniques via `styles.label` without touching the component and added a throwaway comparison story).

Reviewers: #wonder-blocks

Issue: WB-2354

## Test plan:

- `pnpm typecheck` — clean
- `pnpm lint` — clean
- `activity-icon-button.test.tsx` — passes
- Visual review (Chromatic): the locale matrix confirms the launch set stays within two lines without clipping. Story: `Packages / IconButton / Testing / Snapshots / ActivityIconButton` → **Label Fitting (Locale Matrix)**.

## Review plan:

🚨 This PR is **High risk** — it changes the rendered behavior of a shared, user-visible component. Please review this change (see inline comment):

1. 🚨 [`activity-icon-button.tsx`](https://github.com/Khan/wonder-blocks/pull/3122#discussion_r3462410792)

The story file is Low risk (new Chromatic snapshot, validated by story inclusion) and the changeset is Trivial, so they're omitted here. All changed files are covered by the #wonder-blocks reviewer, so no supplemental reviewers are needed.


packages/wonder-blocks-icon-button/src/components/activity-icon-button.tsx
```diff
@@ -139,7 +139,7 @@ export const ActivityIconButton: React.ForwardRefExoticComponent<
                 {hasVisibleLabel && (
                     <BodyText
                         tag="span"
-                        size="medium"
+                        size="small"
```
jandrade, line 142: 🚨 **1 of 1:** Behavior change to a shared, user-visible Wonder Blocks component. Every `ActivityIconButton` visible label now renders at **Body/Small** (was Body/Medium) and switches to locale-aware hyphenation (`hyphens: auto` + `wordBreak: normal` + `overflowWrap: break-word`, replacing `wordBreak: break-word`) — see the label style block further down. This changes the rendered text size and line-breaking of the label for every consumer across all apps, not just the i18n launch surface. There's no unit assertion on the visual output; coverage is the new `Label Fitting (Locale Matrix)` Chromatic snapshot, so please review the snapshot diff (and confirm the smaller label is intended for existing English usages) before merging.

kevinb-khan (approved):
I looked at the Chromatic changes at they seem reasonable to me.  This change appears to have fix the horizontal positioning of tooltips in one of the test cases.  Thanks for adding the `LabelFittingLocaleMatrix` as a stress test for different languages.

marcysutton (approved):
This looks good to me too! It will be interesting to see it integrated into frontend!

PR #3125, opened 2026-06-18
MultiSelect: announce correct count for selected items

marcysutton:
## Summary:

Fix screen reader announcements in MultiSelect and SingleSelect

Several live region announcement bugs are fixed, plus a VoiceOver/Safari workaround for stale combobox values:

- **Wrong item count announced on selection** — the MultiSelect component was announcing total visible options instead of selected count, caused by `labels` prop object identity churn re-firing the effect on every parent render
- **Neighboring MultiSelect announcing on unrelated interactions** — same root cause; inline label objects recreating on parent re-renders triggered announcement effects in unrelated selects on the same page
- **VoiceOver caching a stale combobox value on re-open** — both components make a `polite` announcement of the current selection when a non-filterable dropdown opens
- **Close announcement** — MultiSelecs makes a `polite` announcement of the final selection state when the dropdown closes after a selection change
- Equivalent fixes applied to SingleSelect where applicable. 
- New unit tests and a two-select playtesting story are included, to test the neighboring announcement issue.

Issue: WB-2383

## Test plan:
1. Ensure tests pass
2. Review stories with VoiceOver running: 
   - `/?path=/story/packages-dropdown-testing-snapshots-multiselect--scenarios&globals=theme:thunderblocks`
   - `/?path=/story/packages-dropdown-multiselect--two-multi-selects`
3. Compare to PROD

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/multi-select.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/multi-select.tsx
@@ -553,6 +553,7 @@ const MultiSelect = (props: Props) => {
     const handleAnnouncement = (message: string) => {
         announceMessage({
             message,
+            level: "assertive",
         });
     };
 
@@ -577,26 +578,38 @@ const MultiSelect = (props: Props) => {
         [selectedValues],
     );
 
+    // Refs so the announcement effect re-fires only on meaningful changes,
+    // not on every render caused by inline label object identity churn.
+    const childrenRef = React.useRef(children);
+    childrenRef.current = children;
+    const getMenuTextOrNodeRef = React.useRef(getMenuTextOrNode);
+    getMenuTextOrNodeRef.current = getMenuTextOrNode;
+    const maybeGetOpenerStringValueRef = React.useRef(
+        maybeGetOpenerStringValue,
+    );
+    maybeGetOpenerStringValueRef.current = maybeGetOpenerStringValue;
+
+    // Announces the opener value on open and after each selection change.
     React.useEffect(() => {
         if (isInitialRender.current) {
             isInitialRender.current = false;
             return;
         }
-
+        if (!open) {
+            return;
+        }
         const optionItems = React.Children.toArray(
-            children,
+            childrenRef.current,
         ) as OptionItemComponentArray;
-        const openerContent = getMenuTextOrNode(optionItems);
-        const openerStringValue = maybeGetOpenerStringValue(
+        const openerContent = getMenuTextOrNodeRef.current(optionItems);
+        const openerStringValue = maybeGetOpenerStringValueRef.current(
             optionItems,
             openerContent,
         );
-
         if (openerStringValue) {
-            // opener value changed, so let's announce it
             handleAnnouncement(openerStringValue);
         }
-    }, [children, getMenuTextOrNode, maybeGetOpenerStringValue]);
+    }, [open, selectedValues]);
 
     // If aria-required was supplied, use that. Otherwise, convert `required` to a boolean
     // and apply that value to aria-required.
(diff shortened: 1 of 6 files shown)


beaesguerra:
Thanks for looking into this @marcysutton ! I started with some initial testing and found I am able to reproduce the issue still if I select, de-select, and select the third item in a MultiSelect a few times (it reads out `1 selected` still instead of `2 selected`). Is this expected?

https://github.com/user-attachments/assets/fb4e5598-9bb8-4c2f-93a6-da32eb24f874


Also, do you know why there are changes detected in Chromatic? I don't see anything highlighted in the Chromatic diffs, so I'm curious what's causing that!

marcysutton:
Thanks for testing @beaesguerra! I went through this again yesterday and ultimately decided to go back to `polite` announcements. There were some inconsistent results when selecting and deselecting items where the `assertive` announcements were _sometimes_ beat out by a focus event and associated item announcement. So rather than the `assertive` content getting lost, I went back to `polite` for it to slot into the queue.

As far as Chromatic snapshots go, I think it's re-hashing CSS classes based on imports and source changes? It's not really obvious to me why.

> If no styles were added, the class name churn is more likely caused by a non-deterministic Aphrodite hash across builds — Aphrodite's class name generation can vary between build runs even with identical source, particularly if the bundle entry order shifts slightly (e.g. due to new imports being added, which changes module evaluation order). The new imports brought in by this PR (React.useRef, announceMessage, etc. — though those were likely already imported) could shift module initialization order enough to perturb Aphrodite's internal counter.

> The bottom line is the same: no visual change, safe to accept. The diff is build-order noise, not a real style regression.

beaesguerra:
@claude review once

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -577,26 +575,62 @@ const MultiSelect = (props: Props) => {
...
+    const prevOpenRef = React.useRef(open);
+    const selectionChangedRef = React.useRef(false);
+    const childrenRef = React.useRef(children);
+    childrenRef.current = children;
+    const getMenuTextOrNodeRef = React.useRef(getMenuTextOrNode);
+    getMenuTextOrNodeRef.current = getMenuTextOrNode;
+    const maybeGetOpenerStringValueRef = React.useRef(
+        maybeGetOpenerStringValue,
+    );
+    maybeGetOpenerStringValueRef.current = maybeGetOpenerStringValue;
```
beaesguerra, line 590: The `childrenRef`, `getMenuTextOrNodeRef` and `maybeGetOpenerStringValueRef` refs are set every render so they will stay up to date! Are refs needed for those still?

Since we don't include these refs in the dependency arrays in the hooks, do we need these as refs still?

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
```diff
@@ -577,26 +575,62 @@ const MultiSelect = (props: Props) => {
...
+            if (selectionChangedRef.current) {
+                announceOpenerValue();
+                selectionChangedRef.current = false;
+            }
+        } else if (open) {
+            // Selection changed while open — announce and track for close.
+            selectionChangedRef.current = true;
+            announceOpenerValue();
+        }
+    }, [selectedValues, open]); // eslint-disable-line react-hooks/exhaustive-deps
```
beaesguerra, line 636: For the eslint-disable-line react-hooks/exhaustive-deps comments, could we add a comment explaining why we explicitly only include certain deps in the array? Would be helpful for future us to know that it's intentional!

beaesguerra (approved):
Thanks @marcysutton for digging into this more! I tested it again with VoiceOver + Safari and confirmed that it addresses the reported issue (selecting/deselecting an option item announces the correct number of items selected now)!

https://github.com/user-attachments/assets/b133501f-8508-4373-ab0c-773f2009a00a

(Note: The recording shows the known VoiceOver issue where once the dropdown is closed, the previous value is announced before announcing the new value) 

I left some minor questions in the PR that are non-blocking. Looks good to me!

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: We do! Added a comment.

packages/wonder-blocks-dropdown/src/components/multi-select.tsx
marcysutton: Yes, added a comment!

PR #3126, opened 2026-06-24
Update media-query tokens to fix bugs and cover full range of widths

marcysutton:
## Summary:
The `wonder-blocks-tokens` breakpoint definitions had bugs introduced when ported from `frontend`, that this PR aims to fix.

- The `lg` query was defined using `mdMin` and `lgMin` instead of `lgMin` and `lgMax`, making it overlap with `md` rather than span the large breakpoint. 
- The `xl` query reused `lgMin` instead of `xlMin`, making it identical to `lgOrLarger`.
- The `lgOrSmaller` query also used `lgMin` as its upper bound, cutting off at the start of the large breakpoint instead of the end.
    - As a result, screens between `1024–1199px` were treated as larger than "lg or smaller," causing potential layout bugs in 19 call sites in `frontend`.
- `lgMax` and `xlMin` were never defined in the `width` constants, making the wrong values easy to reach for during the port.

Issue: WB-2384

## Test plan:
1. Ensure linting passes
2. Review mediaQuery docs: `?path=/docs/packages-tokens-media-query-breakpoints--docs`
3. Test usage in `frontend` (WIP)

### DIFF
--- a/.changeset/eighty-suns-taste.md
+++ b/.changeset/eighty-suns-taste.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": minor
+---
+
+Update media-query breakpoint and constant tokens to cover full range of widths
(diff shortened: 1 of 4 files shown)


beaesguerra (approved):
Looks good to me, thanks for the fixes!

PR #3130, opened 2026-06-25
FEI-7954: Enforce mutual exclusivity of location and initialEntries in router adapter

kevinb-khan:
## Summary

Updated the router test harness adapter to enforce mutual exclusivity between `location` and `initialEntries` configuration options at the type level, ensuring TypeScript 6.0 compatibility.

Issue: FEI-7954

## Key Changes

- **Router adapter types**: Added `NotInitialEntries` and `NotLocation` discriminator types to structurally enforce that `location` and `initialEntries` cannot be combined in the same config object
  - Applied `NotLocation` to the `initialEntries`-based union member
  - Applied `NotInitialEntries` to both `location`-based union members
  - This mirrors the existing `NotDataRoutes` pattern for data-routes exclusivity

- **Type tests**: Added two new test cases to verify the mutual exclusivity is enforced:
  - Rejects combining `location` with `initialEntries`
  - Rejects combining `forceStatic` location with `initialEntries`

- **TypeScript upgrade**: Updated `package.json` to use TypeScript 6.0.3 (from 5.7.3)

- **tsconfig-shared.json**: Added explicit `types` array configuration to handle TypeScript 6.0's changed default behavior for `@types/*` packages in composite builds
  - Lists all global-providing `@types/*` dependencies to ensure they remain in scope during declaration emission
  - Includes: aria-query, history, jest, jest-axe, jscodeshift, node, node-fetch, react, react-dom, react-router, react-router-dom, react-window

- **Test file updates**: Added `@ts-expect-error` comments in test files to suppress new TypeScript 6.0 type errors related to `window.location` assignment

## Implementation Details

TypeScript 6.0 relaxed union excess-property checking, which previously prevented combining incompatible union members. The structural `never` guards ensure type safety across both TypeScript 5.7 and 6.0 by making it impossible to satisfy both the positive requirement (e.g., `initialEntries: ...`) and the negative constraint (e.g., `location?: never`) simultaneously.

https://claude.ai/code/session_01CgbaANpEqVYPiEG9kiRjGU

### DIFF
--- a/consistency-tests/__tests__/clickables.test.tsx
+++ b/consistency-tests/__tests__/clickables.test.tsx
@@ -123,6 +123,7 @@ describe.each`
         // Note: window.location.assign and window.open need mock functions in
         // the testing environment, but JSDOM protects assign from being changed
         // so we need to replace the whole location object.
+        // @ts-expect-error [FEI-5019] - TS2322 - Type '{ assign: Mock<any, any, any>; ... }' is not assignable to type 'string & Location'.
         window.location = {...window.location, assign: jest.fn()};
         window.open = jest.fn();
     });
--- a/consistency-tests/__tests__/clickables.test.tsx
+++ b/consistency-tests/__tests__/clickables.test.tsx
@@ -122,9 +122,10 @@ describe.each`
     beforeEach(() => {
         // Note: window.location.assign and window.open need mock functions in
         // the testing environment, but JSDOM protects assign from being changed
-        // so we need to replace the whole location object.
-        // @ts-expect-error [FEI-5019] - TS2322 - Type '{ assign: Mock<any, any, any>; ... }' is not assignable to type 'string & Location'.
-        window.location = {...window.location, assign: jest.fn()};
+        // so we need to replace the whole location object. The DOM types don't
+        // allow reassigning `window.location`, so cast the replacement (a
+        // targeted cast rather than an `@ts-expect-error` line suppression).
+        window.location = {...window.location, assign: jest.fn()} as any;
         window.open = jest.fn();
     });
 
--- a/packages/wonder-blocks-button/src/components/__tests__/button.test.tsx
+++ b/packages/wonder-blocks-button/src/components/__tests__/button.test.tsx
@@ -19,8 +19,10 @@ describe("Button", () => {
     });
 
     afterAll(() => {
-        // @ts-expect-error [FEI-5019] - TS2322 - Type 'Location' is not assignable to type 'string & Location'.
-        window.location = location;
+        // The DOM types treat `window.location` as non-reassignable, so cast
+        // the saved value to restore it (a targeted cast rather than an
+        // `@ts-expect-error` that would suppress the whole line).
+        window.location = location as any;
     });
 
     describe("attributes", () => {
--- a/packages/wonder-blocks-button/src/components/__tests__/button.test.tsx
+++ b/packages/wonder-blocks-button/src/components/__tests__/button.test.tsx
@@ -19,6 +19,7 @@ describe("Button", () => {
     });
 
     afterAll(() => {
+        // @ts-expect-error [FEI-5019] - TS2322 - Type 'Location' is not assignable to type 'string & Location'.
         window.location = location;
     });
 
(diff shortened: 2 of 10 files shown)


jeresig (approved):
It looks like this is upgrading us to Typescript v6? And also fixing a bug? Is this intentional?

consistency-tests/__tests__/clickables.test.tsx
```diff
@@ -123,6 +123,7 @@ describe.each`
         // Note: window.location.assign and window.open need mock functions in
         // the testing environment, but JSDOM protects assign from being changed
         // so we need to replace the whole location object.
+        // @ts-expect-error [FEI-5019] - TS2322 - Type '{ assign: Mock<any, any, any>; ... }' is not assignable to type 'string & Location'.
```
somewhatabstract: Is there a way we can avoid adding new suppressions? I recommend using a cast to `any` instead.

packages/wonder-blocks-button/src/components/__tests__/button.test.tsx
```diff
@@ -19,6 +19,7 @@ describe("Button", () => {
     });
 
     afterAll(() => {
+        // @ts-expect-error [FEI-5019] - TS2322 - Type 'Location' is not assignable to type 'string & Location'.
```
somewhatabstract: Please don't add suppressions to the code base. This can be solved with a simple cast to `any`, for example.

packages/wonder-blocks-icon-button/src/components/__tests__/icon-button-unstyled.test.tsx
```diff
@@ -20,6 +20,7 @@ describe("IconButtonUnstyled", () => {
     });
 
     afterAll(() => {
+        // @ts-expect-error [FEI-5019] - TS2322 - Type 'Location' is not assignable to type 'string & Location'.
```
somewhatabstract: Please don't add suppressions to the code base. This can be solved with a simple cast to `any`, for example.

packages/wonder-blocks-testing-core/src/harness/adapters/__tests__/router.typestest.tsx
```diff
@@ -60,4 +61,87 @@ describe("router adapter config", () => {
         // @ts-expect-error: is not assignable to type 'undefined'
         adapter(null, {path: "/math/*", routes: [{path: "/"}]});
     });
+
+    it("should reject combining location with initialEntries", () => {
+        // @ts-expect-error: is not assignable to type 'undefined'
```
somewhatabstract: I don't think this should be a suppression. This is suppressing all typechecking on this line, not just the problem that we catch.

somewhatabstract (commented):
This looks good to me, but pushing back on the use of type suppressions. It is much better to defer to a targetted cast with a comment explaining it as it ensures we only suppress the typing of the specific thing, rather than an entire line. It also avoids setting a bad precedence that others are likely to follow.

kevinb-khan:
> It looks like this is upgrading us to Typescript v6? And also fixing a bug? Is this intentional?

@jeresig TS 6 is a little more relaxed with how it checks disjoint unions of objects which we use in a few places include the types for the routing adapter.  When I tried updating Khan/frontend to use TS 6 some of our types tests caught the issue there.  The fix though needs to be made in wonder-blocks itself.  Here's a link to the draft PR for updating Khan/frontend to TS 6 for additional context: https://github.com/Khan/frontend/pull/12852.

The reason why I'm updating to TS 6 is that I'm tired of TS checks being slow when using Claude Code's cloud sessions for the stuff I'm working on.  It wants to type check stuff all of the time and unfortunate the changes I'm making result in it having to type check everything which is still pretty slow.  Once we're on TS 6 we can start experimenting with ts-go.  The current status of ts-go is looking very promising, see https://github.com/microsoft/typescript-go.  The nice thing is that we can use them side-by-side until ts-go becuase TS 7.

jeresig:
@kevinb-khan That's fine! The PR title and most of the body emphasized changes to location/initialEntries and only noted the TS 6 upgrade as a bullet point so I was pretty surprised!

PR #3138, opened 2026-07-02
🔷 Remove closeButtonLight prop from PopoverContentCore

jandrade:
## Summary:
Removes the unused `closeButtonLight` prop from `PopoverContentCore`. The
prop toggled the close button's `actionStyles.inverse` styling, but it was
only reachable for an image popover placed at `top` — a combination no
consumer uses in practice. Removing it also drops the sole usage of
`@khanacademy/wonder-blocks-styles`, so that dependency (and its tsconfig
project reference) is removed from the package. Shipped as a major bump
since it removes a public prop from a shared component.

Reviewers: #wonder-blocks

Issue: WB-2389

## Test plan:

Auto-verified: `pnpm typecheck` passes and the existing
`popover-content.test.tsx` suite passes. The removed styling was only
reachable for an image popover placed at `top`, which is never used.

Optional human spot-check in Storybook that popover close buttons still
render correctly:
http://localhost:8228/?path=/docs/packages-popover-popovercontent--docs

## Review plan:

1. ⚠️ `packages/wonder-blocks-popover/src/components/popover-content-core.tsx`: Removes a public prop (`closeButtonLight`) from a shared design-system component, a breaking API change shipped as a major bump. Confirmed dead code (no consumers) and covered by existing unit tests.

### Common patterns:

**4 Files:** Full removal of the `closeButtonLight` prop and its dependency chain — prop type/defaultProps/render usage, the sole `actionStyles.inverse` styling, and the resulting `wonder-blocks-styles` package dependency + tsconfig reference.

```diff
-import {actionStyles} from "@khanacademy/wonder-blocks-styles";
 import CloseButton from "./close-button";

-                        style={[
-                            styles.closeButton,
-                            closeButtonLight && actionStyles.inverse,
-                        ]}
+                        style={styles.closeButton}
```


packages/wonder-blocks-popover/src/components/popover-content-core.tsx
```diff
@@ -10,18 +10,13 @@ import {
     sizing,
 } from "@khanacademy/wonder-blocks-tokens";
 
-import {actionStyles} from "@khanacademy/wonder-blocks-styles";
 import CloseButton from "./close-button";
```
jandrade, line 13: ⚠️ **1 of 1:** This removes the public `closeButtonLight` prop from `PopoverContentCore`, a shared design-system component — a breaking API change shipped as a major bump. It's confirmed dead code (no consumers; the styling was only reachable for an image popover placed at `top`, which is never used) and is covered by the existing `popover-content.test.tsx` suite. Removing it also drops the sole usage of `@khanacademy/wonder-blocks-styles`, so that dependency and its tsconfig reference are removed too.

marcysutton:
One thing I'm seeing in Storybook is a different close button color over an illustration. Will this be up to consumers to handle instead? (sorry just noticed this PR is in Draft!)

Popover with Illustration story with dark close icon button
Popover with Illustration story with light close icon button

jandrade:
This didn't work as I'd expect, reverting this change and I'll revisit with a different approach.

PR #3141, opened 2026-07-07
Switch typecheck scripts to tsgo and fix Text component typing

kevinb-khan:
## Summary

This PR updates the TypeScript type checking tooling to use the faster native TypeScript 7 compiler (`tsgo`) and resolves a type error in the `Text` component by properly casting the dynamic `tag` prop instead of suppressing the diagnostic.

Issue: FEI-7987

## Changes

- **Tooling**: Updated `typecheck` and `typewatch` npm scripts to use `tsgo` (from `@typescript/native-preview@7.0.0-dev.20260707.2`) instead of `tsc` for faster type checking
- **Text Component**: Fixed TypeScript error in `packages/wonder-blocks-core/src/components/text.tsx`:
  - Changed parameter destructuring from `tag: Tag = "span"` to `tag = "span"` to keep `tag` as a string variable
  - Added explicit cast of `tag` to `React.ElementType` before using it as a JSX element, with a clarifying comment
  - Updated `isHeaderRegex.test()` call to use the `tag` variable instead of the `Tag` type
  - Removed the `@ts-expect-error` suppression comment that is no longer needed

## Implementation Details

The `Text` component now properly handles its dynamic `tag` prop by:
1. Keeping the prop as a typed string in the function signature
2. Casting it to `React.ElementType` only when needed for JSX rendering
3. Using the string value for regex testing

This approach maintains type safety while allowing the component to render arbitrary DOM elements with full prop support, and works correctly with both `tsgo` and `tsc`.

https://claude.ai/code/session_01EfYHyhWehuavioXUtuD7s7


somewhatabstract (approved):
LGTM

Can you please link the Jira ticket in the PR description?

PR #3143, opened 2026-07-07
Add dark color-scheme support to preview styles

beaesguerra:
## Summary

Storybook: Added CSS rule to set `color-scheme: dark` on the root element when a dark theme is applied to the body. This ensures that browser-rendered UI elements (scrollbars, etc.) match the dark theme styling.

Note: Created WB-2394 to centralize global styles like this within a WB package! Consumers will need to configure this in the short term. Here's the frontend PR that does this: https://github.com/Khan/frontend/pull/13358

## Changes

- Added `:root:has(body[data-wb-theme$="-dark"])` selector that sets `color-scheme: dark`
- This rule targets any body element with a data attribute ending in "-dark" (e.g., `syl-dark`)

Issue: WB-2395

## Test Plan 

Verify native browser elements like scrollbars show as dark in the `syl-dark` theme: 

| Before | After | 
| --- | --- |
| image | image |

### DIFF
--- a/static/sb-styles/preview.css
+++ b/static/sb-styles/preview.css
@@ -12,7 +12,8 @@ html {
 
 /* Set `color-scheme: dark` on the `:root` element when the body uses a dark
  * theme (e.g. `syl-dark`). This ensures that browser-rendered UI (form
- * controls, scrollbars, etc.) matches the dark theme. */
+ * controls, scrollbars, etc.) matches the dark theme.
+ * This will be moved to WB global styles in WB-2394. */
 :root:has(body[data-wb-theme$="-dark"]) {
     color-scheme: dark;
 }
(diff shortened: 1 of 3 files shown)


static/sb-styles/preview.css
```diff
@@ -10,6 +10,13 @@ html {
     color: var(--wb-semanticColor-core-foreground-neutral-strong);
 }
 
+/* Set `color-scheme: dark` on the `:root` element when the body uses a dark
+ * theme (e.g. `syl-dark`). This ensures that browser-rendered UI (form
+ * controls, scrollbars, etc.) matches the dark theme. */
```
jandrade: maybe you could add the same WB ticket from the `frontend` PR to get to replace this with the upcoming global-styles file?

jandrade (approved):
Nice! :shipit:

static/sb-styles/preview.css
beaesguerra: Good call, added!

PR #3145, opened 2026-07-08
[🔥AUDIT🔥] Update BodyMonospace docs

jandrade:
🖍 _This is an audit!_ 🖍

## Summary:

Renames the `BodyMonospace` docs by removing the "deprecated" label as we
decided the component is not deprecated and should be used for code snippets and
other monospaced text.

Issue: "none"

## Test plan:

Verify that the `BodyMonospace` docs are no longer marked as deprecated and that
the component is still available for use in the documentation.


__docs__/wonder-blocks-typography/body-monospace.stories.tsx
```diff
@@ -24,7 +24,7 @@ export default {
         },
     },
     argTypes: TypographyArgTypes,
-    tags: ["deprecated", "!manifest"],
+    tags: ["!manifest"],
```
beaesguerra: Should we include it in the manifest? So the SB MCP can access these component docs?

beaesguerra (approved):
Looks good to me!

PR #3147, opened 2026-07-09
Add graphics semantic color tokens

beaesguerra:
## Summary

Add a new `graphics` semantic color token tier to support illustrative and decorative UI elements. These tokens map to the Figma "5 - Graphics" collection and provide color values for character art, gems, streaks, progress indicators, role badges, and external brand colors.

Note: Figma also exposes graphic palette token variables. This will be done later on in WB-2396, since we will want to expose the palette tokens with lint rules since it exposes a lot of the primitive values. 

The token table stories for semantic color tokens were also split up to avoid hitting the size limitation in Chromatic

Issue: WB-2167

## Test Plan

1. Confirm new tokens match Figma
2. Confirm new tokens are documented:
- `/?path=/docs/foundations-using-color--docs#graphics`
- `/?path=/docs/packages-tokens-semantic-color--docs#graphics`


__docs__/wonder-blocks-tokens/semantic-color-groups.stories.tsx
```diff
@@ -166,11 +166,6 @@ export const LearningBorder = () => {
                 }}
                 group="learning.border"
             />
-            <Heading size="medium">Characters Flesh</Heading>
```
beaesguerra, line 169: Removing docs for old structure for flesh tokens. WB-2380 will be for removing the old tokens for character flesh! 

Character flesh tokens are now moved into the graphics tier

__docs__/wonder-blocks-tokens/semantic-color.stories.tsx
beaesguerra, line 1: I'm breaking up the semantic color tokens table into a few stories to avoid hitting the Chromatic size limit with the addition of new graphic tokens (which will only get larger once we expose the graphic palette tokens too)

__docs__/wonder-blocks-tokens/semantic-color.stories.tsx
```diff
@@ -85,13 +86,71 @@ export default {
...
+export const Graphics = () => (
+    <SemanticColorsTable
+        tier="graphics"
+        tokens={flattenNestedTokens(semanticColor.graphics)}
+    />
+);
+
+const remainingTokens = Object.fromEntries(
     Object.entries(flattenNestedTokens(semanticColor)).filter(
-        ([key, _]) => !key.includes("action.") && !key.includes("status."),
```
beaesguerra, line 90: This was previously included to work around the Chromatic snapshot size limitation. We can include them again since we've broken the tables into separate stories (I think it's helpful to include them anyways as a reference)

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color.ts
```diff
@@ -160,13 +164,195 @@ const sharedFeedbackStrongTokens = {
     text: core.foreground.knockout.default,
 };
 
+/**
+ * Graphics tokens are used for illustrative and decorative elements such as
+ * character art, gems, streaks, progress indicators, role badges, and
+ * external brand colors.
+ */
+const graphics = {
```
beaesguerra, line 172: Note: I created WB-2396 to expose the graphic palette tokens too, which is done in Figma. 

I will defer to doing this later so we can release it with a lint rule to make sure they only get used with graphic elements (since they're pretty much primitive values). These tokens are not needed for dark mode yet since they don't switch values based on the theme

__docs__/wonder-blocks-tokens/semantic-color.stories.tsx
jandrade: This was removed mostly because these tokens are considered deprecated/internal. I'm going to tackle removing `status` tokens, and `action` tokens are already covered by all Button/IconButton stories as these are the only consumers.

I'd prefer to get rid of them from visual + docs coverage, as coding agents can misunderstand that these are deprecated/disallowed (I've seen recent uses of `status` tokens in `frontend).

beaesguerra: Sounds good, added it back in! 

One thing to note for `action` tokens. In dark mode, `background.instructive.default` (blurple) and `foreground.knockout.default` (blackish) don't work well together, so for any implementations that expect blurple + white will need to use `action` tokens. I have a ticket to go through current usage of this pairing to see if we can switch it to WB components directly or use the action tokens instead https://khanacademy.atlassian.net/browse/FEI-7734

beaesguerra: Oh one other thing to note! The Storybook MCP can't parse dynamic content in stories so it won't get token details from these token tables. It would get it from the tokens map reference (which is static) https://khan.github.io/wonder-blocks/?path=/docs/references-tokens-map-semantic-color--docs 

And we exclude status and action tokens there too! https://github.com/Khan/wonder-blocks/blob/bfc4b3f710bcf172c0aa3a1de52d9e6f0237cefd/packages/wonder-blocks-tokens/src/build/generate-token-docs.ts#L126-L132

beaesguerra:
Also, here's the draft PR where I'm trying out these new tokens with the EOT cards https://github.com/Khan/frontend/pull/13532 !

__docs__/wonder-blocks-tokens/semantic-color-groups.stories.tsx
marcysutton: I hadn't seen the old structure so I'm trying to compare with prod, but confusingly not finding the Semantic Color Groups stories? Were those rendered somewhere? 

Semantic Color docs

beaesguerra: They're in the Using color docs under Learning > Border and Learning > Background! https://khan.github.io/wonder-blocks/?path=/docs/foundations-using-color--docs

image

image

beaesguerra: (We reference these stories from semantic-color-groups.stories.tsx in the mdx so that the colors are reactive to the theme when we switch themes using the Storybook control!)

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-syl-dark.ts
```diff
@@ -549,6 +549,156 @@ export const semanticColor = mergeTheme(thunderblocksSemanticColor, {
     },
     core,
     feedback: feedbackTokens,
+    graphics: {
+        characters: {
+            flesh: {
+                blue: {
+                    background: {
+                        subtle: graphicsPalette.blue_20,
```
marcysutton, line 557: Where are these defined in Figma? So far I've found a Handshake file but the colors don't match up with this object.

__docs__/foundations-color.mdx
```diff
@@ -227,6 +227,48 @@ in the UI.
...
+Neutral background colors used for illustrative surfaces.
+
+<Canvas of={SemanticColorGroups.GraphicsNeutral} layout="fullscreen" />
+
+#### External brands
+
+Colors that reflect external brand palettes, used exclusively when representing
+those brands (e.g. third-party sign-in buttons).
+
+<Canvas of={SemanticColorGroups.GraphicsExternalBrands} layout="fullscreen" />
```
marcysutton, line 270: Lol, I never realized that Google and Microsoft have nearly identical color palettes!

External brands color swatches

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-syl-dark.ts
beaesguerra: They're included in the Figma variables in Foundations (that's where the latest token values should always be in Figma!)

https://www.figma.com/design/9seilz0fBgo3QUj6eXM0yR/%E2%9A%A1%EF%B8%8F-Foundations?m=dev&node-id=3485-50&var-set-id=10-22&view=variables

__docs__/wonder-blocks-tokens/semantic-color.stories.tsx
jandrade: oh great point! thanks for the clarification :)

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-syl-dark.ts
```diff
@@ -549,6 +549,156 @@ export const semanticColor = mergeTheme(thunderblocksSemanticColor, {
...
+                background: {
+                    subtle: graphicsPalette.green_01,
+                    default: graphicsPalette.green_05,
+                },
+                foreground: {
+                    strong: graphicsPalette.green_10,
+                },
+            },
+        },
+        role: {
```
marcysutton, line 660: Are these colors represented in the SYL Light theme? I didn't see the file here and browsing the code, it didn't appear they were in the `semantic-color-thunderblocks.ts` file. I see them on the Using Color docs page though 🤷‍♀️

marcysutton (approved):
This looks good to me! I had one question about some of the colors in the SYL Light theme but that's about it.

packages/wonder-blocks-tokens/src/theme/semantic/semantic-color-syl-dark.ts
beaesguerra: The `default` theme is the base theme in WB right now, and `thunderblocks` gets merged with that! The graphic tokens are the same in these themes, so `thunderblocks` doesn't need to override anything. 

`syl-dark` will also get merged with `thunderblocks`, which gets merged with `default`, but `syl-dark` has some values that are different so they're overridden in `semantic-color-syl-dark.ts`!

jandrade (approved):
LGTM! Thanks for the changes 🚢

PR #3149, opened 2026-07-09
Limit npm upgrade for OIDC support to v11 (avoid upgrading to v12 as its broken)

jeremywiebe:
## Summary:

Duplicated from Khan/perseus#3887 and verified in a [sample PR run](https://github.com/Khan/perseus/actions/runs/29048808663):

> When we switched Perseus to use trusted publishing for npm publishes, we needed to > ensure that the version of `npm` we used was at least v11.5.1 or newer. So the > publish workflow upgrades npm using the command `npm install -g npm@latest`. This > worked great, until npm released v12 just this week. It seems v12 has a bug where it > can't properly detect if a version of a package is published or not. 
> 
> Last week the snapshot publishing worked fine and showed this (as an example):
> 
> ```
> ✅ npm upgraded to 11.18.0
> 
> ... many more lines of output!
> 
> 🦋  warn @khanacademy/kas is not being published because version 2.2.3 is already > published on npm
> ```
> 
> Today, snapshot publishes are failing and we're seeing this in the logs (note that > this is the exact same version as above, but now it tthinks it is not yet published > and needs to be!):
> 
> ```
> ✅ npm upgraded to 12.0.0
> 
> ... many more lines of output 
> 
> 🦋  info @khanacademy/kas is being published because our local version (2.2.3) has > not been published on npm
> ```

Issue: LEMS-4359 

## Test plan:

Edit this PR to include a small change in the `perseus` package
Request a snapshot (add `/snapshot` comment)
Watch the publish run to see if it properly detects that `kas` v2.2.3 is already published!


jandrade (approved):
Looks great! I've confirmed that the right npm version is installed in the current PR: https://github.com/Khan/wonder-blocks/actions/runs/29049502367/job/86226500069?pr=3149


```
Run # npm trusted publishing requires npm CLI v11.5.1+, but NOT

changed 21 packages in 4s

15 packages are looking for funding
  run `npm fund` for details
✅ npm upgraded to 11.18.0
```

Thanks for fixing this! 🚢

PR #3150, opened 2026-07-09
Card: Add support for inline svg backgrounds

beaesguerra:
## Summary:

Currently, the Card component `background` prop supports an image which will be used as a background image. However, when a background image is based on semantic color tokens, it will not adapt based on the theme. This is an issue in dark mode because the background image is usually too light, and the foreground text in the card is swapped. 

To address this, background images can be provided as an inline svg, which will use semantic color tokens to adapt based on the theme. An inline svg can be provided to the `background` prop by setting it to an object with `{ type: "inline-svg", component: InlineSvgBackgroundComponent }` 

Issue: WB-2167

## Test plan:

1. Confirm the docs for using an inline svg for the background prop (`/?path=/docs/packages-card--docs#with-inline-svg-background`)

2. Confirm that a Card that uses an inline svg background changes colors as expected when changing to dark mode (`?path=/story/packages-card-testing-snapshots-card--state-sheet&globals=theme:syl-dark` and `?path=/story/packages-card--with-inline-svg-background&globals=theme:syl-dark`)

https://github.com/user-attachments/assets/88634137-b43e-4c28-9014-ae78f1ea5979

3. Confirm existing Card behaviour is unchanged (for example, the background image behaviour)


__docs__/wonder-blocks-card/card-background-example.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from "react";
```
beaesguerra, line 1: This is an example of a component that renders a card background svg inline. It uses semantic color tokens!

.eslintrc.js
```diff
@@ -224,5 +224,11 @@ module.exports = {
         "@typescript-eslint/no-empty-function": "off",
         "@typescript-eslint/prefer-enum-initializers": "error",
         "@typescript-eslint/prefer-literal-enum-member": "error",
+
+        /**
+         * @khanacademy/wonder-blocks rules
+         */
+        "@khanacademy/wonder-blocks/require-logical-properties-for-rtl":
+            "error",
```
beaesguerra, line 232: Enabling this for the WB repo so we can keep lint rules for using logical properties

beaesguerra:
@claude review

beaesguerra:
Going to go with a pattern masking approach instead to avoid having to render larger svg elements.

PR #3154, opened 2026-07-20
Upgrade repo to Node v24.18.0

jeremywiebe:
## Summary:

There are two high-severity CVE's for Node v24.16.0. We're skipping v24.17.0 because we tried it previously in the Perseus repo and it broke one of our Github Actions. 

There are some related changes here that I thought were useful:

  * Moved the determination of node version _into_ the shared-node-cache action
    * Is there any reason why we'd use different node versions between actions??
  * Moved the installation of npm v11 into shared-node-cache - it's required for trusted publishing and again, it makes sense (to me!) that we centralize it so it can't be forgotten (or if we ever want to remove it, it's a single thing to clean up). 

This PR also applies updated `agent-settings` (given that we bumped the Node version in some of the hooks in that dir).

Issue: INFRA-11108

## Test plan:

Run all Github Actions and if they all pass, we should be good


.github/workflows/chromatic-build.yml
```diff
@@ -74,17 +74,11 @@ jobs:
                   # *.chromatic.com: Chromatic service for uploading and comparing Storybook snapshots
                   # chromatic-builds.s3.us-west-2.amazonaws.com: S3 bucket for Storybook file uploads (presigned URLs returned by index.chromatic.com GraphQL API)
                   extra-domains: registry.npmjs.org index.chromatic.com notify.dev-chromatic.com notify.staging-chromatic.com notify.chromatic.com chromatic.com chromatic-builds.s3.us-west-2.amazonaws.com
-            - name: Use Node.js 24.x
```
jeremywiebe, line 77: The next step (shared-node-cache) already does this, so it's redundant to do it here - plus it adds maintenance overhead when bumping node versions.

.github/workflows/chromatic-build.yml
```diff
@@ -74,17 +74,11 @@ jobs:
...
-              uses: actions/setup-node@v6
-              with:
-                  node-version: 24.x
 
             # Cache and install dependencies
             - name: Install & cache node_modules
               uses: ./.github/actions/shared-node-cache
               timeout-minutes: 5
-              with:
```
jeremywiebe, line 86: I've hard-coded the version of Node we use in `shared-node-cache`. I can't see a reason why we'd want to override this per-workflow. Again, it moves Node version management for Github actions to 1 place (except for a single direct use of `setup-node` outside of shared-node-cache further down).

.github/workflows/chromatic-pr.yml
```diff
@@ -42,10 +42,6 @@ jobs:
         name: Get results on regular PRs
         needs: chromatic_review
         runs-on: ${{ vars.USE_GITHUB_RUNNERS == 'true' && 'ubuntu-latest' || 'ephemeral-runner' }}
-        strategy:
-            fail-fast: false
-            matrix:
-                node-version: [24.x]
```
jeremywiebe, line 48: Using a matrix is only useful when you want to run on multiple versions of node in parallel.  I don't know that we've ever done that. Plus, a matrix embeds this version into the workflow name, which can cause problems if the workflow is used as a Github check for branch protection.

.github/workflows/publish.yml
```diff
@@ -82,19 +82,6 @@ jobs:
             - name: Install & cache node_modules
               uses: ./.github/actions/shared-node-cache
               timeout-minutes: 5
-              with:
-                  node-version: 24.x
-
-            - name: ⬆️ Upgrade npm for OIDC support
```
jeremywiebe, line 88: No longer needed because Node v24.18 ships with a version of `npm` that's newer than 11.5.1!

__docs__/wonder-blocks-form/checkbox-group-testing-snapshots.stories.tsx
```diff
@@ -77,9 +77,9 @@ export const Scenarios: Story = {
                     description: longText,
                     errorMessage: longText,
                     children: [
-                        <Choice label={longText} value="1" />,
-                        <Choice label={longText} value="2" />,
-                        <Choice label={longText} value="3" />,
+                        <Choice key="1" label={longText} value="1" />,
```
jeremywiebe, line 80: I saw in the test run that React was complaining about child list items without keys.

.claude/hooks/pnpm-install.mjs
jeremywiebe, line 1: ALL of the changes to the AI agent directories are the result of running: `pnpm install-settings --env all wonder-blocks` in the Khan/agent-settings directory after [bumping](https://github.com/Khan/agent-settings/pull/36) it to Node 24.18.

somewhatabstract (approved):
@jeremywiebe @jandrade @jeresig what do you think about marking these agent files as binary with `.gitattributes` so they don't get expanded for review when we update them?

jeremywiebe:
@somewhatabstract I'll do it manually for now, but I like @jeresig's idea of generating it via agent-settings.

.gitattributes
```diff
@@ -1,2 +1,34 @@
 docs/* linguist-generated
-package-lock.json linguist-generated
+
+# BEGIN: agent-settings managed files
```
jeremywiebe, line 3: I ran `pnpm install-settings --env all wonder-blocks` in the agent-settings dir and captured all of the files it wrote. I then transformed that list into this list.  

Whenever we automate `agent-settings` writing this list, we at least have comments surrounding this list to key off of so its easy to remove these hand-written entries.

jandrade (approved):
Looks great! thanks for optimizing our workflows 🚀

PR #3155, opened 2026-07-24
Add documentation for dark mode best practices

beaesguerra:
## Summary:

Add docs for best practices around supporting dark mode. This is based on some of the fixes we applied in frontend so far to support dark mode in Classroom. 

The doc emphasizes on using semantic color tokens and how to handle graphics such as background images and icons.

Issue: WB-2402

## Test plan:

Review docs for dark mode best practices (`?path=/docs/best-practices-dark-mode--docs&globals=theme:syl-dark`)

PR #3156, opened 2026-07-27
Update WB examples that use text on a background image

beaesguerra:
## Summary:

Using the dark mode best practices from https://github.com/Khan/wonder-blocks/pull/3155 to update the examples in Wonder Blocks that includes text on top of a background image

Note: We keep the old background image examples in the Card statesheet so we can continue to have tests that confirm the functionality still works (I created https://khanacademy.atlassian.net/browse/WB-2397 for later to consider if we want to systemize background patterns and remove the background image functionality built into cards) 

Issue: WB-2406

## Test plan:
1. Confirm that the card example in the catalog adapts to dark mode `?path=/story/catalog-components--layout-blocks&globals=theme:syl-dark`
2. Confirm that the card background image example adapts to dark mode  `?path=/story/packages-card--with-background-image&globals=theme:syl-dark`
3. Confirm that the card with background split example adapts to dark mode `?path=/story/packages-card--with-split-background-image&globals=theme:syl-dark`
4. Confirm that the card statesheet includes the pre-existing background image example (which doesn't work in dark mode), and the new approach with applying semantic colors to background patterns (`?path=/story/packages-card-testing-snapshots-card--state-sheet&globals=theme:syl-dark`)
5. Confirm that the Flexible Modal example with a background image adapts to dark mode `?path=/story/packages-modal-flexibledialog--with-background-image&globals=theme:syl-dark`


__docs__/wonder-blocks-card/card-testing-snapshots.stories.tsx
```diff
@@ -30,6 +31,41 @@ const backgrounds = [
     {name: "Default", props: {background: "base-default"}},
     {name: "Subtle", props: {background: "base-subtle"}},
     {name: "Image", props: {background: eotBackground}},
```
beaesguerra, line 33: We keep the existing examples in the statesheet so we have visual regression test coverage for the existing functionality. Created https://khanacademy.atlassian.net/browse/WB-2397 to consider if we want to remove support for background images and systemize background patterns instead so we can use semantic color tokens

static/celebration_bg.svg
```diff
@@ -1,7 +0,0 @@
-<svg width="1366" height="640" viewBox="0 0 1366 640" fill="none" xmlns="http://www.w3.org/2000/svg">
```
beaesguerra, line 1: Removed in favour of `celebration_bg-pattern.svg` which has no background

static/EOT-Background-pattern.svg
beaesguerra, line 1: The original background image asset for this is still in the repo since we are still using it in the statesheet to confirm that the existing background image functionality in cards continue to work as expected

marcysutton (approved):
Thanks for this @beaesguerra! Looks great to me. Now we can read that Rainier earned Calculus gems in dark mode 🐶 😂 

Card story in dark mode that says Congrats Rainier McCheddarton, my dog

PR #3042, opened 2026-05-11
Enable require-logical-properties-for-rtl in recommended config (CLASS-14218)

maddy531:
## Summary

- Enables `@khanacademy/wonder-blocks/require-logical-properties-for-rtl` in the plugin's `recommended` config
- Runs `eslint --fix` to auto-fix all existing violations across 

Where are we in our RTL update:
| Step | Body of work | Repo | This PR? | Tracking |
  | --- | --- | --- | --- | --- |
  | 1 | Add `require-logical-properties-for-rtl` rule (TS port of frontend's JS rule), tests, docs, demo, Storybook story, changeset. Ship in plugin but **not enabled in any config**. | wonder-blocks | ✅ 🚢   https://github.com/Khan/wonder-blocks/pull/3040 | [CLASS-14218](https://khanacademy.atlassian.net/jira/software/c/projects/CLASS/boards/35?assignee=624f84e6f3824d006a5d78bb&selectedIssue=CLASS-14218) |
  | 2 | Enable rule in `recommended` config. Run `eslint --fix` against WB. Verify Chromatic. | wonder-blocks | **Here**  https://github.com/Khan/wonder-blocks/pull/3042 | [CLASS-14218](https://khanacademy.atlassian.net/jira/software/c/projects/CLASS/boards/35?assignee=624f84e6f3824d006a5d78bb&selectedIssue=CLASS-14218) |
  | 3 | In frontend: bump plugin, switch `.eslintrc.js` to WB rule, delete internal copy, rename `eslint-disable` comments. No new violations expected. | frontend | ❌ Follow-up | [CLASS-14239](https://khanacademy.atlassian.net/browse/CLASS-14239) |
  | 4 | Extend rule coverage and fix resulting violations in frontend:<ul><li>3- and 4-value `padding`/`margin` shorthand (`padding: "10px 20px 30px"`) — only 2-value is currently handled</li><li>`height` / `width` → `blockSize` / `inlineSize`</li><li>Warn on `translateX` / `transformOrigin` directional use (no auto-fix; manual review)</li><li>Remove unused warning rules</li></ul> | wonder-blocks → frontend | ❌ Future | [CLASS-14240](https://khanacademy.atlassian.net/browse/CLASS-14240) |
  | 5 | Enable in perseus as one full sweep — picks up the complete final rule with no follow-up passes needed. | perseus | ❌ Future | [CLASS-14252](https://khanacademy.atlassian.net/browse/CLASS-14252) |

## Jira

CLASS-14218

[...]

PR #3157, opened 2026-07-27
Configure require-logical-properties-for-rtl explicitly

beaesguerra:
## Summary:
Since `require-logical-properties-for-rtl` is not enabled by default in the WB lint rules plugin, we enable it for the WB project so that we can prevent new instances of non-logical css properties 

Issue: WB-2407

## Test plan:

Confirm PR checks pass, including linting


marcysutton (approved):
Looks good to me! I think this wasn't enabled from the start because we had a bunch of stuff landing at the same time that would be affected. Glad to clean up after it now!!

beaesguerra:
> Looks good to me! I think this wasn't enabled from the start because we had a bunch of stuff landing at the same time that would be affected. Glad to clean up after it now!!

@marcysutton This rule was enabled by default in the plugin in this PR: https://github.com/Khan/wonder-blocks/pull/3042! Existing errors were fixed in that PR too, which is why re-enabling it here doesn't find new errors! I think we had to remove setting it to `error` by default in the plugin when other lint rules were being released around the same time too! 

I only realized it wasn't turned on anymore when I was able to use `top`, `right`, etc properties with no lint errors! Glad to have this enabled as an automated reminder 😄

marcysutton:
Oh yes, I remember now. We disabled it for the downstream effect since there were a ton of fixes to make at the same time I was releasing other lint rules. Thanks again for circling back to this one!

PR #3163, opened 2026-08-03
🚨 Refactor Popover to use wonder-blocks-floating

jandrade:
## Summary:

Migrate `Popover` off PopperJS and the `wonder-blocks-tooltip` dependency to
the new `Floating` component (floating-ui). `Floating` now owns positioning,
portaling, the tail/arrow, dismiss (Esc/outside click), and focus management.

This PR is stacked on #3182, which adds the `Floating` props this migration
needs (`returnFocus`, `closeOnFocusOut`, `onPlacementChange`, `shiftPadding`,
`rootBoundary`). Those `Floating` changes are not part of this diff — this PR is
Popover-only. (#3182 also drops `Floating`'s `472px` `max-inline-size`;
`Popover` is unaffected because `PopoverContentCore` caps its content at 288px.)

Popover keeps its public API. `rootBoundary`/`viewportPadding` are remapped onto
floating-ui's `flip`/`shift` middleware; `autoUpdate` and `initialFocusDelay`
are now deprecated no-ops. Focus management is now floating-ui's non-modal model
(focus flows via focus guards) rather than the previous custom circular
navigation. The popover "bubble" chrome (background/border/shadow/radius) now
comes from `Floating`; `PopoverContentCore` no longer renders its own chrome.
The popover uses `strategy="fixed"` so it positions correctly inside clipping/
scrolling ancestors.

`Floating` resolves the reference (anchor) element from the DOM rather than via a
ref, so `PopoverAnchor` spreads the `FloatingReferenceAttributeName` attribute it
is given onto the trigger. That is the only change to `PopoverAnchor` here, and
it is what lets the popover position at all. Its existing ref plumbing
(`forwardRef` + the hand-rolled `mergeRefs`) is left in place and removed
separately in the follow-up, **#3189**, so the change to the trigger's contract
can be reviewed on its own.

The superseded internal helpers (`focus-manager`, `initial-focus`,
`popover-event-listener`) and the popper-specific parts of `popover-dialog` are
removed. `PopoverDialog` is now a small functional dialog wrapper.

Two smaller consumer-visible changes fell out of the migration: `onClose` is now
guarded to fire at most once per open/close cycle (floating-ui can request a
close through escape, outside click and focus-out, so the callback could
otherwise fire more than once for a single dismissal), and `testId` is now
actually applied — it was previously declared on `Popover`'s props but never
used, and is now forwarded to the floating element.

The changeset (`popover-floating-ui.md`) covers all of the above: the positioning
engine, the `PopoverContentCore` chrome move, the tail now coming from
`Floating`, the `fixed` strategy, the focus model, the remapped
`rootBoundary`/`viewportPadding`, the two deprecated no-ops, and the `onClose` /
`testId` changes.

Also included: the `Placement` and `RootBoundary` types are relocated off the
`wonder-blocks-tooltip` / `@popperjs/core` dependencies into a local
`util/types.ts` (both unions are unchanged); `wonder-blocks-popover`'s
`tsconfig-build.json` project references are updated to match the new dependency
set (floating in; modal/tooltip out); and two Tabs stories are touched — a stale
PopperJS comment in `navigation-tabs.stories.tsx`, and `aria-required-children`
disabled for the `TabLabelRenderFunction` story (floating-ui's focus-guard
scaffolding renders as a sibling inside the tablist).

Issue: WB-2119

## Test plan:

Automated: `pnpm typecheck`, `pnpm lint`, and `pnpm build` all pass. The Popover
jest suite passes (tests updated for the new focus model). Focus-flow behaviors
that jsdom can't faithfully run against floating-ui's focus guards are covered by
the Floating package's unit tests in #3182.

Manual (Storybook — Popover and PopoverContentCore stories):

1. Verify the popover renders with a single border + shadow (no double chrome)
   across placements, and the arrow/tail connects to the bubble.
2. Verify placement flip near viewport edges, and `rootBoundary="document"` /
   `viewportPadding` stories still position correctly.
3. Verify dismiss (Esc + outside click), initial focus, return focus,
   `closedFocusId`, controlled/uncontrolled open, and `portal` on/off.
4. Verify RTL placement (left/right) mirrors correctly.
5. Confirm standalone `PopoverContentCore` stories still look acceptable now
   that its chrome is provided by the parent (Floating) rather than itself.
6. Verify the Tabs `TabLabelRenderFunction` story (popover anchored to a tab)
   still positions and behaves correctly.

## Review plan:

1. 🚨 [`popover.tsx`](https://github.com/Khan/wonder-blocks/pull/3163#discussion_r3706555606)
2. ⚠️ [`popover-content-core.tsx`](https://github.com/Khan/wonder-blocks/pull/3163#discussion_r3706555730)
3. ⚠️ [`popover-dialog.tsx`](https://github.com/Khan/wonder-blocks/pull/3163#discussion_r3706555825)
4. 🔷 [`popover-anchor.tsx`](https://github.com/Khan/wonder-blocks/pull/3163#discussion_r3858693216)

### Common patterns:

**1 component + 1 consumer:** Replace PopperJS/Tooltip-based positioning with the `Floating` component.

```tsx
// Before (popover.tsx)
<TooltipPopper anchorElement={anchorElement} placement={placementProp} ...>
  {(popperProps) => (
    <PopoverDialog {...popperProps} showTail={showTail} onUpdate={setPlacement}>
      {renderContent(uniqueId)}
    </PopoverDialog>
  )}
</TooltipPopper>

// After (popover.tsx)
<Floating open={opened} strategy="fixed" placement={placementProp} content={dialog}
  showArrow={showTail} onPlacementChange={handlePlacementChange}
  dismissEnabled={!!dismissEnabled} closeOnFocusOut={!!dismissEnabled} ...>
  <PopoverAnchor ...>{children}</PopoverAnchor>
</Floating>
```

**3 files deleted:** Remove internal focus/dismiss helpers superseded by Floating's `FloatingFocusManager` + `useDismiss` (`focus-manager.tsx`, `initial-focus.ts`, `popover-event-listener.ts`, plus their tests).

**2 files:** Relocate the `Placement` and `RootBoundary` types off the `wonder-blocks-tooltip` / `@popperjs/core` dependencies into a local `util/types.ts`. Both unions are unchanged.

```ts
// Before (popover-context.ts)
import type {Placement} from "@khanacademy/wonder-blocks-tooltip";
// After
import type {Placement} from "../util/types";
```


packages/wonder-blocks-floating/src/components/floating.tsx
jandrade, line 1: 🚨 ~Shared design-system component. Adds `returnFocus`, `closeOnFocusOut`, `onPlacementChange`, `shiftPadding`, and `rootBoundary`, wiring them into the focus manager and middleware — this changes behavior for all Floating consumers.~

**Outdated:** these `Floating` changes now live in the stacked base PR #3182 and are no longer part of this diff. This PR is Popover-only.

packages/wonder-blocks-popover/src/components/popover.tsx
jandrade, line 1: 🚨 Full rewrite of the user-visible Popover to adopt the Floating component. Changes the positioning engine (PopperJS → floating-ui), focus management (custom → floating-ui non-modal), and dismiss handling.

packages/wonder-blocks-popover/src/components/popover-content-core.tsx
jandrade, line 1: ⚠️ Removes the bubble chrome (border/background/shadow/radius); chrome is now provided by Floating. This changes the appearance of standalone PopoverContentCore usage.

packages/wonder-blocks-popover/src/components/popover-dialog.tsx
jandrade, line 1: ⚠️ Reworked from a class component into a functional dialog wrapper; drops the tail and the wonder-blocks-tooltip dependency (Floating provides the arrow).

packages/wonder-blocks-floating/src/components/floating.tsx
```diff
@@ -209,45 +258,57 @@ export default function Floating({
...
+            // Add offset from the reference element
+            offset({mainAxis: offsetProp}),
+            // Flip to the opposite side if there's not enough space
+            flipProp ? flip({rootBoundary}) : undefined,
+            // Shift along the axis to keep it in view
+            shiftProp
+                ? shift({
+                      padding: shiftPadding,
+                      crossAxis: true,
+                      rootBoundary,
```
jandrade, line 318: ~**note:** The changes here are adding `rootBoundary` to support Popover's prop to change the boundary (document or viewport).~

**Outdated:** moved to #3182.

packages/wonder-blocks-popover/src/components/__tests__/popover.test.tsx
```diff
@@ -568,48 +603,6 @@ describe("Popover", () => {
             ).toHaveFocus();
         });
 
-        it("should close the popover when the user tabs before the trigger element", async () => {
```
jandrade, line 571: **note:** removed some tests below as these are already covered by Floating. Let me know if you prefer to put this back.

packages/wonder-blocks-popover/src/components/popover-anchor.tsx
```diff
@@ -1,37 +1,19 @@
 import * as React from "react";
```
jandrade, line 1: 🔷 `Floating` resolves the reference (anchor) element from the DOM rather than via a ref, so `PopoverAnchor` spreads the `FloatingReferenceAttributeName` attribute it is given onto the trigger. This applies to both the element-children and the function-as-children patterns, and it is what lets the popover position at all.

That attribute spread is the **only** change to this file in this PR. The existing ref plumbing (`React.forwardRef` plus the hand-rolled `mergeRefs` helper) is deliberately left in place here and removed separately in #3189, so the change to the trigger's contract can be reviewed on its own.

somewhatabstract (approved):
Looks great! Excited to see all this code getting deleted!

__docs__/wonder-blocks-tabs/tabs.stories.tsx
```diff
@@ -656,6 +656,26 @@ export const TabLabelRenderFunction: StoryComponentType = {
...
+                    {
+                        // The "Tab with an opened Popover" example anchors a
+                        // Popover to a tab inside the tablist. When open, the
+                        // Popover (via wonder-blocks-floating / floating-ui)
+                        // inserts tab-order and focus-guard scaffolding (e.g.
+                        // `span[aria-owns]`) as siblings of the anchor, which
+                        // trips `aria-required-children` on the tablist. This
+                        // scaffolding is accessible-by-design; the rule is being
+                        // strict about the tablist's allowed DOM children.
+                        id: "aria-required-children",
```
marcysutton, line 679: How does this repro? I wanted to investigate it a bit locally to understand it more. But if I remove this config, there aren't any accessibility violations. I'm looking at this story: `/?path=/story/packages-tabs-testing-tabs-playtesting--popover-with-tabs&globals=theme:syl-dark`

.changeset/popover-floating-ui.md
```diff
@@ -0,0 +1,18 @@
...
+
+Breaking / behavior changes:
+
+- `Popover` now positions its content with floating-ui via the `Floating` component. The `@popperjs/core` and `react-popper` peer dependencies (and the `@khanacademy/wonder-blocks-tooltip` / `@khanacademy/wonder-blocks-modal` dependencies) have been removed.
+- `PopoverContentCore` no longer renders the popover "bubble" chrome (background, border, border radius and shadow). That chrome now comes from `Floating`, so a `PopoverContentCore` rendered on its own — outside a `Popover` — no longer looks like a popover bubble and needs its own container styling.
+- The popover tail is now rendered by `Floating` (as its arrow) instead of `TooltipTail`. The `showTail` prop is unchanged, but the tail's markup and styling now come from `Floating`.
+- The popover is positioned with floating-ui's `fixed` strategy, so it is no longer clipped by scrolling or `overflow`-clipping ancestors the way it could be before.
+- Focus management is now handled by floating-ui's non-modal focus manager rather than the previous custom implementation. Keyboard focus now flows using floating-ui's focus guards instead of the previous circular navigation model.
```
marcysutton: So these dialogs are non-modal, but focus is still cycling through the Popover and any actions by way of the focus guards. Should it? I know we have some difficult-to-reconcile accessibility issues with the `tabindex=0` on non-interactive guard elements, as they fail in axe-core. Looking at the new [native Popover](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/popover) and an example with actions, I'm not so sure that we need to cycle focus at all. WDYT?

marcysutton (commented):
I'm still going through the PR but I thought I'd submit these questions to start, including a potentially big question about focus management and interaction flow. Let me know what you think!

packages/wonder-blocks-popover/src/util/types.ts
```diff
@@ -0,0 +1,18 @@
...
+export type Placement = "top" | "bottom" | "left" | "right";
+
+/**
+ * The root boundary that the popover is kept within when it is repositioned to
+ * stay in view.
+ *
+ * - `"viewport"`: keep the popover within the user's viewport.
+ * - `"document"`: keep the popover within the document body.
+ */
+export type RootBoundary = "viewport" | "document";
```
marcysutton, line 18: Popover with rootBoundary='document'

There's no way to dismiss the popover in the [`document rootBoundary` story](https://5e1bf4b385e3fb0020b7073c-ywoxdqljdb.chromatic.com/?path=/story/packages-popover-popover--with-document-root-boundary) -- clicking neither the trigger button nor the background seems to dismiss. Escape doesn't do anything either. How do those get closed? (I think it predates this PR as I see it in prod, just thought I'd mention it)

packages/wonder-blocks-popover/src/util/types.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * The coarse placement of the popover relative to its anchor.
+ *
+ * NOTE: Wonder Blocks Floating (floating-ui) supports more granular placements
+ * (e.g. `top-start`), but Popover only exposes and reasons about these four
+ * coarse values. The resolved floating-ui placement is normalized down to one
+ * of these before being shared via `PopoverContext`.
+ */
+export type Placement = "top" | "bottom" | "left" | "right";
```
marcysutton, line 9: I'm not seeing these populated in [Storybook args](https://5e1bf4b385e3fb0020b7073c-ywoxdqljdb.chromatic.com/?path=/docs/packages-popover-popover--docs) -- only `top` is showing. Should they be configurable as story options? 

Placement in args table

packages/wonder-blocks-popover/src/components/popover-content-core.tsx
marcysutton: Should we add a note to the stories about the difference in presentation when shown standalone?

This branch:
Image

Prod:
Image

marcysutton (commented):
a few more questions!

.changeset/popover-floating-ui.md
jandrade: I think Claude got it wrong in the description as Popover has never used cyclic navigation.

__docs__/wonder-blocks-tabs/tabs.stories.tsx
jandrade: Good catch that it doesn't repro where you were looking — that's a different story. The playtesting `popover-with-tabs` story starts with its popover **closed**, and the offending markup only exists while the popover is open. The exception is on `Packages / Tabs / Tabs / TabLabelRenderFunction`, where tab-4 (`Tab with an opened Popover`) passes `opened={true}`.

I reproduced it to be sure. While that popover is open, floating-ui's `FloatingFocusManager` inserts a hidden `<span aria-owns="{floating-id}">` immediately after the anchor, so screen readers place the popover next to its trigger in the virtual buffer. Because the anchor is a `Tab`, that span becomes a child of the `tablist`:

```html
<div role="tablist" aria-label="Test">
  <button role="tab" ...>Tab 1</button>
  <button role="tab" aria-expanded="true" ...>Tab with an opened Popover</button>
  <span tabindex="-1" aria-hidden="true" ...></span>
  <span data-type="outside" tabindex="0" aria-hidden="true" data-floating-ui-focus-guard ...></span>
  <span aria-owns=":r4:" ...></span>            <!-- this one -->
  <span data-type="outside" tabindex="0" aria-hidden="true" data-floating-ui-focus-guard ...></span>
</div>
```

and axe reports:

```
aria-required-children: Element has children which are not allowed: span[aria-owns]
```

The focus guards themselves are fine — they're `aria-hidden="true"`, so axe ignores them. It's only the `aria-owns` span, which isn't hidden by design (hiding it would defeat its purpose).

I've pushed a rewrite of the comment in 78e5d48 so it names the actual element and says the popover has to be open — the old wording ("tab-order and focus-guard scaffolding") was vague enough to look unjustified. If you'd rather we chase the root cause, the option I see is teaching `Floating` to place that span outside the anchor's parent, which means diverging from floating-ui's focus manager — happy to file that separately if you think it's worth it.

<sub>One aside in case you try to repro in a unit test: jsdom reports `navigator.vendor` as `"Apple Computer, Inc."`, which makes floating-ui's `isSafari()` true and adds `role="button"` to the focus guards. That produces two extra `aria-command-name` violations that do **not** happen in Chromium. Stub `navigator.vendor` to `"Google Inc."` to match the browser the Storybook a11y test uses.</sub>

.changeset/popover-floating-ui.md
jandrade: Agreed on both counts, and thanks for pushing on it — the changeset was describing behavior that never existed. To be concrete about where we actually land:

**Focus is not cycled or trapped.** The dialog is non-modal (`FloatingFocusManager` with `modal={false}`). Tabbing past the last focusable element in the popover moves focus to the next element after the trigger in the document, and when `dismissEnabled` is set the popover closes on focus out. That matches the native `popover` behavior you linked. The only thing that changed is the *mechanism*: floating-ui's focus guards instead of Popover's own keydown handling.

**The `tabindex="0"` guards are floating-ui's, and they are not what axe fails on.** floating-ui renders them `aria-hidden="true"`, so axe skips them. The one element that does trip a rule is the `<span aria-owns>` next to the anchor, and only when the anchor sits inside a `tablist` — written up in the tabs thread.

So I don't think there's cycling to remove here. Corrected the changeset bullet in dbe796e:

> Focus management is now handled by floating-ui's non-modal focus manager rather than the previous custom implementation. The popover remains non-modal and focus is not trapped or cycled (it never was): tabbing past the last focusable element in the popover moves focus on to the next element after the trigger, and `dismissEnabled` popovers close when focus leaves them. The mechanism behind that is floating-ui's focus guards rather than Popover's own keydown handling.

packages/wonder-blocks-popover/src/util/types.ts
jandrade: You're right, and right that it predates this PR — the story just never gave itself a way to close. It sets neither `dismissEnabled` (so no escape / outside click) nor `closeButtonVisible` (so no close button), and `dismissEnabled` defaults to `false`. Once open, that popover is stuck.

Fixed the story in 78e5d48 by adding both, matching what the other stories do:

```jsx
<Popover
    rootBoundary="document"
    dismissEnabled
    content={() => (
        <PopoverContent
            closeButtonVisible
            title="Popover with rootBoundary='document'"
            ...
        />
    )}
    placement="top"
>
```

The broader footgun is real though: a `Popover` with `dismissEnabled` unset and no close button has no dismissal path at all, and nothing warns you. I'd rather not change the default in this PR since it's a behavior change for every consumer, but it seems worth its own ticket — either flipping `dismissEnabled` to default `true` or dev-warning when a popover has no way to close. Want me to file that?

packages/wonder-blocks-popover/src/util/types.ts
jandrade: They should be, and they were meant to be — the argType was just written in a shape Storybook no longer reads. `options` was nested inside `control`:

```js
placement: {
    control: {
        type: "select",
        options: ["top", "bottom", "right", "left"],   // ignored
    },
    ...
}
```

`options` has to sit at the top level of the argType. Nested under `control` it's silently dropped, so the select had no choices and only rendered the current value — which is the `top` you were seeing. `rootBoundary` had the same bug.

Fixed both in c07a402, and spelled the unions out in the Type column while I was there (`Placement` on its own wasn't telling you much):

```js
placement: {
    control: {type: "select"},
    options: ["top", "bottom", "right", "left"],
    table: {
        type: {summary: `"top" | "bottom" | "right" | "left"`},
        defaultValue: {summary: "top"},
    },
},
```

Note this is pre-existing rather than something this PR introduced. `__docs__/wonder-blocks-core/view.argtypes.ts` has the same nested-`options` bug for `tag` and `dir` — left alone here to keep the diff Popover-only, but happy to file it.

packages/wonder-blocks-popover/src/components/popover-content-core.tsx
jandrade: Yes — that difference is real and worth calling out. Added it in 14ca98a, on the component's JSDoc so the `PopoverContentCore` docs page picks it up above the examples:

> ### Presentation
>
> `PopoverContentCore` only lays out the popover's contents (max width, padding and the optional close button). The popover "bubble" chrome — background, border, border radius, shadow and the tail — is drawn by the `Floating` component that `Popover` renders it inside.
>
> This means that when `PopoverContentCore` is rendered **standalone**, outside of a `Popover` — which is how the examples on this page are rendered — it has no bubble around it and you need to supply your own container styling. Inside a `Popover` it looks like a popover with no extra work.

So your two screenshots are both correct: prod drew the chrome in `PopoverContentCore` itself, this branch moves it up to `Floating`. Inside a `Popover` nothing changes visually (Chromatic shows 0 visual changes); standalone usage is the case that regresses, which is also called out as a breaking change in the changeset.

PR #3164, opened 2026-08-03
🚨 [CSS Modules] Migrate Button to CSS Modules

jandrade:
Migrate `Button` from Aphrodite to CSS Modules (CSS Modules Phase 4 of epic
WB-2120). This is a spike on the worst-case variant matrix (~36 cells across
`kind × actionType × size`) to validate the migration patterns before the
component waves. The public API is unchanged — same props, DOM structure, and
`style` / `styles` / `labelStyle` overrides — so this is an internal styling
refactor. `ActivityButton` (which has its own `_generateStyles`) is deferred.

Styling is organised around a **component-token surface**, following the
convention `NodeIconButton` established with `--wb-c-node-icon-button--`: every
value a variant axis can change is a `--wb-c-button--*` custom property, and
each rule does exactly one of two jobs, never both.

- **Variant classes assign tokens.** One class per axis value (`.primary`,
  `.progressive`, `.small`) sets token values and nothing else, so all three
  axes compose the same way — as class-name strings through the WB `style` prop
  via `processStyleList` (no `clsx`).
- **Base and state rules only read tokens.** Each interactive state is declared
  once instead of once-per-cell of the matrix.

Colour is the one place the axes aren't orthogonal — the semantic tokens are
keyed by `kind` *and* `actionType`
(`--wb-semanticColor-action-<kind>-<actionType>-<state>-*`) — so those nine
cells are spelled out as compound `.primary.progressive` rules. Collapsing them
would mean interpolating token names, which only JS can do.

State rules qualify the kind with `:where(…)`, which contributes no
specificity. Every state rule therefore sits at the same specificity and source
order alone decides precedence, declared hover → active → focus → disabled:
focus has to survive hover/press, and disabled resets everything. `data-kind`
is still set on the element as a consumer/test hook, but no longer drives
styling.

Theming is unchanged — the module references the same `--wb-*` token vars that
switch on `[data-wb-theme]`. The shared reset moves to a CSS Module too, emitted into the
nested layer `shared.reset` so the component styles — which sit directly in
`shared` — outrank it regardless of the order the bundler emits the two
stylesheets in. The package now ships
`dist/index.css` (auto side-effect import) and exposes it via the new
`@khanacademy/wonder-blocks-button/css` subpath.

Issue: WB-2328

## Test plan:

1. Review the Chromatic diff on this PR; any diff must be intentional (target: none).
2. In Storybook, spot-check `Button` across `kind` (primary/secondary/tertiary) ×
   `actionType` (progressive/destructive/neutral) × `size` (small/medium/large) and
   the hover / active / focus-visible / disabled states, in the default,
   thunderblocks, and syl-dark themes.
3. Specifically check the focus ring while *also* hovering and while pressed —
   that combination is what the `:where(…)` specificity flattening protects, for
   every `kind` (primary's `outline`, secondary/tertiary's inset `box-shadow`).
4. Check a disabled button's focus ring, and that hover/press leave a disabled
   button visually unchanged.
5. **Confirm secondary and tertiary buttons show a visible *rest-state* border**
   (all three themes). This one is not covered by CI: the reset now sits in the
   nested layer `shared.reset`, and that only beats the reset's late position in
   the bundle if rules directly in a layer outrank its named sub-layers — CSS
   Cascade 5 behaviour that could not be verified outside a browser. See the
   comment on `button-unstyled.module.css` for the fallback if it's wrong.
`/?path=/docs/packages-button-button--docs`

## Review plan:

1. 🚨 [`button.module.css`](https://github.com/Khan/wonder-blocks/pull/3164#discussion_r3707838201)
2. 🚨 [`button-core.tsx`](https://github.com/Khan/wonder-blocks/pull/3164#discussion_r3707838315)
3. ⚠️ [`button-unstyled.tsx`](https://github.com/Khan/wonder-blocks/pull/3164#discussion_r3707838431)
4. ⚠️ [`package.json`](https://github.com/Khan/wonder-blocks/pull/3164#discussion_r3707838533)
5. ⚠️ [`button-unstyled.module.css`](https://github.com/Khan/wonder-blocks/pull/3164#discussion_r3732194754)

### Common patterns:

**2 Files (button-core.tsx, button-unstyled.tsx):** Replace Aphrodite `StyleSheet.create` / `_generateStyles` with a colocated `*.module.css` and compose the resulting class-name strings through the Wonder Blocks `style` prop (routed to `className` by `processStyleList`).

```tsx
// Before
const styles = StyleSheet.create({reset: {position: "relative", /* … */}});
style={[styles.reset, style]}

// After
import styles from "./button-unstyled.module.css";
style={[styles.reset, style]}
```

### DIFF
--- a/.changeset/css-modules-phase-4-button.md
+++ b/.changeset/css-modules-phase-4-button.md
@@ -18,9 +18,11 @@ styling refactor.
 - `data-kind` is still set on the rendered element and remains available as a
   consumer/test hook, but it no longer drives any styling.
 - The shared element reset (`button-unstyled`) also moves to a CSS Module
-  (`button-unstyled.module.css`) so the reset lives in `@layer shared` alongside
-  the component styles rather than as unlayered Aphrodite (which would otherwise
-  override the layered focus ring / border / underline).
+  (`button-unstyled.module.css`) rather than staying as unlayered Aphrodite
+  (which would otherwise override the layered focus ring / border / underline).
+  It is emitted into the nested layer `shared.reset`, so the component styles —
+  which sit directly in `shared` — always outrank it regardless of the order the
+  bundler emits the two stylesheets in.
 - The package now ships its bundled stylesheet at `dist/index.css` (imported
   automatically as a side-effect of the JS entry) and exposes it explicitly via
   the new `@khanacademy/wonder-blocks-button/css` subpath. `sideEffects` is set
--- a/.changeset/css-modules-phase-4-button.md
+++ b/.changeset/css-modules-phase-4-button.md
@@ -8,11 +8,15 @@ same `style` / `styles` / `labelStyle` overrides — so this is an internal
 styling refactor.
 
 - The `kind × actionType × size` variant matrix is now expressed in
-  `button.module.css` using CSS custom-property "slots" set by the `[data-kind]`
-  attribute plus per-`actionType` classes, so each interactive state is declared
-  once instead of once-per-cell. Theming (default / thunderblocks / syl-dark) is
-  unchanged — the module references the same `--wb-*` token variables that switch
-  on `[data-wb-theme]`.
+  `button.module.css` around a component-token surface: every value a variant
+  axis can change is a `--wb-c-button--*` custom property, assigned by one class
+  per axis value (`.primary`, `.progressive`, `.small`) and read by the base and
+  state rules, so each interactive state is declared once instead of
+  once-per-cell. Theming (default / thunderblocks / syl-dark) is unchanged — the
+  module references the same `--wb-*` token variables that switch on
+  `[data-wb-theme]`.
+- `data-kind` is still set on the rendered element and remains available as a
+  consumer/test hook, but it no longer drives any styling.
 - The shared element reset (`button-unstyled`) also moves to a CSS Module
   (`button-unstyled.module.css`) so the reset lives in `@layer shared` alongside
   the component styles rather than as unlayered Aphrodite (which would otherwise
@@ -23,3 +27,11 @@ styling refactor.
   so bundlers keep the side-effect import. Standard webpack / Vite / Next.js
   setups pick this up with no changes; SSR consumers that can't process CSS
   imports may need a CSS loader / mock in their build.
+- Note for consumers who override Button styles: Aphrodite emitted these rules
+  unlayered and with `!important`, whereas the CSS Modules build emits them in
+  `@layer shared`. Unlayered consumer CSS now wins over Button's own styles
+  regardless of specificity, so overrides that previously needed `!important`
+  may no longer, and stylesheets that were previously losing to Button may now
+  start taking effect. Overrides passed through the `style` / `styles` /
+  `labelStyle` props are unaffected — those still route through Aphrodite and
+  continue to win.
--- a/.changeset/css-modules-phase-4-button.md
+++ b/.changeset/css-modules-phase-4-button.md
@@ -1,0 +1,25 @@
+---
+"@khanacademy/wonder-blocks-button": minor
+---
+
+Migrate `Button` from Aphrodite to CSS Modules (WB-2328, CSS Modules Phase 4).
+The public component API is unchanged — same props, same DOM structure, and the
+same `style` / `styles` / `labelStyle` overrides — so this is an internal
+styling refactor.
+
+- The `kind × actionType × size` variant matrix is now expressed in
+  `button.module.css` using CSS custom-property "slots" set by the `[data-kind]`
+  attribute plus per-`actionType` classes, so each interactive state is declared
+  once instead of once-per-cell. Theming (default / thunderblocks / syl-dark) is
+  unchanged — the module references the same `--wb-*` token variables that switch
+  on `[data-wb-theme]`.
+- The shared element reset (`button-unstyled`) also moves to a CSS Module
+  (`button-unstyled.module.css`) so the reset lives in `@layer shared` alongside
+  the component styles rather than as unlayered Aphrodite (which would otherwise
+  override the layered focus ring / border / underline).
+- The package now ships its bundled stylesheet at `dist/index.css` (imported
+  automatically as a side-effect of the JS entry) and exposes it explicitly via
+  the new `@khanacademy/wonder-blocks-button/css` subpath. `sideEffects` is set
+  so bundlers keep the side-effect import. Standard webpack / Vite / Next.js
+  setups pick this up with no changes; SSR consumers that can't process CSS
+  imports may need a CSS loader / mock in their build.
--- a/packages/wonder-blocks-button/package.json
+++ b/packages/wonder-blocks-button/package.json
@@ -19,12 +19,16 @@
   "main": "dist/index.js",
   "module": "dist/es/index.js",
   "types": "dist/index.d.ts",
+  "sideEffects": [
+    "**/*.css"
+  ],
   "exports": {
     ".": {
       "import": "./dist/es/index.js",
       "require": "./dist/index.js",
       "types": "./dist/index.d.ts"
     },
+    "./css": "./dist/index.css",
     "./styles.css": "./dist/css/vars.css"
   },
   "scripts": {
--- a/packages/wonder-blocks-button/src/components/button-core.tsx
+++ b/packages/wonder-blocks-button/src/components/button-core.tsx
@@ -1,5 +1,4 @@
 import * as React from "react";
-import {CSSProperties, StyleSheet} from "aphrodite";
 
 import {BodyText} from "@khanacademy/wonder-blocks-typography";
 import {View} from "@khanacademy/wonder-blocks-core";
@@ -9,19 +8,11 @@ import type {
     ChildrenProps,
     ClickableState,
 } from "@khanacademy/wonder-blocks-clickable";
-import {focusStyles} from "@khanacademy/wonder-blocks-styles";
-import {semanticColor} from "@khanacademy/wonder-blocks-tokens";
-import type {
-    ButtonActionType,
-    ButtonKind,
-    ButtonSize,
-    ButtonProps,
-    ButtonRef,
-} from "../util/button.types";
+import type {ButtonProps, ButtonRef} from "../util/button.types";
 import {ButtonIcon} from "./button-icon";
 
-import theme from "../theme";
 import {ButtonUnstyled} from "./button-unstyled";
+import styles from "./button.module.css";
 
 type Props = ButtonProps & ChildrenProps & ClickableState;
 
@@ -58,21 +49,23 @@ const ButtonCore: React.ForwardRefExoticComponent<
         ...restProps
     } = props;
 
-    const buttonStyles = _generateStyles(actionType, kind, size);
-
     const disabled = spinner || disabledProp;
 
+    // The `kind` and `actionType` axes of the variant matrix are selected in
+    // CSS via the `[data-kind]` attribute (set by `ButtonUnstyled`) combined
+    // with the `actionType` class below. `disabled` is selected via the
+    // `[aria-disabled="true"]` attribute (also set by `ButtonUnstyled`), which
+    // keeps the element focusable. Class-name strings are composed through the
+    // `style` prop — `processStyleList` routes them to `className`.
     const defaultStyle = [
-        sharedStyles.shared,
-        startIcon && sharedStyles.withStartIcon,
-        endIcon && sharedStyles.withEndIcon,
-        buttonStyles.default,
-        disabled && buttonStyles.disabled,
-        !disabled && pressed && buttonStyles.pressed,
-        // Enables programmatic focus.
-        !disabled && !pressed && focused && buttonStyles.focused,
-        size === "small" && sharedStyles.small,
-        size === "large" && sharedStyles.large,
+        styles.button,
+        actionType && styles[actionType],
+        size === "small" && styles.small,
+        size === "large" && styles.large,
+        // Enable the press / focus states for programmatic (keyboard)
+        // interaction tracked by `ClickableBehavior`.
+        !disabled && pressed && styles.pressed,
+        !disabled && !pressed && focused && styles.focused,
     ];
 
     const label = (
@@ -81,11 +74,11 @@ const ButtonCore: React.ForwardRefExoticComponent<
             weight={size === "large" ? "bold" : undefined}
             tag="span"
             style={[
-                sharedStyles.text,
-                size === "small" && sharedStyles.smallText,
-                size === "large" && sharedStyles.largeText,
+                styles.text,
+                size === "small" && styles.smallText,
+                size === "large" && styles.largeText,
                 labelStyle,
-                spinner && sharedStyles.hiddenText,
+                spinner && styles.hiddenText,
             ]}
             testId={testId ? `${testId}-inner-label` : undefined}
         >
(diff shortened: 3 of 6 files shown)


packages/wonder-blocks-button/src/components/button.module.css
jandrade, line 1: 🚨 **High risk.** New file defining the entire visual matrix for `Button` — a shared component used across many apps. Colour slots (`--button-bg/fg/border-*`) are set per `kind` × `actionType`; structural per-kind rules (primary `outline`, secondary/tertiary inset `box-shadow`, tertiary underline); hover/press/focus/disabled states each declared once. Ported faithfully from the earlier spike (which had 0 Chromatic diffs) with all token var names re-verified against the current token build, but visual correctness ultimately depends on the Chromatic diff on this PR.

packages/wonder-blocks-button/src/components/button-core.tsx
jandrade, line 1: 🚨 **High risk.** Rewrites the styling of a widely-consumed shared library component: removes Aphrodite `StyleSheet`/`_generateStyles` and composes CSS Module class names through the WB `style` prop (`processStyleList` routes strings to `className`). No `_generateStyles` export remains, but nothing outside this file imported it (ActivityButton has its own copy). `disabled` is now selected via `[aria-disabled="true"]` in CSS rather than a JS-applied class.

packages/wonder-blocks-button/src/components/button-unstyled.tsx
jandrade, line 1: ⚠️ **Medium risk.** Moves the shared element reset from Aphrodite to `button-unstyled.module.css` in `@layer shared`, so it no longer beats the layered component styles. The reset keeps `border: none` (as the original Aphrodite reset did) to strip the UA default border. Because the bundle emits `.reset` after `.button` at equal specificity, that would otherwise clear the secondary/tertiary rest border — so `button.module.css` re-asserts the rest border for those kinds at higher specificity via `[data-kind]` (see the dedicated rule there). Primary is unaffected (zero-width border, uses `outline`).

packages/wonder-blocks-button/package.json
jandrade, line 1: ⚠️ **Medium risk.** Adds `sideEffects: [\"**/*.css\"]` and a new `./css` → `./dist/index.css` export subpath (keeps the existing `./styles.css` → token vars). Changes how consumers pick up the component stylesheet: standard webpack/Vite/Next pick up the auto side-effect import with no change; SSR consumers that cannot process CSS imports may need a CSS loader/mock.

packages/wonder-blocks-button/src/components/button.module.css
```diff
@@ -0,0 +1,478 @@
...
+ *   thunderblocks / syl-dark) keeps working unchanged. The build wraps every
+ *   rule below in `@layer shared { … }` via the wrap-in-layer PostCSS plugin.
+ */
+
+@import "@khanacademy/wonder-blocks-styles/focus-styles.css";
+
+.button {
+    /* Layout (was `sharedStyles.shared`). Height is the medium default; the
+     * `.small` / `.large` size classes override it. */
+    block-size: var(--wb-c-button-root-sizing-height-medium);
```
jandrade, line 69: **note:** These `--wb-c-button` vars are still coming from the `theme/default.ts` files as these are transformed to CSS vars during build time.

jandrade:
### Reworked the styling structure around a component-token surface

Pushed `cd058a388`, which restructures `button.module.css` to follow the pattern
`NodeIconButton` established, rather than the `[data-kind]`-attribute approach the
earlier commits used. The PR description is updated to match. Three changes:

1. **Every rule now does exactly one job.** Variant classes (`.primary`,
   `.progressive`, `.small`) *assign* `--wb-c-button--*` tokens and nothing else;
   the base rule and the state rules only *read* them. Previously the kind axis
   was selected by the `[data-kind]` attribute while `actionType` and `size` used
   classes — now all three axes work identically and compose the same way through
   the `style` prop.

2. **State rules qualify the kind with `:where(…)`, which adds no specificity.**
   Every state rule now sits at the same specificity, so source order alone
   decides precedence — declared hover → active → focus → disabled. This is what
   the two earlier commits on this branch were fighting: `.button[data-kind]` had
   to be bolted onto the focus rule just to match the specificity of
   `.button[data-kind="primary"]:hover`. That workaround is gone.

3. **Token names are namespaced.** `--button-bg-default` → `--wb-c-button--bg-default`,
   matching `--wb-c-node-icon-button--*` and staying visibly distinct from the
   codegen'd `--wb-c-button-root-*` design tokens that these read from.

`data-kind` is still set on the element (it has its own tests and is useful as a
consumer/test hook) — it just no longer drives styling.

**No visual or API change is intended.** To check that, I normalized the old and
new modules into `selector → sorted declarations` records and diffed them: 45
rules on both sides, and every declaration block is identical. All 20 differences
are the intended selector changes — variant rules dropping the `.button`
qualifier (they only assign tokens, so they no longer need it) and the two
grouped `:where(.secondary, .tertiary)` selectors replacing the expanded
`[data-kind="secondary"], [data-kind="tertiary"]` lists.

A `tokens` prop like `NodeIconButton`'s is deliberately **not** part of this — it
would be new public API, so it can land as a follow-up.

#### Verified

- `pnpm jest packages/wonder-blocks-button` — all 5 suites pass.
- `pnpm typecheck`, `pnpm build`, and stylelint on the module — clean.
- Emitted `dist/index.css` is still wrapped in `@layer shared` and contains no
  `data-kind` selectors.

#### Not verified — needs eyes on the Chromatic diff

I could not get the Storybook story tests to complete for the button stories
locally (`run-story-tests` timed out repeatedly on `StateSheetStory` and even on
`Kinds`, though the dev server was responding). The normalized-CSS diff above is
strong evidence this is behaviour-preserving, but the specificity reasoning in
point 2 is exactly the kind of thing a static diff can't confirm. **Please treat
the Chromatic snapshots — especially the focus + hover and focus + press cells,
and the disabled states — as the real gate here.**

.changeset/css-modules-phase-4-button.md
```diff
@@ -0,0 +1,25 @@
...
+- The shared element reset (`button-unstyled`) also moves to a CSS Module
+  (`button-unstyled.module.css`) so the reset lives in `@layer shared` alongside
+  the component styles rather than as unlayered Aphrodite (which would otherwise
+  override the layered focus ring / border / underline).
+- The package now ships its bundled stylesheet at `dist/index.css` (imported
+  automatically as a side-effect of the JS entry) and exposes it explicitly via
+  the new `@khanacademy/wonder-blocks-button/css` subpath. `sideEffects` is set
+  so bundlers keep the side-effect import. Standard webpack / Vite / Next.js
+  setups pick this up with no changes; SSR consumers that can't process CSS
+  imports may need a CSS loader / mock in their build.
```
beaesguerra, line 31: Out of curiosity, is perseus already set up to automatically import the new css bundled stylesheet? I think I recall frontend would already handle this?

beaesguerra:
@claude review once

packages/wonder-blocks-button/src/components/button-core.tsx
```diff
@@ -58,21 +49,25 @@ const ButtonCore: React.ForwardRefExoticComponent<
...
-        buttonStyles.default,
-        disabled && buttonStyles.disabled,
-        !disabled && pressed && buttonStyles.pressed,
-        // Enables programmatic focus.
-        !disabled && !pressed && focused && buttonStyles.focused,
-        size === "small" && sharedStyles.small,
-        size === "large" && sharedStyles.large,
+        styles.button,
+        styles[kind],
+        actionType && styles[actionType],
```
beaesguerra: Is there type safety when accessing imported styles like this? I think there was, but thought I'd check!

packages/wonder-blocks-button/src/components/button-unstyled.module.css
```diff
@@ -0,0 +1,35 @@
+/**
+ * Reset styles shared by the rendered `<button>` / `<a>` / `Link` element of
+ * `Button` (CSS Modules migration — WB-2328).
```
beaesguerra: Did we want to keep the ticket details in the comment still? It isn't necessarily documenting future work / a to-do item so I wonder if we can remove it! (We'll have the css module migration ticket in the commit anyways!)

packages/wonder-blocks-button/src/components/button.module.css
```diff
@@ -0,0 +1,522 @@
...
+ * Every value a variant axis can change is funnelled through a
+ * `--wb-c-button--*` custom property ("component token"), following the
+ * convention `NodeIconButton` established with `--wb-c-node-icon-button--`.
+ * The double dash keeps this namespace distinct from the codegen'd design
+ * tokens these read from (`--wb-c-button-root-*`, `--wb-semanticColor-*`).
+ *
+ *   --wb-c-button--padding-inline        inline padding      (kind × size)
+ *   --wb-c-button--border-width-default  per-state widths    (kind)
+ *   --wb-c-button--border-width-hover
+ *   --wb-c-button--border-width-press
```
beaesguerra, line 15: The naming structure kind of reminds me of [bem naming](https://getbem.com/naming/) (block, element, modifier). What do you think about following the naming structure separating the element and the modifier? I find it a bit easier to read with different separators! 

For example, `--wb-c-button__border-width--default`,`--wb-c-button__border-width--hover`, `--wb-c-button__border-width--press`, etc.

packages/wonder-blocks-button/src/components/button.module.css
```diff
@@ -0,0 +1,522 @@
...
+    transition: border-radius 0.1s ease-in-out;
+}
+
+/**
+ * `kind` axis. Sets the border widths, the primary outline offset, the
+ * medium-size inline padding and the disabled-state colours (which depend on
+ * `kind` only, not `actionType`).
+ */
+.primary {
+    --wb-c-button--border-width-default: var(
```
beaesguerra, line 91: I'm curious about this layer of CSS variable definitions! Is there a reason why we don't define the CSS properties in for the `primary` style here? (for example, setting `border-width: var(--wb-c-button-root-border-width-primary-default)` directly here)

packages/wonder-blocks-button/src/components/button.module.css
beaesguerra: Will we be converting the theme styles to css modules too at some point?

beaesguerra (commented):
Nice work, I'm glad we have visual regression tests to check all the different states! I left some questions I wanted to get your thoughts on (mostly to refresh my memory as you continue this work!)

packages/wonder-blocks-button/src/components/button-unstyled.module.css
jandrade: good point! I'll remove that.

packages/wonder-blocks-button/src/components/button.module.css
jandrade: It is mostly because we reuse those "internal" CSS vars in different selectors (e.g. `.button`, `.button[aria-disabled="true"]`, etc), and these change between kinds and themes.

I think this can be simplified once we get rid of the default/OG theme.

.changeset/css-modules-phase-4-button.md
jandrade: Not yet, but I'm planning to create that at some point in the near future, following this `frontend` approach: https://github.com/Khan/frontend/pull/11912/changes#diff-2d9ab5214985b9a608b4cd47ed561a301cfd4c7d5352c7a94ee046d4e1f1af10

.changeset/css-modules-phase-4-button.md
```diff
@@ -0,0 +1,25 @@
...
+
+Migrate `Button` from Aphrodite to CSS Modules (WB-2328, CSS Modules Phase 4).
+The public component API is unchanged — same props, same DOM structure, and the
+same `style` / `styles` / `labelStyle` overrides — so this is an internal
+styling refactor.
+
+- The `kind × actionType × size` variant matrix is now expressed in
+  `button.module.css` using CSS custom-property "slots" set by the `[data-kind]`
+  attribute plus per-`actionType` classes, so each interactive state is declared
+  once instead of once-per-cell. Theming (default / thunderblocks / syl-dark) is
```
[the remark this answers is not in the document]
jandrade: updated!

packages/wonder-blocks-button/src/components/button-core.tsx
jandrade: That's correct! In this case we need to verify that `actionType` is set as it is an optional prop and CSS modules types the definitions inside that file:

Image

packages/wonder-blocks-button/src/components/button.module.css
jandrade: I left it that way to preserve the CSS variables notation as close as we could. I asked Claude about this and it's aligned with this approach: https://claude.ai/share/aa90711b-998a-44dd-b47c-386f48798ca1

packages/wonder-blocks-button/src/components/button.module.css
jandrade: I'm not sure yet as we still have the ergonomics in JS/TS that allow us to merge themes easily. We'll see once we get rid of the OG/default theme.

packages/wonder-blocks-button/src/components/button.module.css
beaesguerra: Thanks for looking into this more! That makes sense, bem is more for DOM nodes, rather than CSS variable names!

This current approach is the "Structural naming order" one right? I remembered we had a structure for naming component tokens, so it makes sense to continue that order! https://khanacademy.atlassian.net/wiki/spaces/WB/pages/3967287319/Component+tokens+structure

beaesguerra (approved):
Looks good to me! 🚀

packages/wonder-blocks-button/src/components/button-unstyled.module.css
```diff
@@ -0,0 +1,34 @@
+/**
```
jandrade, line 1: ⚠️ **The cascade here rests on an assumption I could not verify in this environment. Please confirm it visually before approving.**

**What changed.** This reset and the component styles both end up in `@layer shared` at equal specificity, and the bundler emits this file *after* `button.module.css` — so the unqualified `border: none` was winning and wiping the secondary / tertiary **rest-state border**. That was a live regression on this branch (introduced by the migration, fixed here; `main` never shipped it).

**Why not just fix the order.** Bundle order isn't a lever. `rollup-plugin-styler` sorts extracted stylesheets by the *last* position each file occupies in a breadth-first walk of the module graph (`dist/index.js:1450`, `:1469`), so a shared dependency like this one — pulled in by both `button-core` and `activity-button` — is always reached late. I tried forcing it two ways and instrumented the result:

| Attempt | `lastIndexOf` ranking |
|---|---|
| Side-effect import at the top of `src/index.ts` | button 11, **unstyled 12**, activity 13 |
| Side-effect import beside each component's own CSS | button 14, **unstyled 16**, activity 17 |

Both pushed the reset *further* down — each new import path adds another late occurrence. The plugin exposes no ordering option, and Storybook resolves CSS through Vite rather than this pipeline, so there is no single order to rely on either way.

**How the fix works.** `wrap-in-layer` skips only files already wrapped in `@layer shared`, so wrapping these rules in `@layer reset` yields `@layer shared { @layer reset { … } }` — the sub-layer `shared.reset` — with no build change. Verified from the emitted `dist/index.css`:

```
@layer shared          @char 0        <- button.module.css (directly in `shared`)
@layer shared          @char 19391
  @layer reset         @char 19406    <- this file
```

**The assumption.** This only fixes the bug if rules placed *directly* in a layer outrank that layer's *named sub-layers*. That is CSS Cascade 5 behaviour, but I could not confirm it here — Chromium won't launch in this sandbox and no cascade-layer polyfill was installable. Static analysis can't settle it either: specificity and source order both still say the reset should win, so the layer is doing all the work. It's only observable once an engine renders it.

**So the review ask:** confirm secondary and tertiary Buttons show a visible **rest-state** border, in Storybook and in the Chromatic diff, across all three themes. If those borders are missing, the sub-layer assumption is wrong — the fallback is `:where(.reset)`, which drops the reset to specificity (0,0,0) and is provable from the emitted CSS alone (I built and validated that variant before switching to the layer approach).

PR #3165, opened 2026-08-04
Fix Tooltip tail position in RTL contexts

maddy531:
## Summary

The Tooltip tail is mispositioned in RTL. Because Popover renders `TooltipTail`, this fixes Popover in RTL too, which is what [WB-2069](https://khanacademy.atlassian.net/browse/WB-2069) is really about.

## Root cause

`TooltipTail` spreads Popper.js's arrow offset straight onto its container:

- `tooltip-popper.tsx` copies `popperProps.arrowProps.style.transform` into `tailOffset`
- `tooltip-tail.tsx` spreads it via `{...offset}` onto a `position: relative` element

Popper reports that offset as a **physical** distance from the bubble's left edge. But the tail is a `position: relative` flex item, so its static origin flips to the bubble's **right** edge under `dir="rtl"`. A physical offset applied to a flipped origin puts the tail a constant `bubbleWidth - tailWidth` too far along the inline axis — **248px** at the default popover size, entirely outside the bubble.

The bubble itself is unaffected because it's `position: absolute`, which is direction-independent. Only the tail sits in flow.

This is really Popper's arrow contract: the arrow element is meant to be absolutely positioned. Using `position: relative` in flex flow is what broke it.

## Screenshots

https://github.com/user-attachments/assets/444abd84-c904-43e4-9037-0e7656a2a209

Screenshot 2026-08-04 at 3 12 26 PM
Screenshot 2026-08-04 at 3 13 24 PM
Screenshot 2026-08-04 at 3 17 44 PM
Screenshot 2026-08-04 at 3 20 21 PM
Screenshot 2026-08-04 at 3 22 00 PM

`pnpm run jest --testPathPattern="wonder-blocks-(tooltip|popover)"` passes (16 suites). `pnpm lint` and `pnpm typecheck` clean.

To see it: open `Packages / Popover / Popover` -> `In Corners`, set the toolbar direction to RTL, and open all four popovers.

[WB-509]: https://khanacademy.atlassian.net/browse/WB-509?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WB-2069]: https://khanacademy.atlassian.net/browse/WB-2069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/wb-509-tooltip-tail-rtl.md
+++ b/.changeset/wb-509-tooltip-tail-rtl.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tooltip": patch
+---
+
+Fix the Tooltip tail being mispositioned in RTL contexts. The tail is now positioned from a physical origin, matching the coordinate space Popper.js reports its offset in, instead of relying on the bubble's direction-aware flex flow. This also fixes the Popover tail, which renders `TooltipTail`.
--- a/packages/wonder-blocks-tooltip/src/components/__tests__/tooltip-tail.test.tsx
+++ b/packages/wonder-blocks-tooltip/src/components/__tests__/tooltip-tail.test.tsx
@@ -48,9 +48,14 @@ describe("TooltipTail", () => {
             expect(container).toMatchInlineSnapshot(`
 <div>
   <div
+    aria-hidden="true"
+    class=""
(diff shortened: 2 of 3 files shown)


jandrade:
@Khan/wonder-blocks: @maddy531 and I chatted about this offline. This PR will be closed as we are migrating to Floating UI, which already solves this issue.

See the related PR here: https://github.com/Khan/wonder-blocks/pull/3163

maddy531:
Thank you, @juanjuy !

PR #3168, opened 2026-08-05
[WB-2328] 🚨 [CSS Modules] Migrate ActivityButton to CSS Modules

jandrade:
Migrate `ActivityButton` from Aphrodite to CSS Modules, the piece that was
deliberately deferred when `Button` moved over (it had its own
`_generateStyles`). The public API is unchanged — same props, DOM structure and
`styles.root` / `styles.box` / `styles.startIcon` / `styles.endIcon` /
`styles.label` overrides — so this is an internal styling refactor.

The new `activity-button.module.css` follows the component-token surface
`Button` settled on: every value a variant axis can change is a
`--wb-c-activity-button--*` custom property, variant classes only *assign*
tokens, and the base + state rules only *read* them. The `actionType × kind`
colour matrix is six compound rules (`.progressive.primary`, …) because the
semantic tokens are keyed by both axes; disabled colours are keyed by `kind`
alone so they live on the kind classes. Structural values (border width, the
6px/8px/0 shadow depths) are written literally rather than codegen'd per theme,
matching the note the Aphrodite `theme` object carried — every theme reused
them.

The "chonky" box no longer gets its state styling from classes toggled in JS.
`chonkyPressed` / `chonkyDisabled` are gone; the box is styled by descendant
selectors from the root (`.button:hover .box`, `.button.pressed .box`,
`.button[aria-disabled="true"] .box`), all at the same specificity so source
order alone decides precedence: hover -> active -> disabled -> focus. The
plain `chonky` className stays on the element as a consumer/test hook, it just
no longer drives styling — same treatment `data-kind` got in `Button`.

The icon and label rules are qualified with `.box` on purpose. `PhosphorIcon`
(`.medium`) and `BodyText` (`.mediumSemi`) ship single-class rules in the same
`@layer shared`, and Aphrodite's `!important` used to settle that contest for
us; the compound selector makes the outcome depend on specificity instead of
which package's stylesheet happens to load first.

One intentional behaviour change: a disabled button that is both focused and
hovered now keeps the full focus ring. Previously the disabled `:hover` reset
was emitted after the `:focus-visible` rule at equal specificity and stripped
the outline, leaving only the inner box-shadow. That was an artifact of
Aphrodite's rule ordering, and `Button` already made the same call
(68ea60e34).

Verified: `pnpm jest packages/wonder-blocks-button` (all 5 suites), typecheck,
eslint, stylelint, prettier, and a full rollup build all pass — the build emits
the ActivityButton rules into the package's `dist/index.css` with `@apply
--wb-focus-visible` expanded and everything wrapped in `@layer shared`. Visual
parity still needs a Chromatic diff review on this PR.

Issue: WB-2328

## Test plan:

Automated checks are green and cover behavior/structure/aria:
- `pnpm jest packages/wonder-blocks-button` — all 5 suites pass
- `pnpm typecheck`, `eslint`, `stylelint` — clean
- `pnpm rollup -c ./build-settings/rollup.config.mjs` — emits the 28
  `wb-activity-button-*` selectors into `packages/wonder-blocks-button/dist/index.css`

Manual (visual — not covered by the above):
1. Review the Chromatic diff on this PR. The only expected diff is the disabled
   + focus + hover cell of the ActivityButton StateSheet (see above); anything
   else must be investigated.
2. In Storybook, spot-check `ActivityButton` across `kind`
   (primary/secondary/tertiary) x `actionType` (progressive/neutral) and the
   hover / active / focus-visible / disabled states, in the default,
   thunderblocks and syl-dark themes. Confirm the box still lifts on hover and
   drops flat on press, and that the `WithCustomStyles` story's overrides
   (root/box/startIcon/endIcon/label) still take effect.
   http://localhost:6061/?path=/docs/packages-button-activitybutton--docs

## Review plan:

1. 🚨 [`activity-button.module.css`](https://github.com/Khan/wonder-blocks/pull/3168#discussion_r3721886693)
2. ⚠️ [`activity-button.tsx`](https://github.com/Khan/wonder-blocks/pull/3168#discussion_r3721886876)

### DIFF
--- a/packages/wonder-blocks-button/src/components/activity-button.module.css
+++ b/packages/wonder-blocks-button/src/components/activity-button.module.css
@@ -1,0 +1,380 @@
+/**
+ * ActivityButton styles (CSS Modules migration — WB-2328 / CSS Modules Phase 4).
+ *
+ * Component-token surface
+ * -----------------------
+ * Every value a variant axis can change is funnelled through a
+ * `--wb-c-activity-button--*` custom property ("component token"), following
+ * the convention `NodeIconButton` established with `--wb-c-node-icon-button--`
+ * and `Button` with `--wb-c-button--`.
+ *
+ *   --wb-c-activity-button--box-border-width      structural (all variants)
+ *   --wb-c-activity-button--box-shadow-y-rest
+ *   --wb-c-activity-button--box-shadow-y-hover
+ *   --wb-c-activity-button--box-shadow-y-press
+ *   --wb-c-activity-button--label-color           root text colour (actionType)
+ *   --wb-c-activity-button--box-bg-rest           box colours (actionType × kind)
+ *   --wb-c-activity-button--box-fg-rest
+ *   --wb-c-activity-button--box-border-rest
+ *   --wb-c-activity-button--box-shadow-color-rest
+ *   --wb-c-activity-button--box-bg-hover
+ *   --wb-c-activity-button--box-fg-hover
+ *   --wb-c-activity-button--box-border-hover
+ *   --wb-c-activity-button--box-shadow-color-hover
+ *   --wb-c-activity-button--box-bg-press
+ *   --wb-c-activity-button--box-fg-press
+ *   --wb-c-activity-button--box-border-press
+ *   --wb-c-activity-button--box-shadow-color-press
+ *   --wb-c-activity-button--box-bg-disabled       disabled colours (kind)
+ *   --wb-c-activity-button--box-fg-disabled
+ *   --wb-c-activity-button--box-border-disabled
+ *   --wb-c-activity-button--box-shadow-color-disabled
+ *
+ * Every rule below does exactly one of two jobs, never both:
+ *
+ * - **Variant classes assign tokens.** One class per axis value
+ *   (`.progressive`, `.primary`) sets token values and nothing else.
+ *   `activity-button` composes them through the existing `style` prop (Wonder
+ *   Blocks' `processStyleList` routes class-name strings to `className`), so
+ *   the axes stay orthogonal and no `clsx` dependency is needed.
+ * - **Base and state rules only read tokens.** `.button` / `.box` plus one
+ *   rule per interactive state, so each state is declared once rather than
+ *   once-per-cell of the variant matrix.
+ *
+ * Colour is the one place the axes aren't orthogonal: the semantic tokens are
+ * keyed by `actionType` *and* `kind`
+ * (`--wb-semanticColor-chonky-<actionType>-<property>-<kind>-<state>`), so
+ * those six cells are spelled out as compound `.progressive.primary` rules.
+ * Collapsing them would mean interpolating token names, which only JS can do.
+ * Disabled colours are keyed by `kind` alone, so they live on the kind classes.
+ *
+ * The token values themselves are *not* codegen'd per theme (unlike
+ * `--wb-c-button-root-*`): every theme reuses the same structural values, so
+ * they're written literally here, matching the note on the `theme` object the
+ * Aphrodite version carried. Colour tokens still resolve against the per-theme
+ * `[data-wb-theme]` blocks shipped in `@khanacademy/wonder-blocks-tokens`, so
+ * theming (default / thunderblocks / syl-dark) keeps working unchanged.
+ *
+ * Specificity: the box state rules all sit at (0,3,0) — `.button` plus a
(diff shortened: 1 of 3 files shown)


packages/wonder-blocks-button/src/components/activity-button.module.css
```diff
@@ -0,0 +1,380 @@
+/**
```
jandrade, line 1: 🚨 **High risk — new file defining the whole visual matrix.**

This file is the entire visual output for `ActivityButton` (colour tokens, the chonky box, and the hover / press / disabled / focus states) for a component shipped to every consuming app. Nothing in CI verifies the rendered pixels, so correctness here rests on the Chromatic diff.

What to look at closely:

1. **The colour matrix (6 compound rules).** Each `.progressive.primary`-style rule maps to `semanticColor.chonky[actionType][property][kind][state]` in the deleted `_generateStyles`. Worth spot-checking a couple of cells against the old JS — a transposed `kind`/`actionType` would be invisible in one theme and obvious in another.
2. **Source order is load-bearing.** The box state rules all sit at (0,3,0) specificity, so precedence comes from declaration order alone: hover → active → disabled → focus. Reordering any of these blocks silently changes behavior.
3. **`.box .icon` / `.box .label` are compound on purpose.** `PhosphorIcon` (`.medium`, 2.4rem) and `BodyText` (`.mediumSemi`) ship single-class rules into the same `@layer shared`. Aphrodite's `!important` used to settle that contest; without the `.box` qualifier the winner would come down to which package's stylesheet loads first.
4. **Intentional behavior change.** A disabled button that is both focused *and* hovered now keeps the full focus ring (`.button[aria-disabled="true"]:focus-visible`). Previously the disabled `:hover` reset was emitted after `:focus-visible` at equal specificity and stripped the outline, leaving only the inner box-shadow — an artifact of Aphrodite's rule ordering rather than a design decision. `Button` already made the same call in 68ea60e34.

**Expected Chromatic result:** exactly one diff, in the `hoverAndFocus` cell of the disabled column on the ActivityButton StateSheet. Any other diff needs investigating before this lands.

packages/wonder-blocks-button/src/components/activity-button.tsx
```diff
@@ -1,26 +1,19 @@
 import * as React from "react";
```
jandrade, line 1: ⚠️ **Medium risk — restyles a shared component; the box's state classes are no longer toggled in JS.**

`_generateStyles`, the local `theme` object and the `aphrodite` / `focusStyles` / `tokens` imports are all gone (−235 lines). Styling now comes from `activity-button.module.css` composed through the `style` prop — `processStyleList` routes class-name strings to `className`, so no `clsx`.

The behavioural bits worth a careful read:

1. **`chonkyStyles` collapsed to `[styles.box, stylesProp?.box]`.** `chonkyPressed` / `chonkyDisabled` no longer exist as JS-toggled classes; the box is styled by descendant selectors from the root (`.button:hover .box`, `.button.pressed .box`, `.button[aria-disabled="true"] .box`). Consumer overrides still win, because `stylesProp?.box` goes through Aphrodite (unlayered + `!important`) and beats everything in `@layer shared`.
2. **`disabled` is now selected in CSS, not JS.** The `[aria-disabled="true"]` attribute comes from `ButtonUnstyled`, which keeps the element focusable. Note `ButtonUnstyled` renders a `<button>` (not an `<a>`) whenever `disabled` is set, even with an `href`, so the attribute is always present when it needs to be.
3. **`pressed` / `focused` still gate on `!disabled` in JS**, matching the old array exactly — there is no `.button[aria-disabled="true"].focused` rule in the module because that combination can't occur.
4. **The plain `chonky` className is kept deliberately** as a consumer/test hook even though it no longer drives styling — the same treatment `data-kind` got in the `Button` migration.

Behaviour is covered by the existing `activity-button.test.tsx` suite (roles, mouse handlers, icon `aria-hidden`), which passes unchanged. Visual parity is not covered — see the Chromatic note on the CSS file.

beaesguerra:
@claude review once

packages/wonder-blocks-button/src/components/activity-button.module.css
beaesguerra: Interesting, there wasn't a diff in Chromatic! I checked in Storybook and things seem to be looking okay though! Any thoughts on this?

beaesguerra (approved):
Looks good to me!

PR #3170, opened 2026-08-07
Update SYL Dark action.secondary.neutral background tokens

beaesguerra:
## Summary
Updates the SYL Dark theme's `action.secondary.neutral` background tokens to use `core.background.base.default` instead of `core.transparent` across all interactive states.

This makes it so secondary neutral buttons have a fill. This is also consistent with the thunderblocks theme

Issue: WB-2419

## Test Plan
Confirm secondary neutral buttons in SYL dark have `core.background.base.default` as the background color now
image


marcysutton (approved):
Looks good to me!

PR #3174, opened 2026-08-11
[CLASS-13991] Auto-mirror directional PhosphorIcons in RTL

maddy531:
## Summary

`PhosphorIcon` now auto-mirrors icons whose glyph encodes a **reading** direction in RTL, driven by a central allowlist.
`mirrorInRtl` overrides the allowlist per call site in both directions:
- `true` mirrors a custom directional SVG, `false` opts an allowlisted icon out.
- `false` is for absolute directions — e.g. perseus's graph-editor arrows, where "left" means decreasing x in every locale.

Issue: CLASS-13991
Follow-up: CLASS-14877 (same support in the `Icon` component)

## Breaking change

Released as **major**. Any consumer doing `isRTL ? caretLeft : caretRight` will double-flip in RTL after upgrading, and must pass the LTR-facing icon instead (or `mirrorInRtl={false}`).

## Rollout

1. **wonder-blocks** — this PR. Publish as major.
2. **perseus** — bump `wonder-blocks-icon`, drop the 2 `scrollable-view.tsx` swaps, and opt the graph editor's x-axis arrows out with `mirrorInRtl={false}`. Publish.
- Khan/perseus#4104
3. **frontend** — one PR: bump `@khanacademy/perseus` **and** `@khanacademy/wonder-blocks-icon` together (both are exact-pinned in the same `pnpm-workspace.yaml` catalog), and remove the 11 caret/arrow swaps. 
- Khan/frontend#15137

## Test plan

- [x] `pnpm test`
- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] Chromatic — `ExpectedToMirror` / `ExpectedNotToMirror` cover the allowlist
      in LTR and RTL panels; `CustomIconMirrorInRtl` covers the override


__docs__/wonder-blocks-icon/icons/custom-forward.svg
maddy531, line 1: I added a custom icon for testing purposes/for storybook.

.changeset/phosphor-icon-rtl-mirroring.md
beaesguerra, line 1: Based on the rollout plan in the PR decription, I wonder if we need to treat this like a new `major` version, since it could "break" things if consumers don't make changes on their end (by break, I mean show the wrong direction for icons!). Especially because it is trickier to release a WB version that needs changes in both frontend and perseus

I think we need to release these changes in stages for it to be non-breaking. For example:
1. We add the `mirrorInRtl` prop to PhosphorIcon, but it doesn't do anything yet. We release it, and update Perseus to pass in `mirrorInRtl=false` so that the new behaviour (that will be added later) doesn't break it. Release Perseus.
2. Then, we add in the automatic icon mirroring functionality in WB and support the `mirrorInRtl` prop. Release WB + integrate/deploy in frontend. isRTL checks for icons in frontend can be done in that same PR. Deploying this won't break Perseus since Perseus is already using the `mirrorInRtl=false` prop.
3. After frontend is released, remove the rtl conditionals and `mirrorInRtl=false` in Perseus since the new functionality will be in place already

__docs__/wonder-blocks-icon/accessibility.mdx
```diff
@@ -36,6 +37,10 @@ More information about all these points can be found in the
 
 ### Demo: Right-to-left icons
 
+Directional carets flip automatically in RTL. For the full whitelist and
```
beaesguerra, line 40: nit: What do you think about replacing `whitelist` references to `allowlist`? To align with [inclusive language norms](https://khanacademy.atlassian.net/wiki/x/rQBZMg)! 
```suggestion
Directional carets flip automatically in RTL. For the full allowlist and
```

(same with other references to `whitelist` in the PR!)

__docs__/wonder-blocks-icon/phosphor-icon-rtl.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+### Not mirrored
+
+- Media controls (`play`, `fast-forward`, `rewind`, `skip-forward`, `skip-back`)
+- Rotation (`arrow-clockwise`, `arrow-counter-clockwise`)
+- Charts / trends (`trend-up`, `trend-down`)
+- Vertical or symmetric glyphs (`arrows-down-up`, `arrows-left-right`,
+  `text-align-center`, `text-align-justify`)
+- Physical text alignment (`text-align-left`, `text-align-right`)
+- Launch / pop-out (`arrow-square-out`, `arrow-up-right`,
```
beaesguerra, line 43: Out of curiosity, do you know why arrow-up-right doesn't need to be flipped in RTL? I would have thought anything related to left/right would need to be flipped!

`arrow-square-out` is probably the icon that should get used to represent launching in a new window / pop-outs!

__docs__/wonder-blocks-icon/phosphor-icon-rtl.stories.tsx
```diff
@@ -0,0 +1,303 @@
...
+ */
+export const CommonPatterns: StoryComponentType = {
+    render: () => (
+        <Comparison>
+            <View style={styles.stack}>
+                <PaginationRow />
+                <ListRows />
+            </View>
+        </Comparison>
+    ),
```
beaesguerra, line 203: I think we can disable the chromatic snapshot for this story since this is more of an example use case! We'll have coverage from the other stories :) 

This can be done by adding the following to the story: 
```
  parameters: {
        chromatic: {
            disableSnapshot: true,
        },
    },
```

__docs__/wonder-blocks-icon/phosphor-icon.argtypes.ts
```diff
@@ -227,4 +227,16 @@ export default {
...
+            "When true, mirror in RTL even if the icon is not on the directional whitelist (e.g. custom directional SVGs).",
+        control: {
+            type: "boolean",
+        },
+        table: {
+            type: {
+                summary: "boolean",
+            },
+        },
+    },
```
beaesguerra, line 241: I believe this shouldn't be needed! For simple prop types, the prop description and type should be extracted automatically from the js docs!

packages/wonder-blocks-icon/src/components/__tests__/phosphor-icon.test.tsx
```diff
@@ -202,6 +202,83 @@ describe("PhosphorIcon", () => {
         });
     });
 
+    describe("RTL mirroring", () => {
```
beaesguerra, line 205: These tests can be removed since we use the visual regression tests in Chromatic for test coverage related to styling!

packages/wonder-blocks-icon/src/components/phosphor-icon.tsx
```diff
@@ -55,6 +56,13 @@ type Props = Pick<AriaProps, "aria-hidden" | "aria-label" | "role"> & {
      * - `string`: an import referencing an arbitrary SVG file.
      */
     icon: PhosphorIconAsset | string;
+
+    /**
+     * When true, mirror this icon in RTL even if it is not on the directional
+     * whitelist (e.g. a custom directional SVG). Defaults to whitelist
+     * behavior when omitted. Does not disable mirroring for whitelisted icons.
```
beaesguerra, line 63: > Does not disable mirroring for whitelisted icons.

What do you think about letting `mirrorInRtl` override the built-in functionality? This would be needed for the rollout strategy I mentioned in the review! 

I wonder too if there would be certain use cases where overriding is needed

packages/wonder-blocks-icon/src/components/phosphor-icon.tsx
```diff
@@ -55,6 +56,13 @@ type Props = Pick<AriaProps, "aria-hidden" | "aria-label" | "role"> & {
...
+     */
+    mirrorInRtl?: boolean;
```
beaesguerra, line 65: Idea for later: It would be great to support the `mirrorInRtl` prop in the `Icon` component too (which is used for custom icons that are multi-colored)!

beaesguerra:
@claude review once

beaesguerra (commented):
Thank you for working on this @maddy531! This is a great idea, and will improve our rtl support! 

I left some comments in the PR - The implementation overall looks good to me, though I wanted to discuss more how this will be rolled out since there are considerations for breaking changes!

.changeset/phosphor-icon-rtl-mirroring.md
beaesguerra: I'll raise this in our thread too so we can discuss further :)

packages/wonder-blocks-icon/src/components/phosphor-icon.tsx
maddy531: Yes! Good idea!

.changeset/phosphor-icon-rtl-mirroring.md
maddy531: Great! I will update the PR to reflect the updated plan.

__docs__/wonder-blocks-icon/phosphor-icon-rtl.mdx
maddy531: Good catch, adding arrow-square-up-left to the not-mirrored list.

These are excluded because they're action symbols ("opens elsewhere" / export) rather than direction symbols. Mirroring a box-with-arrow-out can also invert the meaning, since an arrow going into a box reads as import.

packages/wonder-blocks-icon/src/components/phosphor-icon.tsx
maddy531: Created a ticket for this idea!

maddy531:
@claude review once

__docs__/wonder-blocks-icon/phosphor-icon-rtl.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+`mirrorInRtl={false}` suppresses mirroring for an allowlisted icon. This is an
+escape hatch for call sites that still do their own RTL icon swap and would
+otherwise double-flip — prefer removing the swap instead.
+
+```tsx
+<PhosphorIcon icon={caretRightIcon} mirrorInRtl={false} />
+```
+
+<Canvas of={PhosphorIconRtlStories.CustomIconMirrorInRtl} />
```
jandrade: **suggestion:** Could you please add a title and a description for how this is used with custom icons? thanks!

packages/wonder-blocks-icon/src/util/mirrored-icon-names.ts
```diff
@@ -0,0 +1,82 @@
+/**
+ * Icon names mirrored in RTL by `PhosphorIcon`. Single source of truth —
+ * add or remove names in a Wonder Blocks PR rather than special-casing a call
+ * site. `mirrorInRtl` exists for the exceptions. See the PhosphorIcon RTL docs
+ * for deliberate exclusions (media, launch, trends, physical text-align, etc.).
```
jandrade, line 5: **suggestion:** I'd recommend adding a link to https://phosphoricons.com/ so folks know what icon names are expected to be included.

__docs__/wonder-blocks-icon/phosphor-icon-rtl.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+`PhosphorIcon` automatically mirrors icons whose glyph encodes a **reading**
+direction when rendered in right-to-left content. Layout position already
+mirrors via logical CSS; this flips the glyph itself.
+
+Which icons mirror is decided by the allowlist in
+`packages/wonder-blocks-icon/src/util/mirrored-icon-names.ts`. To add or remove
+an icon, open a Wonder Blocks PR against that list rather than handling RTL
+mirroring at the call site. For genuine one-offs, `mirrorInRtl` overrides the
+allowlist in either direction.
```
jandrade, line 21: **suggestion:** I'd recommend adding a link to https://phosphoricons.com/ so folks know what icon names are expected to be included.

jandrade (approved):
Looks great! Thanks for adding this useful functionality in WB 👏 🚀

maddy531:
Hi @jandrade ,

Thanks for the review! I'm seeing some tooltip diffs in the UI tests and I wanted to check in with you before merging my PR to see if this is a known issue.

PR #3176, opened 2026-08-12
Allow more aria attrs on combobox; consume style

nedredmond:
## Problem
I tried to use the combobox like this, and realized that while React allowed me to pass these attributes, they did nothing:

```
...
const id = useId()

return (
    <LabeledField
        id={id}
        field={
            <Combobox
                aria-labelledby={${id}-label}
                aria-describedby={caption ? ${id}-description : undefined}
                selectionType="multiple"
                value={values}
                onChange={handleChange}
            >
                {options.map((option) => (
                    <OptionItem
                        key={option.id}
                        label={option.label}
                        value={option.value}
                    />
                ))}
            </Combobox>
        }
        label={title}
        description={caption}
    />
);
};
```

Then I realized that unlike many of our WB components, the combobox didn't take in all aria-props and pass them through to the underlying component. It specifically only accepted aria-label. This felt fishy to me and also made it incompatible with `LabeledField` features.

## Solution
I am replacing that prop with the AriaAttributes interface so that any aria override will be respected.

Additionally, I noticed that two other props, `loading` and `style`, were completely unused. While `loading` has a comment indicating it's a WIP, `style` had a comment that lied! So I made the comment true by adding it to the combobox text input.

Please let me know if the exclusions were on purpose! I know the combobox has some sensitive aria machinery, but people won't add those attributes unless they know what they are doing, and this allows us to fully take advantage of `LabeledField`.

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/combobox.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/combobox.tsx
@@ -4,7 +4,11 @@ import * as React from "react";
 import caretDownIcon from "@phosphor-icons/core/regular/caret-down.svg";
 import xIcon from "@phosphor-icons/core/regular/x.svg";
 
-import {StyleType, View} from "@khanacademy/wonder-blocks-core";
+import {
+    type AriaAttributes,
+    type StyleType,
+    View,
+} from "@khanacademy/wonder-blocks-core";
 import {TextField} from "@khanacademy/wonder-blocks-form";
 import IconButton from "@khanacademy/wonder-blocks-icon-button";
 import {border, semanticColor, sizing} from "@khanacademy/wonder-blocks-tokens";
@@ -131,11 +135,6 @@ type Props = {
     startIcon?: React.ReactElement<
         React.ComponentProps<typeof PhosphorIcon>
     > | null;
-
-    /**
-     * An optional aria-label to display on the combobox.
-     */
-    "aria-label"?: string;
 };
 
 /**
@@ -165,8 +164,10 @@ export default function Combobox({
     startIcon,
     testId,
     value = "",
-    "aria-label": ariaLabel,
-}: Props) {
+    loading,
+    style,
+    ...ariaAttrs
+}: Props & AriaAttributes) {
     // eslint-disable-next-line import/no-deprecated
     const generatedUniqueId = useId();
     const uniqueId = id ?? generatedUniqueId;
@@ -586,10 +587,9 @@ export default function Combobox({
                 {maybeRenderStartIcon()}
 
                 <TextField
-                    aria-label={ariaLabel}
                     id={textFieldId}
                     testId={testId}
-                    style={styles.combobox}
+                    style={[styles.combobox, style]}
                     value={inputValue}
                     onChange={handleTextFieldChange}
                     disabled={disabled}
@@ -613,6 +613,7 @@ export default function Combobox({
                     // the combobox is already providing suggestions.
                     autoComplete="off"
                     role="combobox"
+                    {...ariaAttrs}
                 />
 
                 {inputValue && !disabled && (
--- a/.changeset/gentle-shoes-arrive.md
+++ b/.changeset/gentle-shoes-arrive.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-dropdown": minor
+---
+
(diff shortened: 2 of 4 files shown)


packages/wonder-blocks-dropdown/src/components/combobox.tsx
```diff
@@ -165,8 +164,10 @@ export default function Combobox({
     startIcon,
     testId,
     value = "",
-    "aria-label": ariaLabel,
-}: Props) {
+    loading,
+    style,
+    ...ariaAttrs
```
beaesguerra, line 169: Could we add a unit test to verify aria attributes get set on the text field?

packages/wonder-blocks-dropdown/src/components/combobox.tsx
beaesguerra, line 1: From the PR description, you mentioned trying to set aria attributes on the combobox like this:
```
 <LabeledField
        id={id}
        field={
            <Combobox
                aria-labelledby={${id}-label}
                aria-describedby={caption ? ${id}-description : undefined}
                selectionType="multiple"
                value={values}
                onChange={handleChange}
            >
                {options.map((option) => (
                    <OptionItem
                        key={option.id}
                        label={option.label}
                        value={option.value}
                    />
                ))}
            </Combobox>
        }
        // ...
/>
```

Was there a reason why aria-labelledby and aria-describedby needed to be set explicitly on Combobox? I'm wondering if you ran into a limitation with the current implementation of LabeledField. Right now, it'll generate a unique id by default and wire up the [labelling](https://github.com/Khan/wonder-blocks/blob/3cc7eca9b5cd224e4b4a323c32c6b45ef28cfbc9/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx#L163-L164) and the [aria-describedby](https://github.com/Khan/wonder-blocks/blob/3cc7eca9b5cd224e4b4a323c32c6b45ef28cfbc9/packages/wonder-blocks-labeled-field/src/components/labeled-field.tsx#L259-L265) attribute for you! So you could use it like this:

```
<LabeledField
    field={
        <Combobox
            selectionType="multiple"
            value={values}
            onChange={handleChange}
        >
            {options.map((option) => (
                <OptionItem
                    key={option.id}
                    label={option.label}
                    value={option.value}
                />
            ))}
        </Combobox>
      }
      label={label}
      description={description}
/>

```

Then with your changes in this PR, it'll make it so the attributes injected by LabeledField will apply to the combobox!

beaesguerra, line 1: For the [combobox.argtypes.ts](https://github.com/Khan/wonder-blocks/blob/main/__docs__/wonder-blocks-dropdown/combobox.argtypes.ts) for Storybook, could you also add the `AriaArgTypes`? 

It'll prevent from having all the aria attributes listed in the prop docs for the component! 

Image

Here's an example of how it's set in `option-item.argtypes.tsx`:

https://github.com/Khan/wonder-blocks/blob/3cc7eca9b5cd224e4b4a323c32c6b45ef28cfbc9/__docs__/wonder-blocks-dropdown/option-item.argtypes.tsx#L4

https://github.com/Khan/wonder-blocks/blob/3cc7eca9b5cd224e4b4a323c32c6b45ef28cfbc9/__docs__/wonder-blocks-dropdown/option-item.argtypes.tsx#L15

beaesguerra (approved):
Thanks for the fix @nedredmond ! The changes make sense to me, I'll cc: @jandrade too in case there was some context he had when working on this! 

I left suggestions for adding a unit test and updating Storybook argtypes, and a question to understand how you're using LabeledField!

packages/wonder-blocks-dropdown/src/components/combobox.tsx
nedredmond: Yeah, with this change, I wouldn't need those extra props. I saw that it wasn't wiring up automatically, which is why I tried to do it manually.

beaesguerra: Ah gotcha! Thanks for clarifying, glad you are not needing to wire things up explicitly 😄

nedredmond:
@jandrade, would you like to take a look or can this be landed?

marcysutton (approved):
This looks good to me! Juan is out until the 24th, so I'd say let's ship it.

PR #3177, opened 2026-08-12
Remove font.body.lineHeight.large token

jandrade:
## Summary

Removes the `font.body.lineHeight.large` token from `@khanacademy/wonder-blocks-tokens`.

This token was originally added to provide backwards compatibility with the default theme while building the ThunderBlocks theme. Now that typography uses the new `Heading` and `BodyText` components — which reference the correct tokens — this token is no longer needed. It has no functional usages anywhere in Wonder Blocks.

## Changes

- Removed `large: sizing.size_220` from `body.lineHeight` in `packages/wonder-blocks-tokens/src/theme/primitive/font.ts` (the unrelated `heading.lineHeight.large` is untouched).
- Regenerated the token-docs snapshot (`generate-token-docs.test.ts.snap`) to drop the now-removed row.
- Added a **major** changeset — removing a public token is a breaking change.

## ⚠️ Cross-repo coordination

This is a breaking change. Based on the Slack discussion:

- **`perseus`**: does not use this token — no action needed.
- **`webapp`/`frontend`**: uses it **once** (in `devadmin`). That usage must be updated to the closest match — `font.body.lineHeight.medium` — in a separate PR in that repo. The `frontend` fix should land **before or together with** this change so its build doesn't break when it picks up the new tokens version.

## Testing

- Tokens package tests pass.
- Full `typecheck` is clean (no consumer breakage inside Wonder Blocks).

## Context

Slack thread: https://khanacademy.slack.com/archives/C8Z9DSKC7/p1786550546633419?thread_ts=1786548443.300549&cid=C8Z9DSKC7

---
_Generated by [Claude Code](https://claude.ai/code/session_019c3ZnWfCeaJrqbNZx1LFjL)_

### DIFF
--- a/.changeset/delete-body-lineheight-large.md
+++ b/.changeset/delete-body-lineheight-large.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-tokens": major
+---
+
+Remove the `font.body.lineHeight.large` token. This token was originally added to provide backwards compatibility with the default theme while building the ThunderBlocks theme. Now that typography uses the new `Heading` and `BodyText` components (which reference the correct tokens), this token is no longer needed.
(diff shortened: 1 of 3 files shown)


somewhatabstract (approved):
LGTM

beaesguerra (approved):
Looks good to me!

PR #3179, opened 2026-08-18
[Modal] Update default animations in DrawerLauncher

marcysutton:
## Summary:

Updates `DrawerLauncher`'s default animation to match the design team's motion spec (WB-2191), and adds two customization points. Previously both phases shared one 400ms linear animation and the panel faded as it slid.

- **Enter:** 300ms, `cubic-bezier(0.05, 0.7, 0.1, 1)`
- **Exit:** 150ms, `cubic-bezier(0.3, 0, 0.8, 0.15)`
- Panel slides without fading, travelling its full width/height. The backdrop keeps its fade and shares the panel's durations.
- Closing unmounts the drawer after the exit duration rather than the enter duration.

New API:

- `easing={{enter, exit}}` — per-phase curve overrides, applied to the panel; the backdrop's fade stays linear.
- `styles.backdrop` — styles the backdrop overlay, applied last. Also the only way to reach its opacity animation.

Also fixes a pre-existing bug: `timingDuration` applied only to the panel, so an override left the panel and backdrop finishing at different times.

Issue: WB-2191

## Test plan:
- `pnpm jest`, `pnpm lint`, `pnpm typecheck` all pass
- Verify computed animation values in Storybook across all three alignments, including RTL sign flipping, and for default vs. overridden timing/easing

### DIFF
--- a/__docs__/wonder-blocks-modal/drawer-launcher.stories.tsx
+++ b/__docs__/wonder-blocks-modal/drawer-launcher.stories.tsx
@@ -55,7 +55,7 @@ export default {
 It can align a dialog on the \`inlineStart\` (left),  \`inlineEnd\` (right), or \`blockEnd\` (bottom).
 
 - Slide animations can be turned off with the \`animated\` prop.
-- Timing of animations can be fine-tuned with the \`timingDuration\` prop, used on enter and exit animations. It is also used to coordinate timing of focus management on open and close.
+- The drawer slides in over 300ms easing out, and leaves over 150ms easing in. Override the durations with \`timingDuration\` and the curves with \`easing\` — see the \`WithCustomAnimation\` example.
 
 **IMPORTANT**: This component should only be used with \`DrawerDialog\`. Using it with other
 dialog components may result in incorrect animations, positioning, and styling.
@@ -161,13 +161,11 @@ export const InlineEndAligned: StoryComponentType = {
 /**
  *
  * An blockEnd-aligned drawer. Uses the `alignment` prop to slide in from the
- * bottom in all writing modes, and a `timingDuration` of 400 milliseconds to
- * allow more time for animating-in vertically.
+ * bottom in all writing modes.
  */
 export const BlockEndAligned: StoryComponentType = {
     args: {
         alignment: "blockEnd",
-        timingDuration: 400,
     },
     render: (args) => (
         <DrawerLauncher modal={DefaultModal} alignment={args.alignment}>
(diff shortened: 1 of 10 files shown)


__docs__/wonder-blocks-modal/drawer-launcher.stories.tsx
```diff
@@ -199,6 +197,79 @@ export const WithNoAnimation: StoryComponentType = {
...
+ *
+ * Two props tune this independently:
+ *
+ * - `timingDuration` overrides the duration of both phases. Since the same value
+ *   coordinates focus management, a longer duration also delays the drawer's
+ *   initial focus and the return of focus to the trigger.
+ * - `easing` overrides the curves per phase, and applies to the panel only — the
+ *   backdrop's fade stays linear.
+ *
+ * This example exaggerates both: it slows the drawer down, overshoots past the
```
jonahgoldsaito, line 213: Does it actually overshoot the resting position on the way in? To me this reads like a *spring-like* behavior, which wasn't the intention, and I don't think it's actually doing.

packages/wonder-blocks-modal/src/util/drawer-animation.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/** Easing for the exit: eases in so the panel accelerates away. */
+export const DRAWER_EXIT_EASING = "cubic-bezier(0.3, 0, 0.8, 0.15)";
+
+/**
+ * How far the panel sits from its resting position at the far end of the slide,
+ * as a percentage of its own size, so it is fully clear of the edge it is docked
+ * to. The panel slides without fading, so this has to cover it completely.
+ */
+export const DRAWER_SLIDE_OFFSET = "100%";
```
jonahgoldsaito, line 27: 1. Makes so much sense that you're keeping this as a percentage of the width and not a fixed or calculated value. Much clearner for sure 👍 
2. This is too jarring though. The translation from the edge is meant to be partially suggested to reduce the intensity. Would love it if you could bump this down to `42%`!!! 🙏😃

jonahgoldsaito (dismissed):
In comments!

jonahgoldsaito:
Thank you for addressing those!!!!

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
```diff
@@ -81,19 +82,40 @@ type BaseProps = Readonly<{
...
     timingDuration?: number;
+    /**
+     * Optional per-phase easing overrides for the slide animation, e.g.
+     * `{enter: "ease-out", exit: "ease-in"}`. Each value is any CSS
+     * `animation-timing-function`, and either phase may be omitted to keep its
+     * default curve.
+     *
+     * Applies to the sliding panel only; the backdrop's fade stays linear.
+     */
+    easing?: DrawerEasing;
```
jandrade, line 118: **question:** What do you think about grouping these two props into a single prop object as both things are related?

packages/wonder-blocks-modal/src/components/drawer-dialog.tsx
```diff
@@ -174,68 +181,130 @@ const DrawerDialog = React.forwardRef(function DrawerDialog(
...
+
+    return {
+        animationName,
+        animationDuration: `${timingDuration ?? phaseDuration}ms`,
+        animationTimingFunction: phaseEasing,
+        // `backwards` holds the off-screen `from` before the entrance starts;
+        // `forwards` holds the off-screen `to` after the exit completes.
+        animationFillMode: isExiting ? "forwards" : "backwards",
+    };
+};
```
jandrade, line 291: **suggestion:** Is there a way to move this to called as inline styles? I'd like to avoid generating dynamic styles into Aphrodite objects to help with the css modules migration.

packages/wonder-blocks-modal/src/components/drawer-backdrop.tsx
```diff
@@ -216,20 +234,27 @@ const styles = StyleSheet.create({
...
+            // @ts-expect-error [FEI-5019] - `animationName` expects a string not an object
+            animationName: isExiting ? keyframes.fadeOut : keyframes.fadeIn,
+            animationDuration: `${
+                timingDuration ??
+                (isExiting ? DRAWER_EXIT_DURATION_MS : DRAWER_ENTER_DURATION_MS)
+            }ms`,
+            animationTimingFunction: "linear",
+            animationFillMode: "forwards",
+        },
+    });
```
jandrade: **suggestion:** I'd suggest moving this to static Aphrodite styles (like before) so the css modules conversion is simplified, then moving the dynamic parts as inline styles inside the component.

jandrade (commented):
Thanks for working on this! Left some initial comments to evaluate changing the API and simplifying the style generation.

packages/wonder-blocks-modal/src/components/drawer-launcher.tsx
marcysutton: Hmmm. It's an interesting thought, but `timingDuration` is already in the component API. This would cause a broader breaking change, and I'm not sure that it's worth it.

packages/wonder-blocks-modal/src/components/drawer-dialog.tsx
marcysutton: Yes! I had included context of the CSS Modules refactor into the implementation but it still ended up that way. I've taken another pass on it to try and avoid dynamic styles into Aphrodite objects. It's pretty complicated (it was hard for me to follow, tbh), so I included some inline comments about the types of objects and how they are used.

packages/wonder-blocks-modal/src/components/drawer-backdrop.tsx
marcysutton: Pushed an update! See the comment above for more info.

packages/wonder-blocks-modal/src/components/drawer-dialog.tsx
```diff
@@ -130,187 +136,221 @@ const DrawerDialog = React.forwardRef(function DrawerDialog(
...
-            },
-        },
-        dialog: {
-            // Override the minBlockSize and minInlineSize on View
-            // And allow BlockEnd content to provide its own height
-            minBlockSize: alignment === "blockEnd" ? "unset" : "100vh",
+// Positioning only — independent of animation state, so this is computed once
+// rather than per render. Aphrodite classes (StyleSheet.create), unlike the
+// animation styles above.
+const STATIC_STYLES = StyleSheet.create({
```
jandrade: **nit:** Would you mind renaming this to `styles`? that way is consistent with all the other style objects in the repo.

jandrade (approved):
Looks great! thanks for addressing my comments :) 🚀

packages/wonder-blocks-modal/src/components/drawer-dialog.tsx
marcysutton: I went with `componentStyles` since it always confused me to have a `styles` prop and a `styles` object.

PR #3180, opened 2026-08-21
Dropdown: Add label to internal listbox

marcysutton:
## Summary:

The listbox rendered by `SingleSelect` and `MultiSelect` lives in a portal, disconnected from its combobox opener, so it had no accessible name — screen reader users navigating into it got no context for the options. Fixes [CLASS-13348](https://khanacademy.atlassian.net/browse/CLASS-13348) (a11y audit finding).

The listbox is now labelled with the same name as the opener, verbatim. `getListboxLabelProps()` picks, in order:

1. `aria-labelledby` on the select → the listbox references the same element
2. the `<label>` associated with the opener (e.g. from `LabeledField`) → the listbox references that label's `id`
3. that label's text, when the label has no `id`
4. the select's `aria-label`

If the select has no label, neither does its listbox — no generic fallback text is invented, so there's nothing new to translate.

Issue: CLASS-13348

## Test plan:
- `pnpm jest`, `pnpm typecheck`, `pnpm lint` all pass
- Verify in Storybook: with `LabeledField`, both components render `<div role="listbox" aria-labelledby="…-labeled-field-label">` pointing at the visible label
- Review MultiSelect and SingleSelect Accessibility docs in Storybook

[CLASS-13348]: https://khanacademy.atlassian.net/browse/CLASS-13348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
+++ b/packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
@@ -14,6 +14,8 @@ import {
 } from "@testing-library/user-event";
 
 import {PropsFor} from "@khanacademy/wonder-blocks-core";
+import {LabeledField} from "@khanacademy/wonder-blocks-labeled-field";
+
 import OptionItem from "../option-item";
 import MultiSelect from "../multi-select";
 import {defaultLabels as builtinLabels} from "../../util/constants";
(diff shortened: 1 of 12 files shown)


packages/wonder-blocks-dropdown/src/components/__tests__/multi-select.test.tsx
```diff
@@ -2566,6 +2568,117 @@ describe("MultiSelect", () => {
...
+            );
+
+            // Act
+            await userEvent.click(await screen.findByRole("combobox"));
+
+            // Assert
+            expect(
+                await screen.findByRole("listbox", {hidden: true}),
+            ).toHaveAccessibleName("Students");
+        });
```
jandrade: **suggestion:** This test can probably be removed as the `id` doesn't do anything related to connecting the label with the control.

packages/wonder-blocks-dropdown/src/components/__tests__/single-select.test.tsx
```diff
@@ -1878,6 +1880,130 @@ describe("SingleSelect", () => {
...
+            // Act
+            await userEvent.click(await screen.findByRole("combobox"));
+
+            // Assert
+            expect(
+                await screen.findByRole("listbox", {hidden: true}),
+            ).toHaveAccessibleName("Students");
+        });
+
+        it("should label the listbox with the text of a label element without an id", async () => {
```
jandrade: same suggestion about dropping this test here.

packages/wonder-blocks-dropdown/src/util/helpers.ts
```diff
@@ -84,3 +84,48 @@ export function getSelectOpenerLabel(
     }
     return props.label;
 }
+
+/**
+ * Returns the aria attributes that label a listbox with the same name as its
+ * opener. The listbox is rendered in a portal, so it needs its own name to give
+ * the options context.
+ */
+export function getListboxLabelProps({
```
jandrade, line 93: **praise:** Neat helper! I wonder if this can be reused with other widgets (to consider as a future improvement).

jandrade (approved):
Looks great! thanks for fixing this! :accessibility: 🚀

PR #3182, opened 2026-08-25
⚠️ Add middleware/focus props and reference registration to Floating

jandrade:
⚠️ Add middleware/focus props and reference registration to `Floating`

Adds the props that consumers like `Popover` need from `Floating`, and changes
how `Floating` resolves its reference (anchor) element so that a trigger no
longer has to forward refs — and so that no wrapper element is added to the DOM.

New props: `returnFocus`, `closeOnFocusOut`, `onPlacementChange`,
`shiftPadding`, `rootBoundary`.

New export: `useFloatingReference` (and the `FloatingReferenceContext` it
reads). A trigger that can't receive a ref (a plain function component) can
attach the returned ref callback to the element it renders, and that element
becomes the reference element. Triggers that *can* receive a ref (host elements,
`forwardRef` and class components) keep getting the reference ref injected
directly, so existing usage is unchanged.

Each `Floating` provides its reference setter to **its own trigger only** —
never to the floating content — so multiple simultaneously open (or nested)
floating elements resolve independent reference elements.

Also removes the floating element's `max-inline-size` (previously `472px`,
carried over from `Tooltip`). That cap forced every `Floating` consumer into a
tooltip-sized bubble; the floating element now sizes to its content, and
consumers that need a width cap set one on their own content (or via the
`styles.floating` prop). `Popover` is unaffected — `PopoverContentCore` already
caps its content at 288px.

**Impact:** `@khanacademy/wonder-blocks-floating` is not yet released (`0.0.1`)
and currently has no production consumers, so this is additive for external
users. Within this stack, `WB-2119.2` (the `Popover` refactor) depends on
`useFloatingReference` and should be rebased on this once it lands.

Issue: WB-2119

## Test plan:

Automated: `pnpm typecheck`, `pnpm lint`, `pnpm jest packages/wonder-blocks-floating`
and `packages/wonder-blocks-popover` all pass, plus Storybook story tests
(with a11y) for every `Floating` story and 8 `Popover` stories.

Manual verification (positioning and focus behavior can't be fully asserted in
jsdom):

1. Run `pnpm start` and open **Packages / Floating → Custom Trigger Component**
   (http://localhost:6061/?path=/story/packages-floating--custom-trigger-component).
   - Click the trigger. The floating element should appear anchored to the
     custom component (which uses `useFloatingReference`, not a forwarded ref).
   - Confirm the trigger has **no wrapper element** around it in the DOM
     inspector.
   - Press `Esc` and click outside — both should dismiss it, confirming the
     props `Floating` injects still reach the trigger.
   - Confirm focus returns to the trigger after closing.
2. Open **Packages / Floating → Placements** and confirm all 12 placements still
   anchor correctly against `IconButton` triggers, in both LTR and RTL (toggle
   the direction global).
3. Open **Packages / Floating → Middlewares** and scroll the container; confirm
   `shift`/`flip`/`hide` still behave as before with a `View` trigger.
4. Confirm the floating element sizes to its content now that the `472px` cap is
   gone: check a long-text story (and a `Popover` story, which should be
   unchanged at 288px) in both LTR and RTL.

## Review plan:

1. ⚠️ [`floating.tsx`](https://github.com/Khan/wonder-blocks/pull/3182#discussion_r3854760344)
2. ⚠️ [`floating-reference-context.ts`](https://github.com/Khan/wonder-blocks/pull/3182#discussion_r3854760544)
3. ⚠️ [`accepts-ref.ts`](https://github.com/Khan/wonder-blocks/pull/3182#discussion_r3854760708)
4. ⚠️ [`floating.tsx` (styles)](https://github.com/Khan/wonder-blocks/pull/3182#discussion_r3863055138) — removes the `472px` `max-inline-size`.

### Common patterns:

**1 File:** New optional props are declared with JSDoc (so they land in the
Storybook props table), defaulted in the destructured signature, and threaded
into the floating-ui config. Five props follow this shape.

```diff
+    /**
+     * The boundary that the floating element should be kept within by the
+     * `flip` and `shift` middleware.
+     * @default "viewport"
+     */
+    rootBoundary?: "viewport" | "document";

-    shiftProp ? shift({padding: SHIFT_PADDING, crossAxis: true}) : undefined,
+    shiftProp
+        ? shift({padding: shiftPadding, crossAxis: true, rootBoundary})
+        : undefined,
```

**3 Files:** Logic that doesn't belong in the component was extracted into
`src/util/` with a colocated test file, then imported by `floating.tsx`.

```diff
-    const trigger = React.useMemo(() => {
-        return React.cloneElement(children, {
-            ref: refs.setReference,
-            ...getReferenceProps(),
-        });
-    }, [children, refs.setReference, getReferenceProps]);
+    const trigger = React.useMemo(() => {
+        return React.cloneElement(children, {
+            ...(acceptsRef(children) ? {ref: setReference} : undefined),
+            ...getReferenceProps(),
+        });
+    }, [children, setReference, getReferenceProps]);
```


packages/wonder-blocks-floating/src/components/floating.tsx
```diff
@@ -23,13 +23,25 @@ import {ARROW_SIZE_INLINE} from "../util/constants";
 import {Arrow, type ArrowStyles} from "./floating-arrow";
 import {Portal} from "./floating-portal";
 import {rtlMirror} from "../util/rtl-mirror-middleware";
+import {acceptsRef} from "../util/accepts-ref";
```
jandrade: ⚠️ **Medium risk** — this changes how the reference element is resolved for every consumer of `Floating`.

The reference ref is now injected *conditionally* (only when the trigger can receive one), and five new props are threaded into the floating-ui middleware and `FloatingFocusManager` config.

Worth scrutinising: a trigger that neither accepts a ref nor calls `useFloatingReference` now fails **silently** — no reference element resolves, so `open && elements.reference` stays false and the floating element simply never renders. Previously React logged a "Function components cannot be given refs" warning in that situation. I left out a dev-mode warning because the check is racy on the first commit (the ref callback and the context registration can both land after the first render); happy to add a deferred check in an effect if you'd prefer the louder failure.

packages/wonder-blocks-floating/src/util/floating-reference-context.ts
```diff
@@ -0,0 +1,48 @@
+import * as React from "react";
```
jandrade: ⚠️ **Medium risk** — new public API (`useFloatingReference` / `FloatingReferenceContext`).

The key line to check is where the provider is placed in `floating.tsx`: it wraps **only the trigger**, never `content`. That's what keeps multiple simultaneously open — and nested — floating elements from claiming each other's reference element, since a trigger rendered inside an outer floating element resolves to its own nearest provider.

There are no ids and no document-wide queries involved, so there is nothing to collide across instances or across React roots. Covered by tests for the multi-instance, nested, and "content can't see the setter" cases in `floating.test.tsx`.

packages/wonder-blocks-floating/src/util/accepts-ref.ts
```diff
@@ -0,0 +1,31 @@
+import * as React from "react";
```
jandrade: ⚠️ **Medium risk** — new functionality that inspects React internals.

`acceptsRef` decides whether a ref can be injected by checking `$$typeof` against `Symbol.for("react.forward_ref")` / `Symbol.for("react.memo")` and `prototype.isReactComponent`. These symbols are stable across React versions, but it is internals-adjacent, so worth confirming the covered cases are the ones we care about.

Unit tests cover host elements, `forwardRef`, `memo(forwardRef(...))`, class components, plain function components, and `memo(function)`.

packages/wonder-blocks-floating/src/components/floating.tsx
```diff
@@ -331,8 +422,6 @@ const styles = StyleSheet.create({
         background: semanticColor.core.background.base.default,
         border: `solid ${border.width.thin} ${semanticColor.core.border.neutral.subtle}`,
         borderRadius: border.radius.radius_040,
```
jandrade, line 452: ⚠️ Drops the `max-inline-size: 472px` cap (carried over from `Tooltip`). The floating element now sizes to its content, so consumers that want a width limit set one on their own content or via `styles.floating`. `Popover` is unaffected — `PopoverContentCore` already caps at 288px — but any future consumer relying on the implicit tooltip width will now render wider.

packages/wonder-blocks-floating/src/components/__tests__/floating.test.tsx
```diff
@@ -50,6 +80,213 @@ describe("Floating", () => {
...
+            // reference element and both are rendered.
+            const firstContent = screen.getByText("First content");
+            const secondContent = screen.getByText("Second content");
+
+            // Assert
+            expect(firstContent).toBeInTheDocument();
+            expect(secondContent).toBeInTheDocument();
+        });
+
+        it("should warn and not render the floating content when the trigger doesn't spread the props it is given", () => {
```
somewhatabstract: **suggestion:** The "and" in this test case, plus the two assertions, shows that this is pulling double duty asserting two different requirements; this should be two separate test cases.

packages/wonder-blocks-floating/src/components/floating.tsx
```diff
@@ -133,6 +167,36 @@ type FloatingProps = {
...
+     */
+    returnFocus?: boolean | React.MutableRefObject<HTMLElement | null>;
+
+    /**
+     * Whether the floating element should close when focus moves outside of it
+     * (e.g. the user tabs past the last focusable element). Only relevant for
+     * non-modal focus management (i.e. when `focusManagerEnabled` is `true`).
+     * @default false
+     */
+    closeOnFocusOut?: boolean;
```
somewhatabstract, line 199: **question:** `focusOut` is when a component that contained focus (like a dialog that has a focused textbox) no longer contains the focus, right? As opposed to `blur` which is the directly focused component losing focus.

somewhatabstract (approved):
LGTM

packages/wonder-blocks-floating/src/components/floating.tsx
jandrade: that's correct!

packages/wonder-blocks-floating/src/components/__tests__/floating.test.tsx
jandrade: fixed!

PR #3183, opened 2026-08-25
Remove node-fetch dependency

somewhatabstract:
## Summary

Node 24's built-in Fetch API makes `node-fetch` unnecessary, and its peer dependency on the testing packages forces every consumer (~90 packages in Khan/frontend) to carry `node-fetch` in their devDependencies just to satisfy peer resolution.

- Use the global `Response` directly in `RespondWith`, dropping the `ResponseImpl` indirection and its `node-fetch` fallback (dead code on Node 18+).
- Drop the body from mocked responses with null-body status codes (204/205/304) — the spec-compliant `Response` constructor throws on those where node-fetch's did not.
- Shim jsdom's missing fetch globals in the jest setup from `undici` (the implementation Node's own fetch comes from) instead of node-fetch.
- Remove `node-fetch`/`@types/node-fetch` from the testing packages' peer dependencies, the root manifest, the pnpm catalog, and the shared tsconfig `types` list.

Consumers' jsdom test environments must provide a global `Response` (Khan/frontend already polyfills one via `whatwg-fetch`).

Issue: FEI-8179

## Test plan:

- `pnpm jest` — full suite passes; testing-core snapshots updated for native fetch error/serialization text.
- `pnpm typecheck` and lint on changed files pass.
- Verify a consumer's tests (e.g. Khan/frontend) still pass against a snapshot build of `wonder-blocks-testing-core`.

---
_Generated by [Claude Code](https://claude.ai/code/session_013BRecrQQFefX75XUnn9XjS)_


jandrade (approved):
This is awesome! thanks for the cleanup :shipit:

PR #3186, opened 2026-08-26
WB-2185 Date Picker Contrast/Weight Styles

marcysutton:
## Summary:
This PR updates the styles for the Date Picker calendar to use color and font weight more intentionally.

- The current day is bolded and blue, no longer relying on color alone
- The selected day is no longer bolded, since it has a circle outline

Issue: WB-2185

## Test plan:
1. Open the date picker stories and measure contrast: `/?path=/story/packages-datepicker-datepicker--open-calendar-overlay&globals=theme:thunderblocks`
2. Note the days fall under text contrast (1.4.3) and should meet 4.5:1, while the previous/next arrow buttons fall under non-text-contrast (1.4.11), only requiring 3:1 ratio

## Screenshots

Before:
Date picker overlay with bolded selected day, and blue regular weight current day

After:
Date picker overlay with regular weight selected day, and blue bolded current day

### DIFF
--- a/.changeset/wise-walls-perform.md
+++ b/.changeset/wise-walls-perform.md
@@ -1,0 +1,5 @@
+---
+"@khanacademy/wonder-blocks-date-picker": patch
+---
+
+Update date picker styling for accessibility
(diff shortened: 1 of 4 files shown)


jandrade (approved):
LGTM! Thanks for the fixes :accessibility:

PR #3188, opened 2026-08-28
DatePicker: Change interaction pattern to manual activation

marcysutton:
## Summary:
When a `DatePicker` sits side-by-side with a native time input, the activation patterns aren't consistent with one another. This PR changes the WB `DatePicker` to use manual activation on a calendar icon button overlaying the `DatePickerInput`. It also includes a change to `IconButton` to add an `onKeyDown` prop, which was unsupported before.

Expected interaction flow:
- Tab onto DatePicker input, and tab again to reach icon button
- Enter/space toggles the date picker overlay
- When date picker overlay is open, arrow down to move focus into the date picker grid (note: except in NVDA or JAWS, where Browse Mode is the default and the virtual cursor will move to the next element outside the date picker)
- Escape or shift tab will return to the icon button
- Tab will continue past the icon button when the overlay is closed

Issue: WB-2425

## Test plan:
1. Review new comparison story: `/?path=/story/packages-datepicker-datepicker--date-and-time-comparison&globals=theme:thunderblocks`
2. Ensure date validation and locales still work (@beaesguerra's [testing doc](https://khanacademy.atlassian.net/wiki/spaces/~712020fa1796fb383b47c5a0cb43a4c0a5ca5f/pages/4571201829/Test+cases+for+date-picker+integration) may help!)
3. Ensure tests pass

### DIFF
--- a/packages/wonder-blocks-date-picker/src/components/__tests__/date-picker.test.tsx
+++ b/packages/wonder-blocks-date-picker/src/components/__tests__/date-picker.test.tsx
@@ -32,7 +32,9 @@ function expectOverlayHidden() {
 // focus/click on the input no longer opens the overlay.
 async function openOverlay() {
     await userEvent.tab();
-    await userEvent.click(screen.getByRole("button", {name: "Show calendar"}));
+    await userEvent.click(
+        screen.getByRole("button", {name: "Toggle calendar"}),
+    );
 }
 
 describe("DatePicker", () => {
@@ -94,14 +96,14 @@ describe("DatePicker", () => {
         await openOverlay();
         expectOverlayVisible();
         await userEvent.click(
-            screen.getByRole("button", {name: "Show calendar"}),
+            screen.getByRole("button", {name: "Toggle calendar"}),
         );
         expectOverlayHidden();
     });
 
     it("reflects aria-expanded on the calendar button as the overlay opens and closes", async () => {
         render(<DatePicker updateDate={() => {}} />);
-        const button = screen.getByRole("button", {name: "Show calendar"});
+        const button = screen.getByRole("button", {name: "Toggle calendar"});
         expect(button).toHaveAttribute("aria-expanded", "false");
         await openOverlay();
         expect(button).toHaveAttribute("aria-expanded", "true");
@@ -113,7 +115,7 @@ describe("DatePicker", () => {
         render(<DatePicker disabled updateDate={() => {}} />);
         await userEvent.tab();
         await userEvent.click(
-            screen.getByRole("button", {name: "Show calendar"}),
+            screen.getByRole("button", {name: "Toggle calendar"}),
         );
         expectOverlayHidden();
     });
@@ -161,7 +163,7 @@ describe("DatePicker", () => {
     it("keeps focus on the calendar button when it's clicked to open the overlay", async () => {
         render(<DatePicker updateDate={() => {}} />);
         const calendarButton = screen.getByRole("button", {
-            name: "Show calendar",
+            name: "Toggle calendar",
         });
         await userEvent.click(calendarButton);
         expect(calendarButton).toHaveFocus();
@@ -170,7 +172,7 @@ describe("DatePicker", () => {
     it("closes the overlay and returns focus to the calendar button when Escape is pressed", async () => {
         render(<DatePicker updateDate={() => {}} />);
         const calendarButton = screen.getByRole("button", {
-            name: "Show calendar",
+            name: "Toggle calendar",
         });
         await openOverlay();
         expect(calendarButton).toHaveFocus();
@@ -194,18 +196,17 @@ describe("DatePicker", () => {
         await userEvent.keyboard("{Escape}");
         await waitFor(() => expectOverlayHidden());
         expect(
-            screen.getByRole("button", {name: "Show calendar"}),
+            screen.getByRole("button", {name: "Toggle calendar"}),
         ).toHaveFocus();
         await userEvent.keyboard("{ArrowDown}");
         expectOverlayHidden();
     });
 
-    it("opens the date picker if Enter is pressed when closed", async () => {
+    it("does not open the overlay when Enter is pressed on the input while closed", async () => {
         render(<DatePicker updateDate={() => {}} />);
         await userEvent.tab();
-        await userEvent.keyboard("{Escape}");
         await userEvent.keyboard("{Enter}");
-        expectOverlayVisible();
+        expectOverlayHidden();
     });
 
     it("closes the date picker if Enter is pressed when open", async () => {
@@ -215,19 +216,21 @@ describe("DatePicker", () => {
         expectOverlayHidden();
     });
 
-    it("opens the date picker if down arrow is pressed when closed", async () => {
+    it("does not open the overlay when ArrowDown is pressed on the input while closed", async () => {
+        // The calendar button is the only way to open the overlay --
+        // ArrowDown on the input matches native date/time inputs, where
+        // arrow keys adjust the focused value segment instead.
         render(<DatePicker updateDate={() => {}} />);
         await userEvent.tab();
-        await userEvent.keyboard("{Escape}");
         await userEvent.keyboard("{ArrowDown}");
-        expectOverlayVisible();
+        expectOverlayHidden();
     });
 
     it("moves focus into the calendar grid when ArrowDown is pressed on the calendar button while the overlay is already open", async () => {
         render(<DatePicker updateDate={() => {}} />);
         await openOverlay();
         expect(
-            screen.getByRole("button", {name: "Show calendar"}),
+            screen.getByRole("button", {name: "Toggle calendar"}),
         ).toHaveFocus();
         await userEvent.keyboard("{ArrowDown}");
         await waitFor(() => {
@@ -242,8 +245,8 @@ describe("DatePicker", () => {
     it("moves focus into the calendar grid when ArrowDown is pressed on the input while the overlay is already open", async () => {
         render(<DatePicker updateDate={() => {}} />);
         const input = screen.getByRole("textbox");
-        await userEvent.tab();
-        await userEvent.keyboard("{ArrowDown}"); // opens; focus stays on input
+        await openOverlay(); // opens via the calendar button
+        await userEvent.click(input);
         expect(input).toHaveFocus();
         await userEvent.keyboard("{ArrowDown}"); // overlay already open
         await waitFor(() => {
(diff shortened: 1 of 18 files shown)


packages/wonder-blocks-date-picker/src/components/date-picker.tsx
```diff
@@ -87,6 +92,15 @@ interface Props {
      * is no visible label associated with the date picker, such as with LabeledField.
      */
     inputAriaLabel?: string;
+    /**
+     * The aria-label for the calendar toggle button that opens/closes the
+     * calendar overlay. Defaults to "Toggle calendar".
+     */
+    calendarButtonAriaLabel?: string;
```
marcysutton, line 99: I weighed whether to move these to a `labels` prop, but it would have been a breaking change with `inputAriaLabel`. Let me know if that's preferred (along with splitting this change off to a separate PR).

jandrade: I think it is fine to keep it this way. We can iterate later if we need to.

packages/wonder-blocks-date-picker/src/components/__tests__/date-picker.test.tsx
```diff
@@ -27,6 +27,16 @@ function expectOverlayHidden() {
...
+// Opens the calendar overlay via the explicit toggle mechanism: focus the
+// input (matching real usage), then click the calendar toggle button. Plain
+// focus/click on the input no longer opens the overlay.
+async function openOverlay() {
+    await userEvent.tab();
+    await userEvent.click(
+        screen.getByRole("button", {name: "Toggle calendar"}),
+    );
+}
+
```
jandrade: **suggestion:** I'd recommend to avoid creating helpers in unit tests to make these easier to understand: https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/98402353/Unit+Testing+Best+Practices#Repeat-yourself

packages/wonder-blocks-date-picker/src/components/__tests__/date-picker.test.tsx
```diff
@@ -51,10 +61,10 @@ describe("DatePicker", () => {
         expectOverlayHidden();
     });
 
-    it("shows the date picker overlay if the input is clicked", async () => {
+    it("does not show the date picker overlay if the input is clicked", async () => {
         render(<DatePicker updateDate={() => {}} />);
         await userEvent.click(screen.getByRole("textbox"));
-        expectOverlayVisible();
+        expectOverlayHidden();
```
jandrade: **suggestion:** Same thing here about adding unnecessary abstraction in unit tests: https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/98402353/Unit+Testing+Best+Practices#Repeat-yourself

packages/wonder-blocks-date-picker/src/components/__tests__/date-picker.test.tsx
```diff
@@ -63,18 +73,53 @@ describe("DatePicker", () => {
...
+
+    it("reflects aria-expanded on the calendar button as the overlay opens and closes", async () => {
+        render(<DatePicker updateDate={() => {}} />);
+        const button = screen.getByRole("button", {name: "Toggle calendar"});
+        expect(button).toHaveAttribute("aria-expanded", "false");
+        await openOverlay();
+        expect(button).toHaveAttribute("aria-expanded", "true");
+        await userEvent.click(button);
+        expect(button).toHaveAttribute("aria-expanded", "false");
+    });
```
jandrade: **suggestion:** This test makes sense, but it is definitively better if each assertion is split into its own test.

https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/98402353/Unit+Testing+Best+Practices#One-Expectation-Per-Test-Case

packages/wonder-blocks-date-picker/src/components/__tests__/date-picker.test.tsx
```diff
@@ -63,18 +73,53 @@ describe("DatePicker", () => {
...
+
+    it("does not open the overlay when the calendar button is clicked if disabled", async () => {
+        render(<DatePicker disabled updateDate={() => {}} />);
+        await userEvent.tab();
+        await userEvent.click(
+            screen.getByRole("button", {name: "Toggle calendar"}),
+        );
+        expectOverlayHidden();
```
jandrade: **suggestion:** I'd suggest following the AAA pattern here (and in all the tests in this file).

https://khanacademy.atlassian.net/wiki/spaces/ENG/pages/98402353/Unit+Testing+Best+Practices#Arrange-Act-Assert

packages/wonder-blocks-date-picker/src/components/__tests__/date-picker.test.tsx
```diff
@@ -110,63 +155,112 @@ describe("DatePicker", () => {
...
+        expect(calendarButton).toHaveFocus();
+    });
+
+    it("closes the overlay and returns focus to the calendar button when Escape is pressed", async () => {
+        render(<DatePicker updateDate={() => {}} />);
+        const calendarButton = screen.getByRole("button", {
+            name: "Toggle calendar",
+        });
+        await openOverlay();
+        expect(calendarButton).toHaveFocus();
```
jandrade: **suggestion:** This expectation is redundant as it is tested in the previous test case.
```suggestion

```

packages/wonder-blocks-date-picker/src/components/__tests__/date-picker.test.tsx
jandrade, line 1: The test cases are a great addition!

Left some suggestions around improvements on the tests cases overall to apply our testing best practices.

packages/wonder-blocks-date-picker/src/components/date-picker-input.tsx
```diff
@@ -426,18 +448,22 @@ const DatePickerInput = React.forwardRef<HTMLInputElement, Props>(
...
+            (inputElement as any).focusCalendarButton = () => {
+                buttonRef.current?.focus();
+            };
+            // Lets the parent (DatePicker) use the calendar toggle button --
+            // not the input -- as the overlay's focus-trap anchor, so
+            // Shift+Tab out of the overlay's first element (a nav button)
+            // returns focus to the button that opened it.
+            (inputElement as any).getCalendarButtonElement = () =>
+                buttonRef.current;
             return inputElement;
```
jandrade: **question:** Is there a different approach for this? this feels really hacky and outside of the React model.

Maybe this logic should be put in `date-picker.tsx` directly? wdyt?

packages/wonder-blocks-date-picker/src/components/date-picker.tsx
```diff
@@ -136,6 +150,18 @@ const customRootStyle = {
...
+// style, so we add it as an extra class alongside the default one (classNames
+// entries replace, rather than merge with, the default per UI element).
+const dayButtonFocusStyles = StyleSheet.create({
+    focus: focusStyles.focus,
+});
+const dayPickerClassNames = {
+    [UI.DayButton]: `${getDefaultClassNames()[UI.DayButton]} ${css(
+        dayButtonFocusStyles.focus,
+    )}`,
+};
```
jandrade, line 164: **suggestion:** If possible, I'd suggest passing this as an inline style object instead of using the `StyleSheet.create/css` pair from Aphrodite.

packages/wonder-blocks-date-picker/src/components/date-picker.tsx
```diff
@@ -239,6 +273,69 @@ const DatePicker = (props: Props) => {
...
+    const getCalendarButtonElement = React.useCallback(():
+        | HTMLButtonElement
+        | null
+        | undefined => {
+        return (
+            datePickerInputRef.current as HTMLInputElement & {
+                getCalendarButtonElement?: () => HTMLButtonElement | null;
+            }
+        )?.getCalendarButtonElement?.();
+    }, [datePickerInputRef]);
```
jandrade: This is the part that is connected to what I mentioned about the custom implementation outside of React. Is there a way to make this work directly here?

packages/wonder-blocks-date-picker/src/components/date-picker.tsx
```diff
@@ -284,54 +381,84 @@ const DatePicker = (props: Props) => {
...
         }
-        if (e.relatedTarget instanceof Node) {
-            return !dayPickerCalendar.contains(e.relatedTarget);
+        // Not leaving if focus moved into the calendar, or to the toggle
+        // button -- both are still part of this widget while the overlay
+        // is open (e.g. Tab from the input to the button).
+        if (dayPickerCalendar?.contains(e.relatedTarget)) {
+            return false;
+        }
+        const calendarButton = getCalendarButtonElement();
```
jandrade: Related: This should probably point to a ref instead of having to create a custom wrapper to get that.

packages/wonder-blocks-icon-button/src/components/icon-button-unstyled.tsx
```diff
@@ -74,8 +66,9 @@ export const IconButtonUnstyled: React.ForwardRefExoticComponent<
                     onPress?.(true);
                 }
             }
+            onKeyDownProp?.(e);
```
jandrade: **suggestion/thought:** I'm not fully sure if order matters here, but maybe this could be moved before the condition with the `onPress` call?

jandrade (changes requested):
Thanks for working on improving this component. I'm requesting changes to the follow ing parts:

- Custom logic to handle focus / element references that create custom functions.
- Modifying unit tests to follow our best practices.

Let me know if you have any questions.

packages/wonder-blocks-date-picker/src/components/__tests__/date-picker.test.tsx
marcysutton: agh! It's annoying how many times I have to remind Claude about that advice too. Thanks for catching it!

packages/wonder-blocks-date-picker/src/components/date-picker.tsx
marcysutton: Tbh I'm not sure of the best workaround for this without an actual stylesheet, if we can't use Aphrodite functions... pinged you in Slack!

packages/wonder-blocks-date-picker/src/components/date-picker-input.tsx
marcysutton: Yes! Cleaned it up with a ref.

packages/wonder-blocks-date-picker/src/components/date-picker.tsx
jandrade: After chatting about this, it makes sense to follow the current approach, then we can refactor it accordingly once we migrate to CSS modules

marcysutton: As discussed, there's no easy way to deal with this until we move to CSS Modules given we're targeting a `:focus-visible` pseudo-class. Added a ticket to follow up: https://khanacademy.atlassian.net/browse/WB-2432

jandrade (approved):
Looks great! thanks for going over the feedback, really appreciate it 👏 🚢

PR #3189, opened 2026-08-28
🔷 Anchor the popover without injecting a ref into the trigger

jandrade:
## Summary:

Now that `Floating` resolves the reference (anchor) element from the DOM, the
popover's anchor no longer needs any ref plumbing. This PR removes it.

This is the follow-up to #3163, which did the `Popover` → `Floating` migration.
That PR made the one change to `PopoverAnchor` needed for the popover to
position at all — spreading the `FloatingReferenceAttributeName` attribute
through to the trigger — and left the rest alone, so this PR is a small, focused
cleanup on top of it.

- `PopoverAnchor` becomes a plain function component. It was a
  `React.forwardRef` component purely to receive the anchor ref, and after
  #3163 nothing passes it a ref anymore.
- The hand-rolled `mergeRefs` helper is deleted. It existed to merge the anchor
  ref with the trigger's own ref; there is no anchor ref to merge, so a
  trigger's `ref` is now left untouched.
- The function-as-children and element-children branches are unified, since ref
  handling was the only thing that differed between them.

The user-visible effect is the reason this is a separate PR: a `Popover` trigger
can now be of **any** component type — host element, `React.forwardRef`
component, class component, or plain function component — and never has to
accept or forward a ref. It only has to spread the props it is given onto its
own element (which it needs to do anyway for the interaction and ARIA props).

This PR carries its own changeset (`popover-anchor-no-ref.md`, major) for the
trigger-contract change, separate from #3163's `popover-floating-ui.md`.

Issue: WB-2119

## Test plan:

Automated: `pnpm typecheck`, `pnpm lint`, and `pnpm build` all pass. The Popover
jest suite passes, with new coverage for a plain function component as the
trigger and for a trigger's own ref being left alone.

Manual (Storybook — Popover stories):

1. Verify the popover still anchors and positions correctly for the
   element-children trigger pattern.
2. Verify the same for the function-as-children trigger pattern.
3. Verify the Tabs `TabLabelRenderFunction` story (popover anchored to a tab)
   still positions and behaves correctly.

## Review plan:

1. 🔷 [`popover-anchor.tsx`](https://github.com/Khan/wonder-blocks/pull/3189#discussion_r3882173776)

### Common patterns:

**1 file:** Drop `forwardRef` + `mergeRefs`; the trigger is identified by the
attribute `Floating` injects rather than by a ref.

```tsx
// Before
const PopoverAnchor = React.forwardRef<HTMLElement, Props>(
    function PopoverAnchor(props, ref) {
        // ...
        return React.cloneElement(children, {
            ...sharedProps,
            ref: childrenRef ? mergeRefs(ref, childrenRef) : ref,
        });
    },
);

// After
export default function PopoverAnchor(props: Props) {
    // ...
    return React.cloneElement(renderedChildren, sharedProps);
}
```


packages/wonder-blocks-popover/src/components/popover-anchor.tsx
```diff
@@ -60,64 +35,54 @@ type Props = AriaProps & {
...
-        const sharedProps = {
-            id: id,
-            "aria-controls": ariaControls,
-            "aria-expanded": ariaExpanded,
-            // Passed along so that the `Floating` component this anchor is
-            // rendered in can find the trigger's element in the DOM and position
-            // the popover against it.
-            [FloatingReferenceAttributeName]: floatingReferenceId,
-        } as const;
+export default function PopoverAnchor(props: Props) {
```
jandrade, line 38: 🔷 This is the file to review in this PR — it changes the contract between `Popover` and the trigger element it wraps.

After #3163, `Floating` resolves the reference (anchor) element from the DOM via the attribute it injects, and nothing passes a ref into `PopoverAnchor` any more. So the ref plumbing that was left behind is removed here:

- `React.forwardRef` → a plain function component. It was a `forwardRef` component purely to receive the anchor ref.
- The hand-rolled `mergeRefs` helper is deleted. It existed to merge the anchor ref with the trigger's own ref; there is no anchor ref to merge, so a trigger's `ref` is now left untouched rather than being wrapped.
- The function-as-children and element-children branches are unified, since ref handling was the only thing that differed between them.

The user-visible consequence, and the reason this is its own PR + changeset: a `Popover` trigger can now be **any** component type — host element, `forwardRef` component, class component, or plain function component — and never has to accept or forward a ref. It only has to spread the props it is given onto its own element, which it needs to do anyway for the interaction and ARIA props.

packages/wonder-blocks-popover/src/components/__tests__/popover-anchor.test.tsx
```diff
@@ -59,6 +66,39 @@ describe("PopoverAnchor", () => {
...
+
+        // Assert
+        const triggerElement = await screen.findByRole("button");
+        expect(triggerElement).toHaveAttribute(
+            FloatingReferenceAttributeName,
+            "reference-id",
+        );
+        // React would warn about `Function components cannot be given refs` if
+        // the anchor injected a ref into the trigger.
+        expect(consoleErrorSpy).not.toHaveBeenCalled();
```
somewhatabstract, line 117: **suggestion:** This feels like it's asserting a stealth requirement that the test case never mentions. I think this last assert should be a separate test case.

somewhatabstract (approved):
LGTM

Small suggestion in the tests as I think they need a little tweak before landing.

PR #3190, opened 2026-08-28
Accordion: animate the first time a section is opened

jonahgoldsaito:
https://github.com/user-attachments/assets/1f81a4c6-f76c-49c8-8775-27596122d7d9

## Summary

The `AccordionSection` component doesn't animate the **first** time it's opened was opened — it snapped open . Every toggle after that animates smoothly as intended

## Root cause

Aphrodite merges each style list into a single generated class and injects that class's rule lazily. Swapping between `wrapperCollapsed` and `wrapperExpanded` pointed the first expand at a rule that hadn't been injected yet, so `grid-template-rows` computed to `none` — which **cannot interpolate** — and the section snapped open.

The caret icons rotation works because but `none → rotate(180deg)` *is* interpolable.

## Fix

Now there's one stable class on the wrapper. Collapsing keys off the `data-expanded` attribute. 

Both rules are injected together on first render, so no toggle ever meets a missing rule. 

Precedent: attribute selectors in `StyleSheet.create` are established WB practice (`action-styles.ts`, `cell-core.tsx`, `tab.tsx`).

## Verification

Claude took screengrabs and measured `grid-template-rows` frame by frame in the browser across the first open:

| | before | after |
|---|---|---|
| first open | `0 → 52px` in one frame | `0 → 0.65 → 3.9 → 11.3 → … → 52px` over ~300ms |

Confirmed the diagnosis before fixing: in a `rounded-per-section` accordion all sections share one merged class, so opening section 1 (snap) injected the rule that let section 2's *first* open animate. Same component, same props — which ruled out a layout-measurement cause.

Checked for all three `cornerKind`s, the collapse direction, `animated={false}` (still instant), mount-expanded (no transition on mount), and `collapsible={false}`. Collapsed content stays `visibility: hidden`; zero frames where the panel's `overflow` left `hidden`.

Jest (4/4 accordion suites, existing transition tests unchanged), typecheck, and lint all pass.

## Reviewer notes

- **The `frontend` blast radius is zero**. There are 40 render sites, but 0 snapshots render an accordion, 0 style-based assertions, 0 `data-expanded` collisions, 0 Cypress selectors. Every accordion test there asserts via `aria-expanded`. Yay!
- The collapsed value now sits behind `:not([data-expanded="true"])`, which is more specific than the base class. A consumer overriding `gridTemplateRows` via `style` would still win when expanded but be ignored when collapsed.

### DIFF
--- a/.changeset/accordion-first-open-animation.md
+++ b/.changeset/accordion-first-open-animation.md
@@ -4,4 +4,6 @@
 
 Fix `AccordionSection` not animating the first time a section is opened when `animated` is true. The expand/collapse row sizing now lives on a single stable Aphrodite class, selected off a `data-expanded` attribute, instead of swapping between two classes. Aphrodite injects a merged class's rule lazily, so the old approach pointed the first expand at a rule that did not exist yet — `grid-template-rows` computed to `none`, which cannot interpolate, and the section snapped open. Every subsequent toggle animated correctly, which is why this only ever showed up on first open.
 
-`AccordionSection`'s wrapper element now renders a `data-expanded="true" | "false"` attribute, and its generated class name changed. Snapshot tests that render an `AccordionSection` will need regenerating. Prefer asserting expanded state via `aria-expanded` on the header rather than via the wrapper's styles.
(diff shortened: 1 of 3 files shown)


jandrade (approved):
Looks great! and it is in line with the ongoing CSS modules migration. Thanks for your contribution! hoping to get more of this 👏 🚢

PR #3191, opened 2026-08-28
DatePicker: Allow date segments to be incremented with the arrow keys in DatePickerInput

marcysutton:
## Summary:
To make the DatePicker work more like the native time input, we allow the arrow keys to adjust numeric date segments in the date picker input.

Issue: WB-2431

## Test plan:
1. Ensure tests pass
2. Compare interaction model to the native time input: `/?path=/story/packages-datepicker-datepicker--date-and-time-comparison&globals=theme:thunderblocks`
3. Ensure text date formats still validate as expected


packages/wonder-blocks-date-picker/src/components/__tests__/date-picker.test.tsx
```diff
@@ -598,6 +598,84 @@ describe("DatePicker", () => {
         ).toBeInTheDocument();
     });
 
+    it("increments the day segment when ArrowUp is pressed in the input", async () => {
```
jandrade: **suggestion:** Could you please add AAA comments on these new tests? thanks :)

packages/wonder-blocks-date-picker/src/hooks/__tests__/use-date-segment-arrow-keys.test.ts
```diff
@@ -0,0 +1,153 @@
...
+            }),
+        {initialProps: {value}},
+    );
+
+    return {input, handleChange, result, rerender};
+}
+
+describe("useDateSegmentArrowKeys", () => {
+    it("returns false for keys other than ArrowUp/ArrowDown", () => {
+        const {input, result} = setup("01/16/2026");
```
jandrade: same suggestion about AAA here

packages/wonder-blocks-date-picker/src/util/adjust-date-segment.ts
```diff
@@ -0,0 +1,33 @@
...
+    if (type === "day") {
+        const {daysInMonth} = date;
+        const newDay =
+            ((((date.day - 1 + delta) % daysInMonth) + daysInMonth) %
+                daysInMonth) +
+            1;
+        return date.with({day: newDay});
+    }
+    if (type === "month") {
+        const newMonth = ((((date.month - 1 + delta) % 12) + 12) % 12) + 1;
```
jandrade, line 31: **suggestion:** It would be nice adding a comment explaining what this does (I'm only assuming it by the const name)

marcysutton:
@jandrade the RTL arrow key handling was a bit involved -- i actually ended up uncovering a bug where default dates weren't rendering for RTL locales in the base input! 

Before: RTL example for `wonder-blocks-date-picker.stories.tsx` in `frontend`: 
with Arabic locale - RTL story

After: rendering a date in the input in WB (in a new story with the `ar` locale) - `/?path=/story/packages-datepicker-datepicker--with-rtl-numeric-locale&globals=theme:thunderblocks;direction:rtl`
Text input with

packages/wonder-blocks-core/src/index.ts
```diff
@@ -4,6 +4,7 @@ export {default as InitialFallback} from "./components/initial-fallback";
 export {default as addStyle} from "./util/add-style";
 export {default as Server} from "./util/server";
 export {Id} from "./components/id";
+export {useDirectionDetection} from "./hooks/use-direction-detection";
```
marcysutton, line 7: This utility lived in `wonder-blocks-modal` which made it a cross-import, unless it got moved! Maybe this could satisfy https://khanacademy.atlassian.net/browse/WB-2060?

jandrade: Good call! yeah, that should satisfy the ticket you shared.

jandrade (approved):
This is an amazing addition! thanks for fixing the RTL case 👏 🚀

packages/wonder-blocks-date-picker/src/util/adjust-date-segment.ts
marcysutton: Good call!

PR #3193, opened 2026-09-02
🚨 [CSS Modules] Migrate IconButton to CSS Modules

jandrade:
Migrate `IconButton` and the shared `IconButtonUnstyled` reset from Aphrodite to
CSS Modules. No public API change — same props, DOM structure and `style` overrides.

**PR 1 of 4 in the WB-2329 stack** (`IconButton` → `ConversationIconButton` →
`ActivityIconButton` → `NodeIconButton`). It carries the shared reset and the
package.json CSS-delivery wiring the other three build on, so it lands first.

- Follows the component-token surface `Button` settled on: variant classes only
  assign `--wb-c-icon-button--*`, base/state rules only read them. The
  `kind × actionType` colour matrix is nine compound rules; `_generateStyles` is gone.
- State rules qualify the kind with `:where(…)` so specificity is flat and source
  order decides: hover → active → focus → disabled.
- Icon size now comes from `--wb-c-icon-button--icon-size` instead of an inline
  theme read.
- The reset moves to `icon-button-unstyled.module.css`, emitted into the nested
  `shared.reset` layer so component styles outrank it regardless of bundle order.

Two parity notes: hover is intentionally **not** gated behind `@media (hover: hover)`
(IconButton's Aphrodite wasn't either — the divergence from `Button` is pre-existing);
and a disabled button that is both focused and hovered now keeps its focus ring,
matching `Button` (68ea60e34) and `ActivityButton` (bfd43dea4).

Issue: WB-2329

## Test plan:

- `pnpm jest` (repo-wide), `pnpm typecheck`, eslint, stylelint, prettier — clean
- Rollup emits the rules into `dist/index.css` with `@apply --wb-focus-visible`
  expanded and the reset nested in `@layer shared { @layer reset }`
- Manual, in default/thunderblocks/syl-dark: spot-check `kind` × `actionType` ×
  `size` and hover/active/focus-visible/disabled; check the focus ring while also
  hovering and while pressed; confirm **rest-state borders** for every `kind` (the
  nested-layer ordering isn't verifiable in CI) and correct icon size per `size`
  (`PhosphorIcon` ships competing rules in the same layer)

## Review plan:

1. 🚨 [`icon-button.module.css`](https://github.com/Khan/wonder-blocks/pull/3193#discussion_r3915197581)
2. 🚨 [`package.json`](https://github.com/Khan/wonder-blocks/pull/3193#discussion_r3915202506)
3. ⚠️ [`icon-button-unstyled.module.css`](https://github.com/Khan/wonder-blocks/pull/3193#discussion_r3915197905)
4. ⚠️ [`icon-button.tsx`](https://github.com/Khan/wonder-blocks/pull/3193#discussion_r3915198061)
5. ⚠️ [`icon-button-unstyled.tsx`](https://github.com/Khan/wonder-blocks/pull/3193#discussion_r3915202720)


packages/wonder-blocks-icon-button/src/components/icon-button.module.css
```diff
@@ -0,0 +1,424 @@
+/**
```
jandrade, line 1: 🚨 **High risk.** This replaces the entire `kind × actionType × size` variant matrix — 36 cells that Aphrodite previously generated and cached per-variant — with hand-written CSS. Nothing in CI checks that the cascade resolves the way it did before, so this file is where a visual regression would hide.

Three things worth your attention:

- **Specificity flattening.** Every state rule qualifies the kind with `:where(…)`, which contributes no specificity. That's deliberate: it puts all state rules at the same specificity so source order alone decides precedence, declared hover → active → focus → disabled. If any of those `:where(…)` wrappers were dropped, the focus ring would lose to the primary `outline` when a button is both focused and hovered.
- **Hover is _not_ gated behind `@media (hover: hover)`**, unlike `Button`'s. IconButton's Aphrodite only reset the background on non-hover devices, and I preserved that exactly. The inconsistency with `Button` is pre-existing — I'm flagging it rather than silently widening the reset, but it's worth reconciling separately.
- **`.button .icon` is compound on purpose.** `PhosphorIcon` ships single-class `.small` / `.medium` rules in the same `@layer shared`. Aphrodite's `!important` used to win that contest for us; the compound selector makes it a specificity question instead of a bundle-order one.

packages/wonder-blocks-icon-button/src/components/icon-button-unstyled.module.css
```diff
@@ -0,0 +1,34 @@
+/**
```
jandrade, line 1: ⚠️ **Medium risk, and the one thing I could not verify outside a browser.** The reset moves into the nested layer `shared.reset` — the build wraps the file in `@layer shared`, and the file's own `@layer reset` becomes the sub-layer. Component styles sit directly in `shared`, and rules directly in a layer outrank that layer's named sub-layers (CSS Cascade 5). If that's wrong, secondary/tertiary rest-state borders would disappear, which is why step 5 of the test plan calls it out specifically.

Note the property set intentionally differs from `button-unstyled.module.css`: this one has `background: none` and lacks `user-select` / the `-webkit-tap-highlight-color` rule. Those differences are pre-existing in the Aphrodite version and preserved rather than harmonised.

packages/wonder-blocks-icon-button/src/components/icon-button.tsx
```diff
@@ -1,22 +1,15 @@
 import * as React from "react";
```
jandrade, line 1: ⚠️ **Medium risk.** `_generateStyles` and its per-variant `StyleSheet.create` cache are gone; variant selection is now four class-name strings composed through the `style` prop (`processStyleList` routes them to `className`).

Two behavioural notes:

- `IconChooser` no longer reads the theme object at all — icon sizing moved to the `--wb-c-icon-button--icon-size` custom property assigned by the size class. The `PhosphorIcon size="small"` vs `size="medium"` switch is unchanged.
- `pressed` is now guarded with `!disabled`. Previously `pressed` came after `disabled` in the Aphrodite array and would have won; in CSS `[aria-disabled="true"]` outranks `.pressed` on specificity. The state is unreachable in practice (`IconButtonUnstyled` only fires `onPress` when not disabled), so this just makes the JS agree with the CSS.

packages/wonder-blocks-icon-button/package.json
```diff
@@ -19,12 +19,16 @@
   "main": "dist/index.js",
   "module": "dist/es/index.js",
   "types": "dist/index.d.ts",
+  "sideEffects": [
```
jandrade, line 22: 🚨 **High risk — consumer-facing.** Adds `sideEffects: ["**/*.css"]` and a `./css` export subpath, so the package now ships `dist/index.css` and imports it as a side-effect of the JS entry. This mirrors what `wonder-blocks-button` did in #3164.

Standard webpack / Vite / Next.js setups pick this up with no changes, but **SSR consumers that can't process CSS imports may need a CSS loader or a mock**. That risk is the same one #3164 took, so the blast radius is partly established — but this is the second package to do it, and the first one that search-field, dropdown and modal all depend on.

packages/wonder-blocks-icon-button/src/components/icon-button-unstyled.tsx
```diff
@@ -6,8 +6,8 @@ import {Link, useInRouterContext} from "react-router-dom-v5-compat";
 
 import {addStyle, keys} from "@khanacademy/wonder-blocks-core";
 import {isClientSideUrl} from "@khanacademy/wonder-blocks-clickable";
-import {StyleSheet} from "aphrodite";
 import type {IconButtonProps, IconButtonRef} from "../util/icon-button.types";
+import styles from "./icon-button-unstyled.module.css";
```
jandrade, line 10: ⚠️ **Medium risk — this file is shared by all four components in the package.** The only change is swapping the Aphrodite `styles.shared` for the CSS Module's `styles.reset`.

Worth noting for the stack: through PRs 2–4 the other three components are still Aphrodite (unlayered) while this reset is layered, so they keep overriding it exactly as before. No intermediate state in the stack is broken.

packages/wonder-blocks-icon-button/src/components/icon-button.module.css
```diff
@@ -0,0 +1,423 @@
...
+ * IconButton styles
+ *
+ * Component-token surface
```
marcysutton: Will these comments get trimmed/stripped out in the distributed CSS? I'm wondering if this as a convention will inflate our file sizes.

marcysutton:
@claude review once

packages/wonder-blocks-icon-button/src/components/icon-button.module.css
```diff
@@ -0,0 +1,423 @@
...
+ *   --wb-c-icon-button--border-default
+ *   --wb-c-icon-button--bg-hover
+ *   --wb-c-icon-button--fg-hover
+ *   --wb-c-icon-button--border-hover
+ *   --wb-c-icon-button--bg-press
+ *   --wb-c-icon-button--fg-press
+ *   --wb-c-icon-button--border-press
+ *   --wb-c-icon-button--bg-disabled        disabled colours          (kind)
+ *   --wb-c-icon-button--fg-disabled
+ *   --wb-c-icon-button--border-disabled
```
beaesguerra, line 26: This had me thinking if we could centralize how the component css variables are defined so we don't need to maintain a list of the variables in the comment. Started a thread around this to discuss this more with the wb team! (not blocking)

packages/wonder-blocks-icon-button/src/components/icon-button.tsx
```diff
@@ -29,18 +22,16 @@ function IconChooser({
...
-    // We set the icon size based on the theme object. This is necessary
-    // because the icon size could change based on the theme.
-    const iconStyle = {
-        width: theme.icon.sizing[size],
-        height: theme.icon.sizing[size],
-    };
+    // The icon's box is sized by the `--wb-c-icon-button--icon-size` component
+    // token, which the `size` class on the root element assigns. This keeps the
+    // icon size theme-driven without the component having to read the theme
+    // object in JS.
```
beaesguerra, line 28: nit: Can this comment be removed since there's no longer code here related to the icon sizing?

beaesguerra (approved):
Looks good to me! Left some minor, non-blocking comments!

PR #3194, opened 2026-09-02
🚨 [CSS Modules] Migrate ConversationIconButton to CSS Modules

jandrade:
Migrate `ConversationIconButton` from Aphrodite to CSS Modules. No public API
change — same props, DOM structure and `style` overrides.

**PR 2 of 4 in the WB-2329 stack**, based on `WB-2329.1` (`IconButton`) — review
that one first, it carries the shared `IconButtonUnstyled` reset and the
package.json CSS-delivery wiring.

- Same component-token surface as the rest of the wave: variant classes only
  assign `--wb-c-conversation-icon-button--*`, base/state rules only read them.
  The `kind × actionType` colour matrix is six compound rules.
- Border widths and radius don't vary by `kind` here (the old `theme` object
  spelled out all three kinds identically), so they're assigned once on `.button`.
  Its unused `root.sizing` entry is dropped.
- State rules are all (0,2,0), so source order decides: hover → active → focus →
  disabled. The press rule keeps the Aphrodite `calc(padding - 1px)` compensation
  that holds the outer size stable against the thicker press border.
- One intentional change, matching the rest of the family: a disabled button that
  is both focused and hovered now keeps its full focus ring (previously the
  disabled `:hover` reset stripped the outline — an Aphrodite rule-ordering artifact).

Issue: WB-2329

## Test plan:

- `pnpm jest packages/wonder-blocks-icon-button`, `pnpm typecheck`, eslint,
  stylelint, prettier — clean
- Rollup emits the 20 `wb-conversation-icon-button-*` selectors into
  `dist/index.css` with `@apply --wb-focus-visible` expanded, inside `@layer shared`
- Manual, in default/thunderblocks/syl-dark: spot-check `kind` × `actionType` and
  hover/active/focus-visible/disabled; **confirm the outer size doesn't shift when
  pressed** (2px press vs 1px rest border — a wrong sign is a subtle 2px jitter);
  check the disabled focus ring; confirm a custom (non-Phosphor) `icon` still
  renders at 1.6rem, since `.button .icon` now has to outrank `PhosphorIcon`'s
  own rules in the same layer

## Review plan:

1. 🚨 [`conversation-icon-button.module.css`](https://github.com/Khan/wonder-blocks/pull/3194#discussion_r3915198349)
2. ⚠️ [`conversation-icon-button.tsx`](https://github.com/Khan/wonder-blocks/pull/3194#discussion_r3915198518)


packages/wonder-blocks-icon-button/src/components/conversation-icon-button.module.css
```diff
@@ -0,0 +1,286 @@
+/**
```
jandrade, line 1: 🚨 **High risk.** Replaces the `kind × actionType` variant matrix and its per-variant `StyleSheet.create` cache with hand-written CSS. Not covered by CI.

Three things worth your attention:

- **Border widths and radius are assigned once on `.button`, not per kind.** The hand-written Aphrodite `theme` object spelled out all three kinds with identical values, so per-kind classes would have been three copies of the same thing. If a future design does make these kind-dependent, they'd move onto the kind classes.
- **The press rule keeps the `calc(padding - 1px)` compensation.** The press border is 2px against a 1px rest border; without the padding adjustment the circle grows by 2px on every press. This is the subtlest thing in the file and a wrong sign here reads as jitter rather than breakage.
- **`root.sizing` is dropped, not ported.** The Aphrodite `theme` object declared `root.sizing.default = size_360`, but no rule ever read it. Please sanity-check that — if the button is _supposed_ to have a fixed 3.6rem target area, then the bug is that it never did, and that's a separate fix.

packages/wonder-blocks-icon-button/src/components/conversation-icon-button.tsx
```diff
@@ -1,18 +1,15 @@
 import * as React from "react";
```
jandrade, line 1: ⚠️ **Medium risk.** Same shape as the `IconButton` change in PR 1: `_generateStyles` out, class-name strings through the `style` prop in.

The custom-icon branch now gets `styles.icon` (a `.button .icon` rule) instead of the Aphrodite `staticStyles.icon`. The compound selector is deliberate — it has to outrank `PhosphorIcon`'s single-class rules in the same `@layer shared`, which Aphrodite's `!important` used to handle.

jandrade:
@claude review once

packages/wonder-blocks-icon-button/src/components/conversation-icon-button.tsx
```diff
@@ -65,12 +62,20 @@ export const ConversationIconButton: React.ForwardRefExoticComponent<
...
-    const buttonStyles = _generateStyles(actionType, !!disabled, kind);
-
-    const styles = [
-        buttonStyles.default,
-        disabled && buttonStyles.disabled,
-        pressed && buttonStyles.pressed,
+    // One class per variant axis. Each of these only assigns the
+    // `--wb-c-conversation-icon-button--*` component tokens that its axis owns;
+    // `styles.button` and the state rules in the module read those tokens.
+    // `disabled` is selected in CSS via the `[aria-disabled="true"]` attribute
```
beaesguerra, line 68: One thing to consider now that we apply disabled styles using `[aria-disabled="true"]` is that the disabled styles will have higher specificity now compared to before. I think this should be okay since this would mean the disabled styles would override any custom styles with the default specificity. Worth mentioning though!

packages/wonder-blocks-icon-button/src/components/conversation-icon-button.tsx
```diff
@@ -85,156 +90,16 @@ export const ConversationIconButton: React.ForwardRefExoticComponent<
...
-        },
-        disabled: {
-            cursor: "not-allowed",
-            ...disabledStatesStyles,
-            // NOTE: Even that browsers recommend to specify pseudo-classes in
-            // this order: link, visited, focus, hover, active, we need to
-            // specify focus after hover to override hover styles. By doing this
-            // we are able to reset the border/outline styles to the default
-            // ones (rest state).
-            // For order reference: https://css-tricks.com/snippets/css/link-pseudo-classes-in-order/
```
beaesguerra, line 223: I assume it's okay that we lose this context, since the previous ordering (hover, active, focus-visible) didn't match this comment! 

I see we persist this order with the same hover, active, focus-visible order in the css module file 👍

packages/wonder-blocks-icon-button/src/components/conversation-icon-button.tsx
```diff
@@ -85,156 +90,16 @@ export const ConversationIconButton: React.ForwardRefExoticComponent<
...
-                border: `${borderWidthKind.hover} solid ${themeVariant.hover.border}`,
-            },
-            // Allow hover styles on non-touch devices only. This prevents an
-            // issue with hover being sticky on touch devices (e.g. mobile).
-            ["@media not (hover: hover)"]: {
-                ":hover": {
-                    // reset hover styles on non-touch devices
-                    backgroundColor: "transparent",
-                },
-            },
```
beaesguerra: This doesn't seem to be in the new css file, is that expected?

packages/wonder-blocks-icon-button/src/components/conversation-icon-button.module.css
```diff
@@ -0,0 +1,286 @@
...
+
+    /* Layout */
+    border-radius: var(--wb-border-radius-radius_full);
+    padding: var(--wb-c-conversation-icon-button--padding);
+    place-self: center;
+
+    /* Theming, driven by the colour tokens set by the variant classes. */
+    border-style: solid;
+    border-width: var(--wb-c-conversation-icon-button--border-width-rest);
+    border-color: var(--wb-c-conversation-icon-button--border-rest);
```
beaesguerra, line 84: What do you think about including "border-color" in the variable name for clarity?


```suggestion
    border-color: var(--wb-c-conversation-icon-button--border-color-rest);
```

packages/wonder-blocks-icon-button/src/components/conversation-icon-button.module.css
```diff
@@ -0,0 +1,286 @@
...
+    background: var(--wb-c-conversation-icon-button--bg-rest);
```
beaesguerra, line 85: question: In component token variables, do we want to follow the naming convention of `bg` and `fg` instead of `background` and `foreground`? 

I'm wondering how we decide which term, since we spell it out for semantic color token names and in the [component token structure format](https://khanacademy.atlassian.net/wiki/x/FwB47)!

packages/wonder-blocks-icon-button/src/components/conversation-icon-button.module.css
```diff
@@ -0,0 +1,286 @@
...
+    --wb-c-conversation-icon-button--border-press: var(--wb-semanticColor-action-tertiary-neutral-press-border);
+}
+
+/**
+ * Hover. The hover border width matches the rest width, so the padding is
+ * unchanged and the box doesn't shift.
+ */
+.button:hover {
+    background: var(--wb-c-conversation-icon-button--bg-hover);
+    border-width: var(--wb-c-conversation-icon-button--border-width-rest);
```
beaesguerra, line 192: This uses the border width rest token, but I think that's because there's no border width hover token! Is this expected? Thought I'd check!

packages/wonder-blocks-icon-button/src/components/conversation-icon-button.module.css
```diff
@@ -0,0 +1,286 @@
...
+ * `PhosphorIcon` at `size="small"`, which sizes itself.
+ *
+ * Qualified with `.button` on purpose: `PhosphorIcon` ships single-class rules
+ * in the same `@layer shared`, so at equal specificity the winner would come
+ * down to which package's stylesheet happens to load first. The compound
+ * selector makes the outcome depend on specificity instead.
+ */
+.button .icon {
+    inline-size: var(--wb-sizing-size_160);
+    block-size: var(--wb-sizing-size_160);
```
beaesguerra, line 285: In the [IconButton PR](https://github.com/Khan/wonder-blocks/pull/3193/changes#diff-4f2b85caf355f0e6833be5815f04db067a308f7a3a26b8dc227ae4d311cdf588R70-R71), we have a css variable for the button sizing. For my own understanding, do we only create css variables for things if the value is expected to change? I'm wondering if there are any benefits to creating component tokens for cases where they don't. Only thing I can think of right now is for design parity if there is an equivalent component token for it in Figma! Though we also want to be mindful of the amount of css variables we add! 

Let me know what you think, as I'm still wrapping my head around these new changes! 😄

beaesguerra (commented):
Overall looks great! Had some questions for you I wanted to get your thoughts on, mostly around naming conventions!

PR #3195, opened 2026-09-02
🚨 [CSS Modules] Migrate ActivityIconButton to CSS Modules

jandrade:
Migrate `ActivityIconButton` from Aphrodite to CSS Modules. No public API change
— same props, DOM structure and `styles.root` / `styles.box` / `styles.label`
overrides.

**PR 3 of 4 in the WB-2329 stack**, based on `WB-2329.2`
(`ConversationIconButton`). PR 1 carries the shared reset and CSS-delivery wiring.

- Mirrors the already-landed `activity-button.module.css` (same
  `semanticColor.chonky` token family): variant classes only assign
  `--wb-c-activity-icon-button--*`, base/state rules only read them, with six
  compound `actionType × kind` colour rules. Structural values (border, the
  6px/8px/0 shadow depths, radius, padding) are written literally, as every theme
  reused them.
- The box is no longer styled by JS-toggled classes (`chonkyPressed` /
  `chonkyDisabled` are gone) but by descendant selectors from the root, all at the
  same specificity: hover → active → disabled → focus. The plain `chonky`
  className stays as a consumer/test hook.
- **One divergence worth a decision:** the `.box` / `.label` base rules are
  qualified with `.button`, because `View` (`.default`) and `BodyText`
  (`.smallSemi`) ship single-class rules in the same `@layer shared` that set the
  same properties — Aphrodite's `!important` used to settle that. This makes the
  outcome depend on specificity rather than bundler emit order.
  `activity-button.module.css` left those as single classes and is still
  load-order-dependent; if you agree with the reasoning, it should get the same
  treatment in a follow-up.
- One intentional change, matching the rest of the family: a disabled button that
  is both focused and hovered now keeps its full focus ring.

Issue: WB-2329

## Test plan:

- `pnpm jest packages/wonder-blocks-icon-button`, `pnpm typecheck`, eslint,
  stylelint, prettier — clean
- Rollup emits the 26 `wb-activity-icon-button-*` selectors into `dist/index.css`
  with `@apply --wb-focus-visible` expanded and the `-webkit-line-clamp` label
  clipping intact, inside `@layer shared`
- Manual, in default/thunderblocks/syl-dark: spot-check `kind` × `actionType` and
  all states; confirm the box lifts on hover and drops flat on press and that the
  custom-styles story's root/box/label overrides still apply; **confirm the box's
  padding, centring and border are unchanged and the label still clamps to 2 lines
  with hyphenation** (a long translated label is the useful case — see WB-2354);
  check the disabled focus ring

## Review plan:

1. 🚨 [`activity-icon-button.module.css`](https://github.com/Khan/wonder-blocks/pull/3195#discussion_r3915198772)
2. ⚠️ [`activity-icon-button.tsx`](https://github.com/Khan/wonder-blocks/pull/3195#discussion_r3915198985)


packages/wonder-blocks-icon-button/src/components/activity-icon-button.module.css
```diff
@@ -0,0 +1,382 @@
+/**
```
jandrade, line 1: 🚨 **High risk**, and the file with the one deliberate divergence in this stack.

- **The `.box` and `.label` _base_ rules are qualified with `.button`.** The already-landed `activity-button.module.css` left the equivalent rules as single classes. `View` (`.default`) and `BodyText` (`.smallSemi`) ship single-class rules in the same `@layer shared` that set some of the same properties (`flex-direction`, `padding`, `border-width`, `align-items`, `margin`), so at equal specificity the winner is whichever package's stylesheet the bundler emits first. Qualifying makes it a specificity question instead. **This is worth a decision, not just a review** — if you agree, `activity-button.module.css` should get the same treatment in a follow-up.
- **Box state styling moved out of JS.** `chonkyPressed` / `chonkyDisabled` are gone; the box is styled by descendant selectors from the root, all at (0,3,0), declared hover → active → disabled → focus so source order alone decides precedence. The plain `chonky` className stays in the DOM as a consumer/test hook.
- **Structural values are written literally** (thin border, the 6px/8px/0 shadow depths, radius, padding). The Aphrodite `theme` object was hand-written in-file precisely because every theme reused them — same call `ActivityButton` made.

packages/wonder-blocks-icon-button/src/components/activity-icon-button.tsx
```diff
@@ -1,21 +1,17 @@
 import * as React from "react";
```
jandrade, line 1: ⚠️ **Medium risk.** `_generateStyles` out, class-name strings in, and `chonkyStyles` collapses from four entries to two — the box's hover / press / disabled styling is now driven entirely by the root's state selectors.

The local `theme` object (shadow depths, padding, radius, label colours) is deleted; its themed value — `themeTokens.activityIconButton.label.color.progressive` — is referenced in CSS as `--wb-c-icon-button-activityIconButton-label-color-progressive`, so the syl-dark override still applies. Worth confirming that in the theme switcher.

jandrade:
@claude review once

packages/wonder-blocks-icon-button/src/components/activity-icon-button.tsx
```diff
@@ -127,12 +128,15 @@ export const ActivityIconButton: React.ForwardRefExoticComponent<
...
+            style={buttonStyles}
             type={type}
             {...(!hasVisibleLabel ? {"aria-label": ariaLabel} : {})}
         >
             <>
-                {/* NOTE: Using a regular className to be able to use descendant selectors to account for the hover and press states */}
+                {/* NOTE: The plain `chonky` className is kept as a
+                consumer/test hook. It no longer drives styling — the box is
+                styled by descendant selectors from the root in the CSS
+                module. */}
```
beaesguerra, line 139: Could we remove the `className` since it is no longer used to apply styling?

beaesguerra (approved):
Looks good to me!

